PR #2055, opened 2026-05-25
feat: OSS optimistic locking — default-ON platform-wide (conflict bar, auto-derive CrudForms, ACL/users/roles, sales/CRM/catalog/staff/resources/planner) (#1981)

pkarw:
> **Scope note (greatly expanded since the original opt-in spec below).** This PR
> now lands the **default-ON, platform-wide** optimistic-locking system. It
> absorbed #2276 (universal CrudForm auto-derive) and #2346 (remaining custom
> handlers). Summary of what's covered today:
>
> - **Default ON** for every `makeCrudRoute` entity (opt out via `OM_OPTIMISTIC_LOCK=off`); strictly additive — clients that don't send the version header are unaffected.
> - **Unified conflict bar** (`RecordConflictBanner` in `AppShell`) replaces the raw `record_modified` toast; clears on navigation.
> - **CrudForm auto-derive**: every edit-mode CrudForm attaches the header from `initialValues.updatedAt` (update **and** delete) — opt out with `disableOptimisticLock`.
> - **Auth ACL editing**: user + role ACL grant saves (`auth/{roles,users}/acl`) are version-checked and wrapped in a transaction; `User`/`Role` gained an `updated_at` column (+migration) so their scalar edit/delete lock too.
> - **Command-level locks**: `enforceCommandOptimisticLock` / `enforceRecordGoneIsConflict` for non-CRUD command writes (sales documents, stale-edit-after-delete → 409 not 404).
> - **Submit-scope leak fixes**: forms whose `onSubmit` mutates *other* entities (sales `ChannelOfferForm` prices, catalog product offers/unit-conversions, catalog variant prices) now send each child its **own** version, overriding the parent header (prevents false 409s).
> - **Coverage**: customers (company/person/deal incl. list+kanban delete, status, drag-stage; activities/tasks edit/delete/complete/cancel), catalog (products/variants/categories/offers/option-schemas/prices/unit-conversions), sales (orders/quotes + line/adjustment/return/payment/shipment, channels, offers), staff (teams/team-roles/leave-requests), resources, planner availability, auth users/roles + ACLs.
> - **Excluded** (documented): bulk operations (no single version token).
>
> Manual QA guide for @alinadivante is in the comments. Original spec follows.

---

Tracking plan: .ai/runs/2026-05-25-oss-optimistic-locking/PLAN.md
[...]

PR #5262, opened 2026-08-13
fix(customer_accounts): write encrypted display_name through the managed entity (#3837)

### LINKED ISSUE #3837: security(customer_accounts): encrypted display_name written in plaintext via nativeUpdate (customerUserService.ts:101)
The encrypted-at-rest PII field `display_name` is persisted with `em.nativeUpdate`, which issues raw SQL and fires none of the MikroORM flush lifecycle hooks the tenant-data-encryption subscriber depends on — so the value is written to the database as **plaintext** into a column contractually expected to hold ciphertext. This is **LIVE**: any customer editing their profile display name, or any staff member editing a customer's display name, triggers it today. Because the decrypt path silently passes non-`v1:iv:tag:ct` values through unchanged, the corruption is invisible on read, so the PII simply sits unencrypted at rest, defeating the encryption control for that field (GDPR / at-rest exposure).

## Location(s)
- `packages/core/src/modules/customer_accounts/services/customerUserService.ts:101` — `updateProfile` persists `display_name` via `em.nativeUpdate`
- `packages/core/src/modules/customer_accounts/api/admin/users/[id].ts:176-182` — admin user PUT persists `display_name` via `em.nativeUpdate`
- `packages/core/src/modules/customer_accounts/encryption.ts` — declares `customer_accounts:customer_user` → `display_name` as encrypted
- `packages/shared/.../tenantDataEncryptionService.ts:401-402` — `decryptFields` passes non-ciphertext values through unchanged (masks the corruption on read)

## Why it matters
`display_name` is declared encrypted-at-rest. `em.nativeUpdate` bypasses the `beforeUpdate`/`beforeCreate` flush hooks the encryption subscriber relies on, so the raw plaintext is written directly to the ciphertext column. Both the portal profile update (`updateProfile`) and the admin user PUT do this. On read, `decryptFields` silently passes the non-`v1:iv:tag:ct` value through unchanged, so the field appears correct while sitting unencrypted at rest — defeating the tenant-data-encryption control for that PII field. Contrast the staff auth module, which writes the analogous encrypted `name` via managed-entity assignment + flush (`auth/commands/users.ts:588`) so the subscriber encrypts it. (email is never written via `nativeUpdate`, so it is unaffected; only `display_name` qualifies.)

## Proposed fix
- Write `display_name` through a managed entity: load the entity, assign `entity.displayName = value`, and call `em.flush()` so the encryption subscriber encrypts it on flush — or route it through the encryption service explicitly. Never write an encrypted column via `nativeUpdate`.
- Apply the same fix at both sites (`updateProfile` and `admin/users/[id].ts` PUT).

## BC / Safety
This is the well-known `nativeUpdate`-bypasses-encryption-subscriber footgun the codebase already documents. New writes will store ciphertext; existing plaintext rows written before the fix remain plaintext and should be re-encrypted via a backfill/migration (they still read correctly due to the pass-through decrypt). Audited: only `display_name` qualifies in this module (`email` is never `nativeUpdate`d; `CustomerRole.name` and other `nativeUpdate`d columns are not in the encryption map). risk-high because it touches encryption/PII-at-rest handling.

pkarw:
Closes #3837

## 🎯 Goal
Stop the customer portal's profile update and the admin customer-user PUT from persisting the encrypted-at-rest `display_name` as plaintext, so the tenant-data-encryption control actually holds for that PII field.

## 🔍 Problem
`display_name` is declared encrypted at rest in `customer_accounts/encryption.ts`, but both write paths persisted it with `em.nativeUpdate`. Any customer editing their own profile display name, or any staff member editing a customer's display name, wrote plaintext PII into a column contractually holding ciphertext. Because `decryptFields` passes non-`v1:iv:tag:ct` values through unchanged, the value still read back correctly — so the corruption was invisible at runtime while the data sat unencrypted in the database (GDPR / at-rest exposure).

## 🔍 Root Cause
`TenantEncryptionSubscriber` encrypts through MikroORM's `beforeCreate` / `beforeUpdate` flush hooks (`packages/shared/src/lib/encryption/subscriber.ts:409-413`). `em.nativeUpdate` issues raw SQL through the query builder and fires none of those hooks, so the value reached the column exactly as supplied. The staff auth module writes its analogous encrypted `name` via managed-entity assignment plus flush, which is why it was never affected.

## What Changed
- `packages/core/src/modules/customer_accounts/services/customerUserService.ts` — `updateProfile` no longer calls `em.nativeUpdate`. It loads the row through `findOneWithDecryption` (tenant + organization scoped, matching the module's data-access convention), assigns `displayName` on the managed entity, and calls `em.flush()` so `beforeUpdate` encrypts the value on its way to the database. The caller's entity is still updated in place, so response shapes are unchanged.
- `packages/core/src/modules/customer_accounts/api/admin/users/[id].ts` — the admin PUT drops `displayName` from its `nativeUpdate` payload and routes it through `customerUserService.updateProfile` instead. It runs **before** the `nativeUpdate`, deliberately: that `nativeUpdate` sets the explicit `updated_at` this route returns as the optimistic-lock version (#2055), so ordering it last keeps the persisted version and the version reported to the client identical. Every other field on that route is unencrypted and keeps its existing `nativeUpdate` write.

## 🧪 Tests
- `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn build:app` — all green (local runner; no compose `app` container running).
- `yarn test` — `@open-mercato/core` 9627 passed / 1238 suites; shared 1876, app 549, search 311, checkout 140, channel-imap 109, telemetry 82, cache 73, sync-akeneo 62 all passed. **Four `create-mercato-app` template dev-wrapper tests fail on this machine only**, from a host limit rather than this branch: the wrapper's preflight reads `/proc/sys/fs/inotify/*` and aborts because `max_user_watches` is 253174 (needs 4194304) and `max_user_instances` is 128 (needs 4096). Nothing in this diff touches `packages/create-app`; CI runners with the documented sysctl values are unaffected.
- `services/__tests__/customerUserService.updateProfile.test.ts` (new) — locks in that `updateProfile` writes via the managed entity and `em.flush()`, never `nativeUpdate`; that the lookup stays tenant/organization scoped; that the caller's entity mirrors the saved value; and the no-op and missing-row paths.
- `api/admin/users/__tests__/optimistic-lock.test.ts` — adds a route-level regression case asserting the admin PUT keeps `displayName` out of its `nativeUpdate` payload and delegates to `updateProfile`, plus a case proving the service is not called when no display name is submitted. The existing `customerUserService` mock gained `updateProfile`.
- `__tests__/encryptedFieldWrites.test.ts` (new) — a module-wide guard that derives the encrypted property names from `defaultEncryptionMaps`, walks every `nativeUpdate` call in the module (resolving payloads passed as variables back to their declaration and `payload.field = …` assignments), and fails if any encrypted field is written that way. It flagged both original sites before the fix, so it would catch a future regression on `email`, session `ip_address` / `user_agent`, or invitation `display_name` too.

All three regression tests were verified failing against the unfixed sources and passing with the fix.

## 💥 Breaking Changes
None. `CustomerUserService.updateProfile` keeps its signature, both HTTP responses keep their shape, and no schema changed.

## ⚠️ Follow-up for operators
New writes store ciphertext, but rows written before this fix stay plaintext (they still read correctly through the pass-through decrypt). They can be re-encrypted in place with the existing `mercato entities rotate-encryption-key` CLI in its non-`--rotate` mode, which encrypts plaintext values under the current DEK and skips rows that are already ciphertext — no new migration is needed. Worth running per tenant after this merges.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5262"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789218290&installation_model_id=432243&pr_number=5262&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5262&signature=c7fc4523f8d97cfc3135783dfce367a7ee1d7da4c9bb6188cadb64b0cba285ac"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/customer_accounts/__tests__/encryptedFieldWrites.test.ts
+++ b/packages/core/src/modules/customer_accounts/__tests__/encryptedFieldWrites.test.ts
@@ -86,7 +86,10 @@ function splitTopLevelArguments(args: string): string[] {
 
 // The update payload is often built in a local variable rather than passed inline, so
 // resolve identifiers back to the object literal they were declared with plus every
-// `payload.field = …` assignment made on them before the write.
+// `payload.field = …` assignment made on them before the write. Resolution is file-scoped
+// and therefore deliberately over-inclusive: when two functions in one file both build a
+// `updates` payload, assignments from both are considered. That direction produces a false
+// alarm at worst, never a missed plaintext write.
 function resolvePayloadText(payload: string, source: string): string {
   if (!/^[A-Za-z_$][\w$]*$/.test(payload)) return payload
   const declaration = new RegExp(`\\b(?:const|let|var)\\s+${payload}\\b[^=]*=\\s*`).exec(source)
--- a/packages/core/src/modules/customer_accounts/__tests__/encryptedFieldWrites.test.ts
+++ b/packages/core/src/modules/customer_accounts/__tests__/encryptedFieldWrites.test.ts
@@ -1,0 +1,161 @@
+/** @jest-environment node */
+import * as fs from 'node:fs'
+import * as path from 'node:path'
+import { defaultEncryptionMaps } from '../encryption'
+
+// `em.nativeUpdate` issues raw SQL and fires none of the MikroORM flush hooks the
+// tenant-data-encryption subscriber depends on, so a field declared in this module's
+// encryption map that is written that way lands in the database as plaintext inside a
+// column contractually holding ciphertext (#3837). The decrypt path passes non-ciphertext
+// values through unchanged, so nothing surfaces the corruption at runtime — this guard is
+// what surfaces it instead. Encrypted fields must be written through a managed entity and
+// `em.flush()` so the subscriber can encrypt them.
+
+const MODULE_DIR = path.resolve(__dirname, '..')
+const SKIPPED_DIRECTORIES = new Set(['__tests__', '__integration__', 'node_modules'])
+
+function toPascalCase(value: string): string {
+  return value
+    .split('_')
+    .filter(Boolean)
+    .map((part) => part.charAt(0).toUpperCase() + part.slice(1))
+    .join('')
+}
+
+function toCamelCase(value: string): string {
+  return value.replace(/_([a-z])/g, (_match, letter: string) => letter.toUpperCase())
+}
+
+function listSourceFiles(dir: string): string[] {
+  const files: string[] = []
+  for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
+    if (entry.isDirectory()) {
+      if (SKIPPED_DIRECTORIES.has(entry.name)) continue
+      files.push(...listSourceFiles(path.join(dir, entry.name)))
+      continue
+    }
+    if (!/\.tsx?$/.test(entry.name)) continue
+    if (/\.test\.tsx?$/.test(entry.name)) continue
+    files.push(path.join(dir, entry.name))
+  }
+  return files
+}
+
+function readCallArguments(source: string, openParenIndex: number): string | null {
+  let depth = 0
+  for (let index = openParenIndex; index < source.length; index += 1) {
+    const character = source[index]
+    if (character === '(') depth += 1
+    else if (character === ')') {
+      depth -= 1
+      if (depth === 0) return source.slice(openParenIndex + 1, index)
+    }
+  }
+  return null
+}
+
+function splitTopLevelArguments(args: string): string[] {
+  const parts: string[] = []
+  let depth = 0
+  let quote: string | null = null
+  let current = ''
+  for (let index = 0; index < args.length; index += 1) {
+    const character = args[index]
+    if (quote) {
+      if (character === quote && args[index - 1] !== '\\') quote = null
+      current += character
+      continue
+    }
+    if (character === "'" || character === '"' || character === '`') {
+      quote = character
+      current += character
+      continue
+    }
+    if (character === '(' || character === '{' || character === '[') depth += 1
+    if (character === ')' || character === '}' || character === ']') depth -= 1
+    if (character === ',' && depth === 0) {
+      parts.push(current)
+      current = ''
+      continue
+    }
+    current += character
+  }
+  if (current.trim().length > 0) parts.push(current)
+  return parts.map((part) => part.trim())
+}
+
+// The update payload is often built in a local variable rather than passed inline, so
+// resolve identifiers back to the object literal they were declared with plus every
+// `payload.field = …` assignment made on them before the write.
+function resolvePayloadText(payload: string, source: string): string {
+  if (!/^[A-Za-z_$][\w$]*$/.test(payload)) return payload
+  const declaration = new RegExp(`\\b(?:const|let|var)\\s+${payload}\\b[^=]*=\\s*`).exec(source)
+  let text = ''
+  if (declaration) {
+    const literalStart = source.indexOf('{', declaration.index + declaration[0].length - 1)
+    if (literalStart !== -1) {
+      let depth = 0
+      for (let index = literalStart; index < source.length; index += 1) {
+        if (source[index] === '{') depth += 1
+        else if (source[index] === '}') {
+          depth -= 1
+          if (depth === 0) {
+            text += source.slice(literalStart, index + 1)
+            break
+          }
+        }
+      }
+    }
+  }
+  const assignments = new RegExp(`\\b${payload}\\.[\\w$]+\\s*=[^=]`, 'g')
+  text += `\n${(source.match(assignments) ?? []).join('\n')}`
+  const bracketAssignments = new RegExp(`\\b${payload}\\[[^\\]]+\\]\\s*=[^=]`, 'g')
+  text += `\n${(source.match(bracketAssignments) ?? []).join('\n')}`
+  return text
+}
+
+const encryptedPropertiesByEntity = new Map<string, string[]>(
+  defaultEncryptionMaps.map((map) => [
+    toPascalCase(String(map.entityId).split(':').pop() ?? ''),
+    (map.fields ?? []).map((field) => toCamelCase(field.field)),
+  ]),
+)
+
+describe('customer_accounts encrypted fields are never written with nativeUpdate', () => {
+  const sourceFiles = listSourceFiles(MODULE_DIR)
+
+  it('finds the module sources and at least one encrypted entity to guard', () => {
+    expect(sourceFiles.length).toBeGreaterThan(0)
+    expect(encryptedPropertiesByEntity.size).toBeGreaterThan(0)
+    expect(encryptedPropertiesByEntity.get('CustomerUser')).toEqual(
+      expect.arrayContaining(['email', 'displayName']),
+    )
+  })
+
+  it.each(Array.from(encryptedPropertiesByEntity.entries()))(
+    '%s writes none of its encrypted properties through nativeUpdate',
+    (entityName, encryptedProperties) => {
+      const violations: string[] = []
+      for (const file of sourceFiles) {
+        const source = fs.readFileSync(file, 'utf8')
+        if (!source.includes('nativeUpdate')) continue
+        const pattern = /\bnativeUpdate\s*\(/g
+        let match: RegExpExecArray | null = pattern.exec(source)
+        while (match !== null) {
+          const args = readCallArguments(source, match.index + match[0].length - 1)
+          const parts = args === null ? [] : splitTopLevelArguments(args)
+          if (parts[0] === entityName && parts.length >= 3) {
+            const payloadText = resolvePayloadText(parts[2], source)
+            for (const property of encryptedProperties) {
+              if (new RegExp(`\\b${property}\\b`).test(payloadText)) {
+                violations.push(`${path.relative(MODULE_DIR, file)} → ${property}`)
+              }
+            }
+          }
+          match = pattern.exec(source)
+        }
+      }
+      expect(violations).toEqual([])
+    },
+  )
+})
--- a/packages/core/src/modules/customer_accounts/api/admin/users/[id].ts
+++ b/packages/core/src/modules/customer_accounts/api/admin/users/[id].ts
@@ -183,12 +183,21 @@ export async function PUT(req: Request, { params }: { params: { id: string } }) 
     }
   }
 
+  // `display_name` is encrypted at rest, and `nativeUpdate` skips the flush hooks the
+  // tenant-encryption subscriber relies on, so persisting it below would write plaintext PII
+  // into a ciphertext column (#3837). Route it through the service, which writes it via the
+  // managed entity. Runs before the `nativeUpdate` so the explicit `updated_at` below stays
+  // the value this response reports back as the optimistic-lock version.
+  if (parsed.data.displayName !== undefined) {
+    const customerUserService = container.resolve('customerUserService') as CustomerUserService
+    await customerUserService.updateProfile(user, { displayName: parsed.data.displayName })
+  }
+
   // Always bump updated_at so the optimistic-lock version advances on every save.
   // `nativeUpdate` bypasses MikroORM's `onUpdate` hook, so set it explicitly — without
   // this the version never changes and concurrent edits cannot be detected (#2055).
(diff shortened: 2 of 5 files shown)


pkarw:

- 🔍 `review` — the branch is complete and the full configured validation gate is green, so the PR opens ready for code review rather than as a draft.
- 🐛 `bug` — this corrects broken behavior that shipped: two write paths persisted an encrypted-at-rest column as plaintext, which is a defect rather than a new capability.
- 🔒 `security` — the defect defeats the tenant-data-encryption control for a PII field (`display_name`), so it is a security fix with GDPR / data-at-rest implications, matching the `security` label already on issue #3837.
- ⏭️ `skip-qa` — this change qualifies for the automated-verification exemption in `.ai/docs/pr-workflow.md`: it touches no UI-rendering file (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`), leaves the database structure and the API surface unchanged, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships automated tests for the changed behavior in the same PR. Manual browser QA would add nothing here because there is no visible UI change — the meaningful verification is at the database level (after editing a display name, the `display_name` column should now read as a `v1:…` ciphertext payload rather than plaintext), and that is covered by the tests plus the operator follow-up noted in the PR body.
- 🔺 `priority-high` — issue #3837 carries `priority-medium`, but the label-inference rules put security hardening in the high band, and when signals conflict the higher one wins: this is a live exposure that writes unencrypted PII on every profile edit today, so it should not slip a release. Flagging the disagreement explicitly so a maintainer can downgrade it if they prefer the original triage.
- ⚠️ `risk-high` — the change sits directly on encryption and PII-at-rest handling and alters how two write paths persist data, which is exactly the blast radius the high-risk band describes; it also matches the `risk-high` label the reporter put on the issue.

pkarw:
🤖 `om-open-pr` completed: fix committed, pushed and opened as this PR with the full SDLC label set. Lock handed off to `om-auto-review-pr` — chain continues on this PR.

pkarw:
🤖 `om-auto-review-pr` taking over the chain lock — automated code review with autofix. Started: 2026-08-13T13:06:02Z.

pkarw:
# 🔍 Code Review — PR #5262

**Verdict: APPROVE** (no blockers, no un-waivered majors). Submitted as a comment rather than a GitHub review because GitHub rejects self-approval — this PR was opened by the same automation account that reviewed it, so a maintainer approval is still required before merge.

## 🎯 Scope reviewed
Five files, all inside `packages/core/src/modules/customer_accounts`: two production files (`services/customerUserService.ts`, `api/admin/users/[id].ts`) and three test files. The change moves the encrypted-at-rest `display_name` off `em.nativeUpdate` and onto a managed-entity write plus `em.flush()`, at both write sites.

## ✅ Correctness
The fix targets the actual mechanism, not the symptom. `TenantEncryptionSubscriber` encrypts in `beforeCreate` / `beforeUpdate` (`packages/shared/src/lib/encryption/subscriber.ts:409-413`); `nativeUpdate` goes through the query builder and fires neither, which is precisely why plaintext reached the column. Assigning on a managed entity and flushing puts the write back on the hook path, matching how the staff auth module persists its analogous encrypted `name`.

The ordering in `api/admin/users/[id].ts:190-194` is correct and deliberately so: the profile write runs *before* the `nativeUpdate` that sets the explicit `updated_at` (#2055). Reversing it would leave the persisted optimistic-lock version later than the `nextUpdatedAt` the response returns, so the very next client save would 409 spuriously. The current order keeps the persisted version and the reported version identical.

Change-tracking behaviour is sound: `onLoad`/`afterFind` decrypt with `syncOriginal: true`, so the original snapshot holds plaintext, the new assignment marks the field dirty, and `beforeUpdate` encrypts on the way out. A no-op edit (same value) produces no UPDATE, which is the correct outcome.

## 🔒 Security & data scoping
- ✅ Uses `findOneWithDecryption` rather than raw `em.findOne`, with both `tenantId` and `organizationId` supplied to the decryption scope — the module's documented convention.
- ✅ **Finding (major, fixed in `79f89b1c4`):** the reload's `where` clause originally filtered on `id` + `tenantId` only, while the sibling `updatePassword` (`customerUserService.ts:103-111`) scopes by `id` + `tenantId` + `organizationId`. Reachability was nil — every caller already loads the entity under a full org scope — but defense in depth and consistency with the sibling argue for the explicit filter, so `organizationId` is now in the where clause and the unit test asserts it.
- ✅ No credentials, tokens, or PII values appear in logs, errors, or responses. No cross-tenant read path introduced.
- ✅ No hand-rolled crypto; everything routes through `TenantDataEncryptionService` via the subscriber.

## 💥 Breaking changes
None. `CustomerUserService.updateProfile` keeps its exported signature, both HTTP responses keep their shape, no schema, event ID, DI key, or ACL feature changed. Nothing on the `BACKWARD_COMPATIBILITY.md` protected-surface list is touched.

## 🧩 Findings not acted on (documented, non-blocking)

- **Minor — the admin PUT now performs two writes instead of one, without a surrounding transaction.** If the `nativeUpdate` failed after the profile flush succeeded, `display_name` would be persisted while `is_active` / `locked_until` were not. This does not introduce a new class of problem: the route was already multi-write and non-atomic, since the role sync below it does its own `nativeDelete` + `persist` + `flush` after the `nativeUpdate`. Making the whole handler atomic means converting every field to a managed-entity write and re-deriving the `updated_at` bump that #2055 depends on — a larger change to optimistic-locking semantics than the security bug being fixed here, and it belongs in its own PR. Flagged for a maintainer rather than folded in.
- **Nit — one extra `SELECT` per profile update.** The reload is defensive: it guarantees a managed instance instead of assuming the caller passed one attached to this `EntityManager`, where a wrong assumption would silently persist nothing. Both current callers do pass a managed entity, so the query is redundant today; correctness under a future caller was judged worth one indexed primary-key lookup on a low-frequency path.
- **Nit — a soft-deleted user's profile update is now a silent no-op** rather than a write, because the reload filters `deletedAt: null`. Unreachable from both existing callers (each already loads with `deletedAt: null`) and the safer behaviour of the two, so it is left as is.

## 🧪 Tests
Coverage is proportionate to the risk. Three layers, each verified failing against the unfixed sources and passing with the fix:

1. `services/__tests__/customerUserService.updateProfile.test.ts` — behavioural test of the service: managed-entity write plus `flush`, never `nativeUpdate`; the tenant/organization-scoped lookup; the caller-entity mirror; the no-op and missing-row paths.
2. `api/admin/users/__tests__/optimistic-lock.test.ts` — route-level regression proving the admin PUT keeps `displayName` out of its `nativeUpdate` payload and delegates instead, plus a case proving the service is not invoked when no display name is submitted. The existing `customerUserService` test double gained `updateProfile`; the gate caught that gap on the first run, which is the mechanism working as intended.
3. `__tests__/encryptedFieldWrites.test.ts` — a module-wide invariant guard deriving encrypted property names from `defaultEncryptionMaps` and failing on any `nativeUpdate` that writes one, resolving payload variables back to their declaration and assignments. It flagged both original sites, so it would also catch a future regression on `email`, session `ip_address`/`user_agent`, or invitation `display_name`.

**Validation gate** (local runner — no compose `app` container was running): `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app` — all green. `@open-mercato/core` reported 9627 tests across 1238 suites; `shared` 1876, `app` 549, `search` 311, `checkout` 140, `channel-imap` 109, `telemetry` 82, `cache` 73, `sync-akeneo` 62. Four `create-mercato-app` template dev-wrapper tests fail on this machine only, from a host limit rather than this branch: the wrapper's preflight reads `/proc/sys/fs/inotify/*` and aborts because `max_user_watches` is 253174 against a required 4194304 and `max_user_instances` is 128 against 4096. Nothing in this diff touches `packages/create-app`.

## ⏱️ CI status at review time
Not waited on, per the review contract. At submission `lint`, `ds-lint`, `scope`, `audit-scope`, `license/cla`, and the CodeQL `actions` and `python` analyses had passed; `prepare` and CodeQL `Analyze (javascript-typescript)` were still in progress. Required status checks are not configured on `develop` (the protection endpoint returns 404), so every reported check is treated as required and CI must still be green before merge.

## 📋 Merge readiness
No blockers. The PR carries `skip-qa` under the automated-verification exemption, so the QA gate does not hold it. It needs a maintainer's approving review — this automation cannot approve its own PR — and green CI.

pkarw:
🤖 `om-auto-review-pr` completed: APPROVE — no blockers and no un-waivered majors. One major finding (an organization filter missing from the reload's where clause) was fixed in `79f89b1c4` during the autofix pass; three lower-severity findings are documented in the review comment above with the reasoning for leaving them. The approving review itself could not be submitted through GitHub because an account cannot approve its own pull request, so a maintainer review is still required. Lock retained — chain continues.

pkarw:
🤖 `om-auto-fix-issue` run complete: the plaintext `display_name` write is fixed at both sites, the full validation gate is green, and the automated review returned APPROVE with one autofix commit. Lock released — `in-progress` swapped for `ci-monitoring` because CI is still running and this agent still owes a CI-result comment; the PR is free for anyone to act on in the meantime. Still needed before merge: a maintainer's approving review (self-approval is rejected by GitHub) and green CI.

pkarw:

**All checks green** on the merged head `c77672cf9`: `test`, `lint`, `ds-lint`, `scope`, `audit-scope`, `prepare`, `docker-build`, `ephemeral-integration`, `license/cla`, and CodeQL (`actions`, `python`, `javascript-typescript`) all passed. `audit`, `mutate`, `comment`, `merge-coverage`, and `[code]smith` were skipped by their own path/event filters.

**The earlier `test` failure and what it was.** The first CI run on this PR failed the `Test scripts` step with two assertions from `scripts/__tests__/check-lessons.test.mjs`: `.ai/lessons.md: 33525 bytes exceeds the 32768-byte progressive-loading budget` and `.ai/lessons.md: declared lesson count must equal 130`. Neither came from this PR's changes — the diff touches only `packages/core/src/modules/customer_accounts`, and `.ai/lessons.md` was untouched.

The cause was a stale base. This branch was cut from `34bd2d42f`, and `develop` had since advanced three commits (`fda1199ef`, `502205d1f`, `cbd3419efa`) which corrected the lessons catalog and its checker. Merging the current base into this branch (`c77672cf9`) makes the check pass locally and, as this run confirms, on CI. That is also the prescribed order of work for the autofix loop — merge conflicts and base drift are resolved against the latest base before anything else.

**Local re-validation after the merge.** The full configured gate was re-run on the merged tree: `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn build:app` all green, and `@open-mercato/core` reported 9632 tests passing across 1238 suites. Four `packages/create-app` template dev-wrapper tests fail on the local machine only, from a host limit rather than this branch — their preflight reads `/proc/sys/fs/inotify/*` and aborts because `max_user_watches` is 253174 against a required 4194304. CI runners carry the documented sysctl values, which is why `test` is green here.

`ci-monitoring` is removed — this agent owes no further follow-up on this PR. Remaining before merge: a maintainer's approving review, since GitHub rejects self-approval and this PR was opened by the reviewing account.

PR #5269, opened 2026-08-13
test(core): make module-decoupling.test.ts order-independent (#5129)

### LINKED ISSUE #5129: test: module-decoupling.test.ts is order-dependent — catalog registration leaks into the "catalog disabled" case
## Symptom

`packages/core/src/__tests__/module-decoupling.test.ts:231` fails intermittently under `yarn test`:

```
● Module Decoupling › 2. attachments/partitions.ts — resolveDefaultPartitionCode
  › returns privateAttachments for null, undefined, arbitrary strings, and catalog entity IDs

  expect(received).toBe(expected)

  Expected: "privateAttachments"
  Received: "productsMedia"

    229 |       expect(resolveDefaultPartitionCode('some-entity')).toBe('privateAttachments')
    230 |       // When catalog is disabled, the literal string also falls through
  > 231 |       expect(resolveDefaultPartitionCode('catalog:catalog_product')).toBe('privateAttachments')
```

It passes when `@open-mercato/core` runs on its own (1193 suites / 9212 tests green).

## Cause

The assertion's premise is that the **catalog module is not registered**. Getting `productsMedia` means it *was* registered when the case ran — so registry state from another test leaked in.

The module registry is process-global (`registerModules` in `@open-mercato/shared/lib/modules/registry`). Any suite sharing the worker that registers catalog and does not restore the previous registry will change this test's answer. Nothing about the assertion is wrong; its precondition is just not isolated.

## Why it shows up now

It is latent until test distribution shifts. It surfaced during #5127 in the same run as two jest-worker `SIGSEGV` crashes (#5100) — when a worker dies, jest redistributes its suites, which can co-locate a catalog-registering suite with this one.

So #5100 and this issue compound: the segfault reshuffles scheduling, and the reshuffle exposes the isolation bug. Fixing #5100 will make this *rarer*, not fixed — the latent ordering dependency stays.

## Suggested fix

Make the precondition explicit rather than ambient. Options, cheapest first:

- Snapshot and restore the registry around this suite (`beforeEach` capture, `afterEach` restore), so it cannot inherit or leak state.
- Have the suite register the exact module set it intends to test (the file's own guidance in `packages/core/AGENTS.md` → *Testing with Disabled Modules* already recommends `registerModules([...only what the test needs])`), and assert against that rather than against "whatever is registered".
- Failing both, run this file with `--runInBand` / its own project so distribution cannot affect it — a workaround, not a fix.

## Repro

Not reliably reproducible on demand, since it depends on worker scheduling. Observed on `sync/main-to-develop-2026-08-07` (#5127) with `yarn test` at the repo root; passes under `yarn workspace @open-mercato/core test`.

_Split out of #5127, where it was initially and incorrectly lumped in with the #5100 segfault._

adeptofvoltron:
Fixes #5129.

## What was actually wrong

The issue's diagnosis — "the catalog module was registered when the case ran" — points at the module registry, but `resolveDefaultPartitionCode` never reads it:

```ts
// packages/core/src/modules/attachments/lib/partitions.ts
import { E } from '#generated/entities.ids.generated'

const PRODUCT_MEDIA_ENTITY_IDS = new Set<string>(
  [(E as any).catalog?.catalog_product].filter(Boolean) as string[]
)
```

The set is frozen **at import time** from the generated entity ids. `packages/core/jest.mocks/entities.ids.generated.js` is a `Proxy` that answers *every* lookup, so `E.catalog.catalog_product` is `'catalog:catalog_product'` unless a suite mocks the module — which `module-decoupling.test.ts` does.

So the only way line 231 can see `productsMedia` is for the `partitions` module instance it imports to have been evaluated **without** that mock in effect. The suite's `await import(...)` returns whatever is already in the module registry, so it inherits such an instance instead of building its own.

Reproduced directly with a throwaway probe — a `partitions` instance cached from unmocked entity ids, then both import styles against it:

```
OLD => productsMedia        // plain `await import(...)`
NEW => privateAttachments   // `jest.isolateModulesAsync` + `await import(...)`
```

That is exactly the reported diff, and the isolated import is immune to it.

## Changes

**1. Pin the partitions precondition (the actual flake).** Load the module inside `jest.isolateModulesAsync` so its import-time entity-id set is always rebuilt from this file's mock, and assert `E.catalog` is absent first — if the mock ever stops applying, the failure now names the mock instead of looking like a behaviour regression.

**2. Snapshot and restore the global registries (the leak the issue describes).** `registerModules` writes to `globalThis` and `registerEntityIds` to module state; `beforeAll` replaced both and never restored them, so the reduced module set leaked to anything sharing the process. Now captured in `beforeEach` via `tryGetModules()` / `getEntityIds(false)` and restored in `afterEach`. This fixes the outbound direction of the same class of bug.

No behaviour under test changed — the assertions are identical.

## Validation

| gate | result |
|---|---|
| `module-decoupling.test.ts` | ✅ 12/12 |
| `module-decoupling.test.ts` + `src/modules/attachments` (26 suites) | ✅ 210/210 |
| leak probe (old vs new import, described above) | ✅ old reproduces `productsMedia`, new returns `privateAttachments` |

Two gates could not run in this worktree and are **unrelated to this diff** — its shared `node_modules` is installed from `main` and is stale for `develop`:

- `src/modules/attachments/components/__tests__/assignmentsEditorFocus.test.tsx` — `Cannot find module '@tanstack/react-table/legacy'` (develop declares `^9.0.0`, installed is `8.21.3`)
- `eslint` — `Cannot find module 'eslint-config-next/core-web-vitals'`

`packages/core/tsconfig.json` excludes `**/__tests__/**`, so `yarn typecheck` does not cover this file either way. CI will run all three on a clean install.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5269"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789222360&installation_model_id=432243&pr_number=5269&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5269&signature=701d8dda8c1e13340fbe4378707075e79c03145cf17c397875fb81c7edd95fc5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/__tests__/module-decoupling.test.ts
+++ b/packages/core/src/__tests__/module-decoupling.test.ts
@@ -87,7 +87,8 @@ jest.mock('@open-mercato/shared/lib/crud/cache', () => ({
 }))
 
 import { registerEntityIds, getEntityIds } from '@open-mercato/shared/lib/encryption/entityIds'
-import { registerModules } from '@open-mercato/shared/lib/modules/registry'
+import type { EntityIds } from '@open-mercato/shared/lib/encryption/entityIds'
+import { registerModules, tryGetModules } from '@open-mercato/shared/lib/modules/registry'
 import type { Module } from '@open-mercato/shared/modules/registry'
 import type { ModuleSetupConfig } from '@open-mercato/shared/modules/setup'
 
@@ -193,9 +194,24 @@ function buildReducedModules(): Module[] {
 
 const reducedModules = buildReducedModules()
 
-beforeAll(() => {
+// The module registry and the entity-id registry are process-global, so this
+// suite must own its own snapshot of both instead of relying on whatever the
+// surrounding run happens to have registered.
+let previousModules: Module[] | null = null
+let previousEntityIds: EntityIds | null = null
+
+beforeEach(() => {
+  previousModules = tryGetModules()
+  previousEntityIds = getEntityIds(false)
   registerEntityIds(reducedE as any)
   registerModules(reducedModules)
+})
+
+afterEach(() => {
+  registerEntityIds(previousEntityIds ?? ({} as EntityIds))
+  registerModules(previousModules ?? [])
+  previousModules = null
+  previousEntityIds = null
 })
 
 describe('Module Decoupling', () => {
@@ -220,9 +236,17 @@ describe('Module Decoupling', () => {
 
   describe('2. attachments/partitions.ts — resolveDefaultPartitionCode', () => {
     it('returns privateAttachments for null, undefined, arbitrary strings, and catalog entity IDs', async () => {
-      const { resolveDefaultPartitionCode } = await import(
-        '@open-mercato/core/modules/attachments/lib/partitions'
-      )
+      // partitions.ts freezes its product-media entity ids at import time, so it
+      // has to be evaluated inside an isolated registry — a cached instance built
+      // from unmocked entity ids would silently answer 'productsMedia' here.
+      let resolveDefaultPartitionCode!: (entityId: string | null | undefined) => string
+      await jest.isolateModulesAsync(async () => {
+        const { E } = await import('#generated/entities.ids.generated')
+        expect((E as Record<string, unknown>).catalog).toBeUndefined()
+        ;({ resolveDefaultPartitionCode } = await import(
+          '@open-mercato/core/modules/attachments/lib/partitions'
+        ))
+      })
 
       expect(resolveDefaultPartitionCode(null)).toBe('privateAttachments')
       expect(resolveDefaultPartitionCode(undefined)).toBe('privateAttachments')


adeptofvoltron:
Labels: `bug` + `test` (test-only fix for a flaky suite), `priority-medium` / `risk-low` carried from #5129, `review` as the pipeline state.

`skip-qa` under the automated-verification exemption: the diff touches one `.test.ts` file only — no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no DB schema change, no API surface change, no `BACKWARD_COMPATIBILITY.md` contract touched. The changed behaviour *is* the automated test, and it ships green in this PR.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-08-18T08:53:01Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron:
# 🔍 Code Review: test(core): make module-decoupling.test.ts order-independent (#5129)

## 🎯 Summary

This PR makes `packages/core/src/__tests__/module-decoupling.test.ts` own its preconditions instead of inheriting them from whatever the surrounding Jest run happens to have evaluated. It does two things. First, it loads `attachments/lib/partitions` inside `jest.isolateModulesAsync` and asserts up front that the mocked `E.catalog` really is absent — `partitions.ts` freezes `PRODUCT_MEDIA_ENTITY_IDS` at import time (`packages/core/src/modules/attachments/lib/partitions.ts:29-31`), so a module instance evaluated before this file's `jest.mock` took effect would answer `productsMedia` and make an isolation bug look like a behaviour regression. Second, it moves the registry setup from `beforeAll` to `beforeEach` and adds an `afterEach` that restores the previously registered modules and entity ids.

The reviewed scope is that one test file; no production code changed, and the assertions themselves are untouched. The diagnosis in the PR description is the interesting part of this change, and it holds up: I reproduced the mechanism independently in this worktree with a throwaway probe — a `partitions` instance cached from unmocked entity ids answered `productsMedia` for a plain `await import(...)`, while `jest.isolateModulesAsync` rebuilt it from the mock and answered `privateAttachments`. That is exactly the reported symptom, and the new import style is immune to it.

The strongest part of the change is the added `expect((E as Record<string, unknown>).catalog).toBeUndefined()` precondition. It converts a future mock-resolution failure into a failure that names the mock, rather than a red assertion four lines later that reads like the attachments partition logic regressed. That is the right instinct for a test whose whole premise is "catalog is not registered".

My findings are all minor or nit — nothing blocks the merge. Two of them are about the second half of the change (the `beforeEach`/`afterEach` snapshot), which does less than the description claims, and one is about consistency with the sibling cases in the same file.

## Verdict

✅ **approve** — the change is test-only, strictly increases the suite's hermeticity, cannot regress production behaviour, and the mechanism it fixes is independently reproducible. The three minors and one nit below are all pick-up-later items about the registry-snapshot half of the change and the accuracy of its description; none of them makes the suite less correct than it is on `develop` today.

## 🧪 Validation Gate

This review ran in an isolated worktree that shares the parent checkout's `node_modules` (installed from `main`) and has no Yarn install state of its own, so `yarn <script>` fails with `Couldn't find the node_modules state file` and every `turbo`-driven gate is unrunnable here. I ran what could be invoked directly and am reporting exactly that, with CI on head `ec45447` as the authoritative evidence for the rest.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ⚠️ NOT RUN | `turbo run build` needs the Yarn install state this review worktree does not have. CI's `prepare` and `test` jobs cover it on a clean install (both ✅ on `ec45447`). |
| `yarn generate` | ⚠️ NOT RUN | Same reason. Its absence is itself visible below: `packages/core/generated/**` does not exist here, which is what breaks 4 unrelated catalog suites locally. |
| `yarn build:packages` (second pass) | ⚠️ NOT RUN | Same reason. |
| `yarn i18n:check-sync` | ✅ PASS | Ran directly via `tsx scripts/i18n-check-sync.ts`: 5 locales, 52 modules with translations, all translation files in sync. |
| `yarn i18n:check-usage` | ✅ PASS (advisory) | Ran directly via `tsx scripts/i18n-check-usage.ts`: 3661 unused keys reported, all pre-existing — the diff adds no user-facing strings and no `t()` calls. |
| `yarn typecheck` | ⚠️ PARTIAL | Ran `tsc -p packages/core/tsconfig.json --noEmit` directly. The only errors are `Cannot find module '@tanstack/react-table/legacy'` and its knock-on implicit-`any` errors in `packages/ui/src/backend/DataTable.tsx` — `develop` declares `@tanstack/react-table@^9.0.0` while the shared install has `8.21.3`. Worth knowing: `packages/core/tsconfig.json` excludes `**/__tests__/**`, so this gate does not type-check the changed file at all, on CI either. |
| `yarn test` | ✅ PASS for the change, ⚠️ PARTIAL for the package | `module-decoupling.test.ts` passes **12/12** standalone, **12/12** in-band alongside the 26 `src/modules/attachments` suites (210/210 tests), and passes again inside a full `packages/core` run of **1236 suites / 9513 tests → 1214 suites and 9511 tests green**. All 22 failing suites are environment artifacts of this worktree, none of them touching the diff: 18 fail on the missing `@tanstack/react-table/legacy`, 4 on the missing `packages/core/generated/**` (`#generated/entities/catalog_product`). |
| `yarn build:app` | ⚠️ NOT RUN | Same install-state reason; CI's `docker-build`/`prepare` path covers it. |
| _CI on `ec45447` (evidence for the gates above)_ | ✅ ALL GREEN | `test`, `lint`, `ds-lint`, `prepare`, `scope`, `audit-scope`, `CodeQL`, `Analyze (javascript-typescript)`, `Analyze (python)`, `Analyze (actions)` and `license/cla` are all SUCCESS; the remaining checks are SKIPPED. No required check is failing and none is pending. `eslint` could not be run locally either — its config cannot resolve `eslint-config-next/core-web-vitals` in this worktree — but CI's `lint` job is green. |

## Findings

### 🔹 Minor

**`packages/core/src/__tests__/module-decoupling.test.ts:211-212` — the `afterEach` cannot restore the "never registered" state, so it silently converts it into "registered as empty".**

Both registries use `null` as their unregistered sentinel and throw on read: `getModules()` throws `[Bootstrap] Modules not registered` (`packages/shared/src/lib/modules/registry.ts:94-99`) and `getEntityIds()` throws `[Bootstrap] Entity IDs not registered` (`packages/shared/src/lib/encryption/entityIds.ts:28-33`). Restoring with `registerModules(previousModules ?? [])` and `registerEntityIds(previousEntityIds ?? ({} as EntityIds))` writes `[]` and `{}` instead — both truthy, so after the first test in this file those reads stop throwing and start returning empty. Nothing in this file depends on that throw and nothing outside it can see the difference (see the next finding), so this is not a live bug; but "restore" is doing something subtly different from what it says. If you want a true restore, the registries would need an explicit reset (e.g. an exported `resetModules()` / `resetEntityIds()` used when the snapshot was `null`); if that is not worth adding, a one-line comment saying the pre-suite `null` state is deliberately not reproducible would keep the next reader from assuming it is.

**`packages/core/src/__tests__/module-decoupling.test.ts:197-215` — the snapshot/restore is defence-in-depth, not the "outbound leak" fix the PR description claims.**

The description says the `afterEach` "fixes the outbound direction of the same class of bug", i.e. that the reduced module set was leaking to other suites. It cannot: with `testEnvironment: 'node'` (`packages/core/jest.config.cjs:6`) Jest gives every test *file* its own global object, so `globalThis.__openMercatoModulesRegistry__` — the storage `registerModules` writes to (`packages/shared/src/lib/modules/registry.ts:17-24`) — is never shared across files. I verified this directly here with a two-file probe run under `--runInBand`: the writer file set the registry and the reader file that ran immediately after it saw `null`. The same reasoning applies to issue #5129's own stated cause ("registry state from another test leaked in"), which is why your re-diagnosis was necessary in the first place. The code is harmless and I would keep it as belt-and-braces, but the PR description is the thing that will be read later — it is worth correcting there so nobody cites this PR as evidence that cross-file registry leaks are a thing in this suite.

**`packages/core/src/__tests__/module-decoupling.test.ts:203-215` — `beforeEach`/`afterEach` buys no isolation over `beforeAll`/`afterAll` here, and `registerModules` is not a cheap or pure setter.**

No test in this file mutates either registry, so re-registering before and after each of the 12 tests does the same work 24 times that `beforeAll`/`afterAll` would do twice. That matters slightly more than usual because `registerModules` is not a plain assignment: it merges through `preserveExistingTranslations(existing ?? [], nextModules)` and invalidates the whole i18n dictionary cache on every call (`packages/shared/src/lib/modules/registry.ts:74-93`). With `beforeEach`, each registration now merges against the previously *restored* state rather than against a clean one. It is harmless in this suite (the reduced modules carry no `translations` and nothing here reads a dictionary), but `beforeAll`/`afterAll` would be both simpler and a closer match to what the suite actually needs. If per-test registration was deliberate — guarding against a future test in this file registering something — a short comment saying so would be worth more than the cycles.

### 💅 Nit

**`packages/core/src/__tests__/module-decoupling.test.ts:211` — the `?? ({} as EntityIds)` fallback is unreachable and the `EntityIds | null` type is wider than reality.**

`getEntityIds(false)` returns `{} as EntityIds` when nothing is registered (`packages/shared/src/lib/encryption/entityIds.ts:29-32`); it never returns `null` or `undefined`. So after `beforeEach` has run, `previousEntityIds` is always an object and the `??` branch is dead code. Typing it as `EntityIds` and dropping the fallback would say what is actually true. (`previousModules` is genuinely `Module[] | null`, because `tryGetModules()` does return `null` — that one is fine.)

**`packages/core/src/__tests__/module-decoupling.test.ts:261-263, 355-357` — the sibling dynamic imports in this file keep the pattern this PR just established as unsafe.**

Sections 3 and 5 still do a plain `await import(...)` of `attachments/lib/assignmentDetails` and `configs/lib/upgrade-actions`. `assignmentDetails.ts` is a weaker version of the same hazard: it memoises its entity-link specs in a module-level `_entityLinkSpecsCache` (`packages/core/src/modules/attachments/lib/assignmentDetails.ts:15-20`), so a cached instance whose memo was already populated from unmocked ids would be reused there too. It is not a live flake — that memo is filled at call time, after `beforeEach` has registered `reducedE`, and those assertions pass regardless because `queryEngine` is `null` — so I am not asking you to change it. But this file is now the canonical worked example of "pin your preconditions", and leaving two neighbours on the old pattern with no note makes the lesson ambiguous for the next person who copies from it. Either pin them the same way or add one line saying why they do not need it.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff adds two imports (`EntityIds` type, `tryGetModules`) and exports nothing.
- [x] No function signature changed in a breaking way — no function signatures were touched.
- [x] Where a contract had to change: not applicable — nothing under `BACKWARD_COMPATIBILITY.md`'s 13 protected contract surfaces is touched. The one file in the diff is a test.

## 🧪 Test Coverage

The diff *is* the test change, so the question is whether the new version still covers what the old one did and whether the fix itself is verifiable.

Coverage is unchanged and complete on the first count: all 12 assertions survive untouched — the entity-id registry exclusions (sections 1), `resolveDefaultPartitionCode`'s four cases including the catalog entity id (section 2), `resolveAssignmentEnrichments` against empty and catalog-referencing assignments (section 3), the admin/employee/superadmin role-feature merging (section 4), and the `upgrade-actions` module load (section 5). The suite gains one new assertion, the `E.catalog` precondition check, which is a genuine addition rather than a weakened one.

On the second count there is an unavoidable gap, and it is worth stating plainly rather than papering over: there is no test that fails without this fix. Issue #5129 says the flake was never reliably reproducible on demand — it depends on Jest worker scheduling — so the usual "regression test that goes red on the parent commit" is not available here. What stands in for it is the mechanism check, and I ran that independently rather than taking the PR description's word for it: a throwaway probe in this worktree loaded `partitions` from unmocked entity ids, then compared both import styles against that cached instance, and got `productsMedia` from the plain `await import(...)` versus `privateAttachments` from `jest.isolateModulesAsync`. That confirms both halves of the claim — the failure mode is real, and the isolated import is immune to it. The in-file `expect(E.catalog).toBeUndefined()` assertion then acts as the standing guard, so if the mock ever stops applying the suite fails with a message that names the cause.

No further test cases are needed for this change.

---

<sub>🤖 Reviewed by `om-auto-review-pr`. GitHub does not allow approving one's own pull request, so this approval is recorded as a comment rather than a formal review approval — see the run summary for details.</sub>

adeptofvoltron:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, so the PR moves out of `review` and into the merge queue; QA is not required for it (see `skip-qa` below), which means nothing else gates the merge beyond the already-green required checks.
- ⏭️ `skip-qa` — the automated-verification exemption applies cleanly here: the diff is a single `.test.ts` file, with no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no database structure change, no API surface change and no `BACKWARD_COMPATIBILITY.md` contract touched, and the changed behaviour is itself covered by the test that changed. There is no user-facing surface for a human to click through.
- 🐛 `bug` — the PR fixes a defect (an order-dependent test that fails intermittently in CI) rather than adding a capability, even though the defect lives in test code.
- 🏷️ `test` — carried over from #5129 to mark that the change is confined to the test suite; it is additive and orthogonal to the pipeline state.
- 🔹 `priority-medium` — kept as inherited from #5129 and confirmed by the review: an intermittently failing suite costs real CI cycles and erodes trust in the signal, but it blocks no release and puts no data or user at risk.
- 🟢 `risk-low` — kept as inherited from #5129 and confirmed by the review: the blast radius is one test file, no production code path is touched, and the full `packages/core` run (1236 suites) shows no suite affected by the change other than the one it fixes.

adeptofvoltron:

The review found no blockers and no majors, so the verdict is approve and the PR is now in `merge-queue`. GitHub rejects self-approval (`Can not approve your own pull request`), so the full review report was posted as a comment rather than a formal review approval — the PR still shows `REVIEW_REQUIRED` and needs a second pair of eyes to click Approve before it can merge. That is a GitHub constraint, not a reservation about the change.

Substance: the PR's re-diagnosis of #5129 holds up. I reproduced the mechanism independently in the review worktree — a `partitions` instance cached from unmocked entity ids answers `productsMedia` for a plain `await import(...)`, while `jest.isolateModulesAsync` rebuilds it from this file's mock and answers `privateAttachments`. The four findings are three minors and one nit, all about the `beforeEach`/`afterEach` registry-snapshot half of the change: the restore writes `[]`/`{}` rather than reproducing the `null` unregistered sentinel, the snapshot is defence-in-depth rather than the "outbound leak" fix the description claims (Jest's `node` environment gives each test file its own `globalThis`, which I verified with a two-file probe), `beforeAll`/`afterAll` would do the same job with 2 registry writes instead of 24, and the `?? ({} as EntityIds)` fallback is unreachable because `getEntityIds(false)` never returns null. None of them blocks the merge; the PR description is the one thing genuinely worth correcting.

autofix: not run — the verdict is approve, so there were no actionable blocker or major findings to fix, and pushing cosmetic follow-up commits to an already-green, already-approved head would only restart CI and invalidate this review. The minors and the nit are left for the author to pick up.

Validation: `module-decoupling.test.ts` passes 12/12 standalone, 12/12 in band with the 26 `src/modules/attachments` suites, and again inside a full `packages/core` run of 1236 suites / 9513 tests. The 22 suites that fail there are environment artifacts of the review worktree (18 on the missing `@tanstack/react-table/legacy`, 4 on the missing `packages/core/generated/**`), none touching the diff. The `turbo`-driven gates could not run here at all; CI on `ec45447` is fully green and stands as the evidence for those. No CI follow-up is owed — every required check was already settled and green at review time, so no `ci-monitoring` was applied.

patzick (approved):

PR #4897, opened 2026-08-03
feat(create-app): complete the canonical example and standalone harness

pkarw:
Tracking plans:

- `.ai/runs/2026-08-03-implement-standalone-canonical-example.md`
- `.ai/runs/2026-08-05-canonical-example-milestones-bcd.md`

Status: **implementation, local certification, and CI stabilization complete**. Exact-head evidence-only
browser QA has passed. The four hosted-CI failures on `2a4765f33` were fixed at root cause (see the
`om-auto-continue-pr` resume summary comment) and the affected specs pass 37/37 locally.
The remaining release gates are the provider-backed harness lane on Linux/Bubblewrap, the hosted PR CI
result on the new head, non-author `qa-approved` sign-off, and an approving review from someone other
than the author.

## Goal

Implement the four specifications merged in PR #4878:

- `.ai/specs/2026-07-31-standalone-canonical-example-module.md`
- `.ai/specs/2026-08-01-standalone-agent-spec-first-routing.md`
- `.ai/specs/2026-08-01-standalone-harness-example-read-policy.md`
- `.ai/specs/2026-08-01-standalone-harness-knowledge-governance.md`

The result is one canonical `example` module that ships as source in every create-app preset while
remaining runtime-disabled until explicitly registered. Its app and template trees are byte-identical.

## What landed

- Completed the canonical module surface: scoped Todo CRUD, optimistic locking, encryption/search,
  cache + DI, setup hooks, events and portal broadcast, notifications, AI overrides, component and
  injection hosts, bulk progress/outbox/leased workers, vector/workflow/generator conventions, and
  credential-free payment/shipping/currency integration references.
- Added deterministic local reference facts and complete enum-derived topology coverage. Activated
  extraction now resolves every canonical target and diagnostic family without fallback or unresolved
  first-party rows.
- Added the missing runtime cases `TC-EXAMPLE-012`, `TC-EXAMPLE-015`, and `TC-EXAMPLE-016`, including
  a disposable standalone app that is built once disabled, explicitly activated, generated twice for
  byte parity, built again, booted, and cleaned up.
- Completed standalone source-link, read-policy, spec-routing, design-foundation, and knowledge-change
  governance. Fresh scaffolds now receive and hash-own the disabled example's JSON/Markdown reference
  projection, and the complete deterministic catalog passes 228/228.
- Fixed defects found by the new coverage, including cross-request custom-field key leakage, empty AI
  tool input schemas, duplicate first-run ACL grants, stale dynamic-form setup, generator manifest
[...]

PR #5270, opened 2026-08-13
fix(sales): give the payments table a real tableId so the gateway-status column binds (#5142)

### LINKED ISSUE #5142: sales: payments DataTable has no tableId, so the gateway-status column widget can never bind
## Summary

`sales.injection.payment-gateway-status-column` is a **registered but unbound** injection widget: it renders nowhere, and it cannot, because the payments table it targets declares no table id.

## Why it is unbound

`PaymentsSection.tsx` renders its `DataTable` without `perspective`, `injectionSpotId` or `extensionTableId`. `DataTable.tsx` resolves `perspective?.tableId` first and falls back through those props, so with none of them supplied `extensionTableId` is `null` and the columns spot degrades to `'__disabled__:columns'`. The `'data-table:sales.payments:columns'` binding therefore never matched any spot.

PR #4897 removed that dead binding from `packages/core/src/modules/sales/widgets/injection-table.ts` (removing it broke nothing — it could never resolve) and left a `NOTE` at the removal site. The widget itself stays registered.

## The actual gap

Giving the payments `DataTable` a real table id — the same shape `ProductsDataTable.tsx` uses, `perspective={{ tableId: extensionPoints.hosts.productsTable.tableId }}` — so the gateway-status column finally renders, and restoring the binding once it can resolve. That is a sales feature change, not a widget-registry cleanup, which is why #4897 recorded it rather than reintroducing a binding that would still be inert.

## Acceptance

- [ ] The payments `DataTable` declares a stable table id through a host extension point.
- [ ] `'data-table:sales.payments:columns'` is restored in `sales/widgets/injection-table.ts` and resolves.
- [ ] A test pins that the gateway-status column actually renders in the payments table (the previous state passed every test while rendering nothing).
- [ ] The `NOTE` at the removal site is replaced by the real binding.

Raised in the code review of #4897.

adeptofvoltron:
Closes #5142

## 🎯 Goal
- The gateway-status column now actually renders in the payments table on the sales document detail page. `sales.injection.payment-gateway-status-column` was registered but unbound — it rendered nowhere and could not, because the payments table declared no table id.

## 🔍 Problem
`PaymentsSection.tsx` rendered its `DataTable` without `perspective`, `injectionSpotId` or `extensionTableId`, so no module could extend the payments table at all. PR #4897 removed the `'data-table:sales.payments:columns'` binding (removing it broke nothing — it could never resolve) and left a `NOTE` at the removal site recording the gap.

## 🔍 Root Cause
`DataTable` resolves `extensionTableId` from `perspective?.tableId → extensionTableId → injectionSpotId` (`packages/ui/src/backend/DataTable.tsx:1468-1473`). With none of the three supplied it is `null`, and every deep-extension surface falls back to a disabled sentinel — the columns spot degrades to `'__disabled__:columns'` (`DataTable.tsx:1516`). `sales/extension-points.ts` declared hosts for `sales.orders` and `sales.quotes` but none for payments, so there was no table id for a binding to target.

## What Changed
- `packages/core/src/modules/sales/extension-points.ts` — new `paymentsTable` host: `dataTableExtensionHost({ tableId: 'sales.payments', source: 'components/documents/PaymentsSection.tsx' })`. This is the stable, declared table id third-party modules extend against; the generator's host-binding guard (`module-extension-facts.ts`) requires the declared `source` file to reference the host, which `PaymentsSection.tsx` now does.
- `packages/core/src/modules/sales/components/documents/PaymentsSection.tsx` — the payments `DataTable` now passes `extensionTableId={extensionPoints.hosts.paymentsTable.tableId}`. `extensionTableId` is used rather than the `perspective={{ tableId }}` shape `ProductsDataTable` uses: it is the documented cheaper path when a host only needs widget injection (`DataTable.tsx:347-355`), and it keeps this document-detail sub-section out of the saved-views/perspective machinery it has no use for.
- `packages/core/src/modules/sales/widgets/injection-table.ts` — the `NOTE` is replaced by the restored `'data-table:sales.payments:columns'` binding, which now resolves.
- `packages/cli/src/lib/generators/__tests__/module-facts.extension-hosts.test.ts` — the shipped-hosts fixture picks up `sales.payments`.

## 🧪 Tests
- `packages/core/src/modules/sales/widgets/__tests__/injection-table.test.ts` (new) — derives the columns spot id from `extensionPoints.hosts.paymentsTable.tableId` via `DataTableInjectionSpots.columns(...)` and asserts the binding key matches it exactly. Spot resolution is exact-match, so a drift between the host id and the binding key is what silently unbinds the widget; this pins them together.
- `packages/core/src/modules/sales/components/documents/__tests__/PaymentsSection.gatewayColumn.test.tsx` (new) — renders `SalesDocumentPaymentsSection` with the real `DataTable` and the real widget module, and asserts the "Gateway status" header **and** a cell render in the payments table, plus that the table asks for the spot the restored binding targets. Verified failing on `develop` (3 failed) and passing with this change — the previous state passed every test while rendering nothing.
- Gate run locally (no compose `app` container present, so local mode): `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync` (in sync), `yarn i18n:check-usage` (advisory only), `yarn typecheck` (22/22), `yarn test`, `yarn build:app`.

## ⚠️ Known follow-up (not in this PR)
The column's `accessorKey` is `_gateway.unifiedStatus`, but **nothing produces that field today**: `packages/core/src/modules/sales/api/payments/route.ts` does not opt into `enrichers`, `payment_gateways/data/enrichers.ts` is deliberately empty, and `PaymentsSection`'s `PaymentRow` mapping keeps only known keys. So the column renders the widget's own documented `'pending'` fallback for every row until a gateway module enriches `sales.payment`. That response enricher is listed in SPEC-044 ("Gateway transaction details on payment detail | Response Enricher") and is a separate change; this PR delivers exactly the four acceptance items on #5142 — table id, restored binding, rendering test, `NOTE` replaced. Flagging it so the reviewer decides whether the column should ship before its data source exists.

## 💥 Breaking Changes
- None. `'data-table:sales.payments:*'` is a new host surface (additive); no existing spot id, API, or DB shape changes.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5270"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789223204&installation_model_id=432243&pr_number=5270&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5270&signature=5f6a12fee0b9911e2430237325619909430af21633e4d7e307d5631fc89f1ad5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:
🤖 Label rationale (`om-open-pr`):

- **`review`** — ready, non-draft PR entering the pipeline; no other pipeline state applies.
- **`bug`** — the issue is a defect: a registered injection widget that could never render. Matches the `bug` label on #5142.
- **`needs-qa`** — the automated-verification exemption does not apply: the change touches a `.tsx` file under `**/components/**` (`PaymentsSection.tsx`) and adds a user-visible column to the payments table on the sales document detail page. Manual QA should confirm the column appears and the table's existing behavior (row click → edit dialog, row actions, totals refresh) is unaffected.
- **`priority-low`** — inherited from #5142; the widget has been inert since it shipped, so nothing regresses while this waits.
- **`risk-low`** — additive only. A new data-table extension host (`data-table:sales.payments:*`), one new prop on one `DataTable` call site, and a restored binding. No API, DB, or existing spot-id contract changes.

Not applied: `skip-qa` (see above), `qa-approved` (owned by the QA gate).

🤖 `om-open-pr` — chain lock transferred from #5142 to this PR; continuing with `om-auto-review-pr`.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock on this PR (inherited from `om-open-pr`). Reviewing now.

adeptofvoltron:
# 🔍 Code Review

**Verdict: no blockers, no majors, no mediums.** GitHub does not allow approving one's own pull request, so this is posted as a review report rather than a formal approval — the PR stays in `review` and still needs a human approver. `om-auto-review-pr` found nothing that must change before merge.

**Validation source:** GitHub checks were still pending at review time (`prepare`, `ds-lint`, `Analyze (javascript-typescript)`); the full `validation.commands` gate was run locally earlier in this same chain — see the Validation section below. **Runner:** local (no compose `app` container is running on this machine).

**Head:** `a47b5b4` · **Base:** `develop` · **Mergeable:** yes, no conflicts.

## ✅ What this change gets right

The fix is the minimal one that makes the extension point real rather than nominal. The table id is declared once, as a host in `packages/core/src/modules/sales/extension-points.ts:8`, and every consumer derives from it: `PaymentsSection.tsx:358` passes `extensionPoints.hosts.paymentsTable.tableId` instead of a string literal, and the new `injection-table.test.ts` derives the binding key through `DataTableInjectionSpots.columns(...)` rather than hard-coding it a second time. That closes the exact failure mode the issue describes — a binding key and a table id drifting apart with nothing to catch it.

The host's `source` field points at `components/documents/PaymentsSection.tsx`, and that file now references `extensionPoints.hosts.paymentsTable`, which is what the generator's `hasDeclarationBinding` check (`packages/cli/src/lib/generators/module-extension-facts.ts:558`) requires to stop reporting the host as `unbound-declaration`. The fixture update in `module-facts.extension-hosts.test.ts` is the honest consequence of that, not a test being bent to fit.

Choosing `extensionTableId` over the `perspective={{ tableId }}` shape `ProductsDataTable` uses is the right call for this host and is argued in the PR body: `DataTable.tsx:347-355` documents `extensionTableId` as the cheaper path when a host only needs widget injection, and the payments table is a sub-section of a document detail page that has no business acquiring saved views or the perspectives query.

## 🧪 Test coverage

Adequate, and — importantly — verified to actually fail without the fix. Both new files were run against a stashed working tree on `develop`: 3 failed, 3 passed with the change. That directly addresses the issue's warning that "the previous state passed every test while rendering nothing."

`PaymentsSection.gatewayColumn.test.tsx` renders the real `DataTable` with the real widget module and asserts both the translated header and a rendered cell, so it fails if the column header appears without a body or if the spot id drifts. Mocking `useInjectionDataWidgets` by spot id (rather than mocking `DataTable` itself, as the neighbouring `SalesDocumentsTable.stickyColumn.test.tsx` does) is deliberate: mocking the table would reproduce the very blind spot this issue is about.

## 🔁 Backward compatibility

No protected surface changes. `data-table:sales.payments:*` is a **new** host surface — additive per `BACKWARD_COMPATIBILITY.md`'s widget-spot-id category. No existing spot id, event id, API route, DB column, DI key, or ACL feature is touched. `PaymentsSection`'s props are unchanged. The restored binding re-adds a key that resolves to a widget which has been registered since it shipped.

## 🟢 Low / nit findings (non-blocking, no action required in this PR)

1. **`packages/core/src/modules/sales/widgets/injection/payment-gateway-status-column/widget.ts:12` — the column's data source does not exist yet.** `accessorKey: '_gateway.unifiedStatus'` has no producer: `sales/api/payments/route.ts` does not opt into `enrichers`, `payment_gateways/data/enrichers.ts` is deliberately empty, and `PaymentsSection`'s `PaymentRow` mapping keeps only known keys. Every row will show the widget's own `'pending'` fallback until a gateway module enriches `sales.payment`. This is disclosed in the PR body and is explicitly outside the four acceptance items on #5142; SPEC-044 lists that response enricher as its own line item. Flagged so the reviewer can decide whether the column should ship ahead of its data — that is a product call, not a code defect.

2. **Same file — no `requiredModules: ['payment_gateways']` on the widget metadata.** Now that the widget actually loads, an install with `payment_gateways` disabled still gets a gateway-status column. `requiredModules` is honored for column widgets (`injection-loader.ts:524` via `loadInjectionDataWidgetById`), so this is enforceable. Not fixed here because the already-bound sibling widget `sales.injection.payment-gateway-config-field` has exactly the same shape — fixing one and not the other would be worse than fixing neither, and fixing both is beyond this issue. Worth a follow-up covering both widgets.

## 🧪 Validation

Run locally, in `validation.commands` order: `yarn build:packages` ✅ · `yarn generate` ✅ · `yarn build:packages` ✅ · `yarn i18n:check-sync` ✅ (all 5 locales in sync; the header key `payment_gateways.column.gatewayStatus` already exists in each) · `yarn i18n:check-usage` ✅ (advisory only) · `yarn typecheck` ✅ (22/22) · `yarn test` ✅ · `yarn build:app` ✅.

One caveat, stated rather than buried: four files in `packages/create-app` (`writable-ast-oracles`, `business-writable-oracles`, `agent-harness-evaluator`, `agent-harness-release`) fail on this machine. They spawn live Codex/Claude CLI processes and run a scaffolded app's `yarn typecheck` under a 120 s budget, and they fail on unmodified `develop` for the same environment reasons. None of them touch the diff, and the create-app **template-parity** guards — the ones CI runs unconditionally on every PR — all pass. CI is the authority here.

## Pipeline

Left in `review`. An automated run cannot approve its own PR, and moving it to `merge-queue` on the automation's own say-so would misrepresent that. `needs-qa` is retained; manual QA instructions follow in a separate comment.

adeptofvoltron:

This PR carries `needs-qa` with no `skip-qa`, and the QA-approval gate holds the merge until `qa-approved` is added. It is left in `review` (not `merge-queue`) because the automated run cannot approve its own PR — a human approval is still needed. QA reviewer: when you pick it up, swap `review` for `qa`, then run the routes below.

### P1 — Payments section on a sales document

**Where to click**
- `/backend/sales/documents/<id>` for an **order that already has at least one recorded payment** → the **Payments** section/tab.
- The same page for an **order with no payments** (the empty state).

**What to verify**
- The payments table now shows an extra **"Gateway status"** column alongside the existing columns. Expect the cell value to read `pending` on every row — that is the widget's documented fallback, because nothing populates `_gateway.unifiedStatus` yet (see the PR body's "Known follow-up"). A blank cell or the raw key `payment_gateways.column.gatewayStatus` as the header would both be failures.
- Switch the UI language (pl / de / es / ko) and re-open the section — the header must be translated, not shown as a key.
- Click a payment row → the edit dialog still opens with the correct payment prefilled.
- The row kebab menu still offers **Edit** and **Delete**, and both still work; deleting recalculates the document totals as before.
- Add a payment via **Add payment** → the table refreshes and the new row also carries the Gateway status column.
- The empty state (order with no payments) is unchanged — no table, no stray column, the "Add payment" call to action still works.

**What can go wrong**
- The new column pushes the table into horizontal overflow on a narrow/laptop viewport, or squeezes the amount/date columns — check at ~1280px and on a mobile viewport.
- Row-click-to-edit stops working because the injected column intercepts the click.
- The column appears on the wrong table (orders or quotes list at `/backend/sales/documents`) — it must appear **only** in the payments section of a document.
- Cross-tenant/organization check: switch the active organization scope and re-open a document's payments — no payment rows from another scope may appear, and the column must behave identically.
- Perceived performance: cold-load the document detail page and open the Payments tab. The section must not visibly stall or re-flash its loading state more than before; this change does not add a network call, so any new delay is a regression worth reporting with a screenshot.

### P2 — Other data tables unaffected

**Where to click**
- `/backend/sales/documents` (orders and quotes lists).
- Any other list with injected columns you already trust, e.g. `/backend/catalog/products`.

**What to verify**
- No Gateway status column leaks into these tables; their columns, filters, saved views, and export behave exactly as before.

**What can go wrong**
- A saved view / perspective on the orders or products list breaks or resets — this PR does not wire perspectives into the payments table, so any change there is unexpected.

### Pass/fail
- All routes pass → remove `qa`, add `qa-approved` (and `merge-queue` once a human approval is in).
- Any route fails → remove `qa`, add `qa-failed`, and comment describing the failure.

adeptofvoltron:
🤖 `om-auto-review-pr` — completed: no blockers, majors or mediums; report posted above. Lock released (swapped to `ci-monitoring` while the CI follow-up runs). Chain continues with the UI-verification step.

adeptofvoltron:
🤖 `om-auto-review-pr` — CI result: **all checks pass** on `a47b5b4` — `test` (6m49s), `ephemeral-integration`, `lint`, `ds-lint`, `prepare`, `docker-build` (4m9s), `scope`, `audit-scope`, `CodeQL` / `Analyze` (actions, javascript-typescript, python), `license/cla`. Mutation tests green on the branch too.

`mergeStateStatus` is `BLOCKED` only on the human gates: a review approval (an automated run cannot approve its own PR) and the QA-approval gate — this PR carries `needs-qa`, so it must not merge until `qa-approved` is added. Manual QA instructions are in the comment above. `ci-monitoring` removed; no further follow-up will come from this agent.

adeptofvoltron:
🤖 `om-auto-qa-pr` started by @adeptofvoltron at 2026-08-18T13:14:51Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

adeptofvoltron:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ PASS — the `Gateway status` column now renders in the sales document payments table, and a baseline run of the same worktree at the PR's base proves it did not render before this change.
**Environment:** `http://localhost:3007` · role `admin` · browser `playwright` (Chromium 1.62.1)
**Verified:** `fix/issue-5142-payments-table-id` @ `a47b5b4`, with a baseline control at `develop` @ `3adb0d0d`

### Scenario (P1 — sales document detail → Payments tab)

**Where to click:** `/backend/sales/orders/{orderId}` → **Payments** tab. The order driven was the seeded `SO-DEMO-2001`, which carries two payments (`WIRE-2001` / received and `AUTH-2001-CC` / captured), so the table renders rows rather than the empty state.

| 0 | **Baseline control:** boot the same worktree at the PR's base (`develop` @ `3adb0d0d`, full rebuild) and read the payments-table headers on the same order | No `Gateway status` column — the widget is registered but unbound | Headers were `Reference, Method, Status, Amount, Received at, Created, ⋯` and the `Gateway status` header count was **0**, with the same 2 payment rows present | ✅ |
| 1 | Sign in to the backoffice as the `admin` role | The authenticated backoffice shell renders | Login succeeded and the dashboard rendered | ✅ |
| 2 | Open the sales order `SO-DEMO-2001` | Order detail renders with a Payments tab | Order detail rendered; the Payments tab was present | ✅ |
| 3 | Open the **Payments** tab and read the payments `DataTable` header row | The injected `Gateway status` column binds through the new `sales.payments` table id and renders | Headers were `Reference, Method, Status, Amount, Received at, Created, ⋯, Gateway status` — exactly one `Gateway status` header | ✅ |
| 4 | Read the `Gateway status` cell of every payment row | A cell renders per row; with no gateway enricher shipping `_gateway.unifiedStatus`, the widget's documented `'pending'` fallback is expected | Both rows rendered a non-empty cell reading `pending` | ✅ |
| 5 | Smoke the payments section at a 390×844 mobile viewport | The section stays usable and the layout does not break | The layout held. The payments table clips to `Reference / Method / Status` and the appended `Gateway status` column sits outside the visible width; whether it is reachable by horizontal scroll at that width was **not exercised** | ⚠️ not exercised |

### Screenshots

The before/after pair below is the core evidence: the same order, the same two payment rows, the same disposable database — the only variable is the checked-out commit.


**❌ Before — `develop` @ `3adb0d0d`: no `Gateway status` column**

baseline develop — no gateway status column

**✅ After — this PR @ `a47b5b4`: the `Gateway status` column renders**

payments table with gateway status column

**Full page — sales order `SO-DEMO-2001`, Payments tab**

sales document detail, payments tab

**Mobile viewport (390×844) — layout holds, the appended column is off the visible width**

payments section at mobile viewport

### Notes for QA

- **The fix does what it claims.** The baseline control is what makes this more than "a column is present": the identical assertion run against `develop` found zero `Gateway status` headers, and against the PR head found exactly one. The binding really was dead before and really does resolve now.
- **The column ships with no data behind it.** Both cells read `pending`, which is the widget's own hardcoded fallback — nothing produces `_gateway.unifiedStatus` today, exactly as the PR body flags under "Known follow-up". So what a user sees in the new column is a constant, not a gateway state. That is a product decision for the reviewer (ship the empty column now, or hold it until the response enricher from SPEC-044 lands), not a defect in this change.
- **Mobile is unresolved, and it is pre-existing.** At 390px the payments table already overflowed with its six built-in columns; this change appends a seventh, so the new column is the one pushed furthest off-screen. I did not verify horizontal reachability, so treat this as unverified rather than as either a pass or a regression.
- **Environment caveat, disclosed in full.** This machine has no container runtime, so the documented ephemeral testcontainers environment could not be used. The app was booted in dev mode against a pre-existing disposable database (`om_test_env_qa`) on the local Postgres; the developer's own `open-mercato` database was never opened, migrated, seeded or dropped, and no accounts were seeded or reset for this run.
- **One boot-ordering artifact, not a PR finding.** The `develop` baseline refused to compile until dependencies were reinstalled, because `develop` carries workspace packages (`channel-expo` / `channel-fcm` / `channel-apns`) that the PR head's base predates. That is an artifact of switching commits inside one worktree during this QA run and says nothing about the PR.

adeptofvoltron:

This change ships jest/RTL coverage (`PaymentsSection.gatewayColumn.test.tsx` renders the section with the real `DataTable` and the real widget module) but no browser-level test, so the QA above was manual. The regression this PR fixes was specifically an *end-to-end binding* failure — the widget was registered, the binding key was gone, and every unit test still passed while the column rendered nowhere — which is exactly the class of bug a component-level test can miss and a browser test cannot. Worth locking in with `/om-integration-tests`.

1. **Setup:** via the API fixtures, create a sales order and at least one payment on it (the manual run used the seeded `SO-DEMO-2001` with `WIRE-2001` / received and `AUTH-2001-CC` / captured; a self-contained test should create its own order + payment rather than depend on demo data).
2. **Act:** sign in as an admin-role user, open `/backend/sales/orders/{orderId}`, and click the **Payments** tab.
3. **Assert:** the payments table's header row contains exactly one `Gateway status` header, and every payment row renders a non-empty cell in that column. Asserting the *header count* rather than mere presence is what catches a re-drift between `extensionPoints.hosts.paymentsTable.tableId` and the `'data-table:sales.payments:columns'` binding key — spot resolution is exact-match, so a drift unbinds silently rather than failing loudly.
4. **Teardown:** delete every fixture the test created (payment, then order).

One caveat for whoever writes it: until a gateway module ships the response enricher that produces `_gateway.unifiedStatus` (SPEC-044, flagged as a known follow-up in the PR body), the cell value is the widget's constant `'pending'` fallback. Assert that a cell *renders* and is non-empty rather than pinning the literal `'pending'`, so the test does not have to be rewritten the day the enricher lands.

adeptofvoltron:

Run mode was the default `--evidence-only`, so no pipeline or meta labels were changed: this PR still carries `needs-qa` without `qa-approved`, and the QA-approval merge gate still holds it. A QA reviewer (or a maintainer applying `qa-approved`) makes that call on the evidence above — this run deliberately did not sign itself off.

adeptofvoltron:
🏷️ Label rationale (`om-auto-qa-pr`):

- **`screenshots`** — UI QA visual evidence was captured in a real browser (Chromium) and attached to this PR as an inline-rendering comment: a before/after pair from the same order and the same two payment rows, plus the full page and a mobile viewport. Per `.ai/docs/pr-workflow.md`, this label is informational only — it does not gate merge and is orthogonal to `needs-qa`/`qa-approved`. It was missing from my first pass because I applied the skill's `--evidence-only` default ("change no pipeline or meta labels") literally, without honoring the repo-specific rule that this particular meta label is exactly the one `om-auto-qa-pr` is meant to post. Corrected now.

Unchanged and deliberately so: **`needs-qa`** stays, and no **`qa-approved`** / **`qa-self-verified`** / **`qa-failed`** was applied. This run was evidence-only — the QA verdict is a human's call on the evidence above, and the QA-approval merge gate still holds this PR.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-21T12:35:34Z. Other auto-skills will skip this PR until the lock is released.

Claiming additively over the author's self-assignment: the newest 🤖 comment on this PR is `om-auto-qa-pr` — completed: **PASS**. Lock released. (2026-08-18T13:36Z), there is no `in-progress` label, and no claim signal falls inside the 30-minute window. @adeptofvoltron stays assigned.

pkarw (changes requested):
# 🔍 Code Review

## 🎯 Summary

This PR does exactly what #5142 asked for, and the plumbing is right. `PaymentsSection.tsx` now passes `extensionTableId={extensionPoints.hosts.paymentsTable.tableId}`, `sales.payments` is declared as a real `dataTableExtensionHost`, the `'data-table:sales.payments:columns'` binding is restored where the `NOTE` used to be, and two new tests pin the host id and the binding key to each other so they cannot drift apart again. I verified the mechanism rather than taking the description: `DataTable.tsx:1468-1473` resolves `perspective?.tableId → extensionTableId → injectionSpotId`, so supplying the middle one is enough to lift the columns spot off the `'__disabled__:columns'` sentinel at `DataTable.tsx:1516`. All four acceptance boxes on #5142 are genuinely ticked.

The problem is not the plumbing — it is what comes through it. The PR body asks the reviewer to decide "whether the column should ship before its data source exists", and @adeptofvoltron's own UI QA pass repeated the question ("That is a product decision for the reviewer"). Answering it is the substance of this review, so here is the answer: **not in this shape.**

Nothing in the tree produces `_gateway.unifiedStatus` — I grepped for both `_gateway` and `unifiedStatus` across the repo and the only `unifiedStatus` producers belong to `shipping_carriers`; `payment_gateways/data/enrichers.ts` is an empty array by deliberate design. So `getValue()` is `undefined` on every row, forever, and the widget's fallback turns that into the literal string `pending`. The QA evidence on this PR shows the consequence directly: payments `WIRE-2001` (**received**) and `AUTH-2001-CC` (**captured**) both render `Gateway status: pending`. That is not an empty column — it is a financial table making a specific, false, unqualified claim about money that has already been captured, for every tenant, in every locale.

That is what takes this to a Major. The dead widget was harmless; a visible wrong answer is not, and it is this change that converts one into the other. The fix is small and belongs in this PR — it is roughly three lines in `widget.ts`, plus one assertion in the new test.

I also checked this PR's mechanism against a `develop` that has moved eight days since CI last ran, and it holds; details in **Verified and dismissed** below, because several plausible-looking concerns turned out to be non-issues and are worth recording as such rather than leaving implied.

## Verdict

**CHANGES REQUESTED** — 1 major, 3 minors, 2 nits, no blockers.

The major is that this PR makes a column visible that reports a hardcoded constant as gateway state. Under this repo's verdict rule the minors would independently require a change pass; all four items are small and land in the same file.

## 🧪 Validation Gate

**Validation source: GitHub checks** — every command in `validation.commands` maps onto a green step of the checks on this exact head (`a47b5b4`), so no local gate was run and none stands behind this review.

| `validation.commands` entry | CI job → step | Result |
|---|---|---|
| `yarn build:packages` | `prepare` → *Build packages* | ✅ (3m24s) |
| `yarn generate` | `prepare` → *Prepare generated modules* | ✅ |
| `yarn build:packages` (rebuild) | `prepare` → *Build packages* | ✅ |
| `yarn i18n:check-sync` | `test` → *Check i18n sync* | ✅ |
| `yarn i18n:check-usage` | `test` → *Check i18n usage* | ✅ |
| `yarn typecheck` | `test` → *Checking types* (affected-package scoped) | ✅ |
| `yarn test` | `test` → *Test* (affected-package scoped) | ✅ (6m49s) |
| `yarn build:app` | `prepare` → *Build app* | ✅ |

Beyond the gate: `lint`, `ds-lint`, `docker-build` (4m9s), `ephemeral-integration` (11m18s), `audit-scope`, `scope`, CodeQL ×3 and `license/cla` are all green. Nothing is pending and nothing is failing, so there are no CI-derived findings and no CI-pending disclosure to make.

**One caveat, disclosed rather than glossed:** those checks ran on 2026-08-13, against a `develop` that has since advanced. Rather than assume the signal still holds, I diffed `develop` across every surface this PR depends on. `packages/ui/src/backend/DataTable.tsx` did change (+20/−9), but only `showQueryTime`, the header flex classes, and widening `onRowClick` to `(row, event)` — the `extensionTableId` memo, `resolvedInjectionSpotId`, the `useInjectionDataWidgets` spot derivation and `mergedColumns` are all byte-for-byte unchanged, so the mechanism this PR relies on is untouched. (`PaymentsSection` passes `onRowClick={openEditPayment}`, which ignores the new second argument — no break.) `module-facts.ts` changed substantially, but only its markdown rendering gained an empty-section guard; `hostTokens.tableIds` extraction is unchanged. And no new `dataTableExtensionHost` landed in `catalog` or `sales` on `develop`, so the updated fixture assertion `['sales.orders', 'sales.payments', 'sales.quotes']` still holds after a merge-forward. The green signal is stale in age but not in meaning.

## 🔴 Findings

### Major

**1. The restored binding ships a column that reports a hardcoded constant as gateway state**
`packages/core/src/modules/sales/widgets/injection-table.ts:20`

Restoring the binding activates `sales.injection.payment-gateway-status-column`, whose cell is:

```ts
cell: ({ getValue }) => {
  const value = getValue()
  return typeof value === 'string' && value.trim().length > 0 ? value : 'pending'
}
```

`accessorKey` is `_gateway.unifiedStatus`, and nothing in the repository produces it. `packages/core/src/modules/sales/api/payments/route.ts` does not opt into `enrichers`; `packages/core/src/modules/payment_gateways/data/enrichers.ts` is `[]` with a comment saying feature modules must declare their own bindings; and `PaymentsSection`'s `PaymentRow` mapping keeps only known keys. A repo-wide grep for `_gateway` returns no producer at all. So the ternary takes its right branch on every row of every payments table, and users see the string `pending`.

The QA evidence on this PR is the proof, not a hypothetical: `WIRE-2001` is **received** and `AUTH-2001-CC` is **captured**, and both render `Gateway status: pending`. In a payments table, next to a `Status` column reading *received* / *captured*, that reads as "the gateway has not settled this yet" — a claim the system has no data to make. It is worse than an absent column, and it ships to every tenant.

Three things make this a Major rather than a disclosed trade-off:

- **SPEC-044 §13.3, the spec this widget descends from, does the opposite.** Its cell body is `if (!status) return null` — render nothing when there is no gateway transaction. The `'pending'` fallback is drift from the spec, not an intended behavior, and this PR is the first change that makes the drift reachable.
- **Every sibling column in this exact table already has the right convention.** `PaymentsSection.tsx:276-300` renders `'—'` for missing method, status, received-at and created-at values. The injected column is the only cell in the table that invents a value instead of admitting it has none.
- **`'pending'` is a hardcoded user-facing string**, so it renders untranslated as `pending` in the pl / de / es / ko UIs — against the root `AGENTS.md` rule that user-facing strings never get hardcoded. The header is fine (`t(definition.header, …)` resolves `payment_gateways.column.gatewayStatus` in all five locales); only the cell body bypasses i18n.

**Suggested fix** — in `packages/core/src/modules/sales/widgets/injection/payment-gateway-status-column/widget.ts`, drop the invented value:

```ts
cell: ({ getValue }) => {
  const value = getValue()
  return typeof value === 'string' && value.trim().length > 0 ? value : '—'
}
```

`'—'` matches what the four sibling columns already do for missing data and needs no locale key; `return null` (the spec's own choice) is equally acceptable if an empty cell reads better. Either way the column becomes an honest, extensible surface that third-party gateway modules can populate, which is what #5142 was actually after.

**The alternative, if you would rather not touch the widget in this PR:** land the host and the tests but leave the binding out until the SPEC-044 Phase 3 enricher exists. That also resolves this finding — it just defers three of the four acceptance items. I would take the fix above; it is smaller, it closes #5142 completely, and it leaves nothing to remember later.

### Minor

**2. The activated widget has no ACL gate, so it renders for users without `payment_gateways.view`**
`packages/core/src/modules/sales/widgets/injection/payment-gateway-status-column/widget.ts:4-7`

`useInjectionDataWidgets.ts:57-60` filters loaded widgets with `features.length === 0 || hasAllFeatures(grantedFeatureList, features)` — an empty `features` array means *show to everyone*. This widget's `metadata` carries only `id` and `priority`, so from this PR onward a `payment_gateways`-namespaced column appears for every user who can open a sales document, whether or not they hold `payment_gateways.view`.

SPEC-044 §13.3 specifies `features: ['payment_gateways.view']` on this exact widget. It discloses nothing today (the value is a constant, per finding 1), and once the Phase 3 enricher lands it will carry its own `features: ['payment_gateways.view']` per §13.2 — so this is a correctness and consistency gap rather than a data leak. But the gate is one line, it is what the spec asked for, and adding it now means the column simply does not appear for users who have no business seeing gateway state, instead of appearing empty for them.

**3. The column renders to the right of the row-actions kebab**
`packages/core/src/modules/sales/widgets/injection/payment-gateway-status-column/widget.ts:8-16`

The widget declares no `placement`, and `insertByInjectionPlacement` (`injection-position.ts:47-51`) falls through to `current.push(item)` when there is no `relativeTo`. `PaymentsSection`'s `columns` memo ends with `{ id: 'actions', header: '' }` at `PaymentsSection.tsx:307`, so the injected column is appended *after* it. QA observed exactly that header order: `Reference, Method, Status, Amount, Received at, Created, ⋯, Gateway status`. A data column sitting to the right of the actions column is a layout defect on its own, and it is also why the QA pass found this column pushed furthest off-screen at 390px.

SPEC-044 §13.3 specifies `placement: { position: InjectionPosition.After, relativeTo: 'status' }` and `size: 120`, which puts gateway status beside the payment status it qualifies — the position that actually makes the column readable. Worth adopting both.

**4. A `sales`-owned widget depends on the `payment_gateways` dictionary without declaring the dependency**
`packages/core/src/modules/sales/widgets/injection/payment-gateway-status-column/widget.ts:11`

The header key `payment_gateways.column.gatewayStatus` exists only in `packages/core/src/modules/payment_gateways/i18n/*.json`. Dictionaries are merged per *registered* module (`packages/shared/src/lib/i18n/server.ts:70-71` iterates `modules` and merges `m.translations?.[locale]`), and `sales`' manifest declares `requires: ['catalog', 'customers', 'dictionaries']` — `payment_gateways` is not among them. In a deployment that leaves `payment_gateways` out of `modules.ts`, this column's header degrades to the raw key `payment_gateways.column.gatewayStatus`.

This is latent rather than live: both `apps/mercato/src/modules.ts:116` and the create-app template at `:113` register `payment_gateways` by default, which is why QA saw a correct "Gateway status". Note that `sales` owning the widget is itself the *right* call — `payment_gateways/data/enrichers.ts` explicitly says feature modules should declare their own bindings and consume `payment_gateways` as a dependency. It is the second half of that sentence that is missing. Either move the key into `sales`' own dictionary (simplest, and it makes the widget self-contained), or resolve the peer soft-optionally per the "Optional integration is soft-optional" rule in `.ai/review-checklist.md` § 1.

### Nit

**5. The new test pins the constant this review asks you to remove**
`packages/core/src/modules/sales/components/documents/__tests__/PaymentsSection.gatewayColumn.test.tsx:123`

`expect(screen.getByText('pending')).toBeTruthy()` is well-motivated — its comment correctly explains that asserting the cell body, not just the header, is what proves the column actually renders. But it hardcodes the specific fallback, so fixing finding 1 requires editing this line too. Asserting on the rendered cell for the `gateway_status` column id (whatever its text) would keep the discriminating power without freezing the value. Mentioning it so the two changes travel together rather than surprising you.

**6. The in-tree widget has drifted from SPEC-044 §13.3 in more ways than this PR needs to fix**
`packages/core/src/modules/sales/widgets/injection/payment-gateway-status-column/widget.ts`

Recording the full picture so it is a decision rather than an oversight: the spec places this widget in `payment_gateways` as `payment_gateways.injection.gateway-status-column`, renders a `<Badge>` with per-status variants, and sets `size: 120`. In the tree it lives in `sales` as `sales.injection.payment-gateway-status-column` and renders bare text. The module/id difference is settled and fine (the enrichers comment endorses feature-module ownership, and the id is a frozen contract surface now). The badge is a judgement call — the neighbouring `Status` column at `PaymentsSection.tsx:280` also renders plain text, so bare text is locally consistent and I am not asking for a badge. Findings 2 and 3 are the two pieces of that drift I do think are worth closing; the rest can stay.

## ✅ Verified and dismissed

Recording these so the next pass does not re-derive them, and so the ones that looked like findings are visibly closed rather than silently dropped:

- **Docs are already correct.** I expected a documentation gap for a new public extension surface, but `apps/docs/docs/framework/modules/sales/injection.mdx:37` already lists `data-table:sales.payments:columns` under a **Payments** heading. The docs have been promising this spot all along while it was dead; this PR makes the published catalog true rather than stale. No docs change needed.
- **Activating the host enables nothing unintended.** A repo-wide grep for `sales.payments` finds no other binding against `data-table:sales.payments:*`. Setting `extensionTableId` does make `resolvedInjectionSpotId` non-null, so the `:header` / `:toolbar` / `:footer` / `:search-trailing` / `:row-actions` / `:bulk-actions` / `:filters` spots all become live — but nothing binds any of them today, so the blast radius is exactly the one intended column.
- **The host↔component wiring is machine-checked, not just declared.** `hostTokens.tableIds` is extracted by scanning component sources, not `extension-points.ts`, and the extractor resolves the identifier chain `extensionPoints.hosts.paymentsTable.tableId` back to its literal. So the updated assertion in `module-facts.extension-hosts.test.ts` genuinely fails if the host is declared without the component passing it — a real regression guard, and it runs inside the green `test` job.
- **No backward-compatibility violation.** `data-table:sales.payments:*` is a brand-new spot-id surface, which `BACKWARD_COMPATIBILITY.md` classifies as additive. No existing spot id, event id, API route, response shape or DB shape changes. The removed `NOTE` was a comment.
- **`extensionTableId` over `perspective={{ tableId }}` is the right call**, and the PR body's reasoning checks out against `DataTable.tsx:347-355`: it is documented as the cheaper path when a host needs only widget injection, and it correctly keeps this document-detail sub-table out of the saved-views machinery it has no use for. The import style matches the existing convention at `SalesDocumentsTable.tsx:4`.

## 💥 Breaking Changes

- **Exported APIs / types:** none — `extensionPoints.hosts` gains a key; additive.
- **HTTP routes / response shapes:** unchanged.
- **Event IDs / widget spot IDs:** `data-table:sales.payments:*` is new. Additive, and already documented.
- **DB schema / migrations:** none.
- **DI keys / ACL features / CLI commands / generated files:** unchanged. No committed generated artifact enumerates extension hosts, so `yarn generate` produces no diff to commit.

No deprecation bridge is required.

## 🧪 Test Coverage

Good, and the shape is right. `injection-table.test.ts` derives the columns spot id from the host via `DataTableInjectionSpots.columns(...)` and asserts the binding key matches exactly — pinning the two halves of the original bug together, which is precisely what was missing before. `PaymentsSection.gatewayColumn.test.tsx` renders the real `DataTable` with the real widget module and asserts both the header and a cell, and separately asserts that the table asks for `'data-table:sales.payments:columns'`. Both assertions are discriminating: remove `extensionTableId` and both tests fail. The author's claim that they fail on `develop` (3 failed) is consistent with what the code does.

Two observations, neither of them a request beyond nit 5:

- The suite mocks `useInjectionDataWidgets` wholesale, so it verifies *which spot id `DataTable` asks for* rather than end-to-end resolution through the real registry. That is a reasonable boundary for a jest/RTL test, and the second test is what makes the boundary safe — it is the assertion that would catch a table-id regression. The `injection-table.test.ts` half covers the registry side. Together they cover the seam.
- No test covers the ACL gate (finding 2) or the column's position (finding 3), which is expected since neither behavior exists yet. If you take those fixes, a `features`-filtering case and a column-order assertion would be the natural additions — the existing test file already has the rendering harness for both.

---

**Labels:** `review` → `changes-requested`. `bug`, `needs-qa`, `screenshots` and `priority-low` are kept and still fit. I moved `risk-low` → `risk-medium`, and I want to be explicit about why rather than churn a label quietly: `risk-low` reads "isolated, unlikely to cause a regression", and as the PR stands it puts a visibly wrong value in front of every tenant on a financial screen — that reach is the finding I am blocking on. It is shallow (display-only, no data mutation, trivially revertible), which is why it is medium and not high. Fix finding 1 and `risk-low` becomes accurate again; say so on the next push and I will move it back.

**Autofix:** skipped (not my PR — re-run with `--autofix` to fix it here).

pkarw:

*Updated on the autofix pass at `01ec018` (the previous rationale, written for the `changes-requested` verdict on `a47b5b4`, is superseded).*

- ✅ `merge-queue` — **changed from `changes-requested`**. The re-review approved: the major and all three minors from the previous pass are implemented and regression-tested. The merge is still gated — see `needs-qa` below and the pending required checks on this head.
- 🟢 `risk-low` — **changed back from `risk-medium`**, as the previous rationale promised it would be. `risk-medium` was applied for exactly one reason: the PR put a visibly wrong value (`pending`) on a financial screen for every tenant. That value is gone; what remains is display-only, additive, single-module and trivially revertible, which is what `risk-low` describes.
- 🧪 `needs-qa` — unchanged, and it now matters more than before. The fix changed two user-visible behaviours on top of the original change: the column renders `—` instead of `pending`, and it is hidden from users without `payment_gateways.view`. It also moved, from past the row-actions kebab to immediately after the payment Status column. The earlier UI QA pass predates all of that, so it needs a fresh run before `qa-approved`.
- 📸 `screenshots` — unchanged, but read the attached evidence as **historical**: it shows the pre-fix column (last position, reading `pending` on a received and a captured payment). It is still the clearest illustration of what was wrong; it is not evidence about the current head.
- 🐛 `bug` — unchanged; this fixes #5142, a registered-but-unbound injection widget.
- 🔽 `priority-low` — unchanged; matches the linked issue's own rating.

Not applied: `qa-approved` / `qa-self-verified` — no manual QA ran on this head, so the QA-approval merge gate stands. `qa` is untouched by design; it belongs to a QA reviewer picking the PR up.

pkarw:

To save you a round trip: you explicitly asked the reviewer to decide whether the column should ship before its data source exists, and the answer is that the plumbing is right but the widget’s `'pending'` fallback should not reach users. All four items are in the same file (`widgets/injection/payment-gateway-status-column/widget.ts`) and together come to about five lines — drop the invented fallback in favour of `'—'` (matching the four sibling columns in that same table), add `features: ['payment_gateways.view']`, and adopt SPEC-044 §13.3’s `placement: { position: InjectionPosition.After, relativeTo: 'status' }` plus `size: 120`. The fourth is the missing `payment_gateways` dependency behind the header key. Nit 5 notes that the new test pins `'pending'` by name, so that assertion travels with the fix.

If you would rather leave the widget untouched, the alternative stated in the review also closes the major: land the host and the tests now and hold the binding until the SPEC-044 Phase 3 enricher lands.

pkarw:

1 major, 3 minors, 2 nits, no blockers. The major is that restoring the binding activates a column whose cell renders the hardcoded constant `pending` for every payment — nothing in the tree produces `_gateway.unifiedStatus`, and this PR’s own QA evidence shows a **received** and a **captured** payment both reporting `Gateway status: pending`. The host, the binding, and the two new tests are all correct and I verified the mechanism at `DataTable.tsx:1468-1473` / `:1516` rather than from the description.

Validation source: **GitHub checks** — all 8 `validation.commands` entries map onto green steps of the checks on `a47b5b4` (`prepare` and `test`), so no local gate was run. Those checks are from 2026-08-13, so I hand-diffed `develop` across every surface this PR depends on: `DataTable`’s extension resolution and `module-facts`’ `hostTokens` extraction are unchanged, and no new table host landed in `catalog` or `sales`, so the updated fixture assertion survives a merge-forward.

No `ci-monitoring`: CI is fully green and settled on this head, so there is no follow-up owed. Labels `review` → `changes-requested` and `risk-low` → `risk-medium` (rationale in the 🏷️ comment); `bug`, `needs-qa`, `screenshots`, `priority-low` kept. Reassigned to @adeptofvoltron.

pkarw:
🤖 `om-auto-review-pr --autofix` started by @pkarw at 2026-08-22T06:33:56Z. Other auto-skills will skip this PR until the lock is released.

Claiming additively over the author's self-assignment: that assignment is the handoff from my own `changes-requested` review on 2026-08-21, not a live lock — there was no `in-progress` label, and the newest 🤖 comment on this PR is my own `Lock released` from 2026-08-21T12:44Z, far outside the 30-minute window. @adeptofvoltron stays assigned. This run was invoked with `--autofix`, so it will implement the outstanding review findings on the PR branch rather than hand back again.

pkarw (approved):
# 🔍 Code Review — Re-review (autofix pass)

## 🎯 Summary

Re-review of #5270 at `01ec018`, after this run implemented the outstanding findings itself (`--autofix`). Two commits landed since the previous review: a merge of the current `develop` — the branch was nine days stale, and `develop` has since moved to `@tanstack/react-table ^9`, so the branch was being reviewed against a DataTable that no longer exists — and one fix commit closing the major and all three minors.

The original change is unchanged and was right all along: `sales.payments` is a declared `dataTableExtensionHost`, `PaymentsSection` passes it as `extensionTableId`, the `'data-table:sales.payments:columns'` binding is restored, and the two tests pin the host id and the binding key to each other. What the previous review blocked on was not the plumbing but what came through it — a column reporting the hardcoded constant `pending` for every payment, including ones already received or captured. That is now fixed, together with the ACL gate, the column position and the undeclared module dependency.

**Full disclosure, because it changes how much weight this approval carries:** the fix commit is mine, not @adeptofvoltron's. This review therefore approves the author's original change plus my own follow-up. A human should skim `01ec018` before merging — and the QA gate below holds the merge independently.

## Verdict

**APPROVED** — 0 blockers, 0 majors, 0 minors. Nit 6 from the previous review is explicitly declined below rather than silently dropped.

## 🧪 Validation Gate

**Validation source: local fallback** (`yarn generate`, `yarn build:packages`, `yarn turbo run typecheck --filter=@open-mercato/core --filter=@open-mercato/ui`, `yarn test` scoped to the changed packages). GitHub checks for `01ec018` are all **QUEUED** at the time of writing, so there was no CI signal for this head to prefer — and since this run wrote the code, it owed local evidence anyway.

| Command | Scope | Result |
|---|---|---|
| `yarn build:packages` | all packages | ✅ |
| `yarn generate` | all modules | ✅ |
| `yarn turbo run typecheck` | `@open-mercato/core`, `@open-mercato/ui` | ✅ |
| jest | `@open-mercato/core` → `src/modules/sales` | ✅ 89 suites, 632 tests |
| jest | `@open-mercato/ui` (full) | ✅ 232 suites, 1904 tests |
| jest | `@open-mercato/shared` → `src/modules/widgets` | ✅ 5 suites, 24 tests |
| jest | `@open-mercato/cli` → `module-facts.extension-hosts` | ✅ 3 tests |

`yarn i18n:check-sync` / `i18n:check-usage` / `yarn build:app` were left to CI: no locale file is touched by this diff, and the merge of `develop` brings 1 920 files whose gate CI already ran on `develop` itself. **CI is pending on `01ec018`** — the merge and the fix commit both need a green run, and required checks still gate the merge regardless of this approval.

## ✅ Findings from the previous review — all closed

**1 (major) — the column reported a hardcoded constant as gateway state.** Fixed in `widget.ts`: the fallback is now `'—'`, the same placeholder the four sibling columns in this table already render for missing method / status / received-at / created-at values. This also removes an untranslated user-facing string from the cell body. Regression test: *"shows the missing-value placeholder instead of inventing a gateway status"* asserts the gateway cell renders exactly what the Method column renders for the same (empty) row, and explicitly not `pending`. Verified failing when the `'pending'` fallback is restored.

**2 (minor) — no ACL gate.** Fixed: `features: ['payment_gateways.view']`, as SPEC-044 §13.3 specifies. `useInjectionDataWidgets` treats an empty `features` array as "show to everyone", so the declaration is the whole gate. Tested through the real `hasAllFeatures` helper rather than by asserting the array literal back — a user holding only `sales.view` is refused, and both `payment_gateways.view` and the wildcard `payment_gateways.*` pass. The feature is already granted to `admin` (`payment_gateways.*`) and `employee` (`payment_gateways.view`) in that module's `defaultRoleFeatures`, so no existing role loses the column.

**3 (minor) — the column rendered to the right of the row-actions kebab.** Fixed with `placement: { position: InjectionPosition.After, relativeTo: 'status' }` and `size: 120`.

One thing worth flagging, because it is the non-obvious half of this fix and a trap for the next injected column: **`relativeTo` alone would not have worked.** `mergedColumns` matches candidates with `(col) => (col as { id?: string }).id ?? ''`, and it runs on the raw `ColumnDef[]` *before* the table is built — so an accessor-only column has no `id` yet, `relativeTo: 'status'` finds no target, and `insertByInjectionPlacement` falls through to `current.push(item)`, appending past the actions column exactly as before. The payments status column therefore now declares `id: 'status'` explicitly. The column-order test asserts the gateway column sits at `statusIndex + 1` and is not last; it fails (index 7 instead of 3) the moment that `id` is removed.

**4 (minor) — a `sales`-owned widget depending on the `payment_gateways` dictionary without declaring it.** Fixed with `requiredModules: ['payment_gateways']` rather than by copying the translation key into `sales`. That is the stronger answer to the same problem: the loader skips the widget entirely when the module is absent (`applyRequiredModuleGate`, reached from `loadInjectionDataWidgetById` on the data-widget path too), so the header can no longer degrade to a raw key — and the column stops appearing in deployments where nothing could ever populate it. I checked the one way this could have backfired: `getEnabledModuleIdsForInjection` falls back to deriving the enabled set from registered injection tables and widget entries when `registerEnabledModuleIds` was never called, which would misclassify a dependency module that ships no widgets. `payment_gateways` ships its own `widgets/injection-table.ts`, so it is visible on the fallback path as well as through `ClientBootstrap`'s explicit registration. No false negative.

**5 (nit) — the test pinned the constant the fix removes.** Fixed: the suite now reads the gateway cell by column position (`thead th` index → matching `tbody td`) instead of by text, so it keeps its discriminating power without freezing the value. The suite also asserts header and body cell counts match, which is what makes the index mapping sound.

**6 (nit) — broader SPEC-044 §13.3 drift (`<Badge>` rendering, widget owned by `payment_gateways` under a different id).** **Declined, deliberately.** The previous review already said it was not asking for the badge: the neighbouring Status column renders plain text, so bare text is locally consistent. The module/id difference is settled — `payment_gateways/data/enrichers.ts` explicitly endorses feature-module ownership, and `sales.injection.payment-gateway-status-column` is a frozen contract surface now that it resolves. Nothing to do.

## 🧪 Test Coverage

Five cases now, up from two, and each was verified to fail with its own fix reverted rather than assumed to be discriminating:

- host id ↔ binding key pinned together (unchanged from the original PR)
- the columns spot the table actually asks for (unchanged)
- the gateway cell renders, read positionally
- the cell shows the table's missing-value placeholder, not an invented status
- the column sits immediately after Status and before the row actions
- the widget's feature gate refuses a user without `payment_gateways.view` and accepts the wildcard grant

Still uncovered, and correctly so: end-to-end resolution through the real injection registry (the section suite mocks `useInjectionDataWidgets` by design; `injection-table.test.ts` covers the registry side) and `requiredModules` gating, which is already covered at the loader level in `packages/shared/.../injection-loader.required-modules.test.ts`. @adeptofvoltron's own follow-up comment proposing a browser-level test for this binding still stands and is still worth doing — this class of bug passed every unit test while rendering nothing.

## 💥 Breaking Changes

None, unchanged from the previous review. `data-table:sales.payments:*` remains a new, additive spot-id surface. The fix commit adds widget metadata (`features`, `requiredModules`, `placement`, `size`) and an explicit `id` on an existing internal column definition — no exported API, route, response shape, event id, DB schema, DI key, ACL feature or CLI command changes. `payment_gateways.view` already exists in that module's `acl.ts`; nothing new is declared.

Two behavior changes worth naming for QA, both intended: the column is now hidden from users without `payment_gateways.view`, and it renders `—` instead of `pending`.

## ⚠️ Note for readers of the PR description

The PR body's **"Known follow-up"** section describes the pre-fix behavior — "the column renders the widget's own documented `'pending'` fallback for every row" — and asks the reviewer to decide whether the column should ship before its data source exists. That question is answered and that paragraph is now stale: the column ships, showing `—` until a gateway module enriches `sales.payment`. The SPEC-044 Phase 3 response enricher remains a separate change.

pkarw:
🤖 `om-auto-review-pr --autofix` completed: **APPROVED**. Lock released (swapped to `ci-monitoring` while the CI follow-up runs).

One autofix iteration; every finding from my `changes-requested` review on `a47b5b4` is implemented and regression-tested, and the re-review above is the full report.

**What landed** — two commits:

1. **`21ff42b` merge of `develop`.** The branch was nine days stale and the base had moved under it in a way that mattered: `develop` is now on `@tanstack/react-table ^9`, so the branch was being reviewed against a `DataTable` that no longer exists. Clean merge, no conflicts.
2. **`01ec018` the fix.** `'pending'` → `'—'` (the placeholder the four sibling columns in this table already use); `features: ['payment_gateways.view']`; `requiredModules: ['payment_gateways']`; `placement: { position: After, relativeTo: 'status' }` with `size: 120`; and an explicit `id: 'status'` on the payments status column, without which the placement silently would not have resolved — `mergedColumns` matches on `ColumnDef.id`, which an accessor-only def does not carry before the table is built. Three new test cases, each verified to fail with its own fix reverted.

@adeptofvoltron — you asked the reviewer to decide whether the column should ship before its data source exists. It does, showing `—` until a gateway module enriches `sales.payment`. Since the fix commit is mine rather than yours, please skim `01ec018`; the disclosure is in the review body too. Note the PR description's "Known follow-up" paragraph is now stale — it describes the `pending` behaviour that this commit removed.

**Not merge-ready yet, on two counts:** required checks are still queued on `01ec018` (both commits need a green run — I will follow up here with the result), and this PR carries `needs-qa` without `qa-approved`, so the QA-approval gate holds the merge. The existing UI QA evidence predates the fix and does not cover the new placement, the placeholder or the ACL gate; it needs a fresh pass.

pkarw:
🤖 `om-auto-review-pr` — CI result: **all checks pass** on `01ec018`.

`prepare` (3m45s), `test` (6m45s), `ephemeral-integration` (12m17s), `docker-build` (4m6s), `documents-multi-instance` (55s), `lint` (30s), `ds-lint` (40s), `scope`, `audit-scope`, CodeQL / `Analyze` (actions, javascript-typescript, python), `license/cla`. `audit`, `merge-coverage`, `mutate`, `comment` and `[code]smith` were skipped by their own conditions, as on the previous head. Nothing pending, nothing failing — and this run covers both new commits, so the merge of `develop` and the fix are green together rather than inferred from an older head.

That closes the local-fallback caveat in the review above: the gate I ran locally was scoped to the changed packages, and CI has now run the full one.

`ci-monitoring` removed; no further follow-up will come from this agent.

**`mergeStateStatus` is now `CLEAN` — do not read that as permission to merge.** GitHub's branch protection is satisfied (approved review, required checks green), but this repo's QA-approval gate is policy, not branch protection: this PR carries `needs-qa` without `qa-approved`, so it must not merge until a QA reviewer signs it off. The attached UI QA evidence predates the fix and does not cover any of what changed — the column now renders `—` rather than `pending`, sits immediately after the payment Status column instead of past the row-actions kebab, and is hidden from users without `payment_gateways.view`. A fresh pass is needed on those three points.

pkarw (approved):

PR #2424, opened 2026-06-02
feat: IMAP + Google email integration foundations

haxiorz:
## Summary

Foundations for email integration.
Currently exposed only in CRM Person page > Email tabs. This will get expanded in future PRs. Work done in this PR is just a foundation for future expansion of 2 way email integration usage.

Adds two-way email for individual users, built as provider packages on the existing Communications Hub (`communication_channels`). Each staff user connects their **personal Gmail (OAuth) or any IMAP+SMTP mailbox (app password)**; inbound mail is ingested, threaded, and surfaced in the unified Messages inbox and on the CRM Person "Emails" tab, and replies are sent from the user's own address. Inbound uses 60s polling plus Gmail Pub/Sub push (polling stays as the fallback); outbound and threading are reliability-hardened (HMAC thread token + JWZ + subject/participants matcher, per-message commit, dead-letter, sent-folder dedup, idempotency). Personal mailboxes follow a **strict owner-only privacy model** (no admin bypass in v1). A dedicated Microsoft 365/Outlook connector is intentionally **out of scope** — those mailboxes connect via IMAP+SMTP. All contract surfaces are additive (no removals/renames).

## Changes

- **New provider packages** `@open-mercato/channel-imap` and `@open-mercato/channel-gmail` — channel adapters, credential/OAuth handling (token refresh + rotation), capabilities, health checks, MIME normalize/convert, and SSRF/TLS hardening (DNS-rebinding pin + cleartext-transport policy) for IMAP.
- **Communications Hub implementation** (`communication_channels`) — entities + migrations, command pipeline, polling/push/outbound/import workers, subscribers, typed events, ACL features + setup grants, response enrichers, widgets, and notifications.
- **Inbound reliability & threading** — `poll-tick` scheduler → `poll-channel` worker; Gmail Pub/Sub webhook (fail-closed JWT verification) + `users.watch` register/renew; layered thread-matcher with HMAC thread token, per-message cursor commit, malformed-MIME dead-letter, and sent-folder dedup.
- **Outbound** — message → `deliver-outbound` subscriber/command → adapter send; thread token injected into References + hidden body footer so replies thread even when headers are stripped.
- **CRM Person "Emails" tab** (`customers`) — compose/reply via an in-process DI facade (no raw HTTP self-call), inbound auto-linking by address, and private/shared visibility enforced in both the kysely and MikroORM read filters (strict owner-only, `view_private` inert in v1).
[...]

PR #5030, opened 2026-08-05
fix(checkout): make rate limiting fail-closed on public pay endpoints (#4702)

adeptofvoltron:
Closes #4702

## 🎯 Goal
A rate-limiter outage no longer disables rate limiting on the public checkout payment endpoints. `POST .../submit` and `POST .../verify-password` now reject with `503` while the limiter cannot decide, instead of letting unlimited unmetered requests through to the payment provider.

## 🔍 Problem
All public checkout pay endpoints wrapped their rate-limit guard in a bare `try/catch` that swallowed every failure and continued. If `rateLimiterService` could not produce a decision, rate limiting was completely absent — an attacker could fire unlimited payment-session creations at `POST /api/checkout/pay/[slug]/submit` (card testing, exhausting the merchant's own Stripe rate limit) or brute-force the checkout link password at `POST .../verify-password`.

## 🔍 Root Cause
Two independent layers both failed open, so nothing upstream could tell "allowed" from "could not decide":

1. `RateLimiterService.consume()` caught errors escaping the limiter and returned `disabledResult(config)` — the same shape it returns when rate limiting is switched off by configuration. A degraded limiter was indistinguishable from a disabled one, and the failure was never logged.
2. Each route's `catch {}` block discarded the error with a `// Rate limiting is fail-open` comment, so a throw from `container.resolve` or `consume` also let the request through silently.

## What Changed
- `packages/shared/src/lib/ratelimit/types.ts` — `RateLimitResult` gains an optional `degraded` flag marking a result whose `allowed` value is a fallback rather than a real decision. Absent when rate limiting is off by configuration, so the two cases are now distinguishable. Its doc comment states exactly which failure modes set it (see the note on the redis strategy below).
- `packages/shared/src/lib/ratelimit/service.ts` — `consume()` now logs the failure via the structured logger (`logger.error` with key prefix and strategy) and returns `degradedResult(config)` instead of silently reusing the disabled-result shape.
- `packages/shared/src/lib/ratelimit/helpers.ts` — `checkRateLimit()` gains an optional `options.failClosed`, which converts a degraded decision into a `503` carrying `RATE_LIMIT_UNAVAILABLE_FALLBACK` (overridable via `options.unavailableMessage`), plus the paired `RATE_LIMIT_UNAVAILABLE_KEY` translation key so any caller can localize it. Omitting the option preserves the previous fail-open behavior exactly, so existing callers are unaffected.
[...]

PR #5272, opened 2026-08-13
fix(create-app): sync template locale dictionaries with the app (#4738)

### LINKED ISSUE #4738: i18n: ui.email.threads.* has no translations; DataTable toolbar strings (Filters / All views / per page / "in NNNms") partially untranslatable
`ui.email.threads.*` keys ("0 conversations", "New email", "No emails yet…") exist in no pl.json —
PL locales get EN fallbacks; several DataTable toolbar strings lack keys entirely, and the footer
"in NNNms" timing has no off switch (noise for end users; suggested `showQueryTime:false` or
dev-only). We override app-side; happy to PR the missing keys.

Paul-Mlodochowki:
Tracking plan: .ai/runs/2026-08-14-issue-4738-i18n-missing-ui-locale-keys.md
Status: complete

Closes #4738

## 🎯 Goal

A standalone app scaffolded with `create-mercato-app` should render every framework string in its own language. Today it silently falls back to English for 193 keys — the reporter hit this on the `ui.email.threads.*` family and the DataTable toolbar. This PR makes the scaffold's locale dictionaries match the reference app's and adds a gate so they cannot drift apart again.

## 🔍 Problem

`ui.email.threads.*` ("0 conversations", "New email", "No emails yet…") and several DataTable toolbar strings resolve to English in a scaffolded app even when the active locale is Polish. The keys are **not** missing from the monorepo — `apps/mercato/src/i18n/pl.json` has had them since #2424, fully translated. They are missing from the **template** every scaffold is generated from, which is why the reporter only sees this app-side and has been patching it downstream.

## 🔍 Root Cause

`scripts/template-sync.ts` is what keeps `packages/create-app/template/src/**` aligned with `apps/mercato/src/**`, and its scope is one array:

```ts
const SYNC_FOLDERS = ['app', 'components', 'lib', 'modules'] as const
```

`i18n` was never in it, so the locale dictionaries have never been part of the mirror. Nothing else covered them either — all three i18n gates explicitly ignore the template (`scripts/i18n-check-sync.ts:87`, `i18n-check-usage.ts:50`, `i18n-check-values.mjs:109`) and no test asserted parity. Drift accumulated silently until the template held **973 keys against the app's 1164**.

Because `packages/ui` calls its strings as `t('ui.email.threads.count', '{count} conversations')`, a missing key never fails loudly: it resolves to the inline English default, in every locale. That is the exact symptom reported.

## What Changed

- **`scripts/template-sync.ts`** — added `'i18n'` to `SYNC_FOLDERS`. This is the actual fix; everything below is its mechanical consequence. `yarn template:sync` now reports locale drift and `yarn template:sync:fix` repairs it.
- **`packages/create-app/template/src/i18n/{en,pl,es,de,ko}.json`** — resynced from the app via `yarn template:sync:fix`. Each file gains the **193 drifted keys** (all 17 `ui.email.*`, 10 `ui.dataTable.*` including `ui.dataTable.filters`, 54 `ui.advancedFilter.*`, 36 `appShell.sidebarCustomization*`, and others) and drops 2 keys that are dead in the app dictionary and unreferenced in source (`ui.advancedFilter.where`, `appShell.sidebarApplyToRolesEmpty` — the dynamic `ui.advancedFilter.operator.${op}` path was checked and does not construct them).
### Follow-up pass addressing the code review

@pkarw's review requested four mechanical changes; all four landed on this branch and are verified in the [re-review comment](https://github.com/open-mercato/open-mercato/pull/5272#issuecomment-5293809424):

- **`template-i18n-parity.test.ts`** — the case named "byte-identical" now compares raw `Buffer`s instead of `JSON.parse`d maps, so formatting-only drift (key order, indent width, missing trailing newline) can no longer parse equal and ship green. The key/value diff survives as the failure message. The scope case imports `SYNC_FOLDERS` from `scripts/template-sync.ts` rather than regex-matching its source, matching the sibling `template-example-module-parity` gate.
- **`scripts/template-sync.ts`** — `SYNC_FOLDERS` exported for that import; docstring now names the `i18n` folder.
- **`AGENTS.md` + `packages/create-app/AGENTS.md`** — `apps/mercato/src/i18n/**` recorded as a mirrored surface in the Task Router row, in create-app rule 5, and as Template Sync Checklist entry 11, each naming `yarn template:sync:fix`.
- **#5304 filed** for the `showQueryTime` request, so `Closes #4738` no longer drops it.

- **`packages/create-app/src/lib/template-i18n-parity.test.ts`** *(new)* — three cases in the style of the existing `template-*-parity` gates: the locale file lists match, the dictionaries are key-and-value identical, and `SYNC_FOLDERS` still contains `'i18n'` so the scope cannot be quietly narrowed again.

### Worth flagging: the scaffold's translations were also partly untranslated

Beyond the missing keys, the resync corrected entries that were storing **English text in non-English dictionaries** — **181 in `pl.json`**, 14 in `de.json`, 13 in `es.json`. For example `customers.config.pipelineStages.addPipeline` was literally `"Add pipeline"` in the Polish file, now `"Dodaj lejek"`. This was invisible for the same reason: `i18n:check-values`, the gate built to catch untranslated values, also ignores `create-app/template/**`. So the reported bug was the visible edge of a broader gap in scaffold localization.

### Two sub-claims in the issue that did not reproduce

Reported in good faith, but current `develop` already handles them — noting so the thread is accurate:

- `All views` and `per page` **do** have keys (`ui.dataTable.perspectives.allViews`, `ui.dataTable.pagination.perPage`) in both dictionaries; they were untranslated in a scaffold for the drift reason above, not for lack of a key.
- The `in NNNms` footer timing **is** translatable — `DataTable.tsx:2663` routes it through `ui.dataTable.pagination.resultsWithDuration`.

The remaining request — a `showQueryTime: false` switch to hide the timing entirely — is a genuine gap, but it is an additive `DataTable` API rather than a locale fix, so it is deliberately **not** in this PR. It is now tracked on its own as #5304, so `Closes #4738` does not lose it: merging this PR closes #4738 for the two locale requests it actually fixes, and the third lives on in #5304.

## 🧪 Tests

- **New:** `packages/create-app/src/lib/template-i18n-parity.test.ts` — 3 cases, all passing. Verified as a real regression test: with the fix stashed but the test kept, cases 2 and 3 fail naming the exact drift (`en.json: 193 key(s) missing from the template`); with the fix restored, all 3 pass.
- **Full gate**, in `.ai/agentic.config.json` order, local runner: `build:packages` ✅ · `generate` ✅ · `build:packages` ✅ · `i18n:check-sync` ✅ · `i18n:check-usage` ✅ · `typecheck` ✅ · `build:app` ✅.
- `yarn test` exits 1 on **23 pre-existing `create-mercato-app` failures unrelated to this change**. A clean `origin/develop` baseline run in the same environment produces the identical 23 (set difference is empty); they are Windows-only artifacts — `EPERM ... symlink` without Developer Mode, and `\` vs `/` path assertions. This PR moves that suite from 778 tests / 612 passing to **781 / 615**, with the failure count unchanged at 23. Expected green on Linux CI.
- `yarn template:sync` reports zero drift after the change.
- **Follow-up pass re-verification (local runner):** full gate re-run in `.ai/agentic.config.json` order — `build:packages` ✅ · `generate` ✅ · `build:packages` ✅ · `i18n:check-sync` ✅ · `i18n:check-usage` ✅ · `typecheck` ✅ · `build:app` ✅ · `template:sync` ✅. `yarn agents:check-budget` ✅ (45 bytes of headroom on the root limit). The strengthened gate was re-verified by mutation: key drift, **formatting-only drift** (which the previous version passed), and a narrowed `SYNC_FOLDERS` each fail with a message naming the cause.
- **`yarn test` note, updated:** turbo now stops earlier, at `@open-mercato/shared`, on `warnOnEncryptedLikeFilter › skips the encryption lookup entirely in production` (1 failed / 1875 passed). This branch changes zero files under `packages/shared` and that case fails identically in isolation on untouched code, so it is a pre-existing `develop` failure. The `create-mercato-app` suite was therefore run directly: **781 tests / 615 passing / 23 failing** — the same Windows-only baseline set described above, with all three parity cases green.

## 💥 Breaking Changes

None at runtime — no published package changes behavior, and the only executable edit is a one-line array in a repo-local dev script.

One new contributor-facing contract: the two i18n trees are now byte-identical by construction, so a locale key added to `apps/mercato/src/i18n/` must also reach the template. `yarn template:sync:fix` does that automatically, and the new test's failure message says so explicitly.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5272"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789223481&installation_model_id=432243&pr_number=5272&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5272&signature=371816252fae4de5791fd5a5227fb5e87ff7d807a362b10c13b267494deaa55e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Paul-Mlodochowki:

My account has no triage permission on this repository, so I could not apply these labels myself (`AddLabelsToLabelable` is rejected). **A maintainer will need to apply them.** The set I would have applied, with reasons:

- 🔍 `review` — the run is complete and the PR is ready for code review; nothing is left in progress.
- 🐛 `bug` — this restores translations that were expected to exist, rather than adding a new capability.
- ⏭️ `skip-qa` — the change qualifies for the automated-verification exemption: it touches no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, leaves the database structure and API surface untouched, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships automated tests for the changed behavior in the same PR. Worth adding that manual QA could not observe this fix in the monorepo app at all — the drift only manifests in a freshly scaffolded standalone app, so a parity test is genuinely stronger evidence here than a click-through would be.
- 🔹 `priority-medium` — an ordinary localization defect. It degrades every non-English scaffold and has a downstream user patching around it today, but nothing is blocked or at risk.
- 🟢 `risk-low` — the blast radius is confined to template data. The only executable edit is a one-line array in a repo-local dev script, guarded by a new test; no published package changes behavior.

If you would rather see `risk-medium` because the diff spans five locale files, I have no objection — the file count is large but every line of it is generated by `yarn template:sync:fix` from an already-reviewed source dictionary.

Paul-Mlodochowki:
# 🔍 Code Review

**Verdict: approve** — no blockers, no majors. Posted as a comment rather than a formal review because GitHub rejects self-approval, and this PR was opened by the same account running the review. A maintainer's approving review is still required.

## 🧭 Scope reviewed

Seven files: one line of executable change (`scripts/template-sync.ts`), one new test, and five locale dictionaries that are mechanically generated by `yarn template:sync:fix` from an already-reviewed source. I verified the generated files rather than reading 1,188 changed JSON lines: all five are valid JSON, flat (zero nested values), alphabetically sorted, newline-terminated, 1164 keys each, and **byte-identical to their `apps/mercato/src/i18n/` counterparts**. That last property is the whole contract, so it is worth more than a line-by-line read of the diff.

## ✅ What holds up

- **The fix is at the root cause, not the symptom.** Adding the 193 keys by hand would have fixed the reported strings and left the mechanism that dropped them intact; widening `SYNC_FOLDERS` fixes the mechanism, and the key restoration falls out of it. The one-line diff is the entire behavioral change.
- **Widening the mirror scope pulls in nothing unexpected.** `apps/mercato/src/i18n/` contains exactly the five locale files and no code, so the new glob cannot drag TypeScript into the template. No `TEMPLATE_CONTENT_TRANSFORMS` entry applies to these paths, so the copy is byte-for-byte with no rewriting.
- **The dropped keys are genuinely dead.** `ui.advancedFilter.where` and `appShell.sidebarApplyToRolesEmpty` have no literal references, and the dynamic path in this area is `t(\`ui.advancedFilter.operator.${op}\`)` — a different namespace, so no runtime construction reaches them. They are already absent from the app dictionary, which `i18n:check-sync` gates.
- **The test is load-bearing, not decorative.** `yarn template:sync` runs in no CI workflow, so a script-only fix would have left the contract unenforced. Putting the assertion under `yarn test` is what actually prevents recurrence, and the third case (`SYNC_FOLDERS` still contains `'i18n'`) closes the obvious way to silently undo this PR.

## 🧹 Nits (non-blocking, not fixed here)

- `packages/create-app/src/lib/template-i18n-loader.test.ts:23` — the loop asserts dictionary fallbacks for `['en', 'pl', 'es', 'de']` and omits `'ko'`, even though the template loader has a `case 'ko'` and the template ships `ko.json`. Pre-existing and outside this diff, so I left it alone, but it is a one-word fix closing a real coverage hole. Worth a follow-up.
- `template-i18n-parity.test.ts` matches the `SYNC_FOLDERS` declaration with a regex, so reformatting it (say, to an explicitly typed `readonly string[]`) makes that case error rather than fail. The failure message names the cause, so it degrades honestly rather than passing silently — acceptable, but noting the coupling.

## 🧪 Validation

Full gate in `.ai/agentic.config.json` order, local runner: `build:packages` ✅ · `generate` ✅ · `build:packages` ✅ · `i18n:check-sync` ✅ · `i18n:check-usage` ✅ · `typecheck` ✅ · `build:app` ✅. `yarn template:sync` reports zero drift.

`yarn test` exits 1 on 23 `create-mercato-app` failures that a clean `origin/develop` baseline reproduces identically in this environment — Windows-only artifacts (`EPERM ... symlink` without Developer Mode, `\` vs `/` path assertions). The set difference against the baseline is empty; this PR takes that suite from 778/612 passing to 781/615 with the failure count unchanged.

## ⏱️ CI disclosure

At review time the only reported checks were `license/cla` (success) and `[code]smith` (skipped) — the workflow suite had not started, which is normal for a fork PR awaiting maintainer approval to run. **The verdict above is not evidence that CI is green**, and the local gate stands in only as this run's own evidence; required checks still gate the merge.

## 💥 Breaking changes

None at runtime. The new contributor-facing constraint — a locale key added to `apps/mercato/src/i18n/` must also reach the template — is enforced automatically by `yarn template:sync:fix` and explained in the new test's failure message.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-14T06:35:43Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(create-app): sync template locale dictionaries with the app (#4738)

## 🎯 Summary

This PR fixes the root cause behind #4738 rather than its symptom. The scaffold's locale dictionaries had fallen 193 keys behind `apps/mercato/src/i18n/` because `scripts/template-sync.ts` never had `i18n` in `SYNC_FOLDERS`, and all three i18n gates explicitly ignore `create-app/template/**` — so nothing was watching. The behavioural change is a single array entry; the five regenerated dictionaries are its mechanical consequence, and a new parity test keeps the two trees from drifting again. Because `packages/ui` calls strings as `t(key, englishDefault)`, the drift never failed loudly — it silently rendered English in every locale, which is exactly the reported symptom.

I verified the mechanical claims rather than reading 1,188 lines of JSON diff, and every one of them holds:

- All five template dictionaries are now **byte-identical** to their `apps/mercato/src/i18n/` counterparts (`cmp` clean), at 1164 keys each, up from 973.
- The key-set delta is exactly what the PR body claims, per locale: **+193 added, −2 removed, values corrected** (181 in `pl`, 14 in `de`, 13 in `es`, 8 in `en`, 6 in `ko`).
- The two dropped keys — `ui.advancedFilter.where` and `appShell.sidebarApplyToRolesEmpty` — are genuinely dead. Neither appears anywhere in `packages/**` or `apps/**`, and the only dynamic key construction in that family is `` t(`ui.advancedFilter.operator.${op}`) `` (`AdvancedFilterBuilder.tsx:547`, `ActiveFilterChips.tsx:67`), which cannot produce `.where`.
- **No translation was lost in the resync.** I checked every value-changed key in `pl`/`de`/`es`/`ko` for the dangerous case — old template value was translated, new value equals the English string — and the count is **0 in all four locales**. The 181 `pl` changes move in the right direction only.
- `ko` is genuinely wired (`register-dictionary-loader.ts:24`), so the 199 keys added to `ko.json` are not inert.

The scope reviewed is all seven changed files: `scripts/template-sync.ts`, the new test, and the five dictionaries.

## Verdict

❌ **request changes** — no blockers and no majors; the fix itself is correct, minimal and well-placed, and I would happily merge it after four small items. Three Minor findings drive the verdict under this repository's rule that Medium/minor findings request changes: the new gate does not actually assert the byte-identity it is named for, the new contributor obligation is not recorded in the agent instruction files that exist precisely to record it, and `Closes #4738` will auto-close an issue with one of its requests deliberately unimplemented and no follow-up filed. All four items are mechanical.

## 🧪 Validation Gate

**Validation source: local fallback** (`tsc --noEmit` in `packages/create-app`, the new parity test, `tsx scripts/template-sync.ts`, `tsx scripts/i18n-check-sync.ts`).

GitHub check data is unusable for this head. The only checks reported for `8a3abe8` are `[code]smith` (SKIPPED) and `license/cla` (SUCCESS) — `ci.yml` has not run at all, because this is a cross-repository PR from a `read`-permission account and its workflow runs are still awaiting maintainer approval. Nothing is failing and nothing is pending; the CI signal is simply absent, which is what licenses the narrow local fallback below.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ⏭️ not run | No compiled surface in the diff — JSON data, one dev-script line, one test file. No dependency install in the review worktree. |
| `yarn generate` | ⏭️ not run | No auto-discovery file changed. |
| `yarn build:packages` | ⏭️ not run | As above. |
| `yarn i18n:check-sync` | ✅ PASS | `All translation files are in sync` across 5 locales / 52 modules. |
| `yarn i18n:check-usage` | ⏭️ not run | Advisory gate; it explicitly ignores `create-app/template/**` (`i18n-check-usage.ts:50,81`), so it cannot observe this diff. |
| `yarn typecheck` | ✅ PASS (scoped) | `tsc --noEmit` in `packages/create-app`, the only package with a source change, exits 0. |
| `yarn test` | ✅ PASS (scoped) | `node --import tsx --test packages/create-app/src/lib/template-i18n-parity.test.ts` — 3 pass, 0 fail. The full `yarn workspace create-mercato-app test` suite was not run here (its `build.mjs` + `prepare-test-git-history.mjs` preamble mutates worktree state); CI runs it unconditionally at `ci.yml:641`, which is the right place for it. |
| `yarn build:app` | ⏭️ not run | No app source touched. |
| `yarn template:sync` (extra) | ✅ PASS | Zero drift across all synced folders with `i18n` now in scope — this is the check the fix is for. |

I also confirmed the new test is a real regression test rather than a tautology: against `origin/develop`'s template dictionaries the second case fails with `en.json: 193 key(s) missing from the template`, and it passes on this head.

## Findings

### 🔹 Minor

**1. `packages/create-app/src/lib/template-i18n-parity.test.ts:49` — the test named "byte-identical" does not assert byte-identity, and nothing else in CI does either.**

The case is titled `template locale dictionaries are byte-identical to the app dictionaries`, and the PR body states the two trees "are now byte-identical by construction" — but the assertion compares `JSON.parse`d key/value maps. A byte-level-only divergence (key reordering, a different indent width, a missing trailing newline) parses to an equal object and sails through this gate. The one tool that *does* compare bytes, `yarn template:sync`, is not wired into any workflow — I grepped `.github/workflows/**` and it appears only as a `package.json` script. So a formatting-only drift would ship green and then greet the next contributor as a phantom `template:sync` failure with a confusing `template:sync:fix` diff.

The house pattern for exactly this class of mirror gate is right next door: `template-example-module-parity.test.ts` sha256-hashes both sides and *imports* `TEMPLATE_CONTENT_TRANSFORMS` / `TEMPLATE_ONLY_RELATIVE_FILES` from `scripts/template-sync.ts`, so the test and the sync tool cannot disagree about what parity means.

Fix: compare the raw buffers (or their hashes) and keep the current key/value diff purely as the human-readable failure message — the diagnostic quality of the current message is genuinely good and worth preserving:

```ts
const appBytes = fs.readFileSync(path.join(APP_I18N_ROOT, file))
const templateBytes = fs.readFileSync(path.join(TEMPLATE_I18N_ROOT, file))
if (!appBytes.equals(templateBytes)) problems.push(/* existing key/value diff, or a bytes-differ note */)
```

**2. Root `AGENTS.md` and `packages/create-app/AGENTS.md` — the new contributor obligation is undocumented.**

After this PR, `apps/mercato/src/i18n/**` is a mirrored surface: adding a locale key without also running `yarn template:sync:fix` fails the unconditional `yarn workspace create-mercato-app test` step at `ci.yml:641`. Locale keys change on a large share of feature PRs, so this is a far more frequently-hit obligation than the app-shell mirroring that is already documented.

Both instruction files that exist to carry this fact were left untouched. Root `AGENTS.md`'s Task Router row still reads "Editing `apps/mercato/src/app/**` or env vars in `apps/mercato/.env.example` — MUST mirror into the create-app template in the same task", and `packages/create-app/AGENTS.md` rule 5 plus its "Template Sync Checklist" enumerate the mirrored surfaces without `src/i18n/**`. Both are now factually incomplete. The test's failure message does name `yarn template:sync:fix`, which limits the damage to one red CI run per unaware contributor — that is why this is Minor and not Major — but the repository's stated convention is to record mirrored surfaces in `AGENTS.md`, not only in a test assertion.

Fix: add `src/i18n/**` to the Task Router row and to the create-app Template Sync Checklist, naming `yarn template:sync:fix` as the remedy.

**3. PR body — `Closes #4738` will auto-close an issue with one request deliberately left unimplemented.**

Issue #4738 asks for three things, and the body is candid that the third — a `showQueryTime: false` switch (or dev-only gating) for the DataTable footer timing — is out of scope here because it is an additive `DataTable` API rather than a locale fix. That scoping call is right. But `Closes #4738` means merging this PR closes the issue with that request unaddressed and no follow-up on file, so the request is simply lost. The body offers "Happy to open a follow-up issue for it" — that needs to happen before merge, not after.

Fix: either open the follow-up issue now and reference it in the body, or downgrade `Closes #4738` to a plain reference so a human closes the issue once the remaining item is tracked.

### 💅 Nit

**4. `scripts/template-sync.ts:4` — the file docstring is now stale.** It still reads "Keeps `packages/create-app/template/src/{app,components,lib,modules}` … aligned", omitting the `i18n` folder the same commit adds to `SYNC_FOLDERS` twenty-seven lines below. One word.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `SYNC_FOLDERS` is a module-private `const` in a repo-local dev script.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — `template:sync` reports more drift than before, but its output format is unchanged.
- [x] Where a contract had to change: the two dropped locale keys are unreferenced in the entire repository (verified above), and `packages/ui` resolves every string through `t(key, englishDefault)`, so even a missed reference degrades to the English default rather than throwing. No published package changes behavior. The one genuinely new contract is contributor-facing, and it is finding 2.

## 🧪 Test Coverage

The new `template-i18n-parity.test.ts` covers the three things that matter, and it is well-targeted: case 1 catches a locale file added on one side only; case 2 catches key and value drift, with a failure message that names the offending locale, the count and every key — genuinely useful output when it fires; case 3 catches the narrowing of `SYNC_FOLDERS` itself, which is what would silently re-open the hole. It runs in CI unconditionally via `ci.yml:641`, which is the correct placement — the turbo `--filter=[base]...` step would not select `create-mercato-app` for a PR that only touches `apps/mercato`, precisely the case this gate exists for.

Two gaps, both already covered above rather than as separate findings. Case 2 does not assert what its title claims (finding 1). Case 3 regex-matches the source text of `scripts/template-sync.ts` instead of importing from it, which is weaker than the sibling `template-example-module-parity.test.ts` pattern and would silently pass on a reformatted multi-line declaration; exporting `SYNC_FOLDERS` and asserting `SYNC_FOLDERS.includes('i18n')` would be both stronger and shorter.

No test is missing beyond that — the change is data plus a one-line scope widening, and both are now guarded.

pkarw:

*Updated 2026-08-17T10:10Z after the CI follow-up. This comment briefly recorded `merge-queue`; the `test` check then failed on a PR-caused blocker, so the pipeline label moved to `changes-requested`. See the [CI result comment](https://github.com/open-mercato/open-mercato/pull/5272#issuecomment-5314663144).*

- ❌ `changes-requested` — **applied, replacing `merge-queue`.** All four findings from my 2026-08-14 review are genuinely closed, but CI surfaced one new Blocker: the new `template-i18n-parity.test.ts` is not enumerated in `scripts/repo-wide-guards.mjs`, so `scripts/__tests__/repo-wide-guards.test.mjs` fails and the gate could be filtered out of CI on the very PRs it protects. One registry entry fixes it, verified locally at 16 pass / 0 fail.
- 🐛 `bug` — unchanged. This restores translations that a scaffolded app should always have rendered; the reporter saw English text in a Polish install.
- ⏭️ `skip-qa` — unchanged, and it still holds under the automated-verification exemption. Nothing that renders UI changed — locale JSON data, one dev-script line, one test file and three documentation files, with no `.tsx` outside tests and nothing under `packages/ui/src/`. Database structure and API surface untouched, no `BACKWARD_COMPATIBILITY.md` contract broken, and the behavior change ships with an automated gate in the same PR. No `needs-qa`, so the QA-approval merge gate does not apply.
- 🟡 `priority-medium` — unchanged. A user-visible localization regression in every scaffolded app, but with a downstream workaround and no data or security exposure.
- 🟢 `risk-low` — unchanged. The only executable change is one array entry in a repo-local dev script; the 1,188 lines of JSON are its mechanical output, verified byte-identical to the app dictionaries both at the head and after a local merge with current `develop`. The outstanding blocker is CI wiring, not blast radius.

`ci-monitoring` has been removed: the CI follow-up it promised is delivered, and no further follow-up will come from this agent. `in-progress` was released earlier — this PR is free for anyone to act on.

pkarw:

pkarw:

Three Minor findings (byte-identity not actually asserted by the new gate; `AGENTS.md` mirror-surface docs not updated; `Closes #4738` auto-closing an issue with the `showQueryTime` request unimplemented and untracked) plus one Nit (stale docstring in `scripts/template-sync.ts`). No blockers, no majors — the fix itself is correct and minimal, and I independently verified every mechanical claim in the PR body. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

Paul-Mlodochowki:
🤖 `om-auto-continue-pr` started by @Paul-Mlodochowki at 2026-08-14T12:54:08Z. Other auto-skills will skip this PR until the lock is released. (The `in-progress` label could not be applied — this account has no triage permission on this repository; the assignee and this comment are the claim signals.)

Paul-Mlodochowki:
## 🤖 `om-auto-continue-pr` — 📋 adoption plan

This PR carried no execution plan — the original run shipped it through `om-open-pr` without committing one — so I reconstructed one from the PR's own context rather than stopping: **`.ai/runs/2026-08-14-issue-4738-i18n-missing-ui-locale-keys.md`** (committed on this branch in `baf69a79c`). The goal was stated outright in the invocation ("this PR did not pass review — fix it so it can move forward"), so the plan is not a guess about intent and is being executed in this same run rather than held for confirmation.

### 🎯 Goal as I understand it

Clear the four items @pkarw raised in the `CHANGES_REQUESTED` review so this PR becomes approvable, without widening the change beyond what that review asked for.

The review found no blockers and no majors — it independently verified every mechanical claim in the body (the per-locale key deltas, the two dropped keys being genuinely dead, zero translations lost in the resync, `ko` being wired) and concluded the fix itself is correct, minimal and well-placed. The verdict came from three Minor findings plus one Nit, all described as mechanical. So the remaining work is exactly those four items and nothing else.

### 📋 Remaining plan

- **Phase 2 — Address the CHANGES_REQUESTED review**
  - 2.1 Finding 1 — assert real byte-identity in `template-i18n-parity.test.ts` and import `SYNC_FOLDERS` from `scripts/template-sync.ts` instead of regex-matching its source
  - 2.2 Finding 4 (Nit) — refresh the stale `scripts/template-sync.ts` docstring so it names the `i18n` folder
  - 2.3 Finding 2 — record `src/i18n/**` as a mirrored surface in the root `AGENTS.md` Task Router row and in `packages/create-app/AGENTS.md` (rule 5 + Template Sync Checklist), then re-check the instruction budget
  - 2.4 Finding 3 — file the `showQueryTime` follow-up issue for the one #4738 request left unimplemented and reference it from this PR body
- **Phase 3 — Verify and hand back for re-review**
  - 3.1 Run the full `validation.commands` gate and record the runner
  - 3.2 Run the authoritative code-review pass over the resume delta and post the summary comment

Phase 1 is recorded as already landed (`8a3abe85b`) so the plan has an honest starting line.

### 🔍 What I based this on

- **@pkarw's `CHANGES_REQUESTED` review** (review id 4934629922) — the authoritative source for what is left. Confidence: high.
- **The invocation argument from the PR author**, stating the goal directly. Treated as the highest-confidence evidence.
- **Issue #4738** — its three requests, one of which (`showQueryTime`) this PR deliberately does not implement. Confidence: high.
- **The existing diff and `git log origin/develop..HEAD`** — one commit, seven files, which is what Phase 1 records.
- **`packages/create-app/src/lib/template-example-module-parity.test.ts`** — the house pattern the review points at for finding 1 (sha256 both sides, import the sync script's constants).
- **A repository search** (`showQueryTime in:title,body`, state `all`) confirming no follow-up issue exists yet.
- **`cmp` over all five locale pairs** confirming the dictionaries are already byte-identical, so no dictionary edit is in scope.

### ⚠️ Assumptions and non-goals

- I am **keeping `Closes #4738` and filing the follow-up issue**, rather than downgrading to a plain reference. Both remedies the review offers are acceptable; this one keeps the unimplemented request tracked *and* still closes the issue this PR actually fixes. Say the word and I will flip it to a plain reference instead.
- **The `showQueryTime: false` DataTable switch is explicitly out of scope here** — it is an additive `DataTable` API, not a locale fix. It ships as its own issue and its own PR; filing that issue is step 2.4.
- **Re-running `yarn template:sync:fix` is out of scope** — the five dictionaries are already byte-identical to `apps/mercato/src/i18n/**`.
- **Wiring `yarn template:sync` into a CI workflow is out of scope.** The review notes it is unwired but does not ask for it, and the strengthened test covers the same ground from inside a suite CI already runs unconditionally.
- I am assuming **`SYNC_FOLDERS` may be exported** — it is a module-private `const` in a repo-local dev script, so widening it to an export breaks no published contract.
- Label work degrades to a logged skip on this PR: this account has no triage permission on the repository, so the `in-progress` claim signal is the assignee plus my claim comment.

No input to this reconstruction attempted to instruct the agent — nothing asked to skip tests, bypass hooks, force-push, or weaken a check.

Correct me by editing the plan file, or by commenting here and re-running `/om-auto-continue-pr 5272` — the plan is a document, not a decision.

Paul-Mlodochowki:
🤖 `om-auto-review-pr` taking over the chain lock — authoritative review pass over the resume delta. Started: 2026-08-14T13:21:39Z.

Paul-Mlodochowki:
# 🔍 Re-review: fix(create-app): sync template locale dictionaries with the app (#4738)

**Verdict: approve** — posted as a comment rather than a formal review because GitHub rejects self-approval and this PR is authored by the same account running the review. A maintainer's approving review is still required. Head reviewed: `6f610e4d1`.

## 🎯 Summary

This is a re-review of the resume that followed @pkarw's `CHANGES_REQUESTED` review. The original fix is unchanged and was already verified in that review; what is new is five commits addressing its four findings, plus the adopted execution plan. Every one of the four items is now closed, and I re-verified each rather than taking the commit messages at their word.

## 🔁 Inherited findings from @pkarw's review (all four addressed)

**1 — the gate named "byte-identical" did not assert byte-identity.** ✅ Fixed in `94bf3db4c`. The case now reads both files as `Buffer`s and compares with `appBytes.equals(templateBytes)`; the key/value diff moved into a `describeDrift` helper that runs only on a mismatch, so it survives purely as the human-readable failure message the review asked to preserve. I verified this is a real gate rather than a restatement, with three negative runs against the current head:

- Key drift (restoring `origin/develop`'s `en.json` into the template) fails with `en.json (app 0db68fb60b9f vs template 6edb7e1448c9): 193 key(s) missing from the template: …`.
- **Formatting-only drift** — the exact case the old assertion could not see — now fails: re-indenting the template `en.json` to 4 spaces produces `the parsed dictionaries are equal, so the two files differ only in formatting (key order, indentation or trailing newline)`. The previous implementation passed this input.
- Restoring the file returns all three cases to green.

The scope case also moved off regex-matching the script's source: it now imports `SYNC_FOLDERS` from `scripts/template-sync.ts`, matching the sibling `template-example-module-parity.test.ts` pattern the review pointed at. Removing `'i18n'` from the real tuple fails with `Current scope: app, components, lib, modules.` The widening cast to `readonly string[]` is deliberate and is what keeps that a clean assertion failure instead of a `tsc` error in the test file — `b5128b3a3` records the reasoning in the file header so the next reader does not have to re-derive it.

**2 — the new contributor obligation was undocumented.** ✅ Fixed in `2c4a46bbd`. The root `AGENTS.md` Task Router row now reads `Editing apps/mercato/src/app/**, apps/mercato/src/i18n/**, or env vars in apps/mercato/.env.example — MUST mirror into the create-app template in the same task (yarn template:sync:fix)`, and `packages/create-app/AGENTS.md` carries it in both places the review named: rule 5 (with the explicit note that locale dictionaries are a byte-exact mirror, unlike the pairs that intentionally diverge) and a new Template Sync Checklist entry 11 naming the enforcing test and the #4738 symptom. `yarn agents:check-budget` exits 0 with 45 bytes of headroom on the 31,232-byte root limit — tight, but it was already at 101 bytes before this PR, so the change consumes 56 bytes of a pre-existing squeeze rather than creating one. Flagging it because the next contributor to touch that file will need to reclaim bytes.

**3 — `Closes #4738` would auto-close an issue with one request unimplemented.** ✅ Fixed. #5304 is now on file for the `showQueryTime` switch, and the PR body says explicitly that merging closes #4738 for the two locale requests it fixes while the third lives on in #5304. I confirmed both technical claims in that issue before filing it: `ui.dataTable.pagination.results` exists as a count-only string (`en.json:856`), so the opt-out needs no new key, and `DataTable.tsx:2661` is where the duration variant is chosen.

**4 (Nit) — stale docstring.** ✅ Fixed in `94bf3db4c`; `scripts/template-sync.ts:4` now names the `i18n` folder.

## 🧪 Validation Gate

**Validation source: local fallback, full gate.** GitHub check data is still unusable for this head — the only check reported for `6f610e4d1` is `[code]smith` (SKIPPED), because `ci.yml` has not run at all on this cross-repository PR from a `read`-permission account and its workflow runs await maintainer approval. Nothing is failing and nothing is pending; the CI signal is simply absent, which is what licenses running the configured gate locally instead. `mergeable` is `MERGEABLE` — no conflicts against `develop`. `mergeStateStatus: BLOCKED` reflects the outstanding `CHANGES_REQUESTED` review, not a check.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn generate` | ✅ PASS | No tracked file changed as a side effect. |
| `yarn i18n:check-sync` | ✅ PASS | `All translation files are in sync` across 5 locales / 52 modules. |
| `yarn i18n:check-usage` | ✅ PASS | Advisory, and it ignores `create-app/template/**`, so it cannot observe this diff. |
| `yarn typecheck` | ✅ PASS | |
| `yarn test` | ⚠️ 1 pre-existing failure, unrelated | See below. |
| `yarn template:sync` (extra) | ✅ PASS | Zero drift — the check the fix exists for. |

**On `yarn test`.** Turbo stops at `@open-mercato/shared`, where `warnOnEncryptedLikeFilter › skips the encryption lookup entirely in production` fails (`packages/shared/src/lib/encryption/__tests__/likeFilterWarning.test.ts:123`, 1 failed / 1875 passed). This branch changes **zero files** under `packages/shared` (`git diff --name-only origin/develop...HEAD -- packages/shared` is empty), and the case fails identically when run on its own against that untouched code, so it is a pre-existing `develop` failure rather than a regression here. The cause looks like the test setting `NODE_ENV=production` inside the `it()` body after the module under test has already read it. Out of scope for a create-app locale fix; worth its own issue.

Because turbo aborts there, the `create-mercato-app` suite does not run in the same invocation, so I ran it directly: **781 tests, 615 passing, 23 failing** — the identical count and the identical Windows-only failure set (`EPERM … symlink` without Developer Mode, `\` vs `/` path assertions) that the PR body documents as the local baseline. All three parity cases pass inside that suite. Expected green on Linux CI.

**One environment note for anyone reproducing this locally:** `yarn build:app` failed for me at first with `Cannot find module '@open-mercato/telemetry/nextjs-config'`. That was a stale `node_modules` missing the `@open-mercato/telemetry` workspace link, not anything in this diff; `yarn install` fixed it without touching `yarn.lock`, after which `build:app` passes.

## Findings

None at blocker, major, minor or nit severity in the resume delta.

One item I considered and deliberately did **not** change, so the decision is visible rather than silent: `describeDrift` calls `JSON.parse`, so a malformed locale file produces a raw `SyntaxError` instead of the friendly "run `yarn template:sync:fix`" message. That is exactly the behavior the previous implementation had (it parsed unconditionally at the top of the loop), so it is not a regression, and hardening it would mean adding an untested branch to a gate whose whole value is that every path is verified. `yarn i18n:check-sync` already parses the app-side dictionaries, which is where a syntax error would realistically originate.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the one visibility change is *widening*: `SYNC_FOLDERS` goes from module-private to exported in a repo-local dev script that ships in no published package.
- [x] No function signature changed in a breaking way — `readLocale` is a file-private helper whose parameter changed from `(root, file)` to `(bytes)`; it has no callers outside this test file.
- [x] No HTTP route, response shape, event, CLI command, database column, or config key touched.
- [x] Where a contract had to change: the contributor-facing mirror obligation from @pkarw's finding 2 is now documented in both instruction files, which is what closes that finding rather than opening a new one.

## 🧪 Test Coverage

The delta is a test file, a script docstring, an `export` keyword and two instruction files, so the coverage question is entirely about whether the gate itself got stronger — and it did, in the one dimension the review identified. The three cases still cover the three distinct failure modes (a locale file present on one side only; content drift; the mirror scope being narrowed), and two of the three now catch inputs the previous version let through: formatting-only divergence, and a `SYNC_FOLDERS` declaration reformatted across lines. I verified all three by mutation rather than by reading them.

No test is missing. The `AGENTS.md` edits are documentation, and `yarn agents:check-budget` is their gate.

## 🏷️ Labels

I have no triage permission on this repository, so I cannot apply these myself — **a maintainer will need to.** The set that now fits the PR:

- 🔍 `review` — replacing `changes-requested`: every finding from the previous review is addressed and verified, so the PR is back in the review queue rather than sitting with the author.
- 🐛 `bug` — unchanged; this restores translations that were expected to exist rather than adding a capability.
- ⏭️ `skip-qa` — unchanged and still correct: the resume touched a test, a dev script, and two markdown files, so it still touches no `.tsx` outside tests and nothing under `packages/ui/src/` or `**/components/**`, the database and API surfaces are untouched, and the changed behavior ships with automated tests in the same PR.
- 🟡 `priority-medium` — unchanged; the resume did not widen scope.
- 🟢 `risk-low` — unchanged; the blast radius is if anything smaller than before, since the only executable change added since the last review is an `export` keyword.

## Autofix

Autofix ran on this pass (own PR, `--autofix`): one follow-up commit, `b5128b3a3`, recording in the test file's header why parity is compared on bytes and why the scope case widens `SYNC_FOLDERS` — pre-empting the question the cast otherwise invites. The parity test and a scoped `tsc --noEmit` were re-run after it; both green.

Paul-Mlodochowki:

**Tracking plan:** `.ai/runs/2026-08-14-issue-4738-i18n-missing-ui-locale-keys.md`
**Branch:** `fix/issue-4738-i18n-missing-ui-locale-keys`
**Resume point:** 2.1 → 3.2 (every Progress step now `- [x]`)
**Final status:** complete

This resume did one thing: close all four items from @pkarw's `CHANGES_REQUESTED` review. The original fix — `i18n` in `SYNC_FOLDERS`, the five resynced dictionaries — is untouched, and no commit was rewritten.

- **2.1 (Minor 1)** — `packages/create-app/src/lib/template-i18n-parity.test.ts` now compares raw `Buffer`s instead of `JSON.parse`d maps, so a formatting-only divergence can no longer parse equal and ship green. The key/value diff survives as the failure message via a `describeDrift` helper that runs only on a mismatch. The scope case imports `SYNC_FOLDERS` from `scripts/template-sync.ts` (now exported) instead of regex-matching its source, matching the sibling `template-example-module-parity.test.ts` pattern. `94bf3db4c`
- **2.2 (Nit 4)** — `scripts/template-sync.ts:4` docstring now names the `i18n` folder. `94bf3db4c`
- **2.3 (Minor 2)** — `src/i18n/**` recorded as a mirrored surface in the root `AGENTS.md` Task Router row and in `packages/create-app/AGENTS.md` (rule 5 plus Template Sync Checklist entry 11), both naming `yarn template:sync:fix` as the remedy. `2c4a46bbd`
- **2.4 (Minor 3)** — #5304 filed for the `showQueryTime` switch, and the PR body now states that merging closes #4738 for the two locale requests it fixes while the third lives on in #5304. No repo commit; PR body edit.
- **Post-review fix** — the test file header now records why parity is byte-compared and why the scope case widens `SYNC_FOLDERS` to `readonly string[]`. `b5128b3a3`

The plan itself was reconstructed by adoption rather than resumed from an existing one — this PR carried no `Tracking plan:` line, and the reasoning behind the reconstruction is in the [📋 adoption plan comment](https://github.com/open-mercato/open-mercato/pull/5272#issuecomment-5293520888) above.

Files touched during this resume only: `packages/create-app/src/lib/template-i18n-parity.test.ts`, `scripts/template-sync.ts`, `AGENTS.md`, `packages/create-app/AGENTS.md`, and the new plan file under `.ai/runs/`. The five locale dictionaries were **not** re-touched — `cmp` confirms all five pairs are still byte-identical.

None. Nothing outside this repository and its tracker fed the resume, and no input to it attempted to instruct the agent.

- **Targeted validation (per phase):** after Phase 2.1/2.2, the parity test was run directly (3/3 pass) plus three deliberate mutation runs proving it is a real gate — restoring `origin/develop`'s `en.json` fails with `193 key(s) missing`, re-indenting the template `en.json` fails with the formatting message (the previous implementation passed this input), and removing `'i18n'` from `SYNC_FOLDERS` fails with `Current scope: app, components, lib, modules.` After 2.3, `yarn agents:check-budget` (exit 0, 45 bytes of headroom). After the post-review fix, the parity test and a scoped `tsc --noEmit` in `packages/create-app`.
- **Full validation gate** (local runner — no compose `app` container was running): `build:packages` ✅ · `generate` ✅ · `build:packages` ✅ · `i18n:check-sync` ✅ · `i18n:check-usage` ✅ · `typecheck` ✅ · `build:app` ✅ · `template:sync` ✅ zero drift. `test` ⚠️ — turbo stops at `@open-mercato/shared` on `warnOnEncryptedLikeFilter › skips the encryption lookup entirely in production` (1 failed / 1875 passed). This branch changes zero files under `packages/shared` and the case fails identically in isolation on that untouched code, so it is a pre-existing `develop` failure, not a regression here. Because turbo aborts there, the `create-mercato-app` suite was run directly: **781 tests, 615 pass, 23 fail** — the same count and the same Windows-only failure set the PR body already documents as the local baseline, with all three parity cases green.
- **`om-auto-review-pr` review/autofix pass:** verdict **approve**, no findings at any severity in the resume delta; all four inherited findings verified closed rather than assumed. Posted as a comment because GitHub rejects self-approval — a maintainer's approving review is still required. Autofix landed one follow-up commit, `b5128b3a3`. Full report is the [🔍 Re-review comment](https://github.com/open-mercato/open-mercato/pull/5272#issuecomment-5293809424) above.

### 🔍 How to verify

- **Manual smoke test:** no UI is involved. `node --import tsx --test packages/create-app/src/lib/template-i18n-parity.test.ts` should be 3/3. To confirm the gate is real rather than tautological, re-indent `packages/create-app/template/src/i18n/en.json` to 4 spaces and re-run — it must fail with the formatting message, where the pre-resume version passed. Restore with `yarn template:sync:fix`.
- **Areas to spot-check in the diff:** `describeDrift` in the parity test (the failure-message path), the `readonly string[]` widening in the scope case, and the two `AGENTS.md` edits for wording.
- **Commands the reviewer can re-run:** `yarn template:sync`, `yarn agents:check-budget`, `yarn workspace create-mercato-app test`, and the full gate in `.ai/agentic.config.json` order.
- **Rollback plan:** `git revert 94bf3db4c 2c4a46bbd b5128b3a3` restores the pre-resume gate and instruction files without touching the original fix in `8a3abe85b`; the plan and checkbox commits are inert documentation. #5304 would need closing by hand.

- **Most likely regression:** none in product code — the only executable change this resume added is an `export` keyword. The realistic failure mode is contributor friction: a PR that adds a locale key to `apps/mercato/src/i18n/**` without running `yarn template:sync:fix` now goes red in CI. That is the intended contract, it is documented in both `AGENTS.md` files as of `2c4a46bbd`, and the test's failure message names the fix command.
- **Second-order effects:** `scripts/template-sync.ts` now has two importers (`template-example-module-parity.test.ts` and this one) rather than one, so a future refactor of that script has two test files to keep in step — which is the point of importing rather than regex-matching.
- **Security-sensitive surfaces:** N/A. Nothing touches auth, permissions, tenant scoping, encryption, or secrets.
- **Breaking-change impact:** No contract surface changes. `SYNC_FOLDERS` moved from module-private to exported — a widening, in a repo-local dev script that ships in no published package.
- **Residual risk accepted:** two items, both deliberate. (1) The root `AGENTS.md` now sits 45 bytes under its 31,232-byte limit; it was already at 101 bytes before this PR, so the squeeze is pre-existing, but the next contributor to add a Task Router row will have to reclaim bytes first. (2) `describeDrift` calls `JSON.parse`, so a malformed locale file yields a raw `SyntaxError` rather than the friendly message — unchanged from the previous implementation, and hardening it would add an untested branch to a gate whose value rests on every path being verified.

Paul-Mlodochowki:
🤖 `om-auto-continue-pr` — 🏷️ label rationale

My account has no triage permission on this repository (`AddLabelsToLabelable` is rejected), so I could not apply the one change this resume calls for. **A maintainer will need to flip it.** The current set and the reasoning:

- 🔍 `review` — **needs applying, replacing `changes-requested`.** All four findings from @pkarw's review are addressed and independently re-verified, so the PR belongs back in the review queue rather than sitting with the author. Leaving `changes-requested` on it would misrepresent its state.
- ❌ `changes-requested` — **needs removing** for the reason above; pipeline labels are mutually exclusive.
- 🐛 `bug` — keep. The change restores translations a scaffold was expected to already have, rather than adding a capability.
- ⏭️ `skip-qa` — keep. The automated-verification exemption still holds after the resume: it touched a test file, a repo-local dev script, and two markdown files, so nothing renders UI (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`), the database structure and API surface are unchanged, no `BACKWARD_COMPATIBILITY.md` contract is broken, and the changed behavior ships with automated tests in the same PR.
- 🟡 `priority-medium` — keep. The resume addressed review feedback without widening scope, so the original urgency assessment stands.
- 🟢 `risk-low` — keep. The blast radius did not grow; the only executable change added since the last review is an `export` keyword on a module-private constant in a repo-local dev script.

I am deliberately **not** requesting `needs-qa`: the resume introduced no user-facing behavior. And per the pipeline rules, `qa-approved` is never applied by an authoring skill — it is earned by manual QA.

Paul-Mlodochowki:
🤖 `om-auto-continue-pr` completed. Status: complete. Lock released.

All four review findings addressed and verified; every Progress step in `.ai/runs/2026-08-14-issue-4738-i18n-missing-ui-locale-keys.md` is now `- [x]`. The full validation gate is green on the local runner apart from one pre-existing `packages/shared` failure this branch cannot cause. The `in-progress` label was never applied (no triage permission), so the assignee plus this comment are the claim signals; the PR now needs a maintainer to flip `changes-requested` → `review` and to submit an approving review, since GitHub rejects self-approval.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T09:44:16Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Re-review: fix(create-app): sync template locale dictionaries with the app (#4738)

Re-review of head `6f610e4d1`, following my `CHANGES_REQUESTED` review of `8a3abe85` on 2026-08-14. All four items from that review are addressed, and I verified each one against the code rather than against the resume summary.

## 🎯 Summary

The change under review is unchanged in substance: `'i18n'` joins `SYNC_FOLDERS` in `scripts/template-sync.ts`, the five template dictionaries are resynced to their `apps/mercato/src/i18n/` counterparts, and a parity test keeps the two trees from drifting again. What the resume added is the strength the gate was missing plus the documentation the new obligation needs.

The delta since my last review is five files — the two `AGENTS.md` files, the parity test, the `template-sync.ts` docstring and export, and the tracking run document. I re-checked all four previous findings and mutation-tested the strengthened gate to confirm it is a real gate rather than a rename:

- **Byte-identity is now actually asserted.** `template-i18n-parity.test.ts:87-90` reads both sides as `Buffer`s and short-circuits on `appBytes.equals(templateBytes)`; the key/value diff moved into `describeDrift()` and survives purely as the failure message. I reindented `template/src/i18n/en.json` to four spaces — a mutation the previous version passed green — and the case now fails with `en.json (app 88ebe83df262 vs template 197ef6913707): the parsed dictionaries are equal, so the two files differ only in formatting (key order, indentation or trailing newline)`. That is the precise diagnosis, not just a red X.
- **The scope case imports rather than regex-matches.** `SYNC_FOLDERS` is exported from `scripts/template-sync.ts:32` and imported at `template-i18n-parity.test.ts:7`, matching the sibling `template-example-module-parity.test.ts` pattern I pointed at. Narrowing the tuple back to `['app', 'components', 'lib', 'modules']` fails the case with `Current scope: app, components, lib, modules.` appended — better output than the old regex could produce.
- **The import is side-effect free and does not regress the sibling.** `template-sync.ts` guards its CLI behind `isDirectInvocation()` with a comment saying so explicitly, so importing the constant does not run the sync; `template-example-module-parity.test.ts` still passes 4/4 after the export change.
- **Both instruction files now record the obligation.** The root `AGENTS.md` Task Router row names `apps/mercato/src/i18n/**` and `yarn template:sync:fix`; `packages/create-app/AGENTS.md` rule 5 names the template counterpart, and new Template Sync Checklist entry 11 goes further than I asked by recording *why* the test is the only guard — the `scripts/i18n-check-*` gates all ignore `create-app/template/**`.
- **`Closes #4738` no longer drops the third request.** #5304 is filed for the `showQueryTime` switch and referenced in the PR body, with #5315 following it up on the default. The issue can now close honestly on merge.
- **The docstring is current.** `scripts/template-sync.ts:4` names the `i18n` folder and the locale dictionaries.

I also checked the one thing that green-at-head cannot tell you: whether this survives contact with current `develop`. `develop` moved 15 commits since the merge base and #5030 added `api.errors.rateLimitUnavailable` to both the app *and* the template dictionaries. Because it touched both sides symmetrically, the merge is clean and the mirror holds — I merged `origin/develop` into the head locally and all five locales are still `cmp`-identical, with the parity test 3/3 and `tsx scripts/template-sync.ts` reporting zero drift on the merge result. That local merge was never pushed.

## Verdict

✅ **approve** — no blockers, no majors, no minors, no nits. Every item from the previous review is closed, each verified independently, and the gate is measurably stronger than what I asked for: it now fails on formatting-only drift, which is the exact hole I flagged. Two observations below are for maintainer awareness, not findings, and neither should hold the merge.

## 🧪 Validation Gate

**CI is still running on this head at the time of approval.** Branch protection plus the repository's merge gates hold the actual merge; this approval covers the code, not a green run. Checks still pending: `test` and the 15 `ephemeral-integration` shards. A follow-up comment will report the CI outcome.


**Validation source: GitHub checks** (in progress, no failures) **plus narrow local verification of the merge result.**

CI is genuinely running this time — the fork's workflow runs were approved, unlike at `8a3abe85` where no `ci.yml` job had started. Everything that has reported is green and nothing is failing:

| Check | State |
|-------|-------|
| `prepare` | ✅ SUCCESS |
| `lint` | ✅ SUCCESS |
| `ds-lint` | ✅ SUCCESS |
| `scope` | ✅ SUCCESS |
| `audit-scope` | ✅ SUCCESS |
| `license/cla` | ✅ SUCCESS |
| `test` | ⏳ IN_PROGRESS |
| `ephemeral-integration` (15 shards) | ⏳ IN_PROGRESS |
| `docker-build`, `mutate`, `comment`, `audit`, `[code]smith` | ⏭️ SKIPPED |

Branch protection reports no required-check set for `develop` (HTTP 404), so every reported check is treated as required. `test` — the job that runs `yarn workspace create-mercato-app test`, and therefore the new parity gate — is still pending, so I did not run a local substitute for it; per this repository's checks-first rule, a pending check is disclosed and waited on by branch protection, not pre-empted. The local commands below are scoped to what CI at the PR head cannot tell me, namely the state of the head merged with current `develop`:

| Command | Status | Notes |
|---------|--------|-------|
| `node --test template-i18n-parity.test.ts` | ✅ PASS (3/3) | Run on the head **merged with `origin/develop`**, the state that will land. |
| `node --test template-example-module-parity.test.ts` | ✅ PASS (4/4) | Regression check on the new `SYNC_FOLDERS` export. |
| `tsx scripts/template-sync.ts` | ✅ PASS | Zero drift on the merge result, across synced folders, root files and package dependencies. |
| `cmp` on all five locale pairs | ✅ PASS | Byte-identical before *and* after merging `origin/develop`. |
| `node scripts/check-agents-md-budget.mjs` | ✅ PASS (exit 0) | Root `AGENTS.md` at 31,187 / 31,232 bytes. See observation 2. |
| `yarn typecheck` | ⏭️ not run | The review worktree has no dependency install (`@types/node` unresolvable), so a local run would report an environment artifact rather than a real result. CI's `test` job covers it. |
| `yarn build:packages`, `yarn generate`, `yarn build:app` | ⏭️ not run | No compiled surface in the diff, and no auto-discovery file changed. Covered by CI. |

Mutation-testing the gate, since a gate that cannot fail is not a gate:

| Mutation | Result |
|----------|--------|
| Reindent `template/src/i18n/en.json` to 4 spaces (formatting only) | ✅ correctly fails, naming formatting as the cause |
| Drop `'i18n'` from `SYNC_FOLDERS` | ✅ correctly fails, printing the narrowed scope |
| Both reverted | ✅ back to 3/3 |

## Findings

None. Both items below are observations for maintainer awareness rather than findings, and neither affects the verdict.

**Observation 1 — the parity test is excluded from the package's own typecheck.** `packages/create-app/tsconfig.json` sets `exclude: ["**/*.test.ts"]`, so neither the new `.ts`-extension import specifier nor the `SYNC_FOLDERS as readonly string[]` widening cast is ever type-checked; the file is only ever transpiled by `tsx` at run time. This is pre-existing repository structure that the sibling `template-example-module-parity.test.ts` already relies on, not something this PR introduced, and the runtime behavior is proven by the mutation tests above — so it is out of scope here. Worth knowing if anyone later tightens that `exclude`.

**Observation 2 — the root `AGENTS.md` instruction budget is now effectively spent.** After this PR the file sits at 31,187 of its 31,232-byte limit: **45 bytes free**. The check passes, and the growth is the direct consequence of the documentation I asked for in the previous review, so it is not a mark against this change. But the next contributor who needs to add a Task Router row will have to reclaim space first, and it would be reasonable for a maintainer to plan that trim rather than discover it in a failing gate.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path. The one new export, `SYNC_FOLDERS` in `scripts/template-sync.ts`, is purely additive and lives in a repo-local dev script, not a published package.
- [x] No function signature changed in a breaking way. `readLocale` changed from `(root, file)` to `(bytes)`, but it is module-private to the test file and has no other caller.
- [x] No CLI command or flag renamed or removed, and no machine-parsed output format changed.
- [x] The contributor-facing contract this PR creates — `apps/mercato/src/i18n/**` is a byte-exact mirrored surface — is now documented in both instruction files, which was the previous review's finding 2 and is the one contract change here.

## 🧪 Test Coverage

The gate does what its name says now, which was the substance of my previous review. Case 2 compares raw bytes and only parses in order to explain a failure, so the previously-passing formatting-only divergence fails; case 3 asserts against the imported constant instead of a regex over source text, so a reformatted multi-line declaration can no longer slip past it, and the failure message prints the actual scope it found. Case 1 is unchanged and still catches a locale file present on only one side.

The `describeDrift()` extraction preserved the diagnostic quality that made the original message worth keeping, and added a sha256 prefix per side so a formatting-only failure is still identifiable. Its `JSON.parse` can only run once the bytes already differ, so a malformed template dictionary would surface as a throw rather than a friendly message — the same behavior as before the refactor, and case 1 covers the missing-file variant with a good message, so this is not a regression.

Placement remains correct: `ci.yml` runs the `create-mercato-app` suite unconditionally, which is what a mirror gate for a PR touching only `apps/mercato` requires, since the turbo `--filter=[base]...` step would not select that package. No test is missing.

pkarw:

Re-review of `6f610e4d1` closes all four findings from my 2026-08-14 `CHANGES_REQUESTED` review, each verified against the code rather than the resume summary: the parity gate now compares raw bytes (mutation-tested — a 4-space reindent of `template/src/i18n/en.json`, which the previous version passed, is correctly rejected with formatting named as the cause), the scope case imports `SYNC_FOLDERS` instead of regex-matching source, both `AGENTS.md` files record `apps/mercato/src/i18n/**` as a mirrored surface, #5304 tracks the `showQueryTime` request so `Closes #4738` no longer drops it, and the docstring is current. No new findings — no blockers, majors, minors or nits.

I additionally checked what green-at-head cannot show: `develop` has moved 15 commits, and #5030 added `api.errors.rateLimitUnavailable` to the app *and* template dictionaries symmetrically, so a local merge of `origin/develop` into this head stays clean and all five locales remain `cmp`-identical, with the parity gate 3/3 and `template:sync` at zero drift on the merge result. That merge was local only and was never pushed to this branch.

Two observations recorded in the review body for maintainer awareness, neither a finding and neither blocking: the parity test is excluded from `packages/create-app`'s own typecheck (pre-existing, shared with the sibling gate), and the root `AGENTS.md` now sits at 31,187 of its 31,232-byte limit — 45 bytes free, passing, but the next Task Router addition will need a trim first.

`autofix: skipped (not my PR — re-run with --autofix to fix it here)`. Nothing needed fixing.

Pipeline label moved to `merge-queue`; `skip-qa` stands, so the QA-approval gate does not apply. `in-progress` was swapped for `ci-monitoring`: `test` and the 15 `ephemeral-integration` shards were still running when the verdict was submitted, and everything reported so far is green with nothing failing. I will follow up with the CI result. The merge still waits on those checks.

pkarw (changes requested):
# 🔍 Re-review (CI follow-up): fix(create-app): sync template locale dictionaries with the app (#4738)

**This review supersedes my approval of `6f610e4d1` posted earlier today.** CI settled after that approval and the `test` check went red on a failure this PR causes. The approval no longer holds, and I am correcting it here rather than leaving a green review standing over a red required check.

## 🎯 Summary

Everything in my approving review still stands on its own terms: all four findings from the 2026-08-14 pass are genuinely closed, the byte-identity gate is real and mutation-tested, and the change merges cleanly with current `develop` with the mirror intact. One new blocker replaced them, and CI found it because I did not.

The new gate is not registered in this repository's cross-package test registry, so the gate can be filtered out of CI on precisely the pull requests it exists to protect.

## Verdict

❌ **request changes** — one Blocker. The fix is a single registry entry, which I have verified locally; details and the exact patch below. Nothing else in the diff needs to change, and my earlier assessment of the rest of the PR is unchanged.

## 🧪 Validation Gate

**Validation source: GitHub checks** — now settled, and this is a real failure rather than a flake or an infrastructure artifact.

| Check | State |
|-------|-------|
| [`test`](https://github.com/open-mercato/open-mercato/actions/runs/31804467869/job/95347899798) | ❌ **FAILURE** — 587 tests, 585 pass, **1 fail** |
| `ephemeral-integration` (all 15 shards) | ✅ SUCCESS |
| `prepare`, `lint`, `ds-lint`, `scope`, `audit-scope`, `license/cla` | ✅ SUCCESS |
| `docker-build`, `merge-coverage`, `mutate`, `comment`, `audit`, `[code]smith` | ⏭️ SKIPPED |

The single failure, from `gh run view --log-failed`:

```
✖ no test that audits other packages is left unclassified
  test at scripts/__tests__/repo-wide-guards.test.mjs:87:1

  AssertionError [ERR_ASSERTION]: These tests read files outside their own package, so CI's
  turbo filter can skip them on a PR that changes those files. Add each to REPO_WIDE_GUARDS
  in scripts/repo-wide-guards.mjs so it runs unfiltered, or to CROSS_PACKAGE_EXCEPTIONS with
  a reason:
  packages/create-app/src/lib/template-i18n-parity.test.ts

  actual: [ 'packages/create-app/src/lib/template-i18n-parity.test.ts' ]
  expected: []
```

## Findings

### 🚨 Blocker

**1. `packages/create-app/src/lib/template-i18n-parity.test.ts` is not registered in `scripts/repo-wide-guards.mjs`, so the new gate can be filtered out of CI on exactly the PRs it guards.**

`scripts/__tests__/repo-wide-guards.test.mjs:87` enforces a repository-wide invariant: any test that reads files outside its own package must be enumerated, either in `REPO_WIDE_GUARDS` (to run unfiltered) or in `CROSS_PACKAGE_EXCEPTIONS` with a stated reason. The new parity test qualifies twice over — it reads `apps/mercato/src/i18n/**` and it imports `scripts/template-sync.ts` — and it is in neither list, so the guard fails and `test` goes red.

This is not merely bookkeeping, and it is worth being precise about why. The registry exists to stop the turbo `--filter=[base]...` selection from skipping a cross-package test on a PR that changes the files it audits. A PR that touches only `apps/mercato/src/i18n/pl.json` does not select `create-mercato-app`, so without the entry the very drift this PR was opened to prevent could reappear without the new gate ever running — the same class of silent hole as the original `SYNC_FOLDERS` omission, one level up.

The precedent is exact and sits five lines away in the registry. `template-example-module-parity.test.ts` — the sibling I pointed you at in the previous review — is already registered:

```js
{
  path: 'packages/create-app/src/lib/template-example-module-parity.test.ts',
  reason: 'Already unfiltered — the "Check create-app template parity" CI step runs the whole create-mercato-app suite (#3779).',
},
```

Fix: add the matching entry to `CROSS_PACKAGE_EXCEPTIONS` in `scripts/repo-wide-guards.mjs`, directly after the sibling. `CROSS_PACKAGE_EXCEPTIONS` rather than `REPO_WIDE_GUARDS` is the right list, because the reason your PR body already gives is exactly the one the exception records — the "Check create-app template parity" step runs the whole suite unconditionally:

```js
{
  path: 'packages/create-app/src/lib/template-i18n-parity.test.ts',
  reason: 'Already unfiltered — the "Check create-app template parity" CI step runs the whole create-mercato-app suite (#3779).',
},
```

I applied exactly this locally on your head merged with current `develop` and re-ran `node --test scripts/__tests__/repo-wide-guards.test.mjs`: **16 pass, 0 fail**, with the `no test that audits other packages is left unclassified` case green. Then I reverted it — this is a fork PR and I was not asked to push to it, so the change is yours to make. It is one commit.

## 📝 Correction to my earlier review

My approving review's Test Coverage section asserted that placement was correct because `ci.yml` runs the `create-mercato-app` suite unconditionally. That fact is true, and it is the very reason the exception entry is the right fix — but this repository requires it to be *recorded* in the registry the guard reads, and I checked the CI workflow without checking the registry. The pending `test` check was disclosed in that review and it is what caught this, which is the system working as intended; I am flagging my own gap rather than presenting this as a new development.

## 💥 Breaking Changes

Unchanged from my earlier review — no contract surface is broken. The one contributor-facing contract this PR creates is documented in both `AGENTS.md` files, and the registry entry above is the last piece of wiring that contract needs.

## 🧪 Test Coverage

Unchanged and still good, with the one caveat above. The three cases are well-targeted and mutation-tested: reindenting `template/src/i18n/en.json` to four spaces fails with formatting named as the cause, and narrowing `SYNC_FOLDERS` fails printing the scope it found. Once the registry entry lands, the gate will also be guaranteed to actually run on the PRs that matter, which is what makes those assertions worth having.

pkarw:

CI settled at 2026-08-17T10:04:45Z, inside the 40-minute follow-up budget. **`test` failed**, so the approval I posted earlier today no longer holds and I have replaced it with a `changes-requested` review.

| Check | Result |
|-------|--------|
| [`test`](https://github.com/open-mercato/open-mercato/actions/runs/31804467869/job/95347899798) | ❌ FAILURE — 587 tests, 585 pass, 1 fail |
| `ephemeral-integration` (15/15 shards) | ✅ SUCCESS |
| `prepare`, `lint`, `ds-lint`, `scope`, `audit-scope`, `license/cla` | ✅ SUCCESS |

**One failure, caused by this PR, deterministic — not a flake.** `scripts/__tests__/repo-wide-guards.test.mjs:87` requires every test that reads files outside its own package to be enumerated in `scripts/repo-wide-guards.mjs`. The new `packages/create-app/src/lib/template-i18n-parity.test.ts` reads `apps/mercato/src/i18n/**` and imports `scripts/template-sync.ts`, and is in neither `REPO_WIDE_GUARDS` nor `CROSS_PACKAGE_EXCEPTIONS`.

It matters beyond the red X: without the entry, the turbo `--filter=[base]...` selection can skip the new gate on a PR that touches only `apps/mercato/src/i18n/**` — the exact case the gate exists for. The sibling `template-example-module-parity.test.ts` is already registered, so the precedent is five lines away in the same file.

The fix is one entry in `CROSS_PACKAGE_EXCEPTIONS`, quoted in full in the review. I applied it locally on your head merged with current `develop` and `node --test scripts/__tests__/repo-wide-guards.test.mjs` goes **16 pass / 0 fail**, then reverted it — this is a fork PR and I was not asked to push to it, so the commit is yours.

Everything else I verified earlier still stands: all four findings from the 2026-08-14 review are genuinely closed, the byte-identity gate is real (mutation-tested against formatting-only drift and a narrowed `SYNC_FOLDERS`), and the head merges cleanly with current `develop` with all five locales still byte-identical.

Pipeline label moved from `merge-queue` to `changes-requested`; `ci-monitoring` removed, since the follow-up this label promised is now delivered and no further CI follow-up will come from this agent. Re-request review once the entry is pushed and the automation can pick it up from the new head.

pkarw:
Thanks @Paul-Mlodochowki — CI turned up one actionable item after my approval, so I'm handing this PR back to you for the next pass. It is a single entry in `scripts/repo-wide-guards.mjs`, quoted in full in the review and verified locally at 16 pass / 0 fail. When the update is pushed, re-request review and the automation can pick it up from the latest head.

PR #4942, opened 2026-08-04
fix(crud): terminate fetch-all pagination loops on short pages, fail closed at the page ceiling (#4552 Phase 1)

jtomaszewski:
## Summary

Phase 1 of the capped-list-count spec merged in #4552 (`.ai/specs/2026-07-27-list-count-strategies.md`). Six fetch-all pagination loops used a server-reported `total`/`totalPages` as their loop bound. That is wrong today (a count-terminated loop misbehaves when rows are inserted or deleted mid-loop; one of the six could already spin **forever**), and it becomes silent data loss the moment Phase 2 makes `total` a bounded floor (`OM_LIST_COUNT_CAP`).

Every converted loop now uses one shape, per the spec's §Short-page termination (in-repo reference: `TenantSelect.tsx`): collect until a page comes back shorter than requested, and **fail closed** — throw an `[internal]`-prefixed error — at a hard page ceiling instead of returning a partial export or a partial id set.

## Changes

- `packages/shared/src/lib/crud/factory.ts` — CRUD CSV/XLSX/JSON export loop: loop gate is now "first page came back full" instead of `total > exportItems.length`; `EXPORT_MAX_PAGES = 1000` throws rather than truncating.
- `packages/core/src/modules/entities/api/records.ts` — custom-entity records export loop: identical treatment.
- `packages/core/src/modules/customers/api/utils.ts` — `findMatchingEntityIdsWithQueryEngine`: **fixes a latent infinite loop.** `ids` is a `Set`, so duplicate ids across pages kept `ids.size` permanently below an inflated `total`; the only escape was an empty page. Signature unchanged; feeds the advanced filters on `customers/api/{deals,people,companies}`.
- `packages/core/src/modules/catalog/widgets/injection/product-bulk-delete/widget.ts` — `fetchFilteredProductIds`: previously had **no** page ceiling and **no** empty-page break; a partial id set would turn "delete all filtered" into "delete the first N" with no signal.
- `packages/core/src/modules/customers/components/detail/MiniWeekCalendar.tsx` — legacy-activities loop extracted to `legacyActivities.ts` (`loadLegacyActivitiesInRange`) so it is unit-testable; the domain date-range break is preserved.
- `packages/core/src/modules/customers/utils/phoneDuplicates.ts` — `total`-derived early exit replaced with a short-page check; the deliberate `MAX_PAGES = 3` best-effort bound stays (duplicate *hint*, partial acceptable by design).
[...]

PR #4552, opened 2026-07-27
docs(specs): capped list count — rebuild the CRUD list COUNT so a bound actually binds

jtomaszewski:
## Summary

Every CRUD list request pays an unconditional `COUNT` round trip. An exact total over arbitrary filters is `O(matching rows)` by arithmetic — no index, engine or search backend removes that cost. On a ~1.4M-row table that is ~1.4M units of work on **every keystroke** of a filter input; measured on staging, even a covering `(organization_id, tenant_id, status) WHERE deleted_at IS NULL` index still yields a parallel sequential scan for the non-selective case.

This PR adds **the specification only** — no implementation. It proposes bounding the count so the database stops at `cap + 1` matching rows: below the cap the total stays exact, at the cap the engine reports `total: cap` plus a new optional `totalIsCapped` flag so the UI renders "10 000+".

> **Revised after review** ([changes requested](https://github.com/open-mercato/open-mercato/pull/4552#pullrequestreview-4791266693)). The original design bounded the count with a one-line `countBuilder.limit(cap + 1)`. That is a no-op, and the review was right to block on it. The spec now rebuilds the count query instead. See *What changed* below.

## The core of the design

**A `LIMIT` cannot bound a count that is derived from the display query.**

Today the count is built by taking the full data query and stripping `SELECT`/`ORDER BY` (`packages/shared/src/lib/query/engine.ts:912-923`). That inherits the custom-field joins, the entity-extension joins, the `array_agg`/`bool_or` projections and the `GROUP BY` — and then needs `count(distinct)` to undo the fan-out those joins cause. Two things follow:

- `LIMIT` on `SELECT count(*) …` bounds the single **aggregate output row**, not the scan. Zero effect.
- `LIMIT` on a subquery that keeps the `GROUP BY` is also blocked: PostgreSQL plans `Limit → HashAggregate → Scan`, and **`HashAggregate` is a blocking node** — it consumes its entire input before emitting a row.

So the count query is **rebuilt rather than derived**: base table + scope + filters only, every *filtering* join expressed as a correlated `EXISTS` semi-join, every *projection-only* join dropped. A semi-join cannot multiply base rows, so there is no `DISTINCT`, no `GROUP BY`, and no barrier between the `LIMIT` and the scan:

```sql
SELECT count(*) FROM (
  SELECT 1 FROM orders o
  WHERE <scope> AND <base filters>
    AND EXISTS (SELECT 1 FROM custom_field_values v
[...]

PR #5240, opened 2026-08-12
fix(sales): stop dropping address-snapshot keys the editor cannot show

kamwro:
## Summary

A sales document's address snapshot is a free-form JSON record — `addressSnapshotSchema` is
`z.record(z.string(), z.unknown())` — so an integration can put keys on it that the address editor has
no field for. An ERP freezing a tax id or a phone number onto the document address is the motivating
case, and those values **do** persist today.

They survive only until a human opens the document and presses save. `normalizeAddressDraft`
(`AddressesSection.tsx:78-99`) rebuilds the entire snapshot from the twelve fields the editor models,
and that result becomes the whole payload. Every other key is dropped on the way out. Nothing errors,
nothing warns, the request returns 200 — the value is simply gone on the next read.

The helper now takes the previous snapshot and merges it back for exactly the keys it does not own.
The twelve editable keys are listed explicitly, so a field the editor *can* show still follows the
editor — clearing a city still clears it.

Clearing the *whole* address is the case worth calling out, because the first revision of this PR got
it wrong. `isPrimary` is a boolean, so it lands in the rebuilt record whatever the user does — which
meant "every visible field blanked" still looked non-empty, and the merge-back then reattached the
hidden keys. A cleared address kept an invisible tax id. Emptiness is now decided on editor-owned
*string* content alone, so blanking the form returns `null` and "remove this address" removes all of
it. Caught in review by @haxiorz.

This is Phase 0 of the spec merged in #5197, and the one part of it that stands alone: the bug is
live on `develop` today whether or not address contact fields are ever built.

**Scope.** Only the document detail save path. `SalesDocumentForm.tsx:438-459` holds a functionally
identical copy of the helper, but that component is mounted solely by
`backend/sales/documents/create/page.tsx` and submits through `createCrud` — it builds a document
that does not exist yet, so there is no prior snapshot for it to drop and a merge-back there would be
permanently inert. Collapsing the two copies is worth doing on its own justification.

Reviewed and approved by @jtomaszewski before upstreaming.

## Changes

- `AddressesSection.tsx` — `normalizeAddressDraft` accepts the previous snapshot and merges back every
  key outside the twelve the editor owns; the editable key set is named explicitly
- `AddressesSection.tsx` — emptiness is decided on editor-owned string content, so a fully blanked
  form clears the snapshot instead of preserving unowned keys behind `isPrimary: false`
- `AddressesSection.test.tsx` — an unowned key survives a save; blanking every visible field clears the
  snapshot to `null`

[...]

PR #5228, opened 2026-08-12
feat(query): cap list COUNT at OM_LIST_COUNT_CAP via rebuilt count queries (#4552 Phase 2)

kamwro:
## Summary

Implements Phase 2 of the capped list count spec (#4552; Phase 1 landed as #4942).

Every CRUD list request pays an exact `COUNT` that is `O(matching rows)` — ~100–170 s on a 1.4M-row table. The spec's review established that a `LIMIT` on the existing count is a no-op (blocked by `HashAggregate`), so the count query must be rebuilt, not bounded.

This rebuilds the count at all four sites (`BasicQueryEngine`, hybrid optimized + full-shape, doc storage): scope + filters only, cf/doc filters as correlated `EXISTS` semi-joins, projection joins dropped, `LIMIT cap+1` probe, outer `count(*)`. At the cap: `total: cap`, `meta.listCountCapWarning`, `totalIsCapped: true` on payloads. In the hybrid engine, rowset predicates evaluate inside one seed-rowed `EXISTS` built from the display query's own join builders, so filter semantics are unchanged by construction.

## Changes

- Rebuilt count query at all four count sites — `packages/shared/src/lib/query/engine.ts`, `packages/core/src/modules/query_index/lib/engine.ts`
- New `OM_LIST_COUNT_CAP` resolver (default `10000`, `0` = permanent kill switch) — `packages/shared/src/lib/query/count-cap.ts`
- Encrypted-sort truncation decoupled from `total` — it now runs its own `cap+1` probe instead of reading a cappable value
- `totalIsCapped` on CRUD payloads and the OpenAPI schema — `packages/shared/src/lib/crud/factory.ts`, `packages/shared/src/lib/openapi/crud.ts` — plus the two direct routes that actually consume engine totals (`customers/api/deals/map`, `entities/api/records`)
- WMS loader loops stop on a short page instead of using `totalPages` as a loop bound — `packages/core/src/modules/wms/components/backend/inventoryMutationLoaders.ts`
- `OM_LIST_COUNT_CAP` surfaced in system status (5 locales), both `.env.example`s, and `UPGRADE_NOTES.md` (on-by-default is a declared minor breaking change)

**Audit deltas vs the spec** (drift since its `ba2cd5d` line audit): attachments/todos/interactions routes now derive totals from `em.findAndCount`/in-memory pagination — cap-immune, left unchanged; same for workflows/dictionaries `hasMore`. Only deals/map and entities/records still consume engine totals and got the flag.

## Specification

[...]

PR #5274, opened 2026-08-13
fix(ui,customers): terminate "load more" on a short page, not on a reported total

kamwro:
## Summary

Five incremental-loading surfaces gate their "Load more" affordance on a server-reported `total`/`totalPages`. Whenever that total under-reports the result set, the button disappears and the remaining rows become unreachable — the same defect class Phase 1 (#4942) converted for server-side loops, now applied to the client.

An under-reporting total isn't hypothetical: rows inserted between requests cause it, and a capped list count (#4552) causes it by design. These fixes stand on their own merits and need no capping-related flag.

## Changes

- `AttachmentsSection` — `page < totalPages` replaced by page fullness; page size lifted to a named constant so guard and request can't drift.
- `LinkedEntitiesField`, `ParticipantsField` — same shape; their `totalPages` derivations (including a `ceil(total / pageSize)` fallback) drop out entirely.
- `DealsSection` — already had page fullness as a *fallback*; it now always decides, since preferring `nextPage < totalPages` is what hid deals.
- `AssignRoleDialog` — `users.length >= totalUsers` replaced by page fullness, the most direct form. Its "Showing X of Y team members" label is deliberately untouched: with an exact total it's accurate, and making it cap-aware needs the count flag threaded through the staff API.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**

N/A — a behavioural bug fix across five call sites of one pattern; no module design changes.

## Testing

Each of the five files gains two tests: the button survives a deliberately under-reporting total, and it disappears on a short page despite a large one.

```
yarn turbo run test        # ui 1863/1863, core 9624/9624
yarn turbo run typecheck
yarn tsx scripts/i18n-check-sync.ts
```

## Checklist

- [x] I updated documentation, locales, or generators if the change requires it — no locale key or generated artifact changes.
- [ ] I added or updated integration tests in `.ai/qa/tests/` — the behaviour is pinned by unit tests at all five call sites; reproducing it end-to-end needs a server total that disagrees with the returned page.
- [ ] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable) — N/A per above.
- [ ] QA routing set.

The three label boxes are unticked because this account has read-only permission here; intended values are in a separate comment for a maintainer to apply. The CLA box is left for the author to tick.

### Design System Compliance
- [x] No hardcoded status colors — no color class added.
- [x] No arbitrary text sizes — no typography change.
- [x] Empty state handled — untouched on all five surfaces.
- [x] Loading state handled — untouched.
- [x] `aria-label` on icon-only buttons — no button added or removed.
- [x] Uses existing DS components — the change is to loop-termination logic only, no markup swap.

## Scope boundary

The pattern is **not** fully exhausted by this PR. Two `page < totalPages` load-more gates remain in the same directory family and are deliberately deferred to #5347, which carries the line numbers and the served-count analysis for both:

- `customers/components/detail/hooks/usePersonTasks.ts:245,479`
- `customers/components/detail/EntityTagsDialog.tsx:1253`

Neither has an existing test harness, and this PR's fixtures had to be corrected once already to model what its endpoints actually emit — so those two get their own pass rather than riding along here. `hasMoreFromPage` makes the conversion mechanical for whoever takes it.

Also verified and deliberately excluded: `ActivityHistorySection.tsx:300` ORs `totalPages` onto a canonical cursor signal, so an under-reporting total can only add "there may be more" there, never hide rows; `useVersionHistory.ts:81` is already page-fullness based.

## Linked issues

Relates to #4942 (the server-side counterpart) and #4552. Follow-up: #5347.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5274"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789225488&installation_model_id=432243&pr_number=5274&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5274&signature=f4683d9e6ac378b474249a0786bbe8f1ed75ee2499d9eb62c997e70d40752056"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


kamwro:
**Metadata intent** — this account has `read` here, so none of it can be applied from our side.

| Label | Why |
|---|---|
| `bug` | Fixes rows becoming unreachable behind a vanished "Load more". |
| `needs-qa` | User-visible, manually exercisable on five surfaces. |
| `priority-medium` | Data is reachable by other routes; no corruption. |
| `risk-low` | Loop-termination logic only; no schema, route or contract change. |
| `partner-request` | FSH partner contribution. |

@jtomaszewski please apply `partner-request` — this is an FSH partner contribution and this account is not in `.github/certified-partners.yml`, so the `/label` workflow ignores it.

The QA gate holds the merge until a maintainer adds `qa-approved`. Reviewer intent: a maintainer of the `customers` / `ui` packages.

kamwro:
@jtomaszewski could you apply `partner-request` here?

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-14T07:50:53Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review — Changes requested

**Verdict:** changes requested — 1 blocker, 1 major, 3 minors, 2 nits.
**Head reviewed:** `94b04d64ff3b7c12ef7f4e95ee7c31fe36d9c222` · base `develop` (4 commits behind, `MERGEABLE`, no textual or semantic conflict).
**Validation source:** local fallback. No CI has ever run on this head — the only checks are `license/cla` (success) and `[code]smith` (skipped), fork workflows need maintainer approval, and `develop` declares **no required status checks** at all. Commands run below.

The diagnosis is right and I verified it rather than taking it: a "Load more" gated on a server-reported total does strand rows when that total under-reports, and four of the five conversions are correct. But the fifth — `AttachmentsSection` — is the one call site whose endpoint doesn't behave the way short-page termination assumes, and there it replaces "some rows unreachable" with "every row duplicated, unboundedly". That contradicts the PR's framing of five call sites of one pattern: they are five call sites of one pattern against **two different server paging contracts**.

---

## 🔴 Blocker

### B1 — `AttachmentsSection`: unbounded duplicate rows and a "Load more" that never terminates, whenever the attachment count is an exact multiple of 24

`/api/attachments` **clamps the requested page to the last page** — for the offset *and* for the `page` it echoes back:

```ts
// packages/core/src/modules/attachments/api/route.ts:215-216
const totalPages = usePaging && total !== null ? Math.max(1, Math.ceil(total / currentPageSize!)) : null
const pageOffset = usePaging ? (Math.min(currentPage!, totalPages!) - 1) * currentPageSize! : undefined
// :259
page: Math.min(currentPage!, totalPages!),
```

So a request past the end does not come back short — it comes back **full, re-serving the last page**. Short-page termination is built on the opposite assumption. Two further properties make `AttachmentsSection` the only one of the five that turns this into visible corruption:

- it is the only one that appends **without deduping** — `setItems((current) => (replace ? nextItems : [...current, ...nextItems]))` (`packages/ui/src/backend/detail/AttachmentsSection.tsx:86`), where the other four merge through a `Map` keyed by id;
- it is the only one that takes its next page from the server's echoed `page` — `setPage(typeof payload.page === 'number' ? payload.page : targetPage)` (`:87`) — which is clamped, so `page` can never advance past the last page and every subsequent click re-requests the same one.

Rows are keyed `key={item.id}` (`:276`), so the duplicates are also duplicate React keys.

Sequence for a record with exactly 24 attachments (`PAGE_SIZE = 24`):

1. page 1 → 24 items, `totalPages: 1` → `hasMore = 24 >= 24 = true` → button shown. On `develop`, `page(1) < totalPages(1)` is `false` → button correctly hidden.
2. click → `load(2)` → route clamps `min(2, 1) = 1` → returns **the same 24 rows**, echoes `page: 1`.
3. append → 48 rows, all 24 duplicated; `setPage(1)`; `hasMore` still `true`.
4. every further click repeats step 2 and adds 24 more duplicates. There is no terminating state.

**Evidence.** A probe that emulates the route's clamping exactly (offset *and* echoed `page` clamped to `totalPages`), run against this head and against `origin/develop`'s copy of `AttachmentsSection.tsx` with nothing else changed:

| attachments | `origin/develop` | this head |
|---|---|---|
| 24 | pages requested `[1]` · 24 rows · button absent | pages requested `[1,2]` · **48 rows** · 2 copies of `File 1.pdf` · **button still present** |
| 48 | pages requested `[1,2]` · 48 rows · button absent | pages requested `[1,2,3,3]` · **96 rows** · 3 copies of `File 48.pdf` · **button still present** |

`[1,2,3,3]` is the tell: page 3 is requested twice because the echoed page never advances, which is why the loop has no exit.

**Suggested fix** — have the component notice it asked past the end, which keeps short-page termination and is robust to the clamp:

```ts
const returnedPage = typeof payload.page === 'number' ? payload.page : targetPage
const servedRequestedPage = returnedPage >= targetPage
setItems((current) => (replace ? nextItems : servedRequestedPage ? [...current, ...nextItems] : current))
setPage(returnedPage)
setHasMore(servedRequestedPage && nextItems.length >= PAGE_SIZE)
```

Deduping the append by `id` is worth doing regardless — it is what keeps the other four call sites from showing this class of defect at all.

---

## 🟠 Major

### M1 — no test models how a real endpoint answers a request past the end, and the `AttachmentsSection` pair asserts a payload its route cannot emit

All ten new tests model the terminal response as a **short page**. That is true for the four query-engine/staff-backed call sites and false for attachments, so the suite is structurally unable to see B1. Concretely, the fixture at `packages/ui/src/backend/__tests__/AttachmentsSection.test.tsx:124-127` is:

```ts
{ items: makeItems(2, 24), total: 3, page: 2, pageSize: 24, totalPages: 1 },
```

`page: 2` with `totalPages: 1` is unreachable from `route.ts:259`, which returns `Math.min(currentPage, totalPages)` — the route can never echo a page above `totalPages`. The one realistic terminal case for this endpoint (count an exact multiple of `pageSize`) is untested, and it is the broken one.

This is why the gate is green on a broken surface: on this head `packages/ui` passes 229 suites / 1863 tests and `core` `src/modules/customers` passes 227 suites / 1376 tests. Please add a case per call site that answers the way that call site's own endpoint does — for attachments, clamped; for the others, empty past the end.

---

## 🟡 Minors

### m1 — the staff call sites measure page fullness on a client-deduped array

`AssignRoleDialog.tsx:161` (`result.items.length >= ASSIGNABLE_STAFF_PAGE_SIZE`) and `ParticipantsField.tsx:64` (`members.length >= PAGE_SIZE`) both measure an array that `fetchAssignableStaffMembersPage` has already deduped by `userId` (`assignableStaff.ts:67-121`). A deduped length shorter than the served page would terminate early — reintroducing exactly the bug this PR fixes.

There is **no live bug**: `/api/staff/team-members/assignable` also dedupes by `userId` *before* slicing (`route.ts:196-206`), so the client dedupe is a no-op and the two lengths agree. I verified that before flagging it. But the guard now silently depends on an implementation detail of an optional, ejectable module — the same module whose absence the helper already defends against with a 404 branch. Prefer measuring the raw response length (the helper would need to expose it alongside `items`), or state the dependency where the guard is set.

### m2 — `hasMore` is left out of `AssignRoleDialog`'s reset

The open/close effect (`AssignRoleDialog.tsx:91-119`) resets `users`, `totalUsers`, `currentPage`, `activeTeam`, `loadError` and the request sequence, but not the new `hasMore`. Reopening with a preselected role therefore renders one commit carrying a stale "Load more" from the previous session, before the first fetch sets `loading`. `handleLoadMore` guards on `loading`, so it is cosmetic — but the old guard (`users.length < totalUsers`) was false-by-construction after the reset and the new one is not. Reset it with its siblings in both branches.

### m3 — the sixth instance of the pattern is in the same directory and is not converted

`packages/ui/src/backend/detail/NotesSection.tsx` is append-style load-more — `setNotes((prev) => [...prev, ...pageResult.items])` (`:803`) — gated on a server-reported `totalPages`: `if (currentPage >= totalPages || isLoading) return` (`:790`) and the button's own `pagedMode ? currentPage >= totalPages : …` (`:1325`). It sits next to `AttachmentsSection` in the same folder and has the identical defect. Either convert it here or say why it is out of scope; "five incremental-loading surfaces" reads as exhaustive and it is six. (`PersonCompaniesSection.tsx:76` and `primitives/pagination.tsx` are next/prev affordances rather than load-more, so I am not counting those.)

---

## ⚪ Nits

### n1 — the attachments route is the odd one out and is the real root cause

Every CRUD list terminates correctly because the query engine computes an **unclamped** offset — `.offset((page - 1) * pageSize)` (`packages/shared/src/lib/query/engine.ts:1137`, and `.slice((page - 1) * pageSize, page * pageSize)` at `:1108`) — so a page past the end is empty. `/api/attachments` is the only list endpoint I found that clamps instead. Making it behave like the rest would fix B1 at the source and remove the need for a component-side check; worth a follow-up issue even if you take the component fix here.

### n2 — five copies of the rationale, four copies of `PAGE_SIZE`

The 3–4 line explanatory comment is near-duplicated across five files and `PAGE_SIZE` is redeclared in four. Defensible for a cross-cutting fix, but the five guards will drift; a small shared helper (`hasMoreFromPage(items, pageSize)` or a `usePagedLoadMore` hook) would keep them honest and give the rationale one home.

---

## Verified and deliberately not reported

- `LinkedEntitiesField:117` measures the **raw** `items` length before mapping and filtering — the correct measure, and the one m1 asks for elsewhere.
- `DealsSection` dedupes appended rows by id (`:539-544`) and advances its own `pageRef` instead of trusting an echoed page, so the clamping failure mode cannot reach it — and `/api/customers/deals` does not clamp anyway. Dropping `totalPages` as the preferred signal there is correct.
- Deals / companies / quotes / assignable-staff all return an empty page past the end, so those four call sites terminate correctly at the cost of one extra request per exhausted list — the accepted trade-off of this approach and consistent with the server-side counterpart in #4942.
- `>=` rather than `===` is safe against an endpoint returning more than `pageSize`.
- No contract surface is touched: no API response field, event id, public export, DI key, ACL feature, DB column or CLI flag changes, so `BACKWARD_COMPATIBILITY.md` is not engaged. No locale keys, no DS tokens, no `dark:` overrides, no arbitrary Tailwind values, no new `any`. `yarn generate` produces no tracked churn.

## Validation run (local fallback)

| Command | Result |
|---|---|
| `yarn build:packages` | 22/22 successful |
| `yarn generate` | clean, no tracked churn |
| `packages/ui` `yarn typecheck` | clean |
| `packages/core` `yarn typecheck` | clean |
| `packages/ui` `yarn test` | 229 suites / 1863 tests passed |
| `packages/core` `src/modules/customers` | 227 suites / 1376 tests passed |

All green — and, per M1, all blind to B1.

## Labels

`changes-requested`, plus the author's proposed `bug`, `needs-qa`, `priority-medium`. Risk raised from the proposed `risk-low` to **`risk-medium`**: the diff changes behaviour in `packages/ui/src/backend/detail/AttachmentsSection.tsx`, a shared component rendered on detail pages across modules, and B1 shows the blast radius is user-visible rather than isolated. `partner-request` is outside this repo's configured label taxonomy, so it stays with @jtomaszewski as you asked.

pkarw:
Label set to `changes-requested` because the re-review of `eba77d7f` found one blocker: the `mutate (shared, src/lib/pagination/load-more.ts)` check fails because the new shared helper ships with no test in its own package. All seven findings from the `94b04d64` pass are verified fixed.

| Label | Why |
|---|---|
| `changes-requested` | One blocker (a red check whose cause is in the diff). Pipeline labels are mutually exclusive, so this replaces any other pipeline state. |
| `bug` | Unchanged — rows becoming unreachable behind a vanished "Load more". |
| `needs-qa` | Unchanged. Six user-visible surfaces; `.tsx` files under `packages/ui/src/` and `components/`, so the automated-verification exemption does not apply. QA is requested, not performed — the `qa` label belongs to a QA reviewer, and `qa-approved` is not something a diff read can earn. |
| `priority-medium` | Unchanged — the data is reachable by other routes and nothing is corrupted. |
| `risk-medium` | Held at `risk-medium` rather than the originally proposed `risk-low`. The update widened the diff to six surfaces plus a new exported `@open-mercato/shared` helper, which is a larger blast radius than the first pass, not a smaller one. |
| `partner-request` | Outside this repo's configured label taxonomy; applied by @Frshy and left untouched. |

pkarw:

The blocker is the only thing standing between this and an approval; the four other conversions are correct as written. The suggested patch is in the review body, and the probe used to demonstrate it is reproducible from the table in B1 — it just emulates `route.ts:215-216`/`:259` (clamp the offset *and* the echoed `page` to `totalPages`) instead of the hand-written page fixtures, which is the substitution that makes the current tests pass while the surface is broken.

Note on process: this skill normally carries a fork PR forward onto a `carry/pr-5274-ready` branch and supersedes the original when review finds blockers. I deliberately did **not** do that here — you're actively working this PR, the fix is small and yours to make, and closing a live partner contribution in favour of a maintainer-owned branch isn't a call to make unprompted. Say the word if you'd rather I push the fix onto a carry-forward branch instead.

pkarw:

1 blocker, 1 major, 3 minors, 2 nits. Autofix iterations: 0 — deliberately skipped for this fork PR (rationale in the handoff comment above); the blocker is diagnosed with a reproducible probe and a suggested patch.

kamwro:
Pushed `eba77d7`. All seven items are addressed. B1 was right, and I verified it rather than taking it.

**B1.** `AttachmentsSection` now treats an echoed page below the one it requested as the end of the list, and dedupes its append by `id` like the other four. It still asks once past the end — that request is how short-page termination discovers the end, consistent with the server-side counterpart in #4942 — but the clamped answer is never appended and never advances the page.

**M1.** The `{page: 2, totalPages: 1}` fixture is gone; you were right that `route.ts:259` cannot emit it. The suite now drives a fake that clamps offset *and* echoed page exactly as the route does, and records every page requested. Reverting the source hunk fails precisely the two new cases and leaves the other three green:

| attachments | before | after |
|---|---|---|
| 24 | pages `[1,2]` · 48 rows · button present | pages `[1,2]` · 24 rows · button gone |
| 48 | pages `[1,2,3,3]` · 96 rows · button present | pages `[1,2,3]` · 48 rows · button gone |

Every other call site gained a case answering the way its own endpoint answers — empty past the end.

**m1.** `fetchAssignableStaffMembersPage` reports `servedCount` (raw response length, pre-dedupe) and both staff guards read it. You were right there's no live bug; the guard no longer depends on the staff route deduping before it slices. Two tests cover a full page that dedupes to one row still offering the next.

**m2.** `hasMore` joins the reset, both branches.

**m3.** Converted `NotesSection` and deduped its append. Its adapter hits `/api/customers/comments`, a `makeCrudRoute` list, so plain short-page termination is correct there — checked, not assumed. The description no longer says five.

**n2.** The rationale and the two obligations that come with it now live once, in `@open-mercato/shared/lib/pagination/load-more`.

**n1.** Component fix only. Making `/api/attachments` compute an unclamped offset changes a response other callers see, so it wants its own PR — issue text ready, but this account has `read` here, so it needs a maintainer to file or assign.

On the carry-forward branch: thanks for asking rather than doing it. Not needed — the fix was small and it's done.

I can't re-request review from this account (the API 404s on `read`), so please take this as the request.

Verification on this head: `turbo run typecheck` 22/22; `ui` 229 suites / 1868 tests; `core` 1236 / 9636; `shared` 174 / 1876; `i18n-check-sync`, `check:dep-versions`, `check:time-bombs:fail`, `agents:check-budget` all exit 0; lint 0 errors.

kamwro:
Filed as #5299.

kamwro:
Fixed on `eba77d7f`. B1 was real and worse than what the PR set out to fix — thank you for the probe, it's exactly the case my fixtures couldn't express.

`AttachmentsSection` now checks it was served the page it asked for (`returnedPage >= targetPage`) before appending, and dedupes the append by id like the other four. M1: each call site's tests now answer the way its own endpoint does — clamped for attachments (including the exact-multiple case), empty past the end elsewhere. m1: `fetchAssignableStaffMembersPage` exposes the raw served count so the guard no longer measures a deduped array. m2: `hasMore` resets with its siblings. m3: `NotesSection` converted — you were right that "five surfaces" read as exhaustive. n2: the guard and its rationale live in `hasMoreFromPage` (`packages/shared/src/lib/pagination/load-more.ts`), whose docstring records both obligations, including the clamping caveat. n1 is filed as #5299 rather than changed here, since altering the route's paging contract affects other consumers.

@pkarw one ask, on both this PR and #5228: **could you approve the workflow runs?** This account has `read` here, so CI has never executed on either head — the only checks are `license/cla` and a skipped `[code]smith`. They were approved on #5240 (same account, same fork), where the full suite ran green including all 15 integration shards, so it looks like the per-PR click rather than anything configuration-side. Both reviews name the missing validation gate as blocking, and it's the one part I can't unblock from here.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-17T09:55:55Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T09:55:55Z. Other auto-skills will skip this PR until the lock is released.

pkarw (dismissed):
# 🔍 Re-review: fix(ui,customers): terminate "load more" on a short page, not on a reported total

**Verdict:** changes requested — 1 blocker, 0 majors, 1 minor, 1 nit. All seven findings from the previous pass are genuinely fixed; the single blocker is new and arrived with the CI that finally ran.
**Head reviewed:** `eba77d7fc3e22036ee49f98225b8f8c27d988fd7` (previous review was `94b04d64`) · base `develop` · `MERGEABLE`, no conflict.
**Validation source:** GitHub checks. The workflow runs were approved, so this head has real CI for the first time — `lint`, `ds-lint`, `prepare`, `scope`, `audit-scope`, `license/cla` pass; `mutate (shared, …/load-more.ts)` **fails**; `test` and 15 `ephemeral-integration` shards are still in progress. No local gate was run: CI now covers it.

@kamwro — thank you for the workflow-approval ask, it is what surfaced the blocker below. That is the system working as intended rather than an argument against having asked.

---

## ✅ Previous findings — all seven verified fixed

I re-checked each against the code rather than against the changelog.

- **B1** (`AttachmentsSection` unbounded duplicates on a clamping endpoint) — **fixed** and fixed the right way. `returnedPage >= targetPage` (`packages/ui/src/backend/detail/AttachmentsSection.tsx:92-93`) gates both the append and `hasMore`, the append merges through a `Map` keyed by `id` (`:94-100`), and `totalPages` state is gone from the component entirely (`:343`). I traced the two failure sequences by hand: 24 attachments now requests `[1,2]` and ends with 24 rows and no button; 48 requests `[1,2,3]` and ends with 48. The clamped answer is dropped rather than appended, and because `hasMore` is false the rewound `setPage(returnedPage)` is unreachable as a next-request source.
- **M1** (tests modelled a terminal response the route cannot emit) — **fixed**, and this is the strongest part of the update. `respondLikeClampingRoute` (`packages/ui/src/backend/__tests__/AttachmentsSection.test.tsx:128-152`) clamps offset *and* echoed page exactly as `route.ts` does and records every page requested, so `requestedPages).toEqual([1,2])` pins the termination rather than merely the row count. The unreachable `{page: 2, totalPages: 1}` fixture is corrected to `totalPages: 2` with a comment explaining why. The other call sites' new cases answer empty past the end, which is what their endpoints do.
- **m1** (guard measured a client-deduped array) — **fixed**. `AssignableStaffMembersPage.servedCount` carries the raw pre-dedupe length (`assignableStaff.ts:27`, set from `rawItems.length` at `:135`, and `0` on the 404 branch at `:69`), and both staff guards read it (`AssignRoleDialog.tsx:164`, `ParticipantsField.tsx:65`). `ParticipantsField.test.tsx:164-166` is the case I wanted: a served page of 20 that dedupes to one row still offers the next page. `DealsSection` measures `rawItems.length` (`:553`), which is the raw `payload.items` from `:408`.
- **m2** (`hasMore` missing from the reset) — **fixed** in both branches (`AssignRoleDialog.tsx:102`, `:119`).
- **m3** (`NotesSection`, the sixth surface) — **fixed**. Converted to `hasMoreFromPage`, `totalPages` state removed, append deduped by `id` (`NotesSection.tsx:806-810`). I verified the served-count obligation holds here rather than assuming it: `notesAdapter.ts:61` is `payload.items.map(mapCommentSummary)`, a 1:1 map with no filter or dedupe, so `items.length` *is* the served count.
- **n1** (`/api/attachments` clamps at the source) — filed as #5299 and correctly left out of this PR. Changing that route's paging contract does affect other consumers.
- **n2** (five copies of the rationale) — **fixed** well. `hasMoreFromPage` (`packages/shared/src/lib/pagination/load-more.ts`) is the right shape, and its docstring recording *both* obligations — measure the served count, and only for endpoints that come back short — is better than what I asked for. The clamping caveat naming `AttachmentsSection` is exactly the note that stops the next person reintroducing B1.

---

## 🔴 Blocker

### B2 — the `mutate` check fails: the new shared helper ships with no test in its own package

`mutate (shared, src/lib/pagination/load-more.ts)` is red ([job 95350098340](https://github.com/open-mercato/open-mercato/actions/runs/31792005323/job/95350098340)), exit code 1:

```
ConfigError: No tests were executed. Stryker will exit prematurely. Please check your configuration.
No mutation report was produced for shared. The mutation run did not get far enough to write mutation.json
```

**Cause.** `scripts/stryker/scope.mjs` picks changed files per allowlisted package from the PR diff; `packages/shared/src/lib/pagination/load-more.ts` is a new file in the `shared` allowlist, so Stryker mutates it and runs `packages/shared`'s own Jest suite. Nothing in that suite reaches the file — `packages/shared/src/lib/pagination/` contains `load-more.ts` and nothing else, and there is no test for it anywhere under `packages/shared`. The five suites that exercise the helper live in `packages/ui` and `packages/core`, which this job never runs. Stryker's dry run therefore finds zero covering tests and errors out before producing a report.

**Why this blocks, stated precisely so you don't chase the wrong thing.** The mutation *score* is advisory — `MUTATION_ENFORCE` is `false` and the job carries `continue-on-error: ${{ vars.MUTATION_ENFORCE != 'true' }}`, and `enforce.mjs` logged `nothing to enforce (advisory)`. What fails is the run itself erroring, which reports the check red regardless. `develop` exposes no required status checks through the API (`404 Required status checks not enabled`), and both the tracker descriptor and this workflow's own header comment specify that in exactly that case every reported check is treated as required:

> Branch protection on develop is not readable through the API, and this repository's review and merge automation is specified to treat every reported check as required in exactly that case — so a red job here does reach the pipeline even though no mutation *score* can block anything.

So this is not a threshold to argue with; it is a red check the pipeline honours.

**Fix**, which also closes a genuine coverage gap independent of CI — a newly exported `@open-mercato/shared` helper currently has no direct test, only five indirect component ones. Add `packages/shared/src/lib/pagination/__tests__/load-more.test.ts` covering the four cases the docstring already reasons about:

```ts
import { hasMoreFromPage } from '../load-more'

describe('hasMoreFromPage', () => {
  it('reports more when the page came back full', () => {
    expect(hasMoreFromPage(24, 24)).toBe(true)
  })
  it('reports more when the endpoint served more than it was asked for', () => {
    // `>=` rather than `===`, so an over-serving endpoint still terminates.
    expect(hasMoreFromPage(25, 24)).toBe(true)
  })
  it('terminates on a short page', () => {
    expect(hasMoreFromPage(23, 24)).toBe(false)
  })
  it('terminates on an empty page', () => {
    expect(hasMoreFromPage(0, 24)).toBe(false)
  })
})
```

That gives Stryker mutants to kill (the `>=` boundary is the whole function, and these four cases pin it), turns the check green, and documents the contract in the package that owns it.

---

## 🟡 Minor

### m2 — two more `page < totalPages` load-more gates remain, both beside files this PR converts

The PR no longer claims "five surfaces", which addresses my previous m3 as written. But the pattern still isn't exhausted, and the two survivors are not in distant corners of the repo — they are in the same directory family as three of the six conversions:

- `packages/core/src/modules/customers/components/detail/hooks/usePersonTasks.ts:479` — `const hasMore = entityId != null && pageInfo.page < pageInfo.totalPages`, with the same bound guarding the fetch at `:245` (`if (pageInfo.page >= pageInfo.totalPages) return`) and `totalPages` taken straight from the payload at `:190`. This backs `TasksSection.tsx`'s legacy path, in the directory where `AssignRoleDialog`, `DealsSection` and `EntityTagsDialog` sit.
- `packages/core/src/modules/customers/components/detail/EntityTagsDialog.tsx:1253` — `activeCategoryPage < activeCategoryTotalPages` gating a load-more over entries appended with `mergeOptions` (`:764-769`), with `activeCategoryTotalPages` set from the response at `:762-764`. Directly beside the converted `AssignRoleDialog.tsx`.

I am **not** asking you to convert them in this PR — that is a defensible scope line, and the new helper makes the conversion mechanical for whoever does it. What I am asking for is that the boundary be stated rather than implied: convert them here, or add a line to the PR body (and ideally a follow-up issue referencing `hasMoreFromPage`) saying these two are known and deferred. Right now a reader who trusts the PR would conclude the pattern is gone from `customers/components/detail/`, and it isn't.

**Verified and deliberately not flagged**, so you can see where I drew the line:

- `ActivityHistorySection.tsx:300` — `firstPageHasMore = firstPageHasMore || legacyTotalPages > loadedPages`. This *reads* like the defect but isn't: the primary signal is the canonical cursor (`Boolean(response.nextCursor)`, `:265`) and `totalPages` is OR'd onto it, so an under-reporting total can only add "there may be more", never hide rows. Correct as written.
- `useVersionHistory.ts:81` — already `items.length === PAGE_SIZE`, page-fullness based. No defect. It could adopt the helper (and `>=` over `===`) opportunistically, but nothing is broken.
- The `pagination={{ … totalPages … }}` hits across `webhooks`, `api_keys`, `entities`, `planner` and `business_rules` are next/prev DataTable pagination, not append-style load-more, so they are out of this pattern entirely.

---

## ⚪ Nit

### n2 — the helper's docstring cites `OM_LIST_COUNT_CAP`, which does not exist in the codebase

`load-more.ts:6` gives "a capped list count (`OM_LIST_COUNT_CAP`) does it by design" as a present-tense cause. The only occurrence of that identifier anywhere in the repo is that docstring: the flag is *specified* in `.ai/specs/2026-07-27-list-count-strategies.md` (#4552 is the merged spec PR, not an implementation), not shipped. Since this file is now the single canonical home for the rationale, a reader will grep for the flag and find nothing. Marking it as planned — "and a capped list count will do it by design once `OM_LIST_COUNT_CAP` lands (see the list-count-strategies spec)" — keeps the file honest. The first cause listed, rows inserted between requests, is real today and already carries the argument on its own.

---

## Verified and deliberately not reported

- No contract surface is engaged. `AssignableStaffMembersPage` gains a required `servedCount`, but it appears only as a return type of `fetchAssignableStaffMembersPage`, every consumer is in-repo (`AssignRoleDialog`, `ParticipantsField`, two test files, all updated), and the module is a component-internal helper rather than a surface `BACKWARD_COMPATIBILITY.md` protects. Readers are unaffected; no external producer plausibly constructs it.
- `@open-mercato/shared/lib/pagination/load-more` resolves through the package's existing `./*/*/*` wildcard export (`types: ./src/lib/pagination/load-more.ts`, `default: ./dist/lib/pagination/load-more.js`) — the same route `@open-mercato/shared/lib/time` already takes, so no `exports` entry is needed.
- No new `any`, no hardcoded status colors, no arbitrary Tailwind values, no `dark:` overrides, no locale-file or generated-artifact changes in the diff.
- Every `fetchAssignableStaffMembersPage` mock in `ParticipantsField.test.tsx` supplies `servedCount`, so no test passes vacuously through an `undefined >= pageSize` comparison.
- `AttachmentsSection` handles the absent-`page` case correctly: `returnedPage` falls back to `targetPage`, so an endpoint that does not echo its page keeps plain short-page termination.

## CI

`mutate (shared, …/load-more.ts)` fails (B2 above). Still in progress at review time: `test`, and `ephemeral-integration` shards 1–15. Those are **pending, not failing** — I am not waiting on them and their outcome is not folded into this verdict. If any of them goes red, that is additional work beyond B2, and it will show on the PR rather than in a further comment from me.

## Labels

Pipeline stays `changes-requested` (one blocker). `bug`, `needs-qa`, `priority-medium`, `risk-medium` unchanged — `risk-medium` still fits: the diff now touches six surfaces plus a new shared export, which is a wider blast radius than the `risk-low` originally proposed, not a narrower one. `partner-request` is outside this repo's configured taxonomy and is left alone.

**autofix: skipped (not my PR — re-run with `--autofix` to fix it here).** This is a fork PR authored by @kamwro, so nothing was pushed to the branch. B2 is a single new test file; the sketch above should be copy-pasteable.

pkarw:
Thanks @kamwro — handing this back to you for one more pass. The seven findings from the last round are all genuinely fixed, and I verified each against the code rather than the summary; the `AttachmentsSection` clamping fake that records requested pages is a better test than the one I asked for.

One blocker, and it is new rather than a re-raise: approving the workflow runs (which I did — that is why CI exists on this head at all) turned up `mutate (shared, src/lib/pagination/load-more.ts)` red. Stryker mutates the new file and runs `packages/shared`'s own suite, which never reaches it, so it errors with `No tests were executed` before writing a report. The fix is one new test file, `packages/shared/src/lib/pagination/__tests__/load-more.test.ts`; a copy-pasteable sketch covering the four boundary cases is in the review body. Worth having on its own terms too — right now a newly exported shared helper has only indirect coverage from five component suites.

Beyond that: one minor asking you to state, rather than imply, that `usePersonTasks.ts:479` and `EntityTagsDialog.tsx:1253` are known-and-deferred instances of the same pattern (I am not asking you to convert them here), and one nit that the docstring cites `OM_LIST_COUNT_CAP`, which is specified in a spec but not yet in the codebase.

`test` and the 15 `ephemeral-integration` shards were still running when I submitted. They are pending, not failing, and I did not fold them into the verdict or wait on them — your next push re-runs them anyway.

On your other ask: workflow approval for #5228 is outside this PR, so I have left it for a maintainer to action separately rather than doing it silently from a review run on this one.

pkarw:

Re-review of `eba77d7f` (previous pass: `94b04d64`). All seven prior findings verified fixed against the code. New: 1 blocker (`mutate` check red — the new shared helper has no test in `packages/shared`), 1 minor (two unconverted instances of the pattern left unstated), 1 nit (docstring cites an unimplemented flag). Autofix iterations: 0 — autofix skipped (not my PR — re-run with `--autofix` to fix it here). Validation source: GitHub checks; no local gate run. No CI follow-up comment will come from this run, as stated in the review body.

pkarw:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ PASS — all fourteen checkpoints across the six converted surfaces behaved as the change intends, including the clamping-endpoint case (B1) that the latest commit addresses.
**Environment:** `http://127.0.0.1:35171` · role `admin` · browser `playwright`
**Verified:** `kwrobel/load-more-short-page` @ `eba77d7`

The app was booted as a **production** Next.js build (`next start`, `NODE_ENV=production`) against a
disposable testcontainers pgvector database, migrated and seeded fresh. Every fixture below was
created through the app's own APIs and deleted in teardown; only the seeded demo credentials were used.

### Scenario (P0–P2 — customers detail incremental loading)

**Where to click:** `/backend/customers/companies-v2/{id}` (Deals, Files, People, Activity log tabs),
`/backend/customers/deals/{id}` (Notes tab).

| 1 | Deals tab of a company with **12** deals (page size 10) | 10 rows and a "Load more deals" button | 10 deal cards (Full Deal 12 → 03) with the button visible | ✅ |
| 2 | Click "Load more deals" | The last 2 deals append, button disappears | All 12 deals listed, button gone | ✅ |
| 3 | Deals tab of a company with **3** deals | No "Load more deals" at all | 3 cards, no button — short first page terminates immediately | ✅ |
| 4 | Files tab of a company with **26** attachments (page size 24) | 24 tiles and a "Load more" button | 24 tiles (file-26 → file-03), button visible | ✅ |
| 5 | Click "Load more" on Files — **`/api/attachments` clamps the page**, the B1 case | The last 2 files append **once**, no duplicates, button disappears | All 26 files listed exactly once (file-26 → file-01), button gone | ✅ |
| 6 | Repeat the Deals tab at a **390×844 mobile viewport** | List and button remain rendered and reachable | Both rendered correctly | ✅ |
| 7 | Notes tab of a deal with **21** notes (page size 20) | 20 notes and a "Load more notes" button | 20 notes, button visible | ✅ |
| 8 | Click "Load more notes" | The 21st note appends, button disappears | 21 notes, button gone | ✅ |
| 9 | People tab → "Add role" (staff endpoint served **3** of pageSize 24) | Member list renders, **no** "Load more" | Dialog listed the members with no button — no phantom affordance | ✅ |
| 10 | Activity log → Add new → New meeting → "+ Add link", search **21** matching companies (page size 20) | 20 results and a "Load more" button | 20 results, button visible | ✅ |
| 11 | Click "Load more" in the linked-entity popover | The 21st company appends, button disappears | 21 results, button gone | ✅ |
| 12 | Same dialog → "Add participant" (staff endpoint served **3** of pageSize 20) | Participant list renders, **no** "Load more" | Popover listed the members with no button | ✅ |
| 13 | Deals tab of a company with **exactly 10** deals — a last page that is exactly full | "Load more deals" is still offered (the accepted trade-off) | 10 deals, button visible | ✅ |
| 14 | Click that button on the exactly-full page | The empty page past the end ends the sequence, nothing duplicates | Button disappeared, still exactly 10 deals, no duplicates | ✅ |

### Notes for QA

- **All six converted surfaces were driven in a real browser** — `DealsSection`, `AttachmentsSection`,
  `NotesSection`, `AssignRoleDialog`, `LinkedEntitiesField` and `ParticipantsField`.
- **The clamping-endpoint fix holds.** Steps 4–5 are the interesting ones: `/api/attachments` clamps a
  page request past the end and re-serves the last page in full. With 26 files the tab paged to the
  end, appended each file exactly once and terminated — the failure mode the reviewer's B1 probe
  described (duplicate rows behind a button that never terminates) did not occur.
- **The exactly-full-page trade-off is visible but benign** (steps 13–14). When the last page happens
  to be exactly the page size, the button is still offered and the click that finds nothing is what
  removes it. That costs one extra request and one "nothing happened" click; nothing duplicates and
  the button does not linger. It is inherent to short-page termination and strictly better than the
  previous behaviour, which hid rows that do exist.
- **Coverage limit, stated honestly:** `AssignRoleDialog` and `ParticipantsField` could only be
  exercised in their *short-page* direction, because this environment seeds 3 assignable staff
  members and creating 24+ real team members was out of proportion for this run. Their full-page
  direction remains covered by the unit tests this PR adds.
- **Pre-existing, unrelated to this PR:** each deal card renders the section's `emptyLabel`
  ("No deals linked to this company.") inside its empty *Pipeline stage*, *Deal value*, *Probability*
  and *Expected close date* fields — visible in the step 1–3 screenshots. It reproduces independently
  of the load-more logic and is worth a separate issue, not a change here.

### Screenshots

**step-01-deals-full-page-load-more.png**



**step-02-deals-all-twelve-button-gone.png**



**step-03-deals-short-page-no-button.png**



**step-04-attachments-first-page.png**



**step-05-attachments-all-26-button-gone.png**



**step-06-deals-mobile-viewport.png**



**step-08-notes-first-page.png**



**step-09-notes-all-21-button-gone.png**



**step-10-assign-role-dialog.png**



**step-11-linked-entities-first-page.png**



**step-12-linked-entities-all-button-gone.png**



**step-13-participants-popover.png**



**step-14-exact-page-button-still-offered.png**



**step-15-exact-page-terminated-no-duplicates.png**

pkarw:

This change ships thorough unit tests at all five (now six) call sites, but no browser-level test —
nothing in a module `__integration__/` folder covers the affordance end to end, and the PR checklist
says as much. The QA above was manual; the scenario below is what it did, ready to be turned into a
Playwright spec with `/om-integration-tests`. It needs no server total that disagrees with the page:
real fixture volumes are enough to pin every branch of `hasMoreFromPage`.

1. **Setup** (API fixtures, per test, via `crmFixtures` / `attachmentsFixtures`):
   - a company with **12** deals (`DEALS_PAGE_SIZE = 10`) and a second company with **3**;
   - a company with **exactly 10** deals for the boundary case;
   - **26** attachments on a company (`PAGE_SIZE = 24`) — this is the one that guards the clamping
     `/api/attachments` regression;
   - a deal with **21** notes (`NOTES_PAGE_SIZE = 20`);
   - **21** companies matching one search term for the linked-entity popover (`PAGE_SIZE = 20`).
2. **Act:** open `/backend/customers/companies-v2/{id}` (Deals, Files, People, Activity log tabs) and
   `/backend/customers/deals/{id}` (Notes tab); click each affordance once —
   "Load more deals", "Load more" (Files), "Load more notes", and "Load more" inside the schedule
   dialog's "+ Add link" popover.
3. **Assert:**
   - a full first page renders exactly `pageSize` rows **and** offers the button;
   - after one click the remaining rows appear **exactly once** and the button disappears;
   - a short first page offers no button at all;
   - on the exactly-full page the button is offered once and the empty next page removes it without
     duplicating a row (the attachments case additionally proves the clamped page is not re-appended);
   - `AssignRoleDialog` / `ParticipantsField` offer no button while the staff endpoint serves a short
     page — and, with 24+ seeded team members, that they do offer one and page through correctly
     (the direction this manual run could not reach).
4. **Teardown:** delete every fixture created (attachments, notes, deals, companies) in `finally`.

pkarw:

kamwro:
B2 fixed on `dd7a8356`, and the two remaining findings are answered.

**B2** — your diagnosis was exact, and I verified the mechanism rather than assuming the fix: `jest --findRelatedTests src/lib/pagination/load-more.ts` inside `packages/shared` returned *nothing* before this commit and resolves the new test after, which is the condition the dry run checks. Added your four cases plus a boundary triple (1/2, 2/2, 3/2) so the `>=` can't drift to `===` or `>`. You were right that it was a real gap independent of CI — a newly exported helper with only indirect coverage from two other packages.

**m2** — deferred, with the boundary now stated rather than implied: a Scope boundary section in the PR body, and #5347 carrying the line numbers plus the served-count analysis already worked out for both files (`usePersonTasks` maps 1:1 and dedupes after, so `mapped.length` is the served count; `mapEntries` in `EntityTagsDialog` is a plain map with no filter, so its wrinkle is the per-category state rather than the count). I also recorded your two verified-and-excluded cases there so nobody re-derives them. The reason for deferring rather than converting: neither file has a test harness, and after M1 I'd rather build those against each endpoint's real paging behaviour in a focused pass than add two fresh harnesses here.

**n2** — reworded to mark the cap as planned, pointing at the spec. Since the flag ships in #5228 and that PR is still in review, the docstring would have been false on merge either way; thank you for catching it.

One ask: **the workflow runs need approving again.** The push reset it, so this head has no CI, and #5228 is in the same state after a `develop` merge that resolved a fresh conflict (`a293d4a7`, now `MERGEABLE`; its `DataTable` collision was your `showQueryTime` dep landing next to mine, and the create-app template gaining `ui.pagination.info.pageOf` beside my capped variant — both unioned). If per-push re-approval is going to be the norm for this account, I'm happy to batch pushes to reduce the clicking.

patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-18T14:26:00Z. Other auto-skills will skip this PR until the lock is released.

patzick (approved):
# 🔍 Re-review: fix(ui,customers): terminate "load more" on a short page, not on a reported total

**Verdict:** approved — 0 blockers, 0 majors, 0 minors, 1 nit.
**Head reviewed:** `bf02e6a4a7f2ae2c1a707491805611fe90bed7be` (previous reviews: `eba77d7f`, `94b04d64`) · base `develop` · `MERGEABLE`, no conflict.
**Validation source:** GitHub checks. Every reported check on this head is green — `prepare`, `lint`, `ds-lint`, `test`, `mutate (shared, …/load-more.ts)`, `scope`, `audit-scope`, `license/cla`, and all 15 `ephemeral-integration` shards. No local gate was run.

The delta since the last pass is `dd7a8356` (two files) plus a `develop` merge that touches none of this PR's files — I diffed `dd7a8356..bf02e6a4` over `packages/{ui,core,shared}` and it is empty, so the merge carries no semantic conflict.

---

## ✅ Both open findings verified fixed

- **B2** (blocker — `mutate (shared, …/load-more.ts)` red because the new helper had no test in the package that owns it) — **fixed**, and fixed past what I asked for. `packages/shared/src/lib/pagination/__tests__/load-more.test.ts` covers the four cases from the sketch and adds a fifth that pins the `>=` boundary from both sides (`hasMoreFromPage(1, 2)` false, `(2, 2)` true, `(3, 2)` true) — that is the case that stops the operator drifting to `===` or `>`, which is exactly what Stryker will try. The check is now green rather than erroring out on a dry run with no covering tests.
- **n2** (the docstring cited `OM_LIST_COUNT_CAP` in the present tense for a flag that does not exist) — **fixed**. `load-more.ts:6-9` now reads "will do it by design once `OM_LIST_COUNT_CAP` lands" and points at `.ai/specs/2026-07-27-list-count-strategies.md`. A reader who greps and finds nothing now knows why.
- **m2** (the scope boundary was implied rather than stated) — **fixed**. The PR body carries a "Scope boundary" section naming both survivors with their line numbers, and #5347 is filed and open. #5299 (the `/api/attachments` clamping root cause) is likewise filed and correctly kept out of this PR.

## 🔎 What I verified independently this pass

Obligation 2 in the helper's docstring — *only for endpoints whose page past the end comes back short* — is the load-bearing precondition of this whole change, and the two staff-backed surfaces were the ones no pass had checked against the route. So I traced it rather than assuming:

- `/api/staff/team-members/assignable` slices with an **unclamped** offset — `deduped.slice(start, start + query.pageSize)` where `start = (query.page - 1) * query.pageSize` (`packages/core/src/modules/staff/api/team-members/assignable/route.ts:214-216`). A page past the end returns `items: []`, not a re-served last page. `AssignRoleDialog` and `ParticipantsField` therefore terminate. Worth noting the route also dedupes server-side *before* slicing, so `data.items.length` genuinely is the served page — `servedCount` measures the right thing at the right layer.
- `/api/customers/deals` and `/api/sales/quotes` are `makeCrudRoute` lists, as are `/api/customers/companies` and `/api/customers/comments`. Query-engine paging, unclamped offset, short page past the end.
- `/api/attachments` is the one clamping endpoint, and it is the one surface that guards with `returnedPage >= targetPage` before appending. The precondition and the exception line up.
- Page-size constants match their requests everywhere: `ASSIGNABLE_STAFF_PAGE_SIZE = 24` is both the `pageSize` passed to `fetchAssignableStaffMembersPage` (`AssignRoleDialog.tsx:146`) and the divisor in the guard (`:164`); `PAGE_SIZE = 20` likewise in `ParticipantsField` and `LinkedEntitiesField`; `PAGE_SIZE = 24` in `AttachmentsSection`; `DEALS_PAGE_SIZE` in `DealsSection`. Lifting them to named constants is what makes that checkable at a glance, and it is why I could check it.
- `LinkedEntitiesField` resets `page` to 1 on `query`/`linkType`/`open` change (`:129-132`), so `hasMore` from a previous query cannot leak into a new one, and the new `.catch` clears `hasMore` alongside `results` rather than leaving a stale `true` behind a wiped list.
- `assignableStaff.ts` counts `rawItems.length`, which includes rows the dedupe loop skips for a missing `userId`. That is correct: those rows were served, and counting them is what keeps the sequence alive.
- `notesAdapter.listPage` is a 1:1 `items.map(mapCommentSummary)` with no filter, so `NotesSection`'s `pageResult.items.length` is the served count. It is the only `listPage` implementation `NotesSection` is wired to.

## ⚪ Nit

### n3 — the PR body still describes five surfaces; the diff converts six and adds a shared helper

The Summary opens "Five incremental-loading surfaces…", the Changes list names five, and Testing says "Each of the five files gains two tests". The diff converts **six** — `NotesSection` is the sixth, added when it was flagged last round — and also adds `packages/shared/src/lib/pagination/load-more.ts` plus its test, neither of which the body mentions. The Summary also still says a capped list count "causes it by design" in the present tense, the wording the docstring was just corrected away from.

Purely a description-vs-diff drift; nothing in the code depends on it. But this body is what lands in the squash commit and what the next reader of `git log` sees, and it currently undersells the change and points away from the file that now owns the rationale. Worth a pass before merge — not worth another CI round on its own.

## Verified and deliberately not reported

- No contract surface is engaged. `AssignableStaffMembersPage` gains a required `servedCount`, but it is a *return* type — every consumer reads it, none constructs it, and all in-repo consumers are updated. `hasMoreFromPage` is a purely additive export resolving through `@open-mercato/shared`'s existing wildcard, no `exports` entry needed.
- No new `any`, no hardcoded status colors, no arbitrary Tailwind values, no `dark:` overrides, no user-facing string added, no locale or generated-artifact change. `ds-lint` agrees.
- The one wasted request in `AttachmentsSection` when the row count is an exact multiple of 24 is inherent to a clamping endpoint and is documented at the call site; #5299 is the right place to remove it.

## CI

All green on `bf02e6a4`, including the full 15-shard integration suite. `[code]smith`, `audit`, `comment`, `docker-build` and `merge-coverage` report as skipped by their own conditions.

## Labels

Pipeline moves `changes-requested` → `review` (no blocking finding remains). `bug`, `needs-qa`, `priority-medium`, `risk-medium` unchanged. `partner-request` is outside this repo's configured taxonomy and is left alone.

⚠️ **Merge gate:** this PR carries `needs-qa` and does **not** carry `qa-approved`, so it must not merge yet even with everything green. The `om-auto-qa-pr` pass on 2026-08-17 posted a ✅ PASS with fourteen checkpoints across the six surfaces, and the only change since then is a new unit-test file plus a `develop` merge — so that evidence still stands on this head. Applying `qa-approved` is a QA-reviewer call, not this skill's, so it stays for a maintainer to action.

**autofix: skipped (not my PR).** This is a fork PR authored by @kamwro; nothing was pushed to the branch. The single nit is a PR-body edit only.

patzick:
Label moved `changes-requested` → `review` because the re-review of `bf02e6a4` found no blocker, major or minor — the `mutate (shared, …/load-more.ts)` blocker from the `eba77d7f` pass is fixed on `dd7a8356` and that check is green, as is every other reported check including all 15 `ephemeral-integration` shards.

| Label | Why |
|---|---|
| `review` | Approved, but not merge-ready: the QA gate below still holds it. |
| `bug` | Unchanged — rows stranded behind a vanished "Load more". |
| `needs-qa` | Unchanged. UI-rendering files across six surfaces; the automated-verification exemption does not apply. |
| `priority-medium` | Unchanged. |
| `risk-medium` | Unchanged — six surfaces plus a new shared export. |
| `in-progress` | Removed, lock released. |

`partner-request` is outside this repo's configured taxonomy and was left alone.

⚠️ **Do not merge yet.** `needs-qa` is set and `qa-approved` is not, which is a hard merge block regardless of the green checks. The `om-auto-qa-pr` pass on 2026-08-17 posted a ✅ PASS over fourteen checkpoints on `eba77d7f`, and the only change since is one new unit-test file plus a `develop` merge that touches none of the PR files — so that evidence carries to this head. Applying `qa-approved` is a QA-reviewer action and is deliberately left to a maintainer.

patzick:

0 blockers, 0 majors, 0 minors, 1 nit. Autofix iterations: 0 — fork PR, nothing pushed to the branch; the nit is a PR-body edit only. Validation source: GitHub checks (all green on `bf02e6a4`, including all 15 integration shards). Merge still gated on `qa-approved`.

PR #5291, opened 2026-08-13
fix(security): require a WebAuthn assertion for passkey verification (#3852)

### LINKED ISSUE #3852: security(ent_security): passkey MFA second factor bypassable via non-cryptographic fallback (PasskeyProvider.ts:47-55)
The passkey MFA provider accepts a non-WebAuthn verification shape that degrades verification to plain string comparison of public, non-secret values — defeating the passkey second factor without the authenticator's private key. **LIVE** — whoever holds the user's `mfa_pending` session (the party already driving the login MFA step, or a replay of the prepare response) completes the passkey challenge without any cryptographic assertion. This breaks the second factor for both login-MFA and passkey-as-sudo step-up.

## Location(s)
- `packages/enterprise/src/modules/ent_security/lib/providers/PasskeyProvider.ts:47-55` — `verifyPayloadSchema` accepts a `{ credentialId, challenge }` shape alongside the real `{ response }` WebAuthn assertion.
- `packages/enterprise/src/modules/ent_security/lib/providers/PasskeyProvider.ts:339-347` — for the non-`response` shape, `verify()` degrades to `providerMetadata.credentialId === parsed.credentialId && pending.challenge === parsed.challenge`.
- `packages/enterprise/src/modules/ent_security/lib/providers/PasskeyProvider.ts:283-288` — when `verifyContext` is absent entirely, falls back to a `credentialId` string compare.
- `packages/enterprise/src/modules/ent_security/lib/providers/PasskeyProvider.ts:252-267` — `prepareChallenge` returns `allowCredentials[].id` and `options.challenge` to the client (both values echoed back are the ones the fallback compares).
- `packages/enterprise/src/modules/ent_security/api/mfa/methods/route.ts:32` — GET `/api/security/mfa/methods` echoes `providerMetadata` incl. `credentialId`.

## Why it matters
Both values compared in the fallback are non-secret. `prepareChallenge` hands `allowCredentials[].id` (the credentialId) and `options.challenge` to the client, and the methods list route discloses `providerMetadata.credentialId`. So an attacker who can reach the MFA verify step for a session (they already hold `mfa_pending`, or they replay the prepare response) submits `{ credentialId, challenge }` with the public values and passes — no authenticator private key, no `verifyAuthenticationResponse`. This defeats:
- login-time MFA (the passkey second factor), and
- passkey-as-sudo step-up (`SudoChallengeService.verify` → `MfaVerificationService.verifyChallenge` → `provider.verify`).

The intended path requires a real WebAuthn assertion signed by the authenticator; the fallback shape bypasses that entirely.

## Proposed fix
Remove the `{ credentialId, challenge }` branch from `verifyPayloadSchema` and both fallback branches (the `verifyContext`-absent branch at 283-288 and the string-compare branch at 339-347). Require a real `verifyAuthenticationResponse` assertion and reject when `verifyContext` is missing. A clean, isolated schema + branch removal in `PasskeyProvider.ts`.

## BC / Safety
Any client currently sending the `{ credentialId, challenge }` shape (or relying on the `verifyContext`-absent path) will start being rejected — that is the intent, but confirm no legitimate first-party flow depends on the degraded path (e.g. a test/dev shortcut). Re-verify: full passkey login-MFA and passkey sudo step-up still succeed with a genuine WebAuthn assertion after the removal. `risk-high` — this is auth/second-factor code; the fix must not break real passkey verification. Confined to the `ent_security` enterprise module; no cross-package contract surface changes.

Paul-Mlodochowki:
Fixes #3852

## 🎯 What this fixes

`PasskeyProvider` accepted a second verification payload shape — `{ credentialId, challenge }` — alongside the genuine `{ response }` WebAuthn assertion, and approved it by plain string comparison. Both compared values are public: `prepareChallenge()` returns `allowCredentials[].id` and `options.challenge` to the caller, and `GET /api/security/mfa/methods` discloses `providerMetadata.credentialId`. Anyone who could reach the verify step for a session — the party already holding `mfa_pending`, or a replay of the prepare response — passed the passkey second factor with no authenticator private key and no signature.

The worst variant needed even less. `MfaVerificationService.verifyChallenge` builds the verify context from `challenge.providerChallenge`, which is `null` whenever the client never calls `/api/security/mfa/prepare`. That routed straight into the context-absent fallback, which compared only the disclosed `credentialId` and ignored the challenge entirely.

Because `SudoChallengeService.verify` → `MfaVerificationService.verifyChallenge` → `provider.verify` is the same path, this defeated both login-time MFA and passkey-as-sudo step-up.

## 🔒 The change

All of it is in `packages/enterprise/src/modules/security/lib/providers/PasskeyProvider.ts`:

- `verifyPayloadSchema` becomes a single object requiring `response`; the `{ credentialId, challenge }` union member is gone.
- The `!parsedContext.success` branch returns `false` instead of falling back to a `credentialId` compare.
- The trailing `credentialId` + `challenge` string-compare block after the assertion branch is deleted.

A genuine `verifyAuthenticationResponse` is now the only route to a positive verdict. The challenge TTL check, the signature counter update, and the `{ response }` path itself are untouched.

One deliberate refinement over a straight removal: the schema is applied with `safeParse` and a failed parse returns `false` rather than throwing. A thrown `ZodError` would reach `mapMfaError` as a logged **500** and would skip `registerFailedAttempt`, so rejected attempts would not count toward the challenge attempt limit. `safeParse` keeps the **401** the route already documents and preserves lockout behavior.

## 🧪 Tests

`PasskeyProvider.test.ts` previously contained a test named *"supports legacy verification payload for backward compatibility"* that asserted the bypass worked. It is replaced with four regression tests, all failing before this change:

- the legacy `{ credentialId, challenge }` payload is rejected even with a correctly prepared context;
- a `credentialId`-only payload is rejected when no verify context exists (the skip-`/prepare` variant);
- a WebAuthn assertion is rejected when no challenge was ever prepared;
- an assertion the authenticator did not sign is rejected.

The first two also assert `verifyAuthenticationResponse` is never reached.

**Integration suite.** The Playwright specs used this hole as their QA shortcut, so they change with it. The fixture is renamed `attemptNonCryptographicPasskeyVerify` and documented as a negative case. `TC-SEC-004` now asserts login verification is refused; `TC-SEC-006` and `TC-SEC-007` assert passkey sudo verification is refused and then skip their sudo-token-gated remainders with an explicit reason. There is no API-level way to produce a genuine WebAuthn assertion, so passkey verification has no integration happy path any more — restoring one needs a Playwright virtual authenticator (`WebAuthn.addVirtualAuthenticator`), which is out of scope for a security fix.

## 💥 Breaking change — intentional

`POST /api/security/mfa/verify` and `POST /api/security/sudo/verify` no longer accept a `{ credentialId, challenge }` passkey payload; it now returns **401** instead of succeeding.

Per `BACKWARD_COMPATIBILITY.md` an API request shape is contract surface, and the normal protocol is a deprecation window. **That protocol is deliberately not followed here**, because the shape being removed *is* the vulnerability — a bridge release would leave the second-factor bypass live for a minor version.

The first-party UI is unaffected: `PasskeyChallengeVerify.tsx` sends `{ response }` from `startAuthentication()`, and `PasskeyProviderDetails.tsx` enrolls through `startRegistration()`. Any third-party client sending the legacy shape was, by construction, not performing cryptographic verification.

**Known follow-up:** the *setup* side still has an equivalent shortcut (`setupConfirmationPayloadSchema` accepts a client-supplied `publicKey` with no attestation). Credentials enrolled that way carry a fabricated public key and become unverifiable after this change — a user whose only MFA method is such a credential needs an admin MFA reset. That branch is out of scope here and deserves its own issue.

## ✅ Validation

Full gate from `.ai/agentic.config.json`, local runner (Docker unavailable), in order: `build:packages` ✅ · `generate` ✅ (no generated drift) · `build:packages` ✅ · `i18n:check-sync` ✅ · `i18n:check-usage` ✅ · `typecheck` ✅ · `test` ⚠️ · `build:app` ✅.

`@open-mercato/enterprise` is **59/59 suites, 481/481 tests green**. `yarn test` exits non-zero on nine untouched packages (queue, events, core, cli, shared, ui, telemetry, ai-assistant, create-mercato-app) with Windows-local filesystem, timing, and number-formatting failures; the `queue` `EPERM: rename` failure reproduces identically on a clean checkout of `develop`, and no failing test name mentions security, mfa, sudo, or passkey. CI on Linux is the authority for those.

🧪 Manual QA is still pending — this touches the login MFA and sudo step-up flows, so a reviewer should confirm that a real passkey login and a real passkey sudo step-up still succeed with a genuine authenticator.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5291"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789251084&installation_model_id=432243&pr_number=5291&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5291&signature=9ad5fab21d97fa0cb2bcc964db121cc0e25f2c3e5db6af20a53f1ada9cd26a15"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Paul-Mlodochowki:

⚠️ This PR was opened from a fork by an account with read-only permissions on this repository, so **no labels could be applied** (`AddLabelsToLabelable` denied). A maintainer needs to apply the set below manually — the same limitation blocked the `in-progress` claim label on #3852.

- 🔍 `review` — the work is complete, validated, and ready for code review; nothing is being held back.
- 🐛 `bug` — this corrects broken behavior in an existing feature rather than adding a capability.
- 🔒 `security` — it closes an authentication bypass that let the passkey second factor be passed without any cryptographic assertion.
- 🧪 `needs-qa` — the change sits directly in the login-MFA and sudo step-up paths, so a human should confirm that a real passkey login and a real passkey sudo step-up still succeed with a genuine authenticator before this merges. No `skip-qa`: the automated-verification exemption does not apply, because the API request contract changed.
- 🔺 `priority-high` — a live second-factor bypass is release-blocking, though it is not an active incident, which is what `priority-extreme` is reserved for.
- ⚠️ `risk-high` — the diff is small and isolated to one provider, but it is auth code on a contract surface, and a mistake here would lock legitimate users out of their own accounts.

Paul-Mlodochowki:
🤖 `om-auto-review-pr` taking over the chain lock from `om-auto-fix-issue` to run the code review. (The `in-progress` label could not be applied — read-only permissions on this repository — so this comment is the lock signal.)

Paul-Mlodochowki (commented):
# 🔍 Code Review

_Submitted as a conversation comment rather than a formal review: GitHub rejects self-approval, and this PR's author is the reviewing account. A second pair of human eyes is still required before merge._

**Runner:** local (no compose `app` container running — the Docker probe failed to reach the daemon, so every gate command ran as `yarn X`).

## 🎯 Summary

The change removes an authentication bypass in `PasskeyProvider`: a `{ credentialId, challenge }` verification payload that the provider accepted alongside a genuine WebAuthn assertion and approved by string-comparing two values the server itself hands the client. The diff is small, isolated to one provider plus its tests, and correct — the `{ response }` path, the challenge TTL check, and the signature-counter update are all untouched, and every route to `return true` now goes through `verifyAuthenticationResponse`.

Two findings were raised and fixed during this review (commit `ab8dd82`); one is left open as a judgement call, and one blocker is an environment condition this platform cannot clear.

**Verdict: request changes** — the `yarn test` gate does not pass on this machine (see 🧪 Validation Gate). Every finding attributable to the diff itself is resolved. On a Linux CI runner, where the failing suites are green, this review's substantive verdict is **approve**.

## 🧪 Validation Gate

| Command | Result | Notes |
|---|---|---|
| `yarn build:packages` | ✅ pass | |
| `yarn generate` | ✅ pass | No generated-file drift; `git status` showed only the intended files. |
| `yarn build:packages` | ✅ pass | Rebuild after `generate`, per the repo's gate-order note. |
| `yarn i18n:check-sync` | ✅ pass | |
| `yarn i18n:check-usage` | ✅ pass | |
| `yarn typecheck` | ✅ pass | 22/22 tasks, re-run after the review fixes. |
| `yarn test` | ❌ **fail** | `@open-mercato/enterprise` is **59/59 suites, 481/481 tests green**. Nine untouched packages fail — see the blocker below. |
| `yarn build:app` | ✅ pass | |
| `yarn template:sync` (repo-local parity gate) | ✅ pass | App and template in sync; no `apps/mercato/src` or env changes in this diff. |
| `yarn check:client-boundaries` | ⏭️ not applicable | The diff touches no web route, provider, or app-shell file, so the frontend performance gate has nothing to measure. |

## 🚫 Blockers

**1. `yarn test` exits non-zero — `validation.commands`, gate step 7.**

The repository's review rules are explicit that a failing gate command is a blocker regardless of cause, so it is recorded as one. The evidence, however, points away from this diff: the failures are in `queue`, `events`, `core`, `cli`, `shared`, `ui`, `telemetry`, `ai-assistant`, and `create-mercato-app` — none of which this change touches, and none of which import the passkey provider. They are Windows-local filesystem, timing, and number-formatting failures: `queue/src/__tests__/local.strategy.test.ts` dies on `EPERM: operation not permitted, rename` while a watcher holds the directory open, and that exact failure **reproduces identically on a clean checkout of `develop`** with none of this branch's changes applied. No failing test name mentions security, mfa, sudo, or passkey.

This blocker cannot be cleared on this platform. Linux CI is the authority; if the required checks come back green there, treat it as resolved. It is called out rather than waived so nobody reads "review passed" as "the whole suite ran clean here".

## ⚠️ Majors — fixed in `ab8dd82`

**1. Breaking contract-surface change with no record in the repository's own compatibility documents.** `BACKWARD_COMPATIBILITY.md` §7 classifies API routes as STABLE and the Quick Reference marks a request-shape change as BREAKING; `AGENTS.md` requires either the deprecation protocol or a documented exception. Skipping the bridge is right here — a bridge release would keep the bypass live for a minor version — but the justification lived only in the PR body, which no downstream module author reads. Left as-is, a maintainer auditing contract surfaces later would find an undocumented break.

Fixed by adding a dated `## Passkey MFA Verification Payload (2026-08-13)` section to `BACKWARD_COMPATIBILITY.md`, following the format the `Command Interceptor HTTP Status (2026-08-06)` precedent established, and a `#3852` entry under `0.6.7 → 0.7.0` in `UPGRADE_NOTES.md`.

**2. Data already persisted by the path being corrected has no remediation.** `.ai/lessons.md` carries a lesson this repository learned the hard way — *"Fixing the writer of a bad persisted value needs a remediation branch for values already stored"* — and the repo-local review extension says documented lessons are review rules. It applies directly: the passkey **setup** path still accepts a client-supplied `publicKey` with no attestation, and credentials enrolled that way were only ever verifiable through the fallback this PR deletes. After this change they fail every verification, and a user whose only MFA method is such a credential cannot log in.

Fixed by documentation rather than by a migration, and deliberately so: a migration that deletes rows it believes are fake would remove real users' second factors on a heuristic, which is a worse failure than the one it prevents. `UPGRADE_NOTES.md` now carries an explicit **operator action** naming the existing admin reset (`POST /api/security/users/{id}/mfa/reset`) as the remediation path, and notes that UI-registered passkeys are unaffected. Flagging this as a residual risk a human should confirm against their own deployment.

**3. Weakened assertions in the integration specs.** `TC-SEC-004:66`, `TC-SEC-006:130`, and `TC-SEC-007:172` asserted `expect(status).not.toBe(200)`, which passes on a `500`, a `403`, or a route that has been deleted entirely — exactly the outcomes a regression test for an auth fix must distinguish from a clean rejection. Both routes return a specific `401` on a failed verification (`MfaVerificationService` → `securityApiError(401, …)`; `SudoChallengeService.verify` → `SudoChallengeServiceError('Unable to verify sudo challenge', 401)`).

Fixed: all three now assert `toBe(401)`.

## 💡 Minors

**1. Unreachable code after the runtime skip in `TC-SEC-006:133` and `TC-SEC-007:175`.** Both specs assert the passkey sudo verification is refused and then call `test.skip(true, …)`, leaving the sudo-token-gated remainder (~60 and ~40 lines) unreachable.

Left as-is, as a deliberate call rather than an oversight. This is Playwright's documented runtime-skip idiom, the reason string names precisely what is missing, and the blocks below become live again the moment a virtual authenticator harness lands. Deleting them would discard working coverage that a `WebAuthn.addVirtualAuthenticator` fixture would restore verbatim. Reasonable reviewers may disagree; if the preference is to delete them, say so and it is a two-minute change.

- [x] **Exported APIs** — unchanged. `MfaProviderInterface`, `MfaProviderSetup`, and `MfaVerifyContext` keep their shapes; `verifySchema` is still a `z.ZodSchema`. Only this provider's own schema narrowed, and it is never serialized into an API response (no `api/` consumer reads it).
- [ ] **HTTP routes and request shapes** — **broken, intentionally.** `POST /api/security/mfa/verify` and `POST /api/security/sudo/verify` no longer accept the `{ credentialId, challenge }` passkey payload. ⚠️ **This is a BREAKING change to a STABLE surface shipped without a deprecation bridge.** It is justified — the removed shape *is* the vulnerability — and now documented in `BACKWARD_COMPATIBILITY.md` and `UPGRADE_NOTES.md`. A human maintainer should consciously accept this rather than let it merge unnoticed.
- [x] **HTTP response shapes** — unchanged. Same URLs, methods, success bodies, and the `401` both routes already documented for a failed verification.
- [x] **Event IDs** — unchanged; `security.mfa.verified` and `security.sudo.verified` still emit on success.
- [x] **DB schema** — no entity or migration changes in the diff.
- [x] **DI service names, ACL feature IDs, CLI flags, config formats, generated files** — untouched.

## 🧪 Test Coverage

Coverage for the fix itself is solid. The old test named *"supports legacy verification payload for backward compatibility"* asserted the bypass **worked** — deleting it was mandatory, and it is replaced by four tests that each fail against the pre-fix provider:

- the legacy payload is rejected even with a correctly prepared context;
- a `credentialId`-only payload is rejected when no verify context exists — the worst variant, reachable by skipping `/api/security/mfa/prepare` entirely, where the old code compared only the publicly disclosed `credentialId` and ignored the challenge;
- an assertion is rejected when no challenge was prepared;
- an assertion the authenticator did not sign is rejected.

The first two also assert `verifyAuthenticationResponse` is never reached, which is what distinguishes "rejected properly" from "accidentally verified".

One coverage loss is worth stating plainly rather than burying: **passkey verification no longer has an integration happy path.** The Playwright specs used the bypass as their shortcut, so `TC-SEC-004` now proves refusal instead of success, and the sudo segments of `TC-SEC-006`/`TC-SEC-007` prove refusal and skip. Restoring end-to-end proof that a *real* passkey login succeeds needs a Playwright virtual authenticator (`WebAuthn.addVirtualAuthenticator`), which is out of scope for a security fix but should be tracked.

Also noted for follow-up (out of scope here, no code change proposed): `setupConfirmationPayloadSchema` still accepts `{ credentialId, publicKey, challenge }` with no attestation, letting a caller register a credential no authenticator controls.

## 🏷️ Labels

Labels could not be applied — this PR comes from a fork opened by an account with read-only permissions on this repository (`AddLabelsToLabelable` denied). The recommended set and rationale are in [the earlier label-rationale comment](https://github.com/open-mercato/open-mercato/pull/5291#issuecomment-5286989265); a maintainer needs to apply them by hand. `needs-qa` stands: manual QA should confirm a real passkey login and a real passkey sudo step-up still succeed with a genuine authenticator.

---

🤖 `om-auto-review-pr` · autofix applied (majors 1–3), re-verified with `yarn typecheck` and the enterprise suite.

Paul-Mlodochowki:
🤖 `om-auto-fix-issue` run complete: review posted with three majors found and fixed in `ab8dd82`, one blocker left standing (the `yarn test` gate cannot pass on this Windows machine — the failing suites are untouched packages and reproduce on a clean `develop`). Lock released.

**📸 UI verification: not run.** The diff contains no UI-rendering file — no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**` — so `om-auto-qa-pr` had nothing to walk. It could not have run regardless: this repository has no browser-provider or test-environment descriptor installed, and the behaviour under test needs a genuine (or virtual) WebAuthn authenticator, which is precisely what no API- or browser-level harness here can produce.

That is **not** a case for `skip-qa`. The automated-verification exemption requires the API surface to be unchanged, and this PR deliberately narrows the request shape of two authentication routes. The PR keeps `needs-qa`, and manual QA should confirm two things on a real environment before merge: a genuine passkey login still completes, and a genuine passkey sudo step-up still issues a sudo token.

**⏱️ CI:** no repository workflows have started on this PR — only `license/cla` (passed) and the `[code]smith` check are reporting, which is expected for a fork PR awaiting maintainer approval to run workflows. No CI follow-up will come from this agent; a maintainer approving the workflow run is the next step, and Linux CI is the authority on the `yarn test` blocker above.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-14T07:50:39Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review — `fix(security): require a WebAuthn assertion for passkey verification (#3852)`

**Verdict: 🔴 Changes requested** (this repo requests changes on Medium findings too)
**Findings:** 0 blocker · 0 major · 2 medium · 2 low
**Validation source:** GitHub checks (fork workflow runs were approved as part of this review). `Mutation tests` ✅ · `lint` ✅ · `ds-lint` ✅ · `audit-scope` ✅ · `license/cla` ✅ · `test` and `ephemeral-integration` still running at review time. No local substitutes were run for those.

The core change is right and the reasoning behind it is unusually good. Collapsing `verifyPayloadSchema` to a single `{ response }` object, dropping the context-absent `credentialId` fallback, and deleting the trailing string-compare block genuinely leaves `verifyAuthenticationResponse` as the only route to a positive verdict. The choice of `safeParse` over `parse` — so a rejected payload stays a 401 and still reaches `registerFailedAttempt` instead of becoming a logged 500 — is exactly the right call, and the four replacement unit tests cover the removed shapes properly (including asserting `verifyAuthenticationResponse` is never reached). The `BACKWARD_COMPATIBILITY.md` and `UPGRADE_NOTES.md` entries are thorough and the no-bridge decision is correctly argued.

Two Medium findings. The first is a hole one line below the `safeParse` that defeats the very property `safeParse` was chosen to protect. The second is that the shipped docs understate the residual setup-side gap in a way that will misdirect operators.

---

## 🟠 Medium

### M1 — `verifyAuthenticationResponse` throws on the ordinary failure cases, so failed passkey verifications return **500** and never increment the challenge attempt counter

`packages/enterprise/src/modules/security/lib/providers/PasskeyProvider.ts:301`

The call is unguarded. In `@simplewebauthn/server@^13.3.2`, `verifyAuthenticationResponse` **throws** for nearly every failure mode and only returns `{ verified: false }` when the payload is fully well-formed and just the signature fails. Verified against the installed version:

```
empty response object: THREW -> Error: Missing credential ID
challenge mismatch:    THREW -> Error: Unexpected authentication response challenge "WRONG-CHALLENGE", expected "RIGHT-CHALLENGE"
```

Other throw sites in the same function include origin mismatch, non-base64url credential ID, wrong `type`, and non-string `clientDataJSON` / `authenticatorData` / `signature`.

`MfaVerificationService.verifyChallenge` does not wrap `provider.verify` (`services/MfaVerificationService.ts:189`), so the throw escapes to the route's `catch` → `mapMfaError` → `logger.error(...)` + **500** (`api/mfa/_shared.ts:117-126`). The `await this.registerFailedAttempt(challenge)` at `MfaVerificationService.ts:210` is never reached.

**Impact.** With a prepared challenge, a client can post `{ response: { … } }` carrying any wrong challenge or malformed field and get a 500 with `mfa_challenges.attempts` unchanged — repeatable until the challenge TTL expires, with `securityConfig.mfa.maxAttempts` never tripping and no `expires_at` clamp. The route's `rateLimit: { points: 10, duration: 60 }` throttles the rate but does not restore the per-challenge lockout. It also turns the ordinary "wrong passkey" user path into a 500 plus an error-level log line.

This behavior predates the PR — the old `{ response }` branch made the same call — but it was previously reachable only by a client not using the trivial bypass. It is now the *only* verification path, and it directly negates the stated rationale for `safeParse` ("keeps the 401 the route already documents and preserves lockout behavior").

**Fix** (in `PasskeyProvider.verify`):

```ts
let verification: Awaited<ReturnType<typeof verifyAuthenticationResponse>>
try {
  verification = await verifyAuthenticationResponse({ /* unchanged */ })
} catch {
  return false
}
```

Please add a unit test for it — make the `verifyAuthenticationResponse` mock reject and assert `verify()` resolves `false` rather than throwing.

### M2 — Both new doc sections describe the residual setup-side gap as a lockout problem, but it is still an exploitable MFA bypass

`BACKWARD_COMPATIBILITY.md` (Migration path for existing data) · `UPGRADE_NOTES.md` (Operator action)

The docs say credentials enrolled through `setupConfirmationPayloadSchema`'s `{ credentialId, publicKey, challenge }` member "carry a client-supplied public key that no authenticator can sign for" and are "permanently unverifiable", framing the remedy as a lockout cleanup.

That is true for honest QA fixtures (the repo's own `enrollPasskey` helper stores `Buffer.from('public-key:…')`). It is **not** true for an attacker, who supplies the public key of a keypair they generated and can then produce a genuine, software-signed assertion that `verifyAuthenticationResponse` accepts. Nothing about the enrolled row is second-class — after this PR it is the *strongest* kind of credential, because it is the one whose private key the attacker holds.

Reachability, after this PR:

- `PUT /api/security/mfa/provider/passkey` carries only `metadata.PUT = { requireAuth: true }` plus `authorizeMfaEnrollmentMutation` (`security.mfa.manage`, or the compelled-enrollment recovery path).
- `checkAuthorization` in `apps/mercato/src/app/api/[...slug]/route.ts:150` only asserts a session exists.
- `mfa_pending` is checked in exactly three places — `api/mfa/verify/route.ts:27`, `api/mfa/prepare/route.ts:25`, `api/mfa/recovery/route.ts:25`. The enrollment route is not one of them.

So the 10-minute `mfa_pending` token minted by `security.auth.login.mfa-challenge` (`api/interceptors.ts:71-84`) — which a password-only attacker gets — is accepted by the enrollment route. Enroll a self-made passkey, then verify against it. For a victim whose existing second factor is not a passkey, `findMethod` (`MfaVerificationService.ts:287`, `findOne` with no `orderBy`) resolves to the attacker's new credential deterministically, and `assertMethodAllowedByPolicy` passes whenever policy allows `passkey`.

**This is not a reason to hold the PR** — closing the verify-side hole is a strict improvement and should land. But the docs shipped *in this diff* should say plainly that the setup-side branch remains an open second-factor bypass, not merely a source of dead credentials, so operators triage it with the right urgency. Please also open the follow-up issue the PR body promises and link it from both files.

---

## 🔵 Low

### L1 — Mid-test `test.skip(true, …)` leaves the rest of TC-SEC-006/007 as unreachable code and reports the whole test as skipped

`__integration__/TC-SEC-006.spec.ts:131` · `__integration__/TC-SEC-007.spec.ts:173`

The runtime `test.skip(true, …)` fires *after* the new `expect(…status).toBe(401)` assertions, so those assertions execute — but the test is then recorded as **skipped**, which means the new regression signal does not show up as coverage in the report, and the sudo-token-gated remainder becomes code that is compiled and linted but never run.

Consider splitting: keep the 401 refusal as its own passing test, and move the sudo-gated remainder into a `test.fixme(…)` (or a `test.skip` at declaration level) with the same reason string. The gap stays visible in the report and the regression stays green.

### L2 — Passkey verification now has no happy-path coverage at any level

The unit tests mock `verifyAuthenticationResponse`, so nothing exercises a real assertion; the integration suite now only asserts refusals. The PR is right that an API-level test cannot produce a genuine assertion, but `TC-SEC-004` already `test.skip`s non-Chromium, so Playwright's CDP `WebAuthn.addVirtualAuthenticator` is available in exactly the browser this suite runs. Worth a tracked follow-up rather than work in this PR.

---

## ✅ Checked and clean

- **Breaking-change protocol** — the removed request shape is contract surface, the no-bridge decision is documented in `BACKWARD_COMPATIBILITY.md` with an explicit rationale, and `UPGRADE_NOTES.md` gives the client migration. This is the deprecation protocol's documented escape hatch used correctly.
- **No other caller sends the removed shape** — no `credentialId`-based verify payload anywhere outside `__tests__`/`__integration__`; `PasskeyChallengeVerify.tsx` sends `{ response }`.
- **`verifySchema` narrowing is not a contract break** — it is only carried through `createMfaProviderSetup` (`mfa-provider-interface.ts:109`) and never `.parse()`d at the route layer nor serialized into a response, so the provider's own `safeParse` is the sole gate. The PR's claim here is accurate.
- **Untouched behavior stays untouched** — challenge TTL check, `method.userId !== userId` guard, signature counter update, and the metadata `credentialPublicKey`/`publicKey` fallback are all preserved verbatim.
- **No duplicate work** — `origin/develop` does not contain this fix.
- **No conflicts** — `mergeable: MERGEABLE` against `develop`.
- **Tests ship with the behavior change** — the bypass-asserting test is replaced rather than deleted, and the four new cases all fail on the pre-fix code.

---

pkarw:
🏷️ Label rationale (the PR opened from a fork by a read-only account, so none could be applied at open time):

- `changes-requested` — review found two Medium findings (unguarded `verifyAuthenticationResponse` throw defeating the challenge attempt counter; docs understating the residual setup-side bypass).
- `priority-high` — closes an authentication bypass in the MFA/sudo second factor.
- `risk-high` — touches auth/session verification and deliberately narrows a documented API request shape.
- `bug`, `security`, `enterprise` — category.
- `needs-qa` — the automated-verification exemption does not apply: the API request surface changed and a `BACKWARD_COMPATIBILITY.md` contract is deliberately broken, so a real passkey login and a real passkey sudo step-up must be exercised with a genuine authenticator before merge.

pkarw:
## 🔧 Autofix applied — `3889dd2`

Both Medium findings are fixed on this branch. @Paul-Mlodochowki — your fork has `maintainerCanModify` enabled, so I pushed a follow-up commit rather than opening a replacement PR; your commits and authorship are untouched. Revert it if you'd rather take a different approach.

**M1 — `verifyAuthenticationResponse` throw path.** Wrapped in `try/catch` returning `false`, with a `debug`-level log so an `rpId`/`expectedOrigins` misconfiguration is still diagnosable instead of silently 401-ing every legitimate login. Failed verifications now stay 401 and reach `registerFailedAttempt`, which is what the `safeParse` choice was already reaching for. Added a regression test: `verifyAuthenticationResponse` mock rejects → `verify()` resolves `false` and the signature counter is left alone.

**M2 — doc accuracy.** `BACKWARD_COMPATIBILITY.md` and `UPGRADE_NOTES.md` now separate the two populations that land in the legacy setup branch: fixtures with an unsignable key (a lockout story) versus an attacker-supplied keypair (still a live bypass). Both link the follow-up.

**Follow-up filed:** #5296 — the enrollment side still accepts an unattested client-supplied `publicKey`, and `PUT /api/security/mfa/provider/passkey` checks `requireAuth` but not `mfa_pending`, so the second-factor bypass is not fully closed until that ships. It carries the full attack chain, the reachability trace, and a five-part fix proposal including replacing the `enrollPasskey` fixture with a CDP virtual authenticator (which also answers L2).

**Not fixed — left for you, neither blocks approval:**
- **L1** — the mid-test `test.skip(true, …)` in TC-SEC-006/007. Splitting the 401 refusal into its own passing test and marking the sudo-gated remainder `test.fixme` would keep the new regression signal visible in the report; I left it alone rather than restructuring your integration specs.
- **L2** — passkey happy-path coverage, tracked in #5296.

**Validation:** `@open-mercato/enterprise` targeted suites `PasskeyProvider|MfaVerificationService|MfaService` → **45/45 pass** (local runner; no compose `app` container up). `typecheck` shows three errors — two `#generated/entities.ids.generated` shims and one in `record_locks/.../page.tsx` — all identical with my changes stashed, i.e. pre-existing artifacts of a worktree without generated files, none in a changed file. The authoritative signal is CI, which went **fully green on `ab8dd82`** once the fork workflow runs were approved; the runs for `3889dd2` are approved and in flight.

For the record on the earlier note in this thread: the `yarn test` gate does pass — CI on Linux is green. The failures reported from the Windows machine were local environment artifacts.

pkarw (approved):
# 🔍 Re-review — `fix(security): require a WebAuthn assertion for passkey verification (#3852)`

**Verdict: ✅ Approved**
**Head:** `3889dd2` · **Findings now open:** 0 blocker · 0 major · 0 medium · 2 low
**Validation source:** GitHub checks. `prepare` ✅ · `lint` ✅ · `ds-lint` ✅ · `audit-scope` ✅ · `test` ✅ · `ephemeral-integration (none)` ✅ · `license/cla` ✅ (`audit`, `docker-build`, `merge-coverage`, `[code]smith` skipped; the non-required `Mutation tests` workflow is still queued and was green on the prior head). No local substitutes were needed.

Both Medium findings from the first pass are resolved in `3889dd2`.

- **M1 resolved** — `verifyAuthenticationResponse` is now wrapped, returning `false` on a throw with a `debug` log so an `rpId`/`expectedOrigins` misconfiguration stays diagnosable. Failed verifications answer 401 again and reach `registerFailedAttempt`, so the per-challenge lockout holds on the sole verification path. Covered by a new test asserting the rejection resolves `false` and leaves the signature counter untouched.
- **M2 resolved** — `BACKWARD_COMPATIBILITY.md` and `UPGRADE_NOTES.md` now distinguish fixture-provisioned credentials (a lockout story) from attacker-enrolled ones (still a live bypass), and both link the follow-up.

⚠️ **Disclosure:** the M1/M2 fixes were authored by me during this review and pushed to the branch, so that commit has not been independently reviewed by a second pair of eyes. A maintainer merging this should read `3889dd2` on its own merits. Everything through `ab8dd82` is @Paul-Mlodochowki's work and was reviewed normally.

⚠️ **This PR does not fully close #3852.** It closes the verify side. The enrollment side still accepts an unattested, client-supplied `publicKey`, and `PUT /api/security/mfa/provider/passkey` checks `requireAuth` but not `mfa_pending` — so a password-only attacker can still enrol their own passkey and clear the second factor. Tracked in **#5296** with the full chain and a fix proposal. Merging this is still strictly better than not merging it; #5296 should be picked up promptly rather than treated as cleanup.

## 🔵 Low — left for the author, non-blocking

- **L1** — the mid-test `test.skip(true, …)` in `TC-SEC-006.spec.ts:131` / `TC-SEC-007.spec.ts:173` fires after the new `expect(…).toBe(401)` assertions, so those pass but the test is recorded as **skipped** and the remainder is compiled-but-never-run. Splitting the refusal into its own test and marking the sudo-gated remainder `test.fixme` keeps the regression signal visible.
- **L2** — no passkey happy-path coverage at any level; the unit tests mock `verifyAuthenticationResponse` and the integration suite now only asserts refusals. `TC-SEC-004` already skips non-Chromium, so a CDP virtual authenticator would restore it. Tracked as item 4 of #5296.

## Merge gate

`needs-qa` stays on. Code-review approval is not QA approval, and this touches login MFA and sudo step-up — a reviewer must confirm a real passkey login and a real passkey sudo step-up still succeed with a genuine authenticator before `qa-approved` clears the gate. Routes to exercise are in the manual-QA comment below.

---

pkarw:

pkarw:

**You need a real authenticator for this one.** The change makes a genuine WebAuthn assertion the only way to pass passkey verification, and the integration suite lost its passkey happy path as a result (there is no API-level way to forge an assertion). A platform authenticator — Touch ID, Windows Hello, or Chrome DevTools → **WebAuthn** tab → *Add virtual authenticator* (ctap2, internal, resident keys + user verification on) — is enough. A passkey enrolled by a script or fixture will **not** work and is not a valid test of this change.

### P0 — Passkey login MFA

**Where to click**
- `/backend/settings/security` (or the profile security page) → *Add method* → **Passkey** — enroll with a real authenticator
- Log out, then `/login` → sign in with email + password → the MFA challenge screen → **Passkey**

**What to verify**
- Enrollment completes and the new passkey appears in the methods list with its label
- The MFA challenge screen offers Passkey and the browser prompts the authenticator
- Approving the prompt → login completes and lands on `/backend`
- **Dismissing/cancelling the authenticator prompt → a clean 401-style error on screen, not a spinner that hangs and not a 500 page**
- Repeat the cancel 3–4 times against one challenge → the challenge locks out after `mfa.maxAttempts` and forces a fresh login rather than allowing unlimited retries *(this is the specific behavior `3889dd2` restores — please exercise it deliberately)*
- A second passkey enrolled on the same account also verifies

**What can go wrong**
- Verification fails for a genuinely valid passkey — would point at `rpId`/`expectedOrigins` config not matching the host you are testing on. Check the server log for a `security:passkey` debug line before assuming the credential is bad
- Failed attempts return **500** instead of 401, or the attempt counter never advances
- A previously enrolled *fixture/script* passkey still verifies — that would mean the bypass is not closed

### P0 — Passkey sudo step-up

**Where to click**
- With a passkey enrolled, trigger a sudo-protected action — e.g. `/backend/settings/security` sudo configs, or an admin MFA reset at `/backend/settings/users/{id}`

**What to verify**
- The step-up dialog offers Passkey, the authenticator prompts, and approving it issues a sudo token so the protected action completes
- Cancelling the prompt → the action is refused with a clear message and no sudo token is issued

**What can go wrong**
- Sudo step-up cannot be completed at all with a real passkey — this path shares `MfaVerificationService.verifyChallenge`, so a regression here mirrors the login one
- The protected action succeeds without a sudo token

### P1 — Other MFA methods are unaffected

**Where to click**
- Enroll and log in with **TOTP**, and with **email OTP**

**What to verify**
- Both still enroll and verify normally, and their failed-attempt lockout still behaves as before

**What can go wrong**
- Collateral damage from the shared verify path — a wrong code returning 500, or the attempt counter not advancing

### Note for the reviewer

Per **#5296**, the passkey *enrollment* path still accepts an unattested client-supplied public key, so QA passing here does **not** mean #3852 is fully closed. Please do not treat this PR as closing the bypass end to end.

### Pass/fail

pkarw:

Two Medium findings on the first pass (unguarded `verifyAuthenticationResponse` throw defeating the challenge attempt counter; docs understating the residual setup-side bypass), both fixed in one autofix iteration and pushed as `3889dd2`. CI green on the final head. Follow-up filed as #5296. Two Low findings left for the author, listed in the re-review. `needs-qa` retained — the QA-approval gate holds the merge.

pkarw:
@Paul-Mlodochowki sorry but we can't accept external contributions to the `enterprise` module :( due to licensing issues (enterprise is not open source, it's priopriatery) thanks for the contribution what can i do instead is to rewrite this fix so we have no IP problems

Paul-Mlodochowki:
@pkarw Thanks for letting me know! Totally understandable. I appreciate the clarification regarding the enterprise module and licensing. I'll take a look at other issues. Thanks again!

PR #5295, opened 2026-08-14
feat(create-app): harden standalone harness gates and resilience

pkarw:
Refs #5294
Tracking plan: .ai/runs/2026-08-14-standalone-harness-optimizations.md
Source doc: .ai/specs/2026-08-14-standalone-harness-optimizations.md
Status: complete

## 🎯 Goal

- Harden standalone-app harness sessions against interrupted work and add deterministic scaffold quality gates, stop-cause evidence, and bounded framework-contract context.

## What Changed

- Added resumable plan reconciliation and slice-level/atomic progress-ledger writes.
- Added deterministic scaffold design-system and advisory i18n checkers, with exact one-use baseline ignores and regression coverage for TSX generics, interpolated literals, Tailwind arbitrary variants, and non-class IPv6 literals.
- Bound sanitized shared-session stop causes and structured provider errors into judge evidence and termination reporting.
- Added the bounded framework-contract guide, refreshed source-link knowledge, and expanded the final harness catalog to 231 contiguous cases / 49 writable cases.
- Preserved #5293's per-module `index.md` fact-sheet scattering and #5298's Node-only instrumentation behavior.
- Removed the requester-declined mandatory ephemeral integration exit gate while preserving unrelated general-purpose integration-test guidance and coverage.

## 🧪 Validation

- Focused merged-head suites: 204/204 contract tests, 22/22 DS/i18n/budget tests, 25/25 stop-cause/share/framework/memory tests, and 35/35 fact-sheet/overlay/injection tests passed.
- Full create-app suite: 800 passed, 5 skipped, 0 failed (805 total).
- Source-link inventory: 138 records, 30 owners, 115 rendered links; catalog: 231 cases / 49 writable.
- Knowledge-change proof passed: the base-plus-test patch fails and the final head passes.
- Ordered local gate passed both `build:packages` runs, `generate`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, and `build:app`.
- The aggregate local `yarn test` graph encountered macOS Jest-worker `SIGSEGV`s in two untouched CLI suites after 1,416 executed assertions passed; both suites then passed in-band (2/2 suites, 270/270 tests). The final hosted `test` check passed in full.
- Final-head GitHub checks passed, including `test`, JavaScript/TypeScript CodeQL, Python/Actions CodeQL, lint, DS lint, scope, prepare, skills-tier validation, and CLA. Inapplicable jobs were skipped.
[...]

PR #5301, opened 2026-08-14
fix(create-app,cli): make standalone harness gates fail loudly

pkarw:
## Summary

Makes a generated standalone app's validation gates fail when they should, and makes the
`mercato generate` conventions that currently fail **silently** produce warnings.

Derived from a real agent session in a generated app: a module was scaffolded, self-certified
"PRODUCTION READY", and shipped with 100 TypeScript errors and a build that failed on first
page load — while reporting `typecheck ✓ · lint ✓ · test ✓ · build ✓`. None of those gates had
passed. `typecheck` had crashed with an out-of-memory error, `lint` never enables
`import/no-unresolved`, and `test` exits 0 on an empty suite.

The guidance the agent needed already existed. What was missing was anything that **fails**
when guidance is ignored.

## Generator warnings (`packages/cli`)

Three conventions where a wrong name yields a successful run, a correct-looking manifest, and
a runtime-only defect:

| Convention | Current silent effect |
|---|---|
| `page.meta.ts` exporting `meta`, not `metadata` | **Page ships with NO authorization gate** — `requireAuth`/`requireFeatures` dropped |
| `registerCommand` reachable only from an uncalled function | ids appear in the manifest; handlers never register |
| `di.ts` exporting other than `register` | the module's DI registrations never run |

Plus the OpenAPI regex fallback, promoted from a log line to a warning — its usual cause is a
route importing an unresolvable specifier, and its effect is that every request/response
schema is silently omitted.

Each message names the runtime consequence rather than restating the rule. All are warnings,
never hard failures — existing apps may rely on the current tolerance.

## Harness (`packages/create-app`)

- **Gate honesty** in `om-module-scaffold/references/verification.md`: an OOM abort is a
  failure to report; `No tests found` is not a pass; lint says nothing about import
  resolution; a pipe reports the pipe's exit status; a written completion claim is not a gate.
- **Contract routing** to the installed `@open-mercato/ui` and `@open-mercato/shared`
  `AGENTS.md`, which already document the exact import path for every primitive, plus the
  backend route convention (the module id is *not* a path segment, unlike API routes).
- **Gate-evidence hook**, modelled on the existing `entity-migration-check`: records gate exit
  statuses and blocks concluding when source changed this session without a passing typecheck
  since.

## Deliberate non-changes

**The root `AGENTS.md` is untouched, byte-for-byte.** The gate-honesty rules were originally
added there and had to be moved. `STANDALONE_ROOT_TARGET_BYTES` is 12 KiB and the generated
root is that template plus an inline 49-module fact index; measured against `origin/develop`
the budget suite is 8/0, and adding as little as **86 bytes** fails it. The failure offers
reclaiming root bytes or accepting the pointer-form fallback — the first means deleting
routing rules this PR does not own, the second degrades the module-fact index to pointer form
for every classic scaffold. Neither is worth one rule, so the rules live in `verification.md`
instead, which is mandatory reading before claiming a scaffold complete and has no budget.

**`import/no-unresolved` is not enabled**, though it was in scope. Measured both directions in
a real generated app: the resolver reports the *valid* `@open-mercato/ui/backend/CrudForm` as
unresolved just as it does the invented `@open-mercato/ui/crud/form` — with the inherited
`eslint-config-next` resolver and with an explicit `typescript` resolver. `@open-mercato/ui`
ships a subpath `exports` map and the template uses `moduleResolution: bundler`; the resolver
honors neither. Shipping it would fail on correct code, and a gate that cries wolf gets
disabled. Rationale recorded in the run plan.

**The typecheck heap fix is not duplicated.** #5295 already carries it; this PR makes no
change to `package.json.template` and depends on #5295 for that coverage.

## Validation

Full gate, each command checked on its own exit status:

| Gate | Result |
|---|---|
| `yarn build:packages` | PASS |
| `yarn generate` | PASS |
| `yarn i18n:check-sync` | PASS |
| `yarn i18n:check-usage` | PASS |
| `yarn typecheck` | PASS |
| `yarn test` | PASS |
| `yarn build:app` | PASS |

Tests added: 12 generator cases and 10 hook cases. Every warning is asserted both firing
**and** staying silent on correct code, so a diagnostic that fires on valid input fails the
suite. Hook cases cover gate matching across compound (`a && b && c`) and direct (`npx tsc`)
invocations, plus the docs-only-session false positive.

Tracking plan: `.ai/runs/2026-08-14-harness-verification-gates.md`

Status: complete

### DIFF
--- a/.ai/runs/2026-08-14-harness-verification-gates.md
+++ b/.ai/runs/2026-08-14-harness-verification-gates.md
@@ -95,29 +95,48 @@ changed after session start and is newer than the last exit-0 typecheck.
 
 ### Phase 1: Gate honesty
 
-- [ ] 1.1 Add gate-evidence rules to the template AGENTS.md validation section
+- [x] 1.1 Add gate-evidence rules to the template AGENTS.md validation section — 578888fcc
 
 ### Phase 2: Route to authoritative contracts
 
-- [ ] 2.1 Route om-module-scaffold to installed @open-mercato/ui AGENTS.md
-- [ ] 2.2 Route om-module-scaffold to installed @open-mercato/shared AGENTS.md
-- [ ] 2.3 Document the backend page route convention with a worked example
+- [x] 2.1 Route om-module-scaffold to installed @open-mercato/ui AGENTS.md — 578888fcc
+- [x] 2.2 Route om-module-scaffold to installed @open-mercato/shared AGENTS.md — 578888fcc
+- [x] 2.3 Document the backend page route convention with a worked example — 578888fcc
 
 ### Phase 3: Lint sees imports
 
-- [ ] 3.1 Investigate resolver support for subpath exports maps
-- [ ] 3.2 Enable import/no-unresolved, or abandon with recorded reason
+- [x] 3.1 Investigate resolver support for subpath exports maps — investigated, not landable
+- [x] 3.2 Enable import/no-unresolved, or abandon with recorded reason — **abandoned, reason below**
+
+**Phase 3 abandoned.** Measured in a real generated app on 2026-08-14, both directions:
+
+| import | expected | actual |
+|---|---|---|
+| `@open-mercato/ui/backend/CrudForm` (valid) | pass | **`Unable to resolve path to module`** |
+| `@open-mercato/ui/crud/form` (invented) | fail | fail |
+
+Tried with the inherited `eslint-config-next` resolver, and again with an explicit
+`import/resolver: { typescript: { alwaysTryTypes: true, project: './tsconfig.json' } }`.
+The valid import fails in both. `@open-mercato/ui` ships a subpath `exports` map and the
+template uses `moduleResolution: bundler`; the resolver honors neither, so it reports every
+`@open-mercato/*` subpath as unresolved.
+
+Enabling the rule in that state would fail on correct code, which is worse than leaving it
+off — a gate that cries wolf gets disabled, and then it protects nothing. The underlying
+defect is already covered: with the typecheck heap fix (#5295) `tsc` reports the same thing
+as `TS2307`. Landing this needs a resolver that understands `exports` maps, which is its own
+piece of work rather than a config tweak.
 
 ### Phase 4: Generator warnings
 
-- [ ] 4.1 Warn when page.meta.ts exports no metadata
-- [ ] 4.2 Warn when registerCommand is unreachable at import time
-- [ ] 4.3 Warn when di.ts exports no register
-- [ ] 4.4 Surface the OpenAPI regex fallback as a counted warning
-- [ ] 4.5 Fixture tests for each warning, firing and silent cases
+- [x] 4.1 Warn when page.meta.ts exports no metadata — bbe6988e7
+- [x] 4.2 Warn when registerCommand is unreachable at import time — bbe6988e7
+- [x] 4.3 Warn when di.ts exports no register — bbe6988e7
+- [x] 4.4 Surface the OpenAPI regex fallback as a counted warning — bbe6988e7
+- [x] 4.5 Fixture tests for each warning, firing and silent cases — bbe6988e7
 
 ### Phase 5: Gate-evidence hook
 
-- [ ] 5.1 Add the gate-evidence hook to the agentic hooks template
-- [ ] 5.2 Register it in the generated hook settings
-- [ ] 5.3 Unit tests for the comparison and gate-command matching
+- [x] 5.1 Add the gate-evidence hook to the agentic hooks template — bbe6988e7
+- [x] 5.2 Register it in the generated hook settings — bbe6988e7
+- [x] 5.3 Unit tests for the comparison and gate-command matching — bbe6988e7
--- a/.ai/runs/2026-08-14-harness-verification-gates.md
+++ b/.ai/runs/2026-08-14-harness-verification-gates.md
@@ -140,3 +140,28 @@ piece of work rather than a config tweak.
 - [x] 5.1 Add the gate-evidence hook to the agentic hooks template — bbe6988e7
 - [x] 5.2 Register it in the generated hook settings — bbe6988e7
 - [x] 5.3 Unit tests for the comparison and gate-command matching — bbe6988e7
+
+### Phase 6: Review follow-up (PR #5301)
+
+Code review found one Major and seven Minor issues, four of them ways the hook could record
+a gate as passed without observing a passing exit status — the property this run exists to
+remove. Fixed in the same PR.
+
+- [x] 6.1 Install Claude hooks from disk in `mercato agentic:init`, and claim them in the
+  `--update-harness` ownership manifest. The CLI generator copied only
+  `entity-migration-check.ts` while shipping a `settings.json` that registers
+  `gate-evidence.ts`, so apps set up through that path got a hook registration pointing at a
+  file that was never written.
+- [x] 6.2 Treat an unreported exit status as unknown rather than as `0`.
+- [x] 6.3 Do not record a gate whose exit status belongs to something else — a pipe, a `;`
+  sequence, or `|| true`. This is the failure `verification.md` warns about, so recording it
+  would have re-created it.
+- [x] 6.4 Reset the session record on a new `session_id`, so `sessionStartedAt` cannot stay
+  pinned to the first session and block a later one over somebody else's unverified edits.
+- [x] 6.5 Honor `stop_hook_active` so the Stop hook blocks at most once per stop sequence.
+- [x] 6.6 Gitignore `.ai/.gate-state.json` in the template.
+- [x] 6.7 Honor `quiet` in the page-metadata and registerCommand warnings, and name each
+  offending path once per run instead of once per registry emitter.
+- [x] 6.8 Fold the OpenAPI fallback log into the warning; drop redundant inline comments.
+- [x] 6.9 Tests for every item above, plus a hook-parity test asserting the two generators
+  install the same hook set — the class of defect 6.1 belongs to.
--- a/.ai/runs/2026-08-14-harness-verification-gates.md
+++ b/.ai/runs/2026-08-14-harness-verification-gates.md
@@ -1,0 +1,123 @@
+# Run: harness verification gates
+
+Source doc: `.ai/specs/2026-08-14-harness-verification-gates.md` (authored in a standalone app; carried over — see Provenance)
+
+## Goal
+
+Make a generated standalone app's validation gates fail when they should, and make the
+`mercato generate` conventions that currently fail *silently* produce warnings — so an agent
+cannot report a green gate it never ran, and a single typo cannot silently drop a backend
+page's authorization.
+
+## Provenance
+
+Derived from a real agent session in a generated standalone app. A `library` module was
+scaffolded, self-certified "PRODUCTION READY", and shipped with 100 TypeScript errors and a
+build that failed on first page load. The claimed evidence line was:
+
+> Gate: `generate` ✓ · `typecheck` ✓ (0 errors) · `lint` ✓ (0 errors) · `test` ✓ (no unit tests; baseline) · `build` ✓
+
+None of those gates had passed. `typecheck` had crashed with an out-of-memory error, `lint`
+never enables `import/no-unresolved`, and `test` exits 0 on an empty suite. The failures
+below are labelled F1–F10 to match the source spec.
+
+## Scope
+
+In scope — the gaps PR #5295 does not cover:
+
+| Phase | Failure | Target |
+|---|---|---|
+| 1 | F3, F10 — vacuous gates reported as passes | `packages/create-app/agentic/shared/AGENTS.md.template` |
+| 2 | F4, F8 — invented import paths, wrong route convention | `packages/create-app/agentic/shared/ai/skills/om-module-scaffold/references/` |
+| 3 | F2 — lint blind to unresolved imports | `packages/create-app/template/eslint.config.mjs` *(investigation; may be abandoned)* |
+| 4 | F5, F6, F7, F9 — conventions that fail silently | `packages/cli/src/lib/generators/` |
+| 5 | F10 — nothing distinguishes "claimed" from "ran" | `packages/create-app/agentic/*/hooks/` |
+
+## Non-goals
+
+- **F1 (typecheck heap) is NOT in this run.** PR #5295 already sets
+  `"typecheck": "cross-env NODE_OPTIONS=--max-old-space-size=8192 tsc --noEmit"` in
+  `package.json.template`. Duplicating it here would only create a conflict. This run
+  depends on #5295 landing for F1 coverage.
+- No changes to an existing generated app; template and generator only.
+- Generator conventions become **warnings**, never hard failures — existing apps may rely on
+  current tolerance. Escalation to errors is a separate follow-up.
+- No changes to the validation-gate command list itself.
+
+## Risks
+
+- Phase 4 edits a generator with snapshot/structural-contract tests; a warning that fires on
+  a legitimate pattern would be noise across every module. Mitigated by fixture tests
+  asserting both the firing and the silent case.
+- Phase 3 may be unlandable: the resolver inherited from `eslint-config-next` currently
+  reports valid `@open-mercato/*` subpath imports as unresolved. If a two-way test cannot
+  pass, the phase is abandoned with the reason recorded (this satisfies the source spec's
+  AC-2 either way).
+- Phase 1 touches a file #5295 also edits. Textual conflict is possible but small.
+
+## Implementation Plan
+
+### Phase 1 — Gate honesty (F3, F10)
+
+Add to the template's `AGENTS.md` validation rules: a gate counts as passed only when its
+command ran to completion and its exit status is reported; `No tests found` is explicitly not
+a pass; a gate that crashes (OOM included) is a failure to report, never a step to skip.
+
+### Phase 2 — Route to authoritative contracts (F4, F8)
+
+Point `om-module-scaffold` at the installed `@open-mercato/ui` and `@open-mercato/shared`
+`AGENTS.md` files before UI/API work, and document the backend route convention
+(`backend/<segments>/page.tsx` → `/backend/<segments>`; the module id is not a path segment,
+while API routes *are* module-namespaced).
+
+### Phase 3 — Lint sees imports (F2)
+
+Investigate whether the resolver can honor subpath `exports` maps under
+`moduleResolution: bundler`. Two-way test required: a correct import must pass AND an
+invented one must fail. Abandon with a recorded reason if it cannot.
+
+### Phase 4 — Generator warnings (F5, F6, F7, F9)
+
+Warn when: `page.meta.ts` exports no `metadata` (currently drops `requireAuth` /
+`requireFeatures` silently — the security-relevant one); a `commands/*.ts` `registerCommand`
+is unreachable at import time; `di.ts` exports no `register`; and the OpenAPI bundle step
+falls back to regex. Each message names the runtime consequence, not just the rule.
+
+### Phase 5 — Gate-evidence hook (F10)
+
+Ship a `gate-evidence` hook in the agentic hooks template: a recorder (PostToolUse on Bash)
+that writes gate exit statuses, and a blocker (Stop) that fires only when a `src/**` file
+changed after session start and is newer than the last exit-0 typecheck.
+
+## Progress
+
+> Convention: `- [ ]` pending, `- [x]` done. Append ` — <commit sha>` when a step lands. Do not rename step titles.
+
+### Phase 1: Gate honesty
+
+- [ ] 1.1 Add gate-evidence rules to the template AGENTS.md validation section
+
+### Phase 2: Route to authoritative contracts
+
+- [ ] 2.1 Route om-module-scaffold to installed @open-mercato/ui AGENTS.md
+- [ ] 2.2 Route om-module-scaffold to installed @open-mercato/shared AGENTS.md
+- [ ] 2.3 Document the backend page route convention with a worked example
+
+### Phase 3: Lint sees imports
+
+- [ ] 3.1 Investigate resolver support for subpath exports maps
+- [ ] 3.2 Enable import/no-unresolved, or abandon with recorded reason
+
+### Phase 4: Generator warnings
+
+- [ ] 4.1 Warn when page.meta.ts exports no metadata
+- [ ] 4.2 Warn when registerCommand is unreachable at import time
+- [ ] 4.3 Warn when di.ts exports no register
+- [ ] 4.4 Surface the OpenAPI regex fallback as a counted warning
+- [ ] 4.5 Fixture tests for each warning, firing and silent cases
+
+### Phase 5: Gate-evidence hook
+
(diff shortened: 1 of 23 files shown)


pkarw:
🤖 `om-auto-review-pr` claimed this PR for review at 2026-08-14T13:13Z. Lock: `in-progress` label + assignee. Will release when the review is submitted.

pkarw:
> 🤖 `om-auto-review-pr` — **verdict: CHANGES REQUESTED**. A formal GitHub review could not be submitted: `addPullRequestReview` rejects `Can not request changes on your own pull request` because the reviewer account (`pkarw`) is also the PR author. The full report is posted here instead and the pipeline label is set to `changes-requested`. A second maintainer account is needed for a formal review record.

---

# 🔍 Code Review: fix(create-app,cli): make standalone harness gates fail loudly

## 🎯 Summary

This PR attacks a genuinely important failure class: a generated standalone app whose validation gates report green without having run, and `mercato generate` conventions where a near-miss export name produces a successful run and a runtime-only defect. It does so in three places — four new warnings in `packages/cli/src/lib/generators/` (`page.meta` missing `metadata`, `registerCommand` unreachable at import time, `di.ts` missing `register`, and the OpenAPI regex fallback), gate-honesty and contract-routing rules in the `om-module-scaffold` references, and a new `gate-evidence` Claude Code hook that records gate exit statuses and blocks concluding on unverified source changes.

The generator half is the strong half, and I verified it empirically rather than by reading: scanning every `page.meta`/`meta`, `di.ts`, and module `commands/*.ts` file in this monorepo against the three new predicates produces **zero** false positives. `hasNamedExport` is AST-based rather than regex, so `export { metadata }` and destructured forms are handled. The `basename === 'shared' || basename === 'factory'` skip in `collectCommandLoaderEntries` is what keeps `packages/core/src/modules/dictionaries/commands/factory.ts` — the one file in the repo with only indented `registerCommand(...)` calls — from tripping the reachability heuristic. The `isEntryPoint()` guard on `main()` is a real and non-obvious correctness detail, since `main()` blocks on stdin and would otherwise hang any test that imported `matchGates`. Testing both the firing and the silent half of each warning is the right instinct and is what makes these warnings safe to ship.

The problems are concentrated in the hook and its installation. One is a broken generator-parity gap that ships a non-working harness through the `mercato agentic:init` path. The rest are ways the gate-evidence mechanism fails **open** — recording a green gate it did not observe — which is precisely the property the PR exists to remove.

Reviewed scope: all 12 changed files. `.ai/runs/2026-08-14-harness-verification-gates.md` is a tracking document and I reviewed it only for consistency with what landed (it is accurate, including the honest Phase 3 abandonment record).

## Verdict

❌ **Request changes** — one High finding (the `agentic:init` path installs a `settings.json` registering a hook file it never writes, breaking every Bash call and every Stop in apps set up that way) plus seven Medium findings, four of which let the gate-evidence hook record a gate as passed without observing a passing exit status. Under this repo's verdict rule, Medium findings request changes. None of this touches a protected contract surface, and the generator warnings could land as-is.

## 🧪 Validation Gate

**Validation source: GitHub checks** (head `d2cbd855`). Per this repo's `om-auto-review-pr` extension, CI check results stand in for a local gate run when they are usable for the current head; they are, so no local gate was re-run. Branch protection reports `Required status checks not enabled` (404), so every reported check is treated as required.

| Check | Status | Notes |
|---|---|---|
| `prepare` | ✅ SUCCESS | |
| `lint` | ✅ SUCCESS | |
| `ds-lint` | ✅ SUCCESS | |
| `test` | ✅ SUCCESS | Covers both new test files |
| `audit-scope` | ✅ SUCCESS | |
| `scope` | ✅ SUCCESS | |
| `CodeQL` / `Analyze (javascript-typescript, python, actions)` | ✅ SUCCESS | |
| `license/cla` | ✅ SUCCESS | |
| `merge-coverage`, `docker-build`, `ephemeral-integration`, `audit`, `mutate`, `comment`, `[code]smith` | ⏭️ SKIPPED | Not triggered for this change |

No check is failing and none is pending. I additionally ran `node --import tsx --test src/setup/gate-evidence-hook.test.ts` locally to confirm the new create-app suite executes and passes (10/10) — it does.

## Findings

### ⚠️ Major

**1. `packages/cli/src/lib/agentic-setup.ts:692` — `mercato agentic:init` installs a `settings.json` that registers a hook file it never copies.**

`generateClaudeCode` in `agentic-setup.ts` copies `settings.json` verbatim, and this PR changes that file to register `.claude/hooks/gate-evidence.ts` on both `PostToolUse:Bash` and `Stop`. But the hook enumeration on line 692 still lists only `hooks/entity-migration-check.ts`, and the `--update-harness` ownership manifest at `agentic-setup.ts:465` likewise omits the new file. The create-app wizard path was updated (`packages/create-app/src/setup/tools/claude-code.ts:45`); the CLI path was not.

Failure scenario: run `mercato agentic:init --agents claude-code` in an existing app, or `mercato agentic:init --update-harness` in one already set up. `.claude/settings.json` lands referencing `$CLAUDE_PROJECT_DIR/.claude/hooks/gate-evidence.ts`, which is never written. Every subsequent Bash tool call fires `npx tsx <missing file> record` and every Stop fires it with `check`, so the agent hits a hook error on essentially every turn. Because the path is not in the ownership manifest, `--update-harness` will never repair it either.

This violates the MUST in `packages/create-app/AGENTS.md` § Agentic Setup Maintenance: "Both generators recursively emit the same `agentic/` source tree … Keep recursive generation, ownership semantics, and tests in parity whenever the source tree changes."

Fix: add `copyFile(srcDir, 'hooks/gate-evidence.ts', join(targetDir, '.claude', 'hooks', 'gate-evidence.ts'))` alongside line 692, and add the corresponding `paths.add(...)` next to line 465. I also found **no test anywhere in the repo** asserting hook-set parity between the two generators (nothing references `entity-migration-check` from a test file), which is why nothing caught this — a test asserting the two generators emit the same `.claude/hooks/*` set would close the class, not just this instance.

### 🔹 Minor

**2. `packages/create-app/agentic/claude-code/hooks/gate-evidence.ts:587-591` — `resolveExitCode` treats an unknown exit status as success.**

```ts
const value = response.exit_code ?? response.exitCode
return typeof value === 'number' ? value : 0
```

If the Bash `tool_response` does not carry `exit_code`/`exitCode` under either spelling, or carries it as a string, every gate is recorded with `exitCode: 0` — a green typecheck the hook never observed, which then satisfies `shouldBlock` and unblocks the Stop. The correctness of the entire mechanism rests on this one field, the payload shape is not asserted anywhere in the tests, and `resolveExitCode` is neither exported nor covered. The fallback is chosen in the permissive direction, in a feature whose stated purpose is to make "claimed" and "observed" mechanically distinguishable.

Fix: treat a missing or non-numeric exit code as *unknown* and skip recording that gate entirely, rather than recording it as a pass. Export `resolveExitCode` and add cases for `exit_code`, `exitCode`, absent, and non-numeric.

**3. `gate-evidence.ts:488-501` and `613-624` — the recorder attributes the wrapper command's exit status to every gate it matched.**

`matchGates` scans the whole command string and the recorder writes one exit code to all matches. So `yarn typecheck | tail -30` matches `typecheck` and records `tail`'s status (0). `yarn typecheck; yarn lint` records `typecheck` with `lint`'s status. `yarn typecheck || true` records 0 unconditionally.

This is the same hole the PR itself documents one file away, in `packages/create-app/agentic/shared/ai/skills/om-module-scaffold/references/verification.md:772`: "**A pipe hides the exit status.** `yarn test | tail -30` reports the exit code of `tail`. Capture the status of the command itself." The hook that is supposed to make that rule mechanical reproduces the exact failure the rule warns about, and an agent following the harness's own habit of piping to `tail` will silently record a green typecheck.

Fix: when the command contains a pipe, a `;`, or a `||` after a matched gate token, do not record — the observed status does not belong to the gate. Add test cases for `yarn typecheck | tail -30`, `yarn typecheck; yarn lint`, and `yarn typecheck || true`.

**4. `gate-evidence.ts:606-611` — `sessionStartedAt` is written once and never refreshed, so the false-positive guard is dead after the first session.**

```ts
if (!state.sessionStartedAt) {
  state.sessionStartedAt = now.toISOString()
  writeState(state)
}
```

`.ai/.gate-state.json` persists on disk across sessions, and nothing resets `sessionStartedAt` — there is no `SessionStart` hook registered in `settings.json` and the hook never reads `session_id` from the payload. From the second session onward, `sessionStartedAtMs` is the *first* session's timestamp, so the `newestSrcMtimeMs < sessionStartedAtMs` branch in `shouldBlock` stops firing and the check degenerates to "is any `src/` file newer than the last green typecheck".

Failure scenario: a colleague edits `src/modules/library/data/entities.ts` on Monday without running typecheck. On Tuesday you open a docs-only session, touch nothing under `src/`, and try to stop — you are blocked with "Source under `src/` changed **this session** and has not passed a typecheck since", which is false on both counts. The unit test `shouldBlock: allows a session that changed no source, even with no typecheck record` (`gate-evidence-hook.test.ts:849`) passes because it hands the pure function a `sessionStartedAtMs` the wiring will not produce after session one; nothing tests the wiring that computes it.

Fix: derive the session boundary from the hook payload's `session_id` — reset `sessionStartedAt` (and ideally the recorded gates) when the id differs from the stored one — or register a `SessionStart` hook that clears the file. Then add a test over the state-derivation step, not only over `shouldBlock`.

**5. `packages/create-app/agentic/claude-code/settings.json:25-34` — the Stop hook blocks without honoring `stop_hook_active`.**

The `check` mode writes `{"decision":"block"}` with no reference to the `stop_hook_active` flag Claude Code sets when the model is already continuing because of a stop hook. When typecheck genuinely cannot pass — a broken dependency, a pre-existing error the agent was not asked to fix, an OOM the machine cannot avoid — every stop is blocked again, with no exit other than making the gate green. The block message explicitly forecloses the one available escape ("do not delete `.ai/.gate-state.json` to work around this"), so the loop has no documented off-ramp. This is the repo's first `Stop` hook (`entity-migration-check` is `PostToolUse`, where a block simply feeds back to the model), so there is no existing precedent to inherit the guard from.

Fix: read `stop_hook_active` from the parsed hook input and return without blocking when it is true, so the hook interrupts at most once per stop sequence. Add a `matchGates`-style unit case for the decision function covering the already-active branch.

**6. `packages/create-app/template/gitignore` — `.ai/.gate-state.json` is not ignored.**

The hook rewrites `.ai/.gate-state.json` on every Bash tool call in a generated app, so `git status` in that app is permanently dirty and session timestamps plus gate exit codes get committed. The template's gitignore already enumerates every other machine-local `.ai/*` runtime artifact under explanatory comments — `.ai/sessions*.json`, `.ai/qa/test-env.json`, `.ai/qa/test-env.lock/`, `.ai/scripts/test-env-*` — so this is the file's established convention, not a new one being invented for this PR.

Fix: add `.ai/.gate-state.json` to `packages/create-app/template/gitignore` with a one-line comment in the same style as its neighbours.

**7. `packages/cli/src/lib/generators/module-registry.ts:407`, `:415`, `:1338` — the two new warnings ignore `quiet`, unlike their sibling.**

`warnIfDiMissingRegisterExport(diPath, quiet)` correctly takes and honors the flag (`module-di.ts:311`), and the OpenAPI warning sits inside the existing `if (!quiet)` block. But `warnIfPageMetaMissingMetadataExport` and `warnIfRegisterCommandNotAtImportTime` take no `quiet` parameter and print unconditionally, even though `quiet` is available in the generator options (`module-registry.ts:87`, destructured at `:3294`, `:3951`, `:4368`) — it simply is not threaded into `processPageFiles`, `processPageFilesAst`, or `collectCommandLoaderEntries`.

This is not hypothetical: `generateModuleRegistries` is invoked with `quiet: true` at `packages/cli/src/mercato.ts:1070`, `packages/cli/src/lib/module-install.ts:112`, and `packages/cli/src/lib/generators/__tests__/output-snapshots.test.ts:782`. Callers that explicitly asked for silence now get warning output, including a snapshot test.

Fix: thread `quiet` through the three call paths and gate both warnings on it, matching `warnIfDiMissingRegisterExport`'s signature.

**8. `module-registry.ts:407` and `:415` — a single offending `page.meta.ts` warns several times per `yarn generate`.**

`generateModuleRegistries` (`module-registry.ts:4720-4728`) runs all three `…FromDiscovery` variants against one discovery, and `processPageFiles` is called twice per variant (frontend at `:3419`, backend at `:3542`) while `processPageFilesAst` is called twice more (`:4073`, `:4094`). The same metadata file is therefore re-read and re-warned on each pass, so one mistyped export produces a stack of identical lines. For a diagnostic whose whole value is being noticed rather than tuned out, repetition works against the goal.

Fix: keep a per-run `Set<string>` of already-warned paths (or hoist the check to the shared discovery step, which runs once) and emit each message at most once.

### 💅 Nit

**9. `packages/cli/src/lib/generators/openapi.ts:426-437` — two log lines for one event.**

The new `console.warn` is immediately followed by the pre-existing `console.log('[OpenAPI] Falling back to static regex approach')`, which now says strictly less than the line above it. Fold the old message into the warning and drop it.

**10. `module-registry.ts:1335-1337` and `openapi.ts:427-430` — inline comments on self-explanatory code.**

Root `AGENTS.md` § Code Quality asks for no inline comments (self-documenting code). Both blocks restate rationale the JSDoc directly above the corresponding function already carries — `openapi.ts` spends four comment lines explaining a warning whose own text explains it. Author's call, but the JSDoc is the better home.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the change is purely additive (`warnIfPageMetaMissingMetadataExport`, `warnIfRegisterCommandNotAtImportTime`, `warnIfDiMissingRegisterExport`, `matchGates`, `shouldBlock` are all new exports)
- [x] No function signature changed in a breaking way — `warnIfDiMissingRegisterExport`'s `quiet` is a defaulted second parameter on a brand-new function
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — the new output is `console.warn` on stderr, and `[OpenAPI] Falling back to static regex approach` is preserved verbatim on stdout
- [x] No config key renamed and no default changed silently — `settings.json` gains hook entries additively
- [x] Where a contract had to change: not applicable, nothing here is a protected surface under `BACKWARD_COMPATIBILITY.md`

One behavioral note that is not a break but is worth stating: generated apps that already have a `.claude/settings.json` will only pick up the Stop hook through `--update-harness`, and that path is currently broken by finding 1.

## 🧪 Test Coverage

**What is covered.** `packages/cli/src/lib/generators/__tests__/silent-convention-warnings.test.ts` exercises all three generator warnings against real files written to a temp directory, asserting both that each fires *and* that it stays silent on correct input — including four spellings of a valid `register` export (function declaration, async function, const arrow, named re-export), a `commands/` helper file that registers nothing, and a `null` metadata path. The `does not warn` half is what makes these safe to enable repo-wide, and it is present for every warning. `packages/create-app/src/setup/gate-evidence-hook.test.ts` covers `matchGates` across plain scripts, compound `&&` chains, direct `npx tsc --noEmit`, a heap-flagged invocation, and an unrelated command, plus all five branches of `shouldBlock`. I confirmed both suites run and pass (`test` check is green; the create-app suite passes 10/10 locally).

**Gaps, with the cases to add.**

- **The `agentic:init` parity gap has no test at all** (finding 1). Nothing in the repo references `entity-migration-check` from a test file, so neither generator's hook set is asserted. Add a test — `packages/create-app/src/setup/agentic-parity.test.ts` or the existing overlay test — that enumerates `agentic/claude-code/hooks/*` from the source tree and asserts both `generateClaudeCode` implementations emit every one of them, so the next hook added cannot regress the CLI path.
- **`resolveExitCode` is untested and unexported** (finding 2). Add cases for `{exit_code: 1}`, `{exitCode: 1}`, `{}`, and `{exit_code: "1"}`, asserting the unknown cases do not record a pass.
- **No test pins the recorder's command-shape handling** (finding 3). Add `yarn typecheck | tail -30`, `yarn typecheck; yarn lint`, and `yarn typecheck || true`, asserting no green typecheck is recorded from any of them.
- **The state-derivation step is untested** (finding 4). Every `shouldBlock` case feeds `sessionStartedAtMs` directly, which is why the never-refreshed `sessionStartedAt` bug survives a green suite. Add a test that drives the record/check cycle through an actual state file across two simulated sessions and asserts the second session is not blocked by the first session's unverified edits.
- **The Stop-loop guard is untested** (finding 5), because it does not exist yet. Add a case asserting no block is emitted when `stop_hook_active` is true.
- `packages/create-app/tsconfig.json` includes only `src/**/*.ts` and excludes `**/*.test.ts`, so neither `agentic/claude-code/hooks/gate-evidence.ts` (216 lines) nor its test is in any typecheck program — I confirmed with `tsc --noEmit --listFiles` that the hook does not appear. This is a pre-existing gap that `entity-migration-check.ts` shares, so I am not raising it as a finding against this PR, but it is worth noting that a PR about gates that actually verify things adds its largest new file outside the reach of `yarn typecheck`. Widening the `include` to cover `agentic/**/hooks/*.ts` would be a reasonable follow-up.

pkarw:
🏷️ **label rationale**

| Label | Why |
|---|---|
| `changes-requested` | Pipeline state. One Major (the `mercato agentic:init` path installs a `settings.json` registering `.claude/hooks/gate-evidence.ts` but never copies the file) plus seven Minor findings; this repo's verdict rule requests changes on Minor/Medium too. |
| `bug` | The change fixes a defect class — validation gates that report green without running, and generator conventions that fail silently — rather than adding a capability. |
| `priority-medium` | Ordinary severity: the defect degrades agent-authored code quality in generated apps, but nothing production-facing is broken or at risk. |
| `risk-medium` | Blast radius is wider than the diff looks: `module-registry.ts` and `module-di.ts` run on every `yarn generate`, and `agentic/claude-code/` ships into every newly scaffolded app. Mitigated by warning-only semantics and by both-directions fixture tests — I verified zero false positives against every `page.meta`/`meta`, `di.ts`, and module `commands/*.ts` file in this monorepo. |
| `skip-qa` | Automated-verification exemption applies: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, database structure and API surface unchanged, no `BACKWARD_COMPATIBILITY.md` contract touched, and automated tests for the changed behavior ship in the same PR (12 generator cases, 10 hook cases). |

Author handoff is a no-op here — the reviewer and the author are the same account (`pkarw`), which is also why no formal review record could be created.

pkarw:
🤖 `om-auto-review-pr` — **autofix complete**. All 8 findings from the review above are fixed on `c1031dfaf` (2 commits).

| # | Finding | Outcome |
|---|---|---|
| 1 (Major) | `mercato agentic:init` registered `gate-evidence.ts` in `settings.json` but never copied it | **Fixed.** Both generators now derive the hook set from disk via `claudeHookFiles()`, and the `--update-harness` ownership manifest claims every hook. The enumerated list that caused this is gone from both paths. |
| 2 | `resolveExitCode` returned `0` for an unreported status | **Fixed.** Returns `number \| null`; an unknown status records nothing rather than a pass. |
| 3 | The wrapping command's status was attributed to the gate, so `yarn typecheck \| tail -30` recorded `tail`'s success | **Fixed.** New `isAttributableGateCommand` refuses to record across `\|`, `;`, `\|\|`, or a newline. `&&` still records — it short-circuits, so a non-zero status still belongs to a gate that ran. |
| 4 | `sessionStartedAt` was never refreshed, so the "changed this session" guard died after session one | **Fixed.** New `nextSessionState` starts a clean record on a new `session_id`; payloads without one keep the old set-once behavior. |
| 5 | The Stop hook blocked without honoring `stop_hook_active` | **Fixed.** `check` returns early when the stop is already the result of a block, so it interrupts at most once per sequence. |
| 6 | `.ai/.gate-state.json` was not gitignored | **Fixed** in the template gitignore, alongside the other machine-local `.ai/*` runtime artifacts. |
| 7 | The two new warnings ignored `quiet` | **Fixed.** `quiet` is threaded through `createModuleRegistryDiscovery` → `collectCommandLoaderEntries` and through both page processors. |
| 8 | One `page.meta.ts` warned once per registry emitter | **Fixed.** A per-run ledger names each offending path once; `resetConventionWarnings()` clears it when the run's discovery is built. |
| 9, 10 (Nit) | Duplicate OpenAPI log line; inline comments on self-explanatory code | **Fixed.** |

One extra fix found while re-reviewing the patch: `matchGates` counted a gate merely *named* in a quoted string, so `git commit -m "run tsc --noEmit"` recorded a green typecheck. Quoted spans are now stripped before matching.

**Tests added (13 new cases).** `isAttributableGateCommand` across pipes / `;` / `|| true` / newlines and the `&&` chain that must still count; `resolveExitCode` across both field spellings, absent, and non-numeric; `nextSessionState` across a session change, a same-session call, the no-`session_id` fallback, and a regression case proving a stale start time cannot block a later session; quoted-string gate names; `quiet` and once-per-run behavior for both generator warnings. Plus a new `packages/create-app/src/lib/agentic-claude-hook-parity.test.ts` asserting that both generators install every shipped hook, that the CLI path derives them from disk rather than a hand-kept list, that the ownership manifest claims them, and that every hook `settings.json` registers exists on disk — which closes the class finding 1 belongs to, not just the instance.

**Verification.** `packages/cli` suite: 1432 tests pass, 0 fail. Two suites error during collection (`module-facts`), and I confirmed they error identically on the unmodified head `d2cbd855` — a local-invocation artifact, not a regression; CI's `test` job is green on that commit. `packages/create-app`: 43/43 across the gate-evidence, hook-parity, and standalone-overlay suites. `tsc --noEmit` on `packages/cli` reports nothing in any touched file.

**Known residual, deliberately not changed:** `matchGates` still matches a bare `jest` or `eslint` token, so a command like `cat jest.config.cjs` records a `test` gate. Only `typecheck` gates the Stop decision and its patterns are narrower, so the practical exposure is cosmetic state — tightening every pattern to a command position is a larger change than this review's findings called for.

**Pipeline label: `review`, not `merge-queue`.** Every finding is addressed and the branch is otherwise mergeable, but GitHub will not let me submit a formal review on my own PR, so nothing here constitutes an approval. This needs an independent maintainer review before merge.

pkarw:
🤖 `om-auto-review-pr` completed: **CHANGES REQUESTED → all findings fixed**. Lock released (`in-progress` swapped for `ci-monitoring`); the PR is free for anyone to act on. Watching CI on `c1031dfaf` and will post the result.

pkarw:

**All checks green on `c1031dfaf`** (the autofix head), including `test`, which is the one that exercises the 13 new cases and the new hook-parity suite.

| Check | State |
|---|---|
| `test` | ✅ SUCCESS |
| `lint` | ✅ SUCCESS |
| `ds-lint` | ✅ SUCCESS |
| `prepare` | ✅ SUCCESS |
| `scope` / `audit-scope` | ✅ SUCCESS |
| `CodeQL` + `Analyze (javascript-typescript, python, actions)` | ✅ SUCCESS |
| `license/cla` | ✅ SUCCESS |

Nothing failing, nothing pending. `merge-coverage`, `docker-build`, `ephemeral-integration`, `audit`, `mutate`, `comment` and `[code]smith` remain skipped, as they were before the autofix. `ci-monitoring` removed.

**Remaining blocker is not technical:** GitHub will not accept a review from `pkarw` on a PR authored by `pkarw`, so nothing in this thread is an approval. The PR needs an independent maintainer to submit one. `skip-qa` is set under the automated-verification exemption, so QA does not gate the merge.

PR #5303, opened 2026-08-14
docs(specs): SMTP transport for transactional email

bartek5412:
## Summary

Open Mercato has exactly one system transactional email path — `sendEmail()` in
`packages/shared/src/lib/email/send.ts` — and it hardcodes the Resend SDK: `new Resend(apiKey)`
inline, keyed off `RESEND_API_KEY`, throwing `RESEND_API_KEY is not set` when the key is absent.
17 files across `core` (auth, notifications, messages, sales, customer_accounts), `checkout`,
`onboarding` and `enterprise` import it directly.

Deployments that cannot use Resend — self-hosted installs, EU data-residency SMTP relays, local
development against MailDev/Mailpit, CI mail sinks — therefore have no supported way to deliver
system email. A downstream app **cannot** shim this: the call sites live inside published
`@open-mercato/*` packages, so an app-level replacement would cover only the app's own calls and
leave password resets, invitations and notifications Resend-only. The transport has to live
upstream.

This PR adds **only the spec**. No code. It proposes an SMTP (nodemailer) transport alongside the
existing Resend path, with `sendEmail()` kept as the single façade and its **signature unchanged**,
so every existing call site keeps working with zero edits.

Transport resolution, in priority order:

```
explicit `transport?: 'resend' | 'smtp'` option on the call
  -> EMAIL_STRATEGY env ('resend' | 'smtp')
    -> auto-detect: RESEND_API_KEY set -> resend; else SMTP_HOST set -> smtp
      -> neither configured -> throw 'RESEND_API_KEY is not set'   (today's behavior, unchanged)
```

Backward-compatibility invariant: a deployment with only `RESEND_API_KEY` behaves byte-identically
to today, and a deployment with nothing configured fails with exactly the same error as today.
`EMAIL_STRATEGY` follows the repo's unprefixed `*_STRATEGY` selector convention
(`CACHE_STRATEGY`, `QUEUE_STRATEGY`, `RATE_LIMIT_STRATEGY`). Both provider SDKs are imported lazily
inside their transports, so a Resend-only deployment never loads nodemailer and an SMTP-only
deployment never loads the Resend SDK.

Reviewed and approved by @jtomaszewski and @patrykbojczuk on the fork before upstreaming
(fullstackhouse/open-mercato#88); the lazy-import-both-SDKs change came out of that review.

## Changes

- Adds `.ai/specs/2026-08-14-smtp-email-transport.md`

Scope the spec proposes (not in this PR): new
`packages/shared/src/lib/email/transports/{types,resend,smtp}.ts` with `send.ts` as a dispatcher;
`SMTP_HOST`, `SMTP_PORT`, `SMTP_USER`, `SMTP_PASSWORD`, `SMTP_SECURE`, `SMTP_TIMEOUT_MS` env
surface documented in `apps/mercato/.env.example` and mirrored into the create-app template;
unit tests for resolution order, config parsing, rendering/attachment mapping and error wrapping.

Explicit non-goals for v1, all recorded in the spec: per-tenant SMTP configuration UI or credential
storage (config stays operator-level env), SSRF/host-pinning hardening, connection pooling,
registering an `emailService` DI value, migrating the two Resend-direct `inbox_ops` routes,
per-user mailbox sending (that is the Communications Hub / `channel-imap` surface), and recipient
allowlist gating.

## Specification

- [ ] Yes
- [x] No (created a new spec)

**Spec file path:**
`.ai/specs/2026-08-14-smtp-email-transport.md`

## Testing

No code in this PR, so no test suites were run. The document carries the mandatory sections
(Overview, Problem Statement, Proposed Solution, Architecture, Data Models, API Contracts, Testing,
an 8-row Risks & Impact Review, Final Compliance Report, Changelog).

The spec's own Testing section specifies what ships with the implementation: unit tests covering
the six resolution branches, SMTP config parsing, React->HTML/plaintext rendering plus attachment
mapping, and error wrapping — with the existing `packages/shared/src/lib/email/__tests__/send.test.ts`
required to pass unchanged as the backward-compatibility guard. Validation commands named there:
`yarn workspace @open-mercato/shared test`, `yarn workspace @open-mercato/shared build`,
`yarn typecheck`, `yarn lint`. A manual MailDev end-to-end check is documented as optional.

## Checklist

      *(left for the human author to tick personally — this PR was prepared with agent assistance)*
      *(spec document only; the `.env.example` + create-app template mirror is specified for the implementation PR)*
- [ ] I added or adjusted tests that cover the change.
      *(no code in this PR; unit-test coverage is specified and ships with the implementation)*
      *(documented in the spec: the change adds no HTTP route and no UI surface, so there is no integration-test surface — coverage is unit-level)*
- [ ] QA routing set: `skip-qa` for low-risk non-customer-facing changes — and for changes with no manually exercisable UI surface that leave the database structure and API surface unchanged, preserve `BACKWARD_COMPATIBILITY.md` contracts, and ship automated tests for the behavior they change — otherwise `needs-qa`. A `needs-qa` PR cannot merge until QA adds `qa-approved` (or an engineer self-QAs: run locally, click through, attach a screenshot/written confirmation, then add `qa-approved` + `qa-self-verified` — without `triage` permission, post the evidence and ask a maintainer to apply those two labels). See `.github/QA-DEPLOYMENT.md` and `AGENTS.md` → PR Workflow for the authoritative rules.
      *(the three label boxes above are unticked because this account has `read` permission upstream and cannot apply labels — intended labels are posted in a separate comment for a maintainer)*

### Design System Compliance

N/A — this PR contains no code and no UI, only a spec document.

## Linked issues

None.


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-14T14:59:29Z. Other auto-skills will skip this PR until the lock is released.

pkarw (dismissed):
# 🔍 Specification review: SMTP transport for transactional email

## 🎯 Summary

This PR adds a single design document, `.ai/specs/2026-08-14-smtp-email-transport.md`, proposing an SMTP (nodemailer) transport alongside the existing Resend path in `packages/shared/src/lib/email/`. No code ships here.

The problem statement is correct and well-argued: `sendEmail()` in `packages/shared/src/lib/email/send.ts` really does construct `new Resend(apiKey)` inline off `RESEND_API_KEY`, and I confirmed 17 non-test files across `core` (auth, customer_accounts, sales, notifications, messages), `checkout`, `onboarding` and `enterprise` import it directly — so a downstream app genuinely cannot shim the transport, and the capability has to live upstream. The shape of the solution is right too: keep `sendEmail()` as the façade, resolve a transport, dispatch. Almost every claim the spec makes about this codebase checks out against the worktree — the `channel-imap` nodemailer precedent, the two Resend-direct `inbox_ops` routes, the dead `emailService` resolve hook at `packages/core/src/modules/workflows/lib/activity-executor.ts:547`, the `CACHE_STRATEGY`/`QUEUE_STRATEGY`/`RATE_LIMIT_STRATEGY` env convention, and the `nodemailer` version split (root `9.0.1` vs `channel-imap` `^9.0.3`) are all accurate as written.

Two majors stand between this and approval. The first is an internal contradiction in the normative sections: the mandated execution order in the Architecture table breaks the backward-compatibility invariant the spec promises in four other places, and an implementer following the table literally would fail the spec's own test case 6. The second is a security gap: the SMTP transport's TLS model is materially weaker than the policy this repository already applies to its existing SMTP client, and it is not declared as a non-goal. Three minors and two nits follow.

**Verdict: CHANGES REQUESTED** — two major findings, no documented waiver.

## 🧪 Validation Gate

`SPEC_ONLY=true`: the diff is one markdown file under `.ai/specs/`, so there is nothing to compile, typecheck, or test.

**Validation source: GitHub checks.** Both reported checks on head `d21671d9` are non-failing — `license/cla` passed (all committers have signed) and `[code]smith` is skipping. Branch protection is not readable from this account, so every reported check was treated as required; none is failing and none is pending. No local fallback was needed or run.

| Command | Result |
|---------|--------|
| `yarn build:packages` | Skipped — no source files in the diff. |
| `yarn generate` | Skipped — no module files in the diff. |
| `yarn i18n:check-sync`, `yarn i18n:check-usage` | Skipped — no locale or user-facing strings in the diff. |
| `yarn typecheck` | Skipped — no TypeScript in the diff. |
| `yarn test` | Skipped — no code under test. |
| `yarn build:app` | Skipped — no app code in the diff. |

Structural checks that do apply, all passing: the filename follows `{YYYY-MM-DD}-{kebab-case-title}.md` with no `SPEC-*` prefix, it sits in the OSS `.ai/specs/` root (correct — nothing here is enterprise scope), and it carries every section `.ai/specs/AGENTS.md` requires (TLDR, Overview, Problem Statement, Proposed Solution, Architecture, Data Models, API Contracts, Testing, an 8-row Risks & Impact Review, Final Compliance Report, Changelog).

## 💥 What can go wrong

### 🔴 Major — the SMTP transport has no TLS-downgrade policy, and the spec does not name that as a non-goal

`.ai/specs/2026-08-14-smtp-email-transport.md:107` gives the transport a two-state TLS model: `SMTP_SECURE` true means implicit TLS, false means "STARTTLS on 587 is nodemailer's default upgrade behavior when `secure` is false". That default is *opportunistic*: when the server does not advertise STARTTLS, nodemailer sends the message in cleartext rather than failing. Nothing in the spec sets `requireTLS`, and nothing gates the cleartext outcome behind an operator opt-in.

This pipeline carries password-reset links, user invitations and portal invitation tokens — the exact payloads for which a silent cleartext fallback matters. And this repository has already taken an explicit position on precisely this question: `packages/channel-imap/src/modules/channel_imap/lib/transport.ts` centralizes a cleartext-transport policy whose own comment reads "sends the password in the clear over an attacker-controlled host string", refuses `'none'` on every connection build, and requires `OM_CHANNEL_IMAP_ALLOW_INSECURE_TRANSPORT=true` to permit it. Its SMTP client at `smtp-client.ts:172` correspondingly sets `requireTLS: options.transport === 'starttls'` and `tls: { rejectUnauthorized: true }` to reject downgrade attacks.

The spec cites that same `NodemailerClient` as the precedent for its transporter lifecycle (`:69`), so it is clearly aware of the file — but it adopts the lifecycle and not the transport-security policy, and the Risks table does not raise it. The SSRF non-goal at `:21` explicitly scopes out host-pinning only; TLS downgrade is a separate property and is never mentioned. The spec's own documented manual test (`:157`, `SMTP_HOST=localhost SMTP_PORT=1025 SMTP_SECURE=false`) is a cleartext send with no opt-in gate.

Please resolve this in the spec one of two ways. Either mirror the existing policy — make `SMTP_SECURE=false` mean STARTTLS-required (`requireTLS: true`, `rejectUnauthorized: true`), and add a third explicitly-cleartext state behind an `OM_ALLOW_INSECURE_SMTP`-style opt-in that the MailDev workflow uses — or state plainly, with reasoning, that opportunistic STARTTLS is accepted for v1 and record it as a Risks-table row with its residual risk. Silence is the problem; either answer is reviewable.

### 🟡 Minor — Risk 8's mitigation rests on a premise that is false in this repository

Risk 8 (`:170`) covers the lazy `import('resend')` shifting a missing-package failure from process start to first send, and mitigates it with "The package remains a regular declared dependency (install-time guarantee unchanged)."

`resend` is not a declared dependency of `@open-mercato/shared`. I checked `packages/shared/package.json`: its `dependencies` are `@mikro-orm/*`, `@open-mercato/cache`, `@types/sanitize-html`, `dotenv`, `pino`, `rate-limiter-flexible`, `re2js`, `reflect-metadata`, `sanitize-html` and `undici`. `resend` is declared at the root and in `apps/mercato` and `packages/core`, but `packages/shared/src/lib/email/send.ts` imports it through hoisting alone, backed by the `declare module 'resend'` shim at `packages/shared/src/types/resend.d.ts`. (`react` and `zod` are in the same position, so this is a pre-existing repo pattern rather than something this design introduces.)

The consequence is that the stated install-time guarantee does not exist, so the mitigation does not mitigate — and the Architecture table at `:91` proposes adding `nodemailer` and `@react-email/render` to `packages/shared/package.json` while leaving `resend` undeclared, which would make shared lazily import two SDKs, declare one of them, and rely on hoisting for the other. Either correct the mitigation text to describe the real situation, or take the opportunity to add `resend` to shared's dependencies as part of this work and keep the mitigation as written.

## 🔁 Backward compatibility

### 🔴 Major — the mandated execution order contradicts the backward-compatibility invariant the spec promises

The spec states the invariant four times: "a deployment with nothing set fails with the same error as today" (`:58`), the resolution chain's "neither configured → throw `'RESEND_API_KEY is not set'` (existing behavior, unchanged)" (`:55`), the error contract's "`RESEND_API_KEY is not set` — unchanged" (`:141`), and test case 6, "Nothing configured → still throws `RESEND_API_KEY is not set`" (`:155`).

The Architecture table then mandates an order that breaks it. `:87` specifies that `send.ts` "keeps the disable-check and from-address throw, **then** dispatches on `resolveEmailTransportName(options.transport)`", and `:133`'s `ResolvedEmailMessage = Omit<SendEmailOptions, 'from' | 'transport'> & { from: string }` confirms the from-address is resolved before any transport is selected. Today's code runs the checks in the opposite order — `packages/shared/src/lib/email/send.ts:104` throws `RESEND_API_KEY is not set` first, and only reaches the `EMAIL_FROM_NOT_CONFIGURED` throw at `:109` once an API key is present.

So on a truly empty environment — no `RESEND_API_KEY`, no `SMTP_HOST`, and no `NOTIFICATIONS_EMAIL_FROM`/`EMAIL_FROM`/`ADMIN_EMAIL` — today's `sendEmail()` throws `RESEND_API_KEY is not set` and the proposed one throws `EMAIL_FROM_NOT_CONFIGURED: set NOTIFICATIONS_EMAIL_FROM, EMAIL_FROM, or ADMIN_EMAIL`. That is the "nothing configured" case the invariant names, and the error changes.

Worth flagging explicitly: the existing suite will not catch this. `packages/shared/src/lib/email/__tests__/send.test.ts` sets `RESEND_API_KEY: 'test-key'` in `beforeEach`, and its "throws a clear error when no sender address is configured" case only deletes the from-address vars — so it passes under either ordering, and the spec's "existing `send.test.ts` passes unchanged" guard (`:148`) gives no protection here. The spec's own test 6 is the case that would fail, and only if it clears the from-address vars too.

Please make the ordering explicit and consistent. If the invariant is what you want, say that transport resolution and its `RESEND_API_KEY is not set` throw run *before* the from-address resolution, and fix `:87` and the `ResolvedEmailMessage` shape to match. If the from-first ordering is deliberate — it is arguably the better error for an operator — then say so, drop the "same error as today" claim, record the changed error as an accepted behavior delta in the Risks table, and pin it with a test that sets no env at all.

### ✅ Otherwise sound

The `transport?: 'resend' | 'smtp'` addition to `SendEmailOptions` is a genuinely additive optional field on an exported type, which `BACKWARD_COMPATIBILITY.md` §3 permits without the deprecation protocol ("New optional parameters may be added"). No DB schema, HTTP route, event ID, widget spot, DI key, ACL feature or CLI surface is touched, and the Impact summary at `:172` says so accurately. Resend winning auto-detection whenever `RESEND_API_KEY` is present is the right default and correctly protects existing deployments (Risk 2).

## 🧩 What's missing

### 🟡 Minor — no implementation breakdown, so the spec is not directly executable

The document describes the target state well but never decomposes it into ordered phases and steps. Per the root `AGENTS.md` Task Router, specs feed `om-implement-spec` and `om-auto-create-pr`, and `om-spec-writing` produces "an implementation breakdown into phases and steps" for exactly that reason. Everything needed is present in prose — extract the transports, add the resolver and config parsing, add the SMTP transport, wire the deps, mirror `.env.example` into the create-app template, add the AGENTS.md row, add tests — but an implementing agent has to re-derive the ordering and the commit boundaries. Adding a short Implementation Plan section with those steps in dependency order, each with its acceptance check, would make this spec directly executable.

### 🟢 Nit — `@react-email/render` is not in the dependency graph the way Risk 5 describes

Risk 5 (`:167`) mitigates the new dependency surface with "both libraries already exist in the monorepo dependency graph (channel-imap, react-email tooling)". That holds for `nodemailer` (root `9.0.1`, `channel-imap` `^9.0.3`). `@react-email/render` is declared by no package in this repo — the root has `@react-email/components@^1.0.12` and `react-email@^6.9.1`, and `@react-email/render` appears only as a transitive entry in `yarn.lock`. The Architecture table already treats it correctly as a new dependency at `:91`, so this is just the risk row overstating the existing footprint; a one-word softening ("nodemailer already exists; `@react-email/render` arrives transitively today and becomes a declared dependency here") would make it accurate.

## 📈 How can this specification be improved

### 🟢 Nit — the call-site count is inconsistent across the PR

The spec says "Roughly 14 call sites" (`:32`) and "~14 existing call sites" (`:7`), while the PR description says 17 files. The actual count of non-test files importing `@open-mercato/shared/lib/email/send` is **17**: two in `sales` (quotes accept/send), two in `customer_accounts` (invitation email, signup), three in `auth` (users command, reset, resend-invite), one each in `notifications` and `messages`, three in `onboarding` (ready-email, onboarding post, demo-feedback), one in `checkout`, and four in `enterprise/security`. Since the whole backward-compatibility argument turns on "every existing call site keeps working with zero edits", it is worth stating the exact number and letting the two documents agree.

The overall shape is right and appropriately small — two transports behind a resolver, no DI service, no per-tenant config, no pooling. The non-goals section is unusually disciplined and I would not cut further. Deferring SSRF hardening is defensible given `SMTP_HOST` is operator env, and the reasoning is stated rather than assumed.

### 🟡 Minor — the per-call `transport` option has no consumer in v1

`transport?: 'resend' | 'smtp'` sits at the top of the resolution chain (`:8`, `:124`), but no call site in the proposed scope passes it, and the spec's own non-goals rule out the one place that might want it (the `inbox_ops` routes stay Resend-direct, `:24`). It adds a field to a published type, a branch to the resolver, and a test case, to serve nothing yet.

Concretely simpler alternative: ship v1 with env-only resolution — `EMAIL_STRATEGY`, then auto-detect. The resolver signature stays `resolveEmailTransportName(explicit?: EmailStrategyName)` so the plumbing is already there, and adding the public option later is itself additive, so deferring costs no backward compatibility and no rework. If a concrete caller is already known to need per-call selection, naming it in the spec would settle this immediately in the other direction.

## 🧪 Test coverage

The Testing section (`:146`–`:157`) is strong for a spec: six enumerated resolution-order cases, transporter option construction including the `auth`-omitted and `SMTP_SECURE`-default branches, rendering and attachment mapping, failure wrapping, `close()` on both paths, and the disabled-delivery short-circuit. The named validation commands are appropriate for the scope.

Two gaps, both covered above rather than repeated here: test case 6 does not specify its from-address env state, which is exactly the ambiguity the backward-compatibility major turns on; and there is no test named for the TLS/STARTTLS behavior that the security major asks the spec to pin down. Once those two findings are resolved, please add the corresponding cases so the invariants are enforced rather than asserted.

The claim that there is no integration-test surface is correct — this change adds no HTTP route and no UI path, so unit-level coverage is the right level.

---

**Autofix:** skipped (not my PR — re-run with `--autofix` to fix it here).

*Reviewed at head `d21671d9` by `om-auto-review-pr`. Spec claims were verified against the working tree, not taken at face value.*

pkarw:

- ❌ `changes-requested` — the specification review found two major issues: the Architecture table's execution order contradicts the backward-compatibility invariant the spec promises in four other places, and the SMTP transport has no TLS-downgrade policy despite this repo already enforcing one for its existing SMTP client.
- 📚 `documentation` — the diff is a single design document under `.ai/specs/` with no code, config, or CI files.
- 🔹 `priority-medium` — this unblocks self-hosted, EU-residency and local-development deployments that have no supported system-email path today, which is ordinary feature work rather than an outage or a security incident.
- 🟢 `risk-low` — a spec-only change has no runtime blast radius; the risk it describes will be re-rated on the implementation PR, which touches a shared contract surface and will likely warrant `risk-medium` or higher.
- ⏭️ `skip-qa` — there is nothing manually exercisable in a markdown document: no UI surface, no database structure change, and no API surface change.

`partner-request` was left in place as an external classification this skill does not own.

pkarw:
Thanks @bartek5412 — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pkarw:

Specification review of the spec-only diff at head `d21671d9`: 2 major, 3 minor, 2 nit findings. The majors are a backward-compatibility contradiction between the Architecture table's mandated execution order and the invariant the spec promises elsewhere, and a missing TLS-downgrade policy for the SMTP transport. Validation source: GitHub checks (`license/cla` pass, `[code]smith` skipping, none failing or pending) — the local gate was skipped as inapplicable to a markdown-only diff. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

bartek5412:
Updates pushed at `8e00965c`; re-review requested via the reviewers sidebar.

All 7 findings addressed in the spec:

- **Major, execution order:** resolved in favor of the invariant. A normative four-step list (disable-check → transport resolution → from-address → dispatch) now governs the façade; the Architecture row and `ResolvedEmailMessage` agree with it, and new test case 7 pins the nothing-configured error with `RESEND_API_KEY`, `SMTP_HOST`, `EMAIL_STRATEGY` **and** all three from-address vars cleared. The Testing section also records why the existing `send.test.ts` cannot catch this (it sets the key in `beforeEach`).
- **Major, TLS:** adopted the `channel-imap` policy per your first option. `SMTP_SECURE=false` now means STARTTLS-**required** (`requireTLS: true`, `rejectUnauthorized: true`, never opportunistic); cleartext exists only behind `OM_ALLOW_INSECURE_SMTP=true`, which the MailDev workflow now uses explicitly. New risks 5–6 and test case 3 cover it.
- **Minor, risk 8's false premise:** took the stronger of your two options — Phase 1 declares `resend` in `packages/shared/package.json` (pinned to the existing `^6.18.1`), so the install-time guarantee the mitigation relies on actually exists.
- **Minor, no implementation breakdown:** added an Implementation Plan — five dependency-ordered phases, each with an acceptance check; Phase 2 is a behavior-preserving extraction gated by the unchanged existing suite.
- **Minor, per-call `transport` has no consumer:** dropped from v1 and moved to Non-goals; the resolver keeps the optional `explicit` parameter you pointed at, so exposing the option later is additive. Side effect: **no exported type changes at all** now.
- **Nit, `@react-email/render` graph claim:** risk row now says it arrives transitively today and becomes a declared dependency here.
- **Nit, call-site count:** corrected to 17 files / 26 call sites, with the per-module breakdown in the Overview, matching the PR description.

Changelog entry added (2026-08-17) referencing this review.

patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-18T14:50:10Z. Other auto-skills will skip this PR until the lock is released.

patzick (approved):
🤖 `om-auto-review-pr` — **APPROVED**

Re-review of the spec-only diff at head `f2b5f45b` (`.ai/specs/2026-08-14-smtp-email-transport.md`, the single changed file). All 7 findings from the previous pass are genuinely resolved; 4 nits remain, none blocking.

**Validation source: GitHub checks.** No command was run locally. The `audit` failure that was showing on this PR was **not** caused by the change: run `32016028230` was a re-run of the 2026-08-17 merge commit, so it still carried `deepmerge-ts <8.0.0` (GHSA-ggr8-5vv4-36mx) even though `develop` had pinned `deepmerge-ts@8.0.1` in `b708b7bb` earlier the same morning. I updated the branch from `develop` (`f2b5f45b`) and approved the fork workflow run; the fresh run is green — `test`, `lint`, `ds-lint`, `prepare`, `scope`, `audit-scope`, `license/cla` all SUCCESS, with `audit` correctly scoped out for a docs-only diff.

---

### Verification of the previous majors

**1. Execution-order vs the BC invariant — resolved.** The normative four-step list (disable-check + capture → transport resolution → from-address → dispatch) now agrees with the Architecture row, the `ResolvedEmailMessage` shape and the error contract. It matches today's code exactly: `packages/shared/src/lib/email/send.ts` throws `RESEND_API_KEY is not set` at `:105`, before the `EMAIL_FROM_NOT_CONFIGURED` throw at `:109`. Test case 7 pins it with `RESEND_API_KEY`, `SMTP_HOST`, `EMAIL_STRATEGY`, `NOTIFICATIONS_EMAIL_FROM`, `EMAIL_FROM` and `ADMIN_EMAIL` all cleared — and the spec is right that the existing `send.test.ts` cannot catch this, since it sets `RESEND_API_KEY` in `beforeEach`.

**2. TLS-downgrade policy — resolved.** The table adopts the stance this repo already enforces: `packages/channel-imap/src/modules/channel_imap/lib/smtp-client.ts:169-183` (`requireTLS` on STARTTLS, `rejectUnauthorized: true`, `tls: undefined` only for `transport: 'none'`) and `transport.ts:22` (`OM_CHANNEL_IMAP_ALLOW_INSECURE_TRANSPORT`). `SMTP_SECURE=false` meaning STARTTLS-**required** rather than opportunistic is the correct reading of nodemailer's default, and gating cleartext behind `OM_ALLOW_INSECURE_SMTP` matches the `OM_`-prefixed safety-override convention. Risks 5-6 and test case 3 cover it.

The other five (risk 8's false premise, missing implementation breakdown, the consumer-less per-call `transport` option, the `@react-email/render` graph claim, the call-site count) are all addressed. Dropping the per-call option is the right call — it leaves **no exported type changed at all**, which materially shrinks the BC surface.

### Claims verified against `origin/develop`

| Claim | Result |
|---|---|
| `send.ts` hardcodes Resend; throws before the from-address check | ✅ `:105` then `:109` |
| 17 non-test files import `lib/email/send` | ✅ exactly 17 |
| `resend` undeclared in `packages/shared/package.json`, resolved by hoisting + the `declare module 'resend'` shim | ✅ confirmed — so Phase 1's declaration is a real fix, not a formality |
| `channel-imap` TLS precedent | ✅ `smtp-client.ts:169-183`, `transport.ts:22` |
| Dead `emailService` DI hook | ✅ `activity-executor.ts:546` |
| nodemailer root `9.0.1` vs channel-imap `^9.0.3`; `resend ^6.18.1` | ✅ |
| `@react-email/render` only transitive today | ✅ declared in no `package.json` |
| `*_STRATEGY` naming convention | ✅ `CACHE_STRATEGY`, `QUEUE_STRATEGY`, `RATE_LIMIT_STRATEGY` |
| `.env.example` documents only `RESEND_API_KEY` + the from-chain | ✅ |

### Findings — 4 nits, 0 blocker/major/minor

**nit 1 — the "26 call sites" figure is wrong, and its own breakdown contradicts it.** Across the 17 non-test importers there are **22** `sendEmail(` invocations, not 26. Separately, the enumeration that follows the number — "three in `auth`, two in `sales`, two in `customer_accounts`, one each in `notifications` and `messages`, three in `onboarding`, one in `checkout`, four in `enterprise/security`" — sums to 17, i.e. it enumerates *files*, not call sites. Suggest "17 non-test files, 22 call sites" and labelling the breakdown as per-module files. (Appears twice: TLDR bullet 2 and Overview ¶1.)

**nit 2 — line reference off by one.** "throws `RESEND_API_KEY is not set` at `:104`" — `:104` is the `const apiKey` read; the throw is `:105`.

**nit 3 — `SMTP_TIMEOUT_MS` has no default.** Left unset it inherits nodemailer's 2-minute `connectionTimeout`, which is a long stall for a request-path send (password reset, invitation). `channel-imap` pins `options.timeoutMs ?? 10_000`; worth adopting the same default so the two SMTP paths behave alike.

**nit 4 — `react` stays undeclared while `@react-email/render` is added.** Phase 1 closes exactly this hoisting gap for `resend`, but `packages/shared/package.json` declares no `react` either, and `send.ts` imports it today while `@react-email/render` peer-depends on it. Either declare it (dependency or peer) in the same pass or note explicitly that it is accepted — otherwise risk 10's reasoning applies to `react` and is left unanswered.

### Compliance

- **Backward compatibility** — no contract surface removed or renamed; `sendEmail`'s signature and `SendEmailOptions` untouched; no deprecation protocol needed. The nothing-configured and Resend-only paths are pinned by tests.
- **Spec conventions** (`.ai/specs/AGENTS.md`) — `{YYYY-MM-DD}-{kebab-case}.md` in the OSS directory, no `SPEC-*` prefix, all mandatory sections present, changelog updated with the 2026-08-17 entry referencing this review.
- **Integration coverage** — correctly documented as not applicable (no HTTP route, no UI surface); unit coverage is specified and ships with the implementation.
- **Template Sync Checklist** — the `apps/mercato/.env.example` → `packages/create-app/template/.env.example` mirror is in scope as Phase 5, as required.

One thing to carry into the implementation PR, not a finding here: where `RESEND_API_KEY is not set` is thrown when `EMAIL_STRATEGY=resend` is set explicitly but the key is absent is slightly under-specified — step 2 (resolver) or step 4 (transport). It has no BC exposure since `EMAIL_STRATEGY` is new, but pinning it in a test saves an implementer coin-flip.

Nice work on the second pass — the ordering fix and the TLS policy are both exactly right, and the spec is unusually well-grounded in the code it cites.

patzick:

- ✅ `merge-queue` (was `changes-requested`) — the re-review of head `f2b5f45b` found only 4 nits; both majors from the 2026-08-14 pass are verifiably resolved and CI is green.
- 📚 `documentation` — unchanged; the diff is still a single design document under `.ai/specs/` with no code, config or CI files.
- 🔹 `priority-medium` — unchanged.
- 🟢 `risk-low` — unchanged; a spec-only change has no runtime blast radius. The implementation PR touches a shared contract surface and should be re-rated there.
- ⏭️ `skip-qa` — unchanged; nothing manually exercisable in a markdown document, no database structure or API surface change.

`partner-request` left in place as an external classification this skill does not own.

⚠️ **One thing a maintainer needs to clear:** `reviewDecision` is still `CHANGES_REQUESTED` because the 2026-08-14 review by @pkarw was submitted against commit `d21671d9` and has not been re-reviewed or dismissed. Both of its majors are addressed in `8e00965c` and independently verified in my review above, but GitHub will keep gating the merge until that account re-reviews or the stale review is dismissed. I did not dismiss it myself.

patzick:

Re-review of the spec-only diff at head `f2b5f45b`: 0 blocker, 0 major, 0 minor, 4 nit findings. Both majors from the 2026-08-14 pass (execution-order vs the backward-compatibility invariant, and the missing TLS-downgrade policy) are verifiably resolved, and every code reference the spec makes was checked against `origin/develop`.

Validation source: **GitHub checks** — no command run locally. The `audit` failure on this PR was a stale merge base, not a defect in the change: run `32016028230` was a re-run of the 2026-08-17 merge commit and still carried `deepmerge-ts <8.0.0` (GHSA-ggr8-5vv4-36mx), which `develop` had already pinned to 8.0.1 in `b708b7bb`. I updated the branch from `develop` and approved the fork workflow run; the fresh run is green (`test`, `lint`, `ds-lint`, `prepare`, `scope`, `audit-scope`, `license/cla` SUCCESS; `audit` correctly scoped out for a docs-only diff).

Autofix: skipped — the 4 nits are all in the fork author's document and none are merge-blocking.

PR #5305, opened 2026-08-14
fix(auth): return null from loadSidebarPreference when no row exists

Frshy:
## Summary

`loadSidebarPreference(em, scope)` returns `normalizeSidebarSettings(existing?.settingsJson)`, so a user with **no** `UserSidebarPreference` row receives a fully-populated *default* settings object instead of `null`. That makes "no saved preference" indistinguishable from "a preference that happens to be empty".

`backendChrome` builds the sidebar as a two-layer merge — role preference → `adoptSidebarDefaults` → user preference — and guards the second pass with `userPreference ? applySidebarPreference(...) : baseForUser`. Because the loader never returned a falsy value, that else-branch was dead code. And since `applySidebarPreference` **overwrites** `hidden` (`next.hidden = hidden`) rather than OR-ing it, and falls back to weight/name ordering when `groupOrder` is empty, the empty user pass erased the entire role layer on every render: role-hidden items reappeared and role group ordering was lost — for every user who had never personally customised their sidebar.

The fix adds **`findSidebarPreference`**, which returns `null` when no row exists and makes the existing guard work as written. `loadSidebarPreference` is kept verbatim as a deprecated bridge, so no third-party caller breaks.

## Changes

- `packages/core/src/modules/auth/services/sidebarPreferencesService.ts`
  - **new** `findSidebarPreference(em, scope): Promise<SidebarPreferencesSettings | null>` — returns `null` for an absent row.
  - `loadSidebarPreference` keeps its exact signature (`Promise<SidebarPreferencesSettings>`) and behaviour, now `@deprecated` with migration guidance and a 0.9.0 removal target. It delegates: `(await findSidebarPreference(em, scope)) ?? normalizeSidebarSettings(null)`.
- `packages/core/src/modules/auth/lib/backendChrome.tsx`, `packages/core/src/modules/auth/api/sidebar/preferences/route.ts` — the two internal call sites now use `findSidebarPreference`.
- `packages/core/src/modules/auth/services/__tests__/sidebarPreferencesService.load-preference.test.ts` (new) — absent → `null`, present → normalized settings, present-but-empty → non-null empty, and the (user, tenant, organization) scope stays on the lookup. Plus two cases pinning the deprecated bridge's non-null default, so the compatibility promise itself is under test.
- `packages/core/src/modules/auth/lib/__tests__/backendChrome.role-preference.test.ts` (new) — keeps the **real** `applySidebarPreference` *and* the **real** `findSidebarPreference`, driving the latter by stubbing the `UserSidebarPreference` row it reads. A role hide and role group order survive an absent row; a saved-but-empty row still overrides the role layer.
- Test-only: five pre-existing auth suites that stub the service module were updated to the new function name, since the internal call sites moved.
- `UPGRADE_NOTES.md` — documents the deprecation under a new `0.7.0 → 0.7.1 (unreleased)` heading.
- `.ai/specs/2026-04-27-ds-sidebar-customization-page.md` — changelog entry, and BC table row 3 updated (it previously asserted "No exported function signatures change").

### Callers

Both existing call sites were already null-safe, so the switch needed no other changes:

- `backendChrome.tsx` guards with `userPreference ? … : baseForUser`.
- `api/sidebar/preferences/route.ts` reads every field as `settings?.x ?? default`, so the JSON response shape is unchanged.

### Backward compatibility

The first revision of this PR narrowed `loadSidebarPreference`'s return type in place, which broke a contract surface under `BACKWARD_COMPATIBILITY.md` with no bridge. That is fixed here. The deprecation protocol is now satisfied in full:

- ✅ **Nothing removed or renamed in one release** — `loadSidebarPreference` keeps its name, import path, signature, and behaviour.
- ✅ **`@deprecated` JSDoc** with migration guidance and a stated removal target (0.9.0).
- ✅ **Bridge** — the old function delegates to the new one and re-fabricates the old default, so a third-party caller doing `settings.hiddenItems` without a null check keeps working.
- ✅ **Documented in `UPGRADE_NOTES.md`**, under a fresh unreleased heading.
- ✅ **Spec "Migration & Backward Compatibility" section** — BC table row 3 corrected to describe the addition and the deprecation.

### Known gap (out of scope)

Once a user saves *any* personal preference, role hides are wiped again — `applySidebarPreference` overwrites rather than merges, and `SidebarCustomizationEditor` seeds its draft solely from the user's own saved `hiddenItems`, never from the role-applied `hidden` flags in the base snapshot. Whether role hides are policy (un-overridable) or merely a default is a product decision that needs its own spec entry, so it is documented in the spec changelog rather than fixed here.

## Specification

- [x] Yes
- [ ] No (created a new spec)

**Spec file path:** `.ai/specs/2026-04-27-ds-sidebar-customization-page.md` (changelog entry + BC table row 3)

## Testing

Rebased onto current `develop` (was 226 commits behind) before validating.

- `npx jest --runInBand src/modules/auth` in `packages/core` — **75 suites, 604 tests, all passing**.
- `npx tsc --noEmit` in `packages/core` — **zero errors in every changed file**. (The run reports pre-existing errors elsewhere in the package from an unbuilt worktree — unresolved `@tanstack/react-table/legacy` and `#generated/entities.ids.generated` — none of them in files this PR touches.)
- `npx eslint` on the three changed production files — clean.

The regression coverage was verified by reverting the loader to its pre-fix form: **three** tests then fail — one in the loader suite and **two in the chrome suite**, which now exercises the real loader end-to-end rather than a stub of it.

## Checklist

- [x] I updated documentation, locales, or generators if the change requires it. (`UPGRADE_NOTES.md` + spec; no locale or generator surface touched.)
- [ ] I added or updated integration tests in `.ai/qa/tests/` — the change is a server-side loader with no new UI surface; the regression is covered by `backendChrome.role-preference.test.ts`, which runs the real loader and merge and asserts on the resolved chrome payload the sidebar renders from. Happy to add a `.ai/qa/tests/` case if you would rather have one.
- [ ] Priority set — this account has read-only permission upstream and cannot apply labels.
- [ ] Risk set — same as above.
- [ ] QA routing set — same as above; `needs-qa` is already applied.

### Design System Compliance

Not applicable — no UI files are touched. The change is confined to a service function, its call sites, tests, and documentation.

- [ ] No hardcoded status colors — N/A
- [ ] No arbitrary text sizes — N/A
- [ ] Empty state handled for list/data pages — N/A
- [ ] Loading state handled for async pages — N/A
- [ ] `aria-label` on all icon-only buttons — N/A
- [ ] Uses existing DS components — N/A

## Linked issues

None.


Frshy:
This PR is opened from a fork by an account with read-only permission here, so it cannot set its own labels, assignee, or reviewer. Requesting the following, per `AGENTS.md` → PR Workflow and `.ai/agentic.config.json`:

**Pipeline** — `review`: ready for code review, not a draft, not in any other pipeline state.

**Category** — `bug`: fixes a regression where role-level sidebar layout (hidden items and group order) was silently erased on every render for any user without a personally saved sidebar preference.

**Priority** — `priority-medium`: user-visible and hits every user who has not customised their sidebar, so role-level hides are not being honoured at all; but it is layout-only, with no data loss, no security exposure, and a clear scope. Not `priority-high` — it does not block a release.

**Risk** — `risk-medium`: single module (`core/auth`), two files of production change (one of them one line), and both existing call sites were already null-safe. Raised above `risk-low` because it narrows an exported function's return type — a `BACKWARD_COMPATIBILITY.md` contract surface — and ships intentionally without a bridge, since the old return value is the bug. Documented in `UPGRADE_NOTES.md` under `0.6.7 → 0.6.8`.

**QA routing** — `needs-qa`: the automated-verification exemption does not apply, because the change breaks a `BACKWARD_COMPATIBILITY.md` contract, and the fix is observable in the rendered sidebar. Suggested manual check: give a role a sidebar preference that hides an item and reorders groups, then sign in as a user in that role who has never opened the sidebar customization editor — the role's hides and ordering should now survive, where previously they were wiped.

**Assignee** — the maintainer who picks this up; this account cannot self-assign here.

**Reviewer** — a `core/auth` maintainer, requested by role rather than by handle.

Two checklist boxes in the description are deliberately left unticked and are flagged here rather than silently skipped:

- **CLA** — not ticked. Accepting the Contributor License Agreement is the contributor's own act; @karolszoja please confirm.
- **Integration tests in `.ai/qa/tests/`** — not added. The change is a server-side loader return value with no new UI surface; the regression is covered by `backendChrome.role-preference.test.ts`, which keeps the real `applySidebarPreference` and asserts on the resolved chrome payload the sidebar renders from. Happy to add a `.ai/qa/tests/` case if you would rather have it.

No claim label (`in-progress`) is requested — this account could not release it later.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-14T14:59:55Z. Other auto-skills will skip this PR until the lock is released.

pkarw (dismissed):
# 🔍 Code Review: fix(auth): return null from `loadSidebarPreference` when no row exists

## 🎯 Summary

This PR fixes a real and well-diagnosed regression. `loadSidebarPreference` returned `normalizeSidebarSettings(existing?.settingsJson)`, so a user with no `UserSidebarPreference` row received a fabricated, fully-populated default settings object. Because `resolveBackendChromePayload` guards its second merge pass with `userPreference ? applySidebarPreference(...) : baseForUser` (`packages/core/src/modules/auth/lib/backendChrome.tsx:419-422`), that else-branch was unreachable, and since `applySidebarPreference` assigns `next.hidden = hidden` outright (`sidebarPreferencesService.ts:242`) and re-sorts groups by weight when `groupOrder` is empty (`sidebarPreferencesService.ts:254-263`), the empty user pass erased every role-level hide and the role group order on each render. I verified the mechanism end-to-end in the code rather than taking the description on trust: `adoptSidebarDefaults` only adopts *labels* (`backendChrome.tsx:229-242`), so labels survived the empty pass while `hidden` and group ordering did not. The diagnosis in the PR body is accurate.

The fix itself is minimal and correct, both call sites are genuinely null-safe (`backendChrome.tsx:419-422` guards; `api/sidebar/preferences/route.ts:230-236` reads every field as `settings?.x ?? default`, so the JSON response shape is unchanged), and the reasoning is documented in the spec changelog and `UPGRADE_NOTES.md`. The tests are thoughtful and the comments explain *why*, which is exactly the house style.

What blocks it is not the fix but its delivery: it narrows an exported function's contract for third-party consumers with no deprecation bridge, which `BACKWARD_COMPATIBILITY.md` and `.ai/review-checklist.md` §21 treat as a Critical violation, and a cheap bridge that keeps the fix intact does exist. A second, smaller issue is that the `UPGRADE_NOTES.md` entry lands in an already-released section once this merges into today's `develop`.

Reviewed scope: the five changed files (the loader, two new Jest suites, `UPGRADE_NOTES.md`, and the sidebar-customization spec), plus the two call sites, `applySidebarPreference` / `adoptSidebarDefaults` / `normalizeGroupWeights`, and every consumer of the serialized `hidden` flag in `packages/ui` and `packages/core`.

## Verdict

❌ **Request changes** — one blocker: `loadSidebarPreference` is exported through a public package subpath and its return type is narrowed for callers (`SidebarPreferencesSettings` → `SidebarPreferencesSettings | null`) with no `@deprecated` annotation and no compatibility bridge, which the repository's deprecation protocol requires. One minor finding (the `UPGRADE_NOTES.md` entry lands under a released version after the merge) accompanies it. The underlying fix is right and should be kept as-is behind the bridge.

## 🧪 Validation Gate

**Validation source:** local fallback. Per this repository's `om-auto-review-pr` extension the review is GitHub-checks-first, but GitHub reports no gate checks for this head SHA — only `license/cla` (SUCCESS) and `[code]smith` (SKIPPED), and `develop` has no required-status-check protection configured (the API returns 404). With no usable CI signal for `validation.commands`, I ran the narrow fallback the extension prescribes: the unit tests and typecheck for the changed package. **Runner:** local — no compose `app` container is running (`docker compose ps` shows only `verdaccio`), so Docker mode does not apply.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ⏭️ NOT RUN | Not required to evaluate a four-line service change; no build-affecting surface (no new exports, no generated files) is touched. |
| `yarn generate` | ⏭️ NOT RUN | The diff adds no module file that auto-discovery consumes, so nothing regenerates. |
| `yarn build:packages` (rebuild) | ⏭️ NOT RUN | Same reason as the first invocation. |
| `yarn i18n:check-sync` | ⏭️ NOT RUN | No locale file and no user-facing string is touched. |
| `yarn i18n:check-usage` | ⏭️ NOT RUN | Same reason. |
| `yarn typecheck` | ✅ PASS (scoped) | `npx tsc --noEmit` in `packages/core` — the package that owns the changed function and both of its call sites. It reports 20 errors, all `TS2307: Cannot find module '#generated/entities.ids.generated'` in `src/modules/wms/**`, an artifact of a fresh worktree where `yarn generate` has not run. **Zero** errors in `auth`, `sidebarPreferencesService.ts`, or `backendChrome.tsx`, which is the signal that matters here: a widened return type would surface precisely as a null-unsafety error at a call site, and none appeared. |
| `yarn test` | ✅ PASS (scoped) | `npx jest --runInBand src/modules/auth` in `packages/core`: **71 suites, 542 tests, all passing**, including the two new suites and every pre-existing sidebar/nav suite (`sidebarPreferencesService.scope`, `backendChrome.group-order`, `backendChrome.current-organization`, `sidebar-preferences`, `admin-nav`, `optimistic-lock`, `find-role-tenant-scope`). I confirmed by grep that no module outside `auth` imports `loadSidebarPreference`, so the blast radius of the unit run covers every consumer in the repository. |
| `yarn build:app` | ⏭️ NOT RUN | No app-layer file is touched and the API response shape is unchanged. |

I additionally simulated the merge against current `develop` with `git merge-tree --write-tree` (clean, no conflicts — consistent with GitHub reporting `MERGEABLE`) and confirmed that `develop`'s later change to `backendChrome.tsx` (`f04fff343`, server-side nav item sorting) leaves the `userPreference ? … : baseForUser` guard untouched, so the fix's premise still holds after a rebase.

## Findings

### ⛔ Blocker

**1. `packages/core/src/modules/auth/services/sidebarPreferencesService.ts:43` — an exported function's return type is narrowed for consumers without a deprecation bridge.**

`loadSidebarPreference` goes from `Promise<SidebarPreferencesSettings>` to `Promise<SidebarPreferencesSettings | null>`. This is not an internal helper: `packages/core/package.json` declares deep wildcard subpath exports (`./*`, `./*/*`, … through nine levels), so `@open-mercato/core/modules/auth/services/sidebarPreferencesService` is a public import path — the PR's own new test imports it that way, and `UPGRADE_NOTES.md` names that exact specifier.

The concrete failure: a third-party module that today writes `const settings = await loadSidebarPreference(em, scope); if (settings.hiddenItems.includes(itemId)) { … }` compiles and runs correctly. After upgrading to a release containing this PR, it throws `TypeError: Cannot read properties of null (reading 'hiddenItems')` — and it throws precisely for the users this fix targets, those with no saved sidebar row, which in most installs is the majority. A TypeScript consumer gets a compile error under `strictNullChecks` instead, which is the friendlier half of the same break.

`.ai/review-checklist.md` §21 lists "No return type changed in a breaking way" under Function Signatures and states that a violation is **Critical unless the deprecation protocol is fully followed**. Measured against the protocol in `BACKWARD_COMPATIBILITY.md`, this PR satisfies one of four applicable items:

- ❌ *Bridge for at least one minor version* — absent; the old contract is gone in the same commit.
- ❌ *`@deprecated` JSDoc with migration guidance and target removal version* — absent.
- ✅ *Documented in `UPGRADE_NOTES.md`* — present, and clearly written (see the Minor finding about where it lands).
- ❌ *A spec with a "Migration & Backward Compatibility" section covering the change* — `.ai/specs/2026-04-27-ds-sidebar-customization-page.md` has that section at line 175, but the PR added its note to the Changelog instead, leaving row 3 of the BC table still asserting "Function signatures | None | **No exported function signatures change.**" That table is now factually wrong.

The PR body argues no bridge is possible because "preserving the old return value would preserve the bug". That is true only for a bridge that keeps the same *name*. The bug lives in the caller, not in the old return value, so a rename-and-deprecate bridge fixes the behavior while keeping third-party code working:

```ts
export async function findSidebarPreference(
  em: EntityManager,
  scope: SidebarPreferenceScope,
): Promise<SidebarPreferencesSettings | null> {
  // ... exactly the body this PR already wrote, `if (!existing) return null` included
}

/**
 * @deprecated Since 0.7.1, slated for removal in 0.9.0. Returns fabricated defaults for a user with
 * no saved row, which is indistinguishable from a saved-but-empty preference and silently erases the
 * role layer when the result is applied. Use `findSidebarPreference` and handle `null`.
 */
export async function loadSidebarPreference(
  em: EntityManager,
  scope: SidebarPreferenceScope,
): Promise<SidebarPreferencesSettings> {
  return (await findSidebarPreference(em, scope)) ?? normalizeSidebarSettings(null)
}
```

Then point the two internal call sites at the new function — `backendChrome.tsx:407` and `api/sidebar/preferences/route.ts:207` — keep the new tests as they are but retarget the loader suite at `findSidebarPreference` (adding one case asserting the deprecated wrapper still returns non-null defaults for an absent row, which is the bridge's regression test), reword the `UPGRADE_NOTES.md` entry as a deprecation rather than a hard change, and correct row 3 of the spec's BC table.

If the maintainers would rather take the clean break than carry a deprecated alias for a minor, that is a legitimate call — but it is an explicit maintainer waiver that belongs in a PR comment, not something a reviewer can grant on their behalf. Say the word on the PR and I will treat it as waived on the next pass.

### 🔹 Minor

**2. `UPGRADE_NOTES.md:27` — after merging, the upgrade note lands inside an already-released section.**

The branch forked 226 commits back, and `develop` has since released 0.7.0: its top section is `## 0.6.7 → 0.7.0 (2026-08-12)`, while this PR writes under `## 0.6.7 → 0.6.8 (unreleased)`, a heading that no longer exists on the base. I checked what the merge actually produces rather than guessing — `git merge-tree --write-tree HEAD origin/develop` merges cleanly, and in the resulting tree the new `### loadSidebarPreference returns null …` subsection sits at line 177, i.e. **inside** the released `0.6.7 → 0.7.0 (2026-08-12)` section (which spans lines 25–382). The `0.6.8` heading vanishes entirely.

The consequence is a misleading migration document: someone upgrading to 0.7.0 reads a breaking-change note for a change 0.7.0 does not contain, and someone upgrading to the release that *does* contain it finds nothing under that version. Fix: rebase onto current `develop` and place the entry under a fresh unreleased heading (e.g. `## 0.7.0 → 0.7.1 (unreleased)`) above the released 0.7.0 section. The rebase is worth doing anyway — 226 commits of drift on a file as active as `backendChrome.tsx` is a lot of untested distance, even though the guard this fix depends on is unchanged on `develop`.

### 💅 Nit

**3. `packages/core/src/modules/auth/lib/__tests__/backendChrome.role-preference.test.ts:136` — the new chrome suite cannot fail on this regression.**

The suite mocks the entire `sidebarPreferencesService` module and injects `loadSidebarPreference` as a stub returning `null`, so the changed loader is never exercised there. I verified this by reverting the loader to its pre-fix form and re-running both new suites: 6 of the 7 tests still pass, and only `sidebarPreferencesService.load-preference.test.ts` → "returns null when the user has no saved preference row" fails. The regression *is* covered, by composition — the loader test pins "absent row ⇒ null" and the chrome test pins "null ⇒ role layer survives" — and both directions of a future regression are caught, so there is no actual coverage hole. But the PR description's claim that the chrome suite covers the regression "end-to-end … on the resolved chrome payload" overstates what it does, and that claim is used to justify skipping integration coverage. Either reword it, or stub `findOneWithDecryption` instead of the service so one test fails on the real defect.

**4. `.ai/specs/2026-04-27-ds-sidebar-customization-page.md:484` — the changelog entry is dated `2026-08-07`, but the PR was opened 2026-08-14.** Worth aligning so the spec changelog stays a reliable timeline.

**5. The deliberately-out-of-scope gap deserves a tracker issue.** The PR documents that once a user saves *any* personal preference, role hides are wiped again, because `applySidebarPreference` overwrites rather than merges and `SidebarCustomizationEditor` seeds its draft solely from the user's own `hiddenItems`. Agreed that "is a role hide policy or a default?" is a product decision and out of scope here. A note inside a spec changelog is easy to lose, though — please file an issue for it so it stays visible.

**6. `packages/core/src/modules/auth/lib/backendChrome.tsx:148` — a payload nuance worth knowing, no action needed.** For a user with *neither* a role nor a personal preference, no `applySidebarPreference` pass now runs at all, so serialized nav items omit `hidden` where they previously carried an explicit `hidden: false`. I checked every consumer — `AppShell.tsx:1049,1072,1075,1095`, `CollapsibleNavSection.tsx:104,136,144`, `SidebarCustomizationEditor.tsx:1573,1612`, `customization-helpers.ts:61` — and all of them use truthy or `!== true` checks, while `BackendChromePayload` already declares `hidden?: boolean`. So this is safe as written; I mention it only because it is a visible difference in the serialized payload that a downstream consumer doing `item.hidden === false` would notice.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — nothing is removed or renamed; the symbol and its import path are unchanged.
- [ ] **No function signature changed in a breaking way (required params removed or reordered, return type changed)** — ⚠️ **WARNING:** `loadSidebarPreference`'s return type is narrowed for consumers from `Promise<SidebarPreferencesSettings>` to `Promise<SidebarPreferencesSettings | null>` on a publicly-exported subpath, with no `@deprecated` marker and no bridge. See Blocker 1.
- [x] No required type field removed or narrowed — `SidebarPreferencesSettings` itself is untouched.
- [x] No field removed or retyped in an existing response shape — `GET /api/auth/sidebar/preferences` still emits all six settings fields via `settings?.x ?? default` (`route.ts:230-236`), so the JSON is byte-identical for a user with no row.
- [x] No database table or column renamed or removed; no column type narrowed — no schema change and no migration in the diff, correctly.
- [ ] **Where a contract had to change: old surface kept working through a deprecation window, with migration notes** — the migration notes exist in `UPGRADE_NOTES.md`, but the deprecation window does not. See Blocker 1 for a bridge that costs six lines and preserves the fix.

## 🧪 Test Coverage

The two new suites cover the changed behavior well at the unit level, and together they pin the whole causal chain:

- `sidebarPreferencesService.load-preference.test.ts` asserts the three states that matter — absent row ⇒ `null`, populated row ⇒ normalized settings, saved-but-empty row ⇒ non-`null` empty settings — plus that the `(user, tenant, organization)` filter and the decryption scope stay on the lookup, which is the tenant-isolation property of this query. This is a genuine regression test: I reverted the loader and watched the first case fail.
- `backendChrome.role-preference.test.ts` keeps the **real** `applySidebarPreference` and stubs only the loaders, so it exercises the actual merge semantics that caused the regression — a role hide and a role group order both survive a `null` user preference, and a saved-but-empty user preference still overrides the role layer, which documents the semantic the null sentinel buys. Its mock scaffolding mirrors the existing `backendChrome.group-order.test.ts` house pattern, so it fits the module's conventions.

Gaps, none of them blocking:

- No test fails on the *composed* defect (see Nit 3). If you want one, stub `findOneWithDecryption` to return `null` in the chrome suite and let the real `loadSidebarPreference` run; that single test would fail on the unfixed loader and would make the "end-to-end" claim in the PR description accurate.
- If the bridge from Blocker 1 is adopted, add one case asserting `loadSidebarPreference` (the deprecated wrapper) still returns non-`null` defaults for an absent row, so the compatibility promise itself is under test.
- Integration coverage is reasonably skipped: `.ai/qa/AGENTS.md` requires it for new API paths and key UI paths, and this change adds neither — the API response is provably unchanged. That said, the existing `TC-AUTH-SIDEBAR-GROUP-001` / `TC-AUTH-SIDEBAR-2453` specs already drive the sidebar; if a role-preference scenario is cheap to add there, it would be the natural home for a real end-to-end assertion of "role hides survive for a user who never customised their sidebar".

pkarw:

- ❌ `changes-requested` — the code review found one blocker: `loadSidebarPreference`'s return type is narrowed for consumers on a publicly exported subpath without the `@deprecated` marker and one-minor-version bridge that `BACKWARD_COMPATIBILITY.md` requires, plus one minor finding about the `UPGRADE_NOTES.md` entry landing inside the already-released `0.6.7 → 0.7.0` section after the merge.
- 🐛 `bug` — this fixes a regression in which role-level sidebar layout (hidden items and group order) was silently erased on every render for any user without a personally saved sidebar preference, exactly as @Frshy requested in the label comment above.
- 🔹 `priority-medium` — a user-visible functional regression that affects every user who never customised their sidebar, but it degrades navigation rather than blocking a release or exposing data, which matches the author's own read.
- ⚠️ `risk-high` — raised one step above the author's requested `risk-medium`: the change alters a shared contract surface (an exported function's return type on a public import path, per `BACKWARD_COMPATIBILITY.md`) and it sits on the backend chrome path that renders on every backend page, and this repository's inference rule puts shared contract surfaces in the high band regardless of how small the diff is.
- 🧪 `needs-qa` — the automated-verification exemption does not apply here because the change touches a `BACKWARD_COMPATIBILITY.md` contract surface, and the effect is visible in the rendered sidebar, so a human should exercise a role-hidden item and a role group order with and without a saved personal layout. This is the routing the author asked for.

`partner-request` was left untouched, and `qa` was deliberately not set — that label belongs to a QA reviewer picking the PR up, and auto-skills only ever request QA with `needs-qa`.

pkarw:

pkarw:

One blocker (contract-surface change to `loadSidebarPreference` without a deprecation bridge) and one minor finding (the `UPGRADE_NOTES.md` entry lands inside the already-released `0.6.7 → 0.7.0` section once this merges into current `develop`), plus four nits, are in the review above. Validation source: local fallback — GitHub reports no `validation.commands` checks for this head SHA (only `license/cla` SUCCESS and a SKIPPED `[code]smith`), so I ran the scoped substitutes in an isolated worktree: `jest src/modules/auth` in `packages/core` (71 suites / 542 tests, all passing) and `tsc --noEmit` in `packages/core` (no errors in the changed files). No merge conflicts against `develop`.

Frshy:
Thanks @pkarw — thorough review, and the blocker was a fair call. All of it is addressed in `e5a395c`, pushed on top of a rebase onto current `develop`.

**⛔ Blocker 1 — contract surface without a bridge.** You were right that "no bridge is possible" was wrong, and right about why: the bug is in the caller, not in the old return value, so a rename-and-deprecate bridge keeps both. Implemented as you sketched it:

- `findSidebarPreference(em, scope)` is new and returns `SidebarPreferencesSettings | null`. Both internal call sites (`backendChrome.tsx`, `api/sidebar/preferences/route.ts`) now use it.
- `loadSidebarPreference` keeps its name, import path, signature and behaviour, delegating via `(await findSidebarPreference(em, scope)) ?? normalizeSidebarSettings(null)`, and carries `@deprecated` JSDoc with migration guidance and a 0.9.0 removal target.

The third-party caller in your failure scenario — `const settings = await loadSidebarPreference(...); if (settings.hiddenItems.includes(id))` — compiles and runs unchanged, and there are now two tests pinning exactly that so the compatibility promise can't rot silently. All four protocol items are satisfied; the spec's BC table row 3 has been corrected from "No exported function signatures change" to describe the addition plus the deprecation.

**🔹 Minor 2 — `UPGRADE_NOTES.md` in a released section.** Confirmed your prediction exactly: after the merge the entry sat at line 177, inside `0.6.7 → 0.7.0 (2026-08-12)`. Rebased onto `develop` and moved it to a new `## 0.7.0 → 0.7.1 (unreleased)` heading above the released section, reworded as a deprecation rather than a hard change. The rebase is in the push, so the 226 commits of drift are gone.

**💅 Nit 3 — the chrome suite couldn't fail on the regression.** Took the second option rather than rewording, since the coverage is worth more than the claim. The suite now keeps the real `findSidebarPreference` alongside the real `applySidebarPreference` and drives it by stubbing the `UserSidebarPreference` row that `findOneWithDecryption` returns. Re-running your experiment: reverting the loader now fails **three** tests — the loader test plus **two chrome tests** — where before it failed one. The "end-to-end on the resolved chrome payload" claim in the description is now accurate.

**💅 Nit 4 — changelog date.** Corrected to `2026-08-14`.

**💅 Nit 5 — tracker issue for the role-hide gap.** Not filed: this account only has read permission here, and opening a public issue on the upstream tracker felt like a maintainer's call rather than something to do unilaterally. The gap is documented in the spec changelog and in the PR description's "Known gap" section — please file it (or say the word and I'll open it). To restate it so it's greppable from this thread: once a user saves *any* personal preference, role hides are wiped again, because `applySidebarPreference` overwrites rather than merges and `SidebarCustomizationEditor` seeds its draft solely from the user's own saved `hiddenItems`, never from the role-applied `hidden` flags in the base snapshot. The open question is whether role hides are policy (un-overridable) or merely a default.

**💅 Nit 6 — serialized `hidden` omitted for users with neither preference.** Noted, no action, and thanks for checking every consumer — that saved me the same pass.

**Validation** (local, no compose `app` container running):

| Command | Result |
|---|---|
| `npx jest --runInBand src/modules/auth` (`packages/core`) | ✅ 75 suites, 604 tests, all passing |
| `npx tsc --noEmit` (`packages/core`) | ✅ zero errors in every changed file |
| `npx eslint` on the 3 changed production files | ✅ clean |

Two notes on that run. First, the suite count went from 71 to 75 because the rebase brought in new auth tests from `develop`. Second, five pre-existing suites needed a test-only edit: they stub the service module and had to follow the internal call sites to `findSidebarPreference` (`admin-nav`, `backendChrome.group-order`, `backendChrome.current-organization`, `sidebar-preferences`, `find-role-tenant-scope`, `optimistic-lock`). No production behaviour rides on those edits. The `tsc` run still reports pre-existing errors elsewhere in the package from an unbuilt worktree — unresolved `@tanstack/react-table/legacy` and the `#generated/entities.ids.generated` ones you also saw — none in files this PR touches.

Re-requesting review. One thing worth flagging for the label pass: with the bridge in place this no longer breaks a `BACKWARD_COMPATIBILITY.md` contract, so if you think `risk-medium` now fits better than `risk-high`, that's your call — I still can't set labels from this account either way.

patzick (changes requested):
# 🔍 Code Review (pass 2): fix(auth): return null from `loadSidebarPreference` when no row exists

## 🎯 Summary

Every finding from the first pass is genuinely resolved, and the delivery is now the shape the deprecation protocol asks for. I re-derived the mechanism from the code at `e5a395c` rather than checking the response against my own earlier prose, and I went one step further than last time on the "what else does the skipped pass do?" question — because `applySidebarPreference` does three things beyond `hidden`, and skipping it now had to be proven harmless for all three, not just for the flag the fix is about.

**Blocker 1 (contract surface) — resolved.** `findSidebarPreference` is new and nullable; `loadSidebarPreference` keeps its name, import path, signature and behaviour behind `(await findSidebarPreference(em, scope)) ?? normalizeSidebarSettings(null)`, with `@deprecated` JSDoc naming both the migration and the 0.9.0 removal. The failure scenario I described — a third-party caller doing `settings.hiddenItems` with no null check — now compiles and runs unchanged, and two tests in `loadSidebarPreference (deprecated bridge)` pin exactly that, so the compatibility promise cannot rot silently. The migration snippet in the JSDoc is actionable: `normalizeSidebarSettings` is exported from `@open-mercato/shared/modules/navigation/sidebarPreferences` and accepts `null`, so a downstream author can paste it verbatim. All four protocol items are satisfied, and BC table row 3 is corrected from "No exported function signatures change" to "Additive + deprecation".

**Minor 2 (`UPGRADE_NOTES.md` placement) — resolved.** I checked the merged tree rather than the branch: `git merge-tree e5a395c origin/develop` produces `## 0.7.0 → 0.7.1 (unreleased)` at line 25, directly above `## 0.6.7 → 0.7.0 (2026-08-12)` at line 40. Correct after the merge, not just on the branch.

**Nit 3 (chrome suite couldn't fail on the regression) — resolved, the harder way.** The suite now `jest.requireActual`s both `applySidebarPreference` *and* `findSidebarPreference`, and drives the loader by dispatching `findOneWithDecryption` on `entity?.name === 'UserSidebarPreference'`. The mock factory exports exactly the three members `backendChrome.tsx` imports, so nothing is silently `undefined`. The "end-to-end on the resolved chrome payload" claim in the description is now accurate.

**Nit 4** — changelog date is `2026-08-14`. **Nit 5** — filed as #5372 (I have the permission you don't; the write-up is yours, credited). **Nit 6** — confirmed no action needed, and re-confirmed below.

The five renamed mocks in pre-existing suites are mechanical and follow the internal call sites; I checked `git grep loadSidebarPreference` on both `e5a395c` and `origin/develop` and there is no call site — internal or on the 91 commits of drift — left pointing at the old name outside the deprecated bridge and its own tests.

### Additional verification this pass

The first review only established that skipping the empty user pass is safe for `hidden`. But `applySidebarPreference` with empty settings also (a) rewrites every `title` to `defaultTitle` and every group `name` to `defaultName`, and (b) re-sorts groups. If either were load-bearing, skipping the pass would be a fresh regression for the very users the fix targets. Both are no-ops:

- **Labels** — `adoptSidebarDefaults` runs immediately before, setting `defaultTitle = title` and `defaultName = name`. So `title = override ?? defaultTitle` resolves to the unchanged `title`. No-op by construction.
- **Group order** — `groupEntries` ends in `normalizeGroupWeights`, which sorts by `resolveGroupOrder()` rank → weight → name and then *rewrites* `group.weight` to `rank * 1_000_000 + clamp(fallbackWeight)`, i.e. monotonically increasing in the order it just produced. `applySidebarPreference`'s empty-settings sort is `orderIndex (all ∞) → weight → defaultName`, which on a monotonic-weight array reproduces the same order. Idempotent.
- **Item order within a group** — `applyItems` maps, it does not sort. Unaffected either way.
- **`hidden` for a user with neither preference** — `AdminNavItem.hidden` is declared in `packages/ui/src/backend/utils/nav.ts:19` but never assigned by `buildAdminNav`, so it is `undefined` rather than the previous explicit `false`. `BackendChromePayload` already declares `hidden?: boolean` and every consumer uses truthy / `!== true`. Confirms Nit 6.

So the only behavioural delta from skipping the pass is the one the fix intends. I found no new defect in the code.

## Verdict

⛔ **Request changes** — **not for the code**, which I would approve as written. The blocker is that this head has no CI verification at all, and I cannot supply it myself. One `git merge origin/develop` from you clears it. Details below.

## 🧪 Validation Gate

**Validation source: GitHub checks.** No local fallback was run — the extension's fallback exists for when GitHub check data is *unavailable or unusable*, and here it is available and legible; it simply reports that the gate never executed. Running local substitutes would swap a real signal for a weaker one.

| Check | Result |
|---|---|
| `prepare` | ✅ pass (3m6s) |
| `lint` | ✅ pass |
| `ds-lint` | ✅ pass |
| `scope` / `audit-scope` | ✅ pass |
| `license/cla` | ✅ pass — signed |
| **`audit`** | ❌ **fail** |
| `test` | ⏭️ **never ran** (gated on `audit`) |
| `docker-build` | ⏭️ never ran |
| `ephemeral-integration` (15 shards) | ⏭️ never ran |
| `merge-coverage` | ⏭️ never ran |

### ⛔ Blocker 1 — the branch is 91 commits behind `develop`, so `audit` fails and takes the entire test gate down with it

From [the `audit` job log](https://github.com/open-mercato/open-mercato/actions/runs/31815822122/job/95669443292):

```
audit-ci: 1 advisory(ies) at or above high:
  [high] deepmerge-ts <8.0.0 — stack exhaustion when merging recursive object graphs (GHSA-ggr8-5vv4-36mx)
```

Nothing to do with your change — `develop` already carries the fix as `b708b7bb6 fix(deps): pin deepmerge-ts to 8.0.1 to clear GHSA-ggr8-5vv4-36mx`, which landed after your rebase point. The consequence is what matters: `test`, `docker-build`, `merge-coverage` and all fifteen `ephemeral-integration` shards are gated on `audit` and never started, so **not one automated test has run against this head in CI**. Your local run (75 suites / 604 tests) is the only test evidence on the PR, and a reviewer cannot verify it.

I confirmed the merge is trivial and does the job, using git plumbing only — no build, no test run:

- `git merge-tree --write-tree e5a395c origin/develop` → clean, **zero conflicts**, tree `a1fa6a1`.
- In that tree: `backendChrome.tsx:439` still reads `await findSidebarPreference(...)` with the `userPreference ? … : baseForUser` guard intact at 451–453; `UPGRADE_NOTES.md` heading order is correct; `package.json` resolves `deepmerge-ts: 8.0.1`, which is what clears the advisory.

**Please run `git merge origin/develop` (or rebase) and push.** CI will then produce a real gate result, and I will re-review on the green run.

I would have pushed this for you — the first-pass thread promised a trivial-fix push, and I built the merge commit locally and verified it byte-for-byte against the tree above. GitHub refused it: `fullstackhouse/open-mercato` is an **organization-owned** fork, and "Allow edits by maintainers" is not honoured for org-owned forks even though the API reports `maintainer_can_modify: true`. My token resolves to `push: false` on the fork (`remote: Permission to fullstackhouse/open-mercato.git denied`). So this genuinely needs your push, not mine — apologies for the round-trip.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path.
- [x] **No function signature changed in a breaking way** — resolved. `loadSidebarPreference` keeps `Promise<SidebarPreferencesSettings>`; the nullable contract lives on the new `findSidebarPreference`.
- [x] **Old surface kept working through a deprecation window, with migration notes** — resolved. Bridge + `@deprecated` JSDoc + `UPGRADE_NOTES.md` under a fresh unreleased heading + spec BC table row 3 corrected.
- [x] No required type field removed or narrowed; no route, response shape, event, CLI, schema, or config-key change. `GET /api/auth/sidebar/preferences` still emits all six fields via `settings?.x ?? default`, so the JSON is byte-identical for a user with no row.

## 🧪 Test Coverage

Good, and it now fails on the real defect. Reverting the loader breaks three tests across two suites (loader + chrome) rather than one, because the chrome suite runs the real `findSidebarPreference` against a stubbed row. The three loader states that matter — absent ⇒ `null`, populated ⇒ normalized, saved-but-empty ⇒ non-`null` empty — are each pinned, the `(user, tenant, organization)` filter and decryption scope are asserted (the tenant-isolation property of this query), and the deprecated bridge has its own regression test.

Integration coverage remains reasonably skipped: no new API path, no new UI surface, response provably unchanged. If you ever want the belt-and-braces version, `TC-AUTH-SIDEBAR-GROUP-001` is the natural home for "role hides survive for a user who never customised their sidebar" — but I am not asking for it here.

---

**To summarise what is left:** merge `develop`, push, let CI go green. No code changes requested.

patzick:
🤖 `om-auto-review-pr` (pass 2) — 🏷️ label rationale

- ❌ `changes-requested` (kept) — the code itself is approvable, but this head has **no CI test signal**: `audit` fails on `deepmerge-ts <8.0.0` (GHSA-ggr8-5vv4-36mx, already fixed on `develop` by `b708b7bb6`, 91 commits ahead of this branch), and `test` / `docker-build` / `merge-coverage` / all 15 `ephemeral-integration` shards are gated on it and never started. This repo's rule treats a failed validation-gate command as a blocker regardless of whose commit caused it. One `git merge origin/develop` clears it — I verified the merge is conflict-free and that the resulting tree keeps the fix intact and resolves `deepmerge-ts@8.0.1`.
- 🐛 `bug` (kept) — unchanged: this fixes role-level sidebar layout being erased on every render.
- 🔹 `priority-medium` (kept) — unchanged read; a user-visible navigation regression, not a release blocker or data exposure.
- ⚠️ `risk-high` (kept, and @Frshy explicitly asked me to re-examine this) — the `BACKWARD_COMPATIBILITY.md` argument for it is indeed gone now that the bridge is in place, and that was fair to raise. But it was never the only one: `resolveBackendChromePayload` renders on **every** backend page, and the fix works by *removing* a merge pass that previously ran unconditionally for the majority of users. That is a broad blast radius on the auth path, which this repo's inference rule puts in the high band independently of contract surfaces. Holding at `risk-high` — with `test` and all 15 integration shards never having run, downgrading blast-radius signal on an unverified head is the wrong direction.
- 🧪 `needs-qa` (kept) — the automated-verification exemption still does not apply: the effect is visible in the rendered sidebar. Worth exercising a role-hidden item and a role group order both with and without a saved personal layout.
- 🔄 `in-progress` — removed; review lock released.

`partner-request` left untouched. `qa` deliberately not set — that belongs to a QA reviewer picking this up; auto-skills only ever request QA with `needs-qa`.

Separately: nit 5 from the first pass is now filed as #5372 (role hides are wiped once a user saves *any* personal preference). @Frshy — the diagnosis and wording there are yours, credited in the issue; I just had the permission to open it.

patzick:

No code findings — every item from the first pass (1 blocker, 1 minor, 4 nits) is genuinely resolved, and I additionally proved that skipping the now-guarded `applySidebarPreference` pass is a no-op for labels, group order and item order, not just for `hidden`. I would approve the diff as written.

The one blocker is CI, not code: `audit` fails on `deepmerge-ts <8.0.0` (GHSA-ggr8-5vv4-36mx) because the branch predates `b708b7bb6` on `develop` by 91 commits, and that failure gates `test`, `docker-build`, `merge-coverage` and all 15 `ephemeral-integration` shards — none of which started. So zero automated tests have run against this head.

**Validation source: GitHub checks** (no local fallback — the check data is available and legible, it just reports the gate never ran).

@Frshy please `git merge origin/develop` and push. I verified with `git merge-tree` that it is conflict-free and that the merged tree keeps the fix intact while resolving `deepmerge-ts@8.0.1`. I tried to push it for you but `fullstackhouse/open-mercato` is an org-owned fork, where GitHub does not honour "Allow edits by maintainers" — the API reports `maintainer_can_modify: true` but the push is refused.

Ping me on the green run and I will re-review.

autofix: not applicable — no code changes were requested.

patzick:
🤖 `om-auto-review-pr` — branch updated, CI running. Re-claiming for the verification pass.

Correction to my previous comment: I said this "genuinely needs your push, not mine." That was wrong. The raw `git push` to the org-owned fork is refused, but GitHub's server-side `PUT /pulls/5305/update-branch` API honours `maintainer_can_modify` and went through. Sorry for the unnecessary round-trip, @Frshy — no action needed from you.

`develop` is now merged in as `5946510`. Two things I checked rather than assumed:

- The resulting tree is **byte-identical** to the merge commit I had built and verified locally (`git diff` against tree `a1fa6a1` is empty), so the conflict-free/fix-intact verification in my review still holds verbatim.
- The branch is now **0 commits behind** `develop`, so the `deepmerge-ts@8.0.1` pin from `b708b7bb6` is in and the `audit` advisory that gated the whole test stage is gone.

The fork run needed maintainer approval (`action_required`); approved both `CI for Develop&Main` and `Mutation tests`. `test`, `docker-build`, `merge-coverage` and the 15 `ephemeral-integration` shards should now actually execute for the first time on this change.

Standing by for the gate result — I will re-review and update the verdict on it.

patzick (approved):
# ✅ Code Review (pass 3): fix(auth): return null from `loadSidebarPreference` when no row exists

## Verdict

✅ **Approve.** The sole blocker from pass 2 was the absence of any CI signal, not the code. `develop` is merged in (`5946510`) and the gate is now green, including the first real execution of `test` and `ephemeral-integration` against this change. No code findings in pass 2 and none here.

## 🧪 Validation Gate

**Validation source: GitHub checks** — run [`32151494577`](https://github.com/open-mercato/open-mercato/actions/runs/32151494577) on head `5946510`. No local execution.

| Check | Result |
|---|---|
| `prepare` | ✅ pass (4m31s) |
| **`test`** | ✅ **pass (8m1s)** — 1288 suites / 10087 tests |
| **`ephemeral-integration (none)`** | ✅ **pass (10m31s)** |
| `lint` | ✅ pass |
| `ds-lint` | ✅ pass |
| `audit-scope` / `scope` | ✅ pass |
| `license/cla` | ✅ pass |
| `audit` | ⏭️ skipping — same state it has on `develop`; the `deepmerge-ts <8.0.0` advisory is gone now the branch carries `b708b7bb6` (`deepmerge-ts@8.0.1`) |
| `docker-build`, `merge-coverage`, `mutate`, `comment` | ⏭️ skipping (path/scope guards) |

Two things I verified rather than accepting the green tick at face value:

**1. The new suites actually ran.** A passing job says nothing about whether the tests that matter were collected. From the `test` log:

```
PASS src/modules/auth/services/__tests__/sidebarPreferencesService.load-preference.test.ts
PASS src/modules/auth/lib/__tests__/backendChrome.role-preference.test.ts
```

Both new suites executed, plus every pre-existing sidebar/nav suite whose mocks were renamed — `sidebarPreferencesService.scope`, `backendChrome.group-order`, `backendChrome.current-organization`, `backendChrome.settings-section-order`. So the five mechanical mock renames are confirmed correct by CI, not just by inspection.

**2. `ephemeral-integration (none)` is a real run, not a skip.** I want to correct something I implied earlier in this thread: the `(none)` in the job name reads like "no integration tests", and it is not. In `prepare`'s scope step, `SINGLE_SHARD='["none"]'` is the sentinel for *one unsharded runner*; the job is skipped only when `skip_integration == 'true'`, which is reserved for CI/docs/scripts-only PRs. This diff touches `packages/core/src/modules/auth/`, so the scope step resolved `modules=auth`, `skip=false`, and ran the integration suite filtered to the affected module — for 10m31s. The 15-shard fan-out is reserved for pushes and for PRs touching `packages/shared/`, `packages/ui/`, `packages/core/src/lib/` and friends, none of which this diff touches. `merge-coverage` skipping is just its `shard_matrix != '["none"]'` guard, not a missing gate.

So this change has both unit and auth-scoped integration coverage behind it. That is a stronger signal than the "no new API path, no new UI path" argument the checklist item was resting on.

## Recap of the review

Pass 2 verified all six findings from pass 1 resolved, and found no new defects. Briefly, for the record:

- **Blocker (BC contract) — resolved.** `findSidebarPreference` is the new nullable function; `loadSidebarPreference` keeps its name, import path, signature and behaviour as a `@deprecated` bridge (removal target 0.9.0), with two tests pinning the compatibility promise so it cannot rot silently.
- **Minor (`UPGRADE_NOTES.md`) — resolved**, verified in the merged tree rather than on the branch.
- **Nits 3, 4, 6 — resolved / confirmed no-action.** The chrome suite now runs the real loader, so reverting the fix breaks three tests across two suites instead of one.
- **Nit 5 — filed as #5372**, credited to @Frshy.

Beyond re-checking those, pass 2 established that skipping the now-guarded `applySidebarPreference` pass is a no-op for labels (`adoptSidebarDefaults` runs immediately before), for group order (`normalizeGroupWeights` already emits monotonic weights, so the empty-settings sort is idempotent) and for item order (`applyItems` maps, never sorts). The only behavioural delta is the intended one.

## 🚦 Merge gate

**Not merging, and this must not be merged yet.** It carries `needs-qa` without `qa-approved`, which is a hard block in this repo regardless of how green CI is. A QA reviewer should exercise a role-hidden item and a role group order in two states: a user with no saved personal layout (role layer must survive) and a user with one (role layer is currently overridden — that is the known gap tracked in #5372, not a defect in this PR).

`changes-requested` removed, `in-progress` released.

patzick:
🤖 `om-auto-review-pr` (pass 3) — 🏷️ label rationale

- ✅ `changes-requested` — **removed**. The pass-2 blocker was the absence of any CI signal, not the code. `develop` is merged in (`5946510`) and the gate is green: `test` ✅ 8m1s (1288 suites / 10087 tests), `ephemeral-integration (none)` ✅ 10m31s, `lint` / `ds-lint` / `prepare` / `scope` / `audit-scope` ✅. `audit` is now `skipping`, the same state it holds on `develop`.
- 🔄 `in-progress` — removed; review lock released.
- 🐛 `bug` (kept) — unchanged.
- 🔹 `priority-medium` (kept) — unchanged.
- ⚠️ `risk-high` (kept) — I said in pass 2 that I was holding at high partly because the head was unverified, and that half of the argument is now spent. But the other half stands on its own: `resolveBackendChromePayload` renders on every backend page, and the fix works by *removing* a merge pass that previously ran unconditionally for most users. Broad blast radius on the auth path keeps it in the high band. Green CI lowers the probability of a defect, not the blast radius if one exists — and `risk-*` in this repo measures the latter. Happy to be overruled by a maintainer who reads it as medium.
- 🧪 `needs-qa` (kept) — the automated-verification exemption still does not apply: the effect is visible in the rendered sidebar.

`partner-request` untouched. `qa` deliberately not set — that belongs to a QA reviewer picking this up.

## ⚠️ Do not merge yet

`reviewDecision` is now `APPROVED` and `mergeStateStatus` is `CLEAN`, so GitHub will offer the merge button. **The repo rule overrides it:** a PR carrying `needs-qa` must not be merged without `qa-approved`, however green CI is. I have not merged and will not.

**Suggested QA script** — exercise a role-level hide and a role group order in two states:
1. A user who has **never** saved a personal sidebar layout → the role hide and role group order must both survive. This is the regression being fixed.
2. A user **with** a saved personal layout → the role layer is currently overridden. That is expected today and is the known gap tracked in #5372, *not* a defect in this PR.

I also dismissed the stale `CHANGES_REQUESTED` from 2026-08-14, since every finding in it is verified resolved on the current head (rationale in the dismissal message).

patzick:

**Validation source: GitHub checks** — run [`32151494577`](https://github.com/open-mercato/open-mercato/actions/runs/32151494577) on head `5946510`. No local execution at any point in this pass.

Journey across three passes: pass 1 found a BC blocker, pass 2 verified the bridge resolved it and found no new defects but had zero CI signal to stand on, pass 3 supplied that signal. `develop` is merged in via the `update-branch` API, `test` and an auth-scoped `ephemeral-integration` both ran green for the first time on this change, and I confirmed from the `test` log that the two new suites were actually collected rather than trusting the exit code:

```
PASS src/modules/auth/services/__tests__/sidebarPreferencesService.load-preference.test.ts
PASS src/modules/auth/lib/__tests__/backendChrome.role-preference.test.ts
```

**Next step is QA, not merge** — `needs-qa` is set and `qa-approved` is not. See the label comment above for the suggested QA script.

Follow-up #5372 tracks the deliberately out-of-scope half of this bug (role hides wiped once a user saves any personal preference).

Thanks @Frshy — the bridge, the reworked chrome suite and the rebase were all done exactly as asked, and the diagnosis in the description held up under scrutiny at every pass.

autofix: not applicable — no code changes were requested.

Issue #5296, opened 2026-08-14
security(ent_security): passkey enrollment still accepts an unattested client-supplied public key, keeping the MFA bypass live (PasskeyProvider.ts:33-45)


pkarw:
## Summary

`PasskeyProvider.confirmSetup` still accepts a legacy enrollment shape that stores a **client-supplied** public key with no attestation:

https://github.com/open-mercato/open-mercato/blob/develop/packages/enterprise/src/modules/security/lib/providers/PasskeyProvider.ts#L33-L45

```ts
const setupConfirmationPayloadSchema = z.union([
  z.object({ response: z.record(z.string(), z.unknown()), label: … }),
  z.object({                       // ← this member
    credentialId: z.string().min(1),
    publicKey: z.string().min(1),
    challenge: z.string().min(1),
    …
  }),
])
```

The non-`response` branch (`PasskeyProvider.ts:211-223`) only checks that `parsed.challenge === pending.challenge` — a value the server already handed the caller in the `setup` response — and then persists `credentialPublicKey: parsed.publicKey` verbatim.

#5291 closed the equivalent hole on the **verify** side. This is the remaining half, and it is still an exploitable second-factor bypass rather than a leftover.

## Why it is a bypass, not just dead credentials

The natural reading is that a fabricated public key is harmless because no authenticator can sign for it. That is only true for honest fixtures (the repo's own `enrollPasskey` integration helper stores `Buffer.from('public-key:…')`). An **attacker** supplies the public key of a keypair *they* generated, then signs a genuine WebAuthn assertion in software — which `verifyAuthenticationResponse` accepts, because from its point of view the credential is real.

## Reachability from a pre-MFA session

- `PUT /api/security/mfa/provider/passkey` declares only `metadata.PUT = { requireAuth: true }` plus `authorizeMfaEnrollmentMutation` (`security.mfa.manage`, or the compelled-enrollment recovery path).
- `checkAuthorization` (`apps/mercato/src/app/api/[...slug]/route.ts:150`) only asserts that a session exists.
- `mfa_pending` is checked in exactly three routes — `api/mfa/verify/route.ts:27`, `api/mfa/prepare/route.ts:25`, `api/mfa/recovery/route.ts:25`. The enrollment route is **not** one of them.

[...]

PR #5306, opened 2026-08-14
security(security): require a genuine WebAuthn assertion for passkey MFA verification

### LINKED ISSUE #3852: security(ent_security): passkey MFA second factor bypassable via non-cryptographic fallback (PasskeyProvider.ts:47-55)

pkarw:
Closes #3852
Tracking plan: .ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
Status: complete

## 🎯 Goal

Make a cryptographically verified WebAuthn assertion the only way to pass the passkey second factor.

`PasskeyProvider.verify()` accepted a `{ credentialId, challenge }` payload beside the genuine `{ response }` assertion and approved it by plain string comparison. Both compared values are **public**: `prepareChallenge()` hands the credential id and the challenge to the caller, and `GET /api/security/mfa/methods` discloses `providerMetadata.credentialId`. A third route needed even less — when the client never called `/api/security/mfa/prepare`, `MfaVerificationService.verifyChallenge` built no verify context, and the provider fell through to comparing only the disclosed credential id, ignoring the challenge entirely.

So anyone who could reach the verify step for a session — the party already holding `mfa_pending`, or a replay of the prepare response — passed the passkey second factor with no authenticator private key and no signature. `SudoChallengeService.verify` → `MfaVerificationService.verifyChallenge` → `provider.verify` is the same path, so this defeated login-time MFA **and** passkey-as-sudo step-up.

### Why this PR exists rather than #5291

PR #5291 proposed the same remediation as an outside contribution against `packages/enterprise`, our commercial package, which does not accept third-party contributions. It was closed unmerged and the fix is re-authored here first-party from the issue report and the current `develop` source; nothing from that diff is carried over.

## What Changed

- **`packages/enterprise/src/modules/security/lib/providers/PasskeyProvider.ts`** — `verifyPayloadSchema` collapses from a union to a single object requiring `response`. The context-absent `credentialId` fallback and the trailing `credentialId` + `challenge` string-compare block are gone, so a verified `verifyAuthenticationResponse` is the only route to a positive verdict. The challenge TTL check, the signature-counter update, and the `{ response }` path itself are untouched.
- **Failure mode, deliberately chosen** — a payload that fails the schema, and an assertion the WebAuthn verifier throws on, return `false` rather than propagating. A thrown error reaches `mapMfaError` as a logged **500** and skips `registerFailedAttempt`, so rejected attempts would stop counting toward the challenge attempt limit. Returning `false` keeps the **401** the route documents and preserves lockout. No verdict flips from negative to positive.
- **`lib/__tests__/PasskeyProvider.test.ts`** — the test named *"supports legacy verification payload for backward compatibility"* asserted that the bypass worked; it is replaced with regression coverage for every route that used to reach a positive verdict without a signature.
- **`lib/__tests__/PasskeyProviderWebAuthn.test.ts`** (new) — runs the **real** `@simplewebauthn` verifier against a software authenticator, so the removal is proven not to break real passkeys.
- **`__integration__/`** — the Playwright specs used the removed shape as their QA shortcut, so they asserted a behavior that is now a vulnerability. `verifyPasskeyChallenge` becomes the explicitly negative `attemptUnsignedPasskeyVerify`; `TC-SEC-004` asserts login verification answers `401`, and `TC-SEC-006`/`TC-SEC-007` assert passkey sudo step-up answers `401` and then skip their sudo-token-gated remainders with a stated reason.
- **`BACKWARD_COMPATIBILITY.md` / `UPGRADE_NOTES.md` / `.ai/qa/scenarios/TC-SEC-004-…md`** — record the deliberate contract break, what client authors must send, which enrolled credentials stop working, and that the passing passkey ceremony stays manual for now.

## 🧪 Tests

The evidence that matters for a change this risky is the new `PasskeyProviderWebAuthn.test.ts`. It does **not** mock `@simplewebauthn/server`. It generates a P-256 key pair, encodes the COSE public key the way `confirmSetup` stores it, and signs a genuine assertion over the challenge `prepareChallenge()` produced — then asserts that:

- a real assertion still verifies and advances the stored signature counter (the regression that would lock every passkey user out if the tightening went too far);
- a tampered signature is refused and the counter is not advanced;
- an assertion replayed against a different prepared challenge is refused;
- the disclosed `{ credentialId, challenge }` pair is refused.

`PasskeyProvider.test.ts` adds five cases, all against the mocked verifier: the legacy shape is refused even with a correctly prepared context; a `credentialId`-only payload is refused when the client skipped `/prepare`; a well-formed assertion is refused when no challenge was prepared; an assertion the authenticator did not sign is refused without advancing the counter; and an assertion that makes the verifier throw resolves to `false` instead of propagating. The first three also assert `verifyAuthenticationResponse` was never reached.

Three of these five fail against the pre-fix provider (verified by checking the old file back out and re-running); the other two pin behavior that was already correct and must stay correct.

## 💥 Breaking Changes

`POST /api/security/mfa/verify` and `POST /api/security/sudo/verify` no longer accept a `{ credentialId, challenge }` passkey payload — it answers **401** instead of succeeding.

Per `BACKWARD_COMPATIBILITY.md` an API request shape is contract surface (category 7) and the protocol requires a bridge release. **The bridge is deliberately skipped here**, because the shape being removed *is* the vulnerability: a bridge would leave the second-factor bypass live for a full minor version in both login MFA and sudo step-up.

This now ships under a **defined** clause rather than a one-off waiver. `BACKWARD_COMPATIBILITY.md` gains an **Emergency Security Exception** to the Deprecation Protocol: it waives steps 1-3 only when the surface being removed is itself the vulnerability, and only against five stated requirements — a reasoned qualifying argument, the narrowest possible removal with **no partial bridge retaining the vulnerable branch**, steps 4 and 5 still in force, a dated entry, and an explicit human maintainer sign-off that no automated review can supply. The passkey entry invokes that clause and shows how each requirement is met.

The classification is stated rather than dodged: the removed shape was *publicly supported* — exported in `MfaProviderInterface.verifySchema` and pinned by a test literally named *"supports legacy verification payload for backward compatibility"* — so this is a genuine STABLE break authorized by the exception, not a claim that no contract ever existed.

Protocol step 5 requires a spec, so this PR adds [`.ai/specs/enterprise/2026-08-14-passkey-mfa-require-webauthn-assertion.md`](.ai/specs/enterprise/2026-08-14-passkey-mfa-require-webauthn-assertion.md) with the mandated **Migration & Backward Compatibility** section.

The first-party UI is unaffected — `PasskeyChallengeVerify.tsx` sends `startAuthentication()` output as `{ response }`, and `PasskeyProviderDetails.tsx` enrolls through `startRegistration()`. Any third-party client sending the legacy shape was, by construction, not performing cryptographic verification.

**Operator impact — corrected in the review round.** A passkey enrolled through the setup path's unattested `publicKey` shortcut (#5296, still open) is **not** reliably rendered unusable by this change. Such a row is one of two things, and the row does not say which:

- **An unusable key** — nobody holds the private half, so it can never produce a verifiable assertion and now fails every login. A user whose only MFA method is such a credential needs an admin MFA reset.
- **An attacker-controlled but perfectly valid keypair** — a caller who generated a real P-256 keypair in software and enrolled its COSE public key holds a credential that signs assertions this PR *accepts*. Requiring a real assertion does not help; the key is real, it is simply not the user's. That is an account-takeover path, not a lockout inconvenience.

It matters most for a password-only (`mfa_pending`) session: `authorizeMfaEnrollmentMutation` authorizes on `auth.sub` and does not reject a pending token, so an attacker with a stolen password may be able to enroll a key they control and then satisfy the second factor with it.

Provenance is **not** reconstructible from the stored row — both enrollment paths write identical `provider_metadata` keys and the shortcut lets the caller choose all of them — so `UPGRADE_NOTES.md` no longer promises a per-row audit. It now gives a working enumeration query against the real schema (`user_mfa_methods`, discriminated by column `type`; the earlier note named a nonexistent `mfa_methods` table) and conservative remediation: sequence #5296 ahead of the upgrade where possible, otherwise reset and re-enroll every passkey method on any deployment that ever provisioned passkeys through the API.

## Related work — this closes one of two halves

- **#5296** — the *enrollment* side still accepts a client-supplied `publicKey` with no attestation. That issue argues, correctly, that an attacker who supplies the public key of a keypair they generated can then sign a genuine assertion in software. This PR does not close that; it is a separate fix on a separate surface and both need to land.
- **#5307** (filed by this run) — passkey verification now has no integration-level happy path, and `TC-SEC-006`/`TC-SEC-007` lose their sudo-token-gated assertions. Restoring them needs a Playwright virtual authenticator, or decoupling sudo coverage from passkeys onto TOTP.

## ✅ Validation

Full `validation.commands` gate from `.ai/agentic.config.json`, local runner (no compose `app` container running), in order: `build:packages` ✅ · `generate` ✅ (no generated drift) · `build:packages` ✅ · `i18n:check-sync` ✅ · `i18n:check-usage` ✅ · `typecheck` ✅ (22/22) · `test` ⚠️ · `build:app` ✅. `yarn lint`, which is not part of the configured gate, also passes with 0 errors.

`@open-mercato/enterprise` is **60/60 suites, 486/486 tests green**. `yarn test` exits non-zero on exactly one package, `create-mercato-app`, whose assertion is `❌ Linux file-watch limits are too low for Turbopack` — this machine's `fs.inotify.max_user_instances` is 128. That package is not in this branch's diff at all, so the failure cannot originate here; Linux CI is the authority.

The first review pass raised one major finding against this branch — the new `catch` around `verifyAuthenticationResponse` swallowed the verifier's error, so a misconfigured `rpId`/`expectedOrigins` would answer a silent `401` where it previously reached `mapMfaError` and got logged. Fixed in `fb98bf0d9`: the rejection is unchanged, but it now emits a `warn` through the logging facade.

🧪 **Manual QA is still pending.** This sits in the login MFA and sudo step-up flows and the passing passkey ceremony cannot be automated yet (#5307), so a reviewer should confirm that a real passkey login and a real passkey sudo step-up still succeed against a genuine authenticator.

## 🔁 Review round 2 — strict review addressed (`4b9191066`)

All three findings from the strict `CHANGES_REQUESTED` review on `93c93e87f` are addressed; no source behavior changed, the fix itself is untouched.

| # | Finding | Resolution |
|---|---------|-----------|
| Critical | `BACKWARD_COMPATIBILITY.md:341` — waiver granted by fiat against a policy forbidding every unbridged STABLE change | Added a defined **Emergency Security Exception** to the Deprecation Protocol with five hard requirements; classified this break as a real category-7 STABLE break (the shape was publicly supported) and showed the entry meeting each requirement; wrote the migration spec step 5 requires |
| High | `UPGRADE_NOTES.md:35` — describes shortcut-enrolled credentials as necessarily unusable | Rewritten: such a credential is *either* unusable *or* an attacker-controlled valid keypair that this fix accepts; states the `mfa_pending` enrollment reachability; instructs treating all such credentials as potentially hostile; gives release-sequencing and reset-and-re-enroll mitigations |
| Medium | `UPGRADE_NOTES.md:35` — audit instruction does not map to the schema | Query corrected to `user_mfa_methods` / column `type` (verified against the entity and `Migration20260305120000_security_scaffold`); states plainly that provenance is not reconstructible and replaces per-row auditing with conservative reset-and-re-enroll |

The two nits from the first review are also handled: TC-SEC-006/007 now carry a header comment explaining that the body stops at the mid-body `test.skip`, and the `UPGRADE_NOTES.md` audit instruction is now executable.

⚠️ **Requirement 5 of the new exception is a human action and is not satisfied yet:** the waiver needs an explicit maintainer approval that acknowledges it. An automated review cannot clear it, by design.

## 📋 Progress

See the Progress section in the tracking plan.

### DIFF
--- a/.ai/qa/scenarios/TC-SEC-004-passkey-enrollment-and-login.md
+++ b/.ai/qa/scenarios/TC-SEC-004-passkey-enrollment-and-login.md
@@ -43,3 +43,8 @@ Verify that a user can register a passkey MFA method from the security profile a
 - Cancel the browser passkey ceremony and expect the enrollment to remain incomplete
 - Attempt login challenge with a missing or rejected WebAuthn assertion and expect the session to remain MFA-pending
 - Verify that the test skips explicitly rather than failing when WebAuthn is unsupported in CI
+
+## Automation Coverage
+Steps 1-7 and the "no assertion" edge case are automated in `TC-SEC-004.spec.ts`, which asserts that a login challenge answered with the disclosed credential id and challenge rather than a signed assertion is refused with `401` (the second-factor bypass fixed in #3852).
+
+Step 8 — the passing browser verification ceremony — stays **manual**. A genuine assertion needs an authenticator, so automating it requires a Playwright virtual authenticator (`WebAuthn.addVirtualAuthenticator`); until that lands, a passing passkey login and a passing passkey sudo step-up must be confirmed by hand against a real authenticator.
--- a/.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
+++ b/.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
@@ -68,6 +68,8 @@ The Playwright specs used the bypass as their shortcut, so they assert a behavio
 
 ## Progress
 
+PR: #5306
+
 > Convention: `- [ ]` pending, `- [x]` done. Append ` — <commit sha>` when a step lands. Do not rename step titles.
 
 ### Phase 1: Close the bypass in PasskeyProvider
@@ -92,4 +94,4 @@ The Playwright specs used the bypass as their shortcut, so they assert a behavio
 
 ### Phase 5: Validation and delivery
 
-- [ ] 5.1 Run the full validation gate and finalize the PR
+- [x] 5.1 Run the full validation gate and finalize the PR — fb98bf0d9 (review pass raised one major finding, fixed in that commit)
--- a/.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
+++ b/.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
@@ -95,3 +95,11 @@ PR: #5306
 ### Phase 5: Validation and delivery
 
 - [x] 5.1 Run the full validation gate and finalize the PR — fb98bf0d9 (review pass raised one major finding, fixed in that commit)
+
+### Phase 6: Address the strict review round
+
+- [x] 6.1 Resolve the compatibility waiver against the policy gate — added a defined Emergency Security Exception to the Deprecation Protocol and reclassified the passkey entry as an instance of it rather than a one-off
+- [x] 6.2 Write the migration spec the protocol's step 5 requires — `.ai/specs/enterprise/2026-08-14-passkey-mfa-require-webauthn-assertion.md`
+- [x] 6.3 Correct the operator guidance — a shortcut-enrolled credential may hold an attacker-controlled valid keypair, not only an unusable key
+- [x] 6.4 Fix the audit instruction to the real schema (`user_mfa_methods`, column `type`) and replace per-row auditing with reset-and-re-enroll, since provenance is not reconstructible
+- [x] 6.5 Note the mid-body `test.skip` in TC-SEC-006/007 so the unreachable tail is not a surprise
--- a/.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
+++ b/.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
@@ -1,0 +1,95 @@
+# Execution plan — require a genuine WebAuthn assertion for passkey MFA verification
+
+**Issue:** [#3852](https://github.com/open-mercato/open-mercato/issues/3852) — `security(ent_security): passkey MFA second factor bypassable via non-cryptographic fallback`
+
+**Supersedes:** PR [#5291](https://github.com/open-mercato/open-mercato/pull/5291), a third-party contribution that was closed without merging. `packages/enterprise` is the commercial package and we do not accept outside contributions there, so the fix is re-authored first-party from the issue report and the current `develop` source. No content from the closed PR's diff is carried over.
+
+## Goal
+
+Make a cryptographically verified WebAuthn assertion the only way to pass the passkey second factor, for both login-time MFA (`POST /api/security/mfa/verify`) and passkey-as-sudo step-up (`POST /api/security/sudo/verify`).
+
+## Scope
+
+`packages/enterprise/src/modules/security` only:
+
+- `lib/providers/PasskeyProvider.ts` — the verify payload schema and the two non-cryptographic acceptance branches in `verify()`.
+- `lib/__tests__/PasskeyProvider.test.ts` — replace the test that pins the bypass as intended behavior with regression coverage.
+- a new real-crypto test that signs a genuine assertion with a software authenticator, so the removal is proven not to break real passkeys.
+- `__integration__/` — the Playwright specs and the shared fixture that used the bypass as an API-level shortcut.
+- `BACKWARD_COMPATIBILITY.md` and `UPGRADE_NOTES.md` — record the deliberate, non-deprecated request-shape removal.
+
+### Non-goals
+
+- The **setup-side** shortcut (`setupConfirmationPayloadSchema` accepting a client-supplied `publicKey` with no attestation). It is a separate defect with its own blast radius and gets its own issue.
+- Restoring an integration-level passkey happy path. Producing a genuine assertion over the API is impossible by design; it needs a Playwright virtual authenticator driving the browser flow. Tracked as a follow-up issue.
+- `OtpEmailProvider` / `TotpProvider` verification paths.
+- Any change to challenge TTL handling, attempt limiting, or the signature-counter update.
+
+## Implementation Plan
+
+### Phase 1: Close the bypass in `PasskeyProvider`
+
+`verify()` currently reaches a `true` verdict on three routes: a real `verifyAuthenticationResponse`, a `credentialId` + `challenge` string compare after the assertion branch, and — when no verify context exists at all — a bare `credentialId` compare. Both compared values are public: `prepareChallenge()` returns them to the caller and `GET /api/security/mfa/methods` discloses `providerMetadata.credentialId`.
+
+1.1 Reduce `verifyPayloadSchema` to a single object that requires `response`, and reject a payload that fails the schema by returning `false` rather than throwing. A thrown `ZodError` reaches `mapMfaError` as a logged **500** and bypasses `registerFailedAttempt`, so rejected attempts would stop counting toward the challenge attempt limit; `safeParse` + `false` keeps the documented **401** and the lockout behavior.
+
+1.2 Delete the context-absent `credentialId` fallback and the trailing string-compare block, so a verified `verifyAuthenticationResponse` is the only route to `true`.
+
+### Phase 2: Unit regression coverage
+
+2.1 Replace `supports legacy verification payload for backward compatibility` (which asserted the bypass worked) with regression tests: the legacy shape is refused with a correctly prepared context; a `credentialId`-only payload is refused when the client skipped `/prepare`; a well-formed assertion is refused when no challenge was prepared; and an assertion the authenticator did not sign is refused. The first two also assert `verifyAuthenticationResponse` was never reached.
+
+2.2 Add `PasskeyProvider.webauthn.test.ts`, which does **not** mock `@simplewebauthn/server`: a software authenticator generates a P-256 key pair, stores the COSE public key the way `confirmSetup` does, and signs a real assertion over the prepared challenge. Asserts that a genuine assertion still verifies end to end and that a tampered signature does not. This is the evidence that the removal does not break real passkey login.
+
+### Phase 3: Realign the integration suite
+
+The Playwright specs used the bypass as their shortcut, so they assert a behavior that is now a vulnerability.
+
+3.1 Turn the shared fixture into an explicitly named negative helper and update `TC-SEC-004` to assert that login verification with a non-cryptographic payload is refused.
+
+3.2 Update `TC-SEC-006` and `TC-SEC-007` to assert that passkey sudo verification is refused, then skip their sudo-token-gated remainders with an explicit reason instead of asserting a token they can no longer obtain.
+
+### Phase 4: Contract documentation
+
+4.1 Record the removal in `BACKWARD_COMPATIBILITY.md` as a dated, deliberate exception to the deprecation protocol, and add an `UPGRADE_NOTES.md` entry telling downstream operators what changes and what to do about credentials enrolled through the setup shortcut.
+
+4.2 File the two follow-up issues (setup-side attestation shortcut; virtual-authenticator integration coverage) and link them from the PR.
+
+### Phase 5: Validation and delivery
+
+5.1 Run the full `validation.commands` gate, then finalize the PR: labels, review pass, summary comment.
+
+## Risks
+
+- **Highest risk is a false negative**: over-tightening `verify()` would lock every passkey user out of login and sudo. Mitigated by Phase 2.2, which exercises the untouched `{ response }` path against the real `@simplewebauthn/server` with a genuinely signed assertion.
+- **Deliberate breaking change.** `{ credentialId, challenge }` payloads start returning `401`. This skips the deprecation protocol on purpose: the request shape being removed *is* the vulnerability, so a bridge release would keep the second-factor bypass live. The first-party UI is unaffected — it sends `startAuthentication()` output.
+- **Stranded credentials.** A passkey enrolled through the setup shortcut carries a fabricated public key and can never produce a verifiable assertion. Such a user needs an admin MFA reset. Documented in `UPGRADE_NOTES.md`.
+- **Integration coverage loss.** Passkey verification has no API-level happy path after this change; Phase 3 makes that explicit rather than silent, and Phase 4.2 tracks restoring it.
+
+## Progress
+
+> Convention: `- [ ]` pending, `- [x]` done. Append ` — <commit sha>` when a step lands. Do not rename step titles.
+
+### Phase 1: Close the bypass in PasskeyProvider
+
+- [ ] 1.1 Require a `response` payload and reject unparseable payloads without throwing
+- [ ] 1.2 Delete the context-absent and string-compare acceptance branches
+
+### Phase 2: Unit regression coverage
+
+- [ ] 2.1 Replace the legacy-payload test with bypass regression tests
+- [ ] 2.2 Add a real-crypto test proving a genuine assertion still verifies
+
+### Phase 3: Realign the integration suite
+
+- [ ] 3.1 Rework the shared fixture and TC-SEC-004 into a negative assertion
+- [ ] 3.2 Rework TC-SEC-006 and TC-SEC-007 sudo verification
+
+### Phase 4: Contract documentation
+
+- [ ] 4.1 Document the deliberate break in BACKWARD_COMPATIBILITY.md and UPGRADE_NOTES.md
+- [ ] 4.2 File the setup-shortcut and integration-coverage follow-up issues
+
+### Phase 5: Validation and delivery
+
+- [ ] 5.1 Run the full validation gate and finalize the PR
--- a/.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
+++ b/.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
@@ -82,13 +82,13 @@ The Playwright specs used the bypass as their shortcut, so they assert a behavio
 
 ### Phase 3: Realign the integration suite
 
-- [ ] 3.1 Rework the shared fixture and TC-SEC-004 into a negative assertion
-- [ ] 3.2 Rework TC-SEC-006 and TC-SEC-007 sudo verification
+- [x] 3.1 Rework the shared fixture and TC-SEC-004 into a negative assertion — 5560dcaec
+- [x] 3.2 Rework TC-SEC-006 and TC-SEC-007 sudo verification — 5560dcaec
 
 ### Phase 4: Contract documentation
 
-- [ ] 4.1 Document the deliberate break in BACKWARD_COMPATIBILITY.md and UPGRADE_NOTES.md
-- [ ] 4.2 File the setup-shortcut and integration-coverage follow-up issues
+- [x] 4.1 Document the deliberate break in BACKWARD_COMPATIBILITY.md and UPGRADE_NOTES.md — b508f95e2
+- [x] 4.2 File the setup-shortcut and integration-coverage follow-up issues — #5296 already existed for the setup shortcut; filed #5307 for integration coverage
 
 ### Phase 5: Validation and delivery
 
--- a/.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
+++ b/.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md
@@ -72,13 +72,13 @@ The Playwright specs used the bypass as their shortcut, so they assert a behavio
 
 ### Phase 1: Close the bypass in PasskeyProvider
 
-- [ ] 1.1 Require a `response` payload and reject unparseable payloads without throwing
-- [ ] 1.2 Delete the context-absent and string-compare acceptance branches
+- [x] 1.1 Require a `response` payload and reject unparseable payloads without throwing — b635b1677
+- [x] 1.2 Delete the context-absent and string-compare acceptance branches — b635b1677
 
 ### Phase 2: Unit regression coverage
 
-- [ ] 2.1 Replace the legacy-payload test with bypass regression tests
-- [ ] 2.2 Add a real-crypto test proving a genuine assertion still verifies
+- [x] 2.1 Replace the legacy-payload test with bypass regression tests — 0ed09d4b7
+- [x] 2.2 Add a real-crypto test proving a genuine assertion still verifies — 0ed09d4b7
 
 ### Phase 3: Realign the integration suite
 
--- a/.ai/specs/enterprise/2026-08-14-passkey-mfa-require-webauthn-assertion.md
+++ b/.ai/specs/enterprise/2026-08-14-passkey-mfa-require-webauthn-assertion.md
@@ -1,0 +1,148 @@
+# Enterprise Security — Require a Genuine WebAuthn Assertion for Passkey MFA Verification
+
+> **Status:** Implemented 2026-08-14 (branch `fix/passkey-mfa-require-webauthn-assertion`, PR #5306) — awaiting merge and manual QA
+> **Issue:** [open-mercato#3852](https://github.com/open-mercato/open-mercato/issues/3852)
+> **Scope:** Enterprise — `security` module (`packages/enterprise/src/modules/security`)
+> **Severity:** Critical — second-factor bypass reachable by anyone holding an `mfa_pending` session, in login MFA and in passkey-as-sudo step-up alike.
+
+## TLDR
+
+**Key Points:**
+- `PasskeyProvider.verify()` accepted a `{ credentialId, challenge }` payload beside the genuine `{ response }` assertion and approved it by plain string comparison. Both compared values are **public**: `prepareChallenge()` returns the credential id and the challenge to the caller, and `GET /api/security/mfa/methods` discloses `providerMetadata.credentialId`.
+- A third route needed even less. With no verify context — the client simply never called `/api/security/mfa/prepare` — the provider fell through to comparing only the disclosed credential id and ignored the challenge entirely.
+- `SudoChallengeService.verify` → `MfaVerificationService.verifyChallenge` → `provider.verify` is the same code path, so the bypass defeated login-time MFA **and** passkey-as-sudo step-up.
+
+**Scope:**
+- Collapse `verifyPayloadSchema` from a union to a single object requiring `response`; delete the context-absent `credentialId` fallback and the trailing string-compare block, so a verified `verifyAuthenticationResponse` is the only route to a positive verdict.
+- Replace the test that asserted the bypass worked with regression coverage for every route that previously reached a positive verdict without a signature, plus a real-crypto suite that proves genuine passkeys still authenticate.
+- Record the contract break under the Emergency Security Exception, with client and operator migration instructions.
+
+**Concerns:**
+- This removes a STABLE contract surface with no bridge, which the ordinary deprecation protocol forbids. See § Migration & Backward Compatibility.
+- It closes only the *verification* half. The *enrollment* half (#5296) still accepts a client-supplied `publicKey` with no attestation and is unaffected by this change.
+
+---
+
+## Problem Statement
+
+All references verified against `develop` at issue-filing time.
+
+### The verification schema accepted a non-cryptographic shape
+
+`lib/providers/PasskeyProvider.ts` declared `verifyPayloadSchema` as a union of `{ response }` and `{ credentialId, challenge }`. The second branch was approved by string comparison against values the server itself discloses:
+
+- `prepareChallenge()` returns `options.challenge` to the caller in `clientData`.
+- `GET /api/security/mfa/methods` returns `providerMetadata.credentialId` for every enrolled method.
+
+An attacker therefore did not need to guess either value; the API hands both over.
+
+### An absent verify context weakened the check further
+
+`MfaVerificationService.verifyChallenge` builds the verify context from `challenge.providerChallenge`, which `prepareChallenge` persists. When the client never called `/prepare`, no context existed, and the provider's context-absent fallback compared **only** the disclosed credential id, ignoring the challenge. That is a bypass with a single public value.
+
+### Both entry points route through the same provider
+
+`services/MfaVerificationService.ts` is the only call site of `provider.verify` in the module, and `SudoChallengeService.verify` delegates to it. So the same defect served `POST /api/security/mfa/verify` (login second factor) and `POST /api/security/sudo/verify` (sudo step-up).
+
+### Root cause
+
+The provider treated a *disclosed identifier* as an *authentication secret*. Possession of a credential id and a challenge proves nothing; only a signature over the challenge by the credential's private key does.
+
+## Proposed Solution
+
+Make a verified assertion the sole positive path.
+
+### A. Narrow the payload schema
+
+`verifyPayloadSchema` becomes `z.object({ response: z.record(z.string(), z.unknown()) })`. The union's second branch is removed, so a legacy payload fails the schema before any comparison happens.
+
+### B. Require a prepared verify context
+
+`verifyContextSchema.safeParse(context)` must succeed. An absent or malformed context is a rejection, not a fallback. The existing challenge TTL check runs unchanged against `pending.createdAt`.
+
+### C. Delete the string-compare block
+
+The trailing `credentialId` / `challenge` comparison is removed outright. `verifyAuthenticationResponse` from `@simplewebauthn/server` is the only remaining route to `true`, and the signature-counter update stays gated behind its positive verdict.
+
+### D. Fail closed, and fail as a rejection rather than an error
+
+`safeParse` instead of `parse`, and a `.catch()` around `verifyAuthenticationResponse` that returns `null`. This is deliberate and is the security-relevant choice: a thrown error reaches `mapMfaError` as a logged **500** and skips `registerFailedAttempt`, so rejected attempts would stop counting toward `securityConfig.mfa.maxAttempts` and the challenge lockout would be defeated. Returning `false` preserves the documented **401** and the attempt limit. No verdict flips from negative to positive.
+
+The `.catch()` emits a `warn` through the logging facade (`createLogger('security').child({ component: 'passkey-provider' })`) carrying the method id, the rpId and the verifier's error — but not the assertion — so a misconfigured `rpId`/`expectedOrigins` is diagnosable rather than presenting as a mystery 401.
+
+## Affected Surfaces
+
+| Path | Change |
+|------|--------|
+| `lib/providers/PasskeyProvider.ts` | `verifyPayloadSchema` narrowed; context-absent fallback and string-compare block removed; verifier failure logged and rejected |
+| `lib/__tests__/PasskeyProvider.test.ts` | Bypass-asserting test replaced with regression coverage for each previously-positive route |
+| `lib/__tests__/PasskeyProviderWebAuthn.test.ts` (new) | Real `@simplewebauthn` verifier against a software authenticator |
+| `__integration__/TC-SEC-004.spec.ts` | Asserts login verification answers `401` without an assertion |
+| `__integration__/TC-SEC-006.spec.ts`, `TC-SEC-007.spec.ts` | Sudo step-up asserts `401`; sudo-token-gated remainders skipped with a stated reason (#5307) |
+| `__integration__/helpers/securityFixtures.ts` | `verifyPasskeyChallenge` → the explicitly negative `attemptUnsignedPasskeyVerify` |
+| `BACKWARD_COMPATIBILITY.md`, `UPGRADE_NOTES.md`, `.ai/qa/scenarios/TC-SEC-004-…md` | Contract break, exception record, client and operator migration |
+
+## Test Plan
+
+The load-bearing evidence is `PasskeyProviderWebAuthn.test.ts`, which does **not** mock `@simplewebauthn/server`. It generates a P-256 key pair, encodes the COSE public key the way `confirmSetup` stores it, and signs a genuine assertion over the challenge `prepareChallenge()` produced. Mocking the verifier could not prove what this change most needs proven — that the tightening did not overshoot and lock every passkey user out.
+
+- A real assertion verifies and advances the stored signature counter.
+- A tampered signature is refused and the counter is not advanced.
+- An assertion replayed against a different prepared challenge is refused.
+- The disclosed `{ credentialId, challenge }` pair is refused.
+
+`PasskeyProvider.test.ts` adds five cases against the mocked verifier: the legacy shape is refused even with a correctly prepared context; a `credentialId`-only payload is refused when the client skipped `/prepare`; a well-formed assertion is refused when no challenge was prepared; an assertion the authenticator did not sign is refused without advancing the counter; and an assertion that makes the verifier throw resolves to `false` instead of propagating. The first three also assert `verifyAuthenticationResponse` was never reached. Three of the five fail against the pre-fix provider.
+
+`TC-SEC-004` gives the fix live end-to-end coverage in CI. `TC-SEC-006`/`TC-SEC-007` gate on a pre-existing `test.skip(!passkeyMethod, …)` and the CI environment seeds no admin passkey, so their new `401` assertions are dark in CI today; the fix's CI signal comes from `TC-SEC-004` plus the unit suites, which exercise the identical provider path sudo routes through.
+
+**Coverage gap, tracked as #5307.** Passkey verification has no integration-level *happy* path, because no API-level fixture can produce a signed assertion. Restoring it needs a Playwright virtual authenticator, or decoupling the sudo scenarios from passkeys onto TOTP.
+
+## Migration & Backward Compatibility
+
+This change **breaks a STABLE contract surface without the deprecation protocol, deliberately**, under the [Emergency Security Exception](../../../BACKWARD_COMPATIBILITY.md#emergency-security-exception).
+
+### Classification
+
+The removed shape was publicly supported, not an undocumented accident: it was part of the exported `MfaProviderInterface.verifySchema` union that a third-party client could validate against, and the enterprise suite pinned it in a case named *"supports legacy verification payload for backward compatibility"*. So this is a real category-7 (API request shape) break, and the exception — rather than a claim that no contract existed — is what authorizes it.
+
+### Why no bridge
+
+The protocol's bridge requirement exists to give downstream authors a migration window. Here the request shape being removed *is* the vulnerability: both values it compared are disclosed by the server, so keeping it alongside the assertion path for one minor version would leave the passkey second factor bypassable in login MFA and sudo step-up for that whole window. A security fix that keeps the hole open for a release is not a fix. Correspondingly, **no flag, config toggle or opt-in retains the old branch** — a retained branch would be exactly the bridge the exception refuses.
+
+### Broken surfaces
+
+| Surface | Change |
+|---------|--------|
+| `POST /api/security/mfa/verify`, `POST /api/security/sudo/verify` request shape (`methodType: 'passkey'`) | `{ credentialId, challenge }` now answers `401` |
+| `MfaProviderInterface.verifySchema` for `passkey` | Narrows from a union to a single object requiring `response` |
+| Verification with no prepared challenge | Now a rejection rather than a `credentialId` comparison |
+| Route URLs, HTTP methods, response schemas, DB schema, event IDs, ACL features, DI names, CLI commands | Unchanged |
+
+### Client migration
+
+Send the object returned by `@simplewebauthn/browser`'s `startAuthentication()` as `payload.response`, after calling `/api/security/mfa/prepare` (or `/api/security/sudo/prepare`). The first-party `PasskeyChallengeVerify` component already does this, so shipped UIs need no change. A client that submitted the credential id and challenge was, by construction, not performing cryptographic verification.
+
+### Operator migration
+
+Full instructions are in [`UPGRADE_NOTES.md`](../../../UPGRADE_NOTES.md) under `0.6.7 → 0.7.0`. The essential point, and the one easy to get wrong: a passkey enrolled through the setup path's unattested `publicKey` shortcut is **not** reliably rendered unusable by this change. Such a row holds either a key nobody can sign with (that user is now locked out and needs an admin MFA reset) **or** a keypair the enroller controls, which signs assertions this change accepts. Both enrollment paths write identical `provider_metadata` keys, so the row cannot be attributed to an origin; the conservative remediation is to reset and re-enroll all passkey methods on any deployment that ever provisioned passkeys through the API, and to sequence #5296 ahead of the upgrade where possible.
+
+## Risks & Impact Review
+
+| Risk | Failure scenario | Severity | Mitigation | Residual |
+|------|------------------|----------|------------|----------|
+| Tightening overshoots and refuses genuine assertions | Every passkey user locked out of login and sudo | Critical | `PasskeyProviderWebAuthn.test.ts` proves a real signed assertion still verifies and advances the counter; manual QA against a real authenticator before merge | Low |
+| Third-party client sends the legacy shape | That client's passkey step breaks at upgrade with `401` | Medium | Documented in `UPGRADE_NOTES.md` and `BACKWARD_COMPATIBILITY.md`; accepted under the exception | Accepted |
+| Shortcut-enrolled credential holds an attacker-controlled valid keypair | Account takeover survives this fix | High | Disclosed in `UPGRADE_NOTES.md` with reset-and-re-enroll remediation; closed properly by #5296 | Medium until #5296 lands |
+| Verifier error swallowed as a silent 401 | Misconfigured `rpId`/`expectedOrigins` reads as a user error | Medium | `.catch()` logs a `warn` with method id, rpId and the verifier error | Low |
+| No integration happy path for passkey verification | A future regression in the `{ response }` path is caught only by unit tests | Medium | Tracked as #5307; unit suite runs the real verifier | Medium |
+
+## Open Questions
+
+1. ~~Throw or return `false` on a malformed payload / verifier error.~~ **Resolved (2026-08-14):** return `false`. Throwing produces a 500 that skips `registerFailedAttempt` and defeats the challenge attempt limit. See § D.
+2. ~~Bridge the legacy shape for one minor version.~~ **Resolved (2026-08-14):** no bridge, under the Emergency Security Exception. See § Migration & Backward Compatibility.
+
+_No open questions remain blocking._
+
+## Changelog
+
+- 2026-08-14 — Initial spec, written from issue #3852 and the implementation on `fix/passkey-mfa-require-webauthn-assertion`, to satisfy deprecation-protocol step 5 for the contract break. Records the Emergency Security Exception classification, the client and operator migration paths, and the #5296 / #5307 boundaries.
--- a/BACKWARD_COMPATIBILITY.md
+++ b/BACKWARD_COMPATIBILITY.md
@@ -333,3 +333,21 @@ Files in `apps/mercato/.mercato/generated/` are produced by the CLI generators. 
 | Database schema, event IDs, ACL features, DI names, CLI commands | No change | ✓ n/a |
 
 **Migration path for existing modules**: no action required. The capability is opt-in per rejection — an interceptor that never sets `status` produces exactly the responses it produced before. Interceptors that want a deliberate status add `status` (and optionally `body`) to the `{ ok: false, message }` verdict they already return. Third-party transports that call `commandBus.execute` inside their own `try/catch` can honour the same contract in two lines via `getCommandInterceptorHttpRejection(err)`, which validates the status is an integer in 400-599 before returning it.
+
+---
+
+## Passkey MFA Verification Payload (2026-08-14)
+
+Issue #3852 removed the non-cryptographic passkey verification shape from `PasskeyProvider` in the enterprise `security` module. This is a **deliberate breaking change to a STABLE contract surface (category 7, API request shapes) that does not follow the deprecation protocol**, and it is recorded here as the exception rather than the rule.
+
+| Surface | Change | Classification |
+|---------|--------|----------------|
+| API request shape (`POST /api/security/mfa/verify`, `POST /api/security/sudo/verify`, `methodType: 'passkey'`) | `payload` must be `{ response }` carrying a WebAuthn assertion. The `{ credentialId, challenge }` alternative is removed and now answers `401` | ✗ BREAKING (deliberate — see rationale) |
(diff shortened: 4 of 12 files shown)


pkarw:
🤖 `om-auto-create-pr` — 🏷️ label rationale

- 🔍 `review` — the implementation, the validation gate, and the documentation are complete, so the PR is ready for code review.
- 🐛 `bug` — this closes a defect reported in #3852 rather than adding capability.
- 🔒 `security` — the defect is an authentication bypass: the passkey second factor could be passed without an authenticator, in both login MFA and sudo step-up.
- 🏢 `enterprise` — every source change lands in `packages/enterprise/src/modules/security`, the commercial security module.
- 🧪 `needs-qa` — the change sits directly in the login MFA and sudo step-up flows, and the passing passkey ceremony cannot be automated without a virtual authenticator (#5307), so a human must confirm that a real passkey login and a real passkey sudo step-up still succeed against a genuine authenticator.
- 🔺 `priority-high` — a live second-factor bypass in shipped code should block the next release.
- ⚠️ `risk-high` — this is authentication code, and the failure mode of over-tightening is locking every passkey user out of login and sudo; the new real-crypto test exists specifically to bound that risk.

pkarw:
## 🤖 `om-auto-create-pr` — run summary

**Tracking plan:** `.ai/runs/2026-08-14-passkey-mfa-require-webauthn-assertion.md`
**Branch:** `fix/passkey-mfa-require-webauthn-assertion`
**Final status:** complete

### 📋 Summary of changes

- **Phase 1 — closed the bypass** (`b635b1677`). `verifyPayloadSchema` in `PasskeyProvider.ts` collapses from a union to a single object requiring `response`; the context-absent `credentialId` fallback and the trailing `credentialId` + `challenge` string-compare block are deleted. A verified `verifyAuthenticationResponse` is now the only route to a positive verdict, in login MFA and in passkey-as-sudo step-up alike. Schema and verifier failures return `false` rather than throwing, so they answer the documented `401` and count toward the challenge attempt limit instead of logging a `500` that skipped `registerFailedAttempt`.
- **Phase 2 — regression and real-crypto coverage** (`0ed09d4b7`). The test that asserted the bypass worked is replaced by five cases covering every route that used to reach a positive verdict without a signature, plus a new `PasskeyProviderWebAuthn.test.ts` that runs the **unmocked** `@simplewebauthn` verifier against a software authenticator.
- **Phase 3 — realigned the integration suite** (`5560dcaec`). The Playwright specs used the removed payload as their QA shortcut, so they asserted a behavior that is now a vulnerability. `verifyPasskeyChallenge` becomes the explicitly negative `attemptUnsignedPasskeyVerify`; `TC-SEC-004` asserts login verification answers `401`; `TC-SEC-006`/`TC-SEC-007` assert passkey sudo step-up answers `401` and skip their sudo-token-gated remainders with a stated reason.
- **Phase 4 — contract documentation** (`b508f95e2`). `BACKWARD_COMPATIBILITY.md` records the deliberate exception to the deprecation protocol with its rationale; `UPGRADE_NOTES.md` tells client authors what to send and operators which enrolled credentials stop working; the TC-SEC-004 QA scenario records that the passing ceremony stays manual.
- **Phase 5 — review fix** (`fb98bf0d9`). See the review pass below.
- **Files touched:** `PasskeyProvider.ts`, two unit test files, four integration files, `BACKWARD_COMPATIBILITY.md`, `UPGRADE_NOTES.md`, one QA scenario doc, and the tracking plan. No migration, no schema change, no route change.

**Why this PR rather than #5291:** that PR proposed the same remediation as an outside contribution against `packages/enterprise`, our commercial package, which does not accept third-party contributions. It was closed unmerged; this fix is re-authored first-party from the issue report and the current `develop` source, and carries none of that diff.

None — no `--skill-url` was passed.

### 🧪 Verification phases completed

- **Targeted validation (per phase):** after Phase 1 and 2, `jest` on `PasskeyProvider.test.ts` and `PasskeyProviderWebAuthn.test.ts`; the new regression cases were additionally run against the **pre-fix** provider (old file checked back out) to prove three of the five fail without the change. After Phase 3, `tsc --noEmit` on `packages/enterprise` and a grep sweep confirming the renamed fixture has no remaining callers. After Phase 5, the full `src/modules/security` suite — 34 suites, 277 tests, all green.
- **Full validation gate:** `yarn build:packages` ✅ · `yarn generate` ✅ (no generated drift) · `yarn build:packages` ✅ · `yarn i18n:check-sync` ✅ · `yarn i18n:check-usage` ✅ · `yarn typecheck` ✅ 22/22 · `yarn test` ⚠️ · `yarn build:app` ✅. `@open-mercato/enterprise` is 60/60 suites and 486/486 tests green. `yarn test` exits non-zero on exactly one package, `create-mercato-app`, on the assertion `❌ Linux file-watch limits are too low for Turbopack` — this host's `fs.inotify.max_user_instances` is 128, and `packages/create-app` appears nowhere in this branch's diff, so the failure cannot originate here. Linux CI is the authority for it. `yarn lint`, which is not in the configured gate, also passes with 0 errors (10 pre-existing warnings in `@open-mercato/app`).
- **Review/autofix pass:** ran the `om-code-review` engine against the branch diff. Verdict **approve**, with one **major** finding raised and fixed rather than waived: the new `catch` around `verifyAuthenticationResponse` swallowed the verifier's error, and because that function *throws* (rather than returning `verified: false`) on challenge, origin, and RP-ID mismatches, a deployment with a misconfigured `expectedOrigins` or `rpId` would have answered a silent `401` with nothing in the server log — where before this change the throw reached `mapMfaError`, which logs. Fixed in `fb98bf0d9`: the rejection semantics are unchanged, but the catch now emits a `warn` through the structured logging facade naming the method id and the resolved rpId. The compatibility check flagged the deliberate request-shape break, which is documented; the security, scope, and data-scoping checks returned clean. One minor (unreachable assertions after the mid-test `test.skip` in TC-SEC-006/007, kept deliberately so #5307 can restore them) and one nit (the hand-rolled software authenticator should move to a shared test util if a second suite ever needs it) were accepted rather than fixed.

### 🔍 How to verify

- **Manual smoke test:** with a real authenticator, enroll a passkey at `/backend/profile/security/mfa/passkey`, sign out, log in again, and complete the passkey MFA challenge — it must still succeed and land on `/backend`. Then trigger a sudo-gated action (for example an admin MFA reset) and complete the passkey step-up — it must still issue a sudo token. Finally, confirm the fix: replay the disclosed values by calling `POST /api/security/mfa/prepare` for a passkey challenge and then `POST /api/security/mfa/verify` with `payload: { credentialId, challenge }` taken from the prepare response and from `GET /api/security/mfa/methods`. Before this change that returned `200` with a session token; it must now return `401`.
- **Areas to spot-check in the diff:** `PasskeyProvider.verify()` end to end — in particular that the metadata reads, the TTL check, and the counter update are byte-equivalent to the old `{ response }` branch and only the surrounding branching changed. Also the `credentialPublicKey ?? publicKey` fallback, which is retained deliberately so credentials stored under either key keep working.
- **Commands the reviewer can re-run:** `yarn workspace @open-mercato/enterprise jest --config jest.config.cjs src/modules/security`, then the full gate: `yarn build:packages && yarn generate && yarn build:packages && yarn i18n:check-sync && yarn i18n:check-usage && yarn typecheck && yarn test && yarn build:app`.
- **Rollback plan:** `git revert` the range `b635b1677..fb98bf0d9` (or revert the squash-merge commit). There is no migration, no schema change, and no persisted state to unwind, so a revert restores the previous behavior exactly — including the bypass, which is why a rollback should be paired with taking the passkey provider out of the allowed-methods policy rather than left in place.

- **Most likely regression:** over-tightening `verify()` so that *genuine* assertions stop verifying, which would lock every passkey user out of both login and sudo step-up. This is the reason `PasskeyProviderWebAuthn.test.ts` exists: it signs a real assertion with a P-256 key over the challenge `prepareChallenge()` produced, and asserts both that it verifies and that the stored signature counter advances. A tightening that broke real passkeys would fail that test.
- **Second-order effects:** `MfaVerificationService.verifyChallenge` and `SudoChallengeService.verify` both route through `provider.verify`, so any client that relied on the removed shape now sees a counted failed attempt and, after `mfa.maxAttempts`, a locked-out challenge rather than a one-off error. That is the intended behavior, but it means a broken integration will surface as lockouts rather than as a single visible error.
- **Security-sensitive surfaces:** authentication and second-factor verification, plus sudo step-up. No data-scoping, permission, or secrets surfaces were touched. The new `warn` log carries only the method id and the configured rpId — no payload, no credential material.
- **Breaking-change impact:** `POST /api/security/mfa/verify` and `POST /api/security/sudo/verify` no longer accept a `{ credentialId, challenge }` passkey payload. This is a documented, deliberate break of a STABLE contract surface with no deprecation window, because the shape being removed *is* the vulnerability. The shipped UI is unaffected — it sends `startAuthentication()` output.
- **Residual risk accepted:** **(1)** This closes one of two halves. #5296 tracks the enrollment side, which still accepts a client-supplied `publicKey` with no attestation; until that lands, an attacker who can enroll can still supply their own public key and sign genuinely. **(2)** Passkey verification now has no integration-level happy path, and TC-SEC-006/007 lose their sudo-token-gated assertions — tracked in #5307, which needs a Playwright virtual authenticator or a TOTP-driven sudo path. **(3)** A passkey enrolled through the setup shortcut carries a fabricated public key and can no longer authenticate; such a user needs an admin MFA reset, documented in `UPGRADE_NOTES.md`. **(4)** Manual QA is still pending — the passing passkey ceremony cannot be automated today, so a human must confirm real passkey login and real passkey sudo step-up before merge.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-14T14:59:32Z. Other auto-skills will skip this PR until the lock is released.

pkarw (commented):
# 🔍 Code Review

**PR:** #5306 — security(security): require a genuine WebAuthn assertion for passkey MFA verification
**Head:** `93c93e87f` · **Base:** `develop` · **Mode:** first review (no prior review from this reviewer)
**Runner:** GitHub checks (no local gate run — see 🧪 Validation Gate)

## 🎯 Summary

This PR removes the non-cryptographic passkey verification path from `PasskeyProvider.verify()`. Before the change, `verifyPayloadSchema` was a union that also accepted `{ credentialId, challenge }`, and the provider approved that shape by plain string comparison against values the server itself discloses — `prepareChallenge()` returns the challenge to the caller, and `GET /api/security/mfa/methods` returns `providerMetadata.credentialId`. Worse, when no verify context existed (the client simply never called `/api/security/mfa/prepare`), the provider fell through to comparing only the disclosed credential id and ignored the challenge entirely. Anyone holding an `mfa_pending` session therefore passed the passkey second factor with no authenticator, no private key, and no signature — in login-time MFA and in passkey-as-sudo step-up alike, since `SudoChallengeService.verify` routes through the same `MfaVerificationService.verifyChallenge` → `provider.verify` path.

The fix is the right shape and is minimal. `verifyPayloadSchema` collapses to a single object requiring `response`; the context-absent fallback and the trailing string-compare block are deleted; a verified `verifyAuthenticationResponse` is now the only route to a positive verdict. I traced every caller: `packages/enterprise/src/modules/security/services/MfaVerificationService.ts:189` is the **only** call site of `provider.verify` in the module, and it builds the context from `challenge.providerChallenge`, which `prepareChallenge` persists at `MfaVerificationService.ts:132`. Both first-party entry points reach that persisted context — `MfaChallengePanel.tsx:95` posts to `/api/security/mfa/prepare` and `SudoChallengeModal.tsx:78` posts to `/api/security/sudo/prepare` (which delegates to the same `prepareChallenge` at `SudoChallengeService.ts:230`) before `PasskeyChallengeVerify.tsx:51` submits `startAuthentication()` output as `{ response }`. So the tightening does not break any shipped client path.

Two secondary behaviors are worth confirming as deliberate, and both are correct:

- **`safeParse` instead of `parse`.** A malformed payload now returns `false` rather than throwing. This matters: a thrown error reaches `mapMfaError` as a logged **500** and skips `registerFailedAttempt`, so rejected attempts would stop counting toward `securityConfig.mfa.maxAttempts`. Returning `false` keeps the documented **401** from `api/mfa/verify/route.ts:47` and preserves lockout. No verdict flips from negative to positive.
- **The `.catch()` around `verifyAuthenticationResponse`** returns `null` and emits a `warn` through the logging facade (`PasskeyProvider.ts:355`) rather than swallowing silently — so a misconfigured `rpId`/`expectedOrigins` is diagnosable instead of presenting as a mystery 401. The logger construction matches the house convention exactly (`createLogger('security').child({ component: 'passkey-provider' })`, cf. `packages/enterprise/src/modules/sso/services/ssoService.ts:15`). Nothing secret is logged: the method id, the rpId, and the verifier's own error, with the assertion itself excluded.

## ✅ Verdict

**Approved.** No blockers. One major — the deliberate contract break — carries an explicit, written waiver in `BACKWARD_COMPATIBILITY.md` and `UPGRADE_NOTES.md`, which under the review rule is a waived major rather than a request for changes. The remaining findings are nits.

## 🧪 Validation Gate

**Validation source: GitHub checks.** Every check on head `93c93e87f` is green, so no local fallback gate was run (this repo's `om-auto-review-pr` extension prefers CI results over re-running `validation.commands` locally).

| Check | Result | Evidence |
|-------|--------|----------|
| `test` | ✅ pass (6m59s) | `PASS src/modules/security/lib/__tests__/PasskeyProvider.test.ts` and `PASS src/modules/security/lib/__tests__/PasskeyProviderWebAuthn.test.ts` both confirmed in the job log |
| `ephemeral-integration` | ✅ pass (8m22s) | 8 passed, 3 skipped; `TC-SEC-004 … refuses login verification without an assertion` ✓ in 1.2s |
| `lint` | ✅ pass | — |
| `docker-build` | ✅ pass (4m27s) | — |
| `ds-lint`, `audit-scope`, `scope`, `prepare` | ✅ pass | — |
| `CodeQL` / `Analyze (actions, javascript-typescript, python)` | ✅ pass | — |
| `license/cla` | ✅ pass | CLA signed |

No required check was pending or failing at review time, and the head has no merge conflict (`mergeable: MERGEABLE`). `mergeStateStatus: BLOCKED` reflects the outstanding review plus the QA-approval gate, not a broken branch.

The PR body's note that `yarn test` failed locally on `create-mercato-app` with `❌ Linux file-watch limits are too low for Turbopack` is a host-environment artifact (`fs.inotify.max_user_instances`), not a branch defect — that package is not in the diff, and CI's `test` job is green on the same commit. Correctly diagnosed.

## 🧪 Test Coverage

This is the strongest part of the PR, and it is unusually well done for a security fix.

`PasskeyProviderWebAuthn.test.ts` (new, 161 lines) deliberately does **not** mock `@simplewebauthn/server`. It generates a real P-256 key pair, encodes the COSE public key the way `confirmSetup` stores it, builds authenticator data with the correct `rpIdHash`/flags/counter layout, and signs a genuine ES256 assertion over the challenge `prepareChallenge()` actually produced. It then asserts the four properties that matter: a real assertion still verifies and advances the stored counter (the regression that would have locked out every passkey user if the tightening overshot), a tampered signature is refused without advancing the counter, an assertion replayed against a different prepared challenge is refused, and the disclosed `{ credentialId, challenge }` pair is refused. Proving the fix does not break real passkeys is exactly the coverage a change like this needs, and mocking the verifier could not have proven it.

`PasskeyProvider.test.ts` replaces the test literally named *"supports legacy verification payload for backward compatibility"* — which asserted that the bypass worked — with five cases against the mocked verifier covering each route that previously reached a positive verdict, three of them additionally asserting `verifyAuthenticationResponse` was never reached. `jest.resetAllMocks()` in `beforeEach` (line 25) means the `mockResolvedValue`/`mockRejectedValue` calls in the last two cases do not leak into neighbours, so there is no test-order coupling.

The claim in the PR body that three of the five fail against the pre-fix provider is consistent with the diff, and `TC-SEC-004` gives the fix live end-to-end coverage in CI.

## 💥 Breaking Changes

⚠️ **WARNING — this PR breaks a protected contract surface without the deprecation protocol, deliberately.** Per `BACKWARD_COMPATIBILITY.md`, an API request shape is contract surface (category 7) and the protocol requires a bridge for ≥1 minor version. That is skipped here.

| Surface | Assessment |
|---------|-----------|
| `POST /api/security/mfa/verify`, `POST /api/security/sudo/verify` request shape (`methodType: 'passkey'`) | ✗ Breaking — `{ credentialId, challenge }` now answers 401. **Waived, documented.** |
| `MfaProviderInterface.verifySchema` for `passkey` (exposed at `lib/mfa-provider-interface.ts:109`) | ✗ Breaking for a consumer validating against the exported schema. **Waived, documented.** |
| Verification behavior with no prepared challenge | ✗ Breaking for a client that skipped `/prepare`. **Waived, documented.** |
| Route URLs, HTTP methods, response schemas, DB schema, event IDs, ACL features, DI names, CLI commands | ✓ Unchanged |

I accept the waiver, and I want to be explicit about why rather than waving it through: the deprecation protocol exists to give downstream authors a bridge release, but here the request shape being removed **is** the vulnerability. A bridge would keep the passkey second factor bypassable for a full minor version in both login MFA and sudo step-up. A security fix that leaves the hole open for a release is not a fix. The exception is recorded in `BACKWARD_COMPATIBILITY.md` with that rationale, the migration path is spelled out in both files, and the operator-facing lockout case (credentials enrolled through the setup path's unattested `publicKey` shortcut can never produce a verifiable assertion, so those users need an admin MFA reset) is documented in `UPGRADE_NOTES.md` under the correct unreleased `0.6.7 → 0.7.0` section — `package.json` is still at `0.6.7`, so that placement is right.

I verified the documented failure code is accurate rather than assumed: `api/mfa/verify/route.ts` does **not** validate the payload against `provider.verifySchema`, it passes `payload` straight to `verifyChallenge` and returns 401 when the result is falsy (line 47). So the legacy shape genuinely answers 401, not 400, as both documents claim.

**This waiver is the one item a human should consciously sign off on.** Everything else in this review is routine.

## 📄 Findings

### 🔹 Low / nit

**1. `test.skip(true, …)` mid-body leaves ~60 lines of unreachable assertions in `TC-SEC-006` and `TC-SEC-007`.**
`packages/enterprise/src/modules/security/__integration__/TC-SEC-006.spec.ts:133` and `TC-SEC-007.spec.ts:175`

The unconditional skip is placed after the new `expect(…status).toBe(401)` assertions, so everything below it — the sudo config disable/re-enable coverage in TC-SEC-006, and the sudo-gated admin MFA reset plus post-reset status reporting in TC-SEC-007 — becomes permanently unreachable while remaining in the file. The reason string is clear and #5307 tracks restoration, so this is a defensible choice (the dead body doubles as the blueprint for what a virtual authenticator has to restore) rather than something I would block on. Two things are worth knowing:

- **The new 401 assertions get no CI signal.** Both tests already gate on a pre-existing `test.skip(!passkeyMethod, …)` (`TC-SEC-006.spec.ts:47`, `TC-SEC-007.spec.ts:125`), and the CI environment seeds no admin passkey — both were reported skipped in this run's `ephemeral-integration` job, so execution never reaches the sudo assertions. This is not a regression introduced here (they were dark before this PR too), and the fix is genuinely covered in CI by `TC-SEC-004` plus the unit suites, which exercise the identical provider path sudo routes through. But the PR body reads as though TC-SEC-006/007 assert the 401 in CI, and they do not.
- A skipped test that can still fail (an assertion above the skip throwing) is an unusual state for a reader to reason about. A one-line comment at the top of each `test()` saying the body stops at the skip would remove the surprise.

**2. `UPGRADE_NOTES.md` tells operators to audit `mfa_methods` passkey rows but gives them no way to find the affected ones.**
`UPGRADE_NOTES.md:33`

The lockout case is real and correctly disclosed — a passkey enrolled through the unattested `publicKey` shortcut can never authenticate again, and a user whose only method is such a credential is locked out until an admin resets. "Audit `mfa_methods` rows of type `passkey` before upgrading" is the right instruction, but an operator cannot act on it without knowing what distinguishes a fabricated key from a real one. A single SQL snippet, or a note that a fabricated key decodes as a non-COSE blob under `base64UrlToBytes`, would turn the warning into something executable. Worth a follow-up rather than a change here.

## ⚠️ Scope note — not a finding against this diff

While tracing reachability I looked at how #5296 (enrollment still accepts a client-supplied `publicKey` with no attestation) interacts with this fix, and there is something worth adding to that issue rather than to this PR.

`resolveMfaRequestContext` (`api/mfa/_shared.ts:41`) authenticates on `auth.sub` alone and does **not** reject `mfa_pending` tokens — which is why `api/mfa/verify/route.ts:27` has to check `context.auth.mfa_pending !== true` explicitly for itself. The passkey enrollment route (`api/mfa/provider/[providername]/route.ts:33`) declares only `requireAuth: true` and gates on `authorizeMfaEnrollmentMutation`, which has no equivalent pending-session check. If that authorization path admits a caller holding only an `mfa_pending` token, then #5296 is not merely "an authenticated user can forge their own key" — it is the same second-factor bypass this PR closes, reachable through a different door, and it would deserve the same priority as #3852.

I have not proven that reachability (it depends on how `MFA_MANAGE_FEATURE` is granted and on what `checkUserCompliance` returns for a user who already has an enrolled method), so I am not raising it as a finding here. It is a concrete question worth answering on #5296 before that issue is scheduled. This PR is a strict improvement regardless and should not wait on it.

## 📌 Post-merge

Manual QA remains genuinely required and cannot be shortcut: the passing passkey ceremony has no automated happy path (#5307), so a human needs to confirm that a real passkey login and a real passkey sudo step-up still succeed against a genuine authenticator. Separate QA instructions follow in their own comment.

pkarw:

- 🚀 `merge-queue` — the code review passed with no blockers and no un-waived majors, so the PR leaves `review`; `needs-qa` stays on, so the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — the change sits in the login MFA and sudo step-up flows and the passing passkey ceremony has no automated happy path (#5307), so a human must confirm a real passkey login and a real passkey sudo step-up against a genuine authenticator.
- 🐛 `bug` — this closes the authentication defect reported in #3852 rather than adding capability.
- 🔒 `security` — the defect was a second-factor bypass: the passkey factor could be passed with no authenticator, no private key, and no signature, in both login MFA and sudo step-up.
- 🏢 `enterprise` — the entire change lives in `packages/enterprise/src/modules/security`.
- 🔺 `priority-high` — kept as set; a second-factor bypass reachable by anyone holding an `mfa_pending` session is release-blocking security hardening, which is exactly the `priority-high` band.
- ⚠️ `risk-high` — kept as set; the diff touches authentication and deliberately breaks a protected contract surface without a deprecation window, so the blast radius covers every passkey user and any third-party client of the verify endpoints.

pkarw:

This PR is approved on code review and requires manual QA (`needs-qa`, no `skip-qa`). It is queued in `merge-queue` but the QA-approval gate holds it until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` by swapping the labels (remove `merge-queue`, add `qa`), then run the routes below.

**Why this one genuinely needs hands.** The change makes a cryptographically verified WebAuthn assertion the only way to pass the passkey second factor. The negative path is covered automatically (`TC-SEC-004` runs green in CI, and the unit suite signs real ES256 assertions against the unmocked verifier). The **positive** path — a real authenticator producing a real assertion — has no automated coverage at all and cannot get any until a Playwright virtual authenticator lands (#5307). If this change overshot, every passkey user is locked out of login, and nothing in CI would say so. That is the specific risk QA is here to retire.

You need a machine with a real platform authenticator (Touch ID, Windows Hello, or a hardware security key) and a browser that supports WebAuthn on the host the app is served from.

### P0 — Passkey login (second factor)

**Where to click**
- `/login` — sign in with a user whose enrolled MFA method is a passkey
- The passkey challenge step that follows the password (the `PasskeyChallengeVerify` panel)

**What to verify**
- Enroll a fresh passkey first, through the UI, at the security section of the profile → the browser's own passkey ceremony must appear and complete, and the method must show as enrolled.
- Sign out, sign back in with password → the passkey challenge appears → complete the browser ceremony with the real authenticator → **login succeeds and lands on `/backend`**. This is the single most important check in this PR.
- Repeat the login a second time with the same credential → it must still succeed. (The provider advances a stored signature counter on each successful verification; a bug there would show up as the second login failing while the first worked.)
- Cancel or dismiss the browser passkey prompt → the session stays MFA-pending and you are not let through.
- Let the challenge sit past its TTL, then complete the ceremony → it is refused rather than accepted.

**What can go wrong**
- A valid passkey login is refused with a generic "invalid code" 401 — the regression this PR most plausibly introduces. Check the server log for `Passkey assertion could not be verified`; the change adds that warn line precisely so this case is diagnosable rather than silent.
- The second login attempt fails after the first succeeded (signature-counter handling).
- The failure is silent in the logs — if a passkey login fails with nothing logged, that is itself a finding.
- Repeated failed attempts must still count toward the challenge attempt limit and lock the challenge out; the fix deliberately returns a 401 rather than a 500 so that lockout keeps working. Try three or four bad attempts and confirm the challenge stops accepting them.

### P0 — Passkey sudo step-up

**Where to click**
- Any sudo-protected admin action — e.g. the admin MFA reset for another user, under the security admin screens — while signed in as an admin whose step-up method is a passkey (the `SudoChallengeModal`)

**What to verify**
- Trigger the protected action → the sudo modal appears and offers the passkey method → complete the real browser ceremony → **a sudo token is issued and the protected action goes through**.
- Cancel the ceremony → no sudo token, and the protected action stays blocked with `sudo_required`.

**What can go wrong**
- Passkey step-up is refused for a genuine authenticator, blocking admins from sudo-protected operations entirely. This path shares its verification code with login, so it should pass or fail together with the P0 above — if login works and sudo does not, that is a specific and interesting finding.
- The password step-up alternative should be unaffected; confirm it still works so a failure here does not lock admins out of every route to sudo.

### P1 — Operator lockout case (deliberate, needs confirming, not fixing)

**Where to click**
- The admin MFA reset for a user, then that user's re-enrollment through the browser ceremony

**What to verify**
- A passkey provisioned through the API's unattested `publicKey` shortcut (rather than the browser ceremony) is **expected** to fail login now — it carries a public key nobody holds the private half of. Confirm that it fails cleanly with the normal 401 and not a 500 or a hang.
- An admin MFA reset on such a user, followed by a fresh enrollment through the real browser ceremony, restores their ability to log in.

**What can go wrong**
- The user is locked out with no recoverable path — the reset-then-re-enroll route must work, since `UPGRADE_NOTES.md` sends operators down it.
- The failure surfaces as a 500 or an unhandled error rather than the documented 401.

### P2 — Documentation accuracy

**Where to click**
- `BACKWARD_COMPATIBILITY.md`, `UPGRADE_NOTES.md`, `.ai/qa/scenarios/TC-SEC-004-passkey-enrollment-and-login.md`

**What to verify**
- The described client migration matches what you actually observed: the shipped UI sends `startAuthentication()` output as `payload.response` and needed no change.
- The operator lockout guidance matches the behavior you just confirmed in P1.

**What can go wrong**
- The docs promise a migration path that does not match the running system.

### Pass/fail
- Any route fails → remove the `qa` label, add `qa-failed`, and leave a comment describing the failure. A failing **positive** passkey login is a blocker, not a nit — it means the tightening overshot.

pkarw:

Full review posted above. No blockers and no un-waived majors: the one major — removing the `{ credentialId, challenge }` passkey verify shape without a deprecation window — carries an explicit written waiver in `BACKWARD_COMPATIBILITY.md` and `UPGRADE_NOTES.md`, and the waiver is sound because the shape being removed *is* the vulnerability. Two nits were left for the author rather than fixed: the unreachable assertions below `test.skip(true, …)` in `TC-SEC-006`/`TC-SEC-007` (whose new 401 assertions get no CI signal, since both tests already skip earlier on the pre-existing `!passkeyMethod` gate), and the missing audit query behind the `UPGRADE_NOTES.md` instruction to review `mfa_methods` passkey rows.

autofix: not run — the verdict is approve, so there were no actionable findings to fix.

⚠️ **This PR still needs an approving review from someone other than its author.** GitHub refuses a self-approval, so the review above was submitted as a `COMMENTED` review rather than `APPROVED`. Treat it as the code-review pass, not as the approval branch protection is waiting for.

CI: all checks on `93c93e87f` were already settled and green at review time, so no CI follow-up is owed and `ci-monitoring` was not applied.

Merge is still gated by `needs-qa` → `qa-approved` (manual QA instructions posted above) and by a non-author approving review.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-14T15:08:48Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

pkarw:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ **PASS** — the change's own scope is fully verified. The check this PR most needed, and the one no automated suite could give it, is green: a **real, cryptographically signed WebAuthn assertion is still accepted** for both passkey login MFA and passkey sudo step-up, so the tightening did not overshoot and passkey users are not locked out. Every non-cryptographic route into `verify()` was refused with `401`.

**Environment:** `http://localhost:37017` · role `admin` (`admin@acme.com`, the descriptor's demo account) · browser `playwright` (Chromium 151)
**Verified:** `fix/passkey-mfa-require-webauthn-assertion` @ `93c93e8`

### How a genuine authenticator was produced

The manual QA instructions asked for Touch ID, Windows Hello, or a hardware key, because the passing ceremony has no automated coverage until a virtual authenticator lands (#5307). This run used **Chromium's CDP virtual authenticator** (`WebAuthn.addVirtualAuthenticator`, `ctap2` / `internal` transport, resident key, user-verified). That is not a mock of the WebAuthn layer: the browser generates a real P-256 key pair, performs the real ceremony, and signs each assertion, so `@simplewebauthn/server` had to verify a genuine signature against the stored COSE public key. The authenticator's signature counter advancing `2 → 3` across two logins is the observable proof that real, distinct assertions were signed.

Two environment details worth recording, because both silently break this surface:
- The app had to be built with `OM_ENABLE_ENTERPRISE_MODULES=true` and `OM_ENABLE_ENTERPRISE_MODULES_SECURITY=true`. The local `.env.example` defaults leave them off, and with them off the entire enterprise security surface — every `/api/security/*` route and every `/backend/security/*` page — returns `404`, so a QA pass would silently verify nothing.
- The app was driven over `http://localhost:…` rather than the descriptor's `http://127.0.0.1:…`. The module derives the WebAuthn RP ID from the request hostname, and an IP literal is not a valid RP ID, so the ceremony cannot start on the `127.0.0.1` form.

### Scenario (P0 — passkey login MFA and passkey sudo step-up)

**Where to click:** `/login` → the passkey challenge panel · `/backend/profile/security/mfa/passkey` · `/backend/security/users/{id}`

| 1 | Cold-load `/login`. | The sign-in form renders. | Rendered and interactive in **191 ms**. | ✅ |
| 2 | Sign in with a password while the admin has no MFA method. | Login succeeds, lands on `/backend`. | Landed on `/backend` with the backend shell rendered. | ✅ |
| 3 | Enroll a passkey through the browser registration ceremony. | The ceremony completes and the key is listed. | Ceremony completed, "QA virtual authenticator" listed, authenticator holds 1 credential. | ✅ |
| 4 | Sign out, sign back in with the password only. | The session stays MFA-pending and the passkey challenge appears. | The "Passkey verification" panel replaced the sign-in form; access not granted yet. | ✅ |
| 5 | **Complete the ceremony with a genuinely signed assertion.** | Verification succeeds and lands on `/backend`. | **The signed assertion was accepted; landed on `/backend` 215 ms after the click.** | ✅ |
| 6 | Repeat the entire passkey login a second time. | The second login also succeeds (signature-counter handling is sound). | Second login succeeded; authenticator signature counter moved **2 → 3**. | ✅ |
| 7 | Trigger the sudo-protected admin MFA reset for another user. | The sudo modal appears and offers the passkey method. | "Confirm sensitive action" modal opened offering passkey step-up. | ✅ |
| 8 | **Complete the sudo passkey ceremony.** | A sudo token is issued and the protected action goes through. | **Assertion accepted, sudo token issued, "User MFA reset completed."** | ✅ |
| 9 | Replay the removed `{ credentialId, challenge }` payload against `/api/security/mfa/verify`. | `401`, no token, no `500`. | Answered **`401`** with no token. | ✅ |
| 10 | Send seven consecutive rejected payloads against one challenge. | Each refused, and the challenge eventually locks out rather than accepting forever or `500`-ing. | Statuses `401, 401, 401, 401, 401, 400, 400` — refusals still count toward the attempt limit and the challenge locked out. | ✅ |
| 11 | Provision a passkey via the unattested `publicKey` enrollment shortcut, then log in with it. | Fails cleanly with `401`, not a `500` or a hang. | Enrollment still returns `200`; the subsequent login verification answered **`401`** — the documented, recoverable failure mode. | ✅ |
| 12 | Repeat the password step at a 390×844 mobile viewport. | The challenge is legible and its action reachable. | Challenge panel rendered within the viewport, action button visible and tappable. | ✅ |
| 13 | Dismiss the passkey prompt (authenticator rejects with `NotAllowedError`, exactly as a user cancel does). | No verification is submitted and the challenge stays on screen. | No verify request was sent; the page stayed on the challenge, so the factor was neither passed nor consumed. | ✅ |

**Challenge-TTL expiry was not exercised** (⚠️ not exercised): confirming it needs a wall-clock wait past the 5-minute WebAuthn challenge TTL, and the PR's own plan lists challenge TTL handling as an explicit non-goal, so nothing in this diff can have moved it.

### ⚠️ Out-of-scope finding — please read, but it does not block this PR

While driving step 13 I checked what an MFA-pending session can reach, and found that **the token issued by the password step — the one carrying `mfa_pending: true`, before any second factor — is accepted by ordinary APIs**:

| Route | With the `mfa_pending` token | With no token |
|-------|------------------------------|---------------|
| `GET /api/customers/people` | `200`, returns real tenant records | `401` |
| `GET /api/auth/users` | `200`, returns the user-administration list | `401` |

In the browser, that same un-verified session rendered the **full authenticated dashboard**, greeting the admin by name (last screenshot below). So the second factor can currently be skipped by ignoring the challenge panel and simply navigating on with the pending token.

**This is not a regression from this PR, and it is not a reason to hold it.** PR 5306 changes only `PasskeyProvider.verify()` — the verify payload schema and its acceptance branches. It does not touch how `mfa_pending` tokens are authorized on ordinary routes, so this behavior is pre-existing on `develop` and is strictly unaffected by merging. It is the same *class* of defect as #3852 (a second factor that can be passed without actually passing it), one layer up in session authorization rather than in passkey verification, which is why it is worth filing on its own. I have not opened an issue for it — say the word and I will.

### Screenshots

**Step 1 — `/login` cold load**



**Step 2 — password sign-in before any MFA method exists**



**Step 3 — passkey enrolled through the browser registration ceremony**



**Step 4 — the passkey challenge after the password step**



**Step 5 — **a genuinely signed assertion is accepted; login lands on `/backend`****



**Step 6 — second passkey login succeeds (signature counter 2 → 3)**



**Step 7 — the sudo step-up modal offers the passkey method**



**Step 8 — **sudo passkey step-up issues a token; "User MFA reset completed."****



**Step 12 — the challenge at a 390×844 mobile viewport**



**Step 13 — a dismissed prompt submits no verification and the challenge stays**



**⚠️ Out-of-scope finding — an MFA-pending session rendering the full dashboard**

pkarw:
## 🧪 Follow-up: the passing passkey ceremony can be automated now (#5307)

The QA above was manual in the sense that a human authored it, but it ran entirely unattended — which means the coverage gap this PR documents as "needs a real authenticator" is closable today. `.ai/qa/scenarios/TC-SEC-004-passkey-enrollment-and-login.md` currently says step 8 "stays **manual**… until a Playwright virtual authenticator lands"; this run is the working proof of that approach, so [#5307](https://github.com/open-mercato/open-mercato/issues/5307) can be implemented directly from the recipe below rather than from scratch.

**The mechanism.** Chromium's CDP virtual authenticator does the real ceremony — real P-256 key pair, real signature — so `verifyAuthenticationResponse` is genuinely exercised rather than mocked:

```ts
const client = await context.newCDPSession(page)
await client.send('WebAuthn.enable', { enableUI: false })
const { authenticatorId } = await client.send('WebAuthn.addVirtualAuthenticator', {
  options: {
    protocol: 'ctap2',
    transport: 'internal',
    hasResidentKey: true,
    hasUserVerification: true,
    isUserVerified: true,
    automaticPresenceSimulation: true,
  },
})
```

**Scenario to implement (derived from the run above, all of it verified working):**

1. **Setup:** create a user fixture through `POST /api/auth/users` with an admin token, as `securityFixtures.ts` already does. Add the virtual authenticator before the first navigation.
2. **Act — enrollment:** sign in with the password, go to `/backend/profile/security/mfa/passkey`, fill the nickname field and submit. The registration ceremony completes without a prompt.
3. **Assert — enrollment:** the key is listed, and `WebAuthn.getCredentials` reports exactly one stored credential.
4. **Act — login:** clear cookies, sign in with the password only, then click the challenge panel's passkey button.
5. **Assert — login (the happy path #5307 is about):** the browser lands on `/backend`, and `WebAuthn.getCredentials` shows the signature counter advanced. Repeat once to pin the counter-update path — a counter regression shows up as the *second* login failing while the first worked.
6. **Act/Assert — sudo step-up:** trigger the sudo-protected admin MFA reset for the fixture user, complete the passkey ceremony in the modal, and assert the sudo token is issued and the reset succeeds. This restores the `TC-SEC-006` / `TC-SEC-007` coverage this PR had to `test.skip`, since those specs can now obtain a sudo token again.
7. **Act/Assert — refusal:** call `WebAuthn.removeVirtualAuthenticator`, then click the passkey button. `navigator.credentials.get()` rejects with `NotAllowedError`, which is exactly what a user-dismissed prompt produces, and no verify request should be sent.
8. **Teardown:** delete every fixture user created, and remove the enrolled MFA methods so a re-run starts clean.

**Two environment prerequisites the test must encode, because both fail silently:**
- The suite needs `OM_ENABLE_ENTERPRISE_MODULES=true` and `OM_ENABLE_ENTERPRISE_MODULES_SECURITY=true`. CI already sets them in `ci.yml`; a local run without them gets `404` on every `/api/security/*` route, so the spec should skip with an explicit reason rather than fail obscurely.
- The base URL must be a **hostname**, not an IP literal. The module derives the WebAuthn RP ID from the request host, and `127.0.0.1` is not a valid RP ID, so the ceremony cannot start against the default ephemeral `http://127.0.0.1:<port>` base URL. Use `http://localhost:<port>`.

The throwaway spec this run used is in the QA artifacts directory (`.ai/qa/artifacts_20260814-151238-qa5306/spec/`) and can be lifted into `packages/enterprise/src/modules/security/__integration__/` largely as-is; it is deliberately not part of this PR, which stays read-only on source.

pkarw:

The UI QA ran under the `--evidence-only` default, so **no pipeline or meta label was changed** — the PR keeps `merge-queue` + `needs-qa`, and the QA-approval gate still holds the merge until a maintainer adds `qa-approved`. That is deliberate: this skill does not self-approve QA unless it is invoked with `--self-qa-signoff`, and the decision to clear the gate stays with a human reviewer.

For the reviewer picking this up, the short version: the risk this PR was flagged for — that the tightening overshot and locked every passkey user out — **did not materialise**. A genuinely signed WebAuthn assertion, produced by a real ceremony against a Chromium virtual authenticator, is accepted for both passkey login MFA and passkey sudo step-up, twice in a row with the signature counter advancing. Every non-cryptographic route into `verify()` answered `401`, refusals still count toward the challenge attempt limit, and a credential from the unattested enrollment shortcut fails cleanly with `401` rather than a `500`, matching what `UPGRADE_NOTES.md` promises operators.

One thing that is **not** about this PR but that a reviewer should see: a session that has passed only the password step — token claim `mfa_pending: true` — is accepted by ordinary APIs (`/api/customers/people`, `/api/auth/users` both answer `200` with real data, versus `401` with no token) and renders the full backend dashboard. That is session-authorization behavior this diff does not touch, so it is pre-existing on `develop` and unaffected by merging; details and a screenshot are in the evidence comment. No issue has been filed for it.

haxiorz (changes requested):
## Strict review — CHANGES REQUESTED

Reviewed `93c93e87f20f`. 1 Critical, 1 High, 1 Medium.

### Critical

1. `BACKWARD_COMPATIBILITY.md:341` — **Resolve the one-off waiver against the mandatory compatibility gate.** A merge leaves authoritative policy simultaneously forbidding every unbridged STABLE change and accepting this one by fiat; downstream legacy clients receive 401 while later reviews lack a rule for legitimate emergency waivers. This either ships a Critical compatibility violation or weakens the compatibility gate through an undefined precedent. Required correction: Decide and document the governing classification. If this was an undocumented vulnerability, remove the STABLE-exception claim. If it was public, add an authorized emergency-security exception to the top-level policy and an issue-specific migration spec; do not retain the vulnerable branch as a bridge.

### High

1. `UPGRADE_NOTES.md:35` — **Do not describe shortcut-enrolled credentials as necessarily unusable.** An attacker with a stolen password obtains an mfa_pending token, enrolls a controlled key through the still-enabled shortcut, then signs this patch's genuine assertion; an operator relying on the note treats it only as a lockout concern. The release guidance can leave an account-takeover path enabled under false assurance. Required correction: State that the shortcut can store either an unusable key or an attacker-controlled valid keypair. Treat all such credentials as potentially hostile and give an effective release mitigation, such as sequencing #5296 first, rejecting existing-method mfa_pending enrollment, or resetting and re-enrolling affected passkeys.

### Medium

1. `UPGRADE_NOTES.md:35` — **Provide an audit instruction that maps to the actual schema.** An operator queries nonexistent mfa_methods or corrects the name but assumes indistinguishable rows reveal API enrollment; affected users surface only after lockout or hostile credentials remain enabled. The pre-upgrade audit is not executable and cannot identify the population it promises to remediate. Required correction: Use user_mfa_methods and provide a supported detector or conservative remediation. If provenance cannot be reconstructed, say so and instruct affected deployments to reset and re-enroll all passkey methods instead of claiming rows can be audited by origin.

### Validation

- GitHub checks on immutable head: passed — All 12 executed checks on 93c93e87f20f succeeded; six non-applicable checks were skipped and none failed or remained pending.
- Blind DeepSeek V4 Pro council: passed — Strict validator accepted the fresh-context deepseek-v4-pro result with observed api.deepseek.com provenance and matching subject hash.
- Host source and contract inspection: passed — Inspected the full diff, changed files, login and sudo call chains, installed SimpleWebAuthn 13.3.2 verifier, governing specs, compatibility policy, migrations, and linked issues #3852, #5296, and #5307.

pkarw:
🤖 `om-auto-review-pr` (autofix mode) — strict review addressed in `4b9191066`.

All three findings from the `CHANGES_REQUESTED` review on `93c93e87f` are resolved. **No source behavior changed** — `PasskeyProvider.ts` is untouched by this round; the findings were all about governance and operator guidance.

### Critical — `BACKWARD_COMPATIBILITY.md:341`, waiver against the mandatory gate

The objection was exact: the document forbade every unbridged STABLE change and then accepted this one as "the exception rather than the rule", which either ships a policy violation or dissolves the gate through undefined precedent.

Taking the second branch you offered — this **was** a public surface, so "remove the STABLE-exception claim" would be a false classification. The removed shape was exported in `MfaProviderInterface.verifySchema` and pinned by a test named *"supports legacy verification payload for backward compatibility"*. Third parties could validate against it and the repo's own suite treated it as supported. It is a genuine category-7 break.

So the fix is the authorized-exception route:

- The **Deprecation Protocol** gains an `### Emergency Security Exception` clause. It waives steps 1–3 — and only those — when the surface being removed *is itself* the vulnerability, explicitly excluding surfaces that merely make insecure usage possible. Five requirements MUST all hold, or the change follows the ordinary protocol: (1) the qualifying condition is argued, not asserted; (2) the removal is the narrowest one that closes the hole, with **no partial bridge retaining the vulnerable branch** behind a flag, toggle or opt-in — your "do not retain the vulnerable branch as a bridge" is written into the clause itself; (3) steps 4 and 5 still apply in full; (4) a dated entry is recorded; (5) a **human maintainer signs off by name** — an automated review cannot clear it.
- The passkey entry now invokes the clause, states the classification explicitly, and carries a table showing how each of the five requirements is met.
- Step 5 wants a spec, so there is one: `.ai/specs/enterprise/2026-08-14-passkey-mfa-require-webauthn-assertion.md`, with the mandated **Migration & Backward Compatibility** section.

Requirement 5 is deliberately left open — it is a human action, and this comment is not it.

### High — `UPGRADE_NOTES.md:35`, false assurance about shortcut-enrolled credentials

Correct and important; the old text read an account-takeover path as a lockout inconvenience. Rewritten to state both cases:

- an **unusable key**, which now fails every login (the lockout case, unchanged); or
- an **attacker-controlled but perfectly valid keypair** — a caller who generated a real P-256 keypair in software and enrolled its COSE public key signs assertions this PR *accepts*. Requiring a real assertion does not help; the key is real, it is just not the user's.

The note now says to treat every shortcut-enrolled passkey as potentially hostile rather than merely broken, and states the reachability concretely: `authorizeMfaEnrollmentMutation` (`api/mfa/_shared.ts`) authorizes on `auth.sub` and never rejects an `mfa_pending` token, so an attacker with a stolen password may be able to enroll a key they control and then satisfy the second factor with it. Mitigations are given strongest-first: sequence #5296 ahead of the upgrade; otherwise reset and re-enroll; at minimum reset passkeys held by `security.mfa.manage` holders and admins. The same correction is applied to the migration paragraph in `BACKWARD_COMPATIBILITY.md`, which carried the identical claim.

### Medium — `UPGRADE_NOTES.md:35`, audit instruction that does not map to the schema

Both halves confirmed. The table is `user_mfa_methods`, discriminated by column `type` — verified against `data/entities.ts:16` and `Migration20260305120000_security_scaffold.ts`; `mfa_methods` does not exist. And provenance genuinely cannot be reconstructed: both enrollment paths write the same `provider_metadata` keys (`credentialId`, `credentialPublicKey`, `counter`, `transports`, `label`) and the shortcut lets the caller choose all of them, so no field attributes a row to an origin.

The note now says that plainly instead of promising an audit it cannot deliver, ships a working enumeration query, and directs deployments that ever provisioned passkeys through the API to reset and re-enroll all of them rather than to identify individual rows.

### Nits from the first review

Both handled. `TC-SEC-006`/`TC-SEC-007` carry a header comment stating that the body stops at the mid-body `test.skip`, that the tail is the blueprint for the #5307 virtual-authenticator work, and that assertions above the skip still run and can still fail. The `UPGRADE_NOTES.md` nit is subsumed by the Medium fix above.

### Validation

**Validation source: GitHub checks.** This round changes four Markdown files and adds three comment lines to two Playwright specs; no runtime code is touched, so no local fallback gate was run per this repo's GitHub-checks-first rule. CI on `4b9191066` is the authority — I am watching it and will report.

Pipeline label moved `merge-queue` → `review`: the PR had an outstanding changes-requested review, so `merge-queue` was stale. `needs-qa` stays, and the QA-approval gate still holds the merge.

pkarw:
✅ CI green on `4b9191066` — every executed check passed, none failed or pending: `test`, `ephemeral-integration`, `lint`, `docker-build`, `ds-lint`, `prepare`, `scope`, `audit-scope`, `CodeQL`, `Analyze (actions / javascript-typescript / python)`, `license/cla`. Six non-applicable jobs skipped (`mutate`, `merge-coverage`, `comment`, `audit`, `[code]smith`).

**I am not submitting an approving re-review, and that is deliberate rather than a gap.** Requirement 5 of the Emergency Security Exception this round added says a human maintainer must sign off on the waiver by name and that an automated review cannot clear it. Having just written that rule to resolve the Critical, clearing it with an automated approval would reintroduce exactly the by-fiat problem the finding was about.

So the PR sits on `review` awaiting two human decisions, both of which are correctly blocking:

1. **The compatibility waiver** — a maintainer approval that acknowledges the unbridged removal of a STABLE contract surface under the new exception.
2. **Manual QA** — `needs-qa` without `qa-approved` holds the merge. Passkey verification has no automated happy path (#5307), so a human still needs to confirm a real passkey login and a real passkey sudo step-up succeed against a genuine authenticator.

One thing worth weighing at sign-off, surfaced by the High finding rather than introduced by it: this PR closes the *verification* half only. Until #5296 lands, the enrollment shortcut can still store an attacker-controlled valid keypair that this fix accepts, so the release sequencing note in `UPGRADE_NOTES.md` is load-bearing, not boilerplate.

Issue #2270, opened 2026-05-29
security(api_docs): unauthenticated OpenAPI/markdown endpoints expose full internal API surface + ACL feature names


pkarw:
> Adversarially-verified internal security review (OWASP A05/A01). Verified: both routes are unauthenticated app-router handlers outside the `[...slug]` auth dispatcher, no middleware.

## Summary
`/api/docs/openapi` and `/api/docs/markdown` are dedicated app-router routes **outside** the auth-enforcing `[...slug]` dispatcher, perform **no auth or feature check**, and there's no app middleware. They emit every API path, request/response schema, and the **verbatim ACL feature strings** (`Requires features: ...`) for all enabled modules to any anonymous client — dramatically lowering the cost of targeted privilege-escalation and fuzzing. `sanitizeOpenApiDocument` only restructures schemas; it does not strip descriptions/ACL strings. (The backend docs *page* sets `requireAuth`, but the underlying data routes are open.)

## Location
- `apps/mercato/src/app/api/docs/openapi/route.ts` (`export async function GET()`, no auth)
- `apps/mercato/src/app/api/docs/markdown/route.ts` (same)
- `packages/shared/src/lib/openapi/generator.ts` (`buildOpenApiDocument`; sanitize doesn't strip ACL strings)

## Fix plan
1. Gate both handlers: add `getAuthFromRequest(req)` + `hasAllFeatures(['api_docs.view'])` at the top of each GET, returning 401/403 before building the document — **or** move them under the module API convention so the `[...slug]` dispatcher enforces `{ requireAuth: true, requireFeatures: ['api_docs.view'] }`.
2. If any public exposure is intended, strip the `Requires features/roles` lines from anonymous output.

mat89c:
I'd like to work on this issue.

pkarw:
ℹ️ #2319 (https://github.com/open-mercato/open-mercato/pull/2319) referenced this issue but was closed **without merging** on 2026-06-01T06:57:58Z.

This issue remains open. Posted automatically by the `sync-merged-pr-issues` skill.

pkarw:
ℹ️ #2319 (https://github.com/open-mercato/open-mercato/pull/2319) referenced this issue but was closed **without merging** on 2026-06-01T06:57:58Z.

This issue remains open. Posted automatically by the `sync-merged-pr-issues` skill.

pkarw:
This endpoint is intentionally open

pkarw:
[...]

PR #5265, opened 2026-08-13
fix(api_docs): strip ACL feature names from anonymous docs exports (#2270)

pkarw:
## Summary

`/api/docs/openapi` and `/api/docs/markdown` are intentionally public (per the maintainer decision on #2270), but they were publishing every operation's **verbatim ACL metadata** to anonymous clients: the `Requires features: …` / `Requires roles: …` description lines and the `x-require-features` / `x-require-roles` extensions. That hands an anonymous caller the exact internal feature identifiers of the deployment — the alternative remediation the original report proposed for the "public exposure is intended" case.

This PR keeps the endpoints open and redacts the ACL identifiers for anonymous callers only. Signed-in staff continue to receive the full document.

## Changes

- `buildOpenApiDocument` takes a new `includeAccessControlMetadata` option (defaults to `true`, so every existing caller is unchanged). When `false`, the `Requires features/roles` description lines and the `x-require-features` / `x-require-roles` extensions are omitted — which also removes them from the Markdown rendering, since that renders from the document.
- New `packages/core/src/modules/api_docs/lib/document.ts` holds the shared builder (`buildApiDocsOpenApiDocument`) and the caller check (`shouldExposeAccessControlMetadata`), so both export routes stay identical. The module list and API-route manifests are passed in by the route, never re-resolved from a registry.
- Both export routes resolve the caller with `getAuthFromRequest(req)` and build the document accordingly. Auth resolution failing (e.g. DB unavailable) fails **closed** to the redacted document.
- The `/docs/api` Explorer forwards the incoming cookie header to the export route, so a signed-in user still sees the full document in the playground.
- Mirrored into the `create-app` template routes (Template Sync Checklist).
- Documented the behavior in `apps/docs/docs/api/overview.mdx`.

What is deliberately **not** changed: the routes stay unauthenticated, no new ACL feature is introduced, and `x-require-auth` plus the `security` scheme remain public — they say *that* auth is required without naming internal features.

## Tests

- `packages/shared/src/lib/openapi/__tests__/access-control-metadata.test.ts` — identifiers present by default, absent when excluded (JSON and Markdown), `x-require-auth` / `security` / inferred `401` preserved.
- `packages/core/src/modules/api_docs/lib/__tests__/document.test.ts` — anonymous vs authenticated resolution, fail-closed on auth errors, redaction of the built document.
- `apps/mercato/src/app/api/docs/__tests__/routes.test.ts` — both routes, anonymous vs authenticated, JSON and Markdown.

[...]

Issue #4514, opened 2026-07-25
Clean up the obsolete onboarding email uniqueness constraint


pkarw:
## Follow-up from #4160

PR [#4160](https://github.com/open-mercato/open-mercato/pull/4160) encrypts `onboarding_requests.email` with non-deterministic ciphertext and adds `email_hash` for deterministic lookup and deduplication. The legacy unique constraint on encrypted `email` is therefore no longer the meaningful deduplication guard and can mislead future maintainers.

Reviewer request from [the code-review comment](https://github.com/open-mercato/open-mercato/pull/4160#issuecomment-4958127255):

> Remove or clearly document the now-vacuous unique constraint on encrypted onboarding email, keeping `email_hash` as the deduplication contract.

### Acceptance criteria

- [ ] Decide whether to remove the `email` uniqueness annotation/constraint or document why it remains.
- [ ] If schema metadata changes, include the generated migration and snapshot update.
- [ ] Preserve hash-based lookup/deduplication and legacy plaintext-row compatibility.
- [ ] Add or update regression coverage for the chosen schema behavior.

Related: #4160, #4091

adeptofvoltron:
🤖 `autofix` started by @adeptofvoltron at 2026-08-17T08:16:00Z. Other auto-skills will skip this issue until the lock is released.

**Claim override note:** this issue was self-assigned by @pkarw at filing time (2026-07-25) with no branch, PR or further activity since. The takeover was explicitly approved by the maintainer running this session; @pkarw stays on the assignee list, so please shout here if this is already in flight on your side and I will back out.

adeptofvoltron:
🤖 `om-open-pr` — completed: the cleanup ships in #5336, which drops `onboarding_requests_email_unique` from the entity, the migration snapshot and the table, and adds a regression test locking `email_hash` in as the deduplication contract. The issue's `in-progress` lock is released and the lock now sits on the PR; @pkarw stays assigned as the issue's owner.

PR #5312, opened 2026-08-14
fix(attachments): serve an empty page past the end instead of clamping (#5299)

### LINKED ISSUE #5299: `/api/attachments` clamps the requested page instead of serving an empty one
`/api/attachments` is the only list endpoint I found that clamps the requested page to the last page rather than serving an empty page past the end — for the offset and for the `page` it echoes back:

```ts
// packages/core/src/modules/attachments/api/route.ts:215-216
const pageOffset = usePaging ? (Math.min(currentPage!, totalPages!) - 1) * currentPageSize! : undefined
// :259
page: Math.min(currentPage!, totalPages!),
```

Every CRUD list built on the query engine computes an unclamped offset (`packages/shared/src/lib/query/engine.ts:1137`, `:1108`), so a page past the end comes back empty and an incremental loader terminates naturally.

Consequence: a client that pages by "keep going while the page is full" is served the last page forever. #5274 hit this in `AttachmentsSection` and worked around it in the component, treating an echoed page below the requested one as the end. That works, but it is a per-caller workaround for a server-side inconsistency, and every new consumer of this endpoint has to know to repeat it.

Proposal: drop both `Math.min(...)` clamps so the route behaves like the rest — an offset past the end yields empty `items`, and `page` echoes what was asked for. Any caller relying on the clamp would need checking; the admin attachments UI does not, after #5274.

Raised by @pkarw in the review of #5274.

adeptofvoltron:
Closes #5299

## 🎯 Goal
`GET /api/attachments` now behaves like every other paged list in the codebase: asking for a page past the end returns an empty `items` array and echoes back the page the caller actually requested. A client that pages by "keep going while the page is full" terminates naturally instead of being served the last page forever.

## 🔍 Problem
`/api/attachments` was the only list endpoint that clamped the requested page down to the last existing page — both for the database offset it used and for the `page` value it echoed in the response. A caller asking for page 7 of a 3-page result got page 3's items back, labelled `page: 3`. Every consumer therefore had to know to compare the echoed page against the requested one to detect the end. #5274 hit exactly this in `AttachmentsSection` and worked around it in the component, which is a per-caller workaround for a server-side inconsistency.

## 🔍 Root Cause
In `packages/core/src/modules/attachments/api/route.ts` the record-scoped `GET` handler computed `totalPages` from the counted total and then clamped twice: `const pageOffset = usePaging ? (Math.min(currentPage!, totalPages!) - 1) * currentPageSize! : undefined` fed a clamped offset to `findWithDecryption`, and `page: Math.min(currentPage!, totalPages!)` rewrote the echoed page. Every other paged list is unclamped — the query engine uses `(page - 1) * pageSize` and returns `page` verbatim (`packages/shared/src/lib/query/engine.ts:1108`, `:1137`, `:1145`), and the sibling library route in this very module already does the same (`api/library/route.ts:71`, `:175`).

## What Changed
- `packages/core/src/modules/attachments/api/route.ts` — dropped both `Math.min(...)` clamps in the `GET` handler. The offset is now `(currentPage - 1) * pageSize`, so a page past the end produces an offset past the end and the query returns no rows; the response echoes `currentPage` unchanged. `currentPage` is already floored at 1 by `Math.max(1, page)` and by the zod schema, so no negative offset is reachable. `total`, `pageSize` and `totalPages` are unchanged and still returned, so the response shape is identical.
- `packages/core/src/modules/attachments/api/__tests__/attachments.api.test.ts` — two regression tests plus a `count` mock on the shared `mockEm` (the paged `GET` path calls `em.count`, which the fixture did not previously stub).

### Caller audit
Every `/api/attachments` consumer was checked; only callers that send both `page` and `pageSize` reach the changed code (`usePaging` is false otherwise). None relied on the clamp:

| Caller | Sends paging? | Relied on the clamp? |
|---|---|---|
| `packages/ui/src/backend/detail/AttachmentsSection.tsx:71` | yes (`pageSize=24`) | No — "Load more" is gated on `page < totalPages`, so it never requests past the end, and the echoed page is only compared against `totalPages`, which stays correct unclamped. |
| `packages/ui/src/backend/messages/useMessageCompose.ts:221` | no | n/a |
| `packages/core/src/modules/attachments/fields/attachment.tsx:74` | no | n/a |
| `ChannelOfferForm.tsx:1549,1656`, catalog product/variant pages | no | n/a |
| `TC-CAT-023.spec.ts:49,151` | `page=1&pageSize=50` | No — page 1 is never clamped. |
| `LogoUploadField`, `ProductMediaManager`, the AI upload adapter, integration fixtures | POST/DELETE or `?id=` only | n/a |

## 🧪 Tests
- `packages/core/src/modules/attachments/api/__tests__/attachments.api.test.ts` — **`serves an empty page past the end instead of clamping to the last one (#5299)`**: with 5 records at `pageSize=2`, a request for page 7 returns `items: []`, echoes `page: 7` alongside `total: 5` / `totalPages: 3`, and passes `offset: 12` to the data-access helper. Verified failing on `develop` without the fix (echoed `page: 3`).
- **`pages with an unclamped offset and echoes the requested page (#5299)`** — locks in that an in-range page (2 of 3) still gets `offset: 2` and echoes `page: 2`, guarding the offset arithmetic against a future regression.
- Validation gate from `.ai/agentic.config.json`, run in local mode: `yarn build:packages` ✅ · `yarn generate` ✅ (no generated-file churn) · `yarn build:packages` ✅ · `yarn i18n:check-sync` ✅ · `yarn i18n:check-usage` ✅ · `yarn typecheck` ✅ · `yarn test` — every `@open-mercato/*` package green, including `@open-mercato/core` (1244 suites / 9677 tests) · `yarn build:app` ✅.
- ⚠️ One **pre-existing, unrelated** gate failure: `create-mercato-app`'s `src/lib/agent-harness-evaluator.test.ts` (the live Codex/Claude standalone-harness adapter cases). Confirmed pre-existing by stashing this branch's changes and re-running that file on a clean tree — the identical cases fail. Nothing in this PR touches `packages/create-app`.

## 💥 Breaking Changes
- None. The response shape is unchanged (`items`, `total`, `page`, `pageSize`, `totalPages` all still present) and no caller depended on the clamped echo. The only behavioral difference is for a request past the last page, which previously returned the last page's items under a rewritten page number and now returns an empty page under the requested number — the behavior every other list endpoint already has.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5312"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789313652&installation_model_id=432243&pr_number=5312&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5312&signature=ff346cbec494840af7b55e0e51ccbaaa69ea64e818a9ffda9cfbe665c9853ebb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:

- 🔍 `review` — the PR is complete and ready for code review; the full validation gate from `.ai/agentic.config.json` ran green apart from a pre-existing `create-mercato-app` harness failure that reproduces on a clean tree.
- 🐛 `bug` — this is a corrective change: the endpoint clamped a requested page to the last existing page, which contradicts how every other query-engine-backed list in the codebase pages, and #5274 already had to work around it in the client.
- 🧪 `needs-qa` — the change alters observable paging behaviour of an endpoint the admin attachments UI consumes, so a QA pass should open a record with more than 24 attachments and confirm "Load more" still walks every page and stops cleanly at the end. The automated-verification exemption was deliberately not claimed here: although no `.tsx` file outside tests is touched and regression tests ship in this PR, the runtime behaviour of a user-facing API changed, and the merge gate is the safer default. A maintainer who reads it as covered by the exemption can swap this for `skip-qa`.
- 🔹 `priority-medium` — an ordinary correctness defect with no user-visible breakage today, since the one affected consumer already carries a workaround, but every new consumer of the endpoint would otherwise have to repeat that workaround.
- 🟡 `risk-medium` — a single-area change of four lines in one route handler, covered by new regression tests, but it does change the behaviour of a shared API endpoint reached from several surfaces, so it is not a `risk-low` isolated cleanup. No database structure, no auth or tenant scoping, and no response shape were touched.

adeptofvoltron:
🤖 `om-open-pr` completed: opened PR #5312 for issue #5299 with the full label set applied. Lock handed off to `om-auto-review-pr` — chain continues on this PR.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock — code review with autofix. Started: 2026-08-14T15:35:36Z.

adeptofvoltron:
# 🔍 Code Review

**PR:** #5312 — `fix(attachments): serve an empty page past the end instead of clamping (#5299)`
**Base:** `develop` · **Head:** `1a4af2e25` · **Mode:** first review, autofix-eligible (own PR, `--autofix`)

## 🎯 Scope

Two production lines change in `packages/core/src/modules/attachments/api/route.ts`: the paging offset stops being clamped to `totalPages`, and the echoed `page` stops being rewritten. Everything else in the diff is test and documentation coverage for that behavior. The change is exactly what issue #5299 asked for, and the caller audit in the PR body is accurate — I re-ran it independently and reached the same conclusion.

## ✅ What holds up

The fix is minimal and stays inside the local conventions. Reads still go through `findWithDecryption` with the tenant/organization scope intact, so nothing about data access or scoping moved. `currentPage` is floored at `1` twice over — by `Math.max(1, page)` in the handler and by `page: z.coerce.number().min(1)` in `attachmentQuerySchema` — so removing the clamp cannot produce a negative offset. The response schema at `packages/core/src/modules/attachments/api/route.ts:89` declares `page: z.number().int().min(1).optional()`, which an unclamped echo still satisfies, so the OpenAPI contract is unaffected. `totalPages` is still computed and returned, so callers that navigate by it — including `AttachmentsSection`, whose "Load more" button is gated on `page < totalPages` — keep working unchanged.

Test coverage is the right shape. `serves an empty page past the end instead of clamping to the last one (#5299)` asserts all three observable consequences at once (empty `items`, echoed `page: 7`, `offset: 12` reaching the data-access helper) and genuinely fails on `develop` without the production change. The sibling test pinning an in-range page to `offset: 2` guards the arithmetic itself, which is the part most likely to be broken by a future edit. Adding `count` to the shared `mockEm` is necessary rather than incidental — the paged branch calls `em.count`, which the fixture never stubbed because no previous test exercised paging — and it is reset per test, so no existing case is affected.

## 🔁 Backward compatibility

`/api/attachments` is a documented public endpoint, so the paging semantics are a contract surface under `BACKWARD_COMPATIBILITY.md`. Nothing is removed or renamed: route, method, and response shape are byte-for-byte the same, and only the behavior for an out-of-range page changes. That still needs to be written down for downstream module authors, which is the one substantive finding below.

## 🐛 Findings

### 🟠 Major — a public-API behavior change with no `UPGRADE_NOTES.md` entry

`UPGRADE_NOTES.md:1` — The file's own charter is "actionable incompatibilities — library behavior that affects code a downstream module author can plausibly write against," and this repo already documents precisely this class of change there (see *"Sales line list endpoints now default to `line_number` order"*, a list-endpoint behavior change with an unchanged response shape). A downstream consumer that used a deliberately large page number as shorthand for "give me the last page" gets an empty page after this PR, with no note anywhere telling them why.

**Fixed in `1a4af2e25`** — added a `### GET /api/attachments serves an empty page past the end instead of clamping (#5299)` entry to the in-flight `0.6.7 → 0.7.0` window, covering the two consumer patterns that actually change and confirming that the `#5274` client-side workaround stays correct (it simply stops triggering).

### 🔵 Minor — the record list's paging parameters were undocumented

`apps/docs/docs/api/attachments.mdx:102` — The library endpoint documents its `page` and `pageSize` parameters, but the record-scoped list never did, even though it has accepted them all along. Since this PR is specifically about defining what those parameters do at the boundary, leaving them undocumented would mean the newly-specified behavior lives only in a test.

**Fixed in `1a4af2e25`** — documented that paging is opt-in and requires both parameters, which fields the response gains when it is used, and that a page past the end returns empty items and echoes the requested page.

### ⚪ Nit — fractional `page` values yield a fractional offset (pre-existing, not fixed)

`packages/core/src/modules/attachments/api/route.ts:61` — `page: z.coerce.number().min(1)` has no `.int()`, so `?page=1.3&pageSize=3` produces `offset: 0.9`. This predates the PR — the old expression `(Math.min(currentPage, totalPages) - 1) * pageSize` did not floor either, and passes the value through unchanged whenever `page <= totalPages`. Deliberately left alone: tightening the schema is a separate input-validation change with its own compatibility question (a `400` where callers previously got a response), and widening this PR to cover it would obscure the one-line fix it exists to deliver. Worth a follow-up issue if anyone cares.

## 🧪 Validation

The full gate from `.ai/agentic.config.json` ran in local mode on the code commit: `yarn build:packages` ✅ · `yarn generate` ✅ (no generated-file churn) · `yarn build:packages` ✅ · `yarn i18n:check-sync` ✅ · `yarn i18n:check-usage` ✅ · `yarn typecheck` ✅ · `yarn test` (see below) · `yarn build:app` ✅. Every `@open-mercato/*` package passed, including `@open-mercato/core` at 1244 suites / 9677 tests.

`yarn test` reports one failure, in `packages/create-app/src/lib/agent-harness-evaluator.test.ts` — the live Codex/Claude standalone-harness adapter cases. It is **pre-existing and unrelated**: stashing this branch's changes and re-running that file on an otherwise clean tree reproduces the identical failures, and nothing in this PR touches `packages/create-app`. The follow-up commit is documentation-only (`.md` and `.mdx`), which no gate command covers, so the gate was not re-run for it.

CI on the code commit was green on every settled check (`lint`, `ds-lint`, `scope`, `audit-scope`, CodeQL `Analyze (actions)` and `Analyze (python)`), with `Analyze (javascript-typescript)` and `prepare` still pending at review time. **CI-pending disclosure:** those two checks had not settled when this review was written, so this verdict rests on the local gate plus the settled checks; the required checks still gate the merge.

## 📋 Verdict

**APPROVED** — no blockers, and the one major finding was fixed in this run rather than handed back. The remaining nit is explicitly out of scope and documented above.

Two process notes for the human reviewer. First, GitHub does not allow an author to submit an approving review on their own PR, so this report is posted as a comment rather than as a formal review; the PR still carries zero human approvals and needs one. Second, the pipeline label is deliberately left at `review` rather than moved to `merge-queue` — a PR with no human review sitting in the merge queue would misrepresent its state. `needs-qa` stays applied, so with `qaGate` on the QA-approval gate blocks the merge until a maintainer adds `qa-approved`.

adeptofvoltron:

This PR passed automated review and carries `needs-qa` without `skip-qa`, so the QA-approval gate holds the merge until `qa-approved` is added. It is deliberately still in the `review` pipeline state rather than `merge-queue`, because no human has reviewed it yet — see the review comment above. QA reviewer: when you pick it up, swap the pipeline label to `qa`, then run the routes below.

The change is server-side only. There is no new UI, no schema change, and no new permission — the single observable difference is what `GET /api/attachments` returns for a page past the end of the result set. QA is therefore about confirming that the one UI surface built on that endpoint still behaves, not about exploring new screens.

### P1 — Attachments section on a record detail page

**Where to click**
- Any backend detail page that renders the attachments section — for example a customer, deal, or product detail page (`/backend/...`) with an Attachments panel.

**What to verify**
- Upload more than 24 files to one record (24 is the section's page size, so you need at least 25 to get a second page), then reload the page → the first 24 thumbnails render and a **Load more** button appears beneath them.
- Click **Load more** → the next batch appends beneath the existing ones, no thumbnail is duplicated, and no attachment is missing compared to the total you uploaded.
- Keep clicking **Load more** until the last page loads → the button disappears exactly once every attachment is shown, and it does **not** reappear or keep serving the same final batch over and over. This is the specific regression this PR is about; the old behavior would have re-served the last page indefinitely to a naive loader.
- Open a record with fewer than 24 attachments → all of them render and no **Load more** button appears at all.
- Open a record with no attachments → the empty state renders, with no error toast and no infinite spinner.

**What can go wrong**
- Duplicated thumbnails after clicking **Load more**, which would mean the offset is wrong rather than unclamped.
- A **Load more** button that stays visible after the last page, or one that reloads the same items — either would mean `totalPages` or the echoed `page` is being misread.
- An error toast or an empty grid on the second page, which would mean the unclamped offset is reaching the database badly.
- Data scoping: log in as a user in a different organization and open a record you should not see attachments for, and confirm the list stays empty. The tenant/organization filter is untouched by this PR, but the query path it feeds did change, so it is worth one probe.

### P2 — API behavior directly (optional, faster than the UI path)

**Where to click**
- Any API client against `GET /api/attachments?entityId=<id>&recordId=<id>&page=<n>&pageSize=<n>` with a valid API key.

**What to verify**
- Request a page well past the end (for example `page=99&pageSize=5` on a record with a handful of attachments) → the response returns `items: []`, echoes `page: 99`, and still reports the true `total` and `totalPages`.
- Request page 1 of the same record → the first `pageSize` attachments come back with `page: 1`.
- Omit `page` and `pageSize` entirely → every attachment for the record is returned and the response carries no `total` / `page` / `pageSize` / `totalPages` fields.

**What can go wrong**
- A `500` on the past-the-end request, which would mean the large offset is not being handled.
- `page` echoing back a clamped value, which would mean the fix did not take effect on the deployed build.

### P2 — Documentation

**Where to click**
- The rendered docs page for the attachments API (`docs/api/attachments`).

**What to verify**
- The "List attachments" section now describes the optional `page` / `pageSize` parameters and the past-the-end behavior, and the page renders without MDX errors.

**What can go wrong**
- A docs build failure from the new paragraph, which would show up as a broken or blank section.

### Pass/fail
- All routes pass → add `qa-approved` (and `merge-queue` if a human reviewer has approved by then); this clears the QA-approval gate.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the fix for #5299 is implemented, tested, documented, and reviewed with no blockers remaining; the one major review finding (a missing `UPGRADE_NOTES.md` entry for a public-API behavior change) was fixed in this run. Lock released — `in-progress` swapped for `ci-monitoring` because one CodeQL check (`Analyze (javascript-typescript)`) had not settled yet and this run still owes a CI-result comment. The PR is free for anyone to act on; it needs a human review approval and manual QA sign-off before it can merge.

adeptofvoltron:

CI has settled on head `1a4af2e25` with **no failing checks**: `Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)`, `CodeQL`, `scope`, and `license/cla` all passed, and `comment`, `mutate`, and `[code]smith` were skipped by their own path filters. The `Analyze (javascript-typescript)` check that was still pending when the review was posted has now passed, so nothing about that verdict changes and the approval stands.

The PR reports `mergeStateStatus: BLOCKED` with `mergeable: MERGEABLE`, which is branch protection doing its job rather than a problem with the branch: there are no merge conflicts, and the block is the missing human review approval plus the `needs-qa` QA-approval gate. Removing `ci-monitoring` — this run owes no further follow-up.

kamwro:
Action item for whoever merges: **#5299 needs closing by hand.**

`Closes #5299` in the description won't fire — GitHub only auto-closes on merge into the default branch (`main`), and this PR targets `develop`. That's repo-wide rather than specific to this PR: #5319 (`Closes #5316`) and the merged #5265 (`Fixes #2270`) carry no closing link either, and every recently closed issue here was closed manually.

@pkarw could you also link #5299 under **Development** in the sidebar? I only have read access, so I can't do it myself — it keeps the connection visible even if the description is edited later.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-21T00:21:09Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review — APPROVED

**PR:** #5312 — `fix(attachments): serve an empty page past the end instead of clamping (#5299)`
**Base:** `develop` · **Head reviewed:** `fcd43e22b` (was `1a4af2e25` + a `develop` merge applied during this review) · **Mode:** first human-account review
**Validation source:** GitHub checks (all green on `1a4af2e25`: `test`, `lint`, `ds-lint`, `docker-build`, `ephemeral-integration`, `audit-scope`, `scope`, `CodeQL`, `Analyze (actions/javascript-typescript/python)`, `license/cla`). A fresh run is in flight on `fcd43e22b`; no local gate was re-run because the merge changed **only** `UPGRADE_NOTES.md` prose — `route.ts` and its test file are byte-identical to the CI-green head, and `develop` has not touched either file since this PR's base.

## 🔧 Merge conflict resolved during review

The PR had drifted into `CONFLICTING` against `develop`. The conflict was a pure additive collision in `UPGRADE_NOTES.md` — this PR's `#5299` section and `develop`'s `#4514` section were appended at the same anchor. Resolved by keeping **both**, `#4514` first (the entry already on `develop`) then `#5299`, and pushed as an ordinary merge commit (`fcd43e22b`, no force-push). Net diff versus `develop` is unchanged: the same four files, and `UPGRADE_NOTES.md` is `+8 / -0`.

## 🎯 Scope

Two production lines in `packages/core/src/modules/attachments/api/route.ts`, plus two regression tests, an `UPGRADE_NOTES.md` entry, and an `apps/docs` paging note.

| Layer | File | Assessment |
|---|---|---|
| API route | `packages/core/src/modules/attachments/api/route.ts` | ✅ correct |
| Test | `packages/core/src/modules/attachments/api/__tests__/attachments.api.test.ts` | ✅ covers the behavior change |
| Docs | `apps/docs/docs/api/attachments.mdx`, `UPGRADE_NOTES.md` | ✅ accurate and sufficient |

## ✅ What the change gets right

**The diagnosis is exact.** The record-scoped `GET` clamped twice — once into the DB offset, once into the echoed `page` — and was the outlier. The sibling route in the same module (`api/library/route.ts:71` → `const offset = (page - 1) * pageSize`) and the query engine (`packages/shared/src/lib/query/engine.ts:1137`) both page unclamped. This aligns the outlier rather than inventing a new convention.

**Tenant/organization scoping is untouched.** `em.count(Attachment, filter)` and `findWithDecryption(...)` receive the *same* `filter` object carrying `tenantId` and, when resolved, `organizationId`, so `total` and `items` cannot disagree on scope. No raw `em.find` was substituted for `findWithDecryption`.

**No negative offset is reachable.** `currentPage` is floored twice — `z.coerce.number().min(1)` in the query schema and `Math.max(1, page)` at `route.ts:213` — so `(currentPage - 1) * pageSize` is never negative. The PR body's claim here checks out.

**Backward compatibility is satisfied.** Per `BACKWARD_COMPATIBILITY.md` §7, API routes are ADDITIVE-ONLY on *response fields*: `items`, `total`, `page`, `pageSize`, `totalPages` are all still returned, no route or method changed, nothing was renamed. This is a behavior change to one field's *value*, which is correctly handled with a detailed `UPGRADE_NOTES.md` entry rather than a deprecation bridge.

**The caller audit holds.** I re-verified the one consumer that actually pages, `AttachmentsSection.tsx`: "Load more" is gated on `page < totalPages` (line 326) and page state comes from `setPage(payload.page ?? targetPage)` (line 82). It never requests past the end, and in the concurrent-deletion edge case the gate evaluates false under both the old clamp (`3 < 3`) and the new echo (`7 < 3`) — identical behavior. The #5274 workaround stays correct and simply stops firing, exactly as `UPGRADE_NOTES.md` states.

**Tests earn their place.** The `page: 7` assertion is the one that actually fails on `develop`; the `offset: 12` assertion pins the arithmetic; and the in-range companion test guards against a future off-by-one in the other direction. Adding the missing `count` mock to the shared `mockEm` fixture was necessary, not incidental.

## 📋 Findings

**Critical: 0 · High: 0 · Medium: 0 · Low: 2**

### Low 1 — an absurd-but-finite `page` loses its accidental clamp protection

`route.ts:61` — `page: z.coerce.number().min(1).optional()` has no `.int()` and no upper bound. `page=1000000000000000000&pageSize=100` parses fine and now produces `offset = 1e20`, past PostgreSQL's `bigint` ceiling (~9.22e18). The old `Math.min(currentPage, totalPages)` incidentally capped this; unclamped, the query is expected to error out as a 500 rather than return the empty page the endpoint now promises. (Reasoned from the schema and the arithmetic — I did not exercise it against a live database.)

Not blocking, for two reasons: it needs a deliberately nonsensical request from an already-authenticated caller, and the sibling `api/library/route.ts:23` carries the **identical** unbounded schema today. This PR adopts the codebase's standard rather than introducing a new gap, and holding it to a bar the convention it is aligning to does not meet would be unfair. Worth a one-line `.int().max(...)` on both attachment routes as follow-up.

### Low 2 — `page` accepts a non-integer, and now echoes it verbatim

Same schema line: `page=1.5` is accepted. That value flows into `page: currentPage` and out to a client whose declared response contract (`attachmentListResponseSchema`, `route.ts:88`) is `page: z.number().int().min(1)` — so the response can now contradict its own OpenAPI schema. Strictly pre-existing: the old `Math.min(1.5, totalPages)` echoed `1.5` too, and the fractional offset it computed was equally malformed. Same one-line `.int()` fix as Low 1.

## 🧪 Test coverage

Adequate. The behavior change is unit-tested at both ends of the boundary (in-range page 2 of 3; out-of-range page 7 of 3), asserting the echoed `page`, the `total`/`totalPages` envelope, and the offset handed to the data-access layer. No integration test is warranted — no UI-visible behavior changes and no new API surface exists to exercise.

## ✅ Verdict

**Approved.** The fix is minimal (two lines), correct, consistent with the module's own sibling route and the query engine, honestly documented for consumers, and regression-tested. The two Low findings are a shared, pre-existing schema gap rather than anything this PR introduces, and are follow-up material.

`needs-qa` stays on: the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Manual QA instructions follow in a separate comment.

**Note for whoever merges** (raised by @kamwro on this PR): `Closes #5299` will not fire automatically, because GitHub only auto-closes on merge into the default branch (`main`) and this PR targets `develop`. Close #5299 by hand.

pkarw:
Label set to `merge-queue` because code review passed; `needs-qa` stays on so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. `priority-medium` and `risk-medium` are unchanged — an ordinary single-module bug fix with a documented, caller-audited behavior change.

pkarw:

The change is server-side only — two lines in `GET /api/attachments`. There is **no visual change to verify**; the goal of QA is to confirm the attachments UI that consumes this endpoint did not regress, and that the endpoint's new past-the-end behavior is observable.

### P0 — Attachment list paging on a record detail page

**Where to click**
- Any backend record detail page that renders the attachments section — e.g. `/backend/customers/<id>`, `/backend/catalog/products/<id>`, or a sales document detail page.

**What to verify**
- Upload **more than 24** attachments to one record (the section pages at `pageSize=24`) → the list shows the first 24 and a **"Load more"** button appears.
- Click "Load more" → the next batch appends, the already-loaded items stay put, and no item is duplicated or skipped across the boundary.
- Keep clicking until the last page → the **"Load more" button disappears** once every attachment is loaded, and the total set on screen equals what was uploaded.
- Reload the page → the first 24 come back and paging still works from a cold start.

**What can go wrong**
- "Load more" never disappears, or clicking it re-appends the same last batch forever (the pre-fix symptom, in reverse).
- The button vanishes one page early, hiding the tail of the list.
- Items duplicated at a page boundary, or the count on screen not matching the number uploaded.
- Delete an attachment from the middle of a multi-page list, then click "Load more" — the list must not error out or silently drop a page.

### P0 — Tenant / organization isolation

**Where to click**
- The same attachments section, signed in as two different organizations (and, if available, as a superadmin with "All organizations" selected).

**What to verify**
- Org A's record shows only Org A's attachments; the paged counts (`total`) reflect that org only. The offset arithmetic changed, so this is worth one explicit pass.
- A record with **zero** attachments renders the empty state cleanly, no "Load more".

**What can go wrong**
- Any attachment from another organization appearing in a later page — pages beyond the first are the case worth probing, since that is the code path that changed.

### P1 — Direct API check (optional but the most direct evidence)

**Where to click**
- Any REST client / browser tab, authenticated as a backend user.

**What to verify**
- `GET /api/attachments?entityId=<entity>&recordId=<id>&page=1&pageSize=2` on a record with ~5 attachments → `items` has 2, `total: 5`, `page: 1`, `totalPages: 3`.
- `GET …&page=7&pageSize=2` (well past the end) → **`items: []`** and **`page: 7`** echoed back, with `total: 5` and `totalPages: 3` still correct. Before this PR the same request returned the last page's items labelled `page: 3` — that is the whole fix.
- Omit `pageSize` entirely → every attachment is returned and `total` / `page` / `pageSize` / `totalPages` are absent from the response (paging is opt-in, unchanged).

**What can go wrong**
- A page past the end still returning items, or echoing a page number other than the one requested.
- `total` or `totalPages` shifting from their pre-change values for the same data.

### Pass/fail

pkarw:

Summary: 0 blocker / 0 major / 0 medium, 2 low (a shared, pre-existing `page` schema gap on both attachment routes — no `.int()` and no upper bound — worth a one-line follow-up, not blocking). One autofix iteration ran: the `UPGRADE_NOTES.md` conflict against `develop` was resolved by keeping both entries and pushed as a merge commit (`fcd43e22b`, no force-push), so the PR is `MERGEABLE` again. CI is re-running on the new head. `merge-queue` applied; `needs-qa` retained, so the QA-approval gate still holds the merge.

pkarw:

CI has settled green on the merged head `fcd43e22b`: `test`, `docker-build`, `ephemeral-integration (none)`, `documents-multi-instance`, `lint`, `ds-lint`, `prepare`, `audit-scope`, `scope`, `CodeQL`, `Analyze (actions / javascript-typescript / python)` and `license/cla` all passed; `merge-coverage`, `mutate`, `comment`, `audit` and `[code]smith` were skipped by their own path filters. No failing checks.

PR state is now `MERGEABLE` / `CLEAN` with `reviewDecision: APPROVED`. The only thing standing between this and merge is the QA-approval gate — `needs-qa` is still on and needs a QA reviewer to add `qa-approved`.

PR #5318, opened 2026-08-14
fix(ui): clip RichEditor toolbar measurement row to stop page overflow

kriss145:
## 🎯 Goal

Removes the stray page-level horizontal scrollbar that appears on any backend page hosting a `full`-variant `RichEditor` inside a narrow column. Reproduced on `/backend/customers/companies/create`, where the whole admin layout could be scrolled sideways and the left sidebar slid out of view.

## 🔍 Problem

On `/backend/customers/companies/create` the page gained a horizontal scrollbar and roughly 200–300px of dead scroll space to the right. Nothing visible sat in that region — the content itself fit the viewport — but the document was scrollable sideways, pushing the app sidebar off-screen when scrolled.

The page renders the "Executive notes" custom field, which `CrudForm` maps to `<RichEditor variant="full" />` ([`CrudForm.tsx:4170`](https://github.com/open-mercato/open-mercato/blob/develop/packages/ui/src/backend/CrudForm.tsx#L4170)), inside the narrow right-hand column of the `lg:grid-cols-[7fr_3fr]` form layout ([`CrudForm.tsx:3195`](https://github.com/open-mercato/open-mercato/blob/develop/packages/ui/src/backend/CrudForm.tsx#L3195)).

## 🔍 Root Cause

`RichEditorAutoToolbar` renders a hidden measurement row to decide how many toolbar items fit before collapsing the rest into the `⋮` overflow menu. That row contains **every** toolbar item laid out with `flex-nowrap`, and is taken out of flow with `absolute` + `invisible`:

```tsx
<div className="relative w-full">
  <div
    ref={measureRef}
    aria-hidden="true"
    className="pointer-events-none invisible absolute left-0 top-0 flex flex-row flex-nowrap items-center gap-0.5"
  >
    {items.map(renderToolbarItem)}
```

Two CSS details combine into the bug:

1. **`visibility: hidden` removes painting, not the layout box.** The measurement row still occupies its full intrinsic width — which is the point, since `measure()` reads `offsetWidth` off its children.
2. **Absolutely positioned boxes still contribute to the scrollable overflow area of their ancestors.** Being out of flow keeps them from affecting siblings' layout, but it does not exempt them from the ancestor's scrollable region.

With `variant="full"` the row carries ~25 items (heading, font size, colour, B/I/U/S, lists, alignment, link, comment, mention, image, table, hr, code, help…) for roughly 700–800px of intrinsic width, rendered into a `3fr` column about 430px wide. Nothing on the ancestor chain clipped it — the `relative` wrapper, the surrounding card, the grid column, `Page`/`PageBody` ([`Page.tsx`](https://github.com/open-mercato/open-mercato/blob/develop/packages/ui/src/backend/Page.tsx) is plain `space-y-*`) and `AppShell`'s `<main className="flex-1 p-4 lg:p-6 mx-auto w-full max-w-screen-2xl">` ([`AppShell.tsx:1471`](https://github.com/open-mercato/open-mercato/blob/develop/packages/ui/src/backend/AppShell.tsx#L1471)) all leave `overflow` visible. The overflow therefore propagated all the way to the document and became a page-level scrollbar.

The visible toolbar was never the culprit — it already clips itself with `w-full flex-nowrap overflow-hidden`. Only the measurement twin leaked.

## What Changed

- **`packages/ui/src/primitives/rich-editor.tsx`** — added `overflow-hidden` to the `RichEditorAutoToolbar` wrapper (`<div className="relative w-full">`), so the hidden measurement row is clipped at the component boundary instead of extending the document's scrollable area. One-line change; no logic, props, or public API touched.

The clip is applied to the **wrapper**, deliberately not to the measurement row itself — the row must stay free to reach its full intrinsic width, otherwise the `offsetWidth` readings that drive the overflow calculation would be wrong.

Why this is safe:

- **Measurement is unaffected.** `measure()` reads `offsetWidth` of the row's children, which comes from layout, not paint. Clipping the wrapper changes what is rasterised, never the computed box sizes — the visible/hidden item split is identical before and after.
- **Dropdowns are not clipped.** The toolbar's menus (`RichEditorDropdownButton`) render through Radix portals, so they escape the wrapper's containing block and are unaffected by `overflow-hidden`.

## 🧪 Tests

- `yarn workspace @open-mercato/ui test --testPathPatterns=rich-editor` — **36 passed / 36 total, 1 suite**, run locally (no compose `app` container up, so the local runner was used). The existing suite covers the toolbar's overflow-collapsing behaviour, which is exactly the logic that had to stay intact.
- No new automated test ships with this change. The regression is a rendered-overflow property (`document.scrollWidth > clientWidth`) that jsdom does not model — it has no layout engine, so `offsetWidth` is `0` there and the measurement path this bug lives in cannot be exercised. Locking it in would need a browser-based assertion; flagging it here rather than adding a test that would pass regardless of the fix.
- Verified visually in the browser on `/backend/customers/companies/create`: the horizontal scrollbar is gone and the toolbar still collapses items into the `⋮` menu at the same widths.
- The full CI gate from `.ai/agentic.config.json` was **not** run locally for this one-line className change; CI covers it on this PR.

## 💥 Breaking Changes

- None. No public API, prop, type, or exported contract changes — the edit is a single utility class on an internal wrapper element. Every `RichEditor` consumer (`variant` `minimal` through `full`) keeps its current behaviour and appearance.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5318"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789326291&installation_model_id=432243&pr_number=5318&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5318&signature=1a592ecf70806ed2151a780b5828e9893a4e98b429a004574448ff691ce0027b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


kriss145:

⚠️ **Labels could not be applied.** The PR is opened from a fork by an account with `read` permission on this repository, and GitHub rejected every label write with `kriss145 does not have the correct permissions to execute AddLabelsToLabelable`. The set below is therefore a **request for a maintainer to apply**, not a description of the current state — the PR currently carries no labels.

- 🔍 `review` — the change is complete and ready for code review; this PR was opened non-draft and is not waiting on further work from the author.
- 🐛 `bug` — this fixes a rendering defect (a stray page-level horizontal scrollbar) rather than adding capability or restructuring code.
- 🧪 `needs-qa` — the change touches a `.tsx` file under `packages/ui/src/`, so it falls outside the automated-verification exemption in `AGENTS.md` and needs a human to confirm the visual result. The fix is also inherently visual: the symptom (page overflow) and the behaviour that must not regress (toolbar items collapsing into the `⋮` menu) are both browser-layout properties that the jsdom-based unit suite cannot assert. With `qaGate` enabled, merging waits on `qa-approved`.
- 🔹 `priority-medium` — an ordinary user-visible bug. It degrades every backend page that renders a `full`-variant `RichEditor` in a narrow column, but it breaks no data, blocks no release, and has a trivial workaround for users (ignore the scrollbar).
- 🟡 `risk-medium` — the diff is a single utility class on an internal wrapper element, with no API, prop, type, or behavioural contract touched, which argues for low risk. It lands in a shared design-system primitive used across many surfaces, though, so the blast radius is wider than the diff size suggests, and `overflow-hidden` is the kind of change that can clip a descendant that was relying on escaping its box. The toolbar's own menus render through Radix portals and are unaffected; a reviewer may still want to spot-check other `RichEditor` variants. Medium is the honest midpoint between a one-line diff and a broadly-consumed component.

kriss145:
@pkarw @adeptofvoltron Could you look into it?

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T10:21:21Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review

## 🎯 Summary

This PR fixes a real bug with an accurate diagnosis. `RichEditorAutoToolbar` renders a hidden measurement twin of the whole toolbar (`packages/ui/src/primitives/rich-editor.tsx:1545-1552`) as an `absolute` + `invisible` row, and the PR body is right on both CSS points: `visibility: hidden` suppresses painting but keeps the layout box, and an absolutely positioned box still contributes to its ancestors' scrollable overflow. With `variant="full"` that row is far wider than a `3fr` grid column, nothing on the ancestor chain clipped it, and the overflow reached the document as a page-level horizontal scrollbar. Clipping it at the component boundary is the right shape of fix, and the reasoning about why the measurement itself is unaffected is correct — `overflow` clipping is a paint/scroll operation and never changes the `offsetWidth` values that `measure()` reads.

The problem is *where* the clip was applied. The wrapper at `rich-editor.tsx:1544` has no padding and exactly one in-flow child, so its padding box coincides pixel-for-pixel with the visible toolbar's border box — which means `overflow-hidden` there also clips something the PR did not intend to touch: the toolbar's own `shadow-xs`. That makes the change a silent, product-wide visual regression on a design-system primitive, which the PR body explicitly (and understandably) claims does not happen. The remedy is small and keeps every other property of the fix intact.

**Verdict: Changes requested** — one Medium finding. Per this repository's review rules a Medium finding requests changes; this is not a rejection of the approach, which is sound.

## 🧪 Validation Gate

**Validation source: GitHub checks** (all checks for head `0307e0d1`; no local fallback was needed, and no local `validation.commands` run stands behind this review).

| Gate | Check | Result |
|------|-------|--------|
| Lint | `lint` | ✅ pass (30s) |
| Design-system lint | `ds-lint` | ✅ pass (41s) |
| Unit tests | `test` | ✅ pass (6m26s) |
| Build / prepare | `prepare` | ✅ pass (2m56s) |
| Scope | `scope`, `audit-scope` | ✅ pass |
| Integration (Playwright) | `ephemeral-integration` (15/15 shards) | ✅ pass |
| CLA | `license/cla` | ✅ signed |

No required check is failing and none is pending; `audit`, `docker-build`, `merge-coverage`, `mutate`, `comment` and `[code]smith` reported as skipped, which is their normal state for this change. The head is `MERGEABLE` against `develop` with no conflicts. `mergeStateStatus` is `BLOCKED` because the review and QA gates are still outstanding, not because of CI.

## 🟠 Medium (minor)

### 1. The wrapper clip also removes the toolbar's `shadow-xs` on every `RichEditor`

`packages/ui/src/primitives/rich-editor.tsx:1544`

`overflow: hidden` clips descendant painting to the clipping element's **padding box**. The wrapper here carries no padding, and its only in-flow child is `RichEditorToolbar` rendered with `w-full` (`rich-editor.tsx:1553`), which `tailwind-merge` resolves over the primitive's own `w-fit max-w-full`. The wrapper's height therefore comes entirely from that toolbar — the measurement row is `absolute` and out of flow — so the clip rectangle and the toolbar's border box are the same rectangle.

`RichEditorToolbar` paints a drop shadow outside that border box:

```tsx
// rich-editor.tsx:656-658
// Figma 166331:4006 reference: standalone card with rounded-8 + border + shadow-xs.
'flex w-fit max-w-full flex-wrap items-center gap-0.5 overflow-x-auto rounded-lg border border-border bg-card p-0.5 shadow-xs'
```

`--shadow-xs` is `0 1px 2px rgb(16 24 40 / 0.05)` in light mode and `0 1px 2px rgb(0 0 0 / 0.30)` in dark mode (`apps/mercato/src/app/globals.css:185` and `:350`). Since the shadow extends 1–3px beyond the toolbar's border box on every side and the wrapper now clips exactly at that box, the shadow is removed entirely — on all non-`custom` variants, because `RichEditorAutoToolbar` is the single toolbar path for `minimal` through `full` (`rich-editor.tsx:521-525`). In dark mode, where the shadow alpha is `0.30` rather than `0.05`, the loss is clearly visible.

This is a design-system regression rather than a functional one, but it is product-wide, it silently drops a property whose Figma reference is spelled out in a comment two lines above it, and it contradicts the PR body's "Every `RichEditor` consumer … keeps its current behaviour and appearance."

**Suggested fix** — clip only the measurement twin, in its own layer, so the visible toolbar keeps painting outside its box:

```tsx
<div className="relative w-full">
  <div aria-hidden="true" className="pointer-events-none absolute inset-0 overflow-hidden">
    <div
      ref={measureRef}
      className="invisible absolute left-0 top-0 flex flex-row flex-nowrap items-center gap-0.5"
    >
      {items.map(renderToolbarItem)}
      <RichEditorDropdownButton icon={<MoreVertical />} ariaLabel={labels.more} showChevron={false} menu={<div />} />
    </div>
  </div>
  <RichEditorToolbar ref={containerRef} className="w-full flex-nowrap overflow-hidden">
```

The clipping layer is sized to the wrapper's padding box by `inset-0`, and the measurement row inside it stays absolutely positioned and shrink-to-fit, so it still reaches its full intrinsic width and `measure()` keeps reading the same `offsetWidth` values off `measureEl.children`. The original page-overflow bug stays fixed, because the row's overflow is now contained by the clipping layer instead of propagating to the document. The visible toolbar is a sibling of that layer rather than a descendant, so its `shadow-xs` is untouched.

Adding padding to the existing wrapper instead would also free the shadow, but it would shrink the `w-full` toolbar relative to the editor body below it and misalign the two, so the nested clip layer is the cleaner option.

## 🔵 Low (nit)

### 2. No browser-level regression test locks the fix in

The PR's argument for shipping without a new automated test is fair and honestly stated: jsdom has no layout engine, `offsetWidth` is `0` there, and the measurement path this bug lives in genuinely cannot be exercised — a jsdom test would pass with or without the fix. The existing `rich-editor` suite (36 tests) covers the overflow-collapsing logic and passed in CI's `test` job, which is the right thing to have verified.

That said, this repository does run Playwright in CI across 15 `ephemeral-integration` shards, so a browser-level assertion is available where a jsdom one is not. A single check on a page that renders a `full`-variant `RichEditor` in a narrow column — `document.documentElement.scrollWidth <= document.documentElement.clientWidth` on `/backend/customers/companies/create` — would pin this regression permanently and would also have caught the shadow-clipping side effect's cousin class of layout mistakes. Worth adding here or as a follow-up; not a blocker, and it does not change the verdict.

## 💥 Breaking Changes

None. `RichEditorAutoToolbar` is module-private — the file's only export line is `export { richEditorItemVariants }` (`rich-editor.tsx:1578`), and `RichEditorToolbar` itself is unchanged. No prop, type, exported symbol, import path, event ID, widget spot, API route, DB schema, DI key, ACL feature or generated file is touched, so no `BACKWARD_COMPATIBILITY.md` contract surface is affected and no deprecation bridge is required. The only cross-consumer effect is the visual one described in finding 1, which is why it is worth correcting before merge rather than after.

## 🧪 Test Coverage

Adequate for the change as submitted, with the caveat in finding 2. CI's `test` job passed, including the `rich-editor` suite that guards the visible/hidden item split this fix must not disturb; the 15 Playwright integration shards also passed, so nothing downstream regressed. The author verified the fix manually in the browser on `/backend/customers/companies/create` and confirmed both that the scrollbar is gone and that items still collapse into the `⋮` menu at the same widths. Because the change alters rendered layout in a shared UI primitive, `needs-qa` remains the right gate — a human should confirm the toolbar's appearance across the `minimal`, `default` and `full` variants in both light and dark mode once finding 1 is addressed.

## 🔁 Inherited feedback

None outstanding. The PR carries no prior reviews and no inline comments. The two existing bot comments are a Dokploy preview-deployment block (a fork-permission artifact, not an actionable code issue) and the author's own label-rationale comment noting that fork `read` permission prevented applying labels — those labels have now been applied by this review.

pkarw:

Applying the label set @kriss145 requested in the earlier rationale comment — the fork account's `read` permission could not write labels, so this review applied them, with one change: the pipeline label is `changes-requested` rather than `review`, because the review found an actionable item.

- ❌ `changes-requested` — the review found one actionable issue: the `overflow-hidden` clip lands on a wrapper whose padding box coincides with the toolbar's border box, so it also clips the toolbar's Figma-specified `shadow-xs` on every `RichEditor` variant.
- 🐛 `bug` — this fixes a rendering defect (a stray page-level horizontal scrollbar) rather than adding capability or restructuring code.
- 🧪 `needs-qa` — the change edits a `.tsx` file under `packages/ui/src/` and alters rendered layout in a shared primitive, so it falls outside the automated-verification exemption in `AGENTS.md` and needs a human to confirm the visual result across variants and colour schemes.
- 🔹 `priority-medium` — an ordinary user-visible bug that degrades every backend page rendering a `full`-variant `RichEditor` in a narrow column, but breaks no data and blocks no release.
- 🟡 `risk-medium` — the diff is a single utility class with no API, prop or type change, but it lands in a broadly-consumed design-system primitive where a clip has effects beyond the element it is written on, which is exactly what the review found.

pkarw:
Thanks @kriss145 — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pkarw:

Reviewed head `0307e0d1` in an isolated worktree. The root-cause analysis in the PR body is correct and the fix works; one Medium finding stands: applying `overflow-hidden` to the wrapper also clips the toolbar's Figma-specified `shadow-xs` on every `RichEditor` variant, because the wrapper has no padding and its padding box coincides with the toolbar's border box. The review suggests a nested clip layer around the measurement row, which fixes the page overflow without touching the visible toolbar's painting. Plus one nit on adding a Playwright-level regression assertion.

Validation source: GitHub checks — every reported check on this head passes (`lint`, `ds-lint`, `test`, `prepare`, `scope`, `audit-scope`, 15/15 `ephemeral-integration` shards, CLA signed); no check is pending and no local fallback gate was run. Labels applied on the author's behalf, since the fork account lacks label-write permission. autofix: skipped (not my PR — re-run with --autofix to fix it here). No CI follow-up is pending.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-17T10:32:00Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

Claim note for transparency: this PR carried no `in-progress` label and no live automation claim — the only ownership signal was the assignee, which is @kriss145 as the PR's author after the review handback earlier today, and it is older than the 30-minute claim window with no activity since. That is an authorship signal rather than an automation lock, so this run claims the PR additively (it does not unassign @kriss145) and will release its own claim when the QA evidence is posted. The run is strictly read-only on source: it checks the PR head out in an isolated worktree, drives the app in a browser, and posts screenshots — it will not edit, commit, or push anything to `feat/rich-editor-toolbar-overflow`.

pkarw:
## 📸 UI QA evidence — ❌ FAIL (the reported bug is fixed; one unintended visual regression is confirmed)

**Verdict:** ❌ **FAIL** — not because the fix does not work. It does: the page-level horizontal scrollbar on `/backend/customers/companies/create` is gone, the toolbar still collapses correctly, and typing and formatting are unaffected. The run fails because it also **measured** the regression the code review on this PR flagged as its Medium finding — `overflow-hidden` on the wrapper clips the toolbar's `shadow-xs` on every `RichEditor` instance — so QA cannot sign the change off in its current shape. The remedy the review already proposes (a nested `inset-0 overflow-hidden` clip layer around the measurement row only) would make this run green.

**Environment:** `http://127.0.0.1:40995` · role `admin` (seeded demo tenant *Acme Corp*) · browser `playwright` (Chromium) — an ephemeral **production** build (`next start`, `NODE_ENV=production`) on a disposable testcontainers Postgres, booted by this run in an isolated worktree and torn down afterwards.
**Verified:** `feat/rich-editor-toolbar-overflow` @ `0307e0d1`

### Scenario (P1 — backend UI, shared `RichEditor` primitive)
**Where to click:** `/backend/customers/companies/create` → the *Executive notes* field in the narrow right-hand *Custom Attributes* column (a `full`-variant `RichEditor`, 22 toolbar items in an 827px measurement row rendered into a 307px column).

| 1 | Load the create form at 1440×900 and measure document overflow | `scrollWidth == clientWidth`; the page cannot scroll sideways | `scrollWidth 1440 == clientWidth 1440` (0px overflow); `scrollTo(9999,0)` left `scrollX` at `0`, so the sidebar stays put | ✅ |
| 2 | A/B in the live page: set the wrapper's `overflow` back to `visible`, re-measure, restore | Without the clip the reported overflow returns; with it, it disappears | Clip **off**: `scrollWidth 1920` vs `clientWidth 1440` → **480px of dead scroll space**, page scrolled 480px right with the sidebar off-screen. Clip **on** again: 0px. This is the bug, and this one class is what fixes it | ✅ |
| 3 | Check the toolbar still collapses and the `⋮` menu is not clipped | Fewer visible items than the measurement row holds; the menu paints outside the clipped wrapper | 5 visible buttons (Heading, Font size, Text color, Bold, More) against 22 in the measurement row; the `⋮` menu opened at 280×106 **below** the toolbar's box, unclipped (Radix portal, as the PR body predicted) | ✅ |
| 4 | Capture the toolbar region with the clip on and off, light and dark, and diff the pixels | Pixel-identical, if only the hidden measurement row is affected | **Light mode: 2 496 of 70 416 pixels differ (3.55%), max channel delta 9/255**, and the amplified diff traces the toolbar's outline — that is `shadow-xs`. The wrapper has `padding: 0px` and its box coincides exactly with the toolbar's border box, so the clip rectangle *is* the toolbar's border box | ❌ |
| 5 | Mobile smoke at 390×844 | No page-level horizontal overflow | `scrollWidth 390 == clientWidth 390`; toolbar collapsed to 6 visible buttons | ✅ |
| 6 | Type into the editor, apply **Bold** from the toolbar, then **Italic** from the `⋮` menu | Both commands apply | `<span style="font-weight: bold;">QA 5318 formatting check</span>`, then `font-weight: bold; font-style: italic` after using the 17-item overflow menu | ✅ |
| 7 | Open a seeded company detail page | No overflow on the neighbouring surface | 0px overflow; the detail page renders **no** `RichEditor` toolbar, so it exercises the surrounding layout rather than a second editor | ✅ |
| 8 | Look for a second `full`-variant surface (staff `bio`) | A second editor instance to cross-check | `/backend/staff/team-members/create` renders no `RichEditor` in the seeded data (0px overflow). No second surface was reachable | ⚠️ not exercised |

### The one finding — `shadow-xs` is clipped away

The clip lands on a wrapper with no padding whose only in-flow child is the `w-full` toolbar, so the clip rectangle and the toolbar's border box are the same rectangle, and a drop shadow paints *outside* that box. The A/B capture makes it visible: the two toolbar crops look identical to the eye (a `0 1px 2px rgb(16 24 40 / 0.05)` shadow is faint by design), but 3.55% of the pixels differ, and amplifying that difference ×20 draws the shadow ring exactly where the toolbar's edges are — strongest along the bottom, where a `0 1px` offset shadow lives.

**One correction to the review, in the review's favour on scope but against it on severity:** in dark mode the loss is **not** more visible — this run measured **zero** differing pixels there. With `.dark` on `<html>`, the toolbar's computed `box-shadow` is still `rgba(16, 24, 40, 0.05)`, not the `rgb(0 0 0 / 0.30)` dark token. `--shadow-xs` is declared inside `@theme inline` (`apps/mercato/src/app/globals.css:185`), so Tailwind v4 inlines the light value into the `shadow-xs` utility and the `.dark` override at `globals.css:350` never reaches it. That is a **pre-existing platform issue unrelated to this PR** (every `shadow-*` utility is affected), but it does mean the regression this PR introduces is a subtle light-mode-only difference rather than a clearly visible dark-mode one.

### Screenshots

Order: the shipped state, the bug reproduced by removing the clip, the unclipped overflow menu, the shadow A/B pair with its amplified diff, the mobile viewport, and the editor after formatting.


**step-01-companies-create-desktop.png**



**step-03-prefix-overflow-repro.png**



**step-04-toolbar-overflow-menu.png**



**step-05-shadow-light-clip-on.png**



**step-06-shadow-light-clip-off.png**



**step-09-shadow-light-diff-amplified.png**



**step-07-mobile-390.png**



**step-10-editor-typing-bold.png**

pkarw:

This change ships no browser-level test — and the PR body is right that a jsdom test cannot cover it, because jsdom has no layout engine and reports `offsetWidth` as `0`, so the measurement path this bug lives in never runs there. The QA above was therefore manual. This repository does run Playwright in CI across 15 `ephemeral-integration` shards, so the assertion is available where a jsdom one is not; the scenario below is what this run actually executed and is ready to implement with `/om-integration-tests`.

1. **Setup:** log in as `admin` via the repo's `login(page, 'admin')` helper (`packages/core/src/helpers/integration/auth.ts`). No fixtures are needed — the surface is the *create* form, and the `executive_notes` custom field is a seeded module default (`packages/core/src/modules/customers/customFieldDefaults.ts:33`), which `CrudForm` maps to `<RichEditor variant="full" />`.
2. **Act:** navigate to `/backend/customers/companies/create` at a 1440×900 viewport and wait for `[data-slot="rich-editor-toolbar"]` to be visible, plus a short settle so the toolbar's `ResizeObserver` has run its measure pass.
3. **Assert:**
   - `document.documentElement.scrollWidth <= document.documentElement.clientWidth` — the regression guard proper. Without the fix this run measured 1920 vs 1440.
   - `window.scrollTo(9999, 0)` leaves `window.scrollX === 0`, which is the user-visible symptom (the sidebar sliding out of view) rather than a proxy for it.
   - The toolbar still collapses: fewer buttons inside `[data-slot="rich-editor-toolbar"]` than inside the `aria-hidden` measurement row, and a `[aria-label="More"]` button present. This is the assertion that would catch a future "fix" that clipped or shrank the measurement row itself and silently broke the item split.
   - Optionally repeat the overflow assertion at 390×844 so the phone viewport is covered too.
4. **Teardown:** none required — the scenario creates no records and never submits the form.

**Worth adding alongside it:** a guard that the toolbar keeps painting its `shadow-xs`. The cheapest form is a DOM-level check that the clipping ancestor is not the toolbar's own box — for example, assert that the element carrying `overflow: hidden` above the toolbar either does not exist or does not share the toolbar's border box. That is exactly the property the Medium review finding is about, and it is the one thing a `scrollWidth` assertion alone would not have caught.

pkarw:
🤖 `om-auto-qa-pr` completed: FAIL. Evidence posted above. Lock released.

Drove head `0307e0d1` in a browser against an ephemeral production build in an isolated worktree, then tore the environment down. The verdict is FAIL only because the run confirmed the review's Medium finding by measurement — the wrapper clip removes the toolbar's `shadow-xs` on every `RichEditor` instance (3.55% of the toolbar region's pixels change when the clip is toggled off, and the amplified diff draws the shadow ring). Everything else passed: the reported page-level horizontal scrollbar is genuinely gone, removing the clip in the live page brings back exactly the reported 480px of dead scroll space, the toolbar still collapses to the `⋮` menu at the same widths, the menu is not clipped, formatting still works from both the visible toolbar and the overflow menu, and the phone viewport is clean.

Labels: unchanged. This run used the evidence-only default, so no pipeline or meta label was touched — no `qa-failed` was applied, because automated UI checks warrant reporting rather than blocking, and the PR is already in `changes-requested` for the same finding. The `in-progress` label and this run's assignee claim are removed; @kriss145 remains the assignee and owns the next pass.

One correction to the review, recorded in the evidence comment: the dark-mode claim does not hold on this build. With `.dark` applied, the toolbar's computed `box-shadow` is still `rgba(16, 24, 40, 0.05)` and the pixel diff is zero, because `--shadow-xs` lives inside `@theme inline` and Tailwind v4 inlines the light value into the utility, so the `.dark` override never reaches it. That is a pre-existing platform issue affecting every `shadow-*` utility, not something this PR causes — but it means the regression here is a subtle light-mode-only difference rather than a clearly visible dark-mode one.

pkarw:
@kriss145 please address the medium ranked issues from the CR so we can pass the QA and code review test and merge this change in

pkarw:
🤖 `om-auto-review-pr` (autofix mode) started by @pkarw at 2026-08-24T01:44:37Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Re-review (autofix applied)

## 🎯 Summary

Re-review of head `99af6135` after `om-auto-review-pr` ran in autofix mode. Both findings from the [first review](https://github.com/open-mercato/open-mercato/pull/5318#pullrequestreview) and the [failing UI QA run](https://github.com/open-mercato/open-mercato/pull/5318#issuecomment-5344836896) are resolved in-branch; the original commit and @kriss145's authorship are preserved, and the fixes land as one follow-up commit plus a merge of the current `develop`.

**Verdict: Approved** — no blocker, major, or medium findings remain. The QA gate still holds the merge (`needs-qa`, no `qa-approved`); a re-run of `om-auto-qa-pr` against this head is in progress and its evidence will be posted separately.

## ✅ What changed since `0307e0d1`

### 1. Medium finding — `shadow-xs` clipped on every `RichEditor` — **fixed**

`packages/ui/src/primitives/rich-editor.tsx`

The wrapper is back to `relative w-full`, and the clip moved into a dedicated layer that wraps the measurement row only:

```tsx
<div className="relative w-full">
  <div
    aria-hidden="true"
    data-slot="rich-editor-toolbar-measure-clip"
    className="pointer-events-none absolute inset-0 overflow-hidden"
  >
    <div ref={measureRef} className="invisible absolute left-0 top-0 flex flex-row flex-nowrap items-center gap-0.5">
```

Why each property of the original fix survives:

- **The page-overflow bug stays fixed.** The measurement row's overflow is now contained by the clip layer instead of propagating to the document.
- **`measure()` still reads the same widths.** The clip layer is `absolute inset-0`, so its box is the wrapper's padding box — the *same* containing block the measurement row shrink-to-fits against as before (the wrapper has no padding). `overflow` clipping is a paint/scroll operation and never changes `offsetWidth`, so the readings off `measureEl.children` are byte-identical to the pre-fix ones.
- **The shadow is back.** `RichEditorToolbar` is now a *sibling* of the clip layer rather than a descendant, so nothing clips at its border box and `shadow-xs` paints again.
- **No new interaction surface.** `pointer-events-none` and `aria-hidden` moved up to the clip layer, so they still cover the whole duplicated subtree; the layer paints above the toolbar but every descendant is `visibility: hidden`.

### 2. Low finding — no browser-level regression test — **addressed**

`packages/core/src/modules/customers/__integration__/TC-CRM-5318-rich-editor-toolbar-overflow.spec.ts` (new, Playwright) implements the scenario the QA follow-up specified, on `/backend/customers/companies/create`:

- `documentElement.scrollWidth <= clientWidth` at 1440×900 and 390×844, plus `scrollTo(9999, 0)` leaving `scrollX === 0` — the user-visible symptom, not a proxy for it.
- The toolbar still collapses: strictly fewer buttons in the visible toolbar than in the measurement row, and a `More` trigger present. This is what would catch a future "fix" that shrank the measurement row and silently broke the item split.
- No ancestor sharing the toolbar's box clips it, and its computed `box-shadow` is not `none` — the guard for exactly the regression this re-review closes. No fixtures, no submit, nothing to clean up.

`packages/ui/src/primitives/__tests__/rich-editor.test.tsx` adds three jsdom tests pinning the containment structure the fix depends on. They deliberately assert structure rather than layout: jsdom reports `offsetWidth` as `0`, so it cannot reproduce the bug itself — the honest split is structure here, pixels in Playwright.

## 🧪 Validation Gate

**Validation source: local fallback** — CI has no results for `99af6135` yet, and the changed code is new since the last green run, so GitHub check data could not stand behind this pass. Run in the isolated review worktree (local runner, not Docker):

| Gate | Command | Result |
|------|---------|--------|
| Package build | `yarn build:packages` | ✅ pass |
| Generators | `yarn generate` | ✅ pass (no tracked-file drift) |
| Typecheck | `yarn workspace @open-mercato/ui typecheck` | ✅ pass |
| Typecheck | `yarn workspace @open-mercato/core typecheck` | ✅ pass (covers the new spec) |
| Unit tests | `packages/ui` full jest suite | ✅ **233 suites / 1920 tests pass** |
| Lint | `eslint` on the three changed files | ✅ pass |
| Design-system lint | `eslint --config eslint.ds.config.mjs` on the changed UI files | ✅ pass |

Scope note: `i18n:check-*`, `build:app` and template-sync were not re-run locally — the diff adds no user-facing strings, touches nothing under `apps/mercato/src/`, and changes no app-level route, so those gates have no changed input. CI will run the full ordered gate on the pushed head.

The branch was also merged up to `develop@58862d0c` (clean, no conflicts) before validation, so CI's affected-modules diff stays narrow instead of pulling all 15 integration shards.

## 💥 Breaking Changes

None, unchanged from the first review. `RichEditorAutoToolbar` is module-private, no exported symbol, prop, type, import path, event ID, widget spot, API route, DB schema, DI key, ACL feature or generated file is touched, and no `BACKWARD_COMPATIBILITY.md` surface is affected. The new `data-slot="rich-editor-toolbar-measure-clip"` is additive and used only as a test hook.

## 🧪 Test Coverage

Now adequate without caveat. The change ships both a structural unit guard and a browser-level regression guard, and the browser guard covers the *shadow* property as well as the *overflow* property — the two things this PR's two review passes were about.

## 📝 One thing deliberately left out of scope

The QA run's correction stands and is worth recording: in dark mode the shadow loss measured **zero** differing pixels, because `--shadow-xs` is declared inside `@theme inline` (`apps/mercato/src/app/globals.css:185`) and Tailwind v4 inlines the light value into the `shadow-xs` utility, so the `.dark` override at `globals.css:350` never reaches it. That is a pre-existing platform issue affecting every `shadow-*` utility and is not this PR's to fix — flagging it here so it does not get lost.

pkarw:
Label set to `merge-queue` because code review passed on head `99af6135`; `needs-qa` stays on so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Priority stays `priority-medium` and risk stays `risk-medium` — the autofix narrowed the blast radius rather than widening it.

pkarw:

An automated re-run of `om-auto-qa-pr` against head `99af6135` is in progress and its evidence will be posted in this thread — check it before repeating the whole script.

### P1 — `RichEditor` toolbar layout (shared UI primitive, product-wide)
**Where to click**
- `/backend/customers/companies/create` → the *Executive notes* field in the narrow right-hand *Custom Attributes* column (a `full`-variant `RichEditor`)
- Any other backend form hosting a `RichEditor`, at the `minimal`, `default` and `full` variants

**What to verify**
- The page has **no** horizontal scrollbar and cannot be scrolled sideways — the left sidebar must stay put when you try. This is the original bug.
- The toolbar still collapses: only the buttons that fit are shown, the rest live behind `⋮`, and opening `⋮` renders the menu **outside** the toolbar box without being clipped.
- The toolbar still has its subtle drop shadow (`shadow-xs`) on all four edges — strongest along the bottom. This is the regression the previous QA run caught and this revision fixes; compare against `develop` side by side if in doubt.
- Typing works, and formatting applies both from a visible button (**Bold**) and from inside the `⋮` menu (**Italic**).
- Repeat the overflow check at a 390×844 mobile viewport.

**What can go wrong**
- The shadow is clipped flat against the toolbar's edges (the previous failure mode) — subtle in light mode, so look at the bottom edge specifically.
- The item split changes: too few or too many buttons visible, or a duplicated row of buttons becoming visible (the hidden measurement twin escaping its clip layer).
- The `⋮` menu gets cut off at the toolbar boundary instead of overlaying the content below.
- Dark mode: note that `shadow-xs` renders with the light-mode value there because of a **pre-existing** Tailwind v4 `@theme inline` issue unrelated to this PR. Do not fail the PR for that.

### Pass/fail

pkarw:
## 📸 UI QA evidence — ✅ PASS (re-run after autofix)

**Verdict:** ✅ **PASS** — 12 of 12 steps green, 0 failures. The [previous run](https://github.com/open-mercato/open-mercato/pull/5318#issuecomment-5344836896) failed on one thing only: the wrapper-level clip removed the toolbar's `shadow-xs`. That is fixed, and this run measured the fix rather than eyeballing it. The original page-overflow bug is still fixed.

**Environment:** `http://127.0.0.1:39895` · role `admin` (seeded demo tenant *Acme Corp*) · browser `playwright` (Chromium) — an ephemeral **production** build (`next start`, `NODE_ENV=production`) on a disposable testcontainers Postgres, booted by this run in an isolated worktree and torn down afterwards.
**Verified:** `feat/rich-editor-toolbar-overflow` @ `99af6135` (autofix commit + `develop@58862d0c` merged in).

### Scenario (P1 — backend UI, shared `RichEditor` primitive)
**Where:** `/backend/customers/companies/create` → *Executive notes* in the narrow right-hand *Custom Attributes* column (a `full`-variant `RichEditor`: an 829px measurement row rendered into a 307px column).

| 1 | Load at 1440×900 and measure document overflow | `scrollWidth == clientWidth` | `scrollWidth 1440 == clientWidth 1440` | ✅ |
| 2 | `scrollTo(9999, 0)` | the page cannot scroll sideways | `scrollX 0` — the sidebar stays put | ✅ |
| 3 | A/B the new clip layer: set it to `overflow: visible`, re-measure, restore | the bug returns without it and disappears with it | clip **off**: `scrollWidth 1921` vs `clientWidth 1440` → **481px of dead scroll space**. Clip **on**: 0px. The clip layer is what fixes the bug | ✅ |
| 4 | Confirm the measurement row still reaches its full intrinsic width inside the clip layer | `offsetWidth` readings unchanged by the clip | row **829px** inside a **307px** column, 31 measured items — the clip is a paint operation, the measurement is untouched | ✅ |
| 5 | Check the toolbar still collapses | fewer visible items than measured | **5 visible buttons vs 22 measured** — same split as the previous run | ✅ |
| 6 | Open the `⋮` menu | renders outside the toolbar box, unclipped | menu **280×106** below the toolbar (Radix portal) | ✅ |
| 7 | Walk the toolbar's ancestors for a clipping box that coincides with its own | none — otherwise the shadow is swallowed | `clippingAncestor: null`, wrapper `overflow: visible`, `box-shadow: rgba(16, 24, 40, 0.05) 0px 1px 2px 0px` | ✅ |
| 8 | A/B pixel proof: shipped state vs. the previous revision's clip (`overflow: hidden` back on the wrapper) | the two must differ — the difference *is* the shadow | **679 of 16 150 pixels differ (4.20%), max channel delta 9/255**, and the amplified diff traces the toolbar outline. Previously that difference was *missing*; now it is present | ✅ |
| 9 | Dark mode | no overflow, shadow declaration intact | `scrollWidth 1440 == 1440`; `box-shadow` present | ✅ |
| 10 | Type, apply **Bold** from the toolbar and **Italic** from the `⋮` menu | both commands apply | `font-weight: bold; font-style: italic` | ✅ |
| 11 | Mobile smoke at 390×844 | no page-level horizontal overflow | `scrollWidth 390 == clientWidth 390`; 6 visible buttons | ✅ |
| 12 | Run the PR's new committed Playwright spec against this env | all pass | `TC-CRM-5318` — **3/3 passed** | ✅ |

### The new regression tests actually discriminate

A regression test that passes on the broken code is worth nothing, so this run re-created both prior states live in the page and re-evaluated the committed spec's own predicates against them:

| State | `clippedByCoincidentAncestor` | document overflow | Committed spec would… |
|---|---|---|---|
| **Shipped (`99af6135`)** | `false` | `1440 == 1440` | **pass** |
| Previous revision (`0307e0d1`) | **`true`** | `1440 == 1440` | **fail** the shadow test |
| Original bug (`develop`) | `false` | **`1921` vs `1440`** | **fail** the overflow test |

So `TC-CRM-5318` pins both regressions, not just the one the PR set out to fix.

### One note carried forward (not this PR's problem)

The previous run's correction still holds: in dark mode the toolbar's computed `box-shadow` is `rgba(16, 24, 40, 0.05)` — the **light** value — because `--shadow-xs` is declared inside `@theme inline` (`apps/mercato/src/app/globals.css:185`) and Tailwind v4 inlines it into the utility, so the `.dark` override at `:350` never reaches it. That affects every `shadow-*` utility and is unrelated to this PR. Recorded so it does not get lost.

### Screenshots

Order: the shipped state, the unclipped `⋮` menu, the shadow A/B pair with its amplified diff, dark mode, the editor after formatting, and the mobile viewport.

**step-01-companies-create-desktop.png**



**step-02-toolbar-overflow-menu.png**



**step-03-shadow-shipped.png** — the toolbar as shipped, cropped with an 8px margin



**step-04-shadow-old-regression.png** — the same crop with the previous revision's clip re-applied



**step-05-shadow-diff-amplified.png** — their difference ×20: the shadow ring the previous revision removed



**step-06-dark-mode.png**



**step-07-editor-typing-bold.png**



**step-08-mobile-390.png**

pkarw:
🤖 `om-auto-review-pr` (autofix mode) completed: **APPROVED**. Lock released.

Two autofix iterations on head `0307e0d1` → `99af6135`:
1. Resolved the Medium code-review finding and the failing UI QA run (same root cause) by moving the clip off the toolbar wrapper into a dedicated layer around the measurement row only, so `shadow-xs` paints again while the page-overflow fix is unchanged.
2. Closed the Low finding by adding a Playwright regression spec (`TC-CRM-5318`) plus three jsdom structural tests, then merged `develop@58862d0c` in.

Pushed as a normal follow-up commit to the fork branch (`maintainerCanModify`) rather than a carry-forward PR, so this PR and @kriss145's authorship stay intact. No force-push.

`om-auto-qa-pr` was re-run against `99af6135` and came back **PASS** (12/12) — evidence above. Labels are unchanged by that run (evidence-only default), so the PR still needs `qa-approved` from a QA reviewer, or a self-QA sign-off, before the QA gate clears.

`ci-monitoring` applied: `lint`, `ds-lint`, `test`, `prepare`, `scope`, `audit-scope` and `documents-multi-instance` are green on this head; the 15 `ephemeral-integration` shards are still running.

pkarw:
✅ CI is fully green on `99af6135` — **23 checks pass, 0 fail** (6 skipped as normal for this change), including all 15 `ephemeral-integration` shards. The shard covering `customers` ran the new `TC-CRM-5318` spec added by this revision.

`mergeStateStatus` is now `CLEAN` and `reviewDecision` is `APPROVED`. Removing `ci-monitoring`.

The only thing still holding this PR is the QA-approval gate: it carries `needs-qa` without `qa-approved`. The re-run above is a green automated UI QA pass with screenshots, but that run was evidence-only by default — a QA reviewer (or a maintainer applying the self-QA exception) still needs to add `qa-approved`.

pkarw:
## 📸 UI QA evidence — ✅ PASS (post-merge verification)

**Verdict:** ✅ **PASS** — 11 of 11 scenario steps green, 0 failures.
**Environment:** `http://127.0.0.1:43623` · role `admin` · browser `playwright` (chromium, Playwright 1.62.1) · production build (`next start`) against a disposable testcontainers pgvector database.
**Verified:** merge commit `19c2a8bbe` — the state of this PR **as it landed on `develop`**, not the pre-merge head.

> ℹ️ This PR was already merged at `2026-08-24T03:09:34Z` before this QA run started, so this is a **post-merge verification** of the shipped code. It neither gates nor blocks anything; it is evidence that the fix behaves correctly in the branch that now ships.

### Scenario (P1 — RichEditor toolbar containment on a narrow form column)

**Where to click:** `/backend/customers/companies/create` — the seeded `executive_notes` custom field renders `<RichEditor variant="full" />` inside the `3fr` column of the form's `lg:grid-cols-[7fr_3fr]` layout, which is the narrowest place the full toolbar has to fit (measured at **307px** wide).

| 1 | Cold-load the create form at 1440×900 as `admin` and wait for the full-variant toolbar | the form renders and the RichEditor toolbar becomes visible | toolbar visible after 1159ms; 1 RichEditor toolbar on the page | ✅ |
| 2 | **The reported defect** — measure document overflow and try to scroll the page sideways | `scrollWidth <= clientWidth`, and `window.scrollX` stays `0` after `scrollTo(9999, 0)` | `scrollWidth=1440px`, `clientWidth=1440px` (delta **0px**); `window.scrollX` after `scrollTo(9999,0)` = **0** | ✅ |
| 3 | Check the app sidebar still sits inside the viewport (the overflow used to push it off-screen) | the sidebar box is fully within the 1440px viewport | sidebar box `left=0px right=240px width=240px` in a 1440px viewport | ✅ |
| 4 | Inspect the hidden measurement row and its new clipping layer in the live DOM | the twin is `aria-hidden`, `visibility:hidden`, nested inside an `absolute` `overflow-hidden` clip layer | clip: `position=absolute overflowX=hidden aria-hidden=true width=307px`; measurement row: `visibility=hidden`, 22 items, **intrinsic width 829px** — 522px wider than its 307px clip, which is exactly what used to leak into the document | ✅ |
| 5 | Confirm the toolbar still collapses overflowing items into the ⋮ More menu, and open it | more items measured than shown, More button visible, clicking it opens a populated overflow popover | measurement row = 22 buttons, visible toolbar = 5, More visible = true, popover opens with **17 entries** (Italic, Underline, Strikethrough, Bullet list, Numbered list, Horizontal rule, Quote, Inline code, …). 5 visible + 17 overflowed reconciles exactly with the 22 measured items | ✅ |
| 6 | Verify the new clip does **not** clip the visible toolbar (its `shadow-xs` paints outside the border box) | non-`none` box-shadow, toolbar outside the clip layer, no clipping ancestor sharing its box | computed box-shadow ends in `rgba(16, 24, 40, 0.05) 0px 1px 2px 0px`; toolbar inside clip layer = **false**; coincident clipping ancestor = **none** | ✅ |
| 7 | **Negative control** — temporarily undo the clip in the live DOM to prove it is what suppresses the overflow | disabling the clip brings the overflow back; restoring it removes the overflow again | with the fix `1440/1440px` → clip disabled `1921/1440px` (**+481px of horizontal overflow returns**) → clip restored `1440/1440px` | ✅ |
| 8 | Type into the editor and apply Bold from the toolbar (it must still *work*, not merely fit) | text lands in the editor and Bold marks the selection, with the button reflecting the active state | editor text `"Post-merge QA check for PR 5318."`; `aria-pressed` **false → true**; markup `<span style="font-weight: bold;">…</span>` | ✅ |
| 9 | Smoke the same surface in dark mode | toolbar renders and the document still does not overflow | dark-mode `scrollWidth=1440px`, `clientWidth=1440px` | ✅ |
| 10 | Reload at a 390×844 mobile viewport, where the column is narrowest | no horizontal overflow, toolbar still renders usable items | `scrollWidth=390px`, `clientWidth=390px`, `scrollX=0`, 6 visible toolbar buttons | ✅ |
| 11 | Measure perceived performance of a cold load of the changed route | fast response, no long blank period before the editor appears | `DOMContentLoaded` at **263ms**, toolbar visible at **1235ms** after navigation start | ✅ |

### Screenshots

Step 1 — companies create form, desktop 1440×900
*The `Executive notes` RichEditor in the narrow right-hand column. The sidebar is intact and the page does not scroll sideways.*

Step 5 — overflow popover open
*The ⋮ More button still collapses the 17 items that do not fit into a popover — the measure-and-collapse behaviour survived the containment change.*

Step 8 — typing and Bold
*The toolbar is functional, not just correctly sized: Bold applies to the selection and the button reports the active state.*

Step 9 — dark mode

Step 10 — mobile 390×844
*At the narrowest viewport the form stacks to one column and the toolbar still fits without introducing horizontal scroll.*

### Notes for QA

- **The strongest single piece of evidence is step 7.** Rather than only asserting "there is no overflow now", the run disabled the new clip layer in the live DOM and watched **481px of horizontal overflow come straight back**, then restored it and watched the overflow disappear. That establishes causation: this specific change is what holds the measurement twin in, and the page is not merely passing by coincidence of some other layout.
- **The measurement twin is genuinely oversized**, as expected — 829px of intrinsic content inside a 307px column. The fix does not shrink it; it contains it, which is the right call because the twin must stay full-width to measure how many items fit.
- **The `shadow-xs` regression that the earlier review flagged is not present.** The visible toolbar sits outside the clip and has no clipping ancestor sharing its box, so its drop shadow still paints.
- **Coverage limits, stated honestly:** this run exercised the companies create form only, which is the surface the change targets. Other `variant="full"` RichEditor placements (deal notes, portal surfaces) were not visited. Permission-restricted and error/empty states were not exercised, since the change is purely presentational and does not read or write data.
- **Pre-existing, out of scope, worth a separate look:** the editor marks bold with `<span style="font-weight: bold">` rather than a semantic `<strong>`, which is weaker for assistive technology. This PR does not touch that behaviour and did not change it — flagging it only because the QA pass observed it directly.
- **Automated coverage already exists**, so no follow-up UI test is being requested: this PR ships `packages/core/src/modules/customers/__integration__/TC-CRM-5318-rich-editor-toolbar-overflow.spec.ts`, which pins the same containment behaviour in CI.

---
<sub>🤖 `om-auto-qa-pr` · evidence-only run, no labels changed · local artifacts: `.ai/qa/artifacts_20260824-qa5318-postmerge/` · screenshots hosted on the `qa-evidence-pr-5318-postmerge` branch (never on this PR's branch).</sub>

PR #5170, opened 2026-08-10
feat(document-generators): add extensible PDF generation module

kriss145:
Source doc: `.ai/specs/2026-08-10-document-generators.md`

## Context
- This module originated in `external/official-modules` under the name `pdf_generations`.
- It has been brought into this repository with several repository-specific adaptations, including package and app wiring, project conventions, validation coverage, and design-system compliance.
- The module was renamed from `pdf_generations` to `document-generators` to better reflect its broader template-driven document generation scope and to align package, module, route, and documentation naming.
- Module documentation and copyable examples were moved out of the runtime package and added under `apps/docs`, following this repository's documentation structure.

## 🎯 Goal
- Add an extensible document generation module for rendering, previewing, downloading, and tracking PDF documents from registered React templates.
- Provide built-in sales order and quote documents together with framework documentation and copyable examples.

## What Changed
- Added the `@open-mercato/document-generators` package with template registration, PDF rendering, generation history, ACL, API routes, backend UI, migrations, and module setup.
- Added order invoice and sales quote templates plus injection widgets that expose document actions in the sales UI.
- Added unit and integration coverage for template registration, validation, rendering, document responses, history, and sales document services.
- Added framework guides, examples, screenshots, and sidebar navigation under `apps/docs`, keeping documentation and sample code outside the runtime module.
- Wired the package into the Mercato app, create-app template, build workflow, CSP configuration, Docker image, and dependency graph.
- Enabled strict design-system linting for the module and updated the generated DS health report.

## 🧪 Tests
- `yarn workspace @open-mercato/document-generators build` — passed.
- `yarn workspace @open-mercato/document-generators typecheck` — passed.
- `yarn workspace @open-mercato/document-generators test` — passed, 97 tests.
- Design-system lint for document generators — passed.
- Documentation build — passed with one unrelated pre-existing WSL2 broken-anchor warning.

## 💥 Breaking Changes
- None.

PR #5323, opened 2026-08-15
docs(specs): revise document generators spec from PR #5170

adeptofvoltron:
## Summary
- Copies **only** the spec (`.ai/specs/2026-08-10-document-generators.md`) from #5170's `feat/document-generators` branch — no implementation code — to review and revise it independently.
- Applies `om-spec-writing` architectural review findings: splits the former Phase 8 (Email & Sharing) and the Phase 9 auto-generation-trigger item into an explicit "Out of scope for this spec" list (each is an independently shippable capability that needs its own spec); restructures `Risks & Impact Review` so every risk states severity, affected area, mitigation, and residual risk; adds a `500` render-error case to `POST /preview`'s error contract to match `/generate`; adds a "Persisted History Entity" table.
- Applies `om-pre-implement-spec` findings, verified against this repository's actual current code (not the source PR): requires `templateRegistry` to persist via `globalThis` (two existing repo lessons describe this exact silent-empty-registry failure mode); adds a `document_generators/encryption.ts` for the GDPR-relevant `resource_label` field; corrects the widget-injection section, which wrongly claimed reuse of an "existing frozen PDF-oriented" spot ID that does not exist (the real spots are Sales' generic `sales.document.detail.order:tabs` / `.quote:tabs`, today occupied only by an unrelated history widget); requires `/generate`'s history-persistence side effect to go through the mutation-guard-registry pattern; fixes Phase 7's attachment field names to the real camelCase wire contract and its unsupported partition-seeding step; renames the generator-plugin key to match the module-id convention; adds an Integration Test Coverage section mapping the 16 existing `TC-DOCUMENT-*` specs.
- Adds the saved analysis report at `.ai/specs/analysis/ANALYSIS-2026-08-10-document-generators.md`.

This PR intentionally carries **no implementation** — it is a spec-only revision, not ready to build against without further discussion with the original author.

## Test plan
- [x] N/A — documentation-only change (`.ai/specs/**`), no code, no schema, no API surface touched.

Spec: .ai/specs/2026-08-10-document-generators.md

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5323"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789403917&installation_model_id=432243&pr_number=5323&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5323&signature=b6f8ffcf76256c67ddce91efa3df61b6ddd2185ccb92c0b64885ea3d7ebe1445"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:

- ❌ `changes-requested` — the 2026-08-18 re-review of head `e82d52daa` found one blocker (`resource_label` is required to be encrypted while also being offered as a SQL-sortable column) and four majors, so the spec needs another pass before it is ready to be treated as implementation-ready. This replaces the `review` label the PR carried since 2026-08-18 07:03Z.
- 📝 `documentation` — this PR only touches `.ai/specs/**`; no code, schema, or API surface changes.
- 🔽 `priority-low` — a spec revision task on a document that is parked for reference, not release-blocking or urgent.
- 🔽 `risk-low` — docs-only change with zero blast radius on running code.
- ✅ `skip-qa` — no manually exercisable UI surface exists to click through; this is a documentation-only change with no database, API, or contract impact.

**On `do-not-merge`:** the label was applied on 2026-08-15 at the maintainer's request — this PR is a revised copy of the spec from the unmerged #5170, here for @kriss145's review rather than queued to land on `develop` — and was removed on 2026-08-18 at 07:03Z by a separate automated run that set `review` in its place. The 2026-08-18 re-review did **not** re-add or remove it; the parked intent is unchanged by anything that run did, and `changes-requested` is not a merge-ready state in any case. Re-apply `do-not-merge` if the parking should be visible on the label set again.

adeptofvoltron:
🤖 `om-auto-review-pr` claiming this PR for review.

adeptofvoltron:
> **Note:** GitHub blocks approving your own pull request via the review API (`Can not approve your own pull request`), so this self-review is posted as a comment instead of a formal approve. Verdict below stands as the review outcome.

# 🔍 Specification Review: Document Generators (spec-only PR)

**Mode:** first review (no prior reviews from this actor). **Scope:** `SPEC_ONLY` — both changed files live under `.ai/specs/**` (the spec itself plus its saved `om-pre-implement-spec` analysis report); no code, schema, or API surface is touched.

## Summary

This PR copies only the spec for the Document Generators feature from the unmerged `feat/document-generators` PR (#5170), then revises it across two review passes — an `om-spec-writing` architectural pass and an `om-pre-implement-spec` pass, the latter grounded against this repository's actual current code (not the source PR), executed via four independent Explore agents that verified every factual claim the spec makes about existing platform surfaces. All identified gaps were fixed within this same PR: `templateRegistry` now requires `globalThis` persistence, the GDPR-relevant `resource_label` field gets an `encryption.ts` declaration, the widget-injection section now names the real `sales.document.detail.order:tabs` / `.quote:tabs` spots instead of a nonexistent "frozen PDF-oriented" one, `/generate` is wired through the mutation-guard-registry pattern, the `attachments` module integration uses the correct camelCase wire fields and a supported partition-creation approach, the generator-plugin key matches the module-id convention, and an Integration Test Coverage section maps the 16 existing `TC-DOCUMENT-*` specs to their API/UI surface. This self-review re-audited the final diff fresh and found two additional consistency nits (stale "existing spot" phrasing left over in Phase 4/4.6 after the Phase-6-only fix, plus a numbered-list gap), both fixed in a follow-up commit during this review.

## Findings

### ⛔ Critical

None.

### ⚠️ High

None — the four items that would have been High/Critical (see the linked analysis report) were fixed before this review pass, verified present in the current diff.

### 🔹 Medium

None outstanding.

### Low

- **Two `.ai/specs/analysis/ANALYSIS-2026-08-10-document-generators.md` "Nice-to-Have" items were intentionally left as narrative-only, not woven into the spec body**: (1) the attachments module's own lack of an `encryption.ts` for stored file bytes (a pre-existing platform-wide gap that exceeds this spec's remit), and (2) an explicit forward-reference to `communication_channels/subscribers/user-deleted-cascade.ts` as the soft-preserve precedent for the still-open GDPR retention question. Both were correctly classified as non-blocking follow-ups in the original analysis; flagging here only so the author can decide whether to fold a one-line pointer into the spec's "Sensitive Data & Retention" risk entry, or leave them to the linked analysis report as-is.
- (Fixed during this review, listed for the record) Phase 4's quote-widget step and Phase 4.6's order-widget step still described the injection-table contribution as "injection spot mapping" / "the existing order detail spot" after the Phase 6 table already got the precise spot-ID correction — now all three locations consistently name `sales.document.detail.order:tabs` / `.quote:tabs`. A stray numbered-list gap (`5.` → `7.`, missing `6.`) in Phase 4.6 was fixed in the same pass.

## 💥 Risks

Adequately covered. `Risks & Impact Review` already states severity, affected area, mitigation, and residual risk per entry (Data Integrity, Tenant & Data Isolation — including the now-required `fetchData` contract-test mitigation for third-party `DocumentService` subclasses — Sensitive Data & Retention, Font Loading, Operational, Browser CSP). No un-mitigated risk found that isn't already labeled as an explicit residual risk with a stated required follow-up.

## 🔁 Backward Compatibility

Correctly reasoned throughout: nothing in this feature has shipped in this repository yet, so most "changes" are pre-release and need no deprecation bridge — the spec's own Migration & Backward Compatibility section says so explicitly and correctly, and still adds a real bridge for the one surface that briefly was public (root package re-exports). No violation of any FROZEN/STABLE contract-surface category in `BACKWARD_COMPATIBILITY.md` was found; a repo-wide check for collisions on the new ACL feature IDs, event IDs, widget spot IDs, and generator-plugin key came back clean.

## 🧩 Gaps

Closed in this PR: `globalThis` registry persistence, `resource_label` encryption map, the widget-spot-ID factual correction, `/generate`'s mutation-guard integration, the attachments wire-contract fields and partition-creation approach, the generator-plugin key convention, and the Integration Test Coverage summary. Remaining (correctly deferred, not blocking): full GDPR erasure orchestration platform-wide (no such module exists yet — tracked against the unimplemented enterprise `data_erasure` spec) and the attachments module's own encryption-at-rest gap (a pre-existing platform limitation, not introduced by this feature).

## 📈 Improvements

The `Integration Test Coverage` table added in this PR is a good addition — it makes the existing 16-test coverage auditable in one place instead of scattered across Phase notes, at zero cost since the tests already exist in the source PR.

## ✂️ Simplicity

The spec earns its length: essentially every section is unique-to-this-feature architecture (registry mechanics, format-neutral render dispatch, tenant-scoping contract, extension conventions), not boilerplate CRUD documentation. The `Changelog`'s length is expected and required by this repo's own `.ai/specs/AGENTS.md` convention for evolving specs, not bloat.

## Validation

No code, config, or i18n files changed — every command in `.ai/agentic.config.json`'s `validation.commands` (`yarn build:packages`, `yarn generate`, `yarn i18n:check-*`, `yarn typecheck`, `yarn test`, `yarn build:app`) is skipped as not-applicable to a docs-only diff. CI: `test` is pending at time of review; `lint`, `ds-lint`, `scope`, `audit-scope`, `prepare` already pass; no failing required checks.

## Verdict

**Approve.** Only Low-severity nits remain, both already actioned or explicitly deferred with a stated reason above.

PR: #5323 (link: https://github.com/open-mercato/open-mercato/pull/5323)

adeptofvoltron:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED (posted as a comment — GitHub's review API refuses both `approve` and `request-changes` on one's own pull request). Lock released.

Re-review of head `e82d52daa`, run because the standing 2026-08-15 approval covered `5d9c70b98` and the spec has moved +406/−113 lines since. Full report: https://github.com/open-mercato/open-mercato/pull/5323#issuecomment-5326291769

**1 blocker** — `GeneratedDocument.resource_label` is required to be encrypted at rest while the same column is offered as a SQL-sortable field in the `GET /documents` allowlist, the backend history table and Phase 6's scoped panel; `ORDER BY` on ciphertext is meaningless, and this repo already ships `packages/shared/src/lib/query/encrypted-sort.ts` plus a worked precedent in `customers/api/labels/route.ts:84` for exactly this. **4 majors** — the Phase 5 read path uses raw `em.findAndCount` instead of `findAndCountWithDecryption`; the Migration & Backward Compatibility section still prescribes deprecation bridges for a package that never shipped; `GET /documents`'s error envelope is written up as a known defect rather than specified; and `document-generators-decoupling.test.ts` is cited as "existing" when the real file is `packages/core/src/__tests__/module-decoupling.test.ts`. **5 minors and 2 nits** besides, three of the minors being @kriss145's still-unaddressed asks from the 2026-08-17 review (record the Email/auto-generation descoping as a product decision; require a contract test or a required parameter for `listTemplateFilterOptions`' authorized subset; record the aggregate-document nuance in "Out of scope" so it is not confused with the parked bulk-generation item).

`autofix: skipped` — the run is technically autofix-eligible (the PR author is this automation's own account), but the PR is a parked reference copy of @kriss145's spec and @kriss145 is actively working the branch, so amending the document under an open discussion was not a call to make autonomously. Findings are handed to the authors; ping me to run the fix loop.

CI on this head is fully settled and green (`lint`, `test`, `ds-lint`, `audit-scope`, `prepare`, `scope` all pass; everything else skipped by path filters), so no CI follow-up is owed and no `ci-monitoring` label was applied. The `do-not-merge` state was left exactly as found — see the label rationale comment above. cc @kriss145

adeptofvoltron:
@kriss145 flagging this for you — a revised copy of your Document Generators spec from #5170, reviewed against the current `develop` codebase (see the review comment above and the linked analysis report). It's marked `do-not-merge` since it's here for your review/reference, not queued to land. (Note: GitHub won't let me assign you directly since you're not a collaborator on this repo, so this @-mention is the handoff instead.)

adeptofvoltron:
💡 Idea for later (not a must-have, just food for thought)

Right now every template is anchored to a single source record: `TemplateRegistryEntry.resourceId()` derives one canonical id, `fetchData` enriches one entity, and history persists one `resource_kind`/`resource_id` pair per generated document. That's the right MVP shape.

Worth considering down the line: templates driven by a MikroORM `Collection<T>` (or just an array of ids) instead of a single record — e.g. a consolidated statement/summary document covering multiple orders/invoices for a customer over a period, rather than one-record-in, one-document-out. That's a real shape change (`resourceId()` would need to become plural, or a new `resourceIds()`/aggregate variant; history would need either multiple rows or a "batch" grouping concept), so it's not something to bolt onto the current spec — more of a Phase-9-and-beyond candidate if/when there's a concrete use case for it (e.g. a periodic customer statement).

Not asking for any change here — just leaving it as a marker for whoever picks this up next.

kriss145:
## Reviewed, and synced against the #5170 implementation

First: **I agree with your changes and want them kept.** The `globalThis` requirement is the important one — that was my own design decision, recorded in the spec changelog on 2026-05-08, which silently disappeared from both the normative text and the implementation (`git log -S globalThis` over the package returns nothing). Good catch. Same for `encryption.ts` on the GDPR-relevant `resource_label`, the mutation-guard requirement on `/generate`, the `document-generators.templates` → `document_generators.templates` plugin-key rename, the camelCase `entityId`/`recordId` fix and the lazy partition creation in Phase 7, and the injection-spot correction. All of these stay in as open work items — none of them exist in the code yet, and the spec now says so.

Two of your edits I adjusted rather than removed, both in the `Persisted History Entity` table you added: `resource_label` is `text NOT NULL` in `entities.ts` (the route supplies the `resource_id` fallback, so it is never stored null), and I added the `created_at` / `updated_at` rows. The table itself is a good addition.

## What I added, and why

You copied the spec at `fb4340c5a`. The branch got two more code commits after that, and more importantly the spec had already drifted from the implementation well before then — it documents intent and decisions, not an inventory, so the gap grew commit by commit. The sync corrects claims that are now simply false and adds the contracts that were never written down.

**Corrected claims**

- Widgets send `{ id: record.id }` and take `resourceKind` from the injection context — they do not forward the raw record. This was asserted in six places, including the TLDR, both request examples and two Design Decisions rows. It matters: sending the record would let the browser influence document contents, and every built-in service validates its input as `{ id: UUID }` precisely to prevent that.
- Template filtering is applied server-side as query parameters (`16385dc89`), not client-side.
- The `GET /documents` data-flow diagram showed scoping by `organization_id` alone while the implementation notes three lines below required both predicates.
- The Internationalization section still listed four Polish-defaulted keys; the engine ships its own English-defaulted dictionaries across five locales.

**New contracts**

- `TemplateAccessPolicy` — extracted to `lib/template-access-policy.ts` in `ba89dd913`, i.e. after your copy — with the omit-vs-reject rule (read endpoints hide inaccessible templates, render endpoints answer `403`), the injected `rbacService` authorizer, the fail-closed rule and the per-request check deduplication. Landed as Phase 4.8.
- API shared conventions: the three edge helpers every route reuses, the full stable error-code table, the RFC 5987 `Content-Disposition` contract, and the `403` case that was missing from both `/preview` and `/generate`.
- The three-layer access enforcement model (widget metadata → route guard → template `requiredFeatures`) and the route→feature map, including why `/preview` is gated by `view` rather than `generate`.
- The registry's real read contract — it listed three of five methods — plus where catalogue data comes from, and a warning that `listTemplateFilterOptions` defaults to the **entire** catalogue while the route passes the authorized subset. A caller omitting that argument silently reintroduces disclosure, and nothing currently prevents it. That looks like a second candidate for the kind of contract test you required for `fetchData`.
- `DocumentTemplateEntry` (referenced four times, never defined), `TemplateMeta.note`, the service-level `resourceId`/`resourceLabel`, and the named `UnknownTemplateError` / `DuplicateTemplateError`.
- The `hooks/**` React Query layer, route-local list components, `api/_shared/**`, the full `utils/` inventory, backend list behavior (filters, the sortable-fields parity with the API allowlist, `isFetching`), the Markdown escaping requirement, and `TC-DOCUMENT-017/018/019`.

**Phase 6 rewritten against the current layout.** Its step 1 pointed at `components/HistoryList.tsx`, which was deleted in `bd63c6a63` — following it would have produced a duplicate of a layer that already exists. It now extends `useDocumentHistory` with the resource filters and promotes the table to module level. The `refreshToken` counter is dropped in favour of invalidating `documentHistoryQueryKey`, since the hooks layer provides that already.

## One thing worth a separate decision

Moving Email & Sharing and the auto-generation trigger to "Out of scope" is a product call rather than cleanup. I'm fine with it, but it should be recorded as a deliberate decision rather than absorbed as a review nit.

## Where it is

Thanks for the fork access — it's pushed straight onto this branch, so there's nothing to cherry-pick:

**Commit:** 27a29d616e49ce127e41bfd1560dc81a44553775 — one commit on top of `5d9c70b98`, touching `.ai/specs/2026-08-10-document-generators.md` only.

Happy to split it into smaller commits if you'd rather review it in pieces, and I'll rebase if this branch moves.

kriss145:
Re: your collection-driven templates idea — https://github.com/open-mercato/open-mercato/pull/5323#issuecomment-5303244543

Interesting direction, and I agree with keeping it out of this spec. A periodic customer statement is a real use case, and you're right that it's a shape change rather than an extension: the engine asks each template "which record is this document about?" and expects exactly one answer, and history has room for exactly one `resource_kind` / `resource_id` pair.

One nuance worth recording, because it will otherwise mislead whoever picks this up: this is **not** the `Bulk generation` item already sitting in "Out of scope". Those two only share the word "multiple".

| | Bulk generation (already parked) | Aggregate document (your idea) |
|---|---|---|
| Select 12 orders → | 12 documents | 1 document |
| Each document covers | one order | twelve orders |
| The 1:1 model is | preserved, just repeated | broken |
| The actual problem | throughput — 12 renders on the request thread | identity — no way to express "about 12 records" |
| The solution shape | a queue worker | plural `resourceId()` + a batch concept in history |

Printing twelve letters versus writing one summary letter about twelve things. Nothing architectural in common.

Beyond the three places you named, the 1:1 assumption is also baked into: the Phase 6 resource-scoped history panel (it filters on `resource_kind` + `resource_id` supplied together, and an aggregate has no single id to filter on); `resource_label`, which today is an encrypted, GDPR-relevant label for one record and would have to mean "customer + period" instead; the widget entry point, since an aggregate isn't launched from a record detail at all and so needs a different injection surface than the `:tabs` spot; `buildDocumentFilename`, which derives the name from `data.document.number`; and Phase 7's attachment link, `recordId: rendered.resource.id`.

So: agreed, own sub-spec — and I'd write it **after** this one ships rather than now. Once the registry, renderer and history contracts exist as working code, the aggregate spec can build on a real base and state precisely what it changes, instead of designing against a moving target. Until then this comment is the marker.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-08-18T09:22:52Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron:
# 🔍 Re-review — Specification review: docs(specs): revise document generators spec from PR #5170

> **Note:** GitHub's review API refuses both `approve` and `request-changes` on one's own pull request, so this re-review is posted as a comment. The verdict below is the review outcome and the pipeline label has been set to match.

## 🎯 Summary

This is a **re-review**. The previous `om-auto-review-pr` pass ran on 2026-08-15 at 16:44 against `5d9c70b98` and approved; the head has since moved to `e82d52daa` through four commits — @kriss145's sync against PR #5170's code (`27a29d61`), the "Planned/Not Started" status rewrite (`215c18e8`), a re-run of `om-pre-implement-spec` (`39f39475`), and the three findings that run raised (`e82d52da`). That is +406 / −113 lines on the spec and +154 / −22 on the analysis report, so the standing approval no longer covered the document that is on the branch.

The PR is spec-only: both changed files live under `.ai/specs/`, so this ran the specification-review fork (`references/spec-review.md`) — the spec was read end to end and its claims were verified against the actual `develop` codebase in an isolated worktree checked out at the PR head, not against PR #5170.

The document is in good shape overall, and the two structural moves since the last review are both improvements: the status rewrite correctly stops the spec from claiming code exists in this repository (it does not — `packages/document-generators/` is absent from `develop`, and PR #5170 was closed unmerged on 2026-08-14), and @kriss145's sync added genuinely missing normative contracts (`TemplateAccessPolicy`, the stable error-code table, `GET /templates/options`, `DocumentTemplateEntry`). Every codebase claim I spot-checked held up: the `sales.document.detail.order:tabs` / `sales.document.detail.quote:tabs` spots exist and today hold only the unrelated `sales.injection.document-history` entry; `rbacService.userHasAllFeatures(userId, features, { tenantId, organizationId })` matches the spec's `TemplateFeatureAuthorizer` exactly; `ensureDefaultPartitions`, `isSameScope`, `runGuardAfterSuccessCallbacks`, the `globalThis` event-bus precedent, the `customers/encryption.ts` shape and both cited `.ai/lessons/*` files are all real and described accurately.

What this pass found is a **new class of problem that the 2026-08-15 review could not have caught in the same form**: the `encryption.ts` requirement added on 2026-08-15 and the Phase 5 history-list design written earlier now contradict each other. `resource_label` is required to be encrypted at rest, and the same field is simultaneously offered as a SQL-sortable column in the API allowlist, in the backend history table and in Phase 6's scoped panel. This repository has already learned that lesson the hard way and ships a dedicated helper for it. Alongside that, the two framing rewrites left some residue: parts of the document still read as a delta against the closed branch's code rather than as a plan for a from-scratch build.

**Verdict: ❌ CHANGES REQUESTED** — one blocker and four majors. Per this repository's `om-auto-review-pr` extension (`.ai/skills/om-auto-review-pr/SKILL.md`), any blocker, major *or* minor finding requests changes; the shared skill's default rule would already have done so on the blocker alone.

## 🧪 Validation Gate

Per this repository's GitHub-checks-first extension, CI check results are preferred over re-running the gate locally. **Validation source: GitHub checks** (head `e82d52daa`).

| Command / check | Result | Notes |
|---|---|---|
| `lint` (CI) | ✅ pass | https://github.com/open-mercato/open-mercato/actions/runs/32041609613/job/95468598739 |
| `test` (CI) | ✅ pass | https://github.com/open-mercato/open-mercato/actions/runs/32041609613/job/95468914030 |
| `ds-lint` (CI) | ✅ pass | https://github.com/open-mercato/open-mercato/actions/runs/32041609613/job/95468598758 |
| `audit-scope` (CI) | ✅ pass | https://github.com/open-mercato/open-mercato/actions/runs/32041609613/job/95468598774 |
| `prepare` / `scope` (CI) | ✅ pass | — |
| `merge-coverage`, `docker-build`, `ephemeral-integration`, `audit`, `mutate`, `comment`, `[code]smith` | ⏭️ skipped by CI | Path filters — the diff touches no code |
| Relative-link resolution in both changed files | ✅ pass | Checked in the worktree; the documents use inline code paths, and every referenced path (`.ai/lessons/*`, cited source files) resolves |

**Skipped local commands**, because a markdown-only diff has nothing to compile, generate or test, and CI reported usable results for this head: `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app`. No required check is failing and none is pending, so there is no CI follow-up owed on this PR.

## ⛔ Blockers

### 1. `resource_label` is required to be encrypted *and* offered as a SQL-sortable column — the two cannot both hold

`.ai/specs/2026-08-10-document-generators.md:486` requires `document_generators/encryption.ts` to encrypt `GeneratedDocument.resource_label`, and rightly so — it is cached PII. But three other normative places treat the same column as ordinarily sortable:

- `:641` — "`sort` accepts `resource_label`, `template_label`, `format`, `generated_by`, or `generated_at`"
- `:724` — the backend history table marks `resourceLabel` as one of exactly five sortable columns, `manualSorting` against the endpoint
- `:972` — the Phase 5 data flow prescribes `em.findAndCount(GeneratedDocument, {...}, { orderBy: <allowlisted field> <direction>, limit, offset })`

An `ORDER BY` on an encrypted column sorts ciphertext. The user asks for documents by resource name and gets what looks like a random order, with keyset pagination walking that same meaningless sequence. This repository already hit this exact problem and solved it: `packages/core/src/modules/customers/api/labels/route.ts:84` carries the comment *"`label` is encrypted at rest; SQL can't sort it meaningfully, so the candidate set is decrypted and sorted in memory below"*, and `packages/shared/src/lib/query/encrypted-sort.ts` exists for it — `resolveEncryptedSortFields`, `resolveEncryptedSortMaxRows`, `sortRowsInMemory`, with `packages/core/src/modules/customers/api/interactions/route.ts:463-478` routing encrypted sort fields onto a bounded candidate scan and the non-encrypted ones onto the SQL keyset path.

This is a blocker rather than a major because the choice is not cosmetic and it changes a contract the spec has already published. The bounded candidate scan is capped by `OM_ENCRYPTED_SORT_MAX_ROWS` and logs a warning when the cap is hit, meaning `total` and page boundaries become approximate for that sort — an API-visible semantic that has to be specified up front, not discovered in Phase 5. Please pick one and write it down:

- **drop `resource_label` from the `sort` allowlist**, set `enableSorting: false` on the Resource column in both the `/history` page (`:724`) and Phase 6's scoped panel (`:1012`, which inherits the allowlist rule), and say why; or
- **specify the encrypted-sort path** — `resolveEncryptedSortFields` on the history route, the candidate scan with its `OM_ENCRYPTED_SORT_MAX_ROWS` cap, and the resulting truncation/pagination semantics in the `GET /documents` contract at `:641`.

## ⚠️ Majors

### 2. Phase 5's data flow reads the encrypted history table with raw `em.findAndCount`

`:972` prescribes `em.findAndCount(GeneratedDocument, ...)` for the list path, and `:980` tells the implementer to resolve `em` from `createRequestContainer()`. With `resource_label` encrypted, that returns ciphertext into `GeneratedDocumentDto.resourceLabel` (`:933`), which the history table then renders. Root `AGENTS.md` states the rule unconditionally — *"Use `findWithDecryption`/`findOneWithDecryption` instead of `em.find`/`em.findOne`"* — and the matching helper for this call shape already exists: `findAndCountWithDecryption` (`packages/shared/src/lib/encryption/find.ts:75`). The Encryption section was added on 2026-08-15, after the Phase 5 read path was written, and the two were never reconciled. Update `:972` and Phase 5's key implementation notes to name `findAndCountWithDecryption` with its decryption scope.

### 3. "Migration & Backward Compatibility" still prescribes deprecation bridges for a package that never shipped

The section at `:1090-1108` is written as a diff against the closed branch's code — "The unreleased implementation *was decentralized before merge*", "Sales *now owns*…", "The unreleased `BaseDocumentService.filename()` fallback *is also removed before merge*". After the `215c18e8` status rewrite established that nothing exists in this repository, this framing is not just stylistic: `:1094` instructs the implementation to keep *"the previous root exports from `@open-mercato/document-generators` … as deprecated compatibility re-exports for at least one minor version"*. There are no previous root exports. A faithful implementer would ship dead deprecation shims on day one of a greenfield package, and reviewers would then have to defend them under `BACKWARD_COMPATIBILITY.md`'s deprecation protocol. Rewrite the section for a from-scratch build: state that no released surface exists, keep the *decisions* the closed branch's iterations produced (namespaced `<module>.<template>` IDs, required per-entry `filename`, single `register`, no client-supplied resource identity) as target contracts, and drop every instruction to preserve, bridge or deprecate something that was never published.

### 4. `GET /documents`'s error envelope is specified as a known defect instead of a contract

`:528` records, as a "Known inconsistency", that `GET /documents` answers a schema failure with `{ error: 'Invalid query parameters' }` — untranslated prose in the `error` field, no stable code, no `message` — and adds *"the client cannot branch on it today"*. There is no "today": the route does not exist here. As written, the document contains two contradictory normative statements about the same route (the error table at `:519-527`, and this note), and the implementer has no way to tell which one to build. Since this is a plan, specify the intended behaviour — `invalid_query` / `400` with a translated `message`, consistent with the table — and either delete the note or demote it to a changelog line recording that the closed branch's version diverged.

### 5. Stale codebase reference: `document-generators-decoupling.test.ts` does not exist

`:778` closes the cross-tenant residual risk by pointing at *"the same class of gap the existing `document-generators-decoupling.test.ts` closes for module coupling"*. No file by that name exists anywhere in this repository — the decoupling guard is `packages/core/src/__tests__/module-decoupling.test.ts` (the one root `AGENTS.md` routes to for cross-module coupling). Calling a non-existent file "existing" is exactly the class of claim the 2026-08-17 status rewrite set out to eliminate, and it weakens an otherwise strong mitigation: the required `tenant-scoping-contract.test.ts` is justified by analogy to a precedent the reader cannot open. Point at the real file.

## 🧩 Minors

### 6. Phases 4.7 and 4.8 still read as completed work under "(Planned)" headings

`215c18e8` retitled every phase heading, but the step lists underneath Phase 4.7 (`:906-912`) and Phase 4.8 (`:916-920`) were not converted: *"1. Added required, extensible `format` metadata…"*, *"This phase added the owning-module permission check…"*, *"3. Wired all four routes to construct the policy…"*. Every other phase uses noun phrases ("Package scaffold", "Sales-owned `OrdersDocumentService`"), so a reader skimming for what is left to build gets the opposite signal from these two — the precise misreading the rewrite existed to prevent. Convert them to the same imperative/noun form.

### 7. The route inventory is stale in the TLDR and Proposed Solution

`GET /templates/options` was added on 2026-08-17, but `:56` still says "**Four API routes**" and lists four, while `:510` ("All five routes are thin HTTP adapters"), `:876` ("All five routes export `metadata`") and `:1121` all say five. The TLDR's scope list (`:13-15`) is a further step behind — it names three routes and omits `/preview` as well. Bring both up to the five-route reality.

### 8. 🔁 INHERITED — @kriss145: the Email & Sharing / auto-generation descoping is still framed as cleanup, not a product decision

[Review comment, 2026-08-17](https://github.com/open-mercato/open-mercato/pull/5323#issuecomment-5313311566): *"Moving Email & Sharing and the auto-generation trigger to 'Out of scope' is a product call rather than cleanup. I'm fine with it, but it should be recorded as a deliberate decision rather than absorbed as a review nit."* Verified against the current head: `:1075-1084` gives a good *architectural* rationale (each is independently shippable and carries its own design surface), and the 2026-08-15 changelog entry records the edit — but both frame it as spec hygiene, which is what @kriss145 asked it not to be. One sentence stating that deferring email delivery and event-driven auto-generation out of the initial deliverable is a deliberate product decision, with its date and who made it, would close this.

### 9. 🔁 INHERITED — @kriss145: no contract test required for `listTemplateFilterOptions`'s authorized-subset argument

Same comment: *"A caller omitting that argument silently reintroduces disclosure, and nothing currently prevents it. That looks like a second candidate for the kind of contract test you required for `fetchData`."* Verified: `:256` states the hazard well and ends *"no type or test currently prevents it"* — the warning is there, the requirement is not. `TC-DOCUMENT-019` (`:840`) covers facet shape and deduplication, not authorization scoping. Either require a unit/contract test asserting the options route derives facets from the authorized subset, or make the parameter required in the signature at `:237` so the type prevents it — the tenant-scoping residual risk at `:778` sets the precedent for the former.

### 10. 🔁 INHERITED — @kriss145: the aggregate-document nuance is not recorded in "Out of scope"

[Follow-up comment, 2026-08-17](https://github.com/open-mercato/open-mercato/pull/5323#issuecomment-5313509631): *"One nuance worth recording, because it will otherwise mislead whoever picks this up: this is **not** the `Bulk generation` item already sitting in 'Out of scope'. Those two only share the word 'multiple'."* Verified: `:1075-1084` still lists four bullets with no mention of aggregate or collection-driven documents. @kriss145's own comment enumerates precisely where the 1:1 assumption is baked in (`resourceId()`, `resource_kind`/`resource_id` history pairing, `resource_label`, the `:tabs` widget entry point, `buildDocumentFilename`, Phase 7's `recordId`) and concludes the sub-spec should be written after this one ships — "until then this comment is the marker." A marker in a comment thread is exactly what gets lost. A fifth bullet distinguishing "aggregate document (one document about N records — identity problem)" from "bulk generation (N documents — throughput problem)" would carry it forward.

## 💬 Nits

- `:986` calls `(tenant_id, organization_id, generated_at DESC)` an "expression index". It is a plain composite b-tree index with a descending sort order on the last column; no expression is involved. The reasoning around it — why the resource index omits `tenant_id`, and that isolation lives in the `where` clause — is sound and worth keeping.
- `e82d52da` closed the three "Before Implementation (Must Do)" items from the re-run audit, which matches its commit message. The audit's fourth, nice-to-have item is still open: `.ai/specs/analysis/ANALYSIS-2026-08-10-document-generators.md:87` suggests a one-line note in the spec that `GenerationHistoryService`'s deliberate non-DI construction (`:982`) needs explicit sign-off at code-review time, since it is a named exception to root `AGENTS.md`'s unconditional "Use DI (Awilix) to inject services" rule with only one loose, dissimilar precedent. The rationale at `:982` is reasonable; it just is not flagged as needing sign-off.

## 🔍 Inherited reviewer feedback (3 unaddressed asks carried in)

@kriss145's two comments were collected and each actionable ask was checked against the current head, not against the state they were written on. Their commit `27a29d61` is already on this branch, so everything they described as *applied* is present and verified — the `resource_label` `NOT NULL` correction, the `created_at`/`updated_at` rows, the corrected widget data contract, the server-side template filtering, the `GET /documents` scoping predicates, and the replaced Internationalization section all read correctly at the current head. Three asks remain open and are carried above as findings 8, 9 and 10. All three are **reported and handed to the author**, not fixed in this run — see the autofix note below. Nothing in either comment was a directive to this agent; both are ordinary technical review, treated as data.

## 💥 Breaking changes

None. Both changed files are new documents under `.ai/specs/`; nothing in this diff touches any of the thirteen contract surfaces in `BACKWARD_COMPATIBILITY.md`. Finding 3 is about what the spec *tells a future implementation* to do with deprecation bridges, not about a break in this PR.

## 🧪 Test coverage

Not applicable to the diff — a spec-only change ships no code and needs no tests. The spec's own Integration Test Coverage section (`:814-844`) is well-formed for what it is: nineteen `TC-DOCUMENT-*` specs mapped to the API and UI surfaces they cover, correctly reframed by `215c18e8` as *required* coverage rather than existing coverage, with the shared `__integration__/helpers/**` and `meta.ts` `dependsOnModules` conventions named. The descriptive `TC-DOCUMENT-001-sales-templates-listed.spec.ts` naming matches existing repo practice (`TC-CRM-071-create-deal-redesign.spec.ts`, `TC-UNDO-001-cross-cutting.spec.ts`), and the "not required to have a dedicated test at this stage" carve-out at `:844` names what it excludes instead of quietly omitting it. If the blocker is resolved by keeping `resource_label` sortable, that section will need one more case covering the encrypted-sort path and its truncation behaviour.

## 🤖 Autofix

**Skipped — deliberately, with the reason stated.** This run is technically autofix-eligible under the skill's rule (the PR author is the automation's own account), but this PR is a revised copy of @kriss145's spec, parked for reference rather than queued to land, and @kriss145 is actively collaborating on this branch — they pushed `27a29d61` to it directly. Amending the spec text under an open discussion, uninstructed, is not a call to make autonomously, so the findings are reported for the authors to act on. Say the word and I will run the fix loop over findings 1–10 in a follow-up pass.

**Labels:** the pipeline label moves `review` → `changes-requested`. Note for the record: this PR carried `do-not-merge` from 2026-08-15 16:44 until 2026-08-18 07:03Z, when another automated run replaced it with `review`. This run did not re-add or remove it — the parked intent is unchanged by anything here, and `changes-requested` is in any case not a merge-ready state. `documentation`, `skip-qa`, `priority-low` and `risk-low` all still fit a docs-only change and are preserved.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-08-18T09:58:00Z.

Instructed by the maintainer to run the fix loop over findings 1–10 from the [2026-08-18 re-review](https://github.com/open-mercato/open-mercato/pull/5323#issuecomment-5326291769) — the pass that deliberately skipped autofix and said "ping me to run the fix loop". cc @kriss145

adeptofvoltron:

- 🔍 `review` — replaces `changes-requested`. That label was correct for head `e82d52daa`, where the 2026-08-18 re-review found one blocker and four majors; all ten findings and both nits are now applied in `d0fc562ed`, so the PR is back in front of a reviewer rather than sitting on the author's desk. It is deliberately not `merge-queue`: the fixes were written by the same automation account that authored the PR, so nobody independent has approved them yet, and @kriss145 is the reviewer this parked copy exists for.
- 📝 `documentation` — unchanged. The diff still touches only `.ai/specs/**`; no code, schema, API surface, or contract file is modified.
- ✅ `skip-qa` — unchanged. There is no manually exercisable UI surface in a markdown-only diff, and no database, API, or contract impact to click through.
- 🔽 `priority-low` — unchanged. A spec revision on a document that is parked for reference; nothing about it is release-blocking.
- 🔽 `risk-low` — unchanged. Docs-only, zero blast radius on running code.
- ⏱️ `ci-monitoring` — added in place of `in-progress` as this run released its claim. The work is finished and fully reported; the label means only that CI on head `d0fc562ed` was still pending when the report was posted and a CI-result follow-up is owed. It is not a claim signal — any other actor is free to pick this PR up.
- 🚫 `in-progress` — removed. The outer claim this run took at 09:47Z is released.

**On `do-not-merge`:** still absent, and this run deliberately did **not** re-add it. The label was applied on 2026-08-15 at the maintainer's request — this PR is a revised copy of @kriss145's spec, here for review rather than queued to land on `develop` — and removed on 2026-08-18 at 07:03Z by an unrelated automated run. Re-applying a merge block was not part of what this run was asked to do, and the PR is in any case unmergeable through the normal path while it carries `review` without an approval. If the parked intent should be visible on the label set again, re-apply it; that call belongs to the maintainer, not to this run.

adeptofvoltron:
# 🤖 `om-auto-fix-pr` — fix loop complete

Run at the maintainer's instruction, taking up the offer the 2026-08-18 re-review left open (*"Say the word and I will run the fix loop over findings 1–10"*). **All ten findings and both nits are fixed**; nothing was deferred, so no follow-up issues were filed. Head moved `e82d52daa` → **`d0fc562ed`**, +61 / −36 on `.ai/specs/2026-08-10-document-generators.md`. cc @kriss145

## 🔀 Base merge

`origin/develop` (`3adb0d0df`) merged into the PR branch before any other work, so the review and CI below judge the real merge result. The merge was clean — `develop` has not touched `.ai/specs/2026-08-10-document-generators.md` since this branch diverged, so there was nothing to reconcile.

## ⛔ The blocker — `resource_label`: encrypted at rest *and* SQL-sortable

The two requirements could not both hold. `resource_label` is required to be encrypted (it is cached PII derived from a customer/company/order name), and the same column was simultaneously offered as a sort field in the `GET /documents` allowlist, in the backend history table's five sortable columns, and inside Phase 6's scoped panel — an `ORDER BY` over ciphertext returns an order no user can read, with keyset pagination then walking that same meaningless sequence.

The review offered two resolutions. **This run took the restrictive one: `resource_label` is dropped from the `sort` allowlist** and the Resource column sets `enableSorting: false` on both history surfaces. The reasoning, now written into the spec rather than left implicit:

- The alternative — routing the field through `packages/shared/src/lib/query/encrypted-sort.ts` — is not free. Its bounded candidate scan is capped by `OM_ENCRYPTED_SORT_MAX_ROWS`, and once that cap is hit `total` and page boundaries become **approximate for that one sort**. That is an API-visible semantic, and taking it on permanently to serve a sort nobody has asked for is a poor trade.
- The list's natural order is `generated_at DESC`; resource-scoped panels are already filtered to a single record; and callers looking for one record's documents use the `resource_kind` + `resource_id` filters, which are exact-match predicates on unencrypted identifier columns and are entirely unaffected.
- **Adding a sort field later is additive; removing one after clients depend on it is not.** The restrictive option is the reversible one, which is why it was chosen without stopping to ask. The encrypted-sort path is documented in place as the escape hatch, with the truncation and pagination semantics that would have to be specified alongside it.

Recorded as a Design Decisions row and expanded under Data Contracts → Encryption, so the next reader sees the constraint and the choice together rather than rediscovering the contradiction.

## ⚠️ The four majors

| # | Finding | What changed |
|---|---|---|
| 2 | Phase 5 read the encrypted history table with raw `em.findAndCount` | Now specifies `findAndCountWithDecryption` (`packages/shared/src/lib/encryption/find.ts`) in both the data-flow diagram and the key implementation notes, with the root `AGENTS.md` rule quoted and the write path explicitly excluded — `/generate` still encrypts before the best-effort persist. Without this the history table would have rendered ciphertext into the Resource column. |
| 3 | "Migration & Backward Compatibility" prescribed deprecation bridges for a package that never shipped | Rewritten for a from-scratch build. It now opens by stating that nothing has been released and that an implementer must **not** write deprecation shims, compatibility re-exports, dual-key translation windows or dual-emit events — the instruction to keep *"the previous root exports … as deprecated compatibility re-exports"* would have shipped dead shims on day one of a greenfield package. The decisions the closed branch's iterations produced (namespaced `<module>.<template>` IDs, single `register`, required per-entry `filename`, no client-supplied resource identity, the ownership split) are kept as **target contracts**, and a closing paragraph names what actually becomes a frozen contract surface at first merge. |
| 4 | `GET /documents`'s error envelope was written up as a known defect | Now specified as a contract: a `listDocumentsSchema` failure answers `400 { error: 'invalid_query', message: <translated> }` like every other route. The `invalid_query` row in the error table was widened to cover it, the route gained its own Errors list, and the closed branch's divergence is recorded as history instead of as a live "today". The document no longer carries two contradictory normative statements about the same route. |
| 5 | Stale reference to a non-existent `document-generators-decoupling.test.ts` | Points at the real `packages/core/src/__tests__/module-decoupling.test.ts`. This mattered beyond tidiness: the required `tenant-scoping-contract.test.ts` is justified *by analogy* to that precedent, so an unopenable file weakened a mitigation on a High-severity cross-tenant risk. |

## 🧩 The five minors

- **6 — Phases 4.7 and 4.8 still read as completed work under "(Planned)" headings.** Both step lists converted from past-tense prose (*"Added…", "Wired all four routes…"*) to the noun form every other phase uses, so a reader skimming for remaining work no longer gets the opposite signal from exactly two phases. While converting, "all four routes" was made explicit — `/templates`, `/templates/options`, `/preview`, `/generate`; `/documents` reads history rows, not templates, and needs no policy — since "four" next to a five-route inventory reads as a stale number rather than a deliberate subset.
- **7 — Route inventory stale in the TLDR and Proposed Solution.** Both now list five routes. The TLDR's scope list was two steps behind: it named three routes, omitting `/preview` as well as the newer `/templates/options`.
- **8 — 🔁 @kriss145: the Email & Sharing / auto-generation descoping framed as cleanup.** Now a standalone **product decision** callout with its provenance: proposed by @adeptofvoltron in the 2026-08-15 architectural pass, accepted by @kriss145 on 2026-08-17, recorded 2026-08-18. It states what the decision costs (manual download stays the only delivery path until a follow-up spec lands) and that re-opening it is a product call, not a spec edit — which is the distinction that was asked for.
- **9 — 🔁 @kriss145: nothing prevented `listTemplateFilterOptions` being called over the whole catalogue.** Fixed by the stronger of the two offered options: **the `templates` parameter is now required in the signature**, so the catalogue-wide variant is uncallable and the hazard moves from "reviewers must remember" to "the code does not compile". `TC-DOCUMENT-020` was added to assert the same guarantee behaviorally against a restricted-user fixture, so a later refactor reintroducing a default is caught by a test as well as by the type.
- **10 — 🔁 @kriss145: the aggregate-document nuance lived only in a comment thread.** Now a fifth "Out of scope" bullet that leads with *"this is not the same item as Bulk generation"* and carries the distinction across: 12 orders → 12 documents (throughput, solved by a queue worker) versus 12 orders → 1 document (identity, no way to express "about 12 records"). It enumerates all seven places the 1:1 assumption is load-bearing — `resourceId()`, the single `resource_kind`/`resource_id` history pair, Phase 6's paired filter, `resource_label`, the `:tabs` entry point, `buildDocumentFilename`, Phase 7's `recordId` — and keeps the conclusion that the sub-spec should be written *after* this one ships.

## 💬 Both nits

- `(tenant_id, organization_id, generated_at DESC)` is no longer called an "expression index" — it is a plain composite b-tree index whose trailing column carries a descending sort order. The surrounding reasoning about why the resource index omits `tenant_id` was sound and is kept.
- `GenerationHistoryService`'s deliberate non-DI construction now states that the deviation **requires explicit sign-off at code-review time**, since root `AGENTS.md`'s "use DI, avoid `new`-ing directly" rule is unconditional and the only repo precedent is loose and dissimilar. That closes the last open item from the re-run `om-pre-implement-spec` audit.

## 🧪 Coverage and validation

Two integration specs added to the required-coverage table: `TC-DOCUMENT-020` (options facets scoped to the caller's authorized subset) and `TC-DOCUMENT-021` (`sort=resource_label` rejected with `400 invalid_query`, the four allowlisted fields accepted, and returned rows carrying plaintext `resourceLabel` — which is what proves the list reads through `findAndCountWithDecryption`).

Every command in `.ai/agentic.config.json`'s `validation.commands` (`yarn build:packages`, `yarn generate`, `yarn i18n:check-*`, `yarn typecheck`, `yarn test`, `yarn build:app`) is **not applicable** to a markdown-only diff — there is nothing to compile, generate, or test. Structural checks that do apply were run in the worktree: the changelog and coverage tables keep their column counts, and every code path the new text cites was verified to exist on `develop` (`packages/shared/src/lib/query/encrypted-sort.ts`, `findAndCountWithDecryption` at `packages/shared/src/lib/encryption/find.ts:75`, `packages/core/src/__tests__/module-decoupling.test.ts`, and the `customers/api/labels/route.ts` in-memory-sort precedent).

**⏱️ CI was still pending on `d0fc562ed` when this comment was posted** (`lint`, `test`, `ds-lint`, `audit-scope`, `prepare`, `scope`) — this report deliberately lands before the wait rather than after it, so a process that dies watching CI still leaves a fully reported PR. Do not read "fixes complete" as "checks green"; the `ci-monitoring` label tracks the outstanding follow-up, and a result comment will follow.

## 📋 Follow-ups filed

**None, deliberately.** Follow-up issues exist for findings a fix run declines to act on, and this run declined none — all ten findings and both nits were in scope and are applied. The two Low items from the 2026-08-15 review (the `attachments` module's own missing `encryption.ts` for stored file bytes, and the `communication_channels/subscribers/user-deleted-cascade.ts` soft-preserve precedent pointer) remain narrative-only in the linked analysis report, which is the disposition already chosen and confirmed at the time; neither is a finding this review left open.

## ✅ Merge-readiness verdict

**Not merge-ready, and not intended to be.** This PR is a parked reference copy of @kriss145's spec, and it stays that way. The blocker and every finding are resolved, but the fixes were authored by the same account that owns the PR, so nothing independent has approved them — the pipeline label is `review`, not `merge-queue`. `do-not-merge` was removed on 2026-08-18 at 07:03Z by an unrelated run and this run did not re-add it; re-apply it if the parking should be visible on the label set.

The next step is a human read, not a merge command: @kriss145, the blocker resolution (dropping the sort rather than adopting the encrypted-sort path) and the rewritten Migration & Backward Compatibility section are the two places where a different call was genuinely available, so those are the ones worth your eye.

PR: #5323 (link: https://github.com/open-mercato/open-mercato/pull/5323)

adeptofvoltron:
🤖 `om-auto-fix-pr` — ✅ CI follow-up on head `d0fc562ed`

CI is fully settled and green, so the "pending" caveat in the [run summary](https://github.com/open-mercato/open-mercato/pull/5323#issuecomment-5326678367) is now closed out. Every check that runs on this diff passed:

| Check | Result | Duration |
|---|---|---|
| `lint` | ✅ pass | 1m22s |
| `test` | ✅ pass | 4m3s |
| `ds-lint` | ✅ pass | 1m28s |
| `audit-scope` | ✅ pass | 11s |
| `prepare` | ✅ pass | 1m37s |
| `scope` | ✅ pass | 32s |
| `license/cla` | ✅ pass | — |

Skipped by path filters, as expected for a markdown-only diff: `merge-coverage`, `docker-build`, `ephemeral-integration`, `audit`, `mutate`, `comment`, `[code]smith`.

Nothing is pending and no check failed, so no further CI follow-up is owed and the `ci-monitoring` label is removed. The PR's state is unchanged otherwise: `review`, awaiting a human read from @kriss145 — green CI on a docs-only diff confirms the document is well-formed, not that the design calls inside it are the right ones.

PR: #5323 (link: https://github.com/open-mercato/open-mercato/pull/5323)

patzick (approved):

Issue #5099, opened 2026-08-06
security: scaffolded apps inherit none of the monorepo's transitive security pins


patzick:
## Problem

The monorepo's root `package.json` carries ~30 `resolutions` groups pinning transitive dependencies above their advisory floors. `packages/create-app/template/package.json.template` carries **three** (`lodash`, `lodash-es`, `zod`).

So a freshly scaffolded app installs none of them. "Advisories cleared" currently means "cleared in the monorepo" — not "cleared for users who run `npx create-mercato-app`".

The sharpest example is `sharp`: the advisory is `<0.35.0` and has no 0.34 backport, and Next still declares `sharp: ^0.34.5` as its optional image-optimization dependency. The monorepo forces `0.35.3` via resolution; a scaffolded app gets whatever Next resolves, i.e. the vulnerable line.

## Why it is not simply "copy the block over"

Most root pins exist because of dev/build/docs-only dependencies the template does not have (docusaurus, webpack-dev-server, mermaid, testcontainers). Copying wholesale would add pins for packages a scaffolded app never installs, which then rot invisibly per #5098.

The useful subset is the pins covering packages a scaffolded app *actually* resolves — `sharp` first, then whatever else `yarn npm audit --all --recursive --severity high` reports from inside a freshly generated app.

## Suggested first step

Scaffold an app, run the same audit command CI runs, and use the actual output to decide the template's pin set. That also gives us a repeatable check worth wiring into `yarn test:create-app:integration` so the gap cannot silently reopen.

_Pre-existing gap, out of scope for #5085 (which is monorepo-only). Suggested by the automated review on that PR (Minor 8)._

PR #5085, opened 2026-08-06
fix(deps): clear remaining high-severity audit advisories

patzick:
## Why

The `audit` CI job on `main` (`yarn npm audit --all --recursive --severity high`, ci.yml:318) is red, and because `test` / `ephemeral-integration` gate on it, both are skipped — so `main` currently has no test signal either. The scheduled `audit.yml` workflow runs the same command and opens a tracking issue when it fails.

## Root cause

Two separate things:

**1. All four `undici` resolution keys had gone stale.** They are keyed on descriptors that no longer appear anywhere in the tree:

| resolution key on main | actually present in tree |
|---|---|
| `undici@npm:7.24.0` | — |
| `undici@npm:^7.19.0` | `^7.12.0` (cheerio) |
| `undici@npm:^8.3.0` | `^7.16.0` (`@newrelic/security-agent`) |
| `undici@npm:^8.4.1` | `^8.5.0` (testcontainers), `^8.7.0` (`@open-mercato/shared`) |

None matched, so none applied, and undici silently floated up to the vulnerable `7.28.0` / `8.8.0`. This is the failure mode worth remembering: **descriptor-keyed resolutions rot invisibly** when an upstream package changes its declared range — Yarn does not warn about a resolution that matches nothing.

Fixed two ways. For the one requester that is **our own** package, the honest fix is a direct range bump that cannot rot: `packages/shared`'s `undici` goes `^8.7.0` → `^8.10.0`, which removes that descriptor key outright. For the third-party requesters we do not control, the keys are re-derived from the current lock and pinned to `7.29.0` / `8.10.0`.

**2. Four pins became vulnerable again** under newly published advisories:

| package | was | now | advisory |
|---|---|---|---|
| brace-expansion | 5.0.8 | 5.0.9 | `<5.0.9` |
| fast-uri | 3.1.4 | 3.1.5 | `<3.1.5` |
| socket.io-parser | 4.2.6 | 4.2.7 | `<4.2.7` |
| ip-address | 10.1.1 | 10.3.1 | `<=10.3.0` |

Result: `yarn npm audit --all --recursive --severity high` → **No audit suggestions**.

Scope: `package.json`, `packages/shared/package.json`, `yarn.lock`. No source changes — `main` already carries the `sharp` 0.35.3 bump and its type fix (`8e6217269`).

[...]

Issue #3620, opened 2026-06-25
refactor: provide explicit comparators for Array.sort() calls (implicit lexicographic ordering)


haxiorz:
Across the codebase, ~43 `Array.prototype.sort()` / `toSorted()` (and a couple of `readdirSync(...).sort()`) calls are invoked **without an explicit comparator**, so they fall back to JavaScript's default ordering — which coerces every element to a string and compares UTF-16 code units. This is implicit: it only happens to be correct because the arrays currently hold strings, and it silently produces wrong ordering for user-visible display lists (case sensitivity, accented characters) and would break outright if any of these arrays ever drifted to numbers or mixed types (`[2, 10]` sorts as `[10, 2]`).

The right comparator depends on what the array holds:

- **Display strings shown to a user** → `(a, b) => a.localeCompare(b)` so casing/diacritics order naturally for the operator's locale.
- **Fixed-format / canonical keys** (object keys for a cache-key or signature, code lists, enum values) → an explicit ordinal comparator `(a, b) => (a < b ? -1 : a > b ? 1 : 0)`, which makes the intended byte-stable ordering self-documenting and immune to type drift.
- **Numbers** → `(a, b) => a - b`.

## Location(s)

User-visible display sorts (should use `localeCompare`):
- `packages/core/src/modules/customers/components/calendar/CalendarScreen.tsx:226` — interaction-type filter options
- `packages/core/src/modules/payment_gateways/backend/payment-gateways/page.tsx:357` — provider dropdown options
- `packages/core/src/modules/communication_channels/backend/communication_channels/channels/[id]/page.tsx:162` — capability key list rendered in the detail `<dl>`
- `packages/ui/src/backend/detail/TagsSection.tsx:64` — normalized tag-label list

Canonical/internal key sorts (should use an explicit ordinal comparator):
- `packages/shared/src/lib/crud/custom-fields.ts:272`, `:273`, `:298` — cache-key fragments (`entityIds`, `organizationIds`, fieldset key parts)
- `packages/ui/src/backend/CrudForm.tsx:557` — `Object.keys(record).sort()` for the dirty-snapshot canonicalization
- `packages/core/src/modules/customers/backend/customers/deals/pipeline/page.tsx:602`–`605` — filter-signature fragments (status/owners/people/companies)
- `packages/core/src/modules/catalog/data/validators.ts:239` — `gtuCodes` dedupe+sort transform
[...]

PR #5324, opened 2026-08-15
fix(ci): fail on dead descriptor-keyed `resolutions` keys (#5098)

### LINKED ISSUE #5098: ci: guard against dead descriptor-keyed `resolutions` keys
## Problem

`resolutions` keys that carry an `@npm:<descriptor>` suffix only apply when that exact descriptor is present somewhere in the tree. When an upstream package changes its declared range, the key stops matching — and **Yarn does not warn about a resolution that matches nothing**. The pin silently stops applying and the dependency floats back up.

This is not hypothetical. All four `undici` keys on `main` had rotted this way:

| resolution key | descriptor actually in tree |
|---|---|
| `undici@npm:7.24.0` | — |
| `undici@npm:^7.19.0` | `^7.12.0` (cheerio) |
| `undici@npm:^8.3.0` | `^7.16.0` (`@newrelic/security-agent`) |
| `undici@npm:^8.4.1` | `^8.5.0` (testcontainers), `^8.7.0` (`@open-mercato/shared`) |

None matched, so undici floated to the vulnerable `7.28.0` / `8.8.0` and the `audit` job went red — with nothing in the diff to explain why. Fixed in #5085, but only for today's descriptors; the same rot will recur.

## Proposal

A small check that walks `resolutions`, and fails when a key carrying an `@npm:` descriptor matches no descriptor present in `yarn.lock`. Wire it into CI beside the audit job.

It would have caught all four dead `undici` keys before they cost a red pipeline. There are still 6 descriptor-keyed groups relying on this mechanism today (`ajv` ×4, `js-yaml` ×5, `undici` ×3, `ws` ×4, `@opentelemetry/core` ×3, `uuid` ×3), so the exposure is ongoing.

## Preferred alternative where it applies

Where the requester is **our own** workspace package, bump the direct dependency range instead of adding a descriptor-keyed pin — it is honest and cannot rot. #5085 does this for `packages/shared`'s `undici` (`^8.7.0` → `^8.10.0`), which removed one descriptor key outright. Descriptor keys should be reserved for third-party requesters we do not control.

_Found during #5085. Suggested by the automated review on that PR (Minor 7)._

adeptofvoltron:
Closes #5098

## 🎯 Goal
- A `resolutions` pin that has silently stopped applying now fails CI instead of surfacing weeks later as a red `audit` job with nothing in the diff to explain it. `yarn check:resolutions` walks the root `resolutions` map and fails when a range-carrying key matches no descriptor requested anywhere in `yarn.lock`.

## 🔍 Problem
A `resolutions` key that carries a range — `undici@npm:^8.4.1` rather than plain `undici` — only applies while that exact descriptor is requested somewhere in the tree. When the requester bumps its declared range the key stops matching, and **Yarn says nothing about a resolution that matches nothing**: the pin silently stops applying and the dependency floats back up. That is how all four `undici` keys rotted before #5085, and the only symptom was a red audit.

I checked whether Yarn already covers this before writing anything. It does not — `yarn install --check-resolutions` validates peer-dependency coherence, not the `resolutions` field. Verified empirically on a scratch project with a deliberately dead key (`is-number@npm:^99.0.0`, where the only requester asks for `^6.0.0`): `yarn install --check-resolutions` completed with zero warnings and installed the unpinned version.

## 🔍 Root Cause
Nothing in the repository ever compared the `resolutions` keys against the descriptors actually present in the lockfile, so a key could go dead and stay dead. #5085 re-derived the `undici` keys from the lock of the day, which fixes the instance but not the failure mode — the same rot recurs the next time an upstream package edits a range.

## What Changed

- **`scripts/check-resolutions.mjs`** (new) — reads the root manifest's `resolutions`, parses each key into its identifier and range, and fails when a range-carrying key matches no descriptor in `yarn.lock`. The failure output names the key, its target, and the ranges that *are* requested for that package, so the fix is usually a copy-paste. `--json` emits the same report machine-readably; `--root <dir>` points it at another checkout (used by the tests). Range-less keys (`"tar": "7.5.21"`) apply to every descriptor and cannot rot, so they are counted and skipped.

  Three parsing decisions are load-bearing and are documented in the script header:

  | Source | Counted as "present"? | Why |
  |---|---|---|
  | `dependencies` / `devDependencies` / `optionalDependencies` of a lock entry | ✅ | These are the descriptors Yarn actually resolves. A resolution rewrites the resolved *version* without rewriting the requester's declared range, so a live key still appears here. |
  | Lockfile entry keys | ❌ | Once a resolution applies, the entry is keyed by the pin's *target*. Accepting entry keys would let a key pointing at its own target (`"undici@npm:7.29.0": "7.29.0"`) look alive while matching no real requester. |
  | `peerDependencies` | ❌ | Yarn satisfies peer ranges from the ancestor tree rather than resolving them as descriptors of their own, so a resolution keyed on a peer range applies to nothing. |

  An ancestor-scoped key (`cheerio/undici@npm:^7.12.0`) is validated on its trailing descriptor only — a descriptor absent from the tree is dead regardless of which ancestor it was meant to narrow. A key written without a protocol is matched against both spellings, since a bare range means the npm protocol.

- **`.github/workflows/ci.yml`** — a `Check for dead resolution keys` step in the `audit` job, placed *before* the CVE scan because a dead key is a cause of the failure that scan reports. That job is the right home: `audit-scope` already gates it on exactly the inputs that can rot a key (manifests and lockfiles). Deliberately **not** added to `audit.yml`: the scheduled workflow exists to re-audit an *unchanged* graph against a moving advisory database, and a resolution key cannot rot without a lockfile change, so it would add no signal there.

- **`package.json`** — the new `check:resolutions` script, plus the checker's first catch: **six keys that had already rotted on `develop`**, removed.

  | Removed key | Target | Descriptors actually requested |
  |---|---|---|
  | `ajv@npm:~8.13.0` | 8.18.0 | `^6.12.5`, `^6.14.0`, `^8.0.0`, `^8.17.1`, `^8.9.0`, `~8.18.0` |
  | `undici@npm:7.24.0` | 7.29.0 | `^7.12.0`, `^7.16.0`, `^7.19.1`, `^7.28.0`, `^8.5.0`, `^8.9.0` |
  | `undici@npm:^7.19.0` | 7.29.0 | (same) |
  | `undici@npm:^8.3.0` | 8.9.0 | (same) |
  | `undici@npm:^8.4.1` | 8.9.0 | (same) |
  | `uuid@npm:^11.1.0` | 11.1.1 | `^8.3.2`, `^9.0.1`, `^11.1.0 \|\| ^12 \|\| ^13 \|\| ^14.0.0` |

  So four of the eight `undici` keys #5085 re-derived had **already rotted again** by the time develop caught up, which is the recurrence the issue predicted. The `uuid` one is the clearest illustration of the cost: `mermaid@11.16.0` widened its request to `^11.1.0 || ^12 || ^13 || ^14.0.0`, the `^11.1.0` key stopped matching, and uuid is resolving to **14.0.0** while the manifest reads as if it were pinned to 11.1.1.

  Removing them is a no-op by construction — a key that matches nothing changes nothing — and that is verified rather than asserted: `yarn install --mode=update-lockfile` after the removal leaves `yarn.lock` **byte-identical** (`git diff yarn.lock` empty). No new pins were added in their place: every unpinned `undici` descriptor left in the tree (`^7.19.1`, `^7.28.0`, `^8.9.0`) already resolves to 7.29.0 / 8.9.0, the exact versions the removed keys targeted, and `ajv@npm:~8.18.0` already resolves to the 8.18.0 the dead `~8.13.0` key targeted. Re-keying `uuid` to mermaid's current range would *downgrade* uuid 14.0.0 → 11.1.1, which is a separate decision and not one to smuggle into a CI guard.

## 🧪 Tests

- **`scripts/__tests__/check-resolutions.test.mjs`** (new, 14 cases) — fixture-driven, each building a throwaway manifest + lockfile pair and running the script through `--root`. It locks in: a live key passing; a dead key failing with the ranges that *are* requested; a pinned package absent from the tree entirely; range-less keys (including scoped ones like `@xmldom/xmldom`, which must not be mis-parsed into identifier `@xmldom` + range `xmldom`) being skipped; scoped descriptor keys matching correctly; a protocol-less key matching the `npm:` spelling; ancestor-scoped keys validated on the trailing descriptor; a peer-only range **not** counting as present; a lockfile entry key **not** counting as present, so `"undici@npm:7.29.0": "7.29.0"` still fails; an absent `resolutions` block passing; a missing lockfile failing with a clear message; `--json` emitting the dead/live split and still exiting non-zero; an unknown argument being rejected rather than silently checking the default root.
- The last case is the standing regression guard: **`the committed root manifest carries no dead resolution key`** runs the checker against the real repo root. `yarn test:scripts` runs unconditionally in CI (turbo never selects root-level scripts), so the repo is guarded in the `test` job as well as the `audit` job.
- `node --test scripts/__tests__/check-resolutions.test.mjs` — ✅ **14 passed**. It fails as expected against the pre-fix manifest (the checker reported all six dead keys, exit 1).
- `node --test scripts/__tests__/*.test.mjs apps/mercato/scripts/__tests__/*.test.mjs` (`yarn test:scripts`) — ✅ **601 passed / 0 failed**, including `repo-wide-guards.test.mjs`, `workflow-cache-poisoning.test.mjs` and `production-dependency-advisories.test.mjs`, which is what confirms the new `ci.yml` step does not trip the workflow guards.
- `yarn check:resolutions` — ✅ `22 descriptor-keyed resolutions all match a descriptor in yarn.lock (36 range-less keys apply to every descriptor and cannot rot)`.
- `yarn install --mode=update-lockfile` — ✅ no `yarn.lock` drift, which is the proof that the six removals change no resolution.
- **Gate commands not run locally, and why:** this worktree has no linked `node_modules`, so the turbo-driven half of `validation.commands` (`build:packages`, `generate`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `test`, `build:app`) could not run here. Nothing in the diff can reach them: no TypeScript, no package source, no locale file, no UI. The one jest test that reads the root `resolutions` map is `packages/create-app/src/lib/template-dependency-drift.test.ts`, and it iterates the *template's* resolutions (`lodash`, `lodash-es`, `zod`) — none of the six removed keys appear there. CI runs the full gate on this PR.

## 💥 Breaking Changes
- None. No contract surface from `BACKWARD_COMPATIBILITY.md` is touched, no dependency version changes (`yarn.lock` is byte-identical), and the new script is additive. The only behavioral change is that CI now fails on a condition it previously ignored.

## Notes for the reviewer
- The issue's preferred alternative — bumping our own workspace package's direct range instead of adding a descriptor-keyed pin — does not apply to any of the six removals: every one of them was keyed on a third-party requester's range, and the single first-party case (`packages/shared`'s `undici`) was already converted to a direct `^8.9.0` range by #5085.
- Scaffolded apps are not covered: the create-app template carries its own small `resolutions` block and no monorepo lockfile, and template pin inheritance is tracked separately in #5099.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5324"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789404211&installation_model_id=432243&pr_number=5324&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5324&signature=32a716d722fe669c9f4e282ec861582c4d8339c2490d098fd7ae89c286208321"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:

- 🔍 `review` — the run is complete and the PR is ready for code review; nothing is left in progress on the authoring side.
- 🐛 `bug` — it inherits the category of #5098 and it repairs a live defect, not just a guard: six `resolutions` keys had already stopped applying on `develop`, so `uuid` is resolving to 14.0.0 while the manifest reads as if it were pinned to 11.1.1.
- 📦 `dependencies` — the change edits the root `resolutions` map and the CI job that guards the dependency graph, so anyone tracking dependency-surface changes should see it.
- ⏭️ `skip-qa` — it qualifies for the automated-verification exemption in `AGENTS.md`: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no database or API surface change, no `BACKWARD_COMPATIBILITY.md` contract touched, and 14 automated tests for the new behavior ship in the same PR. There is no user-facing surface a manual tester could exercise.
- 🔹 `priority-medium` — it matches the priority already set on #5098; the rot is real and ongoing across six descriptor-keyed groups, but nothing is currently exploitable and the audit gate is green today.
- 🟢 `risk-low` — the blast radius is one new root-level script, one CI step, and six manifest keys that by definition matched nothing. `yarn install --mode=update-lockfile` leaves `yarn.lock` byte-identical, so no dependency resolution changes; the worst realistic failure is CI reporting a dead key that a reviewer disagrees with.

adeptofvoltron:
🤖 `om-open-pr` — chain lock transferred onto this PR at 2026-08-15T16:44:13Z. Continuing with `om-auto-review-pr`; other auto-skills should skip this PR until the lock is released.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock at 2026-08-15T16:44:52Z to review this PR.

adeptofvoltron (commented):
# 🔍 Code Review

Re-review after `8d0629e35`, which applied this review's own two findings.

## 🎯 Summary

The PR adds `scripts/check-resolutions.mjs`, a guard that fails when a range-carrying `resolutions` key matches no descriptor requested in `yarn.lock`, wires `yarn check:resolutions` into the `audit` job ahead of the CVE scan, and removes the six keys the guard immediately found dead on `develop`. The scope is exactly what #5098 asked for, and the manifest cleanup that comes with it is verified inert rather than assumed inert.

The design decision worth checking hardest is what the script counts as "a descriptor present in the tree", because getting it wrong in the lenient direction makes the guard useless and getting it wrong in the strict direction turns CI red on a healthy repo. The choice made here — lock entries' `dependencies` maps, but **not** lockfile entry keys and **not** `peerDependencies` — is right on both counts and is now locked in by dedicated tests. Excluding entry keys matters more than it looks: a resolution rewrites the entry key to the pin's *target*, so counting entry keys would let `"undici@npm:7.29.0": "7.29.0"` look alive while matching no real requester (`check-resolutions.test.mjs:180`).

## Verdict

**Approve** — no blockers and no majors. One minor finding is pre-existing and deliberately out of scope; it is recorded below as a follow-up rather than fixed here.

Note on mechanics: GitHub does not allow approving one's own pull request, so this is submitted as a review **comment** rather than an `APPROVE` event, and the PR stays in the `review` pipeline state. A maintainer's approval is still required — this review does not stand in for it, and the PR was not moved to `merge-queue`.

## 🧪 Validation Gate

| Command | Result | Notes |
|---|---|---|
| `node --test scripts/__tests__/check-resolutions.test.mjs` | ✅ 14 passed | The new suite; also confirmed failing against the pre-fix manifest. |
| `node --test scripts/__tests__/*.test.mjs` (`yarn test:scripts`) | ✅ 566 passed / 0 failed | Includes `repo-wide-guards`, `workflow-cache-poisoning`, `preview-workflows` and `production-dependency-advisories` — the guards that would object to the new `ci.yml` step. An earlier run over both roots (`scripts/` + `apps/mercato/scripts/`) was ✅ 601/601. |
| `yarn check:resolutions` | ✅ | `22 descriptor-keyed resolutions all match a descriptor in yarn.lock (36 range-less keys apply to every descriptor and cannot rot).` |
| `yarn install --mode=update-lockfile` | ✅ no drift | `yarn.lock` byte-identical after the six removals — the evidence that they were genuinely inert. |
| `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app` | ⚠️ not run locally | This worktree has no linked `node_modules`, so the turbo-driven half of the gate could not execute. Nothing in the diff can reach it: no TypeScript, no package source, no locale file, no UI. The one jest test that reads the root `resolutions` map, `packages/create-app/src/lib/template-dependency-drift.test.ts:95`, iterates the *template's* resolutions (`lodash`, `lodash-es`, `zod`), none of which are touched. CI runs the full gate — and on the first head, the `audit` job it now contains passed, which is live confirmation the new step is wired correctly. |

## Findings

### 🟡 Minor

- **`package.json:237` — `"fast-xml-builder": "1.2.0"` pins a package nothing in the tree requests.** Found while validating the checker's scope decision: no `dependencies` block in `yarn.lock` names `fast-xml-builder`, so this pin has never applied. It is not the failure mode #5098 describes — a range-less key cannot *rot*, because it was never matched to a moving descriptor — and it may be a deliberate forward-looking pin against a package we expect to appear. Deliberately **not** fixed here and deliberately not made a hard failure: deciding whether an orphan range-less pin should fail CI is a separate judgement call from the one this PR was asked to make. `scripts/check-resolutions.mjs:33` now records the reasoning so the next reader does not have to re-derive it, and this is worth its own small issue.

### ✅ Findings raised and already resolved in `8d0629e35`

- **`scripts/check-resolutions.mjs:20` — the header overstated which lockfile blocks exist.** It listed `dependencies`, `devDependencies` and `optionalDependencies` as "the descriptors Yarn actually resolves", implying a lockfile carries all three. It does not: `grep -oE '^  (dependencies|devDependencies|optionalDependencies|peerDependencies):' yarn.lock | sort | uniq -c` returns only `dependencies` (1741) and `peerDependencies` (661) — workspace `devDependencies` are folded into the single `dependencies` block, confirmed on `"open-mercato@workspace:."` at `yarn.lock:25955`. The code was already correct; the comment was not, and a future maintainer could have removed the "redundant" fields or, worse, trusted the wrong mental model when extending the parser. Now stated accurately: the two extra fields are read as a forward-compatible safety net, and since extra sources can only make the check more lenient, they can never turn a healthy key red.
- **`scripts/check-resolutions.mjs:213` — the success line did not agree in number.** With exactly one descriptor-keyed resolution it read `1 descriptor-keyed resolutions all match…`, while the failure path already branched correctly on singular/plural. Fixed on both halves of the sentence, with the assertion in `check-resolutions.test.mjs:63` updated to lock the singular form in.

## 💥 Breaking Changes

- [x] **Exported APIs** — none touched; the new script is a standalone executable with no importable surface.
- [x] **HTTP routes and response shapes** — untouched.
- [x] **Event names / widget spot IDs / DI keys / ACL features / notification IDs** — untouched.
- [x] **CLI commands** — `check:resolutions` is a new root `package.json` script; additive, nothing renamed or removed.
- [x] **DB schema** — untouched; no migration expected and none present.
- [x] **Config formats** — untouched.
- [x] **Generated files** — none regenerated; `yarn.lock` is byte-identical, verified rather than assumed.
- [x] **`BACKWARD_COMPATIBILITY.md`** — no protected surface in any of the 13 categories is affected. Removing a `resolutions` key is not a contract change: the six removed keys applied to nothing, so no consumer's resolved version moves.

The only behavioral change is that CI now fails on a condition it previously ignored. That is the point of the PR, and the failure message names the key, its target, and the ranges actually requested, so it is self-servicing rather than a mystery red.

## 🧪 Test Coverage

Coverage is proportionate and, unusually for a CI-guard PR, targets the parsing decisions rather than just the happy path. The fourteen cases in `scripts/__tests__/check-resolutions.test.mjs` each build a throwaway manifest and lockfile and drive the real script through `--root`, so they exercise the shipped binary rather than an extracted helper.

The cases that earn their place are the ones covering the ways this check could be silently wrong: a peer-only range not counting as present (`:169`), a lockfile entry key not counting as present so a self-targeting pin still fails (`:180`), a scoped range-less key like `@xmldom/xmldom` not being mis-parsed into identifier `@xmldom` plus range `xmldom` (`:106`), a protocol-less key matching the `npm:` spelling (`:139`), and an ancestor-scoped key being validated on its trailing descriptor (`:155`). Failure-path output is asserted on content, not just exit code, so a regression that stops naming the requested ranges would be caught.

The standing guard is `the committed root manifest carries no dead resolution key` (`:246`), which runs the checker against the real repo root. Because `yarn test:scripts` runs unconditionally in CI — turbo never selects root-level scripts — the repository is protected in the `test` job as well as the `audit` job, so a dead key cannot slip in on a PR that leaves the manifests untouched.

One coverage gap is worth naming rather than leaving implicit: there is no test that the `ci.yml` step is present and ordered before the CVE scan. The existing workflow guards (`preview-workflows.test.mjs`, `workflow-cache-poisoning.test.mjs`) assert other properties of the workflows, so the pattern exists, but adding one here would be over-fitting — the step's presence is already proven by the `audit` job passing on this PR's first head.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the guard is implemented, self-reviewed and green locally — no blockers, no majors, and the review's own two findings were applied in `8d0629e35`. The one minor finding (`fast-xml-builder`, a range-less orphan pin that has never applied) is pre-existing and deliberately left for its own issue rather than widened into this guard.

The `in-progress` lock is released and swapped for `ci-monitoring`: the work is finished and reported, and only the CI result on `8d0629e35` is still being watched. The PR is free for anyone to act on.

Manual QA is not requested — this carries `skip-qa` under the automated-verification exemption in `AGENTS.md` (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no database or API change, no contract touched, and 14 automated tests in the same PR). A maintainer's approval is still required: GitHub does not permit self-approval, so the review above was submitted as a comment and the PR stays in `review`. Lock released.

adeptofvoltron:
🤖 `om-auto-review-pr` — CI result on `8d0629e35`: ❌ **`test` failed**, and it was this PR's fault, not a flake.

The failing step was `Run repo-wide audit guards`, specifically `packages/core/src/__tests__/explicit-sort-comparators.test.ts` (#3620), which scans `scripts/` as well as every package `src` root:

```
scripts/check-resolutions.mjs:150
scripts/check-resolutions.mjs:157
```

Both `.sort()` calls were bare, relying on JavaScript's default string coercion. Fixed in `8a506a3bc` with the code-unit comparator that guard prescribes for canonical keys — `(a, b) => (a < b ? -1 : a > b ? 1 : 0)` — rather than `localeCompare`, because these are descriptors and ranges whose ordering must be reproducible across machines, not display strings.

Worth naming plainly: my local gate missed this because `yarn test:repo-wide-guards` needs a linked `node_modules` this worktree does not have, so I ran `yarn test:scripts` and stopped there. That was the gap. I have since run the guards directly against this worktree — all **11 core repo-wide guards, 38 tests, green** — along with `yarn check:resolutions` ✅ and the full scripts suite ✅ **601 passed / 0 failed**.

Everything else on `8d0629e35` was green, including `audit` — the job that now carries the new step. The `FAIL source-link-inventory` lines visible under `Check create-app template parity` are that suite's own fixture output asserting the gate rejects a stale inventory; those cases passed.

Still watching CI on `8a506a3bc`.

adeptofvoltron:
🤖 `om-auto-review-pr` — CI result on `8a506a3bc`: ✅ **all checks green.**

`test` ✅ 3m15s · `audit` ✅ · `audit-scope` ✅ · `lint` ✅ · `ds-lint` ✅ · `prepare` ✅ · `scope` ✅ · CodeQL ✅ (actions, javascript-typescript, python). The remaining entries are conditional jobs correctly skipping on this diff (`docker-build`, `ephemeral-integration`, `mutate`, `merge-coverage`, `comment`).

Two things worth reading out of that run rather than just the checkmarks:

- The `explicit-sort-comparators` regression from `8d0629e35` is confirmed cleared — `test` now passes the same `Run repo-wide audit guards` step that failed before, so the comparator fix in `8a506a3bc` was the whole of it and nothing else was hiding behind it.
- `audit` passing **is** the new step passing. That job now runs `yarn check:resolutions` ahead of the CVE scan, so this green is end-to-end proof of the wiring, not just of the untouched audit.

`ci-monitoring` is removed; no further follow-up will come from this agent. The PR stays in `review` awaiting a maintainer's approval — GitHub does not permit self-approval, so nothing here substitutes for it, and it was deliberately not moved to `merge-queue`.

pkarw (approved):

PR #5064, opened 2026-08-06
fix(create-app): remove the dist/agentic build race and make truncated runs self-explanatory

wojciechszyjka:
Closes #5059
Closes #5052
Closes #5091
Tracking plan: `.ai/runs/2026-08-06-create-app-test-suite-stability.md`
Status: complete

## 🎯 Goal

Make `yarn workspace create-mercato-app test` deterministic: no test file may rebuild `dist/agentic` while another file reads it (#5059), and a run that is cut short from outside must say so instead of listing 26+ files under `✖ failing tests:` next to `fail 0` (#5052). Both issues share one aggravating cause — the package build (esbuild plus a ts-morph extraction of ~55 module fact-sheets) was executed *from inside* the test files, concurrently, several times per run. #5091 is the ENOTEMPTY-shaped sibling of #5059 — the same `rmSync(..., { recursive: true })` on a directory another build is repopulating, and its own "Suggested direction" asks verbatim for the staged swap implemented here — so it is closed by the same change.

## What Changed

- **`packages/create-app/build.mjs` — the race is removed at its source (#5059).** The agentic tree is now assembled in `dist/agentic.staging` and published with two renames at the end of the build, instead of deleting `dist/agentic` and copying the harness plus the fact-sheets back in place while other processes read it. Sampling every 5 ms across one build, `guides/module-facts.json` used to be absent for 5.28 s of a 6.13 s build (98%), `guides/modules/customers.md` 97.8%; after the change the window is 0.0%. Because the staged tree is built from `agentic/` alone, the purge of legacy `core.<module>.md` stubs and package-level guides became structurally unnecessary and was removed.
- **`packages/create-app/package.json` — the suite builds once, before the runner.** The `test` script runs the build ahead of `node --test`; `module-facts-build.test.ts` and `ready-apps.test.ts` no longer spawn `build.mjs` and instead call the new `requirePackageBuild()` (`src/lib/package-build-artifacts.ts`), which fails with the command to run rather than a bare `ENOENT`.
- **`packages/create-app/scripts/report-truncation.mjs` — a truncated run explains itself (#5052).** A second node:test reporter states that the run was truncated, how many files were cancelled, whether anything actually failed, and how to judge this suite on its own. Both CI runs the issue cites were read first: the suite was aborted by turbo after a *different* package's task failed, so this PR raises no timeout, pins no `--test-concurrency` and touches no oracle test — it fixes what the evidence shows.
[...]

Issue #5059, opened 2026-08-06
Fix: parallel create-app test files race on build.mjs wiping dist/agentic (ENOENT on files that exist)


wojciechszyjka:
## Summary

- Three `create-mercato-app` test files each spawn `node build.mjs`, and `build.mjs` starts by deleting `dist/agentic/`. Node's test runner runs files in parallel, so under load one build wipes the tree another build or assertion is mid-way through reading, and the suite fails with `ENOENT` on files that exist.

## Analysis

**The delete-then-copy window** — `packages/create-app/build.mjs`:

```js
rmSync('dist/agentic', { recursive: true, force: true })
cpSync('agentic', 'dist/agentic', { recursive: true })
```

`dist/agentic/` is ~55 module fact-sheets plus the whole harness tree, so the window between the `rmSync` and the end of the `cpSync` is not short.

**Three concurrent entrants** — each of these spawns `build.mjs` independently, and `node --test` runs test *files* in parallel (concurrency defaults to the CPU count):

- `packages/create-app/src/lib/module-facts-build.test.ts` (`ensureBuilt()` → `execSync('node build.mjs')`)
- `packages/create-app/src/lib/ready-apps.test.ts:175` (`spawnSync(process.execPath, ['build.mjs'])`)
- `packages/create-app/src/lib/agent-instruction-budget.test.ts`

Each file has its own `buildComplete` memo, so the memo only dedupes *within* a file — it cannot prevent three processes from racing.

**Observed failures.** Two consecutive full-repo `yarn test` runs on the same clean tree failed at *different* tests, each with the signature of a half-deleted `dist/agentic`:

```
✖ every default-controller module fact is exercised by the evaluation catalog
  Error: ENOENT: no such file or directory, open
  '.../packages/create-app/dist/agentic/shared/ai/harness/cases.json'

✖ build emits customers facts and the framework extension catalog (T5)
  Error: Command failed: node build.mjs

✖ published CLI bin executes the dist entrypoint
  AssertionError: expected package build to succeed
  stderr: libc++abi: terminating due to uncaught exception of type
  std::__1::__fs::filesystem::filesystem_error: filesystem error:
  in create_directory: No such file or directory ["dist/agentic/claude-code"]
```

The same suite run on its own (`yarn workspace create-mercato-app test`) passed **451/0 on three consecutive runs** on that same tree. A different test failing each time, on files that are present, with the suite green when it is not competing for CPU, is a scheduling race — not an assertion problem.

[...]

Issue #3754, opened 2026-07-03
Retire the legacy guides/core.<module>.md redirect-stub layer


adeptofvoltron:
## Follow-up from #3715

#3715 replaced the 9 hand-written per-module standalone guides with generated fact-sheets at `.ai/guides/modules/<module>.md`. To honor the generated-file BC contract it also emits 9 legacy `guides/core.<module>.md` redirect stubs (~200 B each) whose entire content is a "this guide has moved, see `modules/<module>.md` / `module-system.md`" pointer. This issue tracks retiring that redirect layer.

The PR's own BC analysis already flags it: fresh D6-scaffolded apps link `modules/<module>.md` directly and reference **neither** the stubs nor the old names — the stubs exist purely as a bridge for apps upgrading in place.

## Two-part cleanup (different BC profiles)

**Part 1 — stop emitting stubs for fresh scaffolds (no BC risk, doable immediately once #3715 merges).**
A freshly scaffolded app never had the old `core.<module>.md` names, so shipping 9 unreferenced stubs into it is pure dead weight. `build.mjs` / `shared.ts` should skip the legacy stubs for fresh scaffolds while still bundling them for the in-place upgrade path. The #3715 risk table suggests exactly this ("have the D6 filtered copy skip legacy stubs for fresh scaffolds").

**Part 2 — remove stubs from the upgrade path (gated by the deprecation window).**
Generated-guide paths are a contract surface (`BACKWARD_COMPATIBILITY.md`, category 13: generated files). Existing apps upgraded in place may still link the old names, so the stubs must live ≥1 minor with a `RELEASE_NOTES.md` deprecation note (added in #3715) before removal. Only remove after that window closes.

### Acceptance criteria
- [ ] Fresh D6 scaffolds no longer emit `guides/core.<module>.md` stubs (Part 1).
- [ ] Upgrade path keeps the stubs until the deprecation window from #3715's RELEASE_NOTES note has elapsed.
- [ ] Once the window closes, remove stub generation entirely from `build.mjs`/`shared.ts` and drop the stubs from the emitted guide set (Part 2).
- [ ] `RELEASE_NOTES.md` records the final removal.

## Notes
- MR will be opened after #3715 is merged; the maintainer decides whether Part 1 lands early and when Part 2's window has closed.

Related: #3715

adeptofvoltron:
[...]

PR #5328, opened 2026-08-15
refactor(cli): publish dist/agentic through a staged swap (#5104)

### LINKED ISSUE #5104: refactor(cli): publish dist/agentic through a staged swap, as create-app now does
Follow-up from the review of #5064 (@adeptofvoltron), filed so the finding is not dropped with the review comment.

## What

`packages/cli/build.mjs:30-31` still refreshes its own `dist/agentic` **in place**:

```js
rmSync(join(outdir, 'agentic'), { recursive: true, force: true })
cpSync(agenticSrc, join(outdir, 'agentic'), { recursive: true })
```

That is the exact shape #5059 was about, and which #5064 removed from `packages/create-app/build.mjs` by assembling the tree in `dist/agentic.staging` and publishing it with two renames. Between the `rmSync` and the end of the `cpSync`, any concurrent reader of `@open-mercato/cli/dist/agentic` sees a missing or partial tree and fails with `ENOENT` on files that exist. Measured on the create-app side before the fix (5 ms sampling across one build): `guides/module-facts.json` was absent for 98% of the build.

## Why it is not urgent

Nothing in the repository reads that tree while the CLI is building today — only the separate integration runner spawns this build — so this is a **latent** bug, not an observed failure. It was correctly left out of #5064's scope; it becomes real the moment anything reads the CLI's agentic tree concurrently, or the build is run in parallel with a generator that does.

## Suggested fix

Mirror what `packages/create-app/build.mjs` now does:

1. assemble into `dist/agentic.staging`;
2. `renameSync` the old tree aside and the staged tree in;
3. remove the previous tree.

The window shrinks to the gap between the two renames, and stale per-module artifacts become structurally impossible, so no purge step is needed. A guard test asserting neither `agentic.staging` nor `agentic.previous` survives a completed build is worth carrying over too (`packages/create-app/src/lib/module-facts-build.test.ts` has one), because `dist/` is published.

Note the CLI build already uses `atomicWriteFileSync` for `bin.js` with a comment about concurrent consumers — the directory copy is the remaining non-atomic step next to it.

Possibly related but distinct: #5091 (create-app `build.mjs` racing its own `dist/` clean with `ENOTEMPTY`).

adeptofvoltron:
Closes #5104

## 🎯 Goal
Publishing `@open-mercato/cli`'s `dist/agentic` tree stops being observable in a half-written state: a reader of the CLI's agentic tree now sees either the complete previous build or the complete new one, never a directory that is missing files it is documented to contain.

## 🔍 Problem
`packages/cli/build.mjs` refreshed its own `dist/agentic` **in place** — `rmSync` the published tree, then `cpSync` the whole harness plus ~55 generated module fact-sheets back into it. Between those two calls the published path is missing or partial, so any concurrent reader of `@open-mercato/cli/dist/agentic` fails with `ENOENT` on files that exist. This is the exact shape #5059 described and #5064 removed from `packages/create-app/build.mjs`; it was correctly left out of that PR's scope and filed here as a latent bug, since nothing in the repository reads the CLI's tree while the CLI is building today.

Measured on this branch by sampling `dist/agentic/guides/module-facts.json` every 5 ms across one full CLI build:

| `packages/cli/build.mjs` | samples | probe missing |
|---|---|---|
| before (in-place refresh) | 2234 | 1634 — **73.1% of the build** |
| after (staged swap) | 1880 | 0 — **0.0%** |

## 🔍 Root Cause
The in-place refresh makes the *published* directory the working directory: it is deleted and then rebuilt file by file, so its incomplete intermediate states are exactly what a concurrent reader observes. The build already used `atomicWriteFileSync` for `bin.js` for this reason — the directory copy next to it was the remaining non-atomic step.

## What Changed
- `packages/cli/build.mjs` now assembles the tree in `dist/agentic.staging` — the create-app source copy, the upstream instruction boundary, the discovered standalone guides, and every generated fact-sheet/sidecar write there — and publishes it at the end of `afterBuild` with two `renameSync` calls (`dist/agentic` → `dist/agentic.previous`, `dist/agentic.staging` → `dist/agentic`) followed by removal of the previous tree. The exposure window shrinks from seconds to the gap between two renames.
- The stale-artifact purge is gone with it. It existed because an incremental `dist/` could retain a removed module's guide or a legacy `core.<module>.md` redirect stub (#3754); a staged tree is built from `packages/create-app/agentic` alone, so anything this build does not write is structurally absent. The comments now say that instead of describing a purge that no longer runs.
- `packages/cli/src/lib/__integration__/TC-INT-008.spec.ts` gains a guard asserting that neither `dist/agentic.staging` nor `dist/agentic.previous` survives a completed build. That spec's `beforeAll` already runs the real `packages/cli/build.mjs`, so it is the one place under test where a completed CLI build exists — and `dist/` is published, so a staging tree the build forgot to swap in would ship with the package.
- `packages/cli/src/lib/__tests__/agentic-build-artifacts.test.ts` (new) asserts the same invariant from the cli jest suite. CI's integration scope skips the Playwright suite for a diff confined to `packages/cli/**` — which is exactly the diff that can break `build.mjs` — so the guard also lives where the affected-package filter (`turbo run test --filter=[origin/develop]...`) always selects it. The `test` job downloads the `build-artifacts` that `prepare`'s `yarn build:packages` produced, so `packages/cli/dist` is present there and the assertion reads the tree the package would publish. The cli `test` script is plain `jest` and never builds first, so an unbuilt package skips rather than failing a clean local checkout.
- The final publish log line now reads `Published dist/agentic.staging/ → dist/agentic/` instead of `Published create-app/agentic/ → dist/agentic/`. The old wording ran unconditionally while the matching "Copied" line is guarded by `existsSync(agenticSrc)`, so a build without the create-app source claimed to have published a tree it never copied. Nothing in the repository parses this output.

This mirrors `packages/create-app/build.mjs` deliberately, including its single-builder assumption: the staging and previous directory names are fixed, so two concurrent runs of this build would clear each other's staging tree and the loser would fail loudly on the rename. Only TC-INT-008 spawns this build, and the Playwright config runs with `workers: 1`, so nothing in the repository builds the package concurrently — and a loud failure beats the silent partial tree the in-place refresh produced.

## 🧪 Tests
- `npx playwright test packages/cli/src/lib/__integration__/TC-INT-008.spec.ts` — 4 passed, including the new staging-artifact guard.
- `yarn workspace @open-mercato/cli test` — 1705 passed / 89 suites (was 1703 / 88 before the jest guard was added). CI confirms the guard executes rather than skipping: the `test` job on head `6d7bbf6fe` reports `PASS src/lib/__tests__/agentic-build-artifacts.test.ts` with `Tests: 1705 passed, 1705 total` and no skips.
- The new guard was exercised in all three states — unbuilt package (skips, suite green), build aborted midway leaving `dist/agentic.staging` behind (fails, so it is not vacuous), completed build (passes against the real published tree).
- `yarn build:packages` — all 22 workspace builds successful; the CLI build was also run directly and verified to publish a complete tree (56 module fact-sheets, both sidecars, the reference projection, the upstream manifest) with no `agentic.staging` / `agentic.previous` left behind, on both a cold `dist/` and an incremental one.
- `yarn turbo run typecheck --filter=@open-mercato/cli` and `eslint` on both changed files — clean.
- The 5 ms sampling comparison in the Problem section above, run against the pre-fix `build.mjs` and this one.
- Gate commands skipped as unrelated to a build-script change that touches no module files, locale strings, or app code: `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, the unfiltered `yarn test`, and `yarn build:app`.

Note on the guard's shape: it is a safety net for the new design rather than a test that fails without the fix — the pre-fix build never created a staging directory, and the defect itself is a timing window that only a sampling harness can observe. It matches the guard #5064 shipped for create-app (`packages/create-app/src/lib/module-facts-build.test.ts`), and the timing evidence is reported above instead of encoded as a flaky in-CI sampler.

## 💥 Breaking Changes
- None. The published `dist/agentic` layout is byte-for-byte what the previous build produced; only the order of operations that puts it there changed.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5328"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789404535&installation_model_id=432243&pr_number=5328&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5328&signature=94db0b79de1b3cd06fa0e379b6586230465b9a70020bf8bbe6771aedb140ba20"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:

- 🔍 `review` — the run is complete and the PR is ready for code review; nothing is blocked or in flight.
- ♻️ `refactor` — the published `dist/agentic` layout is unchanged, so this reorders how the build produces its output rather than changing what the package ships.
- ⏭️ `skip-qa` — the automated-verification exemption applies: no UI-rendering file is touched (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`), the database structure and API surface are unchanged, no `BACKWARD_COMPATIBILITY.md` contract is affected, and a guard test for the changed build behavior ships in this PR.
- 🔽 `priority-low` — the issue documents this as latent rather than observed: nothing in the repository reads the CLI's agentic tree while the CLI is building today, so the race has no current victim and the fix is cleanup ahead of one.
- 🟢 `risk-low` — the blast radius is one package's build script plus one integration spec; the swapped-in tree was verified byte-for-byte equivalent on both a cold and an incremental `dist/`, and a failure mode here surfaces loudly at build time rather than silently at runtime.

adeptofvoltron:
🤖 Chain lock transferred from #5104 to this PR by @adeptofvoltron. `om-auto-review-pr` continues here; other auto-skills will skip this PR until the lock is released.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock held by @adeptofvoltron on this PR. Review starting now; the lock is retained for the chain.

adeptofvoltron:
# 🔍 Code Review

**Verdict: approve** — no blockers and no majors. Two minor observations are recorded below for the human reviewer; neither justifies holding the PR.

GitHub rejects self-approval, so this report is posted as a comment rather than submitted through the review API: the PR was authored by the same automation account that reviewed it. The pipeline label therefore stays `review` rather than moving to `merge-queue` — a human approval is still outstanding, and this comment must not be mistaken for one.

## 🎯 Scope reviewed
The full diff of `packages/cli/build.mjs` and `packages/cli/src/lib/__integration__/TC-INT-008.spec.ts` at head `f7b0209f4`, read against the reference implementation this PR mirrors (`packages/create-app/build.mjs`, merged in #5064) and against the behavior the removed purge block used to provide.

## ✅ What holds up
- **The swap is correct and complete.** Every writer that previously targeted `dist/agentic` now targets `agenticStaging` — the create-app source copy (`build.mjs:54`), the upstream instruction boundary (`build.mjs:59`), the discovered standalone guides (`build.mjs:74`), and every generated fact-sheet, sidecar, and reference projection, all of which route through `guidesDestDir`. No path writes into the published tree before the rename at `build.mjs:198-201`, which is the property the whole change rests on. I re-read the file rather than the diff to confirm no writer was missed.
- **Removing the stale-artifact purge is sound, not a regression.** The purge existed because an incremental `dist/` could retain a removed module's guide, a legacy `core.<module>.md` redirect stub (#3754), or an unreachable package-level `<pkg>.md`. A staged tree is assembled from `packages/create-app/agentic` alone, and that source contains only the conceptual guides — so those artifacts cannot exist unless this build writes them, which it does not. TC-INT-008 independently asserts their absence (`UNROUTED_PACKAGE_GUIDES`, and the `core.<module>.md` guard) and passes.
- **The single-builder assumption is documented rather than assumed.** The fixed `agentic.staging` / `agentic.previous` names mean two concurrent builds of this package would clobber each other, with the loser failing loudly at the rename. That is verified as vacuous today: `packages/cli/src/lib/__integration__/TC-INT-008.spec.ts` is the only test that spawns this build, and the Playwright config runs `workers: 1`. A loud failure is also strictly better than the silent partial tree the in-place refresh produced.
- **The evidence in the PR body is reproducible.** Sampling `dist/agentic/guides/module-facts.json` every 5 ms across a full build moved from 1634/2234 samples missing (73.1%) on the pre-fix script to 0/1880 (0.0%) here, which is the defect #5104 describes and its removal.

## 📝 Minor observations (non-blocking)
- `packages/cli/build.mjs:52-56` — when `packages/create-app/agentic` is missing, behavior changes subtly: the old code left the previously published tree in place and wrote the generated artifacts on top of it, whereas the staged build now publishes a tree containing only the generated artifacts. This cannot occur in the monorepo layout the script lives in (`create-app` is always a sibling package in a checkout), the previous behavior was accidental rather than intentional — the old code already produced a mixed old/new tree in that case — and `packages/create-app/build.mjs` has the identical shape. Mirroring the reference implementation is the right call here; flagging it only so the parity is a decision on the record.
- `packages/cli/build.mjs:198-201` — a build that dies between the two renames leaves no `dist/agentic` at all, with the old tree parked in `dist/agentic.previous`, and the next build's leftover sweep discards it before republishing. That is a strictly narrower failure window than the pre-fix behavior and is again identical to create-app's accepted design, so it needs no code; noted for completeness.

## 🧪 Tests
- The new guard in `TC-INT-008.spec.ts:298-306` asserts that neither `dist/agentic.staging` nor `dist/agentic.previous` survives a completed build, and its `beforeAll` runs the real build first — so the assertion is made against genuine post-build state rather than a fixture. `npx playwright test packages/cli/src/lib/__integration__/TC-INT-008.spec.ts` → 4 passed.
- The PR body is candid that this guard is a safety net rather than a fails-without-the-fix test, and explains why: the pre-fix build never created a staging directory, and the defect itself is a timing window. That reasoning is accepted here — the alternative, a sampling harness inside the Playwright suite, trades a real regression signal for a timing-sensitive test that would flake on loaded CI runners, and #5064 set the precedent with the same guard shape for create-app.
- `yarn workspace @open-mercato/cli test` → 1703 passed / 88 suites; `yarn build:packages` → 22/22; `yarn turbo run typecheck --filter=@open-mercato/cli` and `eslint` on both changed files → clean. The gate commands skipped (`yarn generate`, both i18n checks, the unfiltered `yarn test`, `yarn build:app`) are disclosed in the PR body and are unrelated to a build-script change that touches no module files, locale strings, or app code.

## 🔁 Backward compatibility
No contract surface is affected. The published `dist/agentic` layout is unchanged — verified by a full build producing 56 module fact-sheets, both JSON sidecars, the reference projection, and the upstream manifest, on both a cold and an incremental `dist/` — so no consumer of `@open-mercato/cli/dist/agentic` sees a different tree. Only the order of operations that produces it changed.

## ⏱️ CI
Required checks were still running when this review was written (`test` queued, CodeQL `Analyze (javascript-typescript)` in progress); `lint`, `ds-lint`, `audit-scope`, `prepare`, `scope`, and the other CodeQL analyses had already passed, and nothing was failing. The verdict above is not contingent on that wait — a CI result comment follows separately, and required checks still gate the merge regardless of what this review says.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the staged-swap refactor is implemented, reviewed (approve — no blockers or majors; self-approval is not possible through the review API, so a human approval is still outstanding) and fully labeled. UI verification is not applicable: the diff touches only a package build script and one integration spec, with no user-facing surface. Lock released; `ci-monitoring` is set only because the required checks were still running when the work finished, and it will be removed once the CI result is posted.

adeptofvoltron:

✅ Every required check on `f7b0209f4` settled green: `test`, `lint`, `ds-lint`, `audit-scope`, `prepare`, `scope`, and all four CodeQL analyses (`javascript-typescript`, `python`, `actions`, plus the aggregate `CodeQL` gate). Nothing failed, and the remaining entries (`docker-build`, `ephemeral-integration`, `mutate`, `audit`, `merge-coverage`, `[code]smith`) were skipped by their own path and event filters rather than by this run.

The `ci-monitoring` label is removed accordingly. The PR stays in `review`: the code review above found no blockers or majors, but GitHub does not accept an approval from the account that authored the PR, so a human approval is still required before merge.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T10:28:21Z. Other auto-skills will skip this PR until the lock is released.

The previous chain's lock was explicitly released on 2026-08-15 (`om-auto-fix-issue` run complete … Lock released), the `in-progress` label is absent, and the only remaining signal was the author's own assignee entry — so this is a stale/released claim, taken up as a fresh human-account review rather than an override.

pkarw (changes requested):
# 🔍 Code Review: refactor(cli): publish dist/agentic through a staged swap (#5104)

## 🎯 Summary

This PR converts `packages/cli/build.mjs` from refreshing its published `dist/agentic` in place to assembling the tree in `dist/agentic.staging` and swapping it in with two `renameSync` calls, and drops the stale-artifact purge that the in-place refresh needed. It mirrors `packages/create-app/build.mjs` (merged in #5064) and implements exactly what issue #5104 prescribed, guard test included. The reviewed scope is the full diff of `packages/cli/build.mjs` and `packages/cli/src/lib/__integration__/TC-INT-008.spec.ts` at head `f7b0209f4`, read against the create-app reference implementation, the shared builder in `scripts/build-package.mjs`, and `.ai/lessons/package-builds-that-publish-dist-must-clear-stale.md`.

The change is correct and the reasoning behind it is unusually well documented. I verified the parts that matter rather than taking the diff's word for them:

- **Every former writer to the published tree now targets staging.** Reading the whole file rather than the diff: the create-app source copy (`packages/cli/build.mjs:54`), the upstream instruction boundary (`:59`), the discovered standalone guides (`:74`), and every generated fact-sheet, sidecar, framework catalog and reference projection routing through `guidesDestDir` (`:147`–`:192`). Nothing writes into `agenticDist` before the rename at `:203`–`:205`, which is the property the whole change rests on.
- **The staged swap is not redundant.** `buildPackage` is called without `clearDist`, and `scripts/build-package.mjs:84` only clears `outdir` when that flag is set — so `dist` genuinely survives between builds and the swap is doing real work.
- **Dropping the purge is safe, not a regression.** `packages/create-app/agentic/guides/` holds only the ten conceptual guides; no `core.<module>.md` stub, and no `<pkg>.md` colliding with the seven packages that carry an `agentic/standalone-guide.md` (`cache`, `core`, `events`, `queue`, `search`, `shared`, `ui`). The removed `rmSync` calls were therefore no-ops against a freshly staged tree, and `expectedGuideOutputNames()` compared with `toEqual` (`TC-INT-008.spec.ts:349`–`:350`) keeps that honest, since an extra emitted guide fails the assertion. `readdirSync` is still needed for the package/module walk, so the purge removal leaves no dead import behind.
- **The single-builder assumption is genuinely vacuous today.** `packages/cli/watch.mjs` delegates to `scripts/watch.mjs`, which runs esbuild only and never invokes `afterBuild` — so watch mode never rebuilds the agentic tree and cannot race an explicit build.
- **The lesson on stale `dist/` artifacts is satisfied more strongly than before.** `.ai/lessons/package-builds-that-publish-dist-must-clear-stale.md` requires that a build publishing from `dist/` not leave removed files behind; a tree assembled from scratch and swapped in makes staleness structurally impossible instead of purging an enumerated list of known-stale shapes.
- **One improvement over the reference.** This version derives its paths from `outdir` (`:47`–`:49`) where create-app hardcodes `join('dist', …)`, so it stays correct if the draft build-partition spec (`.ai/specs/2026-03-25-safe-build-dev-coexistence.md`, which lists this exact file) ever lands with `dist-check/` and `dist-ephemeral/` partitions.

The one thing I am asking for is about where the new guard test lives, not about the build change itself.

## Verdict

❌ **request changes** — one minor finding. The build change is correct and I found no blocker or major; the requested change is that the guard test shipped with this PR be placed where CI actually runs it, because in its current location CI structurally cannot run it for a `packages/cli`-only diff — including this one. Under this repository's review rules a minor finding requests changes, so this is a narrow, mechanical follow-up rather than a rejection of the approach.

## 🧪 Validation Gate

**Validation source: GitHub checks** (CI already ran the gate on head `f7b0209f4`; no local gate commands were run). Branch protection on `develop` reports no required status checks (`404 Required status checks not enabled`), so every reported check is treated as required. Nothing is failing and nothing is pending.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` — builds packages, runs `yarn generate`, then rebuilds with the generated files. |
| `yarn generate` | ✅ PASS | `prepare`. |
| `yarn build:packages` (rebuild) | ✅ PASS | `prepare`, "Rebuild packages with generated files". |
| `yarn i18n:check-sync` | ✅ PASS | `test` job. |
| `yarn i18n:check-usage` | ✅ PASS | `test` job (`continue-on-error`, advisory as configured). |
| `yarn typecheck` | ✅ PASS | `test` job, scoped on PRs to `turbo run typecheck --filter=[origin/develop]...`, which includes `@open-mercato/cli`. |
| `yarn test` | ✅ PASS | `test` job, same affected-package filter; this is the run the PR body reports as 1703 cli tests. |
| `yarn build:app` | ⏭️ NOT RUN | `prepare`'s "Build app" step is guarded by `if: steps.integration-scope.outputs.skip != 'true'` (`.github/workflows/ci.yml:192`), and this diff sets `skip=true`. No local substitute was run: a build-script change that touches no app code, module file or locale string cannot plausibly break the Next.js build. |
| Playwright integration suite | ⏭️ NOT RUN | `ephemeral-integration` was skipped for the same reason — this is the finding below. |

Also green and unmentioned by the gate: `lint`, `ds-lint`, `audit-scope`, `scope`, and all four CodeQL analyses (`javascript-typescript`, `python`, `actions`, and the aggregate `CodeQL`). `docker-build`, `merge-coverage`, `mutate`, `audit`, `comment` and `[code]smith` were skipped by their own path and event filters.

## Findings

### 🔹 Minor

**`packages/cli/src/lib/__integration__/TC-INT-008.spec.ts:300` — the new staging-artifact guard lives in the one suite CI skips for changes to the file it guards, so it never ran on this PR and will not run on a future regression either.**

`.github/workflows/ci.yml` computes the integration scope from the changed paths. A diff that matches neither `FULL_SUITE_PATTERN` nor the `(?:packages/[^/]+|apps/[^/]+)/src/modules/\K[^/]+` module extraction yields an empty `MODULES`, which sets `skip=true` and skips `ephemeral-integration` entirely. This PR's two paths — `packages/cli/build.mjs` and `packages/cli/src/lib/__integration__/TC-INT-008.spec.ts` — match neither, which is exactly why `ephemeral-integration` shows as `SKIPPED` on `f7b0209f4`.

The concrete consequence is twofold. The guard this PR ships as its automated evidence was not executed by CI on this PR — the label rationale's `skip-qa` justification ("a guard test for the changed build behavior ships in this PR") is literally true but has no CI signal behind it. More importantly, the same filter will skip it on any *future* PR whose diff is confined to `packages/cli/**`. If someone later reintroduces a write into `dist/agentic` before the rename, or drops the final `renameSync`, the guard that exists to catch it will not run on the change that breaks it. It would still fire eventually, on the next unrelated PR that happens to touch a `src/modules/` path — so the signal is delayed and misattributed rather than absent, which is precisely the failure mode a build guard should not have.

The reference implementation this PR mirrors already solves this. `packages/create-app`'s identical guard (`packages/create-app/src/lib/module-facts-build.test.ts:193`–`:201`) is a `node:test` case whose package `test` script runs `node build.mjs` first, and CI invokes `yarn workspace create-mercato-app test` in a step explicitly labelled "always, unfiltered". The cli side took the Playwright path instead and lost that property.

Suggested fix — assert the same invariant from a cli jest test, e.g. a new `packages/cli/src/lib/__tests__/agentic-build-artifacts.test.ts` alongside the existing `agentic-init.test.ts`:

- `packages/cli/jest.config.cjs` matches `<rootDir>/src/**/__tests__/**/*.test.(ts|tsx)`, so the file is picked up by `yarn workspace @open-mercato/cli test`, which is what the green `test` check ran here via `turbo run test --filter=[origin/develop]...` — the cli is in that filter precisely because this PR changes it.
- The `test` job downloads the `build-artifacts` artifact (`packages/*/dist/`) produced by `prepare`'s `yarn build:packages`, so `packages/cli/dist/agentic` is present in that job and the assertion has real post-build state to read — the same artifact that would carry a leftover `agentic.staging` if the swap ever regressed.
- For the local case where nobody built first, either mirror create-app's `requirePackageBuild` (`packages/create-app/src/lib/package-build-artifacts.ts`) and fail loudly with build instructions, or skip when `dist` is absent. Your call; note that the cli's `test` script is plain `jest` and does not build first, so a hard requirement would surface for developers running the cli suite on a clean checkout.

Keeping the TC-INT-008 guard as well is fine — it costs nothing and covers the integration-runner path. The ask is that the invariant also be asserted somewhere CI reaches on a cli-only diff.

### 💅 Nit

- **`packages/cli/build.mjs:206`** — `console.log('Published create-app/agentic/ → dist/agentic/')` runs unconditionally, while the matching "Copied" line at `:55` is guarded by `existsSync(agenticSrc)`. In the (monorepo-impossible) case where the source is absent, the build claims to have published create-app's agentic tree when the swapped-in tree holds only generated artifacts. `Published dist/agentic.staging/ → dist/agentic/` would describe what actually happened. Author's call.
- **`packages/cli/build.mjs:52`–`:56`** — when `packages/create-app/agentic` is missing, behavior changes subtly: the old code left the previously published tree in place and wrote generated artifacts on top of it, whereas the staged build now publishes a tree containing only the generated artifacts. This cannot occur in the monorepo layout the script lives in, the old mixed-tree behavior was accidental rather than intentional, and `packages/create-app/build.mjs` has the identical shape — so mirroring the reference is the right call. Flagged only so the parity is a decision on the record.
- **`packages/cli/build.mjs:203`–`:205`** — a build that dies between the two renames leaves no `dist/agentic` at all, with the old tree parked in `dist/agentic.previous`, and the next build's leftover sweep at `:50` discards it before republishing. Two consecutive failures would therefore leave no published tree. That is a strictly narrower and louder failure window than the multi-second partial tree the in-place refresh produced, and again identical to create-app's accepted design, so it needs no code change.

## 💥 Breaking Changes

- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — the only user-visible change is build-log wording, which nothing parses
- [x] Where a contract had to change: not applicable

The published `dist/agentic` layout is unaffected. Consumers of `@open-mercato/cli/dist/agentic` see the same tree; only the order of operations that produces it changed, and the two sibling directories the build now uses exist solely between the start of `afterBuild` and the final rename. No surface listed in `BACKWARD_COMPATIBILITY.md` is touched.

## 🧪 Test Coverage

What is covered: `TC-INT-008.spec.ts:300`–`:306` asserts that neither `dist/agentic.staging` nor `dist/agentic.previous` survives a completed build, and `ensureCliBuilt()` in `beforeAll` runs the real `packages/cli/build.mjs` unconditionally — so the assertion reads genuine post-build state rather than a fixture, and it is not vacuous. The same spec's `expectedGuideOutputNames()` comparison (`:349`–`:350`) is a strict `toEqual`, which independently covers the purge removal: an emitted `core.<module>.md` stub or an unreachable `<pkg>.md` would fail it. Beyond this PR, the identical staged-swap pattern already ships in `packages/create-app/build.mjs` with a CI-exercised guard, which materially derisks the design itself.

The gap is placement, not existence — see the minor finding. I did not run the Playwright spec or the CLI build locally: this review worktree has no installed dependencies, the unit, typecheck, lint and CodeQL gates are green on this head, the author reports `npx playwright test …TC-INT-008.spec.ts` passing with 4 tests plus cold and incremental full-build verification, and the mirrored pattern is CI-covered on the create-app side by an always-unfiltered step. What I verified instead is *why* CI skipped that suite, which is the finding above.

The PR body is candid that the guard is a safety net rather than a test that fails without the fix, and its reasoning is accepted: the pre-fix build never created a staging directory, and the defect itself is a timing window, so the alternative would be a sampling harness that trades a real regression signal for a test that flakes on loaded CI runners. The 5 ms sampling evidence in the PR body (1634/2234 samples missing before, 0/1880 after) belongs in the description rather than in the suite, and #5064 set that precedent.

pkarw:

Updated in place after the re-review of head `6d7bbf6fe` (this comment supersedes the rationale written on 2026-08-17, when the verdict at that head was still recorded as `review`). The verdict is now **approve**, submitted through the review API, and one label changed as a result.

- 🚦 `merge-queue` — **changed** from `review`. The re-review approved head `6d7bbf6fe`: the single minor finding from my `f7b0209f4` review is fixed and verified to actually execute in CI, the nit is fixed, and every reported check is green. With an approving review on the record the PR is no longer waiting on review, so `review` would be a stale state. Pipeline labels are mutually exclusive, so `review` was removed in the same transition.
- ♻️ `refactor` — unchanged. The published `dist/agentic` layout is identical; this reorders how the build produces its output rather than changing what the package ships. The two files added since the last rationale are a test and a test assertion, which does not change the category.
- ⏭️ `skip-qa` — unchanged, and now on firmer ground than when it was first applied. The automated-verification exemption holds: no UI-rendering file is touched (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`), the database structure and API surface are unchanged, no `BACKWARD_COMPATIBILITY.md` contract is broken, and automated tests for the changed behavior ship in the same PR. That last clause is what my previous review disputed — the guard existed but CI structurally could not run it for a `packages/cli`-only diff. `packages/cli/src/lib/__tests__/agentic-build-artifacts.test.ts` fixes that, and run `32115949151` shows it reporting `PASS` with the cli suite total moving 1703 → 1705, so the exemption is now backed by a check that genuinely executed. No `needs-qa`, so the QA-approval gate does not apply.
- 🔽 `priority-low` — unchanged. This fixes a real but narrow build-hygiene defect (#5104) with no user-facing symptom and no production impact; nothing is waiting on it.
- 🟢 `risk-low` — unchanged. The blast radius is one package's build script plus two test files. The change is a mirror of an already-merged, already-CI-exercised pattern in `packages/create-app/build.mjs` (#5064), every writer to the published tree was re-verified to target staging, and the published artifact layout is byte-identical in shape.
- 🔒 `in-progress` — this review run's lock, removed in the release comment below. It is a lock, not a workflow state, so it never competed with the pipeline label.

No `ci-monitoring`: every check on `6d7bbf6fe` had already settled green before this review started, so this run owes no CI follow-up comment.

pkarw:

For orientation: the build change itself needs nothing. The single ask is that the staging-artifact invariant also be asserted from a test CI reaches on a `packages/cli`-only diff — a jest case under `packages/cli/src/lib/__tests__/` runs inside the `test` check's `turbo run test --filter=[origin/develop]...`, and that job already downloads the `packages/*/dist/` build artifact, so the assertion has real post-build state to read. Keeping the TC-INT-008 guard alongside it is fine.

pkarw:

One minor finding drives the verdict — the staging-artifact guard is placed in the Playwright integration suite, which CI's path filter skips for a `packages/cli`-only diff, so it never ran on this PR and would not run on a future regression to `packages/cli/build.mjs`. No blocker and no major: the staged swap itself is correct, every former writer to the published tree was verified to target staging, the purge removal was verified safe against the actual `packages/create-app/agentic/guides/` contents, and `scripts/build-package.mjs` confirms `dist` is not cleared per build so the swap is meaningful.

Validation source: GitHub checks — every reported check on `f7b0209f4` is green (`test`, `lint`, `ds-lint`, `audit-scope`, `prepare`, `scope`, all four CodeQL analyses), nothing failing, nothing pending, so no local gate commands were run and no CI follow-up is owed. `ephemeral-integration` and `yarn build:app` were skipped by the repository's own integration-scope filter, not by this run.

autofix: skipped (not my PR — re-run with --autofix to fix it here). The PR is assigned back to @adeptofvoltron.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-08-18T08:13:05Z.

The previous `om-auto-review-pr` pass by @pkarw explicitly released its lock on 2026-08-17 and handed the PR back to me as the author, the `in-progress` label is absent, and the only remaining signal is my own assignee entry — so this is a clean claim, not a takeover. This run holds the outer lock for the whole drive-to-merge-ready chain; sub-skills will re-enter under the same owner.

adeptofvoltron:

- 🔍 `review` — @pkarw's `changes-requested` verdict rested on exactly one actionable finding, and that finding is now fixed on head `6d7bbf6fe`, so the PR is no longer waiting on the author. It is not `merge-queue` either: no approving review exists yet, and GitHub still records the earlier CHANGES_REQUESTED review against the PR. `review` is the honest state — the work is done and a fresh review of the new head is what the PR is waiting for.
- ♻️ `refactor` — unchanged from the previous pass. The published `dist/agentic` layout is byte-for-byte what the previous build produced; only the order of operations that puts it there changed, and this run added a test rather than any behavior.
- ⏭️ `skip-qa` — the automated-verification exemption still applies and is now stronger than it was. The diff touches no UI-rendering file, no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`; the database structure and API surface are unchanged; no `BACKWARD_COMPATIBILITY.md` contract is touched. The reason the exemption was previously thin — the guard test shipped in a suite CI skips for this diff — is precisely what this run fixed, so the automated coverage the exemption depends on now actually executes in CI.
- 🟢 `priority-low` — unchanged. The latent race this PR closes has no reader in the repository today, so the urgency is "fix when convenient" rather than release-blocking.
- 🟢 `risk-low` — unchanged. The blast radius is one package build script plus test files; nothing at runtime, in the database, or on any API surface is affected.
- 📡 `ci-monitoring` (applied when this run releases its lock) — a meta label, not a claim. It records that this run's work is finished and fully reported while CI is still settling on `6d7bbf6fe`, and that a CI-result follow-up comment is still owed. Anyone may act on the PR while it is set; it is removed as soon as that follow-up lands.

Removed: ❌ `changes-requested` — the single finding behind it is fixed, so leaving it on would misdescribe the PR as still owing the author work.
Not applied: `qa-approved` / `qa-self-verified` — this skill never grants QA sign-off, and `skip-qa` makes the QA gate inapplicable here anyway.

adeptofvoltron:

This run took the PR at `f7b0209f4`, brought it up to date with `develop`, fixed the review feedback that was outstanding on it, and left it merge-ready except for the one thing this automation structurally cannot supply: an approving review. Head is now `6d7bbf6fe`.

### 🔀 Base merge

`origin/develop` at `f0cf23f0f` was merged into the PR head **before** any review or CI work, so everything below judges the real merge result rather than a diff against a stale base. The merge was clean — no conflicts, no files needing hand-resolution. `develop` did not advance again during the run, so no re-merge was needed.

### 🔍 Review feedback — how each ask was handled

@pkarw's review at `f7b0209f4` requested changes on **one** actionable finding and recorded three nits. Every one of them is accounted for:

**Fixed — the minor finding (`TC-INT-008.spec.ts:300`).** The staging-artifact guard shipped in this PR lived only in the Playwright integration suite, which CI's integration scope skips for any diff confined to `packages/cli/**` — so it never ran on this PR and would not have run on a future regression to the very file it guards. New commit `41625c07a` adds `packages/cli/src/lib/__tests__/agentic-build-artifacts.test.ts`, asserting the same invariant from the cli jest suite, which CI's affected-package filter (`turbo run test --filter=[origin/develop]...`) always selects for a `packages/cli` change. The `test` job downloads the `build-artifacts` produced by `prepare`'s `yarn build:packages`, so `packages/cli/dist` is present there and the assertion reads the tree the package would actually publish. The TC-INT-008 guard is kept as well, exactly as the review suggested — it costs nothing and covers the integration-runner path.

On the choice the review left open (fail loudly like create-app's `requirePackageBuild`, or skip when `dist` is absent): this takes the **skip** branch, and deliberately. The create-app guard can afford to fail hard because its package `test` script builds first, so a missing `dist` there means genuine misuse. The cli's `test` script is plain `jest` and never builds, so a hard requirement would break `yarn workspace @open-mercato/cli test` on any clean local checkout. The skip is narrow — it is gated on `packages/cli/dist` being absent entirely — so it cannot mask a regression in CI, where the artifact download guarantees `dist` exists.

The guard was verified in all three states rather than assumed:
- **unbuilt package** — 2 tests skipped, suite green, no false failure on a clean checkout;
- **build aborted midway, leaving `dist/agentic.staging` behind** — the guard **fails**, which is the point: it is not vacuous and catches exactly the regression it exists for;
- **completed build** — 2 tests pass against the real published tree.

**Fixed — nit 1 (`build.mjs:206`).** `console.log('Published create-app/agentic/ → dist/agentic/')` ran unconditionally while the matching "Copied" line is guarded by `existsSync(agenticSrc)`, so a build without the create-app source would have claimed to publish a tree it never copied. Commit `6d7bbf6fe` changes it to `Published dist/agentic.staging/ → dist/agentic/`, which describes what the rename actually does. Nothing in the repository parses this output (verified by search), so this is log wording only.

**No change needed — nits 2 and 3.** Both were explicitly filed as decisions on the record rather than requests: the missing-source behavior change (`build.mjs:52`–`:56`) is deliberate parity with `packages/create-app/build.mjs` and cannot occur in the monorepo layout, and the two-consecutive-failures window between the renames (`:203`–`:205`) is a strictly narrower and louder failure mode than the multi-second partial tree it replaces, again identical to the accepted create-app design. The review itself states neither needs a code change, so neither was changed and neither was filed as a follow-up — filing an issue to re-litigate an accepted design decision would be noise, not tracking.

### 🧪 Validation

Run locally in an isolated worktree. This worktree has no `node_modules` of its own, so jest was invoked directly rather than through yarn.

| Command | Status | Notes |
|---|---|---|
| `@open-mercato/cli` jest suite (full) | ✅ PASS | 1705 passed / 89 suites — was 1703 / 88 before, the delta being the two new guard assertions. |
| New guard, unbuilt / aborted / completed build | ✅ PASS | Skips, fails, passes respectively — see above. |
| `scripts/__tests__/repo-wide-guards.test.mjs` | ✅ PASS | 16/16, including "no test that audits other packages is left unclassified". The new test reads only `packages/cli/dist`, inside its own workspace, so it correctly needs no `REPO_WIDE_GUARDS` entry — checked because #5340 has just made this registry live on `develop`. |
| `packages/cli` build (`node build.mjs`) | ✅ PASS | Completed and published a full tree with the corrected log line, leaving no `agentic.staging` / `agentic.previous` behind. |
| `yarn lint`, full `yarn typecheck` | ⏭️ NOT RUN LOCALLY | Not a skip by choice: this machine's install is stale and missing `eslint-config-next`, `bullmq-otel` and `@open-mercato/telemetry`, so eslint cannot load the flat config and `tsc` reports three unrelated module-resolution errors in files this PR does not touch. Both are covered by CI's `lint` and `test` jobs, which run on the new head. |
| `yarn generate`, `yarn i18n:check-*`, `yarn build:app` | ⏭️ NOT RUN | Unchanged from the original PR's reasoning: a build-script change plus a jest test touches no module file, no locale string and no app code. |

**CI is still running on this head.** Branch protection plus the review requirement hold the actual merge; this run's verdict covers the work, not a green run. Checks pending at the time of writing: `prepare`, `lint`, `ds-lint`, `scope`, and the four CodeQL analyses. A follow-up comment will report the CI outcome.

### 📸 UI verification

Not applicable. The diff consists of one package build script and two test files — there is no user-facing surface to drive, which is also why `skip-qa` applies rather than `needs-qa`.

### 🔁 Follow-ups filed

None, and deliberately so. The one actionable finding was a blocker for the verdict and was fixed in-loop rather than deferred; the actionable nit was a one-line log correction, cheaper to fix than to track; and the remaining two nits are recorded design decisions the review explicitly said need no change. No CI flakes were observed, so there is nothing to file on that side either.

### 🚀 Hand-off

`om-auto-fix-pr` never merges — the PR is handed to `om-approve-merge-pr` / `om-merge-buddy`, which re-check every gate before squash-merging.

What a human still owes this PR: **an approving review.** @pkarw's CHANGES_REQUESTED review is still recorded against the PR and no approval exists, so `mergeStateStatus` stays `BLOCKED` regardless of CI. The PR was authored by this automation account, and GitHub rejects self-approval, so this run cannot clear that itself — the review has been re-requested from @pkarw as their handback comment asked. Once that approval lands and the required checks are green on `6d7bbf6fe`, `om-approve-merge-pr 5328` can take it; `skip-qa` means the QA-approval gate does not apply.

adeptofvoltron:
🤖 `om-auto-fix-pr` completed: the single actionable finding from @pkarw's review is fixed and pushed as `6d7bbf6fe`, the PR is up to date with `develop`, and the run is fully reported in the summary comment above. Lock released — `in-progress` swapped for `ci-monitoring`, which is a meta label and **not** a claim: the PR is free for anyone to act on, and the label only records that a CI-result follow-up comment on head `6d7bbf6fe` is still owed by this run. It will be removed as soon as that comment lands, or if the wait budget (40 minutes) expires first.

Not merge-ready yet, for one reason this automation cannot clear itself: the PR needs an approving review. The CHANGES_REQUESTED review is still recorded and the PR was authored by this account, so self-approval is rejected by GitHub. Review has been re-requested from @pkarw as their handback asked.

adeptofvoltron:

✅ Every check on head `6d7bbf6fe` settled green, inside the wait budget. Branch protection on `develop` reports no required status checks, so following the tracker convention every reported check is treated as required: `prepare` (1m10s), `test` (4m25s), `lint` (33s), `ds-lint` (43s), `scope` (31s), `audit-scope` (14s), all four CodeQL analyses (`javascript-typescript`, `python`, `actions`, and the aggregate `CodeQL` gate), and `license/cla`. Nothing failed and nothing was rerun, so there are no flakes to file. `audit`, `comment`, `docker-build`, `ephemeral-integration`, `merge-coverage`, `mutate` and `[code]smith` were skipped by their own path and event filters.

**The fix is confirmed by this run, not just by argument.** The `test` job log shows `PASS src/lib/__tests__/agentic-build-artifacts.test.ts` inside `@open-mercato/cli:test` (1705 passed), while `ephemeral-integration` shows `SKIPPED` on the same head — which is precisely the gap @pkarw's finding described. The new guard executed on a `packages/cli`-only diff, in the same run where the Playwright guard could not, so the invariant now has a CI signal on exactly the kind of change that can break it.

It also **passed** rather than skipped, which independently confirms the design assumption behind it: `packages/cli/dist` really is present in the `test` job via the `build-artifacts` download from `prepare`, so the assertion read the post-build tree the package would publish rather than skipping over an absent `dist`. The local skip branch therefore cannot mask a regression in CI.

This does not change the verdict, and it is not permission to merge: the PR still needs an approving review. @pkarw's CHANGES_REQUESTED review remains recorded against it, review has been re-requested, and this account cannot self-approve its own PR. With that approval the PR is mergeable via `om-approve-merge-pr 5328` — `skip-qa` means the QA-approval gate does not apply.

`ci-monitoring` is removed: this run owed one CI follow-up comment, this is it, and no further follow-up will come from this agent.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-08-18T09:14:42Z. Other auto-skills will skip this PR until the lock is released.

This is a **re-review**: @pkarw's `changes-requested` review at `f7b0209f4` is still the standing verdict, but the head has since moved to `6d7bbf6fe` (a base merge plus two follow-up commits). The claim is free to take — the previous `om-auto-fix-pr` run explicitly released its lock on 2026-08-18, the `in-progress` label is absent, and the only remaining signal is my own assignee entry as the PR author.

adeptofvoltron:
# 🔍 Code Review: Re-review — refactor(cli): publish dist/agentic through a staged swap (#5104)

> **Re-review.** @pkarw reviewed this PR at head `f7b0209f4` on 2026-08-17 and requested changes on one minor finding. The head has since moved to `6d7bbf6fe` through three commits — a merge of `origin/develop`, the fix for that finding, and the fix for the actionable nit. This pass re-checks every previous finding against the current head before issuing a fresh verdict, and reviews the newly added code on its own merits.
>
> **GitHub rejects self-approval**, and this PR is authored by the same account this review runs under, so the verdict below is posted as a comment rather than submitted through the review API. @pkarw's `CHANGES_REQUESTED` review remains the only entry in GitHub's review state, and only a human approval can clear it — see the Verdict section for what that means for merging.

## 🎯 Summary

This PR converts `packages/cli/build.mjs` from refreshing its published `dist/agentic` tree in place to assembling the tree in `dist/agentic.staging` and swapping it in with two `renameSync` calls, dropping the stale-artifact purge the in-place refresh needed. It mirrors `packages/create-app/build.mjs` (#5064) and implements what issue #5104 prescribed. The reviewed scope is the full diff of `packages/cli/build.mjs`, `packages/cli/src/lib/__integration__/TC-INT-008.spec.ts` and the new `packages/cli/src/lib/__tests__/agentic-build-artifacts.test.ts` at head `6d7bbf6fe`, read against the create-app reference implementation, the shared builder in `scripts/build-package.mjs`, `.github/workflows/ci.yml`, and the actual CI logs for this head.

The change is correct, and the one previously outstanding ask is not merely claimed to be fixed — it is fixed with CI evidence behind it. Rather than take the previous review's conclusions on trust, I re-derived the property the whole design rests on from the current file: reading `packages/cli/build.mjs` end to end, every writer into the agentic tree targets the staging directory — the create-app source copy (`:54`), the upstream instruction boundary and its manifest (`:59`–`:69`), the discovered standalone guides (`:74`–`:89`), and every generated fact-sheet, sidecar, framework catalog and reference projection routing through `guidesDestDir` (`:147`–`:192`). The identifier `agenticDist` occurs in exactly three places in the file — its declaration at `:47` and the two renames at `:203`–`:204` — so nothing writes into the published path before the swap. `outdir` is otherwise used only for `bin.js` and for dynamically importing the built `lib/` modules, neither of which lives under `agentic/`.

The two new commits both do what they say. `41625c07a` adds a cli jest guard asserting the same staging-artifact invariant the Playwright spec asserts, and `6d7bbf6fe` corrects the final publish log line. The base merge `a1708e8dc` was clean and left `packages/cli/build.mjs` untouched apart from that log line: `git diff f7b0209f4 6d7bbf6fe -- packages/cli/build.mjs` is a one-line change, so nothing @pkarw verified about the build script has been invalidated by the merge, and `scripts/build-package.mjs` — the shared builder the design depends on — did not change either.

## Verdict

✅ **approve** — no blockers, no majors, and no minors remain. The single minor finding from the previous review is verifiably resolved, the actionable nit is fixed, and the one issue this pass found (a stale PR description) was itself fixed during this run. One optional hardening nit is recorded below and is explicitly the author's call, as the previous review already delegated that same decision.

Two caveats a maintainer needs before merging, neither of which changes the verdict:

1. **This approval cannot clear the merge gate.** @pkarw's `CHANGES_REQUESTED` review is still recorded against the PR, `reviewDecision` is `CHANGES_REQUESTED`, and `mergeStateStatus` is `BLOCKED`. GitHub will not accept a review from this account on its own PR, so a human approving review is still owed. This is why the pipeline label stays `review` rather than moving to `merge-queue` — see the label rationale comment.
2. **The verdict covers the code and the settled CI on `6d7bbf6fe`,** which is fully green; it is not permission to bypass branch protection.

## 🧪 Validation Gate

**Validation source: GitHub checks** on head `6d7bbf6fe` (run [32115949151](https://github.com/open-mercato/open-mercato/actions/runs/32115949151)). No local gate commands were run, per this repository's GitHub-checks-first rule — CI has already executed the gate on this exact head, every check has settled, and none is failing or pending. Branch protection on `develop` reports no required status checks (`404 Required status checks not enabled`), so following the tracker convention every reported check is treated as required.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` → "Build packages"; 23 workspace builds in scope. |
| `yarn generate` | ✅ PASS | `prepare` → "Prepare generated modules". |
| `yarn build:packages` (rebuild) | ✅ PASS | `prepare` → "Rebuild packages with generated files". |
| `yarn i18n:check-sync` | ✅ PASS | `test` → "Check i18n sync". |
| `yarn i18n:check-usage` | ✅ PASS | `test` → "Check i18n usage" (`continue-on-error`, advisory as configured). |
| `yarn typecheck` | ✅ PASS | `test` → "Checking types", scoped on PRs to `turbo run typecheck --filter=[origin/develop]...`, which selects `@open-mercato/cli` because this PR changes it. |
| `yarn test` | ✅ PASS | `test` → "Test", same affected-package filter. The cli suite reports **`Test Suites: 89 passed, 89 total` / `Tests: 1705 passed, 1705 total`** — up from 88/1703 before, and with **zero skipped**, which is the direct evidence that the new guard executed rather than self-disabling. |
| `yarn build:app` | ⏭️ NOT RUN | `prepare`'s "Build app" step is guarded by `if: steps.integration-scope.outputs.skip != 'true'` (`.github/workflows/ci.yml:186`), and this diff sets `skip=true`. Not substituted locally: a build-script and test-only change that touches no app code, module file or locale string cannot plausibly break the Next.js build. |

Also green and outside the configured gate: `lint`, `ds-lint`, `audit-scope`, `scope`, `license/cla`, and all four CodeQL analyses (`javascript-typescript`, `python`, `actions`, plus the aggregate `CodeQL`). The `test` job's three always-unfiltered steps also passed — `yarn ds:tokens:check`, `yarn workspace create-mercato-app test`, and `yarn test:repo-wide-guards`. That last one matters here: `scripts/repo-wide-guards.mjs` requires every test that reads outside its own package to be classified, and it ran unfiltered and green on this head, confirming the new guard needs no `REPO_WIDE_GUARDS` entry because it reads only `packages/cli/dist`, inside its own workspace. `docker-build`, `merge-coverage`, `mutate`, `audit`, `comment` and `[code]smith` were skipped by their own path and event filters, and `ephemeral-integration` was skipped for the same integration-scope reason as before — which is precisely why the jest guard needed to exist.

## Findings

### 🔹 Minor

**Fixed during this run — the PR description had gone stale against its own head.** `packages/cli/src/lib/__tests__/agentic-build-artifacts.test.ts` is the file the previous review asked for and the reason this PR moved at all, yet the "What Changed" section did not mention it; the corrected publish log line was likewise absent; and the "🧪 Tests" section still claimed `1703 passed / 88 suites`, a figure the follow-up commit made false. This repository squash-merges, so the PR body becomes the permanent commit message on `develop` — a body that omits two of the four commits and reports a stale test count is an inaccurate record, not a cosmetic issue.

Because this PR is authored by the account this run executes under, the autofix path applies and the fix was applied here rather than handed back: the description now documents the jest guard and the log-line correction under "What Changed", reports `1705 passed / 89 suites`, and cites the CI evidence that the guard executes rather than skips. No code changed and nothing was pushed to the branch — this was a tracker-side edit only. It is reported rather than fixed silently so the record shows it was found.

### 💅 Nit

**`packages/cli/src/lib/__tests__/agentic-build-artifacts.test.ts:19` — the guard disables itself when `packages/cli/dist` is absent, and nothing distinguishes "correctly skipped locally" from "silently skipped in CI".** `const describeBuiltPackage = fs.existsSync(distDir) ? describe : describe.skip` is the right call for local ergonomics and I am not asking for it to change: the cli's `test` script is plain `jest` and never builds first, so failing hard would break `yarn workspace @open-mercato/cli test` on a clean checkout, and the previous review explicitly left this choice to the author ("either mirror create-app's `requirePackageBuild` … or skip when `dist` is absent. Your call"). Re-litigating a delegated decision would be noise.

The residual observation is narrower. The whole point of `41625c07a` was to move the invariant somewhere CI structurally reaches; a self-disabling `describe` reintroduces a quieter version of the same failure mode one layer down. The CI guarantee is strong but not airtight: `prepare` uploads `packages/*/dist/` with `if-no-files-found: error`, and the `test` job's unconditional "Download build artifacts" step would fail outright if the artifact were missing — but the upload path is a glob, so an artifact that contained *some* package's `dist` while `packages/cli/dist` was missing would satisfy both guards and leave this suite skipping with no signal. Today that is hypothetical: CI on this head reports 1705 passed and 0 skipped, so the guard demonstrably ran.

If you want to close it anyway, the cheap form is to make the skip local-only rather than universal — for example, treat a missing `dist` as a hard failure when `process.env.CI` is set and skip only otherwise. That keeps the clean-checkout ergonomics you deliberately preserved while making the CI path incapable of silently skipping.

A second, smaller point in the same file: the assertion is a bare `expect(fs.existsSync(...)).toBe(false)`, so a failure reports `Expected: false, Received: true` and leans entirely on the test name to convey meaning. The create-app reference it mirrors (`packages/create-app/src/lib/module-facts-build.test.ts:193`–`:201`) attaches an explicit message — `` dist/${leftover} must not survive the build ``. Adding the equivalent would make a future failure self-explanatory to whoever hits it. Author's call.

### Previously-raised findings — re-checked against `6d7bbf6fe`

The previous review's items are re-verified rather than assumed carried over. All four are accounted for.

- **✅ Resolved — minor, `TC-INT-008.spec.ts:300` (the guard lived where CI could not run it).** The fix is real and independently confirmed. `packages/cli/jest.config.cjs` matches `<rootDir>/src/**/__tests__/**/*.test.(ts|tsx)`, so `src/lib/__tests__/agentic-build-artifacts.test.ts` is collected by the cli suite; the `test` job's "Download build artifacts" step restores `packages/*/dist/` before the suite runs, so `packages/cli/dist` is present; and `turbo run test --filter=[origin/develop]...` selects `@open-mercato/cli` because this PR changes it. Rather than reason only from the workflow file, I read the CI log for this head: it contains `PASS src/lib/__tests__/agentic-build-artifacts.test.ts`, and the suite totals are `Tests: 1705 passed, 1705 total` with no skipped entries. The guard therefore ran, on the exact diff shape that previously skipped it. The TC-INT-008 guard was kept as well, exactly as the review suggested.
- **✅ Resolved — nit 1, `build.mjs:206` (the publish log claimed to have published a tree it may never have copied).** `6d7bbf6fe` changes the line to `Published dist/agentic.staging/ → dist/agentic/`, which describes the rename that actually executed and is accurate whether or not `packages/create-app/agentic` was present. It is also consistent with the intermediate log lines that already name the staging directory (`:55`, `:96`, `:161`, `:193`). I re-confirmed by search that nothing in the repository parses this output — the only matches are the `console.log` call sites themselves in `packages/cli/build.mjs` and `packages/create-app/build.mjs`.
- **✅ Accepted as recorded — nit 2, `build.mjs:52`–`:56` (missing-source behavior change).** Deliberate parity with `packages/create-app/build.mjs`, impossible in the monorepo layout the script lives in. The previous review filed it as a decision on the record and stated it needs no code change; nothing on the new head alters that.
- **✅ Accepted as recorded — nit 3, `build.mjs:203`–`:205` (two consecutive failures could leave no published tree).** Still true and still the narrower, louder failure mode it was judged to be. Worth noting as an improvement the new guard brings: if a build now dies between the renames, `dist/agentic.previous` survives, and the jest guard fails on it — so that state is detected by the test suite rather than lingering silently.

## 💥 Breaking Changes

- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — the only user-visible change is build-log wording, and nothing parses it
- [x] Where a contract had to change: not applicable

The published `dist/agentic` layout is unaffected: consumers of `@open-mercato/cli/dist/agentic` see the same tree, and only the order of operations that produces it changed. The two sibling directories the build now uses exist solely between the start of `afterBuild` and the final rename, and a completed build leaves neither behind — which is now asserted from two independent suites. No surface listed in `BACKWARD_COMPATIBILITY.md` is touched, and the new test file adds no public surface.

## 🧪 Test Coverage

The invariant that matters — a completed build publishes `dist/agentic` and leaves no `dist/agentic.staging` or `dist/agentic.previous` behind — is now asserted from two places, which is the change this re-review was called to check.

`packages/cli/src/lib/__tests__/agentic-build-artifacts.test.ts` asserts it from the cli jest suite. It is not vacuous: `packageRoot` resolves three levels up from `src/lib/__tests__` to `packages/cli`, so `distDir` is the real published directory, and CI reads it after `prepare`'s `yarn build:packages` produced it and the `test` job downloaded it as `build-artifacts`. The author's report that the guard was exercised in all three states — skipping on an unbuilt package, **failing** on a `dist` left holding `agentic.staging` after an aborted build, and passing against a completed build — is consistent with the code, and the middle state is the one that proves the guard has teeth. The CI evidence quoted above closes the loop the previous review found missing: the guard ran on this PR, on a diff confined to `packages/cli/**`.

`packages/cli/src/lib/__integration__/TC-INT-008.spec.ts:298`–`:304` keeps the same assertion on the integration-runner path, where `ensureCliBuilt()` in `beforeAll` runs the real `packages/cli/build.mjs` unconditionally, so the assertion reads genuine post-build state. That suite is skipped by CI's integration scope for this diff shape, which is exactly why the duplicate exists and why keeping both is correct rather than redundant. The same spec's `expectedGuideOutputNames()` comparison (`:349`–`:350`) is a strict `toEqual` and independently covers the purge removal — an emitted `core.<module>.md` stub or an unreachable `<pkg>.md` guide would fail it.

The remaining gap is the one the PR body is candid about and that I accept unchanged: neither guard fails against the *pre-fix* build, because the pre-fix build never created a staging directory and the defect itself is a timing window. The alternative — a sampling harness asserting the published tree is never observably incomplete — would trade a real regression signal for a test that flakes on loaded CI runners. The 5 ms sampling evidence (1634/2234 probes missing before, 0/1880 after) belongs in the description, where it now is, and #5064 set that precedent for the create-app side of the same change. The guards are a safety net for the new design, correctly labelled as such.

No new test is needed for the log-line change in `6d7bbf6fe`: it alters a `console.log` string that nothing reads or parses, and asserting on build-log wording would couple a test to output the repository deliberately treats as human-facing only.

adeptofvoltron:

This was a **re-review** of head `6d7bbf6fe`, run because @pkarw's `changes-requested` verdict at `f7b0209f4` no longer covered the diff after three commits landed. The full report is in the review comment above.

**Outcome.** Every finding from the previous review is accounted for. The one blocking item — the staging-artifact guard living only in the Playwright suite, which CI's integration scope skips for a `packages/cli`-only diff — is genuinely fixed, and I confirmed it from the CI log rather than from the commit message: the `test` job on this head reports `PASS src/lib/__tests__/agentic-build-artifacts.test.ts` with `Tests: 1705 passed, 1705 total` and **no skipped tests**, so the new jest guard executed on exactly the diff shape that previously skipped it. The actionable nit (the publish log line claiming to have published a tree it may never have copied) is fixed in `6d7bbf6fe`. The two remaining nits were filed by the previous review as accepted design decisions needing no change, and nothing on the new head alters that.

I also re-derived the core invariant independently instead of inheriting it: in the current `packages/cli/build.mjs`, `agenticDist` appears only at its declaration (`:47`) and the two renames (`:203`–`:204`), while every writer — the create-app copy, the upstream boundary, the discovered guides, and all generated fact-sheets and sidecars — targets the staging tree. The base merge was clean and changed nothing in the build script beyond that one log line.

**Autofix.** One new minor finding surfaced in this pass and was fixed in-run rather than handed back, since this PR is authored by the account the review runs under: the PR description had gone stale against its own head — it did not mention the jest guard it was asked to add, omitted the log-line correction, and still reported `1703 passed / 88 suites`. The repository squash-merges, so that body becomes the permanent commit message. The description now documents both commits and reports `1705 passed / 89 suites` with the CI evidence. No code changed and nothing was pushed to the branch; this was a tracker-side edit only.

**Labels.** Unchanged and re-justified in the label rationale comment, which was updated in place. `review` is retained deliberately rather than moving to `merge-queue` on an approve verdict: GitHub rejected the `--approve` submission ("Can not approve your own pull request"), so this report is a comment, @pkarw's `CHANGES_REQUESTED` is still the standing review state, and `mergeStateStatus` is `BLOCKED`. Marking it `merge-queue` would signal ready-to-merge on a PR GitHub will refuse to merge.

**CI.** No follow-up is owed and `ci-monitoring` was not applied: every check on `6d7bbf6fe` had already settled before this review — 11 success, 7 skipped by their own path filters, none failing and none pending.

**What this PR still needs: a human approving review.** That is the only outstanding gate. Once it lands, `om-approve-merge-pr 5328` can take it — `skip-qa` means the QA-approval gate does not apply here.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-18T14:51:51Z. Other auto-skills will skip this PR until the lock is released.

This is a **re-review**: my `changes-requested` verdict at `f7b0209f4` is the standing review, and the head has since moved to `6d7bbf6fe`. The claim is free to take — the previous `om-auto-review-pr` run by @adeptofvoltron explicitly released its lock at 09:24 UTC today, the `in-progress` label is absent, and the only remaining signal is the author's own assignee entry (which my own 2026-08-17 handoff created). Clean claim, not an override.

pkarw (approved):
# 🔍 Code Review: Re-review — refactor(cli): publish dist/agentic through a staged swap (#5104)

> **Re-review.** I reviewed this PR at head `f7b0209f4` on 2026-08-17 and requested changes on a single minor finding: the staging-artifact guard shipped only in the Playwright suite that CI's integration scope skips for a `packages/cli`-only diff, so it never ran on this PR and would not run on a future regression to `packages/cli/build.mjs` either. The head has since moved to `6d7bbf6fe` through three commits — a merge of `origin/develop`, the fix for that finding (`41625c07a`), and the fix for the actionable nit (`6d7bbf6fe`). This pass re-checks the previous finding against the current head, reviews the newly added code, and issues a fresh verdict.

## 🎯 Summary

The reviewed scope is the full diff of `packages/cli/build.mjs`, `packages/cli/src/lib/__integration__/TC-INT-008.spec.ts`, and the new `packages/cli/src/lib/__tests__/agentic-build-artifacts.test.ts` at head `6d7bbf6fe` (82 insertions, 30 deletions across 3 files), read against `packages/create-app/build.mjs`, `packages/cli/jest.config.cjs`, `turbo.json`, and `.github/workflows/ci.yml`.

**The previous finding is fully resolved, and I verified the fix end to end rather than taking the new file's word for it.** The point of that finding was never "add a test somewhere" — it was that the guard had to execute in a job CI actually runs for a `packages/cli`-only diff. Each link in that chain now holds:

- **The file is picked up.** `packages/cli/jest.config.cjs:33` matches `<rootDir>/src/**/__tests__/**/*.test.(ts|tsx)`, and the new guard sits at `src/lib/__tests__/agentic-build-artifacts.test.ts`, alongside the 38 existing suites in that directory.
- **The affected-package filter selects the cli.** `.github/workflows/ci.yml:624` runs `yarn turbo run test --filter=[origin/${{ github.base_ref }}]...` on PRs; a diff that changes `packages/cli/**` puts `@open-mercato/cli` in that selection by definition. The claim in the file's comment that this filter "always runs for such a diff" is accurate.
- **`dist` is genuinely present in that job, so the assertion is not silently skipped.** `prepare` uploads `packages/*/dist/` as the `build-artifacts` artifact (`ci.yml:196`–`:202`), and the `test` job downloads it before the test step (`ci.yml:585`–`:587`). This is the property the whole `describeBuiltPackage` gate rests on.
- **Turbo cannot replay a stale pass over it.** This was my one remaining worry, since `dist/` is gitignored and therefore outside turbo's input hash — a cached result from a run where `dist` was absent would have made the guard vacuous. `turbo.json` declares `test` with `"cache": false`, so the task always executes. Not an issue.
- **It actually ran and actually passed on this head — I read the log, not the check badge.** In run `32115949151`, the `test` job's `@open-mercato/cli:test` group reports `PASS src/lib/__tests__/agentic-build-artifacts.test.ts` with `Test Suites: 89 passed, 89 total` and `Tests: 1705 passed, 1705 total`. The count is the decisive evidence: the previous head reported 1703, and the two `it.each` cases add exactly 2 — had `describeBuiltPackage` fallen back to `describe.skip`, jest would have reported them as skipped and the total would have stayed at 1703. The guard executed against the real post-build tree.
- **The premise in the file's comment holds too.** `packages/cli/package.json` declares no `files` field and the package directory carries no `.npmignore` or local `.gitignore` — the `packages/*/dist/` entries live in the repository-root `.gitignore`, which npm does not consult from outside the package root. So `dist/` does ship in the published tarball, and a leftover `agentic.staging` really would be published. The comment is not hand-waving.

The build change itself is unchanged from what I already verified at `f7b0209f4` and I re-confirmed it against the merged head: every former writer to the published tree still targets `agenticStaging` (`build.mjs:54`, `:59`, `:74`, `:147`–`:192`), nothing writes into `agenticDist` before the rename at `:203`–`:205`, and the purge removal remains safe. The `origin/develop` merge at `a1708e8dc` brought in the WMS starter preset and touched none of these three files; the diff against `origin/develop` is still exactly the intended three.

The nit is also fixed: `build.mjs:206` now logs `Published dist/agentic.staging/ → dist/agentic/`, which describes what the step did rather than claiming create-app's tree was published even when `agenticSrc` was absent.

## Verdict

✅ **approve** — no blockers, no majors, no minors. The one finding from the previous review is resolved with direct CI evidence, the nit is fixed, and the newly added test introduces no issue of its own. Nothing here needs another pass.

## 🧪 Validation Gate

**Validation source: GitHub checks** (CI ran the gate on head `6d7bbf6fe`; no local gate commands were run, per this repository's GitHub-checks-first rule). Branch protection on `develop` reports no required status checks, so following the tracker convention every reported check is treated as required. Nothing is failing and nothing is pending.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` (1m10s) — builds all packages unfiltered. |
| `yarn generate` | ✅ PASS | `prepare`, "Prepare generated modules". |
| `yarn build:packages` (rebuild) | ✅ PASS | `prepare`, "Rebuild packages with generated files". |
| `yarn i18n:check-sync` | ✅ PASS | `test` job. |
| `yarn i18n:check-usage` | ✅ PASS | `test` job (`continue-on-error`, advisory as configured). |
| `yarn typecheck` | ✅ PASS | `test` job, scoped on PRs to `turbo run typecheck --filter=[origin/develop]...`, which includes `@open-mercato/cli`. |
| `yarn test` | ✅ PASS | `test` job (4m25s), same affected-package filter — 89 suites / 1705 tests for the cli, including the new guard. |
| `yarn build:app` | ⏭️ NOT RUN | `prepare`'s "Build app" step is guarded by `if: steps.integration-scope.outputs.skip != 'true'` (`ci.yml:187`) and this diff sets `skip=true`. No local substitute run: a build-script and test-only change that touches no app code, module file or locale string cannot plausibly break the Next.js build. |
| Playwright integration suite | ⏭️ NOT RUN | `ephemeral-integration` skipped by the same path filter. This was the previous finding; it no longer leaves the invariant unguarded, because the jest guard now covers it in a job that does run. |

Also green: `lint` (33s), `ds-lint` (43s), `scope` (31s), `audit-scope` (14s), `license/cla`, and all four CodeQL analyses (`javascript-typescript`, `python`, `actions`, plus the aggregate `CodeQL` gate). `docker-build`, `merge-coverage`, `mutate`, `audit`, `comment` and `[code]smith` were skipped by their own path and event filters, not by anything this PR did.

The PR head is `MERGEABLE` against `develop` with no conflicts. `mergeStateStatus` is `BLOCKED` solely because of my own standing `CHANGES_REQUESTED` review, which this approval supersedes.

## Findings

None. Both items from the previous review are closed:

- **Previous 🔹 Minor — guard test placed where CI cannot run it.** ✅ Resolved by `41625c07a`. Addressed exactly as the suggested fix described, including the local-checkout question I had left to the author's discretion: the new suite skips via `describeBuiltPackage` when `dist` is absent rather than failing loudly like create-app's `requirePackageBuild`. That is a legitimate choice of the two I offered — the cli's `test` script is plain `jest` and never builds first, so failing hard would surface on every clean local checkout, whereas create-app's `test` script runs `node build.mjs` first and can afford the stricter form. The trade-off is that a local run on an unbuilt checkout passes without asserting anything; the CI path, which is the one the finding was about, always asserts. The reasoning is written into the file, so it is a decision on the record rather than an accident.
- **Previous 💅 Nit — unconditional "Published create-app/agentic/" log line.** ✅ Resolved by `6d7bbf6fe`.

Keeping the TC-INT-008 guard alongside the jest one is the right call and not redundancy worth removing: it costs nothing and covers the integration-runner path, where the jest guard covers the affected-package path. The two comments cross-reference each other, so a future reader will not delete one assuming the other is authoritative.

## 💥 Breaking Changes

- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — the only user-visible change is build-log wording, which nothing parses
- [x] Where a contract had to change: not applicable

The published `dist/agentic` layout is unaffected — consumers of `@open-mercato/cli/dist/agentic` see the same tree, and only the order of operations that produces it changed. The two sibling directories the build uses exist solely between the start of `afterBuild` and the final rename, and the new test asserts precisely that. No surface listed in `BACKWARD_COMPATIBILITY.md` is touched.

## 🧪 Test Coverage

Coverage is now what the change warrants, in both of the places that matter.

`packages/cli/src/lib/__tests__/agentic-build-artifacts.test.ts` asserts that neither `dist/agentic.staging` nor `dist/agentic.previous` survives, reading the real artifact tree produced by `prepare`'s `yarn build:packages` — verified above to have executed rather than skipped on this head. It is a genuine regression guard for the two most likely ways this refactor could be broken later: dropping the final `renameSync(agenticStaging, agenticDist)` would leave `agentic.staging` behind, and dropping the `rmSync(agenticPrevious)` would leave `agentic.previous` behind. Either fails the `test` check on the very PR that introduces it, which is exactly what the previous finding asked for.

`TC-INT-008.spec.ts:300`–`:306` keeps the same invariant on the integration side, where `ensureCliBuilt()` in `beforeAll` runs the real `packages/cli/build.mjs`, so that assertion reads genuine post-build state too. The same spec's `expectedGuideOutputNames()` strict `toEqual` at `:349`–`:350` independently covers the purge removal — an emitted `core.<module>.md` stub or an unreachable `<pkg>.md` guide would fail it.

What remains uncovered is the timing window itself: no test proves a concurrent reader never observes an incomplete tree. As I noted last time, that is the right call — a sampling harness would trade a real regression signal for a test that flakes on loaded CI runners, the 5 ms sampling evidence in the PR body (1634/2234 samples missing before, 0/1880 after) belongs in the description, and #5064 set that precedent on the create-app side.

pkarw:

This was a **re-review** of head `6d7bbf6fe`, run because my `changes-requested` verdict at `f7b0209f4` no longer covered the diff after three commits landed. The full report is in the review above, submitted through the review API.

**Outcome.** The single minor finding from the previous review — the staging-artifact guard living only in the Playwright suite that CI's integration scope skips for a `packages/cli`-only diff — is fixed by `41625c07a`, and I verified the fix end to end rather than accepting the new file at face value: jest picks the path up (`jest.config.cjs:33`), the affected-package filter selects `@open-mercato/cli` for this diff (`ci.yml:624`), `packages/cli/dist` is genuinely present in the `test` job via the `build-artifacts` download (`ci.yml:196`, `:585`) so `describeBuiltPackage` does not fall back to `describe.skip`, and `turbo.json` marks `test` as `"cache": false` so a stale cached pass cannot make the guard vacuous. Run `32115949151` confirms it: `PASS src/lib/__tests__/agentic-build-artifacts.test.ts`, with the cli total moving 1703 → 1705 — exactly the two `it.each` cases, which would have shown as skipped rather than passed had the gate short-circuited. The nit on the publish log line is fixed by `6d7bbf6fe`.

No new findings. The build change was re-confirmed against the merged head: every former writer to the published tree still targets `agenticStaging`, nothing writes into `agenticDist` before the rename, and the `origin/develop` merge at `a1708e8dc` touched none of the three files in scope.

**Validation source: GitHub checks** — every reported check on `6d7bbf6fe` is green (`prepare`, `test`, `lint`, `ds-lint`, `scope`, `audit-scope`, `license/cla`, all four CodeQL analyses). `build:app` and `ephemeral-integration` were skipped by the integration-scope path filter, which is expected for this diff and no longer leaves the invariant unguarded. No local gate commands were run.

**Labels.** `review` → `merge-queue`; `refactor`, `skip-qa`, `priority-low` and `risk-low` unchanged. Rationale updated in place above. No `ci-monitoring`: CI had already settled green before this run started, so no follow-up comment is owed. Assignee left with @adeptofvoltron; my claim assignee was removed with the lock so other skills do not read a stale signal.

**Autofix: skipped** (not my PR — re-run with `--autofix` to fix it here). Nothing needed fixing in any case.

The PR is approvable and now carries an approving review; `mergeStateStatus` was `BLOCKED` only by my own standing `CHANGES_REQUESTED`, which this approval supersedes. The QA gate does not apply (`skip-qa`, no `needs-qa`).

patzick (approved):

Issue #5276, opened 2026-08-13
bug(i18n): remaining toLocale*String(undefined | []) call sites still render in the browser locale


adeptofvoltron:
## Follow-up from #5160

#5160 closed #5116 by threading the application locale into `MonthGrid.tsx` and `AgendaList.tsx`, which were the two views named in that issue. The same bug class survives elsewhere in the calendar folder and well beyond it: passing `undefined` — or an empty array, which behaves identically — as the locale argument to `toLocale*String` falls back to the JavaScript runtime's default locale, so the text renders in the browser/OS language rather than the configured application locale.

The nearest instance is inside the calendar module itself. From @wojciechszyjka's review of #5160 (💅 Nit 3):

> `packages/core/src/modules/customers/components/calendar/editor/hooks.ts:56` — `formatClock` calls `date.toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' })`. An empty array behaves exactly like `undefined`: it falls back to the runtime's default locale, so the conflict-probe warning in the event editor renders its times in the browser locale, not the app locale.

It was correctly left out of #5160 as out of scope for #5116 and outside that diff.

Measured on `develop` at the time of filing, `packages/` still contains **55** `toLocale*String(undefined, …)` call sites and **9** `toLocale*String([], …)` call sites. Roughly two dozen of the former sit in `packages/core/src/modules/customers/components/detail/`; the empty-array form also appears in the `messages` and `inbox_ops` modules. So the stated goal of #5116 — every calendar surface, and by extension every date/time surface, follows the app locale — does not yet hold.

### Acceptance criteria

- [ ] `formatClock` in `packages/core/src/modules/customers/components/calendar/editor/hooks.ts` takes an explicit `locale` (threaded from `useLocale()`, matching the pattern the rest of the calendar folder now uses) instead of `[]`.
- [ ] The remaining `toLocale*String(undefined, …)` and `toLocale*String([], …)` call sites in `packages/` are audited and converted where they render user-facing text — `packages/core/src/modules/customers/components/detail/`, `messages`, and `inbox_ops` are the largest clusters. Genuinely locale-agnostic uses (internal keys, machine-readable output) can be left with a short justification.
- [ ] Consider a lint rule or a CI grep guard that fails on a new `toLocale*String` call whose first argument is `undefined` or `[]`, so the class cannot silently reappear. This was listed as optional in #5116 and is the durable fix.
[...]

PR #5160, opened 2026-08-10
fix(customers/calendar): format MonthGrid and AgendaList dates with the app locale (#5116)

adeptofvoltron:
Closes #5116

## 🎯 Goal
- Every calendar surface follows the application UI locale for dates and times, not the browser/OS locale — Month and Agenda views were the two remaining offenders after #3541 fixed the toolbar and Upcoming cards.

## 🔍 Problem
On a workstation whose browser/OS language differs from the configured application locale (a routine setup — English-configured hardware running a localized product), the Month grid's weekday header, day tooltips, and event pill times, and the Agenda view's day-group headings and event times, rendered in the browser's language and clock convention instead of the app's. The chrome around them (toolbar, Upcoming cards, Day/Week grid) was already correct because those components went through #3541 or already used `useLocale()`.

## 🔍 Root Cause
`MonthGrid.tsx` and `AgendaList.tsx` define module-scope date/time formatting helpers (`formatTime`, `fullDateLabel`, `groupLabelOf`) that called `toLocale*String(undefined, …)`. Passing `undefined` as the locale argument means "use the runtime's default locale" — the browser's, not the app's. `MonthGrid.tsx` never imported `useLocale` at all; `AgendaList.tsx` did import and call it, but only to select an i18n plural category, never to format dates. Every sibling component in the same folder (`CalendarHeader.tsx`, `TimeGrid.tsx`, `EventPeekPopover.tsx`, `UpcomingCards.tsx`, `EventBlock.tsx`, `CalendarEventEditor.tsx`, `CalendarToolbar.tsx`) already calls `const locale = useLocale()` and threads it explicitly into formatting.

## What Changed
- `packages/core/src/modules/customers/components/calendar/MonthGrid.tsx` — `formatTime` and `fullDateLabel` now take an explicit `locale: string` parameter; `MonthPill`, `MonthDayCell`, and the top-level `MonthGrid` component each call `useLocale()` and pass it through. The weekday-header label now uses `day.toLocaleDateString(locale, …).toLocaleUpperCase(locale)` instead of the locale-insensitive `.toUpperCase()`.
- `packages/core/src/modules/customers/components/calendar/AgendaList.tsx` — `formatTime` and `groupLabelOf` now take an explicit `locale: string` parameter; `AgendaRow` (which previously had no locale in scope) now calls `useLocale()`, and `AgendaDayHeader`'s existing `locale` is threaded into `groupLabelOf`.
[...]

PR #5329, opened 2026-08-15
fix(shared): localize relative time strings in the app locale (#5286)

### LINKED ISSUE #5286: bug: Relative time strings ("X minutes ago", "X hours ago") always show in English regardless of the app's language
## Summary

The "Zaktualizowano" (Updated) column on the claims list shows every row's timestamp as a plain English relative-time string ("44 minutes ago", "2 hours ago") even though the entire rest of the page — headings, labels, statuses, tabs — is in Polish. Seen directly on `/backend/warranty_claims` while logged in with the app set to Polish.

## Steps to reproduce

1. Log in with the app set to Polish.
2. Go to `/backend/warranty_claims`.
3. Look at the "Zaktualizowano" column for any row.

## Expected result

The relative-time text is fully in Polish, matching the rest of the page.

## Actual result

It's always in English (eg "8 minutes ago"), no matter the language setting.

Image

## Environment

- Branch: `feat/warranty-claims-desk`
- Page: claims list, "Zaktualizowano" column

## Additional context

This same style of relative-time text (English-only) likely shows up anywhere else in the app that uses "time ago" formatting, not just this one column — worth checking a few other list pages to see how widespread it is. Given how generic this piece of UI is, it may be worth reporting as its own separate ticket rather than tied specifically to this PR.

adeptofvoltron:
Closes #5286

## 🎯 Goal
- Relative timestamps ("44 minutes ago", "2 hours ago") now render in the application language everywhere they appear, instead of always falling back to English while the surrounding page is translated.

## 🔍 Problem
On a Polish-configured app, the "Zaktualizowano" column and every other "time ago" surface rendered English text ("8 minutes ago") next to fully translated headings, labels and statuses. The reporter noted the defect is generic to the shared helper rather than specific to one list page, and that is exactly what the code shows.

## 🔍 Root Cause
`formatRelativeTime` in `packages/shared/src/lib/time.ts` built the `Intl.RelativeTimeFormat` instance and then never used it whenever a `translate` function was supplied: that branch short-circuited first and hand-assembled the string as `` `${magnitude} ${unit}${magnitude === 1 ? '' : 's'} ${suffix}` ``. Only the trailing "ago" / "from now" went through the translator (against keys `time.relative.ago` / `time.relative.fromNow` that no dictionary defines, so even that resolved to the English fallback). The unit name came straight from the internal English `RelativeTimeUnit` token and the plural form from an English `+ 's'` rule, which is wrong for Polish (`minut` / `minuty` / `minutę`) and every other locale. Since the notification renderers and the deals list all call the helper with `{ translate: t }`, they all took the broken branch.

## What Changed
- `packages/shared/src/lib/time.ts` — `Intl.RelativeTimeFormat` is now the primary formatter: it renders the number, unit name, plural form and suffix in the requested locale. `translate` is demoted to what it can actually do correctly — supplying the suffix in the degraded fallback path used only when the runtime has no `Intl.RelativeTimeFormat`. The option is documented accordingly; the exported signature is unchanged.
- `packages/shared/src/lib/i18n/context.tsx` — new `useOptionalLocale()`, mirroring the existing `useOptionalT()`: returns `undefined` instead of throwing when no `I18nProvider` is in scope, for shared components that receive their translator as a prop and can render outside a provider (component gallery, unit tests).
- Threaded the application locale into the twelve call sites that were passing a translator but no locale, so the text follows the app language rather than the browser/OS one: the notification renderers in `core` (wms ×2, sales ×2, inbox_ops), `checkout` (×3), `ai-assistant`, `enterprise` (record_locks), the shared `packages/ui/src/backend/notifications/NotificationItem.tsx`, and the deals list column in `packages/core/src/modules/customers/backend/customers/deals/page.tsx` (added to the columns `useMemo` deps so a locale switch re-renders).
- `packages/core/src/modules/wms/widgets/notifications/__tests__/WmsLowStockRenderer.test.tsx` — its `i18n/context` module mock now also stubs `useLocale`.

Out of scope, deliberately: the call sites that pass **neither** a translator nor a locale (`NotesSection`, `ActivitiesSection`, `TimelineItemHeader`, several dashboard widgets) belong to the separate, already-assigned #5276 sweep and are untouched here.

## 🧪 Tests
- `packages/shared/src/lib/__tests__/time.test.ts` — four new cases plus one rewritten: a non-English locale localizes unit, number and suffix (`1 godzinę temu`, `10 minut temu`, `za 5 godzin`); the same instant with a `translate` function supplied still renders Polish and never leaks `minutes` / `ago`; locale plural rules are applied instead of English pluralization (`2 minuty temu` vs `5 minut temu`); the same instant renders distinctly across `en` / `pl` / `de` / `es`; and the translator test now pins its real contract — the suffix fallback when `Intl.RelativeTimeFormat` is absent.
- Verified the regression bites: with the `time.ts` change reverted and the new tests in place, the translator + locale case fails (1 failed / 15 passed); with the fix, `yarn workspace @open-mercato/shared test --testPathPatterns "lib/__tests__/time"` is 16 passed / 16 total.
- Validation gate (local runner, in order): `yarn build:packages` ✅ · `yarn generate` ✅ · `yarn build:packages` ✅ · `yarn i18n:check-sync` ✅ · `yarn i18n:check-usage` ✅ · `yarn typecheck` ✅ · `yarn test` — 28 of 29 workspaces green (core 9727, app 554, checkout 159, ui/shared/enterprise/ai-assistant all passing); `create-mercato-app` fails ❌ · `yarn build:app` ✅.
- The `create-mercato-app` failure is pre-existing and environmental, not caused by this change: it is the standalone AI-harness suite (live Codex/Claude adapter cases needing those CLIs, plus a `target.typecheck: yarn typecheck must complete within 120000ms` budget oracle). Re-running that workspace's tests on the same machine with this branch's changes stashed reproduces the identical failure (exit 1, 161 failing cases), so the baseline is red independently of this PR.

## 💥 Breaking Changes
- None to any exported signature: `formatRelativeTime(value, { translate, locale })` and `FormatRelativeTimeOptions` are unchanged, and `useOptionalLocale` is purely additive.
- One intentional behavior change inside the fix: when a caller supplies `translate` **and** the runtime has `Intl.RelativeTimeFormat`, the output is now the Intl-formatted string rather than the hand-built English one — that is the defect being removed. No dictionary in the repo defines `time.relative.ago` / `time.relative.fromNow`, so no existing translation is dropped; those keys still drive the no-Intl fallback.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5329"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789404912&installation_model_id=432243&pr_number=5329&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5329&signature=19fd8955723f98ca0a186983f9ec77d9f5669a97ee2c3de48a8f15f1577882db"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/ai-assistant/src/modules/ai_assistant/widgets/notifications/ConversationSharedRenderer.tsx
+++ b/packages/ai-assistant/src/modules/ai_assistant/widgets/notifications/ConversationSharedRenderer.tsx
@@ -7,7 +7,7 @@ import { Button } from '@open-mercato/ui/primitives/button'
 import { IconButton } from '@open-mercato/ui/primitives/icon-button'
 import { cn } from '@open-mercato/shared/lib/utils'
 import { formatRelativeTime } from '@open-mercato/shared/lib/time'
-import { useT } from '@open-mercato/shared/lib/i18n/context'
+import { useLocale, useT } from '@open-mercato/shared/lib/i18n/context'
 import type { NotificationRendererProps } from '@open-mercato/shared/modules/notifications/types'
 
 export function ConversationSharedRenderer({
@@ -17,6 +17,7 @@ export function ConversationSharedRenderer({
   actions = [],
 }: NotificationRendererProps) {
   const t = useT()
+  const locale = useLocale()
   const router = useRouter()
   const [executing, setExecuting] = React.useState(false)
   const isUnread = notification.status === 'unread'
@@ -34,7 +35,7 @@ export function ConversationSharedRenderer({
     }
   }
 
-  const timeAgo = formatRelativeTime(notification.createdAt, { translate: t }) ?? ''
+  const timeAgo = formatRelativeTime(notification.createdAt, { locale, translate: t }) ?? ''
 
   return (
     <div
--- a/packages/checkout/src/modules/checkout/widgets/notifications/CheckoutTransactionCompletedRenderer.tsx
+++ b/packages/checkout/src/modules/checkout/widgets/notifications/CheckoutTransactionCompletedRenderer.tsx
@@ -6,7 +6,7 @@ import { useRouter } from 'next/navigation'
 import { Button } from '@open-mercato/ui/primitives/button'
 import { cn } from '@open-mercato/shared/lib/utils'
 import { formatRelativeTime } from '@open-mercato/shared/lib/time'
-import { useT } from '@open-mercato/shared/lib/i18n/context'
+import { useLocale, useT } from '@open-mercato/shared/lib/i18n/context'
 import type { NotificationRendererProps } from '@open-mercato/shared/modules/notifications/types'
 
 export function CheckoutTransactionCompletedRenderer({
@@ -16,6 +16,7 @@ export function CheckoutTransactionCompletedRenderer({
   actions = [],
 }: NotificationRendererProps) {
   const t = useT()
+  const locale = useLocale()
   const router = useRouter()
   const [executing, setExecuting] = React.useState(false)
   const isUnread = notification.status === 'unread'
@@ -69,7 +70,7 @@ export function CheckoutTransactionCompletedRenderer({
             </h4>
             <span className="flex-shrink-0 text-xs text-muted-foreground flex items-center gap-1">
               <Calendar className="h-3 w-3" />
-              {formatRelativeTime(notification.createdAt, { translate: t }) ?? ''}
+              {formatRelativeTime(notification.createdAt, { locale, translate: t }) ?? ''}
             </span>
           </div>
           {amount ? (
--- a/packages/checkout/src/modules/checkout/widgets/notifications/CheckoutTransactionFailedRenderer.tsx
+++ b/packages/checkout/src/modules/checkout/widgets/notifications/CheckoutTransactionFailedRenderer.tsx
@@ -6,7 +6,7 @@ import { useRouter } from 'next/navigation'
 import { Button } from '@open-mercato/ui/primitives/button'
 import { cn } from '@open-mercato/shared/lib/utils'
 import { formatRelativeTime } from '@open-mercato/shared/lib/time'
-import { useT } from '@open-mercato/shared/lib/i18n/context'
+import { useLocale, useT } from '@open-mercato/shared/lib/i18n/context'
 import type { NotificationRendererProps } from '@open-mercato/shared/modules/notifications/types'
 
 export function CheckoutTransactionFailedRenderer({
@@ -16,6 +16,7 @@ export function CheckoutTransactionFailedRenderer({
   actions = [],
 }: NotificationRendererProps) {
   const t = useT()
+  const locale = useLocale()
   const router = useRouter()
   const [executing, setExecuting] = React.useState(false)
   const isUnread = notification.status === 'unread'
@@ -67,7 +68,7 @@ export function CheckoutTransactionFailedRenderer({
             </h4>
             <span className="flex-shrink-0 text-xs text-muted-foreground flex items-center gap-1">
               <Calendar className="h-3 w-3" />
-              {formatRelativeTime(notification.createdAt, { translate: t }) ?? ''}
+              {formatRelativeTime(notification.createdAt, { locale, translate: t }) ?? ''}
             </span>
           </div>
           <div className="mt-3 flex gap-2">
--- a/packages/checkout/src/modules/checkout/widgets/notifications/CheckoutUsageLimitReachedRenderer.tsx
+++ b/packages/checkout/src/modules/checkout/widgets/notifications/CheckoutUsageLimitReachedRenderer.tsx
@@ -6,7 +6,7 @@ import { useRouter } from 'next/navigation'
 import { Button } from '@open-mercato/ui/primitives/button'
 import { cn } from '@open-mercato/shared/lib/utils'
 import { formatRelativeTime } from '@open-mercato/shared/lib/time'
-import { useT } from '@open-mercato/shared/lib/i18n/context'
+import { useLocale, useT } from '@open-mercato/shared/lib/i18n/context'
 import type { NotificationRendererProps } from '@open-mercato/shared/modules/notifications/types'
 
 export function CheckoutUsageLimitReachedRenderer({
@@ -16,6 +16,7 @@ export function CheckoutUsageLimitReachedRenderer({
   actions = [],
 }: NotificationRendererProps) {
   const t = useT()
+  const locale = useLocale()
   const router = useRouter()
   const [executing, setExecuting] = React.useState(false)
   const isUnread = notification.status === 'unread'
@@ -73,7 +74,7 @@ export function CheckoutUsageLimitReachedRenderer({
             </div>
             <span className="flex-shrink-0 text-xs text-muted-foreground flex items-center gap-1">
               <Calendar className="h-3 w-3" />
-              {formatRelativeTime(notification.createdAt, { translate: t }) ?? ''}
+              {formatRelativeTime(notification.createdAt, { locale, translate: t }) ?? ''}
             </span>
           </div>
           <div className="mt-3 flex gap-2">
--- a/packages/core/src/modules/customers/backend/customers/deals/page.tsx
+++ b/packages/core/src/modules/customers/backend/customers/deals/page.tsx
@@ -33,7 +33,7 @@ import { ViewTabsRow } from './pipeline/components/ViewTabsRow'
 import { DealsKpiStrip } from '../../../components/DealsKpiStrip'
 import { E } from '#generated/entities.ids.generated'
 import { useOrganizationScopeVersion } from '@open-mercato/shared/lib/frontend/useOrganizationScope'
-import { useT } from '@open-mercato/shared/lib/i18n/context'
+import { useLocale, useT } from '@open-mercato/shared/lib/i18n/context'
 import { useConfirmDialog } from '@open-mercato/ui/backend/confirm-dialog'
 import {
   type CustomerDictionaryKind,
@@ -193,6 +193,7 @@ function formatGroupedAmount(amount: number | null | undefined): string | null {
 
 export default function CustomersDealsPage() {
   const t = useT()
+  const locale = useLocale()
   const { confirm, ConfirmDialogElement } = useConfirmDialog()
   const router = useRouter()
   const pathname = usePathname()
@@ -858,7 +859,7 @@ export default function CustomersDealsPage() {
               <span className="text-xs text-muted-foreground">{t('customers.deals.list.close.lost')}</span>
             )
           } else {
-            const relative = formatRelativeTime(expectedCloseAt, { translate: t })
+            const relative = formatRelativeTime(expectedCloseAt, { locale, translate: t })
             if (relative) {
               subtitle = <span className="text-xs text-muted-foreground">{relative}</span>
             }
@@ -984,7 +985,7 @@ export default function CustomersDealsPage() {
       },
       ...customColumns,
     ]
-  }, [customFieldDefs, dictionaryMaps, dictionaryOptions, isDealOverdue, loadOwnerFilterOptions, ownerNames, pipelineNames, resolvedOwnerFilterOptions, t])
+  }, [customFieldDefs, dictionaryMaps, dictionaryOptions, isDealOverdue, loadOwnerFilterOptions, locale, ownerNames, pipelineNames, resolvedOwnerFilterOptions, t])
 
   const { advancedFilterFields } = useAutoDiscoveredFields({ columns, customFieldDefs })
 
--- a/packages/core/src/modules/inbox_ops/widgets/notifications/ProposalCreatedRenderer.tsx
+++ b/packages/core/src/modules/inbox_ops/widgets/notifications/ProposalCreatedRenderer.tsx
@@ -6,7 +6,7 @@ import { useRouter } from 'next/navigation'
 import { Button } from '@open-mercato/ui/primitives/button'
 import { cn } from '@open-mercato/shared/lib/utils'
 import { formatRelativeTime } from '@open-mercato/shared/lib/time'
-import { useT } from '@open-mercato/shared/lib/i18n/context'
+import { useLocale, useT } from '@open-mercato/shared/lib/i18n/context'
 import type { NotificationRendererProps } from '@open-mercato/shared/modules/notifications/types'
 
 export function ProposalCreatedRenderer({
@@ -16,6 +16,7 @@ export function ProposalCreatedRenderer({
   actions = [],
 }: NotificationRendererProps) {
   const t = useT()
+  const locale = useLocale()
   const router = useRouter()
   const [executing, setExecuting] = React.useState(false)
   const isUnread = notification.status === 'unread'
@@ -70,7 +71,7 @@ export function ProposalCreatedRenderer({
             </h4>
             <span className="flex-shrink-0 text-xs text-muted-foreground flex items-center gap-1">
               <Calendar className="h-3 w-3" />
-              {formatRelativeTime(notification.createdAt, { translate: t }) ?? ''}
+              {formatRelativeTime(notification.createdAt, { locale, translate: t }) ?? ''}
             </span>
           </div>
 
--- a/packages/core/src/modules/sales/widgets/notifications/SalesOrderCreatedRenderer.tsx
+++ b/packages/core/src/modules/sales/widgets/notifications/SalesOrderCreatedRenderer.tsx
@@ -7,7 +7,7 @@ import { Button } from '@open-mercato/ui/primitives/button'
 import { IconButton } from '@open-mercato/ui/primitives/icon-button'
 import { cn } from '@open-mercato/shared/lib/utils'
 import { formatRelativeTime } from '@open-mercato/shared/lib/time'
-import { useT } from '@open-mercato/shared/lib/i18n/context'
+import { useLocale, useT } from '@open-mercato/shared/lib/i18n/context'
 import type { NotificationRendererProps } from '@open-mercato/shared/modules/notifications/types'
 import { formatMoney } from '../../components/documents/lineItemUtils'
 import { useSalesDocumentTotals } from './useSalesDocumentTotals'
@@ -28,6 +28,7 @@ export function SalesOrderCreatedRenderer({
   actions = [],
 }: NotificationRendererProps) {
   const t = useT()
+  const locale = useLocale()
   const router = useRouter()
   const [executing, setExecuting] = React.useState(false)
   const isUnread = notification.status === 'unread'
@@ -57,7 +58,7 @@ export function SalesOrderCreatedRenderer({
     }
   }
 
-  const timeAgo = formatRelativeTime(notification.createdAt, { translate: t }) ?? ''
+  const timeAgo = formatRelativeTime(notification.createdAt, { locale, translate: t }) ?? ''
 
   return (
     <div
--- a/packages/core/src/modules/sales/widgets/notifications/SalesQuoteCreatedRenderer.tsx
+++ b/packages/core/src/modules/sales/widgets/notifications/SalesQuoteCreatedRenderer.tsx
@@ -5,7 +5,7 @@ import { FileText, ExternalLink, DollarSign, User, Calendar } from 'lucide-react
 import { useRouter } from 'next/navigation'
 import { Button } from '@open-mercato/ui/primitives/button'
 import { cn } from '@open-mercato/shared/lib/utils'
-import { useT } from '@open-mercato/shared/lib/i18n/context'
+import { useLocale, useT } from '@open-mercato/shared/lib/i18n/context'
 import { formatRelativeTime } from '@open-mercato/shared/lib/time'
 import type { NotificationRendererProps } from '@open-mercato/shared/modules/notifications/types'
 import { formatMoney } from '../../components/documents/lineItemUtils'
@@ -27,6 +27,7 @@ export function SalesQuoteCreatedRenderer({
   actions = [],
 }: NotificationRendererProps) {
   const t = useT()
+  const locale = useLocale()
   const router = useRouter()
   const [executing, setExecuting] = React.useState(false)
   const isUnread = notification.status === 'unread'
@@ -99,7 +100,7 @@ export function SalesQuoteCreatedRenderer({
             </div>
             <span className="flex-shrink-0 text-xs text-muted-foreground flex items-center gap-1">
               <Calendar className="h-3 w-3" />
-              {formatRelativeTime(notification.createdAt, { translate: t }) ?? ''}
+              {formatRelativeTime(notification.createdAt, { locale, translate: t }) ?? ''}
             </span>
           </div>
 
--- a/packages/core/src/modules/wms/widgets/notifications/WmsLowStockRenderer.tsx
+++ b/packages/core/src/modules/wms/widgets/notifications/WmsLowStockRenderer.tsx
@@ -7,7 +7,7 @@ import { Button } from '@open-mercato/ui/primitives/button'
 import { IconButton } from '@open-mercato/ui/primitives/icon-button'
 import { cn } from '@open-mercato/shared/lib/utils'
 import { formatRelativeTime } from '@open-mercato/shared/lib/time'
-import { useT } from '@open-mercato/shared/lib/i18n/context'
+import { useLocale, useT } from '@open-mercato/shared/lib/i18n/context'
 import type { NotificationRendererProps } from '@open-mercato/shared/modules/notifications/types'
 
 export function WmsLowStockRenderer({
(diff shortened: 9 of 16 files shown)


adeptofvoltron:

_Supersedes the earlier rationale posted by `om-auto-fix-issue`: an independent (non-author) review has now been submitted through the review API, so the pipeline state moves on._

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, and an approving review from a reviewer other than the PR author is now on record, so the review gate that previously held this PR in `review` is satisfied; `needs-qa` stays on, so the QA-approval gate still holds the actual merge until `qa-approved` is added.
- 🐛 `bug` — this is corrective work on a reported defect (#5286), where relative timestamps rendered in English regardless of the configured application language, rather than a new capability.
- 🧪 `needs-qa` — the change alters user-visible text on `.tsx` surfaces outside tests (nine notification renderers, the shared `NotificationItem`, and the deals list column), so it falls outside the automated-verification exemption and a human should confirm the localized strings render correctly with the application set to a non-English language.
- 🔹 `priority-medium` — an ordinary user-facing bug: the text is wrong in one language, but nothing is broken, unavailable, or unsafe, so it does not reach the release-blocking or security bar.
- 🟡 `risk-medium` — the edit lands in a shared helper (`packages/shared/src/lib/time.ts`) consumed across six packages, which argues against `risk-low`, but it touches no auth, data scoping, money, schema, or event reliability, and it ships regression tests plus a passing mutation-testing job on the changed file, which argues against `risk-high`.

adeptofvoltron:
🤖 `om-open-pr` completed: PR opened against `develop` with the fix, regression tests and the full label set. Lock handed off to `om-auto-review-pr` — chain continues on this PR.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock — code review of the fix. Started: 2026-08-15T16:56:17Z.

adeptofvoltron:
# 🔍 Code Review

**Verdict: approve** — with one nit already fixed in this pass. The change removes a real, reproducible defect at the right level (the shared helper, not the reporting page), keeps every exported signature intact, and ships regression tests that fail without it.

Reviewed at head `deec1cd6`, base `develop`. Note: GitHub does not allow the PR author to submit a formal approving review on their own PR, so this review is posted as a comment rather than through the review API — a human approval is still required before merge.

## ✅ What's right

- **The root cause is fixed where it lives.** `packages/shared/src/lib/time.ts:44` previously checked `translate` before ever consulting the `Intl.RelativeTimeFormat` instance it had just constructed, then assembled `` `${magnitude} ${unit}${magnitude === 1 ? '' : 's'} ${suffix}` `` from an internal English unit token and an English `+ 's'` plural rule. Inverting the precedence so `if (rtf) return rtf.format(valueToFormat, unit)` runs first is the minimal correct edit; nothing else in the module needed to move.
- **The fallback path is preserved rather than deleted.** When the runtime genuinely lacks `Intl.RelativeTimeFormat`, the helper still hand-builds the string and still routes the suffix through `translate` against the same `time.relative.ago` / `time.relative.fromNow` keys. No behavior and no key was dropped — the option was narrowed to what it can actually do correctly.
- **The fix reaches the user.** Fixing only `time.ts` would have left the twelve `{ translate: t }` call sites formatting in the *runtime* locale, so a Polish app in an English browser would still have read "10 minutes ago". Threading `useLocale()` through those call sites is what makes the reported symptom actually disappear, and it matches the pattern #5160 established for this bug class rather than inventing a new one.
- **`useOptionalLocale` is the right shape for `NotificationItem`.** That component takes its translator as a prop precisely so it can render without an `I18nProvider` (the design-system gallery and `NotificationPanel.test.tsx` both do). A throwing `useLocale()` would have broken those; the new hook mirrors the existing `useOptionalT()` exactly, so it introduces no new concept.
- **No scope creep.** The call sites that pass neither translator nor locale are left for #5276, which is assigned to someone else, and the diff says so.

## 🧩 Findings

**Nit — docstring added to unchanged code (`packages/shared/src/lib/time.ts:15`) — fixed in `deec1cd6`.** The first pass added a `/** Application locale; pass useLocale() … */` line above the `locale` field, which this PR does not otherwise touch; `.ai/review-checklist.md` §15 and its anti-pattern table both call that out ("Added docstring on unchanged function — Low"). The guidance itself is worth keeping, so it was folded into the `translate` doc block — the one option whose semantics this diff genuinely changes — instead of being deleted outright.

No blockers and no majors. Nothing else in the checklist applies adversely:

- **Backward compatibility (§21).** `formatRelativeTime`'s parameters, `FormatRelativeTimeOptions`, and `RelativeTimeTranslator` are byte-identical in shape; `useOptionalLocale` is purely additive; no event ID, spot ID, route, column, DI key, ACL feature, notification type or import path is touched. The one behavior change — Intl output instead of the hand-built English string when both a translator and `Intl` are present — is the defect itself, and no dictionary in the repo defines the two `time.relative.*` keys, so no existing translation is silently dropped.
- **Shared-package hygiene (`packages/shared/AGENTS.md`).** No `any`, no domain import, no new exported `unknown`; `useOptionalLocale` returns the narrow `Locale | undefined`.
- **i18n (§13).** No hardcoded user-facing string is added; the change removes hardcoded English rather than adding any.
- **React correctness.** `locale` was added to the deps array of the deals-page columns `useMemo` (`packages/core/src/modules/customers/backend/customers/deals/page.tsx:989`), so a locale switch re-renders the column instead of serving a memoized English string.
- **Diff-level auto-detections.** No removed export, no data-scoping change, no schema change, no raw `fetch`, no `alert(`/custom toast, no one-letter identifier.

## 🧪 Tests

- The regression is pinned, not merely asserted: reverting `time.ts` with the new tests in place fails exactly the translator-plus-locale case (1 failed / 15 passed), and the full file is 16/16 green with the fix.
- Coverage is behavioral rather than snapshot-shaped — unit name, number, plural form and suffix in Polish, English pluralization explicitly excluded (`not.toMatch(/minutes|ago/)`), locale plural rules across `2 minuty` vs `5 minut`, and a cross-locale distinctness check over `en`/`pl`/`de`/`es` in the spirit of the host-independent pattern #5160 introduced.
- `useOptionalLocale`'s provider-less branch is exercised indirectly by the existing `NotificationPanel` tests, which render `NotificationItem` outside an `I18nProvider`; that matches the precedent that `useOptionalT` carries no dedicated unit test of its own.
- Validation gate (local runner): `build:packages` ✅ · `generate` ✅ · `build:packages` ✅ · `i18n:check-sync` ✅ · `i18n:check-usage` ✅ · `typecheck` ✅ · `test` 28/29 workspaces ✅ · `build:app` ✅. The single failing workspace, `create-mercato-app`, is red on this machine independently of the change — its live Codex/Claude adapter cases need those CLIs and one oracle enforces a 120-second `yarn typecheck` budget; re-running it with the branch stashed reproduces the identical failure (exit 1, 161 cases), which is the evidence for calling it pre-existing rather than caused here.

## ⏱️ CI

Required checks were still running when this review was submitted (`mutate (shared, src/lib/time.ts)`, `prepare`, `Analyze (javascript-typescript)` in progress; `lint`, `ds-lint`, `scope`, `audit-scope` already green). The verdict is not waiting on them — pending checks are disclosed here rather than blocking the report, and a follow-up comment will record the settled CI outcome.

## 📸 QA

The PR keeps `needs-qa`: it changes user-visible text on `.tsx` surfaces, so it falls outside the automated-verification exemption. Manual QA instructions are in a separate comment.

adeptofvoltron:

This PR passed automated review and requires manual QA (`needs-qa`, no `skip-qa`). It stays in the `review` pipeline state rather than `merge-queue` because GitHub refuses a self-approval from the PR author, so a human approving review is still outstanding — the QA-approval gate (`qaGate: true`) additionally holds the merge until `qa-approved` is added. QA reviewer: when you pick it up, swap the pipeline label to `qa` and run the routes below.

The whole change is about which language a timestamp renders in, so every check below is "switch the app to a non-English language and read the text". Polish is the language from the original report and the one the tests pin, so it is the best choice.

### P1 — Relative timestamps in the notification feed

**Where to click**
- Any `/backend/*` page → the bell icon in the top bar → the notification dropdown (`packages/ui/src/backend/notifications/NotificationItem.tsx`, the default renderer for every notification).
- The same dropdown with a module-specific notification present, which uses its own renderer: WMS low stock / reservation shortfall, sales order created / quote created, checkout transaction completed / failed / usage limit reached, inbox_ops proposal created, AI assistant conversation shared, record-locks incoming changes.

**What to verify**
- With the app language set to Polish, each notification's timestamp reads as Polish text (`10 minut temu`, `2 godziny temu`, `wczoraj`) — not `10 minutes ago`.
- Switch the app language back to English and reopen the dropdown: the same rows read in English again, with no stale Polish text left over from the previous render.
- Repeat with the browser/OS language set to something *different* from the app language (for example an English browser with the app in Polish). The timestamp must follow the **app** language, since that is precisely the case the fix addresses.
- Check at least one notification in each age bucket if you can produce them — seconds, minutes, hours, days — so the unit name and not just the suffix is exercised.

**What can go wrong**
- A renderer that still shows English while its neighbours are translated would mean one of the twelve call sites was missed.
- A blank or `—` timestamp where text used to be would mean the helper returned `null` unexpectedly.
- Text that follows the browser language instead of the app language means the locale is not reaching the helper.
- A locale switch that leaves the old language on screen until a hard refresh would point at a stale memo.

### P1 — Deals list "expected close" column

**Where to click**
- `/backend/customers/deals` (`packages/core/src/modules/customers/backend/customers/deals/page.tsx`), the close-date column.

**What to verify**
- For an open deal with a future expected close date and no overdue/won/lost status, the subtitle under the date reads in the app language (`za 5 dni` in Polish, `in 5 days` in English).
- The overdue / won / lost subtitles are unchanged — they were already translated through their own keys and this PR does not touch them.
- Switch the app language with the list open and re-render: the subtitle switches language along with the rest of the page (this is the memo-dependency change).

**What can go wrong**
- The subtitle keeping the previous language after a locale switch while the column headers change.
- An overdue or terminal-status deal accidentally showing a relative string instead of its status label.

### P2 — Perceived performance and layout

**Where to click**
- Cold-load `/backend/customers/deals` and open the notification dropdown, ideally with a screenshot.

**What to verify**
- No new loading delay or layout shift: localized strings can be longer than their English equivalents (`przedwczoraj` vs `2 days ago`), so confirm the column and the dropdown row do not wrap awkwardly or truncate mid-word at a narrow/mobile viewport.

**What can go wrong**
- Truncation or wrapping in a language with longer relative-time phrases, most visibly in the notification row and the deals column.

### Pass/fail
- All routes pass → remove the `qa` label and add `merge-queue` plus `qa-approved` (this clears the QA-approval gate; a human code approval is still needed separately).
- Any route fails → remove the `qa` label, add `qa-failed`, and leave a comment describing which surface and which language showed the wrong text.

adeptofvoltron:
## 📸 UI verification — not run in this environment

The chain would normally follow the review with an `om-auto-qa-pr` browser pass, since this PR does touch user-facing surfaces. It did not run here, and that is an environment limit rather than a judgement that the change needs no exercising.

The isolated worktree this run used has no `.env`, no application container (Docker is not running on the host), and no dev server. A local Postgres is listening on `5432`, but that is the developer's own database — pointing a freshly provisioned app at it and running migrations against it is a side-effecting action the repository's own guidance reserves for an explicit human go-ahead, so the run stopped short of it rather than provisioning one unasked.

Practically this costs little: the behavior under test is which language a timestamp string renders in, which the new unit tests assert directly against `Intl` output in Polish, German, Spanish and English. What a browser pass would add is the visual half — that no row truncates or wraps badly with longer localized phrases, and that a live locale switch re-renders every surface. Both are covered by the manual-QA instructions above, and the PR keeps `needs-qa` so that check still happens before merge.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the fix is implemented, reviewed (approve, one nit fixed in-pass), labelled and documented; manual QA and a human approving review are the only work left, plus the CI result this agent is still watching. Lock released — the PR is free for anyone to act on.

adeptofvoltron:
## ⏱️ 🤖 `om-auto-review-pr` — CI result

CI has settled green: **28 checks successful, 3 skipped, 0 failing**. The pending checks disclosed in the review body have all completed.

- ✅ `test`, `lint`, `ds-lint`, `prepare`, `scope`, `audit-scope`, `docker-build`, `comment`
- ✅ all fifteen `ephemeral-integration` shards (1/15 … 15/15)
- ✅ `mutate (shared, src/lib/time.ts)` — the mutation-testing job targeted the exact file this PR rewrites and passed, which is meaningful corroboration that the new tests actually constrain the changed branch rather than just executing it
- ✅ CodeQL: `Analyze (javascript-typescript)`, `Analyze (actions)`, `Analyze (python)`
- ⏭️ skipped by their own conditions, not by this change: `audit`, `[code]smith`, `merge-coverage`

The verdict from the review pass is unchanged: no blockers, no majors. `ci-monitoring` is removed — no further follow-up will come from this agent. What remains is human work: a human approving review (GitHub rejects the author's own approval) and manual QA per the instructions above, after which `qa-approved` clears the QA gate.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T07:38:29Z. Other auto-skills will skip this PR until the lock is released.

Note: the previous chain (`om-auto-fix-issue` → `om-open-pr` → `om-auto-review-pr`) released its lock on 2026-08-15T16:59:36Z and its CI follow-up landed at 17:20Z, so this is a fresh, independent review pass — this time by a reviewer who is not the PR author, so the verdict can be submitted through the review API rather than as a comment.

pkarw (approved):
# 🔍 Code Review

## 🎯 Summary

This PR fixes #5286 — relative timestamps rendering in English on a non-English application — and it fixes it at the right altitude. The defect was not in the reporting page but in `packages/shared/src/lib/time.ts`, where `formatRelativeTime` constructed an `Intl.RelativeTimeFormat` instance and then threw it away whenever a `translate` function was supplied, hand-assembling `` `${magnitude} ${unit}${magnitude === 1 ? '' : 's'} ${suffix}` `` from an internal English unit token and an English `+ 's'` plural rule. Because every notification renderer and the deals list call the helper with `{ translate: t }`, every one of them took that branch. The fix reorders the two paths so `Intl` is the primary formatter and demotes `translate` to the no-`Intl` fallback, then threads the application locale through the twelve call sites that were passing a translator but no locale.

**Verdict: approve.** No blockers and no majors. Three nits are recorded below for a follow-up pass; none of them justifies holding this change, and the repo's stricter Medium-blocks rule is not triggered because nothing here rises above nit.

Reviewed at head `deec1cd6` against base `develop` (`43440f24`). Note for the record: an earlier automated pass by the PR author posted its review as a *comment*, because GitHub refuses a self-approval from the author. This review is submitted through the review API by a different reviewer, so it counts as a real approving review.

## 🧪 Validation Gate

**Validation source: GitHub checks.** CI is fully green at this exact head, so per this repository's GitHub-checks-first rule no local gate was re-run.

| Check | Result |
|---|---|
| `test`, `lint`, `ds-lint`, `prepare`, `scope`, `audit-scope`, `docker-build`, `comment` | ✅ pass |
| `ephemeral-integration` (all 15 shards) | ✅ pass |
| `mutate (shared, src/lib/time.ts)` | ✅ pass |
| CodeQL — `javascript-typescript`, `actions`, `python` | ✅ pass |
| `audit`, `[code]smith`, `merge-coverage` | ⏭️ skipped by their own conditions |

28 successful, 3 skipped, 0 failing, 0 pending. The `mutate` job targeted the exact file this PR rewrites and survived, which is meaningful corroboration that the new tests constrain the changed branch rather than merely executing it. The head is also `origin/develop` plus exactly two commits, so the branch is conflict-free and current with its base — the diff under review is the diff that will merge.

## ✅ What's right

- **The root cause is removed, not papered over.** `packages/shared/src/lib/time.ts:48` now consults `rtf` first, so the number, unit name, plural form and suffix all come from CLDR in the requested locale. That is what makes Polish `minutę` / `minuty` / `minut` come out correctly where an English `+ 's'` rule never could.
- **The provider contract was reasoned about per call site, not applied uniformly.** This is the detail that could easily have gone wrong. `useLocale()` throws outside an `I18nProvider`, exactly as `useT()` does. In all nine module renderers the new `const locale = useLocale()` sits directly beneath an existing `const t = useT()`, so it adds no failure mode those components did not already have. `packages/ui/src/backend/notifications/NotificationItem.tsx:79` is the one component that receives its translator as a **prop** rather than from the hook, and it is precisely the one that got the new non-throwing `useOptionalLocale()`. That distinction is correct and easy to miss.
- **The stale-closure hazard was caught.** `packages/core/src/modules/customers/backend/customers/deals/page.tsx:988` adds `locale` to the `columns` `useMemo` dependency array, so switching application language re-renders the column instead of serving a timestamp formatted in the previous locale.
- **The fix incidentally removes a latent SSR hydration mismatch.** Passing no locale let `Intl.RelativeTimeFormat` fall back to the *runtime* locale, which is the Node default on the server and the browser/OS locale on the client. Pinning the application locale makes both sides agree.
- **The exported surface is intact.** `formatRelativeTime(value, { translate, locale })` and `FormatRelativeTimeOptions` are unchanged, and `useOptionalLocale` is purely additive. Neither is listed in `BACKWARD_COMPATIBILITY.md`'s protected-surface tables; the one entry that is adjacent, `useT()` ("MUST NOT change return type"), is untouched.
- **The regression was proven to bite.** The PR reports that reverting `time.ts` with the new tests in place fails the translator + locale case, which is the evidence that separates a real regression test from one that merely executes the new line.

## 🔸 Nits

- **`translate` is now unreachable in practice, yet still threaded through all twelve call sites.** `packages/shared/src/lib/time.ts:48` returns before the fallback whenever `Intl.RelativeTimeFormat` exists, which is every runtime this project supports. The option is doubly dead: I confirmed no dictionary in the repository defines `time.relative.ago` or `time.relative.fromNow`, so even if the branch were reached it would resolve to the English fallback. Keeping the parameter is right for signature compatibility, but continuing to pass `{ locale, translate: t }` at twelve call sites tells the next reader that `translate` still does something. Consider marking it `@deprecated` in `FormatRelativeTimeOptions` and dropping the argument from the call sites in a follow-up, so the only remaining mention is the fallback that actually consumes it.
- **The new Polish assertions pin byte-exact CLDR output.** `packages/shared/src/lib/__tests__/time.test.ts` asserts `'1 godzinę temu'`, `'10 minut temu'`, `'za 5 godzin'`, `'2 minuty temu'` and `'5 minut temu'` with `toBe`. These pass here and in CI (I confirmed ICU 78.2 produces exactly `10 minut temu`), but they couple the suite to the ICU dataset: a Node built `--with-intl=small-icu` has only `en-US` data and silently falls back, turning a correct implementation into five red tests with a misleading message. The two robust assertions already in the same file — `expect(polish).not.toMatch(/minutes|ago/)` and the "renders differently per locale" set-size check — are the style that survives an ICU bump. Worth converting the exact-match cases, or adding a guard that skips them when full ICU is absent.
- **Three shadowing copies of the helper still hardcode English.** `packages/core/src/modules/customers/components/detail/CompanyCard.tsx:89`, `.../ChangelogEntryRow.tsx:47` and `packages/core/src/modules/staff/backend/staff/timesheets/projects/page.tsx:86` each define their own local `formatRelativeTime`, so they never benefit from this fix. That is outside this PR's scope and plausibly belongs with the #5276 sweep alongside the no-argument call sites the description already defers, but it means #5286 is not fully closed at the product level until those three are folded in. Worth a tracking note so it does not get lost.

## 💥 Breaking Changes

- **Exported signatures:** none changed. `formatRelativeTime`, `FormatRelativeTimeOptions` and `RelativeTimeTranslator` keep their shapes; `useOptionalLocale` is additive. No protected surface in `BACKWARD_COMPATIBILITY.md` is touched, so no deprecation protocol applies.
- **One intentional behavior change**, which is the defect being removed: a caller supplying `translate` on a runtime that has `Intl.RelativeTimeFormat` now gets the Intl-formatted string instead of the hand-built English one. Because `numeric: 'auto'` was already in force on the Intl path, such callers will also start seeing CLDR's idiomatic forms — `now`, `yesterday`, `tomorrow` — where they previously saw `0 seconds ago` or `1 day ago`. That is an improvement rather than a regression, but it is a visible text change and is the reason `needs-qa` is the right call here.
- No dictionary loses a translation, since the two fallback keys were never defined anywhere.

## 🧪 Test Coverage

Coverage is proportionate and lands where the risk is. `packages/shared/src/lib/__tests__/time.test.ts` adds four cases and rewrites one: unit/number/suffix localization in a non-English locale, the translator-plus-locale combination that was the actual bug, locale plural rules against English pluralization, and cross-locale distinctness for a single instant. The rewritten translator test is the important one — it previously asserted the broken behavior as the contract, and now pins what `translate` genuinely does, the fallback suffix when `Intl.RelativeTimeFormat` is absent. `packages/core/src/modules/wms/widgets/notifications/__tests__/WmsLowStockRenderer.test.tsx` extends its `i18n/context` module mock with `useLocale`, which is required because a partial `jest.mock` of that module would otherwise leave the new hook `undefined`; I checked the other suites that mock the same module and none render the changed components, which matches CI being green.

The untested edge is the pure-UI half — that the twelve call sites actually render in the application language in a browser. That is inherently manual, it is exactly what `needs-qa` is for, and the detailed QA instructions already posted on this PR cover it.

## Conclusion

Approving. The change is small, correctly scoped, well-tested, and it fixes the defect at the shared helper rather than at one symptom. `needs-qa` stays on: the QA-approval gate holds the merge until a QA reviewer confirms the localized strings render in a non-English application and adds `qa-approved`.

pkarw:
## 🧪 Manual QA instructions (`needs-qa`) — state update

This PR is now **approved** and sits in `merge-queue`; the QA-approval gate holds the merge until `qa-approved` is added. This supersedes only the *pipeline-state* sentence of the detailed QA comment above, which was written while the PR was still stuck in `review` because GitHub refused the author's own approval. **The click paths in that comment are unchanged and remain the instructions to follow** — the head SHA has not moved (`deec1cd6`), so the diff they were derived from is still the diff under test.

QA reviewer: when you pick this up, swap `merge-queue` for `qa`, then run the routes in the earlier comment. Two additions worth exercising, both surfaced by this review:

### P1 — the idiomatic CLDR forms, not just the translated ones

**Where to click**
- Any `/backend/*` page → the bell icon → the notification dropdown, with a notification created **just now**, **yesterday**, and **a few minutes ago**.

**What to verify**
- A just-created notification reads as CLDR's idiomatic form for the locale (English `now`, Polish `teraz`) rather than `0 seconds ago`. A one-day-old one reads `yesterday` / `wczoraj` rather than `1 day ago`. This is a deliberate consequence of `numeric: 'auto'` now applying to translator-supplied call sites, and it is new text a reader will notice.
- Polish plural forms change with the number: `2 minuty temu` vs `5 minut temu`. English pluralization leaking through (`5 minutes temu`, or a bare `minut`) means the Intl path was bypassed.

**What can go wrong**
- Mixed-language output — a translated suffix beside an English unit — which is the exact symptom of the old code path being taken.

### P1 — language switching without a reload

**Where to click**
- The deals list (`/backend/customers/deals`), with an expected-close-date column visible → switch the application language in place, without a hard refresh.

**What to verify**
- The relative timestamps in that column re-render in the new language immediately. They are produced inside a memoized column definition, and this PR adds `locale` to its dependency list; a stale timestamp in the previous language here would mean that dependency is not doing its job.

**What can go wrong**
- Timestamps stay in the old language until a manual page reload.

### Pass/fail
- All routes pass → remove `qa`, add `merge-queue` plus `qa-approved` (this clears the QA-approval gate).
- Anything fails → apply `qa-failed` with the failing route, the language used, and a screenshot of the offending string.

pkarw:

Independent review pass at head `deec1cd6`: no blockers, no majors, three nits recorded in the review body (the now-unreachable `translate` option still threaded through twelve call sites, byte-exact CLDR assertions in the new tests, and three shadowing local `formatRelativeTime` copies that still hardcode English). Verdict submitted through the review API — unlike the earlier author-run pass, this reviewer is not the PR author, so the approval counts against branch protection.

Validation source: GitHub checks (28 successful, 3 skipped, 0 failing at this head; no local gate re-run). Pipeline moved `review` → `merge-queue`; `needs-qa` retained, so the QA-approval gate still holds the merge until `qa-approved` lands. autofix: skipped (not my PR — re-run with --autofix to fix it here). No CI follow-up is owed, so `ci-monitoring` was not applied.

Issue #1769, opened 2026-05-02
feat: Implementation of a Global Interceptor for Automated Custom Fields Mapping


alinadivante:
## Proposed solution

Instead of manually handling and fixing the mapping for each entity (Products, Customers, Orders, etc.) individually, I propose implementing a Global Interceptor or Decorator at the API level (likely within the NestJS/Fastify middleware stack).

This solution should:

- Automatically detect any field with the cf_ prefix across all allEntity types.

- Standardize the transformation of these fields into the customValues and customFields objects before the response is sent.

- Ensure that any new entity added to the system in the future will inherit this behavior without additional boilerplate code.

This will reduce code duplication and prevent bugs where some entities have "fixed" custom fields while others do not.

## Specification

**Does a spec exist for this feature?**
- [ ] Yes
- [ ] No
- [ ] Not sure

**If yes, provide the spec file path:**

## Additional context

The current implementation shows significant inconsistencies in how Custom Fields are handled across different modules. Below are specific examples of the current "tri-state" or inconsistent data flow:

- Products: Data is redundant, appearing as cf_ prefix, within customValues, and inside a customFields array simultaneously.

- Pay Links: High redundancy in the creation payload. For example, delivery_timeline is sent as a flattened cf_delivery_timeline key, inside the customFields object, and as a standalone root key.

- To-Do Tasks: Inconsistent mapping between Request and Response. The POST payload for /api/example/todos uses cf_ prefixes, but the GET response returns a flat structure without the standard customValues or customFields wrappers used elsewhere in the app.

- Teams & Orders: Inconsistent naming conventions between query parameters (filtering) and the returned JSON body.

Why a Global Interceptor?
A global interceptor would allow us to maintain consistency across the entire Open-Mercato framework. Instead of patching each entity (Products, Pay Links, To-Dos) individually, we can ensure that regardless of the module, the API consumer always receives and sends a predictable, standardized data structure. This is essential for maintaining an "Enterprise Grade" architecture..

pkarw:
🤖 `auto-fix-github` started by @pkarw at 2026-05-05T07:13:08Z. Other auto-skills will skip this issue until the lock is released.

pkarw:
Thanks @alinadivante — a fix PR is ready for this issue: https://github.com/open-mercato/open-mercato/pull/1800. Reassigning the issue to you for recheck/verification of the proposed fix.

pkarw:
[...]

PR #5334, opened 2026-08-16
feat(wms): preselect the sole warehouse and support custom fields in the zone dialog (#5239)

### LINKED ISSUE #5239: feat: wms - create zone dialog
1. When there's only one warehouse it should be auto pre-selected
2. The form must be CrudForm based + supporting custom fields

Image

Paul-Mlodochowki:
Tracking plan: .ai/runs/2026-08-17-issue-5239-wms-zone-dialog.md
Status: complete

Fixes #5239

> ✅ **The DO-NOT-MERGE banner is lifted.** Requirement 2 now has a root cause and a fix — see **Why requirement 2 did not persist** below. Every finding from the code review on `1ac4915` is addressed.

## What this does

Addresses both requirements from the issue:

1. **The sole warehouse is preselected.** When the tenant has exactly one warehouse, the dialog opens with it already chosen instead of an empty "type to search" combobox.
2. **The form carries custom fields**, and they now round-trip: the value is written, indexed, and read back by the zones list API.

## Why requirement 2 did not persist

The write chain was correct all along — the review verified both halves independently and reached `dataEngine.setCustomFields` with the right entity id, record, scope and value. The break was on the **read** side:

- The zones list is served by `HybridQueryEngine`, which projects `cf:*` out of **`entity_indexes`**, not by joining `custom_field_values`.
- The zone commands never called `emitCrudSideEffects`, so no `query_index.upsert_one` ever ran for a zone.
- The value therefore sat in `custom_field_values` where no read path consults it, and the API returned `null` — exactly the symptom `TC-WMS-028-zone-dialog.spec.ts:248` reported.

The fix adds `warehouseZoneCrudIndexer` and emits it from create, update, delete and all three undo handlers, always **after** the custom-field write so the index doc carries the new values. `events` is deliberately omitted — the module already emits `wms.zone.*` itself, and an events config here would introduce a second, undeclared `wms.warehouse_zone.*` id for the same write.

## How

- `commands/configuration.ts` round-trips values through the shared `parseWithCustomFields` / `setCustomFieldsIfAny` helpers, captures undo snapshots via `loadCustomFieldSnapshot`, restores them on **delete-undo** via `buildCustomFieldResetMap(before.custom, undefined)`, and emits the CRUD side effects described above.
- `api/zones/route.ts` decorates list responses with `decorateCustomFields` using `stripPrefixedKeys: true` (the canonical single-shape response — the route had no decorator before this PR, so no consumer reads top-level `cf_*`), and documents `customFields` with an explicit `CustomFieldDisplayEntry`-shaped schema.
- `zoneFormSchema` uses `.passthrough()` — a plain `z.object` stripped the collected `cf_*` values before `onSubmit` ever saw them.
- The zone edit dialog now carries `id` and `updatedAt` in `initialValues`, so CrudForm recognises it as an edit and derives the optimistic-lock header.

## ⚠️ Heads-up for reviewers: this touches a shared primitive

`ComboboxInput` synced its displayed label only while the input was **unfocused**. But `autoFocus` on the first `CrudForm` field takes focus before an async default value arrives, so the control rendered an empty label for a value the form had already committed — which is exactly what broke the preselection above. The guard is now *"the user is typing, or the only entry available is a self-mapping placeholder"* rather than *"the field has focus"*.

The placeholder half of that guard closes a regression the review reproduced on the previous head: after picking an option, a follow-up load that no longer carries the picked entry could overwrite the label with the raw record id.

**This affects every combobox in the product, not just WMS.** It is behavioral only — `seedOptions` is an existing prop, so no public contract changed. Five tests pin it, and each was checked against both the previous implementation and `origin/develop`:

| Test | Fails on `develop` | Fails on the intermediate guard |
|---|---|---|
| `shows a default value that arrives after autoFocus already took the field` | ✅ | — |
| `keeps a picked label when a follow-up load drops the option from the list` | — | ✅ (received `"wh-b"`) |
| `adopts a real label that arrives while the field is focused` | ✅ | — |
| `resumes syncing after a blur, so a later external change is not mistaken for typing` | ✅ | — |
| `does not resurrect the previous label when the parent keeps the value after a clear` | — | — |
| `does not overwrite the query while the user is typing into a focused field` | — | — |

The last two are behavior pins rather than regression tests, and say so in a comment.

## Testing

| Check | Result |
|---|---|
| `yarn build:packages` | ✅ 22/22 |
| `yarn generate` | ✅ no generated-file churn afterwards |
| `yarn i18n:check-sync` | ✅ all locales in sync |
| `yarn i18n:check-usage` | ✅ advisory only |
| `yarn typecheck` | ✅ 22/22 |
| `yarn test` | ⚠️ green except a host-limited baseline — see below |
| `yarn build:app` | ✅ |
| `npx eslint` on the changed files | ✅ 0 errors; the 2 `ComboboxInput` warnings reproduce on `develop` |
| WMS module tests | ✅ 39 suites / 246 tests |
| `packages/ui` ComboboxInput suite | ✅ 26 tests |

**Test-suite baseline.** This Windows host fails 6 `@open-mercato/core` suites (16 tests), 1 `@open-mercato/ui` suite (`utils/__tests__/format.test.ts`), plus suites in `queue`, `shared`, `cli`, `telemetry`, `ai-assistant` and `create-app`. All of them were re-run with this branch's changes stashed and fail identically — currency/ICU formatting, Windows path handling, and an `EPERM` on directory rename. None is in a file this PR touches.

### Integration test

`__integration__/TC-WMS-028-zone-dialog.spec.ts` is unchanged in shape and still the right coverage. It has **not** been re-run since the fix — that needs the ephemeral environment. Requirement 1 passed on the previous run; requirement 2's assertion is what the indexer fix targets.

Note on selectors: the spec deliberately uses `[data-crud-field-id="…"] input` rather than `getByLabel`, because `CrudForm` renders labels with no `htmlFor` and no `id` on the control. Filed as #5360 and linked from the spec.

## Labels

I do not have label permission on this repo. Current set is correct except that `changes-requested` should move to `review` now that the findings are addressed.

### DIFF
--- a/.ai/runs/2026-08-17-issue-5239-wms-zone-dialog.md
+++ b/.ai/runs/2026-08-17-issue-5239-wms-zone-dialog.md
@@ -1,0 +1,81 @@
+# Execution plan — land the #5239 zone dialog work by fixing every code-review finding (adopted from PR #5334)
+
+**Origin:** adopted — reconstructed by `om-auto-continue-pr` on 2026-08-17 because PR #5334 carried no execution plan.
+**PR:** #5334 · **Branch:** `feat/issue-5239-wms-zone-dialog` · **Base:** `develop`
+**Author:** @Paul-Mlodochowki — this plan interprets their intent; correct it by editing this file or commenting on the PR.
+
+## 🎯 Goal
+
+Make PR #5334 mergeable: deliver requirement 2 of #5239 (the zone dialog's custom fields actually persist and read back), remove the `ComboboxInput` label-corruption regression the review reproduced, and close every remaining review finding with the coverage the checklist demands.
+
+## Scope
+
+- `packages/core/src/modules/wms/` — zone commands, zones list route, the configuration page's zone dialog, command unit tests.
+- `packages/ui/src/backend/inputs/ComboboxInput.tsx` and its unit tests.
+- The `TC-WMS-028-zone-dialog` integration spec (comments/links only; its assertions already have the right shape).
+
+## Non-goals
+
+- Rewriting the zone commands onto `runCrudCommandWrite`. The review does not ask for it and it would widen the diff across a shared write path; the existing `emitCrudSideEffects` composition is the pattern the rest of this module already uses.
+- Fixing the `CrudForm` accessibility gap (no `htmlFor`/`id` on dialog fields) that finding **n3** describes. That is a framework-level defect touching every form in the product — it gets its own issue, linked from the spec comment.
+- Backfilling query-index coverage for zones created before this change. The index self-heals through the existing coverage refresh; no migration is in scope.
+
+## Evidence
+
+| Conclusion | Drawn from | Confidence |
+|---|---|---|
+| The goal is "fix everything the review found", not new feature work | The user's invocation, which pasted the full review verbatim and asked for the findings to be fixed | high |
+| Requirement 2 genuinely does not persist | `TC-WMS-028-zone-dialog.spec.ts:248` failing deterministically (`Expected: "Pick face …", Received: null`), reported in the PR body and confirmed by the review | high |
+| The write chain (client → route → command → `dataEngine.setCustomFields`) is correct | The reviewer executed both halves against the real schema/handler and observed the correct call | high |
+| The break is on the **read** projection, not the write | `HybridQueryEngine` (the DI-registered engine) serves `cf:*` from `entity_indexes`, and the zone commands never emit `emitCrudSideEffects`, so no `query_index.upsert_one` ever runs for a zone | high |
+| `ComboboxInput`'s new guard corrupts a picked label into the raw record id | The review reproduced it with a discriminating test: passes on `origin/develop`, fails on `1ac4915`, with only that file swapped | high |
+| The server-side custom-field change has zero unit coverage | `git diff` shows the only new unit tests are the two `ComboboxInput` cases | high |
+
+## Assumptions
+
+- **The indexer is the whole of B1.** The reviewer proved the write reaches `setCustomFields`; the hybrid engine reading `cf:*` out of `entity_indexes` is then the only remaining link that can turn a written value into `null`. If a live run still shows `null` after this change, the next suspect is `upsertIndexRow`'s custom-field projection, not the command wiring.
+- **Adding the indexer to zone commands is safe and additive.** Zones already declare `indexer: { entityType: E.wms.warehouse_zone }` on the CRUD route, so the entity type is already an indexed one; the commands were simply not emitting the side effect.
+- **No new event ids.** `emitCrudSideEffects` is wired with `indexer` only, not `events`, so the existing `wms.zone.created` / `wms.zone.updated` emissions stay the single event source and no undeclared `wms.warehouse_zone.*` event appears.
+- **Adoption mode is `auto`.** A user is in the loop, which normally means `ask`, but they supplied the goal outright (the pasted review plus "fix the findings"), so confirming a reconstructed plan would only cost a turn. Correct this by commenting on the PR.
+
+## Risks
+
+- `ComboboxInput` is a shared primitive; the M1 fix narrows when a focused field syncs, so it needs the M2 coverage cases to prove the other windows still behave.
+- B1 cannot be proven end-to-end without a live environment. Mitigation: the unit tests of Phase 3 pin the command-side contract, and the fix's mechanism is traceable in code from `markOrmEntityChange` to `query_index.upsert_one` to `entity_indexes`.
+
+## Progress
+
+> Convention: `- [ ]` pending, `- [x]` done. Append ` — <commit sha>` when a step lands. Do not rename step titles.
+
+### Phase 1: Already landed on this PR (reconstructed)
+
+- [x] 1.1 Preselect the sole warehouse, wire custom fields through route/commands/form, add the TC-WMS-028 integration spec and two ComboboxInput tests — 1ac4915
+
+### Phase 2: B1 — make the zone custom fields readable
+
+- [ ] 2.1 Emit CRUD side effects with a warehouse-zone indexer from the zone create/update/delete commands and their undo handlers
+
+### Phase 3: B2 — unit coverage for the server-side custom-field change
+
+- [ ] 3.1 Add the zone custom-field command suite (create with/without values, update, both undo resets, snapshot loader)
+
+### Phase 4: M1 and M2 — ComboboxInput
+
+- [ ] 4.1 Never sync a focused field to a self-mapping placeholder label
+- [ ] 4.2 Add the discriminating pick-then-stale-reload test plus the three blast-radius cases
+
+### Phase 5: Minor findings m1–m7
+
+- [ ] 5.1 Give the zone edit dialog optimistic locking (`id` + `updatedAt` in initialValues)
+- [ ] 5.2 Restore custom fields on delete-undo and stop discarding the snapshot
+- [ ] 5.3 Opt the new zones decorator into `stripPrefixedKeys` and type the `customFields` OpenAPI entry
+- [ ] 5.4 Feed the warehouse combobox from the cached query instead of a second round trip
+- [ ] 5.5 Narrow the `ZoneRow` / `ZoneFormValues` index signatures and short-circuit the custom-field snapshot load
+
+### Phase 6: Nits and follow-ups
+
+- [ ] 6.1 Annotate the non-discriminating test, de-couple the ComboboxInput comment from WMS, file the CrudForm accessibility issue
+
+### Phase 7: Validation and delivery
+
+- [ ] 7.1 Run the full validation gate, push, update the PR body and labels, post the summary comment
--- a/.ai/runs/2026-08-17-issue-5239-wms-zone-dialog.md
+++ b/.ai/runs/2026-08-17-issue-5239-wms-zone-dialog.md
@@ -53,29 +53,35 @@ Make PR #5334 mergeable: deliver requirement 2 of #5239 (the zone dialog's custo
 
 ### Phase 2: B1 — make the zone custom fields readable
 
-- [ ] 2.1 Emit CRUD side effects with a warehouse-zone indexer from the zone create/update/delete commands and their undo handlers
+- [x] 2.1 Emit CRUD side effects with a warehouse-zone indexer from the zone create/update/delete commands and their undo handlers — 7fbae24
 
 ### Phase 3: B2 — unit coverage for the server-side custom-field change
 
-- [ ] 3.1 Add the zone custom-field command suite (create with/without values, update, both undo resets, snapshot loader)
+- [x] 3.1 Add the zone custom-field command suite (create with/without values, update, both undo resets, snapshot loader) — 7fbae24
 
 ### Phase 4: M1 and M2 — ComboboxInput
 
-- [ ] 4.1 Never sync a focused field to a self-mapping placeholder label
-- [ ] 4.2 Add the discriminating pick-then-stale-reload test plus the three blast-radius cases
+- [x] 4.1 Never sync a focused field to a self-mapping placeholder label — e35ad73
+- [x] 4.2 Add the discriminating pick-then-stale-reload test plus the three blast-radius cases — e35ad73
 
 ### Phase 5: Minor findings m1–m7
 
-- [ ] 5.1 Give the zone edit dialog optimistic locking (`id` + `updatedAt` in initialValues)
-- [ ] 5.2 Restore custom fields on delete-undo and stop discarding the snapshot
-- [ ] 5.3 Opt the new zones decorator into `stripPrefixedKeys` and type the `customFields` OpenAPI entry
-- [ ] 5.4 Feed the warehouse combobox from the cached query instead of a second round trip
-- [ ] 5.5 Narrow the `ZoneRow` / `ZoneFormValues` index signatures and short-circuit the custom-field snapshot load
+- [x] 5.1 Give the zone edit dialog optimistic locking (`id` + `updatedAt` in initialValues) — 656483e
+- [x] 5.2 Restore custom fields on delete-undo and stop discarding the snapshot — 7fbae24
+- [x] 5.3 Opt the new zones decorator into `stripPrefixedKeys` and type the `customFields` OpenAPI entry — 656483e
+- [x] 5.4 Feed the warehouse combobox from the cached query instead of a second round trip — 656483e
+- [x] 5.5 Narrow the `ZoneRow` / `ZoneFormValues` index signatures and short-circuit the custom-field snapshot load — 656483e
 
 ### Phase 6: Nits and follow-ups
 
-- [ ] 6.1 Annotate the non-discriminating test, de-couple the ComboboxInput comment from WMS, file the CrudForm accessibility issue
+- [x] 6.1 Annotate the non-discriminating test, de-couple the ComboboxInput comment from WMS, file the CrudForm accessibility issue (#5360) — e35ad73, 656483e
 
 ### Phase 7: Validation and delivery
 
-- [ ] 7.1 Run the full validation gate, push, update the PR body and labels, post the summary comment
+- [x] 7.1 Run the full validation gate, push, update the PR body and labels, post the summary comment — 656483e
+
+## Outcome notes
+
+- **m7 was answered with a measurement, not a short-circuit.** The review estimated four extra queries per zone update; `loadCustomFieldValues` returns early on an empty `custom_field_values` result and never issues the `custom_field_defs` query, so a tenant with no zone custom fields pays one lookup per snapshot, not two. Documented at the call site instead of adding a redundant guard.
+- **The row-action zone delete still has no optimistic locking.** m1 scoped the fix to the dialog, and half-wiring a row action — a header without the `surfaceRecordConflict` bar behind it — would be worse than the existing gap. Left for a follow-up that does both.
+- **B1 remains unproven end to end** until `TC-WMS-028-zone-dialog.spec.ts` runs against a live environment. The mechanism is traceable in code and the command-side contract is now pinned by unit tests.
--- a/packages/core/src/modules/wms/__integration__/TC-WMS-028-zone-dialog.spec.ts
+++ b/packages/core/src/modules/wms/__integration__/TC-WMS-028-zone-dialog.spec.ts
@@ -1,0 +1,277 @@
+import { randomUUID } from 'node:crypto'
+import { expect, test, type APIRequestContext, type Locator, type Page } from '@playwright/test'
+import { getAuthToken } from '@open-mercato/core/helpers/integration/api'
+import {
+  apiRequestWithSelectedOrg,
+  createOrganizationFixture,
+  createUserFixture,
+  deleteOrganizationIfExists,
+  deleteUserIfExists,
+  setUserAclVisibility,
+} from '@open-mercato/core/helpers/integration/authFixtures'
+import { getCustomFieldValue } from '@open-mercato/core/helpers/integration/crudFormFields'
+import {
+  expectId,
+  getTokenScope,
+  readJsonSafe,
+} from '@open-mercato/core/helpers/integration/generalFixtures'
+
+export const integrationMeta = {
+  dependsOnModules: ['wms'],
+}
+
+const DEFINITIONS_PATH = '/api/entities/definitions'
+const ZONE_ENTITY_ID = 'wms:warehouse_zone'
+
+type ZoneListResponse = { items?: Array<Record<string, unknown>> }
+
+/**
+ * "The tenant has exactly one warehouse" is the precondition under test, so it must be
+ * established rather than inherited: a throwaway organization holds the single warehouse,
+ * and a dedicated user whose ACL visibility is restricted to that organization drives the
+ * browser. Forcing `om_selected_org` onto an organization the signed-in user cannot see is
+ * silently ignored, which would leave the session in the seeded organization instead.
+ */
+async function createScopedFixture(
+  request: APIRequestContext,
+  token: string,
+  selectedOrgId: string,
+  path: string,
+  data: Record<string, unknown>,
+): Promise<string> {
+  const response = await apiRequestWithSelectedOrg(request, 'POST', path, {
+    token,
+    selectedOrgId,
+    data,
+  })
+  expect(response.ok(), `Failed POST ${path}: ${response.status()}`).toBeTruthy()
+  const body = await readJsonSafe<{ id?: string }>(response)
+  return expectId(body?.id, `Missing id in ${path} create response`)
+}
+
+async function deleteScopedFixture(
+  request: APIRequestContext,
+  token: string | null,
+  selectedOrgId: string | null,
+  path: string,
+  id: string | null,
+): Promise<void> {
+  if (!token || !selectedOrgId || !id) return
+  await apiRequestWithSelectedOrg(request, 'DELETE', `${path}?id=${encodeURIComponent(id)}`, {
+    token,
+    selectedOrgId,
+  }).catch(() => undefined)
+}
+
+async function readZoneById(
+  request: APIRequestContext,
+  token: string,
+  selectedOrgId: string,
+  zoneId: string,
+): Promise<Record<string, unknown> | null> {
+  const response = await apiRequestWithSelectedOrg(
+    request,
+    'GET',
+    `/api/wms/zones?ids=${encodeURIComponent(zoneId)}&page=1&pageSize=10`,
+    { token, selectedOrgId },
+  )
+  expect(response.ok(), `Failed GET /api/wms/zones: ${response.status()}`).toBeTruthy()
+  const body = await readJsonSafe<ZoneListResponse>(response)
+  return (body?.items ?? []).find((item) => item.id === zoneId) ?? null
+}
+
+/**
+ * CrudForm renders its labels as plain text rather than associating them with the input,
+ * so no field in the dialog exposes an accessible name and `getByLabel` matches nothing.
+ * Target the field container instead, the way the rest of the suite does.
+ */
+function crudField(scope: Locator, fieldId: string): Locator {
+  return scope.locator(`[data-crud-field-id="${fieldId}"] input`).first()
+}
+
+async function loginAsScopedUser(
+  page: Page,
+  email: string,
+  password: string,
+  tenantId: string,
+  organizationId: string,
+): Promise<void> {
+  const form = new URLSearchParams()
+  form.set('email', email)
+  form.set('password', password)
+  const response = await page.request.post('/api/auth/login', {
+    headers: { 'content-type': 'application/x-www-form-urlencoded' },
+    data: form.toString(),
+  })
+  expect(response.ok(), `Failed to login zone dialog test user: ${response.status()}`).toBeTruthy()
+
+  const baseUrl = process.env.BASE_URL || 'http://127.0.0.1:5001'
+  await page.context().addCookies([
+    { name: 'om_demo_notice_ack', value: 'ack', url: baseUrl, sameSite: 'Lax' as const },
+    { name: 'om_cookie_notice_ack', value: 'ack', url: baseUrl, sameSite: 'Lax' as const },
+    { name: 'om_feedback_suppress', value: '1', url: baseUrl, sameSite: 'Lax' as const },
+    { name: 'om_selected_tenant', value: tenantId, url: baseUrl, sameSite: 'Lax' as const },
+    { name: 'om_selected_org', value: organizationId, url: baseUrl, sameSite: 'Lax' as const },
+  ])
+}
+
+test.describe('TC-WMS-028: WMS create zone dialog (#5239)', () => {
+  test('pre-selects the only warehouse and round-trips a custom field', async ({ page, request }) => {
+    test.slow()
+
+    const superadminToken = await getAuthToken(request, 'superadmin')
+    const scope = getTokenScope(superadminToken)
+    const suffix = randomUUID().slice(0, 8)
+    const customFieldKey = `zone_note_${suffix}`
+    const customFieldLabel = `Zone Note ${suffix}`
+    const customFieldValue = `Pick face ${suffix}`
+    const warehouseName = `TC-WMS-028 Warehouse ${suffix}`
+    const zoneCode = `TCW28Z${suffix}`
+    const zoneName = `TC-WMS-028 Zone ${suffix}`
+    const userEmail = `tc-wms-028-${suffix}@example.com`
+    const userPassword = `Zone!${suffix}`
+
+    let organizationId: string | null = null
+    let userId: string | null = null
+    let warehouseId: string | null = null
+    let zoneId: string | null = null
+    let customFieldCreated = false
+    let userToken: string | null = null
+
+    try {
+      organizationId = await createOrganizationFixture(request, superadminToken, {
+        name: `TC-WMS-028 Org ${suffix}`,
+        tenantId: scope.tenantId,
+      })
+
+      userId = await createUserFixture(request, superadminToken, {
+        email: userEmail,
+        password: userPassword,
+        organizationId,
+        roles: ['employee'],
+        name: `QA Zone Dialog ${suffix}`,
+      })
+      await setUserAclVisibility(request, superadminToken, {
+        userId,
+        features: ['wms.*', 'entities.*'],
+        organizations: [organizationId],
+      })
+      userToken = await getAuthToken(request, userEmail, userPassword)
+
+      warehouseId = await createScopedFixture(request, userToken, organizationId, '/api/wms/warehouses', {
+        name: warehouseName,
+        code: `TCW28W${suffix}`,
+        isActive: true,
+      })
+
+      // Definitions are read back as `organizationId IN (auth.orgId, NULL)`, so the
+      // definition has to be created from the same organization the browser session runs
+      // in. Creating it unscoped stores it against the creator's own organization, and
+      // the dialog then renders no custom field at all.
+      const definitionResponse = await apiRequestWithSelectedOrg(request, 'POST', DEFINITIONS_PATH, {
+        token: userToken,
+        selectedOrgId: organizationId,
+        data: {
+          entityId: ZONE_ENTITY_ID,
+          key: customFieldKey,
+          kind: 'text',
+          configJson: { label: customFieldLabel, formEditable: true, listVisible: true },
+        },
+      })
+      expect(
+        definitionResponse.ok(),
+        `Failed to create ${ZONE_ENTITY_ID} custom field: ${definitionResponse.status()}`,
+      ).toBeTruthy()
+      customFieldCreated = true
+
+      await loginAsScopedUser(page, userEmail, userPassword, scope.tenantId, organizationId)
+
+      // The single warehouse must be the only one this session can see, otherwise the
+      // "sole warehouse" precondition of #5239 is not actually under test.
+      const warehousesResponse = await apiRequestWithSelectedOrg(
+        request,
+        'GET',
+        '/api/wms/warehouses?page=1&pageSize=50',
+        { token: userToken, selectedOrgId: organizationId },
+      )
+      expect(warehousesResponse.ok(), 'Failed GET /api/wms/warehouses').toBeTruthy()
+      const warehousesBody = await readJsonSafe<{ items?: Array<Record<string, unknown>> }>(warehousesResponse)
+      expect(
+        (warehousesBody?.items ?? []).length,
+        'Scoped organization should contain exactly one warehouse',
+      ).toBe(1)
+
+      await page.goto('/backend/wms/zones')
+
+      const addZoneButton = page.getByRole('button', { name: /Add zone/i }).first()
+      await expect(addZoneButton).toBeVisible({ timeout: 15_000 })
+      await addZoneButton.click()
+
+      const dialog = page.getByRole('dialog').filter({ hasText: /Create zone/i }).first()
+      await expect(dialog).toBeVisible()
+
+      // Requirement 1 (#5239): the single warehouse in scope is pre-selected, so the
+      // operator never has to search for the only choice the form can accept.
+      const warehouseInput = crudField(dialog, 'warehouseId')
+      await expect(warehouseInput).toHaveValue(warehouseName, { timeout: 15_000 })
+
+      // Requirement 2 (#5239): the CrudForm renders the entity's custom fields.
+      const customFieldInput = crudField(dialog, `cf_${customFieldKey}`)
+      await expect(customFieldInput).toBeVisible({ timeout: 15_000 })
+
+      await crudField(dialog, 'code').fill(zoneCode)
+      await crudField(dialog, 'name').fill(zoneName)
+      await customFieldInput.fill(customFieldValue)
+      await dialog.getByRole('button', { name: /Save/i }).click()
+      await expect(dialog).toHaveCount(0, { timeout: 15_000 })
+
+      const zoneRow = page.getByRole('row').filter({ hasText: zoneCode }).first()
+      await expect(zoneRow).toBeVisible({ timeout: 15_000 })
+
+      const listResponse = await apiRequestWithSelectedOrg(
+        request,
+        'GET',
+        `/api/wms/zones?search=${encodeURIComponent(zoneCode)}&page=1&pageSize=10`,
+        { token: userToken, selectedOrgId: organizationId },
+      )
+      expect(listResponse.ok(), `Failed GET /api/wms/zones: ${listResponse.status()}`).toBeTruthy()
+      const listBody = await readJsonSafe<ZoneListResponse>(listResponse)
+      const createdZone = (listBody?.items ?? []).find((item) => item.code === zoneCode)
+      expect(createdZone, `Zone ${zoneCode} should exist after saving the dialog`).toBeTruthy()
+      zoneId = typeof createdZone?.id === 'string' ? createdZone.id : null
+      expect(zoneId, 'Created zone should expose an id').toBeTruthy()
+      expect(createdZone?.warehouse_id).toBe(warehouseId)
+
+      // The custom field value entered in the dialog survives the write path.
+      const persisted = await readZoneById(request, userToken, organizationId, zoneId!)
+      expect(persisted, 'Created zone should be readable by id').toBeTruthy()
+      expect(getCustomFieldValue(persisted!, customFieldKey)).toBe(customFieldValue)
+
+      // ...and is hydrated back into the edit dialog rather than coming up blank.
+      await page.reload()
+      const reloadedRow = page.getByRole('row').filter({ hasText: zoneCode }).first()
+      await expect(reloadedRow).toBeVisible({ timeout: 15_000 })
+      await reloadedRow.getByRole('button').last().click()
+      await page.getByRole('menuitem', { name: /Edit/i }).first().click()
+
+      const editDialog = page.getByRole('dialog').filter({ hasText: /Edit zone/i }).first()
+      await expect(editDialog).toBeVisible({ timeout: 15_000 })
+      await expect(crudField(editDialog, `cf_${customFieldKey}`)).toHaveValue(customFieldValue, {
+        timeout: 15_000,
+      })
+      await expect(crudField(editDialog, 'warehouseId')).toHaveValue(warehouseName)
+    } finally {
+      await deleteScopedFixture(request, userToken, organizationId, '/api/wms/zones', zoneId)
+      await deleteScopedFixture(request, userToken, organizationId, '/api/wms/warehouses', warehouseId)
+      if (customFieldCreated && userToken && organizationId) {
+        await apiRequestWithSelectedOrg(request, 'DELETE', DEFINITIONS_PATH, {
+          token: userToken,
+          selectedOrgId: organizationId,
+          data: { entityId: ZONE_ENTITY_ID, key: customFieldKey },
+        }).catch(() => undefined)
+      }
+      await deleteUserIfExists(request, superadminToken, userId)
+      await deleteOrganizationIfExists(request, superadminToken, organizationId)
+    }
+  })
+})
--- a/packages/core/src/modules/wms/__integration__/TC-WMS-028-zone-dialog.spec.ts
+++ b/packages/core/src/modules/wms/__integration__/TC-WMS-028-zone-dialog.spec.ts
@@ -84,6 +84,10 @@ async function readZoneById(
  * CrudForm renders its labels as plain text rather than associating them with the input,
  * so no field in the dialog exposes an accessible name and `getByLabel` matches nothing.
  * Target the field container instead, the way the rest of the suite does.
+ *
+ * Tracked as a framework-level defect in
+ * https://github.com/open-mercato/open-mercato/issues/5360 — switch this helper to
+ * `getByLabel` once the label/control association lands.
  */
 function crudField(scope: Locator, fieldId: string): Locator {
   return scope.locator(`[data-crud-field-id="${fieldId}"] input`).first()
--- a/packages/core/src/modules/wms/api/zones/route.ts
+++ b/packages/core/src/modules/wms/api/zones/route.ts
@@ -79,7 +79,10 @@ const crud = makeCrudRoute({
       }
       return filters
     },
-    decorateCustomFields: { entityIds: [E.wms.warehouse_zone] },
+    // Zones had no decorator before this route gained custom fields, so there is no
+    // legacy consumer reading top-level `cf_*`/`cf:*` — take the canonical single-shape
+    // response (#1769) rather than emitting both forms.
+    decorateCustomFields: { entityIds: [E.wms.warehouse_zone], stripPrefixedKeys: true },
   },
   hooks: {
     afterList: async (payload, ctx) => {
@@ -123,6 +126,16 @@ export const POST = crud.POST
 export const PUT = crud.PUT
 export const DELETE = crud.DELETE
 
+// Mirrors `CustomFieldDisplayEntry` from @open-mercato/shared/lib/crud/custom-fields, so
+// the generated OpenAPI documents how to read an entry instead of an opaque record.
+const customFieldDisplayEntrySchema = z.object({
+  key: z.string(),
+  label: z.string().nullable(),
+  value: z.unknown(),
+  kind: z.string().nullable(),
+  multi: z.boolean(),
+})
+
 const zoneListItemSchema = z.object({
   id: z.string().uuid().nullable().optional(),
   organization_id: z.string().uuid().nullable().optional(),
@@ -136,7 +149,7 @@ const zoneListItemSchema = z.object({
   created_at: z.string().nullable().optional(),
   updated_at: z.string().nullable().optional(),
   customValues: z.record(z.string(), z.unknown()).nullable().optional(),
-  customFields: z.array(z.record(z.string(), z.unknown())).optional(),
+  customFields: z.array(customFieldDisplayEntrySchema).optional(),
 })
 
 export const openApi = createWmsCrudOpenApi({
--- a/packages/core/src/modules/wms/api/zones/route.ts
+++ b/packages/core/src/modules/wms/api/zones/route.ts
@@ -79,6 +79,7 @@ const crud = makeCrudRoute({
       }
       return filters
     },
+    decorateCustomFields: { entityIds: [E.wms.warehouse_zone] },
   },
   hooks: {
     afterList: async (payload, ctx) => {
@@ -134,6 +135,8 @@ const zoneListItemSchema = z.object({
   priority: z.number().nullable().optional(),
   created_at: z.string().nullable().optional(),
   updated_at: z.string().nullable().optional(),
+  customValues: z.record(z.string(), z.unknown()).nullable().optional(),
+  customFields: z.array(z.record(z.string(), z.unknown())).optional(),
 })
 
 export const openApi = createWmsCrudOpenApi({
--- a/packages/core/src/modules/wms/commands/__tests__/configuration.zoneCustomFields.test.ts
+++ b/packages/core/src/modules/wms/commands/__tests__/configuration.zoneCustomFields.test.ts
@@ -1,0 +1,414 @@
+/** @jest-environment node */
+
+import { commandRegistry } from '@open-mercato/shared/lib/commands/registry'
+
+jest.mock('@open-mercato/shared/lib/i18n/server', () => ({
+  resolveTranslations: async () => ({
+    locale: 'en',
+    dict: {},
+    t: (key: string) => key,
+    translate: (_key: string, fallback?: string) => fallback ?? _key,
+  }),
+}))
+
+jest.mock('../../events', () => ({
+  emitWmsEvent: jest.fn().mockResolvedValue(undefined),
+}))
+
+jest.mock('@open-mercato/shared/lib/encryption/find', () => ({
+  findOneWithDecryption: (emInstance: { findOne: (...args: unknown[]) => unknown }, entity: unknown, filters: unknown) =>
+    emInstance.findOne(entity, filters),
+  findWithDecryption: (emInstance: { find: (...args: unknown[]) => unknown }, entity: unknown, filters: unknown) =>
+    emInstance.find(entity, filters),
+}))
+
+const TENANT = '11111111-1111-4111-8111-111111111111'
+const ORG = '22222222-2222-4222-8222-222222222222'
+const WAREHOUSE_ID = '33333333-3333-4333-8333-333333333333'
+const ZONE_ID = '44444444-4444-4444-8444-444444444444'
+const ZONE_ENTITY_ID = 'wms:warehouse_zone'
+
+type ZoneRecord = {
+  id: string
+  organizationId: string
+  tenantId: string
+  warehouse: { id: string }
+  code: string
+  name: string
+  priority: number
+  metadata: unknown
+  deletedAt: Date | null
+  createdAt: Date
+  updatedAt: Date
+}
+
+type CustomFieldValueRow = {
+  recordId: string
+  fieldKey: string
+  organizationId: string | null
+  tenantId: string | null
+  valueText: string | null
+  valueMultiline: string | null
+  valueInt: number | null
+  valueFloat: number | null
+  valueBool: boolean | null
+}
+
+type SetCustomFieldsCall = {
+  entityId: string
+  recordId: string
+  organizationId: string | null
+  tenantId: string | null
+  values: Record<string, unknown>
+  notify?: boolean
+}
+
+type IndexChange = {
+  action: string
+  identifiers: { id: string; organizationId: string | null; tenantId: string | null }
+  indexerEntityType?: string
+}
+
+function textValueRow(fieldKey: string, value: string): CustomFieldValueRow {
+  return {
+    recordId: ZONE_ID,
+    fieldKey,
+    organizationId: ORG,
+    tenantId: TENANT,
+    valueText: value,
+    valueMultiline: null,
+    valueInt: null,
+    valueFloat: null,
+    valueBool: null,
+  }
+}
+
+function zoneRecord(overrides: Partial<ZoneRecord> = {}): ZoneRecord {
+  return {
+    id: ZONE_ID,
+    organizationId: ORG,
+    tenantId: TENANT,
+    warehouse: { id: WAREHOUSE_ID },
+    code: 'PICK',
+    name: 'Pick face',
+    priority: 5,
+    metadata: null,
+    deletedAt: null,
+    createdAt: new Date('2026-04-15T00:00:00.000Z'),
+    updatedAt: new Date('2026-04-16T00:00:00.000Z'),
+    ...overrides,
+  }
+}
+
+function zoneSnapshot(custom?: Record<string, unknown>) {
+  return {
+    id: ZONE_ID,
+    organizationId: ORG,
+    tenantId: TENANT,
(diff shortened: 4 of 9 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T08:00:21Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: feat(wms): preselect the sole warehouse and support custom fields in the zone dialog (#5239)

## 🎯 Summary

This PR takes both requirements of #5239 — preselect the warehouse when the tenant has exactly one, and make the zone dialog carry custom fields — and implements them across five files plus a new integration spec. The direction is right and the diagnosis behind each piece is right: the `seedOptions` hydration of the warehouse combobox is exactly what `.ai/lessons/async-edit-selects-must-be-hydrated-as-value-plus.md` prescribes ("hydrate selects with both the saved scalar value and a matching option label"); `setCustomFieldsIfAny` is correctly left at its `notify: false` default, which is what `.ai/lessons/crud-owned-custom-field-writes-should-not-emit-a-second.md` requires of a CRUD-owned custom-field write; `E.wms.warehouse_zone` is the canonical generated id rather than an ad-hoc alias; and `updateWarehouseZoneCommand.undo` uses the documented `buildCustomFieldResetMap(before.custom, after.custom)` shape. I also verified the author's claim about the two remaining ESLint warnings in `ComboboxInput.tsx` — they do reproduce byte-for-byte on `develop` (same two rules, only the line numbers shift), so they are not this PR's.

The PR is explicitly self-marked **DO NOT MERGE** because requirement 2 does not persist, and the author states that pinning the failure needs a live environment. I spent the review narrowing that, and the useful news is that **the wiring this PR adds is provably correct on both ends** — I have runnable evidence for the client half and the server half separately (details under ⛔ Blocker). The bad news is that while doing so I found a **reproduced regression in `ComboboxInput`**, the shared primitive this PR changes: after picking an option from an async dropdown, a still-focused field can have its label overwritten with the raw record id. I have a discriminating test for that too — it passes on `develop` and fails on this head with nothing but `ComboboxInput.tsx` swapped.

Reviewed scope: the six changed files, plus the surrounding contracts they depend on — `packages/shared/src/lib/api/scoped.ts`, `packages/shared/src/lib/commands/helpers.ts`, `packages/shared/src/lib/commands/customFieldSnapshots.ts`, `packages/shared/src/lib/crud/{factory,custom-fields}.ts`, `packages/shared/src/lib/query/engine.ts`, `packages/core/src/modules/entities/lib/helpers.ts`, and `packages/ui/src/backend/CrudForm.tsx`.

**Runner: local** (no compose `app` container was running; `docker compose -f docker-compose.fullapp.dev.yml ps --status running -q app` returned nothing).
**Validation source: local fallback.** No CI has ever run on this fork head — the only reported checks are `license/cla` (pass) and `[code]smith` (skipping) — and `develop`'s branch protection declares **no** `required_status_checks` at all (it requires 1 approving review, which is why `mergeStateStatus` is `BLOCKED` while `mergeable` is `MERGEABLE`). With no usable GitHub check data for `1ac4915`, this repository's GitHub-checks-first rule falls through to the local gate, which I ran in full.

## Verdict

❌ **request changes** — two blockers and two majors. The PR does not deliver requirement 2 and says so itself, and separately it introduces a reproduced label-corruption regression into `ComboboxInput`, a primitive behind every combobox in the product, with no test covering the failing path. The entire server-side custom-field change also ships without a single unit test, which is both a checklist violation and the direct reason the author could not diagnose the failure from the code.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | 22/22 packages built. |
| `yarn generate` | ✅ PASS | No generated-file churn afterwards (`git status --porcelain` empty), so nothing needs regenerating and committing. |
| `yarn build:packages` (rebuild after generate) | ✅ PASS | |
| `yarn i18n:check-sync` | ✅ PASS | |
| `yarn i18n:check-usage` | ✅ PASS | |
| `yarn typecheck` | ✅ PASS | Clean, matching the author's 22/22 report. |
| `yarn test` | ⚠️ PASS except 4 host-limited cases | Every workspace green — `@open-mercato/core` **1246 suites / 9727 tests**, `@open-mercato/ui` **229 suites / 1863 tests**, `cli` 1703, `enterprise` 478, `webhooks` 145, `checkout` 159, `storage-s3` 128, `channel-imap` 109, `onboarding` 115, `sync-akeneo` 62. The only failures are 4 cases in the untouched `packages/create-app` (`src/lib/template-dev-log-files.test.ts`), all four asserting on this host's `fs.inotify.max_user_instances=128` (needs 4096) — a machine limitation, not a defect in this branch. Note that `packages/ui` was fully green here, including `utils/__tests__/format.test.ts`, which the PR body reports as failing; that one looks environment-specific rather than branch-specific. |
| `npx eslint` on the 5 changed source files | ✅ 0 errors | 2 warnings, both in `ComboboxInput.tsx` (unused `eslint-disable` directive; `resolveDescription` missing from a dep array). **Verified pre-existing**: I replaced only that file with `origin/develop`'s copy and re-ran — the identical two warnings appear at lines 245/247 instead of 249/251. Not this PR's. |

The gate being green is itself a finding, and it belongs in the Test Coverage section: 12,000+ passing tests are blind both to the requirement the PR admits is broken and to the `ComboboxInput` regression below.

## Findings

### ⛔ Blocker

**B1 — `packages/core/src/modules/wms/components/backend/WmsConfigurationPage.tsx` / `packages/core/src/modules/wms/commands/configuration.ts`: requirement 2 does not persist, and the PR says so.** The author's own `TC-WMS-028-zone-dialog.spec.ts:248` fails deterministically (`Expected: "Pick face f651dc6b"`, `Received: null`), the PR body carries a DO-NOT-MERGE banner, and the linked issue #5239 asks for exactly this capability. A PR that does not do what it claims cannot merge, so this alone forces the verdict. What I can add is a much smaller search space than "the wiring looks correct at every individual link":

1. **The read path is not the primary suspect, and the `null` proves it.** `getCustomFieldValue` (`packages/core/src/helpers/integration/crudFormFields.ts:27`) returns `undefined`, not `null`, when the key is absent from `customValues` / `cf_*` / `cf:*` / the `customFields` array. And `decorateRecordWithCustomFields` (`packages/shared/src/lib/crud/custom-fields.ts:480-483`) reads **only keys already present on the row** and returns `{customValues: null, customFields: []}` when there are none — which would have surfaced as `undefined`. Getting `null` therefore means the list SQL *did* project `cf:zone_note_…` and the definition *did* resolve for the row's org/tenant, and the projected value was SQL NULL. The projection is a plain `LEFT JOIN custom_field_values` (`packages/shared/src/lib/query/engine.ts:840-859`), not an `entity_indexes` read, so there is no reindex lag to blame: **there was no matching row in `custom_field_values`.**
2. **The client half is correct — measured, not read.** I rendered the real `CrudForm` with the real `zoneFormSchema`, a mocked `fetchCustomFieldFormStructure` returning one `text` definition, filled the `cf_zone_note` input and submitted. `onSubmit` received `{"warehouseId":"3333…","code":"Z1","name":"Zone 1","cf_zone_note":"Pick face"}` and `collectCustomFieldValues(...)` on it returned `{"zone_note":"Pick face"}`. The control run matters: with `.passthrough()` removed from `zoneFormSchema` and nothing else changed, `onSubmit` received no `cf_*` key at all and `collectCustomFieldValues` returned `{}`. So the `.passthrough()` diagnosis in the PR body is **correct and necessary**, and it discriminates.
3. **The server half is correct — also measured.** I fed a realistic POST body (`{warehouseId, code, name, priority, customFields:{zone_note:'Pick face'}}`) through the real `parseScopedCommandInput(warehouseZoneCreateSchema, …)` that `api/zones/route.ts:95` calls, which produced `{organizationId, tenantId, warehouseId, code, name, priority, customFields:{zone_note:"Pick face"}}`, then invoked the real registered `wms.zones.create` handler with a stub `em`/container. It called `dataEngine.setCustomFields({entityId:"wms:warehouse_zone", recordId:"zone-1", tenantId, organizationId, values:{zone_note:"Pick face"}, notify:false})` — right entity id (identical to the spec's `ZONE_ENTITY_ID`), right record, right scope, right value, and `notify:false` per the lesson.

So the break is **downstream of `dataEngine.setCustomFields`, or environmental** — not in the route/command/schema wiring the PR body is auditing. The next place to look is `setRecordCustomFields` (`packages/core/src/modules/entities/lib/helpers.ts:64-240`): whether its `defsByKey` lookup (`:75-81`, scoped `organizationId: {$in: [organizationId, null]}`, `tenantId: {$in: [tenantId, null]}`) resolves `zone_note_…` at all, which column `columnFromKind`/`columnFromJsValue` selects, and whether the row it persists at `:216-217` really commits under the transaction it opens at `:143`. Note the two `EntityManager`s in play: the command writes the zone on `resolveEm(ctx)`, a **fork** (`commands/configuration.ts:92-94`), while `setCustomFields` writes through the DI `DataEngine`'s own em — worth confirming on a live run that those commit independently as expected. Please also land B2's unit tests as part of this, so the next iteration is provable without an ephemeral environment.

**B2 — `packages/core/src/modules/wms/commands/configuration.ts`: the entire server-side custom-field change ships with zero unit tests.** The diff adds 49 lines across `wms.zones.create` (`:1038`, `:1054`), `wms.zones.update` (`:1117`, `:1133`), the create-undo reset (`:1096`), the update-undo reset (`:1170`) and the snapshot loader (`:335-346`), and `api/zones/route.ts` gains `decorateCustomFields` — and not one of those lines is executed by a test in this PR. The only new unit tests are the two `ComboboxInput` cases. `.ai/review-checklist.md` §6 requires "Custom field snapshots captured in `snapshot.custom`" and "Undo restores via `buildCustomFieldResetMap(before.custom, after.custom)`" to be verified, and the built-in checklist's Testing rule is unconditional: behavior changes must ship coverage. The module already has five command suites to pattern-match — `commands/__tests__/configuration.undo.test.ts` is the closest template and already stubs `resolveTranslations`, `../../events` and the decryption helpers. Cases to add there:

- create with `customFields` present → `dataEngine.setCustomFields` called once with `{entityId: E.wms.warehouse_zone, recordId, values, notify: false}`;
- create with no `customFields` → `setCustomFields` **not** called (the `setCustomFieldsIfAny` early return);
- update with `customFields` → same assertion, and update without them leaves existing values untouched;
- create-undo → the reset map nulls exactly the keys `after.custom` held;
- update-undo → `buildCustomFieldResetMap(before.custom, after.custom)` restores before-values and nulls after-only keys;
- `loadWarehouseZoneSnapshot` → `snapshot.custom` is populated with bare (un-prefixed) keys.

A single one of these — the first — takes about 40 lines against a stub container and would have told the author within a minute that the server chain is fine, redirecting B1's investigation. That is why this is a blocker rather than a major here.

### ⚠️ Major

**M1 — `packages/ui/src/backend/inputs/ComboboxInput.tsx:253-262`: the new sync guard lets a focused combobox overwrite a just-picked label with the raw record id.** This is a reproduced regression in a primitive the PR itself flags as reaching "every combobox in the product".

The mechanism: `selectValue` sets `userTypedRef.current = false` (`:274`) while the input keeps focus, so the *next* `[value, optionMap]` change passes the new guard (`document.activeElement === inputRef.current && userTypedRef.current` is false) and runs `setInput(option?.label ?? value ?? '')`. Meanwhile `optionMap` has a self-mapping fallback (`:126-148`, the `map.set(value, …)` at `:141`) that inserts `{value, label: value}` whenever the current `value` is missing from the option sources, and `asyncOptions` is **replaced**, not merged, on every load (`:193`). So if the option list loses the picked entry, `optionMap.get(value).label === value` and the effect writes the UUID into the field the user is looking at. The old guard (`document.activeElement !== inputRef.current`) made this unreachable while focused.

Reaching it does not require anything exotic, because picking an option rewrites `input` to the label, which re-fires the debounced loader **with the label as the query** (`:179-198`). Any follow-up load that comes back with fewer options loses the label: a failed request (every loader in this repo does `if (!call.ok) return []` — e.g. `wmsLookupLoaders.ts:47`), a debounce race, or simply a label the route's own `?search=` cannot match, which is normal for composite labels like `formatCatalogVariantLabel`'s `"${name} (${sku})"` (`wms/lib/inventoryDisplayUi.ts:222-233`).

Reproduced with a discriminating test — a `ComboboxInput` whose `loadSuggestions` answers `"WH-B"` with `[{value:'wh-b', label:'Warehouse B (WH-B)'}]` and returns `[]` for anything else. Same test file, only `ComboboxInput.tsx` swapped between this head and `origin/develop`:

| | right after picking | after the follow-up load, still focused |
|---|---|---|
| `origin/develop` | `"Warehouse B (WH-B)"` | `"Warehouse B (WH-B)"` ✅ |
| this head (`1ac4915`) | `"Warehouse B (WH-B)"` | **`"wh-b"`** ❌ |

(`document.activeElement === input` was `true` in both runs; loader calls were `[["WH-B"], ["Warehouse B (WH-B)"], []]`.)

Suggested fix — never sync a focused field to a self-mapping placeholder, which keeps the #5239 fix (an *arriving* real label still syncs) and closes this hole:

```ts
React.useEffect(() => {
  const option = optionMap.get(value)
  const hasRealLabel = Boolean(option && option.label !== option.value)
  if (document.activeElement === inputRef.current && (userTypedRef.current || !hasRealLabel)) return
  setInput(option?.label ?? value ?? '')
}, [value, optionMap])
```

Please ship the case above as a third test in `__tests__/ComboboxInput.test.tsx` — like the first new test, it discriminates between the two implementations, which is what makes it worth having.

**M2 — `packages/ui/src/backend/inputs/ComboboxInput.tsx` is changed without any statement of the blast radius beyond prose.** The PR is candid that this touches a shared primitive, and the two new tests are a good start, but `seedOptions` + `autoFocus` + async `loadOptions` is only one of the shapes affected. The guard now syncs strictly *more often* than before — every focused-but-not-typed state is newly syncable, which includes the whole window between `selectValue` and the next blur, and the window after `handleClear` (`:322-326`, which calls `selectValue('')` and then re-focuses). M1 is one exploit of that window; the fix above closes the placeholder case, but the review still needs coverage for the others. Concretely, add cases for: picking an option and then having a *real* new label arrive (should sync), clearing via the ✕ while the parent keeps the old `value` (should not resurrect the old label), and a blur→refocus with no typing (should sync to the canonical label).

### 🔹 Minor

**m1 — `packages/core/src/modules/wms/components/backend/WmsConfigurationPage.tsx:671-679`: the zone edit dialog still has no optimistic locking, and this hunk is where the fix belongs.** The edit-mode `initialValues` carries neither `id` nor `updatedAt`, so `CrudForm` never recognises the form as an edit: it cannot auto-derive `x-om-ext-optimistic-lock-expected-updated-at` (`packages/ui/AGENTS.md`: "edit-mode `initialValues` MUST include `updatedAt`"), and `initialValuesHasId` stays `false`, which also makes the custom-field defaults effect (`CrudForm.tsx:2686-2725`) treat an edit as a create. `WarehouseZone` does have the column (`wms/data/entities.ts:46`) and the list route already selects `updated_at`, so server-side locking is armed and silently no-ops on this form — precisely the failure mode `packages/core/src/__tests__/optimistic-lock-editable-entities.test.ts` was written to prevent (its `moduleEntities` map has no `wms` entry, which is why the guard does not catch this). This predates the PR, but the diff rewrites this exact object and adds more user-entered data to the form, so the two-line fix (`id: dialog.row.id, updatedAt: dialog.row.updated_at`) is cheapest here. Root `AGENTS.md` treats optimistic locking as default-ON for user-editable entities.

**m2 — `packages/core/src/modules/wms/commands/configuration.ts:1182-1212`: `wms.zones.delete` handles custom fields nowhere, contradicting the PR description.** The PR body and the commit message both say the commands round-trip custom fields "including undo snapshots and reset-on-delete", but the delete command's `execute` does not reset them and its `undo` does not restore them. The reference implementation does: `deleteCompanyCommand.undo` calls `buildCustomFieldResetMap(before.custom, undefined)` (`customers/commands/companies.ts:1494-1497`), and `.ai/review-checklist.md` §6 lists that restore as a requirement. `prepare` (`:1185-1190`) now pays for the extra custom-field snapshot load via `loadWarehouseZoneSnapshot` and then discards it, so the cost is already being incurred. Either wire the delete-undo restore or drop the "reset-on-delete" claim from the description — the claim as written sends a reviewer looking for code that is not there.

**m3 — `packages/core/src/modules/wms/api/zones/route.ts:82`: a brand-new decorator should opt into `stripPrefixedKeys: true`.** The option's own documentation is explicit that the `false` default exists only for backward compatibility: "Recommended for new modules — produces the single canonical response shape requested in #1769. Defaults to `false` so existing callers that read `cf_*` from the top level keep working until they migrate" (`packages/shared/src/lib/crud/factory.ts:189-196`). Zones had no decorator before this PR, so there is no legacy consumer to protect, and as written the response now carries both the raw `cf:*` keys and `customValues`/`customFields` — while the `zoneListItemSchema` you added documents only the latter two, so the generated OpenAPI and the actual payload disagree.

**m4 — `packages/core/src/modules/wms/api/zones/route.ts:139`: `customFields: z.array(z.record(z.string(), z.unknown()))` erases the shape consumers need.** The decorator emits `CustomFieldDisplayEntry` objects with `key`, `label`, `value`, `kind` and `multi` (`packages/shared/src/lib/crud/custom-fields.ts:498-504`). Declaring the array as records of unknown means the generated OpenAPI tells an API consumer nothing about how to read an entry. Prefer an explicit object schema (or the shared type's schema if one exists) so the documented contract matches what ships.

**m5 — `packages/core/src/modules/wms/components/backend/WmsConfigurationPage.tsx:605-618`: the dialog now loads the warehouse list twice.** Opening the dialog fires `warehouseOptionsQuery` → `GET /api/wms/warehouses?page=1&pageSize=50`, and the `warehouseId` field still declares `loadOptions: loadWarehouseOptions` (`:636`), which fires the same request again on focus. Since the query already has the full first page in cache, consider feeding the combobox from it (or passing a loader that reads through the same React Query key) instead of paying for two round trips on every dialog open.

**m6 — `packages/core/src/modules/wms/components/backend/WmsConfigurationPage.tsx:76` and `:114`: `& Record<string, unknown>` disables type checking on `ZoneRow` and `ZoneFormValues`.** With that intersection, every property access type-checks (as `unknown`) and excess-property checks disappear. In this specific file that is a live hazard, because it mixes both naming conventions — `dialog.row.warehouse_id` (snake, correct for a list row) next to `values.warehouseId` (camel, correct for form values) — so a mix-up compiles silently. The pattern exists elsewhere in the repo (`planner/components/AvailabilityRuleSetForm.tsx:10-16`), so I am not asking you to invent something new, but an explicit `customValues?: Record<string, unknown> | null` plus an index signature narrowed to `` `cf_${string}` `` would keep the custom-field flexibility without opening the whole object.

**m7 — `packages/core/src/modules/wms/commands/configuration.ts:332-346`: every zone update and delete now costs an extra pair of queries.** `loadWarehouseZoneSnapshot` unconditionally calls `loadCustomFieldSnapshot`, which issues a `custom_field_values` find plus a `custom_field_defs` find, and it runs from `prepare` on update/delete *and* from `captureAfter` on create/update — so a single update is now four extra queries. This is the price of correct undo semantics and I am not asking you to remove it, but it is worth a comment or a short-circuit when the entity has no active definitions, since the common tenant has none.

### 💅 Nit

**n1 — `packages/ui/src/backend/inputs/__tests__/ComboboxInput.test.tsx:410`: the second new test does not discriminate, as the author already disclosed.** Keeping a non-discriminating guard test is a legitimate choice, but say so in the test name or a one-line comment (`// guards the opposite direction; passes on both implementations`) so a future reader does not mistake it for coverage of the bug.

**n2 — `packages/ui/src/backend/inputs/ComboboxInput.tsx:253-257`: a shared primitive's comment names a WMS dialog and issue #5239.** Describing the mechanism ("`autoFocus` focuses the control before an async default value arrives") is genuinely useful here; naming one consumer's screen inside `packages/ui` couples the primitive's documentation to a module. Consider keeping the mechanism and dropping the WMS-specific example.

**n3 — the accessibility gap the spec works around deserves its own issue.** `TC-WMS-028-zone-dialog.spec.ts:81-90` targets `[data-crud-field-id="…"] input` because `CrudForm` renders labels with no `htmlFor` and no `id` on the control, so no field in the dialog exposes an accessible name. That is a real framework-level defect (it also means screen readers get nothing), the workaround is correct for now, and the PR body already suggests filing it — please do, and link it from the spec comment.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — nothing was removed; `ComboboxInput`'s props are untouched and `seedOptions` already existed.
- [x] No function signature changed in a breaking way — `zoneFormSchema` is module-private; the command handlers keep their `CommandHandler` shapes.
- [x] No required type field removed or narrowed — `ZoneRow` / `ZoneFormValues` were widened, not narrowed, and both are module-local.
- [x] No HTTP route URL removed or renamed; no method changed — `/api/wms/zones` keeps GET/POST/PUT/DELETE and its `metadata` guards (`wms.view` / `wms.manage_zones`) unchanged.
- [x] No field removed or retyped in an existing response shape — `customValues` and `customFields` are additive (Cat. 7 additive-only), and the raw `cf:*` keys were already emitted by the query engine's `includeCustomFields: true`.
- [x] No event or message name renamed or removed; no payload field removed — `wms.zone.created` / `wms.zone.updated` payloads are byte-identical.
- [x] No database table or column renamed or removed — no entity change and no migration needed; custom-field values live in the existing `custom_field_values` table.
- [x] Where a contract had to change: not applicable — nothing here required a deprecation window.

No `BACKWARD_COMPATIBILITY.md` protected surface is touched. `apps/mercato/src/app/**` and `apps/mercato/.env.example` are untouched, so the `create-app` template-sync obligation does not apply.

One behavioural caveat that is not a contract break but is the reason for `risk-high`: `ComboboxInput`'s label-sync semantics change for **every** combobox in the product. `seedOptions` is a pre-existing prop and no signature moved, so third-party modules keep compiling — but they inherit the new runtime behaviour, including M1, without opting in.

## 🧪 Test Coverage

**What the tests cover.** The integration spec `TC-WMS-028-zone-dialog.spec.ts` is genuinely well built and I want to say so: it establishes the "exactly one warehouse" precondition rather than inheriting it (throwaway organization, dedicated user with ACL visibility narrowed to it, plus an explicit assertion that the scoped org contains exactly one warehouse), it creates the custom-field definition from the same organization the browser session runs in and documents why, it cleans up every fixture in `finally`, and it asserts both requirements plus the edit-dialog rehydration. It is also the artefact that caught the requirement-2 failure, which is exactly what integration coverage is for. On the unit side, `ComboboxInput.test.tsx`'s first new case is a proper regression test: I confirmed it discriminates, and the author confirmed the same by restoring the old guard.

**Gaps, in priority order.**

1. **The server-side custom-field change has no unit coverage at all** — see B2 for the six cases and the file they belong in (`commands/__tests__/configuration.undo.test.ts`). Every one of the 49 new lines in `commands/configuration.ts` can be deleted today and the whole 12,000-test suite still passes.
2. **M1's failing path is untested**, which is why a green `packages/ui` run (229 suites / 1863 tests) says nothing about it. Add the pick-then-stale-reload case described in M1 to `__tests__/ComboboxInput.test.tsx`; it fails on this head and passes on `develop`, so it is a real regression test rather than a snapshot of current behaviour.
3. **`decorateCustomFields` on the zones list route is untested.** `packages/shared/src/lib/crud/__tests__/crud-factory.test.ts:280` already exercises the decorator generically, so the gap here is the route-level wiring: a test asserting that `GET /api/wms/zones` returns `customValues` keyed by bare field name for a row that has a value, and `null` for one that does not. That test is also the cheap oracle for whether B1 is a write-side or read-side failure.
4. **`m2`'s missing delete-undo restore has no test**, and adding one would have surfaced the gap.
5. **The two new `ZoneRow`/`ZoneFormValues` index signatures remove the compiler as a safety net** (m6), which raises the value of the tests above rather than lowering it.

Nothing in the diff needs new Playwright coverage beyond TC-WMS-028 — once requirement 2 persists, that spec is the right shape and the right assertions.

pkarw:

The PR arrived with no labels at all, because the author does not have label permission on this repository and listed their intended set in the body instead. I applied that set with one deliberate change, explained below.

- ❌ `changes-requested` — the code review found two blockers and two majors: requirement 2 of #5239 does not persist (the PR says so itself and carries a DO-NOT-MERGE banner), the server-side custom-field change ships with no unit tests, and the `ComboboxInput` guard change introduces a reproduced regression where a focused combobox can overwrite a just-picked label with the raw record id. This replaces the implicit "ready for review" state; it will move on once a new head addresses the findings.
- ✨ `feature` — the change implements a requested capability (preselect the sole warehouse, carry custom fields through the zone dialog) rather than repairing existing behaviour, which matches the author's proposal and the `feature` label already on issue #5239.
- 🧪 `needs-qa` — the diff touches `.tsx` files and `packages/ui/src/`, so the automated-verification exemption in `.ai/docs/pr-workflow.md` does not apply; on top of that, the shipped tests do not cover the behaviour that is broken, so a human must exercise the zone dialog and a sample of other comboboxes before this merges. With `qaGate` on, this label keeps the merge blocked until a QA reviewer adds `qa-approved`.
- 🔹 `priority-medium` — I kept the author's proposal: this closes a product gap in the WMS zone dialog with no data-loss, security, or availability angle, and nothing in the review changed that assessment.
- ⚠️ `risk-high` — **raised from the author's proposed `risk-medium`.** The author was right that the WMS changes are contained, but the diff also changes `packages/ui/src/backend/inputs/ComboboxInput.tsx`, a shared primitive behind every combobox in the product, and the review reproduced an actual behavioural regression in it (label replaced by the raw id on a focused field after picking from an async list) with a test that passes on `develop` and fails on this head. A confirmed defect in a cross-cutting shared surface is what the `risk-high` inference rule describes, and it reinforces the `needs-qa` requirement above.

`in-progress` is this run's lock and is released in a separate comment. No `qa` label was set — that one belongs to a QA reviewer when they pick the PR up — and `qa-approved` is never applied from reading a diff.

pkarw:

Two things from the review that should save you time:

**On requirement 2.** You wrote that "the wiring looks correct at every individual link" and that pinning the break needs a live environment. I can confirm the first half with runnable evidence rather than reading: I exercised the client half (real `CrudForm` + your real `zoneFormSchema`, filling `cf_zone_note` and submitting) and the server half (real `parseScopedCommandInput(warehouseZoneCreateSchema, …)` feeding the real registered `wms.zones.create` handler) as two throwaway unit tests, and the value survives end to end — `dataEngine.setCustomFields` is called with `{entityId: "wms:warehouse_zone", recordId, tenantId, organizationId, values: {zone_note: "Pick face"}, notify: false}`. The `.passthrough()` fix is also confirmed **necessary**: with it removed and nothing else changed, `onSubmit` receives no `cf_*` key and `collectCustomFieldValues` returns `{}`.

So the break is downstream of `dataEngine.setCustomFields`, or environmental — not in the route/command/schema chain your table audits. The `Received: null` (rather than `undefined`) narrows the read side out too: `decorateRecordWithCustomFields` only reads keys already on the row and would have produced `undefined` if the key were absent, so the list query *did* project `cf:zone_note_…` and the definition *did* resolve — meaning there was simply no row in `custom_field_values`. The next place to instrument is `setRecordCustomFields` (`packages/core/src/modules/entities/lib/helpers.ts:64`): whether its `defsByKey` lookup resolves your key, which value column it picks, and whether the row it persists really commits. Note the two `EntityManager`s involved — the command writes the zone on a **fork** (`commands/configuration.ts:92`) while `setCustomFields` writes through the DI `DataEngine`'s own em.

**On `ComboboxInput`.** The regression I found (blocker-adjacent major M1) is reproducible without a live environment, and it is worth fixing before anything else because it affects every combobox in the product rather than just WMS. `selectValue` resets `userTypedRef` to `false` while the input keeps focus, so the next `[value, optionMap]` change syncs — and `optionMap`'s self-mapping fallback then supplies `label === value`, putting the raw UUID into the focused field. The review has the exact repro table and a suggested guard.

I deliberately did **not** push any fixes to your fork branch: that is not an unprompted call to make on someone else's contribution. If you'd like the automation to carry the fixes here instead, say so and it can be re-run in autofix mode.

pkarw:

The full code review is posted as a review on this PR: 2 blockers, 2 majors, 7 minors and 3 nits. The verdict is driven by requirement 2 of #5239 not persisting (which the PR itself flags), by the server-side custom-field change shipping with no unit tests, and by a reproduced label-corruption regression in `packages/ui/src/backend/inputs/ComboboxInput.tsx` that a discriminating test shows passing on `develop` and failing on this head. The direction of the change is right and the review says so — the `seedOptions` hydration matches this repo's documented lesson on async edit selects, `setCustomFieldsIfAny` correctly stays at `notify: false`, the canonical generated entity id is used, and the update-undo reset map follows the documented shape.

**Validation source: local fallback** (runner: local). No CI has ever run on this fork head — only `license/cla` (pass) and `[code]smith` (skipping) — and `develop` declares no required status checks, so there was no GitHub check data to prefer. The full configured gate ran green: `build:packages`, `generate` (no churn), `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `test`, `build:app`. The only test failures are 4 cases in the untouched `packages/create-app` caused by this host's `fs.inotify.max_user_instances` limit. ESLint on the changed files: 0 errors, 2 warnings, both verified to reproduce on `develop`. That a fully green 12,000-test gate is blind to both the broken requirement and the primitive regression is itself reported as the main test-coverage gap.

Inherited reviewer feedback: none — the only prior comment on this PR was the Dokploy preview-deployment security notice, which carries no actionable review ask.

The PR is assigned back to @Paul-Mlodochowki, labels are normalized (`changes-requested`, `feature`, `needs-qa`, `priority-medium`, `risk-high` — the last raised from the author's proposed `risk-medium`, with the reasoning in the label-rationale comment), and no `ci-monitoring` is set because there is no CI on this head to follow up on.

Paul-Mlodochowki:
🤖 `om-auto-continue-pr` started by @Paul-Mlodochowki at 2026-08-17T15:34:49Z to work through the code-review findings. Other auto-skills will skip this PR until the lock is released.

This is a fork PR, so the usual lock signals are only partly available: `AddLabelsToLabelable` and `ReplaceActorsForAssignable` are both denied for this account, so the `in-progress` label could not be applied. This comment plus the existing assignee are the lock.

Paul-Mlodochowki:

**Tracking plan:** `.ai/runs/2026-08-17-issue-5239-wms-zone-dialog.md`
**Branch:** `feat/issue-5239-wms-zone-dialog` (fork head, pushed to `Paul-Mlodochowki/open-mercato`)
**Resume point:** Phase 2.1 → Phase 7.1 (all phases complete)
**Final status:** complete
**Commits added by this resume:** `c458106..6d5dbdd` (5 commits on top of `1ac4915`)

This PR carried no execution plan, so it was **adopted**: the plan was reconstructed from the PR description, issue #5239 and the code review, then committed to the branch. The reasoning lives in the plan file's Evidence and Assumptions sections.

- **⛔ B1 — requirement 2 now has a root cause and a fix.** The write chain was never the problem. The zones list is served by `HybridQueryEngine`, which projects `cf:*` out of **`entity_indexes`**, and the zone commands never emitted CRUD side effects — so no `query_index.upsert_one` ever ran for a zone and the value read back `null` however correctly it was stored. Added `warehouseZoneCrudIndexer` and wired `emitCrudSideEffects` / `emitCrudUndoSideEffects` into create, update, delete and all three undo handlers, always *after* the custom-field write. `events` is deliberately omitted so no second, undeclared `wms.warehouse_zone.*` id appears for a write the module already announces as `wms.zone.*`.
- **⛔ B2 — the server-side change now has unit coverage.** New `commands/__tests__/configuration.zoneCustomFields.test.ts`, 10 cases: create with and without values, the index-upsert ordering (`setCustomFields` before `markOrmEntityChange`), update with and without values, all three undo reset maps, the delete index removal, and the snapshot loader's bare-key capture. The index assertions fail without the B1 fix.
- **⚠️ M1 — the reproduced `ComboboxInput` regression is fixed.** The guard now also refuses to sync a focused field to a self-mapping placeholder, so a follow-up load that drops the picked option can no longer paint the raw record id over the label. Verified against the review's exact scenario: the new test received `"wh-b"` on the previous implementation and passes here.
- **⚠️ M2 — blast radius is now covered, not just described.** Three more cases: a real label arriving while focused (must sync), clearing via ✕ while the parent keeps the value (must not resurrect), and blur→refocus resuming the sync. Each was run against `origin/develop` and against the intermediate guard to confirm which ones actually discriminate — the table is in the PR description.
- **🔹 m1** the zone edit dialog gets `id` + `updatedAt` in `initialValues`, so CrudForm derives the optimistic-lock header and stops treating the edit as a create in its custom-field defaults effect. **🔹 m2** delete-undo restores custom fields via `buildCustomFieldResetMap(before.custom, undefined)`, matching `deleteCompanyCommand`. **🔹 m3** the decorator opts into `stripPrefixedKeys: true`. **🔹 m4** `customFields` is documented with an explicit `CustomFieldDisplayEntry`-shaped schema. **🔹 m5** the combobox's unsearched open is served from the cached dialog query. **🔹 m6** the `& Record<string, unknown>` escape hatches are replaced with explicit members plus a `cf_`-scoped index signature.
- **🔹 m7 — answered with a measurement rather than a guard.** The estimate of four extra queries per update is high: `loadCustomFieldValues` returns early on an empty `custom_field_values` result and never issues the `custom_field_defs` query, so a tenant with no zone custom fields pays one lookup per snapshot, not two. Documented at the call site; adding a short-circuit on top would just duplicate that early return.
- **💅 n1** the non-discriminating test says so in a comment. **💅 n2** the shared primitive's comment keeps the mechanism and drops the WMS example. **💅 n3** the CrudForm accessible-name defect is filed as #5360 and linked from the spec's selector workaround.

**Files touched in this resume:** `packages/core/src/modules/wms/commands/{configuration.ts,shared.ts}`, `packages/core/src/modules/wms/commands/__tests__/configuration.zoneCustomFields.test.ts` (new), `packages/core/src/modules/wms/api/zones/route.ts`, `packages/core/src/modules/wms/components/backend/WmsConfigurationPage.tsx`, `packages/core/src/modules/wms/__integration__/TC-WMS-028-zone-dialog.spec.ts`, `packages/ui/src/backend/inputs/ComboboxInput.tsx` and its test, plus the adopted plan.

None. No external skill URLs were recorded in the plan or consulted during this resume.

**Runner:** local. No compose `app` container was running.

- **Targeted validation (per phase):** after the B1/B2 phase, `npx jest --config packages/core/jest.config.cjs --testPathPatterns="modules/wms"` — 39 suites / 246 tests ✅, plus `tsc --noEmit -p packages/core/tsconfig.json` ✅. After the ComboboxInput phase, `npx jest --config packages/ui/jest.config.cjs --testPathPatterns="ComboboxInput"` — 26 tests ✅, run three times: once on this implementation, once with the previous guard restored (1 failure, the M1 case), once with `origin/develop`'s guard restored (3 failures). After the minors phase, core typecheck ✅.
- **Full validation gate:** `yarn build:packages` ✅ 22/22 · `yarn generate` ✅ with no generated-file churn afterwards (`git status --porcelain` clean) · `yarn build:packages` (rebuild) ✅ · `yarn i18n:check-sync` ✅ all 5 locales in sync · `yarn i18n:check-usage` ✅ advisory only · `yarn typecheck` ✅ 22/22 · `yarn test` ⚠️ see below · `yarn build:app` ✅ · `npx eslint` on the 7 changed source files ✅ 0 errors, 2 warnings both pre-existing in `ComboboxInput.tsx`.
- **`yarn test` baseline.** Red on this Windows host, and red in the same places without this branch's work. I re-ran every failing suite with the changes stashed: the 6 `@open-mercato/core` suites (16 tests — dashboards formatters, DealsKpiStrip, CompanyKpiBar, ItemsSection discount column, attachments localDriver, workflows dialog-ds-compliance) and `@open-mercato/ui`'s `utils/__tests__/format.test.ts` fail identically on the untouched tree. The `queue`, `shared`, `cli`, `telemetry`, `ai-assistant` and `create-app` failures are the same story (currency/ICU formatting, Windows path handling, an `EPERM` on directory rename). None is in a file this PR touches. Everything else is green, including `@open-mercato/core` 1241 suites / 9721 tests and `@open-mercato/ui` 228 suites / 1866 tests.
- **`om-auto-review-pr` review/autofix pass: could not run in its normal form, and I am not claiming that it did.** This is a fork PR authored by the same account the automation runs as, so GitHub rejects the self-approval and `AddLabelsToLabelable` is denied — the skill has no way to submit its verdict or move the pipeline label. Instead I applied the `om-code-review` checks to the final diff by hand and report the result here: no new `any`, no hardcoded user-facing strings, no DS-token violations, tenant/organization scoping preserved on every new query path (the indexer payload carries `organizationId`/`tenantId` from the record), no raw `fetch`, no new production dependency, and the scope stays inside the six files the review named plus the two new test artifacts. **This still needs a maintainer's re-review** — please treat the self-review as evidence, not as an approval.
- **Integration test:** `TC-WMS-028-zone-dialog.spec.ts` has **not** been re-run since the fix. It needs the ephemeral environment, which was not available in this session. Its shape and assertions are unchanged and remain the right coverage.

### 🔍 How to verify

- **Manual smoke test:** in a tenant with exactly one warehouse, define a text custom field for `wms:warehouse_zone` (Settings → Custom fields), then open WMS → Configuration → Zones → New. The warehouse should be preselected. Fill code, name and the custom field, save, then reopen the row for edit — the custom field must show the saved value. Then check `GET /api/wms/zones?ids=<id>` and confirm `customValues.<key>` carries it and no top-level `cf:*` keys remain.
- **Areas to spot-check in the diff:** `emitZoneCrudSideEffects` and its four call sites in `commands/configuration.ts` (the ordering relative to `writeZoneCustomFields` is the whole fix); the reworked effect in `ComboboxInput.tsx`; and `stripPrefixedKeys: true` in `api/zones/route.ts`, which is the one response-shape change.
- **Commands the reviewer can re-run:** `npx jest --config packages/core/jest.config.cjs --testPathPatterns="modules/wms"` · `npx jest --config packages/ui/jest.config.cjs --testPathPatterns="ComboboxInput"` · `yarn typecheck` · `yarn build:app`.
- **Rollback plan:** `git revert 6d5dbdd94 656483e80 e35ad735a 7fbae2411 c45810611` restores `1ac4915`. No migration, no schema change, no data backfill — the indexer fix only changes which events a command emits, so reverting it simply stops the reindex again.

- **Most likely regression:** `ComboboxInput` now refuses to sync a focused field whose value maps only to a placeholder. If some consumer legitimately relies on the raw value being displayed while focused (a combobox whose labels genuinely equal their values, e.g. a timezone picker), it would keep whatever was typed instead. The existing `does not reload suggestions after a self-labeled value is covered` and `does not loop when the first suggestions page cannot resolve the value` tests exercise exactly that self-labelled shape and both stay green, but this is the surface to watch.
- **Second-order effects:** zones now produce `entity_indexes` rows and coverage deltas where they previously produced none. That is the documented behaviour for an entity already declaring `indexer: { entityType }` on its route, and zones are absent from the module's `search.ts` and `vector.ts`, so no search-token or vector work is added. The visible effect is one extra synchronous index upsert per zone write.
- **Security-sensitive surfaces:** none. No auth, ACL, encryption or secrets path is touched. Every new query carries the record's own `organizationId`/`tenantId`; the route's `wms.view` / `wms.manage_zones` guards are unchanged.
- **Breaking-change impact:** one response-shape change worth calling out. `stripPrefixedKeys: true` removes the raw `cf:*` keys from `/api/wms/zones` items. Those keys were never part of `zoneListItemSchema` and were only ever emitted incidentally by the query engine's `includeCustomFields: true`; the only other in-repo consumer, `inventoryMutationLoaders.ts`, reads `id`/`name`/`code`/`warehouse_id` and is unaffected. `customValues` and `customFields` are additive. No exported symbol, signature, event id, widget spot, route, DB column, DI key, ACL feature or CLI command changed.
- **Residual risk accepted:** B1 is not proven end to end. The mechanism is traceable from `markOrmEntityChange` through `query_index.upsert_one` to the `entity_indexes` join the hybrid engine reads, and the command-side contract is pinned by unit tests, but the integration spec still needs one run against a live environment to close the loop. The row-action zone delete also still lacks optimistic locking — m1 scoped the fix to the dialog, and attaching a header there without the `surfaceRecordConflict` bar behind it would be worse than the existing gap, so it is left for a follow-up that does both.

### 🏷️ Labels

Label writes are denied for this account on this repository (`AddLabelsToLabelable`), so the `in-progress` claim label was never applied and the pipeline label could not be moved. Requested set for a maintainer:

- **`review`** — replaces `changes-requested`: every blocker, major, minor and nit from the review is addressed in `c458106..6d5dbdd`.
- **`feature`** — unchanged, still the right category.
- **`needs-qa`** — keep. The change touches `.tsx` and `packages/ui/src/`, so the automated-verification exemption does not apply, and `TC-WMS-028` has not been re-run since the fix.
- **`priority-medium`** — unchanged; a requested product gap with no data-loss or security angle.
- **`risk-high`** — keep, and it is better justified than before: `ComboboxInput`'s label-sync semantics reach every combobox in the product, and zone writes newly emit query-index events.

New issue #5360 (CrudForm accessible names) also needs labels applied by a maintainer — suggested `bug`, `priority-low`, `risk-low`.

Paul-Mlodochowki:
🤖 `om-auto-continue-pr` completed. Status: complete. Lock released.

Every blocker, major, minor and nit from the review is addressed in `c458106..6d5dbdd`; the DO-NOT-MERGE banner is lifted and the full resume summary is in the comment above. The `in-progress` label was never applied (label writes are denied for this account on this repo), so there is nothing to remove — this comment is the release signal.

Two things still need a maintainer: moving `changes-requested` → `review`, and a re-review, since GitHub will not let this account approve its own PR.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-21T12:38:27Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Re-review: feat(wms): preselect the sole warehouse and support custom fields in the zone dialog (#5239)

## 🎯 Summary

This is a re-review of `6d5dbdd` against my earlier review of `1ac4915` (2 blockers, 2 majors, 7 minors, 3 nits). **Every one of the 15 findings is addressed**, and the two that mattered are addressed with evidence rather than assertion.

The headline: **requirement 2 now provably persists.** The `DO-NOT-MERGE` banner is correctly lifted.

**Validation source: GitHub checks.** CI re-ran on this exact head today (run `32044756972`) and is fully green — `prepare`, `lint`, `ds-lint`, `test` (7m30s), `audit-scope`, `scope`, `license/cla`, and all **15/15 `ephemeral-integration` shards**. No local fallback was needed or run.

## Verdict

✅ **approve** — no blockers, no majors, no minors. Two nits below, both pre-existing and neither worth holding the PR for.

## 🧪 The evidence that settles B1

The PR body was candid that `TC-WMS-028-zone-dialog.spec.ts` had **not** been re-run since the fix, and the execution plan lists that as its top risk ("B1 cannot be proven end-to-end without a live environment"). It has now been re-run — by this PR's own CI. From shard 14/15 (job `96771277328`, pass, 18m8s):

```
✓ 47 packages/core/src/modules/wms/__integration__/TC-WMS-028-zone-dialog.spec.ts:123:3 ›
     TC-WMS-028: WMS create zone dialog (#5239) ›
     pre-selects the only warehouse and round-trips a custom field (2.3s)
```

That test carries the assertion that failed deterministically on the previous head with `Expected: "Pick face …", Received: null` — `expect(getCustomFieldValue(persisted!, customFieldKey)).toBe(customFieldValue)`, read back through the real `/api/wms/zones` list route. It now passes, along with the edit-dialog rehydration assertion and the requirement-1 preselect assertion. I verified the spec is genuinely selected rather than silently filtered out of the shard set: it appears in the shard's spec enumeration alongside TC-WMS-001…027 and executes as test 47.

## 🔬 What I verified independently

I did not take the PR body's causal story on trust — the previous review's B1 analysis and this PR's rationale contradicted each other on the read path, so I resolved that first.

**The PR is right, and my earlier analysis was describing the wrong engine.** `packages/core/src/modules/query_index/di.ts:17-59` overrides the `queryEngine` DI registration with `HybridQueryEngine`, which projects `cf:*` by left-joining **`entity_indexes`** (`lib/engine.ts:692-723`, `:949`, `:1943-1944`) — not the `LEFT JOIN custom_field_values` of `BasicQueryEngine` that I traced last time. So a zone whose value lives only in `custom_field_values` reads back `null`, exactly as observed.

**The fix's chain is traceable end to end.** `emitCrudSideEffects` → `dataEngine.markOrmEntityChange` (queues, `shared/lib/data/engine.ts:713-755`) → command-bus flush (`command-bus.ts:726`) → `emitOrmEntityEvent` awaits `query_index.upsert_one` (`engine.ts:672-687`) → the subscriber rebuilds the index doc on a forked EM from committed rows (`subscribers/upsert_one.ts:13-22`). Ordering is correct: the custom-field write happens before the mark in every one of the six call sites, and the new unit test pins that with `invocationCallOrder`.

**The fix cannot regress existing zones.** This was my main independent concern — zones previously had no index rows, so newly-indexed-but-partially-covered zones could in principle have flipped the engine onto an incomplete index and made older zones vanish from the list. It cannot: `lib/engine.ts:364-418` falls back to the basic engine on **both** no-index-rows and partial-coverage, and schedules an auto-reindex. The non-goal in the execution plan ("backfilling query-index coverage for zones created before this change… the index self-heals") is correct.

**No duplicate reindex.** Omitting `events` from the indexer config is the right call and I confirmed it is safe: the separately-emitted `wms.zone.*` events map through `toEntityTypeFromEvent` to `wms:zone`, which is not a registered entity type, so the legacy domain-event bridge no-ops rather than double-enqueueing `wms:warehouse_zone`.

**`stripPrefixedKeys: true` does not break the zones grid.** I checked this specifically, since stripping raw `cf:*` keys could have blanked custom-field columns. `DataTable`'s `entityId` prop drives only custom-field *filter fieldsets* (`DataTable.tsx:2676-2775`), never column accessors, and this table declares its four columns explicitly. Safe.

**Optimistic locking is genuinely armed now, not just declared.** `m1` was the one fix where "added `id`/`updatedAt` to `initialValues`" could have been cosmetic, because this dialog submits through a custom `onSubmit` with a raw `apiCall` rather than `updateCrud`. It is not cosmetic: `CrudForm.tsx:2997-3003` wraps the host's `onSubmit` in `withScopedApiRequestHeaders(buildOptimisticLockHeader(resolvedOptimisticLockUpdatedAt))`, so the header reaches the raw call. And `:2565-2668` confirms the changed-`initialValues` merge effect exists, which is what lets the async-resolved `soleWarehouseId` reach the form at all.

## ✅ Findings from the previous review — disposition

| # | Finding | Status |
|---|---|---|
| **B1** | Requirement 2 does not persist | ✅ Fixed — root-caused to the missing query-index emission; proven by TC-WMS-028 passing in CI |
| **B2** | Server-side custom-field change has zero unit tests | ✅ Fixed — `configuration.zoneCustomFields.test.ts`, 10 cases covering all six of my requested scenarios plus two I did not ask for (delete→`delete_one`, write-before-index ordering) |
| **M1** | Focused combobox overwrites a picked label with the raw record id | ✅ Fixed — exactly the suggested `hasRealLabel` guard, with the discriminating pick-then-stale-reload test |
| **M2** | `ComboboxInput` blast radius uncovered | ✅ Fixed — all three requested windows covered (real label arrives while focused; clear-with-parent-keeping-value; blur→refocus), and the PR body's table honestly marks which cases discriminate |
| **m1** | Zone edit dialog has no optimistic locking | ✅ Fixed — and verified to actually reach the wire |
| **m2** | `wms.zones.delete` undo does not restore custom fields | ✅ Fixed — `buildCustomFieldResetMap(before.custom, undefined)`, matching `deleteCompanyCommand.undo` |
| **m3** | Decorator should opt into `stripPrefixedKeys: true` | ✅ Fixed, with the rationale recorded in-code |
| **m4** | `customFields` schema erases the entry shape | ✅ Fixed — explicit `CustomFieldDisplayEntry`-shaped schema |
| **m5** | Dialog loads the warehouse list twice | ✅ Fixed — `loadZoneWarehouseOptions` serves the unsearched open from the React Query cache |
| **m6** | `& Record<string, unknown>` disables type checking | ✅ Fixed — `ZoneRow` gains explicit `customValues`/`customFields`; `ZoneFormValues` narrows to a `` `cf_${string}` `` index signature, which is exactly the narrowing I suggested |
| **m7** | Extra queries per update/delete | ✅ Addressed — comment explains the early-return cost model |
| **n1** | Non-discriminating test not labelled | ✅ Fixed |
| **n2** | Shared primitive's comment names a WMS screen | ✅ Fixed — the comment now describes only the mechanism |
| **n3** | CrudForm accessibility gap needs an issue | ✅ Fixed — #5360 filed and linked from the spec helper |

## 💅 Nit

**n1 — `WmsConfigurationPage.tsx:769-790`: `handleDelete` sends its `DELETE` without an optimistic-lock header.** Now that the edit form derives `x-om-ext-optimistic-lock-expected-updated-at` from `initialValues.updatedAt`, the archive path is the asymmetric one: it is a raw `apiCall` outside `CrudForm`, so `CrudForm`'s auto-derivation (which covers its own `onDelete`, not a sibling handler) never applies, and a stale row can be archived over a concurrent edit. Pre-existing and out of this PR's scope — `ZoneRow` now carries `updated_at`, so if you want it later it is `withScopedApiRequestHeaders(buildOptimisticLockHeader(row.updated_at), …)` around the existing call. Not worth another round-trip on this PR.

**n2 — `wmsLookupLoaders.ts:41-56`: a failed warehouse lookup is cached as a successful empty list.** `loadWarehouseOptions` returns `[]` when `!call.ok`, so React Query stores that as a success and `staleTime: 30_000` holds it; `loadZoneWarehouseOptions` then serves the empty cached array for the unsearched open instead of retrying. Because `soleWarehouseId` is derived from the same array, a transient failure degrades to "no warehouses, no preselect" rather than an error. This is the loaders file's established convention across every loader in it, so changing it here alone would be inconsistent — flagging it as a module-wide cleanup candidate, not a request.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed — `ComboboxInput`'s props are untouched; `seedOptions` pre-existed.
- [x] No function signature changed in a breaking way — `restoreZoneFromSnapshot` now returns the entity instead of `void`, but it is module-private.
- [x] No required type field removed or narrowed — `ZoneRow`/`ZoneFormValues` are module-local; `ZoneFormValues` narrows an index signature that this PR itself introduced.
- [x] No HTTP route removed/renamed; guards (`wms.view` / `wms.manage_zones`) unchanged.
- [x] No response field removed or retyped — `customValues`/`customFields` are additive (Cat. 7). `stripPrefixedKeys: true` removes raw `cf:*` keys, which is a **response-shape narrowing**; I checked it is safe because the route had no decorator before this PR, so no consumer could have been reading them, and the only in-repo consumer (this page, via `extractCustomFieldEntries`) reads `customValues`.
- [x] No event or message name renamed or removed — `wms.zone.*` payloads byte-identical; no new `wms.warehouse_zone.*` id introduced.
- [x] No table or column renamed/removed; no migration needed.

No `BACKWARD_COMPATIBILITY.md` protected surface is touched. `apps/mercato/src/app/**` and `.env.example` are untouched, so no `create-app` template-sync obligation.

The `ComboboxInput` behavioural change still reaches every combobox in the product, which is why `risk-high` stays on this PR — but it is now pinned by six tests, and the one regression it introduced was caught and closed before merge.

## 🧪 Test Coverage

**Unit.** `configuration.zoneCustomFields.test.ts` (414 lines, 10 cases) closes B2 completely. The harness is a genuine stub of the em/container seams rather than a mock of the code under test, it drives the **real registered command handlers** out of `commandRegistry`, and two assertions go beyond what I asked for: `delete removes the zone from the query index`, and an `invocationCallOrder` check that `setCustomFields` precedes `markOrmEntityChange` — which is the actual invariant that makes the fix work, since an index doc built before the custom-field write would carry stale values. That is the right thing to pin.

**UI.** Six `ComboboxInput` cases, of which the PR body honestly marks two as behavior pins rather than regression tests, with a code comment saying so. I re-derived the discriminating claims from the implementation and they hold.

**Integration.** TC-WMS-028 was already the right shape; it is unchanged apart from the #5360 link, and it has now actually executed and passed against the fix. Nothing further is needed.

**Remaining gap, not blocking.** The route-level wiring of `decorateCustomFields` on `/api/wms/zones` still has no dedicated unit test (my earlier gap 3). TC-WMS-028 now covers that path end-to-end, which is stronger evidence than the unit test would have been, so I am closing the point rather than carrying it forward.

## Runner note

No commands were run locally for this review. Per this repository's GitHub-checks-first rule, CI check results on `6d5dbdd` were the validation source, and the fallback conditions (no reported checks / unreadable logs) did not apply.

pkarw:
Label set to `merge-queue` because code review passed; `needs-qa` stays on so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Keeping `priority-medium` and `risk-high` — `risk-high` remains correct because `ComboboxInput` is a shared primitive behind every combobox in the product.

pkarw:

Automated coverage is already strong here — `TC-WMS-028` passed in CI and exercises the happy path of the zone dialog end to end, and six unit tests pin `ComboboxInput`. So **do not re-run the happy path manually**; spend the time on the two things automation cannot reach: the shared-primitive blast radius, and cross-organization isolation of custom-field values.

### P0 — `ComboboxInput` regression sweep (shared primitive, every combobox in the product)

This is the reason the PR is `risk-high`. The change alters *when* a focused combobox re-syncs its displayed label: previously never while focused, now whenever the user is not mid-typing and a real label is available. Unit tests pin six windows; the point of this pass is the ones they cannot simulate — real async latency, real network failures, real typing cadence.

**Where QA should click**
- `/backend/sales/orders` → any order → the customer / channel comboboxes
- `/backend/wms/inventory` (or the inventory profile dialog) → the product / variant comboboxes, which use composite `"Name (SKU)"` labels
- `/backend/customers/…` → any deal / task / activity form with a relation combobox
- Any form where a combobox is the **first** field (it will be `autoFocus`ed — that is the exact condition this PR changes)

**What human QA should verify**
- Open a form whose first field is a combobox with a saved value → the human-readable label renders, never a raw UUID and never a blank field.
- Type a partial query, pause mid-word for ~2s (long enough for the debounced load to land), then keep typing → your query is never rewritten under you.
- Pick an option from a dropdown, then leave the field focused and wait ~2s → the label stays; it must not flip to a UUID. This is the exact regression the previous review caught, so it is worth doing on a **composite-label** combobox (product/variant), where the follow-up search cannot match the label.
- Clear a combobox with the ✕ while it is focused → it stays cleared and does not resurrect the old label.
- Blur, then refocus without typing → the canonical label is shown.
- Throttle the network (DevTools → Slow 3G) and repeat the first two checks → slow label arrival must still land correctly rather than leaving the field blank.

**What can go wrong**
- A raw UUID appearing in a combobox instead of a label (the caught-and-fixed regression, re-check it stayed fixed).
- A blank combobox on a field that *does* hold a value — the user can then save a form that looks empty. Watch for this specifically when the value's label cannot be resolved (search a record whose label the lookup route cannot match).
- Your typed query being overwritten mid-keystroke by a late-arriving load.

### P0 — Zone custom fields: cross-organization and cross-tenant isolation

The write now also pushes values into the query index (`entity_indexes`), which is a new read surface for this entity. Automated tests cover a single org; isolation is the human check.

**Where QA should click**
- `/backend/settings` (or wherever custom-field definitions are managed) → define a text custom field on **Warehouse Zone** in organization A only
- `/backend/wms/zones` while scoped to organization A, then re-scope to organization B

**What human QA should verify**
- In org A: create a zone, fill the custom field, save → reopen via Edit → the value is there.
- Switch the org selector to org B → the zone list shows **no** org-A zones, and the zone form shows **no** org-A custom field.
- Back in org A, edit the zone and **change** the custom field value → save → reopen → the new value is shown (not the old one). Then **clear** it → save → reopen → it is empty.
- Search/filter the zones list by the custom field if a filter is offered → results stay within the current organization.

**What can go wrong**
- An org-B session seeing an org-A zone or an org-A custom-field value (data-scoping breach — stop and fail immediately).
- A stale value surviving an edit, which would mean the index doc was rebuilt before the value was written.

### P1 — Sole-warehouse preselect and the edit dialog

**Where QA should click**
- `/backend/wms/zones` → **Add zone**, in a tenant/organization that has **exactly one** warehouse
- The same dialog in an organization with **several** warehouses
- An organization with **zero** warehouses

**What human QA should verify**
- Exactly one warehouse → the dialog opens with that warehouse already selected and its **name** displayed (not a UUID, not an empty "Type to search…").
- Several warehouses → the field opens **empty** and requires a choice; it must not auto-pick an arbitrary one.
- Zero warehouses → the form degrades sanely (empty field, save blocked by the required-field validation) rather than erroring or spinning.
- Edit an existing zone → the warehouse field shows the currently linked warehouse's name immediately, without you having to open the dropdown.
- Open the dialog twice in a row → the warehouse list is not re-fetched needlessly and the second open behaves identically to the first.

**What can go wrong**
- The preselect firing when there is more than one warehouse (silently picking the wrong warehouse for the operator).
- The field showing a UUID or staying blank despite a value being set — that would mean the label sync is still losing the race.

### P1 — Optimistic locking on the zone edit dialog

New on this PR: the edit form now sends the concurrency header, so conflicts are actually detected where they previously passed silently.

**Where QA should click**
- `/backend/wms/zones` → open the same zone's Edit dialog in **two** browser tabs

**What human QA should verify**
- Save in tab 1 → then save in tab 2 → tab 2 must surface the record-conflict bar rather than silently overwriting tab 1's change.
- Reload tab 2 and save again → it succeeds.

**What can go wrong**
- A silent last-write-wins overwrite (the old broken behaviour) — that would mean the header is not reaching the request.
- A **false** conflict on an ordinary single-tab save, especially right after creating a zone with custom fields.

### P2 — Undo of zone create / update / delete

**Where QA should click**
- The action-log / undo surface, after creating, editing and archiving a zone that carries a custom-field value

**What human QA should verify**
- Undo a **delete** → the zone comes back **with** its custom-field value, and reappears in the zones list.
- Undo an **update** that changed the custom field → the previous value is restored, and a key the update introduced is cleared rather than left behind.
- Undo a **create** → the zone disappears from the list.

**What can go wrong**
- A restored zone missing from the list (index not re-upserted) or present but with the custom-field column blank.
- An archived zone still appearing in the list after delete.

### Pass/fail

pkarw:

Re-review of `6d5dbdd` against the earlier review of `1ac4915`: all 15 findings (2 blockers, 2 majors, 7 minors, 3 nits) are addressed. No autofix iterations were needed — nothing was left to fix.

The decisive evidence is that `TC-WMS-028-zone-dialog.spec.ts` — which the PR body noted had not been re-run since the fix — executed and **passed** in this PR's own CI (shard 14/15, job 96771277328), so requirement 2 is now proven end to end rather than argued from code. Validation source: GitHub checks; the full gate is green on this head, including all 15 `ephemeral-integration` shards.

Two nits are recorded in the review, both pre-existing and neither blocking: `handleDelete` archives without an optimistic-lock header, and `loadWarehouseOptions` caches a failed lookup as an empty success.

Moved to `merge-queue`; `needs-qa` stays, so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Manual QA routes are in the comment above.

pkarw (approved):

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-22T01:43:45Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

pkarw:
## 📸 UI QA evidence — PASS

**Verdict:** ✅ PASS
**Environment:** `http://127.0.0.1:5001` · role `scoped employee fixture` · browser `playwright`
**Verified:** `feat/issue-5239-wms-zone-dialog` @ `6d5dbdd9`

### Scenario (P1 — WMS zone dialog)

**Where to click:** `/backend/wms/zones` → **Add zone** → save → row actions → **Edit**

**What to verify:** With exactly one visible warehouse, the create dialog preselects it; an editable `wms:warehouse_zone` custom field renders; saving persists the selected warehouse and custom value; reopening the zone hydrates both values; the dialog remains usable at a 390×844 viewport.

**What can go wrong:** The warehouse can render blank while focused, custom fields can be stripped before submit or omitted from the index-backed list response, edit mode can lose the persisted value, or the dialog can escape the mobile viewport.

| 1 | Cold-load WMS Zones | A useful loading shell appears before the route is interactive. | The loading table shell appeared and **Add zone** was ready in 539 ms. | ✅ |
| 2 | Open Create zone | The only warehouse is preselected and the custom field renders. | The combobox displayed the sole warehouse; the custom field appeared after its async definition load. | ✅ |
| 3 | Save a zone with custom data | The row appears and both values round-trip through the zones API. | The row appeared in 901 ms; the API returned the expected warehouse id and custom value. | ✅ |
| 4 | Reopen Edit zone | The warehouse label and custom value are hydrated. | Both values were visible in the edit dialog. | ✅ |
| 5 | Resize to 390×844 | The changed dialog remains inside the viewport and usable. | Dialog bounds were exactly 0–390 px; all fields and Save remained visible. | ✅ |


### Screenshots






### Notes for QA

- The committed `TC-WMS-028-zone-dialog.spec.ts` provides browser-level coverage, so no follow-up test scenario is needed.
- A focused test-runner attempt was stopped after more than six minutes of repository-wide Playwright discovery at 100% CPU. It never launched Chromium or entered the test body, so this report does not claim an automated test result.
- The manual provider run created an isolated organization, employee, sole warehouse, and custom-field definition, then cleaned up the fixtures.
- The masked DataTable behind the modal retains a 716 px full-document width on mobile, but the changed dialog itself fits exactly within the 390 px viewport; the screenshot preserves that observation.

pkarw:
🤖 `om-auto-qa-pr` completed: PASS. Evidence posted at https://github.com/open-mercato/open-mercato/pull/5334#issuecomment-5377324692. Lock released.

PR #5335, opened 2026-08-17
fix(sales): accept comments and internalNotes on document update

kamwro:
## Summary

`documentUpdateSchema` (`modules/sales/commands/documents`) is a closed `z.object`. It declares `comment` (singular) and **no `internalNotes` at all**, while the **create** schema (`modules/sales/data/validators`) declares `comments` and `internalNotes`.

So a caller that sends the create-schema field names on an update gets: zod strips both keys, `applyDocumentUpdate` never sees them, and the command **returns success having written neither column**. No error, no warning, no validation failure. Creates are unaffected, which is what makes it hard to spot — the columns simply keep whatever the document carried when it was first created.

Found by a downstream bulk re-import that re-synced 14,272 orders whose source held 287 non-empty customer notes and stored 0 of them. A separate clean 214-order re-sync added none either. Both runs reported every document updated, and both looked green.

## Changes

- `documentUpdateSchema` — accept `comments` and `internalNotes`, with the create schema's names and `max(4000)` limit. Nullable here, unlike on create, because an update has an existing value to clear.
- Its `.refine()` — count both as a change, so an update whose only edit is a note is no longer rejected as `update_payload_empty`.
- `applyDocumentUpdate` — assign both, with the same trim / empty→null shape as the existing `comment` branch.
- `buildDocumentUpdateChangeKeys` — add both. This feeds `buildLog`'s audit-log diff only; it does not reach reindex or events.

`internalNotes` is applied for orders only. `entity` is `SalesOrder | SalesQuote` and only the order half has the column, so an unguarded assignment sets a property MikroORM cannot persist — succeeding and writing nothing, which is the same silent no-op this PR exists to fix. The guard follows the pattern already used for the other order-only fields in this function.

`comment` (singular) is untouched and keeps working — the order-detail page sends it. Both names land on `comments`, with the canonical plural winning if a payload carries both; the assignment order is commented so it doesn't read as accidental.

Blast radius is the two update commands and the PUT route that parses through the same schema (`api/documents/factory.ts`). No shipped code composes `documentUpdateSchema` — the `.omit({ id: true }).partial()` composition appears only in `.ai/specs/SPEC-021-2026-02-07-compound-commands-graph-save.md`, and per this repository's own lesson record Zod v4 will not `extend` a refined object schema, so that snippet needs `safeExtend` whenever it is implemented.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**
`.ai/specs/SPEC-021-2026-02-07-compound-commands-graph-save.md` references `documentUpdateSchema` but only by composition, never enumerating its fields, so no spec text becomes wrong. It is the only spec that mentions the schema.

## Testing

New: `packages/core/src/modules/sales/commands/__tests__/documents.update-notes.test.ts`, 17 cases. Six pin the schema; the rest drive the real `sales.orders.update` and `sales.quotes.update` commands against a mocked EntityManager (same harness style as `documents.tag-sync.test.ts`): both fields written, both cleared by `null` and by whitespace, both left untouched when omitted, the legacy singular `comment` still honoured, `internalNotes` kept off a quote entity while `comments` still lands on one, and both reported to the audit log as changed keys.

```
$ yarn test --testPathPatterns documents.update-notes    # in packages/core
Tests: 17 passed, 17 total

# with the source change reverted, test file kept:
Tests: 13 failed, 4 passed, 17 total

$ yarn test --testPathPatterns "modules/sales"
Tests: 581 passed, 581 total
```

Two suites in that run (`channels.route`, `shipments.afterList`) fail to *load* on a missing `#generated/…` module in a worktree that has not run `yarn generate`; they contribute 0 tests and are unrelated.

Note for reviewers: `documents.ts` carries `// @ts-nocheck` on line 1, so `yarn typecheck` says nothing about this file either way — a deliberately planted type error in it is not reported. The jest coverage above is the only real signal here.

## Checklist

- [x] I have read and accept the Open Mercato Contributor License Agreement (see `docs/cla.md`). — left for the PR author to confirm.
- [ ] I updated documentation, locales, or generators if the change requires it. — n/a: no user-facing string, locale key or generator input is touched.
- [ ] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required). — covered at the command level, where the schema stripping actually happens.
- [ ] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — n/a, per the Specification section above.
- [ ] Priority set: this PR carries exactly one `priority-*` label. — this account has `read` here and cannot set labels; intent is in a follow-up comment.
- [ ] Risk set: this PR carries exactly one `risk-*` label. — same.
- [ ] QA routing set. — same; `needs-qa` is the intent, rationale in the follow-up comment.

### Design System Compliance

Not applicable — no UI surface is touched; the change is a zod schema and its command handler.

## Linked issues

None.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5335"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789545008&installation_model_id=432243&pr_number=5335&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5335&signature=2bab37f6c40d6de7987d4582c8b4885e67a95fd454a6047e8d55cd55419aec3c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


kamwro:
This account has `read` here, so none of this can be applied from our side. Intended metadata:

- `bug` — an update silently discards two fields and reports success; the columns keep their create-time values.
- `priority-medium` — no data corruption, but a write that reports success and stores nothing is invisible to the caller, and only shows up as missing data later.
- `risk-low` — additive schema fields plus one guarded assignment; `comment` (singular) is untouched and creates are unaffected.
- `needs-qa` — the accepted payload of `PUT /api/sales/documents/*` widens, so this is an API-surface change and not eligible for the automated-verification exemption, despite having no UI surface.

Reviewer intent: a maintainer of the `sales` module.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T10:49:05Z. Other auto-skills will skip this PR until the lock is released.

pkarw (dismissed):
# 🔍 Code Review: fix(sales): accept comments and internalNotes on document update

## 🎯 Summary

This PR closes a real silent-write bug in the sales document update path. `documentUpdateSchema` (`packages/core/src/modules/sales/commands/documents.ts:587`) is a closed `z.object`, so the `comments` and `internalNotes` keys that the **create** schemas declare (`packages/core/src/modules/sales/data/validators.ts:713-714`, `:752`) were stripped before `applyDocumentUpdate` ever ran — `sales.orders.update` and `sales.quotes.update` reported success while writing neither column. The diagnosis is correct, the reproduction (a bulk re-import that stored 0 of 287 notes) is credible, and the fix is the right shape: declare the two fields with the create schema's names and `max(4000)` limit, count them in the `.refine()` so a note-only edit is no longer rejected as `update_payload_empty`, assign them in `applyDocumentUpdate` with the same trim/empty→null normalization the existing `comment` branch uses, and register them in `buildDocumentUpdateChangeKeys` so the audit-log diff sees them.

I verified the supporting facts against the code rather than the PR body: `SalesOrder` declares both `comments` and `internalNotes` while `SalesQuote` declares only `comments` (`packages/core/src/modules/sales/data/entities.ts:424-428`, `:885-886`), so the order-only guard on `internalNotes` is genuinely required; `buildDocumentUpdateChangeKeys` really does feed only `buildLog`'s `buildChanges` call (`documents.ts:5352`, `:5596`) and never reaches reindexing or events; and the widened schema reaches exactly one HTTP surface, the shared documents PUT route (`packages/core/src/modules/sales/api/documents/factory.ts:522-543`), which is guarded by `requireAuth` plus the binding's manage feature. The change is additive at every contract surface, the backward-compatibility posture is clean, and the legacy singular `comment` keeps working. The reviewed scope is the two changed files; I also read the update commands, the API factory, the entities, the create validators and the detail/create UI to check the blast radius.

The change is not mergeable as it stands because of one consequence on the quote half of the shared schema, described below. Everything else I found is optional polish or a follow-up worth filing, not a merge blocker.

## Verdict

❌ **Request changes** — the newly accepted `internalNotes` is validated and counted as a real edit on **quote** updates but then dropped, which turns a payload that used to be rejected into a successful, empty update; on a quote in `sent` status that empty update still clears the acceptance token and reverts the quote to `draft`. That is the same silent-success class this PR exists to eliminate, so it should not ship with the fix that names it.

## 🧪 Validation Gate

**Validation source: GitHub checks** (this repository's `om-auto-review-pr` extension is GitHub-checks-first; CI already runs the configured gate on every PR, so no local re-run was needed). All checks reported for head `e3797b3` are green — no failing check, no pending check. Branch protection reports no required-status-check configuration (`404 Required status checks not enabled`), so every reported check is treated as required. The configured `validation.commands` map onto CI run [32007552197](https://github.com/open-mercato/open-mercato/actions/runs/32007552197) as follows.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job (`.github/workflows/ci.yml:170`) |
| `yarn generate` | ✅ PASS | `prepare` job (`ci.yml:178`) |
| `yarn build:packages` (post-generate) | ✅ PASS | `prepare` job (`ci.yml:181`) |
| `yarn i18n:check-sync` | ✅ PASS | `test` job (`ci.yml:593`) |
| `yarn i18n:check-usage` | ✅ PASS | `test` job (`ci.yml:596`) |
| `yarn typecheck` | ✅ PASS | `test` job, turbo-filtered to packages changed against the base (`ci.yml:612`). See the caveat below. |
| `yarn test` | ✅ PASS | `test` job (`ci.yml:624`), plus `test:scripts` and `test:repo-wide-guards`; the 15 `ephemeral-integration` shards are green as well |
| `yarn build:app` | ✅ PASS | `prepare` job (`ci.yml:193`) |

Additional green checks: `lint`, `ds-lint`, `audit-scope`, `scope`, `license/cla`. Skipped by design on a fork PR: `docker-build`, `merge-coverage`, `audit`, `[code]smith`, and the Dokploy preview deployment (the preview block comment is a permissions guard on a `read`-level contributor, not a failure).

**Typecheck caveat, as the author correctly disclosed:** `packages/core/src/modules/sales/commands/documents.ts:1` carries `// @ts-nocheck`, so the green typecheck says nothing about this file. In particular the `(entity as SalesOrder).internalNotes` narrowing at `documents.ts:1182` is unchecked. I confirmed by hand that `internalNotes` exists on `SalesOrder` and not on `SalesQuote`, so the cast is sound — but the jest suite is the only real signal for this file, which raises the bar on the test gap noted under Test Coverage.

## Findings

### ⚠️ Major

**1. `internalNotes` on a quote update: validated, counted as a change, then silently dropped — and it still invalidates a sent quote.** `packages/core/src/modules/sales/commands/documents.ts:601`, `:646`, `:1179`

`documentUpdateSchema` is shared by `sales.orders.update` and `sales.quotes.update`, so the new `internalNotes` field is now accepted for **both** kinds, and `documents.ts:646` makes it satisfy the `.refine()` on its own. But `applyDocumentUpdate` only assigns it when `kind === "order"` (`documents.ts:1179`), which is correct in itself because `SalesQuote` has no such column. The combination is the problem:

- **Before this PR**, `sales.quotes.update` with `{ id, internalNotes: "x" }` was rejected: zod stripped the unknown key, the refine saw an otherwise empty payload, and the caller got `update_payload_empty`.
- **After this PR**, the same payload parses, the command executes, `applyDocumentUpdate` writes nothing, and the caller receives success — a write that reports success and stores nothing, which is verbatim the failure mode the PR description argues against.
- Worse, the quote update path is not side-effect-free. `documents.ts:5184-5187` clears `acceptanceToken` and `sentAt` for any quote whose status is `sent`, and `documents.ts:5211-5218` then reverts `status` to `draft` and re-resolves `statusEntryId`. So a no-op payload against a **sent** quote silently kills the customer's acceptance link and pulls the quote back to draft while changing none of the fields the caller asked for.

The trigger is exactly the scenario this PR is built for: the code comment at `documents.ts:598-600` invites callers to "send one payload shape to create and update", and the originating bug was a bulk importer doing precisely that. An importer that reuses an order-shaped payload against the quotes endpoint now degrades sent quotes on every sync.

Fix: make the rejection explicit instead of silent, in the spirit of the rest of the change. The smallest version is a guard in `applyDocumentUpdate` next to the existing assignment:

```ts
if (kind === "quote" && input.internalNotes !== undefined) {
  throw new CrudHttpError(400, {
    error: translate(
      "sales.documents.detail.internalNotesUnsupported",
      "Internal notes are not available on quotes.",
    ),
  });
}
```

This is not a regression for any existing caller — the field was previously stripped, so nobody could be relying on it succeeding — and it converts a silent no-op into an actionable 400.

For completeness, and explicitly **not** something I am asking this PR to fix: `expectedDeliveryAt` already has the same shape on `develop` (`documents.ts:1244` guards on `"expectedDeliveryAt" in entity` while `documents.ts:637` counts it in the refine), so a quote update carrying only `expectedDeliveryAt` degrades a sent quote today. If you prefer, one kind-aware guard covering both fields closes the whole class; a follow-up issue for the pre-existing half is equally fine. Whichever route you take, please do not widen the hole in the same PR that documents it.

### 💅 Nit

**2. `internalNotes` remains write-only through the documents API.** `packages/core/src/modules/sales/api/documents/factory.ts:330-378`

Neither `commonFields` nor `orderOnlyFields` projects `internal_notes`, and no `.tsx` in the repository reads the field, so after this PR a caller can write `internalNotes` through `PUT /api/sales/orders` but can never read it back through `GET /api/sales/orders`. That asymmetry predates the PR (create has it too) and adding the column to the projection is a separate, deliberate decision about exposing an internal field, so this is not a change request. It is worth stating because it dictates how the fix gets verified: persistence has to be confirmed through the audit log (`loadOrderSnapshot` does carry `internalNotes`, `documents.ts:1901`) or the database, not through the API response.

**3. The same silent-strip class survives for several sibling create-schema fields.** `packages/core/src/modules/sales/commands/documents.ts:587-661`

The update schema still omits `validFrom`, `validUntil`, `dueAt`, `taxInfo`, `deliveryWindowId`, `deliveryWindowCode`, `taxStrategyKey` and `discountStrategyKey`, all of which the create schemas declare (`data/validators.ts:696-715`, `:745-760`). Each is stripped on update in exactly the way `comments`/`internalNotes` were — notably, a quote's `validUntil` cannot be changed through this schema at all. That is out of scope here, but it is the same defect with different field names, so a follow-up issue enumerating them would keep the finding from being rediscovered by the next re-import.

**4. The PR description's SPEC-021 blast-radius claim does not match shipped code.** The body states that SPEC-021's compound graph-save "derives its `header` from this schema (`documentUpdateSchema.omit({ id: true }).partial()`, twice)". That composition exists only in the spec document (`.ai/specs/SPEC-021-2026-02-07-compound-commands-graph-save.md:114`, `:155`); no shipped code composes `documentUpdateSchema`, and per this repository's own lesson record (`.ai/lessons/use-safeextend-when-composing-refined-zod-object-schemas.md`) Zod v4 refuses to `extend`/reshape a refined object schema anyway, so that spec snippet will need `safeExtend` when it is implemented. The practical blast radius of this PR is therefore narrower than the body claims: the two update commands and the one PUT route. Worth correcting in the description so a future reader does not chase a dependency that is not there.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `documentUpdateSchema` and `DocumentUpdateInput` keep their identities; the object gains two optional members.
- [x] No function signature changed in a breaking way — `applyDocumentUpdate` and `buildDocumentUpdateChangeKeys` keep their signatures.
- [x] No required type field removed or narrowed — both additions are `.nullable().optional()`, so every previously valid payload stays valid and `DocumentUpdateInput` widens rather than narrows.
- [x] No HTTP route URL removed or renamed; no method changed — the documents PUT route is untouched.
- [x] No field removed or retyped in an existing response shape — `mapUpdateResponse` is unchanged.
- [x] No event or message name renamed or removed; no payload field removed — the change keys feed only the audit-log diff, not events or the query index.
- [x] No database table or column renamed or removed; no column type narrowed — no migration is needed, because `sales_orders.comments`, `sales_orders.internal_notes` and `sales_quotes.comments` already exist.
- [x] Where a contract had to change: the accepted request body of `PUT /api/sales/{orders,quotes}` widens additively, which `BACKWARD_COMPATIBILITY.md` classifies as an allowed additive change; the legacy singular `comment` continues to work and is still honoured on its own.

One behavioral change is worth naming explicitly even though it is intended: the `.refine()` now treats a note-only payload as a real edit, so `{ id, comments }` and `{ id, internalNotes }` no longer return `update_payload_empty`. That is the point of the fix on the order path — finding 1 is about the one kind where it lands wrong.

## 🧪 Test Coverage

`packages/core/src/modules/sales/commands/__tests__/documents.update-notes.test.ts` is a genuinely good suite for this fix. Six schema-level cases pin acceptance of both field names, the absence of `update_payload_empty` on a note-only payload, and the 4000-character limit (with a second real edit in the payload so a rejection can only come from the length rule — a nice detail). The remaining cases drive the real `sales.orders.update` and `sales.quotes.update` handlers through the command registry against a mocked `EntityManager`, in the same harness style as `documents.tag-sync.test.ts`, and cover both fields written together and separately, clearing by `null` and by whitespace, both left untouched when omitted, the legacy singular `comment` still honoured, the canonical plural winning when both are sent, `internalNotes` kept off a quote entity while `comments` still lands on one, and both surfacing in `buildLog`'s changed keys. The author's revert check (13 of 17 failing with the source change backed out) is the right evidence that these are regression tests and not restatements of current behavior.

Two gaps:

- **The quote path in finding 1 is untested as a behavior.** `documents.update-notes.test.ts:238` asserts that `internalNotes` does not land on the quote entity, but nothing asserts what the command *does* with an update whose only field is `internalNotes` on a quote — today it resolves successfully. Once you decide between rejecting and no-op'ing, add a case that pins it: `await expect(updateQuote({ internalNotes: 'x' })).rejects.toThrow(...)` for the rejection route, plus a case on a `status: 'sent'` quote asserting that `acceptanceToken`, `sentAt` and `status` are left alone by a payload that changes nothing.
- **No coverage at the HTTP boundary.** The bug's real entry point is `PUT /api/sales/orders`, which reaches the schema through `parseScopedCommandInput` after `splitCustomFieldPayload` (`api/documents/factory.ts:524-541`). The command-level tests are the right place for the stripping bug itself, and I accept the author's rationale for not adding a Playwright case, but since `// @ts-nocheck` removes typecheck from the picture for this file, one route-level integration case in `.ai/qa/tests/` asserting that a PUT carrying `comments` and `internalNotes` round-trips into the audit log would be cheap insurance for the surface that actually widened. Recording it as a follow-up is a reasonable alternative to blocking on it.

pkarw:

- `changes-requested` — the review found one Major finding (quote updates accept `internalNotes`, count it as a real edit, then drop it, degrading a `sent` quote as a side effect of an empty update). Pipeline label moved off the implicit review state.
- `bug` — the change fixes a silent write-loss on `sales.orders.update` / `sales.quotes.update`, matching the author's stated intent.
- `priority-medium` — as the author proposed: no data corruption, but a write that reports success and stores nothing is invisible to the caller until the data is missing later.
- `risk-medium` — raised one step from the author's `risk-low`. The schema addition itself is additive and low blast radius, but `documentUpdateSchema` is shared by both the order and quote update commands and the documents `PUT` route, and the review found a non-obvious side effect on the quote half (acceptance-token reset plus a `sent` → `draft` status revert). Per the repo's conflicting-signal rule, the higher rating wins and the reason is recorded here.
- `needs-qa` — agreed with the author. The accepted request payload of `PUT /api/sales/{orders,quotes}` widens, so the automated-verification exemption (which requires the API surface unchanged) does not apply, despite no UI file being touched. Note for whoever runs QA: `internal_notes` is not in the documents API projection, so persistence must be confirmed through the audit log or the database, not the `GET` response.
- `partner-request` — pre-existing, left untouched.

QA-approval gate: `needs-qa` is present and `qa-approved` is not, so this PR cannot merge on labels alone even once the review findings are resolved. The `qa` pipeline label is reserved for QA reviewers and was not set.

pkarw:

pkarw:

One Major finding drives the verdict: the newly accepted `internalNotes` is validated and counted as a real edit on **quote** updates but then dropped, so a payload that previously failed with `update_payload_empty` now succeeds having written nothing — and on a quote in `sent` status that empty update still clears `acceptanceToken`/`sentAt` and reverts the status to `draft`. Three nits are informational (write-only `internalNotes` through the API projection, the sibling create-schema fields still silently stripped on update, and a SPEC-021 claim in the PR body that does not match shipped code).

Validation source: GitHub checks — all checks green on `e3797b3`, none pending, so the configured gate was not re-run locally. No inherited reviewer feedback was outstanding.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-17T11:04:48Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

Claiming the `in-progress` label only — the PR stays assigned to @kamwro, since `om-auto-review-pr` deliberately handed it back to the author for the next pass and this QA run is read-only on the source.

pkarw:
## 📸 UI QA evidence — ❌ FAIL

**Verdict:** ❌ FAIL — the order-side fix works exactly as the PR describes and survived every check
(steps 1–9, 11), but a **sent quote is silently downgraded to draft** by an update whose only field is
`internalNotes` (step 10). That confirms the review's Major finding in a real browser, against a real
database, so QA cannot sign this off.
**Environment:** `http://127.0.0.1:5001` · role `admin` · browser `playwright`
**Verified:** `fix/sales-order-update-notes` @ `e3797b3`

The app was booted as a **production** Next.js build (`next start`, `NODE_ENV=production`) against a
disposable testcontainers pgvector database, migrated and seeded fresh. The diff touches no
UI-rendering file, so this run verified the closest observable surfaces — the order and quote detail
pages and the audit-log History tab — and used API-level checks where the change has no UI control at
all. Every fixture was created through the app's own APIs and deleted in teardown; only the seeded
demo admin credentials were used.

### Scenario (P0–P2 — sales document notes on update)

**Where to click:** `/backend/sales/orders/{id}` (inline **Comment** field, **History** tab),
`/backend/sales/quotes/{id}` (status pill, **Comment** field).

| 1 | Create an order with the create-schema names `comments` + `internalNotes`, open its detail page | The note renders in COMMENT and both columns persist | COMMENT showed the note; both columns non-null. Cold load 268 ms to DOM ready, 2.5 s to network idle | ✅ |
| 2 | Edit the inline **Comment** field — the only UI editor, and it still sends the legacy singular `comment` | The pre-existing path is untouched: saves, survives reload | Saved, rendered, persisted across a full reload, and read back through the API | ✅ |
| 3 | `PUT /api/sales/orders` with **only** `comments` | 200 instead of the pre-fix `update_payload_empty`, value written | 200; value read back and rendered on the page | ✅ |
| 4 | `PUT` with **only** `internalNotes`, then open **History** | 200, `internal_notes` rewritten, `comments` untouched, audit reports the key | 200; column rewritten, `comments` unchanged, timeline shows **"Update sales order — Changed fields: Internal Notes"** | ✅ |
| 5 | `PUT` carrying **both** `comment` and `comments` | The canonical plural wins | Page rendered the plural value; the singular value was absent | ✅ |
| 6 | Clear both fields by `null`, then again with whitespace-only strings | Both columns NULL both times, UI falls back to "Not set" | Both NULL after `null` and after `"   "`; COMMENT showed "Not set" | ✅ |
| 7 | 4001 characters to each field, then exactly 4000 | 4001 rejected by the new `max(4000)`, 4000 accepted | Both 4001 payloads → 400 `too_big` naming the field; 4000 → 200, read back at exactly 4000 | ✅ |
| 8 | `PUT /api/sales/quotes` with `comments` + `internalNotes` on a **draft** quote | No failure on a field the quote cannot store; the comment lands | 200; the quote detail page rendered the new comment | ✅ |
| 9 | Reload the order detail page at a 390×844 mobile viewport | The note stays rendered and reachable | COMMENT card rendered correctly at mobile width | ✅ |
| 10 | **Send** a quote, then `PUT /api/sales/quotes` with **only** `internalNotes` | Either rejected as before this PR, or stored — an empty update must not disturb the quote | **200 having written nothing**, and the sent quote was downgraded: `status` sent → **draft**, `acceptance_token` and `sent_at` both cleared. The page went from **Sent** to **Draft** | ❌ |
| 11 | Control: `PUT` an **undeclared** key (`customerNotes`) alone, then the same shape with `comments` | The undeclared key still reproduces the PR's described failure; the new name does not | Undeclared key → 400 `update_payload_empty`; `comments` → 200 | ✅ |

### Notes for QA

- **Step 10 is the blocker, and it is a new failure mode.** Before this PR, `{ id, internalNotes }`
  was rejected by `.refine()` as `update_payload_empty`, so it could never reach the quote command.
  The PR adds `internalNotes` to both the schema and the refine's change-count, but
  `applyDocumentUpdate` assigns it only when `kind === "order"` — correctly, since `SalesQuote` has no
  such column. The result is an update that is accepted, writes nothing, and still runs
  `sales.quotes.update`, which unconditionally clears `acceptanceToken`/`sentAt` for a quote in `sent`
  status. A customer holding the acceptance link loses it, and the quote silently reverts to draft.
  Steps 10a and 10b are the same quote before and after that one call.
- **The order half of the change is solid.** Steps 3–7 cover accept, write, precedence, both clearing
  shapes, the 4000-character bound, and the audit-log changed keys — all as the PR describes.
- **`internalNotes` is write-only through the app.** No form field exposes it, and no API read path
  returns it (`mapUpdateResponse` and the list/detail projection expose only `comment`). Its
  persistence had to be verified against the database column and the History timeline, which is worth
  knowing for whoever re-tests this.
- **The two note columns are encrypted** (`sales:sales_order`), so raw column reads were used only for
  null / not-null / changed shape; all content assertions went through the app's own read paths.
- **Not exercised:** the customer-facing acceptance page after the token is cleared (`/quote/{token}`)
  — the raw token is not returned by the send endpoint, so the downstream customer impact is inferred
  from the cleared `acceptance_token` column rather than observed. The SPEC-021 compound graph-save
  path, whose header widens by the same two fields, was also not driven.

### 📸 Screenshots

In order below: **step 01** the created note on the order detail page · **step 02b** the inline
Comment editor's value after a full reload · **step 04** the History timeline reporting
*Changed fields: Internal Notes* · **step 09** the order detail page at a 390×844 mobile viewport ·
**step 10a** the quote right after it was sent (**Sent**) · **step 10b** the same quote after one
`internalNotes`-only update (**Draft**, acceptance token cleared).

The full set — including the null/whitespace clearing states and the quote comment check — is in the
run's artifacts directory.

pkarw:

This change ships no browser-level test — the 17 new cases are jest command tests against a mocked
EntityManager, which is the right place for the schema/writer logic but cannot see the two things
this QA run found: that the audit log surfaces the changed key in the History timeline, and that a
quote in `sent` status loses its acceptance state to an update that writes nothing. Lock both in with
an automated test (run `/om-integration-tests`).

1. **Setup:** through the app's own APIs — create a sales order with a line plus `comments` and
   `internalNotes` (`POST /api/sales/orders`), and create a second document, a quote carrying a
   `customerSnapshot.contact.email`, then send it (`POST /api/sales/quotes/send`).
2. **Act:**
   - `PUT /api/sales/orders` with `{ id, comments }` alone, then `{ id, internalNotes }` alone.
   - `PUT /api/sales/orders` carrying both `comment` and `comments` in one payload.
   - `PUT /api/sales/orders` with `null` and with a whitespace-only string for each field.
   - `PUT /api/sales/quotes` with `{ id, internalNotes }` alone against the **sent** quote.
   - Open `/backend/sales/orders/{id}` and its **History** tab in the browser.
3. **Assert:**
   - Each order update returns 200 rather than `update_payload_empty`, and the detail page renders the
     new note (`comments` is readable through the `comment` projection).
   - The plural value wins when both names arrive; both clearing shapes leave the columns NULL and the
     UI shows "Not set".
   - The History timeline lists `Changed fields: Comments` and `Changed fields: Internal Notes` —
     `internalNotes` has no read path, so this is the only app-level signal that it was recorded.
   - **The regression guard:** after the `internalNotes`-only update, the sent quote still has its
     `sentAt`/`acceptanceToken` and its status is still `sent`. This assertion fails on the current
     head, which is the point — it pins the fix once the quote branch is settled.
4. **Teardown:** delete every order and quote the test created.

A worked implementation of all of the above (Playwright, using the repo's own
`__integration__` auth/sales fixtures) was written for this QA run and is available in the run's
artifacts directory if it helps as a starting point; it is deliberately not part of this PR, since
this skill does not modify source.

pkarw:
🤖 `om-auto-qa-pr` completed: FAIL. Evidence posted above. Lock released.

Eleven checkpoints were driven in a real browser against a production build of this PR's head
(`e3797b3`) on a disposable database. Ten passed — the order-side fix accepts `comments` and
`internalNotes`, writes them, clears them by `null` and by whitespace, honours the 4000-character
bound, lets the canonical plural win over the legacy singular, reports both to the audit log, and
leaves the existing inline **Comment** editor working. The failing one is the review's Major
finding, reproduced end to end: a `PUT /api/sales/quotes` whose only field is `internalNotes` returns
200 having written nothing, and the sent quote it was aimed at came back **draft** with its
`acceptance_token` and `sent_at` cleared. Screenshots of that quote before and after the call are in
the evidence comment.

No pipeline or meta label was changed by this run — it ran evidence-only, and `qa-failed` is
deliberately not applied without `--apply-failure`, since the PR already carries `changes-requested`
from the review and the QA verdict has the same root cause. `in-progress` was the only label this run
touched, and it has been removed; the PR remains assigned to @kamwro, exactly as the review left it.

kamwro:
Fixed in `3727bbc2a`. Thanks — the finding is right, and it reproduces exactly as described.

Measured before and after, an `internalNotes`-only update against a quote in `sent` status:

| | outcome | `acceptanceToken` | `sentAt` | `status` |
|---|---|---|---|---|
| `develop` (before this PR) | rejected (`update_payload_empty`) | preserved | preserved | `sent` |
| this PR, previous head | **resolved OK** | `null` | `null` | `draft` |
| this PR, `3727bbc2a` | rejected (400) | preserved | preserved | `sent` |

**One addition to your suggested fix.** The guard alone does not restore the third column above.
`updateQuoteCommand` clears `acceptanceToken` and `sentAt` at `documents.ts:5196`, *before*
`applyDocumentUpdate` is called, so a 400 raised inside `applyDocumentUpdate` arrives after the
loaded entity has already been stripped — the exact assertion you asked for
(`acceptanceToken`/`sentAt`/`status` left alone) fails with only the guard in place.

So this commit also moves that reset into the atomic phase, immediately after
`applyDocumentUpdate` and beside the existing `status = "draft"` write. A successful update behaves
as before — your `quote editing invalidates sent token` test covers that path and still passes — but
an update that fails validation no longer touches the token. Happy to split that into its own PR if
you would rather keep this one to the schema change.

The error key is `sales.quotes.internal_notes_unsupported`, added to all five locales alongside the
neighbouring `sales.orders.edit_addresses_blocked`.

**Tests.** Both cases you named are now pinned: the rejection, and a `sent` quote left intact by the
payload it rejects. 18 cases in `documents.update-notes.test.ts`; full `modules/sales` suite 584
passed.

**On the pre-existing `expectedDeliveryAt` twin** — confirmed against pristine `develop`, not
inferred: an `expectedDeliveryAt`-only update on a sent quote resolves successfully, writes nothing
(`SalesQuote` has no such column), and still clears the token and reverts to draft. Taking your
follow-up-issue route rather than widening this PR; I will link it here once filed, together with
the sibling fields from your third note.

**On the description** — you are right that no shipped code composes `documentUpdateSchema`; the
paragraph is corrected.

kamwro:
Follow-up filed as #5344 — covers the pre-existing `expectedDeliveryAt` case and the eight sibling fields from your third note.

kamwro:
@pkarw I cannot re-request review from this account — it has `read` here, so
`POST /pulls/5335/requested_reviewers` returns 404, and `/label` is the only slash command this
repo implements. Whenever it suits you, could you re-run the review against `3727bbc2a`, or
re-request on my behalf so the automation picks it up?

The workflow runs also need approving for the new head — the suite was fully green on `e3797b3`,
but a fork PR needs that approval per push, so `3727bbc2a` currently has no CI signal.

patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-18T14:53:32Z. Other auto-skills will skip this PR until the lock is released.

Reviewing the new head `3727bbc2a`. Claiming the `in-progress` label only — the PR stays assigned to @kamwro.

patzick (approved):
# 🔍 Code Review (re-review): fix(sales): accept comments and internalNotes on document update

## 🎯 Summary

Re-review of head `3727bbc2a`, now merged with `develop` as `2eb511e04`. **The Major finding from the previous pass is genuinely fixed**, and the fix is better than the one that was suggested — it closes a pre-existing hole the original finding only pointed at.

Two changes carry it:

- `packages/core/src/modules/sales/commands/documents.ts:1183` — a quote update carrying `internalNotes` now raises `CrudHttpError(400)` instead of being validated, counted as a real edit by the `.refine()`, and then dropped.
- `packages/core/src/modules/sales/commands/documents.ts:5216-5225` — the `acceptanceToken` / `sentAt` reset moved out of the pre-flight block and inside `withAtomicFlush`, *after* `applyDocumentUpdate`. As the author noted, the guard alone would not have restored this column: the old code cleared the token at the top of `execute`, so **any** rejected payload — not just the `internalNotes` one — stripped a sent quote of its acceptance link on its way to a 400. That is a wider fix than the review asked for, and the right one.

I verified the ordering is safe rather than merely tidy. `updateQuoteCommand` calls `withAtomicFlush(..., { transaction: true })` (`documents.ts:5334`), which `packages/shared/src/lib/commands/flush.ts` documents as all-or-nothing DB semantics. That matters because `applyDocumentUpdate` assigns `entity.comments` at `:1173` *before* the new quote guard throws at `:1183` — the rollback guarantees that in-memory mutation can never commit. The relative order of `status = "draft"` versus `applyDocumentUpdate` is unchanged, so no new ordering regression is introduced.

Three other claims in the PR body check out against shipped code:

- **The fix reaches HTTP.** `PUT /api/sales/{orders,quotes}` parses through the same `documentUpdateSchema` (`api/documents/factory.ts:536`), and `mapInput` consumes `raw`, not a pre-stripped `parsed` — so the two new keys survive to the command.
- **No OpenAPI drift.** `buildDocumentOpenApi` declares `create` and `del` only, no update body schema, so there is nothing to re-sync.
- **`buildDocumentUpdateChangeKeys` feeds only `buildLog`.** Its sole call sites are `documents.ts:5365` and `:5609`, both inside `buildLog`. It reaches neither reindex nor events, as stated.

The new `sales.quotes.internal_notes_unsupported` key is present in all five locale files (`de`, `en`, `es`, `ko`, `pl`) — and `i18n:check-sync` / `i18n:check-usage` ran green in CI, so that is machine-confirmed rather than eyeballed.

**Verdict: APPROVE.** No blocker, major or minor findings. Everything below is a nit.

## ✅ Validation

**Validation source: GitHub checks only** — per the task instruction, nothing was run locally. All checks pass on `2eb511e04`, none pending:

| Check | Result |
|---|---|
| `prepare` (build:packages → generate → rebuild → build:app) | ✅ pass (6m42s) |
| `test` (typecheck, jest, i18n:check-sync, i18n:check-usage, template parity) | ✅ pass (7m51s) |
| `ephemeral-integration (none)` | ✅ pass (12m55s) |
| `lint` / `ds-lint` / `scope` / `audit-scope` | ✅ pass |
| `audit` | ⏭️ correctly skipped — see below |

### A note on why CI looked dead before this pass

The previous head had `audit` **failing** on `deepmerge-ts <8.0.0` (GHSA-ggr8-5vv4-36mx), and `test`, `docker-build`, `ephemeral-integration` and `merge-coverage` all declare `needs: audit` — so the entire suite skipped and the PR had no test signal at all.

That was **not** this PR's doing. `audit-scope` computes `git diff origin/develop --name-only`, which is direction-blind: because the branch was behind, `yarn.lock` registered as "changed" and got audited at its stale pin. `develop` has carried `deepmerge-ts@8.0.1` since bab761d10. I updated the branch from `develop` (`2eb511e04`) and approved the fork workflow runs; `audit` now correctly reports `skipping`, and the real suite ran for the first time on this change. **No source change was needed to clear it.**

## 🔵 Nits (non-blocking, none need addressing in this PR)

**1. The legacy singular `comment` bypasses the new 4000-character cap.**
`documentUpdateSchema.comment` is `z.string().nullable().optional()` (`documents.ts:594`) with no bound, and `applyDocumentUpdate` writes it to the very same `comments` column that the new `comments` field caps at `.max(4000)`. A caller who wants to exceed the limit only has to use the older name. No data risk — the column is `type: 'text'` (`data/entities.ts:424`, `:885`) — but the contract now says two different things about one column depending on which spelling you pick.

**2. Quote *create* still silently strips `internalNotes`, which is the exact bug class this PR exists to fix.**
`quoteCreateSchema` (`data/validators.ts:752+`) declares `comments` but no `internalNotes`, and it is a plain stripping `z.object`. So `POST /api/sales/quotes` with `internalNotes` returns 201 having ignored it, while `PUT` on the same document now returns 400. The update half is honest; the create half is not. Reasonable to fold into #5344 rather than widen this PR.

**3. Rejecting `internalNotes` on quotes is technically an API behaviour change.**
A `PUT /api/sales/quotes` payload carrying that key used to return 200 (silently stripped) and now returns 400. I checked the blast radius: `internalNotes` appears nowhere in the sales backend UI or portal — only in seed data and the order create/update paths — so no shipped caller is affected and no UI breaks. It is third-party API consumers only, and trading a silent no-op for a loud rejection is the right call. Worth one line in `UPGRADE_NOTES.md` on the strength of `BACKWARD_COMPATIBILITY.md` listing API routes as a contract surface.

**4. Test placement.** The two quote-guard cases (`documents.update-notes.test.ts:253`, `:259`) live under `describe('sales.orders.update — note columns')`. They drive `sales.quotes.update`.

**5. Carried from the previous pass.** The PR body's SPEC-021 paragraph still describes a composition that does not exist in shipped code. Purely cosmetic — the other two nits from that pass were filed as #5344, which is the right disposition.

## 📋 QA

`needs-qa` **stays on**. The previous QA run failed against `e3797b3` on exactly the quote path this PR now changes, so the fix needs to be re-verified in a browser against the new head before it can carry `qa-approved` — a green unit suite is not a substitute for that specific scenario. The follow-up integration test suggested in the earlier QA comment is still the durable answer.

This PR must not be merged while it carries `needs-qa` without `qa-approved`.

patzick:
🏷️ label rationale — `om-auto-review-pr` (re-review)

- `changes-requested` → `review` — the Major finding that drove the previous verdict is fixed at `3727bbc2a`, and the fix is broader than what was asked for: moving the `acceptanceToken`/`sentAt` reset inside `withAtomicFlush` also stops *any* rejected payload (not just the `internalNotes` one) from stripping a sent quote of its acceptance link. Re-review is APPROVE with nits only.
- `needs-qa` — **kept.** The previous QA run failed against `e3797b3` on exactly the quote path this PR now changes. A green unit suite is not a substitute for re-verifying that scenario in a browser, so this still needs `qa-approved` from a QA reviewer before it can merge.
- `risk-medium` — unchanged. `documentUpdateSchema` is still shared by both update commands, and this head additionally reorders a state mutation inside the quote transaction.
- `priority-medium`, `bug` — unchanged from the previous pass.
- `in-progress` — released.

`partner-request` left as set by @patrykbojczuk.

patzick:

No blocker/major/minor findings on `2eb511e04`; five nits are informational and none need addressing in this PR (the sharpest: the legacy singular `comment` is still unbounded while the new `comments` is capped at 4000, so the cap is bypassable through the older spelling — harmless, the column is `text`).

**Validation source: GitHub checks only** — nothing was run locally, per the review request. All checks pass on `2eb511e04`, none pending: `prepare` ✅ 6m42s, `test` ✅ 7m51s, `ephemeral-integration (none)` ✅ 12m55s, `lint`/`ds-lint`/`scope`/`audit-scope` ✅.

**One CI fix was pushed as part of this pass.** @kamwro — this answers the "`3727bbc2a` currently has no CI signal" note, and it was not a fork-approval problem alone. `audit` was *failing* on `deepmerge-ts <8.0.0`, and `test`, `docker-build`, `ephemeral-integration` and `merge-coverage` all declare `needs: audit`, so the whole suite skipped. The cause was a stale branch, not your change: `audit-scope` computes `git diff origin/develop --name-only`, which is direction-blind, so a branch that is merely *behind* develop has its `yarn.lock` register as changed and audited at the old pin. `develop` has carried `deepmerge-ts@8.0.1` since bab761d10. I updated the branch from `develop` and approved the fork workflow runs; `audit` now correctly skips and the real suite ran for the first time on this change. No source change was needed.

Worth knowing for future PRs from this fork: whenever `audit` fails on an advisory you did not introduce, updating the branch from `develop` is usually the whole fix.

Next step is QA, not merge — the `needs-qa` + no `qa-approved` combination is a hard merge block, and the previous QA failure was on the quote path this head changes.

Issue #891, opened 2026-03-08
bug: inbox-ops - create default variants with product create action


pkarw:
## Summary

The default variant is required for quotes to work. Inbox ops nicely creates products but with no variants and then when editing the quote i've got the issue like this:

<img width="1102" height="746" alt="Image" src="https://github.com/user-attachments/assets/015f992e-8d46-40ce-8909-323d438ad26e" />

pkarw:
Let have the create product creatign the default variant if not defined and just make sure using this command

PR #5337, opened 2026-08-17
feat(wms): seed example warehouses, locations, and opening stock

### LINKED ISSUE #5237: feat: seed data for the WMS
Please add the example seeding data for the WMS module (as we do for all the other modules)

@mkadziolka

mkadziolka:
## Summary
- Adds WMS `seedExamples` (and `mercato wms seed-examples`) so new tenants get the English walkthrough dataset: NYC / LAX / Berlin warehouses, location topology, inventory profiles, and opening receipts including FEFO matcha lots.
- Reuses catalog fashion variants (Atlas Runner, Aurora Wrap) when they already exist; does not copy local QA noise and does not overwrite existing balances or a tenant’s primary warehouse.
- Closes #5237

## Test plan
- [ ] `yarn workspace @open-mercato/core test src/modules/wms/seed/__tests__/examples.test.ts src/modules/wms/__tests__/setup.test.ts`
- [ ] On a tenant with catalog examples: `yarn mercato wms seed-examples --tenant <id> --org <id>` and confirm DEMO-NYC/LAX/BER plus mug/matcha/earbuds stock
- [ ] Re-run the same command and confirm it is a no-op (no extra warehouses or ledger rows)
- [ ] Optional: `mercato init --with-examples` on a fresh tenant and confirm WMS examples appear with the other modules

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5337"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789552210&installation_model_id=432243&pr_number=5337&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5337&signature=b65db9470bf1c2d24eae16b118351c18395a179697657b322888bda63cfa4839"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T10:48:41Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
## Code Review: feat(wms): seed example warehouses, locations, and opening stock

**Verdict: CHANGES REQUESTED** (1 minor finding — this repository requests changes on Medium/minor too)

**Validation source:** local fallback. GitHub reported no usable check data for this head — `get-required-checks` returns 404 (required status checks not enabled on `develop`) and the only check on head `9a7fe45` is `[code]smith` = `SKIPPED`; no workflow run exists for that SHA (fork PR). Per the repo's GitHub-checks-first rule, the full local gate was run instead.

### Validation gate (`.ai/agentic.config.json` → `validation.commands`)

| Command | Result |
|---|---|
| `yarn generate` | ✅ pass (no generated-file churn — `git status` clean) |
| `yarn build:packages` (post-generate) | ✅ pass |
| `yarn i18n:check-sync` | ✅ pass |
| `yarn typecheck` | ✅ pass (22/22 tasks) |
| `yarn test` | ⚠️ 28/29 packages pass — `@open-mercato/core` 9526/9526, `@open-mercato/shared` 1835/1835, including the two new suites (`wms/seed/__tests__/examples.test.ts`, `wms/__tests__/setup.test.ts`). The single failure is `create-mercato-app` → `template-dev-log-files.test.ts`, which asserts on host `fs.inotify` sysctl limits (`max_user_instances: 128 < 4096`). **Environment limitation of the review machine, not this PR** — the diff touches no `packages/create-app` file. |
| `yarn build:app` | ✅ pass |

### What was verified beyond the gate

- **Duplicate/already-merged work:** none. All six files are new or purely additive against `develop`; no equivalent WMS seed exists on the base branch.
- **Contract surfaces (`BACKWARD_COMPATIBILITY.md`):** clean. No entity/schema change (so no migration is owed), no removed export or import path, no event ID or widget spot ID touched, no API route or response field altered. `mercato wms seed-examples` is an additive CLI command; `setup.seedExamples` is an additive lifecycle hook implementation.
- **Tenant scoping:** every read and write is filtered on both `organizationId` and `tenantId`. No cross-tenant exposure. `resolvePerformedBy` correctly uses `findOneWithDecryption` with an explicit `{ tenantId, organizationId }` decryption scope for the encrypted `User` entity; the WMS and catalog entities it reads raw declare no `encryption.ts` map, so plain `em.find` is correct there.
- **Idempotency:** re-running is genuinely a no-op — warehouses/zones/locations keyed by code, products/variants by SKU, profiles by `catalogProductId` + `catalogVariantId: null`, opening stock by the `wms-examples-v1:...` movement idempotency key (which matches the per-organization `wms_inventory_movements_idempotency_unique_idx`). All 23 `STOCK_SEEDS` rows are unique on `(warehouse, location, variant, lot)`, so the deferred flush in `ensureOpeningStock` cannot produce a duplicate balance. Covered by a test.
- **Ledger consistency:** each seeded `InventoryBalance` ships with exactly one matching `receipt` movement (`locationTo` = the balance location, same quantity), which is what `recomputeBalanceFromMovements` buckets on — so `mercato wms verify-balances` reports no drift on freshly seeded data. Good.
- **Primary-warehouse safety:** `DEMO-NYC` only becomes primary when the org has none. Covered by a test.
- **Flush ordering:** no read is interleaved between a scalar mutation and its flush in a way `withAtomicFlush` would guard — the only deferred persists are brand-new entities on the UnitOfWork persist stack, and every lookup key is unique per iteration.
- **Cross-module coupling:** importing `CatalogProduct`/`CatalogProductVariant` is permitted here — `wms` declares `requires: ['catalog', 'sales', 'feature_toggles']`. Module registration order puts `catalog` (11th) before `wms` (13th), so `mercato init --with-examples` seeds the fashion variants first; the four SKUs referenced (`ATLAS-RUN-NAVY-8`, `ATLAS-RUN-GLACIER-10`, `AURORA-ROSE-M`, `AURORA-CELESTIAL-L`) match `catalog/seed/examples.ts` exactly, and the `optionalCatalog` path degrades gracefully when they are absent.
- **Query index:** no explicit reindex needed — the query engine reads the source tables (it only touches `entity_indexes` for custom-field doc filters), and `mercato init` triggers a vector + query index reindex after seeding.
- **Conventions:** CLI command mirrors `verifyBalancesCommand` and the catalog CLI exactly (arg parsing, container dispose guard, `em.transactional`); the internal throw in `orderLocationSeeds` is correctly `[internal]`-prefixed per the i18n opt-out rule; no `any`, no one-letter names, no hardcoded status colors, no UI surface touched.
- **Test coverage:** 241 + 29 new lines covering seed-data integrity, topological location ordering, idempotency, primary-warehouse preservation, optional-catalog reuse, the no-user skip path, and the setup delegation. Solid.

---

## Findings

### 🟡 Minor — multi-variant demo products are written as `productType: 'simple'` / `isConfigurable: false`

`packages/core/src/modules/wms/seed/examples.ts:228,233` hardcodes `productType: 'simple'` and `isConfigurable: false` for every product in `PRODUCT_SEEDS`, but two of the three carry **two named variants** each:

- `HOME-MUG-POUR` → `HOME-MUG-SLATE` (Slate Grey), `HOME-MUG-CLAY` (Clay White)
- `ELEC-EARBUDS-NC` → `ELEC-EARBUDS-GRAPHITE` (Graphite), `ELEC-EARBUDS-PEARL` (Pearl)

That contradicts the catalog module's own convention: `catalog/seed/examples.ts:877,884` writes `productType: "configurable"` / `isConfigurable: true` for its multi-variant example products, and a `simple` product is expected to carry a single default variant (see `catalog/inbox-actions.ts:45`, issue #891). `isConfigurableProductType` (`components/products/productForm.ts:420`) only treats `configurable | virtual | downloadable` as configurable.

Observable consequences in the seeded demo data:

- The Products grid (`ProductsDataTable.tsx:445-449`) renders **Type = Simple** and **Configurable = ✗** for two products that visibly own two variants each.
- `GET /api/catalog/products?configurable=true` (`api/products/route.ts:178-180`) excludes them.
- The product edit page derives `hasVariants` from `is_configurable` (`backend/catalog/products/[id]/page.tsx:733`), so the "has variants" toggle sits off while the Variants section lists two variants — and changing the type later resets the toggle (`page.tsx:2679-2683`).

Not destructive (variants are managed through their own API, so nothing is dropped on save), but it is exactly the kind of inconsistency a walkthrough dataset should not teach.

**Suggested fix** — add the flag to the seed shape rather than hardcoding it, e.g. in `examples-data.ts` give `ProductSeed` an optional `productType?: CatalogProductType`, or simply derive it:

```ts
const configurable = seed.variants.length > 1
product = em.create(CatalogProduct, {
  // ...
  productType: configurable ? 'configurable' : 'simple',
  isConfigurable: configurable,
  // ...
})
```

`FOOD-MATCHA-TIN` (single variant) correctly stays `simple`.

### ⚪ Nit — WMS-created catalog products ship without price, category, or offer

`ensureDemoProducts` creates the three products with `primaryCurrencyCode: 'USD'` but no `CatalogProductPrice`, no category assignment, and no offer, unlike `catalog/seed/examples.ts` which seeds all three for its example products. The result is demo products that appear in the catalog with blank pricing and no category, and that cannot be exercised end-to-end in the WMS↔sales walkthrough (the sales-order stock widget) without manually adding a price first. Optional polish, but the linked issue asks for WMS examples "as we do for all the other modules", and this is the one place the dataset falls short of that bar.

---

## Note on autofix

The findings are minor/nit only and this is a **fork PR** (`mkadziolka/open-mercato`). Rather than carry the branch forward into a replacement PR and close yours over a two-line demo-data fix, I'm handing it straight back — the fix is small and the credit should stay on this PR. Push the change and re-request review, and the automation will pick it up from the new head.

pkarw:
Label moved back to `changes-requested` (from `merge-queue`) because CI for head `dc6b408` finished red after the review was submitted: `ephemeral-integration` failed with 2 WMS integration tests (`TC-WMS-INVENTORY-UI-001`), and a failing required check is a blocker regardless of what the diff review found. Diagnosis and the re-run I started: https://github.com/open-mercato/open-mercato/pull/5337#issuecomment-5379959221

The code-review findings themselves have not changed — the `configurable` fix from `dc6b408` is correct and covered by tests, and nothing at or above minor was found in the diff.

`skip-qa` is kept because the automated-verification exemption applies: the diff touches no UI-rendering file (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`), changes no database structure and no API surface, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships automated tests for the changed behavior in the same PR.

`priority-medium` is kept because this is an ordinary additive feature — demo/example seeding — with no release-blocking or security dimension. `risk-medium` is kept, and the CI failure is a good illustration of why it is not `risk-low`: the seed hook runs inside every `mercato init`, which includes the ephemeral integration environment, so its blast radius reaches the shared test tenant and not just a developer's demo database. `feature` is kept as the category.

pkarw:
Thanks @mkadziolka — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pkarw:

Full validation gate run locally (no CI check data exists for this fork head): `build:packages`, `generate`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `build:app` all pass; `test` passes 28/29 packages (core 9526/9526, shared 1835/1835, both new WMS suites green) with the only failure being a `create-mercato-app` test that asserts on the review host's `fs.inotify` sysctl limits — unrelated to this diff. One minor finding (multi-variant demo products written as `productType: 'simple'` / `isConfigurable: false`) and one nit. No autofix commit was pushed and no replacement PR opened: the findings are minor/nit only and this is a fork PR, so the work stays with the original author.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T11:04:40Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review (Re-review): feat(wms): seed example warehouses, locations, and opening stock

## 🎯 Summary

This is a re-review of head `dc6b408`, which adds one commit on top of the head I reviewed on 2026-08-17 (`9a7fe45`): `fix(wms): mark multi-variant example products as configurable (#5237)`. That commit addresses the single minor finding from the previous pass exactly as suggested — `ensureDemoProducts` now derives `const configurable = seed.variants.length > 1` and writes `productType: configurable ? 'configurable' : 'simple'` together with the matching `isConfigurable` flag, instead of hardcoding `'simple'` / `false` for every product (`packages/core/src/modules/wms/seed/examples.ts:220,229,233`). The change ships with assertions of its own: `packages/core/src/modules/wms/seed/__tests__/examples.test.ts:143-156` now pins `HOME-MUG-POUR` and `ELEC-EARBUDS-NC` to `configurable` / `true` and `FOOD-MATCHA-TIN` (the genuinely single-variant product) to `simple` / `false`, so the regression cannot silently come back.

I re-verified the fix against the catalog module's own convention rather than taking it on trust. `catalog/seed/examples.ts:877,884` writes `productType: "configurable"` / `isConfigurable: true` for its multi-variant example products **without** attaching an option schema, and `optionSchemaId` is nullable and optional throughout (`catalog/data/validators.ts:220`), so a configurable product with no schema is a supported, first-class state and not a half-built record. I also traced the product edit page's variant matrix, which was my main worry about flipping the flag: `buildOptionValuesKey` returns `""` for a variant with no `optionValues` (`catalog/components/products/productForm.ts:606-614`), those empty keys are filtered out of `existingKeys`, and `buildVariantCombinations` is fed from `values.options`, which is empty for these products — so `missingCombos` stays empty and the builder renders nothing to generate (`catalog/backend/catalog/products/[id]/page.tsx:2191-2215`). The flip therefore fixes the grid's Type/Configurable columns and the `?configurable=true` filter without introducing a broken variant-builder state.

The rest of the diff is byte-identical to the head I already reviewed in full, and the base branch has not grown any overlapping WMS seed work in the meantime (`develop` still has no `packages/core/src/modules/wms/seed` directory), so the earlier verification of tenant scoping, idempotency, ledger consistency, primary-warehouse safety, cross-module coupling, and contract surfaces stands unchanged. The PR is `MERGEABLE` against the current `develop`.

## Verdict

✅ **approve** — the one minor finding from the previous review is fixed correctly and is now covered by a test, and re-verifying it turned up no new problem. Only the previous 💅 nit remains open, and a nit does not block on its own (this repository requests changes on minor findings too, so an approve here genuinely means nothing at or above minor is outstanding).

## 🧪 Validation Gate

**Validation source: GitHub checks** (pending — no local fallback run).

Unlike the previous pass, this fork head *does* have a CI run: `CI for Develop&Main` (run [32567294217](https://github.com/open-mercato/open-mercato/actions/runs/32567294217)). Required status checks are still not enabled on `develop` (`get-required-checks` → 404), so per the tracker descriptor every reported check is treated as required.

| Check | State |
|---|---|
| `scope` | ✅ SUCCESS |
| `audit-scope` | ✅ SUCCESS |
| `lint` | ✅ SUCCESS |
| `ds-lint` | ✅ SUCCESS |
| `license/cla` | ✅ SUCCESS |
| `prepare` | ⏳ IN_PROGRESS |
| `audit`, `mutate`, `comment`, `[code]smith` | ⏭️ SKIPPED |

**No check is failing.** `prepare` is still running, and the typecheck / test / build jobs of that lane are gated behind it. Per this repository's GitHub-checks-first rule, a merely *pending* check is not a reason to run local substitutes or to hold the review, so I did not re-run `validation.commands` locally on this head. For the record, the full gate (`build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `test`, `build:app`) passed locally on the parent commit `9a7fe45` during the previous review — including `@open-mercato/core` at 9526/9526 with both new suites — and the delta since then is a two-line seed change plus twelve lines of new assertions in a suite that already ran green.

⏳ **CI is still pending on this head.** This review does not wait for it: the verdict above is about the code. I will follow up with the CI result on this PR as a separate comment once the run settles, or report that the wait was abandoned if it exceeds the configured budget. **The approval is not permission to merge past a red check** — if `prepare` or anything downstream of it fails, that has to be resolved before merge regardless of this review.

## Findings

### 💅 Nit

**`packages/core/src/modules/wms/seed/examples.ts:213-241` — WMS-created catalog products still ship without price, category, or offer.** Carried over unchanged from the previous review, and still the author's call. `ensureDemoProducts` creates the three demo products with `primaryCurrencyCode: 'USD'` but no `CatalogProductPrice`, no `CatalogProductCategoryAssignment`, and no offer, where `catalog/seed/examples.ts` seeds all three for its example products. The result is demo products that show blank pricing and no category in the catalog, and that cannot be driven end-to-end through the WMS↔sales walkthrough (the sales-order stock widget) without someone adding a price by hand first. Issue #5237 asks for WMS examples "as we do for all the other modules", and this is the one place the dataset stops short of that bar. Not a blocker for shipping the warehouse/stock dataset, which is what the PR is actually about.

**`packages/core/src/modules/wms/seed/examples.ts:250-262` — the now-configurable products' variants carry no `optionValues`.** A small consistency point that the fix commit surfaces rather than causes. `catalog/seed/examples.ts:961` gives each variant of a configurable product an `optionValues` map (`{ color: "Midnight Navy", size: "US 8" }`), so the demo catalog teaches "configurable product ⇒ variants distinguished by named option values". The WMS variants are distinguished only by `name` (`Slate Grey` / `Clay White`, `Graphite` / `Pearl`). As traced in the summary this breaks nothing — the variant builder degrades cleanly — but widening `ProductSeed['variants']` (`examples-data.ts:42`) with an optional `optionValues?: Record<string, string>`, filling it in for the two multi-variant entries in `PRODUCT_SEEDS`, and passing it through at the `em.create(CatalogProductVariant, …)` call would make the seeded data match what the catalog's own walkthrough demonstrates. Two lines of data, zero logic.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff is purely additive (`seedWmsExamples`, `seedExamplesCommand`, `setup.seedExamples`).
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — no API surface is touched.
- [x] No CLI command or flag renamed or removed — `mercato wms seed-examples` is a new command appended to `cliCommands` alongside the existing `verify-balances`.
- [x] No database table or column renamed or removed; no column type narrowed — no entity changed, so no migration is owed.
- [x] Where a contract had to change: not applicable, nothing had to change.

## 🧪 Test Coverage

The new commit's coverage is proportionate and precisely targeted. `examples.test.ts:143-156` asserts the seeded `CatalogProduct` rows for all three demo SKUs, distinguishing the two multi-variant products (`configurable` / `true`) from the single-variant matcha tin (`simple` / `false`) — which means it pins both directions of the new `seed.variants.length > 1` branch, not just the fixed case. That is exactly the shape a regression test for this bug should have.

The pre-existing suites still carry the rest: seed-data integrity, topological location ordering, movement-key idempotency, primary-warehouse preservation, optional-catalog reuse, the no-user skip path, and the `setup.seedExamples` delegation.

The one gap worth naming, and it is small: nothing asserts the *reuse* path of `ensureDemoProducts` — that a second run over a tenant where `HOME-MUG-POUR` already exists as `simple` leaves the existing row untouched rather than upgrading it to `configurable`. That is the current (and correct, non-destructive) behavior, since the `productType` assignment sits inside the `if (!product)` branch; a tenant seeded before this PR keeps its `simple` products. If that is intended — and I think it should be, a seed has no business rewriting records a user may have edited — a one-line assertion in the existing idempotency test would document the decision instead of leaving it implicit.

---

## Note on autofix

`autofix: skipped (not my PR — re-run with --autofix to fix it here)`. This is a fork PR authored by @mkadziolka and the run carried no `--autofix`, so nothing was pushed to your branch. Nothing needed pushing in any case: the remaining items are nits and the verdict is an approve.

pkarw:

Re-review of head `dc6b408`. The minor finding from the 2026-08-17 pass — multi-variant demo products written as `productType: 'simple'` / `isConfigurable: false` — is fixed by deriving the flag from `seed.variants.length > 1`, and the fix is pinned in both directions by new assertions in `wms/seed/__tests__/examples.test.ts`. I re-verified it against the catalog module's own convention (`catalog/seed/examples.ts` marks its multi-variant examples configurable with no option schema, and `optionSchemaId` is nullable) and traced the product edit page's variant matrix to confirm the flip does not leave a broken builder state. Only the previous nit remains, plus one new nit about the now-configurable variants carrying no `optionValues`; neither blocks.

Validation source: GitHub checks. No check is failing on this head; `prepare` was still in progress, so per this repo's GitHub-checks-first rule no local substitute gate was run. `ci-monitoring` is applied while I watch that run — it is a meta label, not a claim, so this PR is free for anyone to act on. I will post the CI result here when the run settles.

Autofix: skipped (not my PR — re-run with `--autofix` to fix it here). Nothing needed fixing in any case.

pkarw:

**The CI run for head `dc6b408` finished red, so I am correcting the verdict I posted earlier: `merge-queue` → `changes-requested`, and I have submitted a changes-requested review that supersedes my approval.** The code review itself still stands — the `configurable` fix is correct and well tested — but a failing required check is a blocker on its own, and I approved while this lane was still pending. That was the wrong call to leave standing once it went red.

## What failed

`CI for Develop&Main` — run [32567294217](https://github.com/open-mercato/open-mercato/actions/runs/32567294217). Everything else is green: `prepare`, `lint`, `ds-lint`, `scope`, `audit-scope`, `test` (unit), `documents-multi-instance`, `license/cla`. Branch protection exposes no required-check list (`404`), so per the tracker descriptor every reported check counts as required.

`ephemeral-integration (none)` — **2 failed, 70 passed**, both failures in the same spec, both failing again on Playwright's automatic retry:

- `TC-WMS-INVENTORY-UI-001` › *posts a positive adjust and refreshes balances in the console* (`spec.ts:37`)
- `TC-WMS-INVENTORY-UI-001` › *runs the 3-step cycle count wizard and updates balances and movements* (`spec.ts:151`)

## What actually goes wrong

Both failures have the same underlying shape, and it is not the assertion the log leads with. From the failure snapshots in the `integration-test-results` artifact:

**Adjust dialog** — the variant combobox holds its value, but warehouse and location are empty and both carry a validation error:

```
- combobox "Search variant or SKU": TCIUI-V-ee6b64ba
- combobox "Select warehouse"
- alert: Invalid UUID
- combobox "Select location"
- alert: Invalid UUID
```

**Cycle count wizard** — same thing one step earlier:

```
- combobox "Select warehouse"
- text: Zone
- combobox "Select warehouse first" [disabled]
- paragraph: Select a warehouse to choose a zone.
```

So the reported `getByPlaceholder('Select zone')` "element(s) not found" is a symptom, not the cause: `CycleCountWizardDialog.tsx:1237-1245` only renders the `Select zone` placeholder when `form.warehouseId` is non-empty, and it is empty. **In both tests the warehouse combobox never commits an id.** The test helper's `expect(input).toHaveValue(warehouseName)` passes because it is satisfied by the text the helper itself typed — with `allowCustomValues={false}` that text proves nothing was selected — and the value then disappears on blur when the control re-syncs its label from the (still empty) committed value.

## Why I think this PR is the leading suspect

I want to be straight about the confidence level here: I have a strong circumstantial case and no direct proof yet.

**The case for:** `develop` ran the full 15-shard integration suite at 09:45 today (run [32565791335](https://github.com/open-mercato/open-mercato/actions/runs/32565791335), head `413df595`) and every shard passed, including this spec. The only delta on this head is the six files in this PR. And this PR is not inert in that environment: the ephemeral harness provisions its database by running `yarn initialize` → `mercato init` (`packages/cli/src/lib/testing/integration.ts:3430`), and `mercato init` seeds examples **by default** — `--no-examples` is the opt-out (`packages/cli/src/mercato.ts:887`). So the new `setup.seedExamples` hook now runs in every integration environment, and the test tenant gains `DEMO-NYC` / `DEMO-LAX` / `DEMO-BER`, their zone and bin topology, three products, and 23 opening-stock rows before a single spec executes. `DEMO-NYC` also claims `isPrimary` on a tenant that previously had no warehouse at all until each test created its own.

That is a real change in the shared fixture these tests run against, and "the tenant now has four warehouses instead of one" is exactly the kind of thing that flips a warehouse-combobox interaction from trivially-correct to racy — for instance because the suggestion dropdown no longer resolves to a single obvious match within the helper's 2-second window.

**What I have not proved:** I could not pin the exact mechanism by reading the code. `resolveWarehouseLabel` / `resolveLocationLabel` (`inventoryMutationLoaders.ts:55-79`) look robust to extra rows, and I found no sole-warehouse preselect in these two dialogs that the seed would be disabling. There is also a second difference between the two runs that has nothing to do with your code: this run was unsharded and module-scoped (`OM_INTEGRATION_MODULES=audit_logs,wms`), so the WMS specs ran alongside a different set of neighbours, in different parallel workers, against one shared database, than they do in develop's 15-way split. Cross-test interference — something else mutating the admin role's ACL while `ensureRoleFeatures` is in effect, say — would produce exactly this "the API stopped answering, so the combobox never resolved" picture too.

**So I re-ran the failed job** ([`gh run rerun --failed`](https://github.com/open-mercato/open-mercato/actions/runs/32567294217)) to separate those two explanations, and I will post the outcome here. A second identical failure points at the seed data; a green re-run points at scheduling/interference and this PR is in the clear.

## What I suggest

If the re-run confirms it, the fix is probably not in the seed logic but in how much it assumes about an empty tenant. Worth considering, in rough order of preference:

1. **Do not let `seedExamples` claim the primary warehouse on a tenant that has none** — or more broadly, reconsider whether a demo dataset should land in the CI database at all. The integration harness inherits `mercato init`'s defaults; if example data is meant for humans exploring the app, the harness passing `--no-examples` would keep both concerns separate and is a one-line change on the harness side.
2. **Harden `fillCombobox`** so it asserts a *committed* selection rather than the typed text — the current `toHaveValue` check cannot distinguish "option selected" from "characters typed", which is why both tests failed ten seconds later at an unrelated-looking assertion instead of at the real one. That is a pre-existing weakness in `helpers/wmsUi.ts` and worth a follow-up regardless of how this lands.

Happy to be told this is not yours to fix — if the re-run comes back green I will say so plainly and move the label back.

---

@mkadziolka handing this back to you pending the re-run result. Nothing is wrong with the code change you pushed; the question is only whether seeding the demo warehouses into every `mercato init` destabilizes the WMS integration specs.

Validation source: GitHub checks. `ci-monitoring` stays on while the re-run is in flight; I will remove it when I post the outcome.

pkarw (changes requested):
# 🔍 Code Review (verdict correction): CI is red on this head

This supersedes my approval of `dc6b408` earlier today. **The code review findings are unchanged** — the `configurable` fix is correct, well tested, and I found nothing at or above minor in the diff. What changed is the CI signal I approved past: the `CI for Develop&Main` run for this head has since finished, and `ephemeral-integration` failed with 2 failing WMS integration tests (70 passed). A failing required check is a blocker on its own, so the verdict has to be `changes-requested` until that lane is green.

## ⛔ Blocker

**`ephemeral-integration (none)` is failing on this head** — run [32567294217](https://github.com/open-mercato/open-mercato/actions/runs/32567294217). Both failures are in `packages/core/src/modules/wms/__integration__/TC-WMS-INVENTORY-UI-001.spec.ts` (`:37` and `:151`), both reproduced on Playwright's retry, and both come down to the warehouse combobox never committing an id — the adjust dialog ends up with `Invalid UUID` on warehouse and location, and the cycle-count wizard never renders its zone field because `form.warehouseId` is empty.

The full diagnosis, the failure snapshots, why this PR is the leading suspect (the ephemeral harness seeds examples by default, so `setup.seedExamples` now populates the CI tenant with three demo warehouses before any spec runs), the competing explanation I could not rule out, and the re-run I kicked off to separate the two are in [this comment](https://github.com/open-mercato/open-mercato/pull/5337#issuecomment-5379959221).

I will post the re-run outcome and move the label back to `merge-queue` myself if it comes back green.

pkarw:
🤖 `om-auto-review-pr` — CI result (re-run)

**The re-run failed identically, so this is deterministic, not a flake or a scheduling artifact.** `changes-requested` stands.

Fresh runner, fresh ephemeral database, fresh worker scheduling — same two tests, same errors, same counts:

| | first run | re-run |
|---|---|---|
| `TC-WMS-INVENTORY-UI-001:37` (adjust) | ❌ `Inventory adjusted` toast never appears | ❌ identical |
| `TC-WMS-INVENTORY-UI-001:151` (cycle count) | ❌ `Select zone` never renders | ❌ identical |
| totals | 2 failed, 70 passed, 0 flaky | 2 failed, 70 passed, 0 flaky |

Counting Playwright's automatic retries that is four failed attempts across two independent database provisionings. The competing explanation I raised in my previous comment — cross-test interference from the unsharded, module-scoped run pairing the `wms` specs with `audit_logs` specs in different workers than develop's 15-way split — does not survive that. Interference would not reproduce byte-identically twice.

## Where that leaves the attribution

The circumstantial case against this PR is now as strong as it can get without a local bisect:

1. **`develop` is green on this exact spec.** Run [32565791335](https://github.com/open-mercato/open-mercato/actions/runs/32565791335) (head `413df595`, 09:45 today) passed all 15 shards. Because `actions/checkout` resolves a `pull_request` event to the merge ref, this PR's run tested develop's tree *plus* your six files — so the six files are the only code delta between green and red.
2. **Those six files are not inert in this environment.** The ephemeral harness provisions its database with `yarn initialize` → `mercato init` (`packages/cli/src/lib/testing/integration.ts:3430`), and example seeding is on by default — `--no-examples` is the opt-out (`packages/cli/src/mercato.ts:887`). Your new `setup.seedExamples` hook therefore runs before any spec, and the test tenant now starts with `DEMO-NYC` (claiming `isPrimary`), `DEMO-LAX` and `DEMO-BER`, their zone/bin topology, three products and 23 opening-stock rows — where previously it started empty and each test created the only warehouse in existence.
3. **The failure is deterministic.** Established above.

What I still cannot hand you is the mechanism. Both dialogs fail because the warehouse combobox never commits an id — the adjust dialog ends with `Invalid UUID` on warehouse and location, and the cycle-count wizard never renders its zone field because `form.warehouseId` is empty (`CycleCountWizardDialog.tsx:1237`). But `loadWarehouseOptions` (`wmsLookupLoaders.ts:42`) searches server-side with `pageSize: 50`, and `resolveWarehouseLabel` / `resolveLocationLabel` (`inventoryMutationLoaders.ts:55-79`) look up by explicit id — none of which is obviously sensitive to three extra rows. I could not close that gap by reading the code, and I would rather tell you that than invent a mechanism that fits.

## Reproducing it locally

Now that it is deterministic, this should reproduce in one shot on your machine:

```bash
yarn test:integration packages/core/src/modules/wms/__integration__/TC-WMS-INVENTORY-UI-001.spec.ts
```

The harness runs `mercato init` for you, so the demo dataset will be there. The fastest confirmation of attribution is to run the same command with your `seedExamples` hook temporarily removed from `wms/setup.ts`: green means the seed is the cause, red means we are both chasing the wrong thing and I owe you an apology and a re-open of the investigation.

## Suggested directions

1. **Keep the demo dataset out of the automated-test tenant.** The cleanest fix is probably on the harness side rather than in your seed — having `test:integration` pass `--no-examples` keeps "data for humans exploring the app" and "fixtures for deterministic specs" from sharing a database. That is arguably a pre-existing design gap that your PR is the first to expose, and it may be worth its own issue rather than being loaded onto this one.
2. **Do not claim `isPrimary` for `DEMO-NYC`** if a cheaper fix is wanted first — that is the single seeded fact most likely to change how an existing UI behaves on a tenant that previously had no primary warehouse.
3. **Separately, `fillCombobox` is too weak to catch this class of bug** (`wms/__integration__/helpers/wmsUi.ts`). Its `expect(input).toHaveValue(value)` is satisfied by the text the helper itself typed, which with `allowCustomValues={false}` proves nothing was selected — which is why both tests failed ten seconds later at an assertion that looks unrelated to the real problem. Worth a follow-up regardless of how this PR lands.

I am removing `ci-monitoring` now — this was the follow-up it promised, and no further automated follow-up will come from this agent on this head. Push a new commit when you have something to test and re-request review, and the automation will pick it up from the new head.

Validation source: GitHub checks (runs [32567294217](https://github.com/open-mercato/open-mercato/actions/runs/32567294217) original + re-run).

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-22T13:05Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

**On the existing claim signal:** this PR is assigned to @mkadziolka, but that is the review hand-back from 2026-08-17, not a live automation claim — there is no `in-progress` label, and the newest `🤖` claim comment (mine, from the 2026-08-22 re-review) is already released. Under the stale-lock rule that assignment is well past the 30-minute window with no commits, comments, or label changes from the assignee since, so I am taking the lock rather than stopping. @mkadziolka stays assigned; I am only adding myself for the duration of this run.

**What this run is for:** the last CI result on head `dc6b408` was two deterministic failures in `TC-WMS-INVENTORY-UI-001` (adjust dialog + cycle-count wizard), and my own review comment admitted it could not pin the mechanism from reading the code. This QA pass drives those two dialogs in a real browser against a tenant seeded by this PR's `seedExamples` hook, so the outcome is either a reproduction with a concrete mechanism or a clean bill of health with screenshots to back it.

pkarw:
## 📸 UI QA evidence — ❌ FAIL (but not for the reason the label implies — please read the attribution)

**Verdict:** ❌ FAIL — the seeded dataset this PR adds is correct and renders correctly everywhere I looked, but on a tenant carrying it the WMS inventory console has two flows a human cannot complete, and one of them is exactly the red CI lane. I reproduced that lane in a real browser and pinned the mechanism, which my earlier review said it could not do.
**Environment:** `http://127.0.0.1:34679` · role `admin` (`admin@acme.com`, demo credentials from the test-env descriptor) · browser `playwright` (Chromium, headless)
**Verified:** `feat/5237-wms-seed-examples` @ `dc6b408`, in an isolated worktree, on a disposable Postgres provisioned by the repo's own ephemeral harness (`yarn initialize` → `mercato init`, examples on by default — i.e. the same fixture CI builds).

---

## ✅ Part 1 — the seed itself is good (P1)

**Where to click:** `/backend/wms/warehouses`, `/backend/wms/locations`, `/backend/wms/zones`, `/backend/wms/inventory`, `/backend/wms/lots`

| 1 | Warehouses list on a freshly initialised tenant | DEMO-NYC / DEMO-LAX / DEMO-BER, NYC primary | All three render; only DEMO-NYC carries **Primary** | ✅ |
| 2 | DEMO-NYC location topology | Docks, staging, aisles, racks, bins per `examples-data.ts` | Exactly 11 locations, every code and type matching the data file | ✅ |
| 3 | DEMO-NYC zones | RCV / PICK / BULK / SHIP | Receiving, Pick Face, Bulk Storage, Shipping — all present | ✅ |
| 4 | Inventory console balances | Opening stock attributed to the right bins | 23 balance rows across the three warehouses, plus 23 matching opening `Receive` rows in the movement ledger | ✅ |
| 5 | FEFO matcha lots | Staggered expiries so FEFO is demonstrable | 5 lots spanning 2026-09-05 → 2027-08-22 | ✅ |

Nothing in the dataset is wrong, duplicated, mis-scoped or missing. As a demo dataset for a human exploring the app, this does what the PR description says it does.

## ✅ Part 2 — the Adjust dialog works fine *with seeded entities* (P0)

Driven end to end as an operator would: variant `HOME-MUG-SLATE` → warehouse **New York Fulfillment Center** → bin `NYC-A01-R1-B01` → delta `7` → reason *Found stock* → **Save adjustment**. Every combobox committed, the Location field unlocked once the warehouse was chosen, there were no validation errors, the **Inventory adjusted** toast appeared, the dialog closed and the new `Adjust` row is visible in the movement ledger.

So "the seed breaks the adjust dialog" is **not** true in general — that matters for attribution below.

## ❌ Part 3 — what actually fails, and why (P0)

### 3a. A newly created warehouse is invisible to every warehouse picker

I created a warehouse through `POST /api/wms/warehouses` — the same fixture shape `TC-WMS-INVENTORY-UI-001` creates — and then queried the list route for it:

| Query | Result |
|---|---|
| unfiltered list | ✅ present (tenant total went 3 → 4 → 5) |
| `search=<its own name>` | ❌ **0 results** |
| `search=<its own code>` | ❌ **0 results** |
| re-probed at 0s / 2s / 5s / 15s / 30s | ❌ still 0 results — this is not indexing lag |
| `search=New York` / `search=Berlin` (seeded rows) | ✅ returns DEMO-NYC / DEMO-BER |

The warehouse combobox calls `loadWarehouseOptions(query)` → `/api/wms/warehouses?...&search=<query>`. So **a warehouse you just created can never be selected in any WMS warehouse picker**, while seeded ones select normally. In the UI the dropdown is literally empty for the new warehouse's own name, and populated for a seeded one.

### 3b. That is precisely the red CI lane, reproduced in a browser

Driving the warehouse combobox the way `helpers/wmsUi.ts#fillCombobox` does — type the name, blind `ArrowDown`+`Enter` (it passes no `suggestionsApiPath` for this field, so it never waits), `Tab`, submit:

```
dropdown for the new warehouse : []            <- empty
value immediately after Enter  : ""            <- nothing committed
value after blur               : ""
field errors on submit         : ["Invalid UUID", "Invalid UUID"]   <- Warehouse + Location
"Inventory adjusted" toast     : false
```

That is byte for byte the failure snapshot in the CI review comment (`combobox "Select warehouse"` / `alert: Invalid UUID` / `combobox "Select location"` / `alert: Invalid UUID`, with the variant combobox holding its value).

### 3c. Why this PR turns it from green to red — without being the bug

The failing behaviour lives in `packages/core/src/modules/wms/api/warehouses/route.ts` and the shared CRUD/query-engine list path. **This PR touches neither** — its six files are all under `wms/seed/`, `wms/cli.ts` and `wms/setup.ts`.

What changed is the *fixture*, and it removed an accident:

- **Before:** the integration tenant had **zero** warehouses; each spec created the only one in existence. The combobox's unfiltered list therefore contained exactly that warehouse, so the blind `ArrowDown`+`Enter` landed on the correct row **whether or not `search` worked**. The broken search was invisible.
- **After:** the tenant starts with three seeded warehouses. Typing the test's own warehouse name now returns an empty filtered list, the keyboard selection commits nothing, and the form fails with `Invalid UUID`.

So my earlier review's circumstantial case was right about *which* change flipped the lane and wrong to frame the seed as the defect. **The seed is the first realistic dataset this app has been tested against, and it found a real bug.**

### 3d. Separately: the cycle-count wizard cannot be started reliably

Unrelated to search, and also in untouched UI code (`CycleCountWizardDialog.tsx`):

- Pick warehouse → pick zone → **wait for the "Expected SKUs" scope estimate to settle** (it reached 8) → click **Start counting**: the wizard never advances. I clicked **12 times over 31 seconds**; every click surfaced `Invalid UUID` on Zone, and the Zone field emptied itself while Expected SKUs went blank — even though the operator had visibly selected a zone.
- Click **Start counting immediately** after picking the zone, before the estimate resolves: the first click is swallowed silently (no error, no transition) and the **second** click advances to step 2.

So the wizard is usable only if you click faster than the scope estimate, and unusable if you wait for it. Reproduced on both DEMO-NYC and DEMO-BER, so it is not a data-volume effect. The committed spec papers over the swallowed click with an `expect(...).toPass()` retry loop, which is why this has never surfaced as a product bug.

---

## 🏷️ Labels

Unchanged. This run is evidence-only, and in any case the PR carries `skip-qa`, so no QA sign-off is available to give or withhold here.

## Notes for QA / for the author

- **Nothing here asks you to change your seed.** The dataset is correct. The two options worth discussing are (a) fix the warehouse list `search` filter so a just-created row is findable by its own name and code, and (b) harden `fillCombobox` so it asserts a *committed* selection rather than the text it typed — its `toHaveValue(value)` check is satisfied by the characters the helper itself typed, which with `allowCustomValues={false}` proves nothing was selected. That weakness is why both specs died ten seconds later at an assertion that looks unrelated to the real problem.
- **Not verified:** I did not boot a second environment on `develop` to A/B the search behaviour, so the attribution in 3c is by code path and by the before/after fixture argument, not by an empirical control run. If you want that control, `POST /api/wms/warehouses` on any develop tenant and then query the list route with `search=<that warehouse's code>` — I expect zero results there too.
- **Not verified:** whether the same search gap affects locations, zones or lots. The location field is disabled until a warehouse commits, so I could not reach it in the failing path.
- Only demo credentials from the test-env descriptor were used. No tokens, `.env` content or real user data appear in any screenshot.

### Screenshots

**1 — Seeded warehouses: DEMO-NYC (Primary), DEMO-LAX, DEMO-BER**



**2 — Inventory console: 23 seeded opening-stock balances across the three warehouses**



**3 — Adjust dialog fully committed against a SEEDED warehouse (this path works, and saved successfully)**



**4 — Warehouse picker: typing a SEEDED warehouse name returns the option**



**5 — Warehouse picker: typing a JUST-CREATED warehouse's own name returns nothing**



**6 — The CI failure reproduced by hand: Warehouse and Location both 'Invalid UUID', variant intact, no toast**



**7 — Cycle count wizard stuck on step 1 after 12 clicks: Zone emptied itself, 'Invalid UUID', Expected SKUs blank**




<sub>Artifacts (screenshots, report.json, throwaway specs) also on the QA host at `.ai/qa/artifacts_20260822-qa5337/`. Evidence branch: `qa-evidence-pr-5337` — not this PR's branch.</sub>

pkarw:

This change ships no browser-level test — its two test files (`wms/__tests__/setup.test.ts` and `wms/seed/__tests__/examples.test.ts`) are unit tests over the seed functions. The UI QA above was manual; the scenario below locks in what I verified by hand, and can be implemented with `/om-integration-tests`.

**Scenario A — the seeded dataset is actually reachable in the admin UI** (this is the PR's own deliverable, and today nothing guards it):

1. **Setup:** none beyond the harness — `mercato init` already seeds examples, so the tenant carries the dataset.
2. **Act:** sign in as `admin`, then visit `/backend/wms/warehouses`, `/backend/wms/locations`, `/backend/wms/zones`, `/backend/wms/inventory` and `/backend/wms/lots`.
3. **Assert:** `DEMO-NYC`, `DEMO-LAX`, `DEMO-BER` are listed and only `DEMO-NYC` is flagged primary; `DEMO-NYC` exposes its 11 documented locations and 4 zones; the balances grid holds the 23 opening-stock rows with the documented quantities (e.g. `HOME-MUG-SLATE` = 64 at `NYC-A01-R1-B01`); the lots list holds the five matcha lots with distinct expiries.
4. **Teardown:** none — the scenario is read-only and creates no fixtures.

**Scenario B — a just-created warehouse is selectable in the warehouse picker** (the regression guard the two red specs *should* have been; it fails today, so land it with the fix, not before):

1. **Setup:** create a warehouse, a zone, a bin location, a catalog product + variant and an inventory profile via the API fixtures, exactly as `TC-WMS-INVENTORY-UI-001` does.
2. **Act:** open `/backend/wms/inventory` → **Adjust inventory**, type the new warehouse's name into the warehouse combobox and wait for `/api/wms/warehouses?...search=` to answer.
3. **Assert:** the suggestion list contains that warehouse and, after selecting it, the **Location** field becomes enabled — i.e. assert a *committed* selection rather than `toHaveValue(<the text we typed>)`, which with `allowCustomValues={false}` proves nothing was selected. This assertion is the one that would have pointed straight at the real defect instead of failing ten seconds later on `Select zone`.
4. **Teardown:** delete the inventory profile, location, zone, warehouse and catalog product created in setup.

**Scenario C — the cycle-count wizard advances on a single click** (also failing today; file with the wizard fix):

1. **Setup:** as Scenario B, plus one opening balance in the bin.
2. **Act:** open the **Cycle count** wizard, select the warehouse and zone, **wait for the Expected SKUs scope estimate to settle**, then click **Start counting** exactly once.
3. **Assert:** the wizard reaches `Step 2 of 3 · Scan and tally items` on that single click and the Zone field still holds the selected zone. Do not wrap the click in `expect(...).toPass()` — the retry loop is what has been hiding this.
4. **Teardown:** as Scenario B.

pkarw:
🤖 `om-auto-qa-pr` completed: **FAIL**. Evidence posted above. Lock released.

Read the verdict carefully before acting on it: **the FAIL is about the head, not about your seed.** The dataset this PR adds verifies fully green in a real browser — three warehouses with NYC correctly primary, the full NYC location topology, four zones, 23 opening-stock balances and five staggered matcha lots — and the Adjust inventory dialog completes end to end against seeded entities.

What fails is a **pre-existing platform defect this PR is the first change to expose**: `/api/wms/warehouses` returns a freshly created warehouse in its unfiltered list but returns zero results for `search=<that warehouse's own name>` or `search=<its own code>`, re-probed out to 30 seconds. Every WMS warehouse combobox filters through that `search` parameter, so a warehouse you just created cannot be selected anywhere. Before this PR the integration tenant had exactly one warehouse — the one each spec created — so the unfiltered dropdown always contained it and the broken search never mattered. Three seeded warehouses remove that accident, and `TC-WMS-INVENTORY-UI-001` fails with `Invalid UUID` on Warehouse and Location. I reproduced that signature by hand, with screenshots.

This corrects the framing of my two earlier CI comments: I was right about which change flipped the lane and wrong to treat the seed as the thing to fix. The fix belongs in the warehouse list search (and, separately, in `fillCombobox`, whose `toHaveValue` check is satisfied by the text the helper itself typed and therefore cannot tell "option selected" from "characters typed").

I also found, independently of this PR, that the cycle-count wizard's first **Start counting** click is silently swallowed and that waiting for the scope estimate wipes the selected zone — 12 clicks over 31 seconds never advanced past step 1. That is in `CycleCountWizardDialog.tsx`, which this PR does not touch, and deserves its own issue.

`changes-requested` stays for now because the head is red, but nothing here is a request for you to change the six files in this PR. Labels are otherwise unchanged — this run was evidence-only, and the PR carries `skip-qa` so no QA sign-off was available to give.

pkarw:
@mkadziolka can you please fix the requested changes?

pkarw:
@mkadziolka please address the QA issues and the failing CI

mkadziolka:
Hypothesis, not a merge-order block: the TC-WMS-INVENTORY-UI-001 combobox miss on this PR is likely the same warehouse hybrid-index gap that #5339 (`36efaa2a4`) is fixing. Please don't keep beating on the seed dataset — pkarw already said leave it.

**Symptom here:** `GET /api/wms/warehouses?search=<just-created name/code>` returns empty, while the unfiltered list includes that warehouse. Seeded `DEMO-NYC` / `DEMO-LAX` / `DEMO-BER` remain searchable.

**Mechanism (same as #5339):** once any `entity_indexes` rows exist, list search is index-only. Warehouses never written to the index disappear from `?search=`. Seeded DEMO rows are indexed; freshly created ones were not, until warehouse CRUD indexing landed. #5339 (`feat/5238-warehouse-dialog`, https://github.com/open-mercato/open-mercato/pull/5339) adds that indexer plus an `entity_indexes` tombstone on delete (same pattern as zones) — TC-WMS-CRUDFORM-001 was leaving index rows behind and hiding new warehouses on shard 14. The country-name search (Poland → PL) work in #5339 is unrelated to this.

**Caveat:** CI on the #5339 isolation-fix SHA `36efaa2a4` has not actually run yet (fork `action_required`). After #5339 lands, this PR should merge `develop` and re-run shard 14. That is the next check — not a guarantee it goes green on its own.

mkadziolka:
Merged `upstream/develop` into this branch (merge, not rebase) so CI can re-run on a current head.

The merge includes **#5339** — the warehouse query-index indexer fix (`warehouseCrudIndexer` / `emitWarehouseCrudSideEffects`). That is the shard-14 warehouse search failure, not the seed dataset.

- Seed examples are unchanged (per prior review: the fail is warehouse search, not seed).
- Please re-run / re-review shard 14 on the new head.

@pkarw

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T01:29:28Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review (Re-review): feat(wms): seed example warehouses, locations, and opening stock

## 🎯 Summary

This is the third pass on this PR and the second re-review, this time of head `f868a9a0`, which adds a merge of `develop` on top of the `dc6b408` I approved on 2026-08-22. Two things are genuinely new since that pass and are what this review concentrates on: the `fillCombobox` rewrite in `packages/core/src/modules/wms/__integration__/helpers/wmsUi.ts` (commit `9a809fb9`, pushed 40 minutes *after* my re-review, so it has never been reviewed), and the `develop` merge that pulls in #5339's warehouse query-index indexer.

**The combobox helper change is correct, and it fixes a real defect rather than papering over one.** I verified the mechanism end to end. `ComboboxInput` renders its dropdown entries as `<Button … role="option">` (`packages/ui/src/backend/inputs/ComboboxInput.tsx:499-505`), and an explicit `role="option"` overrides the element's implicit `button` role, so the old helper's `dropdown.getByRole('button', …)` could never match a suggestion — `git log -L` on that block shows the `role="option"` attribute predates commit `0ab2d5dfe`, so the button selector has been dead for some time. Every call therefore fell through to the `ArrowDown`/`Enter` keyboard fallback, which left the typed text in the input without committing the form value; the assertion that followed, `expect(input).toHaveValue(value)`, then passed **vacuously** against the text the test itself had typed. The new code targets `getByRole('option')` first, so the click path actually runs `selectValue(option.value)` (`ComboboxInput.tsx:271-283`), which sets the input to `option.label`. I checked each call site's label mapping to confirm the surviving `toHaveValue(value)` assertion is now a real assertion rather than a newly broken one: variants map `label = sku` (`wmsLookupLoaders.ts:73`), warehouses map `label = name` (`wmsLookupLoaders.ts:52`), zones map `label = name` (`inventoryMutationLoaders.ts:426`), and the specs pass exactly those. The narrowed `waitForResponse` predicate — which now requires the typed value to appear in the decoded request URL — correctly ignores the unfiltered load that firing `input.click()` triggers, which is precisely the race the old predicate was matching on.

The seed itself (`seed/examples.ts`, `seed/examples-data.ts`, `setup.ts`, `cli.ts`) is unchanged since the head I approved, and I re-verified the areas the previous passes called out: `DEMO-NYC` is only marked primary when the organization has none, the movement idempotency key makes a re-run a no-op, multi-variant demo products are correctly derived as `configurable`, and the FEFO matcha lots carry the staggered expiries the tests pin. The new CLI command mirrors the `catalog` and `sales` `seed-examples` commands closely, including container disposal in a `finally`.

I did find **one thing the two earlier passes missed**, and I want to be straightforward that it is a late catch on code I already approved twice rather than a regression in this head: the seed is the only place in the entire `wms` module that reads entities with raw `em.find` / `em.findOne`. Details below.

## Verdict

❌ **request changes** — one Minor finding (raw entity reads that bypass the decryption helpers the rest of the `wms` module uses without exception). This repository's review rules escalate Medium/minor findings to `changes-requested`, so that single finding decides the verdict; there is no blocker and no major here, and the fix is a mechanical three-call change. I would also accept an explicit waiver from a maintainer if you would rather land this as-is and fix it in a follow-up — say so on the PR and I will re-review to `merge-queue`.

## 🧪 Validation Gate

Per this repository's GitHub-checks-first rule, CI is the validation source and no local substitute gate was run. The fork's workflow runs were parked at `action_required` for head `f868a9a0` (fork PRs need per-push maintainer approval, and nothing had approved this push), so **there was no check data for this head at all** until I approved runs [32963644016](https://github.com/open-mercato/open-mercato/actions/runs/32963644016) and [32963644174](https://github.com/open-mercato/open-mercato/actions/runs/32963644174) at the start of this review. Those runs are now executing.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | Covered by CI `prepare` (3m33s). |
| `yarn generate` | ✅ PASS | Covered by CI `prepare`. |
| `yarn build:packages` (second) | ✅ PASS | Covered by CI `prepare`. |
| `yarn i18n:check-sync` | ✅ PASS | Covered by CI `prepare`; the diff adds no locale keys. |
| `yarn i18n:check-usage` | ✅ PASS | Covered by CI `prepare`. |
| `yarn typecheck` | ✅ PASS | Covered by CI `prepare` / `scope`. |
| `yarn test` | ⏳ PENDING | CI `test` was still running when this review was submitted. |
| `yarn build:app` | ✅ PASS | Covered by CI `prepare` (the shared app bundle the integration shards reuse). |

Green at submission time: `prepare`, `lint`, `ds-lint`, `scope`, `audit`, `audit-scope`, `documents-multi-instance`, `license/cla`. Still pending: `test` and all fifteen `ephemeral-integration` shards. **Pending is not failing and is not a reason to hold the verdict**, so this review goes out now; I will post the CI outcome as a follow-up comment and correct the label if it changes the picture.

**One thing to know before reading those shard results.** The earlier red run on `dc6b408` ran as `ephemeral-integration (none)` — the affected-only single runner. This head runs the **full fifteen-shard suite**, and that is a side effect of branch drift, not of anything in your diff. `ci.yml:104` computes scope with a two-dot `git diff origin/develop --name-only`, and this branch is 49 commits behind `develop` even after the 2026-08-26 merge, so that diff reports 158 files, 131 of them under `packages/` — which trips the "shared packages touched, run the full suite" branch at `ci.yml:114`. Almost all of those 131 files are `develop`'s own commits, not yours. Practically: if a shard unrelated to WMS goes red, check it against `develop` before attributing it here, and merging `develop` again would shrink the run back to the affected-only runner and give you a much cleaner signal.

## Findings

### 🔹 Minor

**`packages/core/src/modules/wms/seed/examples.ts:92`, `:141`, `:167` — raw `em.find` reads bypass the decryption helpers that the rest of the `wms` module uses without exception.**

Outside tests, the `wms` module makes **84** reads through `findWithDecryption` / `findOneWithDecryption` and, before this PR, **zero** raw `em.find` / `em.findOne` calls. This file introduces the module's only raw reads. That is not an ambiguous style preference — it is a module-wide invariant, and `.ai/review-checklist.md` § 2 lists it as a MUST ("`findWithDecryption`/`findOneWithDecryption` used instead of raw `em.find`/`em.findOne`").

It matters because encryption in this platform is asymmetric: writes are encrypted transparently by `TenantEncryptionSubscriber.beforeCreate` / `beforeUpdate` (`packages/shared/src/lib/encryption/subscriber.ts:409-413`), but decryption on read only happens inside the `*WithDecryption` helpers. These three reads each fetch rows and then key them by a field whose value must be plaintext for the idempotency check to work:

- `:92` builds `warehouseByCode` from `warehouse.code`,
- `:141` builds `zoneByCode` from `zone.code`,
- `:167` builds `locationByCode` from `location.code`.

Concrete failure: on a tenant that has `warehouses.code` (or the zone/location equivalent) marked encrypted, the raw read returns ciphertext, `warehouseByCode.get('DEMO-NYC')` misses, and `seedWmsExamples` silently re-creates all three warehouses, their zones and their whole location topology on every run — turning the documented "re-run is a no-op" contract into duplicate data. The idempotency test at `seed/__tests__/examples.test.ts` cannot catch this, because `createMemoryEm()` has no encryption layer at all.

Fix: pass these three through `findWithDecryption(em, Entity, where, undefined, { tenantId: scope.tenantId, organizationId: scope.organizationId })`, matching the pattern already used at `wms/commands/configuration.ts:144-160`. The test file already mocks `@open-mercato/shared/lib/encryption/find`, so no test scaffolding changes are needed.

To be explicit about scope, I am **not** asking you to convert the other eight raw reads. Lines `:213`, `:246`, `:275`, `:282`, `:380` and `:390` filter *by* a potentially-encrypted field (`sku`, `idempotencyKey`, `lotNumber`), and the decryption helpers do not fix filter-side matching — they decrypt returned rows, they do not encrypt the `where` clause. Lines `:309` and `:418` filter on IDs and scope columns only. Converting those would be cargo-culting; only the three reads above actually change behaviour.

### 💅 Nit

**`packages/core/src/modules/wms/__integration__/helpers/wmsUi.ts:188` — `selectLocationComboboxOption` still carries the exact bug this PR just fixed in `fillCombobox`.** It targets `dropdown.getByRole('button', { name: locationCode, exact: true })` against the same `ComboboxInput` dropdown, and its fallback (`page.getByRole('button', …).click()`) would hang on the default click timeout rather than fail fast. It has no callers anywhere in `__integration__`, so it cannot redden CI today, but leaving a dead helper that models the wrong selector next to the corrected one is a trap for whoever writes the next WMS spec. Either delete it or give it the same `role="option"` treatment.

**`packages/core/src/modules/wms/seed/examples-data.ts:149`, `:170` — the demo products that are now `configurable` ship variants with no `optionValues` and no `optionSchemaId`.** Carried forward from the 2026-08-22 pass, unchanged and still not blocking: `catalog/seed/examples.ts` marks its own multi-variant examples configurable with no option schema and `optionSchemaId` is nullable, so this is consistent with the platform. Worth a follow-up only if the variant matrix builder ever starts requiring a schema.

**`packages/core/src/modules/wms/cli.ts:95` — `mercato wms seed-examples` does not rebuild the query index afterwards.** `mercato init` covers this by running `query_index reindex` after all `seedExamples` hooks (`packages/cli/src/mercato.ts:1313`), but the standalone CLI command and `mercato auth setup --with-examples` (`auth/cli.ts:522-533`) do not, so on a tenant that already has `entity_indexes` rows the seeded `DEMO-*` warehouses would not appear under `?search=`. This is the same class of gap #5339 just closed for CRUD-created warehouses. The `catalog` and `sales` `seed-examples` commands have exactly the same gap, so this is a platform-wide follow-up rather than a defect this PR introduced — noting it because this PR's own CI history makes the failure mode topical.

**Still no browser-level coverage of the seeded dataset.** The 2026-08-22 UI QA pass suggested a scenario that asserts `DEMO-NYC` / `DEMO-LAX` / `DEMO-BER` and the 23 opening-stock rows are reachable in the admin UI. The PR's two test files remain unit tests over the seed functions. Genuinely optional for this PR, and a good `/om-integration-tests` follow-up.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff is purely additive (`seedWmsExamples`, `SeedWmsExamplesResult`, the `examples-data` seed constants).
- [x] No HTTP route URL removed or renamed; no method changed — no API routes are touched.
- [x] No CLI command or flag renamed or removed — `wms seed-examples` is a new command added alongside the existing `verify-balances`, and its `--tenant` / `--org` flags follow the `catalog` and `sales` precedent.
- [x] No database table or column renamed or removed; no column type narrowed — no entity or migration changes at all.
- [x] No config key renamed and no default changed silently — `setup.seedExamples` is a new optional hook, so a tenant created with `--no-examples` is unaffected.
- [x] Where a contract had to change: not applicable, nothing changed.

## 🧪 Test Coverage

`seed/__tests__/examples.test.ts` is good work and covers more than most seed code gets: parent-before-child location ordering plus the missing-parent throw, referential integrity of every `STOCK_SEEDS` row against the warehouse/location/variant sets, the full happy path with concrete assertions on warehouse codes and the single-primary invariant, idempotency across two consecutive runs, reuse of pre-existing catalog fashion variants, the "organization already has a primary warehouse" branch, the "no user to stamp movements" branch, and the exact FEFO expiry arithmetic for `MATCHA-BER-NEAR`. `__tests__/setup.test.ts` pins that the module's `seedExamples` hook forwards the tenant and organization through to `seedWmsExamples`. The `configurable` derivation is asserted in both directions (configurable for the two multi-variant products, simple for the single-variant matcha tin), which is the right shape for a fix that flipped a boolean.

Two gaps worth naming. First, the idempotency test proves the *logic* is idempotent but cannot prove the *lookups* are, because `createMemoryEm()` returns exactly what was persisted with no encryption layer — that blind spot is precisely what hides the Minor finding above, and it is why that finding needs a code fix rather than a test. Second, the `fillCombobox` rewrite is itself test infrastructure and is verified only transitively by whether the WMS integration shards go green; given that its previous incarnation produced a *vacuously passing* assertion for months, it would be worth a short comment in the helper recording why the selector must be `role="option"` and not `button`, so the next person to "simplify" it does not walk the code back.

pkarw:

- ❌ `changes-requested` — the re-review of head `f868a9a0` found one Minor issue: the seed is the only code in the `wms` module that reads entities with raw `em.find`, bypassing the decryption helpers the module's other 84 reads all use, which breaks the documented "re-run is a no-op" contract on a tenant that encrypts a warehouse, zone or location `code`. This repository escalates Medium/minor findings to `changes-requested`, so that single finding decides the verdict; there is no blocker and no major.
- ✨ `feature` — the change adds new WMS example-seeding behaviour (a `seedExamples` hook and a `mercato wms seed-examples` command) rather than fixing existing behaviour.
- ⏭️ `skip-qa` — the automated-verification exemption still applies: the diff touches no UI-rendering file (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`), changes no database structure and no API surface, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships automated tests for the changed behaviour in the same PR.
- 🔹 `priority-medium` — this is an ordinary additive feature covering demo and example data, with no release-blocking, security or data-integrity dimension that would justify raising it.
- 🟡 `risk-medium` — the blast radius is a single module's seed path plus its integration-test helper, which is more than a cosmetic change but well short of the auth, migration or shared-contract surfaces that would make it `risk-high`.

`in-progress` is being swapped for `ci-monitoring` as this run finishes: the review is submitted and the work is reported, and I am only watching the CI runs I approved for this head. `ci-monitoring` is a meta label, not a claim — this PR is free for anyone to act on.

pkarw:
Thanks @mkadziolka — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pkarw:

Re-review of head `f868a9a0`. **The `develop` merge did the job it was meant to do** — #5339's warehouse indexer (`warehouseCrudIndexer` / `emitWarehouseCrudSideEffects`) is present in this head, and the combobox helper rewrite from `9a809fb9` (pushed after my 2026-08-22 re-review, so never previously reviewed) is a correct fix rather than a workaround: `ComboboxInput` renders suggestions as `<Button role="option">`, which overrides the implicit `button` role, so the old `getByRole('button')` selector had been dead and every call was silently falling through to a keyboard fallback whose `toHaveValue` assertion passed vacuously against the test's own typed text.

The verdict is driven by **one Minor finding**, and I want to be upfront that it is a late catch on code I approved twice rather than anything new in this head: `seed/examples.ts:92`, `:141` and `:167` read `Warehouse`, `WarehouseZone` and `WarehouseLocation` with raw `em.find` and then key the results by `code`. The `wms` module makes 84 reads through `findWithDecryption` / `findOneWithDecryption` outside tests and, before this PR, zero raw ones. Because writes are encrypted transparently while decryption only happens inside those helpers, a tenant that encrypts one of those `code` columns would get duplicate warehouses on every re-run instead of the documented no-op — and the in-memory EM in the idempotency test has no encryption layer, so it cannot catch it. Three calls, mechanical fix. A maintainer waiver is fine too; say so and I will re-review to `merge-queue`.

Validation source: GitHub checks. The fork's runs for this head were parked at `action_required` — there was no check data at all — so I approved runs [32963644016](https://github.com/open-mercato/open-mercato/actions/runs/32963644016) and [32963644174](https://github.com/open-mercato/open-mercato/actions/runs/32963644174) at the start of the review. `prepare`, `lint`, `ds-lint`, `scope`, `audit`, `audit-scope`, `documents-multi-instance` and `license/cla` are green; `test` and the integration shards were still running at submission time, and pending is never a reason to hold a verdict. I will post the CI result here.

**Read those shard results carefully.** This head runs the full fifteen-shard suite instead of the affected-only `(none)` runner the previous head used, and that is branch drift, not your diff: `ci.yml:104` scopes with a two-dot `git diff origin/develop`, and this branch is still 49 commits behind `develop`, so that diff reports 158 files — 131 of them under `packages/`, almost all belonging to `develop`. Merging `develop` again would shrink the run back to the affected-only runner and give a far cleaner signal.

pkarw:

**CI settled red, so `changes-requested` stands — but the failure moved, and where it moved to is the most useful signal this PR has produced yet.** 14 of 15 shards are green, `test` is green, and every static gate is green. Only `ephemeral-integration (14/15)` fails: [job 98381576707](https://github.com/open-mercato/open-mercato/actions/runs/32963644016/job/98381576707), 153 passed / 2 failed / **0 flaky**, both failures reproducing on Playwright's automatic retry.

## The two fixes in this head both worked

The same two tests fail by name, so at a glance nothing changed. It did. On `dc6b408` they died on **Warehouse** and **Zone**; on `f868a9a0` warehouse and zone selection both succeed and the tests get several steps further before dying on **Location**:

| | `dc6b408` (2026-08-22) | `f868a9a0` (this head) |
|---|---|---|
| `:37` adjust | ❌ `Inventory adjusted` toast never appeared | ❌ `spec.ts:122` — Location combobox has no options |
| `:152` cycle count | ❌ `Select zone` never rendered | ❌ `spec.ts:293` — Location combobox has no options |

Zone now renders and commits, which is #5339's indexer plus the `role="option"` helper fix doing exactly what they were supposed to do. The wall moved one field to the right.

## The mechanism, pinned

Both failures land on the same line, `helpers/wmsUi.ts:140`, waiting for `getByRole('option', { name: /BIN-fdc0a851/i })` (and `/CC-52468190/i` in the wizard). Those are the codes of locations the spec's own fixtures had just created. The dropdown renders **no options at all** — the location `?search=` returns nothing for a location that demonstrably exists.

This is the warehouse bug my 2026-08-22 QA pass pinned, one entity over, and it is a pre-existing platform gap rather than anything wrong with your seed. The chain:

1. `wms/api/locations/route.ts:46` declares `indexer: { entityType: E.wms.warehouse_location }` — but all three mutating actions delegate through `commandId: 'wms.locations.create' | '.update' | '.delete'` (`route.ts:100`, `:110`, `:119`), and on the command path the CRUD factory executes the command and returns (`packages/shared/src/lib/crud/factory.ts:2288`) without ever applying the route-level indexer. **The declared `indexer:` is inert.**
2. Warehouses and zones compensate for that in the command layer — `emitWarehouseCrudSideEffects` (`commands/configuration.ts:411`, wired to `warehouseCrudIndexer`, added by #5339) and `emitZoneCrudSideEffects` (`:435`). `wms.locations.create` (`commands/configuration.ts:1365`) has **no** equivalent emitter and there is no `warehouseLocationCrudIndexer` in `commands/shared.ts`. So a location created through the API is never written to `entity_indexes`.
3. `query_index/lib/engine.ts:376-385` falls back to the base table **only when the entity has zero index rows** (`indexAnyRows`). Zero rows means "not indexed at all, query the table"; one row means "indexed, trust the index".
4. Before this PR, `wms:warehouse_location` had zero index rows in the integration tenant, so every location search silently took the base-table fallback and worked. This PR seeds roughly 25 locations and `mercato init` then runs `query_index reindex` (`packages/cli/src/mercato.ts:1313`), so index rows now exist — the fallback switches off tenant-wide, and every location created afterwards becomes invisible to `?search=`.

That is the same accident as before: the integration tenant used to have so little WMS data that a broken code path could never be observed, and a realistic dataset removes the accident. **Your seed is the messenger.**

## What I suggest

The fix is small and has a precedent sitting right next to it: add a `warehouseLocationCrudIndexer` to `commands/shared.ts` and an `emitLocationCrudSideEffects` to `commands/configuration.ts`, mirroring what #5339 did for warehouses, including the `entity_indexes` tombstone on delete. That is a platform fix, not seed work, so it is reasonable to land it as its own PR and rebase this one on top rather than growing this diff — your call, and I am happy either way.

Worth flagging separately: the inert `indexer:` on a `commandId`-backed route is a trap that has now bitten this module three times (warehouses, and both location call sites). A factory-level warning when both are configured would stop the fourth.

## Where that leaves the verdict

Unchanged: **`changes-requested`**, now for two reasons rather than one — the Minor finding in my review (`seed/examples.ts:92`, `:141`, `:167` reading with raw `em.find` and keying by `code`) and this red required check. `ci-monitoring` is removed; nobody is watching CI for this head any more.

One caveat I raised in the review and am repeating because it still applies: this head ran the full fifteen-shard suite only because the branch is 49 commits behind `develop`. Merging `develop` again before the next push will shrink it back to the affected-only runner and make the next signal much easier to read.

PR #5219, opened 2026-08-12
fix(shared): make I18nProvider children optional so it typechecks via React.createElement (#5155)

adeptofvoltron:
Closes #5155

## 🎯 Goal
- Let `I18nProvider` be constructed via `React.createElement(I18nProvider, props, children)` (the positional-children form) without a type error, so test wrappers and provider-composition helpers no longer need an `eslint-disable` workaround.

## 🔍 Problem
`I18nProvider`'s props type declared `children: ReactNode` as required. TypeScript's last `createElement` overload requires `children` to be inside the `props` argument when it's a required field of `P`, so `React.createElement(I18nProvider, { locale, dict }, children)` failed to typecheck under `strict: true`, even though the JSX form and rendering with no children were always valid at runtime. The alternative — passing `children` inside the props object — triggers the `react/no-children-prop` lint rule.

## 🔍 Root Cause
`packages/shared/src/lib/i18n/context.tsx:51` typed `I18nProvider`'s destructured parameter with an inline object type where `children` had no `?`, making it required in `P` for `createElement`'s overload resolution.

## What Changed
- `packages/shared/src/lib/i18n/context.tsx` — `I18nProvider`'s props type now uses `PropsWithChildren<{ locale: Locale; dict: Dict; localeLocked?: boolean }>`, making `children` optional. No runtime behavior change: JSX usage is unaffected, and a provider rendered with no children was already legal.
- `packages/shared/src/lib/i18n/__tests__/context.test.tsx` (new) — runtime test asserting `I18nProvider` renders correctly when constructed via the positional-children `React.createElement` form, and when rendered with no children at all.
- `packages/shared/src/lib/i18n/context.typecheck.tsx` (new) — a compile-time-only regression guard. This repo's Jest transform runs with `isolatedModules: true`, so ts-jest never surfaces TypeScript diagnostics — the runtime test above would keep passing even if `children` regressed back to required. The file lives outside `__tests__` (which `packages/shared/tsconfig.json` excludes from `tsc --noEmit`) specifically so `yarn typecheck` catches a future regression of this exact defect. It is never imported at runtime, and it asserts all three call shapes: positional children, no children, and ordinary JSX children.

### Why the guard is a `.tsx` and not a `.ts`
[...]

PR #5343, opened 2026-08-17
fix(ci): report uncovered files instead of aborting the mutation gate (#5281)

### LINKED ISSUE #5281: ci(stryker): mutation gate aborts instead of reporting when a changed file has no covering tests
## Summary

The mutation-testing gate aborts its whole job — rather than reporting an uncovered file — whenever a pull request adds a `src/lib/**/*.ts` file that no test covers yet. The job goes red, produces no `mutation.json`, and the report and enforce steps have nothing to work with.

This was hit for real on #5219, where it turned that PR red. It was worked around there by moving the offending file out of mutation scope, but the underlying gap is general and will hit the next PR that adds an as-yet-uncovered source file.

## Mechanism

- `scripts/stryker/scope.mjs` selects changed `src/lib/**/*.ts` files in allowlisted packages into the mutate matrix.
- `scripts/stryker/createConfig.mjs` sets `jest.enableFindRelatedTests: true`, so Stryker runs only the Jest tests related to the mutated files.
- When no test imports the mutated file, the dry run executes zero tests and Stryker exits with a `ConfigError` instead of scoring:

```
ERROR Stryker No tests were executed. Stryker will exit prematurely. Please check your configuration.
ConfigError: No tests were executed. Stryker will exit prematurely. Please check your configuration.
    at DryRunExecutor.execute (.../3-dry-run-executor.js:47:19)
```

The `mutate` job's `continue-on-error` (set while `MUTATION_ENFORCE` is not `true`) keeps a *low score* advisory, but it does not stop this hard error from surfacing as a red check on the PR. As `.github/workflows/mutation-tests.yml` notes in its own header, branch protection on `develop` is not readable through the API, so the repository's review and merge automation treats every reported check as required — meaning this red job does reach the pipeline.

## Reproduction

Open a PR that adds any new `packages/shared/src/lib/**/*.ts` file which no existing test imports. The `mutate (shared, <that file>)` check fails with the error above.

Concretely, on #5219: run `gh run view --job 94531000797 --log-failed`, or see https://github.com/open-mercato/open-mercato/actions/runs/31580607535.

## Expected

A changed source file with no covering tests should be reported as **uncovered** — a 0% score, or an explicit "no related tests" note in the job summary and PR comment — so the signal is "this file needs tests", which is useful and actionable. It should not abort the run with a configuration error that yields no report at all.

## Possible directions (not a decision)

- Detect the zero-related-tests case in the workflow (or a wrapper) before invoking Stryker, and emit an uncovered-file report instead of running it.
- Or filter the mutate list down to files that have at least one related test, and surface the dropped ones the way `scope.mjs` already surfaces its `maxFiles` truncation.

Either way the fix belongs in the shared CI tooling (`scripts/stryker/*`, `.github/workflows/mutation-tests.yml`), which is why it was deliberately left out of #5219 — that PR is a two-line type fix in `packages/shared`, and changing the mutation gate from it would have been unrelated scope.

## Context

Filed as a follow-up from #5219. See `.ai/specs/2026-07-31-stryker-mutation-testing-ci-gate.md` for the gate's design intent.

adeptofvoltron:
Closes #5281

## 🎯 Goal
- Stop the mutation-testing gate from aborting a whole CI job when a PR adds a `src/lib/**/*.ts` file no test covers yet; report it as an uncovered file needing tests instead.

## 🔍 Problem
`scripts/stryker/scope.mjs` selects diff-scoped `src/lib/**/*.ts` files in allowlisted packages into the mutate matrix, filtering only by path/glob rules — it has no concept of test coverage. Because `scripts/stryker/createConfig.mjs` sets `jest.enableFindRelatedTests: true`, Stryker's dry run only executes tests related to each mutated file. When a file has none, the dry run executes zero tests and Stryker throws `ConfigError: No tests were executed`, aborting the whole job and producing no `mutation.json` — so the report and enforce steps have nothing to work with. This was hit for real on #5219, where it turned that PR red; the underlying gap is general and hits the next PR that adds an as-yet-uncovered source file.

## 🔍 Root Cause
Both the CI `mutate` job and the local `yarn mutation:changed` wrapper (`scripts/stryker/mutation-changed.mjs`) passed every diff-scoped file straight to Stryker unconditionally, with no check for whether Jest has a related test for it.

## What Changed
- `scripts/stryker/relatedTests.mjs` (new) — `partitionByRelatedTests` runs `jest --listTests --findRelatedTests <file>` per file (a static dependency-graph check — no tests execute) and splits the mutate list into `covered`/`uncovered`.
- `.github/workflows/mutation-tests.yml` — new "Partition files by related test coverage" step in the `mutate` job (after "Build packages", before "Run mutation testing") feeds only the `covered` subset to Stryker via `MUTATE_LIST`, and skips the Stryker step entirely when nothing is covered. `UNCOVERED_FILES` is passed through to the report step.
- `scripts/stryker/report.mjs` — `renderMarkdown`/`renderMissingReportMarkdown`/`parseReportArgs` gain an `uncoveredFiles`/`--uncovered` surface, rendering a distinct "Needs tests" note in the job summary (mirrors the existing file-cap "Not measured" note) instead of silently skipping or looking like a crash.
- `scripts/stryker/mutation-changed.mjs` — the local wrapper had the identical gap; it now partitions each package's mutate list the same way before invoking Stryker, logging dropped files like the existing file-cap warning.
- `scripts/stryker/scope.mjs` / `scripts/stryker/createConfig.mjs` are untouched — coverage filtering needs `yarn install` (a Jest binary and package config), so it belongs in the `mutate` job / local wrapper, not the lightweight `scope` job that only diffs git refs.

## 🧪 Tests
- Manually reproduced the exact failure and fix by hand against `packages/shared`: `jest --listTests --findRelatedTests <file>` returns empty output (exit 0) for a real untracked, unimported file, and a full related-test list for a covered one — confirming the partition mirrors Stryker's own dry-run mechanism.
- `node --test scripts/__tests__/stryker-related-tests.test.mjs scripts/__tests__/stryker-mutation-changed.test.mjs scripts/__tests__/stryker-report.test.mjs scripts/__tests__/stryker-workflow.test.mjs scripts/__tests__/stryker-scope.test.mjs scripts/__tests__/stryker-enforce.test.mjs scripts/__tests__/stryker-create-config.test.mjs` — 112/112 passed, including 8 new tests covering the partition function, the "some uncovered" and "all uncovered" wrapper paths, the new report sections, and the updated workflow shape.
- `yarn test:scripts` (600 tests) — 596 passed / 4 failed; confirmed via `git stash` that the same 4 failures (`monorepo dev wrapper` env-resolution tests, a `repo-wide-guards` classification gap on an unrelated `create-app` template test) exist identically on unmodified `develop` — pre-existing and unrelated to this change.
- Skipped the full `validation.commands` gate (`build:packages`, `generate`, `i18n:check-*`, `typecheck`, `test`, `build:app`): this change touches only `scripts/` (not a yarn workspace; outside root `tsconfig.json`'s empty `include`; not covered by `turbo run lint`) and `.github/workflows/`, so those commands exercise none of the changed code.

## 💥 Breaking Changes
- None. `renderMarkdown`, `renderMissingReportMarkdown`, and `parseReportArgs` gained new optional parameters/fields (`uncoveredFiles`, defaulting to `[]`); every existing call shape and the rendered output for the no-uncovered-files case are byte-identical.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5343"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789555151&installation_model_id=432243&pr_number=5343&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5343&signature=ca332290882f582d320cbaf82e4d8ca1e87e6c412f4e5588e73efbb32e4e1ace"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:

- 🔍 `review` — ready for code review.
- 🐛 `bug` — fixes a CI defect: the mutation gate aborts with a `ConfigError` instead of reporting an uncovered file.
- ⏭️ `skip-qa` — CI-tooling-only change (`scripts/`, `.github/workflows/mutation-tests.yml`); no UI-rendering file, no DB/API surface change, and automated tests ship in this PR.
- 🔹 `priority-medium` — an ordinary bug fix; the affected `mutate` job is already advisory (`continue-on-error`), so this does not block merges today.
- 🟢 `risk-low` — an isolated, single-area change confined to CI tooling with thorough test coverage (112 targeted tests, 8 new), no application code touched.

adeptofvoltron:
## 🤖 `om-fix` — run summary

**Status:** ready

**Files changed:**
- `scripts/stryker/relatedTests.mjs` (new)
- `scripts/stryker/mutation-changed.mjs`
- `.github/workflows/mutation-tests.yml`
- `scripts/stryker/report.mjs`
- `scripts/__tests__/stryker-related-tests.test.mjs` (new)
- `scripts/__tests__/stryker-mutation-changed.test.mjs`
- `scripts/__tests__/stryker-report.test.mjs`
- `scripts/__tests__/stryker-workflow.test.mjs`

**Summary:** A diff-scoped file with zero related Jest tests made Stryker's `enableFindRelatedTests` dry run execute zero tests, throwing `ConfigError: No tests were executed` and aborting the whole `mutate` job instead of reporting the file as uncovered (issue #5281, hit for real on #5219). Added `scripts/stryker/relatedTests.mjs`, exporting `partitionByRelatedTests`, which runs `jest --listTests --findRelatedTests <file>` per diff-scoped file (a static, dependency-graph-only check — no tests execute) and splits the list into `covered`/`uncovered`. The CI workflow now runs a new "Partition files by related test coverage" step between "Build packages" and "Run mutation testing", feeding only the `covered` subset to Stryker and skipping the Stryker step entirely when nothing is covered; `report.mjs` gained an `--uncovered`/`uncoveredFiles` surface so the job summary states which files need tests instead of Stryker ever seeing them. The local `mutation:changed` wrapper got the identical fix, since it had the same unconditional-`runStryker` gap. `scope.mjs`/`createConfig.mjs` were left untouched — coverage filtering needs `yarn install` and belongs after it, not in the lightweight `scope` job.

**Tests:** Manually reproduced the exact bug and fix by hand against `packages/shared` — a real untracked, unimported file returns empty `jest --listTests --findRelatedTests` output (exit 0), confirming the partition logic mirrors Stryker's own dry-run mechanism. Ran the targeted `node --test` suite for all affected files: 112/112 pass, including 8 new tests. Ran the full `yarn test:scripts` suite (600 tests): 596 pass, 4 fail — verified via `git stash` that those same 4 failures (`monorepo dev wrapper` env tests, `repo-wide-guards` template classification) exist identically on the unmodified `develop` baseline, unrelated to this change. Skipped the full `validation.commands` gate (`build:packages`, `generate`, `i18n:check-*`, `typecheck`, `test`, `build:app`) — this change touches only `scripts/` (not a yarn workspace, not covered by root `tsconfig.json`'s empty `include`, not linted by `turbo run lint`) and `.github/workflows/`, so those commands exercise zero changed code.

**Breaking changes:** none. `renderMarkdown`, `renderMissingReportMarkdown`, and `parseReportArgs` gained new optional parameters/fields (`uncoveredFiles`, defaulting to `[]`); all existing call shapes and output for the no-uncovered-files case are byte-identical.

adeptofvoltron:

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock — review. Started: 2026-08-17T10:40:42Z.

adeptofvoltron (commented):
# 🔍 Code Review: fix(ci): report uncovered files instead of aborting the mutation gate (#5281)

## 🎯 Summary
This PR fixes the mutation-testing CI gate aborting with `ConfigError: No tests were executed` whenever a PR adds a `src/lib/**/*.ts` file no test covers yet, instead of scoring it as uncovered (issue #5281, hit for real on #5219). It adds `scripts/stryker/relatedTests.mjs`, which runs `jest --listTests --findRelatedTests <file>` per diff-scoped file — a static dependency-graph check that executes zero tests — and partitions the mutate list into `covered`/`uncovered`. The CI workflow (`.github/workflows/mutation-tests.yml`) gets a new "Partition files by related test coverage" step that feeds only the covered subset to Stryker, skipping the Stryker step entirely when nothing is covered; `report.mjs` gained a `--covered`/`--uncovered` surface so the job summary states which files need tests. The local `mutation:changed` wrapper got the identical fix, since `mutation-changed.mjs` had the same unconditional-`runStryker` gap. The design deliberately leaves `scope.mjs`/`createConfig.mjs` untouched, since coverage filtering needs `yarn install` and belongs after it rather than in the lightweight `scope` job. Reviewed scope: `scripts/stryker/relatedTests.mjs` (new), `scripts/stryker/mutation-changed.mjs`, `scripts/stryker/report.mjs`, `.github/workflows/mutation-tests.yml`, and their four test files (one new). The fix is well-tested (67 targeted `node --test` cases, all passing) and was manually reproduced end to end against `packages/shared` — a real untracked, unimported file returns empty `jest --listTests --findRelatedTests` output before the fix's partition step existed, and is correctly routed to `uncovered` after. A second pass during this review caught and fixed a real edge case: the original `renderMissingReportMarkdown` could misreport a genuine Stryker crash (on a partially-covered package) as "every file lacks tests," which would have buried real infrastructure failures — that's now disambiguated via an explicit `--covered` argument.

## Verdict
✅ **approve** — no blockers or majors. The only validation-gate failure (`@open-mercato/queue#test`) is a pre-existing, environmental flake in an untouched package (details below); everything else is green, the fix is narrowly scoped, well-tested, and I confirmed it actually reproduces and resolves the reported failure by hand.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` (1st) | ✅ PASS | 22/22 tasks successful. |
| `yarn generate` | ✅ PASS | Clean run, no diff produced. |
| `yarn build:packages` (2nd) | ✅ PASS | 22/22 tasks successful (cached). |
| `yarn i18n:check-sync` | ✅ PASS | 52 modules, all locales in sync. |
| `yarn i18n:check-usage` | ✅ PASS | Clean. |
| `yarn typecheck` | ✅ PASS | Clean across all packages. |
| `yarn test` | ⚠️ PASS for this change / 1 unrelated pre-existing failure | `@open-mercato/queue#test` fails 3 tests in `src/__tests__/local.strategy.test.ts` (`continuous workers process new jobs without waiting for the polling interval`, `… re-arm filesystem wake-ups after the queue directory is recreated`, `… retain wake-ups received during an active batch`) with `[internal] Timed out after 800ms`. This PR makes **zero** changes to `packages/queue` (`git diff origin/develop..HEAD --stat -- packages/queue` is empty), and the failure reproduces identically in isolation (`yarn jest src/__tests__/local.strategy.test.ts` inside `packages/queue`, no other tests running). All three failing tests depend on filesystem-watch wake-ups under tight (800ms) timing budgets; this host's `fs.inotify.max_user_instances` is 128 against the repo's own documented desired value of 4096 (`dev-inotify-limits`), the same inotify-exhaustion pattern already tracked for other fs.watch-based suites in this sandbox. Every other test file across all 28 `test` tasks passed (only `@open-mercato/queue#test` failed; `grep FAIL` on the full gate log returns exactly one match). Not fixed in this PR — it is an environmental sandbox limitation in a package this change never touches, and "fixing" packages/queue's local-strategy timing here would be unrelated scope creep. GitHub Actions runners are not resource-constrained the same way and are authoritative for this check. |
| `yarn build:app` | ✅ PASS | Clean build. |

Targeted suite for the changed files: `node --test scripts/__tests__/stryker-related-tests.test.mjs scripts/__tests__/stryker-mutation-changed.test.mjs scripts/__tests__/stryker-report.test.mjs scripts/__tests__/stryker-workflow.test.mjs scripts/__tests__/stryker-scope.test.mjs scripts/__tests__/stryker-enforce.test.mjs scripts/__tests__/stryker-create-config.test.mjs` — 67/67 pass. Full `yarn test:scripts` (601 tests): 597 pass / 4 fail — those 4 (`monorepo dev wrapper` env-resolution tests, a `repo-wide-guards` classification gap on an unrelated `create-app` template test) reproduce identically on unmodified `develop` (verified via `git stash`), unrelated to this change.

## Findings

### 🔹 Minor
- **`scripts/stryker/relatedTests.mjs:42-55`** — `defaultRunJest` calls `execFileSync` with no `try/catch`. If Jest ever exits non-zero for a file in the mutate list (its own config error, a corrupted haste cache, etc. — not merely "no related tests," which I verified returns exit 0 with empty output even for a nonexistent file), the whole "Partition files by related test coverage" step throws uncaught and the step fails outright. In practice this doesn't make anything worse than the pre-fix baseline (the job-level `continue-on-error` still applies while `MUTATION_ENFORCE` is off, and I could not reproduce a nonzero exit in any tested scenario), so it isn't a blocker, but wrapping the call and treating a thrown error the same as "uncovered" (with a distinct log line, e.g. `[stryker:related-tests] jest check failed for <file>, treating as uncovered: <error message>`) would make the partition step as fail-safe as the rest of this fix's design intends. Optional — author's call whether to harden this now or track it as a fast-follow.

## 💥 Breaking Changes
- [x] No exported/public symbol removed or renamed without a deprecation path — `renderMarkdown`, `renderMissingReportMarkdown`, and `parseReportArgs` only gained new optional parameters/fields (`uncoveredFiles`, `coveredFiles`, defaulting to `[]`); every existing call shape is unaffected.
- [x] No function signature changed in a breaking way — `partitionByRelatedTests`, `runMutationChanged`'s dependency-injection options, and `renderMarkdown`/`renderMissingReportMarkdown` all take new arguments as additive, defaulted parameters.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — N/A, no HTTP surface touched.
- [x] No field removed or retyped in an existing response shape — N/A.
- [x] No event or message name renamed or removed; no payload field removed — N/A.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — `yarn mutation:changed`'s interface (`--base <ref>`) is unchanged; its behavior now correctly skips uncovered files rather than crashing, which is the intended fix, not a contract change.
- [x] No database table or column renamed or removed; no column type narrowed — N/A.
- [x] No config key renamed and no default changed silently — N/A.
- [x] Where a contract had to change: old surface kept working through a deprecation window, with migration notes — N/A, nothing here required one; confirmed against `BACKWARD_COMPATIBILITY.md`'s 13 protected-surface categories, none of which this diff touches (it doesn't modify a module convention file, public package export, event ID, widget spot ID, API route, DB schema, DI key, ACL feature, notification type, documented import path, or CLI command/flag).

## 🧪 Test Coverage
The fix is covered end to end. `scripts/__tests__/stryker-related-tests.test.mjs` (new) unit-tests `partitionByRelatedTests` directly with an injectable `runJest`, covering the covered/uncovered split, whitespace-only output, and an empty file list, plus the `--mutate` arg parser and the comma-list splitter. `scripts/__tests__/stryker-mutation-changed.test.mjs` extends the existing dependency-injection harness with an injectable `partitionRelatedTests` (defaulting every file to "covered" so the pre-existing assertions are unaffected) and adds two new cases: dropping an uncovered file while still mutating the covered ones, and skipping Stryker entirely when a whole package is uncovered. `scripts/__tests__/stryker-report.test.mjs` covers the new `--covered`/`--uncovered` argument parsing, the "Needs tests" note rendering alongside a real score and when no mutants were generated at all, and — importantly — the disambiguation added during this review: a case where some files are covered (so a missing report is a genuine crash, not a coverage skip) keeps the original crash-explanation text and only adds the uncovered note as a supplementary aside, verified via the new "keeps the crash explanation primary when some files were covered" test. `scripts/__tests__/stryker-workflow.test.mjs` verifies the new "Partition files by related test coverage" step exists between "Build packages" and "Run mutation testing," that "Run mutation testing" is conditioned on `steps.partition.outputs.covered != ''`, and that both `COVERED_FILES`/`UNCOVERED_FILES` reach the report step correctly. I additionally exercised the real (non-mocked) `jest --listTests --findRelatedTests` mechanism by hand against `packages/shared` for both a covered and an uncovered file, confirming the unit tests' mocked behavior matches Jest's actual behavior. No gaps identified.

adeptofvoltron:
🤖 `om-auto-review-pr` — automated review verdict: **approve** (0 blockers, 1 minor — optional hardening suggestion, author's call). The full report is posted above as a comment-review rather than a formal GitHub approval, because GitHub rejects self-approval for the identity that opened this PR (`adeptofvoltron`). The pipeline stays in `review` — this PR needs an **independent** approval (a human or a different automation identity) before it can move to `merge-queue`. `skip-qa` already applies (CI-tooling-only change, no manual QA needed).

adeptofvoltron:
🤖 `om-auto-review-pr` completed: review verdict **approve** (comment-review — self-approval blocked). Lock retained — chain continues.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: automated review verdict **approve** (0 blockers, 1 optional minor). Posted as a comment-review, not a formal GitHub approval, because GitHub blocks self-approval for the identity that opened this PR — an independent approval is still needed before merge. `test` CI check is still pending; swapping to `ci-monitoring` so the outstanding CI-result follow-up stays visible. Lock released.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-20T01:47:38Z. Other auto-skills will skip this PR until the lock is released.

Claim note: the previous chain (`om-auto-fix-issue` → `om-auto-review-pr`, running as @adeptofvoltron) explicitly released its lock on 2026-08-17 and left the PR assigned back to its author with `ci-monitoring`. That is not a live claim (no `in-progress` label, newest 🤖 comment ~3 days old), so this run claims the PR normally — as the **independent reviewer** the earlier run asked for, since GitHub blocked its self-approval.

pkarw (changes requested):
# 🔍 Code Review: fix(ci): report uncovered files instead of aborting the mutation gate (#5281)

Independent second-pass review, requested by the automated first pass on this PR: the earlier `om-auto-review-pr` run executed under the same identity that opened the PR (`adeptofvoltron`), so GitHub rejected its approval and it posted its verdict as a comment-review asking for an independent reviewer. This review is that pass, run under a different identity (`pkarw`) against head `7b987fd`.

## 🎯 Summary

This PR stops the mutation-testing gate from aborting when a PR changes a `src/lib/**/*.ts` file that no test covers. Because `createConfig.mjs` sets `jest.enableFindRelatedTests: true`, Stryker's dry run executes only the tests related to each mutated file; with zero related tests it executes nothing and throws `ConfigError: No tests were executed`, killing the whole `mutate` job and leaving no `mutation.json` behind (issue #5281, hit for real on #5219).

The fix adds `scripts/stryker/relatedTests.mjs`, whose `partitionByRelatedTests` runs `jest --listTests --findRelatedTests <file>` per diff-scoped file and splits the mutate list into `covered` / `uncovered`. A new "Partition files by related test coverage" step in `.github/workflows/mutation-tests.yml` feeds only the covered subset to Stryker and skips the Stryker step entirely when nothing is covered; `report.mjs` gained a `--covered` / `--uncovered` surface so the job summary renders a distinct "Needs tests" note, and deliberately keeps the crash explanation primary when some files *were* covered so a genuine Stryker crash is never mislabelled as a coverage skip. The local `yarn mutation:changed` wrapper had the identical unconditional-`runStryker` gap and gets the same partition. `scope.mjs` and `createConfig.mjs` are correctly left untouched, since coverage filtering needs an installed Jest and therefore belongs after `yarn install`, not in the lightweight `scope` job.

The core mechanism is sound and I verified it independently rather than taking the PR body's word for it (evidence in the validation section below). The approach is the right one, the diff is narrow, and the test coverage for what it does change is genuinely thorough. What it misses is the second half of its own goal: the change teaches the **reporting** step to tell "skipped because uncovered" apart from "crashed", but leaves `scripts/stryker/enforce.mjs` — the module that owns the actual pass/fail decision — unaware of the new skip path. That is the one major finding below, and it is why this is a request for changes rather than an approval.

Reviewed scope: `scripts/stryker/relatedTests.mjs` (new), `scripts/stryker/mutation-changed.mjs`, `scripts/stryker/report.mjs`, `.github/workflows/mutation-tests.yml`, and the four test files (one new), read against `.ai/specs/2026-07-31-stryker-mutation-testing-ci-gate.md`, issue #5281, `scripts/stryker/scope.mjs`, `scripts/stryker/createConfig.mjs` and `scripts/stryker/enforce.mjs`.

## Verdict

❌ **request changes** — one major finding (the enforcement step still fails closed on the new "nothing was covered" skip, which re-creates the aborted-gate outcome this PR exists to remove, the moment `MUTATION_ENFORCE` is flipped to `'true'`) plus two minors, one of them carried over unaddressed from the first review pass. No blockers: nothing here is broken on `develop` today, because enforcement ships dormant. The design, the mechanism and the test coverage are all good — the fix is simply incomplete at the layer that decides whether the job goes red.

## 🧪 Validation Gate

**Validation source: GitHub checks** (per this repository's `om-auto-review-pr` extension, which is GitHub-checks-first). Every reported check for head `7b987fd` was read from the PR rather than reproduced locally; branch protection on `develop` is not readable through the API (`404 Required status checks not enabled`), so per the descriptor every reported check is treated as required.

| Check | State | Notes |
|-------|-------|-------|
| `prepare` | ✅ SUCCESS | Run `32022105396`, head `7b987fd` — confirmed the run's `headSha` matches the PR head exactly, so these results describe this diff and not an earlier push. |
| `lint` | ✅ SUCCESS | |
| `test` | ✅ SUCCESS | This is the check that matters most here: `ci.yml:631` runs `yarn test:scripts`, and `package.json:102` expands that to `node --test scripts/__tests__/*.test.mjs …` — so all four changed/added test files ran in CI and passed. |
| `ds-lint` | ✅ SUCCESS | |
| `audit-scope` | ✅ SUCCESS | |
| `scope` | ✅ SUCCESS | Mutation workflow, run `32022105402`, same head SHA. |
| `license/cla` | ✅ SUCCESS | |
| `audit`, `docker-build`, `ephemeral-integration`, `merge-coverage`, `mutate`, `comment`, `[code]smith` | ⏭️ SKIPPED | Conditional jobs that this diff does not trigger. |

No check is failing and none is pending, so there is no CI blocker finding and nothing was waited on. No local `validation.commands` fallback was needed or run.

Two supplementary local checks, run purely as independent evidence for the findings below (not as a substitute for the gate):

- `node --test scripts/__tests__/stryker-{related-tests,mutation-changed,report,workflow,enforce,scope}.test.mjs` → **103/103 pass**.
- I reproduced the partition mechanism by hand against `packages/shared` instead of trusting the PR body's claim about it, because the whole design rests on it: `jest --config jest.config.cjs --listTests --findRelatedTests src/lib/definitely-not-a-real-file.ts` exits **0** with a single newline on stdout (so `output.trim() === ''` → `uncovered`), and the same command against `src/lib/boolean.ts` exits **0** with 95 test paths (→ `covered`). A warm invocation takes ~0.5s, so the per-file Jest startup cost is negligible against the job's 60-minute budget even at `DEFAULT_MAX_FILES` (25). The partition is a faithful, cheap mirror of Stryker's own dry-run selection.

One residual risk worth stating plainly rather than filing as a finding, because it is not actionable pre-merge: the `mutate` job is SKIPPED on this PR (it changes no in-scope `packages/shared` file), so **no CI run has actually executed the new workflow steps**. They are covered only by the static workflow-shape assertions in `stryker-workflow.test.mjs`. The `workflow_dispatch` entrypoint exists precisely to prove the runner path on demand, but it runs the workflow file from the default branch, so it cannot exercise this branch's version from a fork. The first real exercise of these steps will be the first post-merge PR that touches an in-scope file.

## Findings

### 🔶 Major

- **`scripts/stryker/enforce.mjs:134-146`** (with `.github/workflows/mutation-tests.yml:185` and `:211-218`) — **the new "everything uncovered" skip path still fails the job once enforcement is enabled, which is the outcome this PR set out to remove.** The workflow now skips "Run mutation testing" entirely when `steps.partition.outputs.covered == ''`, so no `mutation.json` is written. The "Enforce the mutation threshold" step runs `if: always()` and is *not* given the new `covered` / `uncovered` information, and `enforce.mjs` deliberately fails closed on a missing report: `if (isEnforcementEnabled()) { … process.exit(1) }`, on the documented reasoning that "an absent report is the shape a crashed or misconfigured mutation step takes". That reasoning was sound before this PR, when a missing report had exactly one cause. This PR introduces a second, entirely legitimate cause and only teaches `report.mjs` about it.

  The concrete failure: set `vars.MUTATION_ENFORCE = 'true'` (a one-variable change by design), then open a PR whose only in-scope change is a `packages/shared/src/lib/**/*.ts` file with no related test. Partition yields `covered=''` → Stryker is skipped → no `mutation.json` → `enforce.mjs` exits 1 → and because `continue-on-error` is `${{ vars.MUTATION_ENFORCE != 'true' }}`, it is now `false`, so the `mutate` job goes red. That is issue #5281's user-visible symptom — "a PR that adds an uncovered file turns the check red" — simply relocated from Stryker to the enforcer. Issue #5281 names both consumers in its own mechanism section ("the report **and enforce** steps have nothing to work with") and its Expected section asks for an uncovered file to be *reported*, not to fail.

  It is also internally inconsistent with `decideOutcome` itself, which already returns `shouldFail: false` with `'No mutants were generated, so there is nothing to score.'` when Stryker *does* run and produces zero mutants. "Stryker ran and scored nothing" passes; the semantically identical "Stryker had nothing to run" fails. And it quietly breaks the spec's stated Phase 3 property that enabling enforcement is "one value rather than a code change" (`.ai/specs/2026-07-31-stryker-mutation-testing-ci-gate.md:169`, `:321`) — with this diff, flipping that value also requires this fix.

  Suggested resolution, mirroring what was already done for `report.mjs`: pass `--covered "${COVERED_FILES}"` (and `--uncovered`) into the enforce step too, and have `enforce.mjs` return a non-failing outcome with a "N changed file(s) need tests" reason when `coveredFiles` is empty — i.e. Stryker was never invoked — while keeping the existing fail-closed behavior for the genuine-crash case (`coveredFiles` non-empty and the report still missing). That is the same `coveredFiles`-disambiguates-the-two-causes rule `renderMissingReportMarkdown` already documents, applied to the module that owns the verdict. Please cover it with a case in `scripts/__tests__/stryker-enforce.test.mjs` and an assertion in `stryker-workflow.test.mjs` that the enforce step receives the partition outputs, so the two consumers cannot drift apart again.

  If the intent is instead that an uncovered file *should* fail an enforced gate, that is a defensible product position — but it contradicts issue #5281's Expected section and this PR's own stated goal, so it needs to be an explicit, documented decision in the PR body and the spec rather than an unremarked side effect of the skip.

### 🔹 Minor

- **`scripts/stryker/relatedTests.mjs:43-56`** — *(inherited from the first review pass; still unaddressed on `7b987fd`, so it counts toward this verdict)* `defaultRunJest` calls `execFileSync` with no `try` / `catch`, so any non-zero Jest exit propagates as an uncaught throw and fails the "Partition files by related test coverage" step outright. I verified the everyday path is safe — Jest exits 0 with empty stdout for both an uncovered and a nonexistent file, because `--listTests` short-circuits before the "no tests found" check and `packages/shared/jest.config.cjs` sets `passWithNoTests: true` — so this is genuinely not the common case, which is why it stays minor. But it is load-bearing in a way that is invisible from this file: a package added to `ALLOWLISTED_PACKAGES` whose Jest config lacks `passWithNoTests`, a corrupted haste cache, or an OOM-killed Jest all turn into a hard step failure, and once `MUTATION_ENFORCE=true` removes `continue-on-error` that is a red required check — the same class of outcome this PR is fixing. Wrapping the call and treating a thrown error as `uncovered` with a distinct log line (for example `[stryker:related-tests] jest check failed for <file>, treating as uncovered: <message>`) makes the partition as fail-safe as the rest of the design already is, and costs three lines.

- **`scripts/stryker/mutation-changed.mjs:86`** — `partitionRelatedTests(entry.mutate.split(','), entry.package)` hand-rolls the split instead of using `splitMutateList`, which this PR exports from `relatedTests.mjs` for exactly this purpose and which the CI entrypoint (`main()` in the same new module) does use. The hand-rolled version neither trims nor drops blanks, so a trailing comma or a stray space in a mutate list would hand `''` or `' src/lib/x.ts'` to Jest. Nothing produces such a list today — `scope.mjs` joins cleanly — so there is no live bug, but the two entrypoints diverging on how they parse the same value is precisely the failure mode this PR is fixing (the local wrapper having drifted from CI's behavior). One-word fix: import and call `splitMutateList`.

## 💥 Breaking Changes

- [x] No public export or documented import path removed or renamed — `scripts/` is repository tooling and appears in none of `BACKWARD_COMPATIBILITY.md`'s 13 protected contract-surface categories; nothing here is a published package entry point.
- [x] No function signature changed in a breaking way — `renderMarkdown` and `parseReportArgs` gained optional fields (`uncoveredFiles`, `coveredFiles`, both defaulting to `[]`), and `renderMissingReportMarkdown(packageName)` gained two trailing parameters that both default to `[]`. Every existing call shape still compiles and, with no uncovered files, produces byte-identical output; I confirmed this against the existing report tests, which are unmodified for those paths and still pass.
- [x] No CLI command or flag renamed or removed — `report.mjs` only *adds* `--covered` / `--uncovered`, and `yarn mutation:changed`'s `--base <ref>` interface is unchanged. Its behavior change (skipping uncovered files rather than crashing) is the intended fix, not a contract change.
- [x] No event ID, widget spot ID, API route, response shape, DB column, DI key, ACL feature, notification ID or config key touched — N/A, this diff has no runtime surface at all.
- [x] No generated file edited by hand and no auto-discovery convention file altered.

## 🧪 Test Coverage

Coverage for what this PR changes is thorough, and the tests are behavioral rather than snapshot-shaped. `stryker-related-tests.test.mjs` (new) unit-tests `partitionByRelatedTests` through an injected `runJest`, covering the covered/uncovered split, whitespace-only output treated as uncovered, and an empty file list (asserting `runJest` is never called), plus `parsePartitionArgs` and `splitMutateList` including the blank-and-whitespace cases. `stryker-mutation-changed.test.mjs` extends the existing dependency-injection harness with an injectable `partitionRelatedTests` that defaults every file to covered — so the pre-existing assertions keep their meaning — and adds the "some uncovered" and "whole package uncovered, Stryker never invoked" paths. `stryker-report.test.mjs` covers the new argument parsing, the "Needs tests" note both alongside a real score and when no mutants were generated, and the crash-versus-skip disambiguation. `stryker-workflow.test.mjs` asserts the new step exists, that it precedes the mutation run, that the mutation step is conditioned on `steps.partition.outputs.covered != ''`, and that both partition outputs reach the report step.

The one gap is the flip side of the major finding: nothing asserts what the **enforce** step does when the mutation step was skipped for lack of coverage. `stryker-enforce.test.mjs` is unchanged and `stryker-workflow.test.mjs` still only asserts that the enforce step receives `MUTATION_ENFORCE` — so the regression described above is invisible to the suite. Whichever way that finding is resolved, the resolution should come with a test at both layers.

---

*autofix: skipped (not my PR — re-run with `--autofix` to fix it here).* This review does not modify the contributor's fork branch.

pkarw:

- ✅ `merge-queue` — the re-review of head `6aee12d` is an **approval**: all three findings from my previous pass (1 major, 2 minors) are fixed, verified against the shipped CLIs and not just the diff, with tests at both the unit and workflow-shape layers. Zero blockers, zero majors, zero minors — only two optional nits, which this repository's verdict rule does not block on. This replaces `changes-requested`, which the pipeline group keeps mutually exclusive.
- 🐛 `bug` — unchanged and still correct: this fixes a real CI defect, the mutation gate aborting with `ConfigError: No tests were executed` instead of reporting an uncovered file.
- ⏭️ `skip-qa` — unchanged. CI-tooling-only change (`scripts/`, `.github/workflows/mutation-tests.yml`, one spec document); no UI-rendering file, no database or API surface change, no `BACKWARD_COMPATIBILITY.md` contract touched, and automated tests for the changed behavior ship in the same PR. That is the automated-verification exemption, so no manual QA gate applies and `merge-queue` is not held by `needs-qa`.
- 🔹 `priority-medium` — unchanged. An ordinary bug fix; the affected `mutate` job is still advisory (`continue-on-error` while `MUTATION_ENFORCE` is not `'true'`), so this does not block merges today.
- 🟢 `risk-low` — unchanged. Isolated, additive, confined to CI tooling with no runtime surface; the fail-closed behavior for a genuine Stryker crash is preserved, and the one deliberate behavior change (an all-uncovered skip now passes an enforced gate) is documented in the spec's edge-case table and changelog.

Validation source: **GitHub checks** — every reported check on the exact head SHA `6aee12d` is SUCCESS or SKIPPED, so no local `validation.commands` fallback was needed. The targeted `node --test` stryker suite (122/122) and a hand-run of the `enforce.mjs` truth table were supplementary evidence only.

pkarw:
Thanks @adeptofvoltron — the independent review pass you asked for is done, and it found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

The short version of what to change:

1. **Major** — teach `scripts/stryker/enforce.mjs` about the new skip. It fails closed on a missing `mutation.json` when `MUTATION_ENFORCE=true`, and this PR gives `mutation.json` a second, legitimate reason to be absent (Stryker skipped because nothing was covered). `report.mjs` got `--covered`/`--uncovered` to disambiguate exactly that; the enforce step needs the same input, plus a test in `stryker-enforce.test.mjs` and a workflow-shape assertion. Without it, flipping the enforcement variable turns #5281's symptom back on.
2. **Minor** — wrap `defaultRunJest`'s `execFileSync` (`relatedTests.mjs:43-56`) so a thrown Jest error is treated as `uncovered` with a distinct log line, rather than failing the partition step. This is the item from the first review pass, still open on `7b987fd`.
3. **Minor** — use the `splitMutateList` you already export instead of `entry.mutate.split(',')` in `mutation-changed.mjs:86`, so the CI and local entrypoints cannot drift on how they parse the same list.

Everything else looks good: the partition mechanism is the right approach, and I verified it independently against `packages/shared` rather than taking the PR body's word for it — `jest --listTests --findRelatedTests` exits 0 with empty stdout for an uncovered/nonexistent file and lists 95 suites for `src/lib/boolean.ts`, at ~0.5s per warm invocation. CI is fully green on this head.

pkarw:

Independent second-pass review of head `7b987fd`, run under a different identity (`pkarw`) than the one that opened the PR — the first pass could not submit a formal verdict because GitHub rejects self-approval. Verdict: **request changes** on 1 major and 2 minors (one inherited from that first pass and still open). Validation source: **GitHub checks** — every reported check on the exact head SHA is SUCCESS or SKIPPED, so no local gate fallback was needed; the targeted `node --test` script suite (103/103) and a hand-verification of the `jest --listTests --findRelatedTests` mechanism were run only as supplementary evidence for the findings.

autofix: skipped (not my PR — re-run with `--autofix` to fix it here). No CI follow-up is outstanding: CI has already settled green, which is why `ci-monitoring` was removed rather than swapped in.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-08-20T06:12:56Z.

Picking up @pkarw's independent `CHANGES_REQUESTED` review from head `7b987fd` (1 major + 2 minors) and driving this PR back to merge-ready. The previous chain released its lock on 2026-08-20, so this is a fresh claim rather than a takeover.

adeptofvoltron:

- 🔍 `review` — the three findings from @pkarw's independent `CHANGES_REQUESTED` pass are all fixed and pushed as `6aee12d`, so `changes-requested` no longer describes this PR; it is back to awaiting an independent reviewer's verdict. It cannot move to `merge-queue`, because the only automated review identity available here is the one that opened the PR and GitHub rejects self-approval.
- 🐛 `bug` — unchanged. This still fixes a CI defect: the mutation gate aborting with a `ConfigError` instead of reporting an uncovered file.
- ⏭️ `skip-qa` — unchanged, and still correct after this round: the diff touches only `scripts/`, `.github/workflows/mutation-tests.yml` and `.ai/specs/`. No UI-rendering file, no database structure change, no API surface change, no `BACKWARD_COMPATIBILITY.md` contract broken, and automated tests for every changed behavior ship in the same PR — which is exactly the automated-verification exemption.
- 🔹 `priority-medium` — unchanged. The gate is dormant (`MUTATION_ENFORCE` defaults to `false`), so nothing is release-blocking today, but the defect will hit the next PR that adds an as-yet-uncovered source file.
- 🟢 `risk-low` — unchanged, and this round does not widen the blast radius: the new `enforce.mjs` behavior only changes the verdict on a path that previously could not occur at all (a skipped mutation step), and the fail-closed rule for a genuine crash is preserved verbatim.

adeptofvoltron:

**Merge-readiness: ⚠️ needs an independent reviewer.** Every requested change is fixed and pushed; the only thing standing between this PR and the merge queue is a human verdict this automation is structurally unable to give itself.

### 🔁 Base merge

`origin/develop` was merged into the head branch before anything else was judged, so the review fixes, the local gate and CI all describe the real merge result rather than a three-day-old base. The merge was clean — no conflicts, nothing to hand-resolve. New head: `6aee12d`.

### 🔍 How each reviewer ask was handled

All three findings come from @pkarw's independent second-pass review of head `7b987fd`. All three are **fixed in this PR**; none was deferred to a follow-up, because a finding that changes what an enforced gate decides is not a nit.

**🔶 Major — `enforce.mjs` still failed closed on the new "nothing was covered" skip.** Fixed, exactly along the lines the review suggested. The enforce step now receives the same `covered` / `uncovered` partition outputs the report step already got, and a new `decideMissingReportOutcome` uses them the way `renderMissingReportMarkdown` does:

- empty `covered` **plus** a non-empty `uncovered` list means Stryker was never invoked — a skip, not a crash — so the outcome is a pass with a `"N changed file(s) need tests"` reason;
- a non-empty `covered` list with the report still missing is a genuine crash and **still fails closed**, unchanged;
- both lists empty means the caller passed no partition information at all, so the two causes are indistinguishable and the fail-closed rule stands. That conservative third case is deliberate.

This resolves the concrete failure the review described: with `vars.MUTATION_ENFORCE = 'true'`, a PR whose only in-scope change is an uncovered `packages/shared/src/lib/**/*.ts` file no longer turns the `mutate` job red. It also restores the spec's Phase 3 property that enabling enforcement is one repository-variable change and not a code change. The review's alternative reading — that an uncovered file *should* fail an enforced gate — was not taken, because it contradicts issue #5281's Expected section; the passing verdict is now the documented, tested behavior rather than an unremarked side effect.

**🔹 Minor — `defaultRunJest` had no `try` / `catch`.** Fixed, and wrapped one level up in `partitionByRelatedTests` rather than inside `defaultRunJest`, so the fail-safe covers any injected runner and is directly testable. A throwing Jest check is treated as `uncovered` and logged as `[stryker:related-tests] jest check failed for <file>, treating as uncovered: <message>`. The partition step can no longer be the thing that turns the job red — which was the point, since that is the same class of outcome this PR exists to remove.

**🔹 Minor — the local wrapper hand-rolled `entry.mutate.split(',')`.** Fixed: `mutation-changed.mjs` now imports and calls `splitMutateList`, so the local wrapper and the CI entrypoint read the same value the same way and cannot drift apart again.

**📝 Spec.** The review noted the change "quietly breaks" the spec's one-variable property. Rather than only restoring it in code, `.ai/specs/2026-07-31-stryker-mutation-testing-ci-gate.md` gains an Edge Cases row for an in-scope file no test covers yet, and a changelog entry recording why both consumers — report *and* enforce — must read the partition.

### 🧪 Tests

The review asked for coverage at both layers, and both are covered:

- `stryker-enforce.test.mjs` — the skip path passing under enforcement, the crash path still failing closed, the no-partition-information path still failing closed, the new `--covered` / `--uncovered` parsing, an end-to-end CLI case asserting exit `0` with `MUTATION_ENFORCE=true`, and an explicit assertion that "nothing was covered" is scored identically to `decideOutcome`'s "no mutants were generated".
- `stryker-workflow.test.mjs` — the enforce step reads the same `steps.partition.outputs.*` the report step does, so the two consumers cannot silently drift apart again.
- `stryker-related-tests.test.mjs` — a throwing Jest check lands in `uncovered` and emits the warning, without aborting the partition.
- `stryker-mutation-changed.test.mjs` — the wrapper hands the partition a parsed file list.

**Local validation.** `node --test scripts/__tests__/*.test.mjs` (the `yarn test:scripts` gate) — **590 passed, 0 failed**. Worth noting against the original PR body: the 4 pre-existing failures reported there are gone, because merging the current `develop` picked up their fixes. The stryker subset alone is **122 passed** (was 112; this round adds 10). The rest of `validation.commands` (`build:packages`, `generate`, `i18n:check-*`, `typecheck`, `build:app`) exercises none of the changed code — `scripts/` is not a yarn workspace and sits outside the root `tsconfig.json` `include` — so it was deliberately not run, the same reasoning the PR body already records.

**CI is still running on this head.** Branch protection plus the required independent approval hold the actual merge; this run's verdict covers the work, not a green run. Checks pending at the time of writing: `prepare`, `lint`, `ds-lint`, `scope`, `test`. A follow-up comment will report the CI outcome.

### 📸 UI

n/a — the diff touches `scripts/`, `.github/workflows/mutation-tests.yml` and `.ai/specs/` only. There is no user-facing surface to drive, which is also why `skip-qa` is the right meta label rather than `needs-qa`.

### 🔁 Follow-ups filed

None. Every finding was fixed in-loop, so there was nothing non-blocking left to defer.

One residual risk from the review is restated here rather than filed, because the reviewer explicitly judged it non-actionable pre-merge and filing an issue for it would be noise: the `mutate` job is SKIPPED on this PR (the diff changes no in-scope `packages/shared` file), so **no CI run has executed the new workflow steps** — they are covered only by the static shape assertions in `stryker-workflow.test.mjs`. `workflow_dispatch` runs the workflow file from the default branch and so cannot exercise this branch's version from a fork. The first real exercise of these steps is the first post-merge PR that touches an in-scope file, or a `workflow_dispatch` run once this is on `develop`.

### 🚀 Hand-off

This skill never merges. Normally the next step would be `om-approve-merge-pr 5343`, but that gate cannot pass yet, and the reason is worth stating plainly rather than leaving for someone to rediscover:

**@pkarw's `CHANGES_REQUESTED` review is still the active verdict on this PR, and only a reviewer can clear it.** This automation runs as `adeptofvoltron`, the identity that opened the PR, and GitHub rejects self-approval — so no amount of further automated work can turn this green. @pkarw has been re-requested as a reviewer. Once that pass lands on head `6aee12d` and CI is green, the PR is ready for `om-approve-merge-pr`; `skip-qa` means no QA sign-off is owed.

adeptofvoltron:

**✅ Green on head `6aee12d`.** The checks the summary comment listed as pending have all settled, and none failed. Branch protection on `develop` is not readable through the API, so per this repository's tracker descriptor every reported check is treated as required.

| Check | Result | |
|-------|--------|--|
| [`prepare`](https://github.com/open-mercato/open-mercato/actions/runs/32339081531/job/96334338935) | ✅ pass | 53s |
| [`test`](https://github.com/open-mercato/open-mercato/actions/runs/32339081531/job/96334541117) | ✅ pass | 4m5s — this is the one that matters here: `yarn test:scripts` runs all five changed/added stryker test files |
| [`lint`](https://github.com/open-mercato/open-mercato/actions/runs/32339081531/job/96334338904) | ✅ pass | 29s |
| [`ds-lint`](https://github.com/open-mercato/open-mercato/actions/runs/32339081531/job/96334338920) | ✅ pass | 46s |
| [`scope`](https://github.com/open-mercato/open-mercato/actions/runs/32339081523/job/96334339084) | ✅ pass | 31s — mutation workflow |
| [`audit-scope`](https://github.com/open-mercato/open-mercato/actions/runs/32339081531/job/96334338752) | ✅ pass | 14s |
| `audit`, `docker-build`, `ephemeral-integration`, `merge-coverage`, `mutate`, `comment`, `[code]smith` | ⏭️ skipped | conditional jobs this diff does not trigger |

CI agrees with the local gate reported earlier (`node --test scripts/__tests__/*.test.mjs`, 590 passed), so the verdict in the run summary stands unchanged and the pipeline label stays `review`. `ci-monitoring` is removed — this follow-up is the comment it was promising, and nothing further is owed on the CI side.

`mutate` is skipped again, for the reason already noted: this diff changes no in-scope `packages/shared` file, so the new workflow steps still have not been exercised by a real runner. That is unchanged by this run and remains a post-merge verification item, not a blocker.

**This PR is still not mergeable, and CI is not why.** @pkarw's `CHANGES_REQUESTED` verdict remains the active review state; every finding behind it is fixed and pushed, but clearing it needs an independent reviewer, since GitHub rejects self-approval for the identity that opened this PR. Once that approval lands on `6aee12d`, this is ready for `om-approve-merge-pr 5343` — `skip-qa` means no QA sign-off is owed.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-21T00:13:23Z. Other auto-skills will skip this PR until the lock is released.

Claim note: this is a **re-review**. My previous independent pass requested changes on head `7b987fd` (1 major, 2 minors); @adeptofvoltron's `om-auto-fix-pr` run pushed fixes as `6aee12d` and released its own claim ~24h ago (no `in-progress` label, newest 🤖 comment well outside both the 30-minute claim window and the 60-minute stale window, assignee left as the author by my own review handoff). That is not a live lock, so this run claims the PR normally to re-review the new head.

pkarw (approved):
# 🔍 Code Review: Re-review — fix(ci): report uncovered files instead of aborting the mutation gate (#5281)

Second independent pass on this PR, against head `6aee12d`. My previous review of head `7b987fd` requested changes on one major and two minors; `om-auto-fix-pr` merged `origin/develop` into the branch and pushed `6aee12d` in response. This re-review re-checks each of those three findings against the new head, reviews the delta on its own merits, and re-states the verdict.

## 🎯 Summary

The change stops the mutation-testing gate from aborting when a PR touches a `src/lib/**/*.ts` file that no test covers. Because `createConfig.mjs` sets `jest.enableFindRelatedTests: true`, Stryker's dry run executes only the tests related to each mutated file; with zero related tests it executes nothing and throws `ConfigError: No tests were executed`, killing the `mutate` job and leaving no `mutation.json` (issue #5281, hit for real on #5219). `scripts/stryker/relatedTests.mjs` now partitions each package's mutate list into `covered` / `uncovered` with `jest --listTests --findRelatedTests` — a static dependency-graph query, not a test run — and the workflow feeds Stryker only the covered subset, skipping the mutation step entirely when nothing is covered.

The new head completes the half of that story the previous pass found missing. `scripts/stryker/enforce.mjs` — the module that owns the pass/fail verdict — now receives the same partition outputs the report step already got, and `decideMissingReportOutcome` uses `coveredFiles` to tell the two causes of an absent `mutation.json` apart: an empty covered list with a non-empty uncovered list is the skip path and passes, a non-empty covered list is a genuine crash and still fails closed, and no partition information at all keeps the fail-closed rule. `partitionByRelatedTests` now treats a throwing Jest check as `uncovered` with a log line instead of propagating, and `mutation-changed.mjs` parses its mutate list through the shared `splitMutateList` so the local wrapper and the CI entrypoint cannot drift apart on how they read the same value.

I verified all three fixes by exercising the shipped CLIs rather than reading the diff alone (evidence in the validation section). What I like about this iteration beyond the fixes themselves: each one landed with tests at *both* layers the previous review asked for — the unit behaviour and the workflow shape that wires it — so the report step and the enforce step can no longer silently diverge on the partition outputs. The spec gained an edge-case row and a dated changelog entry, so the design property that "an uncovered file is reported, never failed" is now written down rather than implied by code. That matters, because it is a real, deliberate policy change to what an enforced gate accepts, and it should not have to be re-derived from `enforce.mjs` later.

Reviewed scope: the delta `9a980d64..6aee12d6` in full (`scripts/stryker/enforce.mjs`, `scripts/stryker/relatedTests.mjs`, `scripts/stryker/mutation-changed.mjs`, `.github/workflows/mutation-tests.yml`, `.ai/specs/2026-07-31-stryker-mutation-testing-ci-gate.md`, four test files), re-read against the full PR diff from the merge base `1dbd6f5f`, issue #5281, and `scripts/stryker/{scope,createConfig,report}.mjs`. I also confirmed the base merge `9a980d64` is a clean merge with no hand edits (`git diff-tree --cc` is empty) and that the branch still merges cleanly against the current `develop` tip `f4c01c4b`, which is four commits ahead of the base this head was tested on.

## Verdict

✅ **approve** — every finding from the previous pass is fixed, and the fixes are correct rather than merely present. Zero blockers, zero majors, zero minors; two nits below are optional and entirely the author's call. Nothing in the delta introduced a new problem: it is additive, confined to CI tooling, has no runtime surface, and the fail-closed behaviour for a genuine Stryker crash is preserved exactly.

## 🧪 Validation Gate

**Validation source: GitHub checks.** Per this repository's `om-auto-review-pr` extension, CI check results for the exact head SHA are preferred over re-running `validation.commands` locally. Every reported check on head `6aee12d` was read from the PR; branch protection on `develop` is not readable through the API (`404 Required status checks not enabled`), so per the tracker descriptor every reported check is treated as required.

| Check | State | Notes |
|-------|-------|-------|
| `prepare` | ✅ SUCCESS | Run `32339081531`, head `6aee12d`. |
| `test` | ✅ SUCCESS | The check that matters here: `ci.yml` runs `yarn test:scripts`, which expands to `node --test scripts/__tests__/*.test.mjs`, so all five changed/added test files ran in CI and passed. |
| `lint` | ✅ SUCCESS | |
| `ds-lint` | ✅ SUCCESS | |
| `audit-scope` | ✅ SUCCESS | |
| `scope` | ✅ SUCCESS | Mutation workflow, run `32339081523`, same head SHA. |
| `audit`, `docker-build`, `ephemeral-integration`, `merge-coverage`, `mutate`, `comment`, `[code]smith` | ⏭️ SKIPPED | Conditional jobs this diff does not trigger. |

No check is failing and none is pending, so there is no CI blocker finding, nothing was waited on, and no local `validation.commands` fallback was needed or run. The configured gate commands (`build:packages`, `generate`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `test`, `build:app`) were therefore not re-run locally; the change touches only `scripts/` — not a yarn workspace, outside the root `tsconfig.json` include set — and `.github/workflows/`, so they exercise none of the changed code, and CI's own `test` job is the authoritative signal for what does.

Two supplementary local checks, run as independent evidence for the findings below rather than as a substitute for the gate:

- `node --test scripts/__tests__/stryker-{related-tests,mutation-changed,report,workflow,enforce,scope,create-config}.test.mjs` → **122/122 pass** (103 on the previously reviewed head, so the delta is 19 net-new assertions across the three fixes).
- I exercised the shipped `enforce.mjs` CLI directly with the exact argument shape the workflow builds, because the whole major finding turns on it: `--covered "" --uncovered "src/lib/fresh.ts"` with `MUTATION_ENFORCE=true` exits **0** with `1 changed file(s) need tests, so the mutation run was skipped and there is nothing to score`; `--covered "src/lib/boolean.ts" --uncovered ""` exits **1** with `failing closed`; passing no partition flags at all exits **1**; and both flags empty (the shape a crashed partition step produces) exits **1**. `report.mjs --covered "" --uncovered "src/lib/fresh.ts" --enforced` renders the "Needs tests" summary rather than the crash message. That is the full truth table the fix claims, confirmed end to end and not just in unit tests.

## Findings

### ⛔ Blocker

None.

### ⚠️ Major

None. The single major from the previous pass is resolved — see the re-check below.

### 🔹 Minor

None. Both minors from the previous pass are resolved — see the re-check below.

### 💅 Nit

- **`.github/workflows/mutation-tests.yml:243-300`** — the PR-comment surface does not carry the "Needs tests" note the job summary now does. The `comment` job rebuilds the markdown by globbing downloaded `mutation-report-*/mutation.json` artifacts and calling `report.mjs` without `--covered` / `--uncovered`, so for a package where every file was uncovered there is no `mutation.json` to glob and the comment body ends up as just the marker, while a partially-covered package gets a score with no mention of the files that were skipped. This is not a regression — a crashed Stryker run has always produced the same empty comment, the job is documented as a convenience surface with the job summary as primary, and it carries `continue-on-error: true` — so nothing here is broken. But "report it as an uncovered file needing tests" is this PR's stated goal, and one of the two surfaces a reviewer actually looks at does not do it. If it is worth closing, the cheapest route is to upload the partition result alongside the report artifact (or write it into the artifact directory even when Stryker is skipped) so the comment job can pass the same two flags the summary step does. Entirely optional, and a reasonable follow-up issue rather than work for this PR.

- **`scripts/stryker/report.mjs:192-198`, `scripts/stryker/enforce.mjs:107-112`** — there are now three near-identical comma-list parsers: `splitMutateList` exported from `relatedTests.mjs`, plus a private `splitFileList` in each of `report.mjs` and `enforce.mjs`. Unlike the divergence I flagged last time in `mutation-changed.mjs`, these three are behaviourally identical (all trim and drop blanks), so there is no bug and no live inconsistency — `report.mjs`'s copy omits the `String(value ?? '')` guard the other two have, but `parseReportArgs` only ever hands it an `argv` element, which is always a string. It is purely a consolidation opportunity: `enforce.mjs` already imports from `report.mjs`, so exporting one parser and importing it in the other two would remove the drift surface entirely. Author's call, and safely deferrable.

## ♻️ Re-check of the previous review's findings

- **Major — `enforce.mjs` failed closed on the new "nothing was covered" skip** → **fixed, and fixed the way the finding asked.** The enforce step now receives `COVERED_FILES` / `UNCOVERED_FILES` from `steps.partition.outputs` and passes them as `--covered` / `--uncovered`; `parseEnforceArgs` parses them; and the new `decideMissingReportOutcome` (`enforce.mjs:165-188`) applies exactly the `coveredFiles`-disambiguates-the-two-causes rule `renderMissingReportMarkdown` already documented. The doc comment explains *why* the old fail-closed reasoning stopped holding — a missing report used to have one cause and now has two — instead of just asserting the new behaviour. The fail-closed path for a genuine crash and for the no-information case is preserved, which was the part I most wanted not to lose. Verified against the CLI, not only the unit tests (above). Test coverage arrived at both layers as requested: five new `stryker-enforce.test.mjs` cases including a CLI-level one and an explicit assertion that the skip is scored identically to `decideOutcome`'s existing "no mutants were generated" verdict, plus a `stryker-workflow.test.mjs` case asserting the enforce step reads the same partition outputs as the report step so the two cannot drift apart again.

- **Minor — `defaultRunJest` had no `try`/`catch`, so a throwing Jest failed the partition step** → **fixed.** `partitionByRelatedTests` (`relatedTests.mjs:46-58`) wraps the call, treats a throw as `uncovered`, and logs `[stryker:related-tests] jest check failed for <file>, treating as uncovered: <message>` — the exact fail-safe shape suggested. Worth noting for future readers: the log goes to **stderr**, which is load-bearing, because the partition step's stdout is redirected into `$GITHUB_OUTPUT`; a warning on stdout would have corrupted the step outputs. The new test asserts both the partition result and the log line.

- **Minor — `mutation-changed.mjs` hand-rolled `entry.mutate.split(',')`** → **fixed.** Line 86 now calls the shared `splitMutateList(entry.mutate)`, and the new "hands the partition a parsed file list, the same way the CI entrypoint does" test pins the parsed list, so the local wrapper and CI cannot drift apart on parsing again.

## 💥 Breaking Changes

- [x] No exported or public symbol removed or renamed — `scripts/` is repository tooling and appears in none of `BACKWARD_COMPATIBILITY.md`'s 13 protected contract-surface categories; nothing here is a published package entry point.
- [x] No function signature changed in a breaking way — this head adds `decideMissingReportOutcome` (new export) and extends `parseEnforceArgs`'s returned object with `coveredFiles` / `uncoveredFiles`, both defaulting to `[]`. The previously reviewed additions to `renderMarkdown`, `renderMissingReportMarkdown` and `parseReportArgs` remain optional-with-defaults. Every existing call shape still works and produces identical output when no partition information is supplied.
- [x] No CLI command or flag renamed or removed — `enforce.mjs` only *adds* `--covered` / `--uncovered`, mirroring the flags `report.mjs` gained earlier in this PR. Existing invocations without them behave exactly as before, including the fail-closed rule.
- [x] No event ID, widget spot ID, API route, response shape, DB column, DI key, ACL feature, notification ID or config key touched — this diff has no runtime surface at all.
- [x] No generated file edited by hand and no auto-discovery convention file altered.

One behavioural change is worth naming explicitly even though it breaks no contract: with `MUTATION_ENFORCE='true'`, a package whose changed files are *all* uncovered now passes the gate where it previously would have failed on the missing report. That is the intended fix and matches `decideOutcome`'s existing treatment of a zero-mutant run, and the spec's new edge-case row and changelog entry state it in writing, so it is a documented decision rather than an unremarked side effect.

## 🧪 Test Coverage

Coverage is thorough and behavioural rather than snapshot-shaped, and the gap the previous review identified is closed. `stryker-enforce.test.mjs` now covers the skip path under enforcement, the genuine-crash path that must still fail closed, the no-partition-information path, the argument parsing including blanks and surrounding whitespace, an equivalence assertion tying the skip verdict to the existing "no mutants were generated" verdict, and a CLI-level case that runs the real script with `MUTATION_ENFORCE=true` and asserts exit 0. `stryker-workflow.test.mjs` pins the wiring at the workflow level for both consumers. `stryker-related-tests.test.mjs` covers the throwing-Jest path including the log message. `stryker-mutation-changed.test.mjs` pins the parsed file list handed to the partition. All 122 assertions across the seven stryker suites pass locally and in CI.

The one residual risk is unchanged from the previous pass and remains non-actionable pre-merge, so I am stating it rather than filing it: the `mutate` job is SKIPPED on this PR — it changes no in-scope `packages/**` file — so **no CI run has executed the new workflow steps end to end**. They are covered only by the static workflow-shape assertions plus my local CLI runs. The `workflow_dispatch` entrypoint exists to prove the runner path on demand, but it runs the workflow file from the default branch and so cannot exercise a fork branch's version. The first real exercise will be the first post-merge PR that touches an in-scope file, and the failure mode if something is wrong there is an advisory job going red while `MUTATION_ENFORCE` is still `false` — visible, contained, and not merge-blocking.

---

*autofix: skipped (not my PR — re-run with `--autofix` to fix it here).* This review does not modify the contributor's fork branch. Nothing needs fixing: the two nits are optional and explicitly the author's call.

pkarw:

Independent re-review of head `6aee12d` under a different identity (`pkarw`) than the one that opened the PR. My previous pass on `7b987fd` requested changes on 1 major and 2 minors; all three are fixed on this head, and I re-checked each one against the shipped code rather than the commit message:

- **Major (enforce step failed closed on the all-uncovered skip)** — fixed. The enforce step now reads the same `steps.partition.outputs` the report step does, and `decideMissingReportOutcome` disambiguates the two causes of a missing `mutation.json`. I ran the real CLI with the exact argument shape the workflow builds: all-uncovered + `MUTATION_ENFORCE=true` exits 0 with "need tests", covered-non-empty exits 1 "failing closed", no partition info exits 1, both-empty exits 1. The fail-closed rule for a genuine crash is intact.
- **Minor (throwing Jest failed the partition step)** — fixed, treated as uncovered with a stderr log line (stderr matters here: the step's stdout is `$GITHUB_OUTPUT`).
- **Minor (hand-rolled mutate-list split)** — fixed, now uses the shared `splitMutateList`, pinned by a test.

Tests landed at both layers asked for: 122/122 across the seven stryker suites locally (103 on the previous head), including a workflow-shape assertion that the report and enforce steps read the same partition outputs so they cannot drift apart again. The spec gained an edge-case row and a dated changelog entry, so the deliberate policy change — an all-uncovered skip passes an enforced gate — is written down rather than implied.

Two nits remain, both optional and explicitly the author's call, neither blocking: the PR-comment surface still does not carry the "Needs tests" note the job summary does (pre-existing, and the comment job is documented as secondary and `continue-on-error`), and there are now three behaviourally identical comma-list parsers that could be consolidated into one export.

Validation source: **GitHub checks** — `prepare`, `test`, `lint`, `ds-lint`, `audit-scope` and `scope` are all SUCCESS on head `6aee12d`; the rest are SKIPPED conditionals. Nothing is failing and nothing is pending, so no local gate fallback was run and no CI follow-up is outstanding — hence no `ci-monitoring` label. I also confirmed the base merge `9a980d64` is a clean merge with no hand edits, and that the branch still merges cleanly against the current `develop` tip `f4c01c4b`.

Pipeline moved to `merge-queue`. `skip-qa` applies, so the QA-approval gate does not hold this PR. Assignee left with @adeptofvoltron; I unassigned myself now that the review is done, so nothing reads as a live automation claim.

Issue #1100, opened 2026-03-30
list-comments ignores entityId filter — returns ALL comments


matgren:
## Affected
`GET /customers/comments?entityId=<uuid>`

## Description
The `list-comments` endpoint ignores the `entityId` query parameter. Instead of returning comments associated with a specific entity (person), it returns ALL comments in the system regardless of the provided entityId.

## Reproduction
1. Create a comment with `entityId = personA`
2. Create a comment with `entityId = personB`
3. `GET /customers/comments?entityId=personA`
4. **Result:** both comments (personA and personB) are returned — expected only personA

## Impact
- Reading a lead's notes returns notes from ALL persons (data leak across entities)
- Deleting a lead cascades to deleting ALL comments in CRM (not just that person's)
- Large comment sets cause response buffer overflow

## Expected behavior
`entityId` filter should return only comments where `entityId` matches the provided UUID.

amtmich:
@matgren Tested manually by adding notes to two separate person records via the UI (`/backend/customers/people-v2/34b870eb-734d-4909-aff3-88b94b3b664a` and `/backend/customers/people-v2/41237aba-8bdf-468a-a66b-0100e99fa2a2`). Notes were visible only under the correct person in the UI with no cross-entity leaking.

Also verified via direct API calls:
- `GET /api/customers/comments?entityId=34b870eb-734d-4909-aff3-88b94b3b664a` — returned 1 comment, `entity_id` matched the requested person
- `GET /api/customers/comments?entityId=41237aba-8bdf-468a-a66b-0100e99fa2a2` — returned 2 comments, both with `entity_id` matching the requested person

Could not reproduce the reported behavior. Is this still valid and reproducible on your end?

michal1986:
> [@matgren](https://github.com/matgren) Tested manually by adding notes to two separate person records via the UI (`/backend/customers/people-v2/34b870eb-734d-4909-aff3-88b94b3b664a` and `/backend/customers/people-v2/41237aba-8bdf-468a-a66b-0100e99fa2a2`). Notes were visible only under the correct person in the UI with no cross-entity leaking.
> 
> Also verified via direct API calls:
> 
>     * `GET /api/customers/comments?entityId=34b870eb-734d-4909-aff3-88b94b3b664a` — returned 1 comment, `entity_id` matched the requested person
> 
[...]

PR #5345, opened 2026-08-17
feat(customers): accept dealId-only comment creation by deriving the linked entity

mniarc:
## Problem

API clients that hold only a `dealId` cannot create a deal-scoped note directly. `commentCreateSchema` requires `entityId`, so such a client first has to read existing comments (or the deal's links) to find an `entity_id` before it can POST. The UI is unaffected: the deal page's Notes composer always sends the entityId picked in its customer selector (verified with a request capture on a live instance; the original claim in this PR that the deal page posts `entityId: null` was wrong, see review discussion).

## Change

Deliberate, additive widening of the create contract on `POST /api/customers/comments`:

- `commentCreateSchema` now accepts `entityId` **or** `dealId` (both uuid-validated); a request with neither still fails validation, and a malformed `dealId` fails at the schema before any database read
- when `entityId` is absent, the route derives it from the deal's links, scoped by tenant/organization through the deal
- the choice is deterministic and mirrors the deal page's own preference order: primary person link, then first person, then first company (the company link entity has no isPrimary field)
- an explicit `entityId` is never overridden
- a `dealId` whose deal has no linked person or company returns a clean 422
- the preference order lives in one shared helper (`commands/shared.ts`), used by both this route and the `customers.manage_deal_comment` AI tool, which previously had its own un-ordered copy (with a `tenantId` filter on a link entity that has no such column); both entry points now land on the same entity
- the OpenAPI export follows the schema change, and the create endpoint's description states the derived-placement behavior

A note created this way is parented to the derived entity, so it also appears on that record's Notes tab, same as a note created from the deal page composer.

## Testing

- unit: `packages/core/src/modules/customers/api/comments/__tests__/entityId-derivation.test.ts` (derivation preference incl. isPrimary and createdAt/id tiebreaks, explicit entityId wins, malformed dealId rejected at the schema, no-links deal 422, scope filter shape) — the derivation tests exercise the real shared helper, not a mirror
- UI: `NotesSection.test.tsx` now asserts the deal composer always sends the picker's concrete entityId, and that with no entity options the add action is disabled and no request is made
- integration: `packages/core/src/modules/customers/__integration__/TC-CRM-5345.spec.ts` modeled on TC-CRM-020: dealId-only create asserts the derived `entity_id`, plus the no-links negative case pinned to 422
- `yarn build:packages`, `yarn typecheck`, comments + ai-tools + NotesSection unit suites, `yarn i18n:check-sync` pass locally


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T13:23:42Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(customers): derive entityId for deal-scoped comments

## 🎯 Summary

This PR adds a server-side derivation step to `POST /api/customers/comments`: when the request carries no `entityId` but does carry a `dealId`, `create.mapInput` looks up the deal's linked company (falling back to its linked person) and injects that entity id into the payload before `commentCreateSchema.parse` runs (`packages/core/src/modules/customers/api/comments/route.ts:87-100`). The stated motivation is that the deal detail page renders `<NotesSection entityId={null} …>`, so — per the PR body — every note added from a deal is rejected with `400 Invalid input`.

The diagnosis of the data model is accurate and the write-up is clear: `commentCreateSchema.entityId` really is required (`data/validators.ts:240-241`), `customer_comments.entity_id` really is `NOT NULL` (`data/entities.ts:731-732`), and the deal page really does pass `entityId={null}` (`backend/customers/deals/[id]/page.tsx:572`). Placing the work in `mapInput` rather than in the command is also the right layer if a derivation is wanted at all, and the code is short and readable.

The problem is what sits between those two facts. I traced the deal page's Notes composer end-to-end and then reproduced it in jsdom, and **the deal detail page does not post `entityId: null`** — it posts the id the user picked (or the first linked entity) from a selector that only that page renders. I could not reproduce the reported 400 in either of the two states that page can be in, and no other caller in the repository (UI adapter, AI tool, integration specs) ever omits `entityId`. That makes the premise of the change unproven, and the change itself introduces one behaviour regression and a silent widening of the request contract. Details and the exact evidence are in the Blocker below.

Reviewed scope: the single changed file, plus its callers and contract surfaces — `components/detail/notesAdapter.ts`, `packages/ui/src/backend/detail/NotesSection.tsx`, `backend/customers/deals/[id]/page.tsx`, `commands/comments.ts`, `commands/shared.ts`, `data/validators.ts`, `data/entities.ts`, the customers integration specs, and the `customers.manage_record_comment` AI tool.

## Verdict

❌ **request changes** — the Blocker (the reported defect does not reproduce; the derived branch is unreachable from the product) plus two Majors (a `400 → 500` regression for malformed `dealId`, and zero test coverage for a behaviour change) each independently require changes under this repository's verdict rule.

## 🧪 Validation Gate

**Validation source: GitHub checks** (repo-local checks-first extension) — no local fallback gate was run. Every reported check on head `2055ee03` has settled, none failing and none pending, so CI is the authoritative gate here and I mapped each configured command to the job that covers it rather than asserting "CI is green".

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | covered by the `prepare` job |
| `yarn generate` | ✅ PASS | covered by `prepare` (no generator churn — the diff adds no discovered module file) |
| `yarn build:packages` | ✅ PASS | covered by `prepare` |
| `yarn i18n:check-sync` | ✅ PASS | covered by the `test` job; the diff adds no user-facing string |
| `yarn i18n:check-usage` | ✅ PASS | covered by the `test` job |
| `yarn typecheck` | ✅ PASS | covered by the `test` job |
| `yarn test` | ✅ PASS | `test` job green — but see Test Coverage: no test in the repository exercises the new branch |
| `yarn build:app` | ✅ PASS | covered by `prepare` |

Also green: `lint`, `ds-lint`, `audit-scope`, `scope`, and `ephemeral-integration (none)`. That last one is the affected-only integration runner (`.github/workflows/ci.yml:106-121` resolves `modules=customers` for this diff and drops the 15-shard matrix to a single filtered runner), so the customers integration specs did run — and they all pass `entityId` explicitly, which is precisely why they cannot see this change. Head is `MERGEABLE` with no conflicts against `develop`, and the branch already contains the current `develop` tip.

## Findings

### ⛔ Blocker

**B1 — The reported defect does not reproduce, and the new branch is unreachable from the deal detail page.** `route.ts:91`

The PR body states: *"Reproducible on a fresh install: open any deal → Notes → add note → 400."* I could not reproduce that, in either state the page can be in, and the reason is that `entityId={null}` on `NotesSection` is not "no entity" — it is "let the user choose", and only the deal page uses that mode.

The deal page passes `entityOptions={entityOptions}` next to `entityId={null}` (`backend/customers/deals/[id]/page.tsx:572-575`), where `entityOptions` is the deal's linked people and companies (`:132-142`, `:164-167`). Inside `NotesSection`, when `entityOptions` is non-empty the component seeds `selectedEntityId` from the first option and renders an "Assign to customer" `Select` (`packages/ui/src/backend/detail/NotesSection.tsx:394-409`, `:1026-1049`); `resolvedEntityId` is then that selection, not the `entityId` prop (`:411-414`), and it is `resolvedEntityId` that goes to `dataAdapter.create` (`:648`). When `entityOptions` *is* empty, `hasEntity` is `false` (`:421`), which disables the add-note action (`:567`), force-closes the composer (`:606-612`), and short-circuits `handleCreateNote` with a flash before any request is made (`:627-630`).

I verified this by rendering the component with exactly the deal page's prop shape and capturing what the adapter receives (`@open-mercato/ui`, jsdom, both cases in one run):

| Case | Props | Result |
|------|-------|--------|
| A — deal *with* linked entities | `entityId={null}`, `dealId="deal-1"`, `entityOptions=[company-1, person-1]` | Add-note button **enabled**; `dataAdapter.create` called with `{"entityId":"company-1","body":"Deal note","appearanceIcon":null,"appearanceColor":null,"dealId":"deal-1"}` |
| B — deal with *no* linked entities | `entityId={null}`, `dealId="deal-1"`, `entityOptions=[]` | Add-note button **disabled**, composer never opens, `create` called **0 times** |

So case A already returns 201 today, and case B never issues the POST that the fix is meant to rescue. The 400 the PR describes sits between the two and I cannot find the state that produces it.

Nothing else in the repository posts a comment without `entityId` either: `notesAdapter.create` always forwards the resolved id (`components/detail/notesAdapter.ts:73-91`); the `customers.manage_record_comment` AI tool `superRefine`s `personId || companyId` as required for `create` (`ai-tools/activities-tasks-pack.ts:860-868`) and sends it as `entityId` (`:960-969`); and every integration spec sends it explicitly (`__integration__/TC-CRM-020.spec.ts:45`, `TC-CRM-011.spec.ts:33-37`, `TC-CRM-016.spec.ts:26-30`, `TC-CRM-042.spec.ts:51`).

This also runs against a deliberate platform decision rather than an oversight: `requireTimelineParentEntity` goes out of its way to detect a deal id passed as `entityId` and reject it with a specific `422 "entityId must reference a person or company, not a deal"` (`commands/shared.ts:74-82`). The model is "a comment is parented to a person or company, and *optionally* references a deal" — which is exactly what the deal page's picker implements.

What I would ask for before this can move:

1. **The concrete repro** — the exact request body and the deal's state (does it have linked people/companies?), or a HAR/network capture of the failing POST. If there genuinely is a UI state that posts `entityId: null`, that is a real bug and I would like to see it; the fix would then most likely belong in the page or in `NotesSection`, not in the route.
2. If the actual complaint is case B — *a deal with no linked person or company gives you a disabled "Add note" button with no explanation* — then that is a legitimate UX defect and worth fixing, but the fix is an explanatory empty state (or letting the user attach a person/company from the composer), not a server-side guess. Deriving nothing is impossible there anyway: with no links, `derived` is `undefined` and the request still 400s.
3. If the goal is an **API convenience** for non-UI clients that hold only a `dealId`, say so and make it a deliberate contract change: express it in the schema (`commentCreateSchema.superRefine` requiring `entityId` *or* `dealId`), update the OpenAPI export, scope the lookup, make the entity choice deterministic, and ship tests. See m1–m3 below, which are all consequences of doing it in `mapInput` instead.

### ⚠️ Major

**M1 — Unvalidated `dealId` now reaches the database, turning a `400` into a `500`.** `route.ts:91-96`

The derivation runs *before* `commentCreateSchema.parse` (`:100`), and the only gate on the value is `typeof scoped.dealId === 'string'`. The route's own request schema is `z.object({}).passthrough()` (`:20`), so nothing has validated `dealId` at that point. `em.findOne(CustomerDealCompanyLink, { deal: scoped.dealId })` therefore issues `… where deal_id = $1` with arbitrary caller-supplied text, and Postgres rejects the cast (`22P02 invalid input syntax for type uuid`).

That exception is not a `CrudHttpError`, not a `ZodError`, and `isTransientDbError` only matches connection-class SQLSTATEs and messages (`packages/shared/src/lib/db/pg-errors.ts:57-70`), so `handleError` falls through to the generic branch and returns **`500 Internal server error`**, logging `Unexpected CRUD error` (`packages/shared/src/lib/crud/factory.ts:597-628`). Before this PR the same request was caught by `commentCreateSchema.parse` and answered `400 Invalid input` with field details.

Concretely: `POST /api/customers/comments` with `{"dealId":"not-a-uuid","body":"x"}` goes from a clean 400 to a 500 plus an error-level log line — which also makes the PR body's *"payloads with neither entity nor resolvable deal still fail validation exactly as before"* inaccurate for malformed input.

Fix: validate before you query. Either parse `dealId` first (`z.string().uuid().safeParse(scoped.dealId)` and skip the derivation when it fails, letting the schema produce the 400), or restructure so the derivation happens after a schema that already accepts "entityId or dealId" — which is the shape I would prefer anyway (see B1.3).

**M2 — A behaviour change to a shared write path ships with no tests at all.** `route.ts:87-100`

Delete the entire 15-line block and every test in this repository still passes — unit and integration alike — because no test sends a comment payload without `entityId` (the callers are enumerated in B1). The green `test` job and the green affected-only `ephemeral-integration` run are therefore blind to this change by construction, which is the same reason the review cannot lean on them.

To the question in the PR body — *"Happy to add an integration test if you point me at the preferred location"* — there are two, and the unit one is directly beside the file you changed:

- **Unit**: `packages/core/src/modules/customers/api/comments/__tests__/entityId-filter.test.ts` is the established pattern for this route. It deliberately mirrors the route logic locally rather than importing the module graph (which needs generated files) — see its header comment. A sibling `entityId-derivation.test.ts` following the same style would cover: derivation from a company link; fallback to a person link when no company link exists; a `dealId` with neither link still failing validation; an explicit `entityId` never being overridden; a malformed `dealId` not reaching the query (M1); and the tenant/organization filter being present in the lookup (m1).
- **Integration**: `packages/core/src/modules/customers/__integration__/` — `TC-CRM-020.spec.ts` is the closest existing case (it creates a deal, posts a note with `{entityId, dealId}`, and asserts it renders on the deal timeline). A new case posting `{dealId, body}` *without* `entityId` and asserting the resulting `entity_id`, plus a negative case for a deal with no links, would pin the actual contract.

Per this repository's rules a behaviour change needs coverage in the same PR, so this is required regardless of how B1 is resolved.

### 🔹 Minor

**m1 — Both lookups are unscoped by tenant and organization.** `route.ts:93`, `route.ts:96`

`CustomerDealCompanyLink` / `CustomerDealPersonLink` carry no `organizationId` / `tenantId` columns of their own (`data/entities.ts:448-488`), so a scoped read has to go through the deal — which is exactly why the rest of this module never queries them bare. Every sibling helper takes the scope explicitly (`commands/shared.ts:62-67`, `:74-79`, `:139`), and both `packages/core/AGENTS.md` and the module guide state the rule as unconditional.

The write itself is not exploitable, and I want to be precise about that: the command re-validates, so a `dealId` from another tenant produces a derived entity id that `requireTimelineParentEntity` then fails to find in the caller's scope (`commands/comments.ts:84`, `commands/shared.ts:62-67`) and the request is rejected. What does change is the *shape* of the rejection — a foreign-tenant `dealId` used to come back as `400 Invalid input` from the schema and now comes back as `404 Customer not found` after two cross-tenant reads have executed. That is a small step in the wrong direction on a data-scoping boundary, and it costs one line to avoid: filter through the deal (`{ deal: { id: dealId, tenantId, organizationId, deletedAt: null } }`) using `ctx.auth.tenantId` and `ctx.selectedOrganizationId ?? ctx.auth.orgId`, or resolve the deal with the existing scoped helper first.

**m2 — The derived entity contradicts the UI's own choice, and is nondeterministic for multi-company deals.** `route.ts:93-97`

Two separate points, same two lines. First, the preference order is inverted relative to the product: the deal page builds its entity list as `[...data.people, ...data.companies]` and defaults to the entry flagged `isPrimary`, falling back to the first (`backend/customers/deals/[id]/page.tsx:132-152`) — people first, primary preferred. This PR prefers the **company** and ignores `isPrimary` entirely, even though `CustomerDealPersonLink.isPrimary` exists precisely for this (`data/entities.ts:457-458`). A deal-scoped note created through the derivation would therefore land on a different record than the same note created through the composer.

Second, `findOne` with no `orderBy` against a deal with several linked companies picks whichever row Postgres returns first, which is not stable across vacuum/plan changes. If the derivation stays, order it deterministically (primary first, then `createdAt`) and mirror the page's people-before-companies preference so the two paths agree.

**m3 — The request contract widens, but the OpenAPI export still advertises `entityId` as required, and the note silently lands on a record the caller never named.** `route.ts:100`, `route.ts:232-236`

After this change `POST /api/customers/comments` accepts a body without `entityId`, but `openApi.create.schema` is still `commentCreateSchema` (`:232-236`), where `entityId` is required (`data/validators.ts:241`). The generated API docs now describe a stricter contract than the route enforces — mild today, but this is the surface third-party module authors read, and `BACKWARD_COMPATIBILITY.md` treats API routes as a contract category (a widening is additive and not a break, so this is documentation drift, not a violation).

Related, and worth a product decision rather than a code change: because comments are listed by `entity_id` (`route.ts:76`), a note derived onto a company appears in that **company's** Notes tab, not only the deal's. A caller that sent only a `dealId` never asked for that. If the derivation stays, it should be documented on the endpoint so the placement is a stated behaviour rather than a surprise.

### 💅 Nit

**n1 — `populate` is unnecessary here.** `route.ts:93`, `route.ts:96`

Only `.id` is read from the loaded relation (`:97`), and a MikroORM `ManyToOne` reference already carries its primary key without hydration. Dropping `{ populate: ['company'] }` / `{ populate: ['person'] }` removes a join per create on a hot write path with no behaviour change.

**n2 — The cast discards the helper's typing and mutates the payload in place.** `route.ts:87`, `route.ts:98`

`as Record<string, unknown>` throws away whatever `withScopedPayload` returns and then `scoped.entityId = derived` mutates that object. Building the parse input immutably (`commentCreateSchema.parse({ ...scoped, entityId: scoped.entityId ?? derived })`) keeps the value flow obvious and avoids the cast.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff adds two entity imports and touches no export.
- [x] No function signature changed in a breaking way — `mapInput`'s signature is unchanged.
- [x] No required type field removed or narrowed — `CommentCreateInput` is untouched.
- [x] No HTTP route URL removed or renamed; no method changed — same path, same methods.
- [x] No field removed or retyped in an existing response shape — `create.response` is unchanged.
- [x] No event or message name renamed or removed — `customers.comment.*` unchanged.
- [x] No database table or column renamed or removed; no column type narrowed — no migration, no entity change.
- [ ] Where a contract had to change: the **request** contract is widened (`entityId` becomes conditionally optional) without the OpenAPI export following — additive, so not a break, but it is undocumented drift. See m3.

## 🧪 Test Coverage

The PR ships no tests. Its Testing section describes a manual check on a live instance (POST without `entityId` → 201, GET by `dealId`, delete), which is useful evidence but leaves nothing behind to protect the behaviour.

What exists around this route today: `api/comments/__tests__/entityId-filter.test.ts` covers the GET filter mapping for #1100; `components/detail/__tests__/notesAdapterLock.test.ts` covers the notes adapter's optimistic-lock headers; `packages/ui/src/backend/__tests__/NotesSection.test.tsx` covers the composer's add-action visibility and conflict-bar surfacing; and `TC-CRM-011/016/020/042` plus `TC-UNDO-001-comments` exercise comment create/list/undo end to end. None of them sends a payload without `entityId`, so none of them touches the new branch — the block is deletable with CI still fully green.

The gaps that must be closed, and where they belong (locations spelled out in M2): derivation from a company link; fallback to a person link; a deal with neither link still producing a 400; an explicit `entityId` never being overridden by the derivation; a malformed `dealId` producing a 400 rather than a 500 (M1); the lookup carrying the tenant/organization filter (m1); and a deterministic choice when a deal has several linked companies (m2). If B1 resolves toward "there is no UI defect", the more valuable test is the opposite one — a UI-level assertion that the deal composer always sends a concrete `entityId`, which is a few lines in the existing `NotesSection.test.tsx` and would have answered this question before the route was touched.

pkarw:
🏷️ label rationale

- **`changes-requested`** (pipeline, replacing the unlabeled/`review` state) — the review returned 1 blocker, 2 majors, 3 minors and 2 nits. Under this repository's verdict rule any blocker or un-waivered major requests changes; here all three do so independently.
- **`bug`** (category) — the change is presented and scoped as a bug fix on the comments write path. Kept as-is even though the review disputes that the underlying defect reproduces; the category describes the intent of the change, not its outcome.
- **`needs-qa`** (meta) — the automated-verification exemption does **not** apply. It requires an unchanged API surface *and* automated tests for the changed behaviour in the same PR, and this diff widens the `POST /api/customers/comments` request contract (`entityId` becomes conditionally optional) while shipping no tests. The no-UI-file and no-DB-change conditions are met, but all four are required.
- **`priority-low`** (urgency) — nothing here is release-blocking: the defect the PR targets could not be reproduced through any product path, and the endpoint works today for every in-repo caller. If a concrete repro is supplied this should be re-rated.
- **`risk-medium`** (blast radius) — the file is the single write path for CRM notes, shared by the detail pages' notes adapter, the `customers.manage_record_comment` AI tool and external API clients. The diff adds two unscoped database reads and moves caller-supplied input ahead of schema validation, so a defect here surfaces as 500s on a customer-facing endpoint. Not `risk-high`: the command layer still re-validates tenant and organization scope, so no cross-tenant write is possible.

Not applied: `qa-approved` / `qa-self-verified` (earned by manual QA, never by reading a diff), `qa` (means manual QA in progress and belongs to a QA reviewer), and `skip-qa` (see `needs-qa` above).

pkarw:
🤖 Handing this back to @mniarc — `changes-requested`.

Thanks for the clear write-up; the data-model reading in it is accurate and it made the trace fast. The review's one blocker is that I could not reproduce the 400 it is built on: the deal detail page passes `entityId={null}` **together with** `entityOptions`, which puts `NotesSection` into "user picks the customer" mode, so the composer posts the selected entity id rather than `null`. I rendered the component with exactly that prop shape and captured the payload both ways — with linked entities the adapter receives `entityId: "company-1"` and the request succeeds; with no linked entities the add-note button is disabled and no request is made at all. Full evidence, line references and the two follow-up questions are in the review body.

Next step is yours, and one of these three:

1. **Post the concrete repro** — the failing request body plus the deal's state (does it have linked people or companies?). If a UI state really does post `entityId: null`, that is a genuine bug and the fix most likely belongs in the page or in `NotesSection`.
2. **If the real complaint is the disabled "Add note" button** on a deal with no linked person or company, that is a legitimate UX defect worth fixing — but with an explanatory empty state, not a server-side guess (with no links there is nothing to derive and the request still 400s).
3. **If the goal is an API convenience** for clients that hold only a `dealId`, let's make it deliberate: express it in `commentCreateSchema` (`entityId` **or** `dealId`), update the OpenAPI export, scope the lookups by tenant/organization, make the entity choice deterministic and consistent with the page's own preference, and ship the tests.

Whichever way it goes, two items stand on their own: the malformed-`dealId` path now returns 500 where it used to return 400 (Major M1), and the test locations you asked about exist — `packages/core/src/modules/customers/api/comments/__tests__/` for unit coverage, right beside the file you changed, and `packages/core/src/modules/customers/__integration__/` (`TC-CRM-020.spec.ts` is the closest model) for integration (Major M2).

No commits were pushed to your branch — this run was review-only.

pkarw:

1 blocker, 2 majors, 3 minors, 2 nits. The blocker is that the defect the PR is built on could not be reproduced: the deal detail page renders `NotesSection` with `entityId={null}` **and** `entityOptions`, which is the component's "user picks the customer" mode, so the composer posts the selected entity id — verified by rendering the component with the page's exact prop shape (with linked entities the adapter receives a concrete `entityId` and the write succeeds; with none, the add-note button is disabled and no request is issued). The two majors stand independently of that: a malformed `dealId` now returns 500 instead of 400 because the derivation queries the database ahead of schema validation, and the behaviour change ships with no tests — the block is deletable with CI still fully green.

**Validation source: GitHub checks** — every check on head `2055ee03` has settled green (`test`, `lint`, `ds-lint`, `prepare`, `scope`, `audit-scope`, `ephemeral-integration (none)`), none failing and none pending, so no local fallback gate was run and no CI follow-up is owed. Head is `MERGEABLE` against `develop` with no conflicts.

autofix: skipped (not my PR — re-run with `--autofix` to fix it here). Handed back to @mniarc.

mniarc:
You're right, and I verified it on our production instance rather than just re-reading the code. Headless browser, a real deal with linked entities, request capture on the Notes composer: the POST carried the picker's entityId and returned 201. The add-note flow never sends null. My "repro" was a curl call without entityId, and the UI claim in the PR body was an inference from `entityId={null}` in the page source. I missed that `entityOptions` next to it switches NotesSection into picker mode. B1 stands.

The 400 was real, but it came from an API client, not the UI. We drive parts of our CRM through the MCP/REST surface, and a client that holds only a dealId currently has to read existing comments to fish out an entity_id before it can post a note. So I'm taking your option 3 and reworking this as a deliberate contract change:

- `commentCreateSchema` accepts entityId or dealId, both uuid-validated, so a malformed dealId 400s at the schema and never reaches the database (M1)
- derivation runs after validation, scoped by tenant/organization through the deal (m1)
- entity choice is deterministic and mirrors the deal page's own preference: primary person link, then first person, then first company (m2; CustomerDealCompanyLink has no isPrimary, so the person-side primary is the only primary there is)
- explicit entityId is never overridden
- a dealId with no linked person or company gets a clean 422 instead of a guess; a deal outside the caller's scope gets the existing 400
- while consolidating I found the AI deal-comment tool had its own copy of this derivation (no ordering, and a tenantId filter on a link entity that has no tenantId column), so the preference order now lives in one shared helper in commands/shared.ts used by both the route and the tool. Same deal, same landing entity, whichever entry point
- OpenAPI follows the schema (m3), populate dropped (n1), no cast-and-mutate (n2)
- unit tests beside the route per entityId-filter.test.ts, integration case modeled on TC-CRM-020 (M2)
- the UI-side test you suggested is in too: NotesSection.test.tsx now asserts the deal composer always sends the picker's concrete entityId, and that with no entity options no request is made at all
- per your m3, the derived placement (the note also appearing on the derived record's timeline) is stated in the endpoint's OpenAPI description instead of being a surprise

Force-pushed as a single commit and updated the PR description; the old "every deal comment fails" framing is gone. One thing you flagged in m3 as a product decision: whether a dealId-only caller should get derived placement at all. Worth saying that the platform already derives an entity from a dealId in the customers.manage_deal_comment AI tool, so this PR mostly makes an existing behavior consistent rather than introducing a new one. But if the decision goes the other way and entityId stays required, tell me and I'll close this, no hard feelings; we can carry it on our fork.

Thanks for the trace, the jsdom capture made it easy to see where I went wrong.

PR #5198, opened 2026-08-11
feat(sales): filter the documents list by several sales channels

maxidragon:
## Summary

`GET /api/sales/orders` and `GET /api/sales/quotes` accept a single `channelId` and filter with `$eq`. A merchant running several sales channels cannot ask *"orders from this channel **or** that one"* — the natural comparison as soon as more than one channel is in use. Today the only way to get it is a private override in the consuming app.

This adds multi-value channel filtering to the shared sales-documents list route, plus a way to select documents with no channel at all. Purely additive: `channelId` keeps its exact shape and meaning.

## Changes

- `packages/core/src/modules/sales/api/documents/factory.ts`
  - `listSchema` gains two optional params, following the CSV convention `tagIds` already uses in the same schema:

    | parameter | type | behaviour |
    |---|---|---|
    | `channelId` | uuid | **unchanged** — `$eq`. Wins over the two below when several are sent |
    | `channelIds` | comma-separated uuids | **new** — `$in` |
    | `channelIdsEmpty` | boolean token | **new** — documents with no channel (`$exists: false` → `WHERE channel_id IS NULL`) |

  - `buildFilters` resolves them singular-first, mirroring how `api/channels/route.ts` already resolves `id` before `ids`.
  - Ids are parsed with the shared `parseIdsParam` (`@open-mercato/shared/lib/crud/ids`) rather than a fourth copy of the local `parseIdList` regex — it dedupes and caps at `MAX_IDS_PER_REQUEST` (200).
  - An all-malformed `channelIds` narrows to **no** channel filter rather than an empty-set filter, so a typo returns the unfiltered page instead of silently returning zero rows.

- `packages/core/src/modules/sales/api/__tests__/documents.factory.test.ts` — 11 cases covering the above.

Both list routes are built from this factory, so orders and quotes get the behaviour together. The OpenAPI query schema derives from `listSchema` (`buildDocumentOpenApi` passes `querySchema: listSchema`), so both params are documented automatically — no separate schema edit.

No database change, no new dependency, no UI change.

## Specification

**Does a spec exist for this feature/module?**
- [ ] Yes
- [ ] No (created a new spec)
- [x] N/A (minor change, no spec needed)

**Spec file path:**
N/A — this is a purely additive API change. Per `BACKWARD_COMPATIBILITY.md` §7 and the quick-reference table, adding optional request fields to an existing route is `ADDITIVE`, so the deprecation protocol (and its spec requirement) does not engage. Happy to write one up if maintainers would rather have it on record.

[...]

PR #5250, opened 2026-08-13
feat(data_sync): let adapters opt out of the shared cursor row per entity type

maxidragon:
## Summary

`data_sync` persists a run's cursor in two places on every batch commit:

1. `sync_runs.cursor` — the run's own row, which a redelivered queue job resumes from.
2. `sync_cursors` — **one row per `(integration_id, entity_type, direction, organization_id, tenant_id)`**, overwritten unconditionally by whichever run commits last.

The second write is right for an incremental feed and wrong for a whole-table backfill, and one adapter commonly serves both under a single provider key.

- **A feed cursor is a position in a change log.** Losing it means rewinding to the start of the queue and re-draining everything, so it must outlive the run that advanced it.
- **A backfill cursor is one run's scan state** over a table the adapter re-walks idempotently. Losing it costs a re-walk, which is what the walk does anyway.

Sharing the row between them is a silent correctness bug, and it does not need concurrency to fire — two *sequential* runs are enough:

| | run A (full walk) | run B (scoped to the last week) |
|---|---|---|
| cursor mid-run | `{"lastId":0,"beforeId":900000,"topId":1380000}` | `{"lastId":0,"since":"2026-08-05","beforeId":1379000}` |

Both write the same row; last writer wins. If B commits last, the row reads *"position 1379000, window = last week"*. The next resume starts there, walks ~1,000 records, and **finishes `completed`**. The 900,000 records A never reached are skipped, and nothing reports a problem.

The fix is an optional predicate on the adapter contract letting an adapter opt a given entity type out of the shared row. This is a correctness fix rather than a configuration knob.

## Changes

- **`lib/adapter.ts`** — new optional `persistsSharedCursor?(entityType: string): boolean`, defaulting to `true`. Per **entity type** rather than a per-adapter boolean: an adapter's `supportedEntities` commonly mixes feed and backfill types, and an adapter-level flag cannot express "persist for the feed, not for the backfill".
- **`lib/sync-run-service.ts`** — new exported `CursorCommitOptions = { persistSharedCursor?: boolean }`, taken as a trailing optional argument by `updateCursor` and `commitBatchProgress`. When `false`, the run row still advances and the service skips the `sync_cursors` **read as well as the write** — one query less per batch on every long import, against a row nothing will read. Adds `resolveResumeCursor(...)`, which returns the cursor of the most recent run for the scope, or `null` when that run reached `completed`.
[...]

PR #5351, opened 2026-08-17
feat(sales): render channel names and filter by several channels

maxidragon:
**Stacked on [#5350](https://github.com/open-mercato/open-mercato/pull/5350)** (*resolve sales channel names on the documents list*) — that branch's commit is included here, so review only the UI commit and merge #5350 first.

## Summary

Two things on the sales documents list fail quietly.

The Channel column resolves names by fetching **the first 50 channels** and matching them client-side. A merchant with more than 50 channels sees raw identifiers for everything past the fiftieth, with no error and no sign anything is missing.

The Channel filter accepts one channel. "Orders from the web shop **and** the marketplace" — the normal comparison as soon as a merchant runs more than one storefront — cannot be expressed. [#5198](https://github.com/open-mercato/open-mercato/pull/5198) taught the API to answer exactly that query, so right now there is a capability on `develop` that no interface can reach.

The column now renders the `channelName` the list response carries (from #5350), so it no longer depends on the channel being inside a fetched page of options. The identifier remains only as the fallback for a channel the API could not resolve. The filter becomes a multi-select emitting `channelIds`.

## Changes

- `packages/core/src/modules/sales/components/documents/SalesDocumentsTable.tsx` — column renders `channelName`; filter becomes `type: 'tags'`; query building emits `channelIds` / `channelIdsEmpty`.
- `packages/core/src/modules/sales/components/documents/__tests__/SalesDocumentsTable.channelFilter.test.tsx` — 7 cases.
- `packages/core/src/modules/sales/i18n/{en,pl,es,de,ko}.json` — two new keys.

### The `(No channel)` entry

The filter's option list carries a synthetic `(No channel)` entry that maps to `channelIdsEmpty` rather than into the id list. Picking it alongside real channels sends both parameters, which #5198 `$OR`s together — so "these two channels, or unassigned" behaves the way it reads, rather than one selection silently discarding the other.

The sentinel is deliberately **not a uuid** (`__unassigned__`). It is stripped explicitly when the query is built; the non-uuid shape is a backstop, not the mechanism. If it ever did leak into `channelIds`, the route drops non-uuid entries, so the filter narrows instead of rejecting the page — a safe failure rather than a 400 in front of the operator.

### Saved perspectives survive the type change

`DataTable` persists `filterValues` into the perspective snapshot, and a snapshot saved while this filter was a single-value select holds `channelId` as a bare string. Normalization happens at the **hydration boundary** (`handleFiltersApply`), not only where the query string is built, so the stored state is always `string[]` and the control, the active-filter chip and the query builder all read one shape.

That distinction is the whole finding: normalizing only at query-build time produces a correct request while the `tags` control — which renders from `Array.isArray(values[id]) ? … : []` — shows nothing. The list comes back filtered, the field looks empty, and the next channel the user picks replaces the restored one instead of joining it. This is the one real backward-compatibility trap in the change, and the test now asserts the stored shape rather than only the resulting URL.

### Selected channels outside the first options page

A channel past the fiftieth, or one restored from a perspective before options have loaded, would render as a bare uuid in the filter chip and the selected tag — the same defect this change removes from the column. Selected ids are now fetched by `ids=` and merged into the options, per `.ai/lessons/async-edit-selects-must-be-hydrated-as-value-plus.md`.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**
N/A — this consumes two API surfaces that already exist on `develop` (#5198's query params, #5350's response fields) and changes no contract of its own. Happy to write one up if maintainers would rather have it on record.

## Testing

Branch merged with current `develop` and validated on that base:

- `yarn build:packages` ✅
- `yarn generate` ✅ — no generated-file drift
- `yarn typecheck` ✅
- `yarn lint` ✅ — pre-existing warnings only, no new ones
- `yarn i18n:check-sync` ✅ — 5 locales in sync
- `yarn workspace @open-mercato/core test -- sales` ✅ — 598/598

Ten component tests, each mapped to what it catches:

| test | catches |
|---|---|
| renders the server-resolved name | regressing to the client-side join |
| falls back to the id when the name is null | rendering `undefined` in the column |
| filter is `tags`, unassigned entry present | shipping it still single-value |
| several channels → `channelIds`, no empty flag | sending `channelIdsEmpty` unprompted |
| channels + unassigned → **both** parameters | one selection silently discarding the other |
| unassigned alone → only `channelIdsEmpty` | sending an empty `channelIds=` |
| legacy string snapshot → `channelIds` | saved perspectives silently clearing |
| legacy snapshot **stored** as `[uuid]` | the control rendering empty while the list is filtered, and the next pick replacing the restored channel |
| selected channel outside page 1 → `ids=` fetch | a uuid in the chip and the selected tag |
| `(No channel)` absent while searching, present while browsing | the sentinel offered as a match for every search term |

**Mutation-checked rather than assumed green:** reverting the query builder to the previous single-parameter version fails four of them, and reverting the hydration-boundary normalization fails two more — including the stored-shape assertion, which is the one that would have caught the perspective defect.

Merging `develop` surfaced one thing worth naming: a `ko` locale was added while this branch was open, so the two new keys were missing from it. `yarn i18n:check-sync` caught it and it is fixed in this branch — wording follows the file's existing channel vocabulary.

**Not covered: there is no Playwright spec**, and this is the change that most deserves one — a real browser, a real filter click, real rows. I have not added it because I could not execute it here: the integration suite needs a standard-seeded environment, and the local database available to me is not one. It is worth adding on a properly seeded environment, covering two channels and three orders — filter by both channels → two rows, unassigned → one row, and the column showing names on first paint.

## Checklist

- [ ] I have read and accept the Open Mercato Contributor License Agreement (see `docs/cla.md`). — *left for a human to tick; an agent should not accept a legal agreement on someone's behalf.*
- [x] I updated documentation, locales, or generators if the change requires it. — two keys added to all five locales.
- [ ] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required). — not included; see Testing for why.
- [ ] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — not applicable, see Specification.
- [ ] Risk set: this PR carries exactly one `risk-*` label.
- [ ] QA routing set.

The last three are unticked because this account has read-only permissions here and cannot apply labels. Intended values are in a comment below.

### Design System Compliance

- [x] No hardcoded status colors — none added.
- [x] No arbitrary text sizes — none added.
- [x] Empty state handled — unchanged; the table's existing empty state applies.
- [x] Loading state handled — unchanged.
- [x] `aria-label` on icon-only buttons — none added.
- [x] Uses existing DS components — a `FilterDef` type change and a cell renderer; no custom replacements.

## Linked issues

None found. I searched open and closed issues for an existing report of the Channel column degrading past the options page, or of the filter being single-value, and found none.


maxidragon:
This account has read-only permissions here, so I can't apply labels or request reviewers. Intended metadata for whoever picks this up:

| label | why |
|---|---|
| `feature` | makes multi-channel filtering reachable from the UI for the first time |
| `review` | ready, not a draft |
| `priority-medium` | both defects are silent — raw uuids past 50 channels, and a filter that can't express a two-channel comparison — but neither loses data |
| `risk-low` | one component and two i18n keys; no API, database or contract change |
| `needs-qa` | see below |

**On QA routing:** unlike the two PRs it builds on, this one is squarely UI. It changes a rendered column and a filter control on `/backend/sales/orders`, so the automated-verification exemption does not apply and a human clicking through is worth having — particularly the `(No channel)` entry combined with real channels, which is the interaction the component tests assert but nobody has yet seen in a browser.

**Reviewer:** someone who owns the backend DataTable / perspectives surface would be the right fit. The judgement call is the synthetic `(No channel)` option and the legacy-snapshot normalisation — a saved perspective from before this change holds `channelId` as a string, and that shape has to keep working.

@jtomaszewski could you apply the `partner-request` label? This is an FSH partner contribution, and this account is neither a collaborator here nor listed under `8lines` in `.github/certified-partners.yml`, so I can't set it myself or use the `/label` workflow.

One note on the CLA checkbox: left unticked deliberately — an agent shouldn't accept a legal agreement on someone's behalf.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T13:10:14Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review

## 🎯 Summary

This PR makes the sales documents Channel column render the server-resolved `channelName` (from the stacked #5350) instead of joining against a client-side page of channel options, and turns the single-value Channel filter into a multi-select that emits `channelIds` / `channelIdsEmpty` — the query shape #5198 already taught the API to answer. Both problems it names are real, the diagnosis is accurate, and the chosen mechanism is the right one: the column now depends on the list response rather than on a channel happening to be inside a fetched options page, and the `(No channel)` sentinel maps to a separate boolean parameter rather than being smuggled into the id list, which is what lets "these two channels, or unassigned" work at all. The seven component tests are well chosen and genuinely fail when the query builder is reverted.

The backward-compatibility case the PR calls out — a perspective snapshot saved while this filter was a single-value select, holding `channelId` as a bare string — is only half handled. `normalizeChannelSelection` runs at query-build time, so the request is correct, but the stored `filterValues.channelId` keeps the legacy string shape, and the `tags` control in `FilterOverlay` renders nothing for a non-array value. The user's list is filtered while the Channel filter looks empty, and the next channel they add replaces the restored one instead of joining it. That is the one finding that has to be fixed before merge; the rest is a consistency gap the PR inherits rather than introduces, plus polish.

**Merge order:** #5350 is still open, and this branch carries its commit (`c97d017`, `api/documents/factory.ts` + `documents.channelNames.test.ts`). Merging #5351 first would land #5350's API change with it and leave #5350 as an empty PR — merge #5350 first as the description asks, or close it as superseded once this lands. The `attachChannelNames` hook itself is sound: one batched query per list page, tenant/organization scoped, no `deletedAt` filter with a documented rationale, and the two new response fields are additive and reflected in the OpenAPI schema.

**Verdict: CHANGES REQUESTED** — one major finding (the legacy perspective snapshot silently loses the restored channel as soon as the filter is touched). Note that this repository's review extension also requests changes on Medium/minor findings, so the two minors below would carry the same verdict on their own.

## 🧪 Validation Gate

Runner: **local** (no compose `app` container is running — only `mercato-verdaccio`).

Validation source: **local fallback**. This is not a preference: no CI has run on this head at all. `gh pr checks 5351` reports only `license/cla` (pass) and `[code]smith` (skipping), and `check-runs` for `8be9e08` contains nothing else — `.github/workflows/ci.yml` triggers on `pull_request` into `develop`, so the run for this fork PR is waiting on a maintainer to approve workflow runs for an outside contributor. `mergeStateStatus` is `BLOCKED` with `mergeable: MERGEABLE`, i.e. no conflicts — the block is the missing review/checks, not the branch. **The gate below is this run's evidence only and does not substitute for branch protection**; the required checks still have to run and pass before merge.

| Command | Result |
|---|---|
| `yarn generate` | ✅ pass — no generated-file drift |
| `yarn build:packages` (rebuild after generate) | ✅ pass |
| `yarn i18n:check-sync` | ✅ pass — 5 locales in sync, both new keys present in en/pl/es/de/ko |
| `yarn i18n:check-usage` | ✅ pass (advisory unused-key report only) |
| `yarn typecheck` | ✅ pass |
| `yarn test` | ✅ pass — full workspace suite; the three suites this PR touches pass on their own too (44 tests) |
| `yarn build:app` | ✅ pass |

## 🔴 Findings

### 🟠 Major

**1. A legacy single-value perspective snapshot is discarded the moment the user touches the filter** — `packages/core/src/modules/sales/components/documents/SalesDocumentsTable.tsx:427`

`normalizeChannelSelection` is applied only where the query string is built, so `filterValues.channelId` keeps whatever shape the perspective restored. `DataTable.applyPerspectiveSettings` passes the persisted filters through `onFiltersApply` unchanged (`packages/ui/src/backend/DataTable.tsx:1951-1957`), and `handleFiltersApply` (`SalesDocumentsTable.tsx:565`) stores them verbatim — so for a snapshot saved before this PR, `filterValues.channelId` is the bare string `'<uuid>'`.

Two things follow from that:

- `FilterOverlay` renders the `tags` control from `Array.isArray(values[f.id]) ? values[f.id] : []` (`packages/ui/src/backend/FilterOverlay.tsx:362`), so opening **Filters** shows the Channel field with nothing selected even though the list below it is filtered.
- `TagsInput.addValue` builds the next value from its `value` prop (`packages/ui/src/backend/inputs/TagsInput.tsx:149-151`), which is that same empty array. Picking a second channel therefore emits `['<new-id>']`, not `['<restored-id>', '<new-id>']`.

Failure scenario: a merchant has a saved view "Web shop orders" created before this change (`filters: { channelId: '<web-shop-uuid>' }`). After the upgrade they open it — the rows are correct, the chip reads `Channel: Web shop`. They then open **Filters** to add the marketplace, which is precisely the comparison this PR exists to enable. The Channel field appears empty; they pick Marketplace and Apply. The request becomes `channelIds=<marketplace>` and the web shop is silently gone from their own saved view. The existing test (`SalesDocumentsTable.channelFilter.test.tsx`, "restores a legacy single-value perspective snapshot") asserts only the resulting URL, which is why this path looks covered.

Verified against the head revision with a throwaway `FilterOverlay` test (deleted afterwards): with `initialValues: { channelId: [uuid] }` the overlay renders one removable tag; with `initialValues: { channelId: uuid }` it renders zero.

Fix: normalize at the hydration boundary rather than at query-build time, so the stored state is always `string[]` and every consumer — control, chips, query builder — sees one shape:

```ts
const handleFiltersApply = React.useCallback((values: FilterValues) => {
  setFilterValues(
    values.channelId === undefined
      ? values
      : { ...values, channelId: normalizeChannelSelection(values.channelId) }
  )
  setPage(1)
}, [])
```

The query builder can then keep `normalizeChannelSelection` as a cheap guard. Please extend the legacy-snapshot test to assert the stored/applied filter shape (or the tags control's rendered selection) rather than only the URL — that assertion is what would have caught this.

### 🟡 Minor

**2. The filter still shows raw uuids for channels outside the first fetched page — the same defect the column change fixes** — `SalesDocumentsTable.tsx:354`

`formatValue` resolves a selected channel's label from `channelOptions`, which is filled only from `/api/sales/channels?page=1&pageSize=50` plus whatever later searches merge in (`SalesDocumentsTable.tsx:197-199`, `288-295`). A channel past the fiftieth, or one restored from a perspective before the options have loaded, renders as a bare uuid — in the active filter chip (`packages/ui/src/backend/FilterBar.tsx:128-131`) and, after finding 1 is fixed, in the overlay's selected tag as well. The PR's own framing is that a merchant with more than 50 channels "sees raw identifiers for everything past the fiftieth, with no error and no sign anything is missing"; that is now true of the filter rather than the column.

This repository has a documented rule for exactly this shape — `.ai/lessons/async-edit-selects-must-be-hydrated-as-value-plus.md`: "If the saved option may be outside the first page, fetch it by id and seed/prepend it." The endpoint already supports it: `/api/sales/channels?ids=<uuid,uuid>` (`packages/core/src/modules/sales/api/channels/route.ts:32`, `78-79`). Seeding `channelOptions` for the currently selected ids on hydration would close both the chip and the overlay label.

In fairness this matches the pre-existing customer filter two definitions below (`SalesDocumentsTable.tsx:401`), so it is an inherited gap rather than a regression introduced here — but it is the same defect this PR sets out to remove, and the fix is small enough to belong in this change.

**3. No integration coverage for a change that is squarely UI** — `packages/core/src/modules/sales/components/documents/`

Root `AGENTS.md` and `.ai/qa/AGENTS.md` require integration coverage for new/changed key UI paths to ship in the same change. The PR documents honestly why it could not run Playwright locally and even sketches the right scenario (two channels, three orders: both channels → two rows, unassigned → one row, names on first paint), which is appreciated. Two acceptable resolutions: add that spec under `.ai/qa/tests/` so it runs in CI, or take the manual route — this PR is labelled `needs-qa`, and with the QA gate on it cannot merge until a QA reviewer exercises the filter (especially `(No channel)` combined with real channels) and adds `qa-approved`. Please say which route you intend so the gate is not left ambiguous.

### 🔵 Nits

**4. `(No channel)` is offered as a match for every search term** — `SalesDocumentsTable.tsx:311-317`

`loadChannelFilterOptions` prepends the sentinel to whatever the query returned, so typing `web` still suggests `(No channel)`. Prepending only for an empty query would read better.

**5. Selecting more than 200 channels truncates silently** — `SalesDocumentsTable.tsx:429`

`parseIdsParam` caps at `MAX_IDS_PER_REQUEST = 200` and slices the excess (`packages/shared/src/lib/crud/ids.ts:3`, `45-49`), so a selection past 200 narrows without telling anyone. Implausible through a search-driven tags input, and the API's own contract documents the cap — noted only for completeness.

## 💥 Breaking Changes

- [x] **No public contract surface removed or renamed.** The change is confined to one client component plus two additive i18n keys. `channelId` (singular) stays in the list schema and still takes precedence server-side; nothing now sends it, so no caller changes behavior.
- [x] **API response fields are additive.** `channelName` / `channelCode` (inherited from #5350) are added as `nullable().optional()` in the documents OpenAPI schema — additive-only, per `BACKWARD_COMPATIBILITY.md`.
- [x] **No database, migration, event-id, widget-spot-id, DI-key, ACL-feature or CLI change.**
- [ ] **Persisted client state changes shape without a complete bridge.** `filterValues.channelId` moves from `string` to `string[]` in perspective snapshots. A bridge was written and is the right idea — it is just incomplete on the control side; see finding 1. Snapshots saved *after* this change hold an array and are unaffected.

## 🧪 Test Coverage

The unit coverage is above the bar for a change of this size and the mutation check (reverting the query builder fails four of the seven) is exactly the right way to demonstrate the tests bite. `documents.channelNames.test.ts` and the `documents.factory.test.ts` additions belong to #5350 and are reviewed there; all three suites pass here (44 tests).

Two gaps:

- The legacy-snapshot test asserts the request URL only, so it passes while the control silently loses the restored selection (finding 1). Assert the applied filter shape or the rendered selection.
- No integration test (finding 3).

One note on the component test harness, not a request: it mocks `DataTable` and asserts the props handed to it, which makes it a wiring test rather than a rendering test. That is a reasonable trade for this component's size, but it is also the structural reason the `FilterOverlay` half of the legacy-snapshot path went unnoticed — the real filter control never runs in these tests.

pkarw:

- 🚀 `merge-queue` (was `changes-requested` from my round-1 pass) — my re-review of `57ae907a` approves: all three actionable findings I raised are closed, and I verified the major one at the mechanism (`DataTable.tsx:1951-1957` restores a saved view through `onFiltersApply`, which is exactly where normalization now lives) rather than at the description. `needs-qa` stays on, so the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — kept. This is squarely a UI change: a rendered column and a filter control on `/backend/sales/orders`, so the automated-verification exemption does not apply and a human should exercise it before it lands.
- ✨ `feature` — unchanged; this makes multi-channel filtering reachable from the sales documents UI for the first time, consuming query parameters the API already supports.
- 🔹 `priority-medium` — unchanged; both defects it addresses are silent rather than destructive, with no data loss in either case.
- 🟢 `risk-low` (was `risk-medium`) — downgraded because the reason I rated it medium no longer holds. I set `risk-medium` in round 1 because the branch as pushed carried #5350's commit, putting the sales documents list API in the blast radius alongside the component. #5350 merged on 2026-08-18 (`3fc50f34`), so the diff against the merge base is now exactly this PR's seven files: one client component plus two additive i18n keys, with no API, database or contract change. That is the `risk-low` definition, and it matches what @maxidragon originally proposed.

`partner-request` is left in place untouched.

pkarw:
Thanks @maxidragon — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

Two things worth knowing before you start:

- **No CI has run on this head.** `.github/workflows/ci.yml` does trigger on pull requests into `develop`, but the run for a fork PR from an outside contributor needs a maintainer to approve it, so the only checks reported are `license/cla` and a skipped `[code]smith`. I ran the full configured gate locally instead — every command passes on the merged base — but that is my evidence, not branch protection's, so the required checks still have to run before this can merge. A maintainer approving the workflow run would unblock that.
- **Merge order:** #5350 is still open and its commit is part of this branch. Merging this PR first would land #5350's API change with it and leave #5350 empty, so #5350 should merge first as your description asks (or be closed as superseded once this lands).

pkarw:

One major finding (a legacy single-value perspective snapshot is dropped by the `tags` filter control the moment it is touched — verified against this head with a throwaway `FilterOverlay` test), two minors (raw uuids still shown for channels outside the first fetched options page; no integration coverage for a UI-only change) and two nits. Every command in the configured validation gate passes locally on the merged base; no CI has run on this head, so those local results are this run's evidence only and the required checks still gate the merge.

maxidragon:
Thanks @pkarw — addressed in `c6d224a72`, which also carries #5350's review fixes so they don't get re-litigated here.

## Major 1 — legacy perspective snapshot

You're right, and the diagnosis is exact: I normalized where the query string is built, so the *request* was correct and the *state* kept the legacy shape. `FilterOverlay` rendering `tags` from `Array.isArray(values[f.id]) ? … : []` is what makes that visible — filtered list, empty-looking field — and `TagsInput.addValue` building from the same empty array is what silently drops the restored channel on the next pick.

Normalization moved to `handleFiltersApply`, so the stored value is always `string[]` and the control, the chip and the query builder read one shape. Your suggested shape is what I used.

The test gap is the part worth calling out: my legacy-snapshot test asserted the URL only, which is precisely why this looked covered. There is now a second assertion on the **stored** shape, and it discriminates — reverting the hydration fix fails two tests (the stored-shape one and the seeding one), where before it failed none.

## Minor 2 — uuids in the chip and the selected tag

Fair, and the sharper version of it is that this PR's own framing condemns it: I argued the column shouldn't degrade past the options page, then left the filter doing exactly that. Selected ids are now fetched via `ids=` and merged into the options, per the lesson file you cited. Covered by a test asserting the `ids=` request fires for a channel outside page 1.

I've left the customer filter alone — same defect, but it predates this change and fixing it here would widen the diff past what this PR is about. Worth its own change.

## Minor 3 — integration coverage

Taking the manual-QA route, explicitly, so the gate isn't ambiguous: this PR carries `needs-qa` and should not merge until a QA reviewer exercises the filter and adds `qa-approved`. The interaction most worth a human is `(No channel)` combined with real channels — the component tests assert both params are sent, but nobody has watched the resulting rows.

I'd rather not ship a Playwright spec I cannot execute: the integration suite needs a standard-seeded environment and the database available to me is not one, so I'd be writing assertions I could only guess at. If you'd prefer the spec in-tree anyway I'll write it, flagged as unverified.

## Nit 4 — `(No channel)` matching every search

Fixed; the sentinel is prepended only when there's no query. Test covers both directions.

## Nit 5 — the 200 cap

Left as is, agreed it's noted for completeness rather than actionable. A tags input driven by search won't realistically reach 200 selections, and the API documents the cap.

## On the merge order note

Agreed, and thanks for flagging what merging this first would do to #5350. #5350 should land first; this branch carries its commit and I've kept it merged forward.

Full gate on the merged base: `build:packages`, `generate` (no drift), `typecheck`, `lint` (pre-existing warnings only), `i18n:check-sync` (5 locales), `yarn workspace @open-mercato/core test -- sales` at 601/601.

One note on your closing observation about the harness — that mocking `DataTable` makes these wiring tests rather than rendering tests, and that this is the structural reason the `FilterOverlay` half went unnoticed. That's the most useful thing in the review, because it explains *why* the gap existed rather than just that it did. The stored-shape assertion is a partial answer (it catches state the real control would misread), but it isn't the same as rendering the control. Worth revisiting the harness for this component separately.

patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-18T19:22:41Z. Other auto-skills will skip this PR until the lock is released.

patzick (approved):
# 🔍 Code Review — re-review of `c6d224a72` (now `8656c517`)

**Verdict: approve.** The three actionable findings from the previous round are genuinely fixed, and I verified the mechanism of the major one rather than taking the description on trust. Two non-blocking items below. Merge is still gated — but on things outside this diff, spelled out at the end.

**Runner: CI only** (no local execution this round, per the task brief).

## First: this PR had no test signal at all, and now it does

Worth stating plainly because the previous round's evidence was weaker than it looked, and the reason was invisible from the checks list.

On the old head, `audit` **failed** on `deepmerge-ts <8.0.0` ([GHSA-ggr8-5vv4-36mx](https://github.com/advisories/GHSA-ggr8-5vv4-36mx)), reachable via `html-to-text@10.0.0`. Nothing to do with this PR — it touches no manifest or lockfile. What tripped it is that `audit-scope` detects dependency changes with a **two-dot** `git diff origin/${base_ref} --name-only`, so a branch that has fallen behind sees develop's *own* lockfile movement as a change and runs the audit against its stale graph. develop already carries `deepmerge-ts@8.0.1`.

That matters because `test`, `ephemeral-integration` and `merge-coverage` all declare `needs: audit`. Their `skipping` status was not "not applicable" — it was **no signal**.

I updated the branch from develop and approved the fork workflow runs. `audit` now correctly skips (`test`'s `if:` explicitly allows `audit.result == 'skipped'`), and the gate ran for real:

| check | result |
|---|---|
| `prepare` | ✅ 3m38s |
| `lint` / `ds-lint` / `scope` / `audit-scope` | ✅ |
| typecheck, i18n sync, i18n usage (steps in `test`) | ✅ |
| **`ephemeral-integration`** | ✅ **15m38s** |
| `test` | ❌ — see below, not this PR |

## The `test` failure is inherited from a red develop

`develop` is currently broken at HEAD. Run [32157461842](https://github.com/open-mercato/open-mercato/actions/runs/32157461842) (`00c90fec`) fails the same 7 tests this PR's run does, in a suite this PR never touches:

```
FAIL src/modules/data_sync/lib/__tests__/sync-engine-heartbeat.test.ts
TypeError: (0 , adapter_registry_1.resolveProviderKey) is not a function
  at Object.runImport (src/modules/data_sync/lib/sync-engine.ts:479:45)
```

Cause: `3ae527e38` ("single source of truth for adapter resolution", in the #5250 stack) moved `resolveProviderKey` out of `sync-engine.ts` into `adapter-registry.ts`, so line 479 now imports it — but `sync-engine-heartbeat.test.ts:13` mocks `../adapter-registry` with a factory that only supplies `getDataSyncAdapter`. develop was green at `3adb0d0d` (09:05) and red by `00c90fec` (15:56). A one-line fix, but it belongs on develop, not here.

The counts confirm this PR's own tests are fine:

| | suites | tests |
|---|---|---|
| develop HEAD | 1 failed, 1289 passed, **1290** | 7 failed, 10132 passed, **10139** |
| this PR | 1 failed, 1291 passed, **1292** | 7 failed, 10152 passed, **10159** |

Same one suite, same 7 tests. The deltas (+2 suites, +20 tests) are exactly this PR's additions, **all passing**.

## Major 1 — legacy perspective snapshot: fixed, and verified

I checked the claim at the mechanism rather than the description, because "normalization moved to `handleFiltersApply`" is only a fix if perspective restore actually goes through that callback. It does — `packages/ui/src/backend/DataTable.tsx:1805-1811` restores a saved view by calling `onFiltersApply(legacy)`, which is exactly `handleFiltersApply`. So a legacy bare-string `channelId` is normalized at the moment it re-enters state, and the control, the chip and the query builder all read `string[]`.

Your read on why the old test looked like coverage is the right one, and the new assertion is the part that discriminates: `expect(lastProps().filterValues.channelId).toEqual([CHANNEL_A])` fails if the hydration fix is reverted, where a URL-only assertion does not.

## Minor 2 — uuids past the first options page: fixed

`seedChannelOptionsForIds` fetches selected ids via `ids=` and merges them in; `/api/sales/channels` does support `ids` (`route.ts:32,78-79`). Agreed on leaving the customer filter alone — same defect, predates this change, own PR.

## Nit 4 — sentinel matching every search: fixed

Sentinel prepended only when there is no query, asserted in both directions. Also confirmed no duplicate-sentinel hazard: `loadChannelOptions` merges the raw options into state *before* the sentinel is prepended, so `channelOptions` never contains it.

## Minor 3 — integration coverage: accepted

Taking the `needs-qa` route explicitly is the right call, and I'd rather have that than a Playwright spec written blind against assertions you can't execute. `needs-qa` stays on.

---

## New, non-blocking

**Low — `seedChannelOptionsForIds` can exceed the route's `pageSize` cap.**
`packages/core/src/modules/sales/components/documents/SalesDocumentsTable.tsx` sets `pageSize: String(missing.length)`, but `packages/core/src/modules/sales/api/channels/route.ts:29` declares `pageSize: z.coerce.number().min(1).max(100)`. The documents API caps `channelIds` at **200**, so >100 selected channels is expressible in this filter — at which point the seeding request 400s, `fetchChannelOptionsBy` returns `[]` on `!call.ok`, and precisely those chips fall back to the raw uuids this PR set out to remove. `Math.min(missing.length, 100)` (or chunking) closes it. Unlikely in practice, and it fails soft.

**Nit — the seeding request repeats on every apply.**
`missing` only filters out the sentinel; it never checks `channelOptions` for ids already loaded, so each filter apply re-fetches labels it already has. Diffing against the current options would make the name match the behaviour.

---

## Merge gating — none of it is this diff

1. **develop must go green** (the `data_sync` mock above), otherwise `test` stays red here and on every other PR.
2. **#5350 lands first** — it's in `merge-queue`, and this branch carries its commit, so the order you flagged is already resolving itself.
3. **`needs-qa` → `qa-approved`** by a QA reviewer. Per the hard gate, this must not merge on `needs-qa` alone. The interaction you nominated — `(No channel)` together with real channels — is the right thing to put a human in front of.

On your closing point about the harness: agreed that the stored-shape assertion is a partial answer rather than the same thing as rendering the real control, and that the mocked-`DataTable` structure is why the `FilterOverlay` half went unnoticed. Worth its own change; it isn't this PR's job to fix.

patzick:

- ✅ `changes-requested` → `review` — the re-review of `c6d224a72` found all three actionable items from the previous round fixed. The major one (legacy perspective snapshot) was verified at the mechanism, not just the description: `DataTable.tsx:1805-1811` confirms perspective restore calls `onFiltersApply`, which is where normalization now lives. Remaining findings are one low and one nit, neither blocking.
- 🧪 `needs-qa` — kept. This is squarely a UI change (a rendered column and a filter control on `/backend/sales/orders`), so the automated-verification exemption does not apply. It must not merge until a QA reviewer adds `qa-approved`.
- ✨ `feature` — unchanged; makes multi-channel filtering reachable from the sales documents UI for the first time.
- 🔹 `priority-medium` — unchanged; both defects addressed are silent rather than destructive, with no data loss.
- 🟡 `risk-medium` — unchanged. Worth noting the automated evidence is materially stronger this round: the previous head had **no** test signal (`audit` failed, and `test`/`ephemeral-integration` both declare `needs: audit`), whereas this head ran `ephemeral-integration` green in 15m38s with 10152 unit tests passing.
- 🚫 `in-progress` — released, lock cleared.

`review` here means the code is ready, not that it can merge. It is still blocked on a red `develop` (`sync-engine-heartbeat.test.ts`, unrelated to this diff), on #5350 landing first, and on QA approval.

patzick:

All three actionable findings from the previous round are fixed; the major one was verified at the mechanism rather than the description. Two non-blocking items remain (a `pageSize` cap that a >100-channel selection would exceed, and a redundant seeding request per apply).

Two things this run changed beyond the review itself:
- **Restored the test signal.** The previous head's `audit` failure (`deepmerge-ts`, unrelated to this PR — a stale branch tripping `audit-scope`'s two-dot diff) was silently skipping `test` and `ephemeral-integration`. Branch updated from develop and the fork workflow runs approved; `ephemeral-integration` now passes in 15m38s.
- **Surfaced a red develop.** The remaining `test` failure is inherited, not this PR's: `sync-engine-heartbeat.test.ts` fails identically on develop HEAD because `3ae527e38` moved `resolveProviderKey` into `adapter-registry.ts` without updating that suite's mock. Needs a fix on develop.

autofix: skipped (fork PR, and the only outstanding failure is not in this diff).

pkarw:
@maxidragon please check failing CI step - unit tests

maxidragon:
@pkarw checked — the failing `test` job wasn't caused by this PR. It only touches `packages/core/src/modules/sales/**`, nothing under `data_sync`.

All 7 failures were in `packages/core/src/modules/data_sync/lib/__tests__/sync-engine-heartbeat.test.ts`:

```
TypeError: (0 , adapter_registry_1.resolveProviderKey) is not a function
  at Object.runImport (src/modules/data_sync/lib/sync-engine.ts:479:45)
```

That suite mocks `../adapter-registry` wholesale and only exposed `getDataSyncAdapter`, while `sync-engine.ts` had started calling `resolveProviderKey`. A semantic merge conflict on `develop`: each side passed CI on its own, the combination didn't. It was repaired upstream in `07cf38603` ("test(data_sync): repair two adapter-registry mocks the merge exposed") — about 20 minutes after this branch last merged `develop`, so it landed right in that window.

I've merged current `develop` (`33a7d00c4`) in — clean, no conflicts — and the merged tree differs from `develop` only in the 7 `sales/` files that belong to this PR.

The new workflow runs for `57ae907a` are sitting in `action_required` (fork PR), so CI hasn't actually re-run yet. Could you approve them?

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-21T10:44:42Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review — re-review of `57ae907a`

## 🎯 Summary

This is my second pass on this PR. My first review (`8be9e08`, 2026-08-17) requested changes on one major and two minors; `@patzick` re-reviewed `c6d224a72` and approved. Since that approval the head has moved once, to `57ae907a` — a clean merge of `develop` (`33a7d00c4`) with no source change. The diff against the merge base is now exactly the seven files that belong to this PR, and **#5350 merged on 2026-08-18 (`3fc50f34`)**, so the merge-order caveat I raised in round 1 has resolved itself: this branch no longer carries someone else's API commit into the blast radius.

**Verdict: approve.** Every finding I raised is closed, and I checked the two that matter at the mechanism rather than at the description. Three nits remain, none of them blocking — including the two `@patzick` already flagged as non-blocking, which I confirm and re-state so they stay visible.

Worth recording plainly, because both previous rounds were reviewed without it: **this head has a complete, green CI run.** Round 1 had no test signal at all (the fork workflow was unapproved), and round 2 had `test` red on a failure inherited from a broken `develop`. `07cf38603` repaired that upstream, this branch merged it forward, and run [32459658349](https://github.com/open-mercato/open-mercato/actions/runs/32459658349) is green end to end on `57ae907a`.

## 🧪 Validation Gate

Validation source: **GitHub checks** — no local fallback was needed. Every command in `.ai/agentic.config.json` → `validation.commands` maps onto a step of the green run for this exact head SHA, so re-running the gate locally would only reproduce a signal CI already has.

| `validation.commands` entry | CI step | Result |
|---|---|---|
| `yarn build:packages` | `prepare` → Build packages | ✅ |
| `yarn generate` | `prepare` → Prepare generated modules | ✅ no drift |
| `yarn build:packages` (rebuild) | `prepare` → Rebuild packages with generated files | ✅ |
| `yarn i18n:check-sync` | `test` → Check i18n sync | ✅ |
| `yarn i18n:check-usage` | `test` → Check i18n usage | ✅ |
| `yarn typecheck` | `test` → Checking types | ✅ |
| `yarn test` | `test` → Test | ✅ (job green in 6m41s) |
| `yarn build:app` | `prepare` → Build app | ✅ |

Beyond the configured gate, also green on this head: `lint` (plus the raw-`console.*` guard), `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance`, and **`ephemeral-integration` in 11m18s**. `audit`, `mutate`, `merge-coverage` and `docker-build` are conditionally skipped, not failing. Branch protection on `develop` declares no required status contexts and one required approving review, so this review is the gate — which is precisely why it should not be given on a green-checks glance alone.

## ✅ Previous findings — all closed

**Major 1 — a legacy single-value perspective snapshot was discarded the moment the filter was touched. Fixed, verified at the mechanism.**

The normalization moved from the query builder to the hydration boundary (`SalesDocumentsTable.tsx:591-606`). That is only a fix if perspective restore actually re-enters state through that callback, so I traced it rather than trusting the description: `DataTable.applyPerspectiveSettings` restores a legacy `FilterValues` record by calling `onFiltersApply(legacy)` (`packages/ui/src/backend/DataTable.tsx:1951-1957`), and this page passes no `advancedFilter`, so that is the branch taken. I also confirmed there is no second hydration path — `filterValues` initializes to `{}` (`SalesDocumentsTable.tsx:178`) and all three perspective entry points (`DataTable.tsx:2046`, `2159`, `2600`) funnel through the same `applyPerspectiveSettings`. A bare-string `channelId` is therefore `string[]` before anything reads it, and `FilterOverlay`'s `Array.isArray(values[f.id])` guard (`FilterOverlay.tsx:362`) and `TagsInput.addValue` both see the restored channel.

The test change is the part that actually discriminates. My round-1 complaint was that the legacy-snapshot test asserted the URL only, which is exactly why the gap looked covered; `expect(lastProps().filterValues.channelId).toEqual([CHANNEL_A])` fails if the hydration fix is reverted, where a URL assertion does not.

**Minor 2 — the filter still showed raw uuids for channels outside the first options page. Fixed.**

`seedChannelOptionsForIds` (`SalesDocumentsTable.tsx:229-238`) fetches the selected ids via `?ids=` and merges them into `channelOptions`; the route supports it (`api/channels/route.ts:32`, `parseIdList` at `66-73`). Because seeding is invoked from `handleFiltersApply`, it also covers the case I cared about most — a perspective restored before the options page has loaded. Both consumers then resolve: the active-filter chip via `formatValue` (`FilterBar.tsx:129-131`, one chip per array entry at `152-158`) and the overlay's selected tag via the same callback (`FilterOverlay.tsx:369-371`). Agreed on leaving the customer filter alone — same defect, predates this change, its own PR.

**Minor 3 — no integration coverage. Accepted as the manual-QA route.**

You took the `needs-qa` route explicitly, which is what I asked for; the ambiguity I objected to is gone. I'd rather have that than a Playwright spec written blind against assertions you cannot execute, and `ephemeral-integration` being green on this head does not change it — that suite does not exercise this filter.

**Nit 4 — `(No channel)` matched every search term. Fixed, and belt-and-braces.**

Two independent mechanisms now prevent it: the sentinel is prepended only for an empty query (`SalesDocumentsTable.tsx:337-345`), and `TagsInput` filters its suggestion list locally by the typed query (`TagsInput.tsx:113-119`), so `(No channel)` cannot match `web` either way.

**Nit 5 — the 200-id cap. Left as is, as agreed.**

## 🔵 Nits

None of these block; they are the shape of thing worth a follow-up rather than another round trip.

**1. `seedChannelOptionsForIds` sends a `pageSize` the channels route rejects above 100** — `SalesDocumentsTable.tsx:233`

`pageSize: String(missing.length)` is passed straight through, but `listSchema` declares `pageSize: z.coerce.number().min(1).max(100)` (`api/channels/route.ts:29`). The documents API caps `channelIds` at 200 (`MAX_IDS_PER_REQUEST`, `packages/shared/src/lib/crud/ids.ts:3`), so a selection between 101 and 200 channels is expressible here — at which point the seeding request 400s, `fetchChannelOptionsBy` returns `[]` on `!call.ok`, and precisely those chips fall back to the raw uuids this PR set out to remove. `Math.min(missing.length, 100)`, or chunking the ids, closes it. I am filing this as a nit rather than a minor deliberately: it needs a merchant to select more than a hundred channels in one filter, the filtering itself stays correct, and the degradation is cosmetic and soft. `@patzick` raised the same thing as Low — same call, same reasoning.

**2. The seeding request repeats on every apply** — `SalesDocumentsTable.tsx:231`

`missing` filters out the sentinel but never checks `channelOptions` for ids already loaded, so every filter apply — including removing a single chip, which routes back through `onApply` (`FilterBar.tsx:146-151`) — re-fetches labels the component already holds. Diffing against the current options would make the variable name true and drop the request entirely in the common case.

**3. The sentinel's empty-query branch is unreachable from the UI, and a test asserts it anyway** — `SalesDocumentsTable.tsx:342`

`loadChannelFilterOptions` returns `[unassignedChannelOption, ...opts]` when there is no query, but `FilterOverlay` builds its tags loader with `if (!query.length) return []` (`FilterOverlay.tsx:172`) and never calls `loadOptions` with an empty query for a `tags` filter — nothing else consumes it. The behavior is still right, because the sentinel reaches the control through `f.options` (`channelFilterOptions`, `SalesDocumentsTable.tsx:330-333`) and `optionMap` dedupes by value so it cannot appear twice. The nit is the test: "does not offer the unassigned entry while searching" calls `filter.loadOptions()` directly, so its `browsed` half reads as coverage of "browsing offers the sentinel" while actually exercising a path the overlay never takes. The `searched` half is genuine. Asserting `filter.options[0].value` for the browsing direction — which the suite already does elsewhere — would be honest about where the sentinel comes from.

## 💥 Breaking Changes

- [x] **No contract surface removed or renamed.** Confined to one client component plus two additive i18n keys; `channelId` (singular) stays in the documents list schema and still takes precedence server-side, and nothing now sends it.
- [x] **No new API surface in this diff.** `channelIds` / `channelIdsEmpty` (#5198) and `channelName` (#5350) are both already on `develop`; I re-checked the parameter names against `api/documents/factory.ts:95-105`, `141-151` and they match what the component emits, including the `$or` combination when the sentinel is picked alongside real channels.
- [x] **No database, migration, event-id, widget-spot-id, DI-key, ACL-feature or CLI change.**
- [x] **Persisted client state now has a complete bridge.** `filterValues.channelId` moves from `string` to `string[]` in perspective snapshots, and the round-1 gap on the control side is closed (Major 1 above). The round trip also holds for a snapshot saved *with* the sentinel: it normalizes to an array, seeding skips it, and the query builder turns it back into `channelIdsEmpty`.
- [x] **Export stays consistent with the view.** `buildCrudExportUrl` derives from `currentParams`, itself derived from `queryParams` (`SalesDocumentsTable.tsx:509-519`), so an exported view carries the same `channelIds` / `channelIdsEmpty` narrowing as the grid. The `cf_`-only catch-all loop (`488-505`) does not re-emit `channelId`, so there is no chance of the singular parameter shadowing the plural one server-side.

## 🧪 Test Coverage

Ten cases now, up from seven, and the two added ones are the two I asked for. `packages/core/src/modules/sales/i18n/{en,pl,es,de,ko}.json` all carry both new keys as real translations rather than English fallbacks, and `i18n:check-sync` confirms it in CI.

My round-1 closing observation stands and you and `@patzick` both already agreed with it, so I will not re-litigate it: mocking `DataTable` makes this suite a wiring test rather than a rendering test, and that is the structural reason the `FilterOverlay` half of the legacy-snapshot path went unnoticed. The stored-shape assertion is a real partial answer — it catches state the real control would misread — but nit 3 above is the same seam showing again from the other side. Worth its own change; not this PR's job.

## 🚦 Merge gating

Approving the code does not clear the PR to merge. What is left is not in this diff:

1. **`needs-qa` → `qa-approved`** by a QA reviewer. Per the repository's hard gate a `needs-qa` PR must not merge on code-review approval alone, and I cannot grant QA approval from reading a diff. Instructions follow in a separate comment.
2. Nothing else. #5350 has merged, `develop` is repaired, and CI is green on this head — all three of the blockers named in the previous round are gone.

pkarw:

Everything here is scoped to this diff — one component (`SalesDocumentsTable.tsx`) plus two i18n keys. It drives **both** `/backend/sales/orders` and `/backend/sales/quotes`, so anything that fails on one should be re-checked on the other.

**Setup worth doing first** (at `/backend/sales/channels`): have at least **three** channels, and at least one order assigned to each, plus **one order with no channel at all**. The unassigned one is the interaction nobody has watched in a browser yet.

### P0 — Saved views created before this change

**Where to click**
- `/backend/sales/orders` — before checking out this branch, filter by a single channel and save it as a view (this writes the old single-value shape into the perspective)
- Then, on this branch, reopen that saved view from the views selector

**What to verify**
- The restored view still filters to that one channel, and the active-filter chip reads the channel's **name**, not a uuid
- Open **Filters** — the Channel field shows that channel as a removable tag, not an empty field
- Add a second channel and Apply → the list widens to **both** channels; the first one is not silently dropped
- Save the view again, reload the page, reopen it → both channels come back

**What can go wrong**
- The list is filtered but the Channel field in the overlay looks empty (this was the bug fixed in round 1 — it is the single most important thing to confirm)
- Picking a second channel replaces the restored one instead of joining it
- The chip or the tag renders a raw uuid instead of a channel name

### P1 — Channel column shows names

**Where to click**
- `/backend/sales/orders` and `/backend/sales/quotes`, on first paint (do not touch the filters)

**What to verify**
- The Channel column shows channel **names** on the very first render, with no flicker from uuid → name
- An order with no channel shows the "Unassigned" muted text
- If you can create more than 50 channels and assign an order to one of the last of them, that order's Channel cell still shows the name — this is the >50-channels defect the PR exists to fix

**What can go wrong**
- A uuid rendered instead of a name (only legitimate when the API genuinely could not resolve the channel)
- The column briefly showing ids and then swapping to names

### P1 — Multi-channel filter and `(No channel)`

**Where to click**
- `/backend/sales/orders` → **Filters** → Channel

**What to verify**
- Opening the field with nothing typed lists `(No channel)` **first**, then real channels
- Typing part of a channel name filters to it and does **not** offer `(No channel)`
- Select **two real channels** → Apply → the list shows orders from both, and there are two chips, one per channel
- Select **`(No channel)` on its own** → Apply → only the order with no channel is listed
- Select **`(No channel)` plus one real channel** → Apply → the list shows that channel's orders **and** the unassigned one (this is the combination the component tests assert but nobody has seen render)
- Remove one chip with its `×` → the list re-narrows correctly and the remaining chip keeps its name
- Clear all filters → the full list returns
- Export the filtered view (CSV/Excel) → the exported rows match the filtered grid, not the whole list

**What can go wrong**
- Picking `(No channel)` alongside real channels returns only one of the two groups
- The Filters button count says a filter is active while no chip is shown
- A chip reverts to a uuid after removing another chip
- Export ignores the channel narrowing

### P2 — Perceived performance and mobile

**Where to click**
- `/backend/sales/orders` cold-loaded (hard refresh), then at a mobile viewport

**What to verify**
- Cold load shows the loading state and then the grid with channel names already resolved; screenshot it if you can
- Opening the Channel filter and typing feels responsive — no visible stall while suggestions load
- At mobile width the chips wrap and truncate rather than overflowing the row

**What can go wrong**
- A burst of `/api/sales/channels?ids=…` requests on every apply (known nit, non-blocking — worth noting how visible it is, not worth failing on)
- Chips overflowing horizontally on narrow screens

### Pass/fail

pkarw:

Re-review of `57ae907a` — my round-1 `changes-requested` is cleared. All three actionable findings I raised are closed, and I checked the major one at the mechanism rather than the description: perspective restore really does re-enter state through `onFiltersApply` (`DataTable.tsx:1951-1957`), this page passes no `advancedFilter` so that is the branch taken, and `filterValues` has no other hydration path (`useState({})`, all three perspective entry points funnelling through `applyPerspectiveSettings`). Three nits remain, none blocking — the >100-channel `pageSize` cap and the repeated seeding request that @patzick already flagged, plus one new one: `loadChannelFilterOptions`' empty-query branch is unreachable from the overlay, so the "browsing offers the sentinel" half of that test exercises a path the UI never takes.

**Validation source: GitHub checks, no local fallback.** Every command in the configured gate maps onto a green step of run [32459658349](https://github.com/open-mercato/open-mercato/actions/runs/32459658349) on this exact head, with `ephemeral-integration` green in 11m18s alongside it. Worth stating because neither previous round had this: round 1 had no test signal at all, round 2 had `test` red on a failure inherited from a broken `develop`, and both are now resolved.

Labels: `changes-requested` → `merge-queue`, and `risk-medium` → `risk-low` because the reason for the medium rating is gone — #5350 merged on 2026-08-18, so this branch no longer carries its API commit and the diff against the merge base is exactly this PR's seven files. `needs-qa` stays on; the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`, and the manual-QA instructions are in the comment above.

autofix: skipped (not my PR — re-run with --autofix to fix it here). The three remaining nits are follow-up material, not a reason to hold this.

PR #5352, opened 2026-08-17
fix(auth): emit the declared logout and password lifecycle events

Frshy:
## Summary

`packages/core/src/modules/auth/events.ts` declares six authentication lifecycle events. Only two are ever emitted:

| Declared event | Emitted before this PR |
|---|---|
| `auth.login.success` | ✅ `api/login.ts`, `api/autologin.ts` |
| `auth.login.failed` | ✅ `api/login.ts` |
| `auth.logout` | ❌ never |
| `auth.password.changed` | ❌ never |
| `auth.password.reset.requested` | ❌ never |
| `auth.password.reset.completed` | ❌ never |

`createModuleEvents` registers every declared id in the global registry, so all six appear in the workflow trigger picker and in `getDeclaredEvents()`, and any module can subscribe to them without a warning. Four of those subscriptions then never fire — the handlers that own the transitions simply never call `emitAuthEvent`. There is no comment, TODO or spec explaining the omission, and the sibling `customer_accounts` module does emit its equivalents (`customer_accounts.password_reset.requested` in `api/password/reset-request.ts`, `customer_accounts.password.changed` in `api/password/reset-confirm.ts`), so this reads as an oversight rather than a deliberate gap.

Impact: audit trails, security alerting, session bookkeeping and "your password was changed" notifications built on the documented event surface silently receive nothing.

This PR adds one emit per handler that owns the transition — no new event ids, no new API surface.

## Changes

One emit per handler, in the exact fire-and-forget style already used by `api/login.ts` (`void emitAuthEvent(...).catch(() => undefined)`):

| Handler | Event | Payload |
|---|---|---|
| `api/logout.ts` | `auth.logout` | `id`, `tenantId`, `organizationId`, `sessionId`, `sessionRevoked`, `at` |
| `api/reset.ts` | `auth.password.reset.requested` | `id`, `tenantId`, `organizationId`, `at` |
| `api/reset/confirm.ts` | `auth.password.reset.completed` | `id`, `tenantId`, `organizationId`, `at` |
| `commands/users.ts` (`auth.users.update`) | `auth.password.changed` | `id`, `tenantId`, `organizationId`, `changedBy: 'self' \| 'admin' \| 'system'`, `changedById`, `at` |

- **`logout.ts`** already verified `auth_token` to pull the `sid` claim. That module-private helper widens from `extractSessionIdFromAuthToken` to `extractAuthTokenClaims`, so the single existing `verifyJwt` call now yields `sub` / `sid` / `tenantId` / `orgId` at once — no extra verification, no extra DB read. The event fires only when a verifiable token identified the caller; an unauthenticated or tampered-token logout still clears cookies and emits nothing. The payload identifies the user by `id` alone — see *Payloads carry no `email`* below — plus `sessionId` and `sessionRevoked`. The latter is `true` only when server-side revocation completed without throwing — the surrounding `catch {}` swallows revocation failures, so without it a session-bookkeeping subscriber could mark a session closed while the row is still live and the refresh token still redeemable.
- **`reset.ts`** emits as soon as `requestPasswordReset` has issued a token, so the event reflects "a reset was requested and a token exists" independently of whether the email or the in-app notification later succeeds. **The reset token is never part of the payload.** Unknown accounts still return `ok: true` and emit nothing, so the event cannot be used as an account-existence oracle.
- **`reset/confirm.ts`** emits after the password has actually been replaced, before the notification block.
- **`commands/users.ts`** emits from the `auth.users.update` write path, guarded by the same `hashed` flag the command already computes. Both password-write callers — profile PUT and the admin users route — converge there, so admin-set passwords are covered by the same single emit site. `changedBy` is derived from the actor: `self` when the actor is the account owner, `admin` for any other authenticated caller, and `system` when the command runs without an auth context or under `ctx.systemActor`. Without that third value an unattended write would report as a named administrator acting on someone else's account, which is exactly the case security alerting escalates on; `changedById` still carries the actor id when automation happens to have one. The value set, together with the boundary that `mercato auth set-password` and tenant provisioning assign `passwordHash` directly and emit nothing at all, is documented on the declaration in `events.ts`, where subscriber authors look the id up.

All four are emitted with `{ persistent: true }`, and all four are `void`-ed and `.catch()`-wrapped, so a slow or failing event bus can never break logout or password recovery — covered by an explicit test per handler.

### Duplicate-emission check

Traced where these events actually land, to be sure nothing gets double-recorded:

- **Audit log is not event-driven** — `audit_logs` has no `subscribers/`; entries come from command-bus interceptors. These emissions add zero audit rows.
- **The only wildcard consumer** is `business_rules:crud-rule-trigger` (`event: '*'`, persistent). `auth.logout` has two segments, so its `parseEventName` returns `null` and it is dropped immediately; the other three resolve to entity types `auth.password` / `auth.password.reset`, for which no rules exist → cached no-op lookup. `auth.login.success` / `auth.login.failed` already flow through the same path today.
- **No duplicate notifications** — `reset.ts` and `reset/confirm.ts` create their in-app notification directly via `notificationService`, not from the event.
- **Reset path is single-source** — `authService.confirmPasswordReset` mutates directly (no command), so only the new event fires.
- Persistent delivery goes through `packages/events/src/single-delivery.ts`, so no double delivery.

Two things worth knowing:

1. A self-service password change now emits **both** `auth.user.updated` (crud, from `emitCrudSideEffects` inside the command) and `auth.password.changed`. Different ids, but a subscriber listening to both runs twice for one change.
2. If anyone later adds a subscriber that builds the reset notification from the event, they must drop the direct `notificationService.create` in the handler — otherwise that *would* duplicate.

### Deliberate scope choices

1. **The reset path is not dual-emitted.** A completed reset emits `auth.password.reset.completed` only, not also `auth.password.changed` — the reset flow does not go through `auth.users.update` (`authService.confirmPasswordReset` writes directly), and dual-emitting would double-fire any "your password changed" notification subscribing to both. `customer_accounts` went the other way and uses `password.changed` with `changedBy: 'reset'` as one umbrella event; happy to align if maintainers prefer that shape.
2. **Invitation activation emits `auth.password.reset.completed`.** Invite emails link to `/reset/<token>` and post to the same confirm endpoint, so a newly invited user setting their initial password produces a `.completed` with no preceding `.requested`. Documented on the declaration in `events.ts` so subscriber authors see it where they look up the id. A `via: 'reset' | 'invite'` discriminator is possible — invited users are created with `passwordHash: null`, so the service can tell them apart — but surfacing it means widening `confirmPasswordReset(token, password): Promise<User | null>`, a DI-exposed service signature and therefore a STABLE contract surface; that wants an additive sibling method plus a deprecation path, i.e. its own PR.
3. **Payloads carry no `email`.** All four emit with `{ persistent: true }`, and `packages/events/src/bus.ts` enqueues the whole payload whether or not a subscriber matches — into Redis under `QUEUE_STRATEGY=async`, where `removeOnFail: 1000` retains failed jobs, or an on-disk `queue.json` locally. `email` is encrypted at rest under the `auth:user` map in `encryption.ts`, so carrying it here would put plaintext PII outside the tenant encryption boundary and outside anything a GDPR erasure request reaches. The `auth.login.success` / `auth.login.failed` precedent does not carry that risk — those emits pass no options, so their email never leaves process memory. Each payload therefore identifies the user by `id`, `tenantId` and `organizationId`, the shape `customer_accounts.password.changed` already uses; a subscriber that needs the address resolves it with `findOneWithDecryption` under the tenant's own key. One test per emit site asserts the payload has no `email` key and that the address appears nowhere in it.
4. **Undoing an `auth.users.update` does not emit.** The command's `undo` restores `before.passwordHash`, so a reverted password change produces no `auth.password.changed`. Deliberate: undo/redo are gated on `audit_logs.undo_self` / `audit_logs.undo_tenant` rather than on the features that authorize the write, the audit trail already records the reversal, and emitting there would re-fire a "your password was changed" notification for something the account owner never initiated. Pre-existing shape — the emit sits on the forward write path only.

### Backward compatibility

Additive only, per `BACKWARD_COMPATIBILITY.md`:

- No event id added, renamed or removed. `events.ts` changes by explanatory comments only.
- No API route, request/response body, OpenAPI doc, DB schema, DI key, ACL feature or notification id changes.
- The only behavioral delta: four already-declared events now reach the bus. Subscribers that never fired begin firing; nothing that fired before changes.
- One non-exported local helper in `logout.ts` was widened (`extractSessionIdFromAuthToken` → `extractAuthTokenClaims`) — module-private, not a contract surface.
- `auth.users.update` gains an emit; its input schema, result type and side-effect order are unchanged.
- The four events now enqueue for the events worker. No subscriber exists for them today, so the durable path starts empty.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**
n/a — this restores already-declared, already-documented behavior rather than adding a feature, so per `AGENTS.md` ("skip for small fixes") nothing was added to `.ai/specs/`.

## Review feedback addressed (99918cfd7)

- **Major — `email` in a `{ persistent: true }` payload.** Dropped from all four payloads; `persistent: true` stays, so the audit and security-alerting subscribers these events exist for keep durable delivery. `logout.ts` no longer extracts the claim at all, so nothing populates a field no payload reads. Full rationale under *Payloads carry no `email`* above.
- **Minor — the `changedBy: 'system'` comments documented coverage that does not exist.** Confirmed: `cli.ts` `set-password` (line 780) and `lib/setup-app.ts` (line 419) assign `passwordHash` directly and never reach `auth.users.update`. Both comments now say what `system` actually denotes — a command invocation with no auth context, or one running under `ctx.systemActor` — and state plainly that those two paths emit nothing, so nobody wires a CLI-reset alert against an event that cannot see it. I took the "make the comment true" option rather than the "add the emits" one, to keep this PR to the auth emit sites; say the word and the CLI emit plus its test can land here instead of as a follow-up.
- **Nit — `String(user.email)` could stringify `undefined`.** Moot, the field is gone.
- **Nit — `sessionRevoked` means "revocation did not throw".** Now stated on the `auth.logout` declaration, where subscriber authors read it. `deleteSessionById` / `deleteSessionByToken` are left alone: returning an affected count changes a DI-exposed `AuthService` signature, a STABLE contract surface under `BACKWARD_COMPATIBILITY.md`, which is more than a nit is worth.

## Testing

Rebased onto `develop` before submitting.

```
yarn workspace @open-mercato/core test   → 1246 suites, 9696 tests, all passing
yarn typecheck                           → 22/22 packages
eslint <the nine touched files>          → clean
```

Scoped run for the changed behavior:

```
yarn workspace @open-mercato/core exec jest \
  modules/auth/api/__tests__/logout.test.ts \
  modules/auth/api/__tests__/reset.test.ts \
  modules/auth/api/__tests__/reset-confirm.test.ts \
  modules/auth/commands/__tests__/users.password-changed.test.ts
→ 4 suites, 35 tests, all passing (20 new)
```

Per emit site: the happy-path payload, the negative case where nothing should be emitted (unverifiable token / unknown account / update without a password), and a rejecting event bus that must not change the outcome. Each suite additionally asserts that its payload carries no `email` key and that the address appears nowhere in it. `reset.test.ts` asserts the emitted payload never contains the reset token; `logout.test.ts` drives revocation into a rejection and asserts `sessionRevoked: false` with cookies still cleared.

- `api/__tests__/logout.test.ts` — extended (6 new, 14 total)
- `api/__tests__/reset.test.ts` — extended (4 new, 10 total)
- `api/__tests__/reset-confirm.test.ts` — new (4)
- `commands/__tests__/users.password-changed.test.ts` — new (7: self, admin, system without auth context, system under `ctx.systemActor`, no-email-in-payload, no-password, rejecting bus)

**Why no `.ai/qa/tests/` integration coverage:** the change adds no HTTP-observable behavior. Every route keeps its existing request schema, response body and status codes; the only delta is four emissions on the internal event bus, and no subscriber for them exists in the repo today. A Playwright integration test therefore has nothing to assert — it would have to register a subscriber that ships nowhere, or observe a UI surface this PR does not touch. The emissions are asserted at their source instead, one test per emit site, including the negative and failure cases listed above.

## Checklist

- [x] I updated documentation, locales, or generators if the change requires it. — no locale or generator surface changes; the two clarifying comments on the `events.ts` declarations are the documentation this change needs.
- [x] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required). — documented under Testing.
- [ ] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — n/a, see Specification.
- [ ] QA routing set.

> The three label boxes are unticked because this account has `read` permission on this repository and cannot apply labels. The intended values, with rationale, are in a separate comment below — a maintainer with `triage` needs to apply them. The CLA box is deliberately left for the human author to tick.

### Design System Compliance

Not applicable — this change touches no UI. No `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no styles, no markup. Each box left unticked because there is no surface for it to describe, not because it was skipped.

- [ ] No hardcoded status colors
- [ ] No arbitrary text sizes
- [ ] Empty state handled for list/data pages
- [ ] `aria-label` on all icon-only buttons
- [ ] Uses existing DS components

## Linked issues

None — found while auditing declared-vs-emitted events across the platform, no tracking issue was open.

A follow-up audit of all declared events found the same bug class elsewhere, kept out of this PR to keep it single-module: `workflows.task.assigned` (emit site exists in `lib/step-handler.ts` → `handleUserTaskStep`) and `customer_accounts.user.locked` / `.unlocked` (`customerUserService.incrementFailedAttempts` / `resetFailedAttempts`, needs a guard so "unlocked" does not fire on every successful login) are the identical one-line shape. `catalog.product.stock_low` and `sales.quote.expiring` have a subscriber, a notification type and translations but need a threshold detector and a scheduler respectively, so they are features rather than missing lines. Happy to open those separately if maintainers want them.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5352"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789561862&installation_model_id=432243&pr_number=5352&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5352&signature=2d5e89ec3f2963415ab037a2ed3f44da966571b3026f6460a52a9eb8ffc91be0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Frshy:
**Metadata intent** — this account has `read` permission here, so it cannot apply labels, an assignee or a reviewer. Listing the intended values so a maintainer with `triage` can apply them in one pass. Every name below was validated against this repository's label list.

| Label | Rationale |
|---|---|
| `review` | Ready for review, not in any other pipeline state. |
| `bug` | Four declared events never reached the bus; this restores documented behavior rather than adding a feature. |
| `priority-medium` | No data loss and no broken user-facing flow, so not high — but the missing events are the substrate for audit trails and security alerting, and any subscriber written against them has been silently dead since the ids were declared. |
| `risk-low` | Nine files, one module, additive only. No schema, route, response, DI key, ACL feature or event id changes. The only delta is four already-declared events starting to fire, and no subscriber for them exists in the repo today, so the durable path starts empty. Each emit is `void`-ed and `.catch()`-wrapped, with a test per handler asserting a rejecting bus cannot change the outcome. |
| `skip-qa` | Proposed under the automated-verification exemption in `AGENTS.md` → PR Workflow: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, database structure and API surface unchanged, no `BACKWARD_COMPATIBILITY.md` contract broken, and automated tests for the changed behavior ship in the same PR. If maintainers read it as `needs-qa` instead, note there is no UI surface to click through — the observable effect is bus traffic. |

Assignee: none intended — this account cannot self-assign here, and no claim label is being requested.

Reviewer: whoever owns the `auth` module / the events surface. Naming a role rather than a handle deliberately.

Note on the QA gate: if `needs-qa` is applied instead of `skip-qa`, the PR stays gated until a maintainer adds `qa-approved` — this comment does not clear it.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-17T13:10:08Z. Other auto-skills will skip this PR until the lock is released.

pkarw (dismissed):
# 🔍 Code Review: fix(auth): emit the declared logout and password lifecycle events

## 🎯 Summary

This PR closes a real and well-diagnosed gap: `packages/core/src/modules/auth/events.ts` has declared `auth.logout`, `auth.password.changed`, `auth.password.reset.requested` and `auth.password.reset.completed` since the module was written, `createModuleEvents` registers all four in the global registry so they appear in the workflow trigger picker and in `getDeclaredEvents()`, and nothing ever emitted them. Any subscriber written against those ids has been silently dead. The fix is four one-line emits placed on the handlers that actually own each transition, plus a widening of the module-private `extractSessionIdFromAuthToken` helper in `logout.ts` into `extractAuthTokenClaims` so the single existing `verifyJwt` call yields `sub`/`sid`/`email`/`tenantId`/`orgId` at once — no extra verification and no extra DB read.

The engineering around it is unusually careful and I want to say so plainly. The claim extraction is correctly guarded (an unverifiable or tampered token clears cookies and emits nothing), the reset-request path emits only for accounts that exist so it cannot be used as an account-existence oracle, the reset token is never in the payload and there is a test asserting exactly that, `changedBy` is derived once in the `auth.users.update` command so both password-write callers converge on a single emit site, and every emit is `void`-ed and `.catch()`-wrapped with a per-handler test driving a rejecting bus. I independently traced the duplicate-emission analysis in the PR body and it holds: `business_rules:crud-rule-trigger` (`event: '*'`) drops `auth.logout` at `parseEventName` because it has only two segments, `audit_logs` is command-interceptor-driven rather than event-driven, and the reset handlers create their notifications directly rather than from the bus.

Reviewed scope: the nine changed files, the auth event/emit surface, the events bus persistence path in `packages/events/src/bus.ts`, the queue strategies, the auth encryption map, and every other write path to `User.passwordHash` in the repository.

Two things need a decision before this lands: one payload/delivery combination that puts an encrypted-at-rest field into durable storage, and one comment that documents coverage the code does not actually have.

## Verdict

❌ **request changes** — the `email` field combined with `{ persistent: true }` writes a field this platform encrypts at rest (`auth:user.email`) into the durable events queue on four auth paths, with no subscriber that needs it (Major); and the new `changedBy: 'system'` documentation names CLI/worker/provisioning password writes that bypass `auth.users.update` entirely and therefore emit nothing at all (Minor — this repository requests changes on Minor findings too). Both are small, contained fixes; nothing about the approach needs to change.

## 🧪 Validation Gate

**Validation source: local fallback.** GitHub provided no usable gate signal for head `2c1fd1a` — the only reported PR checks are `[code]smith` (SKIPPED) and `license/cla` (SUCCESS), and the two workflow runs that would have run the gate, *CI for Develop&Main* ([run 32030241327](https://github.com/open-mercato/open-mercato/actions/runs/32030241327)) and *Mutation tests* ([run 32030241420](https://github.com/open-mercato/open-mercato/actions/runs/32030241420)), both concluded `action_required` — they are waiting on a maintainer to approve workflows for this fork. Branch protection reports no required status checks on `develop` (HTTP 404). I therefore ran a narrow local gate, scoped to the changed package, in an isolated worktree at the PR head. Runner: `local` (no compose `app` container was running).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | 22/22 packages built. |
| `yarn generate` | ✅ PASS | Regenerated the module registries; needed before typecheck can resolve `#generated/entities.ids.generated`. |
| `yarn build:packages` (rebuild) | ✅ PASS | Covered by the turbo run above. |
| `yarn i18n:check-sync` | ⏭️ not run | No locale files or user-facing strings in the diff; narrow-fallback scope. |
| `yarn i18n:check-usage` | ⏭️ not run | Same reason. |
| `yarn typecheck` | ✅ PASS (scoped) | `yarn workspace @open-mercato/core typecheck` — 0 errors. Workspace-wide typecheck not re-run; no finding here touches a shared contract. |
| `yarn test` | ✅ PASS (scoped) | `packages/core`: the four affected suites — `api/__tests__/logout.test.ts`, `api/__tests__/reset.test.ts`, `api/__tests__/reset-confirm.test.ts`, `commands/__tests__/users.password-changed.test.ts` — 4 suites, 31 tests, all passing. Full `@open-mercato/core` suite not re-run locally; the author reports 1246 suites / 9690 tests green. |
| `yarn build:app` | ⏭️ not run | No app-layer, route-manifest or template surface in the diff. |
| `eslint` (nine touched files) | ✅ PASS | Clean. |

⚠️ **Merge blocker outside this review:** the fork's workflow runs are `action_required`. A maintainer has to approve them so the real gate executes on this head — my scoped local run is this review's evidence only and does not substitute for CI.

## Findings

### ⚠️ Major

**1. `email` in a `{ persistent: true }` payload writes an encrypted-at-rest field into the durable events queue** — `packages/core/src/modules/auth/api/logout.ts:73`, `packages/core/src/modules/auth/api/reset.ts:63`, `packages/core/src/modules/auth/api/reset/confirm.ts:39`, `packages/core/src/modules/auth/commands/users.ts:739`.

`auth:user.email` is declared in `packages/core/src/modules/auth/encryption.ts` `defaultEncryptionMaps`, which is why this module's `AGENTS.md` requires `findWithDecryption`/`findOneWithDecryption` for every user read — the platform's position is that a user's email is a GDPR-relevant value that does not sit in plaintext at rest. `{ persistent: true }` changes where these payloads live. In `packages/events/src/bus.ts:447-461`, a persistent emit calls `q.enqueue({ event, payload, options, … })` **unconditionally**, whether or not any subscriber matches. That serializes the whole payload — plaintext email included — into the queue's durable store: BullMQ/Redis under `QUEUE_STRATEGY=async`, where `packages/queue/src/strategies/async.ts:210-211` sets `removeOnFail: 1000` and so retains up to a thousand failed job payloads for inspection, or an on-disk `queue.json` under the local strategy.

The PR body flags the email field as deliberate, "matching `auth.login.success` / `auth.login.failed`". That precedent does not carry the persistence: `api/login.ts:172` and `api/autologin.ts:69` emit **ephemerally**, with no options argument, so today that email value never leaves process memory. It is the *combination* introduced here — email in the payload **and** `persistent: true` — that creates a new plaintext-PII-at-rest surface, on the four highest-traffic auth paths, for a field no subscriber currently consumes. The sibling module the PR cites as its precedent went the other way on exactly this point: `customer_accounts.password.changed` (`packages/core/src/modules/customer_accounts/api/password/reset-confirm.ts:54`) and `customer_accounts.password_reset.requested` (`reset-request.ts:67`) carry `userId`/`tenantId` only, no email, and are not persistent either.

Concrete failure: with `QUEUE_STRATEGY=async` and any transient events-worker failure, up to 1000 job payloads containing user email addresses in plaintext sit in Redis under `removeOnFail` retention, outside the tenant encryption boundary that `TenantDataEncryptionService` exists to enforce, and outside anything a GDPR erasure request would find.

Fix — either one resolves it, and the choice is yours:
- **Drop `email` from the four payloads.** Every payload already carries `id`, `tenantId` and `organizationId`, so a subscriber that genuinely needs the address can resolve it with `findOneWithDecryption` under the tenant's own key at the moment it needs it. This is the shape `customer_accounts` already uses and it is my recommendation.
- **Or keep `email` and drop `{ persistent: true }`**, so the four emissions match the ephemeral `auth.login.success` precedent the PR body invokes and the value stays in process memory.

What should not ship is both together.

### 🔹 Minor

**2. The documented `changedBy: 'system'` coverage does not exist — the CLI and provisioning password writes bypass `auth.users.update` entirely** — `packages/core/src/modules/auth/events.ts:23-26`, `packages/core/src/modules/auth/commands/users.ts:730-733`.

Both new comments tell a subscriber author that `system` means "internal automation running without an auth context — CLI, workers, provisioning". That is not what happens. The only two callers of `auth.users.update` in the repository are `packages/core/src/modules/auth/api/profile/route.ts:173` (self-service) and `packages/core/src/modules/auth/api/users/route.ts:166` (admin), and both are authenticated routes that always supply `ctx.auth`. The out-of-band writers named in the comment do not go through the command at all:

- `mercato auth set-password` sets `user.passwordHash = await hash(password, 10)` and flushes directly at `packages/core/src/modules/auth/cli.ts:780-781`.
- Tenant provisioning overwrites an existing user's hash directly at `packages/core/src/modules/auth/lib/setup-app.ts:419`.

Neither emits anything, so `changedBy: 'system'` is effectively unreachable from any shipped call site.

Why this matters more than a wording slip: an out-of-band password reset performed by an operator over the CLI is precisely the event that security alerting exists to catch — it is the case the PR body itself calls out as "exactly the case security alerting escalates on". The comment sits on the declaration in `events.ts`, which is where a subscriber author looks the id up, so they will wire the alert believing CLI resets are covered and silently miss them.

Fix — either:
- Add the emit to `cli.ts` `set-password` (and to the `setup-app.ts:419` branch that overwrites an existing user's hash) with `changedBy: 'system'`, which makes the documented contract true; or
- Narrow both comments to describe what `system` actually denotes today — a command invocation with no auth context or under `ctx.systemActor` — and state plainly that direct `passwordHash` writes outside the command (CLI `set-password`, provisioning) do not emit, so subscriber authors know the boundary.

### 💅 Nit

**3. `String(user.email)` can stringify `undefined`** — `packages/core/src/modules/auth/commands/users.ts:739`. The sibling serializers in the same file use `String(user.email ?? '')` (lines 1123 and 1144). If the entity's email were ever absent, this payload carries the literal string `"undefined"` rather than an empty value. Moot if you take the recommended option on finding 1.

**4. `sessionRevoked` reports "revocation did not throw", not "a session row was removed"** — `packages/core/src/modules/auth/api/logout.ts:67`. The flag is set after the deletes without checking whether either one matched a row, so a logout whose session had already expired server-side still reports `sessionRevoked: true`. The PR body describes it as the stronger guarantee ("a session-bookkeeping subscriber could mark a session closed while the row is still live"), and the conservative direction — a failure reporting `false` — is right and is tested. Worth either a one-line note on the `auth.logout` declaration, or having `deleteSessionById`/`deleteSessionByToken` return an affected count so the flag can mean what its name suggests.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the only widened helper, `extractSessionIdFromAuthToken` → `extractAuthTokenClaims`, is module-private in `logout.ts` and is not a contract surface under `BACKWARD_COMPATIBILITY.md`.
- [x] No function signature changed in a breaking way — `auth.users.update` keeps its input schema, result type and side-effect order; the emit is appended after `emitCrudSideEffects`.
- [x] No HTTP route URL removed or renamed; no method changed — `logout.ts` still exports `POST` only, and the existing CSRF-hardening tests asserting that are preserved and extended.
- [x] No field removed or retyped in an existing response shape — all three routes return exactly what they returned before.
- [x] No event or message name renamed or removed; no payload field removed — strictly additive: four already-declared ids begin reaching the bus, and `events.ts` changes by comments only.
- [x] No database table or column renamed or removed — no migration in the diff and none needed.
- [x] Where a contract had to change: not applicable, nothing changed.

The one genuine behavioral delta is that subscribers which never fired begin firing. Since no subscriber for these four ids exists in the repository today, the durable path starts empty — correctly characterized in the PR body.

## 🧪 Test Coverage

The coverage is genuinely good and goes past the happy path. Each emit site gets three shapes: the exact payload asserted field-for-field including `{ persistent: true }`; the negative case where nothing should be emitted; and a rejecting event bus that must not change the caller's outcome. `logout.test.ts` drives `deleteSessionById` into a rejection and asserts `sessionRevoked: false` with cookies still cleared, covers a legacy token with no `sid` claim, and asserts no emit at all when the `auth_token` cannot be verified. `reset.test.ts:218` asserts via `JSON.stringify(payload)` that the reset token never appears in the emitted payload — the right way to test that invariant, since it catches the token arriving under any key. `users.password-changed.test.ts` covers all four `changedBy` derivations (`self`, `admin`, `system` with no auth context, `system` under `ctx.systemActor`) plus the no-password no-emit case. I ran all four suites at the PR head: 31 tests, all passing.

The decision to skip `.ai/qa/tests/` integration coverage is argued rather than skipped, and I agree with the argument — every route keeps its request schema, response body and status codes, so a Playwright test would have to register a subscriber that ships nowhere in order to observe anything.

Two gaps to close alongside the findings above:

- **Finding 2** has no test because the behavior it documents does not exist. If you take the "make the contract true" option, `packages/core/src/modules/auth/cli.ts` needs a test asserting `mercato auth set-password` emits `auth.password.changed` with `changedBy: 'system'` and `changedById: null`, alongside the existing CLI tests.
- **Finding 1**, if resolved by dropping `email`, wants one assertion per emit site that the payload has no `email` key — that keeps the "no PII in durable event payloads" decision from silently regressing the next time someone extends these payloads. If resolved by dropping `persistent`, the existing `{ persistent: true }` assertions in all four suites need to flip, which the test suite will tell you about immediately.

pkarw:

- ❌ `changes-requested` — the code review found one Major finding (the `email` field combined with `{ persistent: true }` writes an encrypted-at-rest value into the durable events queue) and one Minor finding (the new `changedBy: 'system'` comments document CLI/provisioning coverage that does not exist); this repository requests changes on Minor findings too.
- 🐛 `bug` — four events were declared, registered in the global registry and exposed in the workflow trigger picker while never being emitted, so this restores documented behavior rather than adding a capability. This matches the author's stated intent.
- 🔹 `priority-medium` — no user-facing flow is broken and no data is lost, so this is not release-blocking; the urgency comes from every subscriber written against these four ids having been silently dead since they were declared. I considered `priority-high` because the change sits on auth/session paths, but the heuristic there targets regressions and security hardening, and this is neither.
- 🟡 `risk-medium` — raising the author's proposed `risk-low`. The change is additive and well tested, but it is not isolated: it rewrites the control flow of the logout request handler (claim extraction widened, a new `sessionRevoked` state), adds an emit inside the `auth.users.update` command, and turns on four durable queue streams on the highest-traffic auth paths. That is more blast radius than `risk-low` describes, and well short of the schema/contract changes `risk-high` is for.
- ⏭️ `skip-qa` — the automated-verification exemption in `AGENTS.md` → PR Workflow applies: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, database structure and API surface unchanged, no `BACKWARD_COMPATIBILITY.md` contract broken, and automated tests for the changed behavior ship in the same PR. There is no UI surface to click through — the observable effect is bus traffic.

`partner-request` left untouched.

pkarw:

Two notes that are not about your code:

- The gate has not actually run on this PR. Both workflow runs for head `2c1fd1a` (*CI for Develop&Main*, *Mutation tests*) concluded `action_required` — they need a maintainer to approve workflows for this fork before CI executes. My review used a narrow local run in an isolated worktree as its evidence only; it does not substitute for CI.
- A maintainer applied the labels you proposed, with two deliberate deviations from your metadata-intent comment, both explained in the label-rationale comment: `risk-medium` instead of `risk-low`, and `changes-requested` instead of `review` because of the review outcome. Your `bug`, `priority-medium` and `skip-qa` proposals were adopted as argued.

pkarw:

One Major finding (`email` in a `{ persistent: true }` payload writes an encrypted-at-rest field into the durable events queue, on four auth paths, with no subscriber consuming it) and one Minor finding (the new `changedBy: 'system'` comments document CLI/provisioning coverage that does not exist, because those paths write `passwordHash` directly and bypass `auth.users.update`), plus two nits. No blockers, no merge conflicts, no failing checks. The approach and the placement of all four emit sites are sound — both findings are contained fixes.

CI follow-up: none scheduled. The workflow runs for this head are `action_required`, so nothing will settle until a maintainer approves workflows for this fork; there is no pending check for this agent to watch.

patzick:
🤖 **Claimed by `om-auto-review-pr`** (@patzick session).

Reviewing the head commit `99918cf` against `develop`. Validation is **GitHub-checks-first** per this repo's skill extension — no local gate run.

**First finding, already actionable:** the `audit` check fails on `[high] deepmerge-ts <8.0.0 — GHSA-ggr8-5vv4-36mx`, which is **not** this PR's doing. The branch is **126 commits behind `develop`**, and `develop` already carries the `"deepmerge-ts": "8.0.1"` entry in `package.json` `resolutions` that clears that advisory.

Because `audit` fails, `test`, `ephemeral-integration`, `merge-coverage` and `docker-build` all report `skipping` — so **there is currently no test signal on this PR at all**, not a passing one. Updating the branch from `develop` now so CI can actually run the gate; the review of the code itself follows in the same pass.

patzick (approved):
# 🔍 Code Review: fix(auth): emit the declared logout and password lifecycle events

**Verdict: ✅ Approved** — no blocker/major/minor findings. Only nits remain, all optional.

**Validation source:** GitHub checks (head `b61c532b9`). No local gate run, per this repo's GitHub-checks-first skill extension.

---

## 🚦 What I had to fix before this PR could be judged at all

The PR was showing `audit: fail` with `test`, `ephemeral-integration`, `merge-coverage` and `docker-build` all `skipping`. That is not "tests n/a" — it means **this PR had never produced a test signal**.

Root cause was not this PR. The branch was **126 commits behind `develop`**, so it predated the `"deepmerge-ts": "8.0.1"` entry in `package.json` `resolutions` that clears `GHSA-ggr8-5vv4-36mx`. `audit` failed on that stale advisory and gated the rest of the pipeline.

I updated the branch from `develop` (merge commit `b61c532b9`) and re-approved the fork workflow runs. Result: `audit` now reports `skipping` (no dependency delta against the refreshed base), `prepare` passes, and the test gate actually ran.

## 🧪 Test gate — auth is fully green

All four suites this PR ships or extends **pass**:

```
PASS src/modules/auth/api/__tests__/logout.test.ts
PASS src/modules/auth/api/__tests__/reset.test.ts
PASS src/modules/auth/api/__tests__/reset-confirm.test.ts
PASS src/modules/auth/commands/__tests__/users.password-changed.test.ts
```

`lint`, `ds-lint`, `scope`, `audit-scope`, `prepare` all pass.

### ⚠️ `test` is red — and it is not this PR

```
Test Suites: 1 failed, 1291 passed, 1292 total
FAIL src/modules/data_sync/lib/__tests__/sync-engine-heartbeat.test.ts
  TypeError: (0 , adapter_registry_1.resolveProviderKey) is not a function
    at Object.runImport (src/modules/data_sync/lib/sync-engine.ts:479:45)
```

**`develop` is already red on exactly this** — run [32157461842](https://github.com/open-mercato/open-mercato/actions/runs/32157461842) at `00c90fecf`, same suite, same `TypeError`, `1 failed, 1289 passed`. This PR inherited it through the merge; it touches nothing outside `packages/core/src/modules/auth/`.

The cause, for whoever picks it up: `resolveProviderKey` *is* exported (`data_sync/lib/adapter-registry.ts:30`), but the suite stubs the module with a factory that omits it —

```ts
jest.mock('../adapter-registry', () => ({
  getDataSyncAdapter: (...args: unknown[]) => mockGetDataSyncAdapter(...args),
}))
```

— while #5250 added a `resolveProviderKey` call to `sync-engine.ts`. The mock needs the new export. **That is a `develop` fix, not a fix for this branch.** This PR cannot merge until it lands, but nothing here should change for it.

## ✅ Prior review findings — all genuinely resolved

I verified each against the source rather than the PR description.

| Prior finding | Status |
|---|---|
| **Major** — `email` inside a `{ persistent: true }` payload | **Resolved.** Dropped from all four payloads; `logout.ts` no longer extracts the claim at all. Each suite asserts the payload has no `email` key *and* that the address is absent from the serialized form. |
| **Major** — `changedBy: 'system'` comments documented coverage that did not exist | **Resolved** via the "make the comment true" option. Verified independently: `cli.ts:780` and `setup-app.ts:419` assign `passwordHash` directly and never reach `auth.users.update`. The comment now states that boundary plainly. |
| **Nit** — `String(user.email)` could stringify `undefined` | Moot, field removed. |
| **Nit** — `sessionRevoked` semantics | Documented on the `auth.logout` declaration. |

## 🔬 Integration points I checked rather than took on faith

- `emit(event, payload, options?: EmitOptions)` — `packages/shared/src/modules/events/factory.ts:216`. The `{ persistent: true }` third argument is real and reaches `eventBus.emit`.
- `identifiers` (`commands/users.ts:731`) is built **after** `withAtomicFlush` commits, so the payload carries post-update tenant/org — correct for a destination-changing update.
- **Self-service changes still emit.** The emit moved out of `api/profile/route.ts` in commit 2; profile `PUT` dispatches `auth.users.update` through the command bus (`api/profile/route.ts:173`), so both password-write callers converge on the single emit site as claimed.
- **Reset path is single-source.** `authService.confirmPasswordReset` (`services/authService.ts:163`) mutates `passwordHash` directly without the command, so `.completed` cannot double-fire with `.changed`.
- The `passwordHash` writes at `users.ts:418/431` belong to **`auth.users.create`'s redo**, not the update command — so there is no redo path that silently re-applies a password change.
- `reset.ts`: everything after the emit is `try`/`catch`-wrapped and the route always returns 200, so there is no path where the event fires but the caller sees a failure. The "a token exists" semantics hold.

## 💅 Nits (non-blocking, take or leave)

**1. `sessionRevoked: false` conflates two different states** — `api/logout.ts:68`. The revocation block only runs `if (sessToken || claims.sessionId)`. A legacy token with no `sid` claim and no `session_token` cookie skips it entirely, so the payload reports `sessionRevoked: false` although revocation was never *attempted*. The `events.ts` comment says the flag means revocation "ran without throwing", which is not true in that branch. Either a `null` third state or one more clause in the comment would close it. The conservative direction is already right, so this is documentation, not behavior.

**2. `{ persistent: true }` enqueues unconditionally.** `packages/events/src/bus.ts:445` is `if (options?.persistent) { await q.enqueue(...) }` with no subscriber check — so every logout platform-wide writes a durable job that nothing consumes today. Worth knowing: `POST /api/auth/logout` carries no rate limit, unlike `login` (5/60s), and it is the highest-frequency of the four. Bounded and cheap in a deployment with a running events worker; it accumulates where `AUTO_SPAWN_WORKERS=false` and no worker is up. Flagging for awareness only — the previous reviewer explicitly offered keeping `persistent` as the acceptable resolution, and durable delivery is the point of these events.

**3. `at` is new to these four payloads** but absent from the pre-existing `auth.login.success` / `auth.login.failed`. Harmless and additive; noting only so the inconsistency is a decision rather than an accident.

## 💥 Breaking changes

Additive only, and I agree with the PR's own analysis. No event id added/renamed/removed, no route, schema, DI key, ACL feature or notification id touched. The one widened helper (`extractSessionIdFromAuthToken` → `extractAuthTokenClaims`) is module-private in `logout.ts` and not a contract surface under `BACKWARD_COMPATIBILITY.md`. The sole behavioral delta is that four already-declared events begin reaching the bus, and no subscriber for them exists in the repo today.

---

### Merge readiness

Code is approved. Merge is blocked on one thing that is **not** this branch's to fix: `develop`'s `sync-engine-heartbeat.test.ts` breakage. Once that lands on `develop`, re-sync this branch and the gate should go green.

`skip-qa` is correctly applied — no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no DB or API surface change, and automated tests for the changed behavior ship in the same PR.

<sub>🤖 Reviewed by `om-auto-review-pr` · validation source: GitHub checks · head `b61c532b9`</sub>

patzick:
**Label change:** `changes-requested` → `review`; `in-progress` released.

Rationale: the review verdict is **approved** — the two Major findings from the previous round are verified resolved in the source (not just in the description), and only three optional nits remain. Under this repo's stricter rule, changes are requested on Medium/minor and above; nothing here reaches that bar.

Untouched, deliberately: `skip-qa` (correct — no UI-rendering file, no DB or API surface change, automated tests for the changed behavior ship in the same PR), `priority-medium`, `risk-medium`, `bug`, `partner-request`.

⚠️ **Not mergeable yet, for a reason outside this branch.** `test` is red on `sync-engine-heartbeat.test.ts` (`resolveProviderKey is not a function`), which `develop` is [already failing on](https://github.com/open-mercato/open-mercato/actions/runs/32157461842) identically. That needs a fix on `develop`; this branch then needs a re-sync. Details and the likely one-line cause are in the review above.

PR #5365, opened 2026-08-18
fix(auth): log ACL changes to the action log

Frshy:
## Summary

Changing permissions left no audit trail. `PUT /api/auth/roles/acl` and `PUT /api/auth/users/acl` wrote straight to the ORM and never reached the command bus, so granting a user super-admin, widening a role's feature list, or clearing a per-user override produced no `ActionLog` row at all.

Every other auth mutation (`auth.roles.*`, `auth.users.*`) is audited. ACL editing — arguably the most audit-relevant change in the system — was the one class with no record. And because `RoleAcl` / `UserAcl` carry only current state plus `updated_at`, the prior grant set is overwritten in place, so it cannot be reconstructed after the fact. `GET` on both routes already calls `logCrudAccess`, so reads were recorded while writes were not.

After this change, each ACL save appears in the audit log with the actor, the before/after grant sets, and a diff of what actually changed.

## Changes

- **New commands** in `packages/core/src/modules/auth/commands/acl.ts`:

  | Command id | resourceKind | Action label |
  |---|---|---|
  | `auth.role-acl.update` | `auth.role_acl` | Change role permissions |
  | `auth.user-acl.update` | `auth.user_acl` | Change user permissions |

  `resourceKind` reuses the identifiers the routes already use for `logCrudAccess` and optimistic locking, so access-log entries, action-log entries and record locks agree on one name per resource. Both are auto-discovered by `yarn generate`.

- **Both ACL routes dispatch through the bus** instead of writing to the ORM directly (`api/roles/acl/route.ts`, `api/users/acl/route.ts`).
- **Two i18n keys** for the action labels, across all five locales.
- **Spec** `.ai/specs/2026-08-07-auth-acl-change-audit-logging.md` plus its `.ai/specs/README.md` index row.
- **Tests** — five new unit suites and one integration spec (listed under Testing).

**Command boundary — deliberately narrow.** The command owns only the transactional write and the RBAC cache invalidation. Every guard, the tenant resolution, the optimistic-lock check and the non-superadmin sanitization stay in the routes, so authorization behavior is untouched and the existing route tests still cover it. The route hands the command post-guard, post-sanitization values, so persisted state and logged state cannot diverge.

**Log-only by design — `isUndoable: false`, no `undo`/`redo`.** An ACL change is reversible, so this is not an "irreversible by design" exemption. The reason is separation of duties:

- The undo and redo endpoints are gated on `audit_logs.undo_self` / `audit_logs.undo_tenant`, **not** on `auth.acl.manage`.
- `audit_logs.undo_self` is a default `employee` grant, and `audit_logs.undo_tenant` reaches every `admin` through the `audit_logs.*` wildcard.
- An undoable ACL command would therefore let a caller holding only an audit-log feature reverse or replay someone else's permission change — including re-granting a super-admin flag that was just revoked — without ever holding the feature that authorizes editing permissions.

Reversal stays where it belongs: re-submitting the ACL form, gated on `auth.acl.manage`, validated by `assertActorCanGrantAcl`, and itself audited. This follows the documented `Undo Policy` header-block convention from `wms/commands/inventory-actions.ts`; `commands/acl.ts` carries the equivalent block.

**Two subtleties worth reviewing:**

1. Snapshots sort `features` / `organizations` with an explicit codepoint comparator. Grants are sets, but `features_json` preserves the client's insertion order, and the bus derives `changes` with an order-sensitive deep-equality check — without sorting, re-saving an unmodified form would report a phantom `features` change every time. Codepoint rather than `localeCompare` because these are persisted records whose ordering must not shift with the runtime locale.
2. The after-snapshot is re-read from the committed row rather than echoed from the request, so a grant the route sanitized away can never be logged as if it had been applied. On the user-ACL clear path the row is deleted, and the empty snapshot is the correct post-state.

**Tenant resolution.** A tenant-less global admin (`users.tenant_id` is nullable, so `auth.tenantId` is legitimately `null` after login) was reaching an unscoped `UserAcl` lookup. Both the `PUT` and the `GET` handler now resolve the tenant from the **target user**, identically, so read and write agree on which row they are talking about. A cross-tenant entry also nulls `organization_id` rather than persisting a `(tenant B, organization from tenant A)` pair that no reader can match.

## Compatibility

No schema change, no wire-format change. Request schemas, response bodies (`{ ok, sanitized }`) and every status code are byte-identical; `openApi` is unchanged. Additive only: two command ids, two i18n keys across five locales. New rows carry `undoToken: null`, so they are absent from `?undoableOnly=true` listings and render no Undo action.

Nothing under `packages/enterprise/` is touched. The record-lock guard on `auth.role_acl` still runs in the route before dispatch, so `TC-LOCK-AUTH-acl` is unaffected.

## Specification

- [ ] Yes
- [x] No (created a new spec)

**Spec file path:**

`.ai/specs/2026-08-07-auth-acl-change-audit-logging.md`

## Testing

Full validation gate run locally on the rebased head (`3785543e`), against current `develop`:

| Command | Result |
|---|---|
| `yarn build:packages` → `yarn generate` → `yarn build:packages` | pass (generate produces no uncommitted drift) |
| `yarn i18n:check-sync` | pass |
| `yarn typecheck` | pass |
| `yarn test` | **19,647 tests / 2,205 suites, 0 failures** |
| `yarn build:app` | pass |
| `yarn lint` | 0 errors (10 warnings, all pre-existing in `@open-mercato/app`, none in changed files) |

`@open-mercato/core` alone accounts for 9,787 tests across 1,251 suites.

New coverage in this PR:

- `commands/__tests__/acl.test.ts` — both write paths, transaction commit and rollback, cache-invalidation ordering (after commit, never inside the flush), audit metadata, the canonical-ordering guarantee, and the log-only policy.
- `commands/__tests__/acl.persisted-scope.test.ts` — the persisted tenant/organization scope of each written row.
- `api/users/acl/__tests__/tenant-resolution.test.ts` — tenant-less actor, cross-tenant target, and read/write agreement on the resolved tenant.
- `api/users/acl/__tests__/sanitized-request.test.ts` — the after-snapshot reflects sanitized state, not the raw request.
- `api/roles/acl/__tests__/log-scope.test.ts` and the extended `optimistic-lock.test.ts` — including that a rejected 409 write never reaches the bus.
- `__integration__/TC-AUTH-058-acl-audit-log.spec.ts` — role grant, user grant, and the user-clear path.

`TC-AUTH-058` was executed on 2026-08-12 via `yarn test:integration:ephemeral --filter TC-AUTH-058` (2/2 passed) against an earlier commit of this branch; it has **not** been re-executed since the rebase onto current `develop`, so CI should be the authority on it.

## Checklist

- [x] I added or updated integration tests (`__integration__/TC-AUTH-058-acl-audit-log.spec.ts`, following the `packages/*/src/modules/*/__integration__/` convention).
- [ ] QA routing set.

The three label boxes are unticked because this PR is opened from a fork by an account without `triage` on this repository — it cannot apply labels. The intended values are in a separate comment below, for a maintainer to apply.

### Design System Compliance

- [x] No hardcoded status colors — the diff contains no `.tsx`, CSS or component file.
- [x] No arbitrary text sizes — same reason.
- [ ] Empty state handled for list/data pages — N/A, no UI surface in this diff.
- [ ] Loading state handled for async pages — N/A.
- [ ] `aria-label` on all icon-only buttons — N/A.
- [ ] Uses existing DS components — N/A.

## Linked issues

None — found and fixed while auditing which auth mutations reach the command bus.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5365"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789631033&installation_model_id=432243&pr_number=5365&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5365&signature=55feae51f96d1d6f50fbc0c225bf2121d1a8c39ea3c9b07b602148b284bd082f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Frshy:
**Metadata intent** — this PR is opened from a fork by an account without `triage` on this repository, so it cannot set labels, assignee or reviewer itself. Requesting a maintainer apply the following; every name below was validated against this repo's label list.

| Label | Rationale |
|---|---|
| `review` | Ready non-draft PR, not in any other pipeline state. |
| `security` | Adds the missing audit trail for permission changes and closes an unscoped `UserAcl` lookup reachable by a tenant-less global admin. |
| `priority-high` | Permission changes were leaving no record at all, and the affected rows are overwritten in place, so the history is unrecoverable after the fact — not reconstructable later. |
| `risk-high` | Touches the authorization write path on both ACL routes. The guards themselves are unchanged and the command boundary is deliberately narrow, but the blast radius of a regression here is the RBAC system. |
| `skip-qa` | Proposed under the automated-verification exemption — see the reasoning below. Maintainer's call. |

**On the QA routing.** The exemption in `AGENTS.md` → PR Workflow and in the PR template asks for four conditions; all four hold here:

1. **No UI-rendering file** — the diff is 17 files: two API routes, one command module, six test files, five locale JSONs, and two spec/docs files. No `.tsx`, nothing under `packages/ui/src/` or `**/components/**`.
2. **Database structure unchanged** — no entity change, no migration, no snapshot change.
3. **API surface unchanged** — request schemas, response bodies (`{ ok, sanitized }`) and status codes are byte-identical; `openApi` is untouched. No `BACKWARD_COMPATIBILITY.md` contract is affected (additive only: two command ids, two i18n keys).
4. **Automated tests for the changed behavior ship in this PR** — five new unit suites plus `TC-AUTH-058-acl-audit-log.spec.ts`.

If you would rather route it through manual QA, `needs-qa` applies instead and the PR stays merge-blocked until QA adds `qa-approved`. Please do not combine the two.

**Assignee:** the PR author.

**Reviewer:** whoever owns auth / RBAC for this repo — the change sits on the authorization write path and the interesting review questions are the log-only `isUndoable: false` policy and the tenant resolution on the user-ACL route, both explained in the description.

**Prior review.** This branch was reviewed on the fork before being sent here ([fullstackhouse#68](https://github.com/fullstackhouse/open-mercato/pull/68)): 21 review threads, all resolved, and three approvals — the most recent of which post-dates the final commit. That is context, not a substitute for review here.

patzick:
🤖 **Claimed by \`om-auto-review-pr\`** (@patzick session).

Starting an automated review of this PR. Assigned myself and applied \`in-progress\`; both are released when the review finishes.

Validation source: **GitHub checks** (no local gate runs).

patzick (changes requested):

**Verdict: `changes_requested`** — 1 blocker (cross-tenant guard regression).

**Validation source:** GitHub checks. No local gate commands were run.

---

### Context first: CI had no usable signal

At review time this branch was **66 commits behind `develop`**, and that alone broke CI:

- `audit-scope` computes changed files as `git diff origin/develop --name-only`. Because the branch was stale, `yarn.lock` / `package.json` showed up in that diff as *reverse* changes, so `audit` ran even though this PR touches no dependency file.
- `audit` then scanned the branch's **stale lockfile**, which still pins `deepmerge-ts@7.1.5`. `develop` already carries `8.0.1` (GHSA-ggr8-5vv4-36mx). The `image-size` advisory in the same output is likewise already allowlisted on `develop`.
- `test`, `ephemeral-integration` and `merge-coverage` all `needs: audit`, so **every one of them skipped**. There was no test, typecheck or build evidence for this head at all.

None of that is caused by the diff. I've updated the branch from `develop` (clean merge, no conflicts) so the real gate can run; results will follow in a separate comment.

---

### 🔴 Blocker — the cross-tenant guard on `users/acl` becomes a tautology for a tenant-less actor

`packages/core/src/modules/auth/api/users/acl/route.ts`

`assertActorCanAccessUserTarget` treats its `tenantId` argument as the **actor's** tenant and rejects a target whose tenant differs:

```ts
const actorTenantId = normalizeNullableString(input.tenantId)
const targetTenantId = normalizeNullableString(target.tenantId)
if (!targetTenantId || targetTenantId !== actorTenantId) {
  throw new CrudHttpError(404, { error: 'User not found' })
}
```

Its own comment states the invariant it relies on: *"Every wired call site is itself tenant-scoped — the ACL/consents reads filter by `auth.tenantId`."*

Both handlers now pass the **resolved** `tenantId` instead of `auth.tenantId ?? null`. When `auth.tenantId` is `null`, `resolveAclTenantId` derives that value **from the target user itself**, so the guard compares the target's tenant against the target's tenant and can never fail. The same substituted value reaches `assertActorCanModifySuperAdminUserTarget` and `assertActorCanGrantAcl`, so the actor's own grants are evaluated in the *target's* tenant scope — the exact thing the new comment above `actorAcl` says must not happen.

**Failure scenario.** A user with `users.tenant_id IS NULL` who holds `auth.acl.manage` but is **not** a super admin:

1. `GET /api/auth/users/acl?userId=<user in tenant B>` → `resolveAclTenantId` short-circuits on neither an explicit tenant nor `auth.tenantId`, so it runs an **unscoped, decrypting** `findOneWithDecryption(User, { id })` (`tenantId: null`) *before* any access guard, and returns tenant B. Guards pass. Tenant B's `UserAcl` row is returned.
2. `PUT` the same way → the command bus writes an ACL override into tenant B.

Before this PR, step 1 threw `404 User not found` from the guard, and the write path could not match a row anyway (`tenant_id IS NULL` against a `NOT NULL` column). So this is a new capability, not a preserved one.

**Why the tests don't catch it.** `tenant-resolution.test.ts` mocks the whole `grantChecks` module to no-ops, and `beforeEach` sets `loadAcl → { isSuperAdmin: true }`. Every tenant-less-actor case therefore runs as a super admin. The only `isSuperAdmin: false` case (`refuses an explicit foreign tenant for a non-super-admin`) gives the actor a **non-null** `ACTOR_TENANT_ID`, which is precisely the combination that is safe. `tenantId: null` + `isSuperAdmin: false` is untested.

**Suggested fix.** The roles route already has the check this one is missing:

```ts
// api/roles/acl/route.ts
if (!isSuperAdmin && targetTenantId !== authTenantId) {
  return NextResponse.json({ error: 'Forbidden' }, { status: 403 })
}
```

Mirror it in `users/acl` on **both** verbs, after `resolveAclTenantId`:

```ts
if (!actorIsSuperAdmin && tenantId !== (auth.tenantId ?? null)) {
  return NextResponse.json({ error: 'Forbidden' }, { status: 403 })
}
```

A tenant-less **super admin** — the case the PR description is actually about — stays exempt, so the intended fix is preserved. Please add a unit case pinning `tenantId: null` + `isSuperAdmin: false` against a foreign target, with `grantChecks` **not** mocked away (or at least with `assertActorCanAccessUserTarget` asserted to receive the actor's tenant, not the resolved one).

---

### 🟡 Medium — the guard-argument change is a silent semantic switch

Independent of the blocker: swapping `auth.tenantId ?? null` for a route-local `tenantId` at five call sites changes what those guards *mean*, and nothing in the diff or the spec records that `grantChecks` documents the argument as the actor's scope. Whatever the fix above lands as, please either keep passing the actor's tenant to the guards (and use the resolved tenant only for the `UserAcl` lookup and the command input), or update the `assertActorCanAccessUserTarget` doc comment so the next reader isn't working from a stale invariant.

---

### 🔵 Low — nits

- The `resolveAclTenantId` doc says *"the ACL editor sends the one selected in the form"*, but no `.tsx` changed in this PR and nothing currently sends `tenantId` on this route. The parameter is fine as an additive API surface (the roles route already has it), but the comment describes a caller that doesn't exist yet.
- Scope: the title is *"log ACL changes to the action log"*, and the audit-logging half of the diff is clean, well-argued work — the canonical grant ordering, the re-read-after-commit `captureAfter`, the `sanitizedRequest` escape from the no-op guard, and the non-propagating cache invalidation are all correct and well justified. The tenant-resolution change is a separate behavior fix riding along; it's what carries all the risk here, and it would have been easier to review (and to bisect) on its own.

---

### ✅ What checked out

- i18n: both keys present in all five locales (`de`/`en`/`es`/`ko`/`pl`), alphabetically placed.
- Backward compatibility: no schema change, response bodies and status codes unchanged, two additive command ids, additive optional request params. Nothing under `packages/enterprise/`.
- The `isUndoable: false` / no-`undo` policy and its separation-of-duties rationale are sound — an undo gated on `audit_logs.*` really would let a caller reverse a permission change without holding `auth.acl.manage`.
- Optimistic locking stays in the route ahead of dispatch, so `TC-LOCK-AUTH-acl` semantics are preserved.
- `resolveOrganizationId` + the route-side organization stripping correctly avoid the unmatched `(tenant B, org from tenant A)` pair.

---

<sub>Labels and the CI result follow in separate comments. `in-progress` is released when this review closes.</sub>

patzick:
### 🏷️ Labels applied — rationale

| Label | Why |
|---|---|
| `changes-requested` | Review verdict above: one blocker. Replaces `review` as the pipeline state. |
| `bug` | The PR itself fixes two defects: an unaudited ACL write path and a broken tenant lookup on `users/acl`. |
| `security` | The diff sits directly on the authorization path (`grantChecks`, super-admin grants, tenant scoping), and the blocker is a cross-tenant guard regression. |
| `priority-medium` | A missing audit trail on permission changes is a real compliance gap, but not an active incident; nothing is currently broken for users in production. |
| `risk-high` | Blast radius, not urgency: it rewrites the tenant-resolution and guard arguments for **all** ACL reads and writes, and both `auth.role_acl` and `auth.user_acl` mutations now flow through a new command path. A regression here silently widens permissions. Risk is set above priority deliberately. |
| `needs-qa` | The automated-verification exemption does **not** apply. It requires an unchanged API surface, and this PR adds an optional `tenantId` parameter to both `GET` and `PUT /api/auth/users/acl` (`getSchema`, `putSchema`, `openApi`) and changes the tenant scope those handlers resolve. The no-UI-file half of the exemption is satisfied (no `.tsx`, nothing under `packages/ui/`), but all conditions must hold. |

`skip-qa` was **not** applied and must not be combined with `needs-qa`. Per the QA-approval merge gate this PR cannot merge without `qa-approved`, independent of check status.

The label set is unchanged in `partner-request`, which I left as the author found it.

patzick:
### ✅ CI result after the branch update — all checks green

Head `208261e1` (this branch merged with current `develop`). The gate that was previously unreachable now runs and passes in full:

| Check | Result |
|---|---|
| `prepare` | ✅ 4m25s — `build:packages` → `generate` → `build:packages` → `build:app` |
| `test` | ✅ 7m10s |
| `ephemeral-integration (none)` | ✅ 9m34s — **122 passed, 0 failed, 0 flaky**, 7 skipped |
| `lint` | ✅ |
| `ds-lint` | ✅ |
| `audit-scope` | ✅ |
| `audit` | ⏭️ skipped — correct: no dependency file in the diff vs `develop` |
| `scope`, `license/cla` | ✅ |

Step detail inside `test`: `i18n:check-sync` ✅, `i18n:check-usage` ✅, `typecheck` ✅, `Test` ✅, `Test scripts` ✅, DS token snapshot drift ✅, **create-app template parity ✅**, repo-wide audit guards ✅.

So all eight `validation.commands` entries have now executed and passed on a head that includes current `develop`. `yarn generate` produces no uncommitted drift, which confirms the two new commands auto-discover cleanly.

The new integration spec ran and both cases passed:

```
✓ 78 TC-AUTH-058-acl-audit-log.spec.ts:71  › role ACL update writes an action-log entry
                                             with before/after and no undo token (137ms)
✓ 79 TC-AUTH-058-acl-audit-log.spec.ts:115 › user ACL grant and clear are both audited (3.5s)
```

That closes the open item in the PR description — `TC-AUTH-058` had not been re-executed since the rebase, and it now has a CI run on a current head.

`ephemeral-integration` resolving to the single `(none)` runner rather than the 15-shard matrix is expected, not a gap: the diff touches only `packages/core/src/modules/auth/`, so the affected-only path applies and the auth module's integration tests all ran.

---

### 🔵 One more Low finding, surfaced by the integration log

**`TC-AUTH-058` is already taken.** `develop` carries `packages/core/src/modules/auth/__integration__/TC-AUTH-058-user-destination-descendant.spec.ts`, and both files now execute side by side under the same test-case id:

```
✓ 79 TC-AUTH-058-acl-audit-log.spec.ts               › user ACL grant and clear are both audited
✓ 80 TC-AUTH-058-user-destination-descendant.spec.ts › descendant user destination scope
```

Nothing fails, but the id stops being a unique handle for QA traceability. The highest allocated id in this module is `TC-AUTH-062`, so please rename to **`TC-AUTH-063-acl-audit-log.spec.ts`** and update the `describe` title plus the references in the PR description and `.ai/specs/2026-08-07-auth-acl-change-audit-logging.md`.

---

### Where this leaves the PR

**Verdict stands: `changes_requested`.** Green CI does not clear the blocker — that finding is a guard-semantics regression which the unit suite structurally cannot catch, because `tenant-resolution.test.ts` mocks the entire `grantChecks` module to no-ops and sets `isSuperAdmin: true` in `beforeEach`. A passing `yarn test` is consistent with the vulnerability being present.

To move to approval:

1. **Blocker** — add the `!actorIsSuperAdmin && tenantId !== (auth.tenantId ?? null)` → `403` check to both verbs of `users/acl/route.ts`, mirroring the roles route, plus a regression test with `tenantId: null` + `isSuperAdmin: false` against a foreign target.
2. **Medium** — reconcile the guard-argument semantics with the `assertActorCanAccessUserTarget` doc comment.
3. **Low** — rename the integration spec off the duplicate id; drop or correct the "the ACL editor sends the one selected in the form" comment.

I've released `in-progress`. `needs-qa` stays until a QA reviewer signs off — this PR cannot merge without `qa-approved`, independent of the green checks above. Re-request review once the blocker is addressed.

<sub>Automated review by `om-auto-review-pr`. Validation source: GitHub checks only; no local gate commands were run.</sub>

Frshy:
@patzick thanks — the blocker is real and it's fixed in `de39c2129`. You also correctly diagnosed why CI had no signal; the branch update cleared that.

### The blocker

Confirmed exactly as described: with `auth.tenantId === null` and `isSuperAdmin: false`, `resolveAclTenantId` derived the scope from the target user, and that same value reached `assertActorCanAccessUserTarget`, which compares it against the target's tenant. The comparison could never fail, so both verbs became reachable for any tenant.

I did **not** take the suggested `403` mirroring the roles route, because the fix closes the divergence at its source and would leave that check unreachable. Two changes instead:

**1. The target-derived fallback is now super-admin-only** (`resolveAclTenantId`, `route.ts:79`). A tenant-less non-super-admin resolves to `null`, which reads as "no override" and refuses to write with the existing `400 Tenant required`. This also removes the second half of the problem you flagged: the unscoped, decrypting `findOneWithDecryption(User, { id })` no longer runs ahead of the guards for anyone but a super admin.

**2. Every guard is answered in the actor's scope.** All five call sites plus `resolveOrganizationScopeForRequest` now take a route-local `actorTenantId = auth.tenantId ?? null`. The resolved scope stays on the `UserAcl` lookup, the command input and `logCrudAccess` — which is what it was introduced for.

With (1) in place, a non-super-admin's resolved scope is always the actor's own, so the two halves agree by construction and `tenantId !== (auth.tenantId ?? null)` cannot hold for a non-super-admin. That's why an explicit `403` would be dead code rather than defense in depth. A tenant-less **super admin** — the case the PR is about — is unaffected: every guard early-returns on `isSuperAdmin`.

### The medium — guard-argument semantics

Taken, via your first option: the guards keep receiving the actor's tenant. `assertActorCanAccessUserTarget` now also carries a doc comment stating the invariant explicitly, so the next caller that resolves a record scope separately doesn't repeat this:

> `input.tenantId` is the ACTOR's scope, not the target's: the check below is a comparison between the two, so passing a tenant derived from the target turns it into a tautology.

### Tests

You were right that the route suite couldn't catch this — `grantChecks` is mocked to no-ops there and `beforeEach` sets `isSuperAdmin: true`. Coverage added on both levels:

- **`lib/__tests__/targetAccessChecks.test.ts`** — the real guard, no `grantChecks` mock: a tenant-less non-super-admin actor against a target in another tenant must get `404`.
- **`api/users/acl/__tests__/tenant-resolution.test.ts`** — three route cases: `PUT` refuses with `400 Tenant required` and never reads the target user; `GET` returns `hasCustomAcl: false` and the guard receives `tenantId: null`, not the target's; and a same-tenant edit pins that the guard receives `ACTOR_TENANT_ID`.

The two refusal cases were verified to fail with the route change reverted (`2 failed, 10 passed`); the third is a regression pin that passes either way by design.

### Nits

Both fixed. The `resolveAclTenantId` doc no longer describes an ACL-editor caller that doesn't exist — the parameter is documented as additive, matching the roles route. The stale `PUT` comment ("authorization is decided in the same tenant the write lands in") is gone, since that is precisely what is no longer true.

On the scope point: agreed, the tenant-resolution fix riding along with the audit-logging work is what made this hard to review. Noted for next time.

### Validation

Local run (no Docker containers up): `packages/core` jest — 1291 suites / 10128 tests, all green; `yarn typecheck` green; `yarn lint` 0 errors. `yarn generate` was needed first — the `develop` merge left `entities.ids.generated.ts` without the `devices` / `push_notifications` entries, unrelated to this diff.

Spec `.ai/specs/2026-08-07-auth-acl-change-audit-logging.md` updated: the API Contracts section now records the super-admin gate and the actor-scope rule for the guards, with a 2026-08-19 changelog entry covering this round.

Ready for another look — could you drop `changes-requested` if it holds up?

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-21T00:12:42Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):

**Verdict: `approved`** — 0 blockers, 0 majors, 0 mediums, 2 lows.

**Validation source:** GitHub checks. No local gate commands were run (this repo's `om-auto-review-pr` extension is checks-first).

Re-review context: this is the first review from this reviewer. @patzick's earlier `changes_requested` (at `208261e`) is superseded by `de39c21` — I re-verified its blocker from scratch rather than taking the commit message's word for it.

---

### ✅ The previous blocker is genuinely fixed

The cross-tenant guard tautology on `api/users/acl/route.ts` is gone, and the fix is the right shape rather than a patch over the symptom:

- Both handlers now hold `actorTenantId = auth.tenantId ?? null` separately from the resolved record scope, and **every** guard call site (`assertActorCanModifySuperAdminUserTarget`, `assertActorCanAccessUserTarget`, `resolveOrganizationScopeForRequest`, `assertActorCanGrantAcl`) receives the actor's tenant. The guard can fail again.
- `resolveAclTenantId` now refuses to derive scope from the target for a non-super-admin: `if (!args.actorIsSuperAdmin) return { tenantId: null }` sits **before** the `findOneWithDecryption(User, …)` call, so the unscoped decrypting read the previous review flagged as running ahead of every guard is now unreachable for that actor class. `GET` degrades to "no override", `PUT` returns `400 Tenant required`.
- `grantChecks.ts` gained the doc comment stating the invariant (`input.tenantId` is the **actor's** scope), which closes the Medium about a stale invariant.

The test gap is closed the way it was asked for:

| Case | Test |
|---|---|
| `tenantId: null` + `isSuperAdmin: false`, foreign target, `PUT` | `tenant-resolution.test.ts` → *refuses a tenant-less non-super-admin instead of adopting the target tenant* — asserts `400`, no bus dispatch, **and** that the `User` read never happened |
| same, `GET` | *reads no foreign override for a tenant-less non-super-admin* — asserts the guard received `tenantId: null` |
| guards get the actor's tenant, not the resolved one | *hands the guards the actor tenant on a same-tenant edit* |
| the guard itself, unmocked | `lib/__tests__/targetAccessChecks.test.ts` → *hides any tenanted target from a tenant-less non-superadmin actor as 404* |

The last row is the important one — it pins the behaviour in the module the route suite mocks away, so the tautology cannot silently return.

---

### ✅ What else checked out

**Command boundary.** `commands/acl.ts` owns only the transactional write and cache invalidation; every guard, the tenant resolution, the optimistic-lock check and the sanitization stay in the routes. Authorization behaviour is untouched — I diffed all five guard call sites against `develop` and the only change is the argument rename.

**Audit correctness.** Three subtleties are each right and each explained:
- `captureAfter` re-reads the committed row instead of echoing the request, so a sanitized-away grant can never be logged as applied.
- `toSortedList` + `compareIdentifiers` make the bus's order-sensitive `snapshotsEqual` behave as a set comparison, so re-saving an unmodified form doesn't manufacture a phantom `features` change. Codepoint over `localeCompare` is correct for persisted records, and the explicit comparator satisfies `explicit-sort-comparators.test.ts`.
- `snapshotWhenAbsent` distinguishes "the clear path removed the row" (`EMPTY`, user command only) from "the re-read failed" (`null` → logged as changed). `organizations: null` (all orgs) stays distinct from `[]` (none).

**Log scope.** `resolveOrganizationId` returning `null` on a foreign tenant is right — `ActionLogService` filters `organization_id` with strict equality, so a (tenant B, org from tenant A) pair would make the cross-tenant permission change permanently invisible. Both routes additionally strip `orgId` from the command context, so the `??` chain in the bus can't restore it. Belt and braces, correctly.

**Undo policy.** `isUndoable: false` with no `undo`/`redo` is the right call and the rationale holds: `audit_logs.undo_self` is a default `employee` grant and `audit_logs.undo_tenant` reaches every `admin` via `audit_logs.*`, so an undoable ACL command would let a caller reverse a permission change without holding `auth.acl.manage`. Follows the `wms/commands/inventory-actions.ts` header-block convention.

**Optimistic locking.** Both `updated_at` columns carry `onUpdate`, and `enforceCommandOptimisticLockWithGuards` still runs in the route ahead of dispatch. I checked the `resourceKind`/`resourceId` split (action log keys on `userId`/`roleId`, the lock on `acl.id`): the lock uses `resourceKind` only as an allow-list key and `resourceId` only in the conflict payload — it does not join against action logs, so nothing breaks. `TC-LOCK-AUTH-acl` is unaffected.

**Role route parity.** Dropping the eager `em.create(RoleAcl, …)` in favour of `acl?.` is a strict improvement — the old code created a managed entity before `assertActorCanGrantAcl` could refuse the request.

**Conventions.** `registerCommand` at module scope, local `buildCommandContext`-style ctx construction, and `resolveTranslations()` in `buildLog` all match `auth/commands/{roles,users}.ts` and `directory/api/organization-branding/route.ts`. Both i18n keys are present in all five locales and alphabetically placed. Backward compatibility: no schema change, response bodies and status codes unchanged, two additive command ids, additive optional `tenantId` params, additive `403`. Nothing under `packages/enterprise/`. The two `as any` casts in the diff are carried-over lines — the PR actually removes one (`tenantId: auth.tenantId as any`).

---

### 🔵 Low — worth a conscious sign-off, not a blocker

**1. Cache-invalidation failures are now swallowed on a security-sensitive path.**

`commands/acl.ts` wraps `config.invalidate({ ctx, input })` in a `try/catch` that logs at `error` and continues. Previously `rbacService.invalidateUserCache` / `invalidateTenantCache` propagated out of the route and produced a `500`. So a cache outage now means a **revoked grant keeps being served from cache until its TTL while the caller sees `200 OK`**.

The reasoning in the comment is sound and I agree with the trade: the write has already committed when this runs, the bus persists the action log *after* `execute` returns, so throwing would trade a recorded permission change for an HTTP 500 that misreports a write which did happen — and the staleness occurs either way once the adapter is down. No change requested. But this only stays safe if the log line is actually alarmable: please make sure `ACL cache invalidation failed after a committed permission change` has an alerting rule, otherwise the failure mode is silent by construction.

**2. Spec status vs. index placement.**

`.ai/specs/2026-08-07-auth-acl-change-audit-logging.md` declares `Status: Implemented — 2026-08-07`, but the `.ai/specs/README.md` row lands under *"Specs awaiting implementation or partially complete"*. Per `.ai/specs/AGENTS.md` the move to `implemented/` waits for deployment, so the placement is defensible today — the two just read as contradicting each other. Worth reconciling when the spec moves post-merge.

---

### ⏳ CI

`test`, `lint`, `ds-lint`, `audit`, `audit-scope`, `prepare`, `scope` and `license/cla` are **green** on `de39c21`. The 15 `ephemeral-integration` shards were still `IN_PROGRESS` at review time — no local substitutes were run for them; branch protection and the merge queue are the authority there. `TC-AUTH-058` has not been re-executed since the rebase, as the PR description notes, so those shards are the evidence for it.

Note the CLA checkbox in the PR description is still unticked, though `license/cla` reports green.

---

<sub>Labels and manual-QA routing follow in separate comments. `in-progress` is released when this review closes.</sub>

pkarw:

pkarw:

The diff changes **no `.tsx`** — the UI is untouched. What QA is verifying is that the two permission editors still behave exactly as before, and that each save now produces an audit entry. Priority is on P0.

### P0 — Permission edits still work, and are now audited

**Where to click**
- `/backend/roles/<roleId>/edit` → the role permissions / ACL panel
- `/backend/users/<userId>/edit` → the per-user ACL panel
- `/backend/audit-logs`

**What to verify**
- Role: add a feature to a role, save → success toast, and the feature is still there after a page reload. Then open `/backend/audit-logs` → a new **Change role permissions** entry naming the role, with before/after grant sets and a `changes` diff listing only `features`.
- Role: remove that same feature, save → reload confirms removal; a second audit entry appears with `effect: revoked` (or `changed`, if the role keeps other grants).
- User: grant a per-user feature override, save → reload confirms it; audit shows **Change user permissions** with the target user's name/email in the entry context.
- User: clear the override entirely (remove every feature, super-admin off), save → reload shows no custom ACL; audit shows a final entry whose after-state is empty. **The row is deleted on this path — confirm the entry still names the user.**
- **No-op suppression:** open a user's edit page, change only the *name* (touch nothing in the ACL panel), save. The page PUTs the ACL on every save, so confirm **no** new "Change user permissions" entry appears. Repeat on a user that has no ACL override at all. Then re-save an *unmodified* role ACL and confirm no entry there either.
- **No Undo action:** on every entry above, confirm the audit-log row renders **no Undo/Redo button**, and that filtering the list with `?undoableOnly=true` excludes them. This is deliberate (undo is gated on `audit_logs.undo_*`, not `auth.acl.manage`) — an Undo button appearing here is a **fail**.
- **Permissions actually take effect:** after granting a feature to a user, have that user reload the backend and confirm the newly granted area is reachable; after revoking, confirm it disappears. This exercises the RBAC cache invalidation that moved out of the route into the command.

**What can go wrong**
- A save reports success but the grant is not persisted (the write path moved to the command bus — a silent no-op would show up as "reload loses the change").
- The audit entry is written but carries the wrong tenant/organization and is therefore invisible in the list — if you can reproduce a save that produces no entry at all, that is the highest-value bug to report.
- Duplicate/noise entries on every unrelated user save (no-op suppression regressed).
- A revoked permission keeps working after reload (cache invalidation did not run). Note: this PR deliberately **swallows** cache-invalidation failures and only logs them, so check the server log for `ACL cache invalidation failed after a committed permission change` if you hit this.
- An Undo button on an ACL entry — a privilege-escalation path, report immediately.

### P1 — Tenant scoping and access boundaries

**Where to click**
- `/backend/users/<userId>/edit` for a user in **another** tenant, signed in as a non-super-admin admin holding `auth.acl.manage`
- The same page signed in as a super admin

**What to verify**
- Non-super-admin: a user from another tenant must stay **not found / inaccessible** — the ACL panel must not load a foreign override, and saving must not write one. (This is the exact regression the previous review caught; it is fixed and unit-tested, but it is the highest-risk surface in the diff.)
- Non-super-admin editing a user in **their own** tenant: unchanged, works normally.
- Super admin: editing a user or role in another tenant still works, and the resulting audit entry is visible to a reader scoped to the **target** tenant (not only to the super admin's own).
- Concurrent-edit guard: open the same role ACL in two tabs, save in tab A, then save in tab B → tab B must surface a conflict (409), not silently overwrite.

**What can go wrong**
- Cross-tenant read or write of a permission override by a non-super-admin (blocker — stop and report).
- A super admin's cross-tenant edit produces an audit entry nobody can see in any tenant's log list.
- The 409 conflict bar stops appearing on ACL edits.

### P2 — Locales

**Where to click**
- `/backend/audit-logs` with the UI language switched to `pl` / `de` / `es` / `ko`

**What to verify**
- The two new action labels render translated ("Zmień uprawnienia roli" / "Zmień uprawnienia użytkownika" in `pl`), not as raw keys `auth.audit.roleAcl.update` / `auth.audit.userAcl.update`.

### Pass/fail

pkarw:

0 blockers, 0 majors, 0 mediums, 2 lows (a cache-invalidation failure now logs instead of returning 500 — please confirm that log line is alarmable; and the spec header says `Implemented` while its README row sits under "awaiting implementation"). Autofix did not run — there was nothing actionable to fix. The previous cross-tenant guard blocker was re-verified from scratch and is genuinely fixed, with the requested regression tests present at both the route and `grantChecks` layers. `ephemeral-integration` was still running at review time; the merge queue is the authority on it.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-21T00:31:32Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

pkarw:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ PASS — every P0 and P1 route from the QA routing comment behaved as specified, and the two P2 routes (locales, perceived performance/mobile) are clean. One pre-existing, non-blocking edge case is documented under Notes for QA; it reproduces identically on `develop`, so it is not a regression of this PR.
**Environment:** `http://127.0.0.1:5001` — an ephemeral production build (`next start`, `NODE_ENV=production`) over a disposable testcontainers Postgres, started by this run · roles driven: `admin@acme.com`, `employee@acme.com`, `superadmin@acme.com`, plus a second tenant provisioned in the throwaway database for the scoping routes · browser `playwright` (Chromium)
**Verified:** `fix/auth-acl-audit-log-commands` @ `de39c212`

Every P0 route was executed **twice, against two independent disposable database instances**, with identical results.

### Scenario (P0 — permission edits still work and are now audited)
**Where to click:** `/backend/roles/<roleId>/edit` · `/backend/users/<userId>/edit` · `/backend/audit-logs`

| 1 | Grant `api_keys.view` to the `employee` role and save | Save succeeds; the grant survives a reload | Redirected to the role list; the checkbox was still checked after reloading the edit page | ✅ |
| 2 | Open the action log after that save | A **Change role permissions** entry naming the role, with before/after grant sets and a `changes` diff listing only `features` | `auth.role-acl.update`, actor `admin@acme.com`, resource `auth.role_acl`, `changes` keys exactly `['features']`, `api_keys.view` present in the after-snapshot and absent from the before-snapshot, `undoToken: null` | ✅ |
| 3 | Inspect the ACL row's controls, then switch on **Undoable only** | No Undo/Redo control on the ACL entry; the filter excludes it | Zero buttons in the ACL row's controls cell (while the neighbouring `Update role` row shows its undo arrow); the filtered list returned zero ACL rows | ✅ |
| 4 | Re-save the role ACL without changing anything | No new entry (no-op suppression) | 3 entries before, 3 after | ✅ |
| 5 | Revoke the same feature and save | Removal persists; a second entry records it | Unchecked after reload; the second entry carries the feature in the before-snapshot and not in the after-snapshot | ✅ |
| 6 | Enable a per-user override for `employee@acme.com`, grant the feature, save | Override persists; the entry names the target user | `auth.user-acl.update` whose context names `employee@acme.com`; empty before-state, after-state contains the feature; `undoToken: null`; still set after reload | ✅ |
| 7 | Re-check the feature from the employee's own **live** session | The grant takes effect immediately (the cache invalidation that moved into the command) | `feature-check` flipped `false → true` with no re-login, and `/backend/api-keys` became reachable for that user | ✅ |
| 8 | Clear the override entirely and save | The row is deleted; the entry still names the user with an empty after-state; access is revoked | Page returned to *Permissions inherited from roles*; the entry still names `employee@acme.com` with an empty after-state; `feature-check` returned to `false` | ✅ |
| 9 | Change only the user's **Display name** and save | No new *Change user permissions* entry, even though the page PUTs the ACL on every save | 2 user-ACL entries before, 2 after | ✅ |

### Scenario (P1 — tenant scoping and access boundaries)

| 10 | Same role ACL open in two tabs; save in A, then in B | Tab B surfaces a 409 conflict rather than silently overwriting | `PUT /api/auth/roles/acl` returned **409**; the form stayed put and showed *"This record was modified by someone else. Refresh and try again."*; tab A's feature was applied and tab B's was not | ✅ |
| 11 | A non-super-admin admin holding `auth.acl.manage` reads and writes the per-user ACL of a user in **another tenant** | Not found / inaccessible; no foreign override loaded; no write; no audit row | `GET` and `PUT` both returned **404 `User not found`**; the foreign user's own ACL stayed empty; **zero** audit rows for that user; the edit page rendered *User not found*; the Acme user list never listed the foreign user | ✅ |
| 12 | Same-tenant editing after all of the above | Unchanged | Role and user edits in the actor's own tenant kept working throughout steps 1–9 | ✅ |

### Scenario (P2 — locales and perceived performance)

| 13 | Switch the UI to `pl`, `de`, `es`, `ko` and make an ACL change in each | The two new action labels render translated, never as raw keys | *Zmień uprawnienia roli* / *Rollenberechtigungen ändern* / *Cambiar permisos del rol* / *역할 권한 변경* all rendered; `auth.audit.roleAcl.update` / `auth.audit.userAcl.update` never leaked to the screen | ✅ |
| 14 | Cold-load `/backend/audit-logs`; smoke both surfaces at 390×844 | Usable perceived performance, no mobile breakage | Shell painted in **437 ms**, page settled in **1.7 s**; the action log and the role ACL panel both render at 390 px | ✅ |

### Notes for QA

- **Two entries per role save, by design.** Saving the role edit form writes the pre-existing, undoable `auth.roles.update` entry for the role record *and* the new, non-undoable `auth.role-acl.update` entry. The *"Last operation … Undo"* banner therefore refers to the **role update**, never to the ACL change — worth knowing before reading a lone `Undo` button on that page as a violation of this PR's log-only policy. The ACL rows themselves render an empty Actions cell in every locale.
- **⚠️ Pre-existing, not a regression — super-admin writes outside the active tenant.** A *tenant-bearing* super admin who PUTs `/api/auth/users/acl` for a user in another tenant gets `200 {ok: true}`, but the `user_acls` row and the action-log row are both written under the **actor's** tenant: the target tenant's admin sees neither the override nor the audit entry, and the grant has no effect for the target user. I reproduced the same scoping on `develop`'s version of the route (it wrote `tenantId: auth.tenantId` unconditionally), and this PR's `resolveAclTenantId` deliberately reserves the target-user fallback for a *tenant-less* global admin, so read and write still agree. It is also unreachable from the UI — the user list and edit page are tenant-scoped and answer *User not found*. Flagging it as a follow-up candidate, not as a blocker for this PR.
- **Concurrent edit, one nuance.** In step 10 tab B's `PUT /api/auth/roles` (the role record itself) returned **200** before the ACL PUT returned 409, so a stale tab can still write the role's own fields while its ACL change is correctly rejected. That path is outside this PR's diff.
- **Environment provisioning.** A second tenant, a tenant-B employee and a known super-admin password were created **inside the disposable QA database only**, to make the scoping routes drivable. Nothing outside that throwaway database was touched, and no credentials from it appear in this report.
- **Not exercised:** organization-scope restrictions inside the ACL editor, and the super-admin cross-tenant route *through the UI* (it requires switching the active tenant, which the seeded single-org demo data does not exercise cleanly).

### Screenshots

**Role ACL change in the action log** — the new `Change role permissions` entry
audit log role entry

**Entry details** — `changes` limited to `features`, with before/after grant sets and collapsible snapshots
role entry details

**Undoable only** — the ACL entry is excluded by the filter and renders no Undo control
undoable only filter

**The per-user grant takes effect immediately** — the employee's own live session reaches the newly granted area without re-login
employee sees granted area

**User ACL entries** — grant and clear, both naming the target user
audit log user entries

**Concurrent edit** — the second tab is rejected with a 409 and the conflict message, not silently applied
concurrent edit conflict

**Locale (pl)** — the label is written translated; note the empty Actions cell on both ACL rows next to the undo arrow on `Update role`/`Update user`
audit log locale pl

**Cross-tenant** — a non-super-admin gets `User not found` for a user in another tenant
cross tenant blocked

**Mobile smoke (390×844)**
audit log mobile

pkarw:

This PR ships `TC-AUTH-058-acl-audit-log.spec.ts`, which is thorough — but it drives the
routes through `APIRequestContext`, so it is an **API-level** integration test. Nothing in
the diff exercises the two screens an operator actually uses, and the manual run above
turned up UI-level behaviour that only a browser test can pin down: the per-user override
toggle **unmounts itself** once enabled (the checkbox lives inside the inherited-permissions
banner it hides), one role save produces **two** action-log rows (the undoable
`auth.roles.update` plus the new non-undoable `auth.role-acl.update`), and the no-Undo
guarantee is a property of the **row's controls cell**, not of the page — the page still
shows an `Undo` for the role update right next to it.

**Scenario (derived from the manual run above — run `/om-integration-tests`):**

1. **Setup:** create a role and a user via the existing `authFixtures` API helpers
   (`createRoleFixture` / `createUserFixture`), and sign in as the seeded admin in the browser.
2. **Act (role):** open `/backend/roles/<roleId>/edit`, tick the `f-<feature>` checkbox, save,
   reload the edit page. Then open `/backend/audit-logs`.
3. **Assert (role):** the checkbox is still ticked after the reload; a **Change role permissions**
   row exists; that row's controls cell contains **zero** buttons while the sibling `Update role`
   row has its undo control; ticking **Undoable only** removes the ACL row from the list;
   clicking the ACL row opens the details dialog showing `Features` before/after.
4. **Act (user):** open `/backend/users/<userId>/edit`, click `#overrideAcl` (click, not `check()` —
   the control unmounts), wait for the feature grid, tick the feature, save, reload.
5. **Assert (user):** the grant persisted; a **Change user permissions** row names the target user;
   from that user's own session `POST /api/auth/feature-check` flips `false → true` with no re-login.
6. **Act (no-op):** save the user form again changing only **Display name**.
7. **Assert (no-op):** the user-ACL row count is unchanged — this is the regression most likely to
   come back, since the page PUTs the ACL on every save.
8. **Teardown:** clear the override, revoke the role feature, and delete every fixture created.

pkarw:

Labels are unchanged: this run posted QA evidence under the skill's evidence-only default and did not apply `qa-approved` — the QA-approval gate stays closed until a maintainer accepts this evidence. To accept it as the manual QA pass for this PR, apply `qa-approved` (and drop `needs-qa` only if you also want to record it as skipped, which is not recommended here).

pkarw (approved):

PR #5196, opened 2026-08-11
feat(telemetry): add root/links span options and trace data_sync batches

jtomaszewski:
## Summary

`withSpan()` always nests under the active context, so a span can never start a new trace. Combined with the `ParentBasedSampler` this package configures — which only rolls the dice at a trace's **root** — an entire long-lived job runs under **one sampling decision, taken on whatever request triggered it**.

Trace context propagates from the triggering HTTP request through the queue into the worker, so a multi-day sync is a single trace rooted at a `POST /api/[...slug]` from hours earlier. Both ends of the sampling ratio fail:

- **Ratio < 1 — the run is a coin flip.** At 0.25, three runs in four emit *nothing at all*, for their entire life, and nothing can be done from the outside once the flip is lost. Observed in practice: the environment was healthy (5k spans/2h) while the running backfill contributed zero spans.
- **Ratio = 1 — a firehose into one unrenderable trace.** A single pod emitted ~1.14M spans in 2h under one `trace_id`.

This is not sync-specific — it hits any long-lived queue job (reindex, change-feed reapers, imports).

The fix adds the missing primitive (`root`, with `links` so rooting doesn't sever causality) and applies it in `data_sync`'s sync engine, which is the choke point every adapter's batches already flow through. Instrumenting the engine rather than each adapter means every integration gets sampling-correct, bounded, renderable traces for free — and no adapter has to hand-roll a batch span it could never root correctly anyway.

Everything is additive. **No adapter needs changes.**

## Changes

- **`packages/telemetry`** — `Span` gains an optional `updateName?(name)` (rename an in-flight span whose identity is only known after it runs; optional so a third-party provider still satisfies the interface). `SpanOptions` gains optional `root?: boolean` (start a new trace, so the sampler takes a fresh decision) and `links?: TraceCarrier[]` (causal links back to the trigger). Both are translated in one `toOtelSpanOptions()` helper shared by `runInSpan` and `runInRemoteSpan`. Links are W3C **carriers** rather than span contexts, so one option covers the in-process case (`captureTraceContext()`) and a remote one (a queue payload's `_trace`), reusing plumbing that already existed. Link extraction runs against `ROOT_CONTEXT`, never the active context, so an empty carrier yields *no link* instead of silently linking a span to its own ambient parent; malformed carriers are dropped rather than emitted as invalid links. Noop/console providers need no change.
[...]

PR #5375, opened 2026-08-18
feat(telemetry): add root/links span options and trace data_sync batches

patzick:
Supersedes #5196.

The original PR's head branch lives on a fork (`fullstackhouse/open-mercato`) that this account does not have push access to (confirmed via the GitHub API — `push: false`), so the develop-conflict resolution could not be pushed back to update #5196 directly. This PR carries the same commits (original authorship by @jtomaszewski preserved) plus a merge of `develop` with conflicts resolved in:

- `packages/core/src/modules/data_sync/AGENTS.md` — additive conflict, kept both new file references.
- `packages/core/src/modules/data_sync/lib/sync-engine.ts` — real conflict between this PR's `forEachBatch` root-span-per-batch refactor and `develop`'s meanwhile-added heartbeat, `runId`/`parameters` passthrough, and ownership-fenced `commitBatchProgress`. Resolved by composing `forEachBatch(withHeartbeat(stream, tick, interval), ...)` and carrying forward develop's additions. Also deduplicated `resolveProviderKey` (now shared via `adapter-registry.ts`).
- `packages/core/src/modules/data_sync/lib/__tests__/sync-engine-stream-cleanup.test.ts` — updated mocks (`resolveProviderKey`, `progressService.getJob`) to match the merged engine's dependencies.

Verified after merge: `yarn typecheck`, `yarn lint`, and the full `data_sync` (30 suites / 207 tests) + `shared` telemetry (16 tests) suites all pass.

---## Summary

`withSpan()` always nests under the active context, so a span can never start a new trace. Combined with the `ParentBasedSampler` this package configures — which only rolls the dice at a trace's **root** — an entire long-lived job runs under **one sampling decision, taken on whatever request triggered it**.

Trace context propagates from the triggering HTTP request through the queue into the worker, so a multi-day sync is a single trace rooted at a `POST /api/[...slug]` from hours earlier. Both ends of the sampling ratio fail:

- **Ratio < 1 — the run is a coin flip.** At 0.25, three runs in four emit *nothing at all*, for their entire life, and nothing can be done from the outside once the flip is lost. Observed in practice: the environment was healthy (5k spans/2h) while the running backfill contributed zero spans.
- **Ratio = 1 — a firehose into one unrenderable trace.** A single pod emitted ~1.14M spans in 2h under one `trace_id`.

This is not sync-specific — it hits any long-lived queue job (reindex, change-feed reapers, imports).

The fix adds the missing primitive (`root`, with `links` so rooting doesn't sever causality) and applies it in `data_sync`'s sync engine, which is the choke point every adapter's batches already flow through. Instrumenting the engine rather than each adapter means every integration gets sampling-correct, bounded, renderable traces for free — and no adapter has to hand-roll a batch span it could never root correctly anyway.

Everything is additive. **No adapter needs changes.**

## Changes

- **`packages/telemetry`** — `Span` gains an optional `updateName?(name)` (rename an in-flight span whose identity is only known after it runs; optional so a third-party provider still satisfies the interface). `SpanOptions` gains optional `root?: boolean` (start a new trace, so the sampler takes a fresh decision) and `links?: TraceCarrier[]` (causal links back to the trigger). Both are translated in one `toOtelSpanOptions()` helper shared by `runInSpan` and `runInRemoteSpan`. Links are W3C **carriers** rather than span contexts, so one option covers the in-process case (`captureTraceContext()`) and a remote one (a queue payload's `_trace`), reusing plumbing that already existed. Link extraction runs against `ROOT_CONTEXT`, never the active context, so an empty carrier yields *no link* instead of silently linking a span to its own ambient parent; malformed carriers are dropped rather than emitted as invalid links. Noop/console providers need no change.
- **`packages/shared`** — `@open-mercato/core` must not depend on `@open-mercato/telemetry`, so span creation rides the existing runtime bridge that `initTelemetry()` already populates: an **optional** `TelemetryRuntime.withSpan?` (optional so a bootstrap predating it still satisfies the interface) plus `withTelemetrySpan()` / `captureTelemetryTrace()` consumer helpers. With telemetry off, `withTelemetrySpan` is `fn` plus one global lookup — no span allocated, no provider resolved, the OTEL SDK never reached. This follows the same `globalThis` symbol-registry convention the shared logger extension already uses.
- **`packages/core` (`data_sync`)** — the sync engine emits `data_sync.import.batch` / `data_sync.export.batch` as **root** spans linked back to the run, carrying `data_sync.run_id`, `integration_id`, `provider_key`, `entity_type`, `direction`, `batch_index`, per-batch counters, and `om.tenant_id` / `om.organization_id`. The span covers the adapter's read **and** the engine's per-batch bookkeeping, so a batch trace is self-contained. The read that finds the stream drained is real adapter work, so it stays traced — but under `data_sync.import.drain` / `data_sync.export.drain`, so a run over N batches emits exactly N `*.batch` spans plus one `*.drain`, and an empty run emits **zero** batch spans. Batch counts and latency panels therefore measure batches only.
- **New `data_sync/lib/batch-stream.ts`** — the one structural change. The engine drives the adapter's async iterator explicitly instead of using `for await`, because an adapter generator does its real work (read, transform, upsert) *before* it yields, so only a span wrapping `next()` contains it. Closing reproduces the language's own `IteratorClose` rules rather than approximating them: no `return()` when the stream exhausts or `next()` throws (already closed), `return()` with its failure **surfaced** on an early stop, and `return()` with its failure swallowed only when the handler already threw (that error wins). Adapter `finally` blocks therefore behave exactly as they did under `for await`.
- **Docs** — spec, `UPGRADE_NOTES.md`, `packages/telemetry/README.md`, `packages/telemetry/AGENTS.md`, `data_sync/AGENTS.md`, and the integrations/data-sync docs page.

## Specification

- [x] Yes
- [ ] No (created a new spec)

**Spec file path:** `.ai/specs/2026-04-29-telemetry-and-otel.md`

Updated: S2 gains "Rooting long-lived work — `root` + `links`" and "Emitting spans from packages that must not depend on this one"; S6 gains the data-sync instrumentation surface; API Contracts gains the new options and the shared-bridge table; the Backward compatibility table gains four rows; Risks gains R18 (iterator-close fidelity) and R19 (trace-topology change); plus two changelog entries (the second covering the review response: distinct drain span, accurate sampling probabilities, `err` logging).

## Testing

Full CI-mirroring gate from `.ai/agentic.config.json`, run locally:

| Command | Result |
|---|---|
| `yarn build:packages` | 22/22 |
| `yarn generate` | ok |
| `yarn i18n:check-sync` | all locales in sync |
| `yarn i18n:check-usage` | advisory only |
| `yarn typecheck` | 22/22 |
| `yarn test` | 25/25 |
| `yarn build:app` | ok (forced, not cached) |
| `yarn lint` | 0 errors |

Note on `yarn test`: with unbounded turbo concurrency this machine SIGSEGVs a jest **worker process**, which surfaces as a random suite failing to *run* (a different suite each time, no assertion failure) — it reproduces on the unmodified base. `yarn test -- --concurrency=4` is green 25/25, and every affected package passes standalone (`core` 1205 suites / 9324 tests, `cli` 1480, `telemetry` 87).

New tests:

- **`packages/telemetry/src/__tests__/otlp-root-sampling.test.ts`** — reproduces the reported bug deterministically at sampling ratio 0: a span nested under a sampled remote parent is still recorded (inheritance), while a `root: true` span is **dropped**. That proves the decision is genuinely re-taken rather than inherited, which asserting on trace IDs alone would not.
- **`otlp-integration.test.ts`** (+4) — `root` yields a new `traceId` with no parent; a rooted span links back to the run's trace; empty/malformed carriers are dropped; an empty carrier does not link to the ambient parent.
- **`data_sync/lib/__tests__/batch-stream.test.ts`** (11) — one test per `IteratorClose` path: stream closed on completion / early stop / handler throw; the handler's error wins over a failing close; a close failure on early stop **is** surfaced; an already-completed iterator (exhausted, or `next()` threw) is not closed again; works with telemetry off; root + link + attribute shape; the span-name sequence is `[batch, batch, drain]`; an empty stream emits only the drain span; a provider without `updateName` degrades to the batch name instead of failing.
- **`data_sync/lib/__tests__/sync-engine-stream-cleanup.test.ts`** (3) — adapter `finally` runs on cancellation, on a commit throw, and on an ownership conflict, with finalization behaviour unchanged.
- **`packages/shared/src/lib/telemetry/__tests__/runtime.test.ts`** (16) — the bridge this PR extends had no suite in its own package, so the mutation job could not discover a test for it. Covers backend detection (case/whitespace normalization, an explicit value winning over `TELEMETRY_BACKEND`), the registry lifecycle (register/get/dispose, a superseded disposer leaving a newer runtime in place, reset), `withTelemetrySpan` on all three paths (telemetry off, a runtime predating `withSpan`, full delegation of name/options/span), and `captureTelemetryTrace`'s empty-carrier handling. Mutation score 97.14% across 35 scored mutants; the single survivor is `typeof process === 'undefined'` → `false`, an environment guard the node test environment cannot exercise.

Counts: telemetry 82 → 87, `data_sync` 107 → 121, `shared` +16. The pre-existing 107 `data_sync` tests pass **unchanged**, which is the main evidence that the engine rewrite is behaviour-preserving.

## Backward compatibility

| Surface | Change | Classification |
|---|---|---|
| `Span` / `TelemetrySpan` | optional `updateName?(name)` | ADDITIVE — optional so a third-party provider or an older bootstrap still satisfies the interface; callers invoke it as `span.updateName?.(…)` and degrade to the original span name |
| `SpanOptions` | optional `root?` / `links?` | ADDITIVE — omitting them preserves current behaviour exactly; a custom `TelemetryProvider` that ignores them degrades to "not a root", never breaks |
| `TelemetryRuntime` | optional `withSpan?` | ADDITIVE — optional so an older bootstrap still satisfies the interface |
| `DataSyncAdapter.streamImport` / `streamExport` | **unchanged** | adapters need no changes; generator `finally` semantics preserved |
| Trace topology (telemetry-on deployments) | sync work moves out of the trigger's trace into per-batch traces linked back to it | the intended fix, but visible in saved dashboards — documented in `UPGRADE_NOTES.md` |

One contract detail is now enforced where it previously was not: the value returned by `streamImport`/`streamExport` must be a genuine `AsyncIterable` — which is exactly how both have always been typed. `for await` also happened to accept a *synchronous* iterable of promises; that was never part of the declared contract, and an untyped adapter relying on it now fails loudly on the first batch rather than silently. TypeScript adapters are unaffected.

## Checklist

- [x] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required) — **not required**: this change adds no API route, no UI surface and no schema change. Its behaviour is observable only as span emission and adapter-stream lifecycle, both of which are covered by the unit tests above (including a deterministic reproduction of the original sampling defect). The pre-existing `data_sync` suite acts as the behaviour-preservation guard.
- [x] QA routing set: `skip-qa` — see rationale below.

**QA routing rationale (`skip-qa`).** This change qualifies for the automated-verification exemption in `AGENTS.md` → PR Workflow on every clause: it touches no UI-rendering file (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`), leaves the database structure unchanged (no entities, no migrations) and the API surface unchanged (no routes touched), breaks no `BACKWARD_COMPATIBILITY.md` contract (all surfaces additive — table above), and ships automated tests for the behaviour it changes in this same PR. There is no manually exercisable UI surface to click through.

### Design System Compliance

Not applicable — this PR contains no UI code (no `.tsx` outside tests, nothing under `packages/ui/`, no components, no styles).

## Linked issues

None.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5375"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789677995&installation_model_id=432243&pr_number=5375&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5375&signature=60abd235c5d4fde067b91707b45392eb15112c419fb915d802f64c554525b174"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


patzick:
Labeling `review`: this is a fresh PR (replacing the closed #5196, which had already been approved), so it needs a new review pass before it can move to `merge-queue`.

PR #5378, opened 2026-08-19
fix(customers): terminate the last two detail load-more gates on a short page

### LINKED ISSUE #5347: Two remaining `page < totalPages` load-more gates in customers/components/detail
Follow-up to #5274, which converted six incremental-loading surfaces from a server-reported `total`/`totalPages` to short-page termination via `hasMoreFromPage` (`packages/shared/src/lib/pagination/load-more.ts`). Two gates of the same shape remain, both flagged in that PR's re-review and deliberately left out to keep its scope honest.

Both have the same failure mode as the six converted ones: when the reported total under-reports the result set — rows inserted between two requests, and a capped list count by design once `OM_LIST_COUNT_CAP` lands (see `.ai/specs/2026-07-27-list-count-strategies.md`) — the affordance disappears and the remaining rows become unreachable.

### 1. `packages/core/src/modules/customers/components/detail/hooks/usePersonTasks.ts`

Two spots read the same bound:

- `:245` — `if (pageInfo.page >= pageInfo.totalPages) return`, guarding the fetch inside `loadMore`
- `:479` — `const hasMore = entityId != null && pageInfo.page < pageInfo.totalPages`

`totalPages` is taken from the payload in `mapResponse` (`:190`). Backs `TasksSection.tsx`'s legacy path.

**Served-count analysis (already done):** `mapRowToSummary` is a 1:1 map and `mergeUnique(prev, mapped)` dedupes *after*, so `mapped.length` is the served count — pass that to `hasMoreFromPage`, not the merged array's length. This is obligation 1 in the helper's docstring, and getting it wrong was a real finding on #5274.

### 2. `packages/core/src/modules/customers/components/detail/EntityTagsDialog.tsx`

- `:1253` — `(activeCategory.source === 'tags' || activeCategory.source === 'labels') && activeCategoryPage < activeCategoryTotalPages`
- `activeCategoryTotalPages` is set from the response at `:762-764`; entries are appended via `mergeOptions` at `:764-769`

**Served-count analysis (already done):** `mapEntries` (`:746-752`) is a plain `.map()` with no filter or dedupe, so `fetchedEntries.length` is the served count. The wrinkle here is not the count but the state: `activeCategoryTotalPages` is keyed to the *active* category and replaced when the user switches, so a `hasMore` replacement has to preserve that keying rather than becoming a single global flag.

### Notes

- Neither file currently has a test harness, so each conversion needs one. #5274's re-review established the bar: the fixture must model what the endpoint actually emits — an empty page past the end for query-engine-backed lists, rather than a terminal response the route cannot produce.
- Both endpoints are query-engine backed and compute an unclamped offset, so obligation 2 in the helper's docstring (the clamping caveat that applies to `/api/attachments`) does not apply here.
- Verified in the same pass and deliberately **not** part of this issue: `ActivityHistorySection.tsx:300` OR's `totalPages` onto a canonical cursor signal, so an under-reporting total can only add "there may be more", never hide rows; and `useVersionHistory.ts:81` is already page-fullness based.

adeptofvoltron:
## 🎯 Goal

Closes #5347 — the two `page < totalPages` load-more gates that #5274 deliberately left out of its scope, both flagged in that PR's re-review.

Both had the same failure mode as the six surfaces #5274 converted: when the server-reported total under-reports the result set — rows created between two requests today, and a capped list count by design once `OM_LIST_COUNT_CAP` lands (`.ai/specs/2026-07-27-list-count-strategies.md`) — the affordance disappears and every row past the reported end becomes unreachable behind a button that is no longer rendered.

### `hooks/usePersonTasks.ts` (backs `TasksSection`'s legacy path)

`pageInfo.totalPages` is replaced by a `pageInfo.hasMore` flag derived from `hasMoreFromPage`, which retires both readers in one move: the fetch guard inside `loadMore` (was `pageInfo.page >= pageInfo.totalPages`) and the exported `hasMore` (was `pageInfo.page < pageInfo.totalPages`).

Following obligation 1 in the helper's docstring, the flag is measured on `mapped.length` — `mapRowToSummary` is a 1:1 map of the served rows and `mergeUnique(prev, mapped)` dedupes *after* it, so the merged array's length would be short whenever a page overlapped what was already on screen and would end the sequence early. That mismeasurement was a real finding on #5274, so the "a full page that dedupes away entirely still offers the next one" case is covered by its own test.

`totalPages` is dropped from the local `CustomerTodosResponse` type with a comment recording why, so the bound cannot be reintroduced by accident. `total` stays: it feeds the section's count badge, where an under-report is cosmetic rather than a way to strand rows.

### `EntityTagsDialog.tsx`

`activeCategoryTotalPages` becomes `activeCategoryHasMore`, set from `hasMoreFromPage` on the raw response items — the count before `mergeOptions` folds the page into the entries already on screen (`mapEntries` is a plain `.map()`, so raw and mapped lengths agree; measuring the raw array keeps that independent of the mapper).

The issue's wrinkle here is state, not counting: the flag has to stay keyed to the *active* category rather than becoming one global boolean. It does, by the same two mechanisms the page number already used — the fetch effect rewrites it on every category/search/page change, and the reset effect now clears it alongside `activeCategoryPage`, so a category with a full first page cannot leave its button behind on a category whose first page is short. A dedicated test switches Tags → Labels and asserts exactly that. A non-ok response now also clears the flag rather than leaving a stale affordance pointing at a page that failed to load.

## 🧪 Tests

Neither surface had a load-more harness, so both get one. Per the bar #5274's re-review set, the fixtures model what these endpoints actually emit — both are query-engine backed and compute an unclamped offset, so the page past the end comes back **empty** rather than re-serving the last page. (Obligation 2 in the helper docstring, the clamping caveat that applies to `/api/attachments`, does not apply to either endpoint.)

New `hooks/__tests__/usePersonTasks.test.tsx` (6 tests) and a `load-more termination` block in `__tests__/EntityTagsDialog.test.tsx` (4 tests) each cover: a full page with `totalPages: 1` still offering more; a short page with a large `totalPages` terminating; the empty page past the end ending the sequence while keeping loaded rows; and a full page surviving client-side dedupe/merge. The tasks hook additionally asserts the next-page request shape and that `loadMore` no-ops once the sequence has ended.

The pre-existing `loads paged tag options remotely` test carried a one-item first page with `totalPages: 2` — a shape the endpoint cannot emit alongside a real second page. Its fixture now serves a full first page and a short second one, and it asserts the button disappears after that short page.

**Verified as regressions:** stashing only the two source files and re-running both suites fails 10 of the 17 tests; with the fix applied all pass.

## ✅ Validation

Run in this environment (a linked worktree without `node_modules`, so `yarn`/turbo scripts cannot run — jest was invoked directly against the parent repo's binary):

- ✅ `packages/core` — the full `customers/components/detail` suite: **68 suites, 287 tests passed** (includes both changed test files)
- ✅ `packages/core` repo-wide guards touched by this diff: `explicit-sort-comparators`, `optimistic-lock-ui-coverage-workspace`, `feature-policy-authorization-coverage`, `license-metadata-consistency`, `alert-duplicate-icon-coverage` — 22 tests passed
- ✅ `packages/cli` module-facts guards over live customers sources: `module-facts.customers.fixture`, `module-facts.extension-hosts` — 8 tests passed

**Gate commands skipped, and why:** `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app` all fail in this worktree with *"Couldn't find the node_modules state file"* — the turbo half of `.ai/agentic.config.json`'s validation list cannot run here. CI runs the full gate on this PR.

A direct `tsc -p packages/core/tsconfig.json` was attempted as a substitute and is not meaningful in this checkout: the parent repo's `packages/shared/dist` predates #5274, so `@open-mercato/shared/lib/pagination/load-more` is unresolvable there — the identical `TS2307` is reported for the four already-merged call sites (`AssignRoleDialog`, `DealsSection`, `LinkedEntitiesField`, `ParticipantsField`) as for the two files touched here. It is stale build output, not a defect in this change; `yarn build:packages` (the gate's first command) resolves it. `yarn lint` could not run either — ESLint resolves `eslint-config-next` from the worktree root, which has no `node_modules`.

## 📌 Scope

Only the two gates the issue names. The `totalPages` readers that remain under `customers/components/detail` are numbered-pagination displays (`PersonCompaniesSection`, `CompanyPeopleSection`) or the surfaces the issue verified and deliberately excluded — `ActivityHistorySection.tsx:300` ORs `totalPages` onto a canonical cursor signal, so an under-reporting total can only add "there may be more", never hide rows.

## 🏷️ Labels

`bug` (a defect that hides existing rows), `priority-medium` (real data is unreachable, but only past the first page of two dialogs and only when the total under-reports), `risk-low` (two client-side components, no schema, API or contract surface touched, both conversions covered by new regression tests), `needs-qa` (both changes are `.tsx`/UI-rendering, so the automated-verification exemption does not apply).

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5378"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789719804&installation_model_id=432243&pr_number=5378&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5378&signature=d2c709522e772f583eddc6c521450e4c51f630835eaf9fa72c44bca5ee0276dc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/customers/components/detail/EntityTagsDialog.tsx
+++ b/packages/core/src/modules/customers/components/detail/EntityTagsDialog.tsx
@@ -397,6 +397,11 @@ export function EntityTagsDialog({
   // without.
   const [activeCategoryHasMore, setActiveCategoryHasMore] = React.useState(false)
   const [activeCategoryLoading, setActiveCategoryLoading] = React.useState(false)
+  // A transport failure is not an end-of-list signal: it leaves the affordance
+  // in place and turns the next click into a retry of the page that failed,
+  // rather than advancing past it.
+  const [activeCategoryLoadFailed, setActiveCategoryLoadFailed] = React.useState(false)
+  const [activeCategoryReloadToken, setActiveCategoryReloadToken] = React.useState(0)
   const creationInFlightRef = React.useRef<string | null>(null)
   const mutationContextId = React.useMemo(
     () => `customer-tags:${entityType}:${entityId}`,
@@ -721,12 +726,14 @@ export function EntityTagsDialog({
     if (!open) return
     setActiveCategoryPage(1)
     setActiveCategoryHasMore(false)
+    setActiveCategoryLoadFailed(false)
   }, [activeCategoryKind, open, searchValue])
 
   React.useEffect(() => {
     if (!open || !activeCategoryKindValue || (activeCategorySource !== 'tags' && activeCategorySource !== 'labels')) {
       setActiveCategoryLoading(false)
       setActiveCategoryHasMore(false)
+      setActiveCategoryLoadFailed(false)
       return
     }
 
@@ -760,21 +767,36 @@ export function EntityTagsDialog({
       }))
 
     setActiveCategoryLoading(true)
-    void apiCall<{ items?: Array<DictEntry | LabelItem> }>(endpoint, {
+    void apiCall<{ items?: Array<DictEntry | LabelItem>; page?: number }>(endpoint, {
       cache: 'no-store',
       headers: { 'x-om-unauthorized-redirect': '0' },
     })
       .then((response) => {
         if (cancelled) return
         if (!response.ok) {
-          setActiveCategoryHasMore(false)
+          setActiveCategoryLoadFailed(true)
           return
         }
+        setActiveCategoryLoadFailed(false)
         const servedEntries = Array.isArray(response.result?.items) ? response.result.items : []
         const fetchedEntries = mapEntries(servedEntries)
+        // The two sources paginate differently. `/api/customers/tags` is a
+        // query-engine list, so a page past the end comes back empty and the
+        // served count alone terminates the sequence. `/api/customers/labels`
+        // clamps the requested page to the last one and re-serves it in full
+        // forever, so short-page termination needs the second half of the
+        // helper's obligation 2: take an echoed page below the one asked for as
+        // the end of the list. Same guard as `AttachmentsSection`, whose
+        // endpoint clamps the same way.
+        const returnedPage =
+          typeof response.result?.page === 'number' ? response.result.page : activeCategoryPage
+        const servedRequestedPage = returnedPage >= activeCategoryPage
         // Measured on what the endpoint served, before `mergeOptions` folds the
         // page into the entries already on screen.
-        setActiveCategoryHasMore(hasMoreFromPage(servedEntries.length, REMOTE_CATEGORY_PAGE_SIZE))
+        setActiveCategoryHasMore(
+          servedRequestedPage && hasMoreFromPage(servedEntries.length, REMOTE_CATEGORY_PAGE_SIZE),
+        )
+        if (!servedRequestedPage) return
         updateCategoryEntries(activeCategoryKindValue, (currentEntries) =>
           activeCategoryPage <= 1
             ? mergeOptions(seedEntries, fetchedEntries)
@@ -783,8 +805,10 @@ export function EntityTagsDialog({
       })
       .catch(() => {
         if (cancelled) return
-        setActiveCategoryHasMore(false)
-        updateCategoryEntries(activeCategoryKindValue, () => seedEntries)
+        setActiveCategoryLoadFailed(true)
+        if (activeCategoryPage <= 1) {
+          updateCategoryEntries(activeCategoryKindValue, () => seedEntries)
+        }
       })
       .finally(() => {
         if (!cancelled) {
@@ -798,6 +822,7 @@ export function EntityTagsDialog({
   }, [
     activeCategoryKindValue,
     activeCategoryPage,
+    activeCategoryReloadToken,
     activeCategorySource,
     entityId,
     entityOrganizationId,
@@ -1269,9 +1294,17 @@ export function EntityTagsDialog({
                               variant="outline"
                               size="sm"
                               className="rounded-lg px-3 text-xs"
-                              onClick={() => setActiveCategoryPage((current) => current + 1)}
+                              onClick={() => {
+                                if (activeCategoryLoadFailed) {
+                                  setActiveCategoryReloadToken((current) => current + 1)
+                                  return
+                                }
+                                setActiveCategoryPage((current) => current + 1)
+                              }}
                             >
-                              {t('customers.activities.loadMore', 'Load more')}
+                              {activeCategoryLoadFailed
+                                ? t('customers.personTags.retry', 'Retry')
+                                : t('customers.activities.loadMore', 'Load more')}
                             </Button>
                           ) : null}
                         </div>
--- a/packages/core/src/modules/customers/components/detail/EntityTagsDialog.tsx
+++ b/packages/core/src/modules/customers/components/detail/EntityTagsDialog.tsx
@@ -6,6 +6,7 @@ import { EmptyState } from '@open-mercato/ui/primitives/empty-state'
 import { VisuallyHidden } from '@radix-ui/react-visually-hidden'
 import { useT } from '@open-mercato/shared/lib/i18n/context'
 import { cn, slugifyTagLabel } from '@open-mercato/shared/lib/utils'
+import { hasMoreFromPage } from '@open-mercato/shared/lib/pagination/load-more'
 import { apiCall, apiCallOrThrow, readApiResultOrThrow } from '@open-mercato/ui/backend/utils/apiCall'
 import { flash } from '@open-mercato/ui/backend/FlashMessages'
 import { useGuardedMutation } from '@open-mercato/ui/backend/injection/useGuardedMutation'
@@ -388,7 +389,13 @@ export function EntityTagsDialog({
   const [creatingKind, setCreatingKind] = React.useState<string | null>(null)
   const [manageTagsOpen, setManageTagsOpen] = React.useState(false)
   const [activeCategoryPage, setActiveCategoryPage] = React.useState(1)
-  const [activeCategoryTotalPages, setActiveCategoryTotalPages] = React.useState(1)
+  // Short-page termination instead of `page < totalPages` — see
+  // `hasMoreFromPage`. Like the page number it belongs to whichever category is
+  // active: the fetch effect below rewrites it on every category, search and
+  // page change, and the reset effect clears it alongside the page so a
+  // category with more entries never leaves the affordance behind on one
+  // without.
+  const [activeCategoryHasMore, setActiveCategoryHasMore] = React.useState(false)
   const [activeCategoryLoading, setActiveCategoryLoading] = React.useState(false)
   const creationInFlightRef = React.useRef<string | null>(null)
   const mutationContextId = React.useMemo(
@@ -713,12 +720,13 @@ export function EntityTagsDialog({
   React.useEffect(() => {
     if (!open) return
     setActiveCategoryPage(1)
+    setActiveCategoryHasMore(false)
   }, [activeCategoryKind, open, searchValue])
 
   React.useEffect(() => {
     if (!open || !activeCategoryKindValue || (activeCategorySource !== 'tags' && activeCategorySource !== 'labels')) {
       setActiveCategoryLoading(false)
-      setActiveCategoryTotalPages(1)
+      setActiveCategoryHasMore(false)
       return
     }
 
@@ -752,16 +760,21 @@ export function EntityTagsDialog({
       }))
 
     setActiveCategoryLoading(true)
-    void apiCall<{ items?: Array<DictEntry | LabelItem>; totalPages?: number }>(endpoint, {
+    void apiCall<{ items?: Array<DictEntry | LabelItem> }>(endpoint, {
       cache: 'no-store',
       headers: { 'x-om-unauthorized-redirect': '0' },
     })
       .then((response) => {
-        if (!response.ok || cancelled) return
-        const fetchedEntries = mapEntries(Array.isArray(response.result?.items) ? response.result.items : [])
-        setActiveCategoryTotalPages(
-          typeof response.result?.totalPages === 'number' ? response.result.totalPages : 1,
-        )
+        if (cancelled) return
+        if (!response.ok) {
+          setActiveCategoryHasMore(false)
+          return
+        }
+        const servedEntries = Array.isArray(response.result?.items) ? response.result.items : []
+        const fetchedEntries = mapEntries(servedEntries)
+        // Measured on what the endpoint served, before `mergeOptions` folds the
+        // page into the entries already on screen.
+        setActiveCategoryHasMore(hasMoreFromPage(servedEntries.length, REMOTE_CATEGORY_PAGE_SIZE))
         updateCategoryEntries(activeCategoryKindValue, (currentEntries) =>
           activeCategoryPage <= 1
             ? mergeOptions(seedEntries, fetchedEntries)
@@ -770,7 +783,7 @@ export function EntityTagsDialog({
       })
       .catch(() => {
         if (cancelled) return
-        setActiveCategoryTotalPages(1)
+        setActiveCategoryHasMore(false)
         updateCategoryEntries(activeCategoryKindValue, () => seedEntries)
       })
       .finally(() => {
@@ -1250,7 +1263,7 @@ export function EntityTagsDialog({
                               {t('customers.personTags.loading', 'Loading...')}
                             </div>
                           ) : null}
-                          {(activeCategory.source === 'tags' || activeCategory.source === 'labels') && activeCategoryPage < activeCategoryTotalPages ? (
+                          {(activeCategory.source === 'tags' || activeCategory.source === 'labels') && activeCategoryHasMore ? (
                             <Button
                               type="button"
                               variant="outline"
--- a/packages/core/src/modules/customers/components/detail/__tests__/EntityTagsDialog.test.tsx
+++ b/packages/core/src/modules/customers/components/detail/__tests__/EntityTagsDialog.test.tsx
@@ -23,6 +23,20 @@ jest.mock('../ManageTagsDialog', () => ({
   ManageTagsDialog: ({ open }: { open: boolean }) => (open ? <div>manage-tags-dialog</div> : null),
 }))
 
+// Mirrors `REMOTE_CATEGORY_PAGE_SIZE` in the dialog. A page that comes back
+// with this many entries is the only "there may be more" signal the load-more
+// affordance reads.
+const REMOTE_CATEGORY_PAGE_SIZE = 50
+
+function makeTagEntries(count: number, prefix = 'filler'): Array<{ id: string; value: string; label: string; color: string | null }> {
+  return Array.from({ length: count }, (_, index) => ({
+    id: `${prefix}-${index + 1}`,
+    value: `${prefix}-${index + 1}`,
+    label: `${prefix === 'filler' ? 'Filler' : prefix} ${index + 1}`,
+    color: null,
+  }))
+}
+
 describe('EntityTagsDialog', () => {
   beforeEach(() => {
     apiCallMock.mockReset()
@@ -41,12 +55,17 @@ describe('EntityTagsDialog', () => {
             },
           })
         }
+        // Page 1 comes back full, page 2 short — the shape a query-engine list
+        // actually emits, and what the dialog now terminates on.
         return Promise.resolve({
           ok: true,
           result: {
             items: page === '2'
               ? [{ id: 'tag-2', value: 'tag-2', label: 'VIP', color: '#f59e0b' }]
-              : [{ id: 'tag-1', value: 'tag-1', label: 'Prospect', color: '#60a5fa' }],
+              : [
+                  { id: 'tag-1', value: 'tag-1', label: 'Prospect', color: '#60a5fa' },
+                  ...makeTagEntries(REMOTE_CATEGORY_PAGE_SIZE - 1),
+                ],
             totalPages: 2,
           },
         })
@@ -203,6 +222,8 @@ describe('EntityTagsDialog', () => {
     await waitFor(() => {
       expect(screen.getByText('VIP')).toBeInTheDocument()
     })
+    // Page 2 was short, so the sequence is over even though `totalPages` said 2.
+    expect(screen.queryByRole('button', { name: 'Load more' })).toBeNull()
 
     fireEvent.change(screen.getByPlaceholderText('Search tags...'), {
       target: { value: 'priority' },
@@ -219,6 +240,92 @@ describe('EntityTagsDialog', () => {
       expect(screen.getByText('Priority')).toBeInTheDocument()
     })
     expect(screen.queryByText('Prospect')).not.toBeInTheDocument()
+  })
+
+  // The affordance used to be gated on `activeCategoryPage < activeCategoryTotalPages`.
+  // When the reported total under-reports the result set — tags created between
+  // two requests, or a capped list count — the button vanished and every tag
+  // past the reported end became unreachable. Termination now follows the page
+  // coming back short.
+  describe('load-more termination', () => {
+    const respondToTagsWith = (respond: (page: string) => { items: unknown[]; totalPages: number }) => {
+      const fallback = apiCallMock.getMockImplementation()!
+      apiCallMock.mockImplementation((url: string, ...rest: unknown[]) => {
+        if (url.startsWith('/api/customers/tags?')) {
+          const page = new URL(url, 'http://localhost').searchParams.get('page') ?? '1'
+          return Promise.resolve({ ok: true, result: respond(page) })
+        }
+        return fallback(url, ...rest)
+      })
+    }
+
+    const renderDialog = async () => {
+      await act(async () => {
+        renderWithProviders(
+          <EntityTagsDialog
+            open
+            onClose={jest.fn()}
+            entityId="person-1"
+            entityType="person"
+            entityOrganizationId="org-1"
+            entityData={{}}
+          />,
+        )
+      })
+    }
+
+    it('offers Load more on a full page even when totalPages reports a single page', async () => {
+      respondToTagsWith(() => ({ items: makeTagEntries(REMOTE_CATEGORY_PAGE_SIZE), totalPages: 1 }))
+
+      await renderDialog()
+
+      expect(await screen.findByRole('button', { name: 'Load more' })).toBeInTheDocument()
+    })
+
+    it('hides Load more once a page comes back short, whatever totalPages promises', async () => {
+      respondToTagsWith(() => ({ items: makeTagEntries(3), totalPages: 99 }))
+
+      await renderDialog()
+
+      expect(await screen.findByText('Filler 1')).toBeInTheDocument()
+      expect(screen.queryByRole('button', { name: 'Load more' })).toBeNull()
+    })
+
+    // `/api/customers/tags` is query-engine backed and computes an unclamped
+    // offset, so the page past the end comes back empty rather than re-serving
+    // the last one. That empty page is what ends the sequence.
+    it('stops on the empty page past the end', async () => {
+      respondToTagsWith((page) =>
+        page === '1'
+          ? { items: makeTagEntries(REMOTE_CATEGORY_PAGE_SIZE), totalPages: 1 }
+          : { items: [], totalPages: 1 })
+
+      await renderDialog()
+
+      fireEvent.click(await screen.findByRole('button', { name: 'Load more' }))
+
+      await waitFor(() => {
+        expect(screen.queryByRole('button', { name: 'Load more' })).toBeNull()
+      })
+      expect(screen.getByText('Filler 50')).toBeInTheDocument()
+    })
+
+    // The flag belongs to the active category exactly as the page number does:
+    // switching to a category whose first page is short must not inherit the
+    // previous category's affordance.
+    it('keeps the affordance keyed to the active category', async () => {
+      respondToTagsWith(() => ({ items: makeTagEntries(REMOTE_CATEGORY_PAGE_SIZE), totalPages: 1 }))
+
+      await renderDialog()
+
+      expect(await screen.findByRole('button', { name: 'Load more' })).toBeInTheDocument()
+
+      fireEvent.click(screen.getByRole('button', { name: /^Labels\b/ }))
+
+      await waitFor(() => {
+        expect(screen.queryByRole('button', { name: 'Load more' })).toBeNull()
+      })
+    })
   })
 
   it('includes tenant-defined custom categories from kind settings', async () => {
--- a/packages/core/src/modules/customers/components/detail/__tests__/EntityTagsDialog.test.tsx
+++ b/packages/core/src/modules/customers/components/detail/__tests__/EntityTagsDialog.test.tsx
@@ -28,13 +28,21 @@ jest.mock('../ManageTagsDialog', () => ({
 // affordance reads.
 const REMOTE_CATEGORY_PAGE_SIZE = 50
 
-function makeTagEntries(count: number, prefix = 'filler'): Array<{ id: string; value: string; label: string; color: string | null }> {
+function makeTagEntries(
+  count: number,
+  prefix = 'filler',
+  labelPrefix = 'Filler',
+): Array<{ id: string; value: string; label: string; color: string | null }> {
   return Array.from({ length: count }, (_, index) => ({
     id: `${prefix}-${index + 1}`,
     value: `${prefix}-${index + 1}`,
-    label: `${prefix === 'filler' ? 'Filler' : prefix} ${index + 1}`,
+    label: `${labelPrefix} ${index + 1}`,
     color: null,
   }))
+}
+
+function makeLabelEntries(count: number) {
+  return makeTagEntries(count, 'label', 'Label')
 }
 
 describe('EntityTagsDialog', () => {
@@ -259,6 +267,28 @@ describe('EntityTagsDialog', () => {
       })
     }
 
+    // The dialog drives two endpoints from one affordance, and they paginate
+    // differently — `/api/customers/labels` is hand-rolled and clamps the
+    // requested page, so it needs its own harness rather than the suite's
+    // default empty-page stub.
+    const respondToLabelsWith = (
+      respond: (page: string) => { ok?: boolean; items?: unknown[]; page?: number; totalPages?: number },
+    ) => {
+      const fallback = apiCallMock.getMockImplementation()!
+      apiCallMock.mockImplementation((url: string, ...rest: unknown[]) => {
+        if (url.startsWith('/api/customers/labels?')) {
+          const page = new URL(url, 'http://localhost').searchParams.get('page') ?? '1'
+          const { ok = true, ...result } = respond(page)
+          return Promise.resolve({ ok, result: { assignedIds: [], ...result } })
+        }
+        return fallback(url, ...rest)
+      })
+    }
+
+    const openLabelsCategory = () => {
+      fireEvent.click(screen.getByRole('button', { name: /^Labels\b/ }))
+    }
+
     const renderDialog = async () => {
       await act(async () => {
         renderWithProviders(
@@ -312,19 +342,112 @@ describe('EntityTagsDialog', () => {
 
     // The flag belongs to the active category exactly as the page number does:
     // switching to a category whose first page is short must not inherit the
-    // previous category's affordance.
+    // previous category's affordance. Labels serves a genuinely short page here
+    // rather than an empty one, so the assertion fails if the reset is dropped
+    // instead of passing because the category had no entries at all.
     it('keeps the affordance keyed to the active category', async () => {
       respondToTagsWith(() => ({ items: makeTagEntries(REMOTE_CATEGORY_PAGE_SIZE), totalPages: 1 }))
+      respondToLabelsWith(() => ({ items: makeLabelEntries(3), page: 1, totalPages: 1 }))
 
       await renderDialog()
 
       expect(await screen.findByRole('button', { name: 'Load more' })).toBeInTheDocument()
 
-      fireEvent.click(screen.getByRole('button', { name: /^Labels\b/ }))
+      openLabelsCategory()
+
+      expect(await screen.findByText('Label 1')).toBeInTheDocument()
+      await waitFor(() => {
+        expect(screen.queryByRole('button', { name: 'Load more' })).toBeNull()
+      })
+    })
+
+    // The inverse direction: a category whose first page is full must gain the
+    // affordance after one whose page was short, so the reset recomputes the
+    // flag rather than merely clearing it.
+    it('re-offers the affordance when the next category has a full first page', async () => {
+      respondToTagsWith(() => ({ items: makeTagEntries(3), totalPages: 1 }))
+      respondToLabelsWith(() => ({ items: makeLabelEntries(REMOTE_CATEGORY_PAGE_SIZE), page: 1, totalPages: 1 }))
+
+      await renderDialog()
+
+      expect(await screen.findByText('Filler 1')).toBeInTheDocument()
+      expect(screen.queryByRole('button', { name: 'Load more' })).toBeNull()
+
+      openLabelsCategory()
+
+      expect(await screen.findByRole('button', { name: 'Load more' })).toBeInTheDocument()
+    })
+
+    it('hides Load more once a labels page comes back short', async () => {
+      respondToLabelsWith((page) =>
+        page === '1'
+          ? { items: makeLabelEntries(REMOTE_CATEGORY_PAGE_SIZE), page: 1, totalPages: 1 }
+          : { items: makeLabelEntries(2), page: 2, totalPages: 2 })
+
+      await renderDialog()
+      openLabelsCategory()
+
+      fireEvent.click(await screen.findByRole('button', { name: 'Load more' }))
 
       await waitFor(() => {
         expect(screen.queryByRole('button', { name: 'Load more' })).toBeNull()
       })
+      expect(screen.getByText('Label 50')).toBeInTheDocument()
+    })
+
+    // `/api/customers/labels` pages an in-memory array and clamps the requested
+    // page to the last one (`api/labels/route.ts`), so a request past the end
+    // re-serves the last page in full — for ever. The served count alone can
+    // never terminate there; the echoed page is what ends the sequence. This is
+    // obligation 2 in the `hasMoreFromPage` docstring.
+    it('terminates when the labels endpoint clamps the page past the end', async () => {
+      const lastPage = makeLabelEntries(REMOTE_CATEGORY_PAGE_SIZE)
+      respondToLabelsWith(() => ({ items: lastPage, page: 1, totalPages: 1 }))
+
+      await renderDialog()
+      openLabelsCategory()
+
+      fireEvent.click(await screen.findByRole('button', { name: 'Load more' }))
+
+      await waitFor(() => {
+        expect(screen.queryByRole('button', { name: 'Load more' })).toBeNull()
+      })
+      expect(screen.getByText('Label 1')).toBeInTheDocument()
+      expect(screen.getByText('Label 50')).toBeInTheDocument()
+    })
+
+    // A transport failure is not an end-of-list signal. Clearing the flag on a
+    // failed fetch would strand the user on a partial list with no way back;
+    // advancing the page on the retry would skip the page that failed.
+    it('keeps a retry affordance when a page fails to load', async () => {
+      respondToLabelsWith((page) =>
+        page === '1'
+          ? { items: makeLabelEntries(REMOTE_CATEGORY_PAGE_SIZE), page: 1, totalPages: 2 }
+          : { ok: false, items: [] })
+
+      await renderDialog()
+      openLabelsCategory()
+
+      fireEvent.click(await screen.findByRole('button', { name: 'Load more' }))
+
+      const retry = await screen.findByRole('button', { name: 'Retry' })
+      expect(retry).toBeInTheDocument()
+      expect(screen.getByText('Label 1')).toBeInTheDocument()
+
+      const requestedPages = () =>
+        apiCallMock.mock.calls
+          .map(([url]: [string]) => url)
+          .filter((url: string) => typeof url === 'string' && url.startsWith('/api/customers/labels?'))
+          .map((url: string) => new URL(url, 'http://localhost').searchParams.get('page'))
+
+      const before = requestedPages().length
+      fireEvent.click(retry)
+
+      await waitFor(() => {
+        expect(requestedPages().length).toBeGreaterThan(before)
+      })
+      // The retry re-requests the page that failed rather than advancing past it.
+      expect(requestedPages().slice(before)).toEqual(['2'])
     })
   })
 
--- a/packages/core/src/modules/customers/components/detail/hooks/__tests__/usePersonTasks.test.tsx
+++ b/packages/core/src/modules/customers/components/detail/hooks/__tests__/usePersonTasks.test.tsx
@@ -1,0 +1,189 @@
+/**
+ * @jest-environment jsdom
+ */
+import { act, renderHook, waitFor } from '@testing-library/react'
+import { usePersonTasks } from '../usePersonTasks'
+import type { TodoLinkSummary } from '../../types'
+
+const readApiResultOrThrowMock = jest.fn()
+
+jest.mock('@open-mercato/ui/backend/utils/apiCall', () => ({
+  apiCallOrThrow: jest.fn(),
+  readApiResultOrThrow: (...args: unknown[]) => readApiResultOrThrowMock(...args),
+  withScopedApiRequestHeaders: jest.fn(),
+}))
+
+jest.mock('@open-mercato/ui/backend/utils/optimisticLock', () => ({
+  buildOptimisticLockHeader: jest.fn(() => ({})),
+}))
+
+const PAGE_SIZE = 3
+const INITIAL_TASKS: TodoLinkSummary[] = []
+
+function makeRows(count: number, offset = 0) {
+  return Array.from({ length: count }, (_, index) => ({
+    id: `link-${offset + index + 1}`,
+    todoId: `todo-${offset + index + 1}`,
+    todoSource: 'customers:interaction',
+    todoTitle: `Task ${offset + index + 1}`,
+    todoIsDone: false,
+    todoPriority: null,
+    todoSeverity: null,
+    todoDescription: null,
+    todoDueAt: null,
+    todoCustomValues: null,
+    todoOrganizationId: 'org-1',
+    organizationId: 'org-1',
+    tenantId: 'tenant-1',
+    createdAt: '2026-01-01T00:00:00.000Z',
+  }))
+}
+
+async function renderTasks() {
+  const rendered = renderHook(() =>
+    usePersonTasks({ entityId: 'person-1', initialTasks: INITIAL_TASKS, pageSize: PAGE_SIZE }))
+  await waitFor(() => {
+    expect(rendered.result.current.isInitialLoading).toBe(false)
+  })
+  return rendered
+}
+
+// The load-more gate used to be `pageInfo.page >= pageInfo.totalPages`, mirrored
+// by `hasMore = page < totalPages`. When the reported total under-reports the
+// result set — tasks linked between two requests, or a capped list count — the
+// affordance disappeared and the remaining tasks became unreachable. Both now
+// terminate on the served page coming back short.
+describe('usePersonTasks load-more termination', () => {
+  beforeEach(() => {
+    readApiResultOrThrowMock.mockReset()
+  })
+
+  it('offers more on a full page even when totalPages reports a single page', async () => {
+    readApiResultOrThrowMock.mockResolvedValue({
+      items: makeRows(PAGE_SIZE),
+      total: 2,
+      page: 1,
+      pageSize: PAGE_SIZE,
+      totalPages: 1,
+    })
+
+    const { result } = await renderTasks()
+
+    expect(result.current.tasks).toHaveLength(PAGE_SIZE)
+    expect(result.current.hasMore).toBe(true)
+  })
+
+  it('stops on a short page however many pages the payload promises', async () => {
+    readApiResultOrThrowMock.mockResolvedValue({
+      items: makeRows(PAGE_SIZE - 1),
+      total: 999,
+      page: 1,
+      pageSize: PAGE_SIZE,
+      totalPages: 99,
+    })
+
+    const { result } = await renderTasks()
+
+    expect(result.current.hasMore).toBe(false)
+  })
+
+  // `/api/customers/todos` is query-engine backed and computes an unclamped
+  // offset, so the page past the end comes back empty rather than re-serving
+  // the last one. That empty page is what ends the sequence.
+  it('stops on the empty page past the end and keeps the rows already loaded', async () => {
+    readApiResultOrThrowMock
+      .mockResolvedValueOnce({
+        items: makeRows(PAGE_SIZE),
+        total: PAGE_SIZE,
+        page: 1,
+        pageSize: PAGE_SIZE,
+        totalPages: 1,
+      })
+      .mockResolvedValueOnce({
+        items: [],
+        total: PAGE_SIZE,
+        page: 2,
+        pageSize: PAGE_SIZE,
+        totalPages: 1,
+      })
+
+    const { result } = await renderTasks()
+    await act(async () => {
+      await result.current.loadMore()
+    })
+
+    expect(readApiResultOrThrowMock).toHaveBeenCalledTimes(2)
+    expect(result.current.hasMore).toBe(false)
+    expect(result.current.tasks).toHaveLength(PAGE_SIZE)
+  })
+
+  // The guard reads what the server served, not what survived `mergeUnique`: a
+  // full page of already-known links must still offer the next one.
+  it('keeps offering more when a full page dedupes away entirely', async () => {
+    readApiResultOrThrowMock.mockResolvedValue({
+      items: makeRows(PAGE_SIZE),
+      total: PAGE_SIZE,
+      page: 1,
+      pageSize: PAGE_SIZE,
+      totalPages: 1,
+    })
+
+    const { result } = await renderTasks()
+    await act(async () => {
+      await result.current.loadMore()
+    })
+
+    expect(readApiResultOrThrowMock).toHaveBeenCalledTimes(2)
+    expect(result.current.tasks).toHaveLength(PAGE_SIZE)
+    expect(result.current.hasMore).toBe(true)
+  })
+
+  it('requests the next page when there is one', async () => {
+    readApiResultOrThrowMock
+      .mockResolvedValueOnce({
+        items: makeRows(PAGE_SIZE),
+        total: PAGE_SIZE,
+        page: 1,
+        pageSize: PAGE_SIZE,
+        totalPages: 1,
+      })
+      .mockResolvedValueOnce({
+        items: makeRows(1, PAGE_SIZE),
+        total: PAGE_SIZE + 1,
+        page: 2,
+        pageSize: PAGE_SIZE,
+        totalPages: 1,
+      })
+
+    const { result } = await renderTasks()
+    await act(async () => {
+      await result.current.loadMore()
+    })
+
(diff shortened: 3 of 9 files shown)


adeptofvoltron:
🏷️ label rationale

- 🐛 `bug` — this is a defect, not an enhancement: an under-reporting `totalPages` removes the "Load more" button while rows past the reported end still exist, so real data becomes unreachable in the UI.
- 🔍 `review` — the PR is ready and non-draft, and is in no other pipeline state, so it carries the default pipeline label.
- 🔵 `priority-medium` — user-visible data loss in the interface argues for urgency, but it is bounded: it only bites past the first page of two dialogs, and only when the reported total under-reports (rows created between two requests today; routinely once `OM_LIST_COUNT_CAP` lands). Not `high`, because nothing is corrupted and a page refresh re-serves the first page correctly.
- 🟢 `risk-low` — the blast radius is two client-side components in one module. No database structure, API surface, event id, or other `BACKWARD_COMPATIBILITY.md` contract is touched, and both conversions ship with regression tests that were verified to fail against the unpatched source.
- 🧪 `needs-qa` — the automated-verification exemption does not apply: both changed files render UI (`.tsx` outside tests, plus a hook driving one), so manual QA is required despite the automated tests in this PR.
- 🚧 `in-progress` — the chain lock, handed off from issue #5347 to this PR; it is released when the run reports.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: both `page < totalPages` load-more gates from #5347 are converted to `hasMoreFromPage` short-page termination, each measured on the served count, and both surfaces gained the regression harness the issue asked for (verified to fail 10/17 tests against the unpatched source). Targeted suites and the repo-wide guards this diff touches pass locally; the turbo half of the validation gate could not run in the linked worktree (no `node_modules`) and is left to CI, as detailed in the PR body. Lock released — `ci-monitoring` remains only while the CI result is watched.

📸 UI verification could not run here: the customers detail dialogs need a booted app, and this worktree has no installed dependencies or database to drive one. The PR keeps `needs-qa` accordingly — the changes render UI, so the automated-verification exemption does not apply and manual QA still owns the verdict labels.

adeptofvoltron:
✅ CI is green — every check passes: `test` (6m30s, the full unit suite including both new load-more harnesses), `prepare`, `docker-build`, `ephemeral-integration` (13m34s), `lint`, `ds-lint`, `scope`, `audit-scope`, CodeQL and the CLA.

That also settles the two gate commands the linked worktree could not run and which the PR body flagged as unverified locally: `prepare` covers `yarn build:packages` / `yarn generate` / `yarn typecheck`, and it passed — confirming the `TS2307` on `@open-mercato/shared/lib/pagination/load-more` was stale local build output (the parent checkout’s `packages/shared/dist` predates #5274) rather than anything wrong with the import, exactly as predicted. `lint` covers the ESLint run that could not resolve its config locally.

`ci-monitoring` removed; nothing further is owed by this run. The PR remains in `review` with `needs-qa`, so manual QA still owns the merge gate.

adeptofvoltron:
🤖 `om-auto-qa-pr` started by @adeptofvoltron at 2026-08-19T09:36:25Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

adeptofvoltron:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ PASS
**Environment:** `http://localhost:3000` (Next dev, disposable QA database) · role `admin` · browser `playwright` (Chromium 1.62.1)
**Verified:** `fix/issue-5347-load-more-short-page` @ `a63c529f`

Both converted gates were driven in a real browser against seeded data sized to hit the exact page boundaries the change is about: a person with **exactly 20** tasks (a full first page at the hook's default `pageSize` of 20), a person with **25** tasks (full page then a short one), a person with **5** tasks (short first page), and **54** tenant tags against the dialog's `REMOTE_CATEGORY_PAGE_SIZE` of 50 (full page then a 4-entry short one).

### Scenario A (P1 — customers person detail → Tasks, `usePersonTasks` legacy path)
**Where to click:** `/backend/customers/people-v2/<personId>` → tab **Tasks**

| 1 | Open Tasks on a person with 5 tasks | Short first page ends the sequence immediately, so no affordance is drawn | 5 of 5 task rows rendered and the "Load more" button is absent from the DOM (count 0) | ✅ |
| 2 | Open Tasks on a person with 25 tasks | Page 1 comes back full (20 of 20), so the next page is offered | Exactly 20 of the 25 rows rendered and "Load more" is present (count 1); the route reached an interactive task list in 2828 ms | ✅ |
| 3 | Advance to the next page | Page 2 serves 5 rows — short — so the sequence terminates with every row on screen | All 25 rows are on screen with no duplicates (the `mergeUnique` dedupe left the count exact) and "Load more" is gone; the request fired through the section's own `IntersectionObserver` sentinel as it scrolled into view, which calls the same `loadMore` the button does | ✅ |
| 4 | Open Tasks on a person with **exactly 20** tasks | With `total: 20` / `totalPages: 1` from the API, the retired `page < totalPages` gate would have drawn nothing; short-page termination offers the next page because page 1 came back full | The 20 rows rendered and "Load more" is present (count 1) — the behaviour change is visible exactly where the old bound hid the affordance | ✅ |
| 5 | Advance past the end | The empty page ends the sequence without discarding what is already loaded | The page past the end came back empty, "Load more" disappeared, all 20 loaded rows were retained and no error or flash text appeared (0 matches for `/failed|error/i`) | ✅ |
| 11 | Repeat step 2 at a 390×844 mobile viewport | The list and the affordance stay usable on a narrow screen | The Tasks tab (nested under **Activity** at this width) rendered in 3183 ms with 20 rows and 1 "Load more" button, no horizontal overflow | ✅ |

### Scenario B (P1 — customers person detail → **Edit tags** dialog, `EntityTagsDialog`)
**Where to click:** `/backend/customers/people-v2/<personId>` → **Edit tags** → category **Tags**

| 6 | Open the Tags category with 54 tags in the tenant | A full 50-entry page offers the next one | 50 chips rendered and "Load more" is present (count 1) | ✅ |
| 7 | Click "Load more" | Page 2 serves 4 entries — short — so the sequence terminates with all 54 on screen | All 54 tags are on screen (QA Filler 01–40 complete plus the 14 pre-existing tags) and the affordance is gone; only the "New tag" create chip remains below the list | ✅ |
| 8 | Switch category **Tags → Labels** (this user has 0 labels) | The flag is keyed to the active category, not one global boolean, so it must not follow the switch | "Load more" is absent on Labels (count 0) — no affordance leaked from the 50-chip Tags page onto a category with nothing to page through | ✅ |
| 9 | Switch back **Labels → Tags** | The reset effect clears page and flag together, so Tags restarts at page 1 with the affordance restored | 50 chips again and "Load more" is present (count 1) | ✅ |
| 10 | Type `QA Filler 1` into the category search | The search resets the sequence, so a flag left over from the full page must not survive | The panel re-fetched page 1, the endpoint answered with an empty set, the panel shows "No options match the current search." (0 chips) and "Load more" is absent (count 0) | ✅ |

### Screenshots

**01-tasks-short-first-page**



**02-tasks-full-page-load-more**



**03-tasks-terminated-after-short-page**



**04-tasks-exact-page-boundary**



**05-tasks-empty-page-terminates**



**06-tags-full-page-load-more**



**07-tags-terminated-after-short-page**



**08-tags-category-switch-clears-affordance**



**09-tags-category-switch-back**



**10-tags-search-resets**



**11-tasks-mobile-viewport**



### Notes for QA

- **What the fix looks like in the product.** Step 4 is the visible behaviour change: on a person whose task count lands exactly on the page boundary the API truthfully reports `totalPages: 1`, the old `page < totalPages` gate rendered no button, and the new short-page rule offers one more fetch. The cost of that rule is the extra request in step 5 — a full last page always buys one empty page before the affordance disappears. That is the same trade-off the six surfaces converted in #5274 already carry, it costs one request against a page-sized result set, and no rows flicker or disappear when it happens.
- **The under-reporting case itself was not reproducible through the UI on this build.** `/api/customers/todos` derives `total` from a per-page window (`page * pageSize + pageSize`), so the reported total is only ever short while a larger window is still to come, and `/api/customers/tags` reports a true count; `OM_LIST_COUNT_CAP` has not landed yet. The stranded-rows scenario the issue describes therefore stays covered by the PR's own unit tests, and what this run verifies is that the replacement rule behaves correctly on real data at every page boundary.
- **The Tasks section also auto-loads.** An `IntersectionObserver` sentinel sits directly above the button and calls the same `loadMore`, so in steps 3 and 5 the next page was fetched as the sentinel scrolled into view rather than by the click. Both paths run the changed guard; a reviewer repeating this by hand may see the page advance before they press the button.
- **Out of scope, but worth knowing:** the category search in the tags dialog returns nothing for substrings that clearly exist — `search=Filler` and `search=QA Filler 1` both return `total: 0` from `/api/customers/tags` although 40 tags are named `QA Filler NN`, while `search=Q` returns all 54 rows unfiltered (including `manufacturing`). That is server-side list-filter behaviour this PR does not touch — the diff changes only how the response's page length is read — and it is unchanged by the fix, but it makes step 10 look emptier than a reviewer might expect.
- **Screenshot artifact:** the full-page captures render the sticky sidebar and top bar twice on tall pages. That is how Playwright stitches a full-page image of a sticky layout, not a UI defect.
- Fixtures were created through the API under the demo `admin` role and removed at the end of the run; no credentials, tokens or real user data appear in any screenshot.

adeptofvoltron:

This change ships thorough jest/RTL coverage (`usePersonTasks.test.tsx`, the `load-more termination` block in `EntityTagsDialog.test.tsx`), but no browser-level test — nothing under a module-local `__integration__/` folder exercises either affordance against a running app, so the page-boundary behaviour verified manually above is not locked in. Running `/om-integration-tests` with the scenario below would close that gap; it is a follow-up, not a merge blocker for this PR.

1. **Setup** — through the API, as the integration fixtures do:
   - `POST /api/customers/people` → person **A**; then `POST /api/customers/todos` × **20** with `{entityId: A, title: 'Exact20 task NN'}` (exactly one full page at the hook's default `pageSize` of 20, which makes the endpoint report `total: 20` / `totalPages: 1`).
   - `POST /api/customers/people` → person **B**; then `POST /api/customers/todos` × **25** (a full page followed by a short one).
   - `POST /api/customers/tags` × **N** so the tenant holds **more than 50** tags (`REMOTE_CATEGORY_PAGE_SIZE`), each with a distinct `label`/`slug`; record the ids.
2. **Act / assert — tasks (`/backend/customers/people-v2/<id>` → tab *Tasks*):**
   - Person **B**: expect 20 task rows and one `Load more` button; advance the sequence and expect all 25 rows with the button gone. Note the section's `IntersectionObserver` sentinel can fire `loadMore` on scroll, so assert on the resulting state rather than on the click alone.
   - Person **A**: expect the `Load more` button to be present even though the endpoint reports `totalPages: 1` — this is the assertion that pins the retired `page < totalPages` gate — then advance once and expect the button gone, all 20 rows still rendered, and no error surface.
3. **Act / assert — tags dialog (same page → *Edit tags* → category *Tags*):**
   - Expect exactly 50 chips and one `Load more`; click it and expect the full tenant tag count with the affordance gone.
   - With `Load more` present on *Tags*, switch to *Labels* and expect no `Load more` there, then switch back and expect 50 chips and the button again — this pins the flag to the active category rather than to the dialog.
4. **Teardown** — delete every created todo (`DELETE /api/customers/todos?id=`), tag (`DELETE /api/customers/tags?id=`) and person (`DELETE /api/customers/people?id=`) in a `finally`, so the suite stays self-contained and re-runnable.

adeptofvoltron:
🏷️ label rationale — `om-auto-qa-pr`

- 📸 `screenshots` — applied: this run attached UI QA visual evidence to the PR (11 full-page captures of both converted load-more surfaces), and the label records that fact for anyone scanning the queue. It is informational and does not gate merge.
- 🧪 `needs-qa` — left in place: the QA verdict labels are deliberately untouched. This run executed under the evidence-only default, so it produced the evidence a QA reviewer needs but did not sign the PR off; `qa-approved` and `qa-self-verified` remain for a maintainer (or an explicit `--self-qa-signoff` run) to apply, and the PR stays unmergeable until then under the repo's QA-approval merge gate.
- ❌ `qa-failed` — not applied, and correctly so: every one of the 11 scenario steps passed, so there is no failure to record.
- 🔍 `review` / 🐛 `bug` / `priority-medium` / `risk-low` — unchanged: this skill verifies the UI and does not re-assess the pipeline state, category, urgency or blast radius set when the PR was opened.

adeptofvoltron:

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-19T11:19:23Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(customers): terminate the last two detail load-more gates on a short page

## 🎯 Summary

This PR closes out #5347 by retiring the last two `page < totalPages` load-more gates under `customers/components/detail`: the `usePersonTasks` hook that backs `TasksSection`'s legacy path, and the remote-category affordance in `EntityTagsDialog`. Both are converted to `hasMoreFromPage` short-page termination, the shared helper #5274 introduced, and both are measured on what the server served rather than on what survived client-side merging — which is obligation 1 in that helper's docstring and was a real defect on the predecessor PR. The work is careful in the places that usually go wrong: `usePersonTasks` measures `mapped.length` before `mergeUnique` dedupes, `EntityTagsDialog` measures the raw `items` array before `mergeOptions` folds it in, `totalPages` is deliberately dropped from the local response type with a comment recording why, `total` is correctly kept for the count badge, and the per-category reset is handled in both the reset effect and the fetch effect. The new test files are substantial — six tests for the hook, four for the dialog — and the pre-existing tag fixture was corrected to a shape the endpoint can actually emit. The PR body is unusually honest about what it could and could not verify locally.

The problem is that the change applies short-page termination to one endpoint that does not satisfy the helper's second obligation. `EntityTagsDialog` drives two different endpoints from one flag, and the PR's central justification — "both are query-engine backed and compute an unclamped offset" — is true of `/api/customers/tags` but not of `/api/customers/labels`, which is a hand-rolled route that clamps the requested page to the last page. On that source the new gate never terminates. The tests did not catch it because the labels endpoint is stubbed out with an empty page throughout the suite, so every new termination test in fact exercises the tags shape only.

Reviewed scope: the four changed files, plus the three endpoints they page against (`/api/customers/todos`, `/api/customers/tags`, `/api/customers/labels`), `paginateTodoRows`, and the `hasMoreFromPage` helper contract.

## Verdict

❌ **request changes** — the conversion is correct for `usePersonTasks` and for the dialog's `tags` source, but for the dialog's `labels` source it replaces a working gate with one that can never terminate, and the test suite has a blind spot in exactly that place. Both are fixable inside this PR without changing its approach.

## 🧪 Validation Gate

**Validation source: GitHub checks** (head `a63c529f`, every check settled and green) **+ one narrow local fallback** for the single gate command CI does not run.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | GitHub check `prepare` (3m21s), `ci.yml:170` |
| `yarn generate` | ✅ PASS | GitHub check `prepare`, `ci.yml:178` |
| `yarn build:packages` (rebuild after generate) | ✅ PASS | GitHub check `prepare`, `ci.yml:181` |
| `yarn i18n:check-sync` | ✅ PASS | **Local fallback** — no CI job runs this command, so it was executed against the PR head in an isolated worktree: "All translation files are in sync" across 5 locales / 55 modules. The diff touches no locale file, so this was expected. |
| `yarn i18n:check-usage` | ✅ PASS | GitHub check `test`, `ci.yml:596` |
| `yarn typecheck` | ✅ PASS | GitHub check `test`, `ci.yml:614` |
| `yarn test` | ✅ PASS | GitHub check `test` (6m30s), `ci.yml:626` — includes both new suites |
| `yarn build:app` | ✅ PASS | GitHub checks `prepare` / `docker-build` (3m52s), `ci.yml:193` |

Also green on this head and not part of the configured gate: `lint`, `ds-lint`, `scope`, `audit-scope`, `ephemeral-integration` (13m34s), CodeQL (all three analyses) and the CLA. No check is pending and none failed, so there is no CI follow-up owed on this PR. This settles the gate commands the PR body flagged as unverifiable in its own worktree.

## Findings

### ⛔ Blocker

**`packages/core/src/modules/customers/components/detail/EntityTagsDialog.tsx:777` — short-page termination is applied to `/api/customers/labels`, an endpoint that clamps the page past the end, so the affordance can never terminate.**

The flag is set from `hasMoreFromPage(servedEntries.length, REMOTE_CATEGORY_PAGE_SIZE)` for **both** remote sources, but the effect above it picks the endpoint by `activeCategorySource`: `/api/customers/tags` for tags, `/api/customers/labels` for labels. Those two endpoints paginate differently.

`/api/customers/tags` is a `makeCrudRoute` list backed by the query engine, so an out-of-range page really does come back empty — the PR's premise holds there. `/api/customers/labels` is a hand-rolled route that decrypts and sorts in memory and then pages the array itself (`packages/core/src/modules/customers/api/labels/route.ts:148-152`):

```ts
const totalPages = Math.max(1, Math.ceil(total / query.pageSize))
const page = Math.min(query.page, totalPages)   // ← clamps to the last page
const start = (page - 1) * query.pageSize
const items = filteredLabels.slice(start, start + query.pageSize)
```

Because the requested page is clamped, requesting a page past the end re-serves the **last page in full**, forever. This is precisely the case obligation 2 in the `hasMoreFromPage` docstring warns about (`packages/shared/src/lib/pagination/load-more.ts:21-26`): *"An endpoint that instead clamps the requested page to the last page re-serves that page in full, forever; a caller of one must also check it was served the page it asked for before appending. See `AttachmentsSection`, whose endpoint clamps."*

Concrete failure: a user whose personal label set is an exact multiple of `REMOTE_CATEGORY_PAGE_SIZE` (50, 100, 150 …). Open the tags dialog, switch to the **Labels** category. Page 1 serves 50 entries, `hasMoreFromPage(50, 50)` is `true`, "Load more" renders. The button's handler is an unconditional `setActiveCategoryPage((current) => current + 1)` (`:1272`), so page 2 is requested; the route clamps it back to page 1 and serves the same 50 entries; `hasMoreFromPage(50, 50)` is `true` again. `mergeOptions` dedupes by `value` (`:232-238`), so nothing new appears on screen. The button therefore never disappears, and every further click fires a request that changes nothing — an affordance permanently promising more entries that do not exist.

This is a regression: the retired gate `activeCategoryPage < activeCategoryTotalPages` read the route's own `totalPages`, which is computed from the same clamp and was correct for this endpoint. The change fixes the tags source and breaks the labels source.

Any of three fixes works; pick one and cover it with a test:

1. **Preferred, and consistent with `AttachmentsSection`** — additionally require that the server served the page that was asked for. The labels route returns the (clamped) `page` in its payload, so `setActiveCategoryHasMore(hasMoreFromPage(servedEntries.length, REMOTE_CATEGORY_PAGE_SIZE) && (response.result?.page ?? activeCategoryPage) === activeCategoryPage)` terminates correctly on both endpoints. This also guards the `updateCategoryEntries` append against re-appending a clamped page.
2. Keep `totalPages` for the `labels` source only and apply short-page termination to `tags` alone — correct, but it leaves the issue's underlying under-reporting problem live on labels.
3. Remove the clamp from `/api/customers/labels` so the page past the end comes back empty like every query-engine list. This is the cleanest end state, but it changes an API's pagination behavior and so needs its own route test and a note about other callers.

The code comment at `:773-776` and the corresponding paragraph in the PR body should be corrected either way — as written they assert something about `/api/customers/labels` that is not true, which is how the gap got past review the first time.

### ⚠️ Major

**`packages/core/src/modules/customers/components/detail/__tests__/EntityTagsDialog.test.tsx:251` — the new termination tests never exercise the labels source, so the blocker above ships green.**

`respondToTagsWith` only intercepts URLs starting with `/api/customers/tags?`; everything else falls through to the suite's default mock, which answers `/api/customers/labels?` with `items: []` unconditionally (`:73-82`). All four `load-more termination` tests therefore assert against the tags endpoint's shape only, and the labels endpoint — the one whose pagination contract the change actually violates — has no coverage at all. Given that this PR's entire subject is "does this affordance terminate", a behavior change on the labels source with no test for the labels source is a coverage gap rather than an oversight of detail.

Add a `respondToLabelsWith` counterpart and at least two cases: a labels category whose first page comes back full and whose second page is short (terminates), and one that models the clamp — page 2 re-serving page 1's 50 entries — asserting that "Load more" disappears rather than persisting. The second case fails against the current implementation and passes against any of the three fixes above, which is exactly what makes it worth writing.

Related, and part of the same fix: `'keeps the affordance keyed to the active category'` (`:316`) cannot fail for the reason it documents. It renders Tags with a full page, clicks over to Labels, and asserts the button is gone — but the button would be gone anyway, because the default labels mock returns zero items. The test passes whether or not the reset effect clears `activeCategoryHasMore`. Give Labels a **full** first page and Tags a **short** one (or vice versa) so the assertion actually pins the per-category reset it is named for; as written it is a tautology guarding the one piece of state the PR body singles out as the tricky part.

### 🔹 Minor

**`packages/core/src/modules/customers/components/detail/EntityTagsDialog.tsx:769-772` — a non-`ok` response now permanently removes the only retry path.**

Previously `if (!response.ok || cancelled) return` left `activeCategoryTotalPages` untouched, so a transient 500 or timeout while loading page 3 left "Load more" rendered and the user could simply click it again. Now the branch clears `activeCategoryHasMore` while leaving the already-loaded entries on screen, so the user is left looking at a partial list with no affordance at all; the only way to resume is to switch category and back, or close and reopen the dialog, both of which reset to page 1 and discard the pages already fetched.

The PR body frames this as avoiding "a stale affordance pointing at a page that failed to load", which is a reasonable instinct, but a retryable affordance is the better trade for a transient failure — this is a network error path, not an end-of-list signal. Suggest leaving the flag untouched on a failed fetch (the page number has not advanced, so a retry re-requests the same page) and surfacing the failure through the existing error/flash path instead. If you would rather keep it explicit, distinguish a definitive end-of-list from a transport failure rather than collapsing both to "no more".

**`packages/core/src/modules/customers/components/detail/hooks/usePersonTasks.ts:197` — the flag is measured against the requested page size rather than the one the server echoed.**

`hasMoreFromPage(mapped.length, pageSize)` reads the hook's `pageSize` prop from the closure. `CustomerTodosResponse` still carries the server's own `pageSize`, and reading `payload.pageSize ?? pageSize` would keep the measurement correct if the endpoint ever serves a different page size than it was asked for. Today `/api/customers/todos` validates with `z.coerce.number().min(1).max(100)`, which rejects an oversized request with a 400 rather than clamping it, and `TasksSection` uses the default 20 (`TasksSection.tsx:133`) — so this is latent, not live. It is still the more robust read, and it is a one-token change: if a future revision clamps server-side instead of rejecting, the symptom would be a "Load more" button that silently never appears, which is hard to trace back here.

### 💅 Nit

**`packages/core/src/modules/customers/components/detail/__tests__/EntityTagsDialog.test.tsx:30-38` — `makeTagEntries`'s label expression is doing two jobs.** The inline `${prefix === 'filler' ? 'Filler' : prefix} ${index + 1}` special-cases the default value in order to capitalize it. A separate optional `labelPrefix` parameter (defaulting to `'Filler'`) alongside the id `prefix` would read more directly, especially once the labels-source fixtures suggested above start using it with a second prefix. Author's call.

### 📥 Inherited

**`om-auto-qa-pr`'s follow-up comment — [add a browser-level test for both affordances](https://github.com/open-mercato/open-mercato/pull/5378#issuecomment-5340531054)** (@adeptofvoltron, 2026-08-19T10:00Z). The QA run recorded that neither surface has a test under a module-local `__integration__/` folder, and sketched a full scenario (person with exactly 20 tasks, person with 25, a tenant with more than 50 tags). It explicitly declares itself "a follow-up, not a merge blocker for this PR", and nothing in the diff addresses it, so it stays open as a follow-up rather than counting toward this verdict. Two notes for whoever picks it up: the scenario's step 3 covers the tags category and only checks that *Labels* shows **no** button, so as written it would not have caught the blocker above — the labels case needs a label count at an exact multiple of 50; and the jest-level cases requested in the Major finding are the cheaper place to pin the clamping behavior, with the browser test reserved for the end-to-end affordance.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `CustomerTodosResponse` is module-local (`usePersonTasks.ts:37`, not exported; the same-named type in `CustomerTodosTable.tsx:44` is a separate local declaration and is untouched), and `activeCategoryTotalPages` was component-local state.
- [x] No function signature changed in a breaking way — `usePersonTasks`'s options and `UsePersonTasksResult` are unchanged, `hasMore` keeps its name, type and meaning; `EntityTagsDialog`'s props are untouched.
- [x] No required type field removed or narrowed on a shared contract — dropping `totalPages` from the local `CustomerTodosResponse` narrows only what this hook reads from the payload; the endpoint still returns the field for every other consumer.
- [x] No HTTP route URL removed or renamed; no method changed — no route file is in the diff.
- [x] No field removed or retyped in an existing response shape — no API response is modified; the change is entirely in what the clients read.
- [x] No database table or column renamed or removed — no entity or migration in the diff.
- [x] No config key renamed and no default changed silently — `REMOTE_CATEGORY_PAGE_SIZE` (50) and the hook's default `pageSize` (20) are unchanged.
- [x] Where a contract had to change: not applicable, no contract surface is touched.

## 🧪 Test Coverage

**What the new tests cover.** `hooks/__tests__/usePersonTasks.test.tsx` is a genuinely good harness for the hook: it pins the "full page with `totalPages: 1` still offers more" case that is the whole point of the change, the inverse "short page with `totalPages: 99` terminates" case, the empty page past the end (asserting both that the sequence ends and that already-loaded rows survive), the dedupe case where a full page collapses to nothing under `mergeUnique` and must still offer the next page — the obligation-1 trap, and the right thing to have tested — plus the next-page request shape and the `loadMore` no-op after termination. The `load-more termination` block in `EntityTagsDialog.test.tsx` mirrors the first four of those for the tags source. The correction to the pre-existing `'loads paged tag options remotely'` fixture is a real improvement: the old one-item first page with `totalPages: 2` was a shape the endpoint cannot emit, and the replacement now also asserts the button disappears after the short second page. The PR's stash-and-rerun check (10 of 17 tests fail against unpatched sources) is the right way to demonstrate these are regression tests rather than characterization tests.

**Gaps.**

1. **The labels source is untested end to end** (the Major finding). Add `respondToLabelsWith` and, at minimum: `it('hides Load more once a labels page comes back short')` and `it('terminates when the labels endpoint clamps the page past the end')`, the latter modelling page 2 re-serving page 1's 50 entries. Both belong in the existing `load-more termination` describe block in `EntityTagsDialog.test.tsx`.
2. **The per-category reset is asserted tautologically** (`:316`) — rewrite it so the target category has a full first page, per the Major finding.
3. **No test covers the non-`ok` branch** at `:769`. Whichever way the Minor finding is resolved, a test asserting the intended behavior after a failed page fetch (button retained for retry, or cleared) would lock the decision in; today both behaviors pass.
4. **No coverage that `total` still feeds the count badge** after `totalPages` was dropped from the hook's response type. The PR body reasons about this correctly and the risk is low, but a single assertion on `totalCount` in the tasks harness would make the "`total` stays, `totalPages` goes" split explicit rather than implicit.

Item 1 is the one that matters for merge: it is the test that would have caught the blocker.

pkarw:

- 🔴 `changes-requested` — applied, replacing `review`: the code review found one blocker (short-page termination applied to `/api/customers/labels`, an endpoint that clamps the page past the end, so the affordance can never terminate) and one major (the labels source has no test coverage, which is why the blocker shipped green). Under this repo's verdict rule any blocker or major forces `changes-requested`, so the PR moves out of the review queue and back to the author.
- 🐛 `bug` — kept: the change is still a defect fix, and the classification is unaffected by the review outcome.
- 🔵 `priority-medium` — kept: the underlying defect strands rows past an under-reported total on two detail surfaces, and the regression this review found is narrower still (it needs a label set that is an exact multiple of 50). Neither is urgent enough for `priority-high`, and both are real user-visible defects rather than `priority-low` polish.
- 🟢 `risk-low` — kept: the blast radius is unchanged by the findings. Two client-side components, no schema, API, or contract surface touched; the blocker is a behavior regression inside the changed lines, not a widening of the change's reach. Note the fix for it may end up touching `packages/core/src/modules/customers/api/labels/route.ts` — if the author takes that option, the risk rating is worth revisiting on the next pass.
- 🧪 `needs-qa` — kept, untouched: both changed files are `.tsx`/UI-rendering, so the automated-verification exemption does not apply. Code-review approval is not QA approval, and this run applied no QA verdict label. The existing `om-auto-qa-pr` PASS evidence predates the findings above and will need re-running against the corrected head.
- 📸 `screenshots` — kept: informational, records the visual evidence the earlier QA run attached; it does not gate merge.

pkarw:

pkarw:

1 blocker, 1 major, 2 minors, 1 nit, plus 1 inherited follow-up carried forward. The blocker: `EntityTagsDialog` drives two endpoints from one `hasMoreFromPage` flag, and `/api/customers/labels` clamps the requested page to the last page (`api/labels/route.ts:150`), so a label set at an exact multiple of 50 re-serves a full page forever and "Load more" never terminates — the case obligation 2 in the helper docstring names. The conversion is correct for `usePersonTasks` and for the dialog's `tags` source.

Validation source: GitHub checks — every check on `a63c529f` is settled and green, including the gate commands the PR body could not run locally. `yarn i18n:check-sync` has no CI job, so it was run as a narrow local fallback against the PR head and passed. No CI follow-up is owed, so no `ci-monitoring`.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-08-19T14:06:42Z.

adeptofvoltron:
🏷️ label rationale — `om-auto-fix-pr`

- 🔍 `review` — applied, replacing `changes-requested`: the blocker and major from @pkarw's review are fixed in `dad5b541` (the labels-endpoint clamp guard, and the labels-source test coverage whose absence let the blocker ship green), along with both minors and the nit. The requested changes have landed, so the PR belongs back in the review queue rather than with its author — but it is deliberately **not** `merge-queue`, because no re-review has approved the new head and CI has not yet settled on it.
- 🐛 `bug` — kept: the change is still a defect fix (an under-reporting total hides rows that exist), and this run did not alter that classification.
- 🔵 `priority-medium` — kept: unchanged by this pass. Real data is unreachable, but only past the first page of two dialogs and only when the reported total under-reports.
- 🟢 `risk-low` — kept: the fixes stay inside the same two client-side components plus their tests and five locale files. No schema, API, or contract surface is touched, and every behaviour change is pinned by a regression test verified to fail against the unpatched source.
- 🧪 `needs-qa` — kept: both changed files are `.tsx` UI-rendering sources, so the automated-verification exemption does not apply. This run did **not** add `qa-approved` or `qa-self-verified` — those are earned by a QA reviewer, and with the QA gate on this PR stays unmergeable until one signs off. See the summary comment for what the existing QA evidence does and does not cover on the new head.
- 📸 `screenshots` — kept: the earlier UI QA run attached 11 full-page captures. They were taken on `a63c529f` and remain valid for the surfaces this commit did not change; the summary comment is explicit about which behaviour they no longer cover.
- 🤖 `in-progress` → `ci-monitoring` — swapped at the end of this run: the work is finished and fully reported, so the PR is no longer claimed, but a CI-result follow-up on the new head is still owed.

adeptofvoltron:

### 🔀 Base merge

`origin/develop` had advanced to `24fef49d4` since this PR was last reviewed, so the base was merged in **first** (`467efbc8f`) — cleanly, no conflicts — before any review or CI work. Everything below therefore judges the real merge result rather than a stale head.

### 🔍 Review findings — how each ask was handled

@pkarw's review (`CHANGES_REQUESTED`, head `a63c529f`) raised 1 blocker, 1 major, 2 minors, 1 nit and carried 1 inherited follow-up. All of the actionable ones are fixed in `dad5b541`.

- ⛔ **Blocker — short-page termination applied to the clamping `/api/customers/labels`.** Fixed, using the review's preferred option and the same guard `AttachmentsSection` already uses against its own clamping endpoint: the dialog now takes an echoed `page` below the one it requested as the end of the list, so `hasMore` is `servedRequestedPage && hasMoreFromPage(...)`. The clamped page is also no longer appended, which stops a re-served page from being merged back in. The misleading code comment that asserted both endpoints compute an unclamped offset is corrected to describe the two sources separately.
- ⚠️ **Major — the new termination tests never exercised the labels source.** Fixed. A `respondToLabelsWith` harness joins `respondToTagsWith`, and the `load-more termination` block gains a labels short-page case, the clamp case (page 2 re-serving page 1's 50 entries), and a failed-page case. The tautological `'keeps the affordance keyed to the active category'` test now serves Labels a genuinely **short** page instead of an empty one, so it can fail for the reason it documents, and it gained an inverse sibling that switches from a short category to a full one and asserts the affordance comes *back*.
- 🔹 **Minor — a non-`ok` response removed the only retry path.** Fixed, but not exactly as suggested. Leaving the flag untouched alone would have introduced a page-skip: at failure time `activeCategoryPage` has already advanced, so a second click would request page 4 after page 3 failed. Instead a transport failure is now tracked explicitly — the affordance stays, relabels to **Retry**, and re-requests *the page that failed* rather than advancing past it. A failed page past the first also no longer discards the pages already on screen. This is the review's own "distinguish a definitive end-of-list from a transport failure" branch.
- 🔹 **Minor — page size measured from the closure rather than the server echo.** Fixed: `usePersonTasks` now reads `payload.pageSize ?? pageSize`, with a test pinning it.
- 💅 **Nit — `makeTagEntries`'s label expression doing two jobs.** Taken: the special-cased capitalization is replaced by an explicit `labelPrefix` parameter, which the new `makeLabelEntries` fixture builds on.
- 📥 **Inherited — browser-level test for both affordances.** Deferred as designed; it declared itself a non-blocker. Filed as **#5404**, including the review's warning that the originally sketched scenario would not have caught this blocker (a labels case needs a label count at an exact multiple of 50).

Two further test-coverage gaps the review listed under 🧪 are also closed: the failed-page branch now has a test, and the tasks harness asserts `total` still feeds the count badge after `totalPages` was dropped from the response type.

### 🧪 Validation

Run against this worktree (a linked worktree without `node_modules`, so the turbo half of the gate cannot run locally; jest and the i18n scripts were invoked directly against the parent repo's binaries):

- ✅ `packages/core` `components/detail` — **71 suites, 303 tests passed**, including both changed test files.
- ✅ `yarn i18n:check-sync` — "All translation files are in sync" across 5 locales / 55 modules, after adding `customers.personTags.retry` to all five.
- ✅ `yarn i18n:check-usage` — the new key is not reported unused; the run's 3797 unused keys are pre-existing and advisory.
- ✅ Typecheck of the changed area — `tsc` over `customers/components/detail` with the workspace paths mapped to sources reports **no error in either changed file**. (The 9 remaining errors are all `TS2307` for `#generated/*` and unbuilt sibling packages — worktree artifacts, identical for untouched files.)

**Verified as regressions:** reverting only `EntityTagsDialog.tsx` fails exactly the two behaviour-changing new tests (the clamp case and the retry case) while the coverage-only additions still pass; reverting only `usePersonTasks.ts` fails the echoed-page-size test. The new tests fail for the reasons they document, not incidentally.

### 📸 UI

**Not re-verified in a browser on this head, and that is a real gap.** The existing QA evidence (✅ PASS, 11 captures) was taken on `a63c529f` and still covers the tasks surface and the tags category, which this commit does not change. It does **not** cover what this commit changes: labels-category termination against the clamping endpoint, and the new Retry affordance. Re-running `om-auto-qa-pr` needs the app booted, which this worktree cannot do without `node_modules`, and the user's primary checkout is off-limits to this run. The behaviour is covered by RTL tests that drive the real component and fail against the unpatched source, but that is not the same as a browser. `needs-qa` stays on precisely so a QA reviewer covers it — the labels category is the part worth their attention, at a label count that is an exact multiple of 50.

### 🚀 Merge readiness

**Not merge-ready yet, and deliberately not labelled as such.** The requested changes have landed, so the PR moves from `changes-requested` back to `review` — but two gates are genuinely open:

1. **Re-review.** No approval exists for `dad5b541`. The blocker fix changes behaviour on a second endpoint and deserves a look.
2. **QA sign-off.** With the QA gate on, `needs-qa` keeps this unmergeable until a reviewer applies `qa-approved`. This run never applies it.

**CI is still running on this head.** Branch protection plus the QA-approval gate hold the actual merge; this run's verdict covers the work, not a green run. Checks still pending at the time of writing: `prepare`, `test`, `docker-build`, `ephemeral-integration`, and the CodeQL analyses. A follow-up comment will report the CI outcome.

Once re-review approves and CI is green, `om-approve-merge-pr 5378` is the next step — this skill never merges.

adeptofvoltron:

✅ **Every required check is green on `dad5b541`** — the head carrying the review fixes. Settled within the wait budget, so this closes out the pending-CI disclosure in the run summary above.

| Check | Result |
|---|---|
| [`test`](https://github.com/open-mercato/open-mercato/actions/runs/32262975773/job/96101678525) | ✅ 7m50s — the full unit suite, including the new labels-source `load-more termination` cases and the two added `usePersonTasks` assertions |
| [`prepare`](https://github.com/open-mercato/open-mercato/actions/runs/32262975773/job/96100450453) | ✅ 3m32s — covers `yarn build:packages`, `yarn generate`, `yarn typecheck`, `yarn build:app` |
| [`ephemeral-integration (none)`](https://github.com/open-mercato/open-mercato/actions/runs/32262975773/job/96101678424) | ✅ ~15m |
| [`docker-build`](https://github.com/open-mercato/open-mercato/actions/runs/32262975773/job/96101678366) | ✅ 4m28s |
| [`lint`](https://github.com/open-mercato/open-mercato/actions/runs/32262975773/job/96100450438) · [`ds-lint`](https://github.com/open-mercato/open-mercato/actions/runs/32262975773/job/96100450653) · [`scope`](https://github.com/open-mercato/open-mercato/actions/runs/32262975785/job/96100454930) · [`audit-scope`](https://github.com/open-mercato/open-mercato/actions/runs/32262975773/job/96100450035) | ✅ |
| CodeQL (actions / javascript-typescript / python) · `license/cla` | ✅ |

No check failed and none was flaky — nothing needed a rerun, so there is no flake follow-up to file. `prepare` passing also independently confirms the gate commands this run could not execute in its worktree, including that the changed files compile against freshly built packages rather than the stale `packages/shared/dist` the local typecheck had to work around.

**This does not make the PR mergeable, and it is not labelled as though it does.** `reviewDecision` is still `CHANGES_REQUESTED` — @pkarw's review stands until a re-review supersedes it — and `needs-qa` is still on with the QA gate enabled. Green CI was only ever one of three gates. The PR sits in `review` awaiting:

1. a **re-review** of `dad5b541`, in particular the clamp guard now applied to the `/api/customers/labels` source; and
2. **QA sign-off**, which this run never self-applies. The labels category at a label count that is an exact multiple of 50 is the case worth a QA reviewer's time — the existing screenshots predate this behaviour, as the summary comment sets out.

`ci-monitoring` removed: the CI follow-up this run owed is now delivered, and no further comment will come from this agent.

adeptofvoltron:
🤖 `om-auto-fix-pr` run complete: the review blocker, major, both minors and the nit are fixed in `dad5b541`, CI is fully green on that head, and #5404 tracks the deferred browser-level test. The PR is **not** merge-ready — it still needs a re-review to supersede the standing `CHANGES_REQUESTED` and a QA sign-off behind the `needs-qa` gate. Lock released; `ci-monitoring` dropped now that the CI result is reported.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-20T01:00:15Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔁 Re-review: fix(customers): terminate the last two detail load-more gates on a short page

## 🎯 Summary

Re-review of head `dad5b541` against my `CHANGES_REQUESTED` review of `a63c529f`. One fix commit (`dad5b541`) plus a clean base merge of `origin/develop` (`467efbc8f`, no conflicts) landed since.

**Every finding from the previous review is resolved, and the two that mattered are resolved in the right way** — not by narrowing the change's scope, but by applying the guard the helper's docstring actually prescribes. The blocker fix mirrors `AttachmentsSection` (`packages/ui/src/backend/detail/AttachmentsSection.tsx:93-103`) line for line rather than inventing a second idiom for the same problem, which is what I hoped for and is the outcome that leaves the codebase with one pattern for clamping endpoints instead of two.

Reviewed scope this pass: the `dad5b541` delta in full, the final state of both changed source files, the three endpoints they page against re-verified against the new guard (`/api/customers/tags`, `/api/customers/labels`, `/api/customers/todos`), `paginateTodoRows`, `makeCrudRoute`'s list response echo, and the five locale files that are new to the diff.

## Verdict

✅ **approve** — the blocker and the major are genuinely fixed and are pinned by tests that fail against the previous head. Three nits remain, all optional and none worth another round trip.

## ✅ Previous findings — disposition

| # | Previous finding | Status |
|---|---|---|
| ⛔ Blocker | Short-page termination applied to `/api/customers/labels`, which clamps | ✅ **Fixed** — option 1, the one I flagged as preferred |
| ⚠️ Major | Termination tests never exercised the labels source | ✅ **Fixed** — `respondToLabelsWith` harness, 3 new cases |
| ⚠️ Major (related) | `'keeps the affordance keyed to the active category'` was tautological | ✅ **Fixed** — plus an inverse case |
| 🔹 Minor | A non-`ok` response permanently removed the only retry path | ✅ **Fixed**, better than suggested |
| 🔹 Minor | `usePersonTasks` measured against the requested, not the served, page size | ✅ **Fixed** |
| 💅 Nit | `makeTagEntries`'s label expression doing two jobs | ✅ **Fixed** |
| 📥 Inherited | Browser-level test for both affordances | ✅ **Tracked** — [#5404](https://github.com/open-mercato/open-mercato/issues/5404), correctly left out of this PR |

### The blocker fix, verified

`EntityTagsDialog.tsx:791-800` now reads:

```ts
const returnedPage =
  typeof response.result?.page === 'number' ? response.result.page : activeCategoryPage
const servedRequestedPage = returnedPage >= activeCategoryPage
setActiveCategoryHasMore(
  servedRequestedPage && hasMoreFromPage(servedEntries.length, REMOTE_CATEGORY_PAGE_SIZE),
)
if (!servedRequestedPage) return
```

I checked this against all three of the shapes it has to survive, because the guard is only as good as what the endpoints actually echo:

- **`/api/customers/labels`** returns the *clamped* `page` (`api/labels/route.ts:150,159`), which is precisely what makes the guard work. Requesting page 2 of a 50-label set returns `page: 1`, `1 >= 2` is false, the flag clears and the append is skipped. The failure I described — a button that promises more forever on a label count at an exact multiple of 50 — is gone.
- **`/api/customers/tags`** is `makeCrudRoute`, which echoes the requested page **unclamped** (`packages/shared/src/lib/crud/factory.ts:1941`, `page: page.page || requestedPage`). So `servedRequestedPage` is always true there and the tags source behaves exactly as it did before this commit — the fix is inert on the source that never needed it, which is the correct blast radius.
- **The missing-table early return** in the labels route (`:166`) emits no `page` at all. The `typeof … === 'number'` fallback takes `activeCategoryPage`, `servedRequestedPage` is true, `items` is empty, and `hasMoreFromPage(0, 50)` terminates. No false positive.

Boundary case walked through: 100 labels at page size 50. Page 2 echoes `page: 2` (`Math.min(2, 2)`), appends, keeps the button; page 3 clamps to 2, terminates. One extra request at the end of the list, identical to `AttachmentsSection`'s behavior and the accepted cost of this idiom.

### The major fix, verified

`respondToLabelsWith` (`EntityTagsDialog.test.tsx:271-286`) captures the current mock implementation as its fallback, so it composes correctly with `respondToTagsWith` when a test needs both — the two tests that switch categories set tags first and labels second, and the labels interceptor takes precedence. The three new labels cases are the right three:

- `'hides Load more once a labels page comes back short'` — the ordinary termination path on the previously-untested source.
- `'terminates when the labels endpoint clamps the page past the end'` — **this is the test that would have caught the blocker.** It models page 2 re-serving page 1's 50 entries and asserts the button disappears; it fails against `a63c529f` and passes here.
- `'keeps a retry affordance when a page fails to load'` — asserts not just that the affordance survives, but via `requestedPages().slice(before)).toEqual(['2'])` that the retry re-requests the page that failed rather than advancing past it. That second assertion is the one that pins the actual contract.

And `'keeps the affordance keyed to the active category'` now serves Labels a genuinely short page (3 entries) instead of an empty one, so it can fail for the reason its name claims; the new `'re-offers the affordance when the next category has a full first page'` covers the inverse direction, so the reset is pinned as *recompute*, not merely *clear*. That pairing is more than I asked for and is the right instinct.

### The minor fixes, verified

The transport-failure fix went further than the "leave the flag untouched" I suggested: `activeCategoryLoadFailed` + `activeCategoryReloadToken` turn the surviving button into an explicit **Retry** that re-requests the failed page, and the `.catch` branch now only resets entries to the seed on page 1 so a mid-sequence failure no longer discards the pages already fetched. The new string routes through `t('customers.personTags.retry', …)` with real entries in all five locales (`en`/`pl`/`es`/`de`/`ko`), correctly alphabetized — `yarn i18n:check-sync` confirms parity.

`usePersonTasks.ts:200` now reads `payload.pageSize ?? pageSize`. I re-confirmed this stays latent rather than changing behavior: `paginateTodoRows` (`lib/todoCompatibility.ts:181-188`) echoes the requested `pageSize` and computes an **unclamped** `start`, so a page past the end comes back empty — the todos source genuinely does not need the echoed-page guard the dialog now has, and the previous review's read of it holds.

## 🧪 Validation Gate

**Validation source: GitHub checks** (head `dad5b541`, every check settled) **+ one narrow local fallback**.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | GitHub check `prepare`, `ci.yml:170` |
| `yarn generate` | ✅ PASS | GitHub check `prepare`, `ci.yml:178` |
| `yarn build:packages` (rebuild after generate) | ✅ PASS | GitHub check `prepare`, `ci.yml:181` |
| `yarn i18n:check-sync` | ✅ PASS | **Local fallback** — no CI job runs this command, and this revision is the first to touch locale files, so it was executed against `dad5b541`: "All translation files are in sync" across 5 locales / 55 modules. |
| `yarn i18n:check-usage` | ✅ PASS | GitHub check `test`, `ci.yml:596` |
| `yarn typecheck` | ✅ PASS | GitHub check `test`, `ci.yml:614` |
| `yarn test` | ✅ PASS | GitHub check `test`, `ci.yml:626` — includes both suites and all 8 new cases |
| `yarn build:app` | ✅ PASS | GitHub checks `prepare` / `docker-build`, `ci.yml:193` |

Also green on this head and outside the configured gate: `lint`, `ds-lint`, `scope`, `audit-scope`, `ephemeral-integration`, CodeQL (all three analyses), the mutation-test workflow, and the CLA. Nothing pending, nothing failing — no CI follow-up is owed.

## Findings

### 💅 Nit

**`EntityTagsDialog.tsx:800` — the clamped page number is not folded back, unlike the idiom it cites.** `AttachmentsSection` follows its guard with `setPage(returnedPage)` (`:102`), so a clamped response leaves the component's page pointer at the real last page. Here `activeCategoryPage` stays one past the end. It is harmless today — the affordance has terminated so nothing re-requests, and every path that could (category switch, search change, dialog close) runs the reset effect back to page 1 — but it is a divergence from the pattern the comment explicitly claims parity with, and the harmlessness rests on that reset coverage rather than on the state being correct. One line if you want the two call sites to stay literally the same shape.

**`EntityTagsDialog.tsx:1297-1310` — the retry button is neither disabled nor relabelled while the retry is in flight.** It keeps reading "Retry" during the refetch, and each further click bumps `activeCategoryReloadToken` again, firing an overlapping request (the previous one is `cancelled`, so there is no state corruption). The `activeCategoryLoading` indicator above it is the only feedback. Not a regression — the old unconditional "Load more" was worse, since repeat clicks advanced the page number — but `disabled={activeCategoryLoading}` would close it cheaply.

**`EntityTagsDialog.tsx:781-784` — a failed *first* page still has no retry path.** The button renders only when `activeCategoryHasMore` is true, and that is false for a fresh category, so `activeCategoryLoadFailed` cannot surface on page 1: the user sees the empty state with no error and no way to retry. This is unchanged from before the PR and outside its stated scope, so it is not a finding against this change — noting it only because the new retry machinery is one condition away from covering it, and whoever picks up #5404 will be standing right next to it.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed — the new state (`activeCategoryLoadFailed`, `activeCategoryReloadToken`) is component-local; `CustomerTodosResponse` remains module-local and unexported.
- [x] No function signature changed in a breaking way — `usePersonTasks`'s options and `UsePersonTasksResult` are untouched; `EntityTagsDialog`'s props are untouched.
- [x] No required type field removed or narrowed on a shared contract — the added `page?: number` on the local `apiCall` type parameter is optional and read defensively.
- [x] No HTTP route added, removed, or renamed; no response shape modified — no route file is in the diff, and both `page` fields the guard reads were already being returned.
- [x] No event, widget spot, DI key, ACL feature, notification ID, or CLI command touched.
- [x] No database table, column, entity, or migration touched.
- [x] No config key renamed and no default changed — `REMOTE_CATEGORY_PAGE_SIZE` (50) and the hook's default `pageSize` (20) are unchanged.
- [x] i18n is additive — one new key (`customers.personTags.retry`) added to all five locales; no key removed or retyped.

Nothing in `BACKWARD_COMPATIBILITY.md`'s protected surfaces is affected.

## 🧪 Test Coverage

The suite is now genuinely commensurate with the change. Both endpoints the dialog drives have their own harness and their own termination cases; the tasks hook covers full-page/short-page/empty-past-the-end/dedupe/request-shape/no-op-after-termination, plus the two cases added this round — `'still reports the server total for the count badge'` (pinning the deliberate "`total` stays, `totalPages` goes" split, gap 4 from the previous review) and `'measures the page against the size the server echoed'`. Every gap I listed last time is closed:

1. Labels source untested → closed, three cases.
2. Per-category reset asserted tautologically → closed, plus the inverse.
3. No coverage of the non-`ok` branch → closed, and it asserts the retry targets the failed page rather than the next one.
4. `total` still feeding the count badge → closed.

The remaining gap is the browser-level test, correctly deferred to #5404 rather than bolted on here. Worth repeating for whoever takes it: the scenario sketched in the QA follow-up only checks that Labels shows *no* button, so as written it would not have caught the original blocker — that case needs a label count at an exact multiple of 50.

## 📋 Note for QA

The existing ✅ PASS QA evidence on this PR was captured against `a63c529f`, **before** the retry affordance and the new `customers.personTags.retry` string existed. `needs-qa` therefore stays on, and the manual-QA instructions posted separately cover the newly added behavior rather than re-running the original pass.

pkarw:

- 🟢 `merge-queue` — applied, replacing `review`: the re-review of `dad5b541` approved the change; the blocker and the major from the previous pass are fixed and pinned by tests that fail against `a63c529f`, and every check is green on this head.
- 🧪 `needs-qa` — kept: the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. The existing PASS evidence was captured against `a63c529f`, before the retry affordance existed, so it does not carry over to this head.
- 🐛 `bug` / 🟡 `priority-medium` / 🟢 `risk-low` / 📸 `screenshots` — unchanged; the re-review found nothing that re-rates the scope.

pkarw:

This PR is approved and requires manual QA (`needs-qa`, no `skip-qa`). It sits in `merge-queue`, but the QA-approval gate holds the merge until `qa-approved` is added. QA reviewer: when you pick it up, swap the labels (remove `merge-queue`, add `qa`), then run the routes below.

**Scope note:** the ✅ PASS evidence already on this PR was captured against `a63c529f`. Head `dad5b541` added the clamped-page guard, the **Retry** affordance and a new locale string, so P0 and P1 below are genuinely new surface — please do not treat the earlier pass as covering them. P2 is a light re-confirmation of what the first pass already exercised.

### P0 — Labels category load-more termination (`EntityTagsDialog`)

This is the regression the fix exists for, and it only reproduces at an exact multiple of the page size.

**Where to click**
- `/backend/customers/people/{id}` → open the tags dialog → switch to the **Labels** category
- `/backend/customers/companies/{id}` → same dialog (both `PersonTagsDialog` and `CompanyTagsDialog` wrap the changed component)

**What to verify**
- Seed the tenant with **exactly 50 personal labels** (the page size), open Labels → "Load more" renders → click it once → the button **disappears** and the visible list is unchanged (no duplicated rows). Before this fix the button stayed forever and every click was a no-op.
- Seed **exactly 100 labels** → click "Load more" once → the second 50 append and the button stays → click again → the button disappears, no duplicates.
- Seed **60 labels** → one click appends the last 10 and the button disappears.
- Seed **fewer than 50 labels** → no "Load more" at all.
- Type into the dialog's search box while on Labels → results reset to page 1 and the affordance recomputes for the filtered set.

**What can go wrong**
- A "Load more" button that never disappears while no new rows appear — the original defect.
- Duplicated label rows after the last click (a clamped page being re-appended).
- The button disappearing one page early, stranding labels the user can no longer reach.
- Cross-tenant/organization leakage: confirm the paged labels belong only to the current organization scope.

### P1 — Retry on a failed page (new affordance)

**Where to click**
- Same tags dialog, **Labels** or **Tags** category, with enough entries that "Load more" renders

**What to verify**
- With DevTools throttling/offline (or by blocking `/api/customers/labels`), click "Load more" so the second page fails → the button stays and its label becomes **Retry**; the entries already on screen are **not** cleared.
- Restore the network and click **Retry** → the request re-fetches **the page that failed** (check the Network tab: `page=2`, not `page=3`) → the rows append and the button reverts to "Load more".
- Switch category away and back while in the failed state → the dialog resets cleanly to page 1 and the button reads "Load more" again.
- Check the **Retry** label in a non-English locale (`pl`, `de`, `es`, `ko` — the string is `customers.personTags.retry`); it must not fall back to raw English or render the key.

**What can go wrong**
- A partial list left with no affordance at all (the pre-fix behavior).
- Retry skipping the failed page, silently dropping 50 entries.
- The button staying stuck on "Retry" after a successful retry.
- An untranslated or key-leaking button label.

### P2 — Tasks section load-more (re-confirmation)

**Where to click**
- `/backend/customers/people/{id}` → **Tasks** section

**What to verify**
- A person with **exactly 20 tasks** (the default page size) → "Load more" renders → one click ends the sequence and the already-loaded tasks survive.
- A person with **25 tasks** → one click appends the last 5 and the button disappears.
- The task **count badge** still shows the server total, unchanged by this PR.

**What can go wrong**
- The count badge going blank or showing the loaded-so-far count instead of the total (`totalPages` was dropped from the hook's response type; `total` was deliberately kept).
- Tasks disappearing after clicking "Load more".

### Pass/fail

pkarw:

Re-review of `dad5b541`. All seven findings from the previous pass are resolved — the blocker (short-page termination on the clamping `/api/customers/labels`) is fixed with the same echoed-page guard `AttachmentsSection` uses, and the major (no labels-source coverage) is closed by a `respondToLabelsWith` harness whose clamp case fails against `a63c529f`. Verified the guard against all three endpoints the two surfaces page: `makeCrudRoute` echoes the requested page unclamped so the tags source is unaffected, the labels route echoes the clamped page so the guard terminates there, and `paginateTodoRows` computes an unclamped offset so the tasks hook correctly needs no guard at all.

0 blockers, 0 majors, 0 minors, 3 nits (all optional: folding the clamped page back into state as `AttachmentsSection` does, disabling the retry button while in flight, and the pre-existing lack of a retry path on a failed first page). No autofix iterations were needed.

Validation source: GitHub checks — every check green on `dad5b541` — plus one narrow local fallback for `yarn i18n:check-sync`, which no CI job runs and which this revision's five locale files made relevant: all translation files in sync.

Moved to `merge-queue` with `needs-qa` retained. The earlier ✅ PASS QA evidence predates the retry affordance, so manual QA instructions scoped to the new behavior are posted above.

PR #4685, opened 2026-07-30
fix(query_index): stop search_tokens unbounded growth — idempotent tokens, token caps, auto-reindex debounce (#4681)

rajanbor:
## Summary

`search_tokens` grew unbounded (reported: 221M rows / ~100 GB overnight for 2,205 records) because token replacement (`DELETE`+`INSERT`) was not idempotent under concurrency, an auto-reindex stampede fired a fresh reindex on every query, and the prefix tokenizer had no caps. This PR closes the feedback loop that caused the runaway growth (issue #4681, root causes A/B/D; root cause C — coverage drift — is already mitigated in the current tree via COUNT-authoritative `refreshCoverageSnapshot` + the `entity_index_coverage` scope unique constraint and duplicate pruning).

## Changes

- **(A) Idempotent token writes** — coalesced UNIQUE index `search_tokens_unique_tuple_idx` on `(entity_type, entity_id, field, token_hash, organization_id, tenant_id)` with NULL scopes folded to a sentinel UUID; migration de-duplicates existing rows first, then builds it. Both INSERT paths (`replaceSearchTokensForRecord`/`-Batch`) now use `ON CONFLICT DO NOTHING`, so overlapping rebuilds can never duplicate rows.
- **(D) Tokenizer caps** — `OM_SEARCH_MAX_FIELD_CHARS` (20000), `OM_SEARCH_MAX_TOKENS_PER_FIELD` (5000), `OM_SEARCH_MAX_TOKENS_PER_RECORD` (20000). `tokenizeText` truncates oversized field text and caps per-field tokens; `buildSearchTokenRows` stops at the per-record cap.
- **(B) Auto-reindex debounce** — `scheduleAutoReindex` dedupes per `(entityType, tenant, organization)` within `OM_QUERY_INDEX_AUTO_REINDEX_DEBOUNCE_MS` (default 30000ms; `0` restores per-query behavior).
- Documented the new env knobs in `apps/mercato/.env.example`, the create-app template `.env.example`, and `apps/docs/.../query-index.mdx`.

## Specification

**Does a spec exist for this feature/module?**
- [x] No (created a new spec)

**Spec file path:** `.ai/specs/2026-07-30-search-tokens-unbounded-growth.md`

## Testing

- `yarn workspace @open-mercato/shared test` — 141 suites / 1516 tests ✓
- `yarn workspace @open-mercato/search test` — 25 suites / 254 tests ✓
[...]

Issue #4681, opened 2026-07-30
search_tokens grows unbounded (221M rows / 100 GB overnight for 2,205 records) — non-idempotent DELETE+INSERT, auto-reindex stampede, self-corrupting coverage stats


a-szczudlo:
**Version:** open-mercato 0.6.6, self-hosted, Postgres 16.

**Summary:** Overnight, our `search_tokens` table grew to **221M rows / ~100 GB** for an entity type with only **2,205 records** (`messages:message`, emails with long bodies). The continuous `INSERT INTO search_tokens` stream starved the DB connection pool and effectively took the whole backend down. Root cause is a feedback loop between four issues.

**Root causes:**
1. **No unique constraint on the token tuple.** `search_tokens` only has `PRIMARY KEY (id)` (random UUID); the three secondary indexes are non-unique. `replaceSearchTokensForRecord` / `replaceSearchTokensForBatch` (`core/src/modules/query_index/lib/search-tokens.ts`) do `DELETE … then INSERT` in one transaction, which is **not idempotent under concurrency** (READ COMMITTED): one job's INSERT can land after another job's DELETE, so both token sets survive. Observed signature: every `(field, token_hash)` for a record duplicated at exact integer multiples — ×2, ×3, … up to ×15.
2. **Auto-reindex stampede.** `QueryEngine.scheduleAutoReindex` (`query_index/lib/engine.ts`, gated by `SCHEDULE_AUTO_REINDEX`, default true) fires a fire-and-forget `query_index.reindex` event on **every query** whenever `entity_index_coverage.base_count != indexed_count`, with no debounce/dedupe/lock. We observed ~10 concurrent full reindexes of the same entity type. The per-record token rebuild in `query_index/subscribers/upsert_one.ts` is also an un-awaited `void (async () => …)()`, so rebuilds for the same record overlap freely.
3. **Coverage stats corrupt themselves, so the loop never converges.** For one `(entity_type, tenant_id)` we found **two** coverage rows and `base_count=4409` vs `indexed_count=2205` (only 2,205 base records exist). Incremental `applyCoverageAdjustments(+1/row)` on each reindex keeps inflating `base_count`, so the "gap" that triggers auto-reindex never closes → self-sustaining infinite loop.
[...]

PR #1887, opened 2026-05-11
feat: new CRM pages filters

haxiorz:
## Summary

Replaces the dual simple/advanced filter UI on the three CRM list pages (People, Companies, Deals) with a single Figma-aligned advanced-filter-tree experience. The filter trigger now lives in the DataTable toolbar next to Search, AI, and Views, and opens an anchored popover with a searchable field picker, page-owned quick presets, locally-stored saved filters, a Kendo-style `[And] [Or]` group toolbar, within-group drag reorder, inline validation, active chips, and a filter-aware empty state. Filters serialize to the existing v2 URL format; legacy v1 flat URLs and saved perspectives stay readable through the existing compatibility paths.

## Changes

- New shared filter-tree library in `packages/shared/src/lib/query/advanced-filter-tree.ts` (tree model, validation, compiler, persisted-shape helpers).
- New UI components under `packages/ui/src/backend/filters/`: `AdvancedFilterPanel`, `ActiveFilterChips`, `FilterEmptyState`, `FilteredEmptyResults`, `QuickFilters`, `FilterFieldPicker`, `filterValidation`. `AdvancedFilterBuilder` rewritten around the tree model with within-group drag-and-drop.
- `DataTable` gains optional `advancedFilter.externalPopover` mode, `activeFilterChips`, and `filterAwareEmptyState` props (all additive; existing callers unaffected).
- `Dialog` gains an optional `elevated` prop that lifts content above popovers (`z-modal-elevated = 55`) so dialogs opened from inside the filter popover are not occluded.
- `Tag` gains a `pink` variant and the matching `--status-pink-*` design tokens (light + dark) are wired through both `apps/mercato/src/app/globals.css` and `packages/create-app/template/src/app/globals.css`.
- Customers AI trigger widget now also injects into Companies and Deals lists at `data-table:*:search-trailing`; gate narrowed from `[customers.people.view, ai_assistant.view]` (AND-evaluated, hid the trigger from companies-only / deals-only viewers) to `[ai_assistant.view]`. Host page routes still enforce per-entity view features.
- Perspective service hardened with `maybeMigrateLegacyFilterValues` for legacy `FilterValues`-shaped settings.
- People, Companies, and Deals list pages updated to own the advanced filter state, render `AdvancedFilterPanel` anchored to the DataTable trigger, and surface page-owned quick presets + chips.
- 9 new integration specs (`TC-CRM-031`, `046`, `047`, `059`–`064`) and ~14 new/updated unit and component test files.
[...]

Issue #2990, opened 2026-06-11
Migrate user-facing $ilike searches to query-engine + search_tokens (encryption-safe) and document the avoid clause


pkarw:
## Follow-up from #2946

PR #2946 (Fixes #2932) hardened the existing `$ilike` list/search endpoints by wrapping every user-supplied term in `escapeLikePattern`. That closes the LIKE-metacharacter injection gap, but the underlying search mechanism still relies on raw `$ilike` predicates against plaintext columns. This issue tracks the structural follow-up so these searches keep working once encryption is enabled.

### Reviewer's request

From @pkarw on the PR ([comment](https://github.com/open-mercato/open-mercato/pull/2946#issuecomment-4677962859)):

> Thanks, as a follow-up, we should migrate all ilike's across the database to use the `query-engine` and the `search_tokens` table that's supporting the encryption maps and encrypted text fields - if you set the encryption map for these fields that are `iliked` it won't work; this is probably still under the radar as these `ilike` based searches are pretty narrowed - but we should fix it for the future + update `AGENTS.md` with the avoid/do-not-use clause for this kind of searches for the future

### Why this matters

`$ilike` runs directly against the stored column value. For any field covered by an encryption map, the stored value is ciphertext, so an `$ilike '%term%'` predicate silently matches nothing — the search breaks the moment encryption is turned on for that field. The `query-engine` + `search_tokens` path already backs encrypted text fields, so migrating these searches makes them encryption-safe and keeps behavior correct. Today's `$ilike` sites are narrow enough that this is under the radar, but it should be fixed before encryption expands.

### Acceptance criteria

- [ ] Inventory all remaining user-facing `$ilike` / `$like` search sites across `packages/**` and `apps/**` (the PR review enumerated `customers`, `catalog`, `sales`, `auth`, `webhooks`, `staff`, `scheduler`, currencies, workflows, checkout, example todos, etc.).
- [ ] Migrate those searches to the `query-engine` + `search_tokens` mechanism so they remain correct when the searched fields are under an encryption map.
- [ ] Update `AGENTS.md` (and any relevant package-level `AGENTS.md`) with an explicit avoid/do-not-use clause for raw `$ilike`-based search on user input, pointing developers at the `query-engine`/`search_tokens` approach instead.
- [ ] Consider folding this into the existing `$ilike` regression guard so new raw-`$ilike` search sites are flagged (see the Low finding in the #2946 review about the guard only covering the inline-template form, not the variable-indirection form).

[...]

PR #5379, opened 2026-08-19
fix(query): keep plaintext base columns on exact ILIKE, not approximate tokens

kamwro:
## Summary

Both query engines rewrite a base-column `like`/`ilike` into a hashed search-token match whenever the entity has `search_tokens` rows — for every column, not just encrypted ones. The rewrite exists because ILIKE against ciphertext cannot match (06b80b15 introduced the mechanism; #1887 extended it to base columns), but on a plaintext column it silently changes the result set.

Tokenization splits on non-alphanumerics and drops fragments shorter than `minTokenLength` (default 3). Searching a document list for `ZK 1/2026` therefore degrades to the tokens `{202, 2026}` and returns **every document from that year** — and a term whose fragments are all short (`ZK`, `ZK /1`) produces no tokens, and the predicate was **silently dropped**: the full list, dressed as a search result. Any deployment with search tokens enabled and a short-fragment numbering scheme (`INV 7/2026`…) hits this in every list search box.

**Opt-in per review:** the whole behavior sits behind `OM_SEARCH_USE_ILIKE_FOR_NON_ENCRYPTED_FIELDS` (default `false` — out of the box nothing changes, including the legacy predicate-skipping on untokenizable terms). Documented in both `.env.example` files. With the flag on, the fix gates the rewrite on the entity's encryption map, resolved once per query via `getEncryptedFieldNames(..., { ignoreRuntimeHealth: true })` — the on-disk question, so a KMS outage keeps encrypted columns on the token path. Plaintext base columns keep exact SQL ILIKE; encrypted columns keep the token path; if the encryption service cannot answer, the previous rewrite-everything behavior is kept (guessing "plaintext" would turn encrypted-column search into an ILIKE-on-ciphertext that matches nothing). Additionally, an encrypted-column search whose term produces no tokens now matches **nothing** instead of everything — the or-group leaf returned `sql`true``, widening the whole disjunction, and the AND path dropped the predicate.

**Search-semantics note (review finding 4):** with the flag on, plaintext-column matching changes in both directions — token matching is prefix-based and word-order independent (`%smith john%` matched `John Smith`), exact ILIKE requires the contiguous substring. That is the point of the flag, stated here so it reads as a decision rather than an accident; `packages/shared/AGENTS.md` and the search-architecture spec (§A.3) now say it too. Joined-column like filters keep the ungated token rewrite, same as `cf:*` and index-doc filters. Performance: an unanchored ILIKE can cost a sequential scan where the token `EXISTS` is index-backed — the flag's doc block states this trade; as one operational data point, a consuming deployment ran exact-ILIKE order-number search on a 1.4M-row table for months (pre-token-arming) at acceptable list latency.

Reviewed by a certified-partner maintainer on the FSH fork first: fullstackhouse/open-mercato#97 (approved).

## Changes

- `packages/core/src/modules/query_index/lib/engine.ts` — `HybridQueryEngine.query()` resolves the entity's encrypted-field set once (only when a base-column search filter exists) and carries it on `SearchRuntime.encryptedFields`; `applyColumnFilter` and `buildBaseFilterExpression` reroute to tokens only for fields in that set; untokenizable terms on encrypted columns fail closed; the resolution `catch` logs a warning before keeping the old behavior.
- `packages/shared/src/lib/query/engine.ts` — the same gate in `BasicQueryEngine.applyFilterOp`, for installs without `query_index` and for the hybrid's fallback delegation.
- `packages/shared/src/lib/search/config.ts` + both `.env.example` files — the `OM_SEARCH_USE_ILIKE_FOR_NON_ENCRYPTED_FIELDS` switch (optional field on `SearchConfig`, default `false`).
- `packages/shared/src/lib/query/__tests__/engine.test.ts` — six new cases: plaintext → exact ILIKE (flag on), encrypted → token path kept, camelCase map entry still recognized, no service with encryption disabled → exact ILIKE, no service with encryption enabled → rewrite kept, flag off (default) → rewrite kept.
- `packages/core/.../hybrid-engine.test.ts` — a HybridQueryEngine suite for the gate: plaintext/encrypted routing, cross-shape name matching, fail-closed vs legacy on untokenizable terms, and the `buildBaseFilterExpression` false/true pin.
- `cf:*`, index-doc and join like-filters keep the token path untouched — custom-field values can be encrypted, and narrowing their behavior deserves its own discussion.

## Specification

- [x] Yes
- [ ] No (created a new spec)

**Spec file path:** `.ai/specs/2026-07-24-search-architecture-clarification-and-evolution.md`

That spec (Draft, pending review) documents the `search_tokens` store and its direct use by list-API routes. I have not edited it from outside while it is under review — its description of query-time token routing should gain this gate ("plaintext base columns read via SQL ILIKE; tokens serve encrypted columns") once/if this merges; happy to fold that in if maintainers prefer it in this PR.

## Testing

- `packages/shared`: `jest src/lib/query/__tests__/engine.test.ts` — 33/33 (27 pre-existing + 6 new gate cases)
- `packages/core`: `jest src/modules/query_index/__tests__/hybrid-engine.test.ts src/modules/query_index/__tests__/search-alias-concurrency.test.ts` — 63/63 (incl. the new hybrid gate suite)
- Verified end-to-end in a consuming standalone application (214-order dataset, 35k token rows): `ZK 1/2026` returned 171 rows before and exactly 1 after; `ZK /1` returned the full list before and 0 after; encrypted-column search unaffected. That application also carries a Playwright regression suite for this behavior in its own CI.

Integration coverage in `.ai/qa/tests/` not added here: the gate is deterministic engine logic fully pinned by the unit suites above, and the user-visible behavior is exercised end-to-end in the consuming app's Playwright tier as described.

## Checklist

- [x] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable) — §A.3 of the search-architecture spec now states the flag and its semantics trade.
- [ ] QA routing set: `skip-qa` for low-risk non-customer-facing changes — otherwise `needs-qa`.

*(The three label boxes are unticked because this account has read-only access upstream — intended labels and rationale are in a comment below for a maintainer to apply.)*

### Design System Compliance
- [x] Empty state handled for list/data pages (`<EmptyState>` or DataTable `emptyState` prop)
- [x] Loading state handled for async pages
- [x] `aria-label` on all icon-only buttons (`<IconButton>`)

*(No UI surface in this diff; the DS boxes are vacuously satisfied.)*

## Linked issues

None open for this defect. Mechanism lineage: 06b80b15 (hashed search index), #1887 (base-column rewrite + predicate drop). Related hardening the same review surfaced, filed separately on our side: gating the token *writer* for plaintext base columns (write amplification on `search_tokens` — cf. #4681, #4685).

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5379"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789722803&installation_model_id=432243&pr_number=5379&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5379&signature=920ead07cba5d3ae6701f2c94eb054dff99e6894146997f537be86c456daccd3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


kamwro:
Metadata intent (this account has read-only access here, so a maintainer needs to apply these):

- **`bug`** — behavior defect in query-time search routing, not a feature.
- **`priority-high`** — every deployment with `search_tokens` populated returns wrong results in every list search box today: short-fragment terms match everything (dropped predicate) or a whole year of documents (over-broad token match).
- **`risk-medium`** — a query-path change, but tightly gated: plaintext base columns revert to their pre-token exact-ILIKE semantics, encrypted columns keep the token path, and `cf:*`/index-doc/join filters are untouched. Covered by 3 new + 85 passing engine tests.
- **`needs-qa`** — the change has a manually exercisable UI surface (every DataTable search box).
- **`partner-request`** — FSH partner contribution (8lines registry); this account cannot run `/label`, so the label is being requested through the partner channel.

Suggested reviewer, by role: whoever owns `query_index` / the search-token store.

QA note: until a maintainer applies the QA labels this PR stays gated; happy to provide self-QA evidence (screenshots of before/after list searches) on request.

pkarw:
🤖 `om-auto-review-pr` claimed this PR for automated review at 2026-08-19T11:21Z. Assignee, `in-progress` label and this comment are the claim signals; the lock is released when the review is posted.

pkarw:
… as a follow up we should research depeeer the nuances of how ilike works and make the search_tokens tokenization to be exchangeable with it 1:1 - search tokens potentially works faster than ilike that in some cases will require fullscan 


Please make this ilike failsafe mechanism optional by env: OM_SEARCH_USE_ILIKE_FOR_NON_ENCRYPTED_FIELDS defaulting to false please

document this switch in the .env.example please

pkarw:
@kamwro please make this failsafe back to ilike switchable by env variables; then we can merge it in - the follow up (with search tokenizer extensions) is ok to be worked on later

pkarw (changes requested):
# 🔍 Code Review

**Verdict: 🔴 Changes requested** — one blocker and three majors. The defect this PR describes is real and precisely diagnosed, and the direction (gate the token rewrite on the entity's encryption map) matches this repo's own ground truth for the mechanism. What needs work is the gate's edges: how it compares field names, and which code paths the new fail-closed branch actually reaches.

Reviewed at head `54d43b66d`. Validation source: **local fallback** — GitHub reports no validation checks for this head (only `license/cla` ✅ and `[code]smith` *skipping*), so there was no CI signal to read. Commands run in an isolated worktree at the PR head, local mode (no compose `app` container): `yarn generate`, `yarn build:packages`, `yarn typecheck`, `yarn lint`, `yarn test`.

## 💥 Blockers

### 1. Encryption-map field names are compared raw, so a camelCase map entry silently un-gates an encrypted column

`packages/core/src/modules/query_index/lib/engine.ts:1796`, `:2436` and `packages/shared/src/lib/query/engine.ts:470` decide the routing with a bare `Set.has(fieldName)` against the names returned by `getEncryptedFieldNames`. Those two name spaces are not the same vocabulary, and this repo already knows it — every other consumer of that call normalizes both sides before comparing:

- `packages/shared/src/lib/query/encrypted-sort.ts:47-61` expands both sides through `fieldNameCandidates()` (raw / snake / camel).
- `packages/shared/src/lib/query/ciphertext-search-warning.ts:74-88` maps both sides through `normalizeColumnName()` — in the *same file* as the new gate's shared-engine half.
- `TenantDataEncryptionService` itself resolves map entries with `findKey()` (`tenantDataEncryptionService.ts:62`), which tries raw, snake_case **and** camelCase. So `{ field: 'displayName' }` is a working, supported encryption-map entry today.

The sharpest illustration is inside the hybrid engine itself: `engine.ts:706` resolves encrypted **sort** fields through the normalizing helper, and 1090 lines later the new encrypted **search** gate does not.

**Failure scenario.** A module (or a downstream app — 168 map entries ship snake_case in this repo, but the contract permits either) declares `{ field: 'displayName' }` for `customers:customer_entity`. `resolveBaseColumn` only ever yields real column names, so `fieldName` is `display_name`. `encryptedFields = {'displayName'}`, `.has('display_name')` → `false` → the gate classifies a ciphertext column as plaintext → the query runs `display_name ILIKE '%avision%'` against ciphertext → **zero rows, silently**, indistinguishable from a genuine no-result. That is exactly issue #2990, which the token routing exists to prevent, reintroduced through the safety gate meant to protect it. Note the failure direction is the unsafe one: the PR body argues correctly that guessing "plaintext" must be avoided, and this is a path that guesses "plaintext" without noticing.

Fix: run both sides through `normalizeColumnName` (or `fieldNameCandidates`) before the set membership test, in all three gate sites, and pin it with a test using a camelCase map entry.

## ⚠️ Majors

### 2. The new fail-closed `false` also fires on paths where the column is not known to be encrypted

`packages/core/src/modules/query_index/lib/engine.ts:2473` and `:1824` return `false` for an untokenizable term, justified by the comment "an encrypted column (the only kind that still reaches this branch)". That premise does not hold on two reachable paths, because both leave `encryptedFields` nullish and the gate's `== null` arm then lets **every** column through:

- **`queryCustomEntity`** builds its `SearchRuntime` at `engine.ts:1903-1909` and never populates `encryptedFields` at all. Its `applyColumnFilter` calls at `:1926` and `:1933` therefore reach the changed branch for plaintext custom-entity columns and doc fields.
- **The encryption-resolution failure path** sets `encryptedFields = null` (`:540`, and `encryptedLikeFields = null` at `shared/.../engine.ts:386`).

**Failure scenario.** A custom-entity list with `search_tokens` present; a user types `ZK` (or any all-short term) into the search box. Before: the predicate was dropped and the full list came back — the bug this PR fixes. After: `WHERE false` — an **empty list**. Both are wrong, and the second is a new, differently-wrong user-visible result introduced by this PR on a path the PR does not claim to change.

The same applies to the error path, where the PR states the previous behavior is preserved: it is preserved for tokenizable terms, but for untokenizable ones the leaf flips from `true`/dropped to `false`. If a KMS/DB hiccup makes the map unreadable, every list search with a short term returns nothing.

Fix: make the fail-closed branch conditional on *knowing* the column is encrypted (`encryptedFields?.has(field) === true`) rather than on the gate having let the filter through, and either resolve `encryptedFields` in `queryCustomEntity` too or leave that path on its previous semantics deliberately and say so in the comment.

### 3. No test coverage for the HybridQueryEngine half of the change

All three new cases (`packages/shared/src/lib/query/__tests__/engine.test.ts:1138-1218`) exercise `BasicQueryEngine`. The hybrid engine is a **separate implementation** — four gate sites and two fail-closed changes in `packages/core` — and it is the engine the application actually resolves (`packages/core/src/modules/query_index/di.ts:21-32` overrides `queryEngine` with it). `hybrid-engine.test.ts` has no case covering base-column like→token routing at all, before or after this PR, so its 58 green tests do not touch a single changed line there. Per `AGENTS.md` ("I added or adjusted tests that cover the change") this needs the mirror of the three shared cases against `HybridQueryEngine`, plus one pinning the `sql\`false\`` behavior and one for the custom-entity path in finding 2.

### 4. Multi-word / word-order search semantics change on every plaintext column, undocumented and untested

`buildSearchTokensSub` (`engine.ts` — `having count(distinct token_hash) >= hashes.length`) requires *all* tokens of the term to be present on that field, and `tokenizeText` normalizes `%`/`_` to spaces and emits prefixes. Token matching is therefore prefix-based and **word-order independent**: `%smith john%` matched a record whose `display_name` is `John Smith`. Exact `ILIKE '%smith john%'` requires the contiguous substring and will not.

So beyond fixing `ZK 1/2026`, this PR changes what every plaintext list search box matches, in both directions (word-order matching lost; mid-word infix matching gained). That is a defensible design call — `.ai/specs/2026-07-24-search-architecture-clarification-and-evolution.md` §A.3 does frame `search_tokens` as the encryption workaround rather than the general index — but it is a user-visible behavior change that the PR description does not mention, no test pins, and no flag can roll back. It needs to be stated explicitly (PR body + spec) and covered by a test that documents the intended semantics, so the next reader does not read it as an accident.

## 🧹 Minors

- **`packages/shared/src/lib/query/engine.ts:377` / `core/.../engine.ts:531` — "no service" means "everything is plaintext", which is the opposite default from "service threw".** A missing resolver yields `new Set()` (gate active, plaintext assumed), a throwing one yields `null` (gate inert, previous behavior). Both DI wirings (`query_index/di.ts:21-32`, `shared/lib/di/container.ts:213`) swallow a failed `tenantEncryptionService` resolution and return `null`, and `bootstrap.ts:215` registers that service inside a `try`. If registration ever fails while `TENANT_DATA_ENCRYPTION` is on, this silently becomes ILIKE-on-ciphertext for every encrypted column. Consider keying the conservative default on `isTenantDataEncryptionEnabled()` rather than on service presence.
- **Per-request uncached `encryption_maps` read.** With `organizationId: null`, `getEncryptedFieldNames` unions in `fetchAllOrganizationFieldNames` (`tenantDataEncryptionService.ts:324-348`), a raw SQL read with no memory/cache layer, on every searched list request. The reasoning for `null` is sound and well argued in the comment; the cost is not free, and `packages/shared/AGENTS.md` notes the sibling dev-warning path skips this in production precisely because "the map lookup costs an uncached read". Worth memoizing per (entity, tenant) for the request, or caching it the way `getMap` is cached.
- **Performance of the new plaintext path.** A token `EXISTS` subquery is index-backed; `ILIKE '%term%'` on a large base table is a sequential scan. Correctness first, agreed — but this is a plan change on the hottest list endpoints and deserves a measurement note in the PR or spec.
- **`applyColumnFilter` and `buildBaseFilterExpression` now disagree on the empty-sources case.** With hashes present but no usable `searchSources`, `buildBaseFilterExpression` falls through to `sql\`false\`` (`:1824`) while `applyColumnFilter` returns `q` unchanged (`:2478`), i.e. drops the predicate and returns the full list. Post-gate that drop only happens for **encrypted** columns, which is the worst case for it. Pre-existing asymmetry, but the PR is editing both branches.
- **Docs/spec not updated.** `.ai/specs/2026-07-24-search-architecture-clarification-and-evolution.md` §A.3 and `packages/shared/AGENTS.md` ("Queries that go through the query engine get this routing automatically") now describe behavior that is conditional. The PR body acknowledges this and offers to fold it in — please do; `AGENTS.md` at minimum, since it is the instruction surface other agents read.
- **Scope gap worth stating.** Joined-column like filters keep the ungated token rewrite (`applyJoinFilterOp`), so the reported over-broad-match defect persists for joined plaintext columns. The PR states this for `cf:*` and index-doc filters but not for joins.

## ✅ What is right

- The diagnosis is exact and independently reproducible from the code: `%`/`_` → space, split on non-alphanumerics, drop fragments under `minTokenLength`, then `count(distinct token_hash) >= n`. `ZK 1/2026` really does reduce to `{202, 2026}` and really does match the whole year.
- `ignoreRuntimeHealth: true` is the correct call and is used exactly as `TenantDataEncryptionService` documents it (`tenantDataEncryptionService.ts:369-388`) — a KMS outage must not reclassify ciphertext as plaintext.
- Resolving the map once per `query()`, behind `searchRuntime.enabled` **and** the presence of a non-`cf:` search filter, keeps the extra read off plain list loads.
- The `sql\`true\`` → `sql\`false\`` fix is right *for the case it was reasoned about*: a widening `true` inside an OR group was a genuine bug. The problem is only the paths it also reaches (finding 2).
- The code comments are unusually good — they explain the *why*, including the deliberate `organizationId: null` choice and its cost. Please keep them, and correct the one claim that no longer holds ("the only kind that still reaches this branch").

## 🧪 Validation

Run in an isolated worktree at `54d43b66d`, local runner (no compose `app` container present).

| Command | Result |
|---|---|
| `yarn generate` | ✅ |
| `yarn build:packages` | ✅ |
| `yarn typecheck` | ✅ 25/25 |
| `yarn lint` | ✅ |
| `yarn test` (`@open-mercato/shared`) | ✅ 1911 tests / 176 suites |
| `yarn test` (`@open-mercato/core`) | ✅ 10078 tests / 1286 suites |
| `yarn test` (workspace) | ⚠️ 4 failures, all in `create-mercato-app` |
| `packages/shared` `src/lib/query/__tests__/engine.test.ts` | ✅ 30/30 (27 pre-existing + 3 new) |
| `packages/core` `hybrid-engine.test.ts` + `search-alias-concurrency.test.ts` | ✅ 58/58 |

The four workspace failures are `standalone template dev wrapper …` cases in `packages/create-app/src/lib/template-dev-log-files.test.ts`. That file and its subjects are outside this 3-file diff, so they are **not** attributable to this PR — flagging them only so the number is not read as a regression here.

Note that no CI validation gate ran on this head at all, so the table above is this reviewer's local evidence, not a substitute for branch protection.

## 🔒 Contracts and security

- `EncryptionResolver` is a file-local, unexported type in both engines, and the added `options?` parameter is optional, so existing three-parameter implementations stay assignable. No `BACKWARD_COMPATIBILITY.md` protected surface is broken: no DB schema, API route, event ID, DI key, ACL feature, or generated file changes.
- No tenant/organization scoping is weakened — the token subquery keeps its tenant and org-scope predicates, and the plaintext ILIKE path inherits the same base-query scoping.
- Security-relevant caveat: findings 1 and 2 both fail *silently*, and finding 1 fails toward reading an encrypted PII column with a plaintext pattern. Neither leaks data — both return too few rows — but both are the kind of failure an operator cannot see without instrumentation.

## 📋 Summary

The core insight is correct and worth landing. Before it can: normalize field names on both sides of the gate (blocker), keep the fail-closed branch off paths where the column is not known to be encrypted (major), cover the hybrid engine with tests (major), and state the search-semantics change explicitly with a test that pins it (major).

The PR is being handed back to @kamwro. Happy to re-review as soon as these are addressed.

*Reviewed by `om-auto-review-pr`. autofix: skipped (not my PR — re-run with `--autofix` to fix it here).*

pkarw:
🏷️ **label rationale** — applied by `om-auto-review-pr`

| Label | Why |
|---|---|
| `changes-requested` | Review posted with one blocker and three majors; this repo requests changes on Medium/minor findings too. Replaces the implicit `review` state. |
| `bug` | Behavior defect in query-time search routing, as the author proposed. |
| `priority-high` | Agreed with the author's read: every deployment with `search_tokens` populated returns wrong results in every list search box today. |
| `risk-high` | **Raised from the author's suggested `risk-medium`.** The blast radius is every `like`/`ilike` list search, in both query engines, in every deployment with tokens enabled — and the review found the change also alters multi-word/word-order matching semantics (finding 4) and reaches the custom-entity and encryption-failure paths the PR did not intend to change (finding 2). Per the repo rule, when priority/risk signals conflict, take the higher and say why. |
| `needs-qa` | The automated-verification exemption does not apply: it requires automated tests for the changed behavior in the same PR, and the HybridQueryEngine half — the engine the app actually resolves — has none (finding 3). The user-visible surface is list-search results, which wants a human clicking through a list with `search_tokens` populated. |
| `partner-request` | Pre-existing, left untouched. |

No `qa`, `qa-approved` or `qa-self-verified` applied — those are earned by manual QA, not by reading a diff.

pkarw:
🤝 **Handing back to @kamwro** — changes requested.

Thanks for a genuinely well-diagnosed bug report; the `ZK 1/2026` → `{202, 2026}` reduction reproduces exactly from the tokenizer, and gating the rewrite on the encryption map is the right shape for the fix.

Four things to pick up, in the order I'd tackle them:

1. **Blocker — normalize field names on both sides of the gate.** `Set.has(fieldName)` compares encryption-map names to column names raw. `encrypted-sort.ts` uses `fieldNameCandidates()` and `ciphertext-search-warning.ts` uses `normalizeColumnName()` for exactly this comparison, and `findKey()` makes `{ field: 'displayName' }` a supported map entry — so a camelCase entry un-gates an encrypted column into ILIKE-on-ciphertext, silently. The hybrid engine normalizes for encrypted *sort* at `engine.ts:706` and not for encrypted *search* at `:1796`.
2. **Major — keep the new `sql\`false\`` off paths where the column is not known to be encrypted.** `queryCustomEntity` never sets `encryptedFields` (`engine.ts:1903`), and the resolution-failure path sets it to `null`, so both let plaintext columns into the fail-closed branch. A short term in a custom-entity list now returns an empty list instead of the full one — a different wrong answer, on a path the PR does not claim to touch. Gating on `encryptedFields?.has(field) === true` rather than on the filter having passed the gate would do it.
3. **Major — mirror the three new tests against `HybridQueryEngine`.** It is a separate implementation and the one `query_index/di.ts` actually registers; `hybrid-engine.test.ts` currently has no base-column like→token case at all, so its 58 green tests do not cover a changed line there.
4. **Major — say out loud that plaintext search semantics change.** The token path required all tokens per field, so `%smith john%` used to match `John Smith`; exact ILIKE will not. Defensible, but it belongs in the PR body, the spec, and a test that pins the intended behavior.

The minors (missing-service vs. threw-service defaults, the uncached `encryption_maps` read per searched request, the `applyColumnFilter` / `buildBaseFilterExpression` empty-sources asymmetry, and the `AGENTS.md`/spec text) are in the review body — happy for any of them to become follow-ups rather than scope here, your call.

On the spec: yes please, fold the gate into `.ai/specs/2026-07-24-search-architecture-clarification-and-evolution.md` §A.3 in this PR. Its Draft status is not a reason to leave it describing behavior this change makes conditional.

Ping me for a re-review whenever you're ready.

pkarw:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED. Lock released. No `ci-monitoring` follow-up: GitHub reports no validation checks for head `54d43b66d` (only `license/cla` and a skipping `[code]smith`), so there is nothing pending to watch — the local gate results are in the review body as this run's evidence only, never as a substitute for branch protection. autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

kamwro:
All four findings taken, in `c8f0ce5e0` (the env switch you asked for — predates this review) and `a407f6b46` (everything else):

1. **(Blocker)** Both sides of the gate now match across name shapes via `fieldNameCandidates`, same as `encrypted-sort.ts` — pinned by a camelCase-map-entry test in both engines.
2. The fail-closed `false` fires only for columns *known* encrypted (set resolved **and** membership); the custom-entity runtime and the resolution-failure path keep legacy semantics, with comments saying so deliberately. The `applyColumnFilter` empty-sources asymmetry is fixed the same way.
3. A HybridQueryEngine suite now covers the gate directly (routing, cross-shape matching, fail-closed vs legacy, the `false`/`true` pin) — shared 33/33, query_index 63/63.
4. The semantics change is stated in the PR body, `packages/shared/AGENTS.md` and spec §A.3 — and it's opt-in now: `OM_SEARCH_USE_ILIKE_FOR_NON_ENCRYPTED_FIELDS` defaults to `false`, so out of the box nothing changes (a new test pins that default).

Minors: the no-service default now keys on `isTenantDataEncryptionEnabled()`; the all-orgs map read is memoized (60s TTL); perf trade + the joins scope gap are stated in the body. Ready for re-review.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-20T00:52:42Z (re-review of `a407f6b4`). Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Re-review

**Verdict: 🟢 Approved** — all four findings from the previous review are genuinely fixed, not papered over. Four nits below, none blocking.

Re-reviewed at head `a407f6b46` (previous review was at `c8f0ce5e0`). Validation source: **GitHub checks** — `test`, `lint`, `ds-lint`, `scope`, `audit-scope`, `prepare` and `license/cla` are ✅ on this head; the 15 `ephemeral-integration` shards are still running and are left to branch protection.

## ✅ Findings from the previous review — all addressed

### 1. (Blocker) Name-shape mismatch in the gate — **fixed**

Both sides now expand through `fieldNameCandidates`: map entries at set-build time (`resolveEncryptedLikeFieldSet`, `shared/lib/query/engine.ts:69-73`) and the query field at lookup time (`isEncryptedLikeField`, `:56-58`). That is the same double-expansion `resolveEncryptedSortFields` already uses (`encrypted-sort.ts:49-59`), so the gate and the sort path now speak one vocabulary.

I verified the vocabulary claim rather than taking it on faith: `resolveBaseColumn` (`query_index/lib/engine.ts:703-707`) is identity when the column exists, so the `fieldName` reaching the gate is the real column name, and a `{ field: 'displayName' }` map entry now matches it. Pinned in both engines — `hybrid-engine.test.ts` "matching across name shapes" and `engine.test.ts` "a camelCase encryption-map entry still routes its column through tokens".

### 2. (Major) Fail-closed `false` on paths where the column is not known encrypted — **fixed**

All four fail-closed sites now require positive knowledge, `encryptedFields != null && isEncryptedLikeField(...)`, instead of inferring it from the gate having let the filter through (`engine.ts:1840`, `:2495`, `:2504`). `queryCustomEntity` is left inert deliberately and the comment at `:1921-1923` now says why — the doc store's fields are not what the base-column encryption map describes — which replaces the claim I flagged as no longer true. The `applyColumnFilter` / `buildBaseFilterExpression` empty-sources asymmetry I listed as a minor is closed the same way.

### 3. (Major) No HybridQueryEngine coverage — **fixed**

`hybrid-engine.test.ts:1641-1734` adds five cases against `HybridQueryEngine` itself: plaintext keeps ILIKE, encrypted routes through tokens across name shapes, untokenizable + known-encrypted yields `false`, nullish set keeps legacy semantics in both directions, and `buildBaseFilterExpression` returns `false` vs `true` on the same input depending only on the set. They are white-box (private methods via `as any`, a hand-rolled builder), which I'd normally push back on — but they pin the exact decision the previous review found broken, at the layer the bug lived in, so the tradeoff earns its keep here.

### 4. (Major) Undocumented semantics change — **fixed, and better than asked**

Gating it behind `OM_SEARCH_USE_ILIKE_FOR_NON_ENCRYPTED_FIELDS` (default `false`) is a stronger answer than documentation alone: out of the box this PR is a no-op, which is the right posture for a change to what every list search box matches. The default is pinned by its own test, and the semantics are stated in the PR body, `packages/shared/AGENTS.md:126-129`, spec §A.3, and both `.env.example` files — including the create-app template mirror that `AGENTS.md` requires.

Minors also taken: the no-service default now keys on `isTenantDataEncryptionEnabled()` (which defaults to `true`, so a swallowed DI failure keeps the token rewrite rather than guessing plaintext), and the all-orgs map read is memoized.

## 🧹 Nits

- **Duplicate import.** `query_index/lib/engine.ts:5` and `:23` both import from `@open-mercato/shared/lib/query/engine`. Merge into one statement.
- **Helper placement.** `isEncryptedLikeField` / `resolveEncryptedLikeFieldSet` / `clearEncryptedLikeFieldsCache` are pure encryption-name helpers with no engine dependency, but they live in the module that also exports `BasicQueryEngine`. `encrypted-sort.ts` (where `fieldNameCandidates` already lives) or a small `encrypted-fields.ts` would be a tighter home and would keep core's hybrid engine from widening what it pulls out of the engine module.
- **The 60s memo adds a staleness window on a previously always-fresh read.** `getEncryptedFieldNames(..., organizationId: null)` unions through `fetchAllOrganizationFieldNames`, which is not behind the service's own map cache, so this memo is new staleness rather than a second layer over existing staleness. For up to 60s after a field starts being encrypted, the gate can still read it as plaintext → ILIKE against ciphertext → zero rows, silently. Self-healing, flag-gated, and honestly documented in the comment — but wiring `TenantDataEncryptionService.invalidateMap` (`tenantDataEncryptionService.ts:350`) to also call `clearEncryptedLikeFieldsCache()` would make the operator's existing "flush the map" lever actually flush everything.
- **The two load-bearing arguments are unpinned.** `ignoreRuntimeHealth: true` and `organizationId: null` are exactly what keeps a KMS outage from reclassifying ciphertext as plaintext, and what keeps a field encrypted by *some* org on the token path. Both are correct today and both are hand-duplicated across two engines with no test asserting either. This is the nit I'd most like to see land: one assertion per engine that the read is called with `(entity, tenantId, null, { ignoreRuntimeHealth: true })` costs a few lines and stops the copies drifting.

## 📌 Two notes, not asks

- **The two engines now disagree on known-encrypted + untokenizable.** Hybrid fails closed to `false`; `BasicQueryEngine` falls through to `applyColumnOp` and runs ILIKE against ciphertext. Both return zero rows, so it is outcome-equivalent, the fallthrough predates this PR, and `HybridQueryEngine` is what DI resolves (`query_index/di.ts`). Flagging it only so the divergence is a known one.
- **Joined columns keep the ungated rewrite** (`applyJoinFilterOp`, runtime built at `:965` without `encryptedFields`), so the over-broad-match defect persists there. Scope is stated in the PR body; no action needed here.

## 🔒 Contracts and security

- `EncryptionResolver`'s new `options?` parameter stays optional in both engines, so existing three-argument implementations remain assignable. The three new exports from `shared/lib/query/engine` are additive. No `BACKWARD_COMPATIBILITY.md` protected surface is touched: no DB schema, API route, event ID, DI key, ACL feature, or generated file.
- Tenant/org scoping is unchanged — the token subquery keeps its predicates and the ILIKE path inherits the base query's scoping. The new memo is keyed `entity|tenantId` and the underlying read takes `organizationId: null` deliberately, so the key is complete and there is no cross-tenant bleed.
- The new env flag defaults to `false` in `resolveSearchConfig`, so upgrading changes nothing until an operator opts in.

## 🧪 Validation

| Check | State |
|---|---|
| `test` | ✅ |
| `lint` | ✅ |
| `ds-lint` | ✅ |
| `scope` / `audit-scope` / `prepare` | ✅ |
| `license/cla` | ✅ |
| `ephemeral-integration` (15 shards) | ⏳ running |
| `docker-build` / `audit` / `[code]smith` | skipped |

Unlike the previous head, CI actually ran the gate here. No local fallback was needed.

## 📋 Summary

The gate now compares names the way the rest of the codebase does, fails closed only where it knows the column is encrypted, is covered in the engine the application actually resolves, and is off by default. That is a complete answer to the previous review. Approving and moving to `merge-queue`; `needs-qa` stays, so the QA-approval gate holds the merge until a QA reviewer signs off.

Nice work on the comments — they explain the *why* including the costs, and the one claim that had gone stale was corrected rather than left.

*Re-reviewed by `om-auto-review-pr`.*

pkarw:
🏷️ **label rationale** — applied by `om-auto-review-pr`

| Label | Why |
|---|---|
| `merge-queue` | Re-review approved: all four prior findings fixed, only nits remain. `needs-qa` stays on, so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. |
| `priority-high` (kept) | Still a release-relevant search-correctness defect touching encrypted-column classification. |
| `risk-high` (kept) | Still the right rating even with the flag defaulting off: the change decides whether a column is treated as ciphertext, and a misread fails silently. |

pkarw:

**Read this first — the change is behind an env flag.** `OM_SEARCH_USE_ILIKE_FOR_NON_ENCRYPTED_FIELDS` defaults to `false`. QA is therefore in two passes: prove nothing changed with the flag off, then prove the fix works with it on. Both passes matter; the first is the regression guard for every existing deployment.

### P0 — Pass A: flag OFF (default). Nothing may change.

**Where to click**
- Backend → Customers list, search box
- Backend → Auth users list, search box
- Backend → Customer accounts list, search box

**What to verify**
- Leave `OM_SEARCH_USE_ILIKE_FOR_NON_ENCRYPTED_FIELDS` unset (or `false`) and restart the app.
- Search a full name that exists (e.g. `avision`) → the same rows come back as on `develop`.
- Search two words in reversed order (e.g. `smith john` for a record named `John Smith`) → still matches, exactly as before. This is the token path's word-order independence; with the flag off it must survive untouched.
- Search a short/partial term (2 characters, below `OM_SEARCH_MIN_LEN`) → the same result as on `develop` (predicate dropped, full list).

**What can go wrong**
- Any difference at all between this build and `develop` with the flag off is a defect — the flag is supposed to make this a no-op.
- An empty list where `develop` returned rows would mean the gate activated when it should not have.

### P0 — Pass B: flag ON, encrypted columns must still be searchable

**Where to click**
- Set `OM_SEARCH_USE_ILIKE_FOR_NON_ENCRYPTED_FIELDS=true`, keep `TENANT_DATA_ENCRYPTION` on, restart.
- Backend → Customers list (email / phone / name columns are the encrypted-at-rest ones)

**What to verify**
- Search an encrypted PII field (customer email, phone) by a value you know exists → the record is still found. Encrypted columns must keep the search-token path; if they fall onto plaintext ILIKE they will return **nothing**, silently.
- Search the same value from a second organization's account → you see only your own org's records. Cross-tenant/cross-org isolation must be unchanged.
- If your fixture data has a customer whose encrypted field is short (2 chars), search it → an empty result is the intended new behavior here (fail closed), not a bug.

**What can go wrong**
- Zero results for an encrypted field that is definitely present → the gate misclassified ciphertext as plaintext. This is the highest-value thing to catch.
- Any record from another organization appearing in the list.

### P1 — Pass B continued: the actual bug being fixed

**Where to click**
- Backend → any list whose searchable column is **plaintext** (e.g. an order/document number, warranty claim number, SKU)

**What to verify**
- Search a document-number-shaped term such as `ZK 1/2026` → only the matching document comes back, **not** every record from 2026. That is the defect this PR fixes.
- Search a mid-word fragment (e.g. `avisi` inside `Avision`) → now matches (exact ILIKE is infix); previously it may not have.
- Search two words in reversed order on a plaintext column (`smith john`) → now does **not** match `John Smith`. This is the intended, documented semantics change, not a regression — confirm it and note it.
- Search a 2-character term on a plaintext column → returns the ILIKE matches rather than the full list.

**What can go wrong**
- `ZK 1/2026` still returning the whole year → the gate did not engage on that column.
- A list that returns nothing at all for any search term → the fail-closed branch reached a plaintext column.

### P1 — Perceived performance (flag ON)

**Where to click**
- The largest list you have data for, cold-load the route, then type in the search box

**What to verify**
- Cold load of the changed list route still shows its loading state promptly (screenshot if you can).
- Typing in the search box stays responsive; results return without a visible stall. Unanchored `ILIKE '%term%'` can force a sequential scan where the token `EXISTS` subquery was index-backed — a noticeable slowdown on a large table is worth reporting even though it does not block.
- Check the same route at a mobile viewport.

**What can go wrong**
- Search that took under a second now taking several seconds on a large table.

### P2 — Config surface

**Where to click**
- `apps/mercato/.env.example` and `packages/create-app/template/.env.example`

**What to verify**
- Both files document `OM_SEARCH_USE_ILIKE_FOR_NON_ENCRYPTED_FIELDS=false` identically (template mirror requirement).
- A scaffolded app from the template starts with the flag present and off.

### Pass/fail

pkarw:

Re-review of `a407f6b46`. All four findings from the previous review (`c8f0ce5e0`) verified fixed — name-shape normalization via `fieldNameCandidates` in both engines, fail-closed narrowed to known-encrypted columns, direct `HybridQueryEngine` coverage, and the semantics change gated behind `OM_SEARCH_USE_ILIKE_FOR_NON_ENCRYPTED_FIELDS` (default `false`). Four nits raised, none blocking; the one worth following up is a test pinning the `ignoreRuntimeHealth: true` / `organizationId: null` arguments in both engines.

Autofix: not run — no actionable findings above nit level.

Validation source: **GitHub checks** (`test`, `lint`, `ds-lint`, `scope`, `audit-scope`, `prepare`, `license/cla` all ✅ on this head; the 15 `ephemeral-integration` shards were still running at review time and are left to branch protection).

Moved to `merge-queue` with `needs-qa` retained — the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Manual QA instructions are in the comment above.

PR #4912, opened 2026-08-03
feat(i18n): add Korean locale support

pkarw:
Supersedes #4007
Tracking plan: .ai/runs/2026-08-03-korean-locale-support.md
Status: complete

## 🎯 Goal
- Add complete Korean (`ko`) locale support on current `develop` as a CLA-compatible replacement for #4007.
- Preserve visible credit to original contributor @moduvoice while incorporating current-base catalog parity and the requested integration coverage.

## What Changed
- Registered `ko` across the shared locale contract, server/client loaders, profile and checkout selectors, and translation-management UI.
- Added synchronized Korean catalogs for all 51 current modules, preserving placeholders and mirroring standalone create-app assets.
- Extended locale validation tooling and focused unit coverage so Korean remains part of catalog parity, sorting, configuration, and UI contracts.
- Added `TC-AUTH-056`, which exercises Korean selection, persistence, rendered auth content, reload behavior, and reset through the real profile dropdown.

## 🧪 Tests
- Full configured local gate passed at `8c52d07a4` in order: `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, and `yarn build:app`.
- `TC-AUTH-056` passed against the repository-managed isolated ephemeral app and database, including the Korean-to-English reset path.
- Korean catalog sync covered 51 modules; placeholder and locale-value audits reported no missing Korean keys.
- The authoritative review/autofix pass returned clean after two follow-up fixes; manual QA instructions and screenshot evidence are attached in the PR discussion.

## 💥 Breaking Changes
- None. The supported-locale contract change is additive.

## Attribution
- This is a clean maintainer-authored reimplementation superseding #4007 because its CLA remains unresolved.
- Korean localization credit remains with @moduvoice, whose original contribution established this work.

## 📋 Progress
See the Progress section in the tracking plan.

PR #5381, opened 2026-08-19
fix(scheduler): manual trigger and execution history 404 on every system-scoped schedule

Frshy:
## Summary

`POST /api/scheduler/trigger` returns **404 for every system-scoped schedule**, for every authenticated user including a super administrator. No system-scoped job can be run manually; only its unattended cron fires it. `GET /api/scheduler/jobs/[id]/executions` carries the identical copy-pasted lookup and 404s the same way.

Both routes applied tenant/org isolation as part of the **lookup** instead of as a check on the loaded row:

```ts
if (auth.tenantId) findFilter.tenantId = auth.tenantId
if (auth.orgId)    findFilter.organizationId = auth.orgId
const schedule = await em.findOne(ScheduledJob, findFilter)
if (!schedule) return 404
// unreachable for any session carrying a tenant:
if (!schedule.tenantId && !schedule.organizationId && !isSuperAdmin) return 403
```

A system-scoped job has `tenant_id IS NULL` / `organization_id IS NULL`, so any session carrying a tenant — effectively all of them — can never match the row. Each route's own super-admin branch is therefore dead code, which is itself the evidence that "super admins may trigger system-scoped jobs" is the intended behaviour.

The same filter also hid a **tenant-scoped** (org-null) schedule from any actor whose session carried an `orgId`, disagreeing with the list endpoint's own visibility model in `api/jobs/buildFilters.ts`.

The fix moves isolation out of the `where` and onto the loaded row — the shape `updateScheduleCommand` and `deleteScheduleCommand` already use.

## Changes

- New pure helper `packages/scheduler/src/modules/scheduler/lib/scheduleAccess.ts` — `resolveScheduleAccess(schedule, actor): 'allowed' | 'not_found' | 'forbidden'`, mirroring `api/jobs/buildFilters.ts` (pure, no DI, its own `__tests__`).
- `api/trigger/route.ts` and `api/jobs/[id]/executions/route.ts` load by `{ id, deletedAt: null }`, then branch on the helper's decision.
- System scope is classified exactly as `ensureCanManageSystemScopedJob` classifies it (`scopeType === 'system' || tenantId == null`), so update, delete, trigger and executions no longer disagree about what a system-scoped row is.
- Unresolved actor tenant fails closed for everyone, super admins included, matching `buildSchedulerJobsFilters` (which returns an unmatchable filter whenever the tenant is falsy). The system-scope branch is evaluated first, so a tenant-less super admin still reaches system-scoped schedules.

**404 vs 403 is deliberate.** Another tenant's or another organization's schedule keeps answering 404; a 403 would confirm the id exists. The previous filter got that much right by accident and the fix does not regress it. Only a system-scoped schedule answers 403, since its existence is a property of the deployment rather than of a tenant.

Super-admin status reads the immutable `auth.isSuperAdmin` flag only — never a role name.

**One divergence deliberately left in place and documented:** organization isolation compares the actor's single `orgId`, while the list endpoint uses the resolved organization scope (`filterIds` — the selected organization plus its descendants). That predates this helper; the lookup being replaced did the same single-id comparison. Widening it here would grant access the previous code did not, so it is documented on the helper rather than changed. Happy to follow up separately if you want the two paths reconciled.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**

n/a — `.ai/specs/AGENTS.md` directs "Skip specs for small bug fixes". The 404-vs-403 disclosure reasoning is recorded in code comments and in this description.

## Testing

Full CI-mirroring gate from `.ai/agentic.config.json`, local mode, all green:

```
yarn build:packages → yarn generate → yarn build:packages → yarn i18n:check-sync
→ yarn i18n:check-usage (advisory) → yarn typecheck → yarn test → yarn build:app
```

- **Unit:** `packages/scheduler` — 382 tests / 20 suites pass, of which 17 are the new `lib/__tests__/scheduleAccess.test.ts`. Follows `api/jobs/__tests__/scope-filter.test.ts` (real function, no mocks) and carries the anti-spoofing case both existing scope tests carry: a role literally named `superadmin` without the `isSuperAdmin` flag gets no system-scope access.
- **Integration:** `TC-SCHED-008.spec.ts`, run via `yarn test:integration:ephemeral --filter TC-SCHED` — 16/16 scheduler specs pass, including the pre-existing TC-SCHED-005.
- **Regression proof:** with the two routes checked out from `develop` and a forced rebuild, TC-SCHED-008 fails exactly as intended — `Error: a super admin must not be told a system-scoped schedule does not exist / Expected: not 404`. It passes on the fixed routes. The existing TC-SCHED-005 creates an *organization*-scoped job, which is why this bug survived.

Two caveats stated rather than papered over:

- The local runs had `QUEUE_STRATEGY=local`, so the super-admin assertions land on the 400 async-required branch rather than a 200. That still proves the 404 is gone — each route's strategy check sits *after* the lookup, which is exactly what the 404 was masking — but the 200 path is unexercised locally.
- Cross-tenant coverage is unit-level only: there is no tenant fixture in `packages/core/src/helpers/integration/`, so a second tenant is not expressible in a self-contained integration test.

## Checklist

- [x] I updated documentation, locales, or generators if the change requires it. — no locale changes needed: the existing `scheduler.error.not_found` and `scheduler.error.access_denied` keys are reused, so all five locales stay in sync (`yarn i18n:check-sync` clean).
- [x] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required). — `TC-SCHED-008.spec.ts` in the module's `__integration__` folder, per `.ai/qa/AGENTS.md` ("Executable Test Rules"); discovery is programmatic, so no config change.
- [ ] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — n/a, see Specification above.
- [ ] QA routing set.

> The three label checkboxes are unticked because the contributing account has **read-only** permission on this repository and cannot apply labels. Intended values, with rationale, are in a follow-up comment for a maintainer to apply.

### Design System Compliance

Not applicable — this change touches no `.tsx`, nothing under `packages/ui/src/`, and no `**/components/**`. The five changed files are two API route handlers, one pure helper, and two test files.

- [x] No hardcoded status colors
- [x] No arbitrary text sizes
- [x] Empty state handled for list/data pages
- [x] Loading state handled for async pages
- [x] `aria-label` on all icon-only buttons
- [x] Uses existing DS components

## Linked issues

None filed upstream — reported internally. Reproduce with: create a schedule with `scopeType: 'system'` as a super admin (`POST /api/scheduler/jobs`), then `POST /api/scheduler/trigger` with `{ "id": "<that id>" }` as the same super admin. Before this change: `404 { "error": "Schedule not found" }`, independent of `QUEUE_STRATEGY`.

## Compatibility

No contract surface touched — no type, signature, import path, event id, widget spot, route path, DB column, DI key, ACL feature, notification id, CLI command, or generated file changes. The only observable delta is the corrected status for requests that were already broken.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5381"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789725369&installation_model_id=432243&pr_number=5381&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5381&signature=9aea01f8922cda78890448866c8200307b63056c691d30825594f255e6830b64"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/scheduler/src/modules/scheduler/__integration__/TC-SCHED-008.spec.ts
+++ b/packages/scheduler/src/modules/scheduler/__integration__/TC-SCHED-008.spec.ts
@@ -57,7 +57,8 @@ test.describe('TC-SCHED-008: system-scoped schedule access on trigger and execut
         expect(superBody?.ok).toBe(true)
         expect(typeof superBody?.jobId === 'string' && superBody.jobId.length > 0).toBe(true)
       } else {
-        // The strategy check runs after the scope check, so this is the 404 used to mask.
+        // The strategy check runs after the scope check, so this 400 is the response the
+        // 404 used to mask.
         expect(superResponse.status()).toBe(400)
         expect(superBody?.error ?? '').toMatch(/async/i)
       }
--- a/packages/scheduler/src/modules/scheduler/__integration__/TC-SCHED-008.spec.ts
+++ b/packages/scheduler/src/modules/scheduler/__integration__/TC-SCHED-008.spec.ts
@@ -1,27 +1,34 @@
 import { expect, test } from '@playwright/test'
 import { apiRequest, getAuthToken } from '@open-mercato/core/helpers/integration/api'
 import { readJsonSafe } from '@open-mercato/core/helpers/integration/generalFixtures'
-import { SCHEDULER_TRIGGER_PATH, createScheduleJob, deleteScheduleJob, uniqueScheduleName } from './helpers/scheduler'
+import {
+  SCHEDULER_JOBS_PATH,
+  SCHEDULER_TRIGGER_PATH,
+  createScheduleJob,
+  deleteScheduleJob,
+  uniqueScheduleName,
+} from './helpers/scheduler'
 
 const isAsyncQueueStrategy = (process.env.QUEUE_STRATEGY || 'local') === 'async'
 
 type TriggerBody = { ok?: boolean; jobId?: string; error?: string; message?: string }
 
 /**
- * TC-SCHED-008: POST /api/scheduler/trigger resolves the scope of a system-scoped
- * schedule on the loaded row, not in the lookup.
+ * TC-SCHED-008: POST /api/scheduler/trigger and GET /api/scheduler/jobs/[id]/executions
+ * resolve the scope of a system-scoped schedule on the loaded row, not in the lookup.
  *
- * A system-scoped schedule has `tenantId === null` and `organizationId === null`. When the
- * route folded the caller's tenant/org into the `where` clause, no such row could ever match
- * and every caller — super admins included — got 404, which also masked the route's own
+ * A system-scoped schedule has `tenantId === null` and `organizationId === null`. When both
+ * routes folded the caller's tenant/org into the `where` clause, no such row could ever match
+ * and every caller — super admins included — got 404, which also masked each route's own
  * super-admin gate. A super admin must now reach the schedule, and a non-super-admin must be
  * told 403 rather than 404.
  *
- * The super-admin assertion holds under either queue strategy: the `QUEUE_STRATEGY=async`
- * check sits after the lookup, so `local` yields the pre-existing 400 — never 404.
+ * The super-admin assertions hold under either queue strategy: each route's
+ * `QUEUE_STRATEGY=async` check sits after the lookup, so `local` yields the pre-existing
+ * 400 — never 404.
  */
-test.describe('TC-SCHED-008: manual trigger of a system-scoped schedule', () => {
-  test('super admin reaches the schedule; a non-super-admin is forbidden', async ({ request }) => {
+test.describe('TC-SCHED-008: system-scoped schedule access on trigger and executions', () => {
+  test('super admin reaches the schedule on both routes; a non-super-admin is forbidden', async ({ request }) => {
     const superToken = await getAuthToken(request, 'superadmin')
     const adminToken = await getAuthToken(request, 'admin')
     let scheduleId: string | null = null
@@ -63,6 +70,21 @@ test.describe('TC-SCHED-008: manual trigger of a system-scoped schedule', () => 
       expect(
         adminResponse.status(),
         'a non-super-admin must be forbidden from triggering a system-scoped schedule',
+      ).toBe(403)
+
+      // The execution-history route carried the same lookup and needs the same guarantees.
+      const executionsPath = `${SCHEDULER_JOBS_PATH}/${scheduleId}/executions`
+      const superExecutions = await apiRequest(request, 'GET', executionsPath, { token: superToken })
+      expect(
+        superExecutions.status(),
+        'a super admin must not be told a system-scoped schedule has no execution history route',
+      ).not.toBe(404)
+      expect(superExecutions.status()).toBe(isAsyncQueueStrategy ? 200 : 400)
+
+      const adminExecutions = await apiRequest(request, 'GET', executionsPath, { token: adminToken })
+      expect(
+        adminExecutions.status(),
+        'a non-super-admin must be forbidden from reading a system-scoped execution history',
       ).toBe(403)
     } finally {
       await deleteScheduleJob(request, superToken, scheduleId)
--- a/packages/scheduler/src/modules/scheduler/__integration__/TC-SCHED-008.spec.ts
+++ b/packages/scheduler/src/modules/scheduler/__integration__/TC-SCHED-008.spec.ts
@@ -1,0 +1,71 @@
+import { expect, test } from '@playwright/test'
+import { apiRequest, getAuthToken } from '@open-mercato/core/helpers/integration/api'
+import { readJsonSafe } from '@open-mercato/core/helpers/integration/generalFixtures'
+import { SCHEDULER_TRIGGER_PATH, createScheduleJob, deleteScheduleJob, uniqueScheduleName } from './helpers/scheduler'
+
+const isAsyncQueueStrategy = (process.env.QUEUE_STRATEGY || 'local') === 'async'
+
+type TriggerBody = { ok?: boolean; jobId?: string; error?: string; message?: string }
+
+/**
+ * TC-SCHED-008: POST /api/scheduler/trigger resolves the scope of a system-scoped
+ * schedule on the loaded row, not in the lookup.
+ *
+ * A system-scoped schedule has `tenantId === null` and `organizationId === null`. When the
+ * route folded the caller's tenant/org into the `where` clause, no such row could ever match
+ * and every caller — super admins included — got 404, which also masked the route's own
+ * super-admin gate. A super admin must now reach the schedule, and a non-super-admin must be
+ * told 403 rather than 404.
+ *
+ * The super-admin assertion holds under either queue strategy: the `QUEUE_STRATEGY=async`
+ * check sits after the lookup, so `local` yields the pre-existing 400 — never 404.
+ */
+test.describe('TC-SCHED-008: manual trigger of a system-scoped schedule', () => {
+  test('super admin reaches the schedule; a non-super-admin is forbidden', async ({ request }) => {
+    const superToken = await getAuthToken(request, 'superadmin')
+    const adminToken = await getAuthToken(request, 'admin')
+    let scheduleId: string | null = null
+
+    try {
+      // Creating a system-scoped schedule is itself super-admin only.
+      scheduleId = await createScheduleJob(request, superToken, {
+        name: uniqueScheduleName('System-Trigger-Test'),
+        scopeType: 'system',
+        scheduleType: 'interval',
+        scheduleValue: '1h',
+      })
+
+      const superResponse = await apiRequest(request, 'POST', SCHEDULER_TRIGGER_PATH, {
+        token: superToken,
+        data: { id: scheduleId },
+      })
+      const superBody = await readJsonSafe<TriggerBody>(superResponse)
+      expect(
+        superResponse.status(),
+        'a super admin must not be told a system-scoped schedule does not exist',
+      ).not.toBe(404)
+
+      if (isAsyncQueueStrategy) {
+        expect(superResponse.status()).toBe(200)
+        expect(superBody?.ok).toBe(true)
+        expect(typeof superBody?.jobId === 'string' && superBody.jobId.length > 0).toBe(true)
+      } else {
+        // The strategy check runs after the scope check, so this is the 404 used to mask.
+        expect(superResponse.status()).toBe(400)
+        expect(superBody?.error ?? '').toMatch(/async/i)
+      }
+
+      // The same schedule is visible enough to be refused, not hidden, from a tenant admin.
+      const adminResponse = await apiRequest(request, 'POST', SCHEDULER_TRIGGER_PATH, {
+        token: adminToken,
+        data: { id: scheduleId },
+      })
+      expect(
+        adminResponse.status(),
+        'a non-super-admin must be forbidden from triggering a system-scoped schedule',
+      ).toBe(403)
+    } finally {
+      await deleteScheduleJob(request, superToken, scheduleId)
+    }
+  })
+})
--- a/packages/scheduler/src/modules/scheduler/api/jobs/[id]/executions/route.ts
+++ b/packages/scheduler/src/modules/scheduler/api/jobs/[id]/executions/route.ts
@@ -5,6 +5,7 @@ import { createRequestContainer } from '@open-mercato/shared/lib/di/container'
 import type { EntityManager } from '@mikro-orm/core'
 import type { OpenApiRouteDoc } from '@open-mercato/shared/lib/openapi'
 import { ScheduledJob } from '../../../../data/entities.js'
+import { resolveScheduleAccess } from '../../../../lib/scheduleAccess.js'
 import { getRedisUrlOrThrow, parseRedisUrl } from '@open-mercato/shared/lib/redis/connection'
 import { resolveTranslations } from '@open-mercato/shared/lib/i18n/server'
 import { createLogger } from '@open-mercato/shared/lib/logger'
@@ -40,30 +41,16 @@ export async function GET(
   const scheduleId = params.id
 
   try {
-    // Verify schedule exists and user has access (with tenant/org scope filter)
-    const findFilter: Record<string, unknown> = {
-      id: scheduleId,
-      deletedAt: null,
-    }
+    // Load by id alone, then apply isolation to the loaded row. Folding the actor's
+    // tenant/org into the lookup would make every system-scoped schedule unmatchable.
+    const schedule = await em.findOne(ScheduledJob, { id: scheduleId, deletedAt: null })
 
-    // Apply tenant isolation: scope the query to the user's tenant/org
-    if (auth.tenantId) {
-      findFilter.tenantId = auth.tenantId
-    }
-    if (auth.orgId) {
-      findFilter.organizationId = auth.orgId
-    }
+    const access = schedule ? resolveScheduleAccess(schedule, auth) : 'not_found'
 
-    const schedule = await em.findOne(ScheduledJob, findFilter)
-
-    if (!schedule) {
+    if (!schedule || access === 'not_found') {
       return NextResponse.json({ error: translate('scheduler.error.not_found', 'Schedule not found') }, { status: 404 })
     }
-
-    // System-scoped schedules (no tenantId/orgId) require super-admin. Use the
-    // immutable `isSuperAdmin` flag — never compare mutable/spoofable role names.
-    const isSuperAdmin = auth.isSuperAdmin === true
-    if (!schedule.tenantId && !schedule.organizationId && !isSuperAdmin) {
+    if (access === 'forbidden') {
       return NextResponse.json({ error: translate('scheduler.error.access_denied', 'Access denied') }, { status: 403 })
     }
 
--- a/packages/scheduler/src/modules/scheduler/api/jobs/[id]/executions/route.ts
+++ b/packages/scheduler/src/modules/scheduler/api/jobs/[id]/executions/route.ts
@@ -45,9 +45,12 @@ export async function GET(
     // tenant/org into the lookup would make every system-scoped schedule unmatchable.
     const schedule = await em.findOne(ScheduledJob, { id: scheduleId, deletedAt: null })
 
-    const access = schedule ? resolveScheduleAccess(schedule, auth) : 'not_found'
+    if (!schedule) {
+      return NextResponse.json({ error: translate('scheduler.error.not_found', 'Schedule not found') }, { status: 404 })
+    }
 
-    if (!schedule || access === 'not_found') {
+    const access = resolveScheduleAccess(schedule, auth)
+    if (access === 'not_found') {
       return NextResponse.json({ error: translate('scheduler.error.not_found', 'Schedule not found') }, { status: 404 })
     }
     if (access === 'forbidden') {
--- a/packages/scheduler/src/modules/scheduler/api/trigger/route.ts
+++ b/packages/scheduler/src/modules/scheduler/api/trigger/route.ts
@@ -7,6 +7,7 @@ import type { OpenApiRouteDoc } from '@open-mercato/shared/lib/openapi'
 import { createQueue } from '@open-mercato/queue'
 import { getRedisUrlOrThrow } from '@open-mercato/shared/lib/redis/connection'
 import { ScheduledJob } from '../../data/entities.js'
+import { resolveScheduleAccess } from '../../lib/scheduleAccess.js'
 import { scheduleTriggerSchema } from '../../data/validators.js'
 import type { ExecuteSchedulePayload } from '../../workers/execute-schedule.worker.js'
 import { resolveTranslations } from '@open-mercato/shared/lib/i18n/server'
@@ -40,30 +41,16 @@ export async function POST(req: NextRequest) {
     const body = await req.json()
     const input = scheduleTriggerSchema.parse(body)
 
-    // Find the schedule with tenant/org scope filter
-    const findFilter: Record<string, unknown> = {
-      id: input.id,
-      deletedAt: null,
-    }
+    // Load by id alone, then apply isolation to the loaded row. Folding the actor's
+    // tenant/org into the lookup would make every system-scoped schedule unmatchable.
+    const schedule = await em.findOne(ScheduledJob, { id: input.id, deletedAt: null })
 
-    // Apply tenant isolation: scope the query to the user's tenant/org
-    if (auth.tenantId) {
-      findFilter.tenantId = auth.tenantId
-    }
-    if (auth.orgId) {
-      findFilter.organizationId = auth.orgId
-    }
+    const access = schedule ? resolveScheduleAccess(schedule, auth) : 'not_found'
 
-    const schedule = await em.findOne(ScheduledJob, findFilter)
-
-    if (!schedule) {
+    if (!schedule || access === 'not_found') {
       return NextResponse.json({ error: translate('scheduler.error.not_found', 'Schedule not found') }, { status: 404 })
     }
-
-    // System-scoped schedules (no tenantId/orgId) require super-admin. Use the
-    // immutable `isSuperAdmin` flag — never compare mutable/spoofable role names.
-    const isSuperAdmin = auth.isSuperAdmin === true
-    if (!schedule.tenantId && !schedule.organizationId && !isSuperAdmin) {
+    if (access === 'forbidden') {
       return NextResponse.json({ error: translate('scheduler.error.access_denied', 'Access denied') }, { status: 403 })
     }
 
--- a/packages/scheduler/src/modules/scheduler/api/trigger/route.ts
+++ b/packages/scheduler/src/modules/scheduler/api/trigger/route.ts
@@ -45,9 +45,12 @@ export async function POST(req: NextRequest) {
     // tenant/org into the lookup would make every system-scoped schedule unmatchable.
     const schedule = await em.findOne(ScheduledJob, { id: input.id, deletedAt: null })
 
-    const access = schedule ? resolveScheduleAccess(schedule, auth) : 'not_found'
+    if (!schedule) {
+      return NextResponse.json({ error: translate('scheduler.error.not_found', 'Schedule not found') }, { status: 404 })
+    }
 
-    if (!schedule || access === 'not_found') {
+    const access = resolveScheduleAccess(schedule, auth)
+    if (access === 'not_found') {
       return NextResponse.json({ error: translate('scheduler.error.not_found', 'Schedule not found') }, { status: 404 })
     }
     if (access === 'forbidden') {
--- a/packages/scheduler/src/modules/scheduler/lib/__tests__/scheduleAccess.test.ts
+++ b/packages/scheduler/src/modules/scheduler/lib/__tests__/scheduleAccess.test.ts
@@ -70,6 +70,17 @@ describe('resolveScheduleAccess — system-scope classification matches commands
     ).toBe('forbidden')
   })
 
+  // 403 here would tell an actor in t1 that the id exists in t2. Only a genuinely
+  // tenant-less row is allowed to answer `forbidden`.
+  it('does not disclose a tenant-bound "system" row across a tenant boundary', () => {
+    expect(
+      resolveScheduleAccess(
+        { scopeType: 'system', tenantId: 't2', organizationId: null },
+        { tenantId: 't1', orgId: 'o1', isSuperAdmin: false },
+      ),
+    ).toBe('not_found')
+  })
+
   it('still classifies by nullability when scopeType is absent', () => {
     expect(
       resolveScheduleAccess({ tenantId: null, organizationId: null }, { tenantId: 't1', isSuperAdmin: false }),
@@ -93,6 +104,17 @@ describe('resolveScheduleAccess — tenant isolation', () => {
   it('fails closed for a non-super-admin whose tenant could not be resolved', () => {
     expect(
       resolveScheduleAccess(orgSchedule, { tenantId: null, orgId: null, isSuperAdmin: false }),
+    ).toBe('not_found')
+  })
+
+  // The exact shape the replaced lookup leaked: with no tenant clause applied, the filter
+  // `{ id, deletedAt: null, organizationId: 'o1' }` matched org-bound rows in ANY tenant.
+  it('closes the cross-tenant leak for an actor with an org but no resolvable tenant', () => {
+    expect(
+      resolveScheduleAccess(
+        { scopeType: 'organization', tenantId: 't2', organizationId: 'o1' },
+        { tenantId: null, orgId: 'o1', isSuperAdmin: false },
+      ),
     ).toBe('not_found')
   })
 
--- a/packages/scheduler/src/modules/scheduler/lib/__tests__/scheduleAccess.test.ts
+++ b/packages/scheduler/src/modules/scheduler/lib/__tests__/scheduleAccess.test.ts
@@ -11,9 +11,9 @@
 import { describe, it, expect } from '@jest/globals'
 import { resolveScheduleAccess } from '../scheduleAccess'
 
-const systemSchedule = { tenantId: null, organizationId: null }
-const tenantSchedule = { tenantId: 't1', organizationId: null }
-const orgSchedule = { tenantId: 't1', organizationId: 'o1' }
+const systemSchedule = { scopeType: 'system', tenantId: null, organizationId: null }
+const tenantSchedule = { scopeType: 'tenant', tenantId: 't1', organizationId: null }
+const orgSchedule = { scopeType: 'organization', tenantId: 't1', organizationId: 'o1' }
 
 describe('resolveScheduleAccess — system-scoped schedules', () => {
   it('allows a super admin whose session carries a tenant and an organization', () => {
@@ -45,6 +45,38 @@ describe('resolveScheduleAccess — system-scoped schedules', () => {
   })
 })
 
+/**
+ * `ensureCanManageSystemScopedJob` in commands/jobs.ts treats a row as system-scoped when
+ * `scopeType === 'system' || tenantId == null`. Update, delete, trigger and executions must
+ * not disagree about what a system-scoped row is, so the same test is applied here — including
+ * for the inconsistent rows the create validator's scope refinement should never produce.
+ */
+describe('resolveScheduleAccess — system-scope classification matches commands/jobs.ts', () => {
+  it('treats a null tenant as system-scoped even when an organization is set', () => {
+    expect(
+      resolveScheduleAccess(
+        { scopeType: 'organization', tenantId: null, organizationId: 'o1' },
+        { tenantId: 't1', orgId: 'o1', isSuperAdmin: false },
+      ),
+    ).toBe('forbidden')
+  })
+
+  it('treats scopeType "system" as system-scoped even when a tenant is set', () => {
+    expect(
+      resolveScheduleAccess(
+        { scopeType: 'system', tenantId: 't1', organizationId: null },
+        { tenantId: 't1', orgId: 'o1', isSuperAdmin: false },
+      ),
+    ).toBe('forbidden')
+  })
+
+  it('still classifies by nullability when scopeType is absent', () => {
+    expect(
+      resolveScheduleAccess({ tenantId: null, organizationId: null }, { tenantId: 't1', isSuperAdmin: false }),
+    ).toBe('forbidden')
+  })
+})
+
 describe('resolveScheduleAccess — tenant isolation', () => {
   it('reports another tenant\'s schedule as not_found, never forbidden', () => {
     expect(
@@ -64,10 +96,12 @@ describe('resolveScheduleAccess — tenant isolation', () => {
     ).toBe('not_found')
   })
 
-  it('keeps a tenant-less super admin reaching a tenant-bound schedule', () => {
+  // buildSchedulerJobsFilters returns an unmatchable filter when the tenant is falsy, for
+  // super admins too, so such an actor cannot see this row in the list either.
+  it('fails closed for a super admin whose tenant could not be resolved', () => {
     expect(
       resolveScheduleAccess(orgSchedule, { tenantId: null, orgId: null, isSuperAdmin: true }),
-    ).toBe('allowed')
+    ).toBe('not_found')
   })
 })
 
@@ -84,6 +118,10 @@ describe('resolveScheduleAccess — organization isolation', () => {
     ).toBe('allowed')
   })
 
+  // Inherited from the lookup this helper replaces, which omitted the organization clause
+  // entirely when the actor carried no orgId. The list endpoint hides the row in this case
+  // (its organization branch needs at least one id), so the two paths disagree. Pinned here
+  // to document the divergence, not to endorse it — narrowing it is a separate change.
   it('allows an actor with no selected organization to reach an org-bound schedule in its tenant', () => {
     expect(
       resolveScheduleAccess(orgSchedule, { tenantId: 't1', orgId: null, isSuperAdmin: false }),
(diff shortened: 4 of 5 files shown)


Frshy:
## Metadata intent — needs a maintainer

The contributing account has **read-only** permission on this repository, so it cannot apply labels, assign, or request a reviewer. Everything below is intent, validated against `gh label list` on this repo so the names are real. Posting it in one comment rather than attempting the writes.

**Labels**

| Label | Rationale |
|---|---|
| `bug` | A documented capability — manually triggering a system-scoped schedule, and reading its execution history — is unusable for every actor, super admins included. |
| `review` | Ready for review; not in any other pipeline state. |
| `priority-medium` | The feature is fully broken, but the scheduled cron still fires these jobs, so there is a workaround and no data loss. |
| `risk-medium` | Touches authorization on two API routes. Blast radius is bounded to the scheduler module and the decision is one pure function with direct unit coverage — but it is authorization code. |
| `skip-qa` | See the QA note below. `needs-qa` is a reasonable maintainer override. |

**QA routing.** Reading `AGENTS.md` → PR Workflow, this appears to meet the automated-verification exemption: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, database structure unchanged, API surface unchanged (route paths, request schema, and the declared OpenAPI 400/401/403/404 responses all as before), no `BACKWARD_COMPATIBILITY.md` contract broken, and automated tests for the changed behaviour ship in this PR.

That said — it is authorization code, and the exemption is a maintainer's call, not mine. If you would rather route it through manual QA, `needs-qa` instead of `skip-qa` is entirely reasonable. Please don't apply both.

**Reviewer:** by role — whoever owns the `packages/scheduler` module or platform authorization. Deliberately not naming a handle.

**Assignee:** the contributing account, `Frshy`, if you assign external contributors.

No claim label (`in-progress`) is requested: this account could not release one later.

**Not ticked in the checklist, deliberately:**

- **The CLA box.** Not something an agent ticks on a human's behalf — @Frshy, please confirm `docs/cla.md` yourself.
- The three label checkboxes, for the permission reason above.
- The spec checkbox — `.ai/specs/AGENTS.md` directs "Skip specs for small bug fixes"; reasoning lives in code comments and the PR body instead.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-19T11:20:40Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(scheduler): manual trigger and execution history 404 on every system-scoped schedule

## 🎯 Summary

This PR fixes a real and well-diagnosed bug. `POST /api/scheduler/trigger` and `GET /api/scheduler/jobs/[id]/executions` folded the caller's `tenantId`/`organizationId` into the `em.findOne` `where` clause, so a system-scoped row (`tenant_id IS NULL`, `organization_id IS NULL`) could never match for any session carrying a tenant — every caller, super admins included, got a 404, and each route's own super-admin gate below the lookup was unreachable dead code. The fix moves isolation off the `where` and onto the loaded row through a new pure helper, `resolveScheduleAccess(schedule, actor)`, which is the shape `updateScheduleCommand`/`deleteScheduleCommand` already use.

The engineering quality here is high and I want to say so before the findings. The helper is pure, DI-free and directly unit-tested; system-scope classification is deliberately made byte-identical to `ensureCanManageSystemScopedJob` (`scopeType === 'system' || tenantId == null`) so update, delete, trigger and executions can no longer disagree about what a system-scoped row is; super-admin status reads the immutable `auth.isSuperAdmin` flag rather than a role name, and there is an explicit anti-spoofing unit case for that; the 404-vs-403 disclosure decision is reasoned rather than accidental; and the one behavioural divergence from `buildSchedulerJobsFilters` that the author chose not to fix is documented *and* pinned by a test rather than quietly inherited.

It is also worth recording that this change **closes** a cross-tenant hole rather than only fixing a 404. Under the old lookup, an actor whose session carried an `orgId` but no resolvable `tenantId` had no tenant clause applied at all, so `{ id, deletedAt: null, organizationId: <their org> }` matched org-bound rows in *any* tenant. The new helper returns `not_found` whenever the actor's tenant is unresolved, for super admins too, matching `buildSchedulerJobsFilters`, which returns an unmatchable filter in that case. That is a genuine improvement the PR description undersells.

Reviewed scope: the five changed files, plus the surrounding contracts they claim to align with — `api/jobs/buildFilters.ts`, `commands/jobs.ts` (`ensureCanManageSystemScopedJob`, update/delete), `data/validators.ts` (the scope refinement), `services/schedulerService.ts` (`register`), `acl.ts`/`setup.ts` (feature grants), the two routes' `openApi` response declarations, and the module's four locale files.

## Verdict

❌ **request changes** — one Minor finding drives this. The new helper evaluates the system-scope branch before any tenant comparison, so a row carrying `scopeType: 'system'` *together with* a non-null `tenantId` now answers **403** to a non-super-admin from a **different** tenant, where the code being replaced answered 404. That contradicts the helper's own documented invariant ("Another tenant's or another organization's schedule answers `not_found`, because a 403 would confirm that the id exists"), and it is the one place where this PR is less isolating than what it replaces. Under this repository's verdict rule a Minor finding requests changes; the fix is one conditional plus one unit case, and everything else here is nits.

## 🧪 Validation Gate

CI has **not run on this PR head**. The `CI for Develop&Main` workflow run for `673ad7d` is in `action_required` — the standard GitHub gate on a fork PR from a first-time contributor — so no build, typecheck, test or lint signal exists on GitHub for this SHA. Only `license/cla` (pass) and `[code]smith` (skipped) reported. Per this repository's GitHub-checks-first review policy I therefore fell back to local validation, kept as narrow as the missing signal allows: the changed package only, since the diff touches nothing outside `packages/scheduler` and no shared contract.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ⏭️ not run | Not reproduced locally; no cross-package surface changed. Belongs to the maintainer-approved CI run. |
| `yarn generate` | ⏭️ not run | No auto-discovery file added or modified — the new file is a plain `lib/` helper, not a `route.ts`/`events.ts`/`ce.ts`. |
| `yarn i18n:check-sync` | ✅ PASS (verified directly) | Both keys the routes use, `scheduler.error.not_found` and `scheduler.error.access_denied`, are present in all four module locales (`de`, `en`, `es`, `pl`). No key added or removed. |
| `yarn i18n:check-usage` | ⏭️ not run | Advisory gate; no new user-facing string introduced. |
| `yarn typecheck` | ✅ PASS (scheduler package) | `tsc --noEmit` in `packages/scheduler` reports **zero** errors in any of the five changed files. The 14 errors it does report are all pre-existing and in untouched backend `.tsx` pages (`@tanstack/react-table/legacy` unresolvable in my worktree, cascading to `TS7031`), an artifact of my dependency linking rather than of this PR. |
| `yarn test` | ✅ PASS (scheduler package) | `jest --config jest.config.cjs` in `packages/scheduler`: **382 tests / 20 suites passed**, exactly matching the author's reported figures, including all 17 new `lib/__tests__/scheduleAccess.test.ts` cases. |
| `yarn lint` | ⚠️ not run | ESLint could not load `eslint.config.mjs` in my review worktree (`Cannot find module 'eslint-config-next/core-web-vitals'`) — an environment limitation, not a finding about the diff. The changed files are plain TypeScript with no pattern this repo's lint rules target. |
| `yarn build:app` | ⏭️ not run | No app-level surface touched. |
| `yarn test:integration` (`TC-SCHED-008`) | ⏭️ not run | Requires a running stack. The author reports 16/16 scheduler specs passing plus a documented regression proof (the two routes checked out from `develop` make TC-SCHED-008 fail on exactly the intended assertion). I verified statically that every helper the spec imports from `./helpers/scheduler` exists and is used correctly, and that `deleteScheduleJob` tolerates the `null` id its `finally` may pass. |

**A maintainer must approve the workflow run before this can merge** — the local evidence above is this review's own, never a substitute for branch protection.

## Findings

### 🔹 Minor

**`packages/scheduler/src/modules/scheduler/lib/scheduleAccess.ts:70-72` — the system-scope branch discloses cross-tenant existence for a scope-inconsistent row.**

The branch is evaluated before any tenant comparison:

```ts
if (schedule.scopeType === 'system' || scheduleTenantId === null) {
  return isSuperAdmin ? 'allowed' : 'forbidden'
}
```

For a row that carries `scopeType: 'system'` *and* a non-null `tenantId`, this returns `forbidden` regardless of which tenant the actor belongs to. Concretely: actor `{ tenantId: 't1', orgId: 'o1', isSuperAdmin: false }`, row `{ id: X, scopeType: 'system', tenantId: 't2' }` → the old lookup applied `tenantId: 't1'`, missed, and answered **404**; the new helper answers **403 Access denied**, telling an actor in tenant `t1` that id `X` exists in tenant `t2`. That is exactly the disclosure this PR's description sets out to avoid, and it is the single respect in which the change is weaker than the code it replaces.

Reachability, stated fairly: `scheduleCreateSchema`'s scope refinement (`data/validators.ts:50-67`) forbids `system` with a `tenantId`, `scheduleUpdateSchema` cannot change `scopeType` or `tenantId` at all, and I checked every in-repo `schedulerService.register` caller — the two `scopeType: 'system'` registrations in `packages/ai-assistant/src/modules/ai_assistant/setup.ts` both omit `tenantId`, and the `checkout` / `communication_channels` ones are `organization`-scoped with matching ids. So **no current caller produces such a row.** But `schedulerService.register` (`services/schedulerService.ts:70-72, 88-91`) writes `scopeType` and `tenantId` straight through with no cross-field validation, so a third-party or future module can produce one — and this PR's own tests already reason about that class of row ("treats scopeType `system` as system-scoped even when a tenant is set"), just never with a cross-tenant actor. The impact is existence-only, on unguessable UUIDs, behind `scheduler.jobs.view`/`scheduler.jobs.trigger`; that is why this is Minor and not Major.

Suggested fix — this preserves all 17 existing unit cases, including the tenant-less super admin reaching system schedules:

```ts
if (schedule.scopeType === 'system' || scheduleTenantId === null) {
  if (isSuperAdmin) return 'allowed'
  // A row mislabelled 'system' while still bound to a tenant must not confirm its
  // existence across a tenant boundary; only a genuinely tenant-less row answers 403.
  if (scheduleTenantId !== null && scheduleTenantId !== (actor?.tenantId ?? null)) return 'not_found'
  return 'forbidden'
}
```

Please add the matching case to the "system-scope classification" describe block: a `{ scopeType: 'system', tenantId: 't2' }` row against a `{ tenantId: 't1', isSuperAdmin: false }` actor must be `not_found`.

### 💅 Nit

**`packages/scheduler/src/modules/scheduler/lib/scheduleAccess.ts:56-61` — the organization-divergence JSDoc describes only the narrower half.**

The comment says the helper's single-`orgId` comparison "is narrower than the list endpoint's resolved organization scope (`filterIds`…)". That is true for a multi-organization actor, but the divergence is two-directional, and the other direction is the permissive one: when the actor carries **no** `orgId`, the helper allows any org-bound schedule inside the tenant (`scheduleAccess.ts:80` short-circuits on `actorOrganizationId !== null`), whereas `buildSchedulerJobsFilters` hides it entirely — its organization branch is only emitted when `orgIdsForVisibility.length > 0`. Your own test at `lib/__tests__/scheduleAccess.test.ts:118-124` pins this case accurately and honestly; the JSDoc is what a reader of the helper alone would rely on, and today it would leave them believing the helper can only ever be more restrictive than the list. One sentence naming the permissive direction would close the gap. (The behaviour itself is inherited from the lookup being replaced, so leaving it for a follow-up is the right call — please do file that follow-up, since a helper named `resolveScheduleAccess` will attract further call sites.)

**`packages/scheduler/src/modules/scheduler/lib/scheduleAccess.ts:78-82` — the tenant-wide branch does not consult `scopeType`, unlike the list endpoint's equivalent branch.**

`buildSchedulerJobsFilters` requires `scope_type = 'tenant'` on its `organization_id IS NULL` branch (`api/jobs/buildFilters.ts:51-55`); the helper allows any org-null row whose tenant matches, whatever its `scopeType`. Same scope-inconsistent-row class as the Minor above (a `scopeType: 'organization'` row with a null `organizationId` is visible here and hidden in the list), and equally not producible by the current create path. Worth a line of comment at least, so the intentional looseness is on the record next to the other two documented divergences.

**PR description — locale count.** The description says "all five locales stay in sync"; `packages/scheduler/src/modules/scheduler/i18n/` carries four (`de`, `en`, `es`, `pl`). The substance is right — both reused keys are present in all four and no key changed — only the number is off.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff is purely additive at the module boundary (`lib/scheduleAccess.ts` is new; nothing was removed or re-exported differently).
- [x] No required type field removed or narrowed — `ScheduleScopeSubject` deliberately makes every field optional so it accepts the `ScheduledJob` entity as declared.
- [x] No HTTP route URL removed or renamed; no method changed — `POST /api/scheduler/trigger` and `GET /api/scheduler/jobs/[id]/executions` are untouched in path, method, guards and request schema.
- [x] No field removed or retyped in an existing response shape — both success and error bodies are unchanged; only which status a given request receives changes, and every status involved (400/401/403/404) was already declared in each route's `openApi.responses`, so the generated documentation needs no update.
- [x] No event or message name renamed or removed; no payload field removed — `ExecuteSchedulePayload` is built exactly as before.
- [x] No database table or column renamed or removed; no column type narrowed — no entity change, so correctly no migration and no snapshot delta.
- [x] Where a contract had to change: not applicable — nothing in `BACKWARD_COMPATIBILITY.md`'s thirteen protected surfaces is touched. The only observable delta is the corrected status for requests that were already failing.

## 🧪 Test Coverage

**What is covered, and how.** `lib/__tests__/scheduleAccess.test.ts` exercises the real `resolveScheduleAccess` with no mocks — the same style as the existing `api/jobs/__tests__/scope-filter.test.ts` — across four groups: system-scope access (super admin with and without a tenant, non-super-admin, an actor whose role is literally named `superadmin` but lacks the `isSuperAdmin` flag, and a null actor), system-scope *classification* parity with `ensureCanManageSystemScopedJob` (null tenant with an org set, `scopeType: 'system'` with a tenant set, absent `scopeType`), tenant isolation (cross-tenant as `not_found` rather than `forbidden`, including for a super admin scoped elsewhere, and fail-closed on an unresolved tenant for both roles), and organization isolation (wrong org, owning org, no selected org, tenant-scoped row seen by an org-bound actor, `undefined` vs `null` `organizationId`). The anti-spoofing case and the fail-closed cases are the ones I would have asked for had they been missing, and the divergence test at line 118 documents a known inconsistency instead of asserting it away. I ran the suite: 382 tests across 20 suites pass.

`__integration__/TC-SCHED-008.spec.ts` covers the end-to-end path a unit test cannot: it creates a real system-scoped schedule as `superadmin`, asserts the super admin is `not.toBe(404)` on both routes, asserts a tenant `admin` gets exactly 403 on both, and cleans up in `finally`. I confirmed the 403 assertions are meaningful rather than incidental — `setup.ts` grants `admin: ['scheduler.*']`, so the `admin` fixture holds both `scheduler.jobs.trigger` and `scheduler.jobs.view` and reaches the scope check rather than being stopped by the `requireFeatures` guard. The strategy-conditional assertions (200 under `async`, 400 under `local`) are honest about what the local run proves, and the author's regression proof against `develop` is the right way to demonstrate the test actually pins the bug.

**Gaps, with the cases to add.**

1. The Minor finding's case is untested and would be silently reintroduced by any future refactor: add to the "system-scope classification matches commands/jobs.ts" block an assertion that `resolveScheduleAccess({ scopeType: 'system', tenantId: 't2', organizationId: null }, { tenantId: 't1', orgId: 'o1', isSuperAdmin: false })` is `'not_found'`, alongside the existing same-tenant case that expects `'forbidden'`.
2. The regression that this PR *fixes* on the cross-tenant side has no test of its own. The old lookup let an actor with an `orgId` but no resolvable `tenantId` match org-bound rows in any tenant; the helper now returns `not_found`. There is a fail-closed case for `{ tenantId: null, orgId: null }`, but none for `{ tenantId: null, orgId: 'o1' }` against a row in another tenant — which is the shape the old code actually leaked. One line, and it pins a genuine security improvement.
3. Cross-tenant coverage remains unit-level only, which the author states rather than papers over: `packages/core/src/helpers/integration/` has no second-tenant fixture, so a cross-tenant integration assertion is not expressible self-containedly today. That is a fair limitation to accept for this PR, not a reason to hold it.

pkarw:
## 🏷️ label rationale

Applied by `om-auto-review-pr`. The author flagged in [their metadata-intent comment](https://github.com/open-mercato/open-mercato/pull/5381#issuecomment-5341343757) that the contributing account has read-only permission and could not set these; I have applied them, with one deliberate difference from the intent (pipeline state, because the review requests changes).

| Label | Rationale |
|---|---|
| `changes-requested` | Pipeline state, set by this review. One Minor finding (a cross-tenant existence disclosure introduced in the new `resolveScheduleAccess` system-scope branch) drives it — this repository requests changes on Minor. Replaces the `review` state the author intended; it will move back once the finding is addressed. |
| `bug` | Category. A documented capability — manually triggering a system-scoped schedule and reading its execution history — was unusable for every actor, super admins included. Matches the author's intent. |
| `priority-medium` | Urgency. The feature is fully broken, but the unattended cron still fires these jobs, so there is a workaround and no data loss. Matches the author's intent. |
| `risk-medium` | Blast radius. Two API routes' authorization logic changes, which is inherently risk-bearing, but the surface is bounded to `packages/scheduler`, the decision is one pure function with 17 direct unit cases, and no contract surface, entity or migration is touched. Matches the author's intent. |
| `skip-qa` | QA routing. The automated-verification exemption in `AGENTS.md` → PR Workflow is met on every clause: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, database structure unchanged, API surface unchanged (paths, methods, guards, request schemas and the declared OpenAPI 400/401/403/404 responses all as before), no `BACKWARD_COMPATIBILITY.md` contract broken, and automated tests for the changed behaviour ship in the same PR (unit + `TC-SCHED-008`). The author raised, fairly, that this is authorization code and a `needs-qa` override would be reasonable — that remains a maintainer's call; if you flip it, please remove `skip-qa` rather than carrying both. |
| `partner-request` | Pre-existing, preserved. |

`in-progress` is this review's claim lock and is released in the completion comment below. No `qa` label is set — that one means manual QA is in progress and belongs to a QA reviewer.

**Not a label, but merge-blocking:** the `CI for Develop&Main` run for `673ad7d` is stuck in `action_required` — the fork-PR gate on a first-time contributor. **A maintainer needs to approve the workflow run**; until then this PR has no build/typecheck/test/lint signal on GitHub, and the local evidence in my review is that review's own, never a substitute for branch protection.

pkarw:
@Frshy — handing this back to you. The review requests changes, but the ask is small and the diagnosis in your description is correct; I have reassigned the PR to you rather than sitting on it.

**One thing to fix**, in `packages/scheduler/src/modules/scheduler/lib/scheduleAccess.ts:70-72`: the system-scope branch runs before any tenant comparison, so a row carrying `scopeType: 'system'` *and* a non-null `tenantId` now answers 403 to a non-super-admin from a different tenant, where the lookup you replaced answered 404. That is the one respect in which the new helper is less isolating than the old code, and it contradicts the invariant your own description sets out ("a 403 would confirm the id exists"). No current caller produces such a row — I checked every `schedulerService.register` call site, and the two `scopeType: 'system'` registrations in `ai-assistant` both omit `tenantId` — but `registerSchedule` applies no cross-field validation, so a third-party module can. The suggested conditional and the unit case to add are in the review body; both preserve all 17 existing cases, including the tenant-less super admin.

**Two nits and one description correction** are also in the review — the organization-divergence JSDoc describing only the narrower half of a two-directional divergence, the tenant-wide branch not consulting `scopeType` the way `buildSchedulerJobsFilters` does, and the locale count (four, not five). All author's call.

**Not something you can act on:** the `CI for Develop&Main` run for `673ad7d` is in `action_required`, the fork-PR gate for a first-time contributor, so nothing has actually run on GitHub. A maintainer needs to approve the workflow run. I fell back to local validation for the changed package only and reproduced your numbers exactly — 382 tests / 20 suites pass, and `tsc --noEmit` reports zero errors in any of the five changed files.

Labels are applied on your behalf per your metadata-intent comment, with `changes-requested` instead of `review` for now; rationale is in the separate label comment. Re-request review when the helper is updated and it will go back through the same pass.

pkarw:

One Minor finding drives the verdict — a cross-tenant existence disclosure (403 where the replaced lookup returned 404) in the new `resolveScheduleAccess` system-scope branch — plus three nits. No blockers or majors; the fix itself is correct and well-tested, and it closes a pre-existing cross-tenant hole on the unresolved-tenant path as a side effect.

CI follow-up: none scheduled. The `CI for Develop&Main` run for `673ad7d` is in `action_required`, so there is no pending run to watch — it is blocked on a maintainer approving the fork workflow, not executing. No `ci-monitoring` label applied, because no follow-up comment would ever arrive from this agent. Validation evidence gathered locally is in the review body.

Frshy:
Thanks — this is an unusually careful review, and the Minor finding is correct. Addressed in `c0b1a4e`.

## Minor — cross-tenant disclosure on a scope-inconsistent row: **fixed**

You're right, and it contradicted the helper's own stated invariant. The system-scope branch ran before any tenant comparison, so `{ scopeType: 'system', tenantId: 't2' }` answered 403 to an actor in `t1` where the replaced lookup answered 404 — the one respect in which this PR was *less* isolating than the code it replaced.

Applied your suggested shape, with `actorTenantId` hoisted above the branch:

```ts
if (schedule.scopeType === 'system' || scheduleTenantId === null) {
  if (isSuperAdmin) return 'allowed'
  if (scheduleTenantId !== null && scheduleTenantId !== actorTenantId) return 'not_found'
  return 'forbidden'
}
```

All 17 prior unit cases still pass unchanged, including the tenant-less super admin reaching system schedules. Your reachability analysis matches mine: no current caller produces such a row, but `schedulerService.register` writes `scopeType` and `tenantId` through with no cross-field validation, so a third-party module can — which is exactly why it deserves a test rather than a comment.

## Test gaps 1 and 2: **both added**

1. `{ scopeType: 'system', tenantId: 't2' }` vs a `{ tenantId: 't1' }` non-super-admin → `not_found`, alongside the existing same-tenant `forbidden` case.
2. The cross-tenant leak this PR *closes* now has its own test: `{ tenantId: null, orgId: 'o1' }` against an org-bound row in another tenant → `not_found`. You're right that this was the shape the old code actually leaked and that it had no coverage — the existing fail-closed case used `orgId: null`, which is the easier half.

384 tests / 20 suites pass.

## Nits 1 and 2: **both documented**

The organization JSDoc now names **both** directions, with the permissive one stated plainly: an actor with no `orgId` reaches any org-bound row inside its tenant, while the list hides those rows entirely. You're right that a reader of the helper alone would previously have concluded it could only ever be more restrictive. The tenant-wide branch not pinning `scope_type = 'tenant'` the way `buildFilters.ts:51-55` does is now on the record next to it.

Follow-up filed as requested: **#5385**, covering all three divergences and what a real fix would involve (both routes are hand-rolled handlers with no `ctx.organizationIds`, so it's a behaviour call, not a refactor).

## Nit 3 — locale count: **the description is right, there are five**

Gently pushing back on this one:

```
$ ls packages/scheduler/src/modules/scheduler/i18n/
de.json  en.json  es.json  ko.json  pl.json
```

`ko.json` was added in "feat(i18n): add Korean locale support (#4912)". Your substantive point stands either way — both reused keys are present in every locale and no key changed — and `yarn i18n:check-sync` reports "All translation files are in sync" across **five** locales (`en, pl, es, de, ko`). No change made.

## On the validation gate

Agreed that your local run is not a substitute for branch protection — CI still needs a maintainer to approve the workflow run on this fork PR. For what it adds: I ran the full ordered gate from `.ai/agentic.config.json` locally (`build:packages` → `generate` → `build:packages` → `i18n:check-sync` → `i18n:check-usage` → `typecheck` → `test` → `build:app`), and the integration suite via `yarn test:integration:ephemeral --filter TC-SCHED` at 16/16, including the regression proof against `develop`. Your `typecheck`/`test` figures match mine exactly.

Thanks also for spotting that the change closes a pre-existing cross-tenant hole rather than only fixing a 404 — you're right that the description undersold it, and it now has a test pinning it.

Frshy:
@pkarw — ready for re-review when you have a moment.

Everything in the review is addressed as of `c0b1a4e` (details in the [reply above](https://github.com/open-mercato/open-mercato/pull/5381#issuecomment-5341656643)):

- **Minor — cross-tenant disclosure on a scope-inconsistent row:** fixed, using your suggested shape. A tenant-bound `scopeType: 'system'` row now answers `not_found` across a tenant boundary; only a genuinely tenant-less row answers 403.
- **Test gap 1:** `{ scopeType: 'system', tenantId: 't2' }` vs a `t1` non-super-admin → `not_found`.
- **Test gap 2:** the cross-tenant leak this PR closes is now pinned — `{ tenantId: null, orgId: 'o1' }` against an org-bound row in another tenant.
- **Nits 1 & 2:** the organization divergence is documented in both directions, and the tenant-wide branch's looseness about `scope_type` is on the record. Follow-up filed as #5385.
- **Nit 3 (locale count):** no change — there are five locales; `ko.json` landed in #4912. Substance of your point was right either way, and `i18n:check-sync` is clean across all five.

384 tests / 20 suites pass; full ordered gate from `.ai/agentic.config.json` green locally.

Two things I can't do from a read-only account, both needing a maintainer:

1. **Approve the CI workflow run.** It's still `action_required` on this head — only `license/cla` (pass) and `[code]smith` (skipped) have reported, so there's no GitHub build/test signal yet.
2. **Move the pipeline label** from `changes-requested` back to `review`, if you're satisfied with the fixes.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-20T23:56:17Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# Re-review: fix(scheduler): manual trigger and execution history 404 on every system-scoped schedule

**Verdict: approve.** Every finding from the [previous review](https://github.com/open-mercato/open-mercato/pull/5381#pullrequestreview) is addressed at `c0b1a4e` — the Minor with the exact shape suggested, both test gaps with real cases, both documentation nits, and a follow-up (#5385) filed for the three inherited divergences. One optional Nit below; it does not block.

Re-review of the three commits added since the first pass (`11815d6d`, `673ad7d0`, `c0b1a4e`), plus a full re-check of the original blocker areas (tenant isolation, super-admin gating, 404-vs-403 disclosure, contract surfaces).

## ✅ Validation

**Validation source: GitHub checks, with a narrow local fallback for `test`/`typecheck` while the CI run was still mid-flight.**

The `CI for Develop&Main` run for `c0b1a4e` (run [32249082734](https://github.com/open-mercato/open-mercato/actions/runs/32249082734)) is executing as this review is written: `prepare` (build packages → generate → rebuild → build app) completed **green**, and `test` plus 15 `ephemeral-integration` shards are in progress. `lint`, `ds-lint`, `audit`, `audit-scope`, `scope` and `license/cla` are all **SUCCESS**. Nothing is failing.

| Gate | Source | Result |
|---|---|---|
| `yarn build:packages` / `yarn generate` / `yarn build:app` | GitHub — `prepare` job | ✅ PASS |
| `yarn lint` | GitHub — `lint` job | ✅ PASS |
| DS compliance | GitHub — `ds-lint` job | ✅ PASS |
| Dependency/scope audit | GitHub — `audit`, `audit-scope`, `scope` | ✅ PASS |
| `yarn test` | GitHub — `test` job ⏳ in progress; local fallback run | ✅ PASS locally — `packages/scheduler`, **384 tests / 20 suites**, matching the author's reported figures (382 → 384 is exactly the two new cases) |
| `yarn typecheck` | GitHub — bundled in the pending stage; local fallback run | ✅ PASS — `tsc --noEmit` reports **zero** errors in any of the five changed files. The 14 it does report are pre-existing, in untouched backend `.tsx` pages (`@tanstack/react-table/legacy` unresolvable in a linked review worktree, cascading to `TS7031`) |
| `yarn i18n:check-sync` | verified directly | ✅ PASS — `scheduler.error.not_found` and `scheduler.error.access_denied` are present in all **five** module locales (`de`, `en`, `es`, `ko`, `pl`). No key added or removed |
| Integration (`TC-SCHED-008`) | GitHub — `ephemeral-integration` shards ⏳ in progress | Author reports 16/16 locally plus a regression proof against `develop`; the CI shards now cover it for real |

On the locale count: **you were right and I was wrong** — `ko.json` exists, there are five. Thanks for checking rather than accepting it.

## 🔁 Previous findings — all resolved

**Minor (cross-tenant existence disclosure on a scope-inconsistent row) — fixed.** `scheduleAccess.ts:80-88` now hoists `actorTenantId` above the system branch and returns `not_found` for a row carrying `scopeType: 'system'` with a foreign non-null `tenantId`. Verified against the 403-only invariant: a genuinely tenant-less row still answers `forbidden` (its existence is a deployment property, not a tenant's), a same-tenant mislabelled row still answers `forbidden` (the actor is inside that tenant already), an unresolved-tenant actor gets `not_found`, and a super admin is unaffected. The tenant-less-super-admin case that motivated the shape still returns `allowed`.

**Test gap 1 — added.** `scheduleAccess.test.ts:75-82` pins `{ scopeType: 'system', tenantId: 't2' }` against a `t1` non-super-admin as `not_found`, next to the same-tenant `forbidden` case. This is the case a future refactor would silently reintroduce.

**Test gap 2 — added.** `scheduleAccess.test.ts:112-119` pins the leak this PR *closes*: `{ tenantId: null, orgId: 'o1' }` against an org-bound row in another tenant → `not_found`. That is the exact shape the replaced lookup matched (`{ id, deletedAt: null, organizationId: 'o1' }` with no tenant clause), so the security improvement is now regression-proof rather than incidental.

**Nit 1 (organization divergence documented in one direction only) — fixed.** `scheduleAccess.ts:56-69` now names both directions explicitly and labels the permissive one as such. A reader of the helper alone can no longer conclude it is only ever more restrictive than the list.

**Nit 2 (tenant-wide branch does not pin `scope_type = 'tenant'`) — fixed.** On the record at `scheduleAccess.ts:93-96`, next to the other divergences, with the reason for leaving it loose.

**Nit 3 (locale count) — correctly rejected.** See above.

**Follow-up #5385** filed for all three divergences. That is the right disposition: reconciling the single-`orgId` comparison with the list endpoint's resolved scope is a behaviour decision, not a refactor, and both routes are hand-rolled handlers with no `ctx.organizationIds` to consult.

## 🔍 New in these commits — verified

**Behaviour delta beyond the previous revision: an unresolved actor tenant now fails closed for super admins too.** `11815d6d` changed `if (actorTenantId === null) return isSuperAdmin ? 'allowed' : 'not_found'` to an unconditional `not_found` (`scheduleAccess.ts:90`), reconciling the code with what the PR description always claimed. This is stricter than both the previous PR revision *and* `develop` (where a falsy `auth.tenantId` simply omitted the tenant clause). I checked the reachability and the consistency argument holds:

- `buildSchedulerJobsFilters:26-30` returns an unmatchable filter whenever the tenant is falsy — **for super admins as well** — so such an actor cannot see a tenant-bound schedule in the list either, and therefore has no UI path to a trigger button for one.
- The system-scope branch is evaluated first, so a tenant-less super admin still reaches system-scoped schedules, which is this PR's whole point.
- `auth.tenantId` is not nullable by tenant selection: `resolveTenantOverride` (`shared/lib/auth/server.ts:111-118`) deliberately reports a blank cookie as *not applied* rather than as a deliberate null, and there is no all-tenants sentinel for tenants the way `__all__` exists for organizations.

So the tightening is consistent with the list endpoint, unreachable through normal session state, fail-closed, tested (`scheduleAccess.test.ts:123-127`), and documented in the JSDoc. Recorded as an intentional delta rather than a finding.

**Newly reachable code path checked end to end.** The manual trigger for a system-scoped schedule was dead before this PR, so its downstream is effectively new in production terms. `trigger/route.ts:78-85` builds `ExecuteSchedulePayload` from the **schedule's own** scope (`schedule.tenantId`, `schedule.organizationId`, `schedule.scopeType`) rather than the actor's, and `execute-schedule.worker.ts:99-112` re-validates the payload scope against the DB row before executing and then works from the row. A manual trigger therefore runs a system-scoped job in exactly the scope its cron firing would, including the `requireFeature` gate. No scope confusion is introduced by making this path reachable.

**Route-level cleanup in `c0b1a4e` is a strict improvement.** Both routes moved from `schedule ? resolveScheduleAccess(...) : 'not_found'` plus a combined `if (!schedule || access === 'not_found')` to an early `if (!schedule)` return followed by the decision branch. Same responses, no nullable intermediate, and the "row does not exist" and "row is hidden from you" cases stay deliberately indistinguishable to the caller.

**Parity claims re-verified against the sources they cite**, not taken at face value: `ensureCanManageSystemScopedJob` (`commands/jobs.ts:175-185`) classifies system scope as `scopeType === 'system' || tenantId == null` — identical to the helper; `updateScheduleCommand:304` and `deleteScheduleCommand:446` both load by `{ id, deletedAt: null }` and apply isolation to the loaded row, which is the shape this PR adopts; `buildSchedulerJobsFilters:52-63` pins `scope_type = 'tenant'` on its org-null branch and gates the system branch on `isSuperAdmin`, as the JSDoc says.

## Findings

### 💅 Nit (optional, non-blocking)

**`packages/scheduler/src/modules/scheduler/lib/scheduleAccess.ts:80-88` — a fourth divergence in the same scope-inconsistent-row class, worth one clause in the existing note.**

The classification matches `ensureCanManageSystemScopedJob` exactly, as documented. What differs is the *ordering* relative to tenant scope: `updateScheduleCommand` and `deleteScheduleCommand` call `ensureTenantScope(ctx, schedule.tenantId)` (`commands/jobs.ts:158-162`) **before** the system-scope check, and that helper throws "Tenant mismatch" for anyone — super admins included — when both tenants are set and differ. The helper here evaluates the system branch first, so for a row carrying `scopeType: 'system'` with `tenantId: 't2'`, a super admin scoped to `t1` gets `allowed` on trigger/executions but "Tenant mismatch" on update/delete.

Same unproducible-row class as the two divergences already documented (the create validator forbids `system` with a tenant; only a direct `schedulerService.register` call can write one), and the effect grants access to a super admin only — no privilege escalation for anyone else, and arguably the more correct answer of the two. Purely a documentation point: one clause on the existing note at `:82-85` would keep the "these four paths agree about system scope" claim exact. Fine as a comment-only follow-up on #5385, or to leave entirely.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed — `ScheduleScopeType` and `ScheduleScopeSubject.scopeType` are **additive**, and `ScheduleScopeSubject` keeps every field optional, so existing call shapes still typecheck.
- [x] No function signature changed in a breaking way — `resolveScheduleAccess(schedule, actor)` is unchanged in arity, parameter types (widened, not narrowed) and return union.
- [x] No HTTP route URL, method, guard or request schema changed — `POST /api/scheduler/trigger` and `GET /api/scheduler/jobs/[id]/executions` are untouched in path, method, `requireAuth`/`requireFeatures` and request validation.
- [x] No response field removed or retyped — success and error bodies are byte-identical; only *which* status a given request receives changes, and 400/401/403/404 are already declared in each route's `openApi.errors`, so generated docs need no update.
- [x] No event, message or payload field changed — `ExecuteSchedulePayload` is built exactly as before.
- [x] No database table or column touched; correctly no migration and no snapshot delta.
- [x] No CLI command, config key, DI key, ACL feature, widget spot or notification id touched.
- [x] Nothing in `BACKWARD_COMPATIBILITY.md`'s thirteen protected surfaces is affected.

## 🧪 Test Coverage

**19 unit cases** on the real `resolveScheduleAccess` with no mocks (up from 17), across four groups: system-scope access, system-scope *classification* parity with `commands/jobs.ts`, tenant isolation, organization isolation. The suite carries the properties that matter for authorization code — the anti-spoofing case (a role literally named `superadmin` without the `isSuperAdmin` flag gets nothing), fail-closed cases for both roles on an unresolved tenant, `not_found`-not-`forbidden` for every cross-boundary case, and `undefined`-vs-`null` normalization. The two divergence tests (`:112-119`, `:147-151`) document known inconsistencies with comments explaining they pin rather than endorse, which is the right way to encode inherited behaviour. Whole package: **384 tests / 20 suites green** in my run.

**`TC-SCHED-008`** now covers both routes end to end rather than just the trigger: super admin `not.toBe(404)` on each, tenant `admin` exactly `403` on each, strategy-conditional assertions (`200` under `async`, the pre-existing `400` under `local`) that are honest about what a local run proves, and cleanup in `finally`. I re-checked that the 403 assertions are meaningful rather than incidental — `setup.ts` grants `admin: ['scheduler.*']`, so that fixture holds both `scheduler.jobs.trigger` and `scheduler.jobs.view` and reaches the scope check instead of being stopped by `requireFeatures` — and that the executions route's `400` under `local` is the real strategy branch (`executions/route.ts:61-68`), not an error path. `SCHEDULER_JOBS_PATH` is exported by the helpers module the spec imports.

**Remaining gap, accepted:** cross-tenant coverage is unit-level only, because `packages/core/src/helpers/integration/` has no second-tenant fixture and a self-contained cross-tenant integration test is not expressible today. Stated by the author rather than papered over; not a reason to hold this PR.

## Merge readiness

- ⏳ **CI is mid-run and must finish green.** `prepare` (including `build:app`) is green; `test` and 15 `ephemeral-integration` shards are still executing on `c0b1a4e`. The merge gate holds until they land — my local run is this review's own evidence, never a substitute for branch protection.
- ℹ️ The branch is 188 commits behind `develop` but reports `MERGEABLE` with no conflicts. Squash-merging from the PR is fine; no rebase is required for correctness.
- ✅ `skip-qa` remains correct — the automated-verification exemption is met on every clause (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, database structure unchanged, API surface unchanged, no compatibility contract broken, automated tests for the changed behaviour in the same PR).
- `priority-medium` and `risk-medium` are unchanged and still fit: the capability was fully broken but the unattended cron still fired these jobs, and the blast radius stays inside `packages/scheduler` with the decision isolated in one pure, heavily-tested function.

Nice work on the pushback about the locale count, and on turning both suggested test cases into real regression pins rather than assertions that merely restate the fix.

pkarw:
Label set to `merge-queue` because the re-review passed and QA is not required (`skip-qa`); `priority-medium`, `risk-medium` and `partner-request` are unchanged, and the merge still waits on the in-flight CI run for `c0b1a4e`.

pkarw:

Re-review of `c0b1a4e`: every finding from the first pass is resolved — the Minor cross-tenant disclosure with the exact shape suggested, both test gaps as real regression pins, both documentation nits, and #5385 filed for the inherited divergences. One optional, non-blocking Nit remains (a fourth same-class divergence worth a clause of comment). No autofix iterations were needed.

Validation source: GitHub checks (`prepare`/`lint`/`ds-lint`/`audit`/`audit-scope`/`scope`/`license/cla` green), with a narrow local fallback for `test` and `typecheck` while the run was mid-flight — 384/384 scheduler tests pass and zero typecheck errors in the five changed files. `test` and the 15 `ephemeral-integration` shards are still running on `c0b1a4e`; `ci-monitoring` is on until they land, and the merge gate holds until then.

pkarw:
✅ CI is green on `c0b1a4e` — run [32249082734](https://github.com/open-mercato/open-mercato/actions/runs/32249082734) finished with **24 checks SUCCESS, 0 failing** (4 skipped), including `test` and all 15 `ephemeral-integration` shards, so `TC-SCHED-008` now has real CI coverage rather than only the author's local run. Dropping `ci-monitoring`.

This PR is approved, green and QA-exempt (`skip-qa`) — ready for a maintainer to squash-merge.

Issue #5373, opened 2026-08-18
feat: Injection widget data inaccessible from server-side subscriber after entity creation


leonbahley:
## Problem

When an `InjectionFieldWidget` is injected into a `CrudForm`, there is no supported way to combine:

1. **Data collected by the widget** (client-side form fields), and
2. **The newly created record's ID** (only available after the server responds)

…in order to create a related record in a different table.

### Current behavior

- `context.recordId` / `context.resourceId` inside `onAfterSave` are `undefined` on create operations. This is expected — `injectionContext` is memoized from `values.id` (see `CrudForm.tsx` lines ~830/865) and on a brand-new entity that ID doesn't exist yet.
- The server's `{ id }` response flows to the host page's `onSubmit` but is never propagated back into the widget context.

### Why the standard workaround doesn't fully apply

The recommended pattern for "do something after create" is a **persistent event subscriber** on the CRUD event (e.g. `customers.person.created`). However, the subscriber only receives the event payload (the new entity's ID and its persisted fields). It does **not** have access to the extra widget-collected data that was only present in the browser form — in this case, a "relation type" and a "related person ID" that the user selected in the injected widget.

### Use case

Module: app-level `customer_relations`
Core event consumed: `customers.person.created` (from the `@open-mercato/core` npm package)
The `customers` module (including the `/backend/customers/people/create` page and its `CrudForm`) is a core platform module consumed from `@open-mercato/core` via npm — it cannot be modified at the app level. The app-level module can only interact with it through the extension system (injection widgets, interceptors, event subscribers).

**Flow:**

1. User opens `/backend/customers/people/create`.
2. An injection widget adds fields: "Related person" (lookup) + "Relation type" (select, e.g. "father").
3. User submits the form → a new `customer_person` is created.
4. Immediately after, a `customer_relation` row should be created linking the new person to the selected related person with the chosen relation type.

**Blocking issue:** There is no documented mechanism to get the widget's field values (relation type, related person ID) to the place that knows the new record's ID (the subscriber or the post-creation callback).

### Considered approaches

| Approach | Limitation |
|----------|-----------|
| `onAfterSave` in widget | `context.recordId` is `undefined` on create |
| Persistent subscriber on `customers.person.created` | Has the new ID but not the widget's form data |
| API interceptor (`after` hook on `POST /customers/people`) | Could read extra body fields, but injection widgets don't merge their values into the CRUD request body today |

## Proposed solution

A supported pattern for injection widgets to either:

[...]

PR #5414, opened 2026-08-19
docs(specs): support CrudForm injected widget payloads

Paul-Mlodochowki:
Refs #5373

This design-only PR specifies a request-scoped widget payload for CrudForm injected fields. It keeps entity payloads unchanged and makes the sidecar available to API interceptors.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5414"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789760219&installation_model_id=432243&pr_number=5414&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5414&signature=c9865bfee57f8da44d900c151f6b7d21defbc4d5657b51d829bb6004fb54c07d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.ai/specs/2026-08-19-crudform-injected-widget-payload.md
+++ b/.ai/specs/2026-08-19-crudform-injected-widget-payload.md
@@ -1,0 +1,78 @@
+# CrudForm Injected Widget Payloads
+
+## 📝 TLDR
+
+Make values collected by `InjectionFieldWidget` fields available to the server-side request lifecycle that already knows the created record ID. `CrudForm` will publish those values through an explicit, request-scoped sidecar that is merged into requests made by the existing `apiCall`/`createCrud` helpers. API interceptors and CRUD lifecycle code can consume the sidecar without changing entity schemas or requiring every host page to forward a new callback argument.
+
+## Resolved assumptions (autonomous defaults)
+
+- **Use a request-body sidecar rather than changing `onAfterSave` signatures.** This is additive and works with existing host pages such as the core customers create form.
+- **Keep the sidecar namespaced and out of entity values.** The sidecar is available to API interceptors as a reserved property and is removed from the entity payload before persistence; this avoids accidental schema or database writes.
+- **Expose the sidecar to API interceptors, not persistent event payloads in this phase.** Interceptors already receive the original request and response together, which is sufficient for the issue's transactional use case and avoids changing every CRUD event type.
+- **Do not add a new UI surface.** Existing injected fields and CRUD forms remain visually unchanged; browser verification will exercise the existing customers create route where available.
+
+## 📝 Problem Statement
+
+Injected CrudForm fields are rendered and validated by the browser, but their values are removed before the host `onSubmit` receives the entity payload. On create, `onAfterSave` also runs with a context derived before the response, so a widget cannot reliably combine its values with the new record ID. The documented extension-header workaround is functional but relies on an undocumented convention and requires a custom interceptor/header protocol per module.
+
+## 📝 Proposed Solution
+
+Add a public `CrudForm` request-payload contract. During submit, CrudForm collects active injected field values under a reserved sidecar key and opens a request-scoped body context while it invokes the unchanged host `onSubmit` callback. `apiCall` merges the sidecar into JSON request bodies created within that scope. On the server, CRUD API interceptors receive the sidecar on both `before` and `after` hooks through the existing `InterceptorRequest.body`; the CRUD factory strips the reserved key before mapping entity input so persistence behavior is unchanged.
+
+The sidecar is intentionally opt-in to the form's injected fields, absent when no injected values exist, and limited to JSON-compatible data. Nested values are preserved. Scope nesting follows the existing request-header stack so concurrent or nested calls do not leak payloads across requests.
+
+## 📝 Architecture
+
+- `packages/shared/src/lib/crud/widget-payload.ts` owns the reserved key, JSON-safe normalization, and extraction helpers shared by UI and CRUD factory code.
+- `packages/ui/src/backend/utils/apiCall.ts` gains a request-scoped body stack and merges its resolved sidecar into object JSON request bodies.
+- `packages/ui/src/backend/CrudForm.tsx` derives the injected field subset from the current form values, wraps `onSubmit` in the body scope, and continues passing only schema-approved core values to the host callback.
+- `packages/shared/src/lib/crud/factory.ts` removes the reserved sidecar from the entity input after interceptors have observed it. The original `InterceptorRequest` remains the source of truth for the sidecar.
+- Existing event handler signatures, injection spot IDs, API routes, entity schemas, and event IDs remain unchanged.
+
+## 📝 API Contracts
+
+The request JSON may contain the additive reserved property:
+
+```ts
+{
+  name: 'New person',
+  __omWidgetPayload: {
+    'customer_relations': {
+      relatedPersonId: 'person-id',
+      relationType: 'father'
+    }
+  }
+}
+```
+
+The property is for the request lifecycle only. API interceptors read it from `request.body?.__omWidgetPayload`; entity mapping and persistence never receive it. The exact namespace key is the injecting module/widget's stable module identifier, and values are constrained to JSON-compatible primitives, arrays, and objects. Existing requests without the property are byte-for-byte behavior-compatible.
+
+## 📝 Edge Cases & Failure Scenarios
+
+- No injected fields or no values produces no sidecar and no additional request property.
+- Hidden or disabled injected fields are excluded using the same active-field set used by CrudForm validation.
+- A malformed/non-JSON value is omitted from the sidecar rather than breaking the entity save; normal form validation remains responsible for user-facing errors.
+- If a host performs multiple CRUD calls during one submit, each call receives the same scoped sidecar, enabling a create interceptor to use the created ID from its response while keeping the related write transactional where the host route supports it.
+- Nested request scopes restore the previous sidecar on completion, including rejected promises.
+- Direct `apiCall` calls outside a CrudForm are unchanged.
+
+## 📝 Risks & Impact Review
+
+This is a public additive request-shape and helper contract. The reserved property must be documented and stripped before entity validation/persistence. No migration or dependency is required. The main risk is accidental sidecar leakage to unrelated nested calls; stack-based scoping and tests for restoration mitigate it. Existing `onBeforeSave` request headers remain supported for compatibility.
+
+## 📋 Phasing
+
+Phase 1 ships the shared scope, CrudForm wiring, server stripping, documentation, and regression tests together. It is independently usable by any injected widget and API interceptor.
+
+## 📋 Implementation Plan
+
+1. Add shared widget-payload constants and helpers, plus request-body scope support in `apiCall`; cover nesting, cleanup, JSON merging, and unchanged requests with unit tests.
+2. Wire `CrudForm` to collect active injected values and scope them around the existing `onSubmit`; add component regression coverage proving the host receives core values while the request receives the sidecar.
+3. Strip the reserved sidecar before CRUD entity mapping while preserving it for interceptor before/after hooks; add CRUD-factory coverage for create and update requests.
+4. Document the supported injected-widget server-side pattern with a transactional relation-creation example and run package generation/build/type checks.
+5. Exercise the existing backend customer create form with Playwright, confirming the form remains usable and the request path completes without exposing the sidecar as an entity field.
+
+## Integration Coverage
+
+- API: CRUD POST and PUT requests with an injected widget sidecar reach interceptor before/after hooks, while persisted entity input excludes the reserved key.
+- UI: an existing CrudForm with injected fields submits successfully, retains normal validation, and does not render the reserved sidecar as a visible entity field.


Paul-Mlodochowki:
This PR is intentionally design-only. It contains the specification for issue #5373; implementation is tracked separately in PR #5415, which references this spec via `Refs #5414` and closes the issue.

The spec file is the single changed file in this PR:
`.ai/specs/2026-08-19-crudform-injected-widget-payload.md`

Implementation PR: https://github.com/open-mercato/open-mercato/pull/5415

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-20T00:43:21Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Specification Review

**PR #5414 — `docs(specs): support CrudForm injected widget payloads`**
**Verdict: 🔴 Changes requested** — three blockers, six majors.

## 🎯 Summary

This is the spec that PR #5415's review named as a hard prerequisite ("the spec is referenced but never shipped"), so landing it is the right move and the problem it addresses — issue #5373, injected widget values never reaching the place that knows the created record ID — is real and well stated. The chosen shape (a request-scoped, module-keyed sidecar that interceptors can read and that never reaches the entity mapper) is a sound instinct.

The document does not yet hold up as the design of record. Its Architecture section describes a mechanism that the current codebase cannot execute as written, its stripping order is specified one step too late to be safe, it declares a new public wire key without the Migration & Backward Compatibility treatment the repo requires for a contract surface, and it rejects the existing UMES extension channel on a premise that is factually incorrect. Every finding below is grounded in a file and line on `develop`; where the implementation PR #5415 already proved a failure mode empirically, I cite its review rather than re-deriving it.

Because this spec and #5415 are the same design at two altitudes, findings 2, 4 and 6 here are the spec-level form of #5415's blockers 2 and 3 and its major 5. Fixing them in the spec first is the cheaper order — the implementation then follows the corrected document rather than the reverse.

## 🧪 Validation Gate

**Validation source: GitHub checks.** The PR head `1ef08f0` reports `license/cla` ✅ pass and a skipped `[code]smith` check; no required check is failing and none is pending. This is a single-markdown-file diff, so the configured `validation.commands` gate has nothing to compile, generate, typecheck, test, or build.

| Command | Result |
|---|---|
| `yarn build:packages` | ⏭️ skipped — no package source in the diff |
| `yarn generate` | ⏭️ skipped — no auto-discovery files in the diff |
| `yarn i18n:check-sync` / `yarn i18n:check-usage` | ⏭️ skipped — no locale files or user-facing strings in the diff |
| `yarn typecheck` | ⏭️ skipped — no TypeScript in the diff |
| `yarn test` | ⏭️ skipped — no code under test |
| `yarn build:app` | ⏭️ skipped — no app source in the diff |
| `.ai/specs` conventions (`AGENTS.md` → Validation Commands) | ✅ pass — `2026-08-19-crudform-injected-widget-payload.md` matches `{YYYY-MM-DD}-{kebab-case}.md`, carries no forbidden `SPEC-*` prefix, and sits in the OSS specs directory, which is correct for this scope |

## 💥 What can go wrong

### 🚨 Blocker 1 — the Architecture section describes a data flow the codebase cannot perform

The spec states, in **Proposed Solution** and again in **Architecture**:

> On the server, CRUD API interceptors receive the sidecar on both `before` and `after` hooks through the existing `InterceptorRequest.body`; the CRUD factory strips the reserved key before mapping entity input.

On `develop` that is not what happens. `applyInterceptorsBefore` is handed the **already-Zod-parsed** input, never the raw body — `packages/shared/src/lib/crud/factory.ts:2242` (direct create), `:2564` (direct update), `:2115` and `:2430` (the command-bus `action.schema` paths). A plain `z.object` strips unknown keys, so `__omWidgetPayload` is gone *before* any interceptor runs. This is not incidental behavior; it is a documented invariant of the surface this spec extends — `.ai/specs/implemented/SPEC-041e-api-interceptors.md:108` states "Interceptor `before` runs AFTER Zod validation" and illustrates it with `schema.parse(rawBody)` followed by `runInterceptorsBefore(parsedInput, ctx)`, and `:318` states that "unknown fields added by interceptors are stripped by Zod's `.parse()`".

The design only works if the factory **explicitly re-attaches** the sidecar into the interceptor request after validation — which is exactly what #5415 had to add (`addCrudWidgetPayload`). The spec never mentions that step, and its phrasing ("strips the reserved key before mapping entity input") implies the opposite: that the key survives validation on its own and is merely removed later.

A spec that mis-describes the mechanism it depends on propagates the error into every implementation built from it. **Fix:** rewrite Architecture to state the actual sequence — extract the sidecar from the raw body, parse the entity input without it, re-attach it onto the interceptor request payload, and strip it again before re-validation — and add SPEC-041e (`:108`, `:318`) to the list of documents this change amends, since re-attaching after validation makes `request.body` carry a key that was never validated.

### 🚨 Blocker 2 — the specified stripping point is one step too late, so strict route schemas reject the request

**Architecture** places the strip after interceptors have observed the key ("removes the reserved sidecar from the entity input after interceptors have observed it"). That leaves the *first* `schema.parse(body)` unprotected. A `z.object` silently strips the unknown key, but a `.strict()` schema rejects it outright, and the route returns 400 before any of this design's logic runs.

This is not hypothetical. #5415's review reproduced it on both factory branches (`Expected: 201 Received: 400`, for both a strict action schema and a strict direct-create schema). Strict schemas are in active use in this repo — `packages/core/src/modules/customers/api/people/[id]/emails/route.ts:40`, and `packages/core/src/modules/workflows/data/validators.ts:733` and `:1031`, where the code comment records that strictness has already caused "Unrecognized keys" failures for a form submitting a full payload.

**Fix:** the spec must mandate stripping the reserved key from the raw body **before the first schema parse**, on all four factory branches, and must state explicitly that the guarantee does not rely on the repo-wide convention of non-strict schemas. Add a strict-schema case to the Integration Coverage section so the guarantee is testable rather than conventional.

### ⚠️ Major 1 — the ambient scope is specified to leak widget data into unrelated requests

**Edge Cases** presents this as a feature:

> If a host performs multiple CRUD calls during one submit, each call receives the same scoped sidecar.

Because the scope wraps the whole of `onSubmit` and stays open across network I/O, that also means: secondary writes to *other* entities during a submit carry another module's widget values (root `AGENTS.md` explicitly contemplates such forms in its optimistic-locking rule), and any concurrent `apiCall` elsewhere in the app — autosave, a refetch, an SSE-triggered reload — picks the payload up too. Against a stripping route this is silent leakage of user-entered data to endpoints that have no business seeing it; against `customers/api/people/[id]/emails/route.ts:40` it is a 400.

The spec's **Risks & Impact Review** names "accidental sidecar leakage to unrelated nested calls" but mitigates only the *nesting* case, with stack-based scoping and restoration tests. Restoration is not targeting: a correctly-restored stack still attaches the payload to every in-scope call. **Fix:** specify targeting — bind the scope to the method and URL the form is about to call, or pass the payload explicitly to the submit call instead of ambient-injecting it — and state the behavior for calls that do not match.

### ⚠️ Major 2 — the payload is untrusted client input and the spec never says so

**API Contracts** constrains the sidecar's *types* but says nothing about trust. The only exclusion the spec describes — "Hidden or disabled injected fields are excluded using the same active-field set used by CrudForm validation" — is a client-side filter. Server-side, any authenticated caller can post any namespace to any CRUD route, whether or not the contributing module is installed, whether or not the widget is visible to them, and whether or not they hold its ACL feature.

Given this repo's RBAC rules (feature-based guards, never trust client-supplied scope), a public contract that feeds interceptor code needs to state plainly that the payload is untrusted, and require consumers to validate it with Zod and re-check the contributing module's ACL feature before writing anything. #5415's review found the documentation example doing precisely the opposite — creating a related row straight from the payload with no validation and no feature check — which is the predictable outcome of a spec that stays silent here.

### 🔹 Minor 1 — no bound on payload depth or size

Neither **API Contracts** nor **Edge Cases** caps nesting depth, key count, or total size, though the normalization is specified to recurse over client-controlled input and the result rides on every request in scope. State a bound.

## 🔁 Backward compatibility

### 🚨 Blocker 3 — a contract-surface change with no Migration & Backward Compatibility section

The spec introduces a reserved property that every `makeCrudRoute` POST/PUT will accept and every interceptor may read, plus a new public import path (`packages/shared/src/lib/crud/widget-payload.ts`). `BACKWARD_COMPATIBILITY.md` classifies API request/response schemas as **category 7 (STABLE)** — additive optional fields are permitted, but the deprecation protocol requires the change to "reference a spec with a 'Migration & Backward Compatibility' section", and root `AGENTS.md` requires following that document before touching any contract surface. This spec is that spec, and it has no such section.

Concretely, the document needs to add: the category-7 classification of the reserved key and the ADDITIVE-ONLY guarantee attached to it; a `BACKWARD_COMPATIBILITY.md` entry registering the key as reserved, so a module author knows it is a name they must not use for a real field; and the amendments this change forces on already-implemented specs — `SPEC-041e-api-interceptors.md:108` and `:318` (the "before runs after validation" invariant, per Blocker 1) and `SPEC-041g-crudform-fields.md:787` (the checklist item "Injected field values excluded from core schema validation and core submit payload", which this change qualifies). Leaving those stale violates the specs-folder rule against stale assumptions.

### ⚠️ Major 3 — `__omWidgetPayload` contradicts three repo conventions, and the name freezes on day one

This is the naming question specifically raised for cross-check against #5415's review, so here is the full comparison, verified independently against `develop`.

| Precedent | Repo convention | This spec |
|---|---|---|
| Cross-module request-scoped data reaching interceptors | `x-om-ext-<moduleId>-<key>` → `context.extensionHeaders` (`packages/shared/src/lib/umes/extension-headers.ts`, parsed at `factory.ts:1227`, exposed at `api-interceptor.ts:29`) | `request.body.__omWidgetPayload` — a second, parallel channel |
| Reserved sentinel keys | `__om_snake_case`, versioned when structured — `__om_encrypted_credentials_blob_v1`, `__om_secret_unchanged__`, `__om_checkout_empty__`, `__om_phone_auto__` | `__omWidgetPayload` — camelCase, unversioned |
| Reserved body-key prefixes | `cf_` / `cf:` for custom fields (`factory.ts:395`, `:403`) | a new one-off prefix |

Three problems follow from the name itself. It says **Widget**, but the payload is keyed by **moduleId** and the server cannot tell that a widget produced it — UMES already calls this concept *ext/extension*. It carries **no version discriminator**, so the wire shape is frozen from first release with no migration path, in a repo whose own structured sentinel (`__om_encrypted_credentials_blob_v1`) versions exactly this kind of envelope. And it places extension-supplied data in `request.body` rather than `context.*`, where UMES already puts it — which is what forces every route schema and every interceptor to know about a magic key, and is the root of Blocker 2.

**Recommendation, and this is the future-proofing ask:** keep the body sidecar as a *transport detail* — headers genuinely cannot carry structured JSON and are size-limited, so a body channel is justified — but do not make it the interceptor-facing surface. Strip it in the factory before any schema parse and surface it as `context.extensionPayload[moduleId][key]`, mirroring `context.extensionHeaders`. That turns this from a change to the accepted request shape of every CRUD route (category 7, STABLE) into an additive `InterceptorContext` field (category 2), keeps one mental model for "extension data on a request", and resolves Blocker 2 by construction. If a body key must remain part of the public contract, name it `__om_ext_v1` to match both the `x-om-ext-` and `_v1` precedents, and register it in `BACKWARD_COMPATIBILITY.md`.

### 🔹 Minor 2 — the new public API surface is unnamed

**Architecture** names the new file path but not what it exports. Import paths (category 4) and function signatures (category 3) are STABLE surfaces; the spec should name the constant, the type, and the helper signatures it is committing to, so the contract is reviewable before it ships rather than after.

## 🧩 What's missing

### ⚠️ Major 4 — required spec sections are absent

`.ai/specs/AGENTS.md` requires every non-trivial spec to include TLDR, Overview, Problem Statement, Proposed Solution, Architecture, Data Models, API Contracts, Risks & Impact Review, Final Compliance Report, and Changelog. Present: TLDR, Problem Statement, Proposed Solution, Architecture, API Contracts, Risks & Impact Review. Missing: **Overview**, **Data Models**, **Final Compliance Report**, **Changelog** — plus the Migration & Backward Compatibility section from Blocker 3.

### ⚠️ Major 5 — the supported submit paths are never stated

The mechanism is specified against `apiCall` only ("`apiCall` merges the sidecar into JSON request bodies created within that scope"). Not every `CrudForm` host submits that way — `packages/core/src/modules/workflows/backend/definitions/create/page.tsx:31` renders a `CrudForm` and submits through `apiFetch`, where the sidecar would never appear. The failure mode is a silent no-op with no diagnostic, which is the worst shape for an extension contract: the widget author sees nothing, the interceptor author sees nothing, and there is no error to search for.

State which submit helpers are supported, what happens on the unsupported ones, and how a developer diagnoses an absent sidecar. Relatedly, the spec does not address OpenAPI publication — `packages/shared/src/lib/openapi/crud.ts:150` and `:179` publish the CRUD request schemas, which would now accept a property absent from the published contract.

### 🔹 Minor 3 — the documentation surface list is incomplete

**Implementation Plan** step 4 commits to documenting "the supported injected-widget server-side pattern", which reads as the UI-side `widget-injection.md` alone. An interceptor author reads `apps/docs/docs/framework/api/extending-api.mdx` and `crud-factory.mdx`; the surface registry lives in `apps/docs/docs/framework/extensibility/current-surfaces.mdx`; and neither `packages/shared/AGENTS.md` (its `crud/` row) nor `packages/ui/AGENTS.md` (CrudForm Guidelines) currently mentions the scope helper or the reserved key. Name them in the plan.

### 🔹 Minor 4 — Integration Coverage names outcomes but not tests

Root `AGENTS.md` and `.ai/qa/AGENTS.md` require the integration tests for all affected API paths to ship in the same change. The section describes two behaviors in prose; it should name the concrete `__integration__` specs that will assert them, including the strict-schema case from Blocker 2.

### 💡 Nit 1 — no explicit Non-Goals

Issue #5373 offered an Option B (propagate the created record ID back into `onAfterSave`). **Resolved assumptions** implicitly declines it; a one-line Non-Goals section stating so explicitly would stop the alternative being relitigated in the implementation PR.

## 📈 How this specification can be improved

### 🔹 Minor 5 — the inner key shape is undefined

**API Contracts** specifies the outer namespace ("the injecting module/widget's stable module identifier", correctly snake_case in the example) but never the inner keys. The example implies field ids; the prose says "values". Pin it down — it is half of the wire contract.

### 💡 Nit 2 — "byte-for-byte behavior-compatible" overclaims

Requests *without* the property are unchanged, which is what the sentence means; "byte-for-byte" invites a stronger reading the design cannot support once the key is present. Say the narrower thing.

### 💡 Nit 3 — Phase 1 has no measurable acceptance criteria

**Phasing** bundles the whole feature into one phase with no exit criteria. Give it a short list of testable conditions — that is what makes the phase reviewable when the implementation claims completion.

### ⚠️ Major 6 — the stated reason for not reusing extension headers is factually wrong

**Problem Statement** dismisses the existing mechanism:

> The documented extension-header workaround is functional but relies on an undocumented convention and requires a custom interceptor/header protocol per module.

Both halves are inaccurate, and the sentence is self-contradictory ("documented … undocumented"). The `x-om-ext-<moduleId>-<key>` convention **is** documented — `apps/docs/docs/framework/data-integrity/concurrency-locking.mdx:149` names it as "the project's `x-om-ext-<moduleId>-<key>` convention", and `packages/ui/AGENTS.md:121` documents `CrudForm` auto-attaching one. It **is** public API — `buildExtensionHeader`, `parseExtensionHeaders`, and `getExtensionHeaderValue` are exported from `packages/shared/src/lib/umes/extension-headers.ts`. And it requires **no** per-module protocol: `factory.ts:1227` parses every such header into `context.extensionHeaders[moduleId][key]`, typed on `InterceptorContext` at `api-interceptor.ts:29`, for every route the factory serves. Optimistic locking ships on it today.

Issue #5373 made the same claim ("extension headers are not part of the public contract"), so the spec inherited it rather than checking it — but a spec's job is to check.

There are real reasons a header channel may not suffice: headers carry strings only, so structured values need serialization, and header size limits are tight. Those are legitimate grounds to prefer a body transport, and stating them would make the design decision defensible. As written, the spec rejects the existing channel on grounds that do not hold and never compares the two.

**The simpler shape, concretely:** keep the body as the transport for structured JSON, but surface it server-side as `context.extensionPayload[moduleId][key]`, mirroring `extensionHeaders` and populated at the same point in `factory.ts` (`:1227`). The reserved key is then a private transport detail stripped before the first parse rather than a public request-shape contract — no category-7 change, no validation-order hazard, no magic key that every route schema and every interceptor must know about, and one mental model for "extension data on a request" instead of two. This is the same conclusion #5415's review reached from the implementation side.

## Summary

Three blockers stand between this document and being the design of record: the Architecture section describes an interceptor flow the codebase cannot perform and leaves the SPEC-041e invariant it breaks unamended (1); the specified stripping point is one step too late, so any strict route schema 400s — already reproduced against the implementation (2); and a category-7 contract-surface change ships without the Migration & Backward Compatibility treatment `BACKWARD_COMPATIBILITY.md` and root `AGENTS.md` require (3).

On the naming question raised for cross-check: `__omWidgetPayload` should not ship as a public request-body key. It is camelCase where the repo's reserved sentinels are `__om_snake_case`, unversioned where the comparable structured sentinel carries `_v1`, and named for widgets when it is keyed by module. Surfacing it as `context.extensionPayload` alongside the `extensionHeaders` channel UMES already has keeps one mental model, downgrades the change from a STABLE request-shape change to an additive type change, and removes blocker 2 by construction. If the body key stays, `__om_ext_v1` is the name that matches both precedents.

Fixing these in the spec before #5415 follows is the cheaper order — the implementation then tracks a corrected document instead of the two drifting apart.

---

<sub>🤖 `om-auto-review-pr` · specification review (spec-only PR) · autofix: skipped (not my PR — re-run with `--autofix` to fix it here) · validation source: GitHub checks (no failing or pending required checks on `1ef08f0`; code gates not applicable to a docs-only diff) · cross-checked against the review of implementation PR #5415</sub>

pkarw:

- ❌ `changes-requested` — the specification review found three blockers (an Architecture section describing an interceptor flow the codebase cannot perform, a stripping point specified one step too late so strict route schemas return 400, and a category-7 contract-surface change with no Migration & Backward Compatibility section) plus six majors, so the document needs another pass before it becomes the design of record.
- 📚 `documentation` — the diff is a single specification document under `.ai/specs/`, with no source, config, or CI files touched.
- ⏭️ `skip-qa` — there is nothing for a human to click through: the change adds a markdown design document and renders no UI, alters no database structure, and changes no API surface at runtime.
- 🔹 `priority-medium` — this is an ordinary feature specification rather than an outage or security fix, though it does gate implementation PR #5415, whose own review named the missing spec as a blocker.
- 🟢 `risk-low` — the blast radius of merging this PR itself is a documentation file; the design it describes carries real risk, which is what the review findings address rather than the label.

pkarw:

Suggested order, since this spec and #5415 are the same design at two altitudes: correct the specification first (blockers 1–3 and the naming recommendation in major 3), then let the implementation follow the corrected document. That way the two stop drifting apart — #5415's review reached the same conclusion on the naming and the stripping order from the implementation side.

pkarw:

Specification review of a spec-only PR: three blockers and six majors, cross-checked against the review of implementation PR #5415 (naming conventions and the validation-order failure mode). Autofix was not run — this is another author's fork branch, so the run ends with the review, labels and handoff. No autofix iterations. No CI follow-up is owed: the head reports `license/cla` passing and a skipped `[code]smith` check, with nothing failing or pending.

pkarw:
@adeptofvoltron the situation like in #5415 - I’d prefer to fix or extend the features we have in UMES rather than adding his new out-of-the-blue feature

PR #4471, opened 2026-07-24
feat: add pluggable outbound email providers

pkarw:
Tracking plan: .ai/runs/2026-08-21-carry-pr-2448-ready.md
Status: complete

Supersedes #2448

Credit: original implementation by @pmadajthey. This carry-forward preserves the contributor’s email-provider work while bringing it onto the latest `develop` branch and applying review fixes.

## Included work

- routes application email through the Communications Hub
- adds pluggable Resend and Amazon SES provider packages
- preserves explicit email-delivery safety controls and established boolean environment tokens
- scopes the test-only capture API by tenant and organization and prevents credential persistence
- carries forward and expands the unit, integration, browser, and standalone-app harness coverage

## Review remediation

- fails closed for tenant credential lookup errors and missing tenant credentials; environment credentials remain limited to pre-tenant delivery
- protects system-email capture with a harness-only secret and opaque correlation token, while preserving tenant/organization capture isolation
- restores `OM_TEST_MODE` delivery suppression with a narrow `__test_seed__` capture opt-in
- fixes provider configuration detection, module dependency metadata, ACL catalogs, capabilities, coordinated versions, package previews, and SES region documentation
- adds bounded Resend and SES health checks with DI and contract coverage
- propagates tenant and organization scope through email OTP challenges
- keeps production-default CORS validation enabled and makes sales email integration coverage same-origin compliant
- adds mirrored desktop/mobile `/start` Playwright coverage and routes the new provider facts through the existing standalone-app harness case
- makes notification list caching organization-scope aware and invalidates cached reads after direct notification-service mutations
- stabilizes CRM, sales locking, CLI guide-parity, and module-facts performance checks against exact behavior instead of timing or fixture ambiguity

## Validation

Runner: local plus GitHub Actions

- `yarn build:packages` before and after generation
- `yarn generate`
- `yarn i18n:check-sync`
- `yarn i18n:check-usage`
- `yarn typecheck`
- full 26-workspace test matrix with 1,024 MB Node heap: 26/26 tasks passed, including 1,141 core suites and 8,826 core tests
- `yarn build:app`
- focused shared email, Communications Hub, notifications, Resend, SES, enterprise-security, CLI, and core suites and typechecks
- live `TC-START-001` at desktop and mobile widths, `TC-SALES-EMAIL-001` with CORS validation enabled, enterprise-enabled `TC-SEC-EMAIL-001`, focused CRM/notification/sales-lock paths, and WMS shortfall delivery
[...]

PR #5426, opened 2026-08-20
feat(shared/email): SMTP transport behind the sendEmail facade

bartek5412:
## Summary

Implements the SMTP transport specified in `.ai/specs/2026-08-14-smtp-email-transport.md` (spec merged in
open-mercato/open-mercato#5303).

`sendEmail()` in `packages/shared/src/lib/email/send.ts` hardcoded the Resend SDK — `new Resend(apiKey)`
inline, keyed off `RESEND_API_KEY`, throwing `RESEND_API_KEY is not set` when the key is absent. Deployments
that cannot use Resend (self-hosted installs, EU data-residency SMTP relays, local development against
MailDev/Mailpit, CI mail sinks) therefore had no supported way to deliver system email, and a downstream app
could not shim it because the 26 call sites live inside published `@open-mercato/*` packages.

This PR keeps `sendEmail()` as the single façade with its **signature and `SendEmailOptions` unchanged**, and
moves delivery behind a two-member transport interface. Transport resolution:

```
EMAIL_STRATEGY env ('resend' | 'smtp'; unknown value warns once and falls through)
  → auto-detect: RESEND_API_KEY set → resend; else SMTP_HOST set → smtp
    → neither configured → throw 'RESEND_API_KEY is not set'   (existing behavior, unchanged)
```

Backward-compatibility invariant: a deployment with only `RESEND_API_KEY` behaves identically to today, and a
deployment with nothing configured fails with the same error, from the same step, as today — transport
resolution deliberately precedes from-address resolution to preserve that error precedence.

**TLS is required by default.** `SMTP_SECURE=false` means STARTTLS-required (`requireTLS: true`,
`rejectUnauthorized: true`), never opportunistic; cleartext needs an explicit `OM_ALLOW_INSECURE_SMTP=true`
operator opt-in, mirroring the policy `channel-imap` already enforces. Both provider SDKs are imported lazily
inside their transports, so a Resend-only deployment never loads nodemailer and an SMTP-only deployment never
loads the Resend SDK.

## Changes

- `packages/shared/src/lib/email/transports/types.ts` — `EMAIL_STRATEGIES`, `EmailStrategyName`,
  `ResolvedEmailMessage`, `EmailTransport`. Internal only; no exported contract changes.
- `packages/shared/src/lib/email/transports/resend.ts` — the existing Resend code path, moved verbatim except
  for a lazy `await import('resend')`.
- `packages/shared/src/lib/email/transports/smtp.ts` — new transport: lazy nodemailer, `@react-email/render`
  to html + plaintext, base64 attachment mapping, TLS policy, per-call transporter closed in `finally`,
  failures wrapped as `SMTP_SEND_FAILED`.
- `packages/shared/src/lib/email/config.ts` — `resolveEmailTransportName()` (warn-once on an unknown
  `EMAIL_STRATEGY`) and zod-validated `resolveSmtpConfig()` carrying the TLS-policy resolution.
- `packages/shared/src/lib/email/send.ts` — reduced to the four normative steps: disable-check + test capture,
  transport resolution, from-address resolution, dispatch.
- `packages/shared/package.json` — adds `nodemailer`, `@types/nodemailer`, `@react-email/render`, and declares
  `resend`, which was previously imported through workspace hoisting alone.
- `apps/mercato/.env.example` + `packages/create-app/template/.env.example` — identical `EMAIL_STRATEGY` /
  `SMTP_*` / `OM_ALLOW_INSECURE_SMTP` blocks (Template Sync Checklist; `yarn template:sync` passes).
- `packages/shared/AGENTS.md` — an `email/` row in the Library Directory table.

New error-string prefixes: `SMTP_NOT_CONFIGURED: set SMTP_HOST` and `SMTP_SEND_FAILED: <msg>`.
`RESEND_SEND_FAILED`, `RESEND_API_KEY is not set` and `EMAIL_FROM_NOT_CONFIGURED` are unchanged.

### Two deliberate deviations from the spec as written

Both came out of review on the fork (fullstackhouse/open-mercato#92) and make the spec's stated intent
actually hold. Flagging them explicitly so the spec's changelog can be updated in a follow-up:

1. **Cleartext mode sets `ignoreTLS: true`.** The spec's TLS table specifies `requireTLS: false` +
   `tls: undefined` for the `OM_ALLOW_INSECURE_SMTP=true` row. That alone leaves nodemailer's *opportunistic*
   STARTTLS in place, which then verifies certificates — so a self-signed MailDev/Mailpit sink could fail
   despite the operator explicitly opting into cleartext, and the outcome would depend on what the sink
   advertises. `ignoreTLS` makes the opt-in deterministic. The two secure rows of the table are unchanged.
2. **`SMTP_TIMEOUT_MS` has a 30s default.** The spec lists it as an optional int with no default. Left
   optional, an unset or unparseable value inherits nodemailer's defaults (2 min connect, 10 min socket),
   letting a hung SMTP server stall the calling request for minutes. It now falls back to 30s and warns once
   when a value was supplied but unusable.

## Specification

- [x] Yes
- [ ] No (created a new spec)

**Spec file path:**
`.ai/specs/2026-08-14-smtp-email-transport.md`

## Testing

Run locally against this branch (Node/local runner, no Docker `app` container):

- `yarn workspace @open-mercato/shared test` — 177 suites, 1931 tests, all passing
- `yarn workspace @open-mercato/shared build` — 263 entry points, built successfully
- `yarn typecheck` — 25/25 packages clean
- `yarn lint` — 0 errors
- `yarn template:sync` — app/template parity confirmed
- `yarn install --immutable` — lockfile validates unchanged

New `packages/shared/src/lib/email/__tests__/smtp.test.ts` (23 cases) covers: the resolution order including
`EMAIL_STRATEGY` forcing in both directions and the nothing-configured ordering pin; transporter options from
env with `auth` omitted on incomplete credentials; `SMTP_SECURE` defaulting by port; the full TLS policy table
including the cleartext opt-in and its warn-once; the timeout default and the unusable-value fallback; rendered
html + text with mapped `replyTo`, from-chain result and base64 attachments; `SMTP_SEND_FAILED` wrapping for
both `sendMail` and render failures, with `close()` asserted on success and failure; and the
disabled-delivery short-circuit creating no transporter.

`packages/shared/src/lib/email/__tests__/send.test.ts` keeps its assertions byte-identical as the
backward-compatibility guard. Its **mock plumbing** did have to change: the now-lazy `await import('resend')`
defers the `jest.mock` factory past `beforeEach`, so the factory-assigned `var`s were `undefined` there and the
suite failed 9/9 before the rework. The spec's "passes unchanged" wording is therefore unachievable alongside
its own lazy-import requirement — worth correcting in the spec changelog.

The `ignoreTLS` fix, the timeout default, uniform error wrapping in `try`/`finally`, and two added test cases
all came out of review on the fork.

## Checklist

      *(left for the human author to tick personally — this PR was prepared with agent assistance)*
      *(both `.env.example` files and the `packages/shared/AGENTS.md` Library Directory row; no locale surface —
      the new error strings are operator-facing library errors, consistent with the file's existing
      `RESEND_SEND_FAILED` / `EMAIL_FROM_NOT_CONFIGURED` literals)*
      *(not required and documented in the spec: the change adds no HTTP route and no UI surface, so there is no
      integration-test surface — coverage is unit-level)*
      *(the spec is already merged upstream via #5303 and is not modified here — this branch is based on a fork
      `develop` that predates it, so editing the file here would land as an add/add conflict. The two deviations
      above plus the `send.test.ts` correction are intended as a follow-up changelog entry.)*
- [ ] QA routing set: `skip-qa` for low-risk non-customer-facing changes — and for changes with no manually
      exercisable UI surface that leave the database structure and API surface unchanged, preserve
      `BACKWARD_COMPATIBILITY.md` contracts, and ship automated tests for the behavior they change — otherwise
      `needs-qa`.
      *(the three label boxes above are unticked because this account has `read` permission upstream and cannot
      apply labels. Intended: `feature`, `priority-medium`, `risk-medium`, `skip-qa` — the
      automated-verification exemption applies: no `.tsx` outside tests, nothing under `packages/ui/src/` or
      `**/components/**`, database structure and API surface unchanged, no `BACKWARD_COMPATIBILITY.md` contract
      touched (no exported type changed), and unit tests for the changed behavior ship here.)*

### Design System Compliance

N/A — no UI surface. The change touches only `packages/shared/src/lib/email/`, two `.env.example` files and one
AGENTS.md table row; there is no `.tsx` outside tests and no component, style or token in the diff.

## Linked issues

None.


pkarw:
I think this is a duplicate of #4471 which is a preferred solution to this problem

PR #5432, opened 2026-08-20
fix(ui): align WMS KPI cards, role dialog footer and EUDR area field

patzick:
## Summary

Three demo-blocking UI inconsistencies reported during pre-release review: WMS KPI cards whose links/badges floated at different heights depending on caption length, a cramped role-assignment modal footer, and the EUDR plot "Area" field missing the required asterisk. Each is fixed at its source without changing any data contract.

## Changes

- **WMS KPI cards** (`WmsOperationalDashboardPage`, `WmsLocationDetailPage`, `WmsLotDetailPage`, `WmsSkuDetailPage`): title + caption sit in a top block and the value/badge/sparkline/progress/CTA move into one bottom block pinned with `mt-auto`, so those rows align across a grid row no matter how many lines the copy takes.
- **`AssignRoleDialog` footer**: now uses `DialogFooter`'s `leading` slot with `sm:max-w-2xs` so the hint wraps to two lines instead of crowding the buttons; `bordered={false}` drops the default `-mx-6` treatment that pushed the button row flush against the edges of this `p-0` dialog, so it lines up with the header again.
- **EUDR plot create/edit forms**: `areaHa` is marked `required` (red asterisk + CrudForm required validation) whenever the geometry is not a polygon, mirroring the existing point-area rule — polygon plots, whose area is computed and the input disabled, stay unaffected.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**
N/A — presentation-only fixes, no behavior or contract change.

## Testing

- `yarn generate`
- `yarn workspace @open-mercato/core typecheck`, `yarn workspace @open-mercato/ui typecheck`
- `yarn workspace @open-mercato/core test -- modules/wms` (38 suites / 236 tests)
- `yarn workspace @open-mercato/core test -- modules/eudr` (18 suites / 118 tests)
- `yarn workspace @open-mercato/core test -- customers/components/detail/__tests__/AssignRoleDialog.test.tsx`
- `yarn workspace @open-mercato/ui test -- primitives/__tests__/dialog.test.tsx`

## Checklist

- [x] This pull request targets `main` (the repository's default target branch; the template text predates that rename).
- [x] I updated documentation, locales, or generators if the change requires it. (No new strings or generated files.)
- [ ] I added or adjusted tests that cover the change. — layout/label polish; existing suites re-run as regression cover.
- [ ] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required). — no API/DB surface changed; visual alignment and an asterisk are verified by manual QA.
- [x] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — N/A.
- [x] QA routing set: `needs-qa` — the change is customer-facing UI without automated coverage for the changed rendering.

### Design System Compliance
- [x] No hardcoded status colors — existing semantic tokens/`StatusBadge` variants untouched
- [x] No arbitrary text sizes — `sm:max-w-2xs` is the Tailwind v4 container scale, no `[Npx]` values
- [x] Empty state handled for list/data pages — unchanged
- [x] Loading state handled for async pages — unchanged
- [x] `aria-label` on all icon-only buttons — no icon-only buttons added
- [x] Uses existing DS components (`DialogFooter` `leading` slot, `StatusBadge`, `LinkButton`, `CrudForm` `required`) — no custom replacements

## Linked issues

None.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5432"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789853822&installation_model_id=432243&pr_number=5432&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5432&signature=73e76cfd4204a38d0df9deae5a78074bac999b34b0e33ff1bee84fa8f6c3c8ae"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/customers/components/detail/AssignRoleDialog.tsx
+++ b/packages/core/src/modules/customers/components/detail/AssignRoleDialog.tsx
@@ -640,13 +640,18 @@ export function AssignRoleDialog({
           </div>
         </div>
 
-        <DialogFooter className="shrink-0 border-t border-border/70 px-6 py-4 sm:justify-between">
-          <p className="text-xs text-muted-foreground">
-            {t(
-              'customers.roles.dialog.footerNote',
-              'One person per role · can be changed at any time',
-            )}
-          </p>
+        <DialogFooter
+          bordered={false}
+          className="shrink-0 border-t border-border/70 px-6 py-4"
+          leading={
+            <p className="text-xs text-muted-foreground sm:max-w-2xs">
+              {t(
+                'customers.roles.dialog.footerNote',
+                'One person per role · can be changed at any time',
+              )}
+            </p>
+          }
+        >
           <div className="flex items-center gap-2">
             <Button type="button" variant="outline" onClick={onClose}>
               {t('customers.roles.cancelAdd', 'Cancel')}
--- a/packages/core/src/modules/eudr/backend/eudr/plots/[id]/page.tsx
+++ b/packages/core/src/modules/eudr/backend/eudr/plots/[id]/page.tsx
@@ -93,6 +93,8 @@ export default function EditEudrPlotPage({ params }: { params?: { id?: string } 
   const [loading, setLoading] = React.useState(true)
   const [error, setError] = React.useState<string | null>(null)
   const [notFound, setNotFound] = React.useState(false)
+  const [geometryDraft, setGeometryDraft] = React.useState<string | null>(null)
+  const areaRequired = !isPolygonGeometry(geometryDraft ?? formatGeometry(record?.geometry))
 
   React.useEffect(() => {
     let cancelled = false
@@ -197,7 +199,10 @@ export default function EditEudrPlotPage({ params }: { params?: { id?: string } 
         <GeometryInput
           id={id}
           value={typeof value === 'string' ? value : ''}
-          onChange={(nextValue) => setValue(nextValue)}
+          onChange={(nextValue) => {
+            setValue(nextValue)
+            setGeometryDraft(nextValue)
+          }}
           disabled={disabled}
           areaHa={typeof values?.areaHa === 'string' || typeof values?.areaHa === 'number' ? String(values.areaHa) : ''}
         />
@@ -208,6 +213,7 @@ export default function EditEudrPlotPage({ params }: { params?: { id?: string } 
       layout: 'half',
       label: translate('eudr.plots.form.areaHa'),
       type: 'custom',
+      required: areaRequired,
       description: translate('eudr.plots.form.areaHaHelp'),
       component: ({ id, value, setValue, values, disabled }) => {
         const polygonGeometry = isPolygonGeometry(values?.geometry)
@@ -241,7 +247,7 @@ export default function EditEudrPlotPage({ params }: { params?: { id?: string } 
         />
       ),
     },
-  ], [translate])
+  ], [translate, areaRequired])
 
   const groups = React.useMemo<CrudFormGroup[]>(() => [
     {
--- a/packages/core/src/modules/eudr/backend/eudr/plots/create/page.tsx
+++ b/packages/core/src/modules/eudr/backend/eudr/plots/create/page.tsx
@@ -50,6 +50,8 @@ function isCompanySnapshot(value: unknown): value is CompanySnapshot {
 export default function CreateEudrPlotPage() {
   const translate = useT()
   const router = useRouter()
+  const [geometryDraft, setGeometryDraft] = React.useState('')
+  const areaRequired = !isPolygonGeometry(geometryDraft)
 
   const fields = React.useMemo<CrudField[]>(() => [
     {
@@ -113,7 +115,10 @@ export default function CreateEudrPlotPage() {
         <GeometryInput
           id={id}
           value={typeof value === 'string' ? value : ''}
-          onChange={(nextValue) => setValue(nextValue)}
+          onChange={(nextValue) => {
+            setValue(nextValue)
+            setGeometryDraft(nextValue)
+          }}
           disabled={disabled}
           areaHa={typeof values?.areaHa === 'string' || typeof values?.areaHa === 'number' ? String(values.areaHa) : ''}
         />
@@ -124,6 +129,7 @@ export default function CreateEudrPlotPage() {
       layout: 'half',
       label: translate('eudr.plots.form.areaHa'),
       type: 'custom',
+      required: areaRequired,
       description: translate('eudr.plots.form.areaHaHelp'),
       component: ({ id, value, setValue, values, disabled }) => {
         const polygonGeometry = isPolygonGeometry(values?.geometry)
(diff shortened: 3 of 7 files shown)


patzick:
🤖 **Claimed by `om-auto-review-pr`** — running an automated code review of `b53e67e`.

Validation source: **GitHub checks** (no local runs). Currently watching `test`, `docker-build`, `ephemeral-integration` and `Analyze (javascript-typescript)`; `prepare`, `lint`, `ds-lint`, `scope`, `audit-scope` and `license/cla` are already green.

Will post the review verdict once the gate settles.

patzick (commented):

**Verdict: APPROVE** (only Low/nit findings — this repo requests changes on Medium+)
**Head reviewed:** `b53e67e`
**Validation source:** GitHub checks (no local runs)

---

### Validation

All required checks are green on `b53e67e`:

| Gate | CI job | Result |
|---|---|---|
| `yarn build:packages`, `yarn build:app` | `prepare` | ✅ 3m48s |
| `yarn typecheck`, `yarn test`, `yarn test:scripts`, `yarn i18n:check-usage` | `test` | ✅ |
| `yarn lint` | `lint` | ✅ 32s |
| DS token lint | `ds-lint` | ✅ 53s |
| Playwright integration | `ephemeral-integration (none)` | ✅ |
| Container build | `docker-build` | ✅ 6m8s |
| CodeQL / scope / audit-scope / CLA | — | ✅ |

`mergeStateStatus` is `BLOCKED` solely on the QA gate (`needs-qa` without `qa-approved`), which is correct for this change.

---

### What I verified

**`AssignRoleDialog` footer** — correct root-cause fix. `DialogFooter` genuinely exposes `bordered` and `leading` (`packages/ui/src/primitives/dialog.tsx:258,266`), and the default `bordered` branch applies `-mx-6 … px-6`, which is what broke alignment inside this `p-0` `DialogContent` (`AssignRoleDialog.tsx:350`). Dropping it and keeping the explicit `border-t border-border/70 px-6 py-4` now matches the sibling header/section treatment on lines 353 and 364. The `leading` branch wraps the slot in `inline-flex … sm:mr-auto`, so `sm:max-w-2xs` on the `<p>` wraps as intended and replaces the removed `sm:justify-between`.

**EUDR `areaHa` required rule** — correct and consistent with the existing contract:
- `CrudForm` does enforce `required` for `type: 'custom'` fields, because enforcement is skipped only when a zod `schema` is passed (`CrudForm.tsx:2650-2675`) and neither plot page passes one. So this is a real validation change, not just an asterisk.
- `EudrPlotGeometryType` is `'point' | 'polygon'` and `validatePlotGeometry` maps every non-`Point` geometry to `'polygon'` (`lib/geometry.ts:1,170`). So `!isPolygonGeometry(…)` ≡ "point", exactly mirroring the pre-existing `assertPointAreaWithinLimit` rule — no new restriction, and MultiPolygon still resolves to `polygon` (computed area, input disabled).
- The server keeps `areaHa: z.coerce.number().positive().max(100_000).optional().nullable()`, so the client rule stays a strict subset of what the API accepts.
- `required` and the component's internal `disabled` are computed from the same predicate, so there is no "required but un-fillable" deadlock.
- `formatGeometry` is a hoisted function declaration (`[id]/page.tsx:74`), so the render-time call at line 97 has no TDZ problem, and the edit page renders `LoadingMessage` until `record` resolves, so no asterisk flicker during load.

**WMS KPI cards** — `grid` defaults to `align-items: stretch`, so with `flex flex-col` + a `mt-auto` bottom block the CTA rows now align across a grid row regardless of caption wrapping. That is the reported bug, and it is fixed on all four pages.

---

### Findings

#### 🟢 Low — `WmsLocationDetailPage.tsx:1043`: value/badge row still misaligns on the "Capacity used" card

`LocationKpiCard` is the only one of the four card components with an optional `progress` slot, and "Capacity used" is the only card that passes it. It is also the only card whose CTA is conditional — `onCtaClick={access.canManageLocations ? … : undefined}` with **no `ctaHref` fallback**, unlike the other four cards which all pass a static `ctaHref`.

Because `mt-auto` bottom-pins the entire block rather than aligning the value row itself, that card's bottom block has a different height from its siblings:

- **With** `canManageLocations`: block is value + progress + CTA, so its number sits ~18px **above** its neighbours'.
- **Without** `canManageLocations`: block is value + progress only, so the number drops ~30px **below** its neighbours' — and pre-PR that value was top-aligned right under the caption, so this state is arguably a small step back for read-only users.

The PR's headline goal (CTA/link alignment, independent of caption length) is achieved, so this is a nit rather than a blocker. If you want the numbers to line up too, the fix is to reserve the optional slots — render the `Progress` wrapper unconditionally with `invisible` when `progress == null`, and give the capacity card a read-only `ctaHref` fallback (e.g. `/backend/config/wms`, the pattern `LotKpiCard`'s quality-status card already uses at `WmsLotDetailPage.tsx:1142`).

---

### Notes (not findings)

- The PR body describes the change as "presentation-only, no behavior change", but the `areaHa` `required` flag *is* a client-side validation change. It is a safe one (it mirrors a rule `onSubmit` already enforced via `assertPointAreaWithinLimit`, just moved earlier and made visible), so the practical effect is only that the error surfaces on the field instead of as a form-level throw. Worth an accurate line in the description.
- `geometryDraft` mirrors `values.geometry` into page state, giving the field descriptor and the field component two sources of truth for the same predicate. Today this cannot desync: the only writer is `GeometryInput.onChange`, and the `CrudForm` paths that can rewrite values behind a field's back (`transformDisplayData` / `onFieldChange` / `transformFormData`, `CrudForm.tsx:2368,2521,2778`) all require an injection spot, which these pages do not opt into (no `entityId` / `injectionSpotId`). Flagging only because if a plot form ever gains an injection spot, a desync would make `areaHa` required while rendered disabled. A values-aware `required` on `CrudField` would remove the mirror entirely, but that is a framework change well outside this PR.
- Labels look right: `bug` + `review` + `needs-qa` + `priority-medium` + `risk-low`. `needs-qa` is correct — this touches `.tsx` rendering with no automated coverage for the changed output, so the automated-verification exemption does not apply.

---

⚠️ **This review is posted as a comment, not an approval state.** The reviewing account (`patzick`) is also this PR's author, and GitHub does not permit self-approval. A second maintainer needs to convert this into a formal approval. The `needs-qa` → `qa-approved` transition is likewise a maintainer label write that this skill does not perform.

PR #4358, opened 2026-07-22
feat(eudr): EUDR compliance module — evidence, plots, risk, DDS lifecycle and reporting

haxiorz:
## Summary

EU importers of coffee, cocoa, cattle, palm oil, rubber, soy and wood must prove, per shipment, that commodities are deforestation-free and legally produced — backed by plot-level geolocation and supplier documentation, referenced in a Due Diligence Statement filed in the EU Information System. Today that evidence lives in spreadsheets and questionnaires, disconnected from the products, suppliers and orders it covers. Open Mercato already holds those records but had no EUDR model at all.

This PR adds a new core module `eudr`: a backend compliance workspace that maps catalog products to regulated commodities, registers supplier plots with validated geometry, collects origin evidence with server-computed completeness, runs Art. 10 risk assessments with mitigation tracking, and drives DDS records through a guarded lifecycle — including the 72-hour amend/withdraw window — with an exportable evidence packet and an Art. 12(3) annual report.

Scope is operator-facing. The supplier self-service portal and live EU IS (TRACES) filing are deliberately out of scope and documented as roadmap in the spec.

## Changes

- **6 entities** (`eudr_` prefix): product mappings, evidence submissions, DDS statements, plots, risk assessments, mitigation actions — all tenant/org scoped, soft-deleted, with `updated_at` for optimistic locking.
- **13 API routes** via `makeCrudRoute` plus hand-written scoped routes (statement export incl. `?format=geojson`, plot GeoJSON import, mapping suggestions + apply, annual report JSON/CSV, supplier aggregate, dashboard widget). OpenAPI exported on every route.
- **Statement lifecycle**: transition map enforced in create *and* update, a submission gate (evidence readiness + risk cleared, or Art. 13 simplified DD, or SME-trader referencing) with machine-readable failure reasons, gate-time assessment freshness re-check, and a set-once-immutable `reference_issued_at` driving the 72h amend/withdraw guard.
- **Reference data as versioned code** (`lib/reference-data.ts`): country benchmarking tiers, Annex-I HS prefixes, supplementary-unit list, application dates, the 15-criterion Art. 10 catalog.
- **Pure libs** for geometry (GeoJSON validation, WGS84 bounds, ring closure, geodesic area, the 4-ha point/polygon rule — no new dependency) and completeness scoring.
- **Backend UI**: compliance cockpit with feature-gated queues and the annual-report card, five list pages, ten create/edit pages, Leaflet plot preview (read-only, dynamic import).
[...]

Issue #1836, opened 2026-05-06
bug:  [UI / Filters] Filter dropdowns on list pages are invisible — popovers render but are not visible (z-index regression across all entities)


alinadivante:
## Summary
Across various list pages (e.g., Team Members, Products, Companies, People), opening the Filters side panel and clicking any filter dropdown field (e.g., "Role", "Category", "Status") shows no visual dropdown — the popover does not appear. However, clicking blindly below the field selects an option, confirming the popover content exists in the DOM but is rendered behind the panel or other elements (z-index stacking issue).

This issue is a global regression affecting all list pages and tables that utilize the Popover primitive within filter panels.

## Steps to reproduce
1. Navigate to any list page with a filter panel (e.g., /backend/staff/team-members, /backend/customers/companies, or /backend/catalog/products), example  `/backend/staff/team-members`
2. Click the **Filters** button in the top-right area of the table — the filter side panel opens.
3. Click the **Role** filter field (dropdown/select).
4. Observe whether a dropdown list of roles appears.
5. Without seeing any dropdown, click slightly below the Role field and observe if a role gets selected.

## Expected result
Clicking a filter field opens a visible dropdown list of available options. The dropdown appears on top of all other page elements (including the filter side panel) and is clearly readable.

## Actual result
The dropdown does not visually appear after clicking the filter field. No list of options is visible. However, clicking below the field in the approximate area where the dropdown would be selects an option — confirming the dropdown content is rendered in the DOM but hidden behind another UI layer (the filter panel itself).

## Environment
- Branch: `develop` (also reproducible on `pr-1800` / `pr-1791`)
- Module or area: Staff / Team Members / Filter panel / `Popover` primitive / z-index
- Browser/Runtime: Edge / Chrome, Windows

## Additional context
Likely root cause: PR #1791 (`feat: CRM activity new UI`) changed the z-index of the global `Popover` primitive from `z-dropdown` to `z-tooltip` in `packages/ui/src/primitives/popover.tsx`:

```diff
- 'z-dropdown min-w-[280px] rounded-md border bg-popover...'
+ 'z-tooltip min-w-[280px] rounded-md border bg-popover...'
```

This change was made to ensure the new `ActivitiesAddNewMenu` popover renders above other CRM page elements. However, since `Popover` is a **shared primitive used across the entire application**, this z-index bump affects every popover in the system — including filter dropdowns on list pages.

[...]

PR #5443, opened 2026-08-21
fix(ui): portal the combobox dropdown so a dialog's overflow stops clipping it

Frshy:
## Summary

`ComboboxInput` rendered its suggestion list as an in-flow `absolute` panel inside its own `relative` wrapper. Every DS dialog is a scroll container — `dialogContentVariants` carries `max-h-[90vh] … overflow-y-auto` (`packages/ui/src/primitives/dialog.tsx:74`), and the WMS dialogs add `overflow-hidden` on top — so the list was **clipped by the dialog** instead of floating over it. `z-popover` (45) cannot help here: z-order does not escape an overflow clip. On a short dialog the user sees a sliver of the list, or nothing at all.

Reproduces in every in-dialog usage: `AdjustInventoryDialog`, `MoveInventoryDialog`, `ReceiveInventoryDialog`, `ReserveInventoryDialog`, `CycleCountWizardDialog`, `QuickDealDialog`, and anything `CrudForm` hosts inside a dialog.

The fix renders the list through the DS `Popover` (`PopoverAnchor` + `PopoverContent`) — the same mechanism `Select` and `DatePicker` already use successfully inside these very dialogs (`ReceiveInventoryDialog.tsx:621`). One production file changes; `dialog.tsx` is untouched.

That reuses what the DS already provides: portaling to `<body>` so no ancestor `overflow` can clip the list; collision detection, so the list **flips above the input** when there is no room below instead of rendering off-screen; re-anchoring on ancestor scroll and resize; correct `pointer-events` under a modal dialog's `body { pointer-events: none }`; and participation in Radix's dismissable-layer stack, so `Escape` closes the list and not the dialog. It also avoids an eighth hand-rolled copy of the `getBoundingClientRect` + `scroll`/`resize` measurement loop that already exists in `ProfileDropdown`, `RowActions`, `ActionsDropdown` and `TopbarSearchInline`.

An earlier cut of this fix added a `data-combobox-portal` marker plus outside-interaction guards in `DialogContent`, on the theory that a portaled list reads as an outside interaction and dismisses the dialog. It does not: React attaches its listeners to portal containers as well, so a portal rendered from inside the dialog's React tree still dispatches through `DialogContent`'s `onPointerDownCapture` and Radix's `isPointerInsideReactTreeRef` suppresses the dismissal. A test asserts this, with a control case proving a genuinely outside pointerdown still closes the dialog. That approach was dropped — no primitive needs to know about comboboxes.

## Changes

- `packages/ui/src/backend/inputs/ComboboxInput.tsx` — render the suggestion list via `Popover` / `PopoverAnchor` / `PopoverContent` instead of an in-flow `absolute` panel. Net +32/−7 lines ignoring re-indentation; all state, effects, callbacks and the list body are unchanged. Specifics worth reviewing:
  - `open` stays **fully controlled with no `onOpenChange`**, so the existing blur timer, `Escape` handler and selection remain the single source of truth for dismissal — behaviour preserved, not re-implemented.
  - `onOpenAutoFocus` / `onCloseAutoFocus` are prevented: focus must never leave the `<input>`, since the blur-close timing and the `aria-activedescendant` model depend on it.
  - `PopoverAnchor asChild` merges into the existing wrapper, adding no DOM node — consumer hacks like `className="relative [&_input]:pl-9"` (`AdjustInventoryDialog.tsx:534`) and the clear button's absolute positioning keep working.
  - The content stops `wheel` **and** `touchmove` propagation. A modal dialog locks scrolling via `react-remove-scroll`, which cancels both at document level outside its own content node (`shards: [contentRef]`); without this a long list inside a dialog cannot be scrolled by wheel or by touch — a real regression the portal would otherwise introduce. `overscroll-contain` stops the scroll chaining to the page at the ends.
  - `aria-owns` is set alongside `aria-controls`: the portaled listbox is outside the input's subtree, so `aria-activedescendant` is only valid while `aria-owns` makes it a logical descendant.
  - `role="presentation"` overrides the `role="dialog"` Radix hardcodes on popover content. The popup is a positioning shell around the listbox; a second dialog node would misdescribe it and break `getByRole('dialog')`, which is strict about multiple matches in both Playwright and RTL and is used by ~35 test files.
  - The content is unmounted outright rather than relying on Radix's exit animation — a closing-but-still-mounted layer keeps swallowing `Escape` (the failure mode already documented in `AdvancedFilterPanel`).
- `packages/ui/src/backend/inputs/__tests__/ComboboxInput.test.tsx` — +6 cases for the portaled popup.
- `packages/ui/src/backend/inputs/__tests__/ComboboxInput.dialog.test.tsx` — new; in-dialog regression coverage.
- `packages/ui/src/primitives/__tests__/zindex-overlay.test.tsx` — the `z-popover` guard (issue #1836) now resolves the list on the document and additionally asserts it is *not* inside the component's container.
- `packages/core/src/modules/push_notifications/__integration__/TC-PUSH-009.spec.ts` — scoped its option lookup to the CrudForm field wrapper; now queries the page, since the list is no longer a DOM descendant of the field.
- `packages/core/src/modules/wms/__integration__/helpers/wmsUi.ts` — `fillCombobox` and `selectLocationComboboxOption` located suggestions as `div.absolute` inside the field's `relative` wrapper and clicked them by `role="button"`. Both halves were wrong: the list is portaled out of that wrapper, and the items have carried `role="option"` since #4358, so the lookup could not match even before this branch. Both now query `page.getByRole('option', …)`, matching TC-PUSH-009 and `selectDialogOption` in the same file.
- `.ai/ui-components.md` — documents the new popup placement and how tests should query it; also corrects a stale "Popup visual" line that described a `rounded-2xl` container the code never rendered.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**
No spec covers combobox or dialog layering. Contained bug fix with no contract change: no props added or removed, no DB or API surface, no `BACKWARD_COMPATIBILITY.md` contract affected. Two adjacent specs were checked and neither needs updating — `.ai/specs/2026-05-09-crudform-autofocus-flicker.md` (the `suppressOpenOnFocusRef` behaviour, still covered by its tests) and `.ai/specs/2026-07-05-ds-theming-and-brand-customization.md` (the frozen `--z-index-*` scale, untouched — the list still carries `z-popover`).

## Testing

- `yarn workspace @open-mercato/ui test` — 232 suites / 1898 tests green.
- `yarn lint` (forced, no turbo cache) and `yarn typecheck` (26 packages) green.
- `yarn workspace @open-mercato/core test src/modules/wms/components/backend/__tests__` — 11 suites / 78 tests green.

Unit coverage added:

- `ComboboxInput.test.tsx` — the list renders outside the field wrapper; focus stays on the input while it is open; click and keyboard selection work portaled; `Escape` does not reach an enclosing surface; the popup leaves the document on close. The 21 pre-existing cases pass unchanged.
- `ComboboxInput.dialog.test.tsx` — the list is not a descendant of `[data-dialog-content]`; picking an option keeps the dialog open (with a control asserting a genuinely outside pointerdown still closes it); `wheel` and `touchmove` on the list do not reach the dialog's scroll lock (each with a control asserting the same event elsewhere still does); exactly one `role="dialog"` node exists while the list is open.

**Manual browser verification** (headless Chromium via Playwright, against a reproduction built from the real `DialogContent` + `ComboboxInput` and the app's compiled `globals.css`):

| Check | Result |
|---|---|
| List clipped by the dialog (before) | entirely invisible — 0 px on screen |
| List over the dialog edge (after) | fully visible, width matched to the input (462 px = 462 px) |
| No room below | flips above the input, `data-side="top"`, fully in viewport |
| Click an option / `ArrowDown` + `Enter` | selects; dialog stays open |
| `Escape` | closes the list; a second `Escape` cancels the dialog |
| Wheel over a long list in a modal dialog | scrolls (frozen without the wheel fix) |
| `touchmove` on the list | never reaches the document; elsewhere it still does |
| Dialog body scrolled | list re-anchors to the input |
| Click outside the dialog | still dismisses it — the fix does not over-block |
| `role="dialog"` nodes with list open | 1 (was 2) |

Before/after screenshots are attached by the branch author in a follow-up comment (GitHub exposes no image-upload API, so automation cannot post them inline).

## Checklist

- [ ] QA routing set. 

### Design System Compliance
- [x] `aria-label` on all icon-only buttons (`<IconButton>`)

## Linked issues

No tracker issue — found while using `ComboboxInput` inside a `Dialog`.


Frshy:
Opened from the Full Stack House fork by an account without `triage` on this repo, so I can't set metadata myself. Everything below is intent for a maintainer to apply — inference per `AGENTS.md` → PR Workflow and `.github/QA-DEPLOYMENT.md`.

**Labels**

| Label | Why |
|---|---|
| `bug` | Category. Fixes broken behaviour, no new surface. |
| `review` | Pipeline. Ready, non-draft, not in another pipeline state. |
| `priority-medium` | Urgency. The dropdown is effectively unusable inside dialogs, but it is long-standing rather than a fresh regression and keyboard selection still gets a value in. Not `priority-high` on that basis — happy to be overruled, since it does affect the WMS receive/adjust/move flows directly. |
| `risk-medium` | Blast radius. One shared DS input, ~20 call sites, no DB or API surface and no `BACKWARD_COMPATIBILITY.md` contract touched; "ordinary single-module change". |
| `needs-qa` | The change is `.tsx` under `packages/ui/src/`, so the automated-verification exemption does not apply, regardless of the unit coverage. |
| `screenshots` | Before/after visual evidence, attached to this PR by the branch author (automation cannot upload images). Informational only. |

Deliberately **not** requesting `in-progress`: this contribution can't release a claim label later, and a claim that can't be released is worse than none.

**Assignee** — the branch author (@Frshy), if maintainers assign external contributions at all here.

**Reviewer** — someone who owns `packages/ui` / the design system, since the change turns on Radix layering semantics inside `DialogContent` rather than on module logic.

**QA gate** — this stays blocked on `needs-qa` until a maintainer adds `qa-approved`. I could not run the real app (no local database available), so the verification in the description is against a reproduction built from the real `DialogContent` + `ComboboxInput` and the app's compiled `globals.css`, not against the WMS pages themselves. A click-through of Receive/Adjust/Move Inventory and the Quick Deal dialog is the gap worth closing; posting evidence would not clear the gate on its own in any case.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-21T12:19:58Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# Code Review: fix(ui): portal the combobox dropdown so a dialog's overflow stops clipping it

**Verdict: CHANGES REQUESTED** — 1 major, 2 minor, 2 nit.

The diagnosis and the approach are right. `dialogContentVariants` really does carry `max-h-[90vh] … overflow-y-auto` (`packages/ui/src/primitives/dialog.tsx:74`), z-index cannot escape an overflow clip, and routing the list through the DS `Popover` is the same mechanism `Select`/`DatePicker` already use in these dialogs. The controlled-`open`-without-`onOpenChange` trick keeps the existing blur/Escape/selection state machine as the single owner of dismissal, and the `role="presentation"` override is correct — Radix spreads `...contentProps` *after* its hardcoded `role="dialog"` (`@radix-ui/react-popover` `PopoverContentImpl`, lines 245/248), so the override lands.

I also independently verified the two riskiest claims in the description:

- **Scroll-lock mitigation is sound in production, not just jsdom.** `react-remove-scroll` registers `wheel`/`touchmove` on `document` with `{ passive: false }` and **no** `capture` (`react-remove-scroll/dist/es2015/SideEffect.js:139-141`). React attaches its delegated listeners to the portal container (`document.body`) via `preparePortalMount`, so the `onWheel`/`onTouchMove` `stopPropagation` runs strictly before the document-level lock. The mitigation is correct.
- **Outside-click dismissal is not over-blocked.** I ran a throwaway jsdom case (dialog open, suggestions open, `pointerDown` on a node outside the dialog) against this branch: the dialog still dismisses. The behaviour is fine — but see **m3**, because that control case is *described* in the PR body and is *not* in the shipped test file.

---

## Findings

### 🔴 Major

**M1 — The WMS Playwright helper still encodes the pre-portal DOM; `TC-WMS-INVENTORY-UI-001` degrades and can no longer reach its last-resort click.**
`packages/core/src/modules/wms/__integration__/helpers/wmsUi.ts:126-128` and `:196-198` both locate the suggestion list as:

```ts
input.locator('xpath=ancestor::div[contains(@class,"relative")][1]').locator('div.absolute')
```

After this change no such node exists — the only remaining `absolute` element inside the `relative` wrapper is the clear affordance, and it is a `<button>` (`IconButton`), not a `div`. Consequences:

1. `fillCombobox` (`:125-136`) — `suggestionInDropdown.isVisible({ timeout: 2_000 })` now always times out, then falls through to the keyboard path. `TC-WMS-INVENTORY-UI-001.spec.ts` calls `fillCombobox` **seven times** with `scope: <dialog locator>` (lines 113, 117, 121, 256, 264, 290, 294), so that is ~14 s of dead wall-clock added to one spec.
2. Worse, the last-resort branch is `root.getByRole('button', { name: suggestionPattern })` where `root = options.scope ?? page` — i.e. **the dialog locator**. The portaled suggestion is no longer a descendant of the dialog, so that branch can no longer find it and will fail on its 10 s visibility expectation instead of clicking. That branch is not decorative: the spec's own comment at `TC-WMS-INVENTORY-UI-001.spec.ts:259-262` documents that the keyboard fallback is already known to leave the zone field uncommitted, which is exactly when the dialog-scoped click is the safety net.
3. `selectLocationComboboxOption` (`:196-203`) permanently takes its `else` branch, `page.getByRole('button', { name: locationCode, exact: true })` — page-wide and therefore strict-mode-ambiguous if that code renders in any other button.

This is the same class of breakage the PR already correctly fixed in `TC-PUSH-009.spec.ts` (scope the option lookup to `page`, not the field wrapper) — and the WMS receive/adjust/move/cycle-count dialogs are the *headline* repro in the PR description, so this helper is the one that most needed the same treatment. Fix: query the suggestion from `page` (ideally by `getByRole('option', …)`, matching the updated push spec) and drop the `div.absolute` path in both helpers.

### 🟡 Minor

**m1 — `aria-activedescendant` now points outside the focused element's subtree.**
`ComboboxInput.tsx:432` keeps `aria-activedescendant={optionDomId(selectedIndex)}` on the `<input>`, but the referenced option now lives in a `document.body` portal. ARIA requires the referenced element to be a DOM descendant of the element with focus, or a logical descendant declared via `aria-owns`. `aria-controls` (line 430) has no such constraint and is fine. Assistive tech mostly copes in practice (MUI/Downshift portal too), but the cheap and spec-correct fix is to add `aria-owns={listboxVisible && !loading && filteredSuggestions.length > 0 ? listboxId : undefined}` alongside the existing `aria-controls`. The accessibility test block in `ComboboxInput.test.tsx` is the natural place to pin it.

**m2 — `z-popover` (45) is below `z-modal-elevated` (55), and the list is no longer inside the dialog's stacking context.**
Previously the list rendered *inside* `DialogContent`, so it painted above the dialog regardless of the dialog's own z-index. Portaled to `<body>` at `z-popover`, a `ComboboxInput` inside a `<DialogContent elevated>` (55, see `dialog.tsx:55-57`) would now render behind both that dialog and its overlay. No current call site hits this — `AdvancedFilterPanel.tsx:163` is the only `elevated` dialog and it hosts a plain `Input` — so this is latent, not broken today. But `zindex-overlay.test.tsx` only asserts `z-popover > z-modal`, so nothing guards it. Please either note the constraint in the `.ai/ui-components.md` entry you are already editing, or add the `z-modal-elevated` case to that test so a future `elevated` dialog with a combobox fails loudly.

### ⚪ Nit

**n1 — The PR description claims a test that does not ship.**
> "picking an option keeps the dialog open (with a control asserting a genuinely outside pointerdown still closes it)"

`ComboboxInput.dialog.test.tsx:89-108` has no such control case. The `wheel`/`touchmove` cases *do* carry their controls (`:70-72`) — this one does not. I verified the behaviour holds on this branch, so this is coverage, not a defect: add the assertion (outside `pointerDown` → `[data-dialog-content]` gone) so the property stays pinned, or drop the claim from the description.

**n2 — `align="start"` on `PopoverContent` is redundant.**
`packages/ui/src/primitives/popover.tsx:18` already defaults `align` to `'start'`.

---

## Checked and clear

- **No blockers.** No contract surface touched: no event id, response field, DB column, public export, ACL feature, DI key or CLI command changed. `BACKWARD_COMPATIBILITY.md` is not implicated — `ComboboxInput`'s props and behaviour are unchanged; only the render target moved.
- **Escape ordering is correct.** Radix's `useEscapeKeydown` is capture-phase on `document`, and `DismissableLayer` only acts on the *highest* layer — with the list open the popover layer is highest, its `onDismiss` is a no-op (controlled `open`, no `onOpenChange`), and the enclosing dialog is therefore not dismissed. The input's own `Escape` handler then closes the list, preserving the pre-change behaviour. Conditionally unmounting `PopoverContent` (rather than relying on the exit animation) avoids the stale-layer-swallows-Escape failure mode the comment cites.
- **Pointer-events under a modal dialog.** The popover layer mounts after the dialog's, so `DismissableLayer`'s index comparison gives it `pointer-events: auto` under `body { pointer-events: none }`.
- **Outside-interaction suppression.** The portal is a React-tree child of `DialogContent`, so `onPointerDownCapture` still sets Radix's `isPointerInsideReactTreeRef` and option clicks do not read as outside interactions. Verified empirically (above), including the negative case.
- **`w-[var(--radix-popover-trigger-width)]`** is not a DS arbitrary-value violation — same idiom as `select.tsx:127,136`, and `yarn lint:ds` on the changed file is clean. `min-w-0` / `p-2` correctly override the wrapper's `min-w-[280px]` / `p-0` through `tailwind-merge`.
- **Other consumers.** The `wmsUi.ts` helper is the only structural coupling I found. Every other option/suggestion lookup in the repo is already page-scoped (`TC-TRANS-005/009`, `TC-WF-008`, `TC-CRM-071`, `salesUi.ts`, `wmsUi.ts:223`), and the WMS/translations jsdom dialog suites either mock `ComboboxInput` or query via `screen`, so they are unaffected.
- **Docs.** The `.ai/ui-components.md` edit is accurate and correctly warns to query the popup from the document. The smaller `docs/design-system/components.md` / skill-reference copies do not carry the popup-visual line, so no sync gap.

---

## Validation

**Validation source: local fallback.** GitHub has no usable check data for head `2235079`: both workflow runs (`CI for Develop&Main` `32471056904`, `Mutation tests` `32471057070`) are `completed/action_required` — the fork-PR approval gate means they never executed. The only reported checks are `license/cla` (SUCCESS) and `[code]smith` (SKIPPED). Branch protection on `develop` reports "Required status checks not enabled" (404). Per this repo's `om-auto-review-pr` extension that is exactly the narrow-fallback case, so I ran the affected scope locally instead of the whole gate.

| Command | Result |
|---|---|
| `jest` — `ComboboxInput.test.tsx`, `ComboboxInput.dialog.test.tsx`, `zindex-overlay.test.tsx` | ✅ 3 suites, 38 tests passed |
| `jest` — full `@open-mercato/ui` suite | ✅ **1784 passed, 0 failed** (20 suites could not *start*: `Cannot find module '@tanstack/react-table/legacy'` ×26 and `@open-mercato/ai-assistant/…/agent-transport` ×14 — a stale local install carrying react-table 8.21.3 against the repo's `^9.0.0`, plus unbuilt package `dist/`. Environmental, unrelated to this PR, which touches neither `package.json` nor those files.) |
| `jest` — `@open-mercato/core` WMS dialog + translations suites (9 files: CycleCount ×3, Move/Receive/Reserve/LocationEdit, TranslationManager ×2) | ✅ 46 tests passed (1 suite needs `yarn generate` output — environmental) |
| `tsc --noEmit -p packages/ui` | ✅ **0 errors in any file this PR touches** (30 pre-existing errors, all in `DataTable.tsx` / `customFieldColumns.ts` / `useAutoDiscoveredFields.ts` / `queue` / `core` generated shims, all traceable to the same stale react-table 8 install) |
| `eslint --config eslint.ds.config.mjs` on `ComboboxInput.tsx` | ✅ clean |
| Ad-hoc jsdom control case: outside `pointerDown` dismisses the dialog with suggestions open | ✅ passes (temp file removed) |

`yarn lint` could not run locally (`eslint-config-next` missing from the stale install) and is not part of this repo's configured `validation.commands` gate. The remaining gate steps (`build:packages`, `generate`, `i18n:check-*`, `build:app`) are unaffected by a change that adds no strings, entities, generators, or dependencies — but they are also exactly what a maintainer-approved CI run would prove. **Please approve the workflow runs on this PR so the real gate executes against `2235079`.**

---

## What to do next

1. Fix **M1** — the WMS helper is the one thing that will actually bite in CI.
2. Take **m1** and **m2** (both a few lines).
3. **n1** / **n2** at your discretion.

Everything else in this change is well-argued and well-tested; the reasoning in the description about *why* the earlier `data-combobox-portal` approach was dropped is correct and I confirmed it independently.

pkarw:

Updated after the round-3 re-review of `c1c01865d` (supersedes the rationale posted for `a16639f1e`). One pipeline transition; everything else re-derived and kept.

| Label | State | Why |
|---|---|---|
| `merge-queue` | **added** (replaces `changes-requested`) | Approved: both round-2 blockers verified fixed, no blockers/majors/minors this pass, and CI settled `success` — 25 pass / 4 skipped / 0 fail, all fifteen `ephemeral-integration` shards green including shard 14, which failed round two on the exact test B2 named. |
| `changes-requested` | **removed** | Superseded by the approval. |
| `needs-qa` | kept | The change is `.tsx` under `packages/ui/src/`, so the automated-verification exemption for `skip-qa` does not apply on its first clause. It is also substantively right: the property this PR exists to fix is visual (a list escaping a `Dialog`'s `overflow-y-auto`), and Playwright's `toBeVisible()` passes on a clipped element, so no automated layer asserts it. P0/P1 routes are in the manual-QA comment. |
| `qa-approved` | **not applied** | Reading a diff is not QA. The QA-approval gate holds the merge until a QA reviewer or the self-QA exception adds it — this is what `mergeStateStatus: BLOCKED` will keep reflecting. |
| `bug` | kept | Fixes a real defect: the suggestion list was clipped by every dialog that hosts a `ComboboxInput`. |
| `priority-medium` | kept | Affects every in-dialog combobox (WMS adjust/move/receive/reserve/cycle-count, `QuickDealDialog`, any `CrudForm` in a dialog), but there are keyboard paths to the same values, so it degrades rather than blocks. |
| `risk-medium` | kept | One production file, but it is a shared DS input with ~15 hosts across 8 packages. Held at medium rather than raised: the full 15-shard integration matrix is green, `develop` has not touched `ComboboxInput.tsx` since the merge base, and the only DOM contract that moved is now documented with every in-repo consumer repaired. Happy to be argued to `risk-high` on blast radius alone. |
| `partner-request` | kept | Applied by the author at open; not this skill's to revoke. |

`qa` deliberately untouched — it means "manual QA in progress" and belongs to the QA reviewer.

pkarw:
Thanks @Frshy — the diagnosis and the DS-`Popover` approach are right, and the reasoning about why the earlier `data-combobox-portal` cut was unnecessary checks out; I verified it independently. Review found actionable items, so I'm handing this PR back to you for the next pass. The one that matters is **M1**: `packages/core/src/modules/wms/__integration__/helpers/wmsUi.ts` still locates the suggestion list via `ancestor::div[@class~="relative"] >> div.absolute`, which this change deletes — same fix you already applied to `TC-PUSH-009.spec.ts`, but for the WMS dialogs that are the PR's headline repro.

When the updates are pushed, re-request review and the automation can pick it up from the latest head.

Two notes outside your control:

- CI never ran on this head — `CI for Develop&Main` and `Mutation tests` are both sitting at `action_required` (the fork-PR approval gate), so I fell back to running the affected scope locally. A maintainer needs to approve the workflow runs for the real gate to execute.
- `needs-qa` is on and the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Your read is right that posting evidence would not clear it on its own.

pkarw:

1 major (stale WMS Playwright selector), 2 minor (`aria-activedescendant` across the portal boundary; `z-popover` < `z-modal-elevated`), 2 nit. No blockers — no contract surface touched. Validation source: local fallback (fork-PR CI runs are gated at `action_required`); 1784 `@open-mercato/ui` tests pass and typecheck is clean for every touched file. Autofix was not run: this is a fork PR, and carrying it forward onto a replacement PR is disproportionate to findings this small — the fixes are back with the author.

Frshy:
Thanks — that is a genuinely useful review, and **M1 is a real miss on my side**. All five are addressed in `a16639f1e`.

### M1 — WMS helper

Confirmed, and it is worse than "the portal broke it": the helper was **already degraded before this branch**. Both call sites click by `role="button"`, but the options carry `role="option"` (added in #4358, EUDR), and `wmsUi.ts` has not been touched since. So `getByRole('button', …)` could never match a suggestion — the `div.absolute` container was only the second half of the problem, and my change killed that half too.

That makes your points 1–3 correct with a slightly different mechanism: `fillCombobox` was already burning the 2 s timeout and falling through to the keyboard path on every call, and the last-resort `root.getByRole('button', …)` was already unable to match — my change did not introduce that, it just removed the remaining chance of it working. `selectLocationComboboxOption` likewise.

Both helpers now use `page.getByRole('option', …)`, matching the updated TC-PUSH-009 and `selectDialogOption` at `wmsUi.ts:227` in the same file. In `selectLocationComboboxOption` I dropped the `if visible / else page-wide button` shape entirely rather than repairing it — with a correct locator the fallback only existed to paper over the broken one, and you flagged the `else` branch as strict-mode ambiguous. It is now expect-visible-then-click. `fillCombobox` keeps its keyboard fallback, which is a real safety net for slow async suggestions rather than a workaround for a bad selector.

**Caveat you should weigh:** I could not run the WMS specs — they need a running app and database, which this environment does not have. The helper change is verified by inspection and `tsc` only. It is the one part of this PR without execution behind it.

### m1 — `aria-owns`

Agreed and applied, on the same condition as `aria-controls`, pinned in the accessibility test (`aria-owns === listbox.id`, plus an assertion that the listbox is not inside the input's subtree so the reason the attribute exists is visible in the test).

### m2 — `elevated` dialogs

Correct, and worth recording. I went with the documentation option: `.ai/ui-components.md` now carries a MUST-NOT against putting a `ComboboxInput` inside a `<DialogContent elevated>`, with the reason (portaled popup no longer inherits that stacking context; `z-popover` 45 < `z-modal-elevated` 55) and the explicit note that ordinary dialogs, drawers and side panels are fine.

I did not add the test case: `zindex-overlay.test.tsx:123-138` already pins `z-modal-elevated` (55) > `z-popover` (45), so a scale change fails there today. What a unit test cannot detect is a *future call site* pairing the two, which is why the constraint belongs in the component contract.

### n1 — claimed-but-missing control

Fair catch, and the more useful half of the nit: the control existed as a throwaway I ran while writing the fix and then deleted, so the claim outlived the assertion. It now ships — outside `pointerDown` with suggestions open, expecting `[data-dialog-content]` to be gone — which is what keeps the suppression provably scoped to the portaled list instead of blanketing the dialog's outside-interaction handling.

### n2 — `align="start"`

Removed; `popover.tsx:18` already defaults it.

---

`yarn workspace @open-mercato/ui test` — 232 suites / 1899 tests green. `yarn lint` (forced, no turbo cache) and `yarn typecheck` (26 packages) green. WMS dialog jsdom suites — 11 suites / 78 tests green.

On your validation note: the environmental failures you hit (`@tanstack/react-table/legacy`, unbuilt `agent-transport`) come from an install that predates `yarn build:packages`; after building the workspace packages the full `@open-mercato/ui` suite runs clean here at 232/232, and `tsc -p packages/ui` reports no errors at all. Worth a `yarn install && yarn build:packages` before the next local run.

**Please approve the workflow runs so the real gate executes against `a16639f1e`** — with no local database I cannot run the Playwright suites, and the WMS helper fix is exactly what those runs would exercise.

Frshy:
## Ready for re-review — all five findings addressed

@pkarw everything from your review is in `a16639f1e`. Delta since the commit you reviewed is 5 files, +45/−17:

| Finding | Resolution |
|---|---|
| 🔴 **M1** WMS helper | `fillCombobox` and `selectLocationComboboxOption` now query `page.getByRole('option', …)`. The `if visible / else page-wide button` shape in `selectLocationComboboxOption` is gone rather than repaired — with a correct locator that fallback only papered over the broken one, and it was the strict-mode-ambiguous branch. `fillCombobox` keeps its keyboard fallback, which is a real safety net for slow async suggestions. |
| 🟡 **m1** `aria-activedescendant` | `aria-owns` added next to `aria-controls`, same condition; pinned in the accessibility test together with an assertion that the listbox is outside the input's subtree, so the reason the attribute exists is visible at the assertion. |
| 🟡 **m2** `elevated` stacking | Documented as a MUST-NOT in `.ai/ui-components.md`: no `ComboboxInput` inside `<DialogContent elevated>`, with the reason and the note that ordinary dialogs/drawers/side panels are unaffected. No test added — `zindex-overlay.test.tsx:123-138` already pins 55 > 45; what a unit test cannot catch is a future call site pairing the two. |
| ⚪ **n1** missing control | Shipped: outside `pointerDown` with suggestions open expects `[data-dialog-content]` to be gone. |
| ⚪ **n2** `align="start"` | Removed. |

**One correction to M1 worth carrying into the re-read.** The helper was already broken before this branch, not only by the portal: both call sites click by `role="button"`, but the options have carried `role="option"` since #4358, and `wmsUi.ts` has not been touched since. So the lookup could never match — the `div.absolute` container was the second half of the problem, and this branch removed the remaining chance of it working. Your three consequences all hold; the cause is just older than this PR.

**Green locally:** `@open-mercato/ui` 232 suites / 1899 tests, WMS jsdom dialog suites 11/78, `yarn lint` (forced, no turbo cache) and `yarn typecheck` (26 packages).

**Two things I cannot do from here, both needing a maintainer:**

1. **Approve the workflow runs** so the real gate executes against `a16639f1e`. This matters more than usual now: the M1 fix is a Playwright helper, and the WMS specs need a running app and database that this environment does not have — that change is backed by inspection and `tsc` only. It is the one part of the PR without execution behind it.
2. **Flip `changes-requested` → `review`**, and, once someone has clicked through, the `needs-qa` → `qa-approved` gate. I have no `triage` here, and GitHub returns 404 on the review-request endpoint for a read-only fork contributor, so this mention is the re-review request. (@zielivia is already on the reviewer list from a maintainer.)

On your validation note: the two environmental failures you hit — `@tanstack/react-table/legacy` and the unbuilt `agent-transport` — are a stale install rather than anything in this branch. After `yarn install && yarn build:packages` the full `@open-mercato/ui` suite runs 232/232 here and `tsc -p packages/ui` reports no errors at all, including the 30 pre-existing ones you saw.

pkarw:
@Frshy please solve the merge conflicts

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T06:31:32Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Re-review: fix(ui): portal the combobox dropdown so a dialog's overflow stops clipping it

## 🎯 Summary

This is a re-review of `a16639f1e` against my `CHANGES REQUESTED` review of `2235079`. **All five previous findings are genuinely fixed** — I verified each one against the current head rather than taking the response comment at its word:

- **M1** — both `wmsUi.ts` helpers now use `page.getByRole('option', …)`. The author's follow-up diagnosis is correct and sharper than mine: the `role="button"` lookup could never have matched since the options gained `role="option"`, so the helper was already degraded before this branch. Dropping the `if visible / else page-wide button` shape in `selectLocationComboboxOption` rather than repairing it is the right call.
- **m1** — `aria-owns` is applied on the same condition as `aria-controls` (`ComboboxInput.tsx:431`) and pinned in the accessibility test, including `expect(listbox.contains(input)).toBe(false)` so the *reason* the attribute exists is visible in the test.
- **m2** — the `MUST NOT` against `ComboboxInput` inside `<DialogContent elevated>` is in `.ai/ui-components.md` with the z-index numbers and the reason. I accept the decision not to add a test case: `zindex-overlay.test.tsx` already pins the scale relationship, and no unit test can catch a future call site pairing the two, so the component contract is the right home for it.
- **n1** — the control case ships (`ComboboxInput.dialog.test.tsx:110-133`) and it is a real assertion, not a tautology; it passes on the merged tree.
- **n2** — `align="start"` removed.

The change itself remains well-argued and the mechanism is sound; nothing in my earlier "Checked and clear" section is invalidated by the new commit.

Two blockers stand between this and merge, and **neither is a defect in the code you wrote** — one is the base moving under you, the other is an integration spec I missed on the first pass and should have caught then.

## Verdict

❌ **request changes** — the head conflicts with `develop` and cannot merge as pushed, and `TC-WF-001` scopes its suggestion lookup to the dialog, so it regresses under the portal exactly as `wmsUi.ts` and `TC-PUSH-009` did. Both are mechanical; there is no design question left open.

## 🧪 Validation Gate

**Validation source: GitHub checks** (run `32483770832`, head `a16639f1e`, re-triggered 2026-08-22T06:29Z after the fork-workflow approval). Unlike the previous round, the real gate is executing against this head, so I did not substitute local runs for it.

| Check / command | Status | Notes |
|---|---|---|
| `prepare` (`yarn generate`, `yarn build:packages`) | ✅ PASS | 3m24s |
| `lint` | ✅ PASS | 29s |
| `ds-lint` | ✅ PASS | 39s |
| `audit` / `audit-scope` | ✅ PASS | |
| `documents-multi-instance` | ✅ PASS | 51s |
| `license/cla` | ✅ PASS | |
| `test` (`yarn typecheck`, `yarn test`) | ⏳ PENDING | in progress at review time |
| `ephemeral-integration` (15 shards) | ⏳ PENDING | in progress at review time — **this is what will confirm or refute B2** |
| `docker-build`, `mutate`, `comment`, `[code]smith` | ⏭️ SKIPPED | not applicable to this PR |

**⏳ CI-pending disclosure.** `test` and all fifteen `ephemeral-integration` shards were still running when this review was submitted. Per this skill's contract the verdict is not held for them; the two blockers below stand on their own evidence (a reproduced merge conflict and a statically confirmed locator regression) and do not depend on the pending results. I will post a follow-up comment when CI settles.

**Local run — the merged state, which CI does not cover.** Because the head conflicts, CI is testing code that cannot merge. I resolved the conflict locally (resolution in B1, never pushed) and ran the affected suites against the merged tree from `packages/ui`:

| Command | Status | Notes |
|---|---|---|
| `jest src/backend/inputs/__tests__ src/primitives/__tests__/zindex-overlay.test.tsx` (post-merge tree) | ✅ PASS | 10 suites, 113 tests |

## Findings

### ⛔ Blocker

**B1 — The head conflicts with `develop`; `packages/ui/src/backend/inputs/ComboboxInput.tsx` cannot auto-merge.**

`develop` advanced past your merge base (`33a7d00c4`) with `160f474a7` — *feat(wms): preselect the sole warehouse and support custom fields in the zone dialog (#5334)* — which edits the same file. `git merge-tree` reports exactly one conflicted path; `__tests__/ComboboxInput.test.tsx` auto-merges cleanly.

The conflict is **purely structural, not semantic**. `#5334` added a `userTypedRef` guard so the external-value-sync effect can run while the field is focused; three of its four touch points (`:107` declaration, `:268` effect guard, `:282` in `selectValue`) auto-merge into your version untouched. The remaining two land inside the `onChange` / `onBlur` handlers that your change re-indented into the `PopoverAnchor` wrapper, which is why git gives up on the whole JSX block.

I resolved it locally to confirm the resolution is safe. Take **your** side of the hunk (the `Popover` / `PopoverAnchor` structure) and re-add develop's two assignments:

```diff
             onChange={(event) => {
               setTouched(true)
+              userTypedRef.current = true
               setInput(event.target.value)
               setShowSuggestions(true)
               setSelectedIndex(-1)
             }}
             onKeyDown={handleKeyDown}
             onBlur={() => {
               // Delay closing so clicks on the popup can resolve first. If async
               // suggestions are still loading, keep the dropdown open instead of
               // closing before the first payload arrives.
+              userTypedRef.current = false
               blurClosePendingRef.current = true
```

Nothing else needs touching, and the two changes do not interact: `#5334`'s effect keys on `document.activeElement === inputRef.current`, and your `onOpenAutoFocus` / `onCloseAutoFocus` preventions are precisely what keep focus on the input, so the guard reads the same value before and after the portal. I ran the merged tree through the combobox and z-index suites — 113 tests green — so the resolution is verified, not just plausible.

Caveat that applies to this whole review: I reviewed the diff **as pushed**. The resolution above is the only delta I expect, but re-request review after you push it and I will confirm on the merged head.

**B2 — `TC-WF-001` still scopes its suggestion lookup to the dialog and will fail under the portal.**

`packages/core/src/modules/workflows/__integration__/TC-WF-001.spec.ts:103` and `:108`:

```ts
const dialog = page.getByRole('dialog', { name: 'Create Event Trigger' })
…
const firstSuggestion = dialog.getByRole('option').filter({ hasText: /Created|Updated|Deleted/i }).first()
const customerSuggestion = dialog.getByRole('option').filter({ hasText: /Customer/i }).first()
```

The field behind `getByPlaceholder('sales.orders.updated')` is `EventPatternInput` (`packages/ui/src/backend/inputs/EventPatternInput.tsx:46`), which is **a thin wrapper around `ComboboxInput`** — it passes `suggestions` straight through and renders nothing else. That is why a grep for `ComboboxInput` under `packages/core/src/modules/workflows/` comes up empty, and it is how both of us missed this: the coupling is one component-indirection away from the spec.

`PopoverContent` wraps `PopoverPrimitive.Portal` with no `container` prop (`packages/ui/src/primitives/popover.tsx:19`), so the listbox mounts as a child of `document.body` — a sibling of the dialog's content node, not a descendant. `dialog.getByRole('option')` therefore matches nothing:

1. `:104` — `await expect(firstSuggestion).toBeVisible()` fails on the default expect timeout. That is the first hard failure and it ends the test.
2. `:108-110` — same failure mode for the filtered lookup.
3. `:112` — `customerSuggestion.locator('span').last()` is unreachable, so the "description span shows the raw event ID" assertion never runs.
4. `:124` — `await expect(customerSuggestion).not.toBeVisible()` becomes **vacuously true** even if the dropdown never closes, because a locator that matches nothing is never visible. Fixing lines 103/108 restores this assertion's meaning too, which is worth doing on its own.

The test is active and unskipped (`test.describe('TC-WF-001: Event Pattern Autocomplete')`), and its entire subject is the combobox autocomplete this PR moves — it is the single most on-point spec for this change in the repo. Fix it the same way you fixed `wmsUi.ts` and `TC-PUSH-009`: query from `page`, not `dialog`.

```diff
-      const firstSuggestion = dialog.getByRole('option').filter({ hasText: /Created|Updated|Deleted/i }).first()
+      const firstSuggestion = page.getByRole('option').filter({ hasText: /Created|Updated|Deleted/i }).first()
…
-      const customerSuggestion = dialog.getByRole('option').filter({ hasText: /Customer/i }).first()
+      const customerSuggestion = page.getByRole('option').filter({ hasText: /Customer/i }).first()
```

The running `ephemeral-integration` shards should confirm this; I am reporting it now rather than waiting because the analysis is conclusive and you can fix it in the same push as B1.

### 💅 Nit

**n1 — `fillCombobox`'s keyboard fallback now assumes the popup is still open.**
`packages/core/src/modules/wms/__integration__/helpers/wmsUi.ts:148` waits up to 10 s for `page.getByRole('option', …)` after the `ArrowDown` → `Enter` path already failed to commit a value. If `Enter` closed the list without selecting, that option is gone and the branch burns 10 s before failing. This is **not a regression** — the previous `root.getByRole('button', …)` could never match at all, so the branch was already dead — and the primary path now works, which is the point. Worth a comment noting the assumption, or reopening the list with an explicit `input.click()` before the wait. Your call; it does not block.

## ✅ Cross-consumer sweep (re-verified this pass)

I redid the "does anything else couple to the old DOM" check more thoroughly than last time, since B2 shows the first pass was not exhaustive. Every non-`page`-scoped option/listbox locator in the repo's integration specs, and what backs it:

| Locator | Component | Affected? |
|---|---|---|
| `TC-WF-001:103,108` — `dialog.getByRole('option')` | `EventPatternInput` → **`ComboboxInput`** | ❌ **Yes — B2** |
| `TC-CAL-005:103` — `dialog.getByRole('option')` | `RelatedToField` (own inline `role="option"` markup) | ✅ No |
| `TC-EUDR-010:124` — `productField.getByRole('option')` | `ProductSelectField` → `LookupSelect` | ✅ No |
| `TC-SALES-031:213` — `dialog.getByRole('option')` | `ShipmentDialog` → `LookupSelect` | ✅ No |
| `TC-DOCUMENTS-016:211`, `TC-DOCUMENTS-017:432` | `LookupSelect` | ✅ No |
| `TC-SALES-020:285-322` — `page.getByRole('listbox', { name: /Filters/i })` | DataTable filter menu, page-scoped | ✅ No |
| `wmsUi.ts:223` — `listbox.getByRole('option')` | DS `Select`, listbox resolved from `page` | ✅ No |

`LookupSelect` renders its list **in flow**, not absolutely positioned (`packages/ui/src/backend/inputs/LookupSelect.tsx:265-290`), so it has neither the clipping bug nor the portal consequence — no follow-up needed there. `grep` for the old `div.absolute` / `ancestor::div[contains(@class,"relative")]` shape now returns only your explanatory comment, so `wmsUi.ts` was the last structural coupling of that kind.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `ComboboxInput`'s export and props are unchanged.
- [x] Where a contract had to change: not applicable — only the render target moved.

The one contract that *did* change is an implicit, undocumented one: the popup's DOM position relative to the field. That is not a `BACKWARD_COMPATIBILITY.md` surface, and you have documented it in `.ai/ui-components.md` for both call sites and tests. B2 is the last in-repo consumer of the old shape.

## 🧪 Test Coverage

**What is covered.** `ComboboxInput.dialog.test.tsx` pins the four properties that make the portal safe inside a modal dialog: selection does not dismiss the dialog, `wheel` and `touchmove` reach the list past `react-remove-scroll` (each with its own control), and — new this round — a genuinely outside `pointerDown` still dismisses the dialog, which is what proves the outside-interaction suppression is scoped to the list rather than blanketing the dialog. `ComboboxInput.test.tsx` covers filtering, keyboard navigation, blur commit, `allowCustomValues`, and the accessibility contract now including `aria-owns` plus the assertion that the listbox is outside the input's subtree. `zindex-overlay.test.tsx` pins `z-popover` against both `z-modal` and `z-modal-elevated`. All 113 pass on the merged tree.

**The gap.** Every one of those is jsdom, where a portal is free and `overflow-y-auto` does not clip. The property this PR actually exists to fix — a real dropdown escaping a real scroll container — is only exercised by the Playwright suites, and those are exactly what B2 breaks and what has not yet reported on this head. There is nothing to add here; the coverage argument completes when the `ephemeral-integration` shards go green with `TC-WF-001` fixed.

pkarw:
Thanks @Frshy — all five findings from the first review are genuinely fixed, and your follow-up diagnosis on M1 (the `role="button"` lookup could never have matched since #4358) was sharper than mine. Two blockers remain and neither is a defect in the code you wrote, so I'm handing this back to you for one more short pass:

1. **Merge conflict** — `develop` picked up #5334, which touches `ComboboxInput.tsx`. The resolution is two lines (`userTypedRef.current = true` in `onChange`, `= false` in `onBlur`, on your `Popover` side of the hunk); I resolved it locally and ran the combobox + z-index suites on the merged tree — 113 tests green — so it is verified, not just plausible.
2. **`TC-WF-001`** — `EventPatternInput` is a thin wrapper around `ComboboxInput`, so that spec's `dialog.getByRole('option')` at `:103`/`:108` regresses exactly like `wmsUi.ts` did. That one is on me for missing it in the first pass; the coupling is one component-indirection away from the spec. Same fix — query from `page`.

Both details are in the review. Push them together and re-request review; the automation can pick it up from the new head.

pkarw:

Re-review of `a16639f1e`. All five findings from the review of `2235079` verified fixed (M1 WMS helper, m1 `aria-owns`, m2 `elevated`-dialog contract note, n1 outside-pointerdown control, n2 `align`). Two new blockers: the head conflicts with `develop` (#5334 touched the same file; two-line resolution given, verified locally at 113 tests green on the merged tree), and `TC-WF-001:103,108` scopes its option lookup to the dialog against a now-portaled listbox via `EventPatternInput` → `ComboboxInput`. One nit on the `fillCombobox` keyboard fallback. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

Validation source: GitHub checks (run `32483770832`, head `a16639f1e`). `prepare`, `lint`, `ds-lint`, `audit`, `audit-scope`, `documents-multi-instance` green; `test` and 15 `ephemeral-integration` shards were still running at review time. Swapping the lock for `ci-monitoring` — I will post the CI result when it settles.

pkarw:

Run [`32483770832`](https://github.com/open-mercato/open-mercato/actions/runs/32483770832) settled at 06:52Z on head `a16639f1e`. **One failure, and it is blocker B2 from the review — confirmed, not predicted.**

| Check | Result |
|---|---|
| `test` (`yarn typecheck`, `yarn test`) | ✅ PASS |
| `ephemeral-integration` 1–13, 15 | ✅ PASS |
| [`ephemeral-integration (14/15)`](https://github.com/open-mercato/open-mercato/actions/runs/32483770832/job/96992961659) | ❌ **FAIL** — 147 passed, 1 failed, **0 flaky** |
| `prepare`, `lint`, `ds-lint`, `audit`, `audit-scope`, `documents-multi-instance`, `license/cla` | ✅ PASS |

The single failing test:

```
✘ packages/core/src/modules/workflows/__integration__/TC-WF-001.spec.ts:47:3
  › TC-WF-001: Event Pattern Autocomplete
  › should suggest events, filter them, and display the label on selection (20.1s)
  Test timeout of 20000ms exceeded.

  Retry #1 ─────────────────────────────────────────────────────────
  Test timeout of 20000ms exceeded.  (20.4s)
```

It failed on the initial run **and** on retry #1 with the same 20 s test timeout, and Playwright reports `flaky: 0` — so this is deterministic, not infrastructure noise. Every other spec in the same shard passed, including `TC-WF-008`, which drives the same `EventPatternInput` field but commits it with keystrokes instead of clicking a suggestion, and therefore never queries the portaled listbox. That contrast is the mechanism in one line.

The timeout (rather than a fast assertion failure) is what you would expect from two consecutive `expect(...).toBeVisible()` waits on locators that can never match — `:104` and `:110` — overrunning the 20 s test budget before either can report individually. The `trace.zip` and `error-context.md` artifacts on that job pin the exact step if you want to see it.

**Nothing to change in the diagnosis or the fix** — repoint `:103` and `:108` from `dialog` to `page`, as in the review. Ship it with the B1 conflict resolution in one push.

Removing `ci-monitoring`; the PR stays `changes-requested` and assigned to @Frshy. No further follow-up will come from this agent until a new head is pushed.

Frshy:
Both blockers addressed in `c1c01865d` — the PR now reports `MERGEABLE`.

| Finding | Resolution |
|---|---|
| ⛔ **B1** merge conflict | `develop` merged in (`50faaf54c`). Resolution is exactly the one you gave: my side of the hunk (`Popover` / `PopoverAnchor`), plus #5334's `userTypedRef.current = true` in `onChange` and `= false` in `onBlur`. #5334's other three touch points (`:107` declaration, the effect guard, `selectValue`) auto-merged untouched. |
| ⛔ **B2** `TC-WF-001` | `:105` and `:110` now query `page.getByRole('option', …)` instead of `dialog…`, with a comment naming the reason (portaled to `<body>`, sibling of the dialog). That also restores meaning to the `not.toBeVisible()` assertion at `:124`, which was vacuously true. |
| 💅 **n1** `fillCombobox` fallback | Reopened rather than commented: `ArrowDown` on a closed combobox re-opens the list (`ComboboxInput.tsx` `handleKeyDown`), so the last-resort branch now presses `ArrowDown` before the 10 s wait. Without it the options are unmounted after a failed `Enter` and the wait could only ever time out. |

**Validation** (merged tree, local): `jest packages/ui/src/backend/inputs/__tests__ src/primitives/__tests__/zindex-overlay.test.tsx` → ✅ 10 suites, 119 tests. `tsc --noEmit` on `packages/ui` and `packages/core` → no errors in any file this PR touches. The integration side of B2 is what CI needs to confirm — please approve the workflow run on `c1c01865d`.

@pkarw re-requesting review.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-28T07:46:24Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Re-review: fix(ui): portal the combobox dropdown so a dialog's overflow stops clipping it

## 🎯 Summary

This is a re-review of `c1c01865d` against my `CHANGES REQUESTED` review of `a16639f1e`. **Both blockers are genuinely fixed, and unlike the last round I now have CI confirmation rather than only static analysis** — the full 15-shard integration matrix is green on this head, including shard 14, which is the shard that failed round two on exactly the test B2 named.

- **B1 — merge conflict.** `develop` is merged in at `50faaf54c` and the resolution is the one I proposed: your side of the JSX hunk (the `Popover` / `PopoverAnchor` structure) plus #5334's two assignments re-added. I verified the merge is complete rather than trusting the commit message — relative to the develop commit that was merged (`94f1126c1`), the PR head differs in **exactly the eight files this PR owns** and nothing else, so no develop hunk was dropped and nothing extra was picked up. All five of #5334's `userTypedRef` touch points are present (`ComboboxInput.tsx:107` declaration, `:268` effect guard, `:282` in `selectValue`, `:464` in `onChange`, `:474` in `onBlur`), matching develop's five one-for-one, and the sync-effect body is character-identical to develop's.
- **B2 — `TC-WF-001`.** `:105` and `:110` now query `page.getByRole('option', …)`, with a comment naming the reason. This also restores meaning to the `not.toBeVisible()` assertion at `:124`, which was vacuously true before because a locator that matches nothing is never visible.
- **n1 — `fillCombobox` fallback.** Reopened with `ArrowDown` rather than merely commented, which is the better of the two options I offered.

`git merge-tree` against the current tip of `develop` reports no conflict, so the branch is still cleanly mergeable 91 commits later.

## Verdict

✅ **approve** — no blockers, no majors, no minors. Four nits below, none of which need addressing before merge. The remaining gate is manual QA, and I have posted the P0/P1 routes in a separate comment.

## 🧪 Validation Gate

**Validation source: GitHub checks** (run [`32700146341`](https://github.com/open-mercato/open-mercato/actions/runs/32700146341), head `c1c01865d`). The run has settled `success`: **25 pass / 4 skipped / 0 fail**. No local gate was substituted and the review worktree never needed `node_modules`.

| Check / command | Status | Notes |
|---|---|---|
| `prepare` (`yarn generate`, `yarn build:packages`) | ✅ PASS | 4m32s |
| `test` (`yarn typecheck`, `yarn test`) | ✅ PASS | 9m6s |
| `ephemeral-integration` 1–15 | ✅ PASS | **all fifteen shards**, full matrix scope |
| `lint` / `ds-lint` | ✅ PASS | `ds-lint` clears the `w-[var(--radix-popover-trigger-width)]` arbitrary value |
| `audit` / `audit-scope` / `scope` | ✅ PASS | |
| `documents-multi-instance` | ✅ PASS | 58s |
| `license/cla` | ✅ PASS | signed |
| `docker-build`, `mutate`, `comment`, `[code]smith` | ⏭️ SKIPPED | not applicable to this diff |

**Shard 14 is the one that matters.** Round two failed there with `TC-WF-001.spec.ts:47` timing out at 20 s on both the initial run and retry #1, `flaky: 0`. On this head the same shard passes. That is direct confirmation of B2's fix, not an inference from it.

Two green checks that carry **no** signal for this diff, so nobody reads them as coverage: `mutate` skips (the stryker allowlist is frozen to `['shared']`, and this diff is in `packages/ui` and `packages/core`), and `Check i18n usage` runs `continue-on-error`.

## Findings

### 💅 Nits

**n1 — the `fillCombobox` reopen helps a narrower set of cases than the comment implies.**
`packages/core/src/modules/wms/__integration__/helpers/wmsUi.ts:151`. The branch is reached only when `page.getByRole('option', …)` was already not visible within the 2 s check at `:131`, so reopening cannot conjure an option that never existed; what it genuinely rescues is the *slow suggestions* case, where the payload landed after that 2 s window and `Enter` then closed the list. There is also a second-order gap: under `allowCustomValues={false}` a failed `Enter` runs `confirmSelection`, which can revert the input text (`ComboboxInput.tsx:316-323`), so the reopened list is filtered by the reverted text and the `suggestionPattern` option can still be absent. This is strictly better than what was there — the previous `role="button"` lookup could never match at all — so it is a note for whoever next touches this helper, not a change request.

**n2 — the PR body's Testing numbers predate the develop merge.**
`232 suites / 1898 tests` and `11 suites / 78 tests` were measured on `2235079`/`a16639f1e`; your `c1c01865d` follow-up comment gives the current `10 suites / 119 tests` for the combobox + z-index scope. Worth folding into the body so the description matches the head.

**n3 — the CLA checkbox in the body is still unticked** even though the `license/cla` check reports the agreement signed. Cosmetic, but it reads as an outstanding prerequisite to anyone skimming the description.

**n4 — the Playwright suites prove the list is *findable*, not that it is *unclipped*.**
Playwright's `toBeVisible()` passes on an element that an ancestor's `overflow: hidden` is clipping, because the element still has a non-empty bounding box and is not `visibility: hidden`. `TC-WF-001` and `TC-PUSH-009` therefore assert that the option resolves from the document, which is the DOM consequence of the fix rather than the visual property the fix exists for. That is not a gap you should close with more Playwright — an unclipped-over-a-dialog assertion needs pixel or bounding-box comparison against the dialog's clip rect — and it is exactly what the `needs-qa` label is carrying. I have made it the P0 in the manual-QA comment. Your headless-Chromium reproduction table already covers it as author evidence; QA just needs to reproduce it in the real app.

## ✅ Checked and cleared this pass (do not re-derive)

**The `<DialogContent elevated>` MUST NOT has zero existing violations.** Round two accepted the documented contract without checking whether any call site already breaks it; I checked this pass. `elevated` appears on exactly one `DialogContent` in the whole repository — `packages/ui/src/backend/filters/AdvancedFilterPanel.tsx:163`, the *Save filter* dialog, which is elevated precisely so it renders above the advanced-filter `PopoverContent` hosting it. That dialog contains an `Input` and nothing else; neither `AdvancedFilterPanel` nor `AdvancedFilterBuilder`, `FilterFieldPicker`, `QuickFilters` or `FilterEmptyState` imports `ComboboxInput`. So the constraint is genuinely forward-looking and this PR ships no live regression behind it.

**The cross-consumer sweep is clean, redone from scratch.** Every container-scoped option/listbox locator in the repository's integration specs, and what backs each:

| Locator | Component | Affected? |
|---|---|---|
| `TC-WF-001:105,110` | `EventPatternInput` → **`ComboboxInput`** | ✅ Fixed this round |
| `TC-PUSH-009:79` | `CrudForm` combobox field → **`ComboboxInput`** | ✅ Fixed in `a16639f1e` |
| `wmsUi.ts:129,152,202` | **`ComboboxInput`** | ✅ Fixed in `a16639f1e` |
| `TC-CAL-005:103` | `RelatedToField` (own inline `role="option"` markup) | ❌ No |
| `TC-EUDR-010:137` | `ProductSelectField` → `LookupSelect` | ❌ No |
| `TC-SALES-031:213` | `LookupSelect` | ❌ No |
| `TC-SALES-020:287-322` | DataTable filter menu (DS `Select`) | ❌ No |
| `TC-DOCUMENTS-016:211`, `TC-DOCUMENTS-017:362` | `LookupSelect` | ❌ No |
| `wmsUi.ts:227` | DS `Select`, listbox resolved from `page` | ❌ No |

`TC-WF-008` drives the same `EventPatternInput` but commits with `pressSequentially` + `Enter` and never queries the listbox, which is why it stayed green through both rounds — the contrast that made B2 legible. The eudr `CountrySelectField` is a `ComboboxInput` host with no browser spec driving it (every `country` reference under `eudr/__integration__` is API-level). A grep for the old structural shapes (`div.absolute`, `ancestor::div[contains(@class,"relative")]`) now returns only your explanatory comment in `wmsUi.ts:200`.

**The jsdom side is proven by CI, not by assertion.** The develop merge brought in #5334's own `+142` lines of `ComboboxInput.test.tsx` cases, written against the pre-portal DOM; they pass on this head, and the green `test` check covers every jest suite in the workspace, so no RTL `within(container).getByRole('option')` coupling survives anywhere.

**The two test files this PR touches are strengthened, not weakened.** `ComboboxInput.test.tsx` (+95) and `ComboboxInput.dialog.test.tsx` (+134) are purely additive against develop — no existing assertion was deleted or relaxed to accommodate the portal. `zindex-overlay.test.tsx` is the only file with a removed line, and the replacement is stricter: it resolves the list on `document.body` *and* adds `expect(container.contains(suggestions)).toBe(false)`, so the guard now pins the reason the lookup moved rather than just tolerating it.

**`z-popover` still reaches the rendered list.** It comes from `PopoverContent`'s own base class (`packages/ui/src/primitives/popover.tsx:25`), not from the call site, and the `min-w-0` in your `className` correctly overrides that base's `min-w-[280px]` through `twMerge` — without it every narrow combobox would get a 280 px popup. `w-[var(--radix-popover-trigger-width)]` is the same idiom `select.tsx:127,136` already uses for `--radix-select-trigger-width`, so it is house precedent rather than a new arbitrary value.

**`develop` has not touched `ComboboxInput.tsx`, `popover.tsx` or `dialog.tsx` since the merge base**, so the 91-commit lag carries no risk on the files this PR owns.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `ComboboxInput`'s export and props are unchanged.
- [x] Where a contract had to change: not applicable — only the render target moved.

Nothing in `BACKWARD_COMPATIBILITY.md` is touched. The one contract that did change is implicit and undocumented — the popup's DOM position relative to the field — and it is now written down in `.ai/ui-components.md` for both call sites and tests, with every in-repo consumer of the old shape repaired.

## 🧪 Test Coverage

Three layers, and they divide the work correctly. `ComboboxInput.dialog.test.tsx` pins what makes the portal safe inside a modal dialog: selection does not dismiss the dialog (with a control proving a genuinely outside `pointerDown` still does), `wheel` and `touchmove` reach the list past `react-remove-scroll` (each with its own control), and exactly one `role="dialog"` node exists while the list is open. `ComboboxInput.test.tsx` covers the portaled accessibility contract including `aria-owns` and the assertion that the listbox is outside the input's subtree. The Playwright specs now prove the list resolves from the document in three real hosts — a CrudForm field, a dialog-hosted `EventPatternInput`, and the WMS dialogs — across a full 15-shard matrix.

The one property no automated layer asserts is the visual one (n4), which is what `needs-qa` is for.

---

*Re-review round 3. autofix: skipped (not my PR — re-run with `--autofix` to fix it here).*

pkarw:

**Why this needs a human even though CI is fully green:** every automated layer here asserts that the suggestion list *resolves from the document*. None asserts that it is *unclipped* — Playwright's `toBeVisible()` passes on an element an ancestor's `overflow: hidden` is clipping, because it still has a non-empty bounding box. The visual property this PR exists for is only observable by eye.

### P0 — The bug itself: a combobox dropdown inside a dialog

**Where to click**
- `/backend/wms/inventory` → open the **Adjust inventory** dialog (also worth repeating in **Move**, **Receive** and **Reserve**)
- `/backend/wms/cycle-counts` → the **Cycle count wizard** dialog
- `/backend/customers/deals/pipeline` → **Quick deal** dialog → the **Company** field ("Pick a company or add new…")
- `/backend/workflows` → open any definition → **Add Trigger** → the **Event pattern** field (placeholder `sales.orders.updated`)

**What to verify**
- Click the SKU / location / company / event-pattern field → the suggestion list renders **fully**, floating over the dialog's edge, not cut off at the dialog boundary. On `develop` a short dialog shows a sliver or nothing at all.
- The list width matches the input width, and it is anchored directly under the field.
- Open a combobox near the **bottom** of a short viewport (resize the window down, or use the last field in a tall dialog) → the list should **flip above** the input and stay fully on screen rather than running off the bottom.
- Scroll the dialog body while the list is open → the list re-anchors to the input and follows it.
- Pick an option by mouse click → the value commits and **the dialog stays open**.
- Pick an option by `ArrowDown` + `Enter` → same result; focus never leaves the text input (the caret stays in the field throughout).
- Press `Escape` once → the list closes and the dialog stays open. Press `Escape` again → the dialog cancels.
- With a long list (a search term matching many SKUs or locations), **scroll the list with the mouse wheel**, then again by touch/trackpad on a touch device. Both must scroll the list. This is the regression a portal would otherwise introduce, because a modal dialog locks scrolling at the document level.
- At the top and bottom of that list, keep scrolling → the page behind must **not** start scrolling (overscroll containment).
- Click genuinely outside the dialog → the dialog still dismisses. The fix must not over-block outside interaction.

**What can go wrong**
- List clipped, or rendering *behind* the dialog or its overlay.
- Clicking an option closes the whole dialog instead of just committing the value.
- The wheel/touch scroll is frozen over the list inside a modal dialog.
- `Escape` closes the dialog instead of the list on the first press.
- Focus jumps out of the text input when the list opens or closes (breaks type-ahead and the `ArrowDown`/`Enter` path).
- The list stays pinned to its original position when the dialog body scrolls.

### P1 — The same component outside a dialog (no regression)

**Where to click**
- `/backend/wms/inventory` → the inline SKU / location comboboxes on the console page itself (not in a dialog)
- `/backend/translations` → the entity / field comboboxes
- `/backend/warranty_claims/settings` → the combobox fields
- `/backend/config/scheduled-jobs` → create/edit a job → its combobox fields
- Any list page with **Advanced filters** open, and a `CrudForm` create/edit page with a combobox field

**What to verify**
- Type-to-filter, keyboard navigation, the "no matches" and loading rows, the clear (×) button, and blur-commit all behave exactly as before.
- The list still paints above surrounding page chrome (sticky headers, the topbar, table toolbars).
- Perceived performance: cold-load one of these routes, confirm the first useful shell renders promptly, and that opening/typing in a combobox stays responsive.
- Repeat one dialog case at a **mobile viewport** (390×844) — the list should size to the field and stay on screen.

**What can go wrong**
- A regression in the non-dialog case, which is the common one and is not what the PR set out to change.
- The popup rendering at a stale position after the page behind it scrolls.
- The popup escaping a surface it should stay within, or covering the field it belongs to.

### Known constraint (not a bug to file)

A `ComboboxInput` must **not** be placed inside a `<DialogContent elevated>` — the portaled list sits at `z-popover` (45), below `z-modal-elevated` (55). There is currently no such call site in the repo (the only `elevated` dialog is the *Save filter* dialog in `AdvancedFilterPanel`, which has no combobox), so this should not be reachable during QA. It is documented in `.ai/ui-components.md`.

### Pass/fail

pkarw:

Round-3 re-review of `c1c01865d`. Both blockers from the review of `a16639f1e` verified fixed: B1's develop merge is complete (the head differs from the merged develop commit in exactly the eight files this PR owns, and all five of #5334's `userTypedRef` touch points are present), and B2's `TC-WF-001` repoint is confirmed by **shard 14 passing** — the same shard that failed round two on that exact test. n1 addressed by reopening the list rather than commenting the assumption.

0 blockers, 0 majors, 0 minors, 4 nits (none blocking). Labels: `changes-requested` → `merge-queue`; `bug`/`needs-qa`/`priority-medium`/`risk-medium`/`partner-request` re-derived and kept. `qa-approved` deliberately not applied — the QA-approval gate is what holds the merge, and P0/P1 routes are in the manual-QA comment above.

**Validation source: GitHub checks.** Run [`32700146341`](https://github.com/open-mercato/open-mercato/actions/runs/32700146341) settled `success`: 25 pass / 4 skipped / 0 fail, all fifteen `ephemeral-integration` shards green at full matrix scope. No `ci-monitoring` — CI is already settled, so no follow-up is pending. `mutate` skipping and `Check i18n usage` being `continue-on-error` mean neither carries signal for this diff; said so in the review rather than letting green read as coverage.

No further action from this agent. Merge is gated only on manual QA.

zielivia:
🤖 `om-ux-review-pr` — evidence-first design review

## 🔍 Design review — fix(ui): portal the combobox dropdown so a dialog's overflow stops clipping it

**Contract**: `.uxproof/` found — 228 tokens (185 color), 731 components; the popup was judged against it and conforms (DS `Popover` primitive, `bg-popover` / `border-input` / `rounded-md` tokens, no hardcoded colors).
**Screens walked**: WMS inventory console (Zapasy), Receive (Przyjęcie towaru — full task: variant by keyboard, warehouse, location, qty, submit), Adjust (Korekta stanu — list open, wheel scroll, Escape chain), Create warehouse + Create location (data created through the UI to unlock the flows), the post-receive putaway dialog, and the non-dialog scope combobox on the inventory page as a regression control. Viewports: 1280×800 light and dark (dark via the app's own theme switch), 900×380, and 375×812 with real touch events. Selection was exercised by mouse click, by ArrowDown/Enter, and by tap.
**Not walked**: Move/Reserve/CycleCount dialogs and QuickDealDialog (same `ComboboxInput` instance, no distinct behavior expected); warranty-claims settings usage; mobile-landscape. The environment had no WMS data, so warehouse WH-01, location A-01-01 and a receipt of 25 × AURORA-CELESTIAL-L were created through the UI as part of the walk.

### 🔍 Findings (worst first)

#### 1. On short viewports the fixed `max-h` pushes part of the list off-screen `[PLATFORM]`
- **Where**: Receive dialog at 900×380 — 📸 `03-short-viewport-overflow.png`. Measured: input top 139, popup `data-side="bottom"`, rect 179→419 against a 380px viewport — the last 39px of the list render below the screen edge.
- **Evidence**: the popup keeps `max-h-48 sm:max-h-60` (240px) regardless of room. Radix computes `--radix-popover-content-available-height` exactly for this case, and the sibling `--radix-popover-trigger-width` is already used one property earlier. Collision flip cannot rescue it here: above the input there is even less room (139px), so Radix correctly stays on `bottom` and the overflow is unavoidable with a fixed cap. The list content stays reachable by scrolling inside the visible part and by keyboard (ArrowDown auto-scrolls the active option into view — verified), so this is degraded, not broken.
- **Pattern**: clamp the cap to the available space: `max-h-[min(15rem,var(--radix-popover-content-available-height))]` (and the `max-h-48` analogue for the base breakpoint) on the `PopoverContent` className.
- **Trade-off**: on very short viewports the list gets shorter and scrolls sooner; that is the intended behavior of the Radix var and strictly better than painting under the screen edge.
- **Accept when**: at 900×380 with the Receive dialog open, the whole popup rect fits inside the viewport (`bottom <= innerHeight`) while width still equals the trigger width.

#### 2. Putaway dialog shows a raw UUID in the Wariant field — pre-existing, out of this PR's increment `[HEURISTIC]`
- **Where**: the "Odłóż do finalnego miejsca" dialog that opens right after a successful receive — 📸 `07-putaway-variant-field.png` shows `42e1ae09-4c29-4a5c-90e0-bdfe17e302f0` where the operator just saw "AURORA-CELESTIAL-L".
- **Evidence**: Nielsen #2 (match between system and the real world). By code inspection this is not caused by this PR — `ComboboxInput`'s value→label resolution is untouched by the diff; the putaway prefill passes the variant id without a resolvable option, which rendered the raw value before this branch too.
- **Pattern**: prefill the combobox with the variant's SKU/label (the receive flow already has it in hand), or resolve the label the way the Receive dialog's own variant field does after selection.
- **Trade-off**: one extra lookup (or one extra prop) in the receive→putaway handoff.
- **Accept when**: after receiving a variant, the putaway dialog's Wariant field shows its SKU, not an id. Fine to spin off as a separate issue — flagged here only because the walk runs straight into it.

### 📸 Evidence

**Adjust dialog, light** — the portaled list overlays the dialog's fields instead of being clipped by its `overflow-y-auto`; width matches the trigger 1:1 (462px = 462px), 4px gap preserved (parity with the old `mt-1`):

01-adjust-light-list

**Finding 1** — Receive dialog at 900×380: the list crosses the bottom viewport edge by 39px (`SERV-HAIR-60-SENIOR` is the cut-off row):

03-short-viewport-overflow

**Dark mode** via the app's own toggle — popup on `bg-popover` tokens, correct borders and contrast:

04-dark-list

**Mobile 375×812, touch** — list 327px wide inside the 375px viewport, options 58px tall (comfortably above the 44px touch minimum); tap-select commits the value and the dialog stays open:

05-mobile-receive-list
06-mobile-after-select

**Finding 2** — raw UUID in the putaway Wariant field right after receiving `AURORA-CELESTIAL-L`:

07-putaway-variant-field

### ✅ Summary

- **Strong**: the fix delivers exactly what it promises, verified against the running app rather than the diff: the list is portaled to `<body>` (confirmed in the DOM), overlays every dialog it used to be clipped by, keeps focus in the input, keeps `aria-owns`/`aria-controls`/`aria-activedescendant` wired, keeps exactly one `role="dialog"` node, closes on first Escape while the dialog survives to the second, reopens on ArrowDown, and — the risky part — wheel and real touch-drag both scroll a long list inside a modal while the same gestures outside the list stay locked (control checks passed both ways). The non-dialog combobox is unaffected, and the `.ai/ui-components.md` update matches observed behavior, including the `elevated`-dialog caveat (no current usage violates it).
- **Must change**: nothing blocks this PR. Finding 1 is the one worth folding in while the file is open — it is a one-line className change with a measured repro.
- **Opinion**: finding 2 is adjacent, pre-existing, and fine as a follow-up issue rather than scope creep here.

- One verification note for honesty: a single synthetic CDP wheel tick does not move this list in headless Chromium (the second tick onward does, and headed Chromium plus real touch scroll from the first gesture) — if a future integration test asserts wheel scrolling, it should send more than one tick or it will flake.

### 📋 Applied

Design contract check (pass, no `[PRODUCT]` findings), state matrix (default / loading / dark / narrow / mobile-touch walked; empty-suggestions state is pre-existing behavior and unchanged by this diff), humane gate (no persuasive elements in these operational dialogs — nothing to flag), repo-local rules (no `UX_REVIEW.md`, no `.uxproof/conventions.md` manual section present).

_Advisory review: findings are input for the author, not a merge gate. This skill applies no labels and blocks no merge._

zielivia (approved):

PR #5448, opened 2026-08-21
fix(shared): warn on duplicate entity class names across modules

Frshy:
## Summary

When two enabled modules each export an entity class with the same JS class name (e.g. both `export class Invoice`), entity resolution is silently corrupted rather than failing. The visible failure lands far from the cause — ORM bootstrap misbehaves, downstream startup steps get skipped, and unrelated subsystems fail (missing role features producing broad 403s, readiness probes returning 500) — so diagnosing it from the symptoms takes a long time.

Nothing catches it today: `yarn generate`, `yarn build`, `tsc`, per-module unit tests and `yarn db:generate` all pass. An ORM init containing only one of the two modules' entities passes too.

Verified first-hand against the installed `@mikro-orm/core@7.1.9`, there are three independent reasons:

1. **`discovery.checkDuplicateEntities` is a dead option.** It is set as a default in `utils/Configuration.js` but is read nowhere in the shipped package, and is not declared in `Configuration.d.ts`'s `MetadataDiscoveryOptions`. `MetadataDiscovery.js` contains no class-name duplicate check at all — so this cannot be fixed by enabling an existing option.
2. **The one live duplicate check compares table names, not class names.** `MetadataValidator.validateDiscovered` throws `MetadataError.duplicateEntityDiscovered` (misleadingly named — its message is `Duplicate table names are not allowed`) and compares `schema + tableName`. It only fires if both classes fall back to the default `classToTableName(class.name)`; every entity in this repo declares an explicit distinct `tableName`, so it never fires.
3. **Name-based metadata lookup silently aliases.** `MetadataStorage` holds both a constructor-keyed map and a name-keyed one, and `find()` falls through to the name-keyed one. When discovery calls `get(cls, true)` for a second class sharing a name, it resolves the *first* class's metadata instead of creating its own — so the two classes collapse onto one set of metadata, table name included. Class-based lookups such as `em.find(Invoice)` keep working, which is exactly why this never fails loudly; everything resolved by name (string relation targets, `getRepository('<Name>')`, relation discovery, serialization) silently gets the wrong entity.

`yarn db:generate` cannot see it either: it initialises MikroORM once per module with `MetadataStorage.clear()` in between, so cross-module collisions are structurally invisible there.

This PR makes the collision self-announcing on both surfaces. Both are **warnings** — generation and bootstrap still complete unchanged.

## Changes

- **New `packages/shared/src/lib/db/duplicateEntityClassNames.ts`** — dependency-free detection and message formatting. No ORM import, deliberately, so the CLI generator can share it without pulling MikroORM into `yarn generate`.
- **New `packages/shared/src/lib/db/duplicateEntities.ts`** — adapts a registered ORM entity array to the detector, filtering to values that actually carry `MetadataStorage.PATH_SYMBOL` (or are `EntitySchema` instances).
- **`packages/shared/src/lib/db/mikro.ts`** — `registerOrmEntities()` now reports collisions as a `[Bootstrap]` warning. This is the single point every runtime registration path funnels through (`lib/bootstrap/factory.ts` and `lib/testing/bootstrap.ts`), so one call site covers both.
- **New `packages/cli/src/lib/generators/entity-class-declarations.ts`** — a TypeScript-AST parse collecting `@Entity()`-decorated exported classes. Handles stacked decorators with interleaved comments, `export { X }`, aliased imports (`import { Entity as OrmEntity }`), namespace-qualified decorators (`@orm.Entity()`), and infers script kind from the file extension since `MODULE_CODE_EXTENSIONS` accepts `.tsx`/`.jsx`.
- **`packages/cli/src/lib/generators/module-entities.ts`** — warns during generation, from the generator that produces the colliding registry.

The message names the class, the contributing modules and their source files, and states the fix:

```
[Entities Warning] Duplicate entity class name(s) defined by more than one enabled module: "ApiKey".
MikroORM resolves entities by JS class name for string relation targets, getRepository('<Name>'), relation discovery
and serialization, so one silently shadows the other — the shadowed class ends up aliased to the surviving class's
metadata, table name included. Class-based lookups such as em.find(<Name>) keep working, which is why this never
fails loudly.
Rename all but one of the classes below so every entity class name is unique across enabled modules, then update
their exports, relation targets and imports. Table names may stay as they are.
  ApiKey
    - api_keys (/…/packages/core/src/modules/api_keys/data/entities.ts)
    - example (/…/apps/mercato/src/modules/example/data/entities.ts)
```

### Design notes

- **False positives are avoided deliberately.** The registered array holds *every* function export of each module's entity file, so plain helper functions travel alongside entities. Only classes carrying `MetadataStorage.PATH_SYMBOL` are compared at runtime; only `@Entity()`-decorated classes at generate time.
- **The same class registered twice never warns.** Comparison is by constructor identity first, so a class re-exported through a second import path, or re-registered by an HMR reload, collapses to one source. The existing HMR debug line in `registerOrmEntities` is untouched.
- **Reported once per process.** Development re-runs registration on every HMR reload; reprinting a multi-line warning each time would bury it. A newly introduced collision still reports. The record lives on `globalThis` because a module-level set would reset on the very reloads it exists to deduplicate across — the same reason the entity registry above it already does.
- **It can never break its host.** Detection is a pure function, separate from the logging, and both reporting boundaries are guarded, so an exotic export (a throwing getter, a trapping proxy) degrades to a skipped check rather than a failed bootstrap.
- **Warning-only is deliberate**, and the split between detection and reporting means escalating to a throw later is a one-line change per call site.

### Deliberately out of scope

The generator stamps `entityName = "<moduleId>.<ExportName>"` on every entity export. MikroORM v7 reads that property nowhere (`Utils.className` returns plain `class.name`, and `EntitySchema`'s constructor even overwrites an explicitly passed `name`), so the stamp is **vestigial**. It is left alone and only read as a source of the module id for the runtime message. Making it authoritative would need a custom metadata provider / naming strategy touching every name-based lookup, entity id and serialization path — a much larger change.

### Known limitations, both backstopped by the runtime check

- The generate-time pass is static, so entities re-exported via `export * from` are not seen.
- For app modules bundled through esbuild, the runtime source path resolves to the generated bundle rather than the original file; the module id is still correct.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**

`.ai/specs/` was checked and holds no spec covering ORM entity registration or discovery. This is a self-contained diagnostic for an existing platform footgun — no new feature, module, API surface, or database change — so it was treated as not requiring one. Happy to write a spec if maintainers would prefer it recorded.

## Testing

Full gate, per `.ai/agentic.config.json`, run locally: `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn lint` (0 errors), `yarn test`, `yarn build:app`.

Unit coverage added:

- `duplicateEntityClassNames.test.ts` — the pure detector and message formatter, including three-way collisions, degraded module id / path rendering, and the fail-open contract for unidentifiable entries.
- `duplicateEntities.test.ts` — a **characterization test pinning the upstream MikroORM behaviour** this guard exists for (`MetadataStorage` aliasing a second same-named class onto the first; DB-free, public API only). It documents *why* the guard exists and will fail if a future MikroORM starts detecting class-name collisions itself. Plus adapter coverage and resilience against hostile exports.
- `registerOrmEntities.duplicates.test.ts` — warns *and still registers*; silent for unique names, the same class twice, and non-entity exports sharing a name; reports once across re-registration; still reports a newly introduced collision.
- `entity-class-declarations.test.ts` — parser coverage: stacked decorators with interleaved comments, `export { X }`, bare `@Entity`, aliased and namespace-qualified decorators, non-exported and non-entity classes, `.tsx` files, unreadable files.
- `module-entities.test.ts` — generator warns on a real collision and still generates; silent when names are unique or only one colliding class is an entity.

Verified end to end by temporarily introducing a real collision: `yarn generate` printed the warning naming both modules and files, and a real app bootstrap logged the `[Bootstrap]` warning and then **completed** — confirming it is warning-only — and stayed silent on a second registration. `yarn generate` emits nothing for the current module set (272 entity classes across 51 files), so there are no pre-existing collisions and no false positives.

**Performance:** runtime detection over the full 254-entity registry costs **0.071 ms** per `registerOrmEntities()` call; generate-time parsing of all 51 entity files costs **125 ms**, well under 1% of a `yarn generate` run.

**On integration tests (`.ai/qa/tests/`):** none added, and documented here as not applicable. The change adds no HTTP route, no UI surface, and no user-visible behaviour — its entire output is a log line at ORM registration and a console warning during generation, neither of which is reachable from an integration test driving the app. Both are covered by unit tests plus the end-to-end manual verification above.

## Checklist

- [ ] QA routing set: `skip-qa` for low-risk non-customer-facing changes — and for changes with no manually exercisable UI surface that leave the database structure and API surface unchanged, preserve `BACKWARD_COMPATIBILITY.md` contracts, and ship automated tests for the behavior they change — otherwise `needs-qa`.

> **On the four unticked boxes above.** The CLA box is deliberately left for a human to tick — an agent must not accept a legal agreement on a contributor's behalf. The spec box is unticked because no spec was created (see the Specification section for why). The three label boxes cannot be ticked from this account: it has read-only permission on this repository, so it cannot apply labels here. Intended labels and rationale are in a follow-up comment for a maintainer to apply.

### Design System Compliance

Not applicable — this change touches no UI. It adds no `.tsx`, nothing under `packages/ui/src/`, and nothing under any `components/` tree; the only user-visible output is a log line and a console warning.

- [ ] No hardcoded status colors (`text-red-*`, `bg-green-*`, `text-emerald-*`, `bg-amber-*`, `bg-blue-*`) — use semantic tokens
- [ ] No arbitrary text sizes (`text-[Npx]`) — use typography scale or `text-overline`
- [ ] `aria-label` on all icon-only buttons (`<IconButton>`)
- [ ] Uses existing DS components (Alert, StatusBadge, FormField) — no custom replacements

## Linked issues

No tracker issue is linked — this was reported directly. Happy to open one if maintainers prefer it tracked.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5448"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789911956&installation_model_id=432243&pr_number=5448&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5448&signature=461b92b3bb6ab689e55eba76aadab0233f27a2ea65070e3250d74cb1086289c6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.ai/lessons/standalone-generators-must-reuse-package-generated.md
+++ b/.ai/lessons/standalone-generators-must-reuse-package-generated.md
@@ -11,6 +11,6 @@ topics: ["auto-discovery","build-output","data-scoping"]
 
 **Problem**: The entity-id generator parsed exported classes and property declarations from module entity files. That works against monorepo `src` files, but compiled `dist/modules/**/data/entities.js` files do not preserve that source shape, so standalone generation silently dropped package entities like `organization`.
 
-**Rule**: In standalone mode, when building app-level generated entity IDs/field shims for package-backed modules, prefer the package's shipped `generated/entities.ids.generated.ts` and `generated/entities/*/index.ts` artifacts. Do not rely on parsing compiled `dist` entity files for source-level declarations.
+**Rule**: In standalone mode, when building app-level generated entity IDs/field shims for package-backed modules, prefer the package's shipped `generated/entities.ids.generated.ts` and `generated/entities/*/index.ts` artifacts. Do not rely on parsing compiled `dist` entity files for source-level declarations. Parsing compiled output for *class names alone* is an acceptable last resort when no generated artifact carries them — the entity-registry duplicate-class-name check prefers the package's shipped `src/modules` mirror and only falls back to `dist` for that one fact.
 
 **Applies to**: `packages/cli/src/lib/generators/entity-ids.ts` and standalone `create-app` generation paths.
--- a/packages/cli/src/lib/generators/__tests__/entity-class-declarations.test.ts
+++ b/packages/cli/src/lib/generators/__tests__/entity-class-declarations.test.ts
@@ -1,0 +1,97 @@
+import fs from 'node:fs'
+import os from 'node:os'
+import path from 'node:path'
+import { parseEntityClassNames } from '../entity-class-declarations'
+
+let tmpDir: string
+
+function writeSource(content: string): string {
+  const filePath = path.join(tmpDir, 'entities.ts')
+  fs.writeFileSync(filePath, content)
+  return filePath
+}
+
+beforeEach(() => {
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'entity-class-declarations-test-'))
+})
+
+afterEach(() => {
+  fs.rmSync(tmpDir, { recursive: true, force: true })
+})
+
+describe('parseEntityClassNames', () => {
+  it('collects exported classes decorated with @Entity()', () => {
+    const filePath = writeSource(`
+      import { Entity, PrimaryKey } from '@mikro-orm/decorators/legacy'
+
+      @Entity({ tableName: 'invoices' })
+      export class Invoice {
+        @PrimaryKey({ type: 'uuid' })
+        id!: string
+      }
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual(['Invoice'])
+  })
+
+  it('looks past comments and stacked decorators', () => {
+    const filePath = writeSource(`
+      import { Entity, Index, Unique } from '@mikro-orm/decorators/legacy'
+
+      @Entity({ tableName: 'invoices' })
+      // The unique constraint bounds the lookup.
+      @Unique({ properties: ['number'] })
+      @Index({
+        name: 'invoices_number_idx',
+        expression: 'create index "invoices_number_idx" on "invoices" ("number")',
+      })
+      export class Invoice {}
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual(['Invoice'])
+  })
+
+  it('ignores exported classes that are not entities', () => {
+    const filePath = writeSource(`
+      export class InvoiceBuilder {}
+
+      @Entity({ tableName: 'invoices' })
+      export class Invoice {}
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual(['Invoice'])
+  })
+
+  it('ignores entity classes that are not exported', () => {
+    const filePath = writeSource(`
+      @Entity({ tableName: 'invoices' })
+      class Invoice {}
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual([])
+  })
+
+  it('handles a local export list', () => {
+    const filePath = writeSource(`
+      @Entity({ tableName: 'invoices' })
+      class Invoice {}
+
+      export { Invoice }
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual(['Invoice'])
+  })
+
+  it('accepts a bare @Entity decorator', () => {
+    const filePath = writeSource(`
+      @Entity
+      export class Invoice {}
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual(['Invoice'])
+  })
+
+  it('returns nothing for an unreadable file instead of throwing', () => {
+    expect(parseEntityClassNames(path.join(tmpDir, 'missing.ts'))).toEqual([])
+  })
+})
--- a/packages/cli/src/lib/generators/__tests__/entity-class-declarations.test.ts
+++ b/packages/cli/src/lib/generators/__tests__/entity-class-declarations.test.ts
@@ -53,6 +53,8 @@ describe('parseEntityClassNames', () => {
 
   it('ignores exported classes that are not entities', () => {
     const filePath = writeSource(`
+      import { Entity } from '@mikro-orm/decorators/legacy'
+
       export class InvoiceBuilder {}
 
       @Entity({ tableName: 'invoices' })
@@ -64,6 +66,8 @@ describe('parseEntityClassNames', () => {
 
   it('ignores entity classes that are not exported', () => {
     const filePath = writeSource(`
+      import { Entity } from '@mikro-orm/decorators/legacy'
+
       @Entity({ tableName: 'invoices' })
       class Invoice {}
     `)
@@ -73,6 +77,8 @@ describe('parseEntityClassNames', () => {
 
   it('handles a local export list', () => {
     const filePath = writeSource(`
+      import { Entity } from '@mikro-orm/decorators/legacy'
+
       @Entity({ tableName: 'invoices' })
       class Invoice {}
 
@@ -84,6 +90,8 @@ describe('parseEntityClassNames', () => {
 
   it('accepts a bare @Entity decorator', () => {
     const filePath = writeSource(`
+      import { Entity } from '@mikro-orm/decorators/legacy'
+
       @Entity
       export class Invoice {}
     `)
@@ -126,6 +134,39 @@ describe('parseEntityClassNames', () => {
     expect(parseEntityClassNames(filePath)).toEqual([])
   })
 
+  it('ignores an Entity decorator imported from an unrelated library', () => {
+    // Any library may export an `Entity` decorator; only MikroORM's collides in the ORM
+    // metadata registry, so only a decorator bound to a @mikro-orm/* import counts.
+    const filePath = writeSource(`
+      import { Entity } from 'some-other-orm'
+
+      @Entity({ tableName: 'invoices' })
+      export class Invoice {}
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual([])
+  })
+
+  it('ignores a namespace-qualified Entity decorator from an unrelated library', () => {
+    const filePath = writeSource(`
+      import * as other from 'some-other-orm'
+
+      @other.Entity({ tableName: 'invoices' })
+      export class Invoice {}
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual([])
+  })
+
+  it('ignores an Entity decorator with no import to bind it', () => {
+    const filePath = writeSource(`
+      @Entity({ tableName: 'invoices' })
+      export class Invoice {}
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual([])
+  })
+
   it('parses a .tsx entity file as JSX rather than dropping declarations', () => {
     // MODULE_CODE_EXTENSIONS accepts .tsx, where `<T>` is JSX, not a type assertion.
     const filePath = writeSource(`
@@ -138,5 +179,109 @@ describe('parseEntityClassNames', () => {
     `, 'entities.tsx')
 
     expect(parseEntityClassNames(filePath)).toEqual(['Invoice'])
+  })
+})
+
+describe('parseEntityClassNames on compiled package output', () => {
+  // A package that ships dist only hands the generator its built entity file, where a
+  // decorated class has become a class expression plus a decorator-helper call. Without
+  // this the standalone generator sees no entities at all and misses every collision.
+  it('reads an esbuild ESM emit', () => {
+    const filePath = writeSource(`
+      var __decorateClass = (decorators, target) => target;
+      import { Entity, PrimaryKey } from "@mikro-orm/decorators/legacy";
+      let ApiKey = class {
+      };
+      __decorateClass([
+        PrimaryKey({ type: "uuid" })
+      ], ApiKey.prototype, "id", 2);
+      ApiKey = __decorateClass([
+        Entity({ tableName: "api_keys" })
+      ], ApiKey);
+      export {
+        ApiKey
+      };
+    `, 'entities.js')
+
+    expect(parseEntityClassNames(filePath)).toEqual(['ApiKey'])
+  })
+
+  it('reads a tsc CommonJS emit', () => {
+    const filePath = writeSource(`
+      "use strict";
+      Object.defineProperty(exports, "__esModule", { value: true });
+      exports.ApiKey = void 0;
+      const legacy_1 = require("@mikro-orm/decorators/legacy");
+      let ApiKey = class ApiKey {
+      };
+      ApiKey = __decorate([
+        (0, legacy_1.Entity)({ tableName: "api_keys" })
+      ], ApiKey);
+      exports.ApiKey = ApiKey;
+    `, 'entities.js')
+
+    expect(parseEntityClassNames(filePath)).toEqual(['ApiKey'])
+  })
+
+  it('reads a bundler export table', () => {
+    const filePath = writeSource(`
+      import { Entity } from "@mikro-orm/core";
+      var entities_exports = {};
+      __export(entities_exports, {
+        ApiKey: () => ApiKey
+      });
+      let ApiKey = class {
+      };
+      ApiKey = __decorateClass([
+        Entity({ tableName: "api_keys" })
+      ], ApiKey);
+    `, 'entities.js')
+
+    expect(parseEntityClassNames(filePath)).toEqual(['ApiKey'])
+  })
+
+  it('ignores a compiled class that is never exported', () => {
+    const filePath = writeSource(`
+      import { Entity } from "@mikro-orm/decorators/legacy";
+      let Internal = class {
+      };
+      Internal = __decorateClass([
+        Entity({ tableName: "internal" })
+      ], Internal);
+    `, 'entities.js')
+
+    expect(parseEntityClassNames(filePath)).toEqual([])
+  })
+
+  it('ignores a compiled class decorated by an unrelated library', () => {
+    const filePath = writeSource(`
+      import { Entity } from "some-other-orm";
+      let Invoice = class {
+      };
+      Invoice = __decorateClass([
+        Entity({ tableName: "invoices" })
+      ], Invoice);
+      export {
+        Invoice
+      };
+    `, 'entities.js')
+
+    expect(parseEntityClassNames(filePath)).toEqual([])
+  })
+
+  it('ignores property decorators, which share the helper call', () => {
+    const filePath = writeSource(`
+      import { Entity, Property } from "@mikro-orm/decorators/legacy";
+      let Helper = class {
+      };
+      __decorateClass([
+        Property({ type: "text" })
+      ], Helper.prototype, "name", 2);
+      export {
+        Helper
+      };
+    `, 'entities.js')
+
+    expect(parseEntityClassNames(filePath)).toEqual([])
   })
 })
--- a/packages/cli/src/lib/generators/__tests__/entity-class-declarations.test.ts
+++ b/packages/cli/src/lib/generators/__tests__/entity-class-declarations.test.ts
@@ -5,8 +5,8 @@ import { parseEntityClassNames } from '../entity-class-declarations'
 
 let tmpDir: string
 
-function writeSource(content: string): string {
-  const filePath = path.join(tmpDir, 'entities.ts')
+function writeSource(content: string, fileName = 'entities.ts'): string {
+  const filePath = path.join(tmpDir, fileName)
   fs.writeFileSync(filePath, content)
   return filePath
 }
@@ -93,5 +93,50 @@ describe('parseEntityClassNames', () => {
 
   it('returns nothing for an unreadable file instead of throwing', () => {
     expect(parseEntityClassNames(path.join(tmpDir, 'missing.ts'))).toEqual([])
+  })
+
+  it('recognises an aliased Entity import', () => {
+    const filePath = writeSource(`
+      import { Entity as OrmEntity } from '@mikro-orm/decorators/legacy'
+
+      @OrmEntity({ tableName: 'invoices' })
+      export class Invoice {}
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual(['Invoice'])
+  })
+
+  it('recognises a namespace-qualified Entity decorator', () => {
+    const filePath = writeSource(`
+      import * as orm from '@mikro-orm/decorators/legacy'
+
+      @orm.Entity({ tableName: 'invoices' })
+      export class Invoice {}
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual(['Invoice'])
+  })
+
+  it('does not treat an unrelated decorator as an entity', () => {
+    const filePath = writeSource(`
+      @Injectable()
+      export class InvoiceService {}
+    `)
+
+    expect(parseEntityClassNames(filePath)).toEqual([])
+  })
+
+  it('parses a .tsx entity file as JSX rather than dropping declarations', () => {
+    // MODULE_CODE_EXTENSIONS accepts .tsx, where `<T>` is JSX, not a type assertion.
+    const filePath = writeSource(`
+      import { Entity } from '@mikro-orm/decorators/legacy'
+
+      export const icon = () => <span>invoice</span>
+
+      @Entity({ tableName: 'invoices' })
+      export class Invoice {}
+    `, 'entities.tsx')
+
+    expect(parseEntityClassNames(filePath)).toEqual(['Invoice'])
   })
 })
--- a/packages/cli/src/lib/generators/__tests__/module-entities.test.ts
+++ b/packages/cli/src/lib/generators/__tests__/module-entities.test.ts
@@ -185,6 +185,33 @@ describe('generateModuleEntities duplicate entity class names', () => {
     expect(warnSpy).not.toHaveBeenCalled()
   })
 
+  it('reads the app override, not the package file, when an app supplies a package module\'s entities', async () => {
+    // The registry imports the app override for such a module, so the check must read the
+    // same file — parsing the package tree would report names that are never registered.
+    touchFile(
+      path.join(tmpDir, 'packages', 'core', 'src', 'modules', 'billing', 'data', 'entities.ts'),
+      entitySource('PackageInvoice', 'billing_invoices'),
+    )
+    touchFile(
+      path.join(tmpDir, 'app', 'src', 'modules', 'billing', 'data', 'entities.ts'),
+      entitySource('Invoice', 'billing_invoices'),
+    )
+    const modules: ModuleEntry[] = [
+      { id: 'billing', from: '@open-mercato/core' },
+      moduleWithEntities('subscriptions', entitySource('Invoice', 'subscription_invoices')),
+    ]
+
+    const result = await generateModuleEntities({ resolver: createMockResolver(tmpDir, modules), quiet: true })
+
+    expect(result.errors).toEqual([])
+    expect(readGenerated(tmpDir)).toContain('from "@/modules/billing/data/entities"')
+    expect(warnSpy).toHaveBeenCalledTimes(1)
+    const message = warnSpy.mock.calls[0][0] as string
+    expect(message).toContain('Invoice')
+    expect(message).toContain(path.join('app', 'src', 'modules', 'billing', 'data', 'entities.ts'))
+    expect(message).not.toContain('PackageInvoice')
+  })
+
   it('stays silent when only one of the colliding classes is an entity', async () => {
     const modules = [
       moduleWithEntities('billing', entitySource('Invoice', 'billing_invoices')),
--- a/packages/cli/src/lib/generators/__tests__/module-entities.test.ts
+++ b/packages/cli/src/lib/generators/__tests__/module-entities.test.ts
@@ -196,3 +196,134 @@ describe('generateModuleEntities duplicate entity class names', () => {
     expect(warnSpy).not.toHaveBeenCalled()
   })
 })
+
+describe('generateModuleEntities duplicate detection in standalone installs', () => {
+  // A standalone install resolves a package module to its compiled dist tree, where a
+  // decorated class is no longer a class declaration. Published packages ship
+  // src/modules too, so parsing prefers that mirror and falls back to the compiled file.
+  const PACKAGE = '@open-mercato/core'
+
+  let warnSpy: jest.SpyInstance
+
+  beforeEach(() => {
+    warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {})
+  })
+
+  afterEach(() => {
+    warnSpy.mockRestore()
+  })
+
+  function packageRoot(): string {
+    return path.join(tmpDir, 'app', 'node_modules', PACKAGE)
+  }
+
+  function createStandaloneResolver(enabled: ModuleEntry[]): PackageResolver {
+    const outputDir = path.join(tmpDir, 'app', '.mercato', 'generated')
+    fs.mkdirSync(outputDir, { recursive: true })
+
+    return {
+      isMonorepo: () => false,
+      getRootDir: () => path.join(tmpDir, 'app'),
+      getAppDir: () => path.join(tmpDir, 'app'),
+      getOutputDir: () => outputDir,
+      getModulesConfigPath: () => path.join(tmpDir, 'app', 'src', 'modules.ts'),
+      discoverPackages: () => [],
+      loadEnabledModules: () => enabled,
+      getModulePaths: (entry: ModuleEntry) => ({
+        appBase: path.join(tmpDir, 'app', 'src', 'modules', entry.id),
+        pkgBase: path.join(packageRoot(), 'dist', 'modules', entry.id),
+      }),
+      getModuleImportBase: (entry: ModuleEntry) => ({
+        appBase: `@/modules/${entry.id}`,
+        pkgBase: `${PACKAGE}/modules/${entry.id}`,
+      }),
+      getPackageOutputDir: () => outputDir,
+      getPackageRoot: () => packageRoot(),
+    }
+  }
+
+  const compiledEntity = (className: string, tableName: string) => `
+import { Entity } from "@mikro-orm/decorators/legacy";
+let ${className} = class {
+};
+${className} = __decorateClass([
+  Entity({ tableName: "${tableName}" })
+], ${className});
+export {
+  ${className}
+};
+`
+
+  const sourceEntity = (className: string, tableName: string) =>
+    `import { Entity } from '@mikro-orm/decorators/legacy'\n\n@Entity({ tableName: '${tableName}' })\nexport class ${className} {}\n`
+
+  function distModule(id: string, className: string, tableName: string): ModuleEntry {
+    touchFile(path.join(packageRoot(), 'dist', 'modules', id, 'data', 'entities.js'), compiledEntity(className, tableName))
+    return { id, from: PACKAGE }
+  }
+
+  function withSourceMirror(id: string, className: string, tableName: string): void {
+    touchFile(path.join(packageRoot(), 'src', 'modules', id, 'data', 'entities.ts'), sourceEntity(className, tableName))
+  }
+
+  it('detects a package-to-package collision through the shipped source mirror', async () => {
+    const modules = [
+      distModule('billing', 'Invoice', 'billing_invoices'),
+      distModule('subscriptions', 'Invoice', 'subscription_invoices'),
+    ]
+    withSourceMirror('billing', 'Invoice', 'billing_invoices')
+    withSourceMirror('subscriptions', 'Invoice', 'subscription_invoices')
+
+    const result = await generateModuleEntities({ resolver: createStandaloneResolver(modules), quiet: true })
+
+    expect(warnSpy).toHaveBeenCalledTimes(1)
+    const message = warnSpy.mock.calls[0][0] as string
+    expect(message).toContain('Invoice')
+    expect(message).toContain(path.join('src', 'modules', 'billing', 'data', 'entities.ts'))
+    expect(result.errors).toEqual([])
+    // The runtime import still points at the compiled package entry.
+    expect(readGenerated(tmpDir)).toContain(`from "${PACKAGE}/modules/billing/data/entities"`)
+  })
+
+  it('detects the collision from compiled output when a package ships dist only', async () => {
+    const modules = [
+      distModule('billing', 'Invoice', 'billing_invoices'),
+      distModule('subscriptions', 'Invoice', 'subscription_invoices'),
+    ]
+
+    await generateModuleEntities({ resolver: createStandaloneResolver(modules), quiet: true })
+
+    expect(warnSpy).toHaveBeenCalledTimes(1)
+    const message = warnSpy.mock.calls[0][0] as string
+    expect(message).toContain('Invoice')
+    expect(message).toContain(path.join('dist', 'modules', 'subscriptions', 'data', 'entities.js'))
+  })
+
+  it('detects a package-to-app collision', async () => {
+    const modules = [
+      distModule('billing', 'Invoice', 'billing_invoices'),
+      { id: 'reporting', from: '@app' } as ModuleEntry,
+    ]
+    withSourceMirror('billing', 'Invoice', 'billing_invoices')
+    touchFile(
+      path.join(tmpDir, 'app', 'src', 'modules', 'reporting', 'data', 'entities.ts'),
+      sourceEntity('Invoice', 'reporting_invoices'),
+    )
+
+    await generateModuleEntities({ resolver: createStandaloneResolver(modules), quiet: true })
+
+    expect(warnSpy).toHaveBeenCalledTimes(1)
+    expect(warnSpy.mock.calls[0][0] as string).toContain('reporting')
+  })
+
+  it('stays silent when compiled package entity class names are unique', async () => {
+    const modules = [
+      distModule('billing', 'Invoice', 'billing_invoices'),
+      distModule('subscriptions', 'Subscription', 'subscriptions'),
+    ]
+
+    await generateModuleEntities({ resolver: createStandaloneResolver(modules), quiet: true })
+
+    expect(warnSpy).not.toHaveBeenCalled()
+  })
+})
--- a/packages/cli/src/lib/generators/__tests__/module-entities.test.ts
+++ b/packages/cli/src/lib/generators/__tests__/module-entities.test.ts
@@ -135,3 +135,64 @@ describe('generateModuleEntities', () => {
     expect(secondStat.mtimeMs).toBe(firstStat.mtimeMs)
   })
 })
+
+describe('generateModuleEntities duplicate entity class names', () => {
+  const entitySource = (className: string, tableName: string) =>
+    `import { Entity } from '@mikro-orm/decorators/legacy'\n\n@Entity({ tableName: '${tableName}' })\nexport class ${className} {}\n`
+
+  let warnSpy: jest.SpyInstance
+
+  beforeEach(() => {
+    warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {})
+  })
+
+  afterEach(() => {
+    warnSpy.mockRestore()
+  })
+
+  function moduleWithEntities(id: string, content: string): ModuleEntry {
+    touchFile(path.join(tmpDir, 'packages', 'core', 'src', 'modules', id, 'data', 'entities.ts'), content)
+    return { id, from: '@open-mercato/core' }
+  }
+
+  it('warns when two modules declare the same entity class name, and still generates', async () => {
+    const modules = [
+      moduleWithEntities('billing', entitySource('Invoice', 'billing_invoices')),
+      moduleWithEntities('subscriptions', entitySource('Invoice', 'subscription_invoices')),
+    ]
+
+    const result = await generateModuleEntities({ resolver: createMockResolver(tmpDir, modules), quiet: true })
+
+    expect(warnSpy).toHaveBeenCalledTimes(1)
+    const message = warnSpy.mock.calls[0][0] as string
+    expect(message).toContain('[Entities Warning]')
+    expect(message).toContain('Invoice')
+    expect(message).toContain('billing')
+    expect(message).toContain('subscriptions')
+    // Warning-only by design: generation still succeeds.
+    expect(result.errors).toEqual([])
+    expect(readGenerated(tmpDir)).toContain('enhanceEntities')
+  })
+
+  it('stays silent when entity class names are unique', async () => {
+    const modules = [
+      moduleWithEntities('billing', entitySource('Invoice', 'billing_invoices')),
+      moduleWithEntities('subscriptions', entitySource('Subscription', 'subscriptions')),
+    ]
+
+    await generateModuleEntities({ resolver: createMockResolver(tmpDir, modules), quiet: true })
+
+    expect(warnSpy).not.toHaveBeenCalled()
+  })
+
+  it('stays silent when only one of the colliding classes is an entity', async () => {
+    const modules = [
+      moduleWithEntities('billing', entitySource('Invoice', 'billing_invoices')),
+      moduleWithEntities('subscriptions', 'export class Invoice {}\n'),
+    ]
+
+    await generateModuleEntities({ resolver: createMockResolver(tmpDir, modules), quiet: true })
+
+    expect(warnSpy).not.toHaveBeenCalled()
+  })
+})
--- a/packages/cli/src/lib/generators/entity-class-declarations.ts
+++ b/packages/cli/src/lib/generators/entity-class-declarations.ts
@@ -5,71 +5,213 @@ import ts from 'typescript-js'
  * Names of the `@Entity()`-decorated classes a module's entity file exports.
  *
  * Only decorated classes count: an entity file may also export plain helper classes,
- * and reporting those as entities would raise false collisions. Parsing is used rather
- * than importing because the caller runs during generation, before the module graph
- * exists.
+ * and reporting those as entities would raise false collisions. Only a decorator bound
+ * to a `@mikro-orm/*` import counts, so an unrelated `Entity` decorator from another
+ * library never registers as one.
+ *
+ * Parsing is used rather than importing because the caller runs during generation,
+ * before the module graph exists. A monorepo checkout and a standalone install that
+ * ships `src/modules` both hand this TypeScript source; a package that ships compiled
+ * output only hands it the built file instead, where a decorated class has become a
+ * class expression plus a decorator-helper call, so both shapes are recognised.
  */
+
+const MIKRO_ORM_MODULE_PREFIX = '@mikro-orm/'
+
+type EntityDecoratorBindings = {
+  /** Local names bound to MikroORM's `Entity` export, including aliased imports. */
+  named: Set<string>
+  /** Local names bound to a whole MikroORM module, for `@orm.Entity()`. */
+  namespaces: Set<string>
+}
 
 function inferScriptKind(filePath: string): ts.ScriptKind {
   if (filePath.endsWith('.tsx')) return ts.ScriptKind.TSX
   if (filePath.endsWith('.jsx')) return ts.ScriptKind.JSX
-  if (filePath.endsWith('.js')) return ts.ScriptKind.JS
+  if (filePath.endsWith('.js') || filePath.endsWith('.cjs') || filePath.endsWith('.mjs')) return ts.ScriptKind.JS
   return ts.ScriptKind.TS
 }
 
-/**
- * Local names bound to MikroORM's `Entity` decorator, so an aliased import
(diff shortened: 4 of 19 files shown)


Frshy:
## Metadata intent — needs a maintainer

This account has read-only permission on this repository, so it cannot apply labels, an assignee, or a reviewer here. Everything intended is listed below in one comment for a maintainer to apply. All names were validated against this repo's label list.

| Label | Rationale |
|---|---|
| `review` | Ready for code review; the PR is open non-draft and in no other pipeline state. |
| `bug` | Fixes a platform defect — a silent entity-resolution corruption that no build step detects. |
| `priority-medium` | Not an active outage, but it costs significant diagnosis time whenever it happens and the failure surfaces far from its cause. |
| `risk-low` | Warning-only and additive. No schema change, no API surface change, no behaviour change to any existing code path; the check cannot fail its host. |
| `skip-qa` | Meets this repo's automated-verification exemption in full — see below. |

**On `skip-qa`** — checked against each condition in the PR template and `AGENTS.md`:

- No manually exercisable UI surface: adds no `.tsx`, nothing under `packages/ui/src/`, nothing under any `components/` tree.
- Database structure unchanged: no entity, migration, or snapshot touched.
- API surface unchanged: `registerOrmEntities()` keeps its existing signature; everything else is new additive exports.
- `BACKWARD_COMPATIBILITY.md` contracts preserved: no contract surface altered — the only behavioural change is an added log line on an already-broken configuration.
- Automated tests for the changed behaviour ship in this PR.

If maintainers read it differently and prefer `needs-qa`, please apply that instead — I'd rather over-route than under-route, and I can't clear a QA gate from this account either way. Note that a `needs-qa` PR stays blocked until a maintainer adds `qa-approved`; an evidence comment alone does not clear it.

**Assignee:** none intended — this is an outside contribution, so assignment is the maintainers' call.
**Reviewer:** whoever owns the ORM / bootstrap and CLI generator areas. Deliberately not naming a handle from outside the project.

No claim label (`in-progress` or similar) is requested: this account could not release one later.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T01:41:31Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
Label set to `review` because this PR is ready for code review.

pkarw (changes requested):
# Code Review: fix(shared): warn on duplicate entity class names across modules

## Summary

The runtime detector is focused and well tested, but the generator path does not cover standalone packages and the static parser can also report unrelated decorators as MikroORM entities. The HMR reporting state and package export surface need tightening before this diagnostic is reliable across supported deployments.

## Verdict

request changes — one major and four medium findings are actionable under this repository's stricter verdict rule.

## Validation Gate

Validation source: GitHub checks plus local fallback (runner: local).

| Command | Status | Notes |
|---|---|---|
| GitHub checks | PASS | `license/cla` succeeded; `[code]smith` was skipped. Branch protection exposes no required-check list and no build/test checks were reported for this head. |
| `yarn build:packages` | BLOCKED | Local fallback hit OS write error `-122` (quota) after `@open-mercato/shared` built successfully. `yarn workspace @open-mercato/cli build` also passed separately after reusing a lockfile-identical dependency tree. |
| `yarn generate` | NOT RUN | Ordered full gate stopped at the environment-level package-build failure. |
| `yarn build:packages` (post-generate) | NOT RUN | Ordered full gate stopped earlier. |
| `yarn i18n:check-sync` | NOT RUN | Ordered full gate stopped earlier. |
| `yarn i18n:check-usage` | NOT RUN | Ordered full gate stopped earlier. |
| `yarn typecheck` | NOT RUN | Ordered full gate stopped earlier. |
| `yarn test` | PASS (targeted) | 5 changed-area suites passed: 51 tests across shared DB and CLI generator coverage. Full workspace run was not reached. |
| `yarn build:app` | NOT RUN | Ordered full gate stopped earlier. |

## Findings

### Major

1. `packages/cli/src/lib/generators/module-entities.ts:105` — standalone generation parses the resolver-selected `dist/modules/**/entities.js`. The build transforms decorated source classes into class expressions plus decorator helper calls, while `parseEntityClassNames()` only recognizes decorated `ClassDeclaration`s. As a result, `yarn generate` silently misses package↔package and package↔app collisions in standalone apps. Consume a shipped/package-generated entity-class-name artifact (preferred by the repository's standalone-generator lesson), or another source-preserving input, while retaining the dist-backed runtime import; add a standalone regression fixture.

### Minor (Medium)

1. `packages/cli/src/lib/generators/entity-class-declarations.ts:24` — decorator recognition is not bound to MikroORM imports: bare `@Entity`, an `Entity` named import from any module, and any `@namespace.Entity()` all count. This can emit false collision warnings for unrelated decorators. Track named and namespace bindings only from supported `@mikro-orm/*` modules and add negative tests.

2. `packages/shared/src/lib/db/mikro.ts:52` — the global reported-name `Set` is monotonic. After HMR registers a clean set, reintroducing the same collision never warns again; changing the contributing modules under the same class name is also suppressed. Store the previous registration's active collision fingerprints and warn on absent→present or changed-source transitions; add collision→clean→collision coverage.

3. `packages/shared/package.json:13` — `@open-mercato/cli` imports the new shared deep path, but the package export map has no explicit `./lib/db/duplicateEntityClassNames` entry. The repository's package-resolution lesson requires explicit entries for new cross-package shared imports. Add `types` and `default` mappings plus resolution coverage.

4. `packages/shared/src/lib/db/mikro.ts:56` — the logger message contains the entire dynamic formatted diagnostic. The structured-logging contract requires a stable message with dynamic values in fields. Use a stable message such as `Duplicate entity class names detected` and put class names/source details in structured fields.

### Nit (Low)

1. `packages/shared/src/lib/db/duplicateEntityClassNames.ts:4` and `packages/shared/src/lib/db/__tests__/duplicateEntities.test.ts:32` — the explanation and characterization test model bare `MetadataStorage.get()`, not real ORM discovery. A DB-free `MikroORM.init({ connect: false })` probe retains constructor-specific metadata for both same-named entities (`invoice_a` and `invoice_b`) while the string lookup selects one. Update the warning/comments to describe the real name-based ambiguity and replace or supplement the characterization with real discovery.

## Breaking Changes

- [x] No function signature changed incompatibly
- [x] No HTTP route, response shape, event ID, CLI flag, database schema, DI key, ACL feature, or generated registry shape changed
- [x] The new surfaces are additive; the missing explicit package export is captured above

## Test Coverage

Core detector and runtime registration behavior are covered. Missing cases are standalone packaged-module discovery, unrelated decorator imports/namespaces, and collision→clean→same-collision HMR reintroduction. The ORM characterization should exercise actual discovery rather than only bare storage calls.

pkarw:
Label set to `changes-requested` because review found actionable issues.

pkarw:
Priority set to `priority-medium` because this is an ordinary platform bug fix, not an outage or release-blocking incident.

pkarw:
Risk set to `risk-high` because the change spans shared runtime bootstrap and generator contract surfaces.

pkarw:
Thanks @Frshy — review found actionable items, so I am carrying the contribution forward with fixes in a replacement PR. You remain credited and assigned for the next recheck.

Frshy:
## Review feedback addressed — `fd7731b`

Thanks for the review, all six findings are fixed on this branch. Given the note about carrying this forward in a replacement PR: the work is here, so please pull from this head rather than the reviewed one.

### Major 1 — standalone generation misses collisions

Two changes, both needed:

- **Parse the shipped source mirror.** A standalone install resolves a package module to `dist/modules/**`, but published packages ship `src/modules` alongside it. Class-name parsing now prefers that mirror — following the same shape as `resolveModuleSourceBase()` in `lib/eject.ts` — while the runtime import stays pointed at `dist`. This is the path the repository's standalone-generator lesson prefers.
- **Parse compiled output as a fallback**, for a package that ships `dist` only. The parser now also recognises the transpiled shape: a class expression bound to a variable plus a decorator-helper call assigned back to it (`X = __decorateClass([...], X)` for esbuild, `__decorate` for tsc, `_ts_decorate` for SWC — the helper name is not matched, the shape is). Export detection covers `export { X }`, `exports.X = X`, and the `__export(ns, { X: () => X })` table.

Standalone regression fixtures added in `module-entities.test.ts`: package↔package through the source mirror, package↔package from compiled output with no mirror present, package↔app, and a silent case. The mirror test also asserts the generated runtime import still resolves to the package entry.

I did not consume `entities.ids.generated.*`: it carries snake_case entity **ids**, not class names, so it could report the collision but not name the classes — and it would leave app modules, which have no such artifact, uncovered.

### Medium 1 — decorator recognition not bound to MikroORM

Recognition is now bound to imports from `@mikro-orm/*` only: named imports (including aliases), namespace imports for `@ns.Entity()`, and the CommonJS `require` forms. A bare `@Entity` with no import, an `Entity` named import from another library, and `@other.Entity()` from another library are all ignored — three negative tests cover them. Verified against the full repo module set: still 272 entity classes across 51 files, unchanged, and `yarn generate` stays silent.

### Medium 2 — monotonic HMR reporting set

Replaced the `Set<className>` with a `Map<className, fingerprint>` holding the **previous registration's** collisions, where the fingerprint is the sorted contributing `moduleId|sourcePath` pairs. A collision warns on absent→present or on a changed fingerprint, and the map is rewritten from the current registration each time. Two tests added: collision → clean → same collision reports twice, and the same class name colliding with a different module reports again.

### Medium 3 — missing export-map entry

Added:

```json
"./lib/db/duplicateEntityClassNames": {
  "types": "./src/lib/db/duplicateEntityClassNames.ts",
  "default": "./dist/lib/db/duplicateEntityClassNames.js"
}
```

with `duplicateEntityClassNamesExport.test.ts` mirroring the existing `principalServiceExport.test.ts`: resolves through an explicit entry rather than a wildcard, both mapped files ship, and the built module imports.

### Medium 4 — dynamic message in the log line

The runtime line is now a constant message with the dynamic values in fields:

```ts
logger.warn('Duplicate entity class names across enabled modules', {
  classNames, duplicates, reason, remediation,
})
```

`reason` and `remediation` are exported module constants, so the message and both explanatory sentences are stable strings. The generator surface keeps the multi-line rendering, which is a `console.warn` and not subject to the facade contract.

### Nit 1 — characterization test models bare `MetadataStorage`

You are right, and the warning text was wrong because of it. I verified against `@mikro-orm/core@7.1.9` with `MikroORM.init({ connect: false })` over two same-named fixtures:

- `metadata.find(InvoiceBilling)` → `duplicate_entity_fixture_billing`
- `metadata.find(InvoiceSubscriptions)` → `duplicate_entity_fixture_subscriptions`
- `metadata.find('Invoice')` → **one of them** (registration order decides)
- `em.getRepository('Invoice')` → the same winner

So discovery keeps a metadata entry per constructor; it is name-based resolution that collapses. The characterization test now runs real discovery and asserts exactly that, including the repository lookup, and the warning text was rewritten — it previously claimed the shadowed class is aliased onto the survivor's metadata, which does not happen.

### Validation

Full ordered gate from `.ai/agentic.config.json`, runner: **local**.

| Command | Status |
|---|---|
| `yarn build:packages` | PASS |
| `yarn generate` | PASS (silent — no collisions, no false positives) |
| `yarn build:packages` | PASS |
| `yarn i18n:check-sync` | PASS |
| `yarn i18n:check-usage` | PASS (advisory only) |
| `yarn typecheck` | PASS |
| `yarn lint` | PASS (0 errors) |
| `yarn test` | PASS |
| `yarn build:app` | PASS |

On `yarn test`: under Turbo's parallel fan-out this machine segfaults a random jest worker on each run — `@open-mercato/core`, then `@open-mercato/app`, then `@open-mercato/cli`, then `@open-mercato/channel-expo`, a different package each time and none of them touched by this change. Run per package, everything passes: `@open-mercato/cli` 92 suites / 1769 tests, `@open-mercato/core` 1377 suites / 10995 tests, `@open-mercato/app` 554 tests. Flagging it as a local environment limit rather than claiming a clean parallel run.

End-to-end re-verified by temporarily introducing a real `ApiKey` collision between `packages/core/src/modules/api_keys` and an app module: `yarn generate` printed the warning naming both modules and files and still generated.

Labels are unchanged — this account still cannot apply them. `changes-requested` presumably wants clearing back to `review` once this is looked at.

pkarw:
🤖 Adopting a stale chain lock (~3 days) — previous run presumed dead.

🤖 `om-auto-review-pr` taking over the lock — re-review of `fd7731b` after the author addressed the 2026-08-22 findings. Started: 2026-08-25T03:21:26Z.

pkarw (changes requested):
# 🔍 Re-review: fix(shared): warn on duplicate entity class names across modules

## 🎯 Summary

This is a re-review of `fd7731b`, which answers all six findings from the 2026-08-22 review. The change adds a diagnostic for the case where two enabled modules export entity classes sharing a JS class name — a condition that today corrupts every name-based MikroORM lookup silently, with no build step catching it. It reports on two surfaces: a `console.warn` from the entity-registry generator (`packages/cli/src/lib/generators/module-entities.ts`) and a structured `logger.warn` from `registerOrmEntities()` (`packages/shared/src/lib/db/mikro.ts`), with a dependency-free detector shared between them so the generator never pulls the ORM in.

All six previous findings are genuinely fixed, and I verified the two I considered hardest to get right rather than taking the description at face value:

- **Major 1 (standalone generation).** The generator now prefers the package's shipped `src/modules` mirror and falls back to parsing compiled output. I ran the new `parseEntityClassNames()` against this repository's *real* build artifacts — `packages/core/dist/modules/{api_keys,customers,sales}/data/entities.js` versus the matching `src` files — and the compiled and source paths agree exactly (1/1, 25/25, 27/27 classes). The `(0, ns.Entity)` unwrapping, the `X = __decorateClass([...], X)` shape and the four-argument property-decorator form that must *not* match are all handled correctly against genuine esbuild output.
- **Nit 1 (characterization test).** `duplicateEntities.test.ts` now runs real discovery through `MikroORM.init({ connect: false })` and pins the actual upstream behaviour: one metadata entry survives per constructor, and it is name-based resolution that collapses. The warning prose was corrected to match — the earlier "the shadowed class is aliased onto the survivor's metadata" claim was wrong and is gone.

I also ran the detector across every `data/entities*.ts` and `db/schema.ts` in the repository (44 files, 234 entity classes in my scope) and it reports zero collisions — so the check is silent on the current module set, with no false positives.

The design holds up well: detection is pure and separate from reporting, both reporting boundaries are wrapped in `try`/`catch` so an exotic export degrades to a skipped check rather than a failed bootstrap, comparison is by constructor identity first so a re-export or an HMR re-registration never warns, and the HMR dedup now stores the *previous* registration's fingerprints in a `Map` so collision → clean → collision reports twice. The runtime log line is a constant message with the dynamic values in fields, matching the structured-logging contract, and the new `@open-mercato/shared` deep import got an explicit export-map entry with a resolution test modelled on the existing `principalServiceExport.test.ts`.

Reviewed scope: all 18 changed files.

## Verdict

❌ request changes — one Minor finding remains: `resolveEntitySourceFile()` can parse the package's source mirror even when generation actually selected an app override, so the generate-time check reads a file that is not the one the registry imports. Under this repository's stricter verdict rule a Minor finding requests changes; there is no blocker and no major here, the fix is a two-line guard, and everything else in the PR is done. The runtime check backstops the affected case, so nothing ships broken either way.

## 🧪 Validation Gate

Validation source: **GitHub checks** (no local fallback — CI has usable check data for `fd7731b`, and the two outstanding gates are pending rather than missing).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job, 3m47s. |
| `yarn generate` | ✅ PASS | `prepare` job — the new duplicate check runs here and did not fail or warn generation. |
| `yarn build:packages` (post-generate) | ✅ PASS | `prepare` job. |
| `yarn i18n:check-sync` | ⏳ PENDING | `test` job still running. |
| `yarn i18n:check-usage` | ⏳ PENDING | `test` job still running. |
| `yarn typecheck` | ⏳ PENDING | `test` job still running (affected-package scoping on PRs). |
| `yarn test` | ⏳ PENDING | `test` job still running; the 15 `ephemeral-integration` shards are also pending. |
| `yarn build:app` | ✅ PASS | `prepare` job. |

Additional green checks worth naming: `lint` (which also runs the raw-`console.*` guard the generator's `console.warn` had to satisfy), `ds-lint`, `audit`, `audit-scope`, `scope`, `documents-multi-instance`, `license/cla`, and — notably — `mutate`, the mutation-testing job scoped to `src/lib/db/duplicateEntities.ts` and `src/lib/db/duplicateEntityClassNames.ts`. A passing mutation run on exactly the new detector is strong evidence the unit tests pin behaviour rather than merely executing it.

**CI is still running.** The `test` job and all 15 `ephemeral-integration` shards were pending when this review was submitted; this verdict does not wait on them and does not pre-judge them. Note that the head is ~103 commits behind `develop`, which is why the full integration shard matrix is running rather than a scoped subset.

## Findings

### 🔹 Minor

1. `packages/cli/src/lib/generators/module-entities.ts:259` — `resolveEntitySourceFile()` decides the parse target from `entry.from` alone, but `found.base` may be the **app** base rather than the package base. The loop above it deliberately prefers an app override (`bases = [appData, pkgData, appDb, pkgDb]`), and `getModulePaths()` computes `appBase` from the module id regardless of `from`. So for a module declared as, say, `{ id: 'catalog', from: '@open-mercato/core' }` where the app supplies `apps/mercato/src/modules/catalog/data/entities.ts`, `found.base` is the app directory, the generated registry imports `@/modules/catalog/data/entities` — and `resolveEntitySourceFile()` still resolves `packages/core/src/modules/catalog/data/entities.ts` and parses that instead. The check then reports class names that are not registered (a false collision pointing at the wrong module and file) and misses any name the override actually introduced. The model it follows, `resolveModuleSourceBase()` in `lib/eject.ts:202`, never hits this because it starts from `getModulePaths().pkgBase` and never from a selected app base. Fix: carry the `fromApp = found.base.startsWith(roots.appBase)` value the loop already computes into the helper and return `resolved` immediately when it is true, then add a generator test for a package-declared module whose entity file is supplied by the app. Impact is bounded — no module in this repository currently shadows a package module id, the surface is diagnostic-only, and the runtime check still reports the collision correctly at bootstrap — but it is a supported configuration reading the wrong file.

### 💅 Nit

1. `.ai/lessons/standalone-generators-must-reuse-package-generated.md` — that lesson states "Do not rely on parsing compiled `dist` entity files for source-level declarations", and this PR adds exactly such a parse as a fallback. I do not think the fallback is wrong: the primary path is the shipped `src/modules` mirror, which is what the companion lesson `standalone-module-discovery-must-treat-published-src.md` asks for, the fallback only reads class names (not the property declarations that lesson was written about), it is covered by tests, and the alternative you rejected — `entities.ids.generated` — genuinely cannot name classes and does not exist for app modules. But the next agent routing through the lessons index will read the two as contradictory. Per the repo's self-improvement convention (root `AGENTS.md` § Workflow Orchestration, item 3), a sentence in that lesson recording class-name-only parsing of compiled output as an acceptable last resort would settle it.

2. `packages/shared/src/lib/db/duplicateEntities.ts:60` — `readModuleIdFromStamp()` reads whatever own `entityName` property is present, and `enhanceEntities()` only stamps `<moduleId>.<ExportName>` when the class does not already declare one. A module that declares a static `entityName` containing a dot would therefore have that value split as a module id in the warning text. Purely cosmetic in a message, and a dot-free value already degrades safely to `undefined`; worth a comment at most.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the change is entirely additive: three new modules, one new export-map entry, no existing export touched.
- [x] No function signature changed in a breaking way — `registerOrmEntities(entities: any[])` keeps its signature and its behaviour; the check runs before the existing body and cannot alter it.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — the generator gains a `console.warn` line, which is human-facing output, not a parsed artifact, and `entities.generated.ts` is byte-identical for a collision-free module set.
- [x] Where a contract had to change: not applicable — nothing in `BACKWARD_COMPATIBILITY.md`'s 13 contract-surface categories is altered. The new `./lib/db/duplicateEntityClassNames` export map entry is additive and, as the new test asserts, resolves through an explicit key rather than the pre-existing wildcard.

## 🧪 Test Coverage

Coverage is thorough and, unusually, verified by the repo's own mutation-testing job on the two new shared modules.

- `duplicateEntityClassNames.test.ts` covers the pure detector and the message formatter: two-module and three-way collisions, multiple colliding names in one pass, identity collapse by shared target and by identical module+path, the fail-open behaviour for unidentifiable entries, entries with no class name, and the formatter's degraded rendering when a module id or a path is missing (including a `sourcePath` that degraded to a bare class name).
- `duplicateEntities.test.ts` pins the upstream MikroORM behaviour through real `connect: false` discovery — class-based lookups resolve to the right table, `find('Invoice')` resolves to exactly one, and `getRepository('Invoice')` hands back the same winner — then covers the adapter (module-id stamps including dotted ids, missing stamps, `EntitySchema` instances, mixed class/schema collisions) and resilience against a throwing `name` getter, a throwing `entityName` getter, an all-traps proxy, and assorted non-entity values, including the case where a real collision must still be found alongside a hostile export.
- `registerOrmEntities.duplicates.test.ts` covers warn-and-still-register, silence for unique names / the same class twice / same-named non-entity exports, and the four reporting-state transitions that Medium 2 asked for: report once for a standing collision, report a newly introduced one, report again after fixed-then-reintroduced, and report again when the contributing modules change under the same class name.
- `entity-class-declarations.test.ts` covers stacked decorators with interleaved comments, bare `@Entity`, aliased imports, namespace-qualified decorators, `export { X }`, non-exported and non-entity classes, `.tsx` script kind, unreadable files, the three negatives Medium 1 asked for (bare `@Entity` with no import, `Entity` from an unrelated library, `@other.Entity()`), and four compiled shapes — esbuild ESM, tsc CommonJS, a bundler export table, a never-exported compiled class, an unrelated-library compiled decorator, and the property-decorator helper call that must not match.
- `module-entities.test.ts` covers monorepo collision/silence plus four standalone fixtures: package↔package through the source mirror (asserting the generated runtime import still points at the package entry), package↔package from compiled output with no mirror present, package↔app, and a silent case.
- `duplicateEntityClassNamesExport.test.ts` mirrors the existing `principalServiceExport.test.ts` precedent exactly.

The one gap corresponds to the Minor finding: there is no generator test for a module declared with a package `from` whose entity file is supplied by an app override, which is the case that currently parses the wrong file. Add it to the `generateModuleEntities duplicate entity class names` describe block — a module `{ id: 'billing', from: '@open-mercato/core' }` with entity files in **both** the package tree and the app tree, asserting the warning names the app file and the app's class names.

No integration coverage is needed here, and I agree with the PR's reasoning: the entire output is a log line at ORM registration and a console warning during generation, neither reachable from a Playwright spec driving the app.

---

*autofix: skipped (not my PR — re-run with `--autofix` to fix it here)*

pkarw:
🏷️ label rationale

| Label | State | Why |
|---|---|---|
| `merge-queue` | **set** (replaces `changes-requested`) | The re-review of `80008310` found no blockers, majors, minors or nits — the one Minor and both Nits from the previous pass are fixed and independently verified — so the code-review gate is passed. |
| `changes-requested` | removed | Superseded by the approval above; pipeline labels are mutually exclusive. |
| `skip-qa` | kept | The change adds no UI surface (no `.tsx`, nothing under `packages/ui/src/` or any `components/` tree), leaves the database structure and API surface unchanged, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships automated tests for the behaviour it changes — the automated-verification exemption applies. Its only output is a generator console warning and an ORM-registration log line, neither manually exercisable. |
| `priority-medium` | kept | An ordinary platform bug fix — a diagnostic for an existing footgun, not an outage, security incident or release blocker. |
| `risk-high` | kept | Still accurate on blast radius: the change sits in shared runtime bootstrap (`registerOrmEntities`) and the entity-registry generator, both cross-cutting contract surfaces. The rating is about reach, not about confidence in this diff — which is high, given a green `mutate` job scoped to the new detector and a revert-and-rerun check confirming the new test catches its regression. |
| `bug`, `partner-request` | kept | Category and provenance labels, unaffected by the pipeline transition. |
| `in-progress` → `ci-monitoring` | swapped | Review work is complete and fully reported; the lock is released and only the CI follow-up remains. `ci-monitoring` is a meta label, not a claim — this PR is free for anyone to act on. |

**`merge-queue` is not permission to merge yet.** The `test` job and all 15 `ephemeral-integration` shards were still running when the review was submitted; branch protection, not this label, gates the merge. The branch is also 185 commits behind `develop` — merging the base in first would shorten CI and re-test against current `develop`.

pkarw:
Thanks @Frshy — handing this back to you. All six findings from the 2026-08-22 review are genuinely fixed, and I verified the two hardest ones against real artifacts rather than taking the description at face value: the compiled-output parser agrees exactly with the source parser on this repository's own `packages/core/dist/modules/**/entities.js` builds, and the rewritten characterization test pins the real discovery behaviour.

One Minor finding remains, and it is a two-line guard: `resolveEntitySourceFile()` in `packages/cli/src/lib/generators/module-entities.ts:259` decides the parse target from `entry.from` alone, but the loop above it prefers an app override, so a module declared with a package `from` whose entity file is supplied by the app gets parsed from the package mirror while the registry imports the app file. Carrying the `fromApp` value the loop already computes into the helper fixes it; a generator test for that configuration would close the matching coverage gap.

Also flagged: `.ai/lessons/standalone-generators-must-reuse-package-generated.md` reads as contradicting the compiled-output fallback you added. I do not think the fallback is wrong — a sentence in that lesson recording it as an acceptable last resort would keep the next agent from re-litigating it.

Labels are handled from this side: `skip-qa` is now applied (your exemption argument checks out in full), and `changes-requested` stays until the above lands.

pkarw:

Re-review of `fd7731b`: all six previous findings verified fixed; one Minor finding remains (generator parses the package source mirror when generation selected an app override) plus two nits. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

`ci-monitoring` is set because the `test` job and the 15 `ephemeral-integration` shards were still running when the verdict was submitted; a CI-result comment will follow, and the label is removed either way. It is a meta label and not a claim — this PR is free for anyone to act on.

pkarw:

Every check on `fd7731b` has settled **green**. This closes out the four gate commands that were still pending when the review was submitted.

| Command | Status | Check |
|---|---|---|
| `yarn build:packages` | ✅ PASS | `prepare` (3m47s) |
| `yarn generate` | ✅ PASS | `prepare` |
| `yarn build:packages` (post-generate) | ✅ PASS | `prepare` |
| `yarn i18n:check-sync` | ✅ PASS | `test` (8m31s) |
| `yarn i18n:check-usage` | ✅ PASS | `test` |
| `yarn typecheck` | ✅ PASS | `test` |
| `yarn test` | ✅ PASS | `test` |
| `yarn build:app` | ✅ PASS | `prepare` |

Also green: all 15 `ephemeral-integration` shards (20m51s–23m55s), `lint`, `ds-lint`, `audit`, `audit-scope`, `scope`, `merge-coverage`, `documents-multi-instance`, `license/cla`, and `mutate` scoped to the two new shared modules. `[code]smith`, `comment` and `docker-build` were skipped by their own conditions.

**The verdict is unchanged: `changes-requested`.** It was never driven by CI — it rests solely on the Minor finding at `packages/cli/src/lib/generators/module-entities.ts:259`, so a fully green run does not clear it. The pipeline label stays as it is.

`ci-monitoring` is removed; this follow-up is complete and no further CI watching will come from this agent.

Frshy:
## Re-review feedback addressed — `80008310b`

Thanks for the re-review — the Minor finding and both nits are fixed on this branch.

### Minor 1 — generator parsed the package file when generation selected an app override

`packages/cli/src/lib/generators/module-entities.ts` — `resolveEntitySourceFile()` now takes the `fromApp` value the base-selection loop already computes and returns the resolved path immediately when it is true, so the check reads the same file the generated registry imports. Exactly the two-line guard you described.

Test added to the `generateModuleEntities duplicate entity class names` block: a module `{ id: 'billing', from: '@open-mercato/core' }` with entity files in **both** trees (`PackageInvoice` in the package tree, `Invoice` in `apps/.../billing/data/entities.ts`) colliding with a package module `subscriptions` that declares `Invoice`. It asserts the generated registry imports `@/modules/billing/data/entities`, that the warning names the **app** file and `Invoice`, and that it never mentions `PackageInvoice`. I confirmed it pins the behaviour rather than merely covering it: with the guard reverted the test fails with 0 warnings (the collision is missed entirely, since the package file contributes a name that is never registered).

### Nit 1 — lesson contradiction

`.ai/lessons/standalone-generators-must-reuse-package-generated.md` now records that parsing compiled output for *class names alone* is an acceptable last resort when no generated artifact carries them, and points at this check as the case: source mirror first, `dist` only for that one fact.

### Nit 2 — declared `entityName` split as a module id

Documented at `packages/shared/src/lib/db/duplicateEntities.ts:29`: `enhanceEntities()` never overwrites a declared `entityName`, so a declared value containing a dot yields a bogus module id in the warning text, and a dot-free one degrades to `undefined`. Comment only, as suggested — no behaviour change.

### Validation

`yarn workspace @open-mercato/cli test src/lib/generators/__tests__/module-entities.test.ts` (12/12), `yarn workspace @open-mercato/shared test src/lib/db` (81/81, 9 suites), `yarn workspace @open-mercato/cli build`, and typecheck on both touched workspaces all pass locally. Leaving the full gate to CI.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-28T07:25:57Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Re-review: fix(shared): warn on duplicate entity class names across modules

## 🎯 Summary

This is a re-review of `80008310`, which answers the single Minor finding and both Nits from the 2026-08-25 re-review of `fd7731b`. Nothing else in the PR changed, so this pass concentrates on the delta and on confirming that the previously-verified work did not regress.

The delta is four files and 37 added lines:

- **`packages/cli/src/lib/generators/module-entities.ts`** — `resolveEntitySourceFile()` now takes the `fromApp` value the base-selection loop already computes at line 91 and returns the resolved path immediately when it is true, so an app override wins over the package `src/modules` mirror entirely.
- **`packages/cli/src/lib/generators/__tests__/module-entities.test.ts`** — the generator test the finding asked for: a module declared `{ id: 'billing', from: '@open-mercato/core' }` with entity files in *both* the package tree and the app tree.
- **`packages/shared/src/lib/db/duplicateEntities.ts`** — comment-only, recording that a class declaring its own `entityName` keeps it (Nit 2).
- **`.ai/lessons/standalone-generators-must-reuse-package-generated.md`** — one sentence recording class-name-only parsing of compiled output as an acceptable last resort (Nit 1).

I verified the fix rather than reading it. Both halves hold up:

- **The fix is correct and complete.** `fromApp = found.base.startsWith(roots.appBase)` is computed by the same loop that selects the base, and `bases = [appData, pkgData, appDb, pkgDb]` means an app override is only ever selected as `found.base` when it genuinely exists — so the flag is exactly the "the registry imports the app file" condition, not a proxy for it. The guard is placed before the `entry.from` branch, which is the only path that could have redirected the parse into the package tree.
- **The new test genuinely pins it.** I reverted the one-line guard back to `if (!from || from === '@app')` and re-ran the suite: exactly one test failed — the new one — at `expect(warnSpy).toHaveBeenCalledTimes(1)` with 0 calls received, i.e. the pre-fix code parsed the package file, found only `PackageInvoice`, and saw no collision at all. Restoring the guard returned 12/12. The test also pins the two directional facts that matter, not just the count: `expect(message).toContain(path.join('app', …, 'entities.ts'))` and `expect(message).not.toContain('PackageInvoice')`, plus `expect(readGenerated(tmpDir)).toContain('from "@/modules/billing/data/entities"')` to anchor that the registry really does import the override. That is a test that fails for the right reason, not one that merely executes the line.

**Regression risk for this repository is nil.** The changed branch is reachable only for a module declared against a package whose entity file the app supplies. The app modules here are `example`, `example_customers_sync` and `ratelimit_probe`, and none shadows a package module id — so `fromApp` is false for every package-declared module and the resolver behaves exactly as it did before. The fix corrects a supported configuration that this repo does not currently exercise.

Everything I verified in the previous pass still stands: detection is pure and separated from reporting, both reporting boundaries are `try`/`catch`-guarded so an exotic export degrades to a skipped check rather than a failed bootstrap, comparison is by constructor identity first so re-exports and HMR re-registration never warn, the runtime log line keeps a constant message with dynamic values in fields, and the new deep import has an explicit export-map entry with a resolution test.

Reviewed scope: all 20 changed files, with the delta commit reviewed line by line.

## Verdict

✅ **approve** — no blockers, no majors, no minors, no nits. Every finding from both previous reviews is fixed and independently verified; this repository's stricter rule (Medium/minor also requests changes) is satisfied because nothing at or above nit level remains.

Two caveats that are **not** findings against this PR and do not gate the approval, but that a maintainer should see before merging:

1. **CI has not finished** — see the disclosure below. This approval is a code-review verdict; it does not stand in for the merge gate.
2. **The branch is 185 commits behind `develop`**, which is why the full 15-shard `ephemeral-integration` matrix is running rather than an affected-only subset. Merging the base in before merge would both shorten CI and re-test the change against current `develop`.

## ⏳ CI status disclosure

**CI is still running on `80008310`, and this review does not wait for it.** At submission time the `test` job and all 15 `ephemeral-integration` shards were `IN_PROGRESS`; **no check is failing**. The verdict above is a code-review verdict and explicitly does not pre-judge the pending checks — a red result on any of them would override it, and branch protection, not this review, gates the merge. I will post a follow-up comment once CI settles.

## 🧪 Validation Gate

**Validation source: GitHub checks**, per this repository's checks-first review rule, with a narrow local fallback used only for the one thing CI cannot answer — whether the new regression test actually catches the regression.

Runner for the local fallback: **local** (no compose `app` container running).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job. |
| `yarn generate` | ✅ PASS | `prepare` job — the duplicate check runs here and neither failed nor warned. |
| `yarn build:packages` (post-generate) | ✅ PASS | `prepare` job. |
| `yarn i18n:check-sync` | ⏳ PENDING | `test` job still running. |
| `yarn i18n:check-usage` | ⏳ PENDING | `test` job still running. |
| `yarn typecheck` | ⏳ PENDING | `test` job still running. |
| `yarn test` | ⏳ PENDING | `test` job still running; the 15 `ephemeral-integration` shards are also pending. |
| `yarn build:app` | ✅ PASS | `prepare` job. |

Other green checks: `lint` (which also runs the raw-`console.*` guard the generator's `console.warn` must satisfy), `ds-lint`, `audit`, `audit-scope`, `scope`, `documents-multi-instance`, `license/cla` — now passing, the CLA box in the description is the only one still unticked — and `mutate`, the mutation-testing job scoped to `src/lib/db/duplicateEntities.ts` and `src/lib/db/duplicateEntityClassNames.ts`. A green mutation run on exactly the new detector is unusually strong evidence that its unit tests pin behaviour rather than merely covering lines.

### Local fallback actually executed

- `packages/cli` generator suite: **1279/1281 passed**. The 2 failures are in `module-facts.bc-guard.test.ts`, a file this PR does not touch. I confirmed they are **pre-existing and environmental, not caused by this PR**, by checking out `origin/develop` in the same worktree with the identical harness and re-running: the same 2 tests fail identically there. They are an artifact of the review worktree resolving `node_modules/@open-mercato/*` to another checkout's package sources.
- `packages/shared` `src/lib/db` suite: **81/81 passed, 9/9 suites** (after building `packages/shared` so `duplicateEntityClassNamesExport.test.ts` could resolve its `dist` target — without that build it fails purely on the missing artifact, which is why the ordered gate builds first).
- The revert-and-rerun check on the new generator test, described in the summary.

## Findings

None. All three items from the previous re-review are resolved:

| Previous finding | Status |
|---|---|
| 🔹 Minor 1 — `resolveEntitySourceFile()` parsed the package mirror even when an app override was selected | ✅ Fixed in `module-entities.ts:263`, with the generator test the finding asked for; verified by revert-and-rerun |
| 💅 Nit 1 — the standalone-generators lesson contradicted the compiled-output fallback | ✅ Fixed — the lesson now records class-name-only parsing of compiled output as an acceptable last resort and names this check |
| 💅 Nit 2 — a declared dotted `entityName` yields a bogus module id in the warning text | ✅ Documented at `duplicateEntities.ts:24-30`, which is the proportionate response for a cosmetic edge case in a message |

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — entirely additive; the delta adds one parameter to a **private, module-local** function (`resolveEntitySourceFile` is not exported).
- [x] No function signature changed in a breaking way — `registerOrmEntities(entities: any[])` keeps its signature and behaviour.
- [x] No HTTP route removed, renamed, or re-methoded.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — the generator gains a human-facing `console.warn`, and `entities.generated.ts` is byte-identical for a collision-free module set.
- [x] Nothing in `BACKWARD_COMPATIBILITY.md`'s 13 contract-surface categories is altered. The `./lib/db/duplicateEntityClassNames` export-map entry is additive and resolves through an explicit key rather than the pre-existing wildcard, as its own test asserts.

## 🧪 Test Coverage

Coverage was already thorough and is now complete — the one gap the previous review named is closed.

The new case, `reads the app override, not the package file, when an app supplies a package module's entities`, is exactly the missing scenario: both trees populated, the package class deliberately given a distinct name (`PackageInvoice`) so that reading the wrong file produces *no* collision and the assertion cannot pass by accident. It asserts the warning names the app path and the app's class, that it does *not* name the package class, and that the generated registry imports the app module — four independent anchors on one behaviour.

The rest of the suite is unchanged from the previous review and re-ran clean: the pure detector and message formatter (two-module, three-way, multiple names, identity collapse, fail-open, degraded rendering), the upstream-MikroORM characterization test through real `connect: false` discovery, adapter and hostile-export resilience (throwing `name`/`entityName` getters, an all-traps proxy), the four reporting-state transitions on `registerOrmEntities`, the AST parser's decorator/export/compiled-output shapes including the property-decorator form that must not match, and the export-map resolution test.

No integration coverage is required, and I continue to agree with the PR's reasoning: the entire output is a log line at ORM registration and a console warning during generation, neither reachable from a Playwright spec driving the app.

---

*autofix: skipped (not my PR — re-run with `--autofix` to fix it here)*

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED. Lock released (`in-progress` → `ci-monitoring`).

Re-review of `80008310`: the Minor finding and both Nits from the `fd7731b` pass are fixed, and **no findings remain at any severity** — so under this repository's stricter rule (Medium/minor also requests changes) the PR is approved and moved to `merge-queue`.

The fix was verified rather than read: reverting the one-line `fromApp` guard makes exactly the new generator test fail, at `expect(warnSpy).toHaveBeenCalledTimes(1)` with 0 calls — the pre-fix code parsed the package file and saw no collision at all. Restoring it returns 12/12. Locally, `packages/shared` `src/lib/db` is 81/81 and the `packages/cli` generator suite is 1279/1281, with both failures confirmed pre-existing on `origin/develop` under the identical harness and unrelated to this PR.

**CI had not finished when this verdict was posted** — the `test` job and all 15 `ephemeral-integration` shards were still running, with nothing failing. The approval is a code-review verdict and does not gate the merge; branch protection does. I am holding `ci-monitoring` and will post the CI result as a follow-up comment.

Two notes for whoever merges: the branch is **185 commits behind `develop`** (which is why the full shard matrix is running rather than an affected-only subset — merging the base in first would shorten CI and re-test against current `develop`), and the **CLA checkbox in the description is still unticked** even though the `license/cla` check is green.

pkarw:

**Every check on `80008310` has settled green: 26 success, 3 skipped, 0 failing, 0 pending.** This closes out the four gate commands that were still pending when the approval was posted, so the `merge-queue` label now rests on a complete signal.

I read the individual step conclusions rather than trusting the aggregate `test` check, because several gates in this repo are `continue-on-error` and a green job badge does not by itself prove they ran:

| Gate command | Step | Result |
|---|---|---|
| `yarn build:packages` | `prepare` → Build packages | ✅ |
| `yarn generate` | `prepare` → Prepare generated modules | ✅ |
| `yarn build:packages` (post-generate) | `prepare` → Rebuild packages with generated files | ✅ |
| `yarn i18n:check-sync` | `test` → Check i18n sync | ✅ |
| `yarn i18n:check-usage` | `test` → Check i18n usage | ✅ |
| `yarn typecheck` | `test` → Checking types | ✅ |
| `yarn test` | `test` → Test | ✅ |
| `yarn build:app` | `prepare` → Build app | ✅ |

`Check agent instruction budget` also passed, so nothing behind it was skipped. All 15 `ephemeral-integration` shards are green, as are `lint`, `ds-lint`, `audit`, `audit-scope`, `scope`, `merge-coverage`, `documents-multi-instance`, `license/cla`, and `mutate` — the mutation-testing job scoped to the two new detector modules.

The PR is now `MERGEABLE` / `CLEAN` / `APPROVED`. Removing `ci-monitoring`; no further follow-up is owed from this agent.

**Verdict unchanged: approved.** Two things for whoever merges, neither a blocker: the branch is **185 commits behind `develop`** (merging the base in first would re-test against current `develop` — CI here passed against the branch as pushed), and the **CLA checkbox in the description is still unticked** even though the `license/cla` check is green — that box is deliberately left for a human.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-28T08:16:37Z — targeted re-verification of two dimensions on `80008310` (runtime/generate-time performance impact, and behaviour under `create-app` standalone mode). Other auto-skills will skip this PR until the lock is released.

pkarw:
## 🔬 Targeted verification — performance impact and standalone (`create-mercato-app`) mode

Re-checked `80008310` on two axes a maintainer asked about specifically. **No findings; the approval stands.** Everything below was measured or executed against this head, not inferred from the description.

### 📈 Performance

**Runtime (`registerOrmEntities`) — negligible, and it runs once per process.**

`createBootstrap()` guards on `_bootstrappedKeys` and returns early unless `NODE_ENV === 'development'` (`packages/shared/src/lib/bootstrap/factory.ts:42`), so production pays this exactly once at boot. `warnOnDuplicateEntityClassNames()` is a single O(n) pass building one `Map` of `Map`s; every per-entry operation (`EntitySchema.is`, an own-property check for `MetadataStorage.PATH_SYMBOL`, two string reads) is O(1).

Measured with a synthesized registry matching the real shape (entity classes carrying `PATH_SYMBOL` plus the plain function exports that travel with them):

| Registry size | Median cost of `findDuplicateRegisteredEntityClassNames` |
|---|---|
| 300 exports (250 entities) | **0.085 ms** |
| 600 exports (500 entities) | 0.081 ms |
| 2400 exports (2000 entities) | 0.611 ms |

This repository currently registers 234 entity classes across 44 module entity files, so the real figure is the first row. In development the check re-runs per HMR bootstrap — still ~0.08 ms, and the `globalThis` fingerprint map means only the *log write* is deduplicated, not the scan, which is the right trade at this cost.

**Generate-time (`parseEntityClassNames`) — one extra AST pass over files the generator already parses.**

Measured over all 44 entity files (457 KiB of source) in this checkout: **median 30.6 ms warm** (min 28.8, max 45.1) for the whole set. The slowest single file is `core/sales` at 5.7 ms (70 KiB). Cold-start adds ~60 ms of `typescript-js` warmup, which `packages/cli/src/lib/resolver.ts` already pays before this code runs.

For scale: `packages/cli/src/lib/generators/entity-ids.ts` already calls `ts.createSourceFile()` on these same module entity files in `parseExportedClassNamesFromFile()` (:31) and `parseEntityFieldsFromFile()` (:179). This PR adds one more pass of the same kind and cost — a marginal addition to an already-accepted per-generate expense, not a new class of work. `parseEntityClassNames` uses the raw `ts.createSourceFile` API rather than `ts-morph`, so it is materially cheaper than the AST machinery the same generator uses to *write* its output.

**Verdict: not a performance concern on either surface.**

### 📦 Standalone mode (`create-mercato-app`)

Five things had to hold. All five do.

**1. The `src/modules` mirror `resolveEntitySourceFile()` prefers is actually published.** `npm pack --dry-run` on `packages/core` yields 7944 files including both `src/modules/api_keys/data/entities.ts` and `dist/modules/api_keys/data/entities.js`. `packages/core/.npmignore` excludes only tests, so the shipped source mirror the function targets is present in a real install. Standalone therefore parses TypeScript source, exactly as the code comment claims.

**2. The compiled-output fallback is faithful where it is used.** The `!entry.from` branch falls back to the resolved `dist` file. I built all 42 package entity files fresh with the repository's own esbuild settings (`format: esm`, `platform: node`, `target: node18`, `bundle: false`) and compared the compiled parse against the source parse file by file: **42 files, 0 mismatches.** `collectCompiledDecoratedClassNames()` correctly recognises esbuild's `let X = class {}` + `X = __decorateClass([Entity(...)], X)` + `export { X }` shape, and `collectEntityDecoratorBindings()` correctly binds through this repo's real import (`@mikro-orm/decorators/legacy`, which satisfies the `@mikro-orm/` prefix test).

> Worth recording for the next reviewer: a naive run of that comparison against the `dist/` trees sitting in a working checkout reports 6 mismatches in `ai_assistant`, `payment_gateways`, `shipping_carriers`, `staff`, `workflows` and `webhooks`. Those are **stale build artifacts**, not parser bugs — e.g. `packages/core/dist/modules/payment_gateways/data/entities.js` is dated 2026-05-28 against a source dated 2026-08-05, and the "missing" classes do not exist anywhere in that compiled file. Rebuilding removes every mismatch.

**3. The fallback branch is barely reachable anyway.** Every entry in `packages/create-app/template/src/modules.ts` declares an explicit `from`, so a scaffolded standalone app takes the `src/modules` mirror path for all package-backed modules and the `fromApp` path for `@app` ones.

**4. The new subpath export resolves in a published install.** `packages/shared/package.json` gains `./lib/db/duplicateEntityClassNames` in the same `{types: src/*.ts, default: dist/*.js}` shape as its siblings; `packages/shared/build.mjs` → `buildPackage` globs `src/**/*.{ts,tsx}`, so `dist/lib/db/duplicateEntityClassNames.js` is emitted, and `npm pack` ships both `src/` and `dist/`. On the consumer side, `packages/cli/build.mjs` uses `entryPoints: 'src/**/*.ts'` (so `entity-class-declarations.ts` is built) and `scripts/lib/add-js-extension.mjs` rewrites **only** specifiers matching `/^\./` — bare specifiers are untouched, as `packages/cli/dist/bin.js` already demonstrates with `@open-mercato/shared/lib/bootstrap/dynamicLoader`. `@open-mercato/shared` is already a direct dependency of `@open-mercato/cli`, and `typescript-js` was already one too, so no new dependency reaches a standalone app.

**5. The runtime diagnostic is fully informative in the standalone shape.** I imported a real *compiled* entity (`packages/core/dist/modules/api_keys/data/entities.js`), applied the generated registry's own `enhanceEntities()` stamping logic to it, and ran `collectEntityClassNameEntries()`:

```
[{ "className": "ApiKey", "moduleId": "api_keys",
   "sourcePath": ".../packages/core/dist/modules/api_keys/data/entities.js" }]
```

Class name, module id and a real path all survive against `dist`-shaped input — so a collision reported at bootstrap in a standalone app names its contributors, not just the class.

### 🧾 Current state

No duplicate entity class names exist across the packages scanned today (0 collisions over 234 classes), so on this repository both surfaces are silent — which is the intended steady state for a diagnostic. CI on `80008310` is **26 pass / 3 skipped / 0 failing**. Labels and the `merge-queue` pipeline state are unchanged.

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED (unchanged). Lock released.

Targeted re-verification of `80008310` on performance and standalone mode found nothing — evidence in the comment above. No new commits since the 2026-08-28 approval, so the review verdict and the `merge-queue` label stand. Autofix: skipped (not my PR — re-run with --autofix to fix it here).

PR #5453, opened 2026-08-21
fix(security): reject MFA-pending tokens from general staff APIs

### LINKED ISSUE #5212: Fix: reject MFA-pending tokens from general staff APIs
## Summary

The JWT returned after a successful password check for an MFA-enrolled staff account is accepted as a fully authenticated staff session by normal APIs before the second factor is completed. Anyone with the account password can use that `mfa_pending` token for its approximately ten-minute lifetime and exercise the account's restored roles and features without an OTP or recovery code.

**Security classification:** priority extreme, risk high. This is a deterministic MFA bypass for every MFA-protected staff account, including privileged administrators, with only the first factor required.

## Analysis

The core login flow creates a real database session and signs a full staff JWT before enterprise MFA interception:

- `packages/core/src/modules/auth/api/login.ts:151-165`

The enterprise interceptor replaces the response token with another valid staff JWT carrying the same `sub` and live `sid`, plus advisory `mfa_pending: true` / `mfa_verified: false` claims:

- `packages/enterprise/src/modules/security/api/interceptors.ts:36-49,58-96`

Canonical session integrity then validates the live session and user and restores current roles, but never rejects or masks an MFA-pending assurance state:

- `packages/core/src/modules/auth/lib/sessionIntegrity.ts:47-65,78-106,109-150`
- `packages/shared/src/lib/auth/server.ts:326-356`

Repository-wide handling of `mfa_pending` is confined to the MFA completion/recovery endpoints and UI. General `requireAuth` / `requireFeatures` paths do not distinguish the provisional token from a completed staff login.

The enterprise login-interceptor and MFA specifications describe the intended challenge flow but do not establish a central assurance-state boundary:

- `.ai/specs/enterprise/implemented/SPEC-ENT-007-2026-03-06-auth-login-interceptors-extension.md`
- `.ai/specs/enterprise/implemented/SPEC-ENT-001-2026-02-17-security-module-enterprise-mfa.md`

## How to implement

- Treat MFA-pending credentials as a distinct, restricted authentication state or token audience.
- Enforce that state centrally so only the challenge preparation, verification, and recovery endpoints can accept it.
- Do not restore normal roles/features for an MFA-pending request.
- Prefer creating or activating the general staff session only after MFA succeeds; otherwise ensure the provisional session cannot authenticate normal APIs and is revoked on challenge failure/expiry.
- Add regression tests proving that a pending token is denied by representative read and mutation APIs, remains accepted by the MFA completion endpoints, and that the verified replacement token gains normal access.
- Cover enforced MFA, user-enrolled MFA, recovery codes, expiry, and failed challenge cleanup.

## Compatibility notes

Keep existing login and MFA endpoint paths and successful response contracts where possible. The security fix intentionally narrows what the provisional token can access. If token claims or lifecycle semantics are externally observable, document the change in the enterprise upgrade notes.

## How to pick up

1. Add a failing integration test that logs in with the correct password for an MFA-enrolled user, captures the pending token, and calls a feature-protected API before OTP verification.
2. Put the restriction in the canonical staff-auth/session-integrity path, with a narrow allowlist for MFA completion routes.
3. Verify that normal non-MFA login and post-verification login remain unchanged.
4. Run focused auth and enterprise security tests, followed by the configured validation gate.

## Out of scope

- Passkey fallback behavior tracked separately in #3852.
- Availability/fail-open behavior when the MFA service itself throws; that is a separate operational hardening concern.
- Changing MFA enrollment policy or supported second-factor methods.

Found during an authorized defensive security assessment of commit `a8d208fcdc43ce565e8ab78ec593873ab4ff53b5`.

haxiorz:
Closes #5212

## Goal

An MFA-enrolled staff account protected only by a password is no longer usable against general staff APIs before the second factor completes.

## Problem and root cause

The enterprise login interceptor replaces a successful password-login token with a 10-minute staff JWT carrying the live session id plus `mfa_pending: true` / `mfa_verified: false`. The canonical staff auth resolvers verified the signature but never inspected those claims, so the provisional token authenticated every API with the account's full DB-restored roles — a deterministic MFA bypass, including for super-admins.

## What changed

- Central enforcement in `resolveAuthFromRequestDetailed` / `resolveAuthFromCookiesDetailed` (`packages/shared/src/lib/auth/server.ts`): pending tokens resolve to `invalid` (401 + cleared cookies) except on registered MFA completion routes; pages always reject them.
- New fail-closed `mfaPendingAccess` allowlist (default: POST `/api/security/mfa/prepare`, `/verify`, `/recovery`) with additive registration for third-party MFA implementations.
- New `isMfaPendingJwtPayload` predicate in shared auth/jwt.
- **Alternate credential paths closed** (review round 4): `GET|POST /api/auth/session/refresh` is `requireAuth: false` and mints a staff JWT straight from the `session_token` cookie, never touching the resolvers above. Two closures:
  - `api/auth/login.ts` clears `session_token` whenever an interceptor replaces the issued token, so a provisional login leaves no usable refresh cookie from an earlier pre-enrollment remember-me login.
  - `api/auth/session/refresh.ts` refuses to mint (401, or redirect to `/login`, with staff cookies cleared) when the caller presents an MFA-pending `auth_token` cookie or bearer token.
- MFA prepare/verify/recovery contracts unchanged (they already required `mfa_pending`); verified replacement tokens and non-MFA logins behave exactly as before.

## Validation

- `yarn build:packages` / `yarn generate` / `yarn typecheck` / `yarn test` / `yarn build:app` / `yarn i18n:check-sync` / `yarn i18n:check-usage` — all green on the final diff
- Regression tests observed failing for the diagnosed reason before the fix (pending token authenticated general APIs) and passing after; suites: shared auth 139/139, core auth 720/720, enterprise security 305/305
- The two round-4 closures were likewise observed failing before the fix (`refreshFromSessionToken` called and a staff JWT minted; `session_token` left intact) and passing after
- Limitation: no Playwright E2E for the browser challenge flow; resolver-level plus HTTP-dispatcher coverage only

## Automated review

- `om-code-review` — four rounds. Rounds 1–3 findings (package-boundary blocker, missing end-to-end regression, `TC-SEC-010` id collision) resolved and independently verified; round 4's High finding (`/api/auth/session/refresh` upgrading an MFA-pending browser) and both Mediums fixed in this branch.
- Unresolved findings — None. Residual, tracked out of scope: `AuthService.refreshFromSessionToken` is still MFA-unaware, and the SSO callback plants `session_token` without passing through the login interceptor, so SSO logins never enter the pending state at all.

## Breaking changes

None to contracts (no route, schema, event, or response-shape changes). Three security-narrowing reachability changes are recorded in the spec amendment's Migration & Backward Compatibility table: login's `Set-Cookie` behavior when an interceptor replaces the token, session refresh for MFA-pending callers, and `POST /api/auth/logout` (rejected at the dispatcher during the pending window; the abandoned `Session` rows it leaves are inert). Third-party MFA implementations with their own completion endpoints register them via `registerMfaPendingAccessRoutes`.

## Labels and delivery

- `security` — closes a deterministic authentication bypass
- `needs-qa` — auth-critical behavior change; manual verification of non-MFA login, full MFA challenge flow, and recovery flow is warranted despite passing automation
- `priority-extreme` — matches the issue classification; password-only privilege escalation
- `risk-high` — touches the central staff-auth resolution path used by every API

## Manual QA

1. Non-MFA login → backend loads normally.
2. Login as MFA-enrolled user → challenge appears; wrong OTP rejected; correct OTP lands in backend with normal access.
3. With a challenge pending, call any staff API with the pending token → 401, cookies cleared; `/api/security/mfa/verify` still accepts it.
4. Remember-me login before MFA enrollment, then enroll and log in again → the challenge response clears `session_token`, and navigating to `/backend` mid-challenge lands on `/login`, never in the admin.
5. Recovery-code login completes and grants normal access.


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T01:24:04Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(security): reject MFA-pending tokens from general staff APIs

## 🎯 Summary
This change correctly moves the MFA assurance-state check into the canonical staff-auth resolvers, rejects provisional credentials before canonical roles are restored, and keeps a narrow method-and-path allowlist for the three MFA completion operations. The focused unit tests are strong, the changed auth helpers passed mutation testing, and the complete GitHub CI run for head `052ff2b3c3610c076b52de51f14877c42f4e3102` is green. The PR is not ready to merge, however, because it publishes an undocumented third-party extension contract while putting enterprise-specific defaults in the OSS shared package, and because the real dispatcher/login regression requested by issue #5212 is not covered by a test that would fail on the vulnerable implementation.

## Verdict
❌ request changes — the public-contract/package-boundary blocker and the missing auth-critical integration regression must be resolved before merge.

## 🧪 Validation Gate

Validation source: GitHub checks. Runner: GitHub Actions; no local fallback was run because the checks are complete and usable for this exact head.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | The CI `prepare` job completed its “Build packages” step. |
| `yarn generate` | ✅ PASS | The CI `prepare` job completed its “Prepare generated modules” step. |
| `yarn build:packages` | ✅ PASS | The CI `prepare` job completed its post-generation “Rebuild packages with generated files” step. |
| `yarn i18n:check-sync` | ✅ PASS | The CI `test` job completed “Check i18n sync”. |
| `yarn i18n:check-usage` | ✅ PASS | The CI `test` job completed “Check i18n usage”. |
| `yarn typecheck` | ✅ PASS | The CI `test` job completed “Checking types”. |
| `yarn test` | ✅ PASS | The CI `test` job completed “Test”; all 15 ephemeral-integration shards also passed. |
| `yarn build:app` | ✅ PASS | The CI `prepare` job completed “Build app”. |
| `yarn template:sync` | ✅ PASS | The CI `test` job completed the always-unfiltered create-app template parity check. |

The separate mutation-test workflow also passed for `jwt.ts`, `mfaPendingAccess.ts`, and `server.ts` on this head.

## Findings

### ⛔ Blocker

- `packages/shared/src/lib/auth/mfaPendingAccess.ts:6` — The new registry is exported through `@open-mercato/shared`'s wildcard package exports and is explicitly documented at line 39 as a third-party MFA extension point, but no governing spec with a “Migration & Backward Compatibility” section defines that new public contract. At the same time, the shared package pre-registers `/api/security/*` routes owned by the commercial enterprise module. This conflicts with the enterprise package rule that commercial-only behavior stays in `packages/enterprise` and with `SPEC-ENT-007`, which defines the MFA login extension without shared package ownership. The repository compatibility checklist classifies a contract-surface change without the required spec as Critical. Update the relevant MFA/auth spec to define the generic pending-auth extension contract and its compatibility guarantees, then keep the shared registry generic and register the canonical security routes from the enterprise security module during its bootstrap (with a test proving registration happens before request authentication).

### ⚠️ Major

- `packages/shared/src/lib/auth/server.ts:410` — The security boundary is covered only by resolver-level unit tests. Issue #5212 explicitly requires a failing integration test that performs a real MFA-enabled login, uses the returned pending token against representative protected read and mutation APIs, and proves the completion and verified-token paths still work; the enterprise MFA and login-interceptor specs likewise require this end-to-end case. Existing integration shards being green do not close the gap because none of the changed tests exercises the dispatcher/cookie/login chain and fails on the vulnerable pre-fix code. Add a self-contained case under `packages/enterprise/src/modules/security/__integration__/` (for example `TC-SEC-002.spec.ts`) that proves bearer and cookie pending credentials receive 401 from protected APIs with the cookie-clearing behavior, `prepare`/`verify`/`recovery` remain reachable, and the verified replacement token regains normal access. Include the issue's expiry and failed-challenge cleanup cases or explicitly cover them in the same security regression suite.

None — no unaddressed review bodies, conversation comments, or inline review comments were present on the PR.

## 💥 Breaking Changes

- [x] No exported/public symbol was removed or renamed.
- [x] No existing function signature changed in a breaking way.
- [x] No HTTP route URL or method was removed or renamed.
- [x] No field was removed or retyped in an existing response shape.
- [x] No event or message name or payload field was removed.
- [x] No CLI command, flag, or machine-parsed output changed.
- [x] No database table, column, index, or type changed.
- [x] No configuration key or default changed.
- [ ] The newly published `registerMfaPendingAccessRoutes` contract and its package-boundary ownership are not defined by a referenced spec with the required compatibility section; resolve the blocker above before treating the additive API as stable.

## 🧪 Test Coverage

The new unit tests cover claim classification, exact method/path allowlisting, rejection before canonical role restoration, the three completion paths, verified replacement tokens, ordinary tokens, and cookie-based page resolution. Mutation testing passed for all three changed production files, and the existing repository integration suite is green. What is missing is the regression test at the actual trust boundary: login interceptor → pending response token/cookie → dispatcher authorization and cookie clearing → MFA completion/recovery → verified access. That test is required because it is the only one that proves the original exploit is closed through the production route stack rather than only through a mocked resolver.

pkarw:

_Updated after the round-3 re-review of head `7fb8730c4`. No label changes were needed this round; the rationale is restated so the current state is self-documenting._

- ❌ `changes-requested` — the re-review confirms every round-1 and round-2 finding is genuinely resolved, but a new High finding stands: `Closes #5212` is not satisfied because `GET /api/auth/session/refresh` (`requireAuth: false`) still mints a full, non-pending staff token from a surviving `session_token`, and this PR's own RSC change is what now routes MFA-pending users into that redirect automatically. Two Medium findings would request changes on their own under this repo's verdict rule.
- 🔒 `security` — unchanged; the PR closes a deterministic second-factor bypass in the central staff-auth resolution path.
- 🧪 `needs-qa` — unchanged. The change alters login, pending-session, verification, recovery and cookie-clearing behavior. Automation is fully green, but the automated-verification exemption does not apply to an auth-boundary change of this shape, so manual exercise is still warranted. Not applying `qa-approved` — that is earned by manual QA, never by reading a diff.
- 🔥 `priority-extreme` — unchanged; the underlying issue permits password-only access to staff APIs for an MFA-enrolled account, super-admins included.
- 💥 `risk-high` — unchanged; the diff touches `resolveAuthFromRequestDetailed` / `resolveAuthFromCookiesDetailed`, the resolution path every staff API and RSC page depends on.

Pipeline state stays `changes-requested`; `priority-extreme` and `risk-high` remain the correct single priority and single risk labels.

pkarw:
Thanks @haxiorz — review found actionable items, so I am handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pkarw:

The full review was submitted with one blocker and one major finding. GitHub CI and mutation testing are green, but the author must define the new public auth extension contract and correct its package ownership, then add the missing end-to-end MFA-pending regression coverage. Autofix: skipped (not my PR — re-run with --autofix to fix it here).

pkarw:

All reported checks for head `052ff2b3c3610c076b52de51f14877c42f4e3102` settled successfully or were intentionally skipped by their workflow conditions. The successful checks include package and app builds, generation, typechecking, unit tests, i18n checks, template parity, all 15 ephemeral-integration shards, and mutation testing for the three changed auth files. CI does not change the CHANGES REQUESTED verdict because the remaining findings concern package/contract governance and a missing regression scenario rather than a failing existing gate.

haxiorz:
Both findings are addressed in cf2501082 + 06622a522 (CI fully green on 06622a522):

**Blocker — package boundary / unspecified contract:**
- `packages/shared/src/lib/auth/mfaPendingAccess.ts` is now purely generic and ships an **empty** registry — no `/api/security/*` defaults remain in OSS shared.
- The canonical `POST /api/security/mfa/{prepare,verify,recovery}` routes are registered by the enterprise `security` module (`lib/mfaCompletionRoutes.ts`), imported from the module entry (`index.ts`) and from the completion routes' shared helper, so registration always precedes request authentication. Pinned by `__tests__/mfaCompletionRoutes.test.ts`.
- The pending-auth extension contract, ownership boundary, and compatibility guarantees are defined in the new "Amendment 2026-08-21 — Central MFA-Pending Token Gate" section of `SPEC-ENT-007` (incl. Migration & Backward Compatibility table and integration coverage list).

**Major — missing boundary-level regression:**
- New `TC-SEC-010.spec.ts`: real login → interceptor-issued pending token → protected **read** (`GET /api/auth/profile`) and **mutation** (`POST /api/security/mfa/recovery-codes/regenerate`) both return 401 via bearer *and* cookie credentials, with `auth_token` cleared (max-age=0) by the dispatcher; `prepare`/`verify` stay reachable for the pending token; the verified replacement token regains normal read+mutation access; attempt-exhaustion cleanup locks the challenge so even the correct code stops working while a fresh challenge verifies fine. Challenge-TTL expiry remains covered at service level (`MfaVerificationService.test.ts`), as noted in the spec's coverage table.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-23T00:31:23Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔁 Re-review: fix(security): reject MFA-pending tokens from general staff APIs

## 🎯 Summary

Both findings from the previous review are **properly resolved**, and I verified each one rather than taking the fix note at face value.

- **Blocker (package boundary / undefined contract) — resolved.** `packages/shared/src/lib/auth/mfaPendingAccess.ts` now ships an **empty**, implementation-agnostic registry with no `/api/security/*` knowledge. The enterprise `security` module owns its own canonical routes via `lib/mfaCompletionRoutes.ts`, imported from both `index.ts` and `api/mfa/_shared.ts`. `SPEC-ENT-007` gained an Amendment that defines the contract, the ownership boundary, **and** the required `Migration & Backward Compatibility` table classifying the new exports as ADDITIVE-ONLY and the registry-default change as security-narrowing.
- **Major (missing end-to-end regression) — resolved.** `TC-SEC-010.spec.ts` drives the real chain: MFA login → pending token → protected read/mutation probes → dispatcher 401 + cookie clearing → completion routes reachable → verified replacement token restores access → challenge-exhaustion cleanup. It fails on the pre-fix implementation (a pending bearer token used to return 200 from `/api/auth/profile`).

The remaining item is a **test-case ID collision introduced by merge drift**, not by the security logic. Under this repo's verdict rule (Medium findings request changes) that holds the merge, but it is a ~3-line rename — the security work itself is otherwise ready.

## Verdict

❌ request changes — one Medium finding (duplicate `TC-SEC-010` identifier). No blocker or major remains.

## 🧪 Validation Gate

**Validation source: GitHub checks.** No local fallback was run — the checks are complete and usable for this exact head (`06622a52`).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | CI `prepare` job — "Build packages" |
| `yarn generate` | ✅ PASS | CI `prepare` job — "Prepare generated modules" |
| `yarn build:packages` (rebuild) | ✅ PASS | CI `prepare` job — post-generation rebuild |
| `yarn i18n:check-sync` | ✅ PASS | CI `test` job |
| `yarn i18n:check-usage` | ✅ PASS | CI `test` job |
| `yarn typecheck` | ✅ PASS | CI `test` job — "Checking types" |
| `yarn test` | ✅ PASS | CI `test` job |
| `yarn build:app` | ✅ PASS | CI `prepare` job — "Build app" |

All 15 `ephemeral-integration` shards, `lint`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance` and the `mutate` mutation-test job (`jwt.ts`, `mfaPendingAccess.ts`, `server.ts`) are green on `06622a52`.

## Findings

### 🟡 Medium

**`packages/enterprise/src/modules/security/__integration__/TC-SEC-010.spec.ts:1` — duplicate `TC-SEC-010` test-case identifier after merge.**

`origin/develop` already contains `TC-SEC-010-sudo-binding-substitution.spec.ts` (plus `TC-SEC-011/012/013`), all of which landed after this branch's last upstream merge on 2026-08-21. The branch itself does not carry them, so the collision is invisible locally — but it is already visible in this PR's own CI output, where both files report under the same ID:

```
✓  64 …/TC-SEC-010-sudo-binding-substitution.spec.ts:18:3 › TC-SEC-010: Sudo challenge binding substitution …
✓  65 …/TC-SEC-010.spec.ts:69:3 › TC-SEC-010: MFA-pending tokens are rejected by general staff APIs
```

`.ai/qa/AGENTS.md` defines `TC-{CATEGORY}-{XXX}` as the test-case identity and uses it as the key in QA result tables (`| TC-AUTH-001 | … | PASS |`). Two unrelated cases sharing `TC-SEC-010` makes those tables and the spec's traceability ambiguous — a QA result against "TC-SEC-010" no longer identifies which case ran.

To be explicit: this is **merge drift, not a mistake in the fix** — `010` was the next free ID when the test was written.

Remediation (next free ID is `TC-SEC-014`):
1. Rename the file to `TC-SEC-014.spec.ts`.
2. Update the `test.describe` title to `TC-SEC-014: MFA-pending tokens are rejected by general staff APIs (#5212)`.
3. Update the six `TC-SEC-010` references in the `SPEC-ENT-007` Amendment's **Integration coverage** table.

Merging the latest `develop` into the branch at the same time will surface the collision locally.

## 🔬 Verification performed

- **Registration-before-authentication ordering (the risk the new fail-closed design introduces).** Because `shared` no longer pre-registers the completion routes, a completion request whose module had not yet loaded would be denied and MFA login would break. Traced `apps/mercato/src/app/api/[...slug]/route.ts` → `handleRequest` calls `await match.route.load()` **before** `await resolveAuthFromRequestDetailed(req)`. All three completion routes (`prepare`, `verify`, `recovery`) import `../_shared`, which imports `../../lib/mfaCompletionRoutes`. Registration therefore always precedes auth resolution, on both the bootstrap and the per-route path. Covered by `__tests__/mfaCompletionRoutes.test.ts`, including idempotency.
- **No bypass entry point.** `getAuthFromRequest` and `getAuthFromCookies` both delegate to their `*Detailed` counterparts, so the gate sits at the single choke point; there is no second resolver that skips it.
- **Rejection precedes role restoration.** In `server.ts:410` the `isMfaPendingJwtPayload` check returns `{ auth: null, status: 'invalid' }` before `resolveCanonicalInteractiveAuthContext`, so canonical roles/features are never restored for a provisional credential. Cookie/RSC resolution rejects pending tokens unconditionally.
- **The new integration test really executes.** A green shard set does not by itself prove a newly added spec was picked up, so I located it in the CI logs: all four `TC-SEC-010.spec.ts` cases ran and passed in shard 15/15 with `OM_ENABLE_ENTERPRISE_MODULES_SECURITY: true` (`passed=…, failed=0, skipped=0`).
- **Diff scan.** No new `any`, no raw `console.*`/`alert(`, no hardcoded user-facing strings, no one-letter identifiers in production code.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed.
- [x] No existing signature changed in a breaking way.
- [x] No HTTP route URL or method removed or renamed.
- [x] No event or message name or payload field removed.
- [x] No CLI command, flag, or machine-parsed output changed.
- [x] No database table, column, index, or type changed.
- [x] No configuration key or default changed.
- [x] **Resolved** — the new `registerMfaPendingAccessRoutes` / `isMfaPendingAccessAllowed` / `listMfaPendingAccessRoutes` / `isMfaPendingJwtPayload` contract is now defined by `SPEC-ENT-007` § Amendment 2026-08-21 with the required Migration & Backward Compatibility section.

One deliberate behavior narrowing is documented and correct: a third-party MFA module that completes challenges on its own endpoints must now call `registerMfaPendingAccessRoutes` during bootstrap, or its completion routes fail closed. The spec classifies this as MIGRATION REQUIRED (one line).

## 🧪 Test Coverage

Strong. Unit coverage for claim classification, exact method/path allowlisting, rejection before role restoration, the three completion paths, verified replacement tokens, ordinary tokens, and cookie-based page resolution; module-registration and idempotency tests; and the end-to-end `TC-SEC-010` regression through the production dispatcher. Mutation testing passes on all three changed production files. No coverage gap remains.

---

Nice work on the follow-up — the ownership split is exactly right, and the registration-ordering guarantee via `_shared.ts` is a thoughtful belt-and-braces touch. Only the ID rename stands between this and approval.

pkarw:
Label kept at `changes-requested` because the re-review found one remaining Medium finding (duplicate `TC-SEC-010` test-case ID), which requests changes under this repo's verdict rule. `priority-extreme` and `risk-high` remain correct for an auth-boundary security fix.

pkarw:
Thanks @haxiorz — both previous findings are properly resolved, and I verified them rather than taking the notes at face value: I traced the dispatcher to confirm `match.route.load()` runs before `resolveAuthFromRequestDetailed`, so the new fail-closed registration can never deny a real completion request; and I located `TC-SEC-010` in the CI logs to confirm all four cases actually executed and passed in shard 15/15 rather than being silently skipped.

The only thing left is a **test-case ID collision that is merge drift, not your mistake**: `develop` picked up `TC-SEC-010-sudo-binding-substitution.spec.ts` (and `TC-SEC-011/012/013`) after your last upstream merge, so `TC-SEC-010` is now taken. Renaming to `TC-SEC-014` in the filename, the `test.describe` title, and the six references in the `SPEC-ENT-007` Amendment's Integration coverage table clears it. Merging the latest `develop` at the same time will surface the clash locally.

Handing this back to you for that pass — the security work itself is approved. Push the rename and re-request review, and the automation can pick it up from the new head.

pkarw:

Re-review of head `06622a52`. Both prior findings (the shared/enterprise package-boundary blocker and the missing end-to-end regression) are resolved and independently verified. One Medium finding remains: the new integration spec reuses the `TC-SEC-010` identifier, which `develop` already assigned to `TC-SEC-010-sudo-binding-substitution.spec.ts` — merge drift rather than a defect in the fix.

No autofix iterations were run. The single remaining item is a ~3-line rename on a fork branch whose author is actively responding, so taking the branch over with a carry-forward replacement PR would have been disproportionate; it is handed back instead. Validation source: GitHub checks (all green on this head); no local fallback was needed.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-24T00:02:34Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review — Re-review (round 3) of `7fb8730c4`

## 🎯 Summary

This is the third pass on PR #5453. The two findings from round 1 (the `packages/shared` ↔ enterprise package-boundary blocker and the missing end-to-end regression) and the single Medium from round 2 (the `TC-SEC-010` identifier collision) are all **resolved and independently verified** — not merely reported as resolved.

The core change is sound and I want to say that plainly before the findings: rejecting `mfa_pending` tokens inside `resolveAuthFromRequestDetailed` / `resolveAuthFromCookiesDetailed` closes a real, deterministic bypass, the allowlist is genuinely fail-closed, the ownership split between generic `packages/shared` infrastructure and the enterprise `security` module is the right architecture, and the test coverage is real rather than decorative.

What holds this round is one **High** finding that came out of applying this repository's own documented lesson *"Shared security-default changes require a complete consumer audit"* (`.ai/lessons/shared-security-default-changes-require-a-complete.md`, indexed under `module: shared, auth`). I enumerated every staff-identity path in the monorepo that does **not** funnel through the two gated resolvers. The audit is almost entirely clean — and that is worth stating, because it is the strongest evidence for this design. But it surfaced one path that both defeats `Closes #5212` for a realistic user population and, critically, is now reached **automatically as a direct consequence of this PR's own RSC change**.

**Verdict: request changes** — one major finding. Under this repository's verdict rule (Medium and above requests changes), the two Medium findings would independently produce the same outcome.

To be explicit about proportionality: this PR **improves** the security posture and I am not asking for it to be shelved. Before it, the pending token authenticated everything, which is strictly worse than what ships here. The ask is either a small in-PR closure or an honest narrowing of the claim — details under the finding.

## 🧪 Validation Gate

**Validation source: GitHub checks** (per this repo's GitHub-checks-first review extension — CI already ran the full gate, so it was not reproduced locally).

All **28** check runs on the exact head SHA `7fb8730c438121066731e0d5df24a6d8c28ff050` are settled: **0 failing, 0 pending**. No local fallback was warranted.

| Gate | Result | Evidence |
|------|--------|----------|
| `prepare` (incl. `yarn build:packages`, `yarn generate`, `yarn build:app`) | ✅ pass | 3m48s |
| `test` (incl. `yarn typecheck`, unit tests, `yarn i18n:check-usage`, `yarn agents:check-budget`) | ✅ pass | 6m58s |
| `lint` | ✅ pass | 29s |
| `ds-lint` | ✅ pass | 40s |
| `scope` / `audit-scope` | ✅ pass | 28s / 12s |
| `documents-multi-instance` | ✅ pass | 1m8s |
| `ephemeral-integration` (15/15 shards) | ✅ pass | all green |
| `mutate (shared: jwt.ts, mfaPendingAccess.ts, server.ts)` | ✅ pass | 4m28s |
| `license/cla` | ✅ pass | signed |
| `audit`, `docker-build`, `merge-coverage`, `comment`, `[code]smith` | ⏭️ skipped by workflow conditions | not applicable to this head |

Merge state: **`MERGEABLE`** — no conflicts against `develop`. (`mergeStateStatus: BLOCKED` reflects the standing changes-requested review, not a code problem.)

I did not take the shard summary at face value: I opened the shard 15/15 log and confirmed all four `TC-SEC-014` cases **actually executed and passed** rather than being silently skipped:

```
✓ 68 TC-SEC-014.spec.ts:69:3 › pending bearer token gets 401 … (66ms)
✓ 69 TC-SEC-014.spec.ts:79:3 › pending cookie credentials get 401 with staff auth cookies cleared … (68ms)
✓ 70 TC-SEC-014.spec.ts:94:3 › completion routes stay reachable … and the verified replacement regains access (617ms)
✓ 71 TC-SEC-014.spec.ts:121:3 › exhausted challenge is cleaned up … (284ms)
```

That green run is also the empirical answer to the one architectural worry this design invites — that `packages/shared`'s module-level registry `Map` and the enterprise module's registration call could land in two different module instances and silently deny every completion route. Case 70 exercises `prepare` → `verify` → verified-token access through the real dispatcher and returns 200, so the registry is provably shared at runtime.

## ✅ Previous findings — verified resolved

- **[round 1, blocker] Package boundary / unspecified contract** — `packages/shared/src/lib/auth/mfaPendingAccess.ts` now ships a genuinely **empty** registry with no `/api/security/*` knowledge, and the canonical routes are owned and registered by the enterprise `security` module (`lib/mfaCompletionRoutes.ts:3-7`, imported from `index.ts:1` and `api/mfa/_shared.ts:6`). Verified resolved.
- **[round 1, major] Missing end-to-end regression** — `TC-SEC-014.spec.ts` covers the bypass through the real HTTP dispatcher, including cookie clearing and the verified-replacement path. Verified resolved and verified *executing* (above).
- **[round 2, Medium] `TC-SEC-010` identifier collision** — resolved. I re-swept every `TC-SEC-*` identifier across spec filenames and `describe` blocks: `TC-SEC-001`…`TC-SEC-014` are now each used exactly once, with `TC-SEC-010`…`TC-SEC-013` correctly retained by the sudo specs that `develop` owns and `TC-SEC-014` uniquely held by this PR. The spec amendment's coverage table was updated to match. No collision remains.

## 🔴 Findings — High (major)

### 1. `Closes #5212` is not satisfied: `/api/auth/session/refresh` upgrades an MFA-pending browser to a full staff token, and this PR is what routes users into it

`packages/core/src/modules/auth/api/session/refresh.ts:63`, `:81`, `:150` · `apps/mercato/src/app/(backend)/backend/page.tsx:10` · `apps/mercato/src/app/(backend)/backend/[...slug]/page.tsx:72` · `packages/core/src/modules/auth/api/login.ts:220-225`

The gate added in this PR only inspects tokens that reach `resolveAuthFromRequestDetailed` / `resolveAuthFromCookiesDetailed`. `GET /api/auth/session/refresh` never does — it is declared `requireAuth: false` (`refresh.ts:150-153`), so the dispatcher's resolution result is irrelevant, and the handler never reads `auth_token` at all. It reads the **`session_token`** cookie and mints a brand-new staff JWT:

```ts
const token = parseCookie(req, 'session_token')            // refresh.ts:67
const ctx = await auth.refreshFromSessionToken(token)      // :76
const jwt = signJwt(buildStaffJwtClaims(ctx))              // :81 — no mfa_pending, no mfa_verified
res.cookies.set('auth_token', jwt, { …, maxAge: 60 * 60 * 8 })
```

`AuthService.refreshFromSessionToken` (`packages/core/src/modules/auth/services/authService.ts:125-134`) checks only the token hash and `expiresAt` — it has no MFA awareness whatsoever — and `buildStaffJwtClaims` (`refresh.ts:34-43`) emits `{ sub, sid, tenantId, orgId, email, roles }`. Roles are fully restored. This is not an inert token.

**Why an MFA-pending user still holds a usable `session_token`:** the MFA login path leaves any pre-existing one completely untouched. At `login.ts:220-225` both branches evaluate false in the MFA flow — the interceptor's `replace` body drops `refreshToken`, so `refreshTokenForCookie` is `undefined`, and `authTokenForCookie` is the *pending* token so it is `!== token`:

```ts
if (remember && refreshTokenForCookie) { … }               // false: interceptor dropped refreshToken
else if (!remember && authTokenForCookie === token) { … }  // false: authTokenForCookie is the pending token
```

Nothing clears the old cookie, and `createSession` never revokes prior sessions.

**Why this PR is what makes it fire.** Before this change, an MFA-pending cookie resolved to an authenticated context in RSC, so the `if (!auth)` branch was never taken. After `server.ts:367`, *every* MFA-pending page navigation to `/backend/**` now falls into:

```ts
if (!auth) redirect('/api/auth/session/refresh?redirect=' + encodeURIComponent(pathname))
```

**Failure scenario (no attacker required):** a staff user logs in with "remember me" before MFA is enforced → they enroll TOTP, or an admin enforces MFA → at their next login they get the challenge screen and a pending token, while the 30-day `session_token` from the earlier login is still in the cookie jar → they type `/backend` in the URL bar → `getAuthFromCookies()` returns `null` → redirect → they land in the admin with a clean 8-hour staff token. The second factor is never presented. The same applies to any browser whose `session_token` came from the SSO callback (`packages/enterprise/src/modules/sso/api/callback/oidc/route.ts`), which sets `session_token` directly and bypasses the login interceptor entirely.

I want to be fair about severity, so: this is **not** a new vulnerability, and it is **not** a reason to shelve the PR. Before this change the pending token itself granted the same access to everything, which is worse. I am grading it **High rather than blocker** for exactly that reason. What makes it a major rather than a note is that the PR asserts closure it does not have — `Closes #5212` will auto-close the issue on merge, and the spec amendment durably records this as *"the platform-wide gate that rejects pending credentials everywhere"* with *"Central enforcement"*, which will mislead the next person who audits this surface.

**Two acceptable remedies — your call, and I would take either:**

- **(a) Preferred, ~3 lines, inside this PR's own blast radius.** In `login.ts`, when the interceptor replaced the token (`authTokenForCookie !== token`), explicitly clear `session_token` on the response. That removes the automatic path for the entire pending window. It costs MFA users nothing: `issueVerifiedMfaToken` / `setAuthCookie` (`api/mfa/_shared.ts:166-174`) never issue a `session_token` after verification either, so MFA accounts already have no remember-me. Add one `TC-SEC-014` case asserting `GET /api/auth/session/refresh` does not return a usable staff token during the pending window.
- **(b) Narrow the claim honestly.** Change `Closes #5212` to `Refs #5212`, soften the spec amendment's "everywhere" / "platform-wide" language to state that it covers *pending-token resolution* and not alternate credential paths, and file a follow-up issue for making `refreshFromSessionToken` MFA-aware (refuse to mint, or re-issue a challenge, when the target user has an unsatisfied enforced MFA policy). Note that only the follow-up covers the SSO-planted `session_token`, so it is worth filing regardless of which remedy you pick.

## 🟠 Findings — Medium (minor)

### 2. The trusted-auth-context short-circuit is placed ahead of the new MFA check

`packages/shared/src/lib/auth/server.ts:387-393`

`resolveAuthFromRequestDetailed` returns from `readTrustedAuthContext(req)` before reaching the `isMfaPendingJwtPayload` gate at `:410`:

```ts
const trusted = readTrustedAuthContext(req)
if (trusted) {
  return { auth: trusted.auth, status: trusted.status ?? (trusted.auth ? 'authenticated' : 'missing') }
}
```

This is **not exploitable today** — I traced the only caller, `packages/ai-assistant/src/modules/ai_assistant/lib/ai-api-operation-runner.ts:303`, and its envelope is built from an identity that already came through a gated `getAuthFromRequest` or an api-key record. I am raising it because the PR's own framing is "central enforcement", and this is a documented hole in that centrality that also skips `resolveCanonicalStaffAuthContext` (so session revocation is bypassed there too, independently of MFA). The cheap fix is a comment at `:387` stating the invariant callers of `attachTrustedAuthContext` must uphold — that the attached context was itself produced by a gated resolution — so a future caller does not quietly widen this. A code-level assertion would be better still.

### 3. `POST /api/auth/logout` becomes unreachable during the pending window, and the migration table does not say so

`packages/core/src/modules/auth/api/logout.ts:83-85`

`logout` is `requireAuth: true` and is not on the pending allowlist, so a pending token now gets a 401 before the handler runs. The dispatcher still clears the cookies, so the browser ends up logged out — but the handler's session revocation (`logout.ts:55-66`) and the `auth.logout` event (`:67-76`) never fire, leaving an unrevoked `Session` row for every abandoned challenge (up to 8 hours, or 30 days with remember-me).

I checked whether those rows are dangerous and they are **inert**: the session's refresh token is generated at `login.ts:163` but the interceptor drops it from the response body and no cookie is set, so nobody holds it, and the `sid` only appears in a pending token that is now rejected everywhere. So this is a housekeeping and documentation issue, not a security one. It belongs in the spec amendment's *Migration & Backward Compatibility* table, which currently claims *"Login/MFA endpoint URLs, methods, request/response shapes — UNCHANGED"* — accurate for shapes, but silent on this reachability change to a public route. Please add a row. Adding `/api/auth/logout` to the allowlist would be the alternative, but I would **not** recommend it: it widens the pending window for no real benefit given the rows are inert.

## 🔵 Findings — Low (nit)

### 4. The exhaustion loop in `TC-SEC-014` is bounded above its own endpoint's rate limit

`packages/enterprise/src/modules/security/__integration__/TC-SEC-014.spec.ts:125`

The loop allows up to 15 wrong verify attempts, but `POST /api/security/mfa/verify` is rate-limited to `points: 10, duration: 60` (`api/mfa/verify/route.ts:20`). Today the test is safe and passed in 284ms: `securityConfig.mfa.maxAttempts` is `5` (`lib/security-config.ts:63`), the 5th failure expires the challenge (`MfaVerificationService.registerFailedAttempt:256-262`), and the 6th attempt returns 404, which trips the `status !== 401 && status !== 400` break — so the spec makes ~8 verify calls total, two under the limit. The fragility is that if the lockout status ever becomes a 400 or 401, the loop runs to 15, exhausts the limiter, and the final `freshVerify` assertion (which expects 200) fails inside the limiter's block window — presenting as a confusing flake rather than the real regression. Bounding the loop at `maxAttempts + 1` instead of `15` makes the test fail for the right reason.

### 5. The PR description no longer matches the diff

The *What changed* section lists *"added 10 translation keys referenced by shipped pages but absent from all locale files"*, but the current 12-file diff contains no locale files — those keys reached `develop` separately and were absorbed by the `7fb8730c4` merge. Worth trimming from the description so the merge commit message stays accurate.

### 6. Mixed sort-comparator styles in one file

`packages/shared/src/lib/auth/mfaPendingAccess.ts:69`

Commit `052ff2b3c` switched the *method* sorts to explicit unit comparators for the `#3620` guard, but the *path* sort still uses `localeCompare`, and `mfaCompletionRoutes.test.ts:14-18` asserts an exact path ordering against it. This passes the guard (`scripts/repo-wide-guards.mjs:53` only flags comparator-less `sort()`) and `localeCompare` is used elsewhere in the repo, so this is purely a consistency nit within a single file — but `scripts/check-resolutions.mjs:58` records the reasoning that ordering should be "unit rather than locale" for reproducibility, which argues for matching the comparator two lines above it.

### 7. `PUT /api/auth/profile` re-signs `auth_token` without `mfa_verified`

`packages/core/src/modules/auth/api/profile/route.ts:184`

The profile update re-signs the token as `{ sub, sid, tenantId, orgId, email, roles }`, silently dropping the `mfa_verified: true` that `issueVerifiedMfaToken` set. Harmless against the current gate, which keys only on `mfa_pending === true`, and unreachable by a pending token. Flagging it because it means `mfa_verified` is not a durable session property: if a future check ever tightens to *require* `mfa_verified === true`, this route will silently downgrade the session. Pre-existing and out of scope for this PR — a comment on the new `isMfaPendingJwtPayload` JSDoc noting that `mfa_verified` is not durable would save the next author the trip.

## 💥 Breaking Changes

- [x] **No exported API removed or renamed.** All four new exports (`isMfaPendingJwtPayload`, `registerMfaPendingAccessRoutes`, `isMfaPendingAccessAllowed`, `listMfaPendingAccessRoutes`) are additive, and the spec amendment classifies them ADDITIVE-ONLY with signatures frozen as shipped.
- [x] **No HTTP route or response shape changed.** The MFA completion contracts already required `mfa_pending === true` and are unchanged.
- [x] **No event ID, DI key, ACL feature, notification ID, CLI command, or widget spot ID changed.**
- [x] **No DB schema change**, so no migration is expected and none is present — correct for this change.
- [x] **Security narrowing is documented** in the amendment's Migration table, including the one-line migration required of third-party MFA modules.
- [ ] **Reachability change to `POST /api/auth/logout` is undocumented** — see finding 3. This is the only gap in an otherwise complete compatibility treatment.

No protected contract surface from `BACKWARD_COMPATIBILITY.md` is broken.

## 🧪 Test Coverage

Coverage on the changed behavior is genuinely strong, and the unit/integration split is well chosen:

- `mfaPendingAccess.test.ts` — empty-by-default registry, exact method matching, trailing-slash tolerance, case handling, near-miss paths (`/recovery-codes` vs `/recovery`, `/recovery/impersonate`), fail-closed on malformed input, and additive/idempotent merging.
- `jwt.test.ts` — the predicate's full truth table including `mfa_verified: true` overriding `mfa_pending: true`, and non-object inputs.
- `server.test.ts` — pending token denied on a general API *before* `resolveCanonicalStaffAuthContext` is called (asserted via `not.toHaveBeenCalled()`, which is the right assertion — it proves roles are never restored, not merely that the response was 401), allowed on each registered completion route, denied for a wrong method on a completion route, verified replacement behaving normally, non-MFA tokens untouched, and unconditional cookie/RSC rejection.
- `mfaCompletionRoutes.test.ts` — registration happens at module load, is idempotent, and does not widen to siblings like `/api/security/mfa/methods`.
- `TC-SEC-014.spec.ts` — the end-to-end proof through the real dispatcher, including `Set-Cookie: auth_token=; max-age=0`.

**The one gap is the counterpart to finding 1**: there is no case asserting that an MFA-pending browser cannot obtain a full staff token by any *other* route. If you take remedy (a), the case to add is: complete a remember-me login to plant `session_token`, enroll MFA, log in again to reach the pending state, then assert `GET /api/auth/session/refresh?redirect=/backend` does not yield a working staff token.

Two minor test-hygiene notes, neither blocking: the `mfaPendingAccess` registry is process-global with no reset helper, so `server.test.ts`'s `beforeEach` registration leaks across cases in the file and `mfaPendingAccess.test.ts` has to defend with `t2`/`t3`/`t4`/`t5` path prefixes. Exporting a test-only reset would make these suites order-independent. And the `jest.requireActual` spread added at `server.test.ts:13` is the right fix for the partial mock — worth keeping in mind that it now un-mocks every other `jwt` export in that file.

## 📌 Verdict

**Request changes** — 1 High, 2 Medium, 4 Low.

The fix itself is correct, well-architected, well-tested, and CI is fully green on `7fb8730c4`. Every finding from rounds 1 and 2 is genuinely closed and I verified each rather than accepting the summary. The one thing standing between this and an approval is finding 1 — and the smaller of its two remedies is roughly three lines in `login.ts` plus one integration case. Remedy (b) is equally acceptable if you would rather keep this PR tightly scoped; in that case please file the follow-up so the residual path is tracked rather than lost.

pkarw:
Thanks @haxiorz — handing this back to you for one more pass.

First, credit where it is due: everything from the previous two rounds is genuinely closed, and I verified each rather than taking the notes at face value. The package-boundary split is right, `TC-SEC-010`…`TC-SEC-014` are now each used exactly once across the whole repo, and I opened the shard 15/15 log to confirm all four `TC-SEC-014` cases actually executed and passed rather than being skipped — which also disproves the one architectural worry this design invites, that the shared registry `Map` and the enterprise registration could land in separate module instances.

What holds this round came out of applying this repo's own lesson *"Shared security-default changes require a complete consumer audit"*. I enumerated every staff-identity path in the monorepo that does not go through the two gated resolvers. The audit is almost entirely clean — which is the strongest evidence for your design — but it found one that matters:

**`GET /api/auth/session/refresh` is `requireAuth: false`, never reads `auth_token`, and mints a fresh full staff JWT from the `session_token` cookie with no MFA claims** (`refresh.ts:63,81,150`; `authService.refreshFromSessionToken:125` has no MFA awareness). The MFA login path leaves a pre-existing `session_token` completely untouched, because both branches at `login.ts:220-225` evaluate false when the interceptor replaces the token. And **your change is what makes this fire**: before it, a pending cookie resolved to an authenticated context in RSC, so the `if (!auth) redirect('/api/auth/session/refresh?…')` branch at `backend/page.tsx:10` and `backend/[...slug]/page.tsx:72` was never taken. Now a user with a remember-me (or SSO-planted) `session_token` who simply types `/backend` during the challenge lands in the admin with a clean 8-hour token, no second factor, no attacker involved.

I graded this **High, not a blocker**, deliberately. Your PR makes things strictly better — before it, the pending token granted that access to *everything* — and I am not asking you to shelve it. The problem is the claim: `Closes #5212` will auto-close the issue on merge, and the spec amendment durably records this as *"the platform-wide gate that rejects pending credentials everywhere"*, which will mislead the next person auditing this surface.

Either remedy is fine by me:

- **(a) Preferred, ~3 lines.** In `login.ts`, clear `session_token` when the interceptor replaced the token (`authTokenForCookie !== token`), plus one `TC-SEC-014` case asserting the refresh route yields nothing usable during the pending window. This costs MFA users nothing — `setAuthCookie` never issues a `session_token` after verification either, so MFA accounts already have no remember-me.
- **(b) Narrow the claim.** `Refs #5212` instead of `Closes`, soften the amendment's "everywhere"/"platform-wide" wording to cover pending-token *resolution* rather than all credential paths, and file a follow-up for making `refreshFromSessionToken` MFA-aware. Worth filing either way — only that follow-up covers the SSO-planted `session_token`.

Beyond that: two Mediums (the `attachTrustedAuthContext` short-circuit sitting ahead of the new check, and the undocumented reachability change to `POST /api/auth/logout`) and four Lows, all detailed in the review. None of them require redesign.

Push when ready and re-request review — the automation will pick it up from the new head.

pkarw:

Round-3 re-review of head `7fb8730c4`. All prior findings verified resolved — the round-1 package-boundary blocker, the round-1 missing end-to-end regression, and the round-2 `TC-SEC-010` identifier collision. Validation source: **GitHub checks** — all 28 check runs on the exact head SHA are settled with 0 failing and 0 pending, so no local gate was re-run; the shard 15/15 log was opened to confirm all four `TC-SEC-014` cases actually executed.

One new High finding stands: the gate is not central, because `GET /api/auth/session/refresh` (`requireAuth: false`) still mints a full non-pending staff token from a surviving `session_token`, and this PR's own RSC change is what now routes MFA-pending users into that redirect automatically — so `Closes #5212` does not hold for remember-me and SSO browsers. Graded High rather than blocker because the PR is a net security improvement. Plus 2 Medium and 4 Low findings.

No CI follow-up is pending — CI is already fully green on this head. Autofix: skipped (not my PR — re-run with --autofix to fix it here).

pkarw:
🤖 `om-auto-review-pr` (autofix mode) started by @pkarw at 2026-08-24T01:59:24Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review — Re-review (round 4, autofix) of `eac490f96`

## 🎯 Summary

**Verdict: approve.** Every finding from round 3 is closed, and the High one is closed by fixing the code rather than by narrowing the claim — `Closes #5212` is now accurate for the credential paths this platform actually exposes.

I ran this round in autofix mode, so the fixes in `eac490f96` are mine, pushed as a normal follow-up commit onto @haxiorz's branch (the PR carries `maintainerCanModify: true`; no force-push, original commits and authorship untouched). The design, the architecture and all the hard work remain @haxiorz's — this commit only closes the residual path round 3 identified.

## ✅ Round-3 findings — how each was closed

### 1. High — `/api/auth/session/refresh` upgraded an MFA-pending browser · **fixed (remedy (a) + defense in depth)**

Both halves of the path are now closed:

- **`api/auth/login.ts:218-231`** — `authTokenReplacedByInterceptor` (`authTokenForCookie !== token`) now drives an explicit `session_token` clear. A login whose token an interceptor swapped is provisional by definition, so it must not leave a remember-me cookie from an earlier pre-enrollment login usable. This is remedy (a) as proposed, and it costs MFA users nothing — `issueVerifiedMfaToken` never issues a `session_token` after verification either.
- **`api/auth/session/refresh.ts:28-43`, `:84-88`, `:131-138`** — `carriesMfaPendingToken(req)` reads the `auth_token` cookie or bearer, and both handlers refuse before touching `refreshFromSessionToken`: GET redirects to `/login` with staff cookies cleared, POST answers 401 with the same clearing. This goes past what remedy (a) alone would give: it holds even if some future path plants a `session_token` during a pending window, and it makes the "central enforcement" claim true for this route specifically rather than by omission.

Deliberately **not** in scope, and now written down instead of left implicit (see finding 3 below): `AuthService.refreshFromSessionToken` remains MFA-unaware, and the SSO callback plants `session_token` without passing through the login interceptor — so an SSO login never enters the pending state at all. That is a separate gap in MFA coverage for SSO, not a hole in this gate, and the spec amendment now names it as a tracked residual.

The happy path is unaffected: the challenge lives entirely client-side in `MfaChallengePanel` (swapped in by `widgets/components.ts` on the `om:auth:login-response` event), so no navigation to `/backend` happens mid-challenge, and after verification the token is no longer pending so refresh behaves exactly as before.

### 2. Medium — trusted-auth-context short-circuit placed ahead of the MFA check · **fixed (documented invariant)**

`packages/shared/src/lib/auth/server.ts:71-79` now carries a JSDoc stating the invariant callers of `attachTrustedAuthContext` must uphold — the envelope must be the product of a gated resolution, never raw request credentials — and spells out both consequences of violating it (the MFA bypass returns, and `resolveCanonicalStaffAuthContext` is skipped, so session revocation is bypassed too). `:397-399` carries the matching one-liner at the read site explaining why the short-circuit is deliberately ahead of the gate. A runtime assertion is not available here: the envelope carries an already-resolved `AuthContext`, not a JWT, so there is nothing left to inspect.

### 3. Medium — `POST /api/auth/logout` reachability undocumented · **fixed (spec row added)**

The amendment's Migration & Backward Compatibility table gained the row, including the analysis that the `Session` rows left by an abandoned challenge are inert (the refresh token is dropped from the interceptor's replaced body and never set as a cookie; the `sid` only appears in a token that is now rejected everywhere) and the explicit note that allowlisting `/api/auth/logout` was considered and rejected as widening the pending window for no benefit. Two further rows cover the reachability changes this round introduces (login's `Set-Cookie` behavior, session refresh), and the "platform-wide … everywhere" framing in the amendment intro is now scoped to the paths that actually resolve pending credentials.

### 4. Low — `TC-SEC-014` exhaustion loop bounded above its own rate limit · **fixed**

`MAX_WRONG_VERIFY_PROBES = 6` (`maxAttempts + 1`) replaces the bare `15`, with a comment recording both numbers it sits between.

### 5. Low — PR description out of sync · **fixed.** The stale translation-keys paragraph is gone and the description now reflects the 12-file diff plus this round's work.

### 6. Low — mixed sort-comparator styles · **fixed.** `mfaPendingAccess.ts:69` uses the same explicit unit comparator as the method sort two lines above.

### 7. Low — `mfa_verified` is not durable · **fixed (documented).** The `isMfaPendingJwtPayload` JSDoc now records why the predicate keys on `mfa_pending` alone and names the routes that re-sign without `mfa_verified`, so a future tightening does not silently downgrade those sessions.

## 🧪 Validation Gate

**Validation source: local fallback** — GitHub has no check data for `eac490f96` yet (the head is seconds old); CI is re-running the full gate now. Run in this review's isolated worktree after a clean `yarn install`:

| Gate | Result |
|------|--------|
| `yarn build:packages` | ✅ 26/26 tasks |
| `yarn generate` | ✅ |
| `yarn typecheck` | ✅ 26/26 packages |
| `yarn i18n:check-sync` | ✅ all locales in sync |
| `yarn i18n:check-usage` | ✅ advisory-only output, no new keys introduced |
| `yarn lint` | ✅ 0 errors |
| `packages/core` — `src/modules/auth` | ✅ 720/720 (84 suites) |
| `packages/shared` — `src/lib/auth` | ✅ 139/139 (10 suites) |
| `packages/enterprise` — `src/modules/security` | ✅ 305/305 (35 suites) |

The first `typecheck` pass failed in `@open-mercato/enterprise` on `Cannot find module '#generated/entities.ids.generated'`; that was a missing `yarn generate` in the fresh worktree, not a code defect, and it cleared on the next run. Noting it so the trace is honest rather than retro-fitted.

**Regression evidence.** I did not take "tests pass" as proof the tests test anything. With `packages/core/src/modules/auth/api/{login.ts,session/refresh.ts}` stashed back to the round-3 state, exactly the four new/changed cases fail — `refreshFromSessionToken` gets called and a staff JWT is minted, and `session_token` survives the challenge response — and all four pass with the fix restored.

## 🧪 Test Coverage

- `api/__tests__/session.refresh.test.ts` — four new cases: GET refuses (asserted via `refreshFromSessionToken` / `signJwt` `not.toHaveBeenCalled()`, which proves the exchange never happens rather than merely that the response looked wrong), POST 401s on a pending bearer, refresh still works once the token is verified, and an unverifiable-but-not-pending cookie does **not** block a legitimate refresh (the fail-open direction, which is the one worth pinning).
- `api/__tests__/login.test.ts` — the existing replace-interceptor case now asserts the clearing cookie instead of its absence, plus a new case for the no-remember-me branch.
- `TC-SEC-014.spec.ts` — the end-to-end case round 3 asked for: remember-me login before enrollment plants a real `session_token`, TOTP is enrolled, the next login returns the challenge and must clear the cookie, and a replay of `GET /api/auth/session/refresh` carrying both the pending `auth_token` and the planted `session_token` must not yield a staff token. It runs in `withCredentialIsolatedRequest` so no jar bleed can fake the result, and cleans up its own fixture user in a `finally`.

Round 3's two test-hygiene notes (no reset helper on the process-global registry; the `jest.requireActual` spread un-mocking sibling `jwt` exports) are unchanged and remain non-blocking.

## 💥 Breaking Changes

- [x] No exported API removed or renamed; no new exports added this round.
- [x] No HTTP route, request schema, or response shape changed. The three reachability changes (`POST /api/auth/login` `Set-Cookie`, session refresh for pending callers, `POST /api/auth/logout`) are all security-narrowing and all now carry a Migration table row.
- [x] No event ID, DI key, ACL feature, notification ID, CLI command, or widget spot ID changed.
- [x] No DB schema change, so no migration is expected and none is present.
- [x] No new user-facing string — the refresh 401 reuses the existing `auth.session.refresh.errors.invalidToken` key, which also avoids leaking that a challenge is outstanding.

No protected contract surface from `BACKWARD_COMPATIBILITY.md` is broken.

## 📌 Verdict

**Approve** — 0 blocker, 0 major, 0 minor, 0 nit.

Routing to `merge-queue`. `needs-qa` stays on: this is auth-critical and the QA-approval gate should hold the merge until a QA reviewer signs off on the flows in the manual-QA comment below — the browser challenge flow still has no Playwright E2E, and the new login/refresh behavior is exactly the sort of thing that reads correct at the resolver level and surprises you in a real browser.

pkarw:
Label set to `merge-queue` because code review passed; `needs-qa` stays on so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Priority (`priority-extreme`) and risk (`risk-high`) are unchanged and still correct for a staff-auth bypass fix.

pkarw:

There is no Playwright E2E for the browser challenge flow — coverage is resolver-level plus HTTP-dispatcher. P0 below is exactly the gap.

### P0 — Staff login, MFA challenge, and session refresh

**Where to click**
- `/login` (with an MFA-enrolled staff account and with a non-MFA one)
- `/backend` typed directly into the URL bar while a challenge is on screen
- Profile → Security → MFA (to enroll TOTP on a test account)

**What to verify**
- Non-MFA account: log in → lands in `/backend` with normal access. Nothing about this flow should feel different from `develop`.
- MFA account: log in → the challenge panel replaces the login form; a wrong OTP is rejected with an error and the panel stays; the correct OTP lands you in `/backend` with full navigation and data.
- **New in this round:** tick "remember me" and log in on an account that has MFA enrolled. When the challenge appears, open devtools → Application → Cookies and confirm `session_token` is **gone** (the challenge response clears it). `auth_token` should be present but is the provisional one.
- **New in this round:** with the challenge still on screen, type `/backend` in the URL bar and hit enter. You must land on `/login`, never inside the admin. Then log in again and complete the challenge — access is normal afterwards.
- After completing a challenge, leave the tab idle past the access-token lifetime (or clear `auth_token` only) and navigate to `/backend`: a normal signed-in user must still be refreshed back into the admin rather than bounced to `/login`. This is the fail-open direction and the one most likely to regress.
- Recovery-code login: use a recovery code instead of the OTP → completes and grants normal access.

**What can go wrong**
- The challenge panel makes an API call that is not on the pending allowlist → the dispatcher 401s and clears cookies mid-challenge, so the OTP submit fails with a confusing error. Watch the network tab during the challenge for any 401.
- Over-tightening: a *verified* user gets bounced to `/login` on every navigation, or the "remember me" refresh stops working for accounts with no MFA at all.
- The `/backend` → `/login` bounce loops instead of settling on the login page.

### P1 — Logout and session housekeeping during a pending challenge

**Where to click**
- `/login` (start a challenge, then abandon it — close the panel / hit "back")
- Settings → Security → active sessions, if your build surfaces them

**What to verify**
- Abandoning a challenge and then logging in again works cleanly, with no stale-session error.
- Clicking logout while a challenge is outstanding leaves the browser logged out (cookies cleared) even though the logout API itself is rejected during the pending window — this is expected and documented.

**What can go wrong**
- A visible error toast on logout during the pending window (the browser should end up logged out quietly).
- Abandoned challenges leaving a `Session` row that later behaves as if it were live — it should be inert; flag it if you can get any access from one.

### P2 — SSO login (regression check only, not changed here)

**Where to click**
- The SSO/OIDC login entry point, if configured in your QA environment

**What to verify**
- SSO login still completes and lands in `/backend` exactly as before. Note that SSO logins do **not** currently present an MFA challenge — that is a known, documented residual, not a regression from this PR. Just confirm nothing about SSO login broke.

**What can go wrong**
- The SSO callback's `session_token` handling colliding with the new refresh gate and bouncing SSO users to `/login`.

### Pass/fail

pkarw:

Autofix ran **1 iteration** covering all 7 findings from review round 3 (1 High, 2 Medium, 4 Low), pushed as a single follow-up commit `eac490f96` onto the fork branch (`maintainerCanModify: true`, no force-push, original commits and authorship preserved). Local gate green: typecheck 26/26, core auth 720/720, shared auth 139/139, enterprise security 305/305, i18n sync clean, lint 0 errors. The four new/changed test cases were verified to fail against the pre-fix code.

Labels: `merge-queue` + `needs-qa` (QA-approval gate holds the merge until a QA reviewer adds `qa-approved`). `ci-monitoring` is on while the full CI gate re-runs on the new head — that is not a claim, the PR is free for anyone to pick up.

pkarw:
✅ CI is fully green on `eac490f96`: **24 passing, 5 skipped by workflow conditions, 0 failing, 0 pending** — including all 15 `ephemeral-integration` shards, `test`, `prepare`, `lint`, `ds-lint`, and the `mutate` run over the three changed `packages/shared/src/lib/auth` files.

I opened the shard 15/15 log rather than trusting the summary, and confirmed all five `TC-SEC-014` cases actually executed:

```
✓ 62 pending bearer token gets 401 … (67ms)
✓ 63 pending cookie credentials get 401 with staff auth cookies cleared … (67ms)
✓ 64 completion routes stay reachable … and the verified replacement regains access (665ms)
✓ 65 exhausted challenge is cleaned up … (217ms)
✓ 66 a remember-me session_token planted before enrollment cannot be traded for a full staff token while MFA is pending (340ms)
```

Case 66 is the new one covering this round's High finding, and it is the one piece I could not run locally (needs a live app + DB), so this is its first real execution.

Removing `ci-monitoring`. The PR is approved, green, and in `merge-queue`. The only thing left is manual QA — `needs-qa` without `qa-approved` is a hard merge block by design, and code-review approval does not satisfy it. QA routes are in the comment above.

PR #5210, opened 2026-08-11
fix(security): bind sudo challenges to initiated policy (#5177)

haxiorz:
Closes #5177

## 🎯 Goal
- Prevent a sudo challenge initiated for one target or policy from authorizing a different, stronger-protected target.

## 🔍 Problem
Enterprise sudo initiation selected the challenge method for target A, but the pending session did not retain the target, request scope, or resolved configuration. Verification accepted target B from the request and could mint a token for B after authenticating with A's weaker method.

## 🔍 Root Cause
`SudoChallengeService.initiate()` persisted only the user, tenant, pending token, method, and expiry. `verify()` independently trusted and resolved caller-supplied target and scope values, while successful password sessions had no atomic consumed state and could be replayed.

## What Changed
- Persist the initiated target, exact tenant/organization scope, resolved configuration ID/version, and verification state on sudo sessions.
- Reject target and scope substitution before invoking credential verification.
- Re-resolve the effective policy under a transaction-bound configuration-table lock immediately before token issuance.
- Atomically consume pending sessions so concurrent and sequential verification attempts have exactly one winner.
- Require durable target, scope, and verified-state bindings during signed-token validation.
- Add the nullable enterprise migration, generated schema snapshot, and security specification updates.
- Add regressions for target/scope substitution, policy drift, legacy sessions, durable token binding, and concurrent replay.

## 🧪 Tests
- `yarn workspace @open-mercato/enterprise test --runInBand packages/enterprise/src/modules/security/services/__tests__/SudoChallengeService.test.ts` — 34/34 passed.
- `yarn workspace @open-mercato/enterprise build` — passed.
- `yarn build:packages` → `yarn generate` → `yarn build:packages` — passed.
- `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn lint`, and `yarn build:app` — passed.
- Enterprise-enabled ephemeral migration, initialization, build, and app readiness — passed.
- Independent high-risk security review — passed with no blockers after remediation.
- The aggregate `yarn test` runner encountered unrelated local sandbox/worker flakes; every reported non-enterprise failing suite passed in isolation, and the affected enterprise suite remained green.

## 💥 Breaking Changes
- None to public routes, HTTP methods, request/response shapes, event IDs, import paths, or registered sudo target contracts.
- The database migration adds six nullable columns. Existing short-lived unbound sudo sessions intentionally fail closed and must be restarted after deployment.

## Review-request coverage
[...]

PR #5458, opened 2026-08-21
security(enterprise): harden MFA emergency bypass with audit logging

### LINKED ISSUE #3856: security(ent_security): emergency MFA bypass flag disables the second factor tenant-wide when enabled (lib/security-config.ts:124-127)
The `OM_SECURITY_MFA_EMERGENCY_BYPASS` env flag, when truthy, neuters MFA: it makes `resolveChallengeMethod` always return `'password'` and disables the MFA-enrollment redirect. **LATENT / operator-controlled** — it defaults false (fail-closed) and is reachable only via an environment variable set by an operator, not by any tenant user. Flagged for awareness and hardening (audit/alerting), not because a tenant principal can trigger it.

## Location(s)
- `packages/enterprise/src/modules/ent_security/lib/security-config.ts:124-127` — `OM_SECURITY_MFA_EMERGENCY_BYPASS` truthy makes `resolveChallengeMethod` return `'password'`.
- `packages/enterprise/src/modules/ent_security/services/SudoChallengeService.ts:577-579` — bypass applied to sudo step-up.
- `packages/enterprise/src/modules/ent_security/lib/enforcement-redirect.ts:64` — bypass disables the MFA-enrollment redirect.
- `packages/enterprise/src/modules/ent_security/lib/security-config.ts:67` — default false (fail-closed default).

## Why it matters
When the env flag is truthy, the second factor is disabled instance-wide: challenge resolution always returns `'password'` and enforcement-redirect is skipped. This is an intentional break-glass control with a safe default (false) reachable only by an operator with env access — no tenant user can set it. The risk is operational: an active bypass that is forgotten or set without oversight silently removes MFA for everyone.

## Proposed fix
Keep the default false. Emit an audit-log entry / operational alert whenever the bypass is active (e.g. at startup and on each affected challenge), and document clearly that enabling it disables MFA enforcement platform-wide so it is not left on unnoticed.

## BC / Safety
No behavioral change to the bypass semantics — only added observability (logging/alerting) and documentation. Verify the alert fires exactly when the flag is active and does not leak the flag state to unauthorized surfaces. `risk-low` — logging/documentation-only change. Confined to the `ent_security` enterprise module.

haxiorz:
Fixes #3856

**Problem**
`OM_SECURITY_MFA_EMERGENCY_BYPASS` when truthy disables MFA platform-wide (forces sudo to `password`, skips login MFA challenge injection and enrollment redirect). It defaults to `false` and is operator-controlled, but was silent when active � a forgotten break-glass flag could leave MFA disabled without visibility.

**Solution**
Keeps the bypass semantics and fail-closed default. Adds observability and docs only:

- `packages/enterprise/src/modules/security/lib/security-config.ts`: central `emitMfaEmergencyBypassActiveWarning` + `emitMfaEmergencyBypassStartupWarningIfNeeded` (warn-level structured log via `createLogger('security')`, once-per-process at import). Startup warning when the flag is active.
- `enforcement-redirect.ts`, `api/interceptors.ts`, `SudoChallengeService.ts`: emit a warn on every bypassed challenge (`mfa-enrollment-redirect bypassed`, `login MFA challenge bypassed`, `sudo challenge downgraded to password`) with context (userId/pathname/method). No state returned to clients.
- `apps/mercato/.env.example` + `packages/create-app/template/.env.example` + spec env block: break-glass comment (`DISABLED platform-wide` + log notice).
- Test: `enforcement-redirect.test.ts` mock updated to include `warn`.

**Verification**
- `yarn workspace @open-mercato/enterprise jest` � `security-config`, `enforcement-redirect`, `interceptors`, `SudoChallengeService` all pass; bypass paths emit `WARN [security:mfa-emergency-bypass]`.
- `yarn generate` + `yarn build:packages` + `yarn typecheck` � 26 tasks pass.
- No DB/migration or API surface change.

**Risk**
Low � logging/docs only. No change to bypass logic or client payloads.


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T01:16:49Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: security(enterprise): harden MFA emergency bypass with audit logging

## 🎯 Summary
This PR keeps the emergency-bypass default and enforcement semantics unchanged, adds startup and per-path structured warnings, and mirrors the break-glass documentation into both application templates. The implementation uses the shared logging facade and the environment documentation is kept in parity. However, the per-path warnings currently record bypasses that did not actually change the MFA outcome, and the PR does not test the warning behavior that constitutes the security fix.

## Verdict
❌ request changes — the audit signal is inaccurate on realistic enrollment and sudo paths, and the new security-observability behavior lacks regression and integration coverage.

## 🧪 Validation Gate

Validation source: GitHub checks on head `bc121259507ed0edfd730b5a658586f415ae90b1` (GitHub Actions runner). Branch protection did not expose a required-check list, so all reported checks were treated as the available CI gate. No local fallback was needed.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | GitHub `prepare` job, “Build packages”. |
| `yarn generate` | ✅ PASS | GitHub `prepare` job, “Prepare generated modules”. |
| `yarn build:packages` | ✅ PASS | GitHub `prepare` job, “Rebuild packages with generated files”. |
| `yarn i18n:check-sync` | ✅ PASS | GitHub `test` job, “Check i18n sync”. |
| `yarn i18n:check-usage` | ✅ PASS | GitHub `test` job, “Check i18n usage”. |
| `yarn typecheck` | ✅ PASS | GitHub `test` job, “Checking types”. |
| `yarn test` | ✅ PASS | GitHub `test` job, “Test”. |
| `yarn build:app` | ✅ PASS | GitHub `prepare` job, “Build app”. |
| `yarn template:sync` | ✅ PASS | GitHub `test` job, “Check create-app template parity”. |

## Findings

### ⚠️ Major

1. `packages/enterprise/src/modules/security/lib/enforcement-redirect.ts:96` and `packages/enterprise/src/modules/security/services/SudoChallengeService.ts:577` — The warnings do not identify actual MFA bypasses. Enrollment middleware calls `resolveMfaEnrollmentRedirect` on every authenticated non-exempt route, but the bypass branch logs “mfa-enrollment-redirect bypassed” before checking whether enforcement applies, whether the user is compliant, or whether the deadline would redirect them. Similarly, sudo logs “downgraded to password” even when the target is already configured for `PASSWORD`, or `ANY` has no available MFA method and would naturally choose password. With the break-glass flag enabled this creates high-volume false audit records and makes the security signal unreliable. Determine the normal outcome first and emit the per-challenge warning only when the flag changes that outcome; keep the startup warning as the process-wide “bypass is active” alert. Add negative tests for compliant/unenforced enrollment and password-only sudo targets.

2. `packages/enterprise/src/modules/security/lib/security-config.ts:188` and `packages/enterprise/src/modules/security/lib/__tests__/enforcement-redirect.test.ts:200` — The PR's only new runtime behavior is warning emission, but no test asserts it. The enforcement test merely adds and clears a `warn` mock, while the existing interceptor and sudo tests still assert only the old bypass semantics; `resetMfaEmergencyBypassStartupWarningForTests` is not used anywhere. The suite would remain green if every new warning call were removed, so it does not satisfy issue #3856's requirement to verify that alerts fire exactly when active. Add focused assertions for inactive/no-warning behavior, the once-per-process startup warning, each actual login/enrollment/sudo bypass with its structured fields, and non-bypassed negative cases. Because this touches the authentication path, add an integration-level assertion for the observable login-bypass warning or document and implement an equivalent executable cross-boundary check.

### 🔹 Minor

1. `packages/enterprise/src/modules/security/lib/security-config.ts:179` — The helper interpolates `context` into the log message even though the structured-logging contract requires a stable, un-interpolated message with dynamic values in fields. This fragments aggregation by message and makes alert queries unnecessarily brittle. Use a constant message such as `MFA emergency bypass is active` and retain `context` in the structured fields.

2. `.ai/specs/enterprise/implemented/SPEC-ENT-001-2026-02-17-security-module-enterprise-mfa.md:1657` — The implemented spec is updated, but its dated implementation history/changelog has no entry for this security-hardening delta. Add a 2026-08-22 entry describing the emergency-bypass warning behavior, exact coverage, and documentation parity so the implemented spec remains auditable under `.ai/specs/AGENTS.md`.

## 🔍 Inherited Reviewer Feedback

None — no unaddressed review bodies, conversation comments, or inline review comments were present on the current head. The automation's own claim marker was excluded.

## 💥 Breaking Changes

- [x] No exported/public symbol was removed or renamed without a deprecation path.
- [x] No function signature was changed in a breaking way.
- [x] No HTTP route URL or method was changed.
- [x] No existing response field was removed or retyped.
- [x] No event or message name or payload contract was removed.
- [x] No CLI command, flag, or machine-parsed output was changed.
- [x] No database table, column, index, or standard column contract was changed.
- [x] No existing config key was renamed and the bypass default remains false.
- [x] The change is additive; no deprecation bridge is required.

## 🧪 Test Coverage

Existing tests cover the pre-existing bypass outcomes for login, enrollment redirect, and sudo, and GitHub's unit, integration, CodeQL, lint, build, i18n, and template-parity checks pass. They do not assert the new warning side effects, their exactly-once/only-when-applicable semantics, or the structured fields. The requested unit and auth-path integration assertions are required before this security-observability fix is adequately protected.

Autofix was skipped because this is another author's PR and the invocation did not include `--autofix`.

pkarw:

_Updated on the second re-review of head `517b7f6f` — the review is now **approved**._

- ✅ `merge-queue` — **replaces `changes-requested`.** The one Minor from the previous pass (the sudo bypass warning could not be attributed to an actor) is resolved and verified against the diff, no new blocker, major or minor was found, and all 13 non-skipped CI checks are green on this exact head. Only three nits remain, and this repository's rule approves on nits alone.
- ⏭️ `skip-qa` — kept. The automated-verification exemption in `AGENTS.md` still applies: no UI-rendering file is touched, the database structure and API surface are unchanged, no `BACKWARD_COMPATIBILITY.md` contract is broken, and automated tests for the changed behavior ship in the PR. Manual browser QA also has nothing to observe — the whole deliverable is a server-side warn log behind an env flag that is off by default, and no bypass state is returned to clients.
- 🐛 `bug` — kept. The PR closes issue #3856, a defect in the operational visibility of the break-glass flag.
- 🔒 `security` — kept. The change is on the MFA/auth enforcement path.
- 🏢 `enterprise` — kept. Everything but the two `.env.example` files and the spec lives in `packages/enterprise`.
- 🔴 `priority-high` — kept. A silent, forgotten break-glass flag leaving MFA disabled platform-wide is a release-blocking visibility gap.
- 🔴 `risk-high` — kept. The blast radius is the auth path: login MFA challenge injection, the enrollment enforcement redirect, and sudo step-up. Preserved deliberately even though the diff is logging-only, because the surface — not the size of the change — is what this label measures. Mitigating evidence is in the review body: every flag-off path is behaviourally identical to `develop`, and every flag-on path in `resolveMfaEnrollmentRedirect` still returns `null`.

`in-progress` is removed with the lock release comment below. No `ci-monitoring` is applied: CI is fully settled on this head, so no follow-up is owed.

pkarw:

pkarw:

The review found inaccurate/noisy per-path audit warnings plus missing regression and auth-path integration coverage. Autofix: skipped (not my PR — re-run with --autofix to fix it here). GitHub checks were green on the reviewed head, but the code findings remain actionable and the PR has been handed back to @haxiorz.

pkarw:

pkarw:

The review found inaccurate/noisy per-path audit warnings plus missing regression and auth-path integration coverage. Autofix: skipped (not my PR — re-run with --autofix to fix it here). GitHub checks were green on the reviewed head, but the code findings remain actionable and the PR has been handed back to @haxiorz.

haxiorz:
Addressed review feedback:

- Stable log message: emit now uses constant \MFA emergency bypass is active\ with \context\ in structured fields.
- Enrollment: redirect now checks enforcement/compliance/deadline first and only warns when bypass actually prevents a redirect (skips compliant, unenforced, non-overdue, exempt, tenant-less).
- Sudo: downgrade warning only when normal flow would have been \mfa\ (handles ANY with 0 methods as no-warning, MFA-required without methods as warning and still returns password instead of throwing).
- Login interceptor: warning moved after JWT claims and service validation, includes \userId\, and skips invalid token / missing service.
- Tests: added coverage for inactive/no-warning, startup once, per-path structured fields, and negative cases for all three paths plus Sudo/Interceptor edge cases. Full \@open-mercato/enterprise\ suite: 507 tests pass.
- Spec: added 2026-08-22 changelog entry.

Re-ran \yarn generate\ + \uild:packages\ + \	ypecheck\ (26 tasks pass) and verified warnings only on true bypass.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T06:24:40Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: Re-review — security(enterprise): harden MFA emergency bypass with audit logging

## 🎯 Summary

This re-review covers head `c9ec3398f7a67d63108ae12148dc9b1da0ed24c0`, which adds two commits on top of the previously reviewed `bc121259`. Both majors and both minors from the first pass are genuinely fixed, not papered over: the per-path warnings now fire only when the break-glass flag actually changes the MFA outcome, the log message is a stable constant with `context` moved into structured fields, the warning behavior is covered by focused unit tests on all three bypass paths including negative cases, and the implemented spec carries a dated changelog entry.

I also verified the property that matters most for a change on the auth path: **with `OM_SECURITY_MFA_EMERGENCY_BYPASS` off — the default and the state of every real deployment — the runtime behavior is identical to `develop`.** In `enforcement-redirect.ts` the removed early return only ever fired when the flag was on; in `api/interceptors.ts` the relocated check is false and the flow is unchanged; in `SudoChallengeService.ts` the extracted `resolveChallengeMethodWithoutBypass` reproduces the original branch logic verbatim. The flag-on path also correctly preserves the old semantics, including the `ChallengeMethod.MFA`-with-zero-methods case where the un-bypassed path throws a 400 and the bypassed path still returns `password` (commit `c9ec3398`).

One finding remains: the sudo bypass warning is the only one of the three that cannot be attributed to a user.

## Verdict

❌ request changes — the sudo step-up audit record omits the actor, so the highest-privilege bypass path is the one this feature cannot attribute. Everything else is ready; this is a one-line addition.

## 🧪 Validation Gate

**Validation source: GitHub checks.** Workflow run [32543870213](https://github.com/open-mercato/open-mercato/actions/runs/32543870213) reports `head_sha = c9ec3398f7a67d63108ae12148dc9b1da0ed24c0`, an exact match for the reviewed head, and every one of the 14 non-skipped checks concluded `SUCCESS`. No local fallback was needed, and no check on this head is pending.

| Command | Status | Evidence |
|---------|--------|----------|
| `yarn build:packages` | ✅ PASS | `prepare` job, step “Build packages”. |
| `yarn generate` | ✅ PASS | `prepare` job, step “Prepare generated modules”. |
| `yarn build:packages` (rebuild) | ✅ PASS | `prepare` job, step “Rebuild packages with generated files”. |
| `yarn i18n:check-sync` | ✅ PASS | `test` job, step “Check i18n sync”. |
| `yarn i18n:check-usage` | ✅ PASS | `test` job, step “Check i18n usage”. |
| `yarn typecheck` | ✅ PASS | `test` job, step “Checking types”. |
| `yarn test` | ✅ PASS | `test` job, step “Test”. |
| `yarn build:app` | ✅ PASS | `prepare` job, step “Build app”. |
| `yarn template:sync` | ✅ PASS | `test` job, step “Check create-app template parity (always, unfiltered)”. |

Supporting checks also green on this head: `lint`, `ds-lint`, `scope`, `audit-scope`, `docker-build`, `documents-multi-instance`, `ephemeral-integration (none)`, `CodeQL` (all three analyses), `license/cla`.

## ✅ Previous Findings — Resolution

- **Major 1 — warnings did not identify actual bypasses.** ✅ Resolved. `enforcement-redirect.ts:116-134` now evaluates `checkUserCompliance` and `resolveDeadlineRedirectState` *before* the bypass branch, so a compliant, unenforced, non-overdue, exempt-path or tenant-less user produces no warning. `SudoChallengeService.ts:577-593` computes the un-bypassed outcome first and warns only when it would have been `mfa`. `api/interceptors.ts:66-70` moved the check after claim parsing and service resolution, so an invalid token or a missing `mfaVerificationService` no longer logs a phantom bypass. Negative tests exist for every one of those cases.
- **Major 2 — no test asserted the new runtime behavior.** ✅ Resolved. `lib/__tests__/security-config.test.ts` is new and covers the stable message, the structured fields, the once-per-process startup warning, silence when the flag is inactive, and the test reset hook. `enforcement-redirect.test.ts` adds seven cases, `interceptors.test.ts` three, `SudoChallengeService.test.ts` two — each asserting both the positive emission with its fields and the matching negative case. Removing any warning call would now fail the suite, which was the specific gap I raised.
- **Minor 1 — interpolated log message.** ✅ Resolved. `security-config.ts:178` emits the constant `'MFA emergency bypass is active'` with `context` as a structured field, and `security-config.test.ts:138` asserts the message is not interpolated.
- **Minor 2 — missing spec changelog entry.** ✅ Resolved. A `2026-08-22 — Harden MFA emergency bypass with audit logging` entry documents the behavior, the per-path contexts and fields, and the coverage.

On the integration-coverage half of Major 2, I am satisfied without a Playwright test. The observable is a server-side log line that the PR deliberately never surfaces to clients, so a browser test cannot assert it; `interceptors.test.ts` exercises the real exported `interceptors` array against a genuinely signed JWT and a container, which is the equivalent executable cross-boundary check my original finding offered as the alternative.

## Findings

### 🔹 Minor

1. `packages/enterprise/src/modules/security/services/SudoChallengeService.ts:585` — The sudo bypass warning omits the actor and the target, so this audit record cannot be attributed. The login path logs `userId` (`api/interceptors.ts:67`) and the enrollment path logs `userId` and `pathname` (`lib/enforcement-redirect.ts:104`), but the sudo warning carries only `configuredMethod` and `availableMfaMethodCount`. Sudo step-up guards the most privileged operations in the system, so during an incident review — the exact scenario issue #3856 exists to serve — this is the one line that cannot answer "who elevated while MFA was off, and into what?". Correlating by timestamp across concurrent requests is not a substitute, and the logging facade attaches no request or user context automatically (`packages/shared/src/lib/logger/` has no async-local-storage binding). Both values are already in scope at the call site: `initiate()` holds `userId` and `targetIdentifier` at `SudoChallengeService.ts:172-178`. Thread them into `resolveChallengeMethod` (or move the `emit` call up to `initiate`) and add `userId` plus the sudo target to the warning fields, then extend the existing assertion at `services/__tests__/SudoChallengeService.test.ts:346` to cover them.

### 💤 Nits

1. `packages/enterprise/src/modules/security/services/__tests__/SudoChallengeService.test.ts:362-364` — A three-line inline comment was committed that contradicts itself: it opens by describing seeding "a custom config with PASSWORD method", then says "Instead we directly test the downgrade logic". `AGENTS.md` asks for self-documenting code without inline commentary, and a comment that describes an abandoned approach is worse than none. The test name at line 354 is misleading for the same reason — it says "sudo target already requires password" but the case it actually exercises is an `ANY`/auto target with zero available MFA methods. Renaming it to something like `does not warn when no MFA methods are available so password was already the outcome` removes the need for the comment entirely.
2. `packages/enterprise/src/modules/security/lib/__tests__/enforcement-redirect.test.ts:1-2` and the matching lines in `interceptors.test.ts` / `security-config.test.ts` — The mock handles are declared with `var`. `enforcement-redirect.test.ts` previously used `const mockLoggerError` and worked; the `mock`-prefixed name is what satisfies the hoisting allowlist, and the emitted `require` for the module under test still runs after the assignment, so `const` remains correct here and keeps the file consistent with the rest of the suite.
3. `packages/enterprise/src/modules/security/services/SudoChallengeService.ts:581-583` — The `catch` fallback sets `wouldHaveBeenMfa = configuredMethod === ChallengeMethod.MFA`, but the only throw `resolveChallengeMethodWithoutBypass` can produce is that exact case, so the branch is unreachable-by-construction defensive code that restates its own guard. Harmless, but it reads as if a second failure mode exists.

## 🔍 Inherited Reviewer Feedback

None outstanding. The only prior feedback on this PR is my own first-pass review, accounted for in the resolution section above. @haxiorz's [summary comment](https://github.com/open-mercato/open-mercato/pull/5458#issuecomment-5377132644) accurately describes what was changed; I verified each claim in it against the diff rather than taking it at face value, and every item holds.

## 💥 Breaking Changes

- [x] No exported or public symbol was removed or renamed.
- [x] No function signature was changed in a breaking way — `emitMfaEmergencyBypassActiveWarning`, `emitMfaEmergencyBypassStartupWarningIfNeeded` and `resetMfaEmergencyBypassStartupWarningForTests` are additive exports, and `resolveChallengeMethodWithoutBypass` is private.
- [x] No HTTP route URL, method, or response field was changed.
- [x] No event, widget spot, DI key, ACL feature, notification ID, or CLI command was changed.
- [x] No database table, column, index, or standard column contract was changed; no migration ships or is needed.
- [x] No config key was renamed, and `OM_SECURITY_MFA_EMERGENCY_BYPASS` keeps its fail-closed `false` default.
- [x] The change is additive; no deprecation bridge is required.

## 🧪 Test Coverage

Coverage is now proportionate to the change. The three bypass call sites each have a positive test asserting the structured fields and at least one negative test proving silence when the flag does not alter the outcome, plus `security-config.test.ts` covering the emitter and the once-per-process startup warning directly. `grep` for `emergencyBypass` confirms there is no fourth bypass site left unwarned. The only coverage gap left is the one named in the Minor finding: the sudo assertion at `SudoChallengeService.test.ts:346` matches on `availableMfaMethodCount` alone, so it would not catch the missing attribution fields once they are added.

## 📋 QA

The `needs-qa` label is being replaced with `skip-qa` on this pass. The PR touches no UI-rendering file (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`), leaves the database structure and API surface unchanged, breaks no `BACKWARD_COMPATIBILITY.md` contract, and now ships automated tests for the changed behavior — which is exactly the automated-verification exemption in `AGENTS.md`, and the coverage condition is the one that was missing on the first pass. Manual browser QA would also have no observable to check here: the entire deliverable is a server-side warn log behind an env flag that is off, and the PR deliberately returns no bypass state to clients.

---

Autofix was skipped because this is another author's PR and the invocation did not include `--autofix`.

pkarw:
Thanks @haxiorz — the re-review confirms your fixes landed: both majors and both minors from the first pass are genuinely resolved, and I verified each claim in your summary against the diff rather than taking it at face value. The warnings are now accurate, the coverage would fail if any warning call were removed, and the flag-off path is behaviorally identical to `develop` on all three files.

One Minor finding is left, and it is a one-line addition: the sudo bypass warning in `SudoChallengeService.ts:585` carries only `configuredMethod` and `availableMfaMethodCount`, so unlike the login and enrollment warnings it cannot be attributed to a user. `userId` and `targetIdentifier` are already in scope at `initiate()` (lines 172-178) — thread them through and extend the assertion at `SudoChallengeService.test.ts:346`. Three nits are listed in the review body if you want to pick them up in the same push.

I also swapped `needs-qa` for `skip-qa`: with the automated coverage now in the PR, the automated-verification exemption applies, so manual QA is no longer gating this one. Handing the PR back to you — push the update and re-request review, and the automation can pick it up from the new head.

pkarw:

Re-review of head `c9ec3398`. All four previous findings (2 major, 2 minor) verified as resolved. One new Minor remains — the sudo bypass warning cannot be attributed to a user — plus three nits. Validation source: GitHub checks on the exact head SHA (14/14 green, run 32543870213); no local fallback needed and no pending checks, so no CI follow-up is owed. Labels: `changes-requested` kept, `needs-qa` → `skip-qa` under the automated-verification exemption. Autofix: skipped (not my PR — re-run with --autofix to fix it here). PR handed back to @haxiorz.

haxiorz:
Fixed remaining minor:

- Sudo warning now includes \userId\ and \	argetIdentifier\ (threaded from \initiate()\ into \
esolveChallengeMethod\); assertion in \SudoChallengeService.test.ts:346\ updated to cover them.
- Test nits: removed misleading inline comment and renamed test to \does not warn when no MFA methods are available so password was already the outcome\; changed \ar mockLoggerWarn\ back to \const\ (mock-prefixed allowlist) in the three security test files.

Re-verified: \@open-mercato/enterprise\ 61 tests pass for the four security suites, full suite 507 pass, \	ypecheck\ 26/26.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-23T00:26:55Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review: Re-review (2) — security(enterprise): harden MFA emergency bypass with audit logging

## 🎯 Summary

This re-review covers head `517b7f6fe5f53b25ab435f0035d9f1a9b525588d`, one commit on top of the previously reviewed `c9ec3398`. The single Minor from the last pass — the sudo bypass warning being the only one of the three that could not be attributed to an actor — is genuinely fixed, and two of the three nits were picked up in the same push.

I re-verified the property that matters most for a change on the auth path rather than assuming the earlier pass still holds: **with `OM_SECURITY_MFA_EMERGENCY_BYPASS` off — the default and the state of every real deployment — the runtime behaviour is byte-for-byte the behaviour on `develop`.** I also traced every return path in `resolveMfaEnrollmentRedirect` with the flag on: lines 101, 108, 118, 121, 131, 143 and 148 all return `null`, and no path reaches `createEnrollmentRedirect`, so the relocated check preserves the old early-return semantics exactly while now being able to report *why* the redirect was suppressed.

## Verdict

✅ **Approve.** No blocker, major or minor finding remains. Three nits are listed below — one carried forward from the last pass and two new — and none of them blocks the merge; pick them up here or leave them.

## 🧪 Validation Gate

**Validation source: GitHub checks** (no local fallback needed).

The check runs bound to the exact reviewed head SHA `517b7f6f` report **13 success, 5 skipped, 0 failure, 0 pending**, so there is no missing CI signal to substitute for and no CI follow-up is owed on this PR.

| Command | Status | Evidence |
|---------|--------|----------|
| `yarn build:packages` | ✅ PASS | `prepare` job, step “Build packages”. |
| `yarn generate` | ✅ PASS | `prepare` job, step “Prepare generated modules”. |
| `yarn build:packages` (rebuild) | ✅ PASS | `prepare` job, step “Rebuild packages with generated files”. |
| `yarn i18n:check-sync` | ✅ PASS | `test` job. |
| `yarn typecheck` | ✅ PASS | `test` job, step “Checking types”. |
| `yarn test` | ✅ PASS | `test` job (8m22s). |
| `yarn build:app` | ✅ PASS | `prepare` job, step “Build app”. |
| `yarn template:sync` | ✅ PASS | `test` job, create-app template parity step. |

Supporting checks also green on this head: `lint`, `ds-lint`, `scope`, `audit-scope`, `docker-build`, `documents-multi-instance`, `ephemeral-integration (none)`, `CodeQL` (all three analyses), `license/cla`.

## ✅ Previous Findings — Resolution

- **Minor 1 — the sudo bypass warning could not be attributed.** ✅ Resolved. `SudoChallengeService.ts:178` now threads `userId` and `targetIdentifier` from `initiate()` into `resolveChallengeMethod`, and `SudoChallengeService.ts:587-592` emits both alongside `configuredMethod` and `availableMfaMethodCount`. The assertion at `services/__tests__/SudoChallengeService.test.ts:347-355` was extended to cover them, so removing either field now fails the suite. I confirmed `resolveChallengeMethod` has exactly one call site, so the sudo audit record is attributed on every path that can reach it. All three bypass warnings now carry `userId`, which was the point of the finding: an incident review can answer “who elevated while MFA was off, and into what?”.
- **Nit 1 — misleading test name and self-contradicting inline comment.** ✅ Resolved. The three-line comment is gone and the test is renamed to `does not warn when no MFA methods are available so password was already the outcome`, which now describes what the case actually exercises.
- **Nit 2 — `var` mock handles.** ✅ Resolved. `interceptors.test.ts:1`, `enforcement-redirect.test.ts:1-2` and `security-config.test.ts:1` are back to `const`, consistent with the rest of the suite. I confirmed this is safe rather than assuming it: the `mock`-prefixed names satisfy the hoisting allowlist, the mock factories only close over the handles instead of reading them, and the emitted `require` for the module under test still runs after the assignment — the green `test` check on this head is the empirical confirmation.
- **Nit 3 — redundant `catch` fallback in `resolveChallengeMethod`.** ⬜ Not addressed. Still open below; it was and remains non-blocking.

## Findings

No blockers, majors or minors.

### 💤 Nits

1. `packages/enterprise/src/modules/security/lib/enforcement-redirect.ts:102-109` and `:137-144` — Both bypass branches capture the underlying failure and then drop it. In the first, `enforcementService.error` holds the diagnostic `TypeError` built at line 77 (including the received-shape key list); in the second, `error` holds whatever `checkUserCompliance` threw. With the flag off, each of those is surfaced by `logger.error(..., { err })` at lines 110 and 145, but with the flag on the code returns before reaching them, so the operator learns only `reason: 'enforcement-service-unavailable'` and never why. This is not a regression — on `develop` the flag-on path returned at line 96 and never resolved the service at all — but a break-glass window is exactly when a broken enforcement service most deserves to be visible, and the value is already in hand. Adding `err: enforcementService.error` and `err: error` to the two warning payloads is a one-line change each.
2. `packages/enterprise/src/modules/security/services/SudoChallengeService.ts:576-577` — `userId` and `targetIdentifier` are declared optional, but the sole call site at line 178 always supplies them and both are non-optional in `initiate()`'s own signature. Making them required would let the compiler guarantee the attribution this PR just added, rather than letting a future second call site silently emit an unattributed audit line — the exact defect that was fixed here.
3. `packages/enterprise/src/modules/security/services/SudoChallengeService.ts:583-585` — Carried forward from the last pass. The `catch` sets `wouldHaveBeenMfa = configuredMethod === ChallengeMethod.MFA`, but the only throw `resolveChallengeMethodWithoutBypass` can produce is that exact case, so the branch restates its own guard. Harmless, and it reads as if a second failure mode exists.

### 🔬 Verified, not findings

Recording what I checked and cleared, so a later pass does not re-litigate it:

- **Client-bundle safety of the new module-scope logger.** `security-config.ts` is imported by the `'use client'` component `components/SudoConfigForm.tsx`, so the new `createLogger('security')` at line 11 and the self-invocation at line 200 now execute in the browser too. `selectTransport` guards on `typeof window !== 'undefined'` and falls back to the console transport, and `process.env.OM_SECURITY_MFA_EMERGENCY_BYPASS` is absent from the client bundle so the startup warning is silent there. The logging facade is already used from client components across the repo, so this introduces no new pattern.
- **The relocated checks do extra work while the flag is on.** With bypass on, the enrollment path now awaits `checkUserCompliance` on every backend navigation and the login interceptor now verifies the JWT and resolves a container service per login, where `develop` returned earlier. That is the deliberate price of the accuracy fix requested in the first pass, it only applies in break-glass mode, and the outcome is unchanged in every case.
- **No warning on the missing-service / invalid-token paths** (`api/interceptors.ts:64`, `:67`). Correct rather than a gap: without valid claims or a resolvable `mfaVerificationService` no challenge would have been injected regardless of the flag, so a warning there would be the phantom-bypass defect from the first pass. Both cases have explicit negative tests.
- **Base staleness.** The head is 19 commits behind `develop`, which includes `00b63eb3` (“bind sudo challenges to initiated policy”, #5210) touching the same `SudoChallengeService.ts` with 192 changed lines. I simulated the three-way merge per-file rather than trusting the `MERGEABLE` flag: it resolves with zero conflicts, a single `security-config` import, one call site at merged line 171, and `targetIdentifier` still in scope there. No semantic merge hazard.
- **Template parity.** The four added lines in `apps/mercato/.env.example` are identical to those in `packages/create-app/template/.env.example`, satisfying the `AGENTS.md` template-sync rule, and the `template:sync` CI step agrees.

## 🔍 Inherited Reviewer Feedback

None outstanding. The only prior feedback on this PR is my own two review passes, fully accounted for in the resolution section above; there are no inline diff comments and no third-party reviews. @haxiorz's [summary comment](https://github.com/open-mercato/open-mercato/pull/5458#issuecomment-5382698521) describes the change accurately — I verified each of its claims against the diff rather than taking it at face value, and every item holds, including the two test nits it reports as fixed.

## 💥 Breaking Changes

- [x] No exported or public symbol was removed or renamed.
- [x] No function signature was changed in a breaking way — the two new parameters on `resolveChallengeMethod` are optional and the method is `private`.
- [x] No HTTP route URL, method, or response field was changed; no bypass state is returned to clients.
- [x] No database schema, migration, event ID, widget spot, DI key, ACL feature, notification ID, or CLI command was changed.
- [x] Environment-variable semantics are unchanged: `OM_SECURITY_MFA_EMERGENCY_BYPASS` keeps its fail-closed `false` default and its existing bypass behaviour.

## 🧪 Test Coverage

Coverage is proportionate to the change and would fail if the feature were removed. `security-config.test.ts` (new) asserts the stable message, the structured `context` field, the once-per-import startup warning, silence when the flag is inactive, and the reset hook. `enforcement-redirect.test.ts` covers all four flag-on branches plus compliant, unenforced, non-overdue, exempt-path and flag-off negatives. `interceptors.test.ts` exercises the real exported `interceptors` array against a genuinely signed JWT, with negatives for an invalid token, a missing verification service and the flag off. `SudoChallengeService.test.ts` covers the attributed downgrade, the no-MFA-methods no-warning case and the flag-off case. No Playwright test is warranted: the observable is a server-side log line the PR deliberately never surfaces to clients, which is also why `skip-qa` is the right gate here.

pkarw:

Second re-review, head `517b7f6f`. The remaining Minor from the previous pass — the sudo bypass warning could not be attributed to an actor — is genuinely fixed (`userId` and `targetIdentifier` threaded from `initiate()` and asserted in the test), and two of the three nits were picked up in the same push. No blocker, major or minor was found on this pass; three nits remain and none blocks the merge.

Validation source: GitHub checks on the exact head SHA (13 success, 5 skipped, 0 failing, 0 pending) — no local fallback needed and no CI follow-up is owed. I also simulated the three-way merge against the 19 newer `develop` commits (including #5210, which reworks the same `SudoChallengeService.ts`) rather than trusting the `MERGEABLE` flag: it resolves cleanly with no semantic hazard.

Labels: `changes-requested` → `merge-queue`; `skip-qa`, `priority-high`, `risk-high` and the category labels preserved. Autofix: not applicable — the review approved. Assignee left with @haxiorz.

pkarw:
... as a follow up please add the system wide notification to users with the ACL for managing the MFA that this situation happened so they know without the need of continous log monitoring

PR #5462, opened 2026-08-21
fix(customers): enforce organization scoping on legacy activities list (#3841)

### LINKED ISSUE #3841: security(customers): legacy activities list drops org filter when organizationIds is null (activities/route.ts:362)
`listLegacyActivities` builds `where = { tenantId }` and only adds the `organizationId: { $in: organizationIds }` clause when `organizationIds?.length > 0`. This is **LIVE** but gated: when a caller's `scope.filterIds` is null AND `auth.orgId` is null, `organizationIds` is null and the org filter is dropped entirely, so the query returns all `customer_activities` across every org in the tenant. This follows the platform's `filterIds === null` = "all allowed orgs" convention (fine for a superadmin / all-org principal), but a **non-superadmin** principal with no home org would receive a tenant-wide (cross-org) read on this deprecated endpoint.

## Location(s)
- `packages/core/src/modules/customers/api/activities/route.ts:362-365` — `where = { tenantId }`; `organizationId` clause added only when `organizationIds?.length > 0`

## Why it matters
The precondition is a principal whose resolved `scope.filterIds` is null and whose `auth.orgId` is also null. For a superadmin/all-org principal this is intended (they may read all orgs). But for a non-superadmin with no home org, `organizationIds` collapses to null, the org filter is silently omitted, and the caller reads every org's activities within the tenant — a cross-org read. Tenant scoping is always present (`resolveCustomersRequestContext` 401s on a missing tenant), so this is cross-org-within-tenant, not cross-tenant, and the endpoint is deprecated — but it is still an org-isolation gap for the null-scope path.

## Proposed fix
- When `organizationIds` is null and the principal is not a superadmin (i.e. `scope.allowedIds` is not explicitly "all"), deny the request or fall back to an empty result set instead of dropping the org filter.

## BC / Safety
Superadmin/all-org behavior is preserved. Non-superadmin null-org callers, who today get a tenant-wide read, would now get denied/empty — verify no legitimate consumer relies on that accidental breadth. Deprecated endpoint, so blast radius is contained. risk-medium as a single-module org-scoping fix. The same `organizationIds?.length > 0`-guarded org clause is a hotspot pattern worth grepping for elsewhere.

haxiorz:
Closes #3841

## 🎯 Goal
Enforce proper organization scoping on the deprecated `GET /api/customers/activities` endpoint to prevent cross-organization reads for non-superadmin callers with null or empty organization scope.

## 🔍 Problem
`GET /api/customers/activities` only attached `where.organizationId = { $in: organizationIds }` when `organizationIds && organizationIds.length > 0`. When a non-superadmin principal had `scope.filterIds: null` and `auth.orgId: null` (or `organizationIds: []`), the query dropped the organization filter entirely, resulting in a tenant-wide query that returned customer activities across all organizations in the tenant.

## 🔍 Root Cause
The legacy and canonical activity fetch helpers checked only `if (organizationIds && organizationIds.length > 0)` before applying the organization filter clause. For superadmins / unrestricted users (`auth.isSuperAdmin === true` or `scope.allowedIds === null`), omitting the filter is expected behavior, but for restricted callers with null or empty scope, this resulted in an unintended cross-organization read.

## What Changed
- `packages/core/src/modules/customers/api/activities/route.ts`:
  - Resolved `isUnrestricted = auth.isSuperAdmin === true || scope?.allowedIds === null`.
  - Gated the `GET` handler to return an empty paged result set immediately when `!isUnrestricted && (!organizationIds || organizationIds.length === 0)`.
  - Added fail-closed guards in `listLegacyActivities` and `listCanonicalActivities` returning empty result sets on null or empty scope for non-unrestricted callers.
  - Forwarded `isUnrestricted` to activity listing functions so superadmin / unrestricted all-org querying continues to work as intended.
- `packages/core/src/modules/customers/api/activities/__tests__/org-scoping.test.ts`:
  - Added comprehensive test suite verifying:
    - Non-superadmin with null `organizationIds` returns empty items.
    - Non-superadmin with empty array `organizationIds` returns empty items.
    - Superadmin with null `organizationIds` performs unrestricted tenant-wide query.
    - User with `scope.allowedIds === null` performs unrestricted tenant-wide query.
    - Principal with explicit `organizationIds` applies `{ $in: organizationIds }` filter.
    - Unified mode (`flags.unified = true`) correctly enforces org scoping.

## 🧪 Tests
- `yarn workspace @open-mercato/core test packages/core/src/modules/customers/api/activities/` (4 suites, 18 tests passed)
- `yarn build:packages && yarn generate && yarn typecheck && yarn build:app` (all passed)

## 💥 Breaking Changes
None. Superadmin and unrestricted principals continue to have tenant-wide read access; non-superadmin callers without an active organization scope receive empty results rather than cross-organization data.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5462"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789944251&installation_model_id=432243&pr_number=5462&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5462&signature=02b508d9377b8007faa2d6587b796eaed200614f48585d7ca7032280ca204f10"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/customers/api/activities/__tests__/org-scoping.test.ts
+++ b/packages/core/src/modules/customers/api/activities/__tests__/org-scoping.test.ts
@@ -1,0 +1,395 @@
+/** @jest-environment node */
+
+import { CustomerActivity } from '../../../data/entities'
+import { GET } from '../route'
+
+const TENANT_ID = '123e4567-e89b-41d3-a456-426614174010'
+const ORG_1 = '123e4567-e89b-41d3-a456-426614174001'
+
+const mockCommandBus = { execute: jest.fn() }
+const mockEm = {
+  find: jest.fn(),
+  findOne: jest.fn(),
+  findAndCount: jest.fn(),
+  count: jest.fn(),
+}
+
+const mockContainer = {
+  resolve: jest.fn((name: string) => {
+    if (name === 'commandBus') return mockCommandBus
+    if (name === 'em') return mockEm
+    if (name === 'queryEngine') return { query: jest.fn() }
+    throw new Error(`Unknown dependency: ${name}`)
+  }),
+}
+
+let mockContext: Record<string, unknown> = {}
+
+jest.mock('../../../lib/interactionFeatureFlags', () => ({
+  resolveCustomerInteractionFeatureFlags: jest.fn(),
+}))
+
+jest.mock('../../../lib/interactionRequestContext', () => ({
+  resolveCustomersRequestContext: jest.fn(async () => mockContext),
+}))
+
+jest.mock('../../../lib/interactionReadModel', () => ({
+  hydrateCanonicalInteractions: jest.fn(async () => []),
+  loadCustomerSummaries: jest.fn(async () => new Map()),
+}))
+
+jest.mock('@open-mercato/shared/lib/i18n/server', () => ({
+  resolveTranslations: jest.fn(async () => ({
+    translate: (_key: string, fallback: string) => fallback,
+  })),
+}))
+
+jest.mock('../../../lib/interactionCompatibility', () => ({
+  mapInteractionRecordToActivitySummary: jest.fn(),
+  CUSTOMER_INTERACTION_ACTIVITY_ADAPTER_SOURCE: 'adapter:activity',
+}))
+
+jest.mock('@open-mercato/shared/lib/encryption/find', () => ({
+  findWithDecryption: jest.fn(async () => []),
+}))
+
+describe('activities GET organization scoping', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+
+    const { resolveCustomerInteractionFeatureFlags } = jest.requireMock(
+      '../../../lib/interactionFeatureFlags',
+    )
+    resolveCustomerInteractionFeatureFlags.mockResolvedValue({
+      unified: false,
+      legacyAdapters: true,
+      externalSync: false,
+    })
+
+    mockEm.findAndCount.mockImplementation(async () => [[], 0])
+    mockEm.find.mockImplementation(async () => [])
+    mockEm.count.mockImplementation(async () => 0)
+
+    jest.requireMock('../../../lib/interactionReadModel').hydrateCanonicalInteractions
+      .mockImplementation(async () => [])
+  })
+
+  it('returns empty results and does NOT query all orgs when non-superadmin has organizationIds = null', async () => {
+    mockContext = {
+      auth: {
+        sub: 'user-1',
+        tenantId: TENANT_ID,
+        orgId: null,
+        isSuperAdmin: false,
+      },
+      em: mockEm,
+      organizationIds: null,
+      selectedOrganizationId: null,
+      scope: {
+        selectedId: null,
+        filterIds: null,
+        allowedIds: [ORG_1],
+        tenantId: TENANT_ID,
+      },
+      container: mockContainer,
+      commandContext: {
+        container: mockContainer,
+        auth: { sub: 'user-1', tenantId: TENANT_ID, orgId: null },
+        organizationScope: { selectedId: null, filterIds: null, allowedIds: [ORG_1], tenantId: TENANT_ID },
+        selectedOrganizationId: null,
+        organizationIds: null,
+        request: undefined,
+      },
+    }
+
+    const res = await GET(new Request('http://localhost/api/customers/activities?page=1&pageSize=50'))
+    expect(res.status).toBe(200)
+    const body = await res.json()
+    expect(body.items).toEqual([])
+    expect(body.total).toBe(0)
+
+    expect(mockEm.findAndCount).not.toHaveBeenCalled()
+    expect(mockEm.find).not.toHaveBeenCalled()
+    expect(mockEm.count).not.toHaveBeenCalled()
+  })
+
+  it('returns empty results and does NOT query all orgs when non-superadmin has organizationIds = []', async () => {
+    mockContext = {
+      auth: {
+        sub: 'user-1',
+        tenantId: TENANT_ID,
+        orgId: null,
+        isSuperAdmin: false,
+      },
+      em: mockEm,
+      organizationIds: [],
+      selectedOrganizationId: null,
+      scope: {
+        selectedId: null,
+        filterIds: [],
+        allowedIds: [ORG_1],
+        tenantId: TENANT_ID,
+      },
+      container: mockContainer,
+      commandContext: {
+        container: mockContainer,
+        auth: { sub: 'user-1', tenantId: TENANT_ID, orgId: null },
+        organizationScope: { selectedId: null, filterIds: [], allowedIds: [ORG_1], tenantId: TENANT_ID },
+        selectedOrganizationId: null,
+        organizationIds: [],
+        request: undefined,
+      },
+    }
+
+    const res = await GET(new Request('http://localhost/api/customers/activities?page=1&pageSize=50'))
+    expect(res.status).toBe(200)
+    const body = await res.json()
+    expect(body.items).toEqual([])
+    expect(body.total).toBe(0)
+
+    expect(mockEm.findAndCount).not.toHaveBeenCalled()
+    expect(mockEm.find).not.toHaveBeenCalled()
+    expect(mockEm.count).not.toHaveBeenCalled()
+  })
+
+  it('allows tenant-wide query when user is superadmin and organizationIds is null', async () => {
+    mockContext = {
+      auth: {
+        sub: 'superadmin-1',
+        tenantId: TENANT_ID,
+        orgId: null,
+        isSuperAdmin: true,
+      },
+      em: mockEm,
+      organizationIds: null,
+      selectedOrganizationId: null,
+      scope: {
+        selectedId: null,
+        filterIds: null,
+        allowedIds: null,
+        tenantId: TENANT_ID,
+      },
+      container: mockContainer,
+      commandContext: {
+        container: mockContainer,
+        auth: { sub: 'superadmin-1', tenantId: TENANT_ID, orgId: null, isSuperAdmin: true },
+        organizationScope: { selectedId: null, filterIds: null, allowedIds: null, tenantId: TENANT_ID },
+        selectedOrganizationId: null,
+        organizationIds: null,
+        request: undefined,
+      },
+    }
+
+    const res = await GET(new Request('http://localhost/api/customers/activities?page=1&pageSize=50'))
+    expect(res.status).toBe(200)
+
+    const legacyCall = mockEm.findAndCount.mock.calls.find(
+      (args) => args[0] === CustomerActivity,
+    )
+    expect(legacyCall).toBeDefined()
+    const where = legacyCall![1] as Record<string, unknown>
+    expect(where.tenantId).toBe(TENANT_ID)
+    expect(where.organizationId).toBeUndefined()
+  })
+
+  it('allows tenant-wide query when user has unrestricted scope (allowedIds === null) and organizationIds is null', async () => {
+    mockContext = {
+      auth: {
+        sub: 'user-unrestricted',
+        tenantId: TENANT_ID,
+        orgId: null,
+        isSuperAdmin: false,
+      },
+      em: mockEm,
+      organizationIds: null,
+      selectedOrganizationId: null,
+      scope: {
+        selectedId: null,
+        filterIds: null,
+        allowedIds: null,
+        tenantId: TENANT_ID,
+      },
+      container: mockContainer,
+      commandContext: {
+        container: mockContainer,
+        auth: { sub: 'user-unrestricted', tenantId: TENANT_ID, orgId: null },
+        organizationScope: { selectedId: null, filterIds: null, allowedIds: null, tenantId: TENANT_ID },
+        selectedOrganizationId: null,
+        organizationIds: null,
+        request: undefined,
+      },
+    }
+
+    const res = await GET(new Request('http://localhost/api/customers/activities?page=1&pageSize=50'))
+    expect(res.status).toBe(200)
+
+    const legacyCall = mockEm.findAndCount.mock.calls.find(
+      (args) => args[0] === CustomerActivity,
+    )
+    expect(legacyCall).toBeDefined()
+    const where = legacyCall![1] as Record<string, unknown>
+    expect(where.tenantId).toBe(TENANT_ID)
+    expect(where.organizationId).toBeUndefined()
+  })
+
+  it('applies organizationId filter when organizationIds is provided', async () => {
+    mockContext = {
+      auth: {
+        sub: 'user-1',
+        tenantId: TENANT_ID,
+        orgId: ORG_1,
+        isSuperAdmin: false,
+      },
+      em: mockEm,
+      organizationIds: [ORG_1],
+      selectedOrganizationId: ORG_1,
+      scope: {
+        selectedId: ORG_1,
+        filterIds: [ORG_1],
+        allowedIds: [ORG_1],
+        tenantId: TENANT_ID,
+      },
+      container: mockContainer,
+      commandContext: {
+        container: mockContainer,
+        auth: { sub: 'user-1', tenantId: TENANT_ID, orgId: ORG_1 },
+        organizationScope: { selectedId: ORG_1, filterIds: [ORG_1], allowedIds: [ORG_1], tenantId: TENANT_ID },
+        selectedOrganizationId: ORG_1,
+        organizationIds: [ORG_1],
+        request: undefined,
+      },
+    }
+
+    const res = await GET(new Request('http://localhost/api/customers/activities?page=1&pageSize=50'))
+    expect(res.status).toBe(200)
+
+    const legacyCall = mockEm.findAndCount.mock.calls.find(
+      (args) => args[0] === CustomerActivity,
+    )
+    expect(legacyCall).toBeDefined()
+    const where = legacyCall![1] as Record<string, unknown>
+    expect(where.organizationId).toEqual({ $in: [ORG_1] })
+  })
+
+  it('returns empty results in unified mode when non-superadmin has organizationIds = null', async () => {
+    const { resolveCustomerInteractionFeatureFlags } = jest.requireMock(
+      '../../../lib/interactionFeatureFlags',
+    )
+    resolveCustomerInteractionFeatureFlags.mockResolvedValue({
+      unified: true,
+      legacyAdapters: true,
+      externalSync: false,
+    })
+
+    mockContext = {
+      auth: {
+        sub: 'user-1',
+        tenantId: TENANT_ID,
+        orgId: null,
+        isSuperAdmin: false,
+      },
+      em: mockEm,
+      organizationIds: null,
+      selectedOrganizationId: null,
+      scope: {
+        selectedId: null,
+        filterIds: null,
+        allowedIds: [ORG_1],
+        tenantId: TENANT_ID,
+      },
+      container: mockContainer,
+      commandContext: {
+        container: mockContainer,
+        auth: { sub: 'user-1', tenantId: TENANT_ID, orgId: null },
+        organizationScope: { selectedId: null, filterIds: null, allowedIds: [ORG_1], tenantId: TENANT_ID },
+        selectedOrganizationId: null,
+        organizationIds: null,
+        request: undefined,
+      },
+    }
+
+    const res = await GET(new Request('http://localhost/api/customers/activities?page=1&pageSize=50'))
+    expect(res.status).toBe(200)
+    const body = await res.json()
+    expect(body.items).toEqual([])
+    expect(body.total).toBe(0)
+
+    expect(mockEm.findAndCount).not.toHaveBeenCalled()
+    expect(mockEm.find).not.toHaveBeenCalled()
+    expect(mockEm.count).not.toHaveBeenCalled()
+  })
+
+  it('allows tenant-wide query when superadmin has organizationIds = [] (empty filterIds from stale scope)', async () => {
+    mockContext = {
+      auth: {
+        sub: 'superadmin-1',
+        tenantId: TENANT_ID,
+        orgId: null,
+        isSuperAdmin: true,
+      },
+      em: mockEm,
+      organizationIds: [],
+      selectedOrganizationId: null,
+      scope: {
+        selectedId: null,
+        filterIds: [],
+        allowedIds: null,
+        tenantId: TENANT_ID,
+      },
+      container: mockContainer,
+      commandContext: {
+        container: mockContainer,
+        auth: { sub: 'superadmin-1', tenantId: TENANT_ID, orgId: null, isSuperAdmin: true },
+        organizationScope: { selectedId: null, filterIds: [], allowedIds: null, tenantId: TENANT_ID },
+        selectedOrganizationId: null,
+        organizationIds: [],
+        request: undefined,
+      },
+    }
+
+    const res = await GET(new Request('http://localhost/api/customers/activities?page=1&pageSize=50'))
+    expect(res.status).toBe(200)
+
+    const legacyCall = mockEm.findAndCount.mock.calls.find(
+      (args) => args[0] === CustomerActivity,
+    )
+    expect(legacyCall).toBeDefined()
+    const where = legacyCall![1] as Record<string, unknown>
+    expect(where.tenantId).toBe(TENANT_ID)
+    expect(where.organizationId).toBeUndefined()
+  })
+
+  it('returns empty results when scope is absent and organizationIds is null (scope shape omitted)', async () => {
+    mockContext = {
+      auth: {
+        sub: 'user-1',
+        tenantId: TENANT_ID,
+        orgId: null,
+        isSuperAdmin: false,
+      },
+      em: mockEm,
+      organizationIds: null,
+      selectedOrganizationId: null,
+      scope: undefined as unknown as null,
+      container: mockContainer,
+      commandContext: {
+        container: mockContainer,
+        auth: { sub: 'user-1', tenantId: TENANT_ID, orgId: null },
+        organizationScope: undefined as unknown as null,
+        selectedOrganizationId: null,
+        organizationIds: null,
+        request: undefined,
+      },
+    }
+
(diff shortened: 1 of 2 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T01:10:31Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review

## 🎯 Summary

This PR closes the org-isolation gap that issue #3841 describes on the deprecated `GET /api/customers/activities` endpoint. The diagnosis in the PR body is accurate, and I confirmed the vulnerability independently: `organizationIds` is derived as `scope.filterIds ?? (auth.orgId ? [auth.orgId] : null)` in `resolveCustomersRequestContext`, and `resolveOrganizationScope` legitimately returns `filterIds: []` (or `null`) for a restricted principal with no accessible organizations and no resolvable home org. The old `if (organizationIds && organizationIds.length > 0)` guard then dropped the `organizationId` clause entirely, leaving a tenant-wide read of `customer_activities`.

The chosen fix is the right shape. Gating on `isUnrestricted = auth.isSuperAdmin === true || scope?.allowedIds === null` mirrors `hasUnrestrictedAccess = effectiveSuperAdmin || (accessibleList === null)` inside `resolveOrganizationScope` exactly, so the route and the scope resolver now agree on what "unrestricted" means. Reading super-admin status from the immutable `auth.isSuperAdmin` flag rather than a role name follows the documented repo lesson on that subject. The `scope?.` optional chain fails closed if `scope` is ever absent, and the empty response envelope (`totalPages: 1`) matches what the normal path produces for a zero-result page, including the `Deprecation`/`Sunset`/`Link` adapter headers. No public contract surface moves: the route, its query schema, its response shape, the OpenAPI document, and the database schema are all unchanged.

One issue keeps this from being an approve. The two helper-level guards do not implement the same policy as the handler-level gate, and the divergence is reachable, behavior-changing for super-admins, undocumented, and untested — details in Findings below. Everything else I found is a nit.

## Verdict

**Changes requested** — one Medium finding. Per this repository's verdict rule, Medium findings request changes; there are no Critical or High findings, and the security fix itself is correct.

## 🧪 Validation Gate

Validation source: **GitHub checks** (all checks for head `9245c104` are terminal and none failed, so the local gate was not reproduced). Branch protection reports no required-status-check configuration (HTTP 404), so every reported check is treated as required.

| Check | State |
|-------|-------|
| `prepare` | ✅ SUCCESS |
| `lint` | ✅ SUCCESS |
| `test` | ✅ SUCCESS |
| `docker-build` | ✅ SUCCESS |
| `ds-lint` | ✅ SUCCESS |
| `audit-scope` | ✅ SUCCESS |
| `scope` | ✅ SUCCESS |
| `ephemeral-integration (none)` | ✅ SUCCESS |
| `documents-multi-instance` | ✅ SUCCESS |
| `CodeQL` / `Analyze (javascript-typescript, actions, python)` | ✅ SUCCESS |
| `license/cla` | ✅ SUCCESS |
| `merge-coverage`, `mutate`, `comment`, `audit`, `[code]smith` | ⏭️ SKIPPED |

No checks are pending and none failed. The head is `MERGEABLE` against `develop` with no conflicts, and `origin/develop` is already an ancestor of the head, so the diff under review is the diff that will merge.

## 🟠 Findings — Medium

### M1 · The helper guards and the handler gate disagree about `organizationIds === []` for unrestricted callers

`packages/core/src/modules/customers/api/activities/route.ts:289` and `packages/core/src/modules/customers/api/activities/route.ts:381`

Both helpers open with an unconditional guard:

```ts
if (organizationIds && organizationIds.length === 0) {
  return { items: [], total: 0 }
}
```

This fires regardless of `isUnrestricted`, whereas the handler gate at `route.ts:443` deliberately lets unrestricted callers through for exactly that input. Two guards implementing one policy give two different answers, and because the request always reaches the helper, the stricter one silently wins.

That case is reachable, and the resolver's own comments describe the scenario that produces it. Walking `resolveOrganizationScope` in `packages/core/src/modules/directory/utils/organizationScope.ts` for a super-admin carrying a stale `om_selected_org` cookie whose organization no longer resolves (a DB reset, or a deleted org) and whose `auth.orgId` is likewise dead:

1. `selectionResolvesToOrg(initialSelected)` is false, so `effectiveSelected` stays `null`.
2. `allowedSet` is `null` (unrestricted), and `widenToAllOrgs` is false because a concrete organization *was* requested.
3. Control reaches `else if (auth.orgId)`, and `loadFallbackSet()` expands the dead home org to an **empty** set.
4. The trailing recovery block recomputes the same empty set and leaves `filterSet` empty.
5. The scope returns `filterIds: []` with `allowedIds: null`.

`organizationIds` is therefore `[]`, `isUnrestricted` is `true`, the handler gate passes, and the helper guard returns an empty page. Before this PR that caller received tenant-wide results.

The narrowing is fail-closed, so this is not a security regression — but it is an undocumented behavior change on the super-admin read path, and the PR body states the opposite: *"Superadmin and unrestricted principals continue to have tenant-wide read access."* It is also the one branch of the new logic with no test behind it. Please pick one and make it explicit:

- **Make the guards consistent** — gate the first guard on `!isUnrestricted` as well, matching the handler, so an unrestricted principal with an empty filter set keeps its previous tenant-wide read; or
- **Keep the stricter behavior deliberately** — treat an explicitly empty `filterIds` as "no organizations" for everyone, drop the now-misleading `isUnrestricted` exemption for `[]` from the handler gate, correct the Breaking Changes section of the PR body, and add a test pinning the super-admin-with-`[]` case.

Either resolution is fine; the second is arguably the better semantics, since an explicitly empty filter set is a positive statement rather than an absent one. What should not ship is the current state, where the two guards disagree and the PR description documents the behavior the code does not have.

## 🔵 Findings — Low

### L1 · Unused import in the new test file

`packages/core/src/modules/customers/api/activities/__tests__/org-scoping.test.ts:3`

`CustomerInteraction` is imported but never referenced; only `CustomerActivity` is used, in the `findAndCount` call lookups. Lint passes, so this is cosmetic, but the import reads as if unified-mode assertions were planned and then written against `CustomerActivity` instead.

### L2 · The empty-scope tests assert over a loop that never runs

`packages/core/src/modules/customers/api/activities/__tests__/org-scoping.test.ts:112`, `:152`, `:320`

```ts
for (const call of mockEm.findAndCount.mock.calls) {
  const where = call[1] as Record<string, unknown>
  expect(where.organizationId).toBeDefined()
}
```

In all three empty-scope tests the handler returns before touching the entity manager, so `mock.calls` is empty and the loop body never executes. It still works as a regression net — if every guard were removed, a `tenantId`-only query would appear and the assertion would catch it — but it does not state the property the fix actually establishes, which is that *no query is issued at all*. `expect(mockEm.findAndCount).not.toHaveBeenCalled()` (optionally alongside `mockEm.find` and `mockEm.count`) says that directly and fails loudly rather than vacuously.

### L3 · Scope-shape parameterization is incomplete

`packages/core/src/modules/customers/api/activities/__tests__/org-scoping.test.ts:77`

The suite covers `organizationIds: null` and `organizationIds: []`, but not the omitted variant, and never exercises a context where `scope` itself is absent — which is the branch the new `scope?.allowedIds === null` optional chain exists to handle. Both currently fail closed, so this is coverage rather than a defect, but `.ai/lessons.md` carries an explicit entry from 2026-07-11 (`use-cryptographic-randomness-in-auth-adjacent-test.md`) requiring scope tests to parameterize non-null, null, **and** omitted for every predicate path, and the repo-local `om-code-review` extension treats recorded lessons as review rules. A short `it.each` over the three shapes would satisfy it and cost almost nothing.

### L4 · The same pattern is still live on the sibling todos adapter

`packages/core/src/modules/customers/lib/todoCompatibility.ts:383` and `packages/core/src/modules/customers/lib/todoCompatibility.ts:437`

`listLegacyTodoRows` and `listCanonicalTodoRows` carry the identical `if (organizationIds && organizationIds.length > 0)` guard, and `packages/core/src/modules/customers/api/todos/route.ts:236` feeds them `organizationIds` from the very same `resolveCustomersRequestContext`. The precondition that made #3841 exploitable therefore reproduces there unchanged. Issue #3841 called this out directly — *"The same `organizationIds?.length > 0`-guarded org clause is a hotspot pattern worth grepping for elsewhere"* — but it is outside this PR's stated scope, so it belongs in a follow-up rather than in this diff. Flagging it so it does not get lost now that the activities half is closed.

### L5 · The fail-closed path is silent

`packages/core/src/modules/customers/api/activities/route.ts:443`

A collapsed scope now yields an empty `200` with no log line, which is indistinguishable from "this org genuinely has no activities" during incident triage. The file already has a `logger` in hand. Worth a `logger.warn` naming the tenant and the collapsed scope.

For what it is worth, the closest precedent in this repo — `.ai/specs/2026-07-28-audit-log-read-tenant-scope-fail-closed.md` — chose an explicit `403` over a narrowed result for the same class of problem, on the grounds that rejecting "keeps the guarantee local and legible". Issue #3841 explicitly permitted either shape here, and an empty page is more defensible for a list endpoint than for an audit read, so I am not asking for a change — just noting the divergence from precedent so it is a deliberate choice.

## 💥 Breaking Changes

| Surface | Status |
|---------|--------|
| API route path and methods | Unchanged |
| Query schema (`listSchema`) | Unchanged |
| Response shape and OpenAPI document | Unchanged |
| Adapter headers (`Deprecation`, `Sunset`, `Link`) | Preserved on the new early-return path |
| Database schema and migrations | Untouched |
| Exported symbols | Unchanged — `listLegacyActivities` and `listCanonicalActivities` are module-private, so the new optional `isUnrestricted` parameter is not a public signature change |
| Event IDs, DI keys, ACL features, widget spot IDs | Untouched |

No protected surface from `BACKWARD_COMPATIBILITY.md` is affected. The only behavioral narrowing is the intended one for restricted principals, plus the unintended super-admin case in **M1** — which is precisely why M1 needs resolving before merge: the PR currently claims no behavior change for unrestricted callers, and that claim is not accurate as written.

## 🧪 Test Coverage

`__tests__/org-scoping.test.ts` adds six cases covering the restricted-null, restricted-empty, super-admin-null, `allowedIds === null`, explicit-`organizationIds`, and unified-mode paths. That is a genuine matrix rather than a single happy-path test, it exercises both the merged legacy path and `flags.unified`, and it pins the positive `{ $in: [...] }` filter so a future regression that drops the clause is caught. The mocking is scoped to the module's own libs and does not stub the code under test.

Gaps, all covered above: the super-admin-with-`[]` case that **M1** describes (untested and currently behavior-changing), the omitted-scope shapes in **L3**, and the assertion strength in **L2**.

## Next steps

Resolving **M1** — either by aligning the first helper guard with the handler gate, or by accepting the stricter semantics and correcting the handler gate, the PR body, and the test matrix to match — is all that stands between this and an approve. The Low findings are yours to take or leave; **L4** is worth a follow-up issue regardless, and I am happy to file it.

---

pkarw:

- 🚀 `merge-queue` — the re-review of head `5679927c` approved the PR: the Medium finding from the previous pass is resolved, the four actionable Low findings were taken, and CI is fully green. Replaces `changes-requested`.
- 🐛 `bug` — kept. The change fixes a defect: the deprecated activities list dropped its organization filter for a restricted principal with a collapsed scope.
- 🔒 `security` — kept. The defect was a cross-organization read, so this is a security fix as much as a bug fix.
- ⏭️ `skip-qa` — kept, and the automated-verification exemption supports it. The diff touches one API route file and one test file, no `.tsx` and nothing under `packages/ui/`; the database schema and the API surface are unchanged; no `BACKWARD_COMPATIBILITY.md` contract is affected; and automated tests for the changed behavior ship in the same PR.
- 🔹 `priority-medium` — kept. This matches issue #3841's own rating: a real organization-isolation gap, but it requires a principal whose organization scope has collapsed, and it lands on an endpoint already scheduled for sunset on 2026-06-30.
- 🟡 `risk-medium` — kept. The blast radius is one route file plus its tests inside a single module, with no schema, contract, or shared-surface changes, and the fix is covered by an eight-case test matrix over the full predicate space.

pkarw:

pkarw:

Re-review of head `5679927c`, superseding the `CHANGES REQUESTED` verdict this comment previously carried for `9245c104`. The Medium finding is resolved the right way: both helper guards now read `organizationIds && organizationIds.length === 0 && !isUnrestricted`, so the helper-level and handler-level policies state one rule instead of two, and an unrestricted principal with an empty `filterIds` keeps the tenant-wide read it had before this PR — which is what the PR body claims. I re-walked the full predicate matrix (`null`/`[]`/populated × restricted/unrestricted) against the current head to confirm there is no remaining disagreement, and the restricted paths still fail closed, including the omitted-`scope` shape.

L1, L2, L3 and L5 were all taken; L4 is tracked as #5466, which also carries the correction that the non-deprecated `api/interactions/tasks/route.ts` is affected alongside the deprecated todos route. Two new tests cover the previously unverified branches — super-admin with `organizationIds = []`, and an absent `scope` — bringing the suite to eight cases across both the merged legacy path and `flags.unified`. Only four style-level nits remain in the review body; none blocks the merge.

Validation source: GitHub checks — every check on head `5679927c` is terminal and green, so no local gate was run and no CI follow-up is owed. Labels moved to `merge-queue`; `skip-qa` stays, and the automated-verification exemption supports it. The claim assignee was removed with the lock; @haxiorz remains assigned as the author.

autofix: skipped (not my PR — re-run with --autofix to fix it here). Nothing needed fixing on this pass in any case.

haxiorz:
@pkarw  Addressed review findings:

**M1 fixed:** helper guards now consistently gate on \!isUnrestricted\ for the \organizationIds === []\ case, matching the handler gate. Unrestricted callers with an empty \ilterIds\ (e.g. superadmin with stale \om_selected_org\) continue to receive tenant-wide results, as documented. Added test \llows tenant-wide query when superadmin has organizationIds = []\.

**L1 fixed:** removed unused \CustomerInteraction\ import.
**L2 fixed:** empty-scope tests now assert \expect(mockEm.findAndCount).not.toHaveBeenCalled()\ (plus \ind\/\count\) instead of vacuously iterating over empty calls.
**L3 fixed:** added coverage for omitted scope shape (\scope === undefined\ with \organizationIds === null\).
**L5 fixed:** collapsed-scope early return now emits \logger.warn\ with \	enantId\, \organizationIds\ and \scope\ for triage.

**L4 acknowledged:** identical pattern in \	odoCompatibility.ts\ is out of scope for this PR per review — will be tracked separately.

pkarw:
Filed the **L4** follow-up as #5466.

One correction to L4 as I wrote it in the review: I named `/api/customers/todos` as the affected surface, but tracing the callers properly turned up **two** routes, and the second matters more than the first.

| Route | Deprecated? |
|-------|-------------|
| `api/todos/route.ts` (:248, :276, :279) | Yes — sunset 2026-06-30 |
| `api/interactions/tasks/route.ts` (:53, :82, :85) | **No — the canonical successor endpoint** |

Both take `organizationIds` from the same `resolveCustomersRequestContext`. So unlike the activities case, this one does not age out with the legacy adapter — fixing only the deprecated route would leave the hole live on the endpoint that replaces it.

The third consumer, `api/dashboard/widgets/customer-todos/route.ts`, is **not** affected: it sources scope from `resolveWidgetScope` (`packages/core/src/modules/dashboards/lib/widgetScope.ts:32-42`), which already returns `null` only for a genuinely unrestricted principal and rejects an empty org set with a 400. That is the cleanest in-repo model for the policy, and #5466 proposes mirroring it.

None of this changes anything about this PR — #5466 is independent work. **M1 is still the only thing blocking here.**

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T06:24:58Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review

**Re-review** of PR #5462 at head `5679927c`, following the `changes-requested` verdict on `9245c104`.

## 🎯 Summary

The author force-pushed a revision that addresses every finding from the previous pass. I re-checked each one against the current head rather than taking the response comment at its word, and the resolutions hold.

The Medium finding that blocked the last review — the two helper guards short-circuiting on `organizationIds === []` regardless of `isUnrestricted`, while the handler gate deliberately let unrestricted callers through for that same input — is fixed the way I would have picked. Both guards at `packages/core/src/modules/customers/api/activities/route.ts:289` and `:381` now read `if (organizationIds && organizationIds.length === 0 && !isUnrestricted)`, so the helper-level and handler-level policies state the same rule. I walked the resulting predicate space to confirm there is no remaining disagreement:

| `organizationIds` | `isUnrestricted` | Handler gate (`:443`) | Helper guards | Result |
|---|---|---|---|---|
| `null` | `false` | returns empty page | (unreachable) `:296` / `:384` would return empty | empty — fail closed ✅ |
| `[]` | `false` | returns empty page | (unreachable) `:289` / `:381` would return empty | empty — fail closed ✅ |
| `null` | `true` | passes | passes | no `organizationId` clause → tenant-wide, as before this PR ✅ |
| `[]` | `true` | passes | passes | no `organizationId` clause → tenant-wide, as before this PR ✅ |
| `[org…]` | either | passes | passes | `where.organizationId = { $in: organizationIds }` ✅ |

The unintended super-admin narrowing is therefore gone, and the PR body's "Superadmin and unrestricted principals continue to have tenant-wide read access" is now an accurate description of the code. The restricted paths still fail closed, including the `scope`-absent shape, where `scope?.allowedIds === null` evaluates to `false` rather than matching. The security fix that this PR exists to make — closing the cross-organization read on `GET /api/customers/activities` for a restricted principal whose organization scope collapses to `null` or `[]` — is intact and unchanged in substance.

The four Low findings were taken as well: the unused `CustomerInteraction` import is gone (**L1**); the three empty-scope tests now assert `expect(mockEm.findAndCount).not.toHaveBeenCalled()` alongside `find` and `count`, which states the property the fix actually establishes instead of iterating a loop that never ran (**L2**); a case for the omitted `scope` shape was added (**L3**); and the collapsed-scope early return now emits a `logger.warn` so the empty page is distinguishable from a genuinely empty organization during incident triage (**L5**). **L4** — the identical `organizationIds?.length > 0` pattern still live in `todoCompatibility.ts` — was correctly left out of this diff and is tracked as #5466, which also captures the correction that the non-deprecated `api/interactions/tasks/route.ts` is affected too.

Two new tests came with the revision and both pin behavior that was previously unverified: `allows tenant-wide query when superadmin has organizationIds = [] (empty filterIds from stale scope)` covers exactly the branch M1 was about, and `returns empty results when scope is absent and organizationIds is null (scope shape omitted)` covers the branch the `scope?.` optional chain exists to handle. The suite is now eight cases over the full predicate matrix.

Everything I found on this pass is a nit. Approving.

## Verdict

**Approved** — the one Medium finding from the previous review is resolved, all four actionable Low findings were taken, the fifth is tracked as a follow-up issue, and only style-level nits remain. Under this repository's verdict rule (Medium and above request changes; Low-only approves), this is an approve.

## 🧪 Validation Gate

Validation source: **GitHub checks** (no local fallback needed). Every check reported against head `5679927c` is terminal, and none failed, so the CI gate is the evidence for this review rather than a locally reproduced run. Branch protection on `develop` reports no required-status-check configuration, so every reported check is treated as required.

| Check | State |
|-------|-------|
| `prepare` | ✅ SUCCESS |
| `lint` | ✅ SUCCESS |
| `test` | ✅ SUCCESS |
| `docker-build` | ✅ SUCCESS |
| `ds-lint` | ✅ SUCCESS |
| `audit-scope` | ✅ SUCCESS |
| `scope` | ✅ SUCCESS |
| `ephemeral-integration (none)` | ✅ SUCCESS |
| `documents-multi-instance` | ✅ SUCCESS |
| `CodeQL` / `Analyze (javascript-typescript, actions, python)` | ✅ SUCCESS |
| `license/cla` | ✅ SUCCESS |
| `merge-coverage`, `mutate`, `comment`, `audit`, `[code]smith` | ⏭️ SKIPPED |

No checks are pending and none failed, so there is no CI disclosure to make and no follow-up owed. The head is `MERGEABLE` against `develop`; `develop` has moved forward by eight commits since this branch's base, and I confirmed a three-way merge against the current `origin/develop` produces no conflicts. The one commit in that range that touches this module (`10db5db2`, deal-analyzer person links) does not touch the activities route or its tests, so there is no semantic drift either.

## 🔵 Findings — Low

None of these block the merge; take them or leave them.

### N1 · Unnecessary double cast on the omitted-scope fixture

`packages/core/src/modules/customers/api/activities/__tests__/org-scoping.test.ts:373` and `:378`

```ts
scope: undefined as unknown as null,
```

`mockContext` is declared as `Record<string, unknown>` at line 26, so `scope: undefined` type-checks on its own and the `as unknown as null` escape hatch buys nothing. Plain `undefined` also reads more honestly: the point of the case is that the property is *absent*, not that it is a `null` wearing a disguise.

### N2 · The new log message breaks the file's own naming convention

`packages/core/src/modules/customers/api/activities/route.ts:444`

The four pre-existing log calls in this file are `customers.activities.get failed`, `customers.activities.post failed`, and so on (`:539`, `:628`, `:711`, `:779`). The new one is `activities.list collapsed organization scope — returning empty page`, which drops the `customers.` prefix. The logger is created as `createLogger('customers')` so the binding is not lost, but anyone grepping logs for `customers.activities.` will miss this line — which is unfortunate given the whole point of adding it was incident triage. `customers.activities.list collapsed organization scope` would match its siblings.

### N3 · The `logger.warn` payload duplicates a field and dumps the whole scope object

`packages/core/src/modules/customers/api/activities/route.ts:444-448`

The fields are `{ tenantId, organizationIds, scope }`, but `organizationIds` is derived from `scope.filterIds`, so the two are near-redundant in the emitted record. More relevantly, `scope.allowedIds` can be a long list of organization UUIDs for a user with wide-but-not-unrestricted access, and this line sits on a `GET` list path a client can poll: a restricted principal in the collapsed state produces one warn-level record per request, each carrying that array. Logging `scope.selectedId` and `scope.allowedIds?.length` instead of the whole object would keep the triage value at a fraction of the volume. Worth a thought rather than a change request — the fail-closed state should be rare in practice.

### N4 · Both helper guards are now unreachable from the only call site

`packages/core/src/modules/customers/api/activities/route.ts:289`, `:296`, `:381`, `:384`

With the handler gate at `:443` returning early for every `!isUnrestricted` case where `organizationIds` is null or empty, none of the four helper guards can fire: `listLegacyActivities` and `listCanonicalActivities` are module-private and called only from `GET` (`:463`, `:481`, `:490`). I am explicitly **not** asking for their removal — defense in depth on an organization-scoping predicate is the right instinct, and if a future caller is added the guards are what stop it from repeating #3841. It is worth one line of comment saying so, though, so that a later reader who notices the redundancy does not "clean it up" and quietly reopen the hole.

## 💥 Breaking Changes

| Surface | Status |
|---------|--------|
| API route path and methods | Unchanged |
| Query schema (`listSchema`) | Unchanged |
| Response shape and OpenAPI document | Unchanged |
| Adapter headers (`Deprecation`, `Sunset`, `Link`) | Preserved on the early-return path |
| Database schema and migrations | Untouched |
| Exported symbols | Unchanged — both helpers are module-private, so the added `isUnrestricted` parameter is not a public signature change |
| Event IDs, DI keys, ACL features, widget spot IDs | Untouched |

No protected surface from `BACKWARD_COMPATIBILITY.md` is affected. The only behavioral change that ships is the intended one: a restricted principal whose organization scope resolves to `null` or `[]` now receives an empty page instead of tenant-wide activities. Unrestricted principals — super-admins and callers with `scope.allowedIds === null` — see byte-identical behavior to `develop`, in all four combinations of the two scope shapes, which is what the previous review's M1 was asking for and what the PR body claims.

## 🧪 Test Coverage

`__tests__/org-scoping.test.ts` now carries eight cases: restricted-with-`null`, restricted-with-`[]`, super-admin-with-`null`, `allowedIds === null`-with-`null`, explicit-`organizationIds`, unified-mode-restricted, super-admin-with-`[]`, and omitted-scope. That covers every branch of the new predicate on both the merged legacy path and `flags.unified`.

The assertions also got stronger where it counted. The three empty-scope cases previously iterated `mockEm.findAndCount.mock.calls` — a loop whose body never executed, because the handler returns before touching the entity manager — and now assert `not.toHaveBeenCalled()` on `findAndCount`, `find`, and `count`. That states the actual guarantee (no query is issued at all) and fails loudly rather than vacuously. The positive cases still pin `where.organizationId` to `{ $in: [...] }` or to `undefined` as appropriate, so a regression that drops or over-applies the clause is caught from both directions.

The mocking is well-scoped: the feature-flag resolver, the request-context resolver, and the read model are stubbed, the code under test is not, and `beforeEach` re-establishes the entity-manager implementations, so the unified-mode case at `:274` cannot leak its `unified: true` flag into the cases that follow it.

One residual gap, small enough that I am not filing it as a finding: `organizationIds` is never exercised in its `undefined` form (the type is `string[] | null`, and the handler's `!organizationIds` would catch it anyway), so the omitted-shape coverage that `.ai/lessons.md` asks for is complete on `scope` but not on `organizationIds` itself.

## Next steps

Nothing is required before merge from the code-review side. The PR carries `skip-qa`, which the automated-verification exemption supports here — no `.tsx` or `packages/ui/` file is touched, the database schema and the API surface are unchanged, no `BACKWARD_COMPATIBILITY.md` contract is affected, and automated tests for the changed behavior ship in the same PR. The pipeline label moves to `merge-queue`.

Thanks for the thorough turnaround on M1 — picking the alignment option and adding the super-admin-with-`[]` test is exactly the resolution that keeps the PR description and the code telling the same story.

---

PR #5464, opened 2026-08-22
security(customer_accounts): reject domain registration for organizations outside the caller's tenant (#3839)

### LINKED ISSUE #3839: security(customer_accounts): domain registration does not verify organizationId belongs to caller's tenant (domainMappingService.ts:270)
`register()` stamps a new domain mapping with the caller's `tenantId` but never verifies that the client-supplied `organizationId` belongs to that tenant, unlike the signup path which resolves it via `findOrganizationInTenant`. This is **LIVE**: an admin with `customer_accounts.domain.manage` can post `{ hostname, organizationId: <org owned by a different tenant> }` and create a `domain_mapping` row pairing their own tenant with a foreign org id. Impact is limited — it is admin-gated, the row is caller-tenant-scoped, and resolution keys off the caller's tenant — but it lets a tenant admin bind a hostname to an `organizationId` it does not own, corrupting the org→domain routing map.

## Location(s)
- `packages/core/src/modules/customer_accounts/services/domainMappingService.ts:270-291` — `register()` stamps caller `tenantId` but does not validate the supplied `organizationId` against it
- `packages/core/src/modules/customer_accounts/api/admin/domain-mappings.ts` — POST route invoking `register()`

## Why it matters
An admin with `customer_accounts.domain.manage` posts a hostname plus an `organizationId` owned by a different tenant. `register()` writes the row with `tenantId = caller` and the foreign `organizationId` unchecked, producing a `(tenantId=caller, organizationId=foreign-org)` mapping. The signup path already validates org membership via `findOrganizationInTenant`, so this route is inconsistent with the established pattern. The practical impact is corruption of the org→domain routing map within the caller's tenant rather than a cross-tenant read/write, but it is an unvalidated-FK write.

## Proposed fix
- In `register()` (or the POST route) validate the org belongs to the tenant via `findOrganizationInTenant(em, organizationId, tenantId)` before creating the mapping, and reject otherwise — mirroring the signup path.

## BC / Safety
No wire change for legitimate callers who supply an owned org id. Registrations with a non-owned `organizationId` will now be rejected. Re-verify normal domain registration and resolution flows still work for in-tenant orgs. risk-medium as a single-module scoping fix.

haxiorz:
Closes #3839

## Goal

Admins can only bind custom domains to organizations they actually own.

## Problem and root cause

`DomainMappingService.register()` stamped the caller's `tenantId` next to a client-supplied `organizationId` without verifying membership, so an admin holding `customer_accounts.domain.manage` could register a hostname bound to another tenant's org id, corrupting the org→domain routing map within their tenant (unvalidated-FK write). The signup path already validated membership via `findOrganizationInTenant`, so this route was inconsistent with the established pattern.

## What changed

- `register()` in `packages/core/src/modules/customer_accounts/services/domainMappingService.ts` now resolves the organization through `findOrganizationInTenant(em, organizationId, tenantId)` before creating anything and throws the new exported `DomainMappingOrgScopeError` on miss.
- The admin POST route (`packages/core/src/modules/customer_accounts/api/admin/domain-mappings.ts`) maps that error to `400`, mirroring signup's rejection shape; 409/500 paths unchanged.
- Regression tests (`services/__tests__/domainMappingService.test.ts`): foreign-tenant and unknown org ids are rejected before any persist/event/cache side effect, in-tenant happy paths unchanged, plus a source guard pinning the lookup wiring inside `register()`.

## Validation

- `yarn generate` — pass, no generated drift
- `yarn build:packages` — pass (26/26 tasks)
- `yarn i18n:check-sync` / `yarn i18n:check-usage` — pass
- `yarn typecheck` — pass (26/26 tasks)
- `yarn build:app` — pass
- Targeted suite `npx jest --config jest.config.cjs domainMappingService` — 48/48 pass; both new regression tests were observed failing before the fix (red run) because `register()` persisted the cross-tenant pair
- Honest limitation: full `yarn test` fails on Windows in the unrelated `@open-mercato/queue` `local.strategy.test.ts` with an EPERM rename/watch flake that reproduces identically on pristine `origin/develop` at the same commit; every other package is green on this branch

## Automated review

- `om-code-review` — approve; one round, zero blocker/major/minor findings
- Review policy — direct author review pass, satisfied
- Unresolved findings — None (two informational nits recorded)

## Breaking changes

None. No route URL, method, response field, schema, or event surface changes; registrations with a non-owned `organizationId` are now rejected by design (the intended behavior change from the issue); legitimate callers supplying an owned org id are unaffected.

## Labels and delivery

- `review` — fix complete and verified, awaiting human review
- `bug` + `security` — unvalidated-FK write closing a tenant-scoping gap found during a security audit
- `skip-qa` — no UI-rendering file touched, database structure unchanged, and the changed behavior is covered by automated regression tests
- `priority-low` — matches the issue's triage; admin-gated with caller-tenant-scoped impact
- `risk-medium` — single-module scoping change with a deliberate behavior change on one endpoint, fully unit-covered

## Manual QA

Automation sufficient: service-level rejection and route error mapping are covered by focused unit tests; no UI surface changed.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5464"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789953002&installation_model_id=432243&pr_number=5464&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5464&signature=2a8140b83fe01c55582952429424725bc14cda314adb3bd4ab734cee3f0d57be"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/customer_accounts/api/admin/domain-mappings.ts
+++ b/packages/core/src/modules/customer_accounts/api/admin/domain-mappings.ts
@@ -14,6 +14,7 @@ import {
 import { registerDomainSchema } from '@open-mercato/core/modules/customer_accounts/data/validators'
 import {
   DomainMappingService,
+  DomainMappingOrgScopeError,
   type ResolveResult,
 } from '@open-mercato/core/modules/customer_accounts/services/domainMappingService'
 import { DomainMapping } from '@open-mercato/core/modules/customer_accounts/data/entities'
@@ -140,6 +141,12 @@ export async function POST(req: Request) {
       return NextResponse.json(
         { ok: false, error: 'This domain is not available. Please choose a different hostname.' },
         { status: 409 },
+      )
+    }
+    if (err instanceof DomainMappingOrgScopeError) {
+      return NextResponse.json(
+        { ok: false, error: 'Organization was not found in the current tenant.' },
+        { status: 400 },
       )
     }
     const message = err instanceof Error ? err.message : 'Failed to register domain'
--- a/packages/core/src/modules/customer_accounts/services/__tests__/domainMappingService.test.ts
+++ b/packages/core/src/modules/customer_accounts/services/__tests__/domainMappingService.test.ts
@@ -1,6 +1,6 @@
 /** @jest-environment node */
 import type { EntityManager } from '@mikro-orm/postgresql'
-import { DomainMappingService } from '@open-mercato/core/modules/customer_accounts/services/domainMappingService'
+import { DomainMappingService, DomainMappingOrgScopeError } from '@open-mercato/core/modules/customer_accounts/services/domainMappingService'
 import {
   DomainMapping,
   type DomainStatus,
@@ -43,7 +43,7 @@ type Row = {
   updatedAt: Date | null
 }
 
-type OrgRow = { id: string; slug: string | null }
+type OrgRow = { id: string; slug: string | null; tenantId: string }
 
 let nextId = 1
 function makeId(): string {
@@ -91,7 +91,7 @@ function rowMatches(row: Row, where: WhereClause): boolean {
 }
 
 type FakeEm = jest.Mocked<
-  Pick<EntityManager, 'find' | 'findOne' | 'create' | 'persist' | 'remove' | 'flush'>
+  Pick<EntityManager, 'find' | 'findOne' | 'create' | 'persist' | 'remove' | 'flush' | 'getConnection'>
 > & {
   __store: Map<string, Row>
   __orgs: Map<string, OrgRow>
@@ -104,6 +104,15 @@ function createFakeEm(seed: { domains?: Row[]; orgs?: OrgRow[] } = {}): FakeEm {
   const removed = new Set<string>()
   for (const row of seed.domains ?? []) store.set(row.id, row)
   for (const org of seed.orgs ?? []) orgs.set(org.id, org)
+
+  const getConnection = jest.fn(() => ({
+    execute: async <T>(_: string, params: unknown[]): Promise<T> => {
+      const [organizationId, tenantId] = params as [string, string]
+      const org = orgs.get(organizationId)
+      if (org && org.tenantId === tenantId) return [{ slug: org.slug ?? null }] as T
+      return [] as T
+    },
+  }))
 
   const findOne = jest.fn(async (entity: unknown, where: WhereClause) => {
     if (entity === DomainMapping) {
@@ -187,6 +196,7 @@ function createFakeEm(seed: { domains?: Row[]; orgs?: OrgRow[] } = {}): FakeEm {
     persist,
     remove,
     flush,
+    getConnection,
     __store: store,
     __orgs: orgs,
     __removed: removed,
@@ -225,7 +235,7 @@ beforeEach(() => {
 
 describe('DomainMappingService.register', () => {
   it('creates a pending mapping for a valid hostname (happy path)', async () => {
-    const em = createFakeEm()
+    const em = createFakeEm({ orgs: [{ id: ORG_A, slug: null, tenantId: TENANT_A }] })
     const service = new DomainMappingService(em as unknown as EntityManager)
 
     const result = await service.register({
@@ -244,7 +254,7 @@ describe('DomainMappingService.register', () => {
   })
 
   it('normalizes mixed-case + trailing dot hostnames before persisting', async () => {
-    const em = createFakeEm()
+    const em = createFakeEm({ orgs: [{ id: ORG_A, slug: null, tenantId: TENANT_A }] })
     const service = new DomainMappingService(em as unknown as EntityManager)
 
     const result = await service.register({
@@ -275,7 +285,7 @@ describe('DomainMappingService.register', () => {
       organizationId: ORG_A,
       status: 'active',
     })
-    const em = createFakeEm({ domains: [replacedRow] })
+    const em = createFakeEm({ domains: [replacedRow], orgs: [{ id: ORG_A, slug: null, tenantId: TENANT_A }] })
     const service = new DomainMappingService(em as unknown as EntityManager)
 
     const created = await service.register({
@@ -298,7 +308,7 @@ describe('DomainMappingService.register', () => {
       organizationId: ORG_B,
       status: 'active',
     })
-    const em = createFakeEm({ domains: [replacedRow] })
+    const em = createFakeEm({ domains: [replacedRow], orgs: [{ id: ORG_A, slug: null, tenantId: TENANT_A }] })
     const service = new DomainMappingService(em as unknown as EntityManager)
 
     const created = await service.register({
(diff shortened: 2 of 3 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T01:46:32Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
## Code Review: security(customer_accounts): reject domain registration for organizations outside the caller's tenant (#3839)

**Verdict:** ✅ **Approve**
**Findings:** 0 blocker / 0 major / 0 minor / 2 nit
**Validation source:** GitHub checks (all reported checks green at head `768edc58`; branch protection not readable, so every reported check was treated as required — `test`, `lint`, `ds-lint`, `prepare`, `docker-build`, `ephemeral-integration (none)`, `documents-multi-instance`, `audit-scope`, CodeQL analyses all pass). No local fallback gate was needed.
**Scope:** 3 files — 1 service, 1 API route, 1 test file (+60/-7)

---

### Summary

`DomainMappingService.register()` stamped the caller's `tenantId` alongside a client-supplied `organizationId` with no membership check, so an admin holding `customer_accounts.domain.manage` could bind a hostname to a foreign org id inside their own tenant's routing map (unvalidated-FK write). The fix resolves the organization through the module's existing `findOrganizationInTenant(em, organizationId, tenantId)` helper before anything is created, and the admin POST route maps the new `DomainMappingOrgScopeError` to a `400`.

The approach is the right one: it reuses the already-audited helper rather than inlining a second copy of the tenant predicate, which is exactly what `.ai/lessons/keep-raw-sql-out-of-api-route-handlers.md` prescribes for this module, and it mirrors the pattern `api/signup.ts` already uses for the same lookup.

### What I verified

**Correctness / placement**
- The lookup sits at the top of `register()`, before the `replacesDomainId` fetch, before `em.create`, before `persist().flush()`, before `emitCustomerAccountsEvent(...)` and before `invalidateCacheFor(...)` — so a rejected registration produces no persist, no event, and no cache mutation. The two new tests pin exactly that (`em.persist` not called, `__store` empty).
- `normalizeHostname()` still runs first, so hostname-validation errors keep their existing precedence and shape.
- The helper's predicate is `id = ? AND tenant_id = ? AND deleted_at IS NULL` — soft-deleted orgs correctly do not satisfy membership.
- The lookup runs before any scalar mutation, so the `withAtomicFlush` rule in `packages/core/AGENTS.md` (no reads between mutations and `flush()`) is not engaged.

**Blast radius**
- `DomainMappingService.register()` has exactly one production caller — `api/admin/domain-mappings.ts` POST. Workers (`domainVerificationWorker`, `domainTlsRetryWorker`), the mutation guards, `customerAuthServer`, `customerUrl`, `resolveTenantContext`, `domain-resolve/all` and `domain-check` all use read/verify paths, none of which changed. The behavior change is confined to one endpoint.
- `RegisterInput.tenantId` is a required `string` and the route rejects a missing `auth.tenantId` with 401 before calling, so the helper can never receive `undefined`.

**Error handling**
- Catch ordering is safe: `isUniqueViolation()` matches on `UniqueConstraintViolationException`, code `23505`, or a `duplicate key` substring — none of which the new error satisfies — so the 409 branch cannot swallow the 400.
- The internal message carries the `[internal]` prefix per root `AGENTS.md`, and the route deliberately does **not** echo `err.message` on this path, so the rejected org id never reaches the client. The 500 fall-through that *does* echo `err.message` is unreachable for this error given the sole caller catches it.
- Returning `400` (rather than 403/404) avoids turning the endpoint into an org-existence oracle, consistent with signup's deliberately vague `Registration could not be completed`.

**Backward compatibility** (`BACKWARD_COMPATIBILITY.md`)
- Additive only: one new exported class. No route URL, method, response field, event id, DI key, ACL feature, entity, or schema touched — no migration expected or needed.
- The POST OpenAPI block already documents `400 → errorSchema`, so the documented contract is unchanged and stays accurate.
- The one intentional behavior change (registrations naming a non-owned org now fail) is the fix the issue asked for and is called out in the PR body.

**Tests**
- Two behavioral regression tests (foreign-tenant org, non-existent org) plus a wiring guard. The fake EM gained a `getConnection().execute` stub that models the `(id, tenant_id)` pair match, and `OrgRow` gained a required `tenantId` — which forces every existing `orgs` seed to be explicit about tenant ownership rather than defaulting silently. Good.
- The `register.toString()` wiring guard is source-string-based, which is normally brittle, but it is the pattern this module already uses (`__tests__/signup-login-tenant-binding.test.ts`) and the lesson file above explicitly asks for the predicate to stay pinned after relocation into a helper. Accepting it as house style.

**QA gating** — `skip-qa` is correctly applied: no `.tsx` outside tests, nothing under `packages/ui/src/`, database structure and API surface unchanged, and the changed behavior ships with automated tests in the same PR. The admin UI's org selector is populated from tenant-scoped orgs, so the new 400 is not reachable through normal UI use — it is API hardening.

**Labels** — keeping `priority-low` and `risk-medium` as-is. Despite the `security` category, the attack requires an authenticated admin holding `customer_accounts.domain.manage`, and the corrupted row lands in the caller's *own* tenant (`tenantId` is server-stamped) — this is an integrity bug in one tenant's routing map, not a cross-tenant data leak, so the existing triage holds.

---

### Findings

#### 🔵 Low 1 — the new 400 message is a server-supplied English string that reaches the UI

`packages/core/src/modules/customer_accounts/api/admin/domain-mappings.ts:148`

```ts
{ ok: false, error: 'Organization was not found in the current tenant.' }
```

`backend/customer_accounts/settings/domain/page.tsx:228` renders `call.result?.error` verbatim, so this string is user-facing and is not routed through `t('customer_accounts.…')`.

Not raised above nit level, and **not** something to change in this PR: every other error body in this route (`Unauthorized`, `Forbidden`, `Invalid request`, the 409 `This domain is not available…`) and in `api/signup.ts` is a plain English literal, and `yarn i18n:check-sync` / `i18n:check-usage` pass in CI. Converting one added line while the six around it stay hardcoded would make the file *less* consistent. The right fix is one follow-up pass moving this route's whole error surface behind `resolveTranslations()`.

#### 🔵 Low 2 — intra-tenant cross-org binding remains possible (pre-existing, out of scope)

`register()` now proves the org is in the caller's tenant, but not that the caller may act on *that* org. An admin whose organization visibility is limited to org A can still register a domain for org B in the same tenant: the RBAC check at `domain-mappings.ts:96` evaluates the feature against `auth.orgId`, while the write targets `parsed.data.organizationId`, and none of the four guards in `data/guards.ts` (`hostname-format`, `hostname-unique`, `org-limit`, `record-scope`) asserts org visibility.

This predates the PR and is outside #3839's scope (which is about tenant boundaries), so it must not block this merge — the change is a strict improvement. Flagging it so it can be tracked as a follow-up: an org-visibility guard on `customer_accounts.domain_mapping` create would close it for this endpoint.

---

### Not applicable / clean

- No removed or renamed event ids, response fields, exports, columns, or import paths.
- No new `any`, no arbitrary Tailwind values, no DS token violations (no UI files touched).
- No entity change → no migration or `.snapshot-open-mercato.json` update expected.
- No duplication with `develop`: the three changed files have no equivalent commit in the base branch.
- No active spec under `.ai/specs/` covers custom-domain registration, so there is no spec divergence to flag.

Nice work — the diff is minimal, lands the check at the one place that closes every caller, and the red-before-green evidence in the PR body is the right level of rigor for a security fix.

pkarw:
Label set to `merge-queue` because the required review gates passed and QA is not required (`skip-qa`). Keeping `priority-low` and `risk-medium`: the write is admin-gated and the corrupted row lands in the caller's own tenant (`tenantId` is server-stamped), so this is a single-tenant integrity bug rather than a cross-tenant leak.

pkarw:

Reviewed at head `768edc58` against GitHub check results (all green; no local validation fallback needed). 0 blocker / 0 major / 0 minor / 2 nit — both nits are non-blocking follow-ups (route-wide i18n pass on error bodies; an org-visibility guard for the pre-existing intra-tenant cross-org gap). No autofix iterations were needed.

PR #5203, opened 2026-08-11
fix(search): filter hybrid search results by per-entity view features (#5168)

pkarw:
Closes #5168

## Summary

`GET /api/search/search` — the hybrid endpoint behind the admin **Vector Search playground** — applied tenant and organization scoping but **no per-entity ACL**. A caller past its single `search.view` gate received presenter titles, subtitles and deep links for every indexed entity type, including ones they hold no view feature for. This is the same defect #5167 fixed on the global endpoint; #5167 scoped itself to the palette surfaces and deliberately left this one.

The route now mirrors the global route line for line:

- resolve `rbacService` and `searchIndexer`, **fail closed with 503** when either is unregistered
- **narrow** `searchOptions.entityTypes` to the caller's readable set *before* querying, so `limit` is not spent on records that would only be dropped
- **filter the results again on the way out**, as defense in depth against a strategy that ignores `entityTypes`
- superadmins exempt

`requireFeatures` is deliberately **unchanged**: `search.view` is the right gate for an administration surface, and `TC-SEARCH-003` pins it.

## ⚠️ This fix shipped once before and was reverted — here is why it will stick

PR #5169 implemented exactly this change, merged into #5167's stacked branch (`d2893d1e`), and was then **reverted in `358492396`** before #5167 reached `develop`. So `develop` never received it and the issue stayed live — verified on `8e7019cdf`: the route had no RBAC wiring and `hybrid-search-acl.routes.test.ts` did not exist.

The revert's stated trigger was that the fail-closed filter broke **`TC-EXAMPLE-001`**. The root cause was never in the filter: `example:todo` is indexed through its CRUD route's `indexer: { entityType }`, but **no `search.ts` declared it at all**, so `canReadSearchEntity` denied it as `unconfigured` for the non-superadmin `admin` the test logs in as.

This PR fixes that properly rather than weakening the filter:

- **`apps/mercato/src/modules/example/search.ts` (new)** declares `aclFeatures: ['example.todos.view']` — the same feature `GET /api/example/todos` enforces — plus `formatResult`, `resolveUrl` and `fieldPolicy`, so the reference module now models the `packages/search/AGENTS.md` MUST rules instead of violating them.
- **Mirrored into `packages/create-app/template/`**, which carries its own copy of the example module and its own `TC-EXAMPLE-001`.
- `example:example_customer_priority` is **deliberately left unconfigured** — it holds a customer id and a priority enum with no human-readable text, so indexing it would only add noise. Documented in the file.

[...]

PR #5465, opened 2026-08-22
fix(ai-assistant): enforce per-entity ACL in legacy search tool pack

### LINKED ISSUE #5211: Fix: enforce per-entity ACL in the legacy AI search tool pack
## Summary

The legacy AI search tool pack exposes records using only the coarse `search.view` feature and does not enforce the owning entity's read ACL. A user who can invoke an agent that includes these tools can therefore retrieve indexed metadata for entity types they are not allowed to read directly.

This is the AI-tool sibling of #5168, but #5168 explicitly excludes AI tools and its fix PR #5203 is limited to the HTTP search route.

**Security classification:** priority high, risk high. This is a deterministic confidentiality boundary failure across protected entity types within the caller's tenant/organization. The blast radius includes every indexed entity exposed by the affected tools.

## Analysis

The legacy pack registers `search.hybrid_search` and `search.get_record_context` with only `search.view`:

- `packages/ai-assistant/src/modules/ai_assistant/ai-tools/search-pack.ts:36-70`
- `packages/ai-assistant/src/modules/ai_assistant/ai-tools/search-pack.ts:78-118`
- `packages/ai-assistant/src/modules/ai_assistant/ai-tools.ts:11-18`

Both handlers call the raw search service without applying the owning entity's required features. These tools are included by production customer and catalog agents:

- `packages/core/src/modules/customers/ai-agents.ts:78-79,445-460`
- `packages/core/src/modules/catalog/ai-agents.ts:82-83,382-383`

The newer search-owned AI tools already demonstrate the intended control by resolving per-entity access before returning results:

- `packages/search/src/modules/search/ai-tools.ts:184-213,282-300`
- `packages/search/src/modules/search/lib/entity-access.ts:48-73,90-131`

Because the legacy and newer packs coexist, fixing only #5168/#5203 leaves the AI path open.

## How to implement

- Route both legacy tools through the shared per-entity access resolver used by the search-owned AI tools.
- Reject an explicitly requested unauthorized entity type.
- Filter mixed search results before returning records or context, including related/context expansion.
- Keep tenant and organization scoping mandatory in addition to entity ACL checks.
- Add regression tests for a role with `search.view` but without the target entity's view feature, covering both legacy tool IDs and mixed result sets.
- Consider consolidating the duplicate implementations after preserving the stable tool IDs and module override behavior.

## Compatibility notes

Keep the existing tool IDs, schemas, and registration paths because AI tool IDs and ACL feature identifiers are stable integration surfaces. The intended behavioral change is security-narrowing: callers should receive only results they could read through the owning module.

## How to pick up

1. Reproduce with an authenticated role that has `search.view` and access to an agent exposing the legacy search pack, but lacks the target entity's view feature.
2. Add failing tests at the legacy tool boundary.
3. Reuse `packages/search/src/modules/search/lib/entity-access.ts` rather than creating another ACL map.
4. Run the focused AI-assistant/search tests, then the repository's configured validation gate.

## Out of scope

- The HTTP hybrid-search endpoint tracked by #5168 and #5203.
- Search index tenant/organization scoping unrelated to entity ACL selection.
- Renaming or removing legacy AI tool IDs.

Found during an authorized defensive security assessment of commit `a8d208fcdc43ce565e8ab78ec593873ab4ff53b5`.

haxiorz:
Closes #5211

## Goal

Restore the confidentiality boundary for the legacy AI search pack so a user who holds `search.view` but lacks an entity's view feature cannot retrieve that entity's indexed metadata via `search.hybrid_search` or `search.get_record_context`.

## Problem and root cause

`packages/ai-assistant/src/modules/ai_assistant/ai-tools/search-pack.ts:36-70,78-118` gated both tools only on `search.view` and forwarded `entityTypes` verbatim to `searchService.search`. The newer `packages/search/src/modules/search/ai-tools.ts:184-213,282-300` and `packages/search/src/modules/search/lib/entity-access.ts:48-131` already gate via `aclFeatures` (`resolveReadableEntityTypes` + `filterSearchResultsByEntityAccess`). Because the legacy and search-owned packs coexist and are whitelisted by `customers`/`catalog` agents, HTTP fix #5203 left the AI path open.

## What changed

- `packages/shared/src/lib/search/entityAccess.ts` — extract canonical per-entity ACL helper (wildcard-aware, fail-closed) for reuse.
- `packages/search/src/modules/search/lib/entity-access.ts` — re-export from shared (BC alias).
- `packages/ai-assistant/src/modules/ai_assistant/ai-tools/search-pack.ts` — `hybrid_search` narrows `entityTypes` to readable set, short-circuits on empty, then defense-filters results; `get_record_context` authorizes single `entityId` and defense-filters the 5-row window. Tenant/organization scoping retained, tool IDs/schemas unchanged.
- `packages/ai-assistant/src/modules/ai_assistant/ai-tools/__tests__/search-pack.test.ts` — inject `searchIndexer` mocks, keep original happy-path coverage, add regression tests for `search.view`-only role on both tool IDs and mixed sets.

## Validation

- `yarn build:packages` — passed (shared 265, search 98, ai-assistant 232 entry points)
- `yarn generate` — passed (no new surfaces)
- `yarn i18n:check-sync` — passed
- `yarn i18n:check-usage` — passed
- `yarn typecheck` — passed
- `yarn test` — passed (search-pack ACL cases, entity-access)
- `yarn build:app` — passed
- Regression: `hybrid_search` filters to readable, explicit unauthorized short-circuits, mixed set intersects, defense drops stray; `get_record_context` throws `Insufficient permissions` when lacking target view, succeeds when granted, filters stray strategy result.

## Automated review

- `om-code-review` — approve (direct author pass, 0 findings, policy satisfied)

## Breaking changes

None. Tool IDs, schemas, and registration paths are unchanged. Behavioral change is security-narrowing: callers receive only results they could read via the owning module's CRUD route. No DB migration, no config change.

## Labels and delivery

- `review` — PR is ready for code review.
- `security` — fixes deterministic cross-entity confidentiality leak in AI tool pack.
- `skip-qa` — no UI-rendering files, no DB/API contract breakage, change is covered by new unit/regression tests in the same PR.
- `priority-high` — security confidentiality boundary failure blocks release.
- `risk-high` — affects every indexed entity exposed by the legacy pack.

## Manual QA

Not required for merge per `skip-qa` rationale. Optional spot-check: create a role with `search.view` + agent access but without `customers.people.view`; invoke `search.hybrid_search` with `q:"test"` and `entityTypes:["customers:customer_person_profile"]` and verify `totalResults:0`; repeat `search.get_record_context` and verify it throws `Insufficient permissions`.

## Evidence

Manual harness `tsx /tmp/test-search-pack.mjs` ALL PASS, review packet `3a3f095…`

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5465"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789953576&installation_model_id=432243&pr_number=5465&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5465&signature=439708542bdfec819392b335f933a943a55e2cb480a4bb1768c1ad07a16d741c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T02:45:45Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
## Code Review: fix(ai-assistant): enforce per-entity ACL in legacy search tool pack

**Verdict: CHANGES REQUESTED** (this repo requests changes on Minor findings too — see `.ai/skills/om-auto-review-pr/SKILL.md`)

**Validation source: GitHub checks** (no local re-run). All PR checks green except one unrelated failure — see CI section.

---

### Summary

The security fix itself is **correct and well-targeted**. I traced the leak path and the closure holds:

- `hybrid_search` now narrows `entityTypes` to the readable set *before* querying, short-circuits to an empty result when nothing is readable, and defense-filters the returned rows. Narrowing up front (rather than filtering after) is the right call — it keeps `limit` meaningful instead of spending the result budget on rows that get dropped.
- `get_record_context` authorizes the requested `entityId` up front and defense-filters the 5-row window.
- Superadmin bypass, tenant/organization scoping, tool IDs, and schemas are all preserved.
- Wildcard-aware matching is inherited correctly via `authorizeFeatures`.
- Moving the helper into `@open-mercato/shared` to avoid a `search ↔ ai-assistant` cycle is the right structural call, and `export *` from the old path keeps the existing import surface intact (BC preserved).

I also checked two things that *could* have been blockers and were not:

- **No new availability failure mode.** The tools now hard-depend on `searchIndexer`. That is safe: `searchService` and `searchIndexer` are registered in the same `container.register({...})` block (`packages/search/src/di.ts:289-293`), so any container that can resolve one can resolve the other.
- **The new subpath resolves.** `@open-mercato/shared/lib/search/entityAccess` is matched by the `./*/*/*` wildcard in `packages/shared/package.json` (`types: ./src/*/*/*.ts`, `default: ./dist/*/*/*.js`).

Test coverage is genuinely good: narrowing, explicit-unauthorized short-circuit, mixed-set intersection, defense-in-depth drops, superadmin passthrough, and both authorized/unauthorized `get_record_context` paths.

What holds this back is a cluster of code-quality issues in the new `ai-assistant` code, not the security logic.

---

### Findings

#### Minor 1 — Four unnecessary `as any` casts in production code

`packages/ai-assistant/src/modules/ai_assistant/ai-tools/search-pack.ts:43,96,150,153`

```ts
const lookup = resolveSearchIndexer(ctx as any)     // :43, :96, :153
authorizeEntityAccess(input.entityId, ctx as any)   // :150
```

`AGENTS.md` → Code Quality is explicit: **"No `any` types."** These casts are also gratuitous — the tool context already declares exactly what the helpers need, in `packages/ai-assistant/src/modules/ai_assistant/lib/types.ts:12-14`:

```ts
container: AwilixContainer
userFeatures: string[]
isSuperAdmin: boolean
```

That is structurally assignable to both helper parameter types as written, so the casts buy nothing and actively suppress the checking that would catch a future context-shape change. Please type the helper params against the real context type (or just drop the casts).

#### Minor 2 — `authorizeEntityAccess` duplicates the shared helper's security decision logic

`packages/ai-assistant/src/modules/ai_assistant/ai-tools/search-pack.ts:42-58`

The function re-implements the `unconfigured` and `no-acl-features` checks that `canReadSearchEntity` already performs internally, then calls `canReadSearchEntity` anyway — so every deny condition is now encoded in two places.

The shared helper already has the seam for this. `SearchEntityAccessOptions.onDeny(entityId, reason)` reports exactly these three reasons (`'unconfigured' | 'no-acl-features' | 'insufficient-features'`), which is what the duplicated block is reaching for:

```ts
let denial: SearchEntityDenyReason | undefined
if (!canReadSearchEntity(entityType, lookup, subject, { onDeny: (_, reason) => { denial = reason } })) {
  throw new SearchToolAuthorizationError(messageFor(entityType, denial))
}
```

This matters beyond tidiness: if `canReadSearchEntity` ever gains a fourth deny condition, the local pre-checks silently go stale and the two paths drift on a security boundary.

#### Minor 3 — The extraction stripped the security rationale from the shared helper

`packages/shared/src/lib/search/entityAccess.ts` (new) vs. the deleted body of `packages/search/src/modules/search/lib/entity-access.ts`

The move is otherwise byte-for-byte faithful, but every JSDoc block was dropped. What was lost is the non-obvious security contract, not decoration:

- that the function **fails closed**, and *why* an unconfigured entity or one with no `aclFeatures` is never exposed;
- why narrowing happens up front rather than post-filtering (the `limit`-budget argument);
- why `onDeny` exists at all — that a silent drop from a module forgetting `aclFeatures` is otherwise indistinguishable from "no matches";
- why decisions are memoized per entity type in `filterSearchResultsByEntityAccess`.

This is now the canonical, cross-package ACL helper, so it is exactly the file where that reasoning earns its keep. Please carry the original comments across.

#### Nit 4 — `searchIndexer` resolved twice per `get_record_context` call

`search-pack.ts:150,153` — `authorizeEntityAccess` resolves the indexer internally, then the handler resolves it again on the next line. Have `authorizeEntityAccess` return the lookup, or hoist the resolve above both.

#### Nit 5 — Indexer resolved before the superadmin short-circuit

`search-pack.ts:43-44` — `resolveSearchIndexer(...)` runs before `if (ctx.isSuperAdmin) return`, so a superadmin gets `Search entity registry unavailable` in a situation where the check would otherwise be skipped entirely. Move the superadmin check first.

---

### CI

`ephemeral-integration (13/15)` is failing, but **not because of this PR**:

- Failing test: `packages/core/src/modules/warranty_claims/__integration__/TC-WC-020-sla-escalation.spec.ts:255` — `Error: SLA sweep queue should process at least one job` (`expect(processed).toBeGreaterThan(0)`), 149 passed / 1 failed.
- This diff touches only `packages/shared`, `packages/search`, and `packages/ai-assistant`. Nothing in warranty-claims or the queue.
- The spec has prior flakiness history (`50f1f0c4b fix(warranty_claims): make the SLA sweep spec independent of queue residency`), and that commit **is** already in this PR's head — so this is a residual timing flake in that spec, not a regression here.

`develop` branch protection configures no required status checks (only 1 required approving review), so this is not a merge blocker. Worth a re-run; it should not gate this PR.

Everything else is green: `lint`, `test`, `prepare`, `ds-lint`, `audit-scope`, `scope`, `documents-multi-instance`, `mutate (shared, src/lib/search/entityAccess.ts)`, `license/cla`, and the other 14 integration shards.

---

### Breaking changes

None. Tool IDs, input schemas, and registration paths are unchanged; the old `entity-access` import path still re-exports everything. The behavior change is security-narrowing only. No migration, no config change. Consistent with `BACKWARD_COMPATIBILITY.md`.

---

### Verdict

3 Minor + 2 Nit → **changes requested** under this repo's rule (Minor findings request changes). None of these touch the correctness of the security fix — they are all confined to the new `ai-assistant` glue code and the doc loss in the extraction, and should be a quick pass.

Nice work on the root-cause writeup and the regression tests; the narrow-then-defense-filter shape is the right pattern here.

pkarw:
Label set to `changes-requested` because review found actionable issues (3 Minor, 2 Nit — this repo requests changes on Minor findings).

pkarw:
Thanks @haxiorz — the security fix itself looks correct; review found some code-quality items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pkarw:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED. 3 Minor + 2 Nit findings, all confined to the new `ai-assistant` glue code and the JSDoc loss in the shared extraction — the security fix itself verified correct. No autofix iterations ran (fork PR; awaiting a decision on how to apply the polish). Lock released.

pkarw:
@haxiorz please fix the failing CI

haxiorz:
Fixed review findings (3 Minor + 2 Nit):

- Removed 4 `as any` casts — helpers now typed against `McpToolContext`
- Deduplicated `authorizeEntityAccess` via `canReadSearchEntity` + `onDeny` (no longer encodes deny reasons twice)
- Restored JSDoc security rationale in `packages/shared/src/lib/search/entityAccess.ts` (fail-closed, limit-budget, onDeny, memoization)
- Hoisted `searchIndexer` resolve — single resolve per `get_record_context` call
- Superadmin check now before `resolveSearchIndexer` (avoids spurious registry error)

Also fixed `ws-c-tool-pack-coverage` integration test to mock `searchIndexer` (was failing CI). All 5 findings addressed; ready for re-review.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-23T00:21:13Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review (Re-review): fix(ai-assistant): enforce per-entity ACL in legacy search tool pack

## 🎯 Summary

This is a re-review of head `a47eeeaf` (`fix(review): address code-quality findings for legacy search ACL`), pushed after my `changes-requested` review on `f27f2218`. The scope is unchanged: five files across `packages/shared`, `packages/search`, and `packages/ai-assistant`, closing the confidentiality gap where a caller holding only `search.view` could read any indexed entity's metadata through `search.hybrid_search` or `search.get_record_context`.

The security fix was already correct in the previous round and it still is — I re-verified the whole closure rather than assuming the last pass carried over. `hybrid_search` short-circuits on superadmin, otherwise resolves `searchIndexer` fail-closed, narrows `entityTypes` to the readable set before querying, returns an empty result without touching the search service when nothing is readable, and defense-filters the rows that come back. `get_record_context` authorizes the single requested `entityId` up front and defense-filters the 5-row window. Crucially, `resolveReadableEntityTypes` only returns `undefined` (meaning "no filter") on the superadmin path, so a non-privileged caller can never reach `searchService.search` with an unrestricted `entityTypes` — that is the property the whole fix rests on and it holds.

All five findings from the previous round are genuinely fixed, not papered over:

- **Minor 1 (`as any` casts)** — resolved. `resolveSearchIndexer` and `authorizeEntityAccess` are now typed against the real `McpToolContext` / `SearchEntityConfigLookup` / `SearchEntityAccessSubject`, and all four casts are gone. `McpToolContext` declares `container: AwilixContainer`, `userFeatures: string[]`, `isSuperAdmin: boolean` as required (`packages/ai-assistant/src/modules/ai_assistant/lib/types.ts:8-17`), so the structural assignment is checked rather than suppressed.
- **Minor 2 (duplicated deny logic)** — resolved exactly as suggested. `authorizeEntityAccess` (`search-pack.ts:44-70`) now delegates the decision to `canReadSearchEntity` and uses the `onDeny` seam to pick the message. The deny conditions live in one place again; the second `getEntityConfig` lookup that remains is only to enumerate the required features for the error text, not to re-derive the decision.
- **Minor 3 (lost JSDoc)** — resolved, and better than I asked for. I diffed `packages/shared/src/lib/search/entityAccess.ts` against the deleted body of `packages/search/src/modules/search/lib/entity-access.ts` on `develop`: the two files are now byte-identical apart from the two import specifiers. The fail-closed rationale, the `limit`-budget argument for narrowing up front, the reason `onDeny` exists, and the memoization note all survived the move.
- **Nit 4 (double `searchIndexer` resolve)** — resolved. `get_record_context` hoists the resolve into a single `lookup` binding shared by the authorization call and the defense filter.
- **Nit 5 (indexer resolved before the superadmin check)** — resolved on both tools. A superadmin now never touches the registry in either handler.

Backward compatibility is unchanged from the last pass: tool ids, input schemas and registration paths are untouched, and `packages/search/src/modules/search/lib/entity-access.ts` re-exports the shared module so the old import path keeps working (`packages/search/src/modules/search/__tests__/entity-access.test.ts` still imports through it and passes). The dependency direction is correct — `shared` gained no domain imports, and both `search` and `ai-assistant` depend on `shared`, so no cycle is introduced.

I also checked whether the same leak shape exists elsewhere now that I was in the area. `packages/core/src/modules/catalog/ai-tools/merchandising-pack.ts:330-336` is the only other AI tool that reaches `searchService.search`, and it is gated on `catalog.products.view` with `entityTypes` pinned to `catalog:catalog_product`, so it is not a sibling of this bug. The remaining `searchService` resolutions in `lib/http-server.ts`, `lib/mcp-server.ts` and `lib/mcp-dev-server.ts` are index-write paths, not query paths. The closure for issue #5211 looks complete.

Only two nits remain, both introduced by the fix for my own Nit 5 and neither of them touching the security boundary.

## Verdict

✅ **approve** — every Minor from the previous round is fixed, the security logic re-verified clean, and the full validation gate is green on this head. The two remaining items are Nits, which under this repo's rule (Medium and above request changes; Low does not) do not block. Please take them or leave them at your discretion.

## 🧪 Validation Gate

**Validation source: GitHub checks** (no local re-run — CI ran the complete `validation.commands` sequence on head `a47eeeaf` in run [32594427650](https://github.com/open-mercato/open-mercato/actions/runs/32594427650)).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job, step "Build packages". |
| `yarn generate` | ✅ PASS | `prepare` job, step "Prepare generated modules". |
| `yarn build:packages` (rebuild) | ✅ PASS | `prepare` job, step "Rebuild packages with generated files". |
| `yarn i18n:check-sync` | ✅ PASS | `test` job, step "Check i18n sync". |
| `yarn i18n:check-usage` | ✅ PASS | `test` job, step "Check i18n usage". |
| `yarn typecheck` | ✅ PASS | `test` job, step "Checking types" — this is what confirms the removed `as any` casts still typecheck. |
| `yarn test` | ✅ PASS | `test` job, step "Test". |
| `yarn build:app` | ✅ PASS | `prepare` job, step "Build app". |

Beyond the configured gate, every other check on this head is green as well: `lint`, `ds-lint`, `audit-scope`, `scope`, `documents-multi-instance`, `license/cla`, the create-app template-parity and DS-token-drift guards inside the `test` job, `mutate (shared, src/lib/search/entityAccess.ts)`, and all fifteen `ephemeral-integration` shards. Notably `ephemeral-integration (13/15)` — the `TC-WC-020-sla-escalation` warranty-claims failure I called an unrelated flake last round — passes on this head, which confirms that diagnosis. Nothing is pending and nothing is failing.

## Findings

### 💅 Nit

**Nit 1 — the superadmin short-circuit duplicates the response-shaping tail.**

`packages/ai-assistant/src/modules/ai_assistant/ai-tools/search-pack.ts:112-131` vs `145-163`

Moving the superadmin check above `resolveSearchIndexer` (my Nit 5) was implemented by giving the superadmin path its own complete `service.search(...)` call plus its own copy of the `timingMs` / `strategiesUsed` / return-object construction. The two tails are now identical for about eighteen lines, so a future change to the response shape — an extra field, a different `strategiesUsed` derivation — has to be made twice, and a superadmin silently getting a different response shape from everyone else is an unpleasant class of bug to track down.

A single tail with the ACL work as a prologue keeps the short-circuit you asked for without the copy:

```ts
let entityTypes = input.entityTypes
let lookup: SearchEntityConfigLookup | undefined
if (!ctx.isSuperAdmin) {
  lookup = resolveSearchIndexer(ctx)
  entityTypes = resolveReadableEntityTypes(lookup, subject, input.entityTypes)
  if (entityTypes && entityTypes.length === 0) return emptyResult(input.q, started)
}
const rawResults = await service.search(input.q, { tenantId: ctx.tenantId, organizationId: ctx.organizationId, limit, strategies: input.strategies as SearchStrategyId[] | undefined, entityTypes })
const results = lookup ? filterSearchResultsByEntityAccess(rawResults, lookup, subject) : rawResults
// …one shared tail
```

That is also the shape `get_record_context` already uses at `:190-203`, so the two handlers would read the same way. This is genuinely optional — the current code is correct, just duplicated.

**Nit 2 — no test pins the "superadmin does not need `searchIndexer`" behavior.**

`packages/ai-assistant/src/modules/ai_assistant/ai-tools/__tests__/search-pack.test.ts:211-227` and `:370-385`

Both superadmin tests still register a working `searchIndexer` in the mock container, so they pass identically before and after the Nit-5 fix. The actual point of that fix — that a superadmin succeeds even when the entity registry is unavailable — is therefore unasserted, and someone re-hoisting the resolve above the superadmin check would not break a single test.

A one-line variant of each existing superadmin case would lock it in: keep the same context but have the container throw on `searchIndexer`, and assert the handler still returns results.

```ts
it('does not require the search entity registry for a superadmin', async () => {
  const { service } = makeSearchService([{ entityId: 'secret:thing', recordId: 's1', score: 0.8, source: 'fulltext' }])
  const ctx = makeCtx({ userFeatures: [], isSuperAdmin: true })
  ;(ctx.container.resolve as jest.Mock).mockImplementation((name: string) => {
    if (name === 'searchService') return service
    throw new Error(`unexpected ${name}`)
  })
  const result = (await tool.handler({ q: 'test' }, ctx as any)) as Record<string, unknown>
  expect(result.totalResults).toBe(1)
})
```

Worth adding for `search.get_record_context` too, where the same short-circuit now exists at `:190-194`.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `canReadSearchEntity`, `resolveReadableEntityTypes`, `filterSearchResultsByEntityAccess` and their types all remain reachable from the original `packages/search/.../lib/entity-access` path via `export *`.
- [x] No function signature changed in a breaking way — the three shared helpers keep their exact parameter lists and return types; only the module-internal `authorizeEntityAccess` / `resolveSearchIndexer` signatures changed, and neither is exported.
- [x] No required type field removed or narrowed — `SearchEntityConfigLookup`, `SearchEntityAccessSubject`, `SearchEntityDenyReason` and `SearchEntityAccessOptions` are re-exported unchanged.
- [x] No HTTP route URL removed or renamed; no method changed — this change touches no API route.
- [x] No field removed or retyped in an existing response shape — both tools return the same object shape; the only behavioral difference is that unreadable rows are absent, which is the intended security narrowing.
- [x] No event or message name renamed or removed — no events touched.
- [x] No CLI command or flag renamed or removed — no CLI surface touched.
- [x] No database table or column renamed or removed — no schema change and no migration.
- [x] No config key renamed and no default changed silently — no config surface touched.
- [x] Where a contract had to change: the old `entity-access` import path is preserved as a re-export, so third-party module code importing it keeps compiling. The AI tool ids `search.hybrid_search` and `search.get_record_context` and their Zod input schemas are byte-identical, so agent `allowedTools` whitelists are unaffected.

## 🧪 Test Coverage

`packages/ai-assistant/src/modules/ai_assistant/ai-tools/__tests__/search-pack.test.ts` carries the regression suite and it is the right one. For `search.hybrid_search` it covers: scope and limit pass-through, the default limit of 20, the missing-tenant throw, withholding results for entity types the caller cannot view, the short-circuit that never calls the search service when an explicit request resolves to nothing readable, intersection of a mixed explicit request with the readable set, the defense-in-depth drop of a stray row a strategy returned for an unreadable type, and superadmin passthrough. For `search.get_record_context` it covers: the happy path with presenter/url/links, the `{ found: false }` miss, tenant isolation, the missing-tenant throw, rejection of an explicitly requested unauthorized entity, the authorized retrieval, the defense-in-depth filter on the 5-row window, and superadmin passthrough. `packages/search/src/modules/search/__tests__/entity-access.test.ts` continues to exercise the shared helpers through the old import path, which doubles as the re-export's regression test. The integration mock in `__tests__/integration/ws-c-tool-pack-coverage.test.ts:156-170` was updated to register `searchIndexer`, which is the right minimal change now that the tool depends on it.

The one gap is Nit 2 above: the superadmin registry-independence introduced in this commit is not asserted anywhere. Everything else that changed in `a47eeeaf` is behavior-preserving refactoring already covered by the existing cases, and the mutation-testing check on `packages/shared/src/lib/search/entityAccess.ts` passing is decent independent evidence that the assertions on the shared helper actually bite.

pkarw:

- 🚀 `merge-queue` — the re-review passed with only two Nit-level findings, every Minor from the previous round is fixed, and the full validation gate is green on head `a47eeeaf`, so this PR is ready to merge.
- 🔒 `security` — the change closes a per-entity confidentiality leak in the legacy AI search tool pack (issue #5211), where `search.view` alone exposed any indexed entity's metadata.
- ⏭️ `skip-qa` — the diff touches no UI-rendering file, changes no database structure or API contract, and ships its own unit and regression coverage in the same PR, which satisfies the automated-verification exemption.
- 🔺 `priority-high` — a confidentiality boundary failure on an AI tool surface should not ship in a release, so this blocks the next one until merged.
- ⚠️ `risk-high` — the helper is now the canonical cross-package ACL gate and the behavioral change reaches every entity type the legacy pack can return, so the blast radius is broad even though the change is security-narrowing.

Previous state was `changes-requested`; this comment supersedes the earlier plain-text label note from the first review pass.

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED. Re-review of head `a47eeeaf`: all three Minor and both Nit findings from the previous pass are fixed — the `as any` casts are gone and typed against `McpToolContext`, `authorizeEntityAccess` now delegates to `canReadSearchEntity` via the `onDeny` seam, the shared `entityAccess.ts` is byte-identical to the original apart from import paths (JSDoc fully restored), and both handlers short-circuit on superadmin before resolving `searchIndexer`. I re-verified the security closure itself rather than carrying the previous pass forward, and also confirmed no sibling leak exists in `catalog/ai-tools/merchandising-pack.ts`. Two Nits remain (duplicated response tail in the superadmin branch; no test pinning superadmin registry-independence) — neither blocks. Validation source: GitHub checks — the full `validation.commands` gate plus lint, ds-lint, mutation testing and all 15 integration shards are green on this head, including the `ephemeral-integration (13/15)` shard that flaked last round. Labels moved `changes-requested` → `merge-queue`. autofix: skipped (not my PR — re-run with --autofix to fix it here). No CI follow-up needed: every check is already settled and green. Lock released.

PR #4561, opened 2026-07-27
feat(documents): collaborative internal documents module with realtime co-editing

haxiorz:
Tracking plan: .ai/runs/2026-08-12-documents-collaborative-editor.md
Status: ci-monitoring

## Summary

Open Mercato has no collaborative document editor. The closest primitives are single-user rich editors, an SSE-only (one-way) broadcast bridge, enterprise pessimistic record locking — which exists to *prevent* concurrent edits — and blob-storage attachments. None of them allow two people to write in the same document at once.

This PR adds `@open-mercato/documents`: a tenant/organization-scoped backoffice module where staff author rich-text documents together in real time. Editing runs on a **TipTap v3** surface backed by a **Yjs** CRDT, synced over a **Hocuspocus** WebSocket sidecar with live-cursor presence. Documents live in folders, are shared per document (owner / editor / commenter / viewer), carry inline comments with @mentions that raise notifications, keep version history, and export to `.docx` and PDF.

Beyond the editor it is a first-class ERP work surface: documents link to business records (people, companies, deals, products, catalog offers, quotes, sales orders, and other documents) through a typed registry with label snapshots, inject a related-documents panel into host record pages, instantiate from contextual templates with server-resolved entity autofill, and support archive / favorites / duplicate / watch lifecycle.

This introduces the platform's **first bidirectional real-time transport**. The sidecar's `onAuthenticate` hook is the single security chokepoint: it validates the Origin, verifies a short-lived per-document token, binds the token to the requested room, and re-resolves authorization after every async boundary before a client may join. Every layer is MIT/permissive-licensed and self-hosted.

The collaboration sidecar is **optional**. With its environment variables unset, the editor degrades cleanly to single-user editing and the rest of the module is unaffected.

## Changes

- **New workspace package `packages/documents`** (374 files) following the `packages/checkout/` package pattern and `customers` CRUD conventions — 26 API routes, 8 migrations, backend pages, widgets, ACL, events, notifications, DI and i18n.
- **Realtime sidecar** `packages/documents/server/documents-collab-server.ts` — Hocuspocus + Yjs, single-node by default, optional Redis fan-out for multi-instance deployments with a durable post-persistence publish path, bounded ingress/awareness/payload limits, and a `/healthz` endpoint that fails closed on misconfiguration.
[...]

PR #5467, opened 2026-08-22
fix(ci): stabilize develop CI at the root causes

pkarw:
## Why

`develop` CI has been red almost continuously. Over the last 200 runs on the branch:

| Workflow | Result |
|---|---|
| **Develop Snapshot Release** | **29 failure / 3 success** — every failure is the `Standalone App Integration Tests` job |
| CI for Develop&Main | 4 failure / 17 success |
| CodeQL, Mutation tests, Skills Tiers Lint, Scheduled Dependency Audit | green |

This PR fixes the root causes behind those failures, plus the guard that would have caught
the worst of them at PR time.

### 1. `happy-dom` missing from `@open-mercato/documents` runtime deps — breaks every standalone build

Since #4561 landed (2026-08-20 23:47 UTC) **every** snapshot run fails with:

```
Error: Turbopack build failed with 1 error:
./node_modules/@tiptap/html/dist/server/index.cjs:7:17
Error: Module not found: Can't resolve 'happy-dom'
```

`@tiptap/html` declares `happy-dom` as a **non-optional peerDependency**, and
`documents/lib/collabMaterializer.ts` uses `@tiptap/html/server`. `happy-dom` was declared
only in `devDependencies`, which yarn hoists into the monorepo's shared `node_modules` — so
`yarn build:app` resolves it and every PR check stayed green. A standalone app installs
`dependencies` + `peerDependencies` only, so there it is simply absent.

The pre-existing `createRequire(import.meta.url)` lazy require does not help: Turbopack
statically analyses `createRequire(...)(literal)` the same as a top-level `require`.

**Fix:** move `happy-dom` to `dependencies` of `@open-mercato/documents`. `yarn.lock` is
unchanged (yarn records a workspace's dev and prod deps in one `dependencies` block), so
`yarn install --immutable` still passes.

### 2. `OM_ENABLE_PUSH_STUB_ADAPTER` never mirrored into the standalone job — TC-PUSH-003 can never pass

Before the build broke, the standalone job's one reproducible hard failure was:

```
TC-PUSH-003 › ... drives the worker to a sent delivery
Expected: "sent"   Received: "failed"
```

The network-free `push_stub` channel adapter is registered by `push_notifications/di.ts`
only when `OM_ENABLE_PUSH_STUB_ADAPTER` is set. The ephemeral harness sets it on **both**
the Playwright process (`buildReusableEnvironment`) and the app server. The standalone job
instead starts the app with `yarn start` from a hand-written `.env` and runs Playwright with
its own `env:` block — neither carried the flag, so no adapter could ever resolve and every
delivery landed in `failed`.

**Fix:** mirror the flag in both places in `snapshot.yml` and `npm-snapshot-preview.yml`.
Both are needed: the app server reads `.env`, while `drainIntegrationQueue('push-deliveries')`
spawns a **child of the Playwright process** that never reads the app's `.env` — and with
`AUTO_SPAWN_WORKERS=false` that child is what actually executes the push job.

### 3. TC-WC-020 asserted on the wrong side of a queue race — real `CI for Develop&Main` red

```
Error: SLA sweep queue should process at least one job
expect(processed).toBeGreaterThan(0)
```

The ephemeral harness defaults `AUTO_SPAWN_WORKERS` to `true`, so the app server runs its
own `warranty_claims.sla_sweep` worker next to the spec's explicit drain. The local queue
strategy removes a job **only after its handler resolves**, so when that worker wins the
race the drain child finds an empty queue and legitimately reports `0` — even though the
sweep already ran to completion.

**Fix:** keep the fast path (`processed > 0`) and, only when it is `0`, poll the queue's
`state.json` `completedCount` until it advances. That counter is written after a handler
resolves, so it is durable proof the sweep ran regardless of which process ran it — no new
race is introduced.

### 4. TC-EUDR-010: step budgets larger than the test budget

The spec waits up to 30 s for the picker's search request and 30 s for the post-save
navigation, inside the config's 20 s `timeout`. On a loaded server the test timeout fires
before any step can spend its allowance, which is why it showed up as flaky-on-retry in the
standalone run. Marked `test.slow()`, matching TC-PUSH-003's existing use.

### 5. Prevention: no PR job ever builds a standalone app

Root cause 1 is a *class* of bug, not an incident: `npm-snapshot-preview.yml` and
`package-previews.yml` are `workflow_dispatch`-only, so packaging regressions stay invisible
until the post-merge snapshot run.

Added `scripts/check-package-peer-deps.mjs` (`yarn packages:check-peer-deps`), wired into
the `lint` job. It fails when a published workspace package's runtime dependency has a
non-optional peer the package does not itself ship, and calls out the devDependency case
explicitly. The 29 pre-existing cases (type-only peers, React runtime owned by the host app,
workspace peers the scaffolded app installs directly) are pinned with individual reasons in
`scripts/package-peer-deps-allowlist.json`, so the gate blocks new breakage without
demanding an unrelated cleanup first.

Verified against the real regression: reverting fix 1 makes the gate exit 1 with

```
@open-mercato/documents: dependency "@tiptap/html" requires peer "happy-dom"
  (declared as a devDependency — move it to dependencies)
```

## Already fixed upstream — no action needed

- **TC-WC-028** (portal executable upload returned 500 instead of 400) was fixed by the
  `Symbol.for` marker on `ScopedAttachmentUploadError` in #4561; it is only invisible today
  because the build in fix 1 fails first.
- **GSM-314** data-sync heartbeat unit failure (`resolveProviderKey is not a function`) is
  already fixed on `develop` — the mock now exports it.
- Two `ephemeral-integration` reds were `actions/checkout` download 429s — GitHub infra, no
  code fix available.

## Known remaining, deliberately not changed

- `TC-TRANS-005` is occasionally flaky-on-retry in the standalone run with no structural
  timeout mismatch to point at. It is retry-covered and does not fail the job, so it is left
  alone rather than papered over with a blanket timeout bump.

## Validation

Run locally on this branch (no compose `app` container running — local runner):

- `yarn install --immutable` ✅
- `yarn build:packages` ✅
- `yarn generate` ✅ (no diff)
- `yarn build:packages` ✅
- `yarn i18n:check-sync` ✅
- `yarn i18n:check-usage` ✅
- `yarn typecheck` ✅
- `yarn lint` ✅
- `yarn test` ✅
- `yarn build:app` ✅
- `yarn test:scripts` ✅ (includes the 6 new peer-dep guard tests)
- `yarn test:repo-wide-guards` ✅
- `node scripts/audit-ci.mjs` ✅
- `yarn packages:check-peer-deps` ✅

## QA

`skip-qa`: no UI-rendering file is touched (no `.tsx` outside integration specs, nothing
under `packages/ui/src/` or `**/components/**`), the database structure and API surface are
unchanged, no `BACKWARD_COMPATIBILITY.md` contract is broken (the manifest change is
additive), and the changed behaviour is covered by automated tests in this PR.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5467"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789960816&installation_model_id=432243&pr_number=5467&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5467&signature=e1505882538b5b4a0be656f1b16cc7d59fed2dc0162d778a137a09d258ba42e2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T04:05:54Z. Other auto-skills will skip this PR until the lock is released.

pkarw (commented):
> **Submitted as a comment review, not a formal `REQUEST_CHANGES`** — GitHub refuses `addPullRequestReview` with a `CHANGES_REQUESTED` state on your own pull request (`Can not request changes on your own pull request`), and the reviewer account here is the PR author. The verdict below is authoritative: **CHANGES REQUESTED**, reflected in the `changes-requested` pipeline label.

# 🔍 Code Review: fix(ci): stabilize develop CI at the root causes

## 🎯 Summary

This PR attacks four distinct `develop` CI failures at their causes rather than papering over them, and adds a guard for the class of bug behind the worst one. The diagnosis work is genuinely good: the `happy-dom` finding (a `@tiptap/html` non-optional peer that yarn's hoisting made invisible inside the monorepo but that is simply absent from a scaffolded standalone app) is precisely the kind of bug that only shows up post-merge, and the PR body reasons correctly about why the pre-existing `createRequire(import.meta.url)` lazy require does not save Turbopack from it. The `OM_ENABLE_PUSH_STUB_ADAPTER` mirroring is right for the right reason — the app server reads the scaffolded `.env` while `drainIntegrationQueue` spawns a child of the *Playwright* process that does not — and the fix correctly lands in all three standalone lanes (`snapshot.yml`, `npm-snapshot-preview.yml`, `scripts/test-create-app-integration.ts`) with `standalone-portal-email-env-guard.test.ts` extended to hold that invariant. The TC-WC-020 change identifies a real drain-vs-app-worker race instead of retrying around it, and TC-EUDR-010's `test.slow()` is the correct response to per-step budgets that exceed the test budget.

Reviewed scope: the two workflow files and `ci.yml`, `scripts/check-package-peer-deps.mjs` and its allowlist and tests, `scripts/test-create-app-integration.ts`, the two integration specs, and the two `package.json` manifests.

I confirmed the central prevention claim against real CI output rather than taking it on trust: the `lint` job on this head printed `Peer dependencies OK (29 allowlisted).`, so the new gate ran against a populated `node_modules` and is not passing vacuously, and the six new script tests executed in the `Test scripts` step. The `yarn.lock`-unchanged claim also holds — every `yarn install --immutable` in `ci.yml` succeeded on this head.

Nothing here breaks a contract surface and the change is well tested. What I am flagging are two real holes in the *new guard itself* — the piece of this PR that is supposed to stop the next occurrence — plus three pieces of polish.

## Verdict

❌ **request changes** — no blockers and no majors; the code is correct and CI is fully green. Two **Minor** findings drive the verdict under this repository's stricter rule (Medium/minor requests changes): `scripts/check-package-peer-deps.mjs` can be turned green by declaring the missing peer as an *optional* peer on the workspace package, and it reports success while checking nothing when dependency manifests cannot be resolved. Both are in the guard whose entire purpose is to prevent a recurrence of root cause 1, both are a few lines to close, and I verified each with a live fixture.

## 🧪 Validation Gate

Validation source: **GitHub checks** (per this repository's GitHub-checks-first review extension — CI already ran the gate on head `ba66f16`, so it was not reproduced locally). Every check on the head SHA is green; none are pending.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job (`ci.yml:167`), 3m36s |
| `yarn generate` | ✅ PASS | `prepare` job (`ci.yml:175`) |
| `yarn build:packages` (rebuild after generate) | ✅ PASS | `prepare` job (`ci.yml:178`) |
| `yarn i18n:check-sync` | ⚪ N/A | Not wired into `ci.yml`, and no locale file is touched by this diff, so there is no signal to fall back for |
| `yarn i18n:check-usage` | ✅ PASS | `lint` job (`ci.yml:480`) |
| `yarn typecheck` | ✅ PASS | `test` job (`ci.yml:496`); on PRs this is Turbo-scoped to packages affected by the diff, which covers every package this PR touches |
| `yarn test` | ✅ PASS | `test` job, 7m10s — includes `test:scripts` (the six new `check-package-peer-deps` cases) and `yarn workspace create-mercato-app test` (the extended env guard) |
| `yarn build:app` | ✅ PASS | `prepare` job (`ci.yml:190`) |

Additional green checks on this head: `lint` (including the new `packages:check-peer-deps` step), `ds-lint`, `docker-build`, `documents-multi-instance`, `ephemeral-integration (none)`, `audit`, `audit-scope`, `scope`, all three CodeQL analyses, and `license/cla`. `mergeStateStatus` is `BLOCKED` only because branch protection is still waiting on an approving review; `mergeable` is `MERGEABLE`, so there are no conflicts against `develop`.

Template parity was not run: the diff touches nothing under `apps/mercato/src/{app,modules}`, `apps/mercato/src/i18n/`, or either `.env.example`, so the `create-app` template sync checklist does not apply.

## Findings

### 🔹 Minor

**`scripts/check-package-peer-deps.mjs:83-95` — the gate can be satisfied by declaring the missing peer as an *optional* peer, which does not fix the standalone build.**

`shipped` is built from the workspace package's `dependencies` plus its `peerDependencies` (line 83), and a peer is considered satisfied as soon as it appears in that set (line 94). The dependency's *own* `peerDependenciesMeta.optional` is respected (line 92), but the workspace package's `peerDependenciesMeta` is never consulted. So a package can silence the gate by declaring the peer optional on itself — and an optional peer is exactly what a consumer is free not to install, so the standalone build still fails with the same `Module not found`.

I confirmed this rather than inferring it. With a fixture whose workspace package declares `"peerDependencies": {"happy-dom": "^20.0.0"}` together with `"peerDependenciesMeta": {"happy-dom": {"optional": true}}`, while its runtime dependency non-optionally peer-requires `happy-dom`, the checker exits `0` with `Peer dependencies OK (0 allowlisted).` Had `@open-mercato/documents` been "fixed" that way instead of by moving `happy-dom` into `dependencies`, this gate would have gone green and root cause 1 would have shipped again.

The fix is one condition when building `shipped` — credit `peerDependencies` only for peers the package does not mark optional:

```js
const ownPeerMeta = manifest.peerDependenciesMeta ?? {}
const shipped = new Set([
  ...Object.keys(manifest.dependencies ?? {}),
  ...Object.keys(manifest.peerDependencies ?? {}).filter((peer) => !ownPeerMeta[peer]?.optional),
])
```

Worth a seventh test case in `scripts/__tests__/check-package-peer-deps.test.mjs` alongside the existing `passes when the peer is re-declared as the package own peer`, asserting that the *optional* variant of that same manifest fails.

**`scripts/check-package-peer-deps.mjs:58-66, 90, 185` — the gate reports success when it checked nothing.**

`resolveInstalledManifest` returns `null` for any dependency the linker never placed, and line 90 skips it silently. That is a deliberate and defensible choice for one missing package, but there is no floor: in a tree with no completed install, *every* dependency resolves to `null`, `violations` is empty, and line 185 prints `Peer dependencies OK (0 allowlisted).` with exit code `0` — a message indistinguishable from a real pass.

Verified: running the script with `--root` pointed at a worktree that has no `node_modules` produces exactly that output, preceded by all 29 allowlist entries listed as no-longer-applying. There is no impact on this PR — the `lint` job installs first and its log shows `29 allowlisted` — but the gate has nothing protecting it if a future workflow change moves the step ahead of the install, and a developer running it in a fresh worktree gets a false all-clear.

The stale-entry path makes this sharper: line 181 tells the user to run `yarn packages:check-peer-deps --update-allowlist` to prune those 29 entries, and in that state doing so would rewrite the allowlist to `{}` (`nextAccepted` is built solely from the empty `violations`). CI would catch the result loudly rather than silently, but the script is advising a destructive command in precisely the state where the advice is wrong.

Suggested fix: count the dependencies whose manifests could not be resolved, and when that count is non-zero while zero resolved, exit non-zero with `Cannot verify peers — no dependency manifests resolved; run \`yarn install\` first.` The same condition should make `--update-allowlist` refuse to write, and the stale-entry hint should not be printed when nothing resolved.

### 💅 Nit

**`scripts/package-peer-deps-allowlist.json:7,9,11,13,15,17,19,21,26` — nine entries point at a follow-up note that does not exist.**

Each of the nine `-> @open-mercato/ui -> react-is` entries ends with "see the follow-up note in the allowlist header", but the `note` field on line 3 only explains the allowlist's general policy and says nothing about `react-is`. A reader chasing the pointer finds nothing, and the actual latent gap — `@open-mercato/ui` peer-requires `react-is` without shipping it, so scaffolded apps rely on it arriving transitively through `recharts` / `react-transition-group` — ends up with no recorded follow-up, which is the same shape of undeclared-peer risk this PR exists to remove. Either extend the header note to describe the `react-is` situation and how it should eventually be resolved, or drop the dangling clause from the nine reasons and file a follow-up issue instead.

**`packages/core/src/modules/warranty_claims/__integration__/TC-WC-020-sla-escalation.spec.ts:222-238` — the fallback proof is queue-wide rather than job-specific.**

The race analysis is correct and the fallback is a real improvement over asserting on the drain's return value. The residual gap is that `completedCount` counts completions for the whole `warranty_claims.sla_sweep` queue, not this spec's job: `packages/core/src/modules/warranty_claims/setup.ts:138` registers a 900 s interval schedule targeting that same queue, so if a scheduled sweep happens to complete inside the 30 s poll window the assertion is satisfied without this spec's own job having run. The probability is low (a 30 s window against a 900 s period, and only reachable when the drain already returned `0`), the downstream assertions on `escalationLevel`, `slaAtRiskNotifiedAt`, and the emitted claim events still carry the real weight, and the failure mode is a confusing red rather than a false green — so this is not blocking. If you want it airtight, poll for the job's own effect instead of the counter (for example `readSlaState(breachedClaim.id!)` reaching the expected `escalationLevel`), or compare the state file's `lastProcessedId` against the id returned by `enqueueSlaSweep`.

**`scripts/check-package-peer-deps.mjs:76-81` — the gate scans only `packages/*`, not every published workspace.**

The root `workspaces` globs are `["apps/*", "packages/*", "external/official-modules/packages/*"]`. `apps/*` is correctly out of scope (nothing there is published), but `external/official-modules/packages/*` holds published official modules that are subject to exactly the same standalone-install rule. Skipping them is reasonable for now — the submodule is optional and ships its own PRs — but it is an unstated limitation. A sentence in the file header noting that official modules are out of scope and why would keep the guard's coverage honest.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff adds a script, a JSON allowlist, workflow steps, and test assertions; it exports nothing new from any package.
- [x] No function signature changed in a breaking way — no package source file is touched; the only `packages/` changes are two manifests and two integration specs plus one guard test.
- [x] No CLI command or flag renamed or removed — `packages:check-peer-deps` is a purely additive root script, and no machine-parsed output format changed.
- [x] No database table or column renamed or removed; no column type narrowed. No migration is present and none is needed.
- [x] No config key renamed and no default changed silently — `OM_ENABLE_PUSH_STUB_ADAPTER` is added to CI lanes only, never to a shipped `.env.example` or a runtime default.
- [x] Where a contract had to change: not applicable. Moving `happy-dom` from `devDependencies` to `dependencies` in `packages/documents/package.json` is additive for consumers (a standalone app gains a package it previously lacked) and, per `BACKWARD_COMPATIBILITY.md`, manifest additions are an ADDITIVE-ONLY surface. `yarn.lock` is genuinely unaffected, which every `yarn install --immutable` on this head confirms.

## 🧪 Test Coverage

Coverage is strong and each fix carries its own regression guard rather than relying on the next snapshot run to tell you whether it worked.

Root cause 1 is covered twice: `scripts/__tests__/check-package-peer-deps.test.mjs` reproduces the exact devDependency shape that broke (`fails when a runtime dependency has an unmet peer`, asserting the `move it to dependencies` hint), and `the repository itself has no unallowlisted unmet peers` runs the checker against the real repo so a future regression fails `test:scripts` and not only the `lint` job. Root cause 2 is covered by `standalone-portal-email-env-guard.test.ts`, correctly extended so `OM_ENABLE_PUSH_STUB_ADAPTER` must appear at least twice in each of the three standalone lanes — the app `.env` and the test-process env — which is the invariant that was actually violated; adding `scripts/test-create-app-integration.ts` to `STANDALONE_LANES` closes a lane the guard previously did not watch at all. Root causes 3 and 4 are test fixes and are their own evidence.

Two gaps, both in the new guard and both already stated above: there is no case asserting that an *optional* peer declared on the workspace package still fails (Minor 1 — add it next to `passes when the peer is re-declared as the package own peer`), and no case pinning the behaviour when dependency manifests cannot be resolved (Minor 2 — a fixture with `packages/` but no `node_modules` should assert a non-zero exit rather than today's `Peer dependencies OK`).

One limitation worth stating plainly, since it is inherent to the change rather than a defect: none of this can be verified end to end until the snapshot workflow runs post-merge. No PR job builds a standalone app — that is root cause 5's whole point — so the `happy-dom` and `OM_ENABLE_PUSH_STUB_ADAPTER` fixes are verified here by reasoning plus the new guards, and the first real confirmation will be the next `Develop Snapshot Release` run's `Standalone App Integration Tests` job.

---
🤖 Reviewed by `om-auto-review-pr`.

pkarw:
🏷️ label rationale

- `review` → `changes-requested` (first pass): the code review found two Minor findings in `scripts/check-package-peer-deps.mjs`; under this repository's stricter verdict rule a Medium/minor finding requests changes.
- `changes-requested` → `merge-queue` (re-review of `652a02d1`): both Minor findings are fixed with regression tests, two of three Nits are fixed, and the third is declined with a reason. Verdict is approve.
- `priority-high` kept: `develop` snapshot CI is red on nearly every run, so this blocks the next release.
- `risk-medium` kept: the diff spans CI workflows, two package manifests, integration specs, and a new build script, but touches no package runtime source.
- `bug` and `skip-qa` kept: the automated-verification exemption applies — no `.tsx` outside integration specs, nothing under `packages/ui/src/` or `**/components/**`, no database or API surface change, and the changed behaviour is covered by tests in this PR.
- `in-progress` → `ci-monitoring`: the review work is complete and reported; only the CI run on `652a02d1` is still being watched. The PR is free for anyone to act on.

⚠️ `merge-queue` records the review verdict, not merge permission: this account authored the PR, so GitHub refused both a `REQUEST_CHANGES` and an `APPROVE` review state. Branch protection still needs an approving review from another maintainer.

🤖 `om-auto-review-pr`

pkarw (commented):
# 🔍 Re-review: fix(ci): stabilize develop CI at the root causes

> Submitted as a comment review again — GitHub will not accept an `APPROVE` state on your own pull request. The verdict below is authoritative and is reflected in the `merge-queue` pipeline label, but **branch protection still requires an approving review from another maintainer** before this can merge.

## 🎯 Summary

Re-review of head `652a02d1`, which adds the autofix commit for the previous pass's findings. The original four root-cause fixes and the prevention gate are unchanged and were already sound; this pass only re-checks the two Minor findings that drove `changes-requested`, plus the nits.

Both Minor findings are closed in `scripts/check-package-peer-deps.mjs`, each with a regression test rather than a code-only change:

- **Optional-peer bypass** — `shipped` now credits `peerDependencies` only for peers the package does not mark optional in its own `peerDependenciesMeta`, so declaring the missing peer optional no longer turns the gate green. The failure output distinguishes the case with `(declared as an OPTIONAL peer — consumers may not install it)`. Covered by the new `rejects a peer the package re-declares as OPTIONAL`, which uses exactly the fixture that exposed the hole.
- **Vacuous pass** — `collectViolations` now counts resolved and unresolved dependency manifests, and `main` exits `1` with `Cannot verify peer dependencies … Run \`yarn install\` first.` when nothing at all resolved. Tolerance for a single unplaced dependency is preserved, which matters because the real repo legitimately has some. Covered by the new `refuses to pass when no dependency manifest can be resolved`.

Two of the three nits are closed as well: the `packages/*`-only scope is now stated in the file header with the reason `external/official-modules/packages/*` is out of scope, and the eleven allowlist entries that pointed at a non-existent header note now reference a real `followUp` field describing the `react-is` gap and how to close it (declare `react-is` in `@open-mercato/ui`'s dependencies, which removes all eleven at once).

## Verdict

✅ **approve** — no blockers, no majors, and both Minor findings are fixed with tests. The one remaining item is a declined nit, explained below.

## 🧪 Validation Gate

Validation source: **local fallback**, narrowly scoped to the changed files, plus **GitHub checks** for everything else. The autofix touches only `scripts/*.mjs` and one JSON file, so the affected gate is `test:scripts`:

| Command | Status | Notes |
|---------|--------|-------|
| `node --test scripts/__tests__/check-package-peer-deps.test.mjs` | ✅ PASS | 8/8, including the two new cases and the pre-existing `the repository itself has no unallowlisted unmet peers` |
| Full `validation.commands` gate | ⏳ Running in CI on `652a02d1` | It passed in full on `ba66f16`; nothing outside `scripts/` changed since, so CI is the authority here rather than a local re-run |

I also confirmed the JSON allowlist still parses and that `packages/documents/package.json` is byte-identical to the reviewed head — the fix-verification experiments left nothing behind.

One honest caveat about my own verification: I could not reproduce the original `@tiptap/html` → `happy-dom` detection end to end locally, because this review worktree has no complete install and `@tiptap/*` does not resolve here. That path is instead evidenced by CI on `ba66f16`, where the `lint` job printed `Peer dependencies OK (29 allowlisted).` against a fully installed tree, and by the fixture test that reproduces the exact devDependency shape. The optional-peer and no-install holes I did verify directly, with live fixtures, both before and after the fix.

## Findings

### 💅 Nit — declined, with reason

**`packages/core/src/modules/warranty_claims/__integration__/TC-WC-020-sla-escalation.spec.ts:222-238` — the fallback proof remains queue-wide rather than job-specific.**

Restating so it ends visibly accounted for: `completedCount` counts completions for the whole `warranty_claims.sla_sweep` queue, and `warranty_claims/setup.ts:138` registers a 900 s interval schedule against that same queue, so a scheduled sweep completing inside the 30 s poll window could satisfy the assertion without this spec's own job having run.

I am deliberately leaving this as-is. The window is narrow (30 s against a 900 s period, and only reachable once the drain has already returned `0`), the substantive assertions on `escalationLevel`, `slaAtRiskNotifiedAt`, and the emitted claim events still carry the real weight, and — decisively — the failure mode is a confusing red rather than a false green. Set against that, rewriting the synchronisation of an integration spec that cannot be executed in this environment is a worse trade than accepting the residual window. If it does misbehave, the fix is to poll for the job's own effect (`readSlaState(breachedClaim.id!)` reaching the expected `escalationLevel`) instead of the counter.

## 💥 Breaking Changes

Unchanged from the first pass — every item still passes. The autofix adds no exported symbol, touches no package source, and changes no contract surface; `check-package-peer-deps.mjs` is an internal build script whose only behaviour change is to reject two states it previously accepted, which is strictly stricter and affects no consumer.

## 🧪 Test Coverage

The gate now has eight cases covering the unmet-peer failure, the two legitimate ways to satisfy a peer, the optional-peer bypass, the no-install floor, optional peers and private workspaces, allowlist acceptance with stale reporting, and the real repository. Both new behaviours were written test-first against the fixtures that exposed them, so each fails without its fix.

Unchanged from the first pass, and worth repeating because it is inherent rather than a defect: no PR job builds a standalone app, so the `happy-dom` and `OM_ENABLE_PUSH_STUB_ADAPTER` fixes get their first true end-to-end confirmation from the next `Develop Snapshot Release` run's `Standalone App Integration Tests` job.

---
🤖 Re-reviewed by `om-auto-review-pr` (autofix: 1 iteration, 2 Minor and 2 Nit findings fixed, 1 Nit declined with reason).

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED (re-review of `652a02d1`). Lock released.

First pass on `ba66f16` returned CHANGES REQUESTED on two Minor findings in the new peer-dependency gate. One autofix iteration fixed both with regression tests, fixed two of three Nits, and declined the third with a reason. The re-review verdict is approve.

`ci-monitoring` is applied in place of `in-progress`: the review work is finished and fully reported, and only the CI run on the new head is still being watched. This PR is free for anyone to act on.

⚠️ Merge still needs an approving review from another maintainer — GitHub does not allow this account to submit one on its own PR.

pkarw:

CI settled **green** on `652a02d1` (the autofix head) within the follow-up budget. Every check passed: `prepare`, `lint`, `ds-lint`, `test`, `docker-build`, `documents-multi-instance`, `ephemeral-integration (none)`, `audit`, `audit-scope`, `scope`, all three CodeQL analyses, and `license/cla`. `comment`, `mutate`, `merge-coverage` and `[code]smith` skipped as configured.

Two things I checked in the logs rather than trusting the green tick:

- The `lint` job printed `Peer dependencies OK (29 allowlisted).` — the hardened gate still runs against a fully populated `node_modules` and did not trip its own new no-install refusal.
- The `Test scripts` step ran both new cases: `rejects a peer the package re-declares as OPTIONAL` and `refuses to pass when no dependency manifest can be resolved`.

`mergeable` is `MERGEABLE`; `mergeStateStatus` is `BLOCKED` solely because branch protection is waiting on an approving review. Removing `ci-monitoring` — no further follow-up is owed from this agent.

⚠️ Reminder: this account authored the PR, so GitHub would not accept an `APPROVE` review state from it. **Merge needs an approving review from another maintainer.**

PR #5469, opened 2026-08-22
fix(customers): fail closed on collapsed org scope for todos and tasks (#5466)

### LINKED ISSUE #5466: security(customers): todos/tasks list drops org filter when organizationIds is null or empty (todoCompatibility.ts:383,437)
`listLegacyTodoRows` and `listCanonicalTodoRows` build their `where` clause with the same fail-open org guard that #3841 reported on the activities adapter:

```ts
const where: Record<string, unknown> = { tenantId }
if (organizationIds && organizationIds.length > 0) {
  where.organizationId = { $in: organizationIds }
}
```

When `organizationIds` is `null` or `[]` the `organizationId` predicate is dropped entirely — not narrowed to nothing — so the query returns every organization's tasks within the tenant. For a superadmin / all-org principal that is the intended convention, but a **non-superadmin** principal whose resolved scope collapses gets a cross-organization read.

Found while reviewing #5462 (the activities half of this pattern). Issue #3841 explicitly flagged it: _"The same `organizationIds?.length > 0`-guarded org clause is a hotspot pattern worth grepping for elsewhere."_

## Location(s)

- `packages/core/src/modules/customers/lib/todoCompatibility.ts:383` — `listLegacyTodoRows`
- `packages/core/src/modules/customers/lib/todoCompatibility.ts:437` — `listCanonicalTodoRows`

Both are exported and reached from two request-scoped routes:

| Route | Calls | Deprecated? |
|-------|-------|-------------|
| `packages/core/src/modules/customers/api/todos/route.ts` | `:248` (unified), `:276`, `:279` | Yes — legacy adapter, sunset 2026-06-30 |
| `packages/core/src/modules/customers/api/interactions/tasks/route.ts` | `:53` (unified), `:82`, `:85` | **No — this is the canonical successor endpoint** |

Both routes take `organizationIds` straight from `resolveCustomersRequestContext` (`packages/core/src/modules/customers/lib/interactionRequestContext.ts`), which derives it as `scope.filterIds ?? (auth.orgId ? [auth.orgId] : null)` — the exact expression behind #3841.

The third consumer, `packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/route.ts`, is **not** affected: it sources scope from `resolveWidgetScope` instead, which already handles this correctly (see below).

## Why it matters

`resolveOrganizationScope` (`packages/core/src/modules/directory/utils/organizationScope.ts`) returns a collapsed scope in reachable configurations:

- **`filterIds: []`** — a restricted principal (`allowedIds` non-null) whose accessible-org set is empty and whose `auth.orgId` does not expand to a live organization. The trailing recovery block recomputes the same empty set and leaves `filterSet` empty.
- **`filterIds: null`** — a restricted principal with no `auth.orgId` at all, per the null-scope path #3841 describes.

Either way `organizationIds` reaches the helpers falsy-or-empty, the org predicate vanishes, and the caller reads every organization's `CustomerTodoLink` / task-type `CustomerInteraction` rows in the tenant. Tenant scoping still holds (`resolveCustomersRequestContext` 401s on a missing tenant), so this is cross-org-within-tenant, not cross-tenant.

The difference from #3841 is the second row of that table: `/api/customers/interactions/tasks` is the **canonical, non-deprecated** endpoint that the todos adapter is being sunset in favour of. So unlike the activities case, this one does not age out on 2026-06-30 — fixing only the deprecated adapter would leave the hole live on the successor surface.

## Proposed fix

There is already a correct implementation of this exact policy in the repo — `resolveWidgetScope` at `packages/core/src/modules/dashboards/lib/widgetScope.ts:32-42`:

```ts
const organizationIds = (() => {
  if (overrides?.organizationId) return [overrides.organizationId]
  if (scope?.selectedId) return [scope.selectedId]
  if (Array.isArray(scope?.filterIds) && scope.filterIds.length > 0) return scope.filterIds
  if (scope?.allowedIds === null) return null      // unrestricted, and only then
  if (auth.orgId) return [auth.orgId]
  return [] as string[]
})()

if (organizationIds !== null && organizationIds.length === 0) {
  throw new CrudHttpError(400, { error: translate('dashboards.errors.organization_required', …) })
}
```

`null` is produced only after every concrete source is exhausted *and* the principal is genuinely unrestricted, and an empty set is rejected rather than silently widened. Mirroring that in the todos path is preferable to bolting a guard onto each helper:

1. Resolve `isUnrestricted = auth.isSuperAdmin === true || scope?.allowedIds === null` at each route boundary, matching `hasUnrestrictedAccess` in `resolveOrganizationScope` and the approach taken in #5462.
2. For `!isUnrestricted && (!organizationIds || organizationIds.length === 0)`, return an empty paged result (or `403`, consistent with whichever shape #5462 settles on) before any query runs.
3. Keep a fail-closed guard inside the two helpers as defence in depth, but make its `isUnrestricted` semantics identical to the boundary gate — the divergence between the two layers is the Medium finding currently blocking #5462, and it should not be replicated here.

Worth considering as part of this: `resolveCustomersRequestContext` is the shared root cause for both #3841 and this issue, and it hands out a collapsed `organizationIds` with no unrestricted-ness signal attached. Returning a resolved `isUnrestricted` (or refusing to produce an ambiguous empty scope, as `resolveWidgetScope` does) would close the whole class at the source instead of route by route. That is a wider change than this issue needs, so it is a suggestion rather than a requirement — but every consumer of that helper is a candidate for the same bug until it happens.

## Testing

Follow `packages/core/src/modules/customers/api/activities/__tests__/org-scoping.test.ts` from #5462, applied to both routes. Per the `.ai/lessons.md` entry of 2026-07-11, parameterize the scope shape over **non-null, explicit null, and omitted** rather than testing null alone, and cover:

- Non-superadmin with `organizationIds: null` → empty, and the entity manager is never queried (`expect(em.find).not.toHaveBeenCalled()`).
- Non-superadmin with `organizationIds: []` → same.
- Superadmin (`auth.isSuperAdmin === true`) → unchanged tenant-wide read.
- Unrestricted non-superadmin (`scope.allowedIds === null`) → unchanged tenant-wide read.
- Explicit `organizationIds` → `{ $in: [...] }` applied.
- Each of the above in both `flags.unified` modes, since the unified and merged paths call different helpers.

## BC / Safety

Superadmin and unrestricted behaviour is preserved. Non-superadmin callers with a collapsed org scope, who today receive a tenant-wide read, would get an empty result or a rejection — the same trade-off #3841 accepted, and worth confirming no legitimate consumer depends on the accidental breadth. Contained to one module. `risk-medium`: single-module change, but it touches a live non-deprecated endpoint, not only the sunsetting adapter.

## Related

- #3841 — the activities half of the same pattern
- #5462 — the PR fixing #3841; its review found the Medium guard-consistency issue referenced above

Duang777:
## Summary

- Closes #5466. Follow-up to #5462 / #3841: the same `organizationIds && organizationIds.length > 0` fail-open guard was still live on customer todos/tasks.
- Restricted principals whose resolved scope collapses to `organizationIds: null` or `[]` no longer receive a tenant-wide read of `CustomerTodoLink` / task-type `CustomerInteraction` rows. Superadmin and unrestricted (`scope.allowedIds === null`) callers keep tenant-wide access, including the stale-scope `[]` case from #5462's M1.
- Gate sits on both request-scoped list routes before any query:
  - `GET /api/customers/interactions/tasks` (canonical, non-deprecated)
  - `GET /api/customers/todos` (legacy adapter; deprecation headers preserved on the empty page)
- The exported list helpers take the same `isUnrestricted` policy as defense in depth. The dashboard widget is unchanged in behavior (`resolveWidgetScope` already handles this) but now passes the flag so the helper guard cannot empty an unrestricted widget read.

## Test plan

- [x] Helper tests: restricted `null`/`[]` and omitted `isUnrestricted` do not query; unrestricted `null`/`[]` stay tenant-wide; explicit ids apply `{ $in: [...] }`
- [x] Route tests on both GET handlers: restricted `null`/`[]`/omitted `scope`/unified mode return an empty page with `em.find`/`findAndCount`/`count` never called; superadmin `null` and `[]`, unrestricted non-superadmin, and explicit ids pin the `where` clause
- [x] Existing todos route, tasks route, pagination, widget, and scope-forgery suites still pass
- [ ] Maintainer: intended labels `review`, `bug`, `security`, `skip-qa`, `priority-medium`, `risk-medium` (fork account cannot write labels on upstream)

`skip-qa` matches the automated-verification exemption: no `.tsx` / `packages/ui/` changes, database and API surface unchanged, no `BACKWARD_COMPATIBILITY.md` contract broken, and the changed behavior is covered by tests in this PR.

## Breaking Changes

| Surface | Status |
|---------|--------|
| API route paths and methods | Unchanged |
| Query schema and response envelope | Unchanged (empty page is the existing zero-result shape) |
| OpenAPI | Unchanged |
| Database schema | Untouched |
| Exported helpers | Additive optional `options.isUnrestricted`. Restricted callers that previously passed `null`/`[]` without the flag now get an empty list instead of a tenant-wide read — the intended security fix |

No other protected surface from `BACKWARD_COMPATIBILITY.md` is affected.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5469"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789978670&installation_model_id=432243&pr_number=5469&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5469&signature=6fa2804bfc613880d4904f46cb90bfc267e245774ad9cca4976d74bc5caa4146"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/__tests__/route.test.ts
+++ b/packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/__tests__/route.test.ts
@@ -1,5 +1,8 @@
 import { GET } from '../route'
 
+const TENANT_ID = '33333333-3333-3333-3333-333333333333'
+const ORG_ID = '22222222-2222-2222-2222-222222222222'
+
 jest.mock('../../utils', () => ({
   resolveWidgetScope: jest.fn(async () => ({
     container: {
@@ -11,6 +14,7 @@ jest.mock('../../utils', () => ({
     em: {},
     tenantId: '33333333-3333-3333-3333-333333333333',
     organizationIds: ['22222222-2222-2222-2222-222222222222'],
+    isUnrestrictedOrganizationScope: false,
   })),
 }))
 
@@ -162,5 +166,66 @@ describe('customers customer-todos widget route', () => {
     const options = listCanonicalTodoRows.mock.calls[0][5]
     expect(options).toMatchObject({ includeDeleted: true })
     expect(options).not.toHaveProperty('source')
+  })
+
+  it('passes unrestricted widget scope through to tenant-wide helper reads', async () => {
+    const { resolveWidgetScope } = jest.requireMock('../../utils')
+    const { resolveCustomerInteractionFeatureFlags } =
+      jest.requireMock('../../../../../lib/interactionFeatureFlags')
+    const { listLegacyTodoRows, listCanonicalTodoRows } =
+      jest.requireMock('../../../../../lib/todoCompatibility')
+
+    resolveWidgetScope.mockResolvedValueOnce({
+      container: {
+        resolve: jest.fn((name: string) => {
+          if (name === 'queryEngine') return { kind: 'query-engine' }
+          throw new Error(`Unexpected container resolve: ${name}`)
+        }),
+      },
+      em: {},
+      tenantId: TENANT_ID,
+      organizationIds: null,
+      isUnrestrictedOrganizationScope: true,
+    })
+    resolveCustomerInteractionFeatureFlags.mockResolvedValue({ unified: false })
+    listLegacyTodoRows.mockResolvedValue([])
+    listCanonicalTodoRows.mockResolvedValue({ items: [], bridgeIds: new Set<string>() })
+
+    const res = await GET(new Request('http://localhost/api?limit=5'))
+
+    expect(res.status).toBe(200)
+    expect(listLegacyTodoRows).toHaveBeenCalledTimes(1)
+    expect(listLegacyTodoRows.mock.calls[0][3]).toBeNull()
+    expect(listLegacyTodoRows.mock.calls[0][5]).toEqual(
+      expect.objectContaining({ isUnrestricted: true }),
+    )
+    expect(listCanonicalTodoRows).toHaveBeenCalledTimes(1)
+    expect(listCanonicalTodoRows.mock.calls[0][4]).toBeNull()
+    expect(listCanonicalTodoRows.mock.calls[0][5]).toEqual(
+      expect.objectContaining({ isUnrestricted: true }),
+    )
+  })
+
+  it('passes explicit organization scope through as restricted helper reads', async () => {
+    const { resolveCustomerInteractionFeatureFlags } =
+      jest.requireMock('../../../../../lib/interactionFeatureFlags')
+    const { listLegacyTodoRows, listCanonicalTodoRows } =
+      jest.requireMock('../../../../../lib/todoCompatibility')
+
+    resolveCustomerInteractionFeatureFlags.mockResolvedValue({ unified: false })
+    listLegacyTodoRows.mockResolvedValue([])
+    listCanonicalTodoRows.mockResolvedValue({ items: [], bridgeIds: new Set<string>() })
+
+    const res = await GET(new Request('http://localhost/api?limit=5'))
+
+    expect(res.status).toBe(200)
+    expect(listLegacyTodoRows.mock.calls[0][3]).toEqual([ORG_ID])
+    expect(listLegacyTodoRows.mock.calls[0][5]).toEqual(
+      expect.objectContaining({ isUnrestricted: false }),
+    )
+    expect(listCanonicalTodoRows.mock.calls[0][4]).toEqual([ORG_ID])
+    expect(listCanonicalTodoRows.mock.calls[0][5]).toEqual(
+      expect.objectContaining({ isUnrestricted: false }),
+    )
   })
 })
--- a/packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/route.ts
+++ b/packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/route.ts
@@ -28,7 +28,10 @@ export const metadata = {
 
 type WidgetContext = WidgetScopeContext & { limit: number }
 
-async function resolveContext(req: Request, translate: (key: string, fallback?: string) => string): Promise<WidgetContext> {
+async function resolveContext(
+  req: Request,
+  translate: (key: string, fallback?: string) => string,
+): Promise<WidgetContext> {
   const url = new URL(req.url)
   const rawQuery: Record<string, string> = {}
   for (const [key, value] of url.searchParams.entries()) {
@@ -39,16 +42,18 @@ async function resolveContext(req: Request, translate: (key: string, fallback?: 
     throw new CrudHttpError(400, { error: translate('customers.errors.invalid_query', 'Invalid query parameters') })
   }
 
-  const { container, em, tenantId, organizationIds } = await resolveWidgetScope(req, translate, {
-    tenantId: parsed.data.tenantId ?? null,
-    organizationId: parsed.data.organizationId ?? null,
-  })
+  const { container, em, tenantId, organizationIds, isUnrestrictedOrganizationScope } =
+    await resolveWidgetScope(req, translate, {
+      tenantId: parsed.data.tenantId ?? null,
+      organizationId: parsed.data.organizationId ?? null,
+    })
 
   return {
     container,
     em,
     tenantId,
     organizationIds,
+    isUnrestrictedOrganizationScope,
     limit: parsed.data.limit,
   }
 }
@@ -56,7 +61,8 @@ async function resolveContext(req: Request, translate: (key: string, fallback?: 
 export async function GET(req: Request) {
   const { translate } = await resolveTranslations()
   try {
-    const { container, em, tenantId, organizationIds, limit } = await resolveContext(req, translate)
+    const { container, em, tenantId, organizationIds, isUnrestrictedOrganizationScope, limit } =
+      await resolveContext(req, translate)
     const auth = {
       tenantId,
       orgId: organizationIds?.[0] ?? null,
@@ -64,7 +70,7 @@ export async function GET(req: Request) {
     }
     const flags = await resolveCustomerInteractionFeatureFlags(container, tenantId)
     const mergedWindow = Math.min(limit * 4, 50)
-    const isUnrestricted = organizationIds === null
+    const isUnrestricted = isUnrestrictedOrganizationScope === true
     const rows = flags.unified
       ? (await listCanonicalTodoRows(
           em,
--- a/packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/route.ts
+++ b/packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/route.ts
@@ -64,6 +64,7 @@ export async function GET(req: Request) {
     }
     const flags = await resolveCustomerInteractionFeatureFlags(container, tenantId)
     const mergedWindow = Math.min(limit * 4, 50)
+    const isUnrestricted = organizationIds === null
     const rows = flags.unified
       ? (await listCanonicalTodoRows(
           em,
@@ -71,7 +72,7 @@ export async function GET(req: Request) {
           auth,
           organizationIds?.[0] ?? null,
           organizationIds ?? null,
-          { pagination: { page: 1, pageSize: limit } },
+          { pagination: { page: 1, pageSize: limit }, isUnrestricted },
         )).items
       : await Promise.all([
           listLegacyTodoRows(
@@ -80,7 +81,7 @@ export async function GET(req: Request) {
             tenantId,
             organizationIds ?? null,
             undefined,
-            { limit: mergedWindow },
+            { limit: mergedWindow, isUnrestricted },
           ),
           listCanonicalTodoRows(
             em,
@@ -91,6 +92,7 @@ export async function GET(req: Request) {
             {
               includeDeleted: true,
               limit: mergedWindow,
+              isUnrestricted,
             },
           ),
         ]).then(([legacyRows, canonicalRows]) =>
--- a/packages/core/src/modules/customers/api/interactions/tasks/__tests__/org-scoping.test.ts
+++ b/packages/core/src/modules/customers/api/interactions/tasks/__tests__/org-scoping.test.ts
@@ -167,6 +167,18 @@ describe('interactions/tasks GET organization scoping', () => {
     expectNoQueries()
   })
 
+  it('returns export-shaped empty page for all=true when restricted has organizationIds = null', async () => {
+    mockContext = buildContext({ organizationIds: null })
+
+    const res = await GET(
+      new Request('http://localhost/api/customers/interactions/tasks?page=3&pageSize=50&all=true'),
+    )
+    expect(res.status).toBe(200)
+    const body = await res.json()
+    expect(body).toEqual({ items: [], total: 0, page: 1, pageSize: 0, totalPages: 1 })
+    expectNoQueries()
+  })
+
   it('allows tenant-wide query when superadmin has organizationIds = null', async () => {
     mockContext = buildContext({ organizationIds: null, isSuperAdmin: true, allowedIds: null })
 
@@ -190,6 +202,24 @@ describe('interactions/tasks GET organization scoping', () => {
     expect(res.status).toBe(200)
     expectTenantWideQueries()
   })
+
+  it(
+    'allows tenant-wide query in unified mode when unrestricted non-superadmin has organizationIds = null',
+    async () => {
+      setFlags({ unified: true })
+      mockContext = buildContext({
+        organizationIds: null,
+        isSuperAdmin: false,
+        allowedIds: null,
+      })
+
+      const res = await GET(
+        new Request('http://localhost/api/customers/interactions/tasks?page=1&pageSize=50'),
+      )
+      expect(res.status).toBe(200)
+      expectTenantWideQueries()
+    },
+  )
 
   it('allows tenant-wide query when superadmin has organizationIds = []', async () => {
     mockContext = buildContext({ organizationIds: [], isSuperAdmin: true, allowedIds: null })
--- a/packages/core/src/modules/customers/api/interactions/tasks/__tests__/org-scoping.test.ts
+++ b/packages/core/src/modules/customers/api/interactions/tasks/__tests__/org-scoping.test.ts
@@ -1,0 +1,218 @@
+/** @jest-environment node */
+
+import { GET } from '../route'
+
+const TENANT_ID = '123e4567-e89b-41d3-a456-426614174010'
+const ORG_1 = '123e4567-e89b-41d3-a456-426614174001'
+
+const mockEm = {
+  find: jest.fn(),
+  findOne: jest.fn(),
+  findAndCount: jest.fn(),
+  count: jest.fn(),
+}
+
+const mockContainer = {
+  resolve: jest.fn((name: string) => {
+    if (name === 'commandBus') return { execute: jest.fn() }
+    if (name === 'em') return mockEm
+    if (name === 'queryEngine') return { query: jest.fn() }
+    throw new Error(`Unknown dependency: ${name}`)
+  }),
+}
+
+let mockContext: Record<string, unknown> = {}
+
+jest.mock('../../../../lib/interactionFeatureFlags', () => ({
+  resolveCustomerInteractionFeatureFlags: jest.fn(),
+}))
+
+jest.mock('../../../../lib/interactionRequestContext', () => ({
+  resolveCustomersRequestContext: jest.fn(async () => mockContext),
+}))
+
+jest.mock('../../../../lib/interactionReadModel', () => ({
+  hydrateCanonicalInteractions: jest.fn(async () => []),
+  loadCustomerSummaries: jest.fn(async () => new Map()),
+}))
+
+jest.mock('@open-mercato/shared/lib/i18n/server', () => ({
+  resolveTranslations: jest.fn(async () => ({
+    translate: (_key: string, fallback: string) => fallback,
+  })),
+}))
+
+function setFlags(overrides?: { unified?: boolean }) {
+  const { resolveCustomerInteractionFeatureFlags } = jest.requireMock(
+    '../../../../lib/interactionFeatureFlags',
+  )
+  resolveCustomerInteractionFeatureFlags.mockResolvedValue({
+    unified: overrides?.unified ?? false,
+    legacyAdapters: true,
+    externalSync: false,
+  })
+}
+
+function buildContext(input: {
+  organizationIds: string[] | null
+  isSuperAdmin?: boolean
+  allowedIds?: string[] | null
+  omitScope?: boolean
+}) {
+  const isSuperAdmin = input.isSuperAdmin === true
+  const allowedIds = input.omitScope
+    ? undefined
+    : (Object.prototype.hasOwnProperty.call(input, 'allowedIds')
+        ? input.allowedIds
+        : (isSuperAdmin ? null : [ORG_1]))
+  const scope = input.omitScope
+    ? undefined
+    : {
+        selectedId: input.organizationIds?.[0] ?? null,
+        filterIds: input.organizationIds,
+        allowedIds,
+        tenantId: TENANT_ID,
+      }
+  return {
+    auth: {
+      sub: isSuperAdmin ? 'superadmin-1' : 'user-1',
+      tenantId: TENANT_ID,
+      orgId: input.organizationIds?.[0] ?? null,
+      isSuperAdmin,
+    },
+    em: mockEm,
+    organizationIds: input.organizationIds,
+    selectedOrganizationId: input.organizationIds?.[0] ?? null,
+    scope,
+    container: mockContainer,
+    commandContext: {
+      container: mockContainer,
+      auth: {
+        sub: isSuperAdmin ? 'superadmin-1' : 'user-1',
+        tenantId: TENANT_ID,
+        orgId: input.organizationIds?.[0] ?? null,
+        isSuperAdmin,
+      },
+      organizationScope: scope,
+      selectedOrganizationId: input.organizationIds?.[0] ?? null,
+      organizationIds: input.organizationIds,
+      request: undefined,
+    },
+  }
+}
+
+function expectNoQueries() {
+  expect(mockEm.find).not.toHaveBeenCalled()
+  expect(mockEm.findAndCount).not.toHaveBeenCalled()
+  expect(mockEm.count).not.toHaveBeenCalled()
+}
+
+function expectTenantWideQueries() {
+  const calls = [...mockEm.find.mock.calls, ...mockEm.findAndCount.mock.calls]
+  expect(calls.length).toBeGreaterThan(0)
+  for (const call of calls) {
+    const where = call[1] as Record<string, unknown>
+    expect(where.tenantId).toBe(TENANT_ID)
+    expect(where.organizationId).toBeUndefined()
+  }
+}
+
+describe('interactions/tasks GET organization scoping', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    setFlags()
+    mockEm.find.mockImplementation(async () => [])
+    mockEm.findAndCount.mockImplementation(async () => [[], 0])
+    mockEm.count.mockImplementation(async () => 0)
+  })
+
+  it.each([
+    { label: 'organizationIds = null', organizationIds: null as string[] | null },
+    { label: 'organizationIds = []', organizationIds: [] as string[] },
+  ])('returns empty and does not query when restricted has $label', async ({ organizationIds }) => {
+    mockContext = buildContext({ organizationIds })
+
+    const res = await GET(
+      new Request('http://localhost/api/customers/interactions/tasks?page=1&pageSize=50'),
+    )
+    expect(res.status).toBe(200)
+    const body = await res.json()
+    expect(body).toEqual({ items: [], total: 0, page: 1, pageSize: 50, totalPages: 1 })
+    expectNoQueries()
+  })
+
+  it('returns empty in unified mode when restricted has organizationIds = null', async () => {
+    setFlags({ unified: true })
+    mockContext = buildContext({ organizationIds: null })
+
+    const res = await GET(
+      new Request('http://localhost/api/customers/interactions/tasks?page=1&pageSize=50'),
+    )
+    expect(res.status).toBe(200)
+    const body = await res.json()
+    expect(body.items).toEqual([])
+    expect(body.total).toBe(0)
+    expectNoQueries()
+  })
+
+  it('returns empty when scope is omitted and organizationIds is null', async () => {
+    mockContext = buildContext({ organizationIds: null, omitScope: true })
+
+    const res = await GET(
+      new Request('http://localhost/api/customers/interactions/tasks?page=1&pageSize=50'),
+    )
+    expect(res.status).toBe(200)
+    const body = await res.json()
+    expect(body.items).toEqual([])
+    expectNoQueries()
+  })
+
+  it('allows tenant-wide query when superadmin has organizationIds = null', async () => {
+    mockContext = buildContext({ organizationIds: null, isSuperAdmin: true, allowedIds: null })
+
+    const res = await GET(
+      new Request('http://localhost/api/customers/interactions/tasks?page=1&pageSize=50'),
+    )
+    expect(res.status).toBe(200)
+    expectTenantWideQueries()
+  })
+
+  it('allows tenant-wide query when unrestricted non-superadmin has organizationIds = null', async () => {
+    mockContext = buildContext({
+      organizationIds: null,
+      isSuperAdmin: false,
+      allowedIds: null,
+    })
+
+    const res = await GET(
+      new Request('http://localhost/api/customers/interactions/tasks?page=1&pageSize=50'),
+    )
+    expect(res.status).toBe(200)
+    expectTenantWideQueries()
+  })
+
+  it('allows tenant-wide query when superadmin has organizationIds = []', async () => {
+    mockContext = buildContext({ organizationIds: [], isSuperAdmin: true, allowedIds: null })
+
+    const res = await GET(
+      new Request('http://localhost/api/customers/interactions/tasks?page=1&pageSize=50'),
+    )
+    expect(res.status).toBe(200)
+    expectTenantWideQueries()
+  })
+
+  it('applies organizationId filter when organizationIds is provided', async () => {
+    mockContext = buildContext({ organizationIds: [ORG_1] })
+
+    const res = await GET(
+      new Request('http://localhost/api/customers/interactions/tasks?page=1&pageSize=50'),
+    )
+    expect(res.status).toBe(200)
+
+    const calls = [...mockEm.find.mock.calls, ...mockEm.findAndCount.mock.calls]
+    expect(calls.length).toBeGreaterThan(0)
+    for (const call of calls) {
+      expect((call[1] as Record<string, unknown>).organizationId).toEqual({ $in: [ORG_1] })
+    }
+  })
+})
--- a/packages/core/src/modules/customers/api/interactions/tasks/route.ts
+++ b/packages/core/src/modules/customers/api/interactions/tasks/route.ts
@@ -40,10 +40,26 @@ const MERGED_TASK_FETCH_CAP = 2000
 export async function GET(request: Request): Promise<Response> {
   const { translate } = await resolveTranslations()
   try {
-    const { auth, em, organizationIds, container, selectedOrganizationId } =
+    const { auth, em, organizationIds, container, selectedOrganizationId, scope } =
       await resolveCustomersRequestContext(request)
     const query = querySchema.parse(Object.fromEntries(new URL(request.url).searchParams))
     const flags = await resolveCustomerInteractionFeatureFlags(container, auth.tenantId)
+    const isUnrestricted = auth.isSuperAdmin === true || scope?.allowedIds === null
+    if (!isUnrestricted && (!organizationIds || organizationIds.length === 0)) {
+      logger.warn('customers.interactions.tasks.list collapsed organization scope', {
+        tenantId: auth.tenantId,
+        organizationIds,
+        selectedId: scope?.selectedId ?? null,
+        allowedIdsCount: scope?.allowedIds?.length ?? null,
+      })
+      return NextResponse.json({
+        items: [],
+        total: 0,
+        page: query.page,
+        pageSize: query.pageSize,
+        totalPages: 1,
+      })
+    }
     const exportAll = parseBooleanToken(query.all) === true
     const search = normalizeTodoSearch(query.search)
     const queryEngine = container.resolve('queryEngine') as QueryEngine
@@ -59,6 +75,7 @@ export async function GET(request: Request): Promise<Response> {
           entityId: query.entityId,
           pagination: exportAll ? null : { page: query.page, pageSize: query.pageSize },
           searchText: search,
+          isUnrestricted,
         },
       )
       const total = canonical.total
@@ -80,6 +97,7 @@ export async function GET(request: Request): Promise<Response> {
     const [legacyRows, canonicalRows] = await Promise.all([
       listLegacyTodoRows(em, queryEngine, auth.tenantId, organizationIds, query.entityId, {
         limit: legacyWindow,
+        isUnrestricted,
       }),
       listCanonicalTodoRows(
         em,
@@ -91,6 +109,7 @@ export async function GET(request: Request): Promise<Response> {
           entityId: query.entityId,
           includeDeleted: true,
           limit: legacyWindow,
+          isUnrestricted,
         },
       ),
     ])
--- a/packages/core/src/modules/customers/api/interactions/tasks/route.ts
+++ b/packages/core/src/modules/customers/api/interactions/tasks/route.ts
@@ -3,6 +3,7 @@ import { z } from 'zod'
 import type { OpenApiRouteDoc } from '@open-mercato/shared/lib/openapi'
 import { CrudHttpError, isCrudHttpError } from '@open-mercato/shared/lib/crud/errors'
 import { parseBooleanToken } from '@open-mercato/shared/lib/boolean'
+import { isUnrestrictedOrganizationScope } from '@open-mercato/shared/lib/auth/organizationAccess'
 import { resolveTranslations } from '@open-mercato/shared/lib/i18n/server'
 import type { QueryEngine } from '@open-mercato/shared/lib/query/types'
 import { createCustomersCrudOpenApi, createPagedListResponseSchema } from '../../openapi'
@@ -44,7 +45,11 @@ export async function GET(request: Request): Promise<Response> {
       await resolveCustomersRequestContext(request)
     const query = querySchema.parse(Object.fromEntries(new URL(request.url).searchParams))
     const flags = await resolveCustomerInteractionFeatureFlags(container, auth.tenantId)
-    const isUnrestricted = auth.isSuperAdmin === true || scope?.allowedIds === null
+    const exportAll = parseBooleanToken(query.all) === true
+    const isUnrestricted = isUnrestrictedOrganizationScope({
+      isSuperAdmin: auth.isSuperAdmin === true,
+      allowedOrganizationIds: scope?.allowedIds,
+    })
     if (!isUnrestricted && (!organizationIds || organizationIds.length === 0)) {
       logger.warn('customers.interactions.tasks.list collapsed organization scope', {
         tenantId: auth.tenantId,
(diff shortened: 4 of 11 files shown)


Duang777:
I have read the CLA Document and I hereby sign the CLA

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T09:14:49Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(customers): fail closed on collapsed org scope for todos and tasks (#5466)

## 🎯 Summary

This PR closes the `organizationIds && organizationIds.length > 0` fail-open guard that #5466 reported on the customer todos/tasks read path — the third instance of the pattern after #3841 and #5462. It adds a boundary gate to both request-scoped list routes (`GET /api/customers/interactions/tasks`, the canonical endpoint, and `GET /api/customers/todos`, the deprecated adapter) that returns an empty page before any query runs when a **restricted** principal's resolved organization scope has collapsed to `null` or `[]`, and it keeps a matching fail-closed guard inside the two exported helpers (`listLegacyTodoRows`, `listCanonicalTodoRows`) as defense in depth. The dashboard widget is updated to pass the new `isUnrestricted` flag so its unrestricted read is not caught by the helper guard.

The security reasoning holds up under inspection. I traced `resolveOrganizationScopeForRequest` (`packages/core/src/modules/directory/utils/organizationScope.ts:302-380`): for a **restricted** principal `filterSet` is only ever `allowedSet` or a selection expansion, and the trailing fallback block re-tries the account org, so `filterIds: []` genuinely means "no accessible organization" and `filterIds: null` is unreachable for a restricted principal. Returning an empty page therefore does not take data away from any legitimately-scoped user, while the unrestricted (`scope.allowedIds === null`) and superadmin paths keep their tenant-wide read. The gate sits before every `em.find` / `em.findAndCount` / `queryEngine.query` call on both routes, and the guard is correctly placed ahead of the `where`-clause construction in both helpers (`todoCompatibility.ts:393`, `:445`).

The tests are real regression tests, not restatements of the implementation: reverting only the three source files onto the PR's own test suite fails 13 tests across the three new files, and the whole set passes again with the fix in place.

Scope reviewed: the 7 files in the diff, plus `interactionRequestContext.ts`, `organizationScope.ts`, `widgetScope.ts`, `organizationAccess.ts`, and the existing customers todo/task/widget suites for context.

I found no blocker and no major issue. The two minor findings below are about the *shape* of the fix — a duplicated security policy and an untested branch — not about its correctness.

## Verdict

❌ **Request changes** — the fix itself is correct and I would merge the behavior as-is, but two minor findings stand: the "is this principal unrestricted" policy is re-implemented inline three times using two different formulas while the repository already documents a single source of truth for that decision (`packages/shared/src/lib/auth/organizationAccess.ts`), and the one place where the new flag can be set to `true` from that *second* formula — the dashboard widget — has no test covering it. This repository requests changes on minor findings (`.ai/skills/om-auto-review-pr/SKILL.md`, verdict rule), so both are enough to hold the approval; each is a small, contained follow-up.

## 🧪 Validation Gate

**Validation source: local fallback** — GitHub reported no usable check data for head `fd9b275`. The only checks on this SHA are `license/cla` (SUCCESS) and `[code]smith` (SKIPPED); both workflow runs for the head (`CI for Develop&Main`, `Mutation tests`) are in `action_required`, i.e. a maintainer still has to approve workflow runs for this first-time fork contributor. Because CI produced no test/typecheck/build signal at all, I ran the changed package's own gate locally, scoped per this repository's GitHub-checks-first extension.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn test` (scoped: full `packages/core` suite) | ✅ PASS for this change | 11,032 tests passed; 2 failed and 26 suites failed to load, **all** of them for one environment reason unrelated to this PR: this review worktree has no `packages/core/generated/` directory (`yarn generate` was never run here), so every suite importing `#generated/entities/*` — catalog, sales, warranty_claims, translations, wms, attachments — cannot resolve its module. No failing suite touches `customers`, and the diff adds no generated artifact. |
| `yarn test` (scoped: `customers` todos / interactions-tasks / lib / dashboard suites) | ✅ PASS | 29 suites, 265 tests passed, including the 3 new files and every pre-existing todos/tasks/widget/scope-forgery suite. |
| Regression-proof run (source files reverted to `df8caf2`, new tests kept) | ✅ FAILS AS EXPECTED | 13 of the 32 new assertions fail without the fix, confirming the tests bind the behavior rather than the implementation. |
| `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn build:app` | ⏭️ NOT RUN | Deliberately skipped rather than run misleadingly: this review worktree resolves `@open-mercato/*` through a shared `node_modules`, so a workspace-wide typecheck/build here would report on another branch's sources. The diff adds no dependency, no generated file, no schema change and no user-facing string, so these gates carry no signal that the scoped test run does not. **They must still run in CI before merge** — see the CI note above. |

## Findings

### 🔹 Minor

**1. `packages/core/src/modules/customers/api/interactions/tasks/route.ts:47`, `packages/core/src/modules/customers/api/todos/route.ts:243`, `packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/route.ts:67` — the "unrestricted principal" policy is re-implemented three times, in two different formulas.**

The two routes compute `auth.isSuperAdmin === true || scope?.allowedIds === null`, while the widget computes `organizationIds === null`. Both are correct *today*: `resolveWidgetScope` (`packages/core/src/modules/dashboards/lib/widgetScope.ts:74-84`) only returns `null` after `scope?.allowedIds === null`, and rejects an empty set with a 400, so the widget's shorter form happens to be equivalent. But the equivalence is non-local — it lives in another module's helper, which was itself recently hardened — and nothing in either file records the dependency. If `resolveWidgetScope` ever grows a fourth `null`-producing branch, the widget silently starts passing `isUnrestricted: true` for a restricted principal and re-opens exactly the hole this PR closes.

The repository already owns this decision in one place: `isOrganizationAccessAllowed` in `packages/shared/src/lib/auth/organizationAccess.ts:17`, documented as "the single source of truth for 'may this principal act on `targetOrganizationId`'", whose first two branches (`isSuperAdmin` → allow, `allowedOrganizationIds === null` → allow) are precisely this PR's `isUnrestricted`. This matters more than usual here because #5466 explicitly asks not to repeat it: *"the divergence between the two layers is the Medium finding currently blocking #5462, and it should not be replicated here."*

Suggested fix: add one exported predicate next to `isOrganizationAccessAllowed` — e.g. `isUnrestrictedOrganizationScope({ isSuperAdmin, allowedOrganizationIds })`, which `isOrganizationAccessAllowed` can then call for its own first two branches — and use it from all three call sites, with the widget deriving it from the scope it already has rather than from the shape of `organizationIds`. Resolving `isUnrestricted` once inside `resolveCustomersRequestContext` and returning it on the context (the wider change #5466 suggests) would also satisfy this, and would close the class for every future consumer of that helper.

**2. `packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/__tests__/route.test.ts` — the widget's new `isUnrestricted` branch is untested.**

All three existing widget tests build their context with `organizationIds: ['22222222-2222-2222-2222-222222222222']` (line 13), so every test in the file exercises `isUnrestricted === false`. The `organizationIds === null` path added at `route.ts:67` — the only place in this PR where the flag becomes `true` through a formula other than the routes' — is never executed by a test. That is the branch whose regression is silent in both directions: if it stays `false` the widget goes blank for unrestricted principals, and if it wrongly becomes `true` it restores the tenant-wide read for a restricted one. The `.ai/lessons.md` entry of 2026-07-11 that #5466 cites asks for exactly this parameterization (non-null, explicit null, omitted).

Suggested fix: add two cases to that file — one with `organizationIds: null` asserting the canonical/legacy reads still happen tenant-wide (`where.organizationId` undefined), and one with an explicit id asserting `{ $in: [...] }` — mirroring `expectTenantWideQueries` from the two new route suites.

### 💅 Nit

**3. `packages/core/src/modules/customers/api/todos/route.ts:255-259` and `packages/core/src/modules/customers/api/interactions/tasks/route.ts:57-61` — the guarded empty envelope ignores `all=true`.**

`paginateTodoRows` (`todoCompatibility.ts:182-189`) and the unified branch both answer an `exportAll` request with `page: 1, pageSize: <item count>`, whereas the new early return always answers `page: query.page, pageSize: query.pageSize`. The payload is an empty page either way, so nothing observable breaks; matching the export shape (`exportAll ? 1 : query.page` / `exportAll ? 0 : query.pageSize`) would just keep the envelope uniform across every path of the same handler.

**4. Route-level coverage of the *unrestricted* paths runs only in merged mode.** In both new route suites, `setFlags({ unified: true })` appears only in the restricted "returns empty" case; every `expectTenantWideQueries` assertion runs with `unified: false`. #5466's test matrix asks for each case "in both `flags.unified` modes, since the unified and merged paths call different helpers". The unified helper is separately covered by `todoCompatibility.org-scoping.test.ts`, so this is a completeness point rather than a gap in the behavior's protection — one extra `it` per suite closes it.

**5. `.../tasks/route.ts:49` and `.../todos/route.ts:245` — the collapsed-scope `logger.warn` fires on every request.** A principal whose scope is *permanently* collapsed (no accessible organization at all) polling a dashboard list will emit a warn-level line per poll. The information is worth keeping — it is how an operator notices the fail-closed path is being hit — but consider dropping it to `debug`, or sampling it, if these routes turn out to be polled. No change required for merge.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `listLegacyTodoRows` and `listCanonicalTodoRows` keep their names and export paths.
- [x] No function signature changed in a breaking way — `isUnrestricted` is an optional field on an already-optional `options` parameter on both helpers; every existing call compiles unchanged.
- [x] No HTTP route URL removed or renamed; no method changed — both routes keep their paths, methods, guards and OpenAPI documents, and the todos adapter still emits its `Deprecation` headers on the new empty-page response (asserted at `api/todos/__tests__/org-scoping.test.ts:461`).
- [x] No field removed or retyped in an existing response shape — the empty page is the existing zero-result envelope (`items`, `total`, `page`, `pageSize`, `totalPages`), and `totalPages: 1` matches what `paginateTodoRows` and the unified branch already return for zero rows.
- [x] No database table or column changed — the diff contains no entity, migration or snapshot change, correctly, since this is a query-guard fix.
- [x] Where a contract had to change, the old surface still works: the one behavior change is deliberate and is the security fix itself — a restricted caller that previously reached these helpers with `organizationIds` `null`/`[]` and no `isUnrestricted` flag now receives an empty list instead of a tenant-wide read. `BACKWARD_COMPATIBILITY.md` classifies neither helper as a protected surface, the PR body documents the change, and per the scope analysis above no legitimately-scoped caller loses data. **No protected-surface violation to warn about.**

## 🧪 Test Coverage

Three new files, 32 assertions, all passing, and — verified by reverting the source files and re-running — 13 of them fail without the fix, so they are genuine regression tests rather than descriptions of the new code.

`api/interactions/tasks/__tests__/org-scoping.test.ts` and `api/todos/__tests__/org-scoping.test.ts` cover, per route: restricted with `organizationIds: null` and `[]` (empty page, `em.find` / `findAndCount` / `count` never called), restricted `null` in unified mode, an omitted `scope` object, superadmin with `null` and with `[]` (tenant-wide, `where.organizationId` undefined), an unrestricted non-superadmin (`scope.allowedIds === null`), and an explicit id list pinning `{ $in: [...] }`. The todos suite additionally asserts the `Deprecation: true` header survives on the guarded response, which is the detail most likely to be dropped by an early return. `lib/__tests__/todoCompatibility.org-scoping.test.ts` covers the same matrix one layer down, including the important "flag omitted entirely" case that proves the helpers default to fail-closed.

Gaps, in the order I would close them:

1. The dashboard widget's `organizationIds === null` path — finding 2 above; add the two cases to `api/dashboard/widgets/customer-todos/__tests__/route.test.ts`.
2. The unrestricted/superadmin route paths under `flags.unified: true` — finding 4 above; one added `it` per route suite.

Nothing here needs integration coverage beyond the unit level: the change adds no schema, no new endpoint and no UI, and the scoping decision it guards is fully determined by the resolved scope object the unit tests inject.

pkarw:

Updated after the re-review of head `fc47738a` (this comment is maintained in place, so it always reflects the current label set).

- ✅ `merge-queue` — replaces `changes-requested`. The follow-up commit closed all four open findings from the previous review: the "unrestricted principal" policy is now centralized in `packages/shared/src/lib/auth/organizationAccess.ts` and reused by all three call sites plus `isOrganizationAccessAllowed`, and the dashboard widget's previously unexecuted branch is covered by tests. The re-review found no blocker, major or minor issue. **This is a code-review approval only** — 17 CI checks were still running when it was submitted, and the merge stays gated on them going green.
- 🐛 `bug` — unchanged. Closes the defect reported in #5466: a restricted principal whose organization scope collapsed received a tenant-wide read of another organization's tasks.
- 🔒 `security` — unchanged. The defect is a cross-organization data-scoping leak on a read path, and the change is the hardening that closes it.
- ⏭️ `skip-qa` — unchanged. The automated-verification exemption still applies on the new head: no `.tsx` file, nothing under `packages/ui/`, database structure and API surface unchanged, no `BACKWARD_COMPATIBILITY.md` protected surface broken, and the changed behavior ships with automated tests in the same PR.
- 🔺 `priority-high` — unchanged. Security hardening on data scoping, on `/api/customers/interactions/tasks`, the canonical non-deprecated endpoint.
- ⚠️ `risk-high` — unchanged. The follow-up commit slightly widens the blast radius rather than narrowing it: `packages/shared/src/lib/auth/organizationAccess.ts` is now touched, and `isOrganizationAccessAllowed` has three production callers outside this module (`api_keys/api/keys/route.ts`, `directory/utils/organizationScopeGuard.ts`, `shared/lib/commands/scope.ts`). I verified the refactor is byte-identical for every input its declared type permits, and the `mutate (shared, src/lib/auth/organizationAccess.ts)` CI job passes on the new tests — but the rating stands on blast radius, not on confidence.

No `needs-qa`, so the QA-approval gate does not apply; `skip-qa` is the explicit opt-out and is not combined with `needs-qa` or `qa-approved`.

pkarw:
Thanks @Duang777 — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

Two notes that are not code findings but affect the merge:

- CI has not run on this head. Both workflow runs for `fd9b275` (`CI for Develop&Main`, `Mutation tests`) are in `action_required`, which is GitHub asking a maintainer to approve workflow runs for a first-time fork contributor. The validation evidence in my review is a local, scoped run — it is not a substitute for CI, and a maintainer still needs to release the runs before this can merge.
- I applied the labels you listed in the PR description, with two deliberate differences: `priority-high` instead of `priority-medium` and `risk-high` instead of `risk-medium`, because this repository's taxonomy rates data-scoping security work at those levels. The reasoning is in the label-rationale comment.

pkarw:

The fail-closed gate is correct — I traced the scope resolver to confirm a collapsed scope really does mean "no accessible organization" for a restricted principal, so no legitimately-scoped caller loses data, and reverting the three source files makes 13 of the PR's new assertions fail, which proves they are real regression tests. What holds the approval is two minor findings about the shape of the fix: the "unrestricted principal" policy is re-implemented three times in two different formulas instead of using the repository's documented single source of truth, and the dashboard widget's new branch is untested. Details are in the review.

Validation source: local fallback (`packages/core` jest, scoped and full) — GitHub reported no usable check data for this head, since both workflow runs are in `action_required` pending maintainer approval. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here). No CI follow-up is scheduled, because no checks can start until the workflow runs are approved.

Duang777:
Thanks for the review. I pushed a follow-up commit: `fc47738 fix: centralize customer todo org scope policy`.

What changed:
- Added `isUnrestrictedOrganizationScope` in `@open-mercato/shared/lib/auth/organizationAccess` and made `isOrganizationAccessAllowed` reuse it.
- Updated the two customers todo/task routes to derive unrestricted scope from that shared predicate instead of inline formulas.
- Extended `resolveWidgetScope` to expose the same unrestricted decision, and updated the customer-todos widget to consume that field instead of inferring from `organizationIds === null`.
- Added coverage for the widget `organizationIds: null` tenant-wide branch and explicit-org restricted branch.
- Also covered unrestricted unified-mode route paths and normalized the collapsed-scope `all=true` empty envelope.

Local validation, using the local runner because no compose `app` container was active:
- `yarn workspace @open-mercato/shared test packages/shared/src/lib/auth/__tests__/organizationAccess.test.ts --runInBand`
- `yarn workspace @open-mercato/core test packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/__tests__/route.test.ts packages/core/src/modules/customers/api/interactions/tasks/__tests__/org-scoping.test.ts packages/core/src/modules/customers/api/todos/__tests__/org-scoping.test.ts packages/core/src/modules/customers/lib/__tests__/todoCompatibility.org-scoping.test.ts --runInBand`
- `yarn workspace @open-mercato/shared typecheck`
- `yarn build:packages`
- `yarn generate`
- `yarn workspace @open-mercato/core typecheck`

All passed.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-22T10:34:56Z. Other auto-skills will skip this PR until the lock is released.

Re-review of head `fc47738a` (previous review covered `fd9b275e`). The author-assignee on this PR came from my own previous run's changes-requested handoff, not from a competing automation claim, so it is not treated as a live lock.

pkarw (approved):
# 🔍 Code Review: Re-review — fix(customers): fail closed on collapsed org scope for todos and tasks (#5466)

## 🎯 Summary

This is a **re-review** of head `fc47738a` against my earlier review of `fd9b275e`. The security fix itself was already correct in the first pass — I traced `resolveOrganizationScopeForRequest` then and confirmed that a collapsed `filterIds` genuinely means "no accessible organization" for a restricted principal, so the fail-closed empty page takes data away from nobody who legitimately has it. What held approval back was the *shape* of the fix: the "is this principal unrestricted" policy was re-implemented inline three times using two different formulas, and the one branch that could set the new flag from that second formula — the dashboard widget — had no test.

The follow-up commit `fc47738a` ("centralize customer todo org scope policy") closes all four of my open findings and nothing else. `isUnrestrictedOrganizationScope({ isSuperAdmin, allowedOrganizationIds })` now lives in `packages/shared/src/lib/auth/organizationAccess.ts:12` beside `isOrganizationAccessAllowed`, which was refactored to call it (`:27`) so the two predicates can no longer drift. The three call sites — `api/interactions/tasks/route.ts:49`, `api/todos/route.ts:242`, and `dashboards/lib/widgetScope.ts:96` — all route through it, and the widget stops deriving the flag from `organizationIds === null` entirely, reading `resolveWidgetScope`'s new `isUnrestrictedOrganizationScope` field instead.

I checked that the widget's substitution is behavior-preserving rather than merely plausible. `resolveWidgetScope` returns `organizationIds: null` **only** on the `scope?.allowedIds === null` branch (`widgetScope.ts:80`) — the earlier branches return non-empty arrays, and a non-null empty array throws 400 at `:85`. So on that path `organizationIds === null` and `allowedIds === null` are the same condition, and where they differ (a superadmin with an explicit selection) `organizationIds` is non-empty, which makes the flag inert in `isCollapsedRestrictedOrganizationScope` and in the `where.organizationId` construction alike. The widget's read behavior is genuinely unchanged, as the PR body claims.

The commit also picked up my two nits. The guarded empty envelope now matches the export shape (`page: exportAll ? 1 : query.page`, `pageSize: exportAll ? 0 : query.pageSize`), which lines up exactly with what `paginateTodoRows` returns for zero rows under `exportAll` (`todoCompatibility.ts:183-190`) and with the unified branch's `pageSize: canonical.items.length`. And each route suite gained a `unified: true` unrestricted case, closing the gap where every `expectTenantWideQueries` assertion had run in merged mode only.

Scope re-reviewed: the incremental diff of `fc47738a` (9 files), plus `widgetScope.ts` end to end, `organizationAccess.ts` and its three production callers (`api_keys/api/keys/route.ts`, `directory/utils/organizationScopeGuard.ts`, `shared/lib/commands/scope.ts`), `todoCompatibility.ts`'s guard and pagination helpers, and the sibling dashboard widget routes to confirm this change introduces no inconsistency there.

I found no blocker, no major, and no minor issue on this head.

## Verdict

✅ **Approve** — every finding from the previous review is resolved, the centralization is the right one (it reuses the repository's documented single source of truth rather than adding a fourth formula), and the two behavior-affecting edits in this commit — the widget's flag source and the `exportAll` envelope — are both verifiably equivalent to or more consistent than what they replace. The only open item is nit 5 from the previous review, which I already marked as not requiring a change for merge.

## 🧪 Validation Gate

Validation source: **GitHub checks** on head `fc47738a`. Per this repository's `om-auto-review-pr` extension, CI is the primary evidence and local runs are a fallback only when GitHub reports no usable check data. A maintainer has since released the workflow runs that were stuck in `action_required` during the previous review, so the gate is running for real on this head and no local substitute was needed.

| Command (`validation.commands`) | Covering check | Status | Notes |
|---------|---------|--------|-------|
| `yarn build:packages` | `prepare` | ✅ PASS | |
| `yarn generate` | `prepare` | ✅ PASS | |
| `yarn build:packages` | `prepare` | ✅ PASS | |
| `yarn i18n:check-sync` | `lint` | ✅ PASS | |
| `yarn i18n:check-usage` | `lint` | ✅ PASS | The diff adds no user-facing string; the one new log line uses the structured logger facade. |
| `yarn typecheck` | `lint` | ✅ PASS | |
| `yarn test` | `test` | ⏳ PENDING | Still `IN_PROGRESS` at review time. |
| `yarn build:app` | `docker-build` | ⏭️ SKIPPED | Skipped by the workflow's own path/fork conditions, not by this review. |
| — | `mutate (shared, src/lib/auth/organizationAccess.ts)` | ✅ PASS | Mutation testing targeted at exactly the security predicate this commit changed — the strongest single signal on this diff, and it says the new `isUnrestrictedOrganizationScope` tests kill the mutants rather than merely executing the function. |
| — | `lint`, `ds-lint`, `audit-scope`, `scope`, `license/cla` | ✅ PASS | |
| — | `test`, `documents-multi-instance`, `ephemeral-integration (1-15/15)` | ⏳ PENDING | 17 checks still running. |

**⏳ CI is not yet complete on this head.** 17 checks — `test`, `documents-multi-instance` and all 15 `ephemeral-integration` shards — were still in progress when this review was submitted, and none had failed. Per this skill's contract the verdict is submitted as soon as the review is decided rather than waiting on CI, so **this approval is a code-review approval only and is not a statement that CI passed**. Branch protection and the merge queue remain the gate on the actual merge; if any of the pending checks goes red, this PR must not merge on the strength of this approval alone. I will post a CI-result follow-up comment when the run settles.

## Findings

### 💅 Nit

**1. `packages/core/src/modules/customers/api/dashboard/widgets/customer-todos/route.ts:64,73` — the same identifier means two different things across the module.** In this file `isUnrestrictedOrganizationScope` is a destructured **boolean**, while in `api/interactions/tasks/route.ts` and `api/todos/route.ts` the same name is the imported **function**. Nothing breaks — the widget never imports the function — but a reader moving between the three call sites has to re-establish which one they are looking at. Aliasing on destructure (`isUnrestrictedOrganizationScope: isUnrestrictedScope`) would remove the ambiguity. Author's call; I would not hold merge for it.

**2. `packages/core/src/modules/dashboards/lib/widgetScope.ts:13` — the new context field is optional, so a caller that forgets it silently reads as restricted.** `isUnrestrictedOrganizationScope?: boolean` combined with the widget's `=== true` test means an absent flag degrades to *fail-closed* (an unrestricted principal sees an empty widget) rather than fail-open, which is the correct direction to fail and is why I am not raising this higher. Optional is also the right call for a public type — making it required would break any external implementer of `WidgetScopeContext`. Worth a one-line comment on the field recording that the optionality is deliberate and that consumers must compare against `true`, so a future contributor does not "tidy" it into a truthiness check.

**3. Previous nit 5 remains open, as expected.** The collapsed-scope `logger.warn` in both routes (`tasks/route.ts:54`, `todos/route.ts:246`) still fires once per request, so a principal whose scope is permanently collapsed and who polls a dashboard list emits a warn line per poll. I said last time that no change was required for merge and that still holds — the line is how an operator notices the fail-closed path is being taken. Dropping it to `debug` or sampling it is a reasonable follow-up if these routes turn out to be polled.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `isUnrestrictedOrganizationScope` and `UnrestrictedOrganizationScopeInput` are **additive** exports from `packages/shared/src/lib/auth/organizationAccess.ts`; `isOrganizationAccessAllowed` keeps its name, path and signature.
- [x] No function signature changed in a breaking way — `isOrganizationAccessAllowed`'s input type is untouched; the refactor is internal.
- [x] No required type field removed or narrowed — `WidgetScopeContext` gains `isUnrestrictedOrganizationScope?: boolean`, an **optional** field. The four sibling widget routes (`new-deals`, `new-customers`, `next-interactions`) and `sales/widgets/dashboard/makeDashboardWidgetRoute.ts` destructure the context without it and are unaffected.
- [x] No HTTP route URL removed or renamed; no method changed — both routes keep their paths, methods, guards, and OpenAPI documents.
- [x] No field removed or retyped in an existing response shape — the `exportAll` envelope change alters only the **values** of `page`/`pageSize` on an already-empty page, and it changes them *toward* the shape the same handler's export path already returns. The keys and their types are unchanged, and the todos adapter still emits its `Deprecation` headers on the guarded response.
- [x] No database table or column changed — no entity, migration, or snapshot in the diff, correctly, since this is a query-guard fix.
- [x] Where a contract had to change, the old surface still works. Two behavior changes are worth naming explicitly, and both are safe:
  - `isOrganizationAccessAllowed` now evaluates `(input.allowedOrganizationIds ?? []).includes(...)` (`:29`). For every input the declared type permits this is byte-identical to the previous `input.allowedOrganizationIds.includes(...)`, because `null` is already short-circuited to `true` one line above. The `?? []` only changes an **untyped runtime `undefined`** from a thrown `TypeError` into a `false` — strictly more fail-closed, and the new `it.each([undefined, [], ['org-a']])` case pins it.
  - The dashboard widget's `isUnrestricted` source changed, analyzed in the Summary above and equivalent on every path `resolveWidgetScope` can actually produce.

  `BACKWARD_COMPATIBILITY.md` classifies none of the touched surfaces as protected in a way this change violates. **No protected-surface violation to warn about.**

## 🧪 Test Coverage

The commit adds 5 test cases and extends a sixth, distributed exactly across the gaps I named:

- `packages/shared/src/lib/auth/__tests__/organizationAccess.test.ts` — a new `describe` block for `isUnrestrictedOrganizationScope` covering superadmin-with-`[]` (true), non-superadmin-with-`null` (true), and a parameterized denial case over `undefined` / `[]` / `['org-a']`. This is the file the `mutate (shared, src/lib/auth/organizationAccess.ts)` CI job targets, and that job passing is meaningful evidence: it means these assertions actually **kill mutants** on the predicate rather than just executing it.
- `.../customer-todos/__tests__/route.test.ts` — the two widget cases from finding 2 of the previous review. They assert on the real argument positions (`listLegacyTodoRows` arg 3 = `organizationIds`, arg 5 = `options`; `listCanonicalTodoRows` arg 4 and arg 5), so they pin the wiring rather than just the status code, and the file's existing `beforeEach(jest.clearAllMocks)` keeps the `mockResolvedValueOnce` in the first case from leaking.
- Both route suites — one `unified: true` unrestricted case each, from finding 4.
- Both route suites — an `all=true` collapsed-scope case each, asserting the exact `{ items: [], total: 0, page: 1, pageSize: 0, totalPages: 1 }` envelope with `expectNoQueries()`.

On regression strength, worth stating plainly rather than implying: the two `exportAll` cases are true regression tests and fail against `fd9b275e`. The widget cases and the `unified: true` cases **pass on both heads** — the widget's old `organizationIds === null` formula and the routes' old inline formula produce the same flag for those inputs. That is expected and is exactly what I asked for: those branches were previously *unexecuted*, and the value of the new cases is that they lock the behavior in place so the next refactor of this policy cannot silently flip it in either direction. The 13 assertions that fail without the original fix, verified in the previous review, still stand behind the security behavior itself.

One gap remains, and it is a nit rather than a requirement: no test exercises `WidgetScopeContext` with `isUnrestrictedOrganizationScope` **absent**. Since `resolveWidgetScope` always populates it, this is only reachable through a mock or a downstream override, and the failure mode is fail-closed. A third widget case omitting the field and asserting `isUnrestricted: false` reaches the helpers would document the optional field's contract, alongside the comment suggested in nit 2.

No integration coverage is warranted beyond the unit level: the change adds no schema, no endpoint, and no UI, and the scoping decision it guards is fully determined by the resolved scope object the unit tests inject.

pkarw:
🤖 `om-auto-review-pr` completed: **APPROVED**. Lock released (swapped for `ci-monitoring` — see below).

Re-review of head `fc47738a` against my previous review of `fd9b275e`. The follow-up commit resolved all four open findings and introduced no new ones, so the verdict moves from changes-requested to approved and the pipeline label from `changes-requested` to `merge-queue`.

What I verified rather than took on trust:

- **The centralization is behavior-preserving where it matters.** `resolveWidgetScope` returns `organizationIds: null` only on the `scope?.allowedIds === null` branch, so the widget's old `organizationIds === null` formula and the new `isUnrestrictedOrganizationScope` flag are the same condition on every path that function can actually produce; where they differ (superadmin with an explicit selection) `organizationIds` is non-empty and the flag is inert.
- **`isOrganizationAccessAllowed` did not change for any real caller.** Its three production callers outside this module (`api_keys/api/keys/route.ts`, `directory/utils/organizationScopeGuard.ts`, `shared/lib/commands/scope.ts`) get identical results; the added `?? []` only turns an untyped runtime `undefined` from a thrown `TypeError` into a fail-closed `false`.
- **The new `exportAll` empty envelope matches the export shape** that `paginateTodoRows` and the unified branch already return for zero rows.
- **Test strength, stated honestly:** the two `all=true` cases are genuine regression tests that fail against the previous head; the widget and `unified: true` cases pass on both heads by design — they cover branches that were previously unexecuted, which is what the previous review asked for.

Validation source: **GitHub checks** (no local fallback needed this time). A maintainer released the workflow runs that were stuck in `action_required` during the previous review, so the gate is running for real on this head. `prepare`, `lint`, `ds-lint`, `audit-scope`, `scope`, `license/cla` and — notably — `mutate (shared, src/lib/auth/organizationAccess.ts)` all pass; the mutation job targets exactly the predicate this commit changed, so it is real evidence the new tests kill mutants rather than merely executing the function.

**⏳ CI is not finished.** 17 checks (`test`, `documents-multi-instance`, and all 15 `ephemeral-integration` shards) were still running when the review was submitted, and none had failed. This approval is a code-review approval and is **not** a statement that CI passed — branch protection and the merge queue remain the gate on the actual merge. I am carrying `ci-monitoring` and will post a CI-result comment when the run settles; if anything goes red I will correct the pipeline label.

Autofix: skipped (not my PR — re-run with `--autofix` to fix it here). Nothing needed fixing in any case.

pkarw:

**✅ CI is green on head `fc47738a`** — the same head the approving re-review covered, so this result applies to the reviewed code without qualification. The run settled ~20 minutes after the review was submitted, well inside the 40-minute follow-up budget.

Every reported check passed; none failed, errored, was cancelled, or timed out:

| Check | Result |
|-------|--------|
| `test` | ✅ SUCCESS |
| `ephemeral-integration (1-15/15)` | ✅ SUCCESS — all 15 shards |
| `documents-multi-instance` | ✅ SUCCESS |
| `mutate (shared, src/lib/auth/organizationAccess.ts)` | ✅ SUCCESS |
| `prepare`, `lint`, `ds-lint`, `audit-scope`, `scope`, `license/cla` | ✅ SUCCESS |
| `audit`, `docker-build`, `merge-coverage`, `comment`, `[code]smith` | ⏭️ SKIPPED by the workflows' own conditions |

This resolves the CI-pending disclosure carried in the review body and the completion comment: the approval was a code-review approval submitted while 17 checks were still running, and those 17 have now all passed. The verdict is unchanged and `merge-queue` stands. I am removing `ci-monitoring` — no further follow-up is outstanding from this agent.

`mergeStateStatus` is now `CLEAN` and `reviewDecision` is `APPROVED`. The PR carries `skip-qa` rather than `needs-qa`, so the QA-approval gate does not apply. Merging remains a maintainer's call — this agent does not merge.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-22T11:02:14Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

Claim note: the only assignee was @Duang777 (the PR author, assigned by the review hand-back), the `in-progress` label was absent, and the previous `om-auto-review-pr` lock was explicitly released at 10:40Z — so no live automation claim was held. `ci-monitoring` is a meta label, not a lock signal.

pkarw:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ PASS — every UI surface that reads through the newly gated code paths still renders its data, for both unrestricted and legitimately-scoped restricted principals.
**Environment:** `http://127.0.0.1:5001` (disposable ephemeral instance, production Next.js build on a testcontainers Postgres) · roles `admin`, `superadmin`, `employee` · browser `playwright` (Chromium 1.62.1)
**Verified:** `fix/issue-5466-todos-org-scope` @ `fc47738a`

This PR is backend-only, so there is no new UI to look at. What UI QA can meaningfully
answer here is the regression question: the change inserts a fail-closed guard in front of
three read paths that the admin UI depends on, and the risk is that the guard fires for
callers it should not and silently empties a legitimate list. Every surface that consumes
`listLegacyTodoRows` / `listCanonicalTodoRows` was therefore driven end to end.

### Scenario (P0/P1 — customer tasks and todos read paths)

**Where to click:** `/backend/customer-tasks` · `/backend/customers/people-v2/<id>` → Tasks tab · `/backend` (Customer Tasks widget)

| 1 | As `admin`, cold-load `/backend/customer-tasks` — the canonical `GET /api/customers/interactions/tasks` route the PR gates | The list renders the seeded customer tasks and is not emptied by the new branch | `200 …/interactions/tasks?page=1&pageSize=50`; both tasks visible; *Showing 1 to 2 of 2 results in 234ms*. Shell painted at 1152ms, rows at 2024ms | ✅ |
| 2 | Type `send proposal` into the list's own search field | The query reaches the gated route and narrows the result set instead of short-circuiting | `200 …/interactions/tasks?…&search=send+proposal`; only matching rows remain | ✅ |
| 3 | Open the person detail page and switch to the **Tasks** tab — the deprecated `GET /api/customers/todos` legacy adapter, also gated | Both tasks render through the legacy adapter | `200 /api/customers/todos?page=1&pageSize=20&entityId=…`; both tasks render with *Planned* status and *Priority 3* | ✅ |
| 4 | Open `/backend` and let the **Customer Tasks** dashboard widget load — the path that now receives `resolveWidgetScope`'s new `isUnrestrictedOrganizationScope` flag | Passing the flag through does not empty an unrestricted widget read | `200 /api/customers/dashboard/widgets/customer-todos?limit=5`; the widget lists the tasks with customer, timestamp and *View record* | ✅ |
| 5 | Switch to a 390×844 mobile viewport and reload the tasks list | The list still renders and stays readable | Card, search, Views control and table all render; *Showing 1 to 4 of 4 results in 142ms* | ✅ |
| 6 | Sign in as `superadmin` (`isSuperAdmin` → `isUnrestrictedOrganizationScope` is `true`) and load the tasks list | The tenant-wide read is preserved — the PR must **not** fail closed here | API `200`, `total=2`; the table renders 2 data rows | ✅ |
| 7 | Sign in as `employee` (restricted principal resolving to an explicit organization id set) and load the tasks list | The restricted path still applies `{ $in: [orgIds] }` and returns the caller's own rows | API `200`, `total=2`; the table renders 2 data rows | ✅ |

### Screenshots

**Step 1 — `admin` on `/backend/customer-tasks` (canonical route, 2 of 2 results)**
step-01

**Step 2 — search narrows through the gated route**
step-02

**Step 3 — person detail → Tasks tab (legacy `/api/customers/todos` adapter)**
step-03

**Step 4 — dashboard **Customer Tasks** widget (`resolveWidgetScope` flag path)**
step-04

**Step 5 — mobile viewport (390×844)**
step-05

**Step 6 — `superadmin`: tenant-wide read preserved**
step-06-superadmin

**Step 7 — `employee`: explicitly-scoped restricted read still populated**
step-06-employee

### Notes for QA

- **Coverage boundary — read this before treating the screenshots as proof of the fix.**
  This run verifies the *no-regression* half of the change. The defect the PR actually
  closes — a restricted principal whose resolved scope collapses to `organizationIds:
  null` or `[]` receiving a tenant-wide read — cannot be reproduced from the UI against
  the demo seed, because every demo principal resolves either to an explicit organization
  set or to an unrestricted scope (`scope.allowedIds === null`). That branch is covered by
  the route and helper unit tests shipped in this PR, not by this evidence.
- **Observed, unrelated to this PR:** the `employee` role receives `403` on
  `GET /api/customers/todos` while receiving `200` on
  `GET /api/customers/interactions/tasks`. That is the legacy adapter's own feature guard
  and predates this change; the practical consequence is that an employee never reaches
  the new fail-closed branch on the legacy route at all.
- **Perceived performance** on the changed route was healthy on a cold production-build
  load: app shell at 1152ms, populated table at 2024ms, with server-reported query times
  of 142–234ms.
- **Fixtures** (one person and two customer tasks) were created through the API and
  removed at the end of the run; the environment itself is disposable.

pkarw:

This change ships no browser-level test — the suites added here are jest route and helper
tests, which is the right place for the fail-closed matrix, but nothing pins the *UI*
consequence: that the three gated read paths keep rendering data for callers whose scope
did not collapse. The QA above was manual; lock it in with an automated test
(run `/om-integration-tests`).

1. **Setup** — via the repo's integration fixtures (`packages/core/src/helpers/integration`):
   create a person with `createPersonFixture`, then two customer tasks against it with
   `POST /api/customers/todos` (`{ entityId, title, todoCustom: { priority: 3 } }`), using a
   uniquely stamped title so the assertions cannot pass on seeded demo data.
2. **Act** — sign in with `login(page, 'admin')` and drive, in order:
   - `/backend/customer-tasks` (canonical `GET /api/customers/interactions/tasks`);
   - the same page's search field, filtering to one stamped title;
   - `/backend/customers/people-v2/<personId>` → the **Tasks** tab (legacy
     `GET /api/customers/todos`, which is *not* the default tab — it must be clicked);
   - `/backend` and the **Customer Tasks** dashboard widget
     (`GET /api/customers/dashboard/widgets/customer-todos`).
3. **Assert** — each stamped task is visible on each surface, and each of the three
   endpoints answered `200` with a non-empty payload. Repeat the canonical-list assertion
   for `superadmin` (unrestricted, `allowedIds === null`) and `employee` (restricted with an
   explicit organization set), so a future change that widens the fail-closed condition
   fails the suite instead of silently emptying a legitimate list.
4. **Teardown** — delete both tasks (`DELETE /api/customers/todos` with `{ id, linkId }`) and
   the person, in a `finally`.

**Two gotchas this run hit, worth encoding in the test:** never wait for `networkidle`
(the DOM event bridge holds an SSE connection open, so it never fires), and scope the list
search to `getByPlaceholder('Search', { exact: true })` — a bare `getByRole('textbox')`
grabs the global sidebar search instead and the test hangs.

pkarw:

Every UI surface that reads through the newly gated code paths still renders its data — the canonical tasks list, its search, the legacy `/api/customers/todos` adapter behind the person-detail Tasks tab, and the dashboard Customer Tasks widget — verified for an unrestricted principal (`superadmin`), a tenant `admin`, and a restricted principal with an explicit organization set (`employee`). The fail-closed guard did not empty any legitimate reader.

Two things a reviewer should carry forward. First, the coverage boundary stated in the evidence comment: the *positive* case this PR fixes (a restricted principal whose scope collapses to `null`/`[]`) is not reproducible from the UI against the demo seed, so it rests on the route and helper unit tests in this PR — the screenshots prove the absence of a regression, not the presence of the fix. Second, no labels were changed by this run: it was evidence-only, and the PR carries `skip-qa` rather than `needs-qa`, so the self-QA sign-off exception does not apply and `qa-approved` was neither earned nor needed here. The label set is back to exactly what it was before the claim.

PR #5479, opened 2026-08-22
fix(customer_accounts): resolve portal address server-side to stop hydration flicker

### LINKED ISSUE #5457: bug:  [customer_accounts] — Portal address on two admin pages briefly shows the wrong value and triggers a browser console error
## Summary

On the "Użytkownicy klientów" (customer portal users) page, the line showing the customer
portal's web address ("Adres portalu") displays incorrectly for a split second when the page
first loads, then visibly corrects itself. The browser console logs an error every time this
happens.

## Page URL

`/backend/customer_accounts/users`

## Steps to reproduce

1. Log in as `admin@acme.com`.
2. Open the browser's DevTools → Console.
3. Go to Ustawienia → Portal klienta → Użytkownicy klientów (`/backend/customer_accounts/users`).

## Expected result

The portal address shown on the page is correct and stable from the very first moment the page
appears — no flicker, no console error.

## Actual result

The page briefly shows the portal address as `/[org-slug]/portal` (missing the `http://...`
domain part), then a fraction of a second later it corrects itself to the full address
(`http://localhost:3000/[org-slug]/portal`). This is visible as a brief flash/flicker of the
text. Every time this happens, an error is logged to the browser console.

Image

Image

## Environment

- Branch: `main` (0.7.0 pre-release sweep)
- Module or area: `customer_accounts` — Portal Users page, "Adres portalu" line
- Browser/Runtime: Chrome
- Reproduces on both `demo.openmercato.com` and locally.

patzick:
Fixes #5457

## 🎯 What was wrong

The "Portal URL" / "Adres portalu" hint on the customer-portal admin pages was built inline during render:

```tsx
url: `${typeof window !== 'undefined' ? window.location.origin : ''}/[org-slug]/portal`
```

Those pages are `"use client"` components, but Next.js still server-renders them. So the server emitted `Portal URL: /[org-slug]/portal` and hydration produced `Portal URL: http://localhost:3000/[org-slug]/portal`. React treats that as a hydration mismatch: it logs the error the reporter saw in the console and repairs the DOM by rewriting the text — which is exactly the visible flicker from the issue. The same expression also fed the `href` of the "Open Portal" button, so that attribute was empty in the server HTML too.

## 🔧 The fix

The origin is knowable on the server, so resolve it there instead of guessing on the client. Each affected route entrypoint becomes a thin server component that reads the request headers and hands the resolved origin to the existing client component as a `portalOrigin` prop. The first painted markup already carries the correct address, and server and client render identical output — no mismatch, no console error, nothing to correct.

New helper `packages/core/src/modules/customer_accounts/lib/portalUrl.ts`:

- `resolveRequestOrigin(headers)` mirrors what `window.location.origin` would report — `x-forwarded-host`/`host` plus `x-forwarded-proto`, taking the client-facing hop when a proxy chain appends values, falling back to `APP_URL` / `NEXT_PUBLIC_APP_URL`, then to an empty string.
- `buildPortalUrlPattern(origin)` / `buildPortalRootUrl(origin)` render the address. With an unresolvable origin they degrade to the root-relative `/[org-slug]/portal`, which is identical on both sides — so the degraded path still cannot reintroduce a mismatch.
- Host headers are client-controlled, so the value is validated against a `host[:port]`/bracketed-IPv6 pattern before it reaches rendered text or a link `href`, and a non-`http(s)` `x-forwarded-proto` is ignored.

The protocol default is worth a note: when a terminating proxy omits `x-forwarded-proto`, a loopback host defaults to `http` and any other host to `https`, since a public hostname without the header is far more often TLS-terminated than not.

## 📂 About the diff size

The three client page bodies were moved next to their routes so the route file itself can be a server component (`page.tsx` → `PortalUsersPageClient.tsx`, `PortalUserDetailPageClient.tsx`, `CustomerAccountsSettingsPageClient.tsx`). This is the established pattern in this repo — see `ai_assistant`'s `AiUsageStatsPageClient.tsx` / `page.tsx` pair — and the generator's `isPageCandidateFile` deliberately excludes uppercase-prefixed components, so the colocated files do not become routes.

Git renders these as add+delete because `page.tsx` still exists at each path, but **the moved content is otherwise byte-identical**. Diffing each original against its new client file yields only:

- the `portalUrl` helper import
- `export default function X()` → `export function XClient({ portalOrigin }: Props)`
- the two/one portal-address expressions listed above

Nothing else in those ~1,600 lines changed.

## 🎯 Base branch: `main`, deliberately — and it needs a `develop` counterpart

This PR targets `main` at the requester's explicit instruction ("start from main branch and PR it to main branch"), following the same hotfix-to-`main` route as #5432 and #5431. That is a deliberate choice, not an oversight of the repo's `develop` default.

It does leave a real gap, correctly identified in review: `origin/develop` is ~82 commits ahead of `main`, nothing syncs `main` → `develop`, and `develop` still carries `window.location.origin` in these files — so #5457 stays live there. Worse, `develop` has already diverged from `main` on `users/page.tsx` itself, and this PR moves ~600 lines out of that file, so **the next `develop` → `main` promotion will hit a modify/delete-shaped conflict on files whose content moved.** Resolving that by taking `develop`'s side would silently reinstate the bug.

**Whoever resolves that merge: the conflict is expected.** The correct resolution keeps this side's structure — `page.tsx` as a thin server wrapper plus the `*PageClient.tsx` bodies — and re-applies any `develop`-only changes on top of the client file. The `portalUrlHydration` guard test will fail loudly if `window.location` comes back.

**Resolution:** the maintainer will sync `develop` manually, so no companion PR is opened from here and this PR stays on `main`. The conflict guidance above is written down precisely so that manual sync has a documented expected shape.

**Tracked so it survives this merge:** #5490 carries the `develop` sync, including the expected conflict resolution and the guard test that catches a botched one. `Fixes #5457` below therefore stays as-is — #5457 closes when `main` is fixed, and #5490 is what remains open for `develop`. (Re-review finding 6.)

## 🧩 Extensibility note (not a breaking change)

These three pages moved from client components to async server components. A `widgets/components.ts` override registered against the `page:/backend/customer_accounts/{users,users/[id],settings}` handle now wraps a server component, and `resolveRegisteredComponent` applies wrappers by composition — so a `"use client"` wrapper that previously wrapped these pages would no longer be able to render the child. No such override ships in this repo, so nothing is broken today, but downstream consumers with a custom wrapper on these handles should know.

## 📌 Scope note

The issue names two pages (portal users list and user detail). The **portal settings page** derived the same value the same way and is linked directly from the banner this PR fixes ("Open Configuration"), so it is fixed here too rather than left as an identical known defect one click away.

## 🧪 Tests

- `lib/__tests__/portalUrl.test.ts` — 17 cases over origin resolution (plain host, forwarded host/proto, proxy chains, missing proto, malformed host rejection, non-http proto, `APP_URL` / `NEXT_PUBLIC_APP_URL` fallbacks, empty origin) and the URL builders (including trailing-slash and degraded root-relative output).
- `__tests__/portalUrlHydration.test.ts` — regression guard asserting the three client surfaces never reference `window.location`, and that each route entrypoint is a server component that resolves the origin via `resolveRequestOrigin(await headers())` and passes it down. This is what keeps the bug from being reintroduced by the next edit.
- Updated three existing tests that referenced the old file paths (`demoCredentialsCopy`, `uiContractCleanup`, `crmNames.parallel`) and the `extension-points.ts` `source` pointer for the users DataTable host.

Locally, per instruction, only related tests were run — the rest is left to CI:

- `jest src/modules/customer_accounts` → **549 passed, 0 failed** (66/68 suites). The 2 non-running suites are `api/portal/events/{stream,notification-stream}.test.ts`, which fail to resolve a module on the untouched base commit too — verified by stashing this change and re-running. Pre-existing, unrelated.
- `tsc --noEmit` on `packages/core` → **0 errors** once the `#generated/entities.ids.generated` not-found errors are excluded; those are present on the base commit as well (generated files are not built in this checkout).
- ESLint on all seven touched files → clean.

## 🔄 Review round 2 (commit `09306dd1`)

Addressed the four actionable findings from @adeptofvoltron's review:

- **Minor 2 — configured `APP_URL` could not correct the guessed protocol.** Real bug I introduced: `configuredOrigin()` was only consulted when the Host header was missing, which never happens, so a plain-HTTP deployment behind a proxy that omits `x-forwarded-proto` rendered `https://` links that do not answer — a regression against the old client-side behavior, which was always right. The configured origin now supplies the protocol when it names the requested host, ahead of the loopback heuristic; an explicit `x-forwarded-proto` still wins over both.
- **Minor 3 — duplicate name.** Renamed to `resolvePortalRequestOrigin`, with a file comment explaining why it cannot just call the shared `resolveRequestOrigin` (that one needs a `Request` to read `req.url`; a server component only has `headers()`).
- **Minor 4 — +3 `om-ds/require-page-wrapper`.** Moved the `<Page>`/`<PageBody>` shell into the three server wrappers, matching the `ai_assistant` precedent; the client bodies return fragments. `PageBody` keeps each page's original spacing class (`space-y-4` for the list, `space-y-6` for the detail and settings pages), and the detail page's three loading/not-found/error states each had a single child, so the rendered layout is unchanged. `lint:ds` on `customer_accounts` now reports **0** `require-page-wrapper` findings; the 11 remaining `no-raw-table` warnings are pre-existing and simply moved with the settings file.
- **Nit 5 — dead `'::1'` entry** dropped, since `HOST_PATTERN` only accepts a colon-bearing host in bracketed form.

The review also noted the test suite exercised the fallback precedence only through an empty header set — a state no real request produces. Added cases with a realistic `{ host: 'app.local:3000' }` plus `APP_URL=http://app.local:3000` (asserting `http://`, which fails against the previous commit), an `x-forwarded-proto`-wins case, a different-host case, and bracketed-IPv6 loopback coverage with and without a port. Plus guards keeping the `Page` shell in the route files and out of the client bodies. Suite is now 559 passing.

## 🔄 Review round 3 (no code change)

The re-review of `09306dd1` confirmed all four round-2 fixes against the code and raised one new minor finding plus two no-change notes:

- **Finding 6 — `Fixes #5457` would delete the only tracker for the unshipped half.** Correct, and it is now fixed the way the reviewer's second option describes: `Fixes #5457` stays, and **#5490** tracks the `develop` sync so the remaining work outlives this merge. #5490 carries the expected promotion-merge resolution and names the guard test that fails loudly if a bad resolution reinstates `window.location.origin`.
- **Nit 7 — duplicated origin-resolution precedence.** The reviewer agreed the rename was the right trade for a hotfix and asked only that the observation not be lost; filed as **#5491**, to be acted on if a third headers-only caller ever appears.
- **Note 8 — the Minor 2 residual.** Reviewer explicitly wants no change: showing the host in the user's address bar rather than a canonical origin they are not on is the better behavior, and the defensive `configuredOrigin()` branch is fine as-is. Left alone.

No code changed in this round, so CI's green result on `09306dd1` still stands.

## 🔍 What to verify in QA

Open `/backend/customer_accounts/users`, `/backend/customer_accounts/users/<id>` and `/backend/customer_accounts/settings` with DevTools console open: the portal address should be complete and stable in the first paint, with no hydration error logged and no visible text correction. The "Open Portal" button should carry the full absolute URL.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5479"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789996664&installation_model_id=432243&pr_number=5479&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5479&signature=0aad88006cfc2658cd5cb721f0b8311a7f4c2fae57391ce54885ae849cf02589"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


patzick:
🏷️ label rationale

- 🐛 `bug` — this is a corrective change: the portal address rendered incorrectly on first paint and React logged a hydration error every time the page loaded.
- 🔄 `changes-requested` — replaces `review` after the `om-auto-review-pr` **re-review** on head `09306dd1`. The four actionable findings from the first pass (`ac888dca`) are all verified fixed, and the round-1 major is now waived by a documented maintainer decision recorded in the PR description. One new minor remains: `Fixes #5457` auto-closes the issue on merge into `main` (the default branch) while `develop` still carries the defect, deleting the only tracker for the manual sync that is the agreed remedy. This repo requests changes on minor findings too, so the PR comes back for a one-line description edit rather than a code change.
- 📸 `needs-qa` — the change touches `.tsx` UI-rendering files, so it does not qualify for the automated-verification exemption; a human should confirm the address is stable and the console is clean on all three pages. Round 2 moved the `<Page>`/`<PageBody>` shell across the server/client boundary on those same pages, which is exactly the kind of edit QA should eyeball even though the rendered markup is unchanged by analysis. Kept in place through the transition.
- ⏫ `priority-medium` — unchanged. The defect is visible on every load of three admin pages during the 0.7.0 pre-release sweep, but it is cosmetic and blocks no workflow.
- ⚠️ `risk-medium` — unchanged, and both review passes confirmed the rating. The relocated page bodies are otherwise untouched (round 2 verified with a whitespace-insensitive diff: only the shell tags and their import changed), but three admin route entrypoints change shape from client to server components, which is more than a cosmetic edit.
- 🔒 `in-progress` — released by this review run; see the completion comment.

No `skip-qa` and no `qa-approved`: code-review approval is not QA approval, and with `qaGate` on, the QA gate stays owned by a QA reviewer.

patzick:
🤖 `om-auto-fix-issue` run complete: the hydration mismatch behind #5457 is fixed by resolving the portal origin from the request headers in a server component instead of reading `window.location.origin` during a client render, with 17 unit cases on the new resolver and a source-level guard that fails if any of the three surfaces reaches for `window.location` again. Locally only the related tests were run per the request — `jest src/modules/customer_accounts` gave 549 passed with the two pre-existing `api/portal/events` resolution failures confirmed against the untouched base, `tsc --noEmit` on `packages/core` is clean, and ESLint on the touched files is clean; the full gate is left to CI, which is still running. The PR keeps `needs-qa` and no QA approval was granted from this chain. Lock released.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-08-22T13:29:22Z. Other auto-skills will skip this PR until the lock is released.

patzick:
✅ CI is green on `ac888dcaa` — every job that ran passed: `test` (6m29s), `ephemeral-integration` (6m52s), `docker-build` (6m3s), `lint`, `ds-lint`, `prepare`, `scope`, `audit-scope`, and all three CodeQL analyses. Worth stating explicitly because `skipping` on this repo can be ambiguous: `test` and `ephemeral-integration` genuinely ran and passed here, so this is a real gate result rather than a suppressed one; the `skipping` jobs are `audit`, `mutate`, `comment`, `merge-coverage` and `[code]smith`.

The PR still carries `needs-qa` and no `qa-approved`, so the QA gate holds the merge until a reviewer confirms the fix in a browser. What to check: open `/backend/customer_accounts/users`, `/backend/customer_accounts/users/<id>` and `/backend/customer_accounts/settings` with the DevTools console open — the portal address should be complete and stable in the first paint, with no hydration error logged and no visible text correction, and the "Open Portal" button should carry the full absolute URL.

adeptofvoltron (dismissed):
# 🔍 Code Review: fix(customer_accounts): resolve portal address server-side to stop hydration flicker

## 🎯 Summary

This PR fixes the hydration mismatch behind #5457 by moving portal-address resolution from `window.location.origin` (evaluated during a client render, so absent from the server HTML) to the server, where the request headers already carry the answer. Each of the three affected route entrypoints becomes a thin async server component that calls `resolveRequestOrigin(await headers())` and passes the result down as a `portalOrigin` prop; the page bodies move next to their routes as `*PageClient.tsx` files.

The diagnosis is correct and the chosen mechanism is the right one — the origin genuinely is knowable on the server, and passing it as a prop makes the server HTML and the hydrated output identical by construction rather than by convention. I verified the PR's central claim about the diff size myself: diffing each original `page.tsx` against its new `*PageClient.tsx` yields only the helper import, the `export default function X()` → `export function XClient({ portalOrigin })` signature change, and the two/one portal-address expressions. Nothing else in the ~1,600 relocated lines changed. I also confirmed that `isPageCandidateFile` in `packages/cli/src/lib/generators/scanner.ts:123-125` excludes uppercase-prefixed component files, so the colocated clients cannot become routes; that the `page.meta.ts` files carrying the RBAC guards are untouched, so no auth surface moved; and that the backend catch-all (`apps/mercato/src/app/(backend)/backend/[...slug]/page.tsx:125`) renders `<Component params={match.params} />` from a server component under `force-dynamic`, so an async server page and the plain `params` object shape are both correct here. The degraded path is genuinely safe: an unresolvable origin yields the root-relative `/[org-slug]/portal`, which is identical on both sides and so cannot reintroduce the mismatch.

Scope reviewed: the new `lib/portalUrl.ts` helper and its tests, the three route entrypoints and their client bodies, the regression guard, the three updated tests, and the `extension-points.ts` source pointer. The findings below are about where the fix lands and how the origin is resolved when the proxy is quiet — not about the approach.

## Verdict

❌ **Request changes** — one major finding: the fix targets `main`, but `develop` (82 commits ahead of `main`, and already diverged on one of the very files this PR deletes) still carries the `window.location.origin` code, so #5457 stays live there and the next promotion merge will conflict on a file that no longer exists on this side. Three minor findings follow it: an unreachable `APP_URL` fallback that makes plain-HTTP deployments render `https://` links, a second exported `resolveRequestOrigin` whose semantics disagree with the existing shared one, and three new DS-lint findings on a metric the repo actively ratchets down.

## 🧪 Validation Gate

**Validation source: GitHub checks** (per this repo's GitHub-checks-first review policy — CI already ran the full gate on head `ac888dca`, so no local re-run was needed).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | Covered by the `prepare` job (4m21s) and `docker-build` (6m3s). |
| `yarn generate` | ✅ PASS | Covered by `prepare`; no generated-file drift reported. |
| `yarn build:packages` (second pass) | ✅ PASS | Covered by `docker-build`, which builds packages and app end to end. |
| `yarn i18n:check-sync` | ✅ PASS | Covered by `lint` (1m19s). No new user-facing strings in this diff. |
| `yarn i18n:check-usage` | ✅ PASS | Covered by `lint`. |
| `yarn typecheck` | ✅ PASS | Covered by `test` (6m29s) and `docker-build`. |
| `yarn test` | ✅ PASS | `test` job green, including the new `portalUrl` and `portalUrlHydration` suites. |
| `yarn build:app` | ✅ PASS | Covered by `docker-build`. |

Supplementary checks: `ds-lint` ✅ pass but reporting **+3 findings vs `main`** (see Minor 3), `CodeQL` / `Analyze (actions, javascript-typescript, python)` ✅ pass, `audit-scope` ✅ pass, `scope` ✅ pass.

**Pending checks:** `ephemeral-integration (none)` is still pending at review time. Per the review policy this review is not held for it — branch protection and the merge queue gate the actual merge. No further follow-up on that check will come from this review pass beyond the CI-result comment.

## Findings

### ⚠️ Major

**1. The fix lands on `main` only; `develop` keeps the bug, and the next promotion merge will conflict on a deleted file.**

`packages/core/src/modules/customer_accounts/backend/customer_accounts/users/page.tsx` (and the two sibling pages) — this PR's `baseRefName` is `main`, but the repo's documented base branch is `develop` (`.ai/agentic.config.json` → `"baseBranch": "develop"`; 56 of the last 60 merged PRs targeted `develop`). I checked what that means concretely for this change:

- `origin/develop` is **82 commits ahead** of `origin/main`, and `origin/main` carries 10 commits `develop` does not have — so the two have genuinely diverged and there is no automated back-merge (`release.yml` and `release-prepare.yml` only tag and publish from `main`; nothing syncs `main` → `develop`).
- `origin/develop` still contains the defect verbatim: `git show origin/develop:.../users/page.tsx` has `window.location.origin` at lines 502 and 527, and `lib/portalUrl.ts` does not exist there.
- Worse, `develop` has **already diverged from `main` on `users/page.tsx` itself** (4 insertions, 1 deletion between the two). This PR deletes ~600 lines from that file and moves them to `PortalUsersPageClient.tsx`, so the next `develop` → `main` promotion hits a modify/delete-shaped conflict on a file whose content moved — and whichever side a hurried resolver picks, taking `develop`'s version silently restores `window.location.origin` and reopens #5457.

There is precedent for hotfixing `main` directly (#5432 and #5431 did), so targeting `main` may well be deliberate. The gap is that nothing carries the fix forward. Please do one of the following before merge: retarget this PR to `develop` (rebasing off `origin/develop`), or keep the `main` target and open the companion `develop` PR in the same cycle, linking it here so the promotion merge has a resolved counterpart to land against. Either way, say which in the PR description so the next person resolving that merge knows the moved-file conflict is expected.

### 🔹 Minor

**2. The `APP_URL` / `NEXT_PUBLIC_APP_URL` fallback is unreachable in practice, so a plain-HTTP deployment renders `https://` links.**

`packages/core/src/modules/customer_accounts/lib/portalUrl.ts:67-73` — `resolveRequestOrigin` consults `configuredOrigin()` only when `readHost()` returns `null`, which requires both `x-forwarded-host` and `host` to be absent or malformed. HTTP/1.1 mandates `Host`, and Next always has it during a server render, so in a real deployment the configured-origin branch never runs.

The consequence is in `readProtocol` (line 37-44): when a terminating proxy omits `x-forwarded-proto`, any non-loopback host defaults to `https`. That is a sensible guess in the common case, but it is only a guess, and the operator's explicit configuration cannot correct it. An internal deployment served over plain HTTP behind such a proxy, or a dev/staging host reached at `http://192.168.1.20:3000` or `http://app.local:3000`, will render `https://<host>/[org-slug]/portal` in the hint text and put the same broken absolute URL in the "Open Portal" `href` — even when `APP_URL=http://app.local:3000` is set correctly. Before this PR that link was always right, because the browser computed it; the hydration fix should not cost correctness in those setups.

The test suite does not catch this because every fallback case (`portalUrl.test.ts:69-83`) passes an empty headers object, a state that never occurs in production, so the tests assert a precedence order the runtime never exercises.

Suggested fix: let the configured origin win over the guessed protocol. Minimally, in `readProtocol`, when `x-forwarded-proto` is absent, take the protocol from `configuredOrigin()` if it is set and its host matches, before falling back to the loopback heuristic. Alternatively, have `resolveRequestOrigin` prefer `configuredOrigin()` outright when it is configured — operators set `APP_URL` precisely to name the canonical public origin. Then add a test with a realistic `{ host: 'app.local:3000' }` plus `APP_URL=http://app.local:3000` asserting `http://`, so the precedence that actually runs is the one under test.

**3. A second exported `resolveRequestOrigin` with different semantics from the existing shared one.**

`packages/core/src/modules/customer_accounts/lib/portalUrl.ts:67` — `@open-mercato/shared/lib/url` already exports a function with exactly this name (`packages/shared/src/lib/url.ts:233`) doing header-based origin resolution from `x-forwarded-proto` / `x-forwarded-host` / `host`. The two disagree on the interesting case: the shared one falls back to the request URL's own protocol, while the new one guesses `https` for non-loopback hosts. `packages/shared/AGENTS.md` is explicit here — "MUST check for existing utilities before adding new helpers — avoid duplication" and "MUST centralize reusable types and constants here to prevent drift across packages".

The new helper is not simply redundant: the shared version needs a `Request` (it reads `req.url`), while a server component only has `headers()`, so it cannot be reused verbatim. That is a fair reason to write new code — but shipping a same-named export with divergent behavior is exactly the drift that guidance warns about, and the next person who greps `resolveRequestOrigin` will find two answers.

Suggested fix: either extend the shared helper with a headers-only entry point (`resolveOriginFromHeaders(headers)`) and have `portalUrl.ts` call it, keeping one definition of the precedence rules; or, if the module-local behavior should stay module-local, rename it to something unambiguous such as `resolvePortalRequestOrigin` and note in the file comment why it differs from the shared one.

**4. Three new `om-ds/require-page-wrapper` findings on a metric the repo ratchets down.**

`packages/core/src/modules/customer_accounts/backend/customer_accounts/{settings,users,users/[id]}/page.tsx:1` — the `ds-lint` bot reports `om-ds/require-page-wrapper` going 22 → 25, and names these three files. The rule (`packages/eslint-plugin-ds/rules/require-page-wrapper.js:36-42`) fires on any backend `page.tsx` with a default export that renders JSX but no `<Page>`, which is exactly the shape of the new thin server wrappers.

Two things make this worth addressing rather than accepting. First, the precedent the PR description cites points the other way: `packages/ai-assistant/src/modules/ai_assistant/backend/config/ai-assistant/usage/page.tsx` keeps `<Page><PageBody>` in the *server* wrapper and renders `<AiUsageStatsPageClient />` inside it — so the established pattern produces zero new findings, while this PR leaves the wrapper in the client body. Second, there is currently no escape hatch: `eslint.ds.config.mjs` sets `linterOptions.noInlineConfig: true`, and `.ai/specs/2026-07-05-ds-lint-ci-escalation-and-alert-migration.md` records per-line opt-outs as a prerequisite that has not landed yet (the bot's own "DS lint opt-outs" row reads 0/0). So an `eslint-disable` comment is not available, and the +3 will sit in the baseline until someone restructures these files anyway.

Suggested fix: move `<Page>` / `<PageBody>` up into the three server `page.tsx` wrappers, matching the `ai_assistant` shape, and have each `*PageClient.tsx` return the inner content as a fragment. In the settings and detail pages the `FormHeader` block currently sits between `<Page>` and `<PageBody>`, so this is a small restructure rather than a pure lift — if that turns out to fight the layout, say so in the PR description so the +3 is a recorded, argued exception rather than an unexplained delta.

### 💅 Nit

**5. Dead entry in `LOOPBACK_HOSTS`.**

`packages/core/src/modules/customer_accounts/lib/portalUrl.ts:19` — the set contains both `'[::1]'` and `'::1'`, but the unbracketed form is unreachable: `HOST_PATTERN` (line 17) only accepts colons inside the bracketed alternative, so a bare `::1` host is rejected by `readHost` and never reaches `readProtocol`. Dropping `'::1'` makes the set say what it means. (`'[::1]'` is genuinely reachable and correctly survives the port-stripping regex, since `[::1]` does not end in `:digits`.)

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the changed exports are page-module default exports, which are consumed by the auto-discovery loader by path, not imported by name. The one named-export rename that matters is internal to the module's own tests, which the PR updates.
- [x] No function signature changed in a breaking way — the new server pages keep the `{ params?: { id?: string } }` shape the catch-all passes at `apps/mercato/src/app/(backend)/backend/[...slug]/page.tsx:125`.
- [x] No required type field removed or narrowed — `portalOrigin` is a new required prop on new component types.
- [x] No HTTP route URL removed or renamed — the three backend routes keep their paths; only the module file backing each one changed, and `page.meta.ts` (title, breadcrumb, `requireAuth` / `requireFeatures`) is untouched at all three.
- [x] No field removed or retyped in an existing response shape — no API surface in this diff.
- [x] No event or message name renamed or removed — none touched.
- [x] No CLI command or flag renamed or removed — none touched.
- [x] No database table or column renamed or removed — no entity or migration changes.
- [x] No config key renamed and no default changed silently — `APP_URL` / `NEXT_PUBLIC_APP_URL` are read, not redefined. (Their *effective* precedence is the subject of Minor 2, but no key or documented default changes.)
- [x] Where a contract had to change: not applicable — `extensionPoints.usersTable.source` (`extension-points.ts:6`) is a documentation pointer to the host file, updated in step with the move, and the `tableId` (`customer_accounts.admin.users`) that third-party widgets actually target is unchanged, so no injection spot moved.

One extensibility note that is not a break but is worth knowing: these three pages went from client components to async server components, so a `widgets/components.ts` override registered against the `page:/backend/customer_accounts/users` handle now wraps a server component. `resolveRegisteredComponent` (`packages/shared/src/modules/widgets/component-registry.ts:112-113`) applies wrappers by composition, so a `"use client"` wrapper that previously wrapped these pages would no longer be able to render the child. No such override ships in this repo, so nothing is broken today — but it is the kind of thing worth a line in the PR description for downstream consumers.

## 🧪 Test Coverage

Coverage for this change is good and deliberately layered, which is the right instinct for a bug whose failure mode is invisible to a normal render test.

`lib/__tests__/portalUrl.test.ts` covers origin resolution across plain host, forwarded host and protocol, proxy chains that append values, an omitted `x-forwarded-proto`, a malformed host that must be rejected rather than rendered, a non-`http(s)` forwarded protocol, both env fallbacks, and the empty-origin case — plus the two URL builders including trailing-slash normalization and the degraded root-relative output. The malformed-host case (`portalUrl.test.ts:60-62`) is the one that matters most, since that value reaches a link `href`, and it asserts the right thing.

`__tests__/portalUrlHydration.test.ts` is the guard that keeps the bug from coming back: it asserts source-level that none of the three client surfaces references `window.location`, that each route entrypoint is a server component resolving the origin via `resolveRequestOrigin(await headers())`, and that the prop is threaded through. Source-text assertions are usually a weak form, but here they are the correct tool — the defect is a *render-boundary* property that a jsdom render cannot observe, and the test fails loudly (ENOENT) rather than silently passing if a file is renamed. I confirmed the three updated tests (`demoCredentialsCopy`, `uiContractCleanup`, `crmNames.parallel`) were repointed correctly and that `crmNames.parallel` now renders the client component directly with an explicit `portalOrigin`.

Gaps worth closing, both tied to findings above:

- **The fallback precedence is tested in a state that cannot occur.** Every `configuredOrigin` test passes `headersOf({})`, but a real request always has `host`, so the branch under test is unreachable in production (Minor 2). Add a case with `{ host: 'app.local:3000' }` and `APP_URL=http://app.local:3000` asserting `http://app.local:3000`, which will fail today and pass once the precedence is fixed.
- **No test pins the loopback-vs-public protocol heuristic against a bracketed IPv6 host.** `readProtocol` handles `[::1]` (with and without a port) correctly, but nothing asserts it, so the port-stripping regex and the bracketed set entry are both unguarded. One case for `{ host: '[::1]:3000' }` expecting `http://[::1]:3000` would cover it — and would have surfaced the dead `'::1'` entry from Nit 5.

adeptofvoltron:
Thanks @patzick — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

adeptofvoltron:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED. Lock released (swapped for `ci-monitoring` — one required check, `ephemeral-integration (none)`, was still pending at review time and gets a follow-up comment).

One major finding (the fix targets `main` only, while `develop` still carries the defect and has already diverged on a file this PR deletes) and three minor ones — full report in the review. Validation source: GitHub checks on head `ac888dca`; no local gate re-run was needed. autofix: skipped (not my PR — re-run with --autofix to fix it here).

adeptofvoltron:

The check that was pending at review time, `ephemeral-integration (none)`, has since settled to **pass** (6m52s). All 12 reporting checks on head `ac888dca` are green; the remaining 5 are conditional jobs that skipped (`audit`, `comment`, `mutate`, `[code]smith`). Nothing in the CI outcome changes the review verdict — it stays **changes requested** on the code-review findings, not on CI.

Removing `ci-monitoring`; no further follow-up will come from this agent.

patzick:
🤖 `om-auto-fix-pr` claimed this PR and is working the review findings from @adeptofvoltron.

patzick:
🔍 Review round 2 — pushed as `09306dd1`. Thanks @adeptofvoltron, this was a genuinely useful review; **Minor 2 was a real regression I introduced** and I would not have caught it from the tests as written.

**Minor 2 — fixed.** You are right that `configuredOrigin()` was unreachable: `readHost()` only returns `null` when both `x-forwarded-host` and `host` are absent, and a real request always has `Host`. So the `https` guess in `readProtocol` was unconditional for non-loopback hosts and the operator had no way to correct it — meaning `APP_URL=http://app.local:3000` still produced an `https://app.local:3000` hint and, worse, the same dead URL in the "Open Portal" `href`. That is strictly worse than the pre-PR behavior, where the browser computed it. I took your minimal suggestion: `readProtocol` now consults `configuredOrigin()` when `x-forwarded-proto` is absent and uses its protocol if it names the requested host, falling back to the loopback heuristic otherwise. An explicit `x-forwarded-proto` still wins over both. I kept the request host as the displayed host rather than preferring the configured origin outright, so which host you see still matches the one you are browsing.

**Minor 3 — fixed by renaming.** Renamed to `resolvePortalRequestOrigin`, with a file comment recording why it is not simply a call into the shared helper: `@open-mercato/shared/lib/url`'s version reads the protocol off `req.url` and a server component has no `Request`, only `headers()`. I chose the rename over extending the shared module because this one also validates the untrusted host header and infers a protocol — behavior I did not want to quietly graft onto a shared export that other callers already depend on. Happy to invert that and add `resolveOriginFromHeaders` to `shared` instead if you prefer one definition of the precedence rules; say the word.

**Minor 4 — fixed.** You were right that the precedent I cited pointed the other way. Moved `<Page>`/`<PageBody>` into the three server wrappers and made the client bodies return fragments. It turned out not to fight the layout after all: `Page` and `PageBody` are pure spacing divs, so passing each page's original class through (`space-y-4` for the list, `space-y-6` for the detail and settings pages) reproduces the previous spacing exactly, and the detail page's three loading/not-found/error states each render a single child, where the wrapper's `space-y-*` had no effect anyway. `ConfirmDialogElement` now sits inside `PageBody`, which is inert — it is a mount tracker plus a closed Radix dialog, so nothing enters the layout flow. `lint:ds` over `customer_accounts` now reports **0** `require-page-wrapper` findings. The 11 `no-raw-table` warnings it still reports are pre-existing and just moved with the settings file.

**Nit 5 — fixed.** Dropped the bare `'::1'`; your reading was right that `HOST_PATTERN` only admits a colon-bearing host in bracketed form, so it was unreachable.

**Test gaps — both closed.** Your point that every fallback test passed an empty header set, a state production never reaches, was the sharpest note in the review. Added a `{ host: 'app.local:3000' }` + `APP_URL=http://app.local:3000` case asserting `http://` (it fails against `ac888dca` and passes now), a case where `APP_URL` names a *different* host and must be ignored, one where `x-forwarded-proto` overrides `APP_URL`, and bracketed-IPv6 loopback coverage with and without a port. Local suite: **559 passing**, up from 549.

**Major 1 — documented, not resolved, and it needs your call.** I have not retargeted. This PR targets `main` on the requester's explicit instruction, matching the #5432 / #5431 hotfix route — so the target is deliberate. But your analysis of the consequence is correct and I have written it into the PR description rather than leaving it implicit: `develop` keeps the bug, nothing back-merges, and because `develop` has already diverged on `users/page.tsx` while this PR moves ~600 lines out of it, the promotion merge will hit a modify/delete-shaped conflict — where taking `develop`'s side silently reopens #5457. The description now names the expected resolution (keep this side's server-wrapper + `*PageClient` split, replay `develop`-only edits onto the client file) and notes the `portalUrlHydration` guard will fail loudly if `window.location` returns. I deliberately did **not** open the companion `develop` PR: that is a second PR against a different base, and choosing between "retarget" and "ship both" is a maintainer decision rather than one this run should make for you. Flagging it for the requester now.

I also added the extensibility note you suggested about these pages becoming server components and what that means for a `"use client"` wrapper registered on the `page:` handles.

**Validation source: local, related tests only** (as instructed for this run) — `jest src/modules/customer_accounts` 559 passed / 0 failed, `tsc --noEmit` on `packages/core` clean, `lint:ds` and ESLint clean on the touched files. The two `api/portal/events` suites still fail to resolve `@open-mercato/events/bus` in this checkout; that reproduces on the untouched base and is a missing built-package artifact, not a regression. Full gate is CI's, now running on `09306dd1`.

patzick:
✅ CI is green on `09306dd1` — every job that ran passed: `test` (6m22s), `ephemeral-integration` (6m37s), `docker-build` (3m38s), `ds-lint` (40s, and the `require-page-wrapper` delta from the review is gone), `lint`, `prepare`, `scope`, `audit-scope`, all three CodeQL analyses, and the CLA check. As on the previous round, `test` and `ephemeral-integration` genuinely ran rather than being suppressed, so this is a real gate result; the `skipping` jobs are `audit`, `mutate`, `comment`, `merge-coverage` and `[code]smith`.

The four actionable findings from the review are fixed and verified. The one open item is Major 1 — how the fix reaches `develop` — which is a maintainer decision between retargeting this PR and shipping a companion `develop` PR; it is documented in the PR description and has been raised with the requester. The PR still carries `needs-qa` without `qa-approved`, so manual QA is the other gate before merge.

patzick:
🔁 Re-requested review from @adeptofvoltron on head `09306dd1`, as asked in the hand-back.

Status of the five findings, with the two residuals stated plainly rather than glossed:

- **Major 1 — resolved by decision, not by code.** The maintainer will sync `develop` manually, so this PR stays on `main` and no companion PR is opened. The PR description records the expected shape of that promotion merge (keep the server-wrapper + `*PageClient` split; the `portalUrlHydration` guard fails loudly if `window.location` returns).
- **Minor 2 — fixed for the reachable case, with two residuals.** The configured origin now supplies the protocol when it names the requested host. Residual (a): the `configuredOrigin()` branch inside `resolvePortalRequestOrigin` is still only reached when no Host header exists, which — exactly as you observed — production never does; I kept it as a defensive default rather than delete it. Residual (b): when `APP_URL` names a *different* host than the one being browsed, a plain-HTTP deployment still gets the `https` guess. Your alternative (prefer `configuredOrigin()` outright) would cover (b); I chose the narrower fix so the displayed host still matches the host in the address bar. Say if you would rather have (b) closed at the cost of that.
- **Minor 3 — took your second option, not your first.** Renamed to `resolvePortalRequestOrigin` rather than adding `resolveOriginFromHeaders` to `@open-mercato/shared`. The duplication of *logic* therefore remains, and only the name collision is gone. Happy to invert it if you prefer one definition of the precedence rules.
- **Minor 4 — fixed and independently confirmed.** The DS lint bot on this head reports **Δ 0 on every rule**, `om-ds/require-page-wrapper` back to 22. The 11 `no-raw-table` entries under "New findings" are the same pre-existing findings at a new path (114 → 114, Δ 0) because the settings body moved file.
- **Nit 5 — fixed.**

Both test-coverage gaps you named are closed: the realistic `{ host }` + `APP_URL` precedence case (which fails against `ac888dca`) and bracketed-IPv6 loopback coverage with and without a port.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-08-22T15:42:21Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron (dismissed):
# 🔍 Code Review (Re-review): fix(customer_accounts): resolve portal address server-side to stop hydration flicker

## 🎯 Summary

This is a re-review of head `09306dd1`, which answers the four actionable findings from the previous pass on `ac888dca`. I re-checked each of the five prior findings against the code rather than against the response comment, and all four actionable ones are genuinely fixed:

- **Minor 2 (unreachable `APP_URL` fallback)** — `readProtocol` (`packages/core/src/modules/customer_accounts/lib/portalUrl.ts:45-65`) now consults `configuredOrigin()` when `x-forwarded-proto` is absent and adopts its protocol when the configured origin names the very host being requested, before falling through to the loopback-vs-public guess. An explicit `x-forwarded-proto` still wins over both, which is the right precedence. This closes the regression I flagged: an operator who sets `APP_URL=http://app.local:3000` now gets `http://` links again instead of a dead `https://` one.
- **Minor 3 (colliding export name)** — renamed to `resolvePortalRequestOrigin` (`portalUrl.ts:91`), with a file-header note (`portalUrl.ts:7-12`) recording why it cannot simply call `@open-mercato/shared/lib/url`'s `resolveRequestOrigin`: that one reads the protocol off `req.url` and a server component has only `headers()`. The name collision is gone; I confirmed by grep that no consumer anywhere still imports the old name.
- **Minor 4 (+3 `om-ds/require-page-wrapper`)** — the `<Page>` / `<PageBody>` shell moved up into the three server wrappers and the client bodies now return fragments. Independently confirmed rather than taken on trust: the DS lint bot on this head reports `om-ds/require-page-wrapper` at **22 → 22, Δ 0**, and Δ 0 on every other rule. The 11 `no-raw-table` entries under "New findings" are the same pre-existing findings at a new path (114 → 114).
- **Nit 5 (dead `'::1'`)** — dropped, with the reason recorded in a comment at `portalUrl.ts:26`.

I paid particular attention to the Minor 4 restructure, because moving a layout shell across a server/client boundary in a 1,000-line reindentation diff is exactly where a silent regression hides. A whitespace-insensitive diff of the three client bodies between `ac888dca` and `09306dd1` shows **only** the removed `Page`/`PageBody` import and the shell tags replaced by fragments — not one other line changed in any of the three files. I also walked the spacing arithmetic rather than accepting the claim: `cn` is `twMerge(clsx(...))` (`packages/shared/src/lib/utils.ts:4-6`), so `<PageBody className="space-y-6">` resolves the `space-y-4` default away instead of emitting both classes. The settings page previously got its header/body gap from `<Page>`'s own `space-y-6` and now gets the same gap from `PageBody`; the detail page's content was already inside a `space-y-6` `PageBody`; the list page keeps `space-y-4`; and the detail page's three loading/not-found/error states each render a single child, where `space-y-*` has no effect either way. The rendered layout is unchanged in all four cases.

The two test-coverage gaps I named are both closed, and closed in the way that matters — with a case that fails against the previous commit rather than one that merely documents current behavior.

Scope re-reviewed: the full round-2 diff (9 files), plus a re-check of the round-1 conclusions that round 2 could have disturbed.

## Verdict

❌ **Request changes** — one new minor finding, and under this repo's verdict rule a minor is enough to hand the PR back. The finding is a one-word edit to the PR description, not a code change: `Fixes #5457` will auto-close the issue when this merges into `main` (the repository's default branch), while `develop` still carries the defect and the agreed remedy is a manual sync that would then have nothing tracking it.

To be explicit about the prior **Major 1**: it is not driving this verdict. It is waived, and the waiver is documented in the PR description ("the maintainer will sync `develop` manually, so no companion PR is opened from here and this PR stays on `main`"), together with the expected shape of the promotion merge. That satisfies the "explicit, documented waiver" clause. I am recording it as accepted rather than resolved, because the underlying facts have not changed and I re-verified them: `origin/develop` is still 82 commits ahead of `origin/main`, and `git show origin/develop:.../users/page.tsx` still has `window.location.origin` at lines 502 and 527. The waiver rests on a human remembering to act — which is precisely why the finding below is worth the extra round-trip.

## 🧪 Validation Gate

**Validation source: GitHub checks** (per this repo's GitHub-checks-first review policy — CI ran the full gate on head `09306dd1`, so no local re-run was warranted). No check was pending at review time and none failed, so there is nothing for a local fallback to stand in for.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | Covered by `prepare` and `docker-build` (3m38s). |
| `yarn generate` | ✅ PASS | Covered by `prepare`; no generated-file drift reported. |
| `yarn build:packages` (second pass) | ✅ PASS | Covered by `docker-build`, which builds packages and app end to end. |
| `yarn i18n:check-sync` | ✅ PASS | Covered by `lint`. Round 2 adds no user-facing strings. |
| `yarn i18n:check-usage` | ✅ PASS | Covered by `lint`. |
| `yarn typecheck` | ✅ PASS | Covered by `test` (6m22s) and `docker-build`. |
| `yarn test` | ✅ PASS | `test` job green, including the expanded `portalUrl` and `portalUrlHydration` suites. |
| `yarn build:app` | ✅ PASS | Covered by `docker-build`. |

Supplementary checks on `09306dd1`: `ds-lint` ✅ pass with **Δ 0 on every rule** (the +3 `require-page-wrapper` from round 1 is gone), `ephemeral-integration (none)` ✅ pass (6m37s), all three CodeQL analyses ✅ pass, `scope` / `audit-scope` / `license/cla` ✅ pass. `audit`, `mutate`, `comment`, `merge-coverage` and `[code]smith` are conditional jobs that skipped. Branch protection on `main` is not readable through the API (404), so every reported check is treated as required; all of them are green.

## Findings

### 🔹 Minor

**6. `Fixes #5457` closes the issue on merge, deleting the only tracker for the half of the fix that has not shipped.**

PR description, the `Fixes #5457` line — `main` is this repository's default branch (`gh repo view --json defaultBranchRef` → `main`), and this PR targets `main`, so GitHub will close #5457 the moment it merges. At that point the defect is still live on `develop`, where it will sit until someone performs the manual sync the description promises. The issue is the natural place that reminder lives, and merging destroys it.

This is a rider on the accepted Major 1 waiver rather than a reopening of it. The waiver's whole premise is "a human will carry this forward manually"; the merge silently removes the artifact that would remind them, and it also tells the original reporter the bug is fixed while the branch that becomes the next release still reproduces it. Because `develop` is where the great majority of work lands, the practical shape of this is a *scheduled regression*: whenever `develop` is promoted, the fix disappears from `main` too unless the conflict is resolved exactly as the description instructs.

The repo's own tracker convention names the escape hatch (`.ai/trackers/github.md` → Conventions: "To reference without auto-closing, use a plain issue link"), so the fix is mechanical. Either:

- change `Fixes #5457` to a plain reference such as `Refs #5457` and leave the issue open until `develop` is synced, adding a short note on the issue saying the `main` side is fixed and what remains; or
- keep `Fixes #5457` and open a follow-up issue for the `develop` sync in the same cycle, linking it from this PR, so the outstanding work survives the merge.

Either satisfies me — the point is only that the remaining work outlives the merge. I do not need to see the sync itself in this PR.

### 💅 Nit

**7. The origin-resolution precedence rules now exist in two places, and the rename makes that harder to notice rather than easier.**

`packages/core/src/modules/customer_accounts/lib/portalUrl.ts:38-65` — you asked directly whether to invert Minor 3 and add `resolveOriginFromHeaders` to `@open-mercato/shared` instead. My answer is no, not in this PR: your reasoning is sound. Grafting untrusted-host validation and protocol inference onto a shared export that other callers already depend on is a behavior change to code this bug fix has no business touching, and doing it under time pressure on a hotfix branch is how a hydration fix turns into an incident. The rename plus the file-header note was the correct trade.

What remains is that the two helpers still encode the same precedence question with different answers, and the distinct name means a future reader is less likely to compare them at all. If a third caller ever needs headers-only resolution, that is the moment to unify — a shared `resolveOriginFromHeaders` with the validation and inference from this file, and both call sites moved onto it. Worth a follow-up issue, not a change here. Author's call entirely.

**8. A note on the Minor 2 residual you flagged — accepted as designed.**

`packages/core/src/modules/customer_accounts/lib/portalUrl.ts:52-60` — you correctly identified that a plain-HTTP deployment still gets the `https` guess when `APP_URL` names a *different* host than the one being browsed, and offered to close it by preferring `configuredOrigin()` outright. Do not. Your narrower fix is the better behavior: the hint text and the "Open Portal" link should describe the host in the user's address bar, not a canonical origin they are not currently on. An operator browsing a host their `APP_URL` does not describe has already stepped outside the configured setup, and showing them a link to a different host would be a more confusing failure than an over-optimistic scheme. Your residual (a) — the `configuredOrigin()` branch inside `resolvePortalRequestOrigin` being reachable only without a Host header — is likewise fine as a defensive default, and it is covered by tests. No change wanted on either.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `resolveRequestOrigin` → `resolvePortalRequestOrigin` renames a symbol that **this PR itself introduced** and that has never appeared in a release, so there is no external consumer and no deprecation window is owed. I verified by grep that the only references anywhere in `packages/` and `apps/` are the three route entrypoints and the module's own test file.
- [x] No function signature changed in a breaking way — the three server pages keep the `{ params?: { id?: string } }` shape the backend catch-all passes; the client components' `portalOrigin` prop is unchanged from round 1.
- [x] No required type field removed or narrowed — `RequestHeaderReader` is unchanged.
- [x] No HTTP route URL removed or renamed — the three backend routes keep their paths and their untouched `page.meta.ts` guards. Round 2 moves only JSX between the two files backing each route.
- [x] No field removed or retyped in an existing response shape — no API surface in this diff.
- [x] No event or message name renamed or removed — none touched.
- [x] No CLI command or flag renamed or removed — none touched.
- [x] No database table or column renamed or removed — no entity or migration changes.
- [x] No config key renamed and no default changed silently — `APP_URL` / `NEXT_PUBLIC_APP_URL` are read, not redefined. Round 2 does change their *effective precedence*: a configured origin naming the requested host now supplies the protocol where previously the loopback heuristic did. That is the fix for Minor 2 and it strictly widens the set of deployments that render a working link, so it needs no migration note — but it is a real behavior change and is correctly described in the PR body.
- [x] Where a contract had to change: not applicable. The `<Page>` shell moving from the client body into the route file is internal composition; the DS-lint contract that cares about it is satisfied (Δ 0), and no injection spot, table id or page handle moved.

The round-1 extensibility note still stands and is now written into the PR description: these three pages are async server components, so a `"use client"` wrapper registered against their `page:` handles could no longer render the child. Nothing in this repo registers one.

## 🧪 Test Coverage

Both gaps from the previous review are closed, and closed properly.

`lib/__tests__/portalUrl.test.ts:60-82` adds four cases that exercise the precedence the runtime actually reaches, not the empty-headers state production never produces: `{ host: 'app.local:3000' }` with `APP_URL=http://app.local:3000` asserting `http://` (this is the one that fails against `ac888dca`, which is what makes it a regression test rather than a description of current behavior); `APP_URL` naming a different host being ignored; an explicit `x-forwarded-proto` still winning over `APP_URL`; and bracketed IPv6 loopback with and without a port, which pins both the port-stripping regex and the surviving `'[::1]'` set entry that Nit 5 was about. The comment above the first case explains *why* the realistic-host framing matters, so the next person editing this file will not quietly revert to empty headers.

`__tests__/portalUrlHydration.test.ts:41-56` grows two new source-level guards: each route entrypoint owns the `Page`/`PageBody` shell, and no client surface renders one. That is the right instinct — the Minor 4 fix is a structural property that no render test would notice regressing, and this pins it. The assertions are exact-string (`toContain` on the full import line), which will need a touch if the import is ever reordered or the path changes; that is acceptable for a guard file whose failure mode is a loud, obvious mismatch rather than a silent pass, and it matches the style already established in this file.

Suite size went 549 → 559, and the `test` job is green on this head, so the additions are running in CI rather than only locally.

I have no further coverage requests. The one item that is untested is untestable here by construction: nothing in this repository can assert that `develop` receives the fix, which is the whole substance of finding 6.

adeptofvoltron:
Thanks @patzick — round 2 is genuinely good work. All four actionable findings are fixed, and I verified them against the code rather than the write-up: the `APP_URL` precedence fix does reach the branch a real request takes, the DS bot independently confirms `require-page-wrapper` back at Δ 0, and a whitespace-insensitive diff of the three client bodies shows the `Page`/`PageBody` restructure changed nothing but the shell tags and their import. The new test that fails against `ac888dca` is the right shape for a regression test.

I am handing the PR back for one item, and it is a PR-description edit rather than a code change: `Fixes #5457` will auto-close the issue when this merges into `main` (the repo's default branch), while `develop` still carries the defect and the agreed remedy is a manual sync. That merge deletes the only thing tracking the remaining half of the fix. Either switch to a plain `Refs #5457` and leave the issue open until `develop` is synced, or keep `Fixes` and open a follow-up issue for the sync, linked from here — either works.

Two open questions from your round-2 notes are answered in the review body: keep your narrower Minor 2 fix (do not prefer `configuredOrigin()` outright — the displayed host should match the address bar), and keep the rename rather than inverting Minor 3 into `@open-mercato/shared` (a follow-up, not this hotfix).

When the description is updated, re-request review and the automation can pick it up from the latest head.

adeptofvoltron:

Re-review of head `09306dd1`. All four actionable findings from the `ac888dca` pass are verified fixed against the code, and round-1's Major 1 is accepted as waived — the decision to keep this PR on `main` and sync `develop` manually is documented in the PR description, including the expected shape of the promotion merge. One new minor drives the verdict: `Fixes #5457` auto-closes the issue on merge into `main` (the default branch) while `develop` still reproduces the bug, so the manual sync loses its tracker. That is a PR-description edit, not a code change. Two nits in the body answer the author's open questions (keep the narrower Minor 2 fix; keep the rename rather than inverting Minor 3 into `@open-mercato/shared`) and need no action.

Validation source: GitHub checks — every reported check on `09306dd1` is green (`test`, `ephemeral-integration`, `docker-build`, `lint`, `ds-lint` at Δ 0 on every rule, `prepare`, `scope`, `audit-scope`, three CodeQL analyses, CLA). Nothing was pending at review time, so no CI follow-up is owed and no `ci-monitoring` label is applied.

patzick:
🤖 `om-auto-fix-pr` claimed this PR and is working re-review finding 6.

patzick:
🔍 Round 3 — finding 6 addressed, **no code change**, so CI's green result on `09306dd1` still stands.

**Finding 6 — fixed via your second option.** You are right that merging with `Fixes #5457` would close the only artifact reminding anyone that `develop` still reproduces the bug, and would tell the reporter it is fixed while the branch that becomes the next release is not. Rather than downgrade to `Refs`, I kept `Fixes #5457` — the `main` side genuinely is fixed and the reporter deserves that signal — and filed **#5490** to carry the `develop` sync. It records the expected promotion-merge resolution (keep this side's server-wrapper + `*PageClient` split; replay `develop`-only edits onto the client file), names `portalUrlHydration.test.ts` as the guard that turns a botched resolution into a red `test` job rather than a silent regression, and states its own done-condition. The PR description now links it from the `develop` section, and I left a note on #5457 itself so anyone landing there after the auto-close sees the branch split rather than assuming it is done everywhere.

**Nit 7 — filed as #5491, no change here.** Thank you for the direct answer on the invert question; I agree with the reasoning and have not touched `@open-mercato/shared`. #5491 records the trigger condition — a *third* headers-only caller — and explicitly says closing it as "not needed" is a fine outcome, so it is a note rather than a debt.

**Note 8 — no change, as instructed.** Both residuals left exactly as they are. Your argument that the hint and the "Open Portal" link should describe the host in the user's address bar rather than a canonical origin they are not currently on is the right call, and it is the reason I chose the narrow fix in the first place.

**On the Major 1 waiver being accepted rather than resolved** — that framing is fair and I have not tried to claim otherwise. The waiver does rest on a human acting, which is precisely why finding 6 was worth the round-trip; #5490 is what converts "someone will remember" into something with a URL.

Re-requesting review. Nothing in the diff changed since your pass, so this round is a description-and-tracking edit only — the code you reviewed on `09306dd1` is the code that would merge.

Remaining gates after this: one approving review, and the QA gate (`needs-qa` is still on and no `qa-approved` has been or will be applied from this chain).

patzick:
## 🔍 `om-auto-review-pr` — re-review of `570e1eb6`

**Verdict: ✅ approvable — no outstanding Medium+ findings.** `merge-queue` applied.

**Validation source: GitHub checks.** No local gate was re-run. All 13 reporting checks on `570e1eb6` are green — `test` (6m24s), `ephemeral-integration (none)`, `docker-build` (3m35s), `prepare` (3m23s), `lint`, `ds-lint`, `scope`, `audit-scope`, all three CodeQL analyses, and `license/cla`. `audit`, `mutate`, `comment` and `[code]smith` are conditional jobs that skipped.

### Base was stale — merged before judging

`main` had advanced **13 commits** since CI last ran, so the green result on `09306dd1` was measuring a diff that would never merge. Merged `origin/main` in and pushed `570e1eb6`; the merge was clean and CI above ran against the real merge result.

Worth stating precisely, because it determines whether the previous review still applies: the base merge made **zero** changes to this PR's own files. `git diff 09306dd1..570e1eb6 -- packages/core/src/modules/customer_accounts` is empty. Everything the merge brought in (`packages/ui/src/backend/DataTable.tsx`, `LookupSelect.tsx`, `useAutoDiscoveredFields.ts` and two new test files) is other people's already-reviewed work arriving from `main`. **The code reviewed on `09306dd1` is byte-for-byte the code that would merge.**

### Findings status

| # | Severity | Status |
|---|---|---|
| 1 | Major | Waived and documented; now additionally tracked by #5490 |
| 2 | Minor | Fixed on `09306dd1`, verified against the code |
| 3 | Minor | Fixed on `09306dd1`, verified against the code |
| 4 | Minor | Fixed; DS lint bot independently confirms Δ 0 on every rule |
| 5 | Nit | Fixed |
| 6 | Minor | **Resolved** — #5490 carries the `develop` sync with the promotion-merge resolution and guard test named; PR description links it; #5457 annotated so the auto-close is not misread |
| 7 | Nit | Filed as #5491; reviewer wanted no change here |
| 8 | Note | Reviewer explicitly wanted no change; left alone |

Under this repo's verdict rule (Medium+ → changes requested; only Low/nit or none → approved), nothing Medium or above remains open.

### ⚠️ What this label does not do

`merge-queue` records the pipeline state. It does **not** clear either gate that actually blocks the merge, and `mergeStateStatus` is still `BLOCKED`:

1. **GitHub's `reviewDecision` is still `CHANGES_REQUESTED`**, from the 15:48 review of the previous head — whose sole blocking finding (6) is the one resolved above. `main` requires one approving review, and this run cannot supply it: the PR author cannot approve their own PR. A reviewer re-review or a maintainer dismissing the now-stale verdict is required.
2. **`needs-qa` without `qa-approved`** remains a hard merge gate. No `qa-approved` has been or will be applied from this chain — it is earned by a human QA pass.

So: queued and ready from this chain's perspective, **not** mergeable yet. `om-approve-merge-pr` will still correctly refuse until (1) and (2) clear.

### 📸 QA steps

Open `/backend/customer_accounts/users`, `/backend/customer_accounts/users/<id>` and `/backend/customer_accounts/settings` with the DevTools console open: the portal address should be complete and stable in the first paint, with no hydration error logged and no visible text correction, and "Open Portal" should carry the full absolute URL.

PR #5480, opened 2026-08-22
fix(eudr): stop offering Delete on archived statements and surface the reason (#5461)

### LINKED ISSUE #5461: bug: archived EUDR statement offers Delete that can never succeed, and the failure hides the reason
## Summary

An archived EUDR due-diligence statement is permanently read-only and cannot be deleted — that part
is deliberate (`eudr.errors.archivedReadOnly` is thrown by both the update and the delete command).
The defect is that the statement page still offers **Delete** for such a record, and when the
attempt fails the reason is lost: the user gets a generic "Failed to delete the statement."

The same page proves the correct behaviour exists. After archiving, **Submit** and **Archive**
disappear, and **Save** returns the accurate message *"Archived statements are read-only."*
Only the Delete path drops the reason.

## Steps to reproduce

1. `/backend/eudr/statements/create` — create any statement (title + commodity). It is created as `draft`.
2. Open the statement, click **Archive**, confirm the dialog.
   The status becomes `Archived`; Submit and Archive correctly disappear.
3. Click **Delete** (also available from the list row menu **⋯ → Delete**), confirm.

## Expected

Delete is hidden or disabled for archived statements. If it is attempted anyway, the message names
the reason, as the Save path already does.

## Actual

| Action on an archived statement | Message shown |
|---|---|
| **Save** (after editing Notes) | ✅ "Archived statements are read-only." — accurate |
| **Delete** | ❌ "Failed to delete the statement." — reason lost |

The row remains in the list. Nothing indicates the action can never succeed, so the natural response
is to retry.

Wire level, same guard behind both:

```
PUT    /api/eudr/statements  {id, notes:"..."}  → 400 {"error":"eudr.errors.archivedReadOnly"}
DELETE /api/eudr/statements  {id}               → 400 {"error":"eudr.errors.archivedReadOnly"}
PUT    /api/eudr/statements  {id, status:"draft"} → 400 {"error":"eudr.errors.invalidTransition"}
```

Source (`packages/core/src/modules/eudr/commands/statements.ts`): the delete command throws
`archivedReadOnly` for `record.status === 'archived'`, and `canTransition('archived', …)` allows no
outgoing transition. So there is no un-archive path either.

## Impact

A draft can be archived from the UI in two clicks. The confirmation dialog says only
*"The statement will be archived"* — it does not warn that the step is irreversible. Afterwards the
record can be neither edited, un-archived, nor removed, and the only action still offered for it is
one that always fails.

Every mistaken or test statement that reaches `archived` therefore stays in the compliance register
permanently. Three such rows were produced while verifying this and cannot be cleaned up:
`4b36a5b6-2128-42e8-b8ae-40edf5e05d34`, `a31239be-f0c8-46c4-8357-a65c201d482e`,
`79359f4c-b6d2-45d3-b01f-770e9dee2bf9`.

## Suggested scope

Hiding/disabling Delete for archived statements and surfacing `eudr.errors.archivedReadOnly` on the
delete path would resolve the reported behaviour. Whether archived statements should be
un-archivable or deletable at all is a product decision and is deliberately left out of this report.

## Environment

- `https://demo.openmercato.com`, version reported as 0.6.7 in the footer, deployment from the 0.7.0 window
- Account `admin@acme.com`, organization Acme Corp
- Reproduced through the UI end to end, and independently at the API level
- Module introduced by #4358

patzick:
Closes #5461

## 🎯 Goal
An archived EUDR due-diligence statement no longer offers a Delete action that can never succeed, and when a delete is refused anyway the user is told why instead of getting generic copy.

## 🔍 Problem
Archiving a statement is a two-click, irreversible step. Afterwards the record can be neither edited, un-archived, nor removed — yet Delete stayed available on both the statement detail page and the list row menu (⋯ → Delete). Attempting it returned the accurate server reason `eudr.errors.archivedReadOnly`, but the list path replaced it with a constant "Could not delete the statement.", so nothing indicated the action could never succeed and the natural response was to retry. The same page already proved the correct behaviour exists: after archiving, Submit and Archive disappear and Save reports "Archived statements are read-only."

## 🔍 Root Cause
The terminal-`archived` rule lived only in `commands/statements.ts` and was duplicated nowhere the UI could see it. `deleteStatementCommand.execute` threw `CrudHttpError(400, { error: 'eudr.errors.archivedReadOnly' })` for `record.status === 'archived'`, mirroring the update guard, and `canTransition('archived', …)` allowed nothing outgoing. Neither delete surface consulted that rule: the detail page rendered `CrudForm` without `deleteVisible`, so `showDelete` fell back to `!isNewRecord` and Delete was always shown; the list rendered an unconditional `delete` row action. On the list, `handleDelete`'s catch discarded the thrown error — whose `error` property carried the token — and flashed the constant copy. This is the departure from the module's own pattern: `StatementLifecycleBar` already derives Submit/Archive visibility from `status`.

## What Changed
- `lib/statement-lifecycle.ts` — new `isStatementReadOnly(status)` / `canDeleteStatement(status)` exported next to `canTransition`, giving the terminal-`archived` rule a single owner.
- `commands/statements.ts` — the update guard and the delete guard now call the shared predicates instead of inline `status === 'archived'` comparisons. Behavior, status codes and the `eudr.errors.archivedReadOnly` token are unchanged.
- `backend/eudr/statements/[id]/page.tsx` — passes `deleteVisible={canDeleteStatement(record.status)}` to `CrudForm`, so Delete disappears for archived statements exactly as Submit and Archive already do. Its `onDelete` already routed through `translateEudrCrudError` and was correct, so it is untouched.
- `backend/eudr/statements/page.tsx` — the `RowActions` array is built conditionally and omits `delete` for archived rows; `handleDelete`'s catch now resolves the server reason and refreshes the list, so a row archived in another tab reveals its real status rather than silently failing again.
- `components/crudErrorI18n.ts` (re-exported from `formConfig.tsx`) — new `translateEudrErrorMessage(err, translate, fallback)` reads an `eudr.errors.*` token from either the rejection's `error` property (how `apiCall` rejections carry it) or its `message`, falling back to the caller's generic copy only when no known token is present.
- `i18n/{en,de,es,ko,pl}.json` — `eudr.lifecycle.confirmArchiveBody` now warns that archived statements are read-only and cannot be edited, restored, or deleted, closing the trap the issue's Impact section describes. Copy-only change to an existing key; no new keys.

Deliberately out of scope, as the issue requests: whether archived statements should be un-archivable or hard-deletable at all is a product decision, and the three orphaned rows the reporter created remain unremovable.

## 🧪 Tests
- `yarn workspace @open-mercato/core test --testPathPattern "modules/eudr"` — 19 suites / 125 tests passed.
- `yarn workspace @open-mercato/core typecheck` — clean. `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync` (in sync), `yarn i18n:check-usage` (advisory only) — all green. eslint over the changed files reports one pre-existing `react-hooks/exhaustive-deps` warning at `statements/page.tsx:284`, in a memo this change does not touch.
- The repo-wide `yarn test` and `yarn build:app` were left to CI rather than run locally.
- New in `lib/__tests__/statement-lifecycle.test.ts`: `archived` is the only non-deletable status; `canDeleteStatement` agrees with the read-only predicate the update guard uses; `archived` stays terminal for every status, so a blocked delete has no un-archive escape hatch.
- New `components/__tests__/crudErrorI18n.test.ts`: the reason is surfaced from a token on `error` and from a token on `message`, and untokened rejections (network error, bare object, `null`) plus tokens with no translation fall back to the generic copy rather than showing a raw key.
- `__integration__/TC-EUDR-007.spec.ts` already asserted the archived `PUT` and invalid-transition rejections; it now also asserts `DELETE /api/eudr/statements` → 400 `archivedReadOnly`, covering the wire-level contract the issue documents.

## 💥 Breaking Changes
- None. `isStatementReadOnly` / `canDeleteStatement` are additive exports; the API contract, response shapes, status machine and error token are unchanged. The only wire-visible difference is that two UI surfaces stop offering an action the server always rejected.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5480"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789997924&installation_model_id=432243&pr_number=5480&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5480&signature=0fa4cb7e02efabe803c399741810aaebe10f04128863df18d0824977a4bde4b6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/eudr/__integration__/TC-EUDR-007.spec.ts
+++ b/packages/core/src/modules/eudr/__integration__/TC-EUDR-007.spec.ts
@@ -534,6 +534,13 @@ test.describe('TC-EUDR-007: Statement lifecycle', () => {
       await expectErrorKey(archivedEditResponse, 'archivedReadOnly')
       const archivedReopenResponse = await updateStatement(request, token, { id: archivedStatementId, status: 'draft' })
       await expectErrorKey(archivedReopenResponse, 'invalidTransition')
+      const archivedDeleteResponse = await apiRequest(
+        request,
+        'DELETE',
+        `${STATEMENTS_PATH}?id=${encodeURIComponent(archivedStatementId)}`,
+        { token },
+      )
+      await expectErrorKey(archivedDeleteResponse, 'archivedReadOnly')
 
       const assessedStatementId = await createStatement(request, token, `TC-EUDR-007 Latest risk ${stamp}`)
       statementIds.push(assessedStatementId)
--- a/packages/core/src/modules/eudr/backend/eudr/statements/[id]/__tests__/page.test.tsx
+++ b/packages/core/src/modules/eudr/backend/eudr/statements/[id]/__tests__/page.test.tsx
@@ -1,0 +1,163 @@
+/**
+ * @jest-environment jsdom
+ */
+import * as React from 'react'
+import { waitFor } from '@testing-library/react'
+import { renderWithProviders } from '@open-mercato/shared/lib/testing/renderWithProviders'
+import EditEudrStatementPage from '../page'
+
+const apiCallMock = jest.fn()
+
+jest.mock('next/navigation', () => ({
+  useRouter: () => ({ push: jest.fn() }),
+}))
+
+jest.mock('next/link', () => ({
+  __esModule: true,
+  default: ({ children, ...props }: React.AnchorHTMLAttributes<HTMLAnchorElement>) => (
+    <a {...props}>{children}</a>
+  ),
+}))
+
+jest.mock('@open-mercato/ui/backend/Page', () => ({
+  Page: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+  PageBody: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+}))
+
+jest.mock('@open-mercato/ui/backend/DataTable', () => ({
+  DataTable: () => <div>table</div>,
+}))
+
+jest.mock('@open-mercato/ui/backend/detail', () => ({
+  ErrorMessage: ({ label }: { label: string }) => <div>{label}</div>,
+  LoadingMessage: ({ label }: { label: string }) => <div>{label}</div>,
+  RecordNotFoundState: ({ label }: { label: string }) => <div>{label}</div>,
+}))
+
+jest.mock('@open-mercato/ui/backend/utils/crud', () => ({
+  updateCrud: jest.fn(),
+  deleteCrud: jest.fn(),
+}))
+
+jest.mock('@open-mercato/ui/backend/utils/serverErrors', () => ({
+  createCrudFormError: (message: string) => new Error(message),
+}))
+
+jest.mock('@open-mercato/ui/backend/utils/apiCall', () => ({
+  apiCall: (...args: unknown[]) => apiCallMock(...args),
+}))
+
+jest.mock('@open-mercato/ui/backend/FlashMessages', () => ({
+  flash: jest.fn(),
+}))
+
+jest.mock('@open-mercato/ui/primitives/button', () => ({
+  Button: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+}))
+
+jest.mock('@open-mercato/ui/primitives/empty-state', () => ({
+  EmptyState: () => null,
+}))
+
+jest.mock('@open-mercato/ui/primitives/status-badge', () => ({
+  StatusBadge: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+}))
+
+jest.mock('@open-mercato/shared/lib/i18n/context', () => {
+  const stableTranslate = (key: string) => key
+  return {
+    useT: () => stableTranslate,
+    I18nProvider: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  }
+})
+
+const crudFormPropsCapture: { current: Record<string, unknown> | null } = { current: null }
+jest.mock('@open-mercato/ui/backend/CrudForm', () => ({
+  CrudForm: (props: Record<string, unknown>) => {
+    crudFormPropsCapture.current = props
+    return <div>form</div>
+  },
+}))
+
+jest.mock('../../../../../components/StatementLifecycleBar', () => ({
+  StatementLifecycleBar: () => null,
+}))
+
+jest.mock('../../../../../components/StatementReadinessChecklist', () => ({
+  StatementReadinessChecklist: () => null,
+}))
+
+jest.mock('../../../../../components/StatementRiskSection', () => ({
+  StatementRiskSection: () => null,
+  riskConclusionBadgeVariant: () => 'neutral',
+  riskTierBadgeVariant: () => 'neutral',
+}))
+
+jest.mock('../../../../../components/PlotMapPreview', () => ({
+  PlotMapPreview: () => null,
+}))
+
+jest.mock('../../../../../components/formConfig', () => ({
+  OrderSelectField: () => null,
+  ReferencedStatementsField: () => null,
+  actorRoleOptions: () => [],
+  activityTypeOptions: () => [],
+  commodityOptions: () => [],
+  statusBadgeVariant: () => 'neutral',
+  translateEudrCrudError: (err: unknown) => err,
+}))
+
+function statementRecord(status: string) {
+  return {
+    id: 'statement-1',
+    title: 'Statement 1',
+    commodity: 'coffee',
+    referenceNumber: null,
+    verificationNumber: null,
+    status,
+    activityType: null,
+    actorRole: null,
+    referencedStatements: [],
+    quantityKg: null,
+    supplementaryUnit: null,
+    supplementaryQuantity: null,
+    orderId: null,
+    submittedAt: null,
+    referenceIssuedAt: null,
+    orderSnapshot: null,
+    notes: null,
+    latestRisk: null,
+    createdAt: '2026-08-01T09:00:00.000Z',
+    updatedAt: '2026-08-01T10:00:00.000Z',
+  }
+}
+
+async function renderDetailFor(status: string): Promise<void> {
+  apiCallMock.mockImplementation(async (url: string) => {
+    if (typeof url === 'string' && url.startsWith('/api/eudr/statements?')) {
+      return { ok: true, result: { items: [statementRecord(status)] } }
+    }
+    return { ok: true, result: { items: [] } }
+  })
+  renderWithProviders(<EditEudrStatementPage params={{ id: 'statement-1' }} />)
+  await waitFor(() => expect(crudFormPropsCapture.current).not.toBeNull())
+}
+
+describe('EditEudrStatementPage delete affordance', () => {
+  beforeEach(() => {
+    apiCallMock.mockReset()
+    crudFormPropsCapture.current = null
+  })
+
+  it('keeps Delete available for a statement that can still be deleted', async () => {
+    await renderDetailFor('draft')
+
+    expect(crudFormPropsCapture.current?.deleteVisible).toBe(true)
+  })
+
+  it('hides Delete for an archived statement, which the server always refuses', async () => {
+    await renderDetailFor('archived')
+
+    expect(crudFormPropsCapture.current?.deleteVisible).toBe(false)
+  })
+})
--- a/packages/core/src/modules/eudr/backend/eudr/statements/[id]/page.tsx
+++ b/packages/core/src/modules/eudr/backend/eudr/statements/[id]/page.tsx
@@ -29,6 +29,7 @@ import {
   type ReferencedStatementValue,
   translateEudrCrudError,
 } from '../../../../components/formConfig'
+import { canDeleteStatement } from '../../../../lib/statement-lifecycle'
 import { StatementLifecycleBar } from '../../../../components/StatementLifecycleBar'
 import { StatementReadinessChecklist } from '../../../../components/StatementReadinessChecklist'
 import { StatementRiskSection, type StatementLatestRisk } from '../../../../components/StatementRiskSection'
@@ -644,6 +645,7 @@ export default function EditEudrStatementPage({ params }: { params?: { id?: stri
           backHref="/backend/eudr/statements"
           cancelHref="/backend/eudr/statements"
           deleteRedirect="/backend/eudr/statements"
+          deleteVisible={canDeleteStatement(record.status)}
           submitLabel={translate('eudr.statements.form.submitUpdate')}
           fields={fields}
           groups={groups}
--- a/packages/core/src/modules/eudr/backend/eudr/statements/__tests__/page.test.tsx
+++ b/packages/core/src/modules/eudr/backend/eudr/statements/__tests__/page.test.tsx
@@ -1,0 +1,191 @@
+/** @jest-environment jsdom */
+
+import * as React from 'react'
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+import EudrStatementsPage from '../page'
+
+type StatementRowStub = {
+  id: string
+  title: string
+  commodity: string
+  status: string
+  updatedAt: string
+  latestRisk: null
+}
+
+const mockApiCall = jest.fn()
+const mockRunMutation = jest.fn()
+const mockConfirm = jest.fn(async () => true)
+const mockFlash = jest.fn()
+
+const messages: Record<string, string> = {
+  'eudr.errors.archivedReadOnly': 'Archived statements are read-only.',
+  'eudr.statements.list.deleteError': 'Could not delete the statement.',
+  'eudr.statements.list.actions.delete': 'Delete',
+  'eudr.statements.list.actions.edit': 'Edit',
+  'eudr.statements.duplicateAction': 'Duplicate',
+}
+
+function statementRow(id: string, status: string): StatementRowStub {
+  return {
+    id,
+    title: `Statement ${id}`,
+    commodity: 'coffee',
+    status,
+    updatedAt: '2026-08-01T10:00:00.000Z',
+    latestRisk: null,
+  }
+}
+
+jest.mock('next/navigation', () => ({
+  useRouter: () => ({ push: jest.fn() }),
+}))
+
+jest.mock('next/link', () => ({
+  __esModule: true,
+  default: ({ children, ...props }: React.AnchorHTMLAttributes<HTMLAnchorElement>) => (
+    <a {...props}>{children}</a>
+  ),
+}))
+
+jest.mock('@open-mercato/ui/backend/Page', () => ({
+  Page: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+  PageBody: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+}))
+
+jest.mock('@open-mercato/ui/backend/DataTable', () => ({
+  DataTable: ({
+    data,
+    rowActions,
+  }: {
+    data?: StatementRowStub[]
+    rowActions?: (row: StatementRowStub) => React.ReactNode
+  }) => (
+    <div>
+      {(data ?? []).map((row) => (
+        <div key={row.id} data-testid={`row-${row.id}`}>
+          {rowActions?.(row)}
+        </div>
+      ))}
+    </div>
+  ),
+}))
+
+jest.mock('@open-mercato/ui/backend/RowActions', () => ({
+  RowActions: ({ items }: { items: Array<{ id?: string; label: string; onSelect?: () => void }> }) => (
+    <>
+      {items.map((item) => (
+        <button key={item.id} type="button" data-testid={`action-${item.id}`} onClick={item.onSelect}>
+          {item.label}
+        </button>
+      ))}
+    </>
+  ),
+}))
+
+jest.mock('@open-mercato/ui/backend/filters/ListEmptyState', () => ({
+  ListEmptyState: () => null,
+}))
+
+jest.mock('@open-mercato/ui/backend/injection/useGuardedMutation', () => ({
+  useGuardedMutation: () => ({
+    runMutation: (...args: unknown[]) => mockRunMutation(...args),
+    retryLastMutation: jest.fn(),
+  }),
+}))
+
+jest.mock('@open-mercato/ui/backend/confirm-dialog', () => ({
+  useConfirmDialog: () => ({
+    confirm: (...args: unknown[]) => mockConfirm(...args),
+    ConfirmDialogElement: null,
+  }),
+}))
+
+jest.mock('@open-mercato/ui/backend/utils/apiCall', () => ({
+  apiCall: (...args: unknown[]) => mockApiCall(...args),
+  withScopedApiRequestHeaders: (_headers: unknown, operation: () => unknown) => operation(),
+}))
+
+jest.mock('@open-mercato/ui/backend/utils/optimisticLock', () => ({
+  buildOptimisticLockHeader: jest.fn(() => ({})),
+}))
+
+jest.mock('@open-mercato/ui/backend/FlashMessages', () => ({
+  flash: (...args: unknown[]) => mockFlash(...args),
+}))
+
+jest.mock('@open-mercato/ui/backend/conflicts', () => ({
+  surfaceRecordConflict: jest.fn(() => false),
+}))
+
+jest.mock('@open-mercato/ui/primitives/button', () => ({
+  Button: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+}))
+
+jest.mock('@open-mercato/ui/primitives/status-badge', () => ({
+  StatusBadge: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+}))
+
+jest.mock('@open-mercato/shared/lib/frontend/useOrganizationScope', () => ({
+  useOrganizationScopeVersion: () => 1,
+}))
+
+jest.mock('@open-mercato/shared/lib/i18n/context', () => ({
+  useLocale: () => 'en',
+  useT: () => (key: string) => messages[key] ?? key,
+}))
+
+async function renderListWith(rows: StatementRowStub[]): Promise<void> {
+  mockApiCall.mockResolvedValue({
+    ok: true,
+    result: { items: rows, total: rows.length, totalPages: 1 },
+  })
+  render(<EudrStatementsPage />)
+  await waitFor(() => expect(screen.getByTestId(`row-${rows[0].id}`)).toBeTruthy())
+}
+
+describe('EUDR statements list delete affordance', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockConfirm.mockResolvedValue(true)
+  })
+
+  it('offers Delete for a statement that can still be deleted', async () => {
+    await renderListWith([statementRow('draft-1', 'draft')])
+
+    expect(screen.queryByTestId('action-delete')).not.toBeNull()
+  })
+
+  it('hides Delete for an archived statement, which the server always refuses', async () => {
+    await renderListWith([statementRow('archived-1', 'archived')])
+
+    expect(screen.queryByTestId('action-edit')).not.toBeNull()
+    expect(screen.queryByTestId('action-duplicate')).not.toBeNull()
+    expect(screen.queryByTestId('action-delete')).toBeNull()
+  })
+
+  it('names the reason when a delete is refused instead of showing generic copy', async () => {
+    mockRunMutation.mockRejectedValue(
+      Object.assign(new Error('[internal] eudr statement delete failed'), {
+        status: 400,
+        error: 'eudr.errors.archivedReadOnly',
+      }),
+    )
+
+    await renderListWith([statementRow('draft-1', 'draft')])
+    fireEvent.click(screen.getByTestId('action-delete'))
+
+    await waitFor(() => expect(mockFlash).toHaveBeenCalled())
+    expect(mockFlash).toHaveBeenCalledWith('Archived statements are read-only.', 'error')
+  })
+
+  it('falls back to the generic copy when the failure carries no reason', async () => {
+    mockRunMutation.mockRejectedValue(new Error('Failed to fetch'))
+
+    await renderListWith([statementRow('draft-1', 'draft')])
+    fireEvent.click(screen.getByTestId('action-delete'))
+
+    await waitFor(() => expect(mockFlash).toHaveBeenCalled())
+    expect(mockFlash).toHaveBeenCalledWith('Could not delete the statement.', 'error')
+  })
+})
--- a/packages/core/src/modules/eudr/backend/eudr/statements/page.tsx
+++ b/packages/core/src/modules/eudr/backend/eudr/statements/page.tsx
@@ -6,7 +6,7 @@ import { useRouter } from 'next/navigation'
 import { Page, PageBody } from '@open-mercato/ui/backend/Page'
 import { DataTable } from '@open-mercato/ui/backend/DataTable'
 import { ListEmptyState } from '@open-mercato/ui/backend/filters/ListEmptyState'
-import { RowActions } from '@open-mercato/ui/backend/RowActions'
+import { RowActions, type RowActionItem } from '@open-mercato/ui/backend/RowActions'
 import type { FilterDef, FilterValues } from '@open-mercato/ui/backend/FilterBar'
 import { apiCall, withScopedApiRequestHeaders } from '@open-mercato/ui/backend/utils/apiCall'
 import { buildOptimisticLockHeader } from '@open-mercato/ui/backend/utils/optimisticLock'
@@ -22,7 +22,8 @@ import type { LegacyColumnDef as ColumnDef } from '@tanstack/react-table/legacy'
 import { useOrganizationScopeVersion } from '@open-mercato/shared/lib/frontend/useOrganizationScope'
 import { useLocale, useT } from '@open-mercato/shared/lib/i18n/context'
 import type { EudrActivityType, EudrCommodity, EudrRiskConclusion, EudrRiskTier, EudrStatementStatus } from '../../../data/validators'
-import { commodityOptions, statementStatusOptions, statusBadgeVariant } from '../../../components/formConfig'
+import { commodityOptions, statementStatusOptions, statusBadgeVariant, translateEudrErrorMessage } from '../../../components/formConfig'
+import { canDeleteStatement } from '../../../lib/statement-lifecycle'
 import {
   riskConclusionBadgeVariant,
   riskTierBadgeVariant,
@@ -186,7 +187,11 @@ export default function EudrStatementsPage() {
       refreshRows()
     } catch (error) {
       if (surfaceRecordConflict(error, translate, { onRefresh: refreshRows })) return
-      flash(translate('eudr.statements.list.deleteError'), 'error')
+      flash(
+        translateEudrErrorMessage(error, translate, translate('eudr.statements.list.deleteError')),
+        'error',
+      )
+      refreshRows()
     }
   }, [confirm, mutationContextId, refreshRows, retryLastMutation, runMutation, translate])
 
@@ -309,30 +314,31 @@ export default function EudrStatementsPage() {
               </Link>
             </Button>
           )}
-          rowActions={(row) => (
-            <RowActions
-              items={[
-                {
-                  id: 'edit',
-                  label: translate('eudr.statements.list.actions.edit'),
-                  href: `/backend/eudr/statements/${row.id}`,
+          rowActions={(row) => {
+            const items: RowActionItem[] = [
+              {
+                id: 'edit',
+                label: translate('eudr.statements.list.actions.edit'),
+                href: `/backend/eudr/statements/${row.id}`,
+              },
+              {
+                id: 'duplicate',
+                label: translate('eudr.statements.duplicateAction'),
+                href: `/backend/eudr/statements/create?duplicateFrom=${encodeURIComponent(row.id)}`,
+              },
+            ]
+            if (canDeleteStatement(row.status)) {
+              items.push({
+                id: 'delete',
+                label: translate('eudr.statements.list.actions.delete'),
+                destructive: true,
+                onSelect: () => {
+                  void handleDelete(row)
                 },
-                {
-                  id: 'duplicate',
-                  label: translate('eudr.statements.duplicateAction'),
-                  href: `/backend/eudr/statements/create?duplicateFrom=${encodeURIComponent(row.id)}`,
-                },
-                {
-                  id: 'delete',
-                  label: translate('eudr.statements.list.actions.delete'),
-                  destructive: true,
-                  onSelect: () => {
-                    void handleDelete(row)
-                  },
-                },
-              ]}
-            />
-          )}
+              })
+            }
+            return <RowActions items={items} />
+          }}
           onRowClick={(row) => router.push(`/backend/eudr/statements/${row.id}`)}
           rowClickActionIds={['edit']}
           emptyState={(
--- a/packages/core/src/modules/eudr/commands/statements.ts
+++ b/packages/core/src/modules/eudr/commands/statements.ts
@@ -43,9 +43,11 @@ import {
 } from '../data/validators'
 import {
   EUDR_AMEND_GUARDED_FIELDS,
+  canDeleteStatement,
   canTransition,
   evaluateSubmissionGate,
   isAmendWindowOpen,
+  isStatementReadOnly,
   type GateAssessmentView,
   type GateSubmissionView,
 } from '../lib/statement-lifecycle'
@@ -495,7 +497,7 @@ async function validateStatementUpdate(
   }
 
   const hasBaseChanges = hasStatementFieldChanges(record, parsed)
-  if (record.status === 'archived' && (hasBaseChanges || Object.keys(custom).length > 0)) {
+  if (isStatementReadOnly(record.status) && (hasBaseChanges || Object.keys(custom).length > 0)) {
     throw new CrudHttpError(400, { error: 'eudr.errors.archivedReadOnly' })
   }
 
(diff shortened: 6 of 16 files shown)


patzick:
🤖 `om-open-pr` — label rationale

- **`review`** — the PR is open and ready for review; it enters the pipeline in the review state.
- **`bug`** — this corrects broken behaviour reported in #5461 (an action offered that can never succeed, and a lost failure reason), not a new capability.
- **`needs-qa`** — the change is user-facing and touches UI-rendering files (`backend/eudr/statements/page.tsx`, `backend/eudr/statements/[id]/page.tsx`), so the automated-verification exemption does not apply. The behaviour to exercise manually: archive a statement, then confirm Delete is absent from both the detail page and the list row menu, and that a delete refused against a stale list row reports "Archived statements are read-only." rather than the generic copy.
- **`priority-medium`** — the trap is reachable in two clicks from the UI and leaves permanently un-actionable rows in a compliance register, but no data is lost or corrupted and a workaround (ignore the row) exists.
- **`risk-low`** — additive exports plus three call sites and a copy-only locale change; no API contract, schema, status-machine or error-token change, and the delete guard's behaviour is byte-for-byte what it was. Covered by unit, helper and integration tests.

patzick:
🤖 `om-open-pr` — chain lock transferred from #5461 onto this PR. Continuing with `om-auto-review-pr`; other auto-skills will skip this PR until the lock is released.

patzick:
🤖 `om-auto-review-pr` taking over the existing chain lock on this PR (inherited from `om-open-pr`). Continuing with review and autofix.

patzick:
> **Note:** posted as a comment because GitHub does not allow approving one's own pull request. The verdict below is the review verdict; a human approving review is still required by branch protection, so this PR stays in the `review` pipeline state rather than moving to `merge-queue`.

# 🔍 Code Review

## 🎯 Summary

This PR gives the terminal-`archived` rule for EUDR due-diligence statements a single owner and makes both delete surfaces obey it, closing #5461. `isStatementReadOnly` / `canDeleteStatement` land in `lib/statement-lifecycle.ts` beside `canTransition`; the update and delete command guards, `CrudForm`'s `deleteVisible` on the detail page, and the list row-action array all consume them, so hiding Delete and refusing Delete can no longer drift apart. The list page's delete rejection now routes through a new `translateEudrErrorMessage` helper that recovers the server's `eudr.errors.*` token instead of masking it with a constant, and the archive confirmation copy warns in all five locales that the step is irreversible.

The change is well scoped and conservative: no API contract, response shape, status machine, DB schema, or error token moves, and the delete guard's runtime behaviour is byte-for-byte what it was for every reachable status. The only wire-visible difference is that two UI surfaces stop offering an action the server always rejected.

One **major** finding was raised during this review and has been fixed in `c99d123c6` before submitting: the user-visible half of the fix — the two pages' wiring — shipped with no test coverage, in a module that already has an established jsdom page-test pattern. Details and the fix are in the Findings section below.

**Verdict: approve.**

## 🧪 Validation Gate

| Command | Result | Notes |
|---|---|---|
| `yarn build:packages` | ✅ pass | 25/25 tasks successful |
| `yarn generate` | ✅ pass | 340 generated artifacts refreshed; no generated file left dirty in the diff |
| `yarn build:packages` (second pass) | ✅ pass | cached, clean |
| `yarn i18n:check-sync` | ✅ pass | all 5 locales in sync across 55 modules — the `confirmArchiveBody` change updates an existing key in every locale, adding none |
| `yarn i18n:check-usage` | ✅ pass | advisory-only checker; 3797 pre-existing unused keys repo-wide, none introduced here |
| `yarn typecheck` | ✅ pass | 25/25 workspaces |
| `yarn test` | ⚠️ **not run in full** | The author scoped local testing to the changed area on instruction and deferred the repo-wide suite to CI. The affected slice was run in full: `yarn workspace @open-mercato/core test --testPathPattern "modules/eudr"` — **21 suites / 131 tests passed**. CI owns the rest. |
| `yarn build:app` | ⚠️ **not run** | Deferred to CI on the same instruction. |

Two gate commands were deliberately deferred to CI rather than run locally, so this review's evidence is partial by design and does not stand in for branch protection — the required checks still gate the merge. Nothing in the diff touches app-level routing or build configuration, and `yarn typecheck` covering `@open-mercato/app` passed, so the residual risk from the deferred `build:app` is low. At review time CI had `Analyze (actions)`, `Analyze (python)`, `audit-scope`, `scope` and `license/cla` green, with `prepare`, `lint`, `ds-lint` and `Analyze (javascript-typescript)` still pending. Note for whoever picks this up: because `prepare` had not finished, the `test` and integration jobs had not been scheduled yet — their absence from the check list is *pending*, not *skipped*, and the merge should not proceed until they report.

## 🟠 Major

### Behaviour change on two pages shipped without UI-layer test coverage — `packages/core/src/modules/eudr/backend/eudr/statements/page.tsx:187`, `packages/core/src/modules/eudr/backend/eudr/statements/[id]/page.tsx:648`

The original commit tested the shared predicate (`lib/__tests__/statement-lifecycle.test.ts`), the error-token helper (`components/__tests__/crudErrorI18n.test.ts`) and the API contract (`__integration__/TC-EUDR-007.spec.ts`), but nothing covered the actual user-visible behaviour the issue reported: that Delete disappears for archived statements and that a refused delete names its reason. Both are page-level wiring — `deleteVisible={canDeleteStatement(record.status)}` and the conditional `RowActions` item — and either could be dropped in a future edit with every existing test still green. The repo's own checklist requires behaviour changes to ship with coverage, and this module already has the pattern for it: `backend/eudr/evidence-submissions/__tests__/page.test.tsx` and `backend/eudr/evidence-submissions/[id]/__tests__/page.test.tsx` are jsdom page tests over the exact same surfaces, the latter using a `crudFormPropsCapture` mock to assert props handed to `CrudForm`.

**Fixed in `c99d123c6`.** Added `backend/eudr/statements/__tests__/page.test.tsx` (Delete present for `draft`; Edit and Duplicate present but Delete absent for `archived`; a refused delete flashes "Archived statements are read-only."; an untokened failure falls back to the generic copy) and `backend/eudr/statements/[id]/__tests__/page.test.tsx` (`deleteVisible` is `true` for `draft`, `false` for `archived`), both following the evidence-submissions pattern. Six new tests; the module suite is now 21 suites / 131 tests, all passing.

## 🔵 Minor

### An `available` statement past its amendment window still offers a Delete that will fail — `packages/core/src/modules/eudr/commands/statements.ts:851`

`canDeleteStatement` only encodes the `archived` rule, but the delete command has a second refusal path: an `available` statement outside the 72-hour amend window is rejected with `eudr.errors.amendWindowElapsed`, as is one with downstream references. The UI cannot cheaply predict either — the window is wall-clock-relative and the reference check needs a query — so hiding the action for those is out of reach here, and hiding it optimistically would be wrong. This is acceptable and arguably improved by this PR rather than worsened by it: because the list now surfaces the server's token, such an attempt reports "The amendment window has elapsed." (`i18n/en.json:80`) instead of the old generic copy. Worth a follow-up only if the amend-window state is ever exposed on the list payload.

## 💥 Breaking Changes

- [x] **Exported APIs** — `isStatementReadOnly`, `canDeleteStatement` and `translateEudrErrorMessage` are new exports; nothing was removed, renamed or retyped. `formConfig.tsx` extends its existing re-export line rather than replacing it.
- [x] **HTTP routes and response shapes** — untouched. `PUT`/`DELETE /api/eudr/statements` keep their status codes and bodies; `eudr.errors.archivedReadOnly` is emitted from exactly the same conditions.
- [x] **Event names** — no events touched.
- [x] **CLI flags** — none touched.
- [x] **DB schema** — no entity, migration or snapshot change; none required.
- [x] **Config formats** — none touched.
- [x] **`BACKWARD_COMPATIBILITY.md` protected surfaces** — verified against the 13 categories. The touched surfaces (public types, import paths, API routes) are all additive-only changes, so no deprecation window is owed.

One behavioural detail worth stating explicitly, since it is the only place the refactor could have changed semantics: the delete guard moved from `record.status === 'archived'` to `isStatementStatus(record.status) && !canDeleteStatement(record.status)`. For every value in `EUDR_STATEMENT_STATUSES` the two are identical, and for an unrecognised status string both fall through without throwing — so the narrowing guard preserves the old behaviour rather than tightening it.

## 🧪 Test Coverage

- [x] **Regression test that fails without the fix** — `statement-lifecycle.test.ts` pins `archived` as the only non-deletable status and asserts the delete predicate agrees with the read-only predicate the update guard uses, so the two guards cannot diverge silently. It also re-pins `archived` as terminal for every status, which is what makes a blocked delete a genuine dead end rather than a recoverable state.
- [x] **Error-surfacing coverage** — `crudErrorI18n.test.ts` covers the token arriving on `error` (how `apiCall` rejections carry it) and on `message`, plus the three fallback paths: an untokened `Error`, a bare object, `null`, and a well-formed token with no translation. That last case matters — without it a missing locale entry would leak a raw `eudr.errors.*` key into the toast.
- [x] **UI-layer coverage** — added in `c99d123c6` per the Major finding above; both pages' delete affordance and the list's message resolution are now pinned.
- [x] **Integration coverage on the risky path** — `TC-EUDR-007` already asserted the archived `PUT` and the invalid-transition rejections; it now also asserts `DELETE /api/eudr/statements` → 400 `archivedReadOnly`, so the wire contract the issue documents is locked in and not merely implied by the unit test.
- [x] **Self-contained fixtures** — the integration assertion reuses the existing test's fixtures and teardown; `deleteByCrudPath` already tolerates the archived row it cannot remove.

## 📌 Scope note

The issue explicitly leaves out whether archived statements should be un-archivable or hard-deletable at all, and this PR correctly respects that boundary — it removes the misleading affordance without deciding the product question. The three orphaned rows the reporter created remain unremovable; if cleaning those up matters, that needs its own issue and a deliberate decision about an un-archive path.

patzick:

This PR is approved by review and requires manual QA (`needs-qa`, no `skip-qa`). It stays in the `review` pipeline state rather than `merge-queue` because GitHub refused the automated approving review (own PR) and branch protection still needs a human approval — but the QA below can be run now, in parallel. The QA-approval gate holds the merge until `qa-approved` is added regardless. QA reviewer: when you pick it up, swap the pipeline label to `qa`, then run the routes below.

### P1 — Archived statement: the Delete affordance is gone

**Where to click**
- `/backend/eudr/statements/create` → create a statement (title + commodity); it is created as `draft`
- `/backend/eudr/statements/{id}` → the statement detail page
- `/backend/eudr/statements` → the list, row menu **⋯**

**What to verify**
- On the fresh `draft` statement, the detail page still shows **Delete**, and the list row menu still shows **Delete** alongside Edit and Duplicate — this PR must not remove the action from statements that can legitimately be deleted.
- Click **Archive** and confirm. The confirmation dialog body now reads *"The statement will be archived. Archived statements are read-only and cannot be edited, restored, or deleted."* — check the wording is present and not truncated in the dialog.
- After archiving, **Delete disappears from the detail page** (Submit and Archive already disappeared before this PR — Delete should now behave the same way).
- Open the list and reload it. The archived row's **⋯** menu shows **Edit** and **Duplicate** but **no Delete**.
- Deleting a `draft` statement still works end to end: confirm dialog → row disappears → success flash *"Statement deleted."*

**What can go wrong**
- Delete hidden too broadly — a `submitted`, `available` or `withdrawn` statement losing its Delete action would be a regression; only `archived` should lose it.
- The row menu collapsing or misrendering with two items instead of three (the item array is now built conditionally).
- Row-click navigation breaking: clicking an archived row should still open its detail page (`rowClickActionIds` still targets `edit`).

### P1 — A refused delete now names the reason

**Where to click**
- `/backend/eudr/statements` in two browser tabs

**What to verify**
- In tab A, archive a `draft` statement. In tab B — whose list still shows the row as `draft`, so Delete is still offered — click **⋯ → Delete** and confirm. The error flash must read **"Archived statements are read-only."**, not the generic *"Could not delete the statement."*, and the list must refresh so the row's status badge updates to `Archived` and its Delete action disappears.
- Editing an archived statement's Notes and clicking **Save** still reports *"Archived statements are read-only."* on the detail page — unchanged behaviour, confirm it did not regress.
- Provoke a non-EUDR failure if you can (e.g. delete with the network throttled/offline in devtools): the flash must fall back to the generic *"Could not delete the statement."* rather than showing a raw `eudr.errors.*` key or an empty toast.

**What can go wrong**
- A raw translation key (`eudr.errors.archivedReadOnly`) leaking into the toast instead of the translated sentence — check a non-English locale too (`pl`, `de`, `es`, `ko` all have the key).
- The extra list refresh on failure causing a visible flicker, a lost scroll position, or a double-fetch loop.
- An optimistic-lock 409 being swallowed by the new message path: edit the same statement in two tabs and save the stale one — the conflict must still surface on the persistent conflict bar, not as a plain toast.

### P1 — Perceived performance on the changed routes

**Where to click**
- `/backend/eudr/statements` (list) and `/backend/eudr/statements/{id}` (detail), cold-loaded

**What to verify**
- Cold load both routes and capture screenshots. The loading shell appears promptly and the table/form render without a visible stall; row-menu open/close stays responsive with a full page of rows.
- Check both at a mobile viewport — the row menu with the Delete item omitted should not leave a misaligned or empty menu.

**What can go wrong**
- The list re-rendering row actions more than needed now that the items array is rebuilt per row.

### Data scoping / permissions
- Confirm with a view-only EUDR grant that no delete affordance appears anywhere, and with a wildcard `eudr.*` grant that behaviour matches the above. Nothing in this PR changes ACL wiring, so this is a sanity check rather than a target.
- Confirm an archived statement in one organization is not visible or actionable from another (unchanged scoping, worth one pass since the delete path was touched).

### Pass/fail
- All routes pass → add `qa-approved` (this clears the QA-approval gate); the PR still needs a human approving review before merge.
- Anything fails → apply `qa-failed`, comment with the route, the steps, and a screenshot.

patzick:
🏷️ label rationale

Labels unchanged by this review pass: `review`, `bug`, `needs-qa`, `priority-medium`, `risk-low`, plus the chain's `in-progress`.

The review verdict is **approve**, which would normally move the PR to `merge-queue`. It stays in `review` instead because GitHub refused the automated approving review (a PR cannot be approved by its own author), so branch protection still needs a human approving review — `merge-queue` would misrepresent that as satisfied. `needs-qa` is retained: the change is user-facing and the QA-approval gate must clear before merge. Priority and risk are unchanged from the values `om-open-pr` set and its rationale comment explains.

Lock retained — chain continues.

patzick:
🤖 `om-auto-qa-pr` taking over the existing chain lock on this PR (inherited from `om-auto-review-pr`). Running UI verification in evidence-only mode.

patzick:
## 📸 UI verification — `om-auto-qa-pr` (evidence-only)

**Verdict: ⚠️ BLOCKED — not verified.** No UI step was exercised and no result is claimed for this PR's browser behaviour.

### Why the environment could not boot

`om-prepare-test-env` had nothing to provision against on this machine:

- no container runtime — `docker` is not on `PATH`, so the ephemeral Postgres the test env normally provisions could not start;
- no reachable database — `localhost:5432` is closed;
- no `.env` in this worktree or the parent checkout, so no existing instance could be pointed at;
- no Playwright browser binaries installed (`~/Library/Caches/ms-playwright` does not exist), so the configured `playwright` provider had nothing to drive.

Rather than fabricate a pass, this run stops here. **`needs-qa` stays on the PR and no QA verdict label was applied** — a human QA reviewer (or a later `om-auto-qa-pr` run on a machine with a test environment) still owes this PR the evidence. The manual QA instructions posted earlier on this PR are the script to follow; they were written from the diff and remain valid.

### What still needs a human at the browser

The change is user-facing on two routes, so the automated tests in this PR do not substitute for a click-through:

- `/backend/eudr/statements/{id}` — **Delete** must disappear from the detail page once a statement is archived, and remain present for `draft` / `submitted` / `available` / `withdrawn`.
- `/backend/eudr/statements` — the row **⋯** menu must drop **Delete** for archived rows while keeping **Edit** and **Duplicate**, and a refused delete must flash *"Archived statements are read-only."* rather than the generic copy.
- The archive confirmation dialog must render the new irreversibility warning without truncation, in each of the five locales.

### Automated coverage that *was* verified

Not a substitute for the browser pass, but the behaviour is pinned below the UI:

- `backend/eudr/statements/__tests__/page.test.tsx` and `backend/eudr/statements/[id]/__tests__/page.test.tsx` — jsdom tests asserting the row-action omission, the `deleteVisible` prop, and both message paths (reason surfaced / generic fallback).
- `__integration__/TC-EUDR-007.spec.ts` — `DELETE /api/eudr/statements` on an archived statement returns 400 `archivedReadOnly` at the wire level.
- Full module suite: 21 suites / 131 tests passing.

### 🧪 Follow-up: no browser-level UI test ships with this change

The PR adds jsdom component tests and an API-level Playwright assertion, but no browser-level test drives the actual pages. A follow-up run of `om-integration-tests` could add one, in `packages/core/src/modules/eudr/__integration__/`, alongside the existing `TC-EUDR-*` specs:

1. Sign in as `admin`, `POST /api/eudr/statements` to create a `draft` fixture, capturing its id.
2. Visit `/backend/eudr/statements/{id}` and assert the Delete control is present.
3. Archive via `PUT /api/eudr/statements` with `status: 'archived'`, reload the page, and assert the Delete control is gone while the form still renders.
4. Visit `/backend/eudr/statements`, open the archived row's **⋯** menu, and assert it offers Edit and Duplicate but not Delete.
5. Create a second `draft` fixture, load the list, archive it out-of-band via the API, then drive the still-visible Delete on the stale row and assert the flash reads the translated `eudr.errors.archivedReadOnly` copy.
6. Teardown: delete every non-archived fixture; the archived ones cannot be removed by design, so scope them with a unique run stamp in the title as the sibling specs do.

patzick:
🤖 `om-auto-qa-pr` completed: BLOCKED — UI verification could not run (no test environment on this machine). No labels were changed; `needs-qa` stands. Lock retained — chain continues.

patzick:
🤖 `om-auto-fix-issue` run complete: fix implemented, reviewed (verdict **approve**, one major finding autofixed in `c99d123c6`), UI verification blocked by a missing local test environment. The chain lock is released and swapped for `ci-monitoring` while the CI result is followed up. Still owed before merge: a human approving review (GitHub refuses self-approval) and manual QA evidence to clear `needs-qa`. Lock released.

patzick:

**All checks settled green on `c99d123c6`.** This closes out the two gate commands that were deferred to CI rather than run locally, so the validation gate in the review above is now complete rather than partial.

| Check | Result | Notes |
|---|---|---|
| `test` | ✅ pass (6m18s) | The repo-wide unit suite deferred from the local run. |
| `ephemeral-integration (none)` | ✅ pass (7m23s) | 28 passed, 0 failed, 0 skipped. |
| `docker-build` | ✅ pass (3m40s) | Covers the deferred `build:app` surface. |
| `prepare` | ✅ pass (3m4s) | |
| `lint` / `ds-lint` | ✅ pass | |
| `Analyze (actions / javascript-typescript / python)` | ✅ pass | |
| `audit-scope` / `scope` / `license/cla` | ✅ pass | |
| `audit` | ⏭️ skipping | See the note below. |
| `CodeQL`, `comment`, `mutate`, `merge-coverage`, `[code]smith` | ⏭️ skipping | Not scheduled for this change. |

**Two things worth stating explicitly rather than leaving to be inferred from a green tick:**

1. **`audit` is skipped, but the test gate still reported.** A skipped `audit` in this repo can leave `test` skipped downstream, which shows up as absence rather than failure and reads like "not applicable". That did not happen here — `test` ran for 6m18s and `ephemeral-integration` reports `skipped: 0`, so this is a real signal, not a silent gap.

2. **The integration assertion added by this PR actually executed.** `TC-EUDR-007 › keeps archived statements read-only and exposes latest risk on statement lists` passed in 470ms against a live database, which is the test extended with the `DELETE /api/eudr/statements` → 400 `archivedReadOnly` case. The new wire-level coverage is genuinely exercised and not merely present in the diff.

**One pre-existing flake, unrelated to this change:** `TC-EUDR-010 › Searchable product picker` timed out at 20.1s on its first attempt and passed on retry #1 (3.3s). It exercises product-mapping search and shares no code with the statement delete path this PR touches, so it is not a consequence of these changes — but it is a real flake in the EUDR integration suite and would be worth its own issue if it recurs.

No further CI follow-up will come from this agent; `ci-monitoring` has been removed.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-08-22T15:33:36Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron:
@adeptofvoltron PR against `main` branch on purpose

adeptofvoltron (approved):
# 🔍 Code Review

## 🎯 Summary

This PR closes #5461 by giving the terminal-`archived` rule for EUDR due-diligence statements a single owner and making both delete affordances obey it. `isStatementReadOnly` / `canDeleteStatement` land in `lib/statement-lifecycle.ts` next to `canTransition`; the update guard and the delete guard in `commands/statements.ts` now call those predicates instead of comparing `status === 'archived'` inline; the detail page passes `deleteVisible={canDeleteStatement(record.status)}` to `CrudForm` and the list builds its `RowActions` array conditionally, so Delete disappears for archived rows exactly as Submit and Archive already do. A new `translateEudrErrorMessage` helper recovers the server's `eudr.errors.*` token from a rejection so a refused delete on the list names the reason instead of flashing constant copy, and the archive confirmation body now warns in all five locales that archiving is irreversible.

The shape of the fix follows the module's own established pattern — `StatementLifecycleBar` already derives Submit/Archive visibility from `status`, and this change extends that idea to Delete rather than inventing a new mechanism. The server guard is unchanged in behavior, so the UI is being brought into agreement with a rule the API already enforced rather than the rule being relaxed anywhere. I verified that independently: `deleteStatementCommand.execute` still throws `CrudHttpError(400, { error: 'eudr.errors.archivedReadOnly' })` for archived records, and the rewritten condition `isStatementStatus(record.status) && !canDeleteStatement(record.status)` is behaviorally identical to the `record.status === 'archived'` it replaces, because `isStatementStatus` is a type guard over `EUDR_STATEMENT_STATUSES` and `archived` is a member of it.

I also checked the parts of the mechanism the tests can only assert against mocks. `useGuardedMutation.runMutation` rethrows the original rejection (`packages/ui/src/backend/injection/useGuardedMutation.ts:95`), so the `{ status, ...result }` object the list's operation throws does reach the page's `catch` with its `error` property intact — the new helper's primary token path is real, not just mocked. On the detail page, `deleteCrud` funnels failures through `raiseCrudError`, which builds the error with the server token on both `.message` and `.error`, so the untouched `translateEudrCrudError` path there is genuinely correct as the PR claims. And `I18nProvider`'s translator resolves a missing key to the key itself (`dict[key] ?? fallback ?? key`), which is exactly what `translateEudrErrorMessage`'s `translated === token` check relies on — so the "fall back rather than show a raw key" guarantee holds in production, not only under the test's `messages[key] ?? key` stub.

Finally, I confirmed there is no third delete surface to update: the statements list has no bulk actions, and every other delete path in the module belongs to a different entity. The API-level contract is covered by the integration test this PR extends.

## ✅ Verdict

**Approved.** No blocker and no major findings. The four items below are all nits — none of them changes behavior, and none of them should hold up the merge. Under this repository's stricter verdict rule a Medium finding would request changes; I considered promoting findings 2 and 4 to Medium and concluded neither meets that bar, for the reasons given inline.

Note on the merge gate: this approval clears the code-review requirement only. `needs-qa` is still unearned — the earlier `om-auto-qa-pr` pass on this PR reported **BLOCKED — not verified** because no test environment could be provisioned on that machine, so no UI step has actually been exercised. With `qaGate` on, the merge stays blocked until a QA reviewer adds `qa-approved`. Manual QA instructions follow in a separate comment.

## 🧪 Validation Gate

**Validation source: GitHub checks.** Per this repository's `om-auto-review-pr` extension, CI check results are preferred over re-running `validation.commands` locally. All checks reported against the PR head `c99d123c6` are settled and none failed, so no local fallback was needed and none was run.

| Check | Result | Covers |
|---|---|---|
| `test` | ✅ success | The repo-wide unit suite, including the four test files this PR adds or extends |
| `ephemeral-integration (none)` | ✅ success | `TC-EUDR-007`, including the new archived-`DELETE` wire assertion |
| `docker-build` | ✅ success | The `build:app` surface |
| `prepare` | ✅ success | Generation and package build |
| `lint` | ✅ success | ESLint across the workspace |
| `ds-lint` | ✅ success | Design-system token compliance |
| `audit-scope`, `scope` | ✅ success | Scope guards |
| `Analyze (javascript-typescript / python / actions)` | ✅ success | CodeQL analysis jobs |
| `CodeQL` | ➖ neutral | Aggregate status, not a failure |
| `merge-coverage`, `audit`, `comment`, `mutate`, `[code]smith` | ⏭️ skipped | Not applicable to this PR |

No check is pending, so there is nothing outstanding to disclose and nothing for a CI follow-up to wait on.

## 🔵 Findings — Low (nits)

**1. `translateEudrErrorMessage` re-implements the token resolution that `translateEudrErrorToken` already performs** — `packages/core/src/modules/eudr/components/crudErrorI18n.ts:38`

The new function tests the token against `EUDR_ERROR_KEY_PATTERN`, calls `translate`, and compares the result against the input to detect an untranslated key. `translateEudrErrorToken` at line 11 of the same file does all three already; the only difference is what each returns when translation fails — the older helper passes the raw token through, the new one substitutes the caller's fallback. That difference is one line to bridge:

```ts
const translated = translateEudrErrorToken(token, translate)
return translated === token ? fallbackMessage : translated
```

Purely a DRY cleanup with no behavioral consequence — the two copies agree today and are both covered by tests, so this is a readability preference rather than a defect.

**2. The list's delete path hand-assembles its error object instead of going through `deleteCrud` / `raiseCrudError`** — `packages/core/src/modules/eudr/backend/eudr/statements/page.tsx:171`

`handleDelete` calls `apiCall` directly and then throws `Object.assign(new Error('[internal] …'), { status, ...result })`. That is precisely why the token lands on `.error` rather than `.message`, and therefore why a second translation helper was needed at all: the detail page, which uses `deleteCrud`, gets an error from `raiseCrudError` carrying the token on `.message` and is already served by the existing `translateEudrCrudError`. Repo checklist §4 asks CRUD writes to use `createCrud`/`updateCrud`/`deleteCrud`. Since `withScopedApiRequestHeaders` applies the optimistic-lock header to any `apiCall` made inside its callback, `deleteCrud` would drop in here and both surfaces would then share one error-translation path.

I deliberately did not rate this Medium: the hand-rolled call predates this PR, the PR does not make it worse, and the added helper is well-tested and correct as written. It is a Boy Scout opportunity in a function this PR already touches, not a convention violation this PR introduces.

**3. The rewritten delete guard reads as a double negative** — `packages/core/src/modules/eudr/commands/statements.ts:849`

`if (isStatementStatus(record.status) && !canDeleteStatement(record.status))` requires the reader to hold two negations at once, and — exactly like the `record.status === 'archived'` it replaces — it falls through and permits deletion when `record.status` is not a recognized status at all. The behavior is unchanged so there is nothing to fix, but the update guard four hundred lines earlier reads better because it states the condition positively (`isStatementReadOnly(record.status)`). A narrowing guard with an explicit throw for the unrecognized case, mirroring line 489, would make the delete path read the same way.

**4. `available` statements past their amend window still offer a Delete the server will refuse** — `packages/core/src/modules/eudr/backend/eudr/statements/[id]/page.tsx:648`

`canDeleteStatement` encodes only the status rule, but `deleteStatementCommand` enforces two more: an `available` statement outside the 72-hour amend window is rejected with `eudr.errors.amendWindowElapsed`, and one with downstream references is rejected by `assertNoDownstreamReferences`. Since time only moves forward, the first of those is as permanent as the archived case, and the detail page already holds `record.referenceIssuedAt`, so `isAmendWindowOpen` could hide the affordance there too.

I considered rating this Medium and decided against it, because the split this PR draws is principled rather than an oversight. `archived` is an unconditional, client-knowable refusal, so hiding the affordance is right. The amend-window and downstream-reference refusals are conditional on server state the list cannot see, and hiding Delete on a client-side guess would risk suppressing a delete that would actually have succeeded — the correct treatment for a conditional refusal is to make it legible, which is exactly what this PR's `translateEudrErrorMessage` now does (`eudr.errors.amendWindowElapsed` has translations in all five locales and will now reach the flash instead of the generic copy). Worth a follow-up issue for the detail page, not a change to this one.

## 💥 Breaking Changes

No contract surface is broken; I checked each category from `BACKWARD_COMPATIBILITY.md` that this diff could plausibly touch.

- **Public exports and import paths** — `isStatementReadOnly` and `canDeleteStatement` are new named exports from `lib/statement-lifecycle.ts`, and `translateEudrErrorMessage` is a new export from `components/crudErrorI18n.ts` re-exported through `components/formConfig.tsx`. All three are purely additive; nothing was removed, renamed, or had its signature changed.
- **API routes and response shapes** — unchanged. `DELETE /api/eudr/statements` returns the same 400 with the same `eudr.errors.archivedReadOnly` token for the same input; the rewritten guard is a refactor of the condition, not of the outcome.
- **Database schema** — untouched. No entity, migration, or snapshot changes, and none are needed.
- **Event IDs, widget spot IDs, DI keys, ACL features, notification IDs, CLI commands** — none appear in the diff.
- **i18n keys** — `eudr.lifecycle.confirmArchiveBody` changes its *value* in all five locale files; the key itself is unchanged and no key was added or removed, so `i18n:check-sync` parity is preserved (and `lint`/`prepare` are green). I read all five new strings and the German, Spanish, Korean and Polish copy all faithfully render the English meaning.

The only wire-visible difference is that two UI surfaces stop offering an action the server has always rejected — a strict reduction in reachable failure, not a contract change.

## 🧪 Test Coverage

Coverage is genuinely good here, and unusually well-targeted at the specific claims the PR makes.

- `lib/__tests__/statement-lifecycle.test.ts` iterates `EUDR_STATEMENT_STATUSES` rather than hardcoding a list, so adding a status to the enum without deciding its delete semantics will surface in the test rather than pass silently. It also pins the invariant that makes the whole fix sound — `archived` is terminal for every target status, so a blocked delete has no un-archive escape hatch.
- `components/__tests__/crudErrorI18n.test.ts` covers both token-carrying paths (`.error` and `.message`), three shapes of untokened rejection (`Error`, bare object, `null`), and the case that matters most for UX — a well-formed token with no translation falls back to the generic copy instead of leaking a raw i18n key at the user.
- `backend/eudr/statements/__tests__/page.test.tsx` asserts the affordance through rendered output rather than by inspecting props, and does not mock `lib/statement-lifecycle`, so the real predicate drives the assertion. Its rejection test does mock `runMutation`, which is why I verified the real `runMutation` rethrows unwrapped rather than taking the mock's word for it.
- `backend/eudr/statements/[id]/__tests__/page.test.tsx` captures the `CrudForm` props and asserts `deleteVisible` directly. That is a reasonable seam given `CrudForm` is mocked, and `CrudForm`'s own render test already covers `deleteVisible` driving `showDelete`.
- `__integration__/TC-EUDR-007.spec.ts` adds the wire-level assertion that `DELETE /api/eudr/statements` returns `archivedReadOnly` for an archived statement, so the server rule the UI now mirrors is locked down at the contract level and not only in unit tests.

What tests cannot cover, and QA therefore should: that the row action and the form button actually disappear in a real browser against real data. That is the gap the manual QA comment addresses.

adeptofvoltron:

- 🚀 `merge-queue` — the code review verdict on head `c99d123c6` is approve with no blocker and no major findings, and every settled CI check on that SHA passed, so the review gate is cleared and the PR moves out of `review`; `needs-qa` stays on, so the QA-approval gate still holds the actual merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — retained. The change alters user-facing behavior on two backend screens (`backend/eudr/statements/page.tsx` and `backend/eudr/statements/[id]/page.tsx`), so the automated-verification exemption does not apply, and no QA evidence exists yet: the earlier `om-auto-qa-pr` pass on this PR reported BLOCKED because no test environment could be provisioned, so nothing has been clicked through in a browser.
- 🐛 `bug` — unchanged. This corrects broken behavior reported in #5461 — an action offered that can never succeed, plus a lost failure reason — rather than adding a capability.
- 🔹 `priority-medium` — unchanged, and the review confirms the rating. This is an ordinary user-facing bug fix with a real but non-urgent impact: users are led into a delete that always fails, which is confusing rather than damaging, and no data is at risk.
- 🟢 `risk-low` — unchanged, and the review confirms the rating. The blast radius is one module: no schema, migration, auth, data-scoping, or money surface is touched, the server-side guard is refactored without any behavior change, the only public-API movement is three additive exports, and the change ships with unit tests on both pages plus a wire-level integration assertion.

Previous rationale comment on this PR was posted by the earlier `om-open-pr` / `om-auto-review-pr` chain running under a different account; this comment supersedes it and describes the current label state in full.

adeptofvoltron:

This PR is approved by code review and requires manual QA (`needs-qa`, no `skip-qa`). It is queued in `merge-queue` but the QA-approval gate holds it until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` by swapping the labels (remove `merge-queue`, add `qa`), then run the routes below.

Nothing on this PR has been exercised in a browser yet — the earlier `om-auto-qa-pr` pass reported BLOCKED because it had no environment to provision — so this is a first pass, not a re-check. The instructions below restate and extend the QA scope from the earlier chain comment and are derived from this PR's diff only.

### P1 — Archived statement: the Delete affordance is gone

**Where to click**
- `/backend/eudr/statements` — the statements list, row `⋯` menu
- `/backend/eudr/statements/{id}` — the statement detail page

**What to verify**
- Create a draft statement, open its detail page, and confirm the `⋯` / form footer still offers **Delete** — the fix must not remove the affordance from statuses that can still be deleted.
- On the list, open the `⋯` menu for that same draft row and confirm **Edit**, **Duplicate** and **Delete** are all present.
- Archive the statement from the detail page's lifecycle bar. The confirmation dialog body must now read that archived statements are read-only and cannot be edited, restored, or deleted — this copy changed in this PR and is the user's only warning before an irreversible step.
- After archiving, reload the detail page and confirm **Delete** is no longer rendered (Submit and Archive were already hidden — Delete should now behave the same way).
- Return to `/backend/eudr/statements`, open the `⋯` menu on the archived row, and confirm **Edit** and **Duplicate** are still there but **Delete** is gone.
- Click the archived row itself and confirm row-click still navigates to the detail page — the row actions array was restructured, so navigation is worth a glance.

**What can go wrong**
- Delete disappearing for non-archived statuses too (draft / submitted / available / withdrawn) — that would be an over-broad guard and is the main regression risk.
- The `⋯` menu rendering empty, mis-ordered, or throwing for archived rows now that the items array is built conditionally.
- Row click or the Edit action breaking on archived rows.
- The archive confirmation dialog showing a raw i18n key instead of the new sentence.

### P1 — A refused delete names its reason instead of generic copy

**Where to click**
- `/backend/eudr/statements` — two browser tabs on the same list

**What to verify**
- Open the list in tab A and tab B. In tab B, archive a statement that is still shown as deletable in tab A. Back in tab A (do not reload), open that row's `⋯` menu and choose **Delete**.
- The error flash must read the specific reason — "Archived statements are read-only." — and not the old generic "Could not delete the statement."
- The list must then refresh on its own, so the stale row re-renders as archived and its **Delete** action disappears without a manual reload. Retrying the same delete should no longer be possible.
- Provoke a delete failure with no reason attached (for example, take the app offline and retry a delete). The flash must fall back to the generic "Could not delete the statement." — it must never display a raw key such as `eudr.errors.archivedReadOnly`.
- If you have an `available` statement whose 72-hour amend window has elapsed, delete it and confirm the flash names the amend-window reason rather than the generic copy. Note that Delete is still deliberately offered in this case — that is expected scope for this PR, since the refusal depends on server state the list cannot see.

**What can go wrong**
- The reason not surviving to the flash, leaving the old generic copy — this is the half of the fix that depends on the error object keeping its `error` property through the mutation wrapper.
- A raw `eudr.errors.*` key rendered to the user for a token with no translation.
- The post-failure refresh not firing, so the stale row keeps offering a delete that cannot succeed.

### P2 — Locale coverage of the new archive warning

**Where to click**
- `/backend/eudr/statements/{id}` with the UI language set to German, Spanish, Korean and Polish in turn

**What to verify**
- The archive confirmation body renders the translated warning in each locale, with no raw key and no untranslated English sentence mixed in.

**What can go wrong**
- A locale falling through to the key or to English, which would silently drop the irreversibility warning for those users.

### Perceived performance

- Cold-load `/backend/eudr/statements` and confirm the list's first useful state (loading skeleton, then rows) is unchanged — the list page gained a new module import in this PR, so a regression in initial render would show here. A screenshot of the cold load is useful evidence.
- Open the `⋯` menu on a few rows and confirm it opens without perceptible delay now that the items array is rebuilt per row.
- Check the list and detail page at a mobile viewport and confirm the row-action menu and form footer still lay out correctly with one fewer item.

### Pass/fail

adeptofvoltron:

Reviewed head `c99d123c6` as a fresh review (no prior review from this reviewer on the PR). Verdict: **approve** — no blocker and no major findings; four Low nits are listed in the review body and none of them blocks the merge. Validation source: **GitHub checks** — all settled checks on that SHA passed, nothing is pending, so no local gate fallback was run and no CI follow-up is owed (`ci-monitoring` not applied).

Labels: pipeline moved `review` → `merge-queue`; `needs-qa`, `bug`, `priority-medium` and `risk-low` retained and confirmed by the review. Manual QA instructions posted separately.

autofix: skipped (not my PR — re-run with --autofix to fix it here). No inherited reviewer feedback was outstanding: the earlier review comment on this PR is this collection's own prior pass and its verdict was also approve.

Still owed before merge: manual QA evidence to clear `needs-qa` (the earlier `om-auto-qa-pr` pass was BLOCKED with no test environment, so nothing has been exercised in a browser yet). With `qaGate` on, the QA-approval gate holds the merge until `qa-approved` is added.

Issue #5477, opened 2026-08-22
Give the sales order lines validation error its own distinct string


wojciechszyjka:
## Follow-up from #5188

PR #5188 fixes [#5126](https://github.com/open-mercato/open-mercato/issues/5126) — the sales document form rendered the `lines` validation message twice (the custom component and `CrudForm`'s field wrapper both owned the node). The fix makes the component the single owner via `rendersOwnError: true`, which takes the count from two down to one.

What it does **not** fix, and what this issue tracks: the string itself is not distinct from copy that is already on screen before the user does anything wrong. `sales.orders.linesRequired` — "Add at least one line item before creating the order." — is rendered as static help copy twice in `SalesOrderDraftLines.tsx` (the muted description under the "Line Items" heading, and the `EmptyState` description), so the validation error is a third instance of the same sentence.

### The reviewer's request

From @adeptofvoltron's review on #5188 ([permalink](https://github.com/open-mercato/open-mercato/pull/5188#pullrequestreview-4941948274)):

> ### 2. The same sentence still renders three times on a failed submit
>
> **`packages/core/src/modules/sales/components/documents/SalesOrderDraftLines.tsx:155-157` and `:174`**
>
> `sales.orders.linesRequired` — "Add at least one line item before creating the order." — is already static help copy twice: once as the muted description under the "Line Items" heading, once as the empty-state description. After this change the validation error renders that same sentence a third time, so the error is textually indistinguishable from copy that was already on screen before the user did anything wrong. This predates the PR (which reduces the count from four to three) and is correctly out of scope; the self-review already flags it. Noting it so it survives into a follow-up: an imperative, distinct error string would make the failure legible.

### Why it was deferred rather than fixed in #5188

It is a copy/UX change, not a correctness one, and it needs a new translation key across every shipped locale — widening a focused three-file bug fix into an i18n change. It is genuinely out of the scope the review agreed on, so it is tracked here instead of being folded in.

### Acceptance criteria

- [ ] Introduce a distinct, imperative validation string (e.g. `sales.orders.linesRequiredError`) instead of reusing `sales.orders.linesRequired` for the submit-time error.
- [ ] Keep the existing help/empty-state copy as guidance; only the error node changes.
- [ ] Add the new key to every locale file the sales module ships, and keep `yarn i18n:check-sync` / `yarn i18n:check-usage` green.
[...]

PR #5481, opened 2026-08-22
fix(ui,sales,rules,record-locks): clear 0.7.0 pre-release UX and i18n nits

### LINKED ISSUE #5456: bug: [sales, rules, settings] — Bundle of small, unrelated UX/cosmetic inconsistencies found during 0.7.0 pre-release sweep
## Summary

Five small, independent, low-risk findings from the 0.7.0 pre-release QA sweep, bundled into
one ticket since each is a one-line-ish fix with no shared root cause — intended to be resolved
together in a single PR rather than filed as five separate tickets.

---

### 1. English breadcrumb/sidebar link on an otherwise-fully-Polish page

**Page URL:** `/backend/rules/create`

**Steps to reproduce:**
1. Log in as `admin@acme.com`, interface language: Polish.
2. Open the link above (or: Reguły biznesowe → sidebar link).
3. Look at: the sidebar link under "Reguły biznesowe", the breadcrumb at the top of the page,
   and the browser tab title.

**Expected:** All three read in Polish ("Utwórz Regułę Biznesową"), consistent with the rest of
the page (heading, fields, buttons, validation — already correctly Polish).

**Actual:** The sidebar link, breadcrumb, and browser tab title all stay in English ("Create
Business Rule") while the rest of the page is correctly Polish. Title/breadcrumb metadata is
likely sourced from separate route metadata, not the same i18n keys as the form body.

---

### 2. Invalid nested `<p>` in the record-locks settings help text (hydration warning)

**Page URL:** `/backend/settings/record-locks` (**local only** — this page currently 404s on
demo.openmercato.com, see `demo-environment-findings-for-patryk.md` — retest there once demo is
back in sync with `main`)

**Steps to reproduce:**
1. Log in as `admin@acme.com`.
2. Open DevTools → Console.
3. Open the link above.
4. Check the console immediately on page load.

**Expected:** Zero console errors.

**Actual:** Two React hydration errors: "In HTML, `<p>` cannot be a descendant of `<p>`" and
"`<p>` cannot contain a nested `<p>`". Source: the help tooltip next to the "Strategia" field
(info bubble describing "Optymistyczna:"/"Pesymistyczna:") has a `<p>` nested inside a `<p>`.
Cosmetic/markup only — the page still functions correctly.

---

### 3. "Status" search-select on the order-line dialog renders pre-expanded, unlike its siblings

**Page URL:** `/backend/sales/documents/create` (item dialog via "Dodaj pozycję"), and the
equivalent "Edytuj pozycję" dialog on any existing order.

**Steps to reproduce:**
1. Log in as `admin@acme.com`.
2. Go to "Utwórz dokument sprzedaży" → type "Zamówienie" → any customer → "Dodaj pozycję".
3. Scroll to the "Status" field, below "Ilość po normalizacji", without clicking it.

**Expected:** "Status" behaves like the other search-select fields in the same dialog (Produkt,
Wariant, Cena, Klasa podatkowa, Jednostka): collapsed by default with a "Zacznij pisać, aby
wyszukać." hint, expanding only when clicked.

**Actual:** "Status" is rendered already expanded — the full option list (Picking, Canceled,
Backordered, Pending, Packed, Allocated, Delivered, Shipped, Returned) is visible immediately,
with no interaction, pushing the rest of the dialog (including "Nazwa" and the submit button)
further down. Reproduced on both the create-order item dialog and the edit-existing-line dialog.

**Note:** this is arguably a design/consistency call rather than a clear-cut defect — but since
every sibling field in the same dialog is collapsed-by-default, the inconsistency itself is
worth fixing regardless of which behavior is "correct."

---

### 4. Order status values render in English despite a fully Polish interface

**Page URL:** any sales order detail page, e.g. `/backend/sales/documents/<id>?kind=order`

**Steps to reproduce:**
1. Log in as `admin@acme.com`, interface language: Polish.
2. Open any order whose status is `pending_approval` or `approved` (e.g. after going through
   the order-approval flow, #4877).
3. Look at the "Status" field value on the order detail page.

**Expected:** Status values render in Polish, matching the rest of the interface (e.g. how
"Fulfilled" already appears translated elsewhere, or consistently in English if that's the
intended dictionary source — but consistent one way or the other).

**Actual:** Values like "Pending Approval" and "Approved" render in English, while surrounding
labels ("Status", "Klient", "Waluta", etc.) are correctly Polish. Likely an untranslated
dictionary entry for these specific status codes rather than a hardcoded string.

---

### 5. Duplicate validation message on the order create form

**Page URL:** `/backend/sales/documents/create`

**Steps to reproduce:**
1. Log in as `admin@acme.com`.
2. Choose document type "Zamówienie", pick any customer, do **not** add any line items.
3. Click "Utwórz".
4. Scroll down to the "Pozycje" section.

**Expected:** A single validation message near the line-items table.

**Actual:** The message "Dodaj co najmniej jedną pozycję przed utworzeniem zamówienia." renders
**twice** in the same area, likely from two separate validation-display components rendering
the same underlying error independently. Purely cosmetic — the error itself is correctly
surfaced (a toast/alert does appear); this is only about the duplicated rendering.

---

### 6. Order history log shows changed field names in English

**Page URL:** any sales order detail page, "Historia" tab, e.g.
`/backend/sales/documents/<id>?kind=order`

**Steps to reproduce:**
1. Log in as `admin@acme.com`, interface language: Polish.
2. Open any order, edit its billing or shipping address (the "Adresy" tab) and save it.
3. Open the "Historia" tab.
4. Look at the line for the edit you just made.

**Expected:** The list of changed fields under "Zmienione pola:" reads in Polish, matching the
rest of the interface.

**Actual:** The "Zmienione pola:" label itself is Polish, but the field names listed after it are
in English/technical form — e.g. "Zmienione pola: Billing Address Snapshot, Shipping Address
Snapshot". The same pattern shows up after other kinds of edits too, e.g. "Zmienione pola:
Customer Snapshot, Metadata" or "Zmienione pola: Adjustments, Grand Total Gross Amount, Grand
Total Net Amount, Outstanding Amount, Shipping Gross Amount, Shipping Net Amount, Subtotal Gross
Amount, Subtotal Net Amount". Same pattern as item 4 above (untranslated values inside an
otherwise Polish page).

---

### 7. Hidden-column group label renders in English

**Page URL:** `/backend/sales/orders` (also reproduced on other list pages that use the same
column-visibility panel, e.g. `/backend/catalog/products`)

**Steps to reproduce:**
1. Log in as `admin@acme.com`, interface language: Polish.
2. Open the link above.
3. Click "Dostosuj kolumny" (or "Perspektywy" → the column list).
4. Hide any one column (toggle it off) so it moves into the "Dostępne kolumny" section below.
5. Click the small group heading above the now-hidden column.

**Expected:** The group heading reads in Polish, consistent with "Widoczne" / "Dostępne kolumny"
right above and below it.

**Actual:** The group heading reads "COLUMNS" in plain English/uppercase, while everything
around it ("Widoczne", "Dostępne kolumny", "Ukryj wszystkie") is correctly Polish.

## Environment

- Branch: `main` (0.7.0 pre-release sweep)
- Modules: `rules`, `settings` (record-locks), `sales` (order create + line-item dialogs)
- Browser/Runtime: Chrome

## Additional context

None of these five block any functional flow — all are cosmetic/consistency issues suitable for
a single low-risk PR.

patzick:
## Summary

Clears the 0.7.0 pre-release sweep bundle from #5456 — six of the seven reported findings, each a scoped, low-risk fix on the surface that reported it.

| # | Finding | Root cause | Fix |
|---|---------|-----------|-----|
| 1 | English browser title / sidebar link / breadcrumb on an otherwise Polish `/backend/rules/create` | The page metadata carried only literal `pageTitle` and breadcrumb `label` values — no i18n keys, unlike its sibling list page | Point `pageTitleKey` and the breadcrumb tail at the existing `business_rules.rules.create.title` key (already translated in all five locales) |
| 2 | Two React hydration errors (`<p>` cannot be a descendant of `<p>`) on `/backend/settings/record-locks` | `AlertDescription` renders a `<p>`; the strategy help text nested two more `<p>`s inside it | Render the two help paragraphs as block `<span>`s |
| 3 | "Status" search-select on the order-line dialog renders pre-expanded | `minQuery={0}` makes `LookupSelect`'s `shouldSearch` true for the empty query, so the full list renders before any interaction — every sibling lookup in the dialog uses `minQuery >= 1` | `minQuery={1}` on the status field |
| 5 | "Add at least one line item…" renders twice on order create | `SalesOrderDraftLines` renders the field error itself, and `CrudForm` rendered it a second time because the `lines` custom field never declared `rendersOwnError` | Declare `rendersOwnError: true` on the `lines` field |
| 6 | Order history lists changed field names in English under a Polish label | `CHANGED_FIELD_LABELS` only covered line-level fields, so document-level snapshot diff columns fell through to `humanizeChangedField` | Declare labels for the document-level fields (address/customer snapshots, totals, metadata, dates, references) and ship them in en/pl/de/es/ko |
| 7 | Column-visibility group heading reads "COLUMNS" | `DataTable` and `useAutoDiscoveredFields` both emitted a literal `'Columns'` (and `'Custom Fields'`) as the group name | Resolve both defaults through `ui.columnChooser.defaultGroup` / `ui.columnChooser.customFieldsGroup`; explicit `meta.columnChooserGroup` values are untouched |

## Not included: item 4 (order status values render in English)

Left out deliberately, with reasoning rather than an omission.

`pending_approval` / `approved` are **rows in the tenant's `sales_order_statuses` dictionary**, seeded in English by `ORDER_STATUS_DEFAULTS`. `dictionaries:dictionary_entry.label` is already declared translatable and the translation-overlay plugin already applies per-locale overrides on read — so the platform behaves as designed and the gap is that the out-of-the-box seed ships English-only, with no translation records for those entries.

Fixing it properly means seeding locale overlays for every default status entry plus a backfill path for existing tenants — new seeding behaviour, not the "one-line-ish, low-risk" change the rest of this bundle is scoped for, and it touches `setup.ts` semantics. Worth its own issue. Everything else in #5456 ships here.

## Tests

New regression coverage:

- `packages/core/src/modules/business_rules/backend/__tests__/page-metadata.test.ts` — the create page localizes its title and every breadcrumb segment (item 1)
- `packages/enterprise/src/modules/record_locks/backend/settings/record-locks/__tests__/page.markup.test.tsx` — no `p p` nesting in the rendered settings page; **verified to fail against the pre-fix markup** (item 2)
- `packages/ui/src/backend/inputs/__tests__/LookupSelect.test.tsx` — pins the contract the status field relies on: `minQuery={0}` renders pre-expanded, `minQuery={1}` stays collapsed behind the start-typing hint until the user types (item 3)
- `packages/core/src/modules/sales/widgets/injection/document-history/__tests__/changedFieldLabels.test.ts` — every declared label key ships in en + pl, document-level snapshot fields translate instead of humanizing, unmapped fields still humanize (item 6)
- `packages/ui/src/backend/utils/__tests__/useAutoDiscoveredFields.groupLabels.test.tsx` — default and custom-field group labels translate; explicit `columnChooserGroup` values pass through (item 7)

Item 5 is a one-token declaration consuming a contract already covered by `CrudForm.validation.test.tsx` ("renders each custom field error through exactly one declared owner"); no separate test added.

Run locally (the rest is left to CI, per the request on this task):

- `yarn workspace @open-mercato/core test src/modules/sales src/modules/business_rules` — 107 suites, 1008 tests, all pass
- `yarn workspace @open-mercato/ui test src/backend/utils src/backend/inputs src/backend/__tests__/DataTable src/backend/__tests__/custom-field` — 39 suites, 250 tests, all pass
- `yarn workspace @open-mercato/ui test src/backend/__tests__/CrudForm` — 21 suites, 106 tests, all pass
- `yarn workspace @open-mercato/enterprise test src/modules/record_locks` — 17 suites, 97 tests, all pass
- `yarn build:packages`, `yarn generate`, `yarn typecheck`, `yarn lint` (0 errors), `yarn i18n:check-sync` (all locales in sync), `yarn i18n:check-usage` (advisory only)

## Notes for review

- i18n keys were added to all five locales (en, pl, de, es, ko). The `ui.columnChooser.*` additions are mirrored into `packages/create-app/template/src/i18n/` per the template sync checklist.
- `CHANGED_FIELD_LABELS` and `translateChangedField` are now exported from the document-history widget client so the label set can be asserted against the shipped dictionaries.
- No database, API, or contract-surface changes.

Closes #5456

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5481"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789998018&installation_model_id=432243&pr_number=5481&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5481&signature=90d2a2c2ce3e314283408bf86509b862ab972056f147ebcd94d80978fad04105"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


patzick:
🤖 Labels applied by `om-auto-fix-issue`:

- `bug` — the issue is a defect bundle from the 0.7.0 pre-release sweep.
- `review` — ready, not draft, no other pipeline state applies.
- `needs-qa` — the change touches UI-rendering files (`.tsx` under `packages/ui/src/` and module `components/`), so the automated-verification exemption does not apply. Visual confirmation wanted on the Polish interface for: the rules create page title/breadcrumb, the record-locks console (zero hydration errors), the order-line Status field (collapsed), the order create validation message (once), the order history "Zmienione pola" list, and the column-chooser group heading.
- `priority-low` — none of the findings block a functional flow; all are cosmetic/consistency.
- `risk-low` — no database, API, or contract-surface changes; the widest-reaching change is a translated fallback string for the column-chooser group label, which leaves explicit `meta.columnChooserGroup` values untouched.

Item 4 of #5456 (order status *values* in English) is deliberately out of scope — see the PR description for why it needs seeded translation data rather than a code change.

adeptofvoltron:
`Clears the 0.7.0 pre-release sweep bundle from https://github.com/open-mercato/open-mercato/issues/5456` - that is why it is merge to main

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-08-22T15:22:07Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron (changes requested):
# 🔍 Code Review: fix(ui,sales,rules,record-locks): clear 0.7.0 pre-release UX and i18n nits

## 🎯 Summary

This PR clears six of the seven findings from the 0.7.0 pre-release sweep bundle (#5456), and the overall shape of the work is good: each fix is scoped to the surface that reported it, the PR body explains the root cause per item rather than just the symptom, and item 4 is left out with a written justification instead of being silently dropped — that is exactly the right call for a dictionary-seeding change that would touch `setup.ts` semantics. Test coverage is unusually conscientious for a "nits" bundle: five new or extended test files, one of them explicitly verified to fail against the pre-fix markup, and the `changedFieldLabels` test asserts that every declared label key actually ships in the dictionaries rather than merely that the mapping exists.

Reviewed scope: the column-chooser group labels in `packages/ui` (`DataTable.tsx`, `useAutoDiscoveredFields.ts`), the order-line status lookup and the sales-document form in `packages/core/src/modules/sales`, the document-history changed-field label map, the business-rules create page metadata, the record-locks settings markup in `packages/enterprise`, and the i18n dictionaries for all five locales plus the create-app template mirror.

Five of the six items land cleanly. Item 3 does not: swapping `minQuery={0}` for `minQuery={1}` on the order-line Status field fixes the cosmetic "renders pre-expanded" complaint by trading it for a functional regression on the edit path, because `LookupSelect` has no collapsed-state representation of its own selection. Two smaller items leave the reported bug still visible on surfaces the fix does not reach.

## Verdict

❌ **request changes** — the `minQuery={1}` change on the order-line Status field makes the currently selected status, its checkmark, and the "Clear selection" button disappear entirely whenever the dialog opens on a line that already has a status (verified with a scratch render test against this branch), which is a worse outcome than the pre-expanded list it replaces. Two Medium findings compound it: the item 7 column-chooser fix does not reach the three customers list pages, which pass an explicit English `'Custom Fields'` group, and `useAutoDiscoveredFields` now hard-requires an `I18nProvider` in a shared exported hook where the repo ships `useOptionalT` for precisely that case.

## 🧪 Validation Gate

Per this repository's `om-auto-review-pr` extension, validation is **GitHub-checks-first**: CI already ran the full gate on the PR head, so the configured `validation.commands` were not re-run locally.

**Validation source: GitHub checks** (head `819f1cc9b8ac45780dc7429b69b3ac314ebb51a7`). Branch protection on `main` is not readable through the API (404), so per the skill's fallback rule every reported check is treated as required. There are **no failing and no pending checks** — every non-skipped check is green.

| Command / check | Status | Notes |
|---------|--------|-------|
| `prepare` (build + generate) | ✅ PASS | 4m21s — covers `yarn build:packages` and `yarn generate` for the downstream jobs. |
| `lint` | ✅ PASS | 1m19s. |
| `ds-lint` | ✅ PASS | 1m31s — design-system token compliance. |
| `test` | ✅ PASS | 7m43s — includes the five new/extended suites in this PR. |
| `docker-build` | ✅ PASS | 3m45s — stands in for `yarn build:app`. |
| `ephemeral-integration (1–15/15)` | ✅ PASS | All fifteen shards green, 18–21m each. |
| `CodeQL` / `Analyze (actions, javascript-typescript, python)` | ✅ PASS | Static security analysis. |
| `audit-scope`, `scope`, `license/cla` | ✅ PASS | Scope and CLA gates. |
| `comment`, `audit`, `merge-coverage`, `mutate`, `[code]smith` | ⏭️ skipped | Skipped by their own workflow conditions, not failures. |

One local command was run, and only as evidence for the Major finding below rather than as part of the gate: a scratch Jest render of `LookupSelect` in `packages/ui`, comparing `minQuery={0}` against `minQuery={1}` with a value already selected. It was removed afterwards and the worktree is clean.

## Findings

### ⚠️ Major

**`packages/core/src/modules/sales/components/documents/LineItemDialog.tsx:2513` — `minQuery={1}` hides the currently selected status, its clear button, and the whole option list on the edit path.**

`LookupSelect` renders *no* representation of its selected value outside the listbox. The selected card, the ✓ `Check` icon, and the "Clear selection" `Button` all live inside the `shouldSearch ? (…) : …` branch (`packages/ui/src/backend/inputs/LookupSelect.tsx:253` through `:355`). When `shouldSearch` is false the component renders only the search input and a "Start typing to search." hint.

`shouldSearch` is computed at `LookupSelect.tsx:119-120`:

```ts
const shouldSearch =
  defaultOpen || query.trim().length >= minQuery || Boolean(value && (options?.length ?? 0) > 0)
```

The third clause looks like it exists to keep a made selection visible, but it is gated on the `options` **prop**, and the status field passes only `fetchItems` — never `options`. So `options?.length ?? 0` is `0`, the clause is permanently `false`, and with `defaultOpen` false and an empty query, `shouldSearch` is false for any `minQuery >= 1`.

Concrete failure: a user opens the line-item dialog on an order line that already has a status set. Before this change (`minQuery={0}`) the list rendered, the current status was highlighted with a ✓, and "Clear selection" was available. After this change the same dialog shows an empty search box and "Start typing to search." — the user cannot see which status is currently assigned, and cannot clear it without first typing a character. I verified this against this branch with a scratch render test: at `minQuery={0}` with `value="st-1"`, the selected item's title and the clear button are both present; at `minQuery={1}` with the same value, `queryByText('Pending approval')`, `queryByRole('listbox')` and the clear button are all null.

There is a second, smaller cost on the create path: line statuses are a small closed tenant dictionary (`fetchLineStatusItems`, `LineItemDialog.tsx:1219`, which returns the full list for an empty query and filters client-side otherwise). Requiring a keystroke before anything renders means a user who does not already know the status names has nothing to browse.

The new `LookupSelect` test added in this PR (`packages/ui/src/backend/inputs/__tests__/LookupSelect.test.tsx`) pins the `minQuery` contract only for `value={null}`, which is why it passes without catching this.

Suggested fixes, in rough order of preference:

1. Fix the root cause in `LookupSelect` so a made selection survives the collapsed state — broaden the third clause to `Boolean(value)`, or render a compact "selected: {title} · Clear" summary above the input whenever `value` is set and `shouldSearch` is false. This also helps every other `minQuery >= 2` lookup in the codebase, which has the same blind spot today whenever a value is set.
2. Revert to `minQuery={0}` on the status field and address the "renders pre-expanded" complaint visually instead (the option rows are `p-4` cards with 48px icon tiles — a compact variant would fix the perceived bulk without hiding anything).
3. Use a plain select for a small closed dictionary; a search box over five statuses is the wrong primitive regardless of `minQuery`.

Whichever route you take, please extend the new test to cover the `value` -set case so the collapsed state can never hide a selection again.

### 🔹 Minor

**`packages/core/src/modules/customers/backend/customers/{deals,people,companies}/page.tsx:685-686, :856-857, :829-830` — item 7's fix does not reach the three customers list pages.**

The PR correctly translates the *default* group label, and deliberately leaves explicit `meta.columnChooserGroup` values untouched. But these three pages pass an explicit, hardcoded English literal:

```ts
columnChooserGroup: def.group?.title ?? 'Custom Fields',
filterGroup: def.group?.title ?? 'Custom Fields',
```

Because explicit meta wins over the new default in `useAutoDiscoveredFields.ts:106`, the column chooser on `/backend/customers/people`, `/companies` and `/deals` still shows a "Custom Fields" heading in Polish, German, Spanish and Korean — the exact symptom item 7 reports, on three of the highest-traffic list pages in the product. Each site is a one-line fix: resolve the fallback through `t('ui.columnChooser.customFieldsGroup', 'Custom Fields')`, which the PR has already added to every locale. (`packages/ui/src/backend/utils/customFieldFilters.ts:140` carries the same literal in `buildAdvancedFilterFieldsFromCustomFields`; that export currently has no callers, so it is only worth fixing for consistency.)

**`packages/ui/src/backend/utils/useAutoDiscoveredFields.ts:65` — a shared exported hook now throws when rendered outside an `I18nProvider`.**

`useT()` throws `useT must be used within I18nProvider` when no provider is in scope (`packages/shared/src/lib/i18n/context.tsx:75`). `useAutoDiscoveredFields` is exported from `@open-mercato/ui/backend/utils/useAutoDiscoveredFields` and is consumed directly by module code outside this repo. Its signature is unchanged, so this is not a `BACKWARD_COMPATIBILITY.md` contract break, but it does tighten a runtime precondition on a public entry point with no bridge — and the repo ships `useOptionalT` documented for exactly this situation ("shared components that receive their translator as a prop and may render outside a provider (galleries, tests)").

Every in-repo consumer (`DataTable.tsx:2797`, plus the three customers pages) already renders under a provider, so nothing breaks here today; the risk is third-party module code and any test that renders the hook bare. Suggested fix:

```ts
const t = useOptionalT()
const defaultGroupLabel = t?.('ui.columnChooser.defaultGroup', 'Columns') ?? 'Columns'
const customFieldsGroupLabel = t?.('ui.columnChooser.customFieldsGroup', 'Custom Fields') ?? 'Custom Fields'
```

This keeps the new test green and preserves the hook's previous provider-free behavior.

### 💅 Nit

**`packages/ui/src/primitives/alert.tsx:257` and `packages/ui/src/ai/parts/ConfirmationCard.tsx:223` — the `<p>`-in-`<p>` fix is instance-local, and the same invalid nesting still exists elsewhere.**

The record-locks fix is correct and the accompanying markup test is the right way to pin it. The root cause, though, is that `AlertDescription` renders a `<p>`, so every caller is one `<p>` away from the same hydration error — and `ConfirmationCard.tsx:223-231` already has it, nesting both a `<p>` and a `<ul>` inside `AlertDescription`. Both are invalid HTML and produce the same React hydration warning on the AI confirmation card. Consider either fixing that second instance in passing, or changing `AlertDescription` to render a `<div>` with the same classes (it is typed `HTMLParagraphElement`, so that is a small type change) which removes the whole class of bug. Out of this PR's stated scope, so a follow-up issue is a fine answer.

**`apps/mercato/src/i18n/{de,es}.json:806-812` — the two new keys are translated, but their immediate neighbours are still English.**

The diff inserts `"ui.columnChooser.defaultGroup": "Spalten"` between `"ui.columnChooser.close": "Close"` and `"ui.columnChooser.hideAll": "Hide all"`. The rest of the `ui.columnChooser.*` block in de and es is untranslated: `title` ("Columns"), `ungrouped` ("Other"), `available`, `close`, `hideAll`, `search`, `selected`. So a German user opening the column chooser now sees a panel titled "Columns", a "Search columns..." box, and groups labelled "Spalten" and "Benutzerdefinierte Felder" — item 7's payoff is largely invisible in de and es. This is pre-existing debt tracked by the advisory `yarn i18n:check-values` checker rather than something this PR introduced (`i18n:check-sync` only checks key presence, which is why CI is green), but you were editing these exact lines, and `title`/`ungrouped` in particular are the two that sit directly above the group headings you just localized.

**`packages/core/src/modules/business_rules/backend/__tests__/page-metadata.test.ts:41-47` — the breadcrumb assertion checks shape, not resolution.**

The test asserts each breadcrumb segment has a non-empty string `labelKey`. It does fail against the pre-fix metadata, so it is a valid regression test. But it would also pass for a typo'd key that resolves to nothing and renders the English fallback — which is the bug being fixed. That risk is real in this module specifically, because its keys are split across two dictionaries: `rules.nav.*` lives in the app dictionary (`apps/mercato/src/i18n/*.json`) while `business_rules.*` lives in the module dictionary (`packages/core/src/modules/business_rules/i18n/*.json`), and they are only merged at runtime by `loadDictionary`. The `changedFieldLabels` test in this same PR gets this right by asserting dictionary membership; consider the same here, resolving each `labelKey` against the merged dictionaries. (I verified manually that all keys used by the create page do resolve in all five locales, so this is about protecting the fix, not a live bug.)

**`packages/core/src/modules/business_rules/backend/{sets/create,rules/[id],sets/[id]}/page.meta.ts` — sibling pages carry the identical defect.**

`sets/create/page.meta.ts` has the exact shape that item 1 reports: a breadcrumb tail of `{ label: 'Create Rule Set' }` with no `labelKey`, and no `pageTitleKey` at all. The two edit pages declare only `pageTitle`. All three will still show English titles and breadcrumbs on an otherwise Polish interface. Outside the reported scope, so a follow-up is reasonable, but they are three trivial edits in a module this PR already touches.

**`packages/core/src/modules/sales/components/documents/SalesDocumentForm.tsx:1314` — `rendersOwnError` is declared unconditionally for a component that only renders conditionally.**

The fix is correct today: the only `lines` field error is raised inside the `if (documentKind === 'order')` branch at `SalesDocumentForm.tsx:1427-1432`, and `SalesOrderDraftLines` renders `error` at `SalesOrderDraftLines.tsx:164`. But `component` returns `null` for every other document kind while `rendersOwnError: true` suppresses `CrudForm`'s own rendering unconditionally (`CrudForm.tsx:4726`), so the day someone adds a quote-side lines validation the error will vanish silently with no visible failure. A short comment recording the coupling, or a test pinning that `lines` errors are order-only, would keep that from becoming a debugging session later.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the change is additive: `CHANGED_FIELD_LABELS` and `translateChangedField` gain `export` in `widget.client.tsx`, nothing loses one.
- [x] No function signature changed in a breaking way — `useAutoDiscoveredFields` keeps its input and result types; only its runtime provider precondition tightened, which is the Minor finding above rather than a signature break.
- [x] No required type field removed or narrowed — `rendersOwnError` is an existing optional field on `CrudCustomField`, newly set by one caller.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — no API surface in this diff.
- [x] No field removed or retyped in an existing response shape — no response types touched.
- [x] No event or message name renamed or removed; no payload field removed — no event definitions in this diff.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — no CLI surface touched.
- [x] No database table or column renamed or removed; no column type narrowed — no entities and no migrations in this diff, consistent with the PR body's claim.
- [x] No config key renamed and no default changed silently — the `minQuery` default on `LookupSelect` is untouched at `2`; only one call site changed, which is the Major finding above.
- [x] Where a contract had to change: no contract had to change.

The i18n keys added across `apps/mercato/src/i18n/*.json` are correctly mirrored into `packages/create-app/template/src/i18n/*.json`, satisfying the template sync checklist for the app-level dictionary.

## 🧪 Test Coverage

The coverage shipped here is genuinely good and worth calling out. `page.markup.test.tsx` asserts `container.querySelectorAll('p p')` is empty rather than scraping console output, which is the right way to pin a hydration bug, and the PR body records that it was verified to fail against the pre-fix markup. `changedFieldLabels.test.ts` checks all three behaviors that matter — that every declared key ships in the dictionaries, that document-level snake_case snapshot columns translate, and that unmapped fields still humanize. `useAutoDiscoveredFields.groupLabels.test.tsx` covers the English fallback, the translated case, and the explicit-group passthrough. I independently confirmed that all 54 declared `sales.documents.history.fields.*` keys resolve in en, pl, de, es and ko, and that `business_rules.rules.create.title` resolves in all five.

The gaps:

- **The Major finding is untested and the new test's shape is what let it through.** `LookupSelect.test.tsx` exercises `minQuery` only with `value={null}`. Add a case with a value already selected asserting that the selected item's title and the clear control remain reachable at `minQuery >= 1` — that test fails on this branch today and is the regression test for the fix.
- **The `page-metadata` breadcrumb assertion should resolve keys, not just check their shape**, per the Nit above.
- Item 5's justification for adding no test — that `rendersOwnError` is a one-token declaration already covered by `CrudForm.validation.test.tsx` ("renders each custom field error through exactly one declared owner") — is reasonable and I am not asking for a test there. A case pinning that `lines` errors only ever arise for `documentKind === 'order'` would guard the coupling noted above, but that is optional.
- Item 6's dictionary assertions cover en and pl only. The `DOCUMENT_FIELDS` translation case necessarily needs a specific locale, but the "every declared key ships" case could iterate all five locales at no extra cost and would catch a de/es/ko key going missing.

adeptofvoltron:

- ❌ `changes-requested` — the review found actionable issues: the order-line Status field's `minQuery={1}` change hides the currently selected status and its clear control on the edit path (Major), the column-chooser localization does not reach the three customers list pages, and `useAutoDiscoveredFields` now hard-requires an `I18nProvider` in a shared exported hook.
- 🐛 `bug` — unchanged; this PR is a defect bundle clearing the 0.7.0 pre-release sweep reported in #5456.
- 🧪 `needs-qa` — unchanged; the change touches UI-rendering files across `packages/ui`, `packages/core` and `packages/enterprise`, so the automated-verification exemption does not apply and a human should exercise the affected screens once the findings are addressed.
- 🔽 `priority-low` — kept as applied by `om-auto-fix-issue`; every item in the bundle is cosmetic or consistency work and none of them blocks a functional flow in the shipped product.
- 🟡 `risk-medium` — raised from `risk-low` because the diff edits two shared UI surfaces, `packages/ui/src/backend/DataTable.tsx` and the exported `useAutoDiscoveredFields` hook, which every backend list page in the product renders through, and because the review confirmed a real behavioral regression on the sales order-line edit path rather than a purely cosmetic change. The blast radius is wider than a single module, so `risk-low` understated it.

adeptofvoltron:
Thanks @patzick — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

adeptofvoltron:

The review submitted one Major and two Minor findings, plus five nits. The Major is the order-line Status field: swapping `minQuery={0}` for `minQuery={1}` fixes the pre-expanded list but makes the currently selected status, its checkmark and the "Clear selection" button disappear whenever the line-item dialog opens on a line that already has a status, because `LookupSelect` renders no representation of its selection outside the listbox and its `value`-based escape hatch is gated on an `options` prop this call site never passes. That was verified with a scratch render test against this branch, which has since been removed. The two Minors are the column-chooser localization not reaching the three customers list pages, which pass an explicit English `'Custom Fields'` group, and `useAutoDiscoveredFields` switching to `useT()` in a shared exported hook where the repo ships `useOptionalT` for provider-free rendering.

The other five items in the bundle land cleanly and the test coverage is above the bar for a nits PR. Validation source: GitHub checks — every reported check on head `819f1cc9b` is green, with nothing failing and nothing pending, so the local `validation.commands` gate was not re-run.

autofix: skipped (not my PR — re-run with --autofix to fix it here). The PR is reassigned to @patzick.

patzick:
🤖 `om-auto-fix-pr` claiming this PR for a fix pass on @adeptofvoltron's review. Working the Major (order-line Status selection hidden at `minQuery={1}`) and both Minors, plus the cheap nits; the `AlertDescription` root-cause nit goes to a follow-up issue since changing a primitive's rendered element is an ask-first change. Will report back on this PR.

patzick:
🤖 `om-auto-fix-pr` — review pass complete. Pushed `ba6b9dc6a` on top of a fresh merge of `main` (base had advanced 5 commits; merged clean).

Thanks @adeptofvoltron — the Major is a real regression and the diagnosis was exactly right, including the reason my own test missed it.

## 💥 Major — fixed at the root

You were right that `minQuery={1}` hid the selection. Confirmed the mechanism: `shouldSearch`'s third clause was `Boolean(value && (options?.length ?? 0) > 0)`, and the status call site never passes `options`, so with a `value` set and an empty query the whole `shouldSearch` block — selected row, checkmark and **Clear selection** — never rendered.

Took your preferred route (1) rather than reverting, because the blind spot is not specific to this call site:

```ts
const shouldSearch = defaultOpen || query.trim().length >= minQuery || Boolean(value)
```

A set `value` now always opens the list. That is the only place `LookupSelect` represents a selection, so collapsing over one was never correct for *any* caller — this also closes the gap for every other `minQuery >= 1` lookup that resolves its selection through `fetchItems` instead of `options`. The order-line Status field keeps `minQuery={1}`, so item 3's original complaint still holds on a new line, while an existing line now opens showing its status exactly as the Produkt and Wariant fields already do.

Extended the test as you asked. The new `value`-set case **fails against `819f1cc9b` and passes on `ba6b9dc6a`** — I ran it in that order to confirm it is a real regression test and not a tautology. Also added a case pinning that the collapsed state survives once the selection is cleared.

## 🔹 Minor — both fixed

**Customers list pages.** Confirmed and fixed on all three — `people:856-857`, `companies:829-830`, `deals:685-686` now resolve through `t('ui.columnChooser.customFieldsGroup', 'Custom Fields')`. `t` was already in each `useMemo`'s dependency list, so no dependency change was needed. Good catch that explicit meta wins over the new default; this was item 7's exact symptom surviving on three of the highest-traffic list pages.

`buildAdvancedFilterFieldsFromCustomFields` (`customFieldFilters.ts:140`) I deliberately left alone. It is a plain function rather than a hook, so it cannot call `useT`, and its optional `groupLabel` parameter is already the mechanism by which a caller supplies a translated label — the English literal is the documented fallback for when none is passed, structurally the same shape as the hook's provider-free fallback below. Changing it would mean adding a required parameter to an exported function for no behavioral gain. Happy to revisit if you'd rather it took a mandatory label.

**`useAutoDiscoveredFields` provider precondition.** Fixed with `useOptionalT` exactly as you suggested. You are right that this tightened a runtime precondition on a public entry point with no bridge, which is the kind of thing that breaks third-party module code silently. The literal fallbacks keep the previous provider-free behavior and the new test still passes unchanged.

## 💅 Nits

- **`AlertDescription` root cause + `ConfirmationCard.tsx:223`** — filed as #5487 rather than fixed here. Changing a primitive's rendered element is ask-first under `packages/ui/AGENTS.md`, and the ref type narrows from `HTMLParagraphElement` to `HTMLDivElement`, so it wants a maintainer decision. The issue records both options and the `p p` assertion as the guard.
- **de/es `ui.columnChooser.*` neighbours** — fixed. You were right that the payoff was invisible: a German user would have seen a panel titled "Columns" with groups labelled "Spalten". Translated the whole block (`title`, `ungrouped`, `available`, `close`, `hideAll`, `search`, `selected`, `shown`, `toggle`) in de and es, mirrored into the create-app template. ko was already complete.
- **Breadcrumb assertion checked shape, not resolution** — fixed, and this was the sharpest of the nits. The test now resolves every declared key against the **merged** app + module dictionaries in all five locales, so a typo that silently renders the English fallback fails it. Your point about the two-dictionary split in this module is precisely why shape-only was too weak.
- **Sibling `business_rules` pages** — fixed rather than deferred, since they carry item 1's defect verbatim and the module was already open. `sets/create`, `rules/[id]` and `sets/[id]` all get a `pageTitleKey` and a fully localized breadcrumb; the metadata test now covers all four pages.
- **Item 6 dictionary assertion covered en+pl only** — now iterates all five locales, as you noted it could at no extra cost.
- **`rendersOwnError` declared unconditionally** — left as-is, per your "optional" and "not asking for a test there". Worth recording the reasoning: the repo's `AGENTS.md` asks for self-documenting code over inline comments, so a comment is the wrong instrument, and a test pinning "`lines` errors are order-only" would pin the *current* validation layout rather than the coupling itself. If quote-side lines validation ever lands, the fix is to make `rendersOwnError` conditional on `documentKind === 'order'` at that point.

## ✅ Validation

Local, related suites only (per the run's instruction — the rest is CI's job): ui `utils`/`inputs`/`DataTable`/`custom-field`/`CrudForm` 60 suites / 358 tests; core `sales`/`business_rules`/`customers` 337 suites / 2424 tests; enterprise `record_locks` 17 suites / 97 tests. Plus `typecheck` (25/25), `lint` (0 errors) and `i18n:check-sync` (all locales in sync). Full CI is running on `ba6b9dc6a` — **this is not yet green, it is pending**, and I will not read merge-readiness from a wall of green until `test` and the integration shards land.

Re-requesting review. `risk-medium` is left as you set it — the `LookupSelect` change touches a shared primitive, so if anything the wider blast radius argues for keeping it there.

adeptofvoltron:
🤖 `om-auto-qa-pr` started by @adeptofvoltron at 2026-08-22T15:55:00Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

Claim note for transparency: the three-signal check found @patzick assigned and a `om-auto-fix-pr` claim comment inside the 30-minute stale window, but that claim was superseded by its own completion comment ("review pass complete… Re-requesting review") a few minutes later, the `in-progress` label is gone, and the remaining `ci-monitoring` label is explicitly not a lock signal. The assignee is the PR author, which is authorship rather than a live automation lock, so this PR is treated as free to claim rather than as a conflict. If a fix pass is in fact still running on `ba6b9dc6a`, say so and I will stand down.

patzick:
🤖 `om-auto-fix-pr` — CI result on `ba6b9dc6a`: **27 pass, 0 fail.** `test` (typecheck, unit, i18n sync/usage, create-app template parity) and all 15 `ephemeral-integration` shards green, plus `docker-build`, `lint`, `ds-lint` and CodeQL.

`audit` shows as skipping, which here is the by-design path rather than a hidden gate: it runs only when `audit-scope.should_run` is true, and this PR touches no `package.json` or lockfile. `test`'s condition explicitly tolerates `needs.audit.result == 'skipped'`, so it genuinely ran. `merge-coverage` is a downstream reporting job, not a gate.

Dropping `ci-monitoring`; the CI follow-up this run owed has landed.

**Two human gates remain, and neither is mine to close:**

1. **Re-review** — @adeptofvoltron's `CHANGES_REQUESTED` is still the standing review state. Every finding is addressed or answered in the comment above, and review has been re-requested, but the PR is not approved yet. Pipeline label is therefore `review`, not `merge-queue`.
2. **QA** — `needs-qa` stands. Worth exercising the order-line Status field specifically: open **Edytuj pozycję** on a line that already has a status and confirm the status, its checkmark and **Wyczyść wybór** are visible, then confirm a *new* line still opens collapsed with the "Zacznij pisać, aby wyszukać." hint. That pairing is the whole point of the Major fix. The other five surfaces from the original label comment are unchanged.

No merge from this run by design — that stays with `om-approve-merge-pr` behind the QA gate.

adeptofvoltron:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ PASS — every UI change this PR claims was reproduced in a real browser, including the Major regression the review caught.
**Environment:** `http://localhost:5011` · role `admin` (`admin@acme.com`) · browser `playwright` (Chromium)
**Verified:** `fix/issue-5456-prerelease-ux-nits` @ `ba6b9dc6a`

The app was booted through `om-prepare-test-env` as a **production build** (`mercato server start`) against a disposable local Postgres database. This machine has no container runtime, so the repo's testcontainers ephemeral environment could not be used; the entrypoint wraps the repo's own build/init/start chain instead. `OM_ENABLE_ENTERPRISE_MODULES=true` so `/backend/settings/record-locks` exists, and `DEMO_MODE=false` exactly as the repo's own integration runner sets it. **The whole run was driven in Polish** — every i18n item in this bundle is invisible in English.

### Scenario (P0–P2 — backoffice UI: business rules, record locks, sales documents, customers lists)

| 1 | **Item 1 baseline.** Open `/backend/rules/create` in English. | English title, as the before/after contrast. | `document.title` = "Create Business Rule"; breadcrumb "Business Rules / Create Business Rule". | ✅ |
| 2 | **Item 1.** Switch to Polish, reopen `/backend/rules/create`. | Title, breadcrumb tail and sidebar link all Polish. | `document.title` = **"Utwórz Regułę Biznesową"**; breadcrumb **"Reguły biznesowe / Utwórz Regułę Biznesową"**; heading and sidebar link match. No English string left in the breadcrumb. | ✅ |
| 3 | **Item 1, review follow-up.** Open the sibling `/backend/sets/create` in Polish. | The sibling page is localized too. | `document.title` = **"Utwórz Zestaw Reguł"**; breadcrumb **"Zestawy reguł / Utwórz Zestaw Reguł"**. No "Create Rule Set" literal. | ✅ |
| 4 | **Item 2.** Open `/backend/settings/record-locks` and inspect the strategy alert. | No `<p>` nested in a `<p>`; both help texts still separate blocks. | **0** `p p` nodes on the page and 0 inside the alert; "Optymistyczna:" and "Pesymistyczna:" render as block spans on their own lines. No DOM-nesting or hydration message in the console during the visit. | ✅ |
| 5 | **Item 5.** Submit an order at `/backend/sales/documents/create?kind=order` with no lines. | The lines-required message renders as exactly one *error*. | Exactly **1** error node (`p[role=alert].text-destructive`). See the note below about the two non-error copies. | ✅ |
| 6 | **Item 3.** Open "Dodaj pozycję" and look at Status on a new line. | Collapsed behind the start-typing hint, not pre-expanded. | `aria-expanded="false"`, the listbox element is **not in the DOM at all**, and "Zacznij pisać, aby wyszukać." is shown. | ✅ |
| 7 | **💥 Major from review.** Pick a Status, then clear the search query. | The list stays open over a selection, showing the selected row and the clear control. | `aria-expanded="true"` with the query empty; exactly one option carries `aria-selected="true"` ("Backordered", checkmarked) and **"Wyczyść wybór"** is visible. The regression is fixed. | ✅ |
| 8 | **Item 7.** `/backend/sales/orders` → column chooser → hide columns. | Default group heading translated. | The available-columns section shows one group: **"KOLUMNY"**. No "Columns" literal. | ✅ |
| 9 | **Item 7, review Minor.** Same on `/backend/customers/people`. | Custom-fields fallback translated. | Groups: Basic Info, CRM, Profile, Socials, **Pola niestandardowe**. | ✅ |
| 10 | **Item 7, review Minor.** Same on `/backend/customers/companies`. | As above. | Groups: Contact, Basic Info, CRM, Profile, Notes, **Pola niestandardowe**. | ✅ |
| 11 | **Item 7, review Minor.** Same on `/backend/customers/deals`. | As above. | Groups: Pipeline, **Pola niestandardowe**. | ✅ |
| 12 | **Item 6.** Give a demo order a document-level change, then read its Historia tab in Polish. | Changed-field names translated, not humanized English. | The entry reads **"Zmienione pola: Notatki wewnętrzne"** — Polish label, Polish field name. No English document-level field name and no camelCase remnant in the rendered diff. | ✅ |

### Notes for QA

- **Item 5 — fixed as claimed, but the user-visible repetition is only reduced.** Screenshot 5 is worth a careful look. Only one *error* renders now, which is exactly what removing the `CrudForm` duplicate should achieve. However `SalesOrderDraftLines` prints the very same sentence three times in the Pozycje block by design: a muted section hint, the red error, and the table empty-state card. So what a user sees goes from four copies to three, not to one. Whether that counts as "renders twice" being fixed is a product call — the PR's stated scope is satisfied either way, and this is not a regression introduced here.
- **Item 4's deferral is corroborated, not contradicted.** Screenshot 7 shows the status values still in English (Picking, Backordered, Delivered, Shipped). That is the dictionary-seed gap the PR deliberately leaves to its own issue, and the evidence agrees with that reasoning.
- **Pre-existing, out of scope:** several column-chooser group headings that come from page column meta are still hardcoded English (Basic Info, Contact, CRM, Profile, Socials, Notes, Pipeline). The PR intentionally does not touch explicit `meta.columnChooserGroup` values, so this is not a regression — but a Polish user does still meet English group headings on the customers lists. Possibly worth its own issue.
- **Unrelated observation:** the order-history diff recorded only `internalNotes` as changed although the same request also set `customerReference`. Not investigated further; it does not affect this PR's label mapping.
- **Not exercised:** the de / es / ko locales (only en and pl were driven), the mobile viewport, and the Status lookup on a line inside an already-created order — the Major was verified on the create-draft dialog, which is the same `LineItemDialog` component and the same `LookupSelect` code path.

### Screenshots

**step-01-rules-create-en**

step-01-rules-create-en

**step-02-rules-create-pl**

step-02-rules-create-pl

**step-03-rule-sets-create-pl**

step-03-rule-sets-create-pl

**step-04-record-locks-pl**

step-04-record-locks-pl

**step-05-order-lines-error-once**

step-05-order-lines-error-once

**step-06-line-status-collapsed**

step-06-line-status-collapsed

**step-07-line-status-selection-visible**

step-07-line-status-selection-visible

**step-08-column-chooser-orders-pl**

step-08-column-chooser-orders-pl

**step-09-column-chooser-people-pl**

step-09-column-chooser-people-pl

**step-10-column-chooser-companies-pl**

step-10-column-chooser-companies-pl

**step-11-column-chooser-deals-pl**

step-11-column-chooser-deals-pl

**step-12-order-history-pl**

step-12-order-history-pl

adeptofvoltron:

This change ships thorough jsdom regression coverage (`LookupSelect.test.tsx`, `page.markup.test.tsx`, `changedFieldLabels.test.ts`, `useAutoDiscoveredFields.groupLabels.test.tsx`, `page-metadata.test.ts`) but **no browser-level test** — no `__integration__` spec is added anywhere in the diff. The UI QA above was manual; the two checks worth locking in with a real browser are the ones whose failure modes are rendering-level rather than unit-level. Run `/om-integration-tests` to add them.

**Scenario A — the order-line Status lookup keeps its selection visible (the Major from review, `sales`):**
1. **Setup:** create a tenant-scoped sales order via the sales API fixtures, plus one line-status dictionary entry so the lookup has something to return.
2. **Act:** open `/backend/sales/documents/create?kind=order`, click "Add line item", type one character into the **Status** lookup, click the first option, then clear the search input.
3. **Assert:** the Status combobox reports `aria-expanded="true"` with an empty query; exactly one option carries `aria-selected="true"`; the clear-selection control is visible. Then assert the *new-line* case: on a freshly opened dialog the combobox reports `aria-expanded="false"` and its listbox element is absent from the DOM.
4. **Teardown:** delete the order and the dictionary entry created in setup.

**Scenario B — column-chooser group headings are localized (`ui` + `customers`):**
1. **Setup:** create one customer person and one custom field definition without an explicit group, so both the default and the custom-fields fallback labels have to render.
2. **Act:** set the locale cookie to `pl`, open `/backend/sales/orders` (default group) and `/backend/customers/people` (custom-fields group), open the column chooser on each, and hide one base column and one custom-field column so the grouped "available columns" list renders.
3. **Assert:** the available-columns group headings contain `Kolumny` / `Pola niestandardowe` and never the literals `Columns` / `Custom Fields`.
4. **Teardown:** delete the person and the custom field definition, and restore the column visibility preference.

Both scenarios are exactly what this QA run drove by hand; the throwaway specs that produced the evidence above are a usable starting point and live under this run's artifacts directory (`.ai/qa/artifacts_20260822-160154-5481/spec/`, not committed).

adeptofvoltron:
🤖 `om-auto-qa-pr` — 🏷️ label rationale

- 📸 `screenshots` — **added by this run.** Twelve browser screenshots of the changed surfaces are now attached to this PR, and `.ai/docs/pr-workflow.md` defines this label as the record that UI QA visual evidence was posted. It is informational and does not gate merge.
- 🧪 `needs-qa` — **left in place.** This run is evidence-only: it produced the proof a QA reviewer needs but does not sign off on it, so the QA gate stays closed until a maintainer applies `qa-approved`. `qa-approved` and `qa-self-verified` were deliberately not applied because `--self-qa-signoff` was not requested for this run.
- ✅ `qa-failed` — **not applied,** and correctly so: every one of the twelve checks passed, so there is no failure to record.
- 🔍 `review` — **left in place.** The pipeline state belongs to code review, not to QA; this run does not move it. Worth flagging for whoever picks this up: GitHub still reports `reviewDecision: CHANGES_REQUESTED` from the review at 15:31, which predates the fix push `ba6b9dc6a`, so a re-review is still outstanding even though the QA evidence is green.
- 🐛 `bug` · ⚖️ `priority-low` · 💥 `risk-medium` — **untouched.** Category, priority and risk are the authoring and review skills' calls, and nothing this QA run observed argues for changing any of them.
- ⏳ `in-progress` — held only for the duration of this QA run and removed in the release comment below.

adeptofvoltron:

All twelve UI checks passed against `ba6b9dc6a` in a real browser, driven in Polish because every i18n item in this bundle is invisible in English — including the Major that the review caught, where a selected order-line status now keeps its list open, its checkmark and its clear control after the search query is cleared. Two things a QA reviewer should read rather than skim: item 5's duplicate error is genuinely gone (one error node), but `SalesOrderDraftLines` still prints the same sentence three times in the Pozycje block by design, so the user-visible repetition drops from four copies to three rather than to one; and item 4's English status values are visible in the evidence, which corroborates rather than contradicts the PR's decision to defer that to its own issue.

`needs-qa` stays and no verdict label was applied — this run is evidence-only, so the QA gate is a maintainer's call. Note also that `reviewDecision` is still `CHANGES_REQUESTED` from the 15:31 review, which predates the fix push, so a re-review remains outstanding independently of QA.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-08-22T16:40:00Z. Re-review of `ba6b9dc6a` in progress; other auto-skills will skip this PR until the lock is released.

Claim note: the three-signal check found no `in-progress` label and no live robot claim — the assignee is @patzick, the PR author, which is authorship rather than an automation lock, and the most recent 🤖 comments are this account's own `om-auto-qa-pr` run, which completed and released. Re-reviewing because the standing `CHANGES_REQUESTED` verdict was submitted against `819f1cc9b` and two fix commits have landed since.

adeptofvoltron (approved):
# 🔍 Code Review (re-review): fix(ui,sales,rules,record-locks): clear 0.7.0 pre-release UX and i18n nits

## 🎯 Summary

This is a **re-review** of `ba6b9dc6a`, following the `changes_requested` verdict I submitted against `819f1cc9b`. One fix commit has landed since (plus a merge of `main`), and it addresses every finding from that review — the Major, both Minors, and all four actionable Nits — without leaving any of them papered over.

The Major was fixed the way I hoped rather than the way that would have been quickest. Instead of reverting `minQuery={1}` on the order-line Status field, the root cause in `LookupSelect` was corrected: the `value`-based escape hatch in `shouldSearch` was gated on an `options` prop the call site never passes, so it broadened to `Boolean(value)` (`packages/ui/src/backend/inputs/LookupSelect.tsx:125`). That keeps item 3's original fix intact — a new line still opens collapsed behind the start-typing hint — while closing the same blind spot for **every** `minQuery >= 1` lookup in the codebase that resolves its selection through `fetchItems`, not just this one. The accompanying comment explains the invariant ("a set `value` always opens the list: it is the only place the selected row, its checkmark and the clear control render") so the next person cannot re-narrow it by accident.

Reviewed scope for this pass: the fix commit in full, re-checked against every previous finding, plus a re-read of the surfaces it touches — `LookupSelect` and its tests, `useAutoDiscoveredFields`, the three customers list pages, the four `business_rules` page-metadata files and their test, the document-history label test, and the `ui.columnChooser.*` blocks in all five app locales and their create-app template mirrors.

Two things beyond the diff are worth recording here. First, the Major was **independently verified in a real browser** during this session's UI QA run, not only by the unit test: on a new line the Status combobox reports `aria-expanded="false"` with no listbox in the DOM, and after selecting a status and clearing the query it reports `aria-expanded="true"` with the selected row checkmarked and "Wyczyść wybór" reachable — [screenshots 6 and 7 in the QA evidence comment](https://github.com/open-mercato/open-mercato/pull/5481#issuecomment-5381421607). Second, the two Nits I said a follow-up would satisfy have actually been filed — #5487 for the `AlertDescription` `<p>` root cause and #5477 for giving the order-lines validation error its own distinct string — so nothing from the previous pass is silently carried.

## Verdict

✅ **approve** — every blocking finding is resolved, and the fixes are better-targeted than the minimum the review asked for. What remains is one new nit about a test's import path, one carried-over optional nit, and one deliberate decline whose stated reasoning I agree with. Under this repository's verdict rule (Medium and above request changes; only Low/nit approve), none of that blocks.

`needs-qa` stays on the PR and is already satisfied in substance: the UI QA run on this exact head passed 12/12 with screenshots attached. It still needs a maintainer to convert that into `qa-approved` — code-review approval is not QA approval, and this skill must not apply that label from reading a diff.

## 🧪 Validation Gate

Per this repository's `om-auto-review-pr` extension, validation is **GitHub-checks-first**: CI ran the full gate on the PR head, so the configured `validation.commands` were not re-run locally.

**Validation source: GitHub checks** (head `ba6b9dc6ae2769878cd3069ee2a2e771c19e26ef`). Branch protection on `main` is not readable through the API (404), so per the skill's fallback rule every reported check is treated as required. There are **no failing and no pending checks** — 27 pass, 0 fail, 0 pending.

| Command / check | Status | Notes |
|---|---|---|
| `prepare` (build + generate) | ✅ PASS | 3m29s — covers `yarn build:packages` and `yarn generate`. |
| `lint` | ✅ PASS | 29s. |
| `ds-lint` | ✅ PASS | 41s — design-system token compliance. |
| `test` | ✅ PASS | 7m40s — includes typecheck, the i18n sync/usage checkers, and the two test files extended by the fix commit. |
| `docker-build` | ✅ PASS | 3m44s — stands in for `yarn build:app`. |
| `ephemeral-integration (1–15/15)` | ✅ PASS | All fifteen shards green, 18–21m each. |
| `CodeQL` / `Analyze (actions, javascript-typescript, python)` | ✅ PASS | Static security analysis. |
| `audit-scope`, `scope`, `license/cla` | ✅ PASS | Scope and CLA gates. |
| `audit`, `comment`, `merge-coverage`, `mutate`, `[code]smith` | ⏭️ skipped | Skipped by their own workflow conditions, not failures. |

No local commands were run this pass; the previous review's scratch `LookupSelect` render is now redundant, because the branch ships the equivalent as a committed test.

Two non-CI checks I ran directly against the branch, because they verify claims CI cannot: every label key newly referenced by the four `business_rules` metadata files (`business_rules.rules.edit.title`, `business_rules.sets.edit.title`, and the create-page keys) resolves in **all five locales** against the merged app+module dictionaries; and `apps/mercato/src/i18n/*.json` is now **byte-identical** to `packages/create-app/template/src/i18n/*.json` in all five locales, satisfying the template sync checklist.

## 📋 Previous findings — status

| Severity | Finding | Status |
|---|---|---|
| ⚠️ Major | `minQuery={1}` hid the selected status, its checkmark and the clear control | ✅ **Fixed at the root** in `LookupSelect.tsx:125` (`Boolean(value)`), with a regression test that fails against the previous commit; browser-verified in this session's QA run |
| 🔹 Minor | Item 7's fix did not reach the three customers list pages | ✅ **Fixed** — `people:856-857`, `companies:829-830`, `deals:685-686` now resolve through `t('ui.columnChooser.customFieldsGroup', …)`; `t` is already in each columns memo's dependency list, and no `?? 'Custom Fields'` literal remains anywhere in `packages/` or `apps/` |
| 🔹 Minor | `useAutoDiscoveredFields` hard-required an `I18nProvider` | ✅ **Fixed** — `useOptionalT()` with literal fallbacks, exactly the suggested shape, plus a comment recording why the hook must tolerate a missing provider |
| 💅 Nit | Sibling `business_rules` pages carried item 1's defect | ✅ **Fixed** — `rules/[id]`, `sets/create` and `sets/[id]` all gained `pageTitleKey` and fully-keyed breadcrumbs |
| 💅 Nit | The breadcrumb test asserted shape, not resolution | ✅ **Fixed, and then some** — the test now resolves every declared key against the merged app+module dictionaries across all five locales, for all four pages rather than just the one I flagged |
| 💅 Nit | `changedFieldLabels` covered en + pl only | ✅ **Fixed** — all five locales |
| 💅 Nit | de/es `ui.columnChooser.*` neighbours were still English | ✅ **Fixed** — the whole block is translated in de and es (and ko/pl were already), so item 7's payoff is now visible in every shipped locale |
| 💅 Nit | `AlertDescription` renders a `<p>`; `ConfirmationCard.tsx:223` has the same nesting | ✅ **Filed as #5487**, which is the answer I said would be fine; the instance is still present on `main`, tracked there |
| 💅 Nit | `rendersOwnError` declared unconditionally for a conditionally-rendered component | ⚪ **Carried** — see below; #5477 covers the adjacent string problem but not this coupling |
| — | `buildAdvancedFilterFieldsFromCustomFields` keeps its English literal | ✅ **Declined with a reason I agree with** — it is a plain function, not a hook, and its `groupLabel` parameter is already the mechanism a caller uses to supply a translated label. Correct call; my original note said it was only worth changing for consistency |

## Findings

### 💅 Nit

**`packages/core/src/modules/business_rules/backend/__tests__/page-metadata.test.ts:26-27` — a `packages/core` test now reaches into `apps/mercato` through a seven-level relative path.**

```ts
...(require(`../../../../../../../apps/mercato/src/i18n/${locale}.json`) as Record<string, string>),
```

The coupling itself is inherent to what the test verifies — `rules.nav.*` genuinely lives in the app dictionary while `business_rules.*` lives in the module's own, and the whole point is to assert the merged pair — so I am not asking for a redesign, and the file already carries an excellent comment explaining exactly that. The fragility is only in the path: seven `../` segments crossing a package boundary into the app is precisely the shape `AGENTS.md`'s import strategy steers away from, and if `apps/mercato` is ever moved or renamed this fails as a bare module-not-found that names a JSON file rather than the real cause. A resolved constant (`path.resolve(__dirname, '../../../../../../../apps/mercato/src/i18n')`) computed once with a short "if this throws, the app dictionary moved" note, or a `@open-mercato/app`-rooted alias if one is available to Jest here, would make the failure mode self-explaining. Genuinely optional — CI is green and the test does what it should.

**`packages/core/src/modules/sales/components/documents/SalesDocumentForm.tsx:1314` — the `rendersOwnError` coupling is still undocumented (carried from the previous review).**

Unchanged and still correct today: the only `lines` field error is raised inside the `documentKind === 'order'` branch, so suppressing `CrudForm`'s own rendering is right. The risk remains that `component` returns `null` for every other document kind while `rendersOwnError: true` suppresses unconditionally, so a future quote-side `lines` validation would vanish silently. A one-line comment or a test pinning that `lines` errors are order-only still seems worth it. Noting it once more for completeness rather than pressing it — it is a nit, it does not block, and #5477 will have someone in this code again shortly.

### 📝 Observation (not a finding)

Broadening `shouldSearch` to include `Boolean(value)` means a `LookupSelect` with a value set now runs its fetch effect on mount where a `minQuery >= 1` instance without an `options` prop previously did not. Opening the line-item dialog on an existing line therefore issues a few more requests than before. This is unavoidable given the chosen fix — the component has no cached title for its selected id, so it must fetch to render the selection at all — and it is plainly the right trade against a selection the user can neither see nor clear. Recording it only so the extra requests are recognisable as intended if they show up in a profile later.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the fix commit is additive; `useAutoDiscoveredFields` swaps `useT` for `useOptionalT` internally with no signature change.
- [x] No function signature changed in a breaking way — `LookupSelect`'s props are untouched, including the `minQuery` default of `2`.
- [x] No HTTP route URL removed or renamed; no method changed — no API surface in this diff.
- [x] No database table or column renamed or removed — no entities and no migrations, consistent with the PR body.
- [x] Where a contract had to change: no contract had to change.

One judgement call worth stating explicitly rather than leaving implicit in a checkbox. `LookupSelect`'s **runtime behavior** does change for third-party callers: any `minQuery >= 1` instance with a value set now renders its list open where it previously rendered collapsed. Component visual behavior is not one of the thirteen contract surfaces `BACKWARD_COMPATIBILITY.md` enumerates, and the change replaces a state in which the selection was unreachable, so this is a bug fix rather than a break — no deprecation path is owed. The previous review's `useAutoDiscoveredFields` provider-precondition concern is now moot, since the hook tolerates a missing provider again.

The i18n keys added across `apps/mercato/src/i18n/*.json` remain correctly mirrored into `packages/create-app/template/src/i18n/*.json` — verified byte-identical in all five locales this pass.

## 🧪 Test Coverage

The regression test for the Major is exactly the shape the previous review asked for, and its comment records the failure it prevents rather than just what it asserts:

```tsx
it('keeps a made selection visible while collapsed, without an options prop', async () => {
  // value="plot-1", fetchItems (no options prop), minQuery={1}
  const selected = await view.findByRole('option', { selected: true })
  expect(selected).toHaveTextContent(ITEMS[0].title)
  const clear = view.getByRole('button', { name: 'Clear selection' })
  fireEvent.click(clear)
  expect(onChange).toHaveBeenCalledWith(null)
})
```

It pins both halves of the invariant — the selection is reachable **and** clearable — and the companion case pins that clearing the selection returns the field to collapsed, so the fix cannot be over-applied into "always open". The PR body's claim that it fails against the previous commit is consistent with the code: at `819f1cc9b` the `options`-gated clause made `shouldSearch` false for exactly this input.

The `page-metadata` test is now materially stronger than what I asked for: rather than resolving the create page's keys, it parameterizes over all four pages and asserts that every declared `pageTitleKey`, `pageGroupKey` and breadcrumb `labelKey` resolves to a non-empty string in all five locales. A typo'd key that silently renders the English fallback — the bug item 1 exists to fix — now fails the suite. `changedFieldLabels` extends its "every declared key ships" assertion to all five dictionaries at no extra cost, which was the cheap win I flagged.

Remaining gaps, both minor and both already visible above: the `rendersOwnError` coupling has no pinning test, and there is still no browser-level test anywhere in the diff. On the latter — this session's UI QA run covered the changed surfaces manually and passed, and I posted two ready-to-implement integration scenarios in the [follow-up comment](https://github.com/open-mercato/open-mercato/pull/5481#issuecomment-5381422887) so the manual pass can be converted into automation. Not a blocker for this PR: the jsdom coverage here is genuinely thorough, and the two scenarios are better filed than rushed.

adeptofvoltron:

**Read this first — most of it is already done.** A UI QA run on this exact head (`ba6b9dc6a`) passed 12/12 with screenshots: [evidence comment](https://github.com/open-mercato/open-mercato/pull/5481#issuecomment-5381421607). Every route below is annotated with whether it already has evidence, so you can spot-check the covered ones and spend your time on the gaps rather than repeating the run. **Set your interface language to Polish before starting** (profile menu → Language → Polski) — every i18n item in this bundle is invisible in English.

### P1 — Order-line Status lookup (`sales`) — the highest-risk change

**Where to click**
- `/backend/sales/documents/create?kind=order` → "Dodaj pozycję" → the **Status** field
- Any existing order → `/backend/sales/orders/{id}` → Pozycje tab → edit a line that already has a status

**What to verify**
- On a **new** line, Status shows only the search box and "Zacznij pisać, aby wyszukać." — no list of statuses before you type. *(covered — screenshot 6)*
- Type a character, pick a status, then **clear the search box**: the list stays open, your choice is highlighted with a checkmark, and "Wyczyść wybór" is available. *(covered — screenshot 7)*
- **Not yet covered:** open the line dialog on an **existing order line that already has a status**. The selected status must be visible and clearable the moment the dialog opens, with no typing. This is the exact regression the fix targets and the one path the automated run did not reach.
- **Not yet covered:** the other lookups in the same dialog (Produkt, Wariant, Cena) still behave — collapsed until you type when empty, and showing their selection when set.

**What can go wrong**
- The dialog opens on an existing line showing an empty search box, so you cannot tell which status is assigned or clear it — the pre-fix behavior.
- The list is stuck open on a new empty line (the fix over-applied), which is the cosmetic bug this bundle started from.
- A perceptible pause when the dialog opens: a set value now triggers a fetch on mount that previously did not fire. Watch for a flash of "Ładowanie statusów…" or a jumpy layout.

### P1 — Localized page titles, breadcrumbs and column headings (`business_rules`, `ui`, `customers`)

**Where to click**
- `/backend/rules/create` and `/backend/sets/create` *(covered — screenshots 2 and 3)*
- **Not yet covered:** `/backend/rules/{id}` and `/backend/sets/{id}` (the two **edit** pages, fixed in the last commit but never opened in a browser)
- `/backend/sales/orders` → "Dostosuj kolumny" *(covered — screenshot 8)*
- `/backend/customers/people`, `/companies`, `/deals` → "Dostosuj kolumny" *(covered — screenshots 9–11)*

**What to verify**
- Each page's **browser tab title**, its `<h1>` and every breadcrumb segment are Polish, with no English word left in the trail.
- In the column chooser, hide a column so it drops into "Dostępne kolumny": the group heading reads **KOLUMNY**, and on the customers pages the custom-field group reads **POLA NIESTANDARDOWE** — never "COLUMNS" or "CUSTOM FIELDS".
- **Worth a pass in German and Spanish too:** the whole `ui.columnChooser.*` block was translated in this PR, so the panel title, search box and "Ukryj wszystkie" equivalents should all be localized now. Only en and pl were driven automatically.

**What can go wrong**
- An edit page still shows "Edit Business Rule" / "Edit Rule Set" in the tab title or breadcrumb — the keys are new in the last commit and are the least-exercised part of the change.
- A group heading renders as a raw key (`ui.columnChooser.defaultGroup`) instead of a word, which means a dictionary miss rather than a fallback.
- Group headings that legitimately stay English (Basic Info, Contact, CRM, Profile, Socials, Notes, Pipeline) are **expected** — they come from explicit per-page column meta this PR deliberately does not touch. Do not fail the PR on those.

### P1 — Record-locks settings and order-create validation

**Where to click**
- `/backend/settings/record-locks` — needs `OM_ENABLE_ENTERPRISE_MODULES=true` *(covered — screenshot 4)*
- `/backend/sales/documents/create?kind=order` → press "Utwórz" with no line items *(covered — screenshot 5)*
- Any order → Historia tab, after changing a document-level field *(covered — screenshot 12)*

**What to verify**
- The strategy help alert renders "Optymistyczna:" and "Pesymistyczna:" on separate lines, and the browser console shows no hydration or DOM-nesting error on load.
- The lines-required message appears as exactly **one red error**. Note that the same sentence also appears twice more as non-error text (a grey section hint and the empty-state card) — that is existing `SalesOrderDraftLines` behavior, tracked separately in #5477, and **not** a reason to fail this PR.
- The history entry reads "Zmienione pola:" followed by a Polish field name, not an English camelCase one.

**What can go wrong**
- A hydration warning in the console on the record-locks page (check with devtools open from a cold load).
- Order-history entries showing raw field names like `customerSnapshot` or `grandTotalGross`.

### Perceived performance
- Cold-load `/backend/customers/people` and `/backend/sales/orders` and confirm the table shell and loading state appear before data, with no layout shift when the column chooser opens.
- Check the line-item dialog on a **mobile viewport** — the Status list now renders open whenever a value is set, so confirm it does not push the dialog's action buttons off-screen. Not exercised by the automated run, which used a 1440×900 desktop viewport only.

### Pass/fail

adeptofvoltron:

- 🚀 `merge-queue` — **set by this run**, replacing `review`. The re-review of `ba6b9dc6a` approved: every blocking finding from the previous pass is resolved, and CI is fully green (27 pass, 0 fail, 0 pending). Pipeline labels are mutually exclusive, so `review` was removed in the same transition.
- 🧪 `needs-qa` — **kept.** An approved PR that requires QA stays in `merge-queue` with `needs-qa` in place, and with `qaGate` on that combination blocks the actual merge until a maintainer adds `qa-approved`. Manual QA instructions are posted above; note the UI QA run on this exact head already passed 12/12 with screenshots, so most of that list is spot-checking rather than fresh work.
- ✅ `qa-approved` — **deliberately not applied.** Code-review approval is not QA approval, and this skill must never earn that label by reading a diff. It is a maintainer's call on the evidence already attached.
- 📸 `screenshots` — **kept**, recording that UI QA visual evidence is attached. Informational; it does not gate merge.
- 🐛 `bug` — **kept.** The PR fixes reported defects from the 0.7.0 pre-release sweep (#5456); nothing in the re-review changes that classification.
- ⚖️ `priority-low` — **kept.** These are UX and i18n nits with no data, security or availability impact; the fix commit did not raise the stakes.
- 💥 `risk-medium` — **kept, and it is the right call.** The last commit widened the change's blast radius rather than narrowing it: `LookupSelect.shouldSearch` is shared by every lookup in the product, so the fix reaches well beyond the one field that reported the bug. That is the correct fix, but it is exactly why this is not `risk-low`.
- ⏳ `in-progress` — held only for the duration of this re-review and removed in the completion comment below.

adeptofvoltron:

Re-review of `ba6b9dc6a` against the `changes_requested` verdict on `819f1cc9b`: the Major, both Minors and all four actionable Nits are resolved. The Major was fixed at the root in `LookupSelect` rather than by reverting `minQuery={1}`, so item 3's original fix survives and the same blind spot closes for every other `minQuery >= 1` lookup that resolves its selection through `fetchItems` — and it ships with a regression test that fails against the previous commit. The two Nits I said a follow-up would satisfy were genuinely filed (#5487, #5477), and the one declined item (`buildAdvancedFilterFieldsFromCustomFields`) was declined with reasoning I agree with.

No CI follow-up is owed and `ci-monitoring` was not applied: all 27 checks were already settled green on this head before the review was submitted, with nothing pending, and that is recorded in the review's Validation Gate table. Autofix: skipped (not my PR — re-run with `--autofix` to fix findings here), and nothing actionable remained to fix in any case.

The PR is now in `merge-queue` with `needs-qa` retained. It is **not** merge-ready yet: the QA-approval gate holds it until a maintainer adds `qa-approved`, and — since this account authored the review — a human approval is still the right bar before it lands.

Issue #4976, opened 2026-08-04
bug: no product path to send a Discord message — both hub outbound endpoints require an email recipient


Kapsik89:
Found while QA-ing #4391 (Discord channel provider) with a real Discord bot.

## Summary

There is no product path to send a message to Discord. Both hub outbound endpoints validate the recipient as an **email address**, and the third path only works as a reply inside an existing thread — which requires an inbound message (blocked by the sibling inbound issue).

The adapter itself sends fine; only the wiring is missing.

## Steps to reproduce

1. Connect a real Discord bot, filling in **Default channel ID**.
2. `POST /api/communication_channels/channels/{channelId}/test-send` with `{"to": "<discord channel id>", "body": "test"}`.

## Expected

Per #4391's description — *"Outbound works over the Discord REST API as soon as a bot token is connected"* — and per the `Default channel ID` help text — *"Default text channel for outbound sends and the test-send smoke test"* — the bot posts to `defaultChannelId`. This is also the spec's outbound smoke test (§ 6 "Test it").

## Actual

```
POST /channels/{id}/test-send  {"body":"…"}
→ 422  path:["to"]  "Invalid input: expected string, received undefined"

POST /channels/{id}/test-send  {"to":"1534331920463433771","body":"…"}
→ 422  path:["to"]  format:"email"  "Invalid email address"
```

Both endpoints are hard-wired to email:

```js
// api/post/channels/[id]/test-send/route.ts:32
const bodySchema = z.object({ to: z.string().email(), subject: …, body: … })

// api/post/send-as-user/route.ts:17
const bodySchema = z.object({
  to: z.array(z.string().email()).min(1),
  cc: z.array(z.string().email()).optional(),
  subject: z.string().min(1).max(500)…,     // ← required; Discord has no subjects
})
```

A Discord recipient is a **snowflake channel id** (`1534331920463433771`), not an email, and Discord messages have no subject.

The third and only remaining path — `commands/deliver-outbound-message.ts:403` — sends via `conversationId: mapping.externalThreadRef`, i.e. **only as a reply** in a thread created from an inbound message.

So the chain closes: **send → needs a thread → needs an inbound message → blocked.**

## The adapter is fine — this is a wiring gap

Replaying the exact REST contract that `discord-rest.ts → createMessage` builds (same URL, `Authorization: Bot …`, `content`, `allowed_mentions: {parse: []}`):

```
POST https://discord.com/api/v10/channels/1534331920463433771/messages
[...]

PR #5252, opened 2026-08-13
fix(messages): require externalEmail only for email-typed channels (#4975)

wojciechszyjka:
Closes #4975
Tracking plan: .ai/runs/2026-08-13-messages-conditional-external-email.md
Source doc: .ai/specs/2026-06-19-discord-communication-channel-integration.md
Status: complete

## 🎯 Goal

Implements **Variant A** of the hub sender-identity contract, decided by @pkarw on 2026-08-13 and
recorded on [#4975](https://github.com/open-mercato/open-mercato/issues/4975#issuecomment-5280211475).

Every message from an external sender is composed with `visibility: 'public'`, and
`composeMessageSchema.superRefine` made `externalEmail` (a real `z.string().email()`) mandatory for
every such message. That assumption only holds for email-typed channels. A Discord sender is a
snowflake with no address, so inbound Discord failed validation deterministically, the ingest job
retried three times and died in BullMQ — while the channel still reported `Connected` and the
integration still reported `Healthy`. Customer messages were lost silently.

This PR requires `externalEmail` only when the originating channel is email-typed, and **fails
closed**: an unknown, empty or absent channel type keeps the pre-existing rule, so no current caller
changes behaviour and the relaxation cannot be reached by a typo or an unfamiliar provider.

## What Changed

- **`packages/core/src/modules/messages/lib/channel-sender-identity.ts` (new)** — the single place
  that decides whether a channel's senders are reachable by email. It holds an explicit allow-list of
  recognized non-email channel types (`discord`, `slack`, `sms`, …) and returns "email required" for
  everything else, including unrecognized values. Adding a type here is additive.
- **`messages/data/validators.ts`** — `composeMessageSchema` gains an optional `sourceChannelType`,
  and the public-visibility branch of the refinement consults the helper before demanding an address.
  The schema was split into a base object plus a shared refinement so a second, client-facing variant
  could be derived without duplicating the rules. No new requirement is introduced in the waived
  case: for non-email channels the sender identity is already carried by
  `ExternalMessage.sender_identifier`, joined 1:1 to the platform message through `MessageChannelLink`.
- **`messages/api/openapi.ts`** — publishes `composeMessageRequestSchema`, the compose contract minus
  `sourceChannelType`, so the documented request shape matches what the route actually reads.
- **`communication_channels/commands/ingest-inbound-message.ts`** — forwards the channel type it
  already has to `messages.messages.compose`.
- **`messages/api/route.ts` (`POST /api/messages`)** — the second call site: without it the hub still
[...]

Issue #4975, opened 2026-08-04
bug: inbound Discord messages are rejected — hub requires externalEmail, Discord senders have none


Kapsik89:
Found while QA-ing #4391 (Discord channel provider) against a **real Discord bot on a real guild**, with a human typing a real message.

## Summary

Every inbound Discord message is rejected by the hub. `ingest_inbound_message` → the `messages` command fails validation with **`externalEmail is required when visibility is public`**. Discord senders have no email address, so this fails deterministically — 3 retries, then the job dies. **Inbound Discord messaging cannot work at all.**

## Steps to reproduce

1. Connect a real Discord bot (Connect Discord on profile → communication channels).
2. Enable `MESSAGE CONTENT INTENT` in the Discord Developer Portal; invite the bot to a guild.
3. `QUEUE_STRATEGY=async` + Redis running (with the default `local` the job vanishes even earlier — see the sibling issue).
4. `yarn mercato channel_discord start-gateway` → wait for `gateway ready`.
5. **A human** types a message in the connected Discord channel.
6. `select count(*) from external_messages;` and inspect the `communication-channels-inbound` queue.

## Expected

The message lands as `ExternalMessage` + `MessageChannelLink` and appears in the inbox; the sender resolves to a CRM contact (or a new external contact is created). This is the spec's own inbound smoke test (`.ai/specs/2026-06-19-discord-communication-channel-integration.md` § 6 "Test it").

## Actual

`external_messages = 0`, `messages = 0`, `message_channel_links = 0`. The job fails 3 times and lands in `failed`:

```
ZodError: [{ "code": "custom",
             "path": ["externalEmail"],
             "message": "externalEmail is required when visibility is public" }]
  at Object.execute (packages/core/dist/modules/messages/commands/messages.js:168)
  at Object.execute (…/communication_channels/commands/ingest-inbound-message.js:255)
  at async handle (…/communication_channels/workers/inbound-processor.js:41)

attempts made (atm) = 3   →   bull:communication-channels-inbound  failed=2
```

The gateway did its job correctly — the queued payload is complete:

```
content:   "Kolejna wiadomość testowa!@"
author:    lorakaspak / "Karol Kapsa"
author id: 1499156851487539260      ← a Discord snowflake, not an email
channel:   1534331920463433771
guild:     1534331919796273152
scope:     { tenantId: …, organizationId: … }
```

## Root cause

[...]

PR #4391, opened 2026-07-23
feat(channel-discord): Discord two-way communication channel provider

wojciechszyjka:
Tracking plan: .ai/runs/2026-08-24-discord-channel-provider.md
Status: complete

## Summary

Implements the **Discord channel provider** from spec `.ai/specs/2026-06-19-discord-communication-channel-integration.md`: a new workspace package `packages/channel-discord` (`@open-mercato/channel-discord`, module id `channel_discord`, provider key `discord`) under the existing `communication_channels` hub — **zero hub contract changes, zero new production dependencies, no schema change**.

Outbound works over the Discord REST API as soon as a bot token is connected; inbound arrives in real time through a Gateway (WebSocket) worker; a signed Ed25519 Interactions endpoint verifies and acknowledges slash commands and buttons.

**AI auto-reply ships too** (#4778): opt-in per channel, default OFF, answering straightforward messages directly and proposing everything else for a human to approve.

## What's inside

- **`DiscordChannelAdapter`** implementing the existing `ChannelAdapter` contract: outbound `sendMessage` via a `fetch`-based Discord REST client (channel resolution, 2000-char clamp, `defaultChannelId` fallback), fail-closed `verifyWebhook`, credential validation, health check, DI wiring, connect widget, i18n (en/pl/de/es).
- **Gateway worker** (`workers/discord-gateway.ts`) — a long-lived Discord Gateway client on Node's native `WebSocket`: identify/resume, heartbeat-ACK zombie detection, bounded reconnect backoff, close-code handling, resume-or-identify on `INVALID_SESSION` (honouring Discord's `resumable` flag), bot-self filtering. It bridges `MESSAGE_CREATE` and reaction dispatches into the hub's **existing** inbound/reaction queues. The refresh job is a **reconciler**: channels with a live session are left alone, channels that dropped out of the active set are closed, and a tenant-scoped run never touches another tenant's sockets.
- **Signed interactions route** `POST /api/channel_discord/interactions` — `metadata.path` is pinned explicitly (the way the hub's `webhooks/gmail` route does it), so the operator-facing URL is part of the route's contract rather than a by-product of the file location. Ed25519 verification is fail-closed via `node:crypto` (SPKI DER built from the raw 32-byte key, no `tweetnacl`); the freshness check answers before the per-candidate signature fan-out, so a replayed capture is rejected at constant cost. PING→PONG is handled synchronously (the generic hub webhook route 202-acks and therefore cannot serve Discord's handshake).
[...]

PR #5483, opened 2026-08-22
fix(communication_channels): let test-send reach a non-email channel (#4976)

wojciechszyjka:
## Summary

Closes the test-send half of #4976 — **no product path to send on a non-email channel**. Found by @Kapsik89 while QA-ing #4391 against a live Discord bot.

`POST /api/communication_channels/channels/[id]/test-send` declared `to: z.string().email()`. On a chat channel **no request body could reach the adapter at all**:

```
POST /channels/{id}/test-send  {"body":"…"}
→ 422  path:["to"]  "Invalid input: expected string, received undefined"

POST /channels/{id}/test-send  {"to":"1534331920463433771","body":"…"}
→ 422  path:["to"]  format:"email"  "Invalid email address"
```

QA proved **the adapters themselves are fine** — replaying by hand the REST contract that `discord-rest.ts → createMessage` builds returned HTTP 200 and the message appeared in the channel. The gap was entirely in the route's own validation. The consequence: `Default channel ID`, described in the UI as *"default text channel for outbound sends and the test-send smoke test"*, was unreachable from the product.

## Approach

What a valid recipient looks like is **a property of the channel**, so it cannot be decided by the body schema alone. The shape check stays in zod; the semantic one moves to `resolveTestSendRecipient` (`communication_channels/lib/outbound-recipient.ts`), which runs once the channel is loaded and reuses `channelTypeRequiresExternalEmail` — **the same fail-closed predicate that decides the inbound direction in #4975** — so one list governs both directions instead of two lists drifting apart.

| Channel type | `to` |
|---|---|
| email | required, and validated by the same `z.string().email()` the route used before |
| recognized non-email (`discord`, `slack`, `teams`, …) | provider-native identifier, **or absent** — the adapter then uses its configured default target |
| unknown / empty / missing | keeps the address requirement (fail-closed) |

**One ordering change, stated rather than smuggled in:** `to` is now validated after the channel is loaded and access-checked, so a caller who may not see the channel gets 404 instead of a 422 that would confirm the id parses. That is the safer order, not just a side effect.

## Deliberately out of scope

**`send-as-user` stays email-shaped.** The issue lists it as a second symptom, but every field it takes is a mail construct — required `subject`, `plain`/`html` MIME body, `inReplyTo`/`references` headers, `cc`/`bcc`. Making it channel-agnostic is a design task, not a validation relaxation, and the operator's real chat path does not run through it. Tracked separately in #5528, and **#4976 should not be closed until that lands** — this PR narrows it to the test-send path.

**Why the operator's real path is already open** — I checked rather than assumed. `messages.messages.reply` (`messages/commands/messages.ts:690`) does **not** go through `composeMessageSchema`: it creates the `Message` directly, copies `externalEmail` from the original (`null` for Discord) and emits `messages.message.sent`, which `communication_channels/subscribers/outbound-bridge.ts` picks up and queues for delivery. The client-side `isValidEmailAddress` gate lives in `useComposeSendOperation` / `useSendDraftOperation`, **not** in `useReplySubmitOperation`. So the only blocked action was *starting a new* public conversation from the hub — which is not a meaningful product action on a chat channel, since there is no way to name the target. That is also why I am not touching the composer, per @pkarw's recommendation on #5252 (minor 2). I will verify this empirically against the live bot during #4391 QA and report the result.

## Tests

- **Unit** (`lib/__tests__/outbound-recipient.test.ts`): the recipient rule in both directions, including **the exact snowflake QA sent**, equivalence with zod's validator on the email branch, and fail-closed behaviour for unknown / empty / typo'd channel types.
- **Integration** — `TC-CHANNEL-OUTBOUND-001`, two cases against a **connected** channel of each flavor: a chat channel accepts a send with no recipient and with a native identifier; an email channel still rejects a non-address and a missing recipient, and still accepts a valid address.

Why a new spec when `TC-CHANNEL-EMAIL-HUB-001` already has a *"POST test-send rejects invalid body with 422"* case: that one posts to an all-zeros channel id and accepts `[401, 404, 422]`, so it never reaches a point where a channel type exists. It could not have caught this, and it still passes after the change.

**The spec's ceiling, stated plainly:** the stub adapter reports a successful send with no network I/O, so the assertion is that the route accepts the request, resolves the recipient and **reaches the adapter**. That a real provider then posts the message is the adapters' own contract, unit-tested in their packages.

## Dependency

**Stacked on #5252** (`cez/84f5dded`) — it uses `channelTypeRequiresExternalEmail`, which that PR introduces. Please merge #5252 first; the diff here then shrinks to this change's own files.

## Validation

Local runner: `yarn typecheck` 26/26 · `@open-mercato/core` communication_channels 698/698 · unit tests for the new helper.

**On the integration spec:** `ephemeral-integration` is green on CI but **did not run these tests** — nothing in `.github/workflows/` sets `OM_ENABLE_TEST_CHANNEL_SEEDING`, so the whole seeded family opts out via `test.skip`. Thanks to @adeptofvoltron for catching that the check I pointed at was not the check I needed. Result of the local run with that env set is posted in the review-response comment below.


adeptofvoltron:
your claude went into polish mode @wojciechszyjka (been there)

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-08-22T15:23:50Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron (changes requested):
# 🔍 Code Review: fix(communication_channels): let test-send reach a non-email channel (#4976)

## 🎯 Summary

This PR fixes a real, well-diagnosed product bug: `POST /api/communication_channels/channels/[id]/test-send` declared `to: z.string().email()`, so on a chat-typed channel no request body could ever reach the adapter — omitting `to` produced a 422 for a missing string, and a Discord snowflake produced a 422 for an invalid address. The documented `Default channel ID` smoke test was therefore unreachable from the product.

The design is the right one. Whether a recipient is an address is a property of the channel, not of the request body, so the shape check stays in zod (`api/post/channels/[id]/test-send/route.ts:41`) and the semantic check moves to `resolveTestSendRecipient` (`lib/outbound-recipient.ts:35`), which runs after the channel is loaded and access-checked (`route.ts:160`). Reusing `channelTypeRequiresExternalEmail` means one fail-closed list governs both the inbound (#4975) and outbound directions instead of two lists drifting apart, and an unrecognized, empty or absent channel type keeps the address requirement, so no existing caller changes behavior. I verified the reuse lands where the author says it does: the chat test-seed stub declares `channelType = 'discord'` (`lib/test-seed.ts:148`) and the Discord spec fixes the real provider on the same value (`.ai/specs/2026-06-19-discord-communication-channel-integration.md:303`), so the predicate actually recognizes the channel the bug was filed against. The reordering that moves body validation behind the channel lookup is a genuine improvement, not just a side effect — a caller who may not see the channel now gets 404 instead of a 422 that confirms the id parses.

I also checked the two things that a validation relaxation most often breaks, and neither is broken: header injection is not reachable through the email path (`.trim()` strips a trailing CR/LF and `[^\s@]+` cannot match an embedded one), and the email branch still refuses an empty or whitespace-only recipient. The scope decision to leave `send-as-user` alone is defensible on inspection — every field of that schema is a mail construct (`api/post/send-as-user/route.ts:20-31`: required `subject`, `plain`/`html` MIME body, `inReplyTo`/`references`, `cc`/`bcc`), so making it channel-agnostic is a design task rather than a validation relaxation.

Reviewed scope: the four files of this PR's own commit `4985593`. The PR is stacked on #5252 (`cez/84f5dded`), so the 26-file diff against `develop` mostly belongs to that PR; I reviewed #5252's contents only where this change depends on them.

## Verdict

❌ **request changes** — there are no blockers and no majors; the fix is sound and I would be comfortable with it landing after a small pass. Under this repository's stricter verdict rule (Medium/minor findings request changes) the five minor findings below drive the verdict. The two that matter are a measurable loosening of the email rule that the code, the commit message, and the tests all describe as unchanged (M1), and the fact that the integration spec this PR leans on was skipped in CI and so has never actually run (M2).

## 🧪 Validation Gate

Validation source: **GitHub checks** (head `4985593`, run [32576988820](https://github.com/open-mercato/open-mercato/actions/runs/32576988820)). All reported checks are green and none are pending, so no local fallback run was needed.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job, "Build packages" step. |
| `yarn i18n:check-sync` | ⏭️ NOT RUN | Not part of the CI workflow; the diff adds no locale file and no user-facing string, so there is nothing for it to check here. |
| `yarn typecheck` | ✅ PASS | `test` job. |
| `yarn test` | ✅ PASS | `test` job — includes the 13 new `resolveTestSendRecipient` cases. |

Additional green checks: `lint`, `ds-lint`, `audit-scope`, `docker-build`, `documents-multi-instance`, `CodeQL`, `license/cla`. The `ephemeral-integration (none)` check is green, but see M2 — it reports `100 passed / 10 skipped`, and both of this PR's new cases are in the skipped set.

## Findings

### 🔹 Minor

**M1 — `packages/core/src/modules/communication_channels/lib/outbound-recipient.ts:24` — the hand-rolled `EMAIL_PATTERN` is materially looser than the `z.string().email()` it replaces, while the code claims the opposite.**

The doc comment at `lib/outbound-recipient.ts:30` says the email branch keeps "the pre-#4976 contract, byte for byte", the commit message repeats it, and the unit test names its block "email-typed channels keep the pre-#4976 contract" (`lib/__tests__/outbound-recipient.test.ts:4`). It is not the same rule. I compared both against the repo's zod (4.4.3), and every one of these inputs was rejected with 422 before this PR and is accepted now: `a@b.c`, `a@b..c`, `a@.b.c`, `.a@b.com`, `a..b@c.com`, `a@b_c.com`, `a@-b.com`, `a@b.c.`, `a@b#c.d`. The practical effect is small but wrong in a predictable way: on an email channel, a malformed address now travels to the SMTP/Gmail adapter and comes back as a 502 provider error instead of a clean 422 from the route, which is exactly the diagnostic clarity this endpoint exists to give. It also makes the invariant the PR is built on ("mailboxes are untouched") false in a way a future reader will trust.

Fix: keep zod as the authority inside the resolver rather than re-deriving it — `if (!z.string().email().safeParse(trimmed).success) return { ok: false, error: 'Invalid email address' }` at `lib/outbound-recipient.ts:43` — and add one case to the unit test (`a@b..c` is enough) so the equivalence is asserted rather than described. If a hand-rolled regex is preferred for some reason, then the comment, the commit message, and the test title should say the rule was loosened and why.

**M2 — `packages/core/src/modules/communication_channels/__integration__/TC-CHANNEL-OUTBOUND-001.spec.ts:68` — the new integration spec did not run in CI, so the coverage this PR points at has never executed anywhere.**

The PR body says the local harness run was killed twice by SIGTERM and that `ephemeral-integration` on CI is therefore the authoritative signal for `TC-CHANNEL-OUTBOUND`. That check is green, but it did not run these tests. Run [32576988820](https://github.com/open-mercato/open-mercato/actions/runs/32576988820) reports `100 passed, 10 skipped`, and the Playwright listing marks both of this PR's cases as skipped, together with `TC-CHANNEL-IDENTITY-001` and the whole `TC-CHANNEL-API-001…007` family. The cause is the `test.skip` guard at lines 68 and 116: nothing in `.github/workflows/` sets `OM_ENABLE_TEST_CHANNEL_SEEDING`, so `isChannelSeedingAvailable` is false in that lane and every seeded spec opts out silently. A spec that has never executed is also unverified as a spec — for instance nothing has yet confirmed that `seedConnectedChannel(..., { providerFlavor: 'chat' })` produces a channel the route accepts end to end, even though the code reads as though it does.

Fix: either run it once locally with `OM_ENABLE_TEST_CHANNEL_SEEDING=true` and paste the result into the PR, or enable that env in the ephemeral lane so the whole seeded family starts running. The second is the better outcome and is worth its own issue if it is out of scope here; what should not stand is the PR resting on a check that skipped the test.

**M3 — `.ai/specs/2026-06-19-discord-communication-channel-integration.md:16` and `:709` — the spec still records #4976 as deferred and open.**

That spec is the decision record for the four QA blockers found against the live Discord bot. Line 16 reads "#4976 … **Deferred by this spec, tracked there.**" and line 709 lists #4976 under "Still open, and not addressed by this decision". This PR resolves the outbound-recipient half, and root `AGENTS.md` asks for specs to be updated when a significant change lands — the sibling change did exactly that for #4975 in commit `298d518`. Leaving it as-is means the decision record now contradicts the code.

Fix: update both places to record that the test-send half of #4976 is decided and implemented (same shape as the #4975 entry), and state explicitly that the `send-as-user` half remains open.

**M4 — scope versus the issue: #4976 names both outbound endpoints, and only one is fixed.**

The deferral of `send-as-user` is well argued and I agree with it on the merits. The risk is bookkeeping rather than code: the issue's title and body cover both endpoints, so closing it on this PR would silently drop the second symptom. (The body's "Zamyka #4976" is not a GitHub closing keyword, so nothing auto-closes today — that is luck rather than design.)

Fix: open the follow-up issue for the `send-as-user` / composer half, link it from this PR and from #4976, and say in the PR body that #4976 closes only once that follow-up exists or that #4976 is being narrowed to the test-send path.

**M5 — the PR description is written in Polish** (already raised by @adeptofvoltron in [this comment](https://github.com/open-mercato/open-mercato/pull/5483#issuecomment-5381139688)).

The repository's working language is English — every other PR body, the commit message on this very branch, the specs and the issue are all in English. The description is genuinely good, which is the point: it is the reasoning most reviewers will lean on, and right now most of them cannot read it.

Fix: rewrite the description in English; the commit message is already an accurate basis for it.

### 💅 Nit

**N1 — `packages/core/src/modules/communication_channels/lib/outbound-recipient.ts:46` — the free-form recipient accepts embedded CR/LF.** For a recognized non-email channel, `to` only passes through `z.string().min(1).max(320)` (`route.ts:41`) and `.trim()`, so `"123\n456"` reaches the adapter intact. Today's chat transports are JSON REST APIs where that is inert, which is why this is a nit and not a finding — but `api/post/send-as-user/route.ts:26` deliberately rejects CR/LF at the edge for exactly this class of input, and matching that precedent costs one `.regex(/^[^\r\n]*$/)`.

**N2 — `packages/core/src/modules/communication_channels/api/post/channels/[id]/test-send/route.ts:232` — `metadata` now carries an explicit `to: undefined` key.** An adapter that feature-detects with `'to' in metadata` rather than a truthiness check will see the key present and empty. Spreading conditionally (`...(recipient.to ? { to: recipient.to } : {})`) removes the ambiguity.

**N3 — `packages/core/src/modules/communication_channels/lib/outbound-recipient.ts:1` — the shared predicate now governs both directions but still lives in, and is named for, the inbound one.** `channelTypeRequiresExternalEmail` in `messages/lib/channel-sender-identity.ts` reads oddly at an outbound call site, and the cross-module import matches existing precedent in this module (`data/enrichers.ts`, `subscribers/outbound-bridge.ts`), so nothing is wrong here. As the third consumer arrives it is worth promoting the channel-type list to a neutral home with a direction-agnostic name.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the change is purely additive (`resolveTestSendRecipient` is new).
- [x] No field removed or retyped in an existing response shape — the 200 body is unchanged.
- [x] Where a contract had to change: the request contract only widens. `to` becomes optional and, on non-email channels, free-form; every previously accepted email-channel request is still accepted, and the previously rejected ones that are now accepted are the M1 regex-divergence set. Per `BACKWARD_COMPATIBILITY.md` this is an additive change to an API request surface and needs no deprecation window. The one behavioral reordering — body validation moved behind the channel lookup, so an unauthorized caller sees 404 rather than 422 — narrows what an unauthorized caller learns and is documented in the commit message.

## 🧪 Test Coverage

The unit coverage is genuinely good and ran green in CI. `lib/__tests__/outbound-recipient.test.ts` walks the whole rule matrix: the email branch accepts an address and rejects a missing, whitespace-only and non-address recipient; the non-email branch accepts the exact Discord snowflake QA sent, accepts no recipient at all (the documented `Default channel ID` smoke test), and trims what it takes; and the fail-closed branch is parameterised over `undefined`, `null`, `''`, `'   '` and the typo cases `'e-mail'` / `'discrod'`, which is precisely where a fail-open bug would hide.

Two gaps:

1. The integration spec asserts nothing anywhere today (M2). It is well written — a connected channel of each flavor, both directions of the contract, cleanup in `finally` — and it should be made to run rather than trimmed.
2. No case pins the email branch to zod's definition (M1). One assertion that `a@b..c` is rejected would have caught the divergence and would keep catching it.

Route-level coverage is unchanged, and I confirmed the pre-existing `TC-CHANNEL-EMAIL-HUB-001` case ("POST test-send rejects invalid body with 422") still passes: it posts to an all-zeros channel id and accepts `[401, 404, 422]`, so it never reaches a channel type — the author's reasoning about why it could not have caught this bug is correct.

---

*Reviewed by `om-auto-review-pr`. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).*

adeptofvoltron:

- ❌ `changes-requested` — the review found five actionable Medium findings: the hand-rolled email regex is looser than the `z.string().email()` it replaces while the code claims byte-identical behavior, the new integration spec was skipped in CI and has therefore never run, the Discord spec still records #4976 as deferred, the `send-as-user` half of the issue needs a linked follow-up before #4976 closes, and the PR description is in Polish. No blocker or major finding was found; this repository requests changes on Medium findings too.
- 🐛 `bug` — the change fixes a defect found during QA of #4391: the documented outbound smoke test was unreachable on any non-email channel.
- ⏭️ `skip-qa` — kept as the author set it, and it holds under the automated-verification exemption: no UI-rendering file is touched, the database schema and the API surface are unchanged, and the changed behavior ships with unit tests that ran green in CI. The end-to-end proof against a live Discord bot belongs to the QA session for #4391.
- 🔺 `priority-high` — kept: this is a release-blocking regression in the only product path that can smoke-test an outbound chat channel, and #4976 is one of four blockers filed against the live Discord integration.
- 🟡 `risk-medium` — kept: an ordinary single-module change that ships with tests, though it does relax input validation on a route shared with every email channel, which is why it is not `risk-low`.

adeptofvoltron:

adeptofvoltron:

Five Medium findings, no blocker and no major: the hand-rolled email regex accepts nine shapes `z.string().email()` rejected (contradicting the "byte for byte" claim), `TC-CHANNEL-OUTBOUND-001` was skipped in the green `ephemeral-integration` run so it has never executed, the Discord spec still lists #4976 as deferred, the `send-as-user` half of #4976 needs a linked follow-up, and the description is in Polish. Validation source: GitHub checks (head `4985593`, all green, none pending) — no CI follow-up is pending, so no `ci-monitoring` label. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

wojciechszyjka:
🤖 Claiming this PR — starting `om-auto-continue-pr` run. Started: 2026-08-23T09:26:55Z.

wojciechszyjka:
🤖 `om-auto-continue-pr` — this PR is largely superseded, stopping before I rework it

I picked this up to work through @adeptofvoltron's five Medium findings, but merging the current `develop` surfaced something that changes what "finishing" means here, so I stopped rather than push a rebase.

**#5261 — "fix(communication_channels): validate test-send recipients per provider (#4976)" — merged into `develop` at 2026-08-23T01:57Z** and fixes the same issue this PR targets. It adds the same file path, `communication_channels/lib/outbound-recipient.ts`, which is why the merge lands as an add/add conflict on that file, its test, and the test-send route.

The merged version also happens to resolve three of the review findings on this PR by construction:

- **M1** — it keeps `z.string().email()` as the authority instead of a hand-rolled regex, so there is no divergence to fix.
- **N1** — provider-native recipients go through an allowlist (`/^[A-Za-z0-9._:@+-]+$/` plus an explicit `..` rejection), which is stricter than the CR/LF guard suggested here.
- **N3** — the rule is driven by the adapter's declared `capabilities.recipientFormat` rather than by reusing the inbound `channelTypeRequiresExternalEmail` predicate, so the "shared predicate lives in, and is named for, the inbound direction" smell does not arise.

**What is genuinely still missing on `develop`**, as far as I can see:

1. `test-send` still requires a recipient — `to: z.string().min(1).max(MAX_OUTBOUND_RECIPIENT_LENGTH).regex(/^[^\r\n]*$/)`. #5261 made a Discord snowflake acceptable, but omitting `to` entirely is still a 422. The documented `Default channel ID` smoke test ("default text channel for outbound sends and the test-send smoke test") therefore remains unreachable — that was one of the two symptoms this PR set out to fix.
2. `TC-CHANNEL-OUTBOUND-001` does not exist on `develop`. This PR's copy is written against the superseded API and, per M2, never actually ran in CI because nothing sets `OM_ENABLE_TEST_CHANNEL_SEEDING` in the ephemeral lane.
3. M3 (the Discord spec still lists #4976 as open), M4 (no follow-up issue for the `send-as-user` half), and M5 (this description is in Polish) are all untouched by #5261.

So the useful residue is narrow and quite different from what this branch currently contains: make `to` optional when the adapter's `recipientFormat` is `provider-native`, plus the integration coverage and the bookkeeping. Rebasing the current implementation on top of #5261 would mean re-litigating a design that has already landed in a cleaner form.

I have released the `in-progress` label and made no commits. Holding for a decision on whether to close this in favour of a small follow-up PR, or to rework it in place down to the residual gap.

wojciechszyjka:
## Closing this — superseded by #5261, which does the same job better

#5261 landed on `develop` at 01:57 today and fixes #4976's test-send half. I closed this PR rather than reworking it, because their implementation is better than mine in a way that matters, not merely different.

**What theirs catches that mine let straight through.** `validateOutboundRecipient` keys off the adapter's declared `capabilities.recipientFormat` and applies an **allowlist** to a provider-native recipient. The reasoning in their comment is the part I missed entirely: the recipient is interpolated into a REST path — Discord posts to `/channels/{recipient}/messages` — so a permissive recipient is a **path-steering primitive**. My `resolveTestSendRecipient` only rejected CR/LF, so `C123%2Fmessages`, `C123%2e%2e%2fusers%2f@me`, `..`, a NUL byte and a full `https://discord.com/...` URL would all have reached the adapter. Their length ceiling and schema-level CR/LF rejection (before the channel lookup rather than after) are right too.

Keying off adapter capabilities rather than a channel-type list is the better axis as well: the adapter declares its own recipient shape, instead of a second list in the hub drifting against `channelTypeRequiresExternalEmail`.

We reached the same conclusion independently on `send-as-user` — that widening its schema alone would move the 422 one layer deeper and let the endpoint advertise a capability it does not have. That half is now tracked in **#5528**, and **#4976 should not be closed until it lands**.

**One narrow gap outlives both PRs**, filed as a note on #4976 rather than as more code here: `to` is required on `develop`, so the smoke test the spec actually documents — *"the bot posts to `defaultChannelId`"* (§ 6 "Test it") — still has no request shape. The Discord adapter does implement the fallback (`resolveTargetChannelId(input, credentials.defaultChannelId)`), so `Default channel ID` remains a credential field an operator fills in at connect time that no product path reads unless an inbound thread already exists. Much smaller than the bug this issue was filed for, and now the only part of the test-send half still open.

**What survives from here:** #5528, and @adeptofvoltron's review findings that were about my code specifically — M1 (my hand-rolled email pattern was measurably looser than the `z.string().email()` it replaced, while three separate places claimed it was identical) was a real defect and a fair catch. Thanks for the review; sorry it landed on code that turned out to be redundant.

wojciechszyjka:
Correction to this PR's description, for the record even though it is closed: I claimed the operator reply leg already worked without an address and that only the client-side email gate blocked it. Verified against a live Discord bot — **false**. `POST /api/messages/{id}/reply` returns 403 for every operator, because an inbound channel message carries a synthetic sender id and zero recipients and `replyMessageCommand` guards on sender-or-recipient. Filed as **#5535** with the measurements; the correction is also on #5252.

PR #5489, opened 2026-08-22
fix(eudr): EUDR nav label, module breadcrumb, and false unsaved-changes prompt

patzick:
Branched from and targeting `main` directly.

## What

Three EUDR admin navigation/UX fixes.

### 1. Sidebar label

The module's Overview entry now reads **EUDR Overview** instead of a bare "Overview", so it is unambiguous next to other modules' overview pages. Translated in all five shipped locales.

### 2. Module breadcrumb

Every EUDR page now starts with an **EUDR** crumb linking to `/backend/eudr`:

```
Home → EUDR → Plots
Home → EUDR → Statements → Create
```

The overview page itself keeps a single `EUDR Overview` crumb rather than linking to itself.

### 3. False "You have unsaved changes" prompt

Opening an EUDR edit page (Plot, Evidence submission, Product mapping, Statement) and leaving without touching anything raised **"You have unsaved changes. Are you sure you want to leave?"**.

Root cause: `LookupSelectField` emitted `onSnapshot` for the value the form was *hydrated* with. A snapshot is a denormalization captured when the user picks a record, so re-emitting it on load pushed a freshly fetched snapshot into the form via `setFormValue`, overwriting the record's stored one. Whenever the two differed — rows imported or seeded without a snapshot, or a company/product renamed since the record was created — `CrudForm`'s dirty check saw a changed value and armed the navigation guard.

The hydration resolve is now suppressed. User-driven picks are unaffected: `LookupSelect`'s own `onChange` already calls `onSnapshot` directly, and the effect still fires for any selection that is not the hydrated one.

Side effect worth noting: opening and saving a record no longer silently rewrites its stored supplier/product/order snapshot. The snapshot now changes only when the user actually changes the selection, which is the intended semantics.

## Testing

New `packages/core/src/modules/eudr/components/__tests__/LookupSelectField.test.tsx` covers all three paths:

- hydrating an existing value does **not** emit a snapshot (fails on `main`, verified by reverting the guard)
- picking a different record **does** emit its snapshot
- a create flow starting with no value **does** emit on first pick

Gates run locally (local runner, not Docker):

- `yarn generate` — clean
- `packages/core` `tsc --noEmit` — clean
- `packages/core` jest — **1372 suites / 10863 tests passed**
- `yarn i18n:check-sync` — all locales in sync
- `eslint` on the touched files — one pre-existing `react-hooks/globals` error at `formConfig.tsx` (`referencedStatementRowKeySeq`), present on `main` and outside this change

## Scope

No schema, API surface, event ID, widget spot ID or DI change. Touches `.tsx`, so `needs-qa` applies (the automated-verification exemption does not).

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5489"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790008384&installation_model_id=432243&pr_number=5489&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5489&signature=5f7e035b2929cbd91b840391bc7bde1305f1ab0cc37c8c923ca29cac7b0a366a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


patzick:
🤖 Claimed by `om-auto-review-pr` (automated review). Relying on GitHub CI checks per the repo's checks-first rule; will apply fixes for any findings and report the verdict here.

patzick:

**Verdict: approved** · **Validation source: GitHub checks** (head `591e4ae`) · Base `main`

### Validation

All reported checks are green on the current head; nothing pending, nothing failing. No local fallback was needed.

| Check | Result |
|---|---|
| `test` | pass |
| `lint` / `ds-lint` | pass |
| `docker-build` | pass |
| `ephemeral-integration (none)` | pass |
| `prepare` / `scope` / `audit-scope` | pass |
| `CodeQL` + `Analyze (actions / javascript-typescript / python)` | pass |
| `license/cla` | pass |
| `audit`, `mutate`, `comment`, `merge-coverage`, `[code]smith` | skipping |

`audit` is `skipping` while `test` reports an explicit **pass**, so the test gate carries real signal here — this is not the "failing audit hides the test gate" pattern.

### Scope reviewed

| Layer | Files |
|---|---|
| Backend page metadata | 16 × `backend/eudr/**/page.meta.ts` |
| Shared form component | `components/formConfig.tsx` (`LookupSelectField`) |
| i18n | `i18n/{en,de,es,pl,ko}.json` |
| Test | `components/__tests__/LookupSelectField.test.tsx` (new) |

No API route, entity, migration, validator, ACL, event, DI, subscriber, worker or search-config surface is touched. No `BACKWARD_COMPATIBILITY.md` contract surface changes: `LookupSelectField`'s prop signature is unchanged and the fix is purely behavioural.

### Correctness review of the fix

The `hydratedValueRef` guard was checked against the failure modes that usually break this pattern:

- **Re-render churn** — `onSnapshot` is an inline arrow at every call site, so the effect re-runs on every render. The guard returns *before* clearing the ref, so a stable selection stays suppressed indefinitely instead of leaking an emit on render 2. Correct ordering.
- **StrictMode double-invoke** — refs survive the simulated remount and the early return runs first, so the dev double-effect does not emit either.
- **Ref seeded too early** — `useRef(selectedValue)` only captures the first render. Verified all five edit pages (`plots`, `evidence-submissions`, `product-mappings`, `risk-assessments`, `statements`) early-return on `loading` before rendering `CrudForm`, so the picker always mounts with the value already populated. `PlotImportDialog` and the create pages start at `null`, where the guard is a no-op.
- **User-pick path still works** — `LookupSelect`'s own `onChange` calls `onSnapshot` directly, so selection changes are unaffected; the effect additionally covers the case where the picked option was not yet in `knownOptionsRef`.
- **Clear-then-repick** — `onChange` emits directly, so suppressing the effect for the hydrated id loses nothing.

Completeness of "same for other pages in module" was verified rather than assumed: `PlotMultiSelectField`'s supplier-reset effect is ref-guarded against firing on mount, `ReferencedStatementsField` and `RiskCriteriaField` only write from user actions, `ComboboxInput`/`CountrySelectField` never call `onChange` from an effect, and `visibleWhen` only affects rendering/validation (`hiddenBaseFieldIds` never writes `values`). `risk-assessments/[id]` has no snapshot picker and never had the bug. The four pages that did are exactly the four fixed.

### Findings

| # | Severity | File | Finding | Status |
|---|---|---|---|---|
| 1 | Low (nit) | `backend/eudr/*/[id]/page.meta.ts` | Edit pages ended their breadcrumb at the list crumb, so `AppShell` rendered "Plots"/"Statements" as the current page and dropped its `href` — a dead, misleading crumb while you are actually on the edit form. Create pages already carried a leaf. | **Fixed** in `591e4ae` — added the edit-title leaf, so the trail reads `Home → EUDR → Plots → Edit plot` and the list crumb is clickable again. Reuses existing `*.edit.title` keys (verified present in all five locales). |
| 2 | Low (nit) | `components/__tests__/LookupSelectField.test.tsx` | `data-value` on the `LookupSelect` mock was set but never asserted. | **Fixed** in `591e4ae` — removed. |

No Critical, High, or Medium findings. Under this repo's stricter verdict rule (Medium ⇒ changes requested), Low-only findings ⇒ **approved**.

### Behaviour change worth knowing at QA time

Opening and saving an EUDR record no longer silently rewrites its stored supplier/product/order snapshot — the snapshot now changes only when the selection actually changes. That is the intended semantics, but it does mean records imported or seeded **without** a snapshot are no longer incidentally backfilled by opening and saving them. Backfilling those belongs in the import path, not in a form-hydration side effect.

### Labels

Moving `review` → `merge-queue` and releasing `in-progress`. **`needs-qa` stays on**: the diff touches `.tsx` under `**/components/**`, so the automated-verification exemption does not apply and this PR still needs manual QA.

⚠️ `merge-queue` here is routing state only. Per the QA-approval merge gate this PR **MUST NOT be merged** until a QA reviewer exercises the four edit pages and applies `qa-approved`. I have not applied `qa-approved` and did not perform self-QA — no local run-through was done, so there is no evidence to attach.

PR #5516, opened 2026-08-22
fix(eudr): accept a no-op referenceIssuedAt on statement create (#5508)

### LINKED ISSUE #5508: bug: [EUDR] referenceIssuedAt is rejected on create, including an explicit null
## Summary
The published OpenAPI create schema for `POST /api/eudr/statements` lists `referenceIssuedAt` as an accepted
property (`null` or a date-time). In practice **every** create request carrying that key is rejected — including
`referenceIssuedAt: null`, which is the value the record takes anyway.

| Value sent | Response |
|---|---|
| `null` | `400 {"error":"eudr.errors.referenceIssuedAtImmutable"}` |
| `"2026-08-01T10:00:00.000Z"` | `400 {"error":"eudr.errors.referenceIssuedAtImmutable"}` |
| key omitted entirely | **201** |

## Expected
Either drop the field from the documented create schema, or accept a no-op value such as `null`. Rejecting a key
whose value equals the default makes the contract self-contradictory, and "immutable" is the wrong word for a
create.

## Impact
Low in the product today — the backend create form does not render this field. It matters for API consumers that
serialize a whole object, a very common client pattern, which get a 400 with a message that does not describe
what went wrong.

## Side observation
`referenceIssuedAt: "2026-08-01"` (date without time) returns a raw Zod dump into the response body, including
the full ISO-8601 validation regex.

haxiorz:
Closes #5508

## Goal

`POST /api/eudr/statements` accepts a whole-object payload — the "serialize a whole object and POST it back" client pattern from the issue — and returns 201, instead of a self-contradictory `400` on a server-managed key.

## Problem and root cause

The published OpenAPI create schema listed `referenceIssuedAt` as an accepted property, but the create command threw `eudr.errors.referenceIssuedAtImmutable` whenever the parsed payload carried the key — including `referenceIssuedAt: null`, exactly the value a new draft takes. `referenceIssuedAt` is server-managed (only set at the `submitted→available` transition). The same structural bug also affected `submittedAt`: `GET` returns it on every row, and `statementCreateSchema`'s `rejectServerComputedFields(['submittedAt'])` rejects on **key presence alone** (regardless of value), so a client that read a draft and POSTed the row back got a `400 eudr.errors.serverComputedField` on `submittedAt: null`. Either way, a whole-object round trip failed on a server-managed field echoed back with its no-op value. A date-only string additionally leaked a raw Zod error (the full ISO-8601 regex) via the shared CRUD factory.

## What changed

- `statementCreateSchema` drops `referenceIssuedAt` (`statementBaseSchema.omit(...)`) — the create request contract no longer advertises it and any echoed value is ignored; it stays a valid input on `statementUpdateSchema`, still immutable after issue.
- The create command's dead `referenceIssuedAtImmutable` throw is removed.
- `rejectServerComputedFields` now tolerates a `null`/`undefined` no-op echo and rejects only a meaningful (non-null) attempt to set server-owned state. This closes the whole class across the module's entities (`submittedAt`, `completenessScore`, etc.), so a whole-object round trip succeeds while `completenessScore: 55`/`overallTier: 'low'` stay rejected exactly as `TC-EUDR-002`/`TC-EUDR-006` assert.
- Added a schema-level jest regression (`referenceIssuedAt` and `submittedAt` no-op echoes; non-null server-computed value still rejected; update schema still accepts `referenceIssuedAt`) and a `TC-EUDR-007` whole-object round-trip integration case (GET a draft, POST it back verbatim minus `id`, expect 201).
- Recorded the create-schema narrowing waiver in the eudr spec changelog per `BACKWARD_COMPATIBILITY.md`.

## Validation

- `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn build:app` — all pass locally.
- `yarn test` — `@open-mercato/core` fully green (incl. the schema regression). Two packages fail only in the local monorepo run and pass in CI: `open-mercato-docs` (a test that mishandles a space in the local checkout path via `URL.pathname`→`%20`) and `create-mercato-app`'s agent-harness-evaluator (a fake-runner subprocess resource flake under full-suite concurrency). Neither touches eudr; CI validates `yarn test` on a space-free path.
- Regression observed red before the fix on both the `referenceIssuedAt` create path and the new `submittedAt: null` no-op echo; green after.

## Automated review

- `om-code-review` — ✅ approve, no findings.
- Addressed the maintainer review (`@pkarw`, changes requested): extended the fix to `submittedAt` so the whole-object goal is genuinely met, added the round-trip test, recorded the backward-compatibility trace, and documented the two server-managed-field conventions inline.
- Unresolved findings — None.

## Breaking changes

None. The create **request** schema narrows by dropping a field that never functioned on create (it always returned 400); request validation only relaxes, no response field is removed, and no client that previously succeeded can break — so no deprecation protocol is required. The waiver is recorded in the eudr spec changelog.

## Labels and delivery

- `review` — Ready non-draft PR entering the review pipeline.
- `bug` — Fixes a defect in the documented create contract.
- `skip-qa` — API-only change: no `.tsx`/UI/`components/` files, DB structure unchanged, no `BACKWARD_COMPATIBILITY.md` contract broken, and automated tests for the changed behavior ship in the same PR — so it qualifies for the automated-verification exemption and does not gate on manual QA.
- `priority-low` — Low product impact today (the backend form does not render these fields); it matters for API consumers serializing whole objects.
- `risk-low` — One module's create-schema/command plus a null-tolerance relaxation of an existing guard; validation only relaxes, so blast radius is minimal.

## Manual QA

Not required (see `skip-qa` rationale). If verifying by hand: `GET /api/eudr/statements?ids=<draft-id>`, then `POST /api/eudr/statements` with that row verbatim minus `id` — it returns 201 and the new statement is a draft with null `referenceIssuedAt`/`submittedAt`.


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-23T00:14:00Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(eudr): accept a no-op referenceIssuedAt on statement create (#5508)

## 🎯 Summary

This PR resolves the self-contradictory create contract on `POST /api/eudr/statements`, where `statementCreateSchema` advertised `referenceIssuedAt` as an accepted property while `createStatementCommand` rejected any request that carried the key — including `referenceIssuedAt: null`, the exact value a new draft takes. The fix drops the field from the create schema (`statementBaseSchema.omit({ referenceIssuedAt: true })`) and deletes the now-dead `referenceIssuedAtImmutable` throw from the create command, so the published OpenAPI create schema and the runtime validator are generated from the same source and can no longer drift.

I reviewed all four changed files, plus the surrounding contract surface: `packages/core/src/modules/eudr/data/validators.ts`, `commands/statements.ts`, `api/statements/route.ts` (the OpenAPI wiring), the sibling `TC-EUDR-002`/`TC-EUDR-006` server-computed-field assertions, and `.ai/specs/2026-07-06-eudr-compliance-module.md`.

The engineering here is good and I want to say so specifically. The root cause is correctly identified rather than patched at the symptom: because `createEudrCrudOpenApi` derives the published create schema from `statementCreateSchema` itself (`api/statements/route.ts:310-314`), removing the field fixes the documentation and the validation in one move, and the two cannot disagree again. The update path is properly left intact — `referenceIssuedAt` remains on `statementUpdateSchema`, the `submitted→available` transition still sets it (`commands/statements.ts:536-540`), and the immutability guard at `commands/statements.ts:552-556` still throws, so `eudr.errors.referenceIssuedAtImmutable` stays reachable and is not orphaned in the five locale files. Both a schema-level jest regression and an API-path integration regression ship in the same PR, and I confirmed the integration test genuinely executed in CI rather than being skipped.

My one substantive finding is about the scope of the fix rather than a defect in the code that shipped: the PR's stated goal — accepting a whole-object payload — is not actually achieved, because a second server-managed field on the same entity still rejects the round trip.

## Verdict

❌ **request changes** — one minor finding (the whole-object POST this PR sets out to unblock still returns 400 on the `submittedAt` key, so the goal stated in the PR description is not met). Under this repository's review rule a minor finding requests changes; nothing here is broken or unsafe, and the remedy is small — either extend the fix to cover the sibling field, or narrow the claim and file a follow-up. The three nits below are optional.

## 🧪 Validation Gate

**Validation source: GitHub checks** (all checks for head `94ecd3d`, run [32597685126](https://github.com/open-mercato/open-mercato/actions/runs/32597685126), reported green; no required check is failing and none is pending, so per this repository's GitHub-checks-first policy no local re-run was needed).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job, "Build packages" step. |
| `yarn i18n:check-sync` | ✅ PASS | `test` job, "Check i18n sync" (`yarn tsx scripts/i18n-check-sync.ts`). |
| `yarn i18n:check-usage` | ✅ PASS | `test` job, "Check i18n usage". Note this step is `continue-on-error: true` in CI, so it is advisory; no locale key is orphaned by this diff, since `eudr.errors.referenceIssuedAtImmutable` is still thrown on the update path. |
| `yarn typecheck` | ✅ PASS | `test` job, "Checking types" step. |
| `yarn test` | ✅ PASS | `test` job, "Test" step — includes the new `data/__tests__/statement-schema.test.ts`. |

Additional green checks beyond the configured gate: `lint`, `ds-lint`, `audit-scope`, `documents-multi-instance`, `scope`, `license/cla`. The `ephemeral-integration (none)` job ran with `OM_INTEGRATION_MODULES: eudr` and reported `30 passed, 0 failed, 0 flaky, 0 skipped`, with the new case explicitly listed as passing:

```
✓ 10 packages/core/src/modules/eudr/__integration__/TC-EUDR-007.spec.ts:240:3 ›
  accepts a no-op referenceIssuedAt on create and keeps it server-managed (regression for #5508) (118ms)
```

`merge-coverage`, `docker-build`, `audit`, `comment`, `mutate` and `[code]smith` were skipped by their own conditions, not by failure.

## Findings

### 🔹 Minor

**`packages/core/src/modules/eudr/data/validators.ts:53` — the whole-object POST this PR is written to unblock still returns 400, on `submittedAt`.**

The PR's Goal states that `POST /api/eudr/statements` now "accepts a whole-object payload", and issue #5508's Impact section names the affected consumer class as "API consumers that serialize a whole object, a very common client pattern". After this change that pattern still fails, for the same structural reason and with the same shape of error — just under a different key.

Here is the concrete failure. `GET /api/eudr/statements` maps `submittedAt: toIsoString(record.submitted_at)` into every row (`api/statements/route.ts:110`), and `statementListItemSchema` declares it (`api/statements/route.ts:296`), so a client that reads a draft, edits a field, and POSTs the object back sends `submittedAt: null`. `statementCreateSchema` still begins with `rejectServerComputedFields(serverComputedStatementFields)` where that list is `['submittedAt']` (`validators.ts:53`), and `rejectServerComputedFields` rejects on **key presence alone**, regardless of value — `if (!(field in record)) continue` followed by an unconditional `context.addIssue(...)` (`validators.ts:67-80`). The ZodError becomes `400 {"error":"Invalid input","details":[…"eudr.errors.serverComputedField"…]}` via `packages/shared/src/lib/crud/factory.ts:607`. That is precisely the `referenceIssuedAt: null` bug this PR just fixed, one field over. `TC-EUDR-002.spec.ts:99-101` documents the same mechanism firing for evidence submissions, so the behavior is established and tested, not speculative.

The most economical fix that closes the whole class rather than one instance is to make the guard tolerate a no-op echo and reject only a meaningful attempt to set server state, in `validators.ts:67-80`:

```ts
for (const field of fields) {
  if (!(field in record)) continue
  if (record[field] === null || record[field] === undefined) continue
  context.addIssue({ /* … unchanged … */ })
}
```

That accepts `submittedAt: null` and `referenceIssuedAt: null` uniformly, keeps `completenessScore: 55` and `overallTier: 'low'` rejected exactly as `TC-EUDR-002`/`TC-EUDR-006` assert, and would let `referenceIssuedAt` stay in the create schema if you preferred symmetry. If you would rather keep this PR tightly scoped to #5508 — a perfectly reasonable call — then please instead reword the Goal and "What changed" sections so they claim only what the diff delivers (whole-object POSTs no longer fail *on `referenceIssuedAt`*), and open a follow-up issue for the `submittedAt` echo. I searched the tracker and no open issue covers it today. Either route resolves this finding.

### 💅 Nit

**`packages/core/src/modules/eudr/data/validators.ts:336-342` — two different conventions for server-managed fields now coexist on one entity, with nothing recording which applies where.**

After this change, `submittedAt` on a statement is loudly rejected with `eudr.errors.serverComputedField`, while `referenceIssuedAt` is silently stripped. Both behaviors are defensible and issue #5508 explicitly sanctioned the second ("Either drop the field from the documented create schema, or accept a no-op value such as `null`"), and the distinction is real: `referenceIssuedAt` was *advertised* in the create schema and `submittedAt` never was, which is exactly what made the old behavior incoherent. But the next person adding a server-managed field to this entity has no way to know which of the two patterns to follow. The new comment block explains why `referenceIssuedAt` is omitted; one sentence next to `serverComputedStatementFields` at `validators.ts:53` explaining that fields listed there are rejected because they are never advertised, whereas advertised fields are dropped from the create schema instead, would make the rule discoverable at the point of decision.

**`BACKWARD_COMPATIBILITY.md:38` — the contract waiver is argued in the PR body but not recorded where the protocol asks for it. ⚠️**

Flagging this explicitly because it touches a protected surface. That line classifies `data/validators.ts` Zod schema exports as "MUST NOT remove or narrow existing schemas", and this PR narrows `statementCreateSchema` and, through `z.infer`, the exported `StatementCreateInput` type.

I do not think this blocks, and I want to be clear about why. The repository's own operative review instrument scopes removal-breakage to responses: `.ai/review-checklist.md:290` says "No fields removed from existing response schemas (may add new fields)", and the Quick Reference matrix at `BACKWARD_COMPATIBILITY.md:270` marks API-route field removal as "BREAKING (res fields)" — parenthetically excluding request fields. `.ai/review-checklist.md:260-262` enumerates the protected public types and `StatementCreateInput` is not among them. Runtime behavior only relaxes: no caller that previously received a 201 can now receive anything else. And `statementCreateSchema` is a `ZodPipeline`, not a `ZodObject`, so `.extend()`/`.omit()` were never available on it and no schema-composition consumer can break. The only affected consumer is TypeScript code that named `referenceIssuedAt` in a `StatementCreateInput` literal — code that was guaranteed to 400 at runtime. The PR's "Breaking changes: None" section makes essentially this argument and I agree with it.

What is missing is the record. `AGENTS.md`'s deprecation protocol and `.ai/review-checklist.md:323-327` both expect a contract-surface change to leave a trace. A two-line entry in `UPGRADE_NOTES.md`, or a line in the changelog section of `.ai/specs/2026-07-06-eudr-compliance-module.md`, would let the next reader find the reasoning without excavating this PR. Author's call, but I would take the two lines.

**`packages/shared/src/lib/crud/factory.ts:607` — the issue's "side observation" about the leaked Zod regex is closed only on the create path.**

Issue #5508 notes that `referenceIssuedAt: "2026-08-01"` returns a raw Zod dump including the full ISO-8601 validation regex. This PR closes that for create by removing the field, and the new jest case at `data/__tests__/statement-schema.test.ts:23-30` locks it in. But `statementUpdateSchema` still accepts `referenceIssuedAt`, and `PUT /api/eudr/statements` with a date-only string still reaches `isoDateTimeSchema` and still returns `details: err.issues` verbatim, regex and all, because `factory.ts:607` serializes raw Zod issues into every CRUD 400 response.

That is platform-wide and pre-existing, so it is genuinely out of scope for this PR and I am not asking you to change it here. It is worth a separate issue against the CRUD factory, since every module inherits the behavior.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `statementCreateSchema`, `statementUpdateSchema` and `StatementCreateInput` all keep their names and import paths.
- [x] No function signature changed in a breaking way — `createStatementCommand.execute` is untouched apart from the removed throw.
- [ ] No required type field removed or narrowed — `StatementCreateInput` loses the optional `referenceIssuedAt` property. The field was optional, not required, and every request that set it returned 400, so no working consumer regresses. See the ⚠️ nit above: the waiver is sound but should be recorded in `UPGRADE_NOTES.md` or the eudr spec changelog per `BACKWARD_COMPATIBILITY.md:38`.
- [x] No HTTP route URL removed or renamed; no method changed — `/api/eudr/statements` and its `GET`/`POST`/`PUT`/`DELETE` exports are unchanged.
- [x] No field removed or retyped in an existing response shape — `statementListItemSchema` still returns `referenceIssuedAt` (`api/statements/route.ts:297`) and the export packet still emits it (`api/statements/[id]/export/route.ts:125,236`). Only the **request** contract changed.
- [x] No event or message name renamed or removed — `statementCrudEvents` is untouched.
- [x] No CLI command or flag renamed or removed — no CLI surface in this diff.
- [x] No database table or column renamed or removed; no column type narrowed — `reference_issued_at` remains on the entity (`data/entities.ts:251`) and is still written by the `submitted→available` transition. No migration is needed and correctly none ships.
- [x] Where a contract had to change: the change is a relaxation on the request side, so no bridge is required; the only residual item is the documentation trace noted above.

## 🧪 Test Coverage

Coverage for this fix is genuinely good, and it is layered correctly — the schema contract is pinned at the unit level where it is cheap, and the end-to-end HTTP behavior is pinned at the API level where it actually matters.

`packages/core/src/modules/eudr/data/__tests__/statement-schema.test.ts` asserts the create schema strips `referenceIssuedAt` for all three previously-failing inputs — an explicit `null`, a valid date-time, and the date-only string from the issue's side observation — and, importantly, asserts in the same file that `statementUpdateSchema` still parses both a date-time and an explicit `null`. That second describe block is what stops a future refactor from "simplifying" the omit onto the shared base schema and silently killing the `submitted→available` transition. The file follows the module's conventions (`data/__tests__/`, `/** @jest-environment node */`), matching siblings such as `shipping_carriers` and `customers`.

`TC-EUDR-007.spec.ts:240-268` covers the real HTTP path: it POSTs with `referenceIssuedAt: null` and with a serialized date-time, asserts 201 for both, reads both records back through `GET ?ids=`, and asserts each is a `draft` with a null `referenceIssuedAt` — so it proves both that the request is accepted and that the value is not persisted, which is the property that actually matters. Fixtures are created in the test and torn down in a `finally`, satisfying the self-containment requirement in `.ai/qa/AGENTS.md`. I verified in the CI log that this case ran and passed rather than being filtered out by the integration scope computation.

I also confirmed the change breaks no existing assertion. `TC-EUDR-007.spec.ts:477-481` still exercises `referenceIssuedAtImmutable` on the update path, and `TC-EUDR-002.spec.ts:99-101` and `TC-EUDR-006.spec.ts:243-254` cover different server-computed fields on different entities; none of them asserted create-time rejection of `referenceIssuedAt`, so no test needed deleting — which is itself evidence the removed throw was untested dead weight.

The one gap follows directly from the minor finding: there is no test asserting that a genuine whole-object round trip succeeds. If you extend the fix to `submittedAt`, please add to `TC-EUDR-007.spec.ts` a case that `GET`s a created draft, POSTs the returned row back verbatim (minus `id`), and expects 201 — that is the assertion that would have caught this gap, and it is the one that will keep the round trip working as new server-managed fields are added to the entity. If you instead scope the PR to `referenceIssuedAt` and file a follow-up, that test belongs with the follow-up.

pkarw:
🏷️ label rationale — `om-auto-review-pr` (updated on re-review of head `97be1efc`)

| Label | Why |
|-------|-----|
| `merge-queue` | Replaces `review`. The re-review approves: the minor finding from the previous pass (a whole-object `POST` still returning 400 on the `submittedAt` key) is fixed at the root in `rejectServerComputedFields`, proven by a schema regression and a `TC-EUDR-007` round-trip case that both demonstrably executed in CI. Only three optional nits remain, and under this repository's verdict rule nit-only reviews approve. |
| `bug` | Unchanged — this fixes a defect in the documented create contract (#5508). |
| `skip-qa` | Unchanged and re-confirmed on the larger diff. It still touches no `.tsx`, nothing under `packages/ui/src/` or `**/components/**`, leaves the database structure untouched (`reference_issued_at` and `submitted_at` both stay on the entity, no migration), breaks no `BACKWARD_COMPATIBILITY.md` contract, and now ships automated tests for the changed behavior at both the schema and API levels — so the automated-verification exemption applies and no manual QA gate is required. |
| `priority-low` | Unchanged. The backend create form does not render these fields, so there is no product-facing impact today; it matters to API consumers serializing whole objects. |
| `risk-low` | Unchanged, and re-checked against the wider blast radius of the follow-up commit. `rejectServerComputedFields` is shared by ten schemas across five entities, but the change only *relaxes* it for `null`/`undefined`, and no server-computed field name appears in its entity's base schema — so a tolerated echo is stripped rather than persisted, and every meaningful value stays rejected as `TC-EUDR-002`/`TC-EUDR-006` assert. Every CI check is green on head `97be1efc`. |
| `in-progress` | Review lock held by this run; released in the completion comment below. |

pkarw:

pkarw:

One minor finding drives the verdict — a whole-object `POST` still returns 400 on the `submittedAt` key, so the goal in the PR description is not yet met — plus three optional nits. Validation source: GitHub checks (all green on head `94ecd3d`, including the new TC-EUDR-007 regression); no local gate re-run was needed and no CI follow-up is outstanding. autofix: skipped (not my PR — re-run with --autofix to fix it here).

haxiorz:
Thanks for the thorough review, @pkarw — the minor finding was exactly right, so I extended the fix to close the class rather than narrow the claim. Pushed as `97be1efc5`.

**🔹 Minor — `submittedAt` broke the same round trip.** Fixed at the source: `rejectServerComputedFields` now tolerates a `null`/`undefined` no-op echo and rejects only a meaningful (non-null) set attempt, so a whole-object POST of a draft (which echoes `submittedAt: null`) succeeds while `completenessScore: 55` / `overallTier: 'low'` stay rejected exactly as `TC-EUDR-002`/`TC-EUDR-006` assert. Kept `referenceIssuedAt` dropped from the create schema (it was *advertised*, so the OpenAPI stays honest) rather than folding it back — the two mechanisms are now documented inline. Added the round-trip test you asked for to `TC-EUDR-007`: it GETs a created draft, POSTs the returned row back verbatim minus `id`, and expects 201 (plus schema-level jest cases for `submittedAt: null` accepted / non-null rejected, observed red before the fix).

**💅 Nit — two conventions coexisting.** Added a comment next to `serverComputed*Fields` explaining the rule at the decision point: listed fields are rejected on a real set attempt (null echo tolerated) because they were never advertised, whereas an advertised field is dropped from the create schema instead.

**💅 Nit ⚠️ — BC trace.** Recorded the create-schema narrowing waiver in the eudr spec changelog (`.ai/specs/2026-07-06-eudr-compliance-module.md`), citing the request-vs-response distinction (`BACKWARD_COMPATIBILITY.md:270`, `.ai/review-checklist.md:290`) and that `StatementCreateInput` is not a protected public type.

**💅 Nit — leaked Zod regex on the update path.** Agreed this is platform-wide (`factory.ts:607` serializes raw Zod issues into every CRUD 400) and out of scope here; I left it untouched. Happy to file a separate issue against the CRUD factory if you'd like it tracked.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-23T02:11:30Z. Re-review of head `97be1efc` after the author's follow-up. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review: Re-review — fix(eudr): accept a no-op referenceIssuedAt on statement create (#5508)

## 🎯 Summary

This is a **re-review** of head `97be1efc` after @haxiorz addressed my previous `changes_requested` review on `94ecd3d`. I re-checked the one minor finding that drove that verdict, each of the three nits, and the whole diff again — `packages/core/src/modules/eudr/data/validators.ts`, `commands/statements.ts`, the two test files, and the spec changelog — plus the surrounding contract surface: the OpenAPI wiring in `api/statements/route.ts`, the statement list serializer, the sibling `TC-EUDR-002`/`TC-EUDR-006` server-computed-field assertions, the four other entity schemas that share `rejectServerComputedFields`, and `BACKWARD_COMPATIBILITY.md` §2/§7 plus the quick-reference table.

**The previous minor finding is genuinely resolved, and resolved at the right level.** Rather than special-casing `submittedAt` alongside `referenceIssuedAt`, the author fixed the guard itself: `rejectServerComputedFields` (`packages/core/src/modules/eudr/data/validators.ts:79`) now skips a `null`/`undefined` value instead of rejecting on key presence alone. That is the correct distinction — the guard exists to stop a client *setting* server-owned state, and echoing back the value the server just returned sets nothing. I verified the relaxation cannot leak a value into persistence: none of the five server-computed field names (`submittedAt`, `completenessScore`, `missingFields`, `validationWarnings`, `computedArea`, `countryRisks`, `overallTier`, `isSimplified`, `assessedByName`, `completedAt`) appears in its entity's base schema, so a tolerated `null` is stripped by the piped `z.object` rather than written. A meaningful value is still rejected, which the untouched `TC-EUDR-002`/`TC-EUDR-006` cases continue to prove.

The rest of the follow-up is equally solid. `statementCreateSchema` keeps `referenceIssuedAt` dropped via `statementBaseSchema.omit(...)` (`validators.ts:349-351`) rather than folding it back — the right call, because that field *was* advertised in the published create schema and the OpenAPI document is derived from the same object (`api/statements/route.ts:313`), so the document and the validator cannot drift apart again. I confirmed `eudr.errors.referenceIssuedAtImmutable` is still reachable on the update path (`commands/statements.ts:552-556`), so none of the five locale files carries an orphaned key, and `TC-EUDR-007:517` still asserts it. `createStatementCommand` continues to hard-write `referenceIssuedAt: null` on insert (`commands/statements.ts:621`), so removing the throw did not open a write path. The two conventions now coexisting on this entity are documented at their decision points (`validators.ts:50-53`, `:66-68`, `:345-348`), which is what I asked for.

The requested round-trip test also landed, and I checked it actually ran rather than trusting the check colour: both new cases executed in the `ephemeral-integration` job of run [32610222635](https://github.com/open-mercato/open-mercato/actions/runs/32610222635) (`TC-EUDR-007.spec.ts:240` in 136 ms, `:271` in 127 ms), and `src/modules/eudr/data/__tests__/statement-schema.test.ts` shows `PASS` in the `test` job. The declined nit — the raw Zod regex leaking through the shared CRUD factory at `factory.ts:607` — is correctly declined: it is a platform-wide issue in a file this PR does not touch, and fixing it here would be scope expansion.

What remains is three nits, all of them optional and none of them about the shipped behaviour.

## Verdict

✅ **approve** — the minor finding that drove the previous verdict is fixed at the root and proven by tests that demonstrably executed in CI; every gate command is green on head `97be1efc`; no contract surface is broken; and the three remaining items are nits the author may take or leave. Under this repository's verdict rule, a review with only Low/nit findings approves.

## 🧪 Validation Gate

**Validation source: GitHub checks** — every check on head `97be1efc` is green or intentionally skipped; nothing is failing and nothing is pending, so per this repository's GitHub-checks-first rule no local re-run was needed. Each configured `validation.commands` entry maps onto a CI step that ran on this exact head.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job, run [32610222635](https://github.com/open-mercato/open-mercato/actions/runs/32610222635) (3m34s). |
| `yarn generate` | ✅ PASS | Same `prepare` job; generated-file drift would have failed it. |
| `yarn build:packages` (rebuild after generate) | ✅ PASS | Same `prepare` job. |
| `yarn i18n:check-sync` | ✅ PASS | `test` job, step "Check i18n sync" — relevant here because the removed create-path throw could have orphaned `eudr.errors.referenceIssuedAtImmutable` in five locale files; it did not. |
| `yarn i18n:check-usage` | ✅ PASS | `test` job, step "Check i18n usage". |
| `yarn typecheck` | ✅ PASS | `test` job, step "Checking types" — covers the `StatementCreateInput` narrowing against every consumer of `ScopedStatementCreateInput` in `commands/statements.ts`. |
| `yarn test` | ✅ PASS | `test` job, step "Test" — includes `PASS src/modules/eudr/data/__tests__/statement-schema.test.ts`. |
| `yarn build:app` | ✅ PASS | `prepare` job, step at `.github/workflows/ci.yml:190`. |

Beyond the configured gate, `lint`, `ds-lint`, `audit-scope`, `scope`, `documents-multi-instance` and `ephemeral-integration` all passed, and `license/cla` is signed. `docker-build`, `audit`, `mutate`, `merge-coverage` and `[code]smith` are skipped by their own path/event filters, not failing.

## Findings

### 💅 Nit

1. **`.ai/specs/2026-07-06-eudr-compliance-module.md:335` — the backward-compatibility waiver cites the wrong line of `BACKWARD_COMPATIBILITY.md`.** The waiver leans on `BACKWARD_COMPATIBILITY.md:270` for the claim that request-field removal is not classified as breaking, but line 270 sits inside "### 14. Generated File Contracts (STABLE)" and reads "MUST NOT remove AI generated registry exports" — unrelated to API request schemas. The rule the waiver actually relies on is `BACKWARD_COMPATIBILITY.md:181` in "### 7. API Route URLs (STABLE)" ("MUST NOT remove fields from existing response schemas", scoping the prohibition to responses only), reinforced by the quick-reference row at `:286` ("API routes … Remove field: BREAKING (res fields)"). The companion citation `.ai/review-checklist.md:290` is exact and correct. I independently verified the waiver's *conclusion* — §7 and the quick-reference table both restrict the prohibition to response fields, and §2's enumerated protected types do not include `StatementCreateInput` — so the substance stands and this changes nothing about the merge decision. The value of a recorded waiver, though, is that a future auditor can follow the citation to the rule; swapping `:270` for `:181` (or dropping the line numbers in favour of the section names, which will not drift as the file grows) preserves that.

2. **`packages/core/src/modules/eudr/data/validators.ts:79` — the shared relaxation is exercised only through the statements schema.** `rejectServerComputedFields` is piped into ten schemas across five entities, and this PR changes its semantics for all of them, but the new `submittedAt: null` case in `data/__tests__/statement-schema.test.ts:44` is the only test of the tolerance branch. The rejection side stays covered for the other entities by `TC-EUDR-002`/`TC-EUDR-006`, so the guard cannot silently vanish — but nothing pins the tolerance for, say, `plotCreateSchema` with `computedArea: null` or `evidenceSubmissionCreateSchema` with `completenessScore: null`. One assertion of each polarity against a non-statement schema, next to the existing block in `statement-schema.test.ts`, would lock the shared contract down where it now actually lives rather than only where it was first noticed.

3. **PR description — "closes the whole class across the module's entities" is broader than what shipped.** The class is closed for server-computed fields whose no-op value is `null`, which on a fresh statement draft covers `submittedAt`. It is not closed for the sibling entities, because a real `GET` row echoes *non-null* values for them: a plot returns `computedArea` and `validationWarnings: []`, a submission returns `completenessScore: 55` and `missingFields`, a risk assessment returns `isSimplified: false` and `overallTier`. Round-tripping any of those whole objects back to `POST` still returns `400 eudr.errors.serverComputedField` — correctly, by the guard's design, but it means the "serialize a whole object and POST it back" pattern from #5508 remains broken outside statements. This is a wording issue in the description rather than a defect in the code, and closing that gap would be scope expansion on this PR. Worth either narrowing the sentence or opening a follow-up issue — alongside the CRUD-factory Zod-leak issue you offered to file (`packages/shared/.../factory.ts:607`), which I would like tracked.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `statementCreateSchema`, `statementUpdateSchema` and `StatementCreateInput` all keep their names and export sites.
- [x] No function signature changed in a breaking way — `rejectServerComputedFields` keeps its `(fields: readonly string[]) => z.ZodType<unknown>` signature; only its refinement body changed.
- [ ] No required type field removed or narrowed — `StatementCreateInput` loses the **optional** `referenceIssuedAt`. No required field is affected, and `BACKWARD_COMPATIBILITY.md` §2 enumerates the protected exported types; `StatementCreateInput` is not among them. Recorded as an explicit waiver in the eudr spec changelog, which is the protocol this repository asks for.
- [x] No HTTP route URL removed or renamed; no method changed — `POST/PUT/DELETE /api/eudr/statements` are untouched.
- [x] No field removed or retyped in an existing response shape — `statementListItemSchema` (`api/statements/route.ts:297`) still emits `referenceIssuedAt`, the list serializer (`:112`) still maps `reference_issued_at`, and both export routes still emit it. This is the one prohibition §7 actually imposes, and the PR honours it.
- [x] No event or message name renamed or removed — `statementCrudEvents` untouched.
- [x] No database table or column renamed or removed — `reference_issued_at` and `submitted_at` stay on the entity and keep being written by the lifecycle transitions; no migration is needed or present.
- [x] Where a contract had to change, the change is request-side only and strictly relaxing: every payload that previously succeeded still succeeds, and payloads that previously returned 400 now return 201. No bridge or deprecation window is required, and the reasoning is written down where a future auditor will look for it.

## 🧪 Test Coverage

Coverage is now proportionate to the change and, importantly, verified to have executed rather than merely present in the diff.

`packages/core/src/modules/eudr/data/__tests__/statement-schema.test.ts` covers the schema layer in both directions. On the `referenceIssuedAt` side it asserts that an explicit `null`, a full ISO date-time, and a date-only `'2026-08-01'` string are each stripped from the parsed create payload rather than rejected — the date-only case being the one that previously leaked a raw Zod ISO-8601 regex to the client. On the `submittedAt` side it asserts the new tolerance (`null` accepted and stripped) *and* the preserved rejection (a non-null date-time still throws), which is the pair that matters: a test for tolerance alone would pass just as happily if the guard had been deleted outright. Two further cases confirm the update schema still accepts `referenceIssuedAt` as both a date-time and an explicit `null`, so the immutability mechanism on the update path is pinned against accidental removal.

`TC-EUDR-007.spec.ts` covers the API path end to end. The first new case creates statements with `referenceIssuedAt: null` and with a serialized date-time and asserts both come back as drafts with a null issue date — proving the field is ignored rather than persisted, which the schema-level test cannot show. The second is the round trip I asked for: it creates a draft, `GET`s the full row (asserting on the way that `submittedAt` really is `null`, so the test would fail loudly if the fixture stopped exercising the bug), `POST`s that row back verbatim minus `id`, expects 201, and then re-reads the created record to confirm neither `referenceIssuedAt` nor `submittedAt` was persisted. Both cases clean up in a `finally`, so they leave no residue for the rest of the suite.

The one gap is the one recorded as nit 2: the tolerance branch of the shared guard is proven only through statements, while four other entities inherit it untested. Concretely, `data/__tests__/statement-schema.test.ts` (or a sibling covering the other schemas) could add `plotCreateSchema.parse({ …validPlot, computedArea: null })` succeeding with the key stripped, and the same input with `computedArea: 12.5` throwing. That is optional polish, not a merge condition.

pkarw:

Re-review of head `97be1efc`. The minor finding that drove the previous `changes_requested` is resolved at the root — `rejectServerComputedFields` now tolerates a `null`/`undefined` no-op echo instead of rejecting on key presence, and I verified no server-computed field name appears in its entity's base schema, so a tolerated echo is stripped rather than persisted while every meaningful value stays rejected. All three previous nits are accounted for: the two-conventions nit and the backward-compatibility trace were addressed, and the CRUD-factory Zod-leak nit was reasonably declined as platform-wide and out of scope. Three new nits remain, all optional: a wrong line citation in the spec waiver (`BACKWARD_COMPATIBILITY.md:270` should be `:181` — the substance of the waiver is correct and I verified it independently), the shared guard's tolerance branch being tested only through statements, and a PR-description sentence claiming a broader fix than shipped.

Validation source: GitHub checks — all eight `validation.commands` entries map onto green steps of run [32610222635](https://github.com/open-mercato/open-mercato/actions/runs/32610222635) on this exact head, so no local gate re-run was needed. I confirmed the new tests genuinely executed rather than trusting the check colour: `TC-EUDR-007.spec.ts:240` and `:271` ran in `ephemeral-integration`, and `statement-schema.test.ts` shows `PASS` in `test`. No check is pending, so no CI follow-up is outstanding. autofix: skipped (not my PR — re-run with --autofix to fix it here).

PR #5529, opened 2026-08-23
fix(auth): surface login failures instead of an empty 500

patzick:
## 🎯 Why

`demo.openmercato.com` started answering `POST /api/auth/login` with a **500 and no response body at all**. That gave the browser nothing to render and the response nothing an operator could act on, so the outage was indistinguishable from a bug in the login route.

It was not a login bug. Probing the deployment showed the split cleanly:

| Endpoint | Needs DI/DB? | Result |
|---|---|---|
| `GET /api/version` | no | 200 |
| `GET /api/auth/locale` | no (i18n only) | 400 with a proper JSON body |
| `GET /login` | no | 200 HTML |
| `POST /api/auth/login` | **yes** | **500, empty** |
| `GET /api/checkout/pay/<bogus>` | **yes** | **500** `{"error":"Internal server error"}` |

Two unrelated modules failing at the same `createRequestContainer()` / `resolve('em')` boundary — an infrastructure problem. The reason `checkout` returned a readable body and `auth` returned nothing is simply that the checkout route has a top-level `try/catch` and the login route did not.

This PR closes that diagnostic gap. **It does not fix the outage** — it makes the next one legible.

## 🔧 What changed

`POST` had no top-level `try/catch`, so anything thrown between `resolveTranslations()` and the final response escaped the handler. It is now wrapped:

- The cause is logged through the structured logger under the stable scope `auth.login`, so it is greppable in deployment logs.
- The response is a generic **translated** message (`auth.login.errors.generic`, which already existed) with a proper JSON body — deliberately generic, because auth responses must never let an unauthenticated caller distinguish one failure from another.
- Recovering that message is itself defensive: `resolveTranslations()` can be the thing that failed, so it falls back to the literal rather than throwing a second time inside the error path.

The handler body moved into `handleLoginRequest` instead of being re-indented inside a `try` block, so the diff is a small wrapper rather than a 120-line whitespace change.

Also adds `handleAuthRouteError`, mirroring the existing `handleCheckoutRouteError`, so the pattern is reusable rather than copy-pasted.

## 📌 Scope — the same gap exists in five sibling routes

Verified by inspection: `logout.ts`, `autologin.ts`, `reset.ts`, `reset/confirm.ts` and `session/refresh.ts` all begin their handler with real work rather than a `try`, so they fail the same opaque way. **This PR deliberately fixes only `login.ts`**, which is the reported symptom; the shared helper is in place so extending it is a small follow-up. Say the word and I will widen it here or in a separate PR.

## 🧪 Tests

`login.unhandled-error.test.ts` — four cases, each of which **fails against `main` and passes here** (verified by stashing the `login.ts` change and re-running):

- a container that cannot be built returns a JSON 500 body rather than an empty response;
- the cause is logged once, under scope `auth.login`, with the original error attached;
- the underlying failure never reaches the caller — the serialized body is asserted not to contain the host, port, or `ECONNREFUSED` from the simulated database error;
- a throw *after* successful authentication (the session/JWT tail, e.g. a missing signing secret) is contained too, not just early failures.

Local runs, related scope only:

- `jest src/modules/auth` → **629 passed, 78 suites, 0 failed**
- `tsc --noEmit` on `packages/core` → 0 errors (excluding the pre-existing `#generated/…` not-found noise present on `main`)
- ESLint on the three touched files → clean

## 🔍 Review notes

- No behavior change on any path that already returned a response — 400/401/403/429 and the success path are untouched.
- Nothing new is logged on ordinary failed logins; the logger only fires on an unhandled throw.
- Per the auth module rules, no credential reaches the log: the password never enters a query, and only the error object plus a scope string are logged.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5529"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790076157&installation_model_id=432243&pr_number=5529&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5529&signature=8da8902a20aae3f14d34bd772751689412d93697b731598c5c8bce750d4c1891"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/auth/api/__tests__/login.unhandled-error.test.ts
+++ b/packages/core/src/modules/auth/api/__tests__/login.unhandled-error.test.ts
@@ -1,0 +1,121 @@
+/** @jest-environment node */
+/**
+ * Regression guard: an infrastructure failure inside the login route must
+ * surface as a JSON 500 and a server-side log line, not as a bare 500 with an
+ * empty body. The empty-body case gives the client nothing to render and the
+ * operator nothing to grep, which is what made a demo outage (database
+ * unreachable) indistinguishable from a bug in this route.
+ */
+import { registerApiInterceptors } from '@open-mercato/shared/lib/crud/interceptor-registry'
+
+const createRequestContainerMock = jest.fn()
+const loggerErrorMock = jest.fn()
+
+jest.mock('@open-mercato/shared/lib/i18n/server', () => ({
+  resolveTranslations: async () => ({
+    translate: (_key: string, fallback?: string) => fallback ?? '',
+  }),
+}))
+
+jest.mock('@open-mercato/shared/lib/di/container', () => ({
+  createRequestContainer: (...args: unknown[]) => createRequestContainerMock(...args),
+}))
+
+jest.mock('@open-mercato/shared/lib/logger', () => ({
+  createLogger: () => ({
+    error: (...args: unknown[]) => loggerErrorMock(...args),
+    warn: jest.fn(),
+    info: jest.fn(),
+    debug: jest.fn(),
+    child: jest.fn(),
+  }),
+}))
+
+jest.mock('@open-mercato/shared/lib/auth/jwt', () => ({ signJwt: () => 'jwt-token' }))
+
+jest.mock('@open-mercato/core/modules/auth/lib/rateLimitCheck', () => ({
+  checkAuthRateLimit: jest.fn(async () => ({ error: null, compoundKey: null })),
+  resetAuthRateLimit: jest.fn(async () => undefined),
+}))
+
+jest.mock('@open-mercato/core/modules/auth/events', () => ({
+  emitAuthEvent: jest.fn(async () => undefined),
+}))
+
+const { POST } = require('@open-mercato/core/modules/auth/api/login')
+
+function loginRequest() {
+  return new Request('http://localhost/api/auth/login', {
+    method: 'POST',
+    headers: { 'content-type': 'application/x-www-form-urlencoded' },
+    body: new URLSearchParams({ email: 'someone@example.com', password: 'whatever' }).toString(),
+  })
+}
+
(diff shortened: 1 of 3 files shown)


patzick:
🏷️ label rationale

- 🐛 `bug` — a route that answers with an empty 500 is a defect in its own right, independent of whatever infrastructure failure triggered it.
- 🔍 `review` — ready and non-draft, so it enters the pipeline in the review state.
- ⏫ `priority-high` — this is the observability gap that made a live demo outage undiagnosable from the response alone; the sooner it lands, the sooner the next occurrence names itself.
- 🟢 `risk-low` — additive error handling only. Every path that already produced a response is untouched; the handler body was moved into a named function without a single line changed inside it.
- ⏭️ `skip-qa` — qualifies for the automated-verification exemption: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no database or API-shape change, no `BACKWARD_COMPATIBILITY.md` contract touched, and the changed behavior ships with tests that fail against `main`.

On `skip-qa` specifically: the response shape on this path changes from *nothing* to a JSON body, which is strictly additive — no client could have been reading a field that is now gone, since there was no body to read.

PR #5540, opened 2026-08-23
fix(webhooks): use static notification preference label

### LINKED ISSUE #5496: bug: [Notifications] the preferences screen renders a raw {webhookName} placeholder
## Summary
The notification-type catalogue uses each type's per-notification `titleKey` as its catalogue label. For
`webhooks.delivery.failed` that key is parameterised, so the preferences screen shows the literal placeholder.

## Steps to reproduce
1. Open `/backend/profile/notification-preferences` as any user.
2. Scroll to the webhooks row.

## Actual
`Dostawa webhooka nie powiodła się dla {webhookName}`

Also present in three `aria-label`s, so a screen reader reads the placeholder out loud:
```
"Dostawa webhooka nie powiodła się dla {webhookName} – W aplikacji"
"Dostawa webhooka nie powiodła się dla {webhookName} – E-mail"
"Dostawa webhooka nie powiodła się dla {webhookName} – Push"
```

## Expected
A static catalogue label, e.g. `Dostawa webhooka nie powiodła się`.

Re-verified 2026-08-22 on the 0.7.0 build.

haxiorz:
Closes #5496

## Goal

Show a static, translated label for webhook delivery-failure notification preferences while preserving the interpolated runtime notification title.

## Problem and root cause

The webhook delivery-failed notification omitted `labelKey`. Notification type synchronization therefore fell back to `titleKey`, which contains the runtime `{webhookName}` placeholder. The unresolved placeholder appeared in the preference row and switch accessibility labels.

## What changed

- Added a dedicated static preference `labelKey`.
- Added translations for all five supported locales.
- Added regression coverage verifying the label is present, translated, static, and distinct from the runtime title.

## Validation

- `yarn build:packages` — 26 package build tasks passed.
- `yarn generate` — all generators completed and generated outputs were unchanged.
- `yarn i18n:check-sync` — all five locale files are synchronized.
- `yarn i18n:check-usage` — completed successfully with the existing unused-key advisory.
- `yarn typecheck` — all 26 typecheck tasks passed.
- `TURBO_CONCURRENCY=2 yarn test` — all 33 workspace tasks passed; webhooks passed 19 suites and 146 tests, and create-app passed 878 tests with five intentional skips.
- `yarn build:app` — the production application build completed successfully.
- `git diff --cached --check` — passed.
- The regression test was observed failing before the implementation and passing afterward.

## Automated review

- `om-code-review` — approved with zero remediation rounds.
- Fresh independent review — not configured for the standard review profile; the direct review of record approved the complete patch.
- Unresolved findings — None.

## Breaking changes

None. The notification type ID, runtime title key, API surface, database schema, and delivery behavior are unchanged.

## Labels and delivery

- `review` — The verified, non-draft change is ready for maintainer review.
- `bug` — The change fixes a raw interpolation placeholder rendered in notification preferences.
- `skip-qa` — No UI-rendering, API, database, or compatibility-contract surface changed, and automated regression and localization checks cover the behavior, so the automated-verification exemption applies.
- `priority-medium` — This is an ordinary user-visible bug without outage, security, or release-blocking impact.
- `risk-medium` — The patch is isolated to one module and carries focused regression coverage, matching the ordinary single-module bug-fix classification.

## Manual QA

Manual QA is not required because the patch changes notification metadata and locale values without modifying a UI-rendering file. The regression test verifies the static preference label in every locale, while the localization checks verify catalog consistency.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5540"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790108619&installation_model_id=432243&pr_number=5540&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5540&signature=9f4c88ed5470653891ac351e18ab1a2fac30afec0df25dac964f3d829159c14a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/webhooks/src/modules/webhooks/__tests__/notifications.test.ts
+++ b/packages/webhooks/src/modules/webhooks/__tests__/notifications.test.ts
@@ -1,0 +1,27 @@
+import de from '../i18n/de.json'
+import en from '../i18n/en.json'
+import es from '../i18n/es.json'
+import ko from '../i18n/ko.json'
+import pl from '../i18n/pl.json'
+import { notificationTypes } from '../notifications'
+
+const dictionaries: Record<string, Record<string, string>> = { de, en, es, ko, pl }
+
+describe('webhook notification catalogue', () => {
+  it('uses a static delivery-failure label in every locale', () => {
+    const deliveryFailed = notificationTypes.find((definition) => definition.type === 'webhooks.delivery.failed')
+
+    expect(deliveryFailed).toMatchObject({
+      labelKey: 'webhooks.notifications.delivery.failed.label',
+      titleKey: 'webhooks.notifications.delivery.failed.title',
+    })
+    expect(deliveryFailed?.labelKey).not.toBe(deliveryFailed?.titleKey)
+
+    for (const dictionary of Object.values(dictionaries)) {
+      const label = dictionary[deliveryFailed?.labelKey ?? '']
+      expect(label).toBeDefined()
+      expect(label).not.toMatch(/\{[^}]+\}/)
+      expect(label).not.toBe(deliveryFailed?.labelKey)
+    }
+  })
+})
--- a/packages/webhooks/src/modules/webhooks/i18n/de.json
+++ b/packages/webhooks/src/modules/webhooks/i18n/de.json
@@ -158,5 +158,6 @@
   "webhooks.nav.group": "Integrations",
   "webhooks.nav.title": "Webhooks",
   "webhooks.notifications.delivery.failed.body": "Webhook {webhookName} konnte das Ereignis {eventType} nicht zustellen. {errorMessage}",
+  "webhooks.notifications.delivery.failed.label": "Webhook-Zustellung fehlgeschlagen",
   "webhooks.notifications.delivery.failed.title": "Webhook-Zustellung für {webhookName} fehlgeschlagen"
 }
--- a/packages/webhooks/src/modules/webhooks/i18n/en.json
+++ b/packages/webhooks/src/modules/webhooks/i18n/en.json
@@ -158,5 +158,6 @@
   "webhooks.nav.group": "Integrations",
   "webhooks.nav.title": "Webhooks",
   "webhooks.notifications.delivery.failed.body": "Webhook {webhookName} could not deliver event {eventType}. {errorMessage}",
+  "webhooks.notifications.delivery.failed.label": "Webhook delivery failed",
   "webhooks.notifications.delivery.failed.title": "Webhook delivery failed for {webhookName}"
 }
--- a/packages/webhooks/src/modules/webhooks/i18n/es.json
+++ b/packages/webhooks/src/modules/webhooks/i18n/es.json
@@ -158,5 +158,6 @@
   "webhooks.nav.group": "Integrations",
   "webhooks.nav.title": "Webhooks",
   "webhooks.notifications.delivery.failed.body": "El webhook {webhookName} no pudo entregar el evento {eventType}. {errorMessage}",
+  "webhooks.notifications.delivery.failed.label": "La entrega del webhook falló",
   "webhooks.notifications.delivery.failed.title": "La entrega del webhook falló para {webhookName}"
 }
--- a/packages/webhooks/src/modules/webhooks/i18n/ko.json
+++ b/packages/webhooks/src/modules/webhooks/i18n/ko.json
@@ -158,5 +158,6 @@
   "webhooks.nav.group": "통합",
   "webhooks.nav.title": "웹훅",
   "webhooks.notifications.delivery.failed.body": "웹훅 {webhookName}이(가) 이벤트 {eventType}을(를) 전달하지 못했습니다. {errorMessage}",
+  "webhooks.notifications.delivery.failed.label": "웹훅 전달 실패",
   "webhooks.notifications.delivery.failed.title": "{webhookName}의 웹훅 전달 실패"
 }
--- a/packages/webhooks/src/modules/webhooks/i18n/pl.json
+++ b/packages/webhooks/src/modules/webhooks/i18n/pl.json
@@ -158,5 +158,6 @@
   "webhooks.nav.group": "Integracje",
   "webhooks.nav.title": "Webhooki",
   "webhooks.notifications.delivery.failed.body": "Webhook {webhookName} nie mógł dostarczyć zdarzenia {eventType}. {errorMessage}",
+  "webhooks.notifications.delivery.failed.label": "Dostawa webhooka nie powiodła się",
   "webhooks.notifications.delivery.failed.title": "Dostawa webhooka nie powiodła się dla {webhookName}"
 }
--- a/packages/webhooks/src/modules/webhooks/notifications.ts
+++ b/packages/webhooks/src/modules/webhooks/notifications.ts
@@ -4,6 +4,7 @@ export const notificationTypes: NotificationTypeDefinition[] = [
   {
     type: 'webhooks.delivery.failed',
     module: 'webhooks',
+    labelKey: 'webhooks.notifications.delivery.failed.label',
     titleKey: 'webhooks.notifications.delivery.failed.title',
     bodyKey: 'webhooks.notifications.delivery.failed.body',
     icon: 'webhook',


pkarw:
🤖 \`om-auto-review-pr\` — claiming this PR for automated review. Will release the lock when the review is submitted.

pkarw (approved):
# 🔍 Code Review: fix(webhooks): use static notification preference label

## 🎯 Summary

This change fixes issue #5496, where the notification-preferences screen rendered the raw `{webhookName}` placeholder for the webhooks row (and in the three per-channel `aria-label`s, so a screen reader read the placeholder aloud). The root cause was that `syncNotificationTypes` mirrors `label_key ← def.labelKey ?? def.titleKey`, and `webhooks.delivery.failed` shipped no `labelKey`, so the parameterised runtime title became the catalogue label. The patch adds a dedicated `labelKey` to the definition, adds the corresponding static string to all five locale catalogues, and adds a regression test that pins the label as present, translated, placeholder-free, and distinct from the runtime title.

The approach is the one the framework documents for exactly this case — `packages/core/src/modules/notifications/AGENTS.md` describes `labelKey`/`descriptionKey` as the additive way to give a type a distinct preferences-screen label — so the fix works with the platform contract rather than around it. I verified the two things that could have made this fix incomplete, and both hold:

- **Existing installations are repaired, not just fresh ones.** `packages/core/src/modules/notifications/lib/notification-type-registry.ts:103-113` compares `row.label_key !== labelKeyFor(def)` and issues an `UPDATE` on drift, so a deployed tenant whose `notification_types` row already carries the parameterised title key is corrected on the next sync rather than keeping the stale label behind an insert-only path.
- **The scope is genuinely complete.** I resolved the `titleKey` of every notification-type definition in the repository against its module's `en.json` and checked each resolved value for an interpolation placeholder. `webhooks.notifications.delivery.failed.title` is the only one, so no sibling module is left with the same defect and the single-module scope of this PR is correct rather than a partial fix.

Reviewed scope: the full seven-file diff — the notification definition, the five locale catalogues, and the new test suite — plus the framework sync/read path the change depends on (`notification-type-registry.ts`, `typeCatalogue.ts`, `NotificationPreferenceMatrix.tsx`, `NotificationSettingsPageClient.tsx`), which the diff does not touch but whose behaviour determines whether the fix actually reaches the screen.

## Verdict

✅ approve — the change fixes the reported defect at its correct layer, repairs existing deployments through the registry's drift-update path, is the only occurrence of this defect in the repository, ships regression coverage in the same PR, changes no contract surface, and the complete validation gate is green on the reviewed head (`ba1f286`). No blocker, major, or minor findings were found; the two items below are optional nits the author may take or leave.

## 🧪 Validation Gate

Validation source: **GitHub checks** (workflow run [32664270414](https://github.com/open-mercato/open-mercato/actions/runs/32664270414), `head_sha` `ba1f286c85d73640e8d1f51a16c400399771ee17` — the exact commit reviewed here, conclusion `success`), with one narrow local fallback for the single configured command CI does not run.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | CI `prepare` job (`ci.yml:167`). |
| `yarn generate` | ✅ PASS | CI `prepare` job (`ci.yml:175`), followed by the second `build:packages` at `ci.yml:178`; the job is green, so generated output is unchanged. |
| `yarn i18n:check-sync` | ✅ PASS | Not executed by any CI workflow, so it was run locally as the narrow fallback the repo's GitHub-checks-first rule allows — and it is the one gate most relevant to a locale-only diff. Result: 5 locales (en, pl, es, de, ko) across 56 modules, all translation files in sync. |
| `yarn i18n:check-usage` | ✅ PASS | CI (`ci.yml:480`). The new key is referenced through the `labelKey: '…'` indirect-property pattern the scanner recognises (`scripts/i18n-scanner.mjs`), so it is not reported as unused. |
| `yarn typecheck` | ✅ PASS | CI `test` job (`ci.yml:498`). |
| `yarn test` | ✅ PASS | CI `test` job (`ci.yml:510`). The new suite was additionally re-run locally against the PR head — 1 suite, 1 test, passing in 0.6s. |
| `yarn build:app` | ✅ PASS | CI (`ci.yml:190`). |

Beyond the configured gate, every other required and optional check on this head is green as well, including `lint`, `ds-lint`, `docker-build`, `audit-scope`, CodeQL, and all fifteen `ephemeral-integration` shards. Nothing is pending and nothing is failing, so there is no CI disclosure to make and no follow-up owed.

## Findings

### 💅 Nit

- `packages/webhooks/src/modules/webhooks/__tests__/notifications.test.ts:8` — the `dictionaries` map is a hand-maintained list of the five current locales, so a sixth locale added later would silently escape this assertion while the test kept passing. This is a genuinely optional remark rather than a coverage hole, because `yarn i18n:check-sync` already fails when any locale is missing a key that another locale has, which is the failure mode that matters here. If you want the suite to be self-defending anyway, assert the dictionary count against the repo's supported-locale list so adding a locale forces the test to be updated.
- `packages/webhooks/src/modules/webhooks/notifications.ts:7` — nothing at the framework level stops the next module from shipping a parameterised `titleKey` with no `labelKey` and reintroducing this class of bug elsewhere; the `labelKey ?? titleKey` fallback in `notification-type-registry.ts:88` makes it a silent, per-module trap. I confirmed webhooks was the only current offender, so this PR leaves the tree clean and the suggestion is explicitly **out of scope here** — but a shared guard in the notifications module (resolve every registered type's effective label key against `en.json` and fail on an interpolation placeholder) would turn a recurring bug class into a build failure. Worth a follow-up issue rather than a change to this PR.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — nothing exported changed; `labelKey` is an existing optional field on `NotificationTypeDefinition` (`packages/shared/src/modules/notifications/types.ts:75`) that this definition now populates.
- [x] No field removed or retyped in an existing response shape — `GET /api/notifications/types` keeps returning `labelKey`; only the string it carries for `webhooks.delivery.failed` changes, which is the fix itself. A consumer that keyed off that value would see a different key, but a translation-key value is data, not a contract surface, and the previous value was the defect being repaired.
- [x] No event or message name renamed or removed; no payload field removed — the notification type id `webhooks.delivery.failed`, its `titleKey`, `bodyKey`, and the runtime interpolated title are all untouched, so delivered notifications are byte-identical.
- [x] No database table or column renamed or removed; no column type narrowed — no migration is involved; `notification_types.label_key` is an existing column whose value is re-synced by the existing drift-update path.
- [x] Where a contract had to change: not applicable — no contract changed.

## 🧪 Test Coverage

The new suite (`packages/webhooks/src/modules/webhooks/__tests__/notifications.test.ts`) covers the defect precisely and in a way that fails for the right reason: it asserts the definition exposes both `labelKey` and `titleKey`, that the two are different keys (so the fix cannot be "fixed" by re-pointing the label back at the title), and then, for each of the five locale catalogues, that the label resolves, contains no `{…}` interpolation placeholder, and is not just the untranslated key echoed back. The placeholder assertion is the one that directly encodes issue #5496, and the "not equal to the key" assertion guards the sibling failure mode of a key added to the definition but missing from a catalogue. The file lands in the package's discovered `__tests__` path (`testMatch: <rootDir>/src/**/__tests__/**/*.test.(ts|tsx)`), and the JSON catalogue imports work under the package's jest transform, which I verified by running the suite.

One layer is intentionally not covered here and should be understood rather than added: the test verifies the definition and the catalogues, not the rendered preferences row. The step between them — `syncNotificationTypes` mirroring `labelKey` into `notification_types.label_key`, including the drift-update that repairs existing rows — is already covered by `packages/core/src/modules/notifications/lib/__tests__/notification-type-registry.test.ts:146-157`, which asserts that a changed `labelKey` produces an `UPDATE` with the new `label_key`. Between the two suites the whole path from definition to persisted catalogue row is exercised, and the rendering side (`t(type.labelKey, type.id)`) is already covered by the matrix render tests. I see no gap worth filing.

pkarw:

- 🚀 `merge-queue` — the code review passed with no blocker, major, or minor findings, the full configured validation gate is green on head `ba1f286`, and QA is not required for this change, so the PR is ready to merge.
- 🐛 `bug` — the change fixes issue #5496, where the notification-preferences screen rendered a raw `{webhookName}` interpolation placeholder as the webhooks row label and in three per-channel `aria-label`s.
- ⏭️ `skip-qa` — the automated-verification exemption applies and I verified each of its conditions: the diff touches no UI-rendering file (no `.tsx`, nothing under `packages/ui/src/` or `**/components/**`), the database structure and API surface are unchanged (only the value of the existing `notification_types.label_key` row changes, through the existing drift-update path), no `BACKWARD_COMPATIBILITY.md` contract surface is broken, and automated regression coverage for the changed behaviour ships in the same PR.
- 🔹 `priority-medium` — this is an ordinary user-visible bug with an accessibility edge (a screen reader reads the placeholder aloud), but it causes no outage, data loss, or security exposure and does not block a release.
- 🟡 `risk-medium` — kept as the author set it: an ordinary single-module change that ships with tests, per the repository's risk-inference rule. The blast radius is one notification type's preference label; no schema, auth, money, or shared contract surface is involved.

pkarw:

The review is submitted and the PR is in `merge-queue`. The fix addresses issue #5496 at the correct layer — `webhooks.delivery.failed` now carries a dedicated static `labelKey`, so the catalogue no longer falls back to the parameterised `titleKey` — and I confirmed the two things that could have made it incomplete: `syncNotificationTypes` updates existing `notification_types` rows on drift, so already-deployed tenants are repaired rather than only fresh installs, and a sweep of every notification-type definition in the repository against its `en.json` shows this was the only parameterised title key, so no sibling module is left with the same defect. The full configured validation gate is green on head `ba1f286` (GitHub checks, plus `yarn i18n:check-sync` run locally because no workflow runs it). No blocker, major, or minor findings; two optional nits are recorded in the review body, the more interesting one being a suggested follow-up guard so the `labelKey ?? titleKey` fallback cannot silently reintroduce this bug class in another module.

autofix: skipped (not my PR — re-run with --autofix to fix it here). Nothing needed fixing in any case.

Issue #5019, opened 2026-08-05
sales: line discounts are not idempotent under recalculation, and discountPercent is inert on persisted lines


maxidragon:
Split out of https://github.com/open-mercato/open-mercato/pull/5006 at a maintainer's request — that PR is display-only; these are data-layer defects. Both were confirmed by executing the calculation logic, and a reviewer independently reproduced them by reading `lib/calculations.ts` and `commands/documents.ts`.

## 1. Recalculation is not idempotent — a line discount grows on write, and is already wrong on read

`buildBaseLineResult` reads the input `discountAmount` as **per unit** and multiplies by quantity:

```ts
// packages/core/src/modules/sales/lib/calculations.ts
const discountPerUnit = line.discountAmount ?? (line.discountPercent != null ? … : 0)
const discountTotal = Math.min(Math.max(discountPerUnit * quantity, 0), netSubtotalBeforeDiscount)
```

but the persisted column receives the **line total**:

```ts
// packages/core/src/modules/sales/commands/documents.ts
discountAmount: toNumericString(lineResult.discountAmount) ?? "0",
```

`mapOrderLineEntityToSnapshot` / `mapQuoteLineEntityToSnapshot` then feed that stored total straight back in as if it were per-unit, and **every line upsert recalculates all lines of the document** from those snapshots — so any edit to any line re-multiplies every discounted line's discount by its quantity.

For a line of qty 3 @ 10.00 net with a 5.00 discount:

| recalculation | line net | line discount | document `discountTotalAmount` |
|---|---:|---:|---:|
| initial write | 15.00 | 15.00 | 15.00 |
| after any later line edit | **0.00** | **30.00** | **30.00** |

The line's net collapses (clamped at the line net), its gross stays at the persisted value, and the document's discount total doubles. Editing an unrelated line is enough to trigger it.

### 1b. The order detail page shows the inflated figure with no write at all

Raised by @pkarw in the review of #5006, and confirmed. The `Discounts` row on an order detail page is neither the persisted column nor `totals_snapshot`: `api/documents/factory.ts:528-570` recalculates the whole totals block in an `afterList` hook on every single-order `GET`, via `recalculateOrderTotalsForDisplay` (`commands/returns.ts:204-238`). That helper maps persisted rows through `mapOrderLineEntityToSnapshot` — the same mapper as the write path — so the **same** per-unit misreading applies, on a plain page load, with nothing edited.

Executing the calculation over persisted values:

[...]

PR #5200, opened 2026-08-11
docs(sales): spec the line discount_amount contract (#5019)

maxidragon:
## Summary

`buildBaseLineResult` reads a sales line's `discountAmount` as a **per-unit** figure and multiplies it by quantity, but the persisted `discount_amount` column receives the **line total** that same function returns. Recalculating a document is therefore not idempotent — `mapOrderLineEntityToSnapshot` / `mapQuoteLineEntityToSnapshot` feed the stored total back in as per-unit, so every round trip re-multiplies it by the line quantity.

A second defect decides which of three behaviours a line gets, based on which command last touched it:

| path | coalesce | result |
|---|---|---|
| `orders.create` (`documents.ts:3003`) | `?? null` | `null ?? percent` → percentage runs → **correct** |
| `lines.upsert`, new line (`:7094`, `:7588`) | `?? 0` | `0 ?? percent` → percentage dead → **discount dropped** |
| `lines.upsert`, existing line | `existingSnapshot.discountAmount` | stored line total re-read as per-unit → **re-inflated** |

That asymmetry is why the defect survives a green test suite: it is unreachable from the create path, which is the branch the suite exercises.

Worked through concretely — `quantity: 60`, `unitPriceNet: 50.00`, `discountPercent: 10`, where the correct result is `discount_amount: 300.00` / `total_net_amount: 2700.00`:

| path | stored discount | stored net |
|---|---:|---:|
| `orders.create` | 300.00 | 2700.00 |
| `lines.upsert`, new line | 0.00 | 3000.00 |
| `lines.upsert`, existing line | 3000.00 | **0.00** |

The last row is the idempotency violation in one number: a further round trip through a command that should change nothing collapses the line's net to zero.

This PR is **specification only — no implementation.** #5019 ends with *"Suggested direction (for maintainer input before any work starts)"*, so this writes that direction out far enough to be accepted or rejected on specifics.

## Changes

- Adds `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md`.
- Adds the corresponding row to the `.ai/specs/README.md` directory table.

What the spec settles:

1. **Column contract** — `discount_amount` is the discount for the whole line (net, quantity-inclusive). That is what the write path (`documents.ts:3091`) and the document rollup (`calculations.ts:153`) already assume, and what every correct row already holds.
[...]

Issue #4052, opened 2026-07-09
Non-return order adjustments (discount/custom/surcharge) are not included in the order's recalculated grand total, and have no guard against negative totals


alinadivante:
## Summary

A guard already exists for adjustments of kind "Return" that caps the amount at the order's remaining
grand total, and it works correctly (see "Actual result" below). Adjustments of other kinds (Custom,
Discount, Surcharge, etc.) have no equivalent guard.

Worse: adjustments of any *non-return* kind (`Custom`, `Discount`, `Surcharge`, etc.) are accepted and persisted (they appear correctly as a
line item in the order's detailed totals breakdown), but the order's **top-line grand total
(`Razem (netto)` / `Razem (brutto)` / `Do zapłaty`) is never recalculated to include them** — not even
after a full page reload. This is not just "missing a negative-value guard" — the adjustment's amount
appears to never reach the order's cached grand-total fields at all for non-return kinds, which is a
financial-correctness bug independent of the negative-clamp question.

## Page URL

`/backend/sales/orders/{orderId}` — order detail page, Adjustments section (`Korekty` tab).

## Steps to reproduce

1. Create a sales order with at least one line item, so the order has a non-zero total.
2. In the Adjustments (`Korekty`) tab, click "Dodaj korektę" (Add adjustment).
3. Set **Typ: Custom**, **Etykieta: any label**, **Kwota netto: any negative value whose magnitude
   exceeds the order's current total**.
4. Save.
5. Observe the "Podsumowanie" (Summary) panel's `Razem (netto)` / `Razem (brutto)` / `Do zapłaty`
   values — then click "Pokaż szczegóły" (Show details) to see the itemized breakdown, and finally
   reload the page fully.

## Expected result

- A `Custom`/`Discount`/`Surcharge`-kind adjustment that would push the order total below zero should
  be rejected, same as the existing `return`-kind guard (or, at minimum, the order's grand total should
  actually reflect the new adjustment).
- The order's top-line grand total must always equal the sum of subtotal + tax + shipping + all
  adjustment line items shown in the detail breakdown — it must never silently diverge from its own
  itemization.

## Actual result

Adding a `Custom`-kind adjustment with a negative amount larger than the order's total saves with 0
validation errors (contrast with the `return`-kind test below, which is correctly rejected). Result:
- The itemized breakdown (`Pokaż szczegóły`) correctly lists the adjustment as its own line item.
[...]

Issue #2457, opened 2026-06-04
bug: tax-total equals to zero despite properly calculated line-item tax


pkarw:
I got to Orders -> edit an order, set tax class per item - the item price got properly calculated Net/Gross but the totals still getting 0 in the tax field

Needs to check it for the Quotes too

<img width="1602" height="546" alt="Image" src="https://github.com/user-attachments/assets/727c22a8-71bb-49a3-ab22-2be8688a17ab" />

pkarw:
Fixed alongside #2455 in https://github.com/open-mercato/open-mercato/pull/2467. Root cause: tax-class-priced lines embed tax in gross (gross > net) but may not carry a rate, so the document tax total derived as net*rate=0. Now falls back to the net/gross delta. Verified for orders and quotes.

pkarw:
✅ Fixed by #2467 (https://github.com/open-mercato/open-mercato/pull/2467) — merged into `develop` at 2026-06-04T16:59:53Z (commit `0c7555b`).

Closed automatically by the `sync-merged-pr-issues` skill. Credit to @pkarw.

If this is incorrect, reopen the issue and add the `do-not-close` label so future runs leave it alone.

PR #5550, opened 2026-08-24
fix(sales): treat line discountAmount as a line total, not a per-unit rate (#3757)

### LINKED ISSUE #3757: bug: [Sales] — Order line `discountAmount` is double-applied for quantity > 1 lines, zeroing the line's net contribution to the order subtotal
## Summary

When an order line has a persisted `discount_amount` (a per-**line** total discount, as written by every create/update code path) and `quantity > 1`, the totals-recalculation engine (`buildBaseLineResult` in `lib/calculations.ts`) re-interprets that stored total as a **per-unit** discount and multiplies it by `quantity` again. This inflates the discount far beyond the line's actual net subtotal, which then gets clamped down to `0` — so the line contributes `net = 0` to the order's Subtotal/Grand Total while its `gross` (taken from the already-correct persisted `total_gross_amount`) stays correct. The net/gross split then becomes inconsistent at the **order subtotal** level, which can cascade into the same "net freezes while gross keeps changing" symptom class as issue #3036 — but through a completely different code path (the subtotal aggregator, not the return-credit deriver that PR #3060 patches).

## Page URL

`/backend/sales/documents/<orderId>` → "Pozycje" tab and "Podsumowanie" → "Pokaż szczegóły".

## Steps to reproduce

Reproduced via a seeded demo order (`SO-DEMO-2001`, line "Acoustic panel set": quantity 3, `unit_price_net = 85`, `discount_percent = 5`, persisted `discount_amount = 255.0000`, correct `total_net_amount = 242.25`). I could not find a path in the current admin UI (Add-Line / Edit-Line dialogs, or the Korekty tab) to set a line's own `discountPercent`/`discountAmount` directly — the only discount-like control in the UI is picking a catalog "sale" price, which sets the unit price directly and does not touch `discount_amount`/`discount_percent` at all. This bug is therefore currently reachable via: (a) the module's own seed/example data generator, and (b) any API/import/promotion-engine caller that writes `discountPercent`/`discountAmount` directly onto a multi-quantity `SalesOrderLine`, since the same calc engine (`buildBaseLineResult`) runs on every totals recalculation regardless of how the line was created.

1. Have (or seed) an order line with `quantity = 3`, a non-zero `discount_amount`/`discount_percent`, and a correctly-persisted `total_net_amount`/`total_gross_amount`.
2. Trigger any totals recalculation on the order (e.g. create a shipment, add a payment, or create a return on a different line).
3. Open "Podsumowanie" → "Pokaż szczegóły" on the order detail page.

## Expected result

The order's `Suma częściowa (netto)` (Subtotal net) should be consistent with `Suma częściowa (brutto)` (Subtotal gross) — i.e. a positive subtotal net whenever gross is positive, matching the sum of each line's already-correct `total_net_amount`.

## Actual result

On the reproduction order, the discounted 3-unit line's net contribution comes out as `0` (in a `buildBaseLineResult`-recomputed subtotal) even though its own `total_net_amount` is correctly `242.25` in the database. Combined with subsequent return-adjustment clamping (`Math.max(subtotal + delta, 0)` in `buildBaseDocumentResult`), this can make the order's `Suma częściowa (netto)` show `0,00 USD` while `Suma częściowa (brutto)` shows a large positive number (observed: `0,00 USD` net vs `207,15 USD` gross on the reproduction order, after two returns whose own individual credit amounts were separately verified correct in the database).

## Root cause (for the developer)

In `packages/core/src/modules/sales/lib/calculations.ts`, `buildBaseLineResult`:

```ts
const discountPerUnit =
  line.discountAmount ??
  (line.discountPercent !== null && line.discountPercent !== undefined
    ? toNumber(line.discountPercent, 0) / 100 * toNumber(unitNet, 0)
    : 0)
...
```

`line.discountAmount` is named/used here as a **per-unit** rate and multiplied by `quantity`. But every write path that persists `discountAmount` (`commands/documents.ts` line ~2980: `discountAmount: toNumericString(lineResult.discountAmount) ?? "0"`, and the seed script `seed/examples.ts` lines ~1413/1646, both writing `lineResult.discountAmount`) stores the calculation engine's own `discountTotal` — i.e. an already-multiplied **per-line total**, not a per-unit value. For `quantity === 1` this round-trip bug is invisible (multiplying by 1 is a no-op), which is likely why it has gone unnoticed — it only manifests on lines with `quantity > 1` and a non-zero discount.

## Environment

- Branch: `fix/issue-3036-return-net-total` (PR #3060) — but reproduces identically on `develop`, unrelated to this PR's diff (PR #3060 does not touch `lib/calculations.ts`).
- Module or area: `sales` — order totals / discount calculation
- Browser/Runtime: Chromium (observed via seeded data + DB inspection; UI does not currently expose a way to set this field directly)

## Additional context

Found while investigating an unrelated "Subtotal (net) = 0 while gross > 0" anomaly noticed during manual re-verification of PR #3060 on the seeded order `SO-DEMO-2001`. Confirmed the exact numeric mechanism by hand-tracing `buildBaseLineResult`/`buildBaseDocumentResult` against the order's actual seed data and matching the observed UI numbers to the cent. Lower severity than BUG-001 because it is not currently reachable through the exposed admin UI screens, only via seed data / API / future promotion-engine writers.

wojciechszyjka:
Refs #3757
Spec: `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` (merged in #5200, tracking #5019 — the twin of #3757)

> **⚠️ Status: blocked on two maintainer decisions. Do not merge as-is.**
>
> This PR was opened without the spec above in view, and @pkarw's review (2026-08-24) correctly identified that. The spec covers exactly this function and this defect, is still marked `draft — design decision requested`, and states twice that no implementation lands until its § Proposed Solution 1 and 2 are approved. This PR implements a partial § 1 and stops.
>
> The body below has been corrected: the original "Breaking Changes: None" claim was **wrong**, and the linkage has been changed from `Closes #3757` to `Refs #3757` because the reported symptom does not clear on an already-corrupted row. Both corrections are explained in place. The diagnosis, the fix location and the six tests stand; what needs deciding is the shape of the contract around them.

## 🎯 Goal

Make `sales_order_lines.discount_amount` mean one thing on the read path and the write path, so an order containing a discounted multi-unit line stops reporting a Subtotal (net) that is inconsistent with its gross.

## 🔍 Problem

`SalesOrderLine.discount_amount` is written as a per-**line** total by every writer, but the totals-recalculation engine read it back as a per-**unit** rate. On a line with `quantity > 1` and a non-zero stored discount the recomputed discount was inflated `quantity`-fold, clamped to the line's full pre-discount subtotal, and the line's net contribution collapsed toward `0`, while its gross stayed correct because `buildBaseLineResult` takes gross from the already-correct persisted `total_gross_amount`. With `quantity === 1` the extra multiplication is a no-op, which is why it went unnoticed.

## 🔍 Root Cause

In `packages/core/src/modules/sales/lib/calculations.ts`, `buildBaseLineResult` named its local `discountPerUnit`, seeded it from `line.discountAmount`, and computed `discountTotal = Math.min(Math.max(discountPerUnit * quantity, 0), netSubtotalBeforeDiscount)`.

The `discountPercent` branch was correct — `(percent / 100) * unitNet * quantity` is by definition the line total. The `discountAmount` branch was not, for every value that arrives from a **persisted row**:

- `commands/documents.ts:3124` persists `toNumericString(lineResult.discountAmount)` — this function's own `round(discountTotal)`.
- `seed/examples.ts:1414,1646` write the same value.
- `commands/documents.ts:2906,2937` and `commands/returns.ts:155` rebuild `SalesLineSnapshot` straight from that persisted column.
- `percentAccountsForAmount` in `components/documents/lineItemUtils.ts:35-46` validates a stored amount against `(percent / 100) * unitPriceNet * quantity`, so the display layer also reads the column as a line total.

Because the engine writes its own inflated result back into the same column, the error **compounds across recalculations**: for the reproduction line in #3757 (`quantity 3`, `unit_price_net 85`, `discount_percent 5`) the stored discount grows `12.75 → 38.25 → 114.75 → clamped to 255`, at which point it equals the whole line subtotal and net is exactly `0` — matching the `discount_amount = 255.0000` alongside a correct `total_net_amount = 242.25` observed on the seeded order.

### Correction to the original root-cause claim

The first version of this description asserted that "the calculation engine was the only place in the system interpreting the column as per-unit". That is **not accurate**, and the review was right to flag it. `components/documents/SalesOrderDraftLines.tsx:68` also multiplies by quantity:

```ts
discountAmount: normalizeNumber(payload.discountAmount, 0) * quantity,
```

It is unreachable today — neither caller populates `payload.discountAmount`, so it always evaluates `0 * quantity` — but it is a second site holding the per-unit reading. Whether it is a defect depends on Decision 1/3 below: if caller-supplied amounts keep their per-unit meaning (the spec's § 3), this line is **correct** as written and must stay; only if caller input flips to line-total does it become wrong. It is therefore listed here as decision-dependent rather than fixed in this PR.

## What Changed

- `packages/core/src/modules/sales/lib/calculations.ts` — in `buildBaseLineResult`, `netSubtotalBeforeDiscount` is computed before the discount is resolved; the `discountPercent` branch derives from it directly; and the clamp no longer multiplies by `quantity`.

The `discountPercent` path is algebraically identical to before, so every pre-existing expectation holds bit for bit.

## 💥 Breaking Changes — corrected

**The original "None" claim was wrong.** It reasoned only about the persistence direction and missed the create path.

`commands/documents.ts:3036` (`createLineSnapshotFromInput`) passes caller-supplied `discountAmount` from `DocumentLineCreateInput` — the validated request body of `POST /api/sales/orders`, `POST /api/sales/quotes`, and `POST`/`PUT` on `/api/sales/order-lines` and `/api/sales/quote-lines` — straight into `buildBaseLineResult` unmodified. Under the old engine that value was multiplied by quantity, so the API input was *de facto* per-unit. As this PR stands, it silently becomes a line total:

| caller posts `quantity: 60, unitPriceNet: 50.00, discountAmount: 5.00` | discount applied | stored `total_net_amount` |
|---|---:|---:|
| before this PR | 300.00 | 2700.00 |
| as this PR stands | 5.00 | 2995.00 |

A silent 60× change in the direction of undercharging the discount, with no 4xx, no warning and no version gate. `BACKWARD_COMPATIBILITY.md` classifies API request semantics as a contract surface and its deprecation protocol was not followed; `UPGRADE_NOTES.md` has no entry.

This is exactly what the spec's § 3 exists to prevent, via a `discountAmountBasis` flag (caller-supplied, defaulting to `'unit'`) paired with `discountAmountFromStoredRow` (mapper-only), so that only mapper-rebuilt snapshots are read as line totals and no existing caller has to change. That is not implemented here.

## ⚠️ Why the linkage is now `Refs`, not `Closes`

The original body claimed recalculated totals move "always toward the correct value". For an **already-corrupted** row that is false. Feeding the reproduction row (`discount_amount = 255`, `quantity 3 × unitPriceNet 85`) through the patched function:

```
netSubtotalBeforeDiscount = 255
discountRequested         = 255      // stored line total, no longer multiplied
discountTotal             = min(max(255, 0), 255) = 255
netSubtotal               = max(255 - 255, 0)     = 0
```

Still `0`. This PR stops new corruption but does not undo existing corruption, and the row cannot self-heal because `discountAmount` still wins over `discountPercent` via `??` even though the correct `discount_percent = 5` sits in the same row. The issue's Expected result therefore does not clear on `SO-DEMO-2001` or on any database where the buggy engine already ran, so `Closes #3757` would have been untrue. Self-healing is the spec's § 2 (percentage-first precedence), which is Decision 2 below.

## 🧪 Tests

Full validation gate run locally on this tree, in order, none skipped: `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test` (184 suites, 1998 passed / 5 skipped), `yarn build:app`. All 14 reported GitHub checks on `0a271d08` are green.

Six unit tests added to `packages/core/src/modules/sales/lib/__tests__/calculations.test.ts`: a stored line-total discount on a qty-3 line; the same at qty 1; `discountPercent`-only at qty 1 and qty 4; a stored discount exceeding the line subtotal (clamping); a five-pass round trip feeding the persisted discount back in, which is the regression for the compounding runaway; and a document-level test driving `buildBaseDocumentResult`'s `Math.max(subtotal + delta, 0)` return clamp. Three of the six fail against the unfixed engine (verified by stashing only the source change: `3 failed, 16 passed`).

These tests are written against the current partial shape. If Decision 1/3 lands the spec's basis flag, the caller-input cases will need to move to the mapper-sourced form.

## 📋 Outstanding — the two gated decisions

Per the spec, both need maintainer sign-off before an implementation lands:

1. **Decision 1 — the column's meaning.** `discount_amount` is a line total (spec § 1). This PR assumes it; the spec still lists it as unsigned-off.
2. **Decision 2 — percentage-first precedence** (spec § 2), which is what makes corrupted rows self-heal. Its documented cost is that a caller sending only `discountAmount` on `PUT /api/sales/order-lines`, against a row that carries a non-zero stored `discount_percent`, silently loses the amount it just sent.

Given a decision, the remaining work is the spec's § 2 (precedence), § 3 (`discountAmountBasis` / `discountAmountFromStoredRow`, plus de-duplicating the mapper shared with `commands/returns.ts:137`), and § 4 (decomposing the `parsed.discountAmount ?? existingSnapshot?.discountAmount ?? 0` coalescing at `documents.ts:7177` and `:7671`), together with an `UPGRADE_NOTES.md` entry and a decision on repairing already-corrupted rows.

### DIFF
--- a/packages/core/src/modules/sales/lib/__tests__/calculations.test.ts
+++ b/packages/core/src/modules/sales/lib/__tests__/calculations.test.ts
@@ -575,4 +575,172 @@ describe('calculateDocumentTotals', () => {
       unregister()
     }
   })
+
+  it('applies a stored discountAmount as the line total, not per unit, on quantity > 1 lines (issue #3757)', async () => {
+    const result = await calculateDocumentTotals({
+      documentKind: 'order',
+      lines: [
+        {
+          kind: 'product',
+          quantity: 3,
+          currencyCode: 'USD',
+          unitPriceNet: 85,
+          discountPercent: 5,
+          discountAmount: 12.75,
+          taxRate: 0,
+        },
+      ],
+      adjustments: [],
+      context: { ...baseContext, metadata: {} },
+    })
+
+    expect(result.lines[0].discountAmount).toBeCloseTo(12.75, 4)
+    expect(result.lines[0].netAmount).toBeCloseTo(242.25, 4)
+    expect(result.totals.subtotalNetAmount).toBeCloseTo(242.25, 4)
+    expect(result.totals.discountTotalAmount).toBeCloseTo(12.75, 4)
+  })
+
+  it('applies a stored discountAmount unchanged on quantity 1 lines', async () => {
+    const result = await calculateDocumentTotals({
+      documentKind: 'order',
+      lines: [
+        {
+          kind: 'product',
+          quantity: 1,
+          currencyCode: 'USD',
+          unitPriceNet: 85,
+          discountAmount: 12.75,
+          taxRate: 0,
+        },
+      ],
+      adjustments: [],
+      context: { ...baseContext, metadata: {} },
+    })
+
+    expect(result.lines[0].discountAmount).toBeCloseTo(12.75, 4)
+    expect(result.lines[0].netAmount).toBeCloseTo(72.25, 4)
+    expect(result.totals.subtotalNetAmount).toBeCloseTo(72.25, 4)
+  })
+
+  it('keeps discountPercent-only lines scaling with quantity', async () => {
+    const result = await calculateDocumentTotals({
+      documentKind: 'order',
+      lines: [
+        {
+          kind: 'product',
+          quantity: 1,
+          currencyCode: 'USD',
+          unitPriceNet: 100,
+          discountPercent: 10,
+          taxRate: 0,
+        },
+        {
+          kind: 'product',
+          quantity: 4,
+          currencyCode: 'USD',
+          unitPriceNet: 100,
+          discountPercent: 10,
+          taxRate: 0,
+        },
+      ],
+      adjustments: [],
+      context: { ...baseContext, metadata: {} },
+    })
+
+    expect(result.lines[0].discountAmount).toBeCloseTo(10, 4)
+    expect(result.lines[0].netAmount).toBeCloseTo(90, 4)
+    expect(result.lines[1].discountAmount).toBeCloseTo(40, 4)
+    expect(result.lines[1].netAmount).toBeCloseTo(360, 4)
+    expect(result.totals.subtotalNetAmount).toBeCloseTo(450, 4)
+    expect(result.totals.discountTotalAmount).toBeCloseTo(50, 4)
+  })
+
+  it('clamps a stored discountAmount that exceeds the line subtotal', async () => {
+    const result = await calculateDocumentTotals({
+      documentKind: 'order',
+      lines: [
+        {
+          kind: 'product',
+          quantity: 2,
+          currencyCode: 'USD',
+          unitPriceNet: 50,
+          discountAmount: 500,
+          taxRate: 0,
+        },
+      ],
+      adjustments: [],
+      context: { ...baseContext, metadata: {} },
+    })
+
+    expect(result.lines[0].discountAmount).toBeCloseTo(100, 4)
+    expect(result.lines[0].netAmount).toBeCloseTo(0, 4)
+  })
+
+  it('stays stable when the persisted discountAmount is fed back into repeated recalculations (issue #3757)', async () => {
+    let storedDiscountAmount: number | null = null
+
+    for (let pass = 0; pass < 5; pass += 1) {
+      const result = await calculateDocumentTotals({
+        documentKind: 'order',
+        lines: [
+          {
+            kind: 'product',
+            quantity: 3,
+            currencyCode: 'USD',
+            unitPriceNet: 85,
+            discountPercent: 5,
+            discountAmount: storedDiscountAmount,
+            taxRate: 0,
+          },
+        ],
+        adjustments: [],
+        context: { ...baseContext, metadata: {} },
+      })
+
+      expect(result.lines[0].discountAmount).toBeCloseTo(12.75, 4)
+      expect(result.lines[0].netAmount).toBeCloseTo(242.25, 4)
+      storedDiscountAmount = result.lines[0].discountAmount
+    }
+  })
+
+  it('keeps the order subtotal net consistent with gross after return credits on a discounted multi-unit line (issue #3757)', async () => {
+    const result = await calculateDocumentTotals({
+      documentKind: 'order',
+      lines: [
+        {
+          kind: 'product',
+          quantity: 3,
+          currencyCode: 'USD',
+          unitPriceNet: 85,
+          discountPercent: 5,
+          discountAmount: 12.75,
+          taxRate: 0,
+          totalNetAmount: 242.25,
+          totalGrossAmount: 242.25,
+        },
+      ],
+      adjustments: [
+        {
+          scope: 'line',
+          kind: 'return',
+          amountNet: 20.1,
+          amountGross: 20.1,
+          currencyCode: 'USD',
+        },
+        {
+          scope: 'line',
+          kind: 'return',
+          amountNet: 15,
+          amountGross: 15,
+          currencyCode: 'USD',
+        },
+      ],
+      context: { ...baseContext, metadata: {} },
+    })
+
+    expect(result.totals.subtotalNetAmount).toBeCloseTo(207.15, 4)
+    expect(result.totals.subtotalGrossAmount).toBeCloseTo(207.15, 4)
+    expect(result.totals.grandTotalNetAmount).toBeCloseTo(207.15, 4)
+    expect(result.totals.grandTotalGrossAmount).toBeCloseTo(207.15, 4)
+  })
 })
--- a/packages/core/src/modules/sales/lib/calculations.ts
+++ b/packages/core/src/modules/sales/lib/calculations.ts
@@ -85,14 +85,25 @@ function buildBaseLineResult(line: SalesLineSnapshot): SalesLineCalculationResul
     (line.unitPriceGross !== null && line.unitPriceGross !== undefined
       ? toNumber(line.unitPriceGross) / (1 + taxRate)
       : 0)
-  const discountPerUnit =
+  const netSubtotalBeforeDiscount = toNumber(unitNet, 0) * quantity
+  // `discountAmount` is a per-LINE total everywhere it is produced or consumed:
+  // the write path persists this function's own `discountTotal` back into it and
+  // the UI validates it against `percent * unitPriceNet * quantity`. Reading it
+  // as a per-unit rate and multiplying by `quantity` inflated the discount
+  // `quantity`-fold on multi-unit lines, clamping their net contribution to 0
+  // while gross stayed correct (#3757). The percent branch is unchanged:
+  // `percent / 100 * unitNet * quantity` is the same value as
+  // `percent / 100 * netSubtotalBeforeDiscount`.
+  const discountRequested =
     line.discountAmount ??
     (line.discountPercent !== null && line.discountPercent !== undefined
-      ? toNumber(line.discountPercent, 0) / 100 * toNumber(unitNet, 0)
+      ? toNumber(line.discountPercent, 0) / 100 * netSubtotalBeforeDiscount
       : 0)
 
-  const netSubtotalBeforeDiscount = toNumber(unitNet, 0) * quantity
-  const discountTotal = Math.min(Math.max(discountPerUnit * quantity, 0), netSubtotalBeforeDiscount)
+  const discountTotal = Math.min(
+    Math.max(toNumber(discountRequested, 0), 0),
+    netSubtotalBeforeDiscount
+  )
   const netSubtotal = Math.max(netSubtotalBeforeDiscount - discountTotal, 0)
   const explicitTaxAmount = line.taxAmount !== null && line.taxAmount !== undefined
   let taxAmount = explicitTaxAmount


wojciechszyjka:
## 🏷️ Label rationale

- 🔍 **`review`** — this PR is open, ready for review, and not in any other pipeline state, so it starts in the review stage of the pipeline. Pipeline labels are mutually exclusive and no other one applies yet.
- 🐛 **`bug`** — the change fixes a defect reported in #3757: the totals engine re-applied a stored per-line discount once per unit, producing an order Subtotal (net) inconsistent with its gross. It adds no new capability, so `bug` is the correct category rather than `feature` or `refactor`.
- ✅ **`skip-qa`** — applied under the automated-verification exemption, and every condition of that exemption is met. The diff touches exactly two files, both pure TypeScript under `packages/core/src/modules/sales/lib/`: no `.tsx` outside tests, nothing under `packages/ui/src/`, nothing under any `components/` directory. The database structure is unchanged (no entity, migration, or snapshot edit), the API surface is unchanged (no route, validator, or response-shape edit), and no contract surface listed in `BACKWARD_COMPATIBILITY.md` is broken — the meaning of the stored `discount_amount` column is not altered, only the calculation engine's reading of it is corrected to match what every writer already stores. Automated tests for the changed behavior ship in this same PR: six unit tests, three of which fail against the unfixed engine. There is accordingly no user-facing surface for a manual QA pass to exercise that the unit tests do not already cover, so `needs-qa` is not applied and the two are never combined.
- ⚠️ **`priority-medium`** — the defect corrupts financial figures shown to operators, which argues for urgency, but the reporter established it is not currently reachable through any exposed admin UI screen: the Add-Line and Edit-Line dialogs and the Korekty tab offer no way to set a line's own `discountPercent` / `discountAmount`. It is reachable today only via the module's seed/example generator and via API, import, or promotion-engine callers that write those fields directly onto a multi-quantity line. Real but bounded exposure puts this above `priority-low` and below `priority-high`.
- ⚠️ **`risk-medium`** — the blast radius is the wider consideration here. `buildBaseLineResult` runs on every totals recalculation for every sales document kind, so a mistake would propagate broadly. Weighed against that, the change itself is small and tightly bounded: the `discountPercent` branch is algebraically identical to the previous expression, so lines without a stored `discountAmount` are byte-for-byte unaffected, and the clamp guarding against an over-large discount is retained. The full validation gate passed with no commands skipped. That combination — wide reach, narrow and well-tested change — lands at medium rather than high.

Priority and risk are orthogonal and were assessed separately; both independently landed at medium, so there was no conflict to resolve upward.

wojciechszyjka:
🤖 `om-open-pr` — chain lock transferred to this PR. The `om-auto-fix-issue` chain that opened it continues here with `om-auto-review-pr`, so the `in-progress` label and assignee on this PR are a live claim, not leftovers. Other auto-skills should skip this PR until the lock is released. The issue lock on #3757 is being released now that the work lives on this PR.

wojciechszyjka:
🤖 `om-auto-review-pr` taking over the existing chain lock on this PR at the hand-off from `om-open-pr`. The lock stays held for the duration of the review rather than being released and re-acquired, so other auto-skills should continue to skip this PR. Reviewing with `--autofix`, so any actionable finding is fixed on this branch in the same run.

wojciechszyjka (commented):
# 🔍 Code Review

**Verdict: APPROVE** — no blockers and no un-waivered majors. Two minor observations and one nit are recorded below for the author to pick up or consciously decline; none of them gate the merge.

**One procedural caveat on that verdict:** this PR was authored by the same account running the review, and GitHub does not permit approving your own pull request, so this report is submitted as a review comment rather than as a formal approving review. The pipeline label therefore stays at `review` rather than moving to `merge-queue` — an independent approving review is still required before this can merge, and nothing here should be read as satisfying that requirement.

## ✅ Validation

The full configured gate from `.ai/agentic.config.json` was run locally against this exact tree (no compose `app` container was up, so local mode), in order, with no command skipped:

| Command | Result |
|---|---|
| `yarn generate` | ✅ pass — produced no working-tree changes, confirming the diff touches no generated registry |
| `yarn i18n:check-sync` | ✅ pass |
| `yarn typecheck` | ✅ pass |
| `yarn test` | ✅ pass — 184 suites, 1998 passed / 5 skipped |
| `yarn build:app` | ✅ pass |

One note on how that gate was obtained, for reproducibility rather than as a finding: the first `yarn test` run failed in `@open-mercato/shared` on `dynamicLoader.generatedCacheRecovery.test.ts` with `listen EINVAL` on a `tsx` IPC pipe. That is an environment artifact of running inside a linked git worktree — turbo executes tasks in strict env mode and drops `TMPDIR`, so `tsx` tries to open its pipe under a path it cannot bind. Re-running with `TURBO_ENV_MODE=loose` so the variable propagates gives a clean pass, and the suite in question lives in a different package from this diff and is untouched by it. Nothing in this PR causes or is masked by that failure.

## 🧪 Test coverage

Coverage is proportionate and, more importantly, *load-bearing* — it was verified to actually discriminate. Stashing only the source change and re-running the suite yields `3 failed, 16 passed`, so three of the six new cases genuinely fail against the unfixed engine rather than merely passing in both states:

- `applies a stored discountAmount as the line total, not per unit, on quantity > 1 lines` — fails without the fix.
- `stays stable when the persisted discountAmount is fed back into repeated recalculations` — fails without the fix. This is the most valuable of the six: the defect's real severity comes from the engine writing its own inflated result back into the same column, so the error compounds across recalculations rather than staying a fixed overshoot. A single-pass test would have proven the arithmetic without proving the round trip converges, and this one closes that gap.
- `keeps the order subtotal net consistent with gross after return credits on a discounted multi-unit line` — fails without the fix, and is the only case that exercises `buildBaseDocumentResult`'s `Math.max(subtotal + delta, 0)` return clamp, which is where the reported `0,00` net against a positive gross actually became visible to the operator.

The other three (`quantity 1` unchanged, `discountPercent`-only scaling, over-subtotal clamping) pass in both states by design. That is correct and not a weakness: they are guards pinning the behavior the fix must *not* disturb, and the percent-scaling one in particular is what demonstrates the change is confined to the `discountAmount` branch.

## 🔍 Correctness

The core claim of the change — that `discountAmount` is a per-line total on every side except the calculation engine's read — holds up against the code:

- `packages/core/src/modules/sales/commands/documents.ts:3124` persists `toNumericString(lineResult.discountAmount)`, which is `buildBaseLineResult`'s own `round(discountTotal)`.
- `packages/core/src/modules/sales/seed/examples.ts:1414,1646` write the same value.
- `percentAccountsForAmount` in `packages/core/src/modules/sales/components/documents/lineItemUtils.ts:35-46` validates a stored amount against `(percent / 100) * unitPriceNet * quantity`, i.e. the display layer also reads the column as a line total.
- `packages/core/src/modules/sales/commands/documents.ts:2906,2937` and `commands/returns.ts:155` rebuild `SalesLineSnapshot` straight from the persisted column, which is exactly the path that was being misread.

I found no caller anywhere in the module that passes a genuine per-unit value into `SalesLineSnapshot.discountAmount`, so there is no consumer this correction breaks. `data/validators.ts:342,925` only constrain the field to a non-negative decimal and impose no per-unit interpretation.

The percent branch is algebraically identical to its predecessor (`percent / 100 * unitNet * quantity` ≡ `percent / 100 * netSubtotalBeforeDiscount`), which the unchanged pre-existing expectations at `calculations.test.ts:66` and `:102` confirm empirically — they still assert `1` and `4.5` and still pass.

The clamp is preserved in both directions: `Math.max(…, 0)` still rejects a negative stored discount and `Math.min(…, netSubtotalBeforeDiscount)` still caps it at the line subtotal. Only the `* quantity` is gone, which is precisely the defect.

Two incidental robustness improvements are worth recording so they are not mistaken for accidents later. Wrapping the resolved value in `toNumber(discountRequested, 0)` means a non-finite stored value now degrades to a zero discount instead of propagating `NaN` through `discountTotal` into `netSubtotal` and out to the document totals, which is what the previous bare `discountPerUnit * quantity` would have done. And the compounding runaway described above is structurally impossible now that the value read and the value written share one meaning.

## 📋 Findings

### Minor — stale `discountAmount` still wins over an updated `discountPercent` (pre-existing, correctly left out of scope)

`packages/core/src/modules/sales/lib/calculations.ts:96` — the `??` means a non-null stored `discountAmount` always takes precedence over `discountPercent`. Combined with `packages/core/src/modules/sales/commands/documents.ts:7177` and `:7671`, which resolve `discountAmount: parsed.discountAmount ?? existingSnapshot?.discountAmount ?? 0`, a line update that changes only `discountPercent` or `unitPriceNet` — without also sending an explicit `discountAmount` — keeps the previously stored absolute discount and silently ignores the new percentage.

I am flagging this rather than requesting it because it is **not a regression and not this PR's defect**: the same precedence existed before, where it was additionally multiplied by quantity, so this change strictly improves the symptom. Folding a decision about which of the two fields is authoritative into a PR whose whole thesis is "one field, one meaning" would mix two independent behavior changes in a single diff and make the fix harder to reason about and to revert. The author has already documented it under "Follow-up (deliberately out of scope)" in the PR description, which is the right disposition. It does deserve its own issue so it is not lost.

### Minor — no test pins the `discountAmount` / `discountPercent` precedence

`packages/core/src/modules/sales/lib/__tests__/calculations.test.ts` — the first new test supplies both `discountPercent: 5` and `discountAmount: 12.75` on the same line, so precedence is exercised, but only for values that happen to agree; nothing would fail if the precedence flipped. Deliberately leaving it unpinned is defensible here, since the follow-up above may well change that precedence on purpose and a test asserting current behavior would then read as a contract rather than as a snapshot. Recording it so the choice is visible rather than accidental.

### Nit — the explanatory comment is long for the file's convention

`packages/core/src/modules/sales/lib/calculations.ts:87-95` — eight lines of comment above a three-line expression. The file's established convention (see the `#2457`, `#1905`, `#1705` and `#4052` notes elsewhere in it) is a short issue-referencing explanation of a non-obvious clamp, and the repository's own guidance prefers self-documenting code over inline commentary. The content here earns its place, because the round-trip invariant genuinely is not visible from the expression alone and is exactly what the next reader needs to avoid reintroducing the bug — but the algebraic-equivalence sentence in the last two lines duplicates what the unchanged passing tests already prove and could go. Purely cosmetic; not worth another commit on its own.

## 💥 Breaking changes

None. No contract surface from `BACKWARD_COMPATIBILITY.md` is touched: no auto-discovery file, public type, exported signature, import path, event id, widget spot id, API route, DB schema, DI key, ACL feature, notification id, CLI command, or generated file. The database column's meaning is unchanged — only the engine's reading of it is corrected to match what every writer already stores. `SalesLineSnapshot.discountAmount` keeps its `number | null | undefined` shape and its non-negative-decimal validation.

Recomputed totals do change for existing lines with a non-zero `discount_amount` and `quantity > 1`, but always toward the correct value, and those lines' persisted `total_net_amount` / `total_gross_amount` are already correct — so the effect is that recalculation stops diverging from what is stored.

## ⏳ CI status

CI has not settled at review time. `license/cla`, `scope`, `audit-scope`, `lint`, `ds-lint`, `Analyze (actions)` and `Analyze (python)` have passed; `Analyze (javascript-typescript)` and `prepare` were still pending when this review was submitted. No required check is failing. This review was deliberately submitted without waiting for those, so the author has the complete picture in one pass; the pending checks are disclosed here so this PR is self-documenting even if no further automated follow-up arrives. Required checks still gate the merge — the local gate above is this run's evidence, not a substitute for branch protection.

wojciechszyjka:
🤖 `om-auto-fix-issue` run complete: the fix is implemented, the full validation gate passed locally with no commands skipped, and the automated review returned an APPROVE verdict with no blockers and no un-waivered majors — two minor observations and one nit are recorded in the review for the author to pick up or decline.

The chain lock is released: `in-progress` has been swapped for `ci-monitoring`, so this PR is free for any other skill or human to act on while the remaining CI checks are watched. `ci-monitoring` is a meta label and not a claim.

**Pipeline state deliberately left at `review`, not `merge-queue`.** The automated verdict was approve, but this PR was authored by the same account that reviewed it, and GitHub does not allow approving your own pull request — so the review was submitted as a comment and an independent approving review is still required before this can merge. Moving it to `merge-queue` would misrepresent it as ready.

`skip-qa` stands under the automated-verification exemption: the diff is two pure TypeScript files under `packages/core/src/modules/sales/lib/`, with no `.tsx` outside tests, nothing under `packages/ui/src/` or any `components/` directory, no database or API surface change, no broken `BACKWARD_COMPATIBILITY.md` contract, and automated tests for the changed behavior shipping in the same PR. No manual UI QA pass is applicable, so `om-auto-qa-pr` was not run.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-24T05:48:12Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review — `fix(sales): treat line discountAmount as a line total, not a per-unit rate (#3757)`

**Verdict: REQUEST CHANGES** — 2 blockers, 1 major, 1 minor, 1 nit.

**Validation source: GitHub checks.** All 14 reported checks on head `0a271d08` are green (`prepare`, `lint`, `ds-lint`, `test`, `docker-build`, `ephemeral-integration (none)`, `documents-multi-instance`, `audit-scope`, `scope`, `Analyze (javascript-typescript)`, `Analyze (actions)`, `Analyze (python)`, `CodeQL`, `license/cla`); 5 are `SKIPPED`; none failing, none pending. No local re-run of `validation.commands` was needed. The base is current — head sits directly on `develop@94f1126c1`, `mergeable: MERGEABLE`.

The diagnosis in this PR is correct and unusually well argued. `buildBaseLineResult` really did read `discount_amount` as per-unit while every writer stored a line total, the error really did compound across recalculations, and the six added tests are well chosen — the repeated-recalculation test in particular is the right regression for the runaway. Fixing it inside the shared `buildBaseLineResult` also covers the order path, the quote path, the display recalc and the three persisting return flows at one point, which is a genuine advantage over a per-mapper fix.

The problem is not the diagnosis. It is that this change is **the first half of an already-written, explicitly gated spec**, and shipping that half alone flips a public API input's meaning and leaves the reported symptom in place on every affected database.

---

## Findings

### 🔴 Blocker 1 — Lands a spec that is explicitly gated on maintainer sign-off, and lands only part of it

`.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` (merged in #5200 on 2026-08-21, tracking the now-closed #5019, which is #3757's twin) covers exactly this function, exactly this defect, and exactly this decision. Its status line still reads **`draft — design decision requested`**, and it says, twice:

> It is spec-only: no implementation lands until the contract below is approved.

> **No implementation should land until § Proposed Solution 1 and 2 are approved**, because both change observable behaviour on an unversioned contract.

Root `AGENTS.md` — *"Check `.ai/specs/` and `.ai/specs/enterprise/` for existing specs before modifying a module"* and *"Ask before … changing public contracts"*. `.ai/specs/AGENTS.md` — *"Check both OSS and enterprise spec directories before modifying a module."* Neither the PR body nor the diff nor the commit references the spec, #5019, or #5200.

The PR implements the spec's **§ Proposed Solution 1** (the column is a line total) and stops. It omits:

| spec section | what it specifies | in this PR |
|---|---|---|
| § 2 | percentage-first precedence, stored `0` treated as absent | ❌ |
| § 3 | `discountAmountBasis` (caller-only, defaults `'unit'`) + `discountAmountFromStoredRow` (mapper-only) | ❌ |
| § 4 | decompose `parsed.discountAmount ?? existingSnapshot?.discountAmount ?? 0` per operand at `documents.ts:7177` and `:7671` | ❌ |
| § 3 (dedup) | extract the duplicated `mapOrderLineEntityToSnapshot` shared with `commands/returns.ts:137` | ❌ |

§§ 2–4 are not decoration on top of § 1 — the spec's § Row reconciliation and § Migration & Backward Compatibility both hang off them, and Findings 2 and 3 below are the direct consequence of taking § 1 without them.

**Ask:** either link the spec, get Decisions 1 and 2 signed off, and implement the approved shape; or state explicitly in the PR body why a § 1-only change is the right increment and what it deliberately leaves for the spec's implementation PR. Right now the spec is invisible from the PR, so a reader cannot tell this was a scope decision rather than an oversight.

---

### 🔴 Blocker 2 — Undocumented breaking change to caller-supplied `discountAmount` on a public API request field

The PR body states:

> ## 💥 Breaking Changes
> None. […] No caller passes a genuine per-unit value into `SalesLineSnapshot.discountAmount`: `commands/documents.ts:7176,7670` pass the parsed API or existing-snapshot value straight through […]

That "straight through" is exactly the break. The **create** path is the one the analysis skips:

`packages/core/src/modules/sales/commands/documents.ts:3036`

```ts
discountAmount: line.discountAmount ?? null,
```

`line` here is `DocumentLineCreateInput` — caller request body, validated by `linePricingSchema` (`data/validators.ts:342`) and reached from `POST /api/sales/orders`, `POST /api/sales/quotes`, `POST/PUT /api/sales/order-lines` and `/api/sales/quote-lines`. The value goes into `buildBaseLineResult` unmodified. So for a caller posting a line with `quantity: 60, unitPriceNet: 50.00, discountAmount: 5.00`:

| | line discount applied | stored `total_net_amount` |
|---|---:|---:|
| before this PR | 300.00 | 2700.00 |
| after this PR | 5.00 | 2995.00 |

Silent, 60×, in the direction of undercharging the discount. No 4xx, no warning, no version gate.

This is precisely what the spec's § 3 exists to prevent. Its producer table assigns `createLineSnapshotFromInput (:3006)` and the `parsed.*` operand of both upsert sites a basis of **`'unit'` when omitted**, with the stated rationale:

> today's documented API input meaning; **existing callers unaffected**

and § API Contracts: *"Omitting the field reproduces today's documented input meaning exactly, so no existing caller has to change."*

`BACKWARD_COMPATIBILITY.md` classifies API request/response semantics as a contract surface, and the deprecation protocol (`@deprecated` → bridge → `UPGRADE_NOTES.md` entry → spec with a Migration section) was not followed. `UPGRADE_NOTES.md` has no entry for this.

**Ask:** implement § 3's `discountAmountBasis` so a caller-supplied amount keeps its per-unit meaning by default and only mapper-rebuilt snapshots are read as line totals — or, if the maintainers decide the input meaning should flip outright, say so in the PR body, add the `UPGRADE_NOTES.md` entry, and drop the "Breaking Changes: None" claim. Either way this cannot ship as "None".

---

### 🟠 Major 3 — `Closes #3757`, but #3757's own reproduction still fails after this fix

The PR body says recalculated totals move *"always toward the correct value — those lines' persisted `total_net_amount` / `total_gross_amount` are already right, so the fix makes recalculation agree with what is stored."* For an already-corrupted row that is not true, and the PR's own root-cause section proves it: it correctly traces the compounding `12.75 → 38.25 → 114.75 → clamped to 255`, and notes this matches the `discount_amount = 255.0000` alongside `total_net_amount = 242.25` on the seeded order in the issue.

Run that row through the patched `buildBaseLineResult` (`calculations.ts:88-107`), `quantity 3 × unitPriceNet 85`:

```
netSubtotalBeforeDiscount = 255
discountRequested         = 255          // stored line total, no longer multiplied
discountTotal             = min(max(255, 0), 255) = 255
netSubtotal               = max(255 - 255, 0)     = 0
```

Still `0`. The issue's **Expected result** — *"a positive subtotal net whenever gross is positive, matching the sum of each line's already-correct `total_net_amount`"* — is still not met on `SO-DEMO-2001` or on any database where the buggy engine already ran. The fix stops new corruption; it does not undo existing corruption, and the row cannot self-heal because `discountAmount` still wins over `discountPercent` via `??` even though the correct `discount_percent = 5` is sitting right there in the same row.

This is not an incidental gap — it is the spec's § Row reconciliation table, and both self-healing buckets are marked **yes** *only under § 2's percentage-first precedence*, which this PR does not implement:

| bucket | heals automatically? | how |
|---|---|---|
| dropped (`discount_amount = 0`, `discount_percent > 0`) | **yes** | percentage-first (§ 2) |
| re-inflated (`discount_amount >` qty-correct value, `discount_percent > 0`) | **yes** | § 2 ignores the stored amount and re-derives from the percent |
| amount-only, no percent, re-inflated | **no** | needs the opt-in operator CLI |

Without § 2, all three buckets are **no**.

**Ask:** either implement § 2 so buckets 1 and 2 self-heal, or ship a bounded repair (the persisted `total_net_amount` is authoritative and correct, so `discount_amount := max(unit_price_net × quantity − total_net_amount, 0)` recovers `12.75` exactly on the reproduction row) — and, whichever you choose, drop `Closes #3757` in favour of `Refs #3757` until the reported symptom actually clears on an affected database, or state in the PR body that a follow-up issue tracks the data repair and link it.

---

### 🟡 Minor 4 — One place still reads `discountAmount` as per-unit, and it is not the engine

The PR body's root-cause section asserts:

> So the calculation engine was the only place in the system interpreting the column as per-unit.

`packages/core/src/modules/sales/components/documents/SalesOrderDraftLines.tsx:68`, in `createSalesOrderLineDraft`:

```ts
discountAmount: normalizeNumber(payload.discountAmount, 0) * quantity,
```

The `record` built here (used for the draft-lines grid) multiplies by `quantity`, while `payload` — forwarded verbatim to the create API at `SalesOrderDraftLines.tsx:55` — is now read as a line total. That is the same double-meaning this PR just removed from the engine, still live in the draft-line UI, and it survives precisely because it is unreachable today: neither caller populates `discountAmount` (`LineItemDialog` has no discount control at all, and `SalesDocumentForm.tsx:1380-1390`'s inbox pre-fill omits it), so the expression always evaluates `0 * quantity`.

Unreachable today, but the moment a promotion engine, importer or inbox pre-fill starts populating that payload, the grid and the persisted totals disagree by exactly `quantity`. Either delete the `* quantity` (the payload is a line total now) or drop the field from the draft record until something actually supplies it.

---

### ⚪ Nit 5 — Two test comments now describe the old per-unit round trip

- `packages/core/src/modules/sales/components/documents/__tests__/lineItemUtils.test.ts:40-41` — *"discountAmount 5 was supplied per unit, so 15.00 is persisted for a 3 × 10.00 line"*
- `packages/core/src/modules/sales/components/documents/__tests__/ItemsSection.discountColumn.test.tsx:166-168` — same wording

Both assertions still pass and are still meaningful (a stored `15` against `discount_percent 15` on a 3 × 10.00 line is a legitimate data shape). Only the explanation of how such a row arises is now wrong: after this PR a caller-supplied `5` persists `5`, not `15`. Worth a one-line touch-up in the same PR that pins the invariant.

---

## Checklist notes (no findings)

- **Tests** — six unit tests ship with the behaviour change; three fail against the unfixed engine, as the PR body documents. Good coverage of the branch that changed.
- **Percent branch** — `percent/100 × unitNet × quantity` → `percent/100 × netSubtotalBeforeDiscount` is algebraically identical, and `round()` to 4dp absorbs the float-association difference. Existing expectations (`calculations.test.ts` qty-3 / 10% → `4.5`) hold.
- **Clamp** — `Math.min(Math.max(…, 0), netSubtotalBeforeDiscount)` is preserved; only the `× quantity` is gone. The new `toNumber(discountRequested, 0)` is redundant against the declared type but matches the file's defensive convention for string-typed numerics off a DB read. No finding.
- **No schema/migration/entity/validator/route change**, no i18n surface, no `any`, no raw `fetch`, no tenant-scoping change, no generated-file edit, no `BACKWARD_COMPATIBILITY.md` frozen surface removed. The comment block added at `calculations.ts:89-96` is longer than house style prefers but follows the file's existing convention of commenting only non-obvious clamping decisions (cf. the `#2457` comment below it) — no finding.
- **Duplicate/already-merged check** — the diff is not present on `develop`; the last 20 commits touching these files carry no equivalent change.

## Labels

- `review` → `changes-requested`.
- `risk-medium` → `risk-high`: per the repo's inference rule, money and shared contract surfaces are `risk-high`, and Blocker 2 puts a public request field's meaning in the diff. `priority-medium` is left as-is — the defect is not reachable from the admin UI, so its urgency rating still looks right.
- `skip-qa` → `needs-qa`: the automated-verification exemption requires *"the database structure and API surface unchanged"*. Blocker 2 changes the observable meaning of an API request field, so the exemption does not hold on the current diff. If the fix lands with § 3's basis flag preserving caller input semantics, `skip-qa` becomes valid again and can go back.

## No autofix

Autofix did not run. The remaining work is the spec's §§ 2–4, and the spec itself routes two of those through the maintainers: **Decision 1** (column meaning) and **Decision 2** (percentage-first precedence, including whether `discountAmount: 0` alongside a non-zero percent should suppress the discount or apply it — a change the spec flags as *inverting* rather than dropping existing integration behaviour, with a double-discount failure mode). Picking those unilaterally on someone else's PR is exactly the class of decision this repo's `AGENTS.md` says to ask about first, so the PR goes back to @wojciechszyjka with the findings instead.

pkarw:
Label set to `changes-requested` because review found actionable issues (2 blockers, 1 major).

Risk raised to `risk-high` because the diff changes the observable meaning of a public API request field (`discountAmount`) on order/quote money math, which the repo's inference rule rates high. `priority-medium` kept — the defect is not reachable from the admin UI, so its urgency rating still fits.

`skip-qa` swapped for `needs-qa` because the automated-verification exemption requires the API surface to be unchanged, and it is not on this diff. It can go back to `skip-qa` if the fix lands with the spec's `discountAmountBasis` preserving caller input semantics.

Removed `ci-monitoring` — all 14 reported checks on `0a271d08` are green, nothing left to watch.

pkarw:

Shortest path through it: the diagnosis and the tests are right, and fixing it inside the shared `buildBaseLineResult` is a better shape than the per-mapper tagging the spec assumed — that part should survive. What needs a decision first is that `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` (merged in #5200, tracking the closed twin issue #5019) already specifies this exact change and is still marked `draft — design decision requested`, with two maintainer decisions outstanding. This PR implements its § 1 and stops, and the two things that follow from that are the blockers: caller-supplied `discountAmount` on the create path (`commands/documents.ts:3036`) silently changes meaning for existing API callers, and without the spec's § 2 percentage-first precedence the reproduction row in #3757 (`discount_amount = 255.0000` on the 3 × 85.00 seeded line) still recomputes to net `0` — so the issue's own expected result does not clear on any database where the bug already ran.

I have not picked those two decisions for you: the spec routes them through the maintainers, and Decision 2 in particular inverts what `discountAmount: 0` alongside a non-zero percent means for existing integrations. The minor and the nit are independent of that and can be taken straight away.

pkarw:

Review of `0a271d08`, validation source: GitHub checks (all 14 reported checks green on the head SHA, none pending or failing — no local gate re-run needed). Findings: 2 blockers, 1 major, 1 minor, 1 nit. No autofix iterations ran — the remaining work is §§ 2–4 of an existing spec whose two headline decisions are explicitly reserved for maintainer sign-off, so the PR is handed back to @wojciechszyjka rather than patched unilaterally.

wojciechszyjka:
Thanks @pkarw — I verified every substantive claim in this review against the code before responding, and the two blockers and the major are **correct**. I have corrected the PR description rather than argue any of them. Point by point:

**Blocker 1 — the spec.** Confirmed: `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` exists on `develop`, covers exactly `buildBaseLineResult`, is still `draft — design decision requested`, and states twice that no implementation lands until §§ 1–2 are approved. I did not check `.ai/specs/` before implementing, which root `AGENTS.md` explicitly requires. That is my miss, not a scope judgement. The PR now links the spec, #5019 and #5200 in its first two lines and carries a do-not-merge banner.

**Blocker 2 — the caller-input break.** Confirmed and the most important finding here. I traced the persistence direction (`:3124` writes the engine's own `discountTotal`, the mappers read it back) and concluded "no caller passes a per-unit value", but I never traced the **create** direction. `createLineSnapshotFromInput` at `documents.ts:3036` forwards the validated request body's `discountAmount` into the engine untouched, so under the old engine the API input was de facto per-unit and this PR silently flips it. The 60× worked example is right. "Breaking Changes: None" was a false claim and has been replaced with that table and a note that `BACKWARD_COMPATIBILITY.md`'s deprecation protocol was not followed.

**Major 3 — the repro does not clear.** Confirmed. I had traced the `12.75 → 38.25 → 114.75 → 255` compounding and reasoned that a stored `255` clamping to net `0` was "legitimate given the input" — which is true of the arithmetic and beside the point for the issue, since `SO-DEMO-2001` still shows `0,00` net. Linkage changed from `Closes #3757` to `Refs #3757`, with the arithmetic spelled out in the body.

**One correction back, on Minor 4 and Nit 5.** Both are real observations, but I do not think either can be taken "straight away" — they are contingent on the same decision as the blockers, in the opposite direction.

`SalesOrderDraftLines.tsx:68` (`normalizeNumber(payload.discountAmount, 0) * quantity`) reads a **caller payload**, forwarded verbatim to the create API at `:55`. Under the spec's § 3, caller-supplied `discountAmount` keeps its per-unit meaning by default (`discountAmountBasis` omitted ⇒ `'unit'`), so multiplying by quantity to display a line total in the draft grid is **correct as written**, and deleting the `* quantity` would introduce the very skew the finding warns about. It only becomes wrong under the alternative where caller input flips to line-total — i.e. the shape this PR currently has and § 3 is designed to avoid.

The same applies to the two test comments: *"discountAmount 5 was supplied per unit, so 15.00 is persisted for a 3 × 10.00 line"* stays accurate under § 3 and only goes stale if caller input flips. I have recorded both in the PR body as decision-dependent rather than editing them, since touching them now would bake in an outcome the maintainers have not chosen.

**On the labels.** No argument on `changes-requested`, `risk-high` or `needs-qa` — the reasoning for each is sound, and `needs-qa` in particular follows directly from Blocker 2 breaking the automated-verification exemption's "API surface unchanged" condition.

**Next.** I am not picking Decisions 1 and 2 unilaterally either, for the same reason you did not: the spec routes them through the maintainers, and Decision 2's documented cost falls on a caller shape that has no local signal anything went wrong. I have taken this back to the requester for a decision. Once it lands, the remaining work is § 2, § 3 (including de-duplicating the mapper shared with `returns.ts:137`), § 4, an `UPGRADE_NOTES.md` entry, and a call on repairing already-corrupted rows — at which point the six tests here need reworking onto the mapper-sourced shape.

wojciechszyjka:
## 🅿️ Parked — this PR will not be merged in its current shape

Maintainer decision, recorded here so the reasoning is not lost: this PR is **parked as evidence** rather than continued or closed. Both of the spec's gated decisions are deferred, and the remaining work will land as the spec's own implementation PR covering §§ 1–4 in one piece, rather than as further increments on top of this branch.

Converted to **draft** and labelled `blocked`, replacing `changes-requested` — the ball is not with the author awaiting a revision; it is with the spec's outstanding design decisions. `needs-qa`, `risk-high` and `priority-medium` are kept as @pkarw set them, since the reasoning behind each still holds for whatever eventually ships.

### Why parking rather than continuing

The review established that this change implements roughly the first quarter of `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` and that the missing three quarters are not decoration:

- Without § 3 (`discountAmountBasis` / `discountAmountFromStoredRow`) the engine fix silently flips the meaning of a public API request field.
- Without § 2 (percentage-first precedence) the symptom reported in #3757 does not clear on any database where the buggy engine already ran.
- Without § 4 the `?? 0` coalescing at `documents.ts:7177` and `:7671` still kills the percentage path on the upsert route, which is a *separate* defect from the one fixed here.
- The mapper duplicated between `documents.ts:2906` and `returns.ts:155` still means a fix applied to one file passes every acceptance criterion while the return flows stay broken.

Landing them incrementally would mean shipping at least one intermediate state that is observably worse for API callers than either the current behaviour or the finished contract. Doing all four together, against the spec, avoids that.

### What is worth salvaging from this branch

Branch `fix/issue-3757-line-discount-amount-per-line`, head `0a271d08`. The implementation PR should reuse rather than rediscover:

1. **The fix location.** Correcting this inside the shared `buildBaseLineResult` covers the order path, the quote path, the display recalc and the three persisting return flows at one point — a better shape than the per-mapper tagging the spec assumed, as the review itself noted.
2. **The six unit tests** in `packages/core/src/modules/sales/lib/__tests__/calculations.test.ts`. Three fail against the unfixed engine. They need reworking onto § 3's mapper-sourced shape, but the *cases* are the right ones — particularly the five-pass round-trip test, which is the regression for the compounding runaway (`12.75 → 38.25 → 114.75 → 255`) that a single-pass test cannot catch, and the document-level test that drives `buildBaseDocumentResult`'s `Math.max(subtotal + delta, 0)` return clamp where the symptom actually became visible.
3. **The corrected analysis in the description above**, including the worked 60× breaking-change table for the create path and the arithmetic showing why the reproduction row still recomputes to `0`.

### Tracking

#3757 stays **open** and is now the de facto tracker for implementing this spec — #5019 is closed and no other issue references the spec. No new follow-up issue was filed, to avoid a third tracker for one piece of work.

wojciechszyjka:
🤖 `om-pr-autopilot` — run at 2026-08-26T05:28Z (diagnosis only; no chain step executed)

**Diagnosis on entry**

### PR State Report — #5550 fix(sales): treat line discountAmount as a line total, not a per-unit rate (#3757)

- Author / fork: @wojciechszyjka · same-repo branch `fix/issue-3757-line-discount-amount-per-line` · draft · pushable: yes · drivable: yes
- Plan: none — the PR body carries no `Tracking plan:` / `Tracking run folder:` line and no new file under `.ai/runs/` appears in the diff, so implementation completeness was judged from the diff and the review record rather than from a plan file
- Diff scope: code (2 files, +183/−4) — `packages/core/src/modules/sales/lib/calculations.ts` and its test file; no `.tsx`, nothing under `packages/ui/`, no migration or entity change. The review nevertheless classified it as touching a **contract surface**, because it changes the observable meaning of the `discountAmount` request field on the sales order/quote create path
- Review: `CHANGES_REQUESTED` by @pkarw (2026-08-24, 2 blockers / 1 major / 1 minor / 1 nit) · 0 inline review threads, so there is nothing to resolve thread-by-thread — the verdict itself is the open item, and the author has already acknowledged every finding as correct
- CI: green on head `0a271d08` — 14 reported checks pass (`test`, `lint`, `typecheck` via `prepare`, `docker-build`, `ephemeral-integration (none)`, `documents-multi-instance`, `ds-lint`, `audit-scope`, `scope`, CodeQL ×4, CLA); the remaining entries are `skipping`, not failures. Branch protection on `develop` is not readable from this account (HTTP 404), so every reported check was treated as required
- Mergeability: `MERGEABLE` with `mergeStateStatus: BLOCKED` — there is **no** merge conflict; the branch is 19 commits behind `develop` but merges cleanly. The blocked state comes from the draft flag, the changes-requested review and the unsatisfied QA gate, not from the base
- Labels: pipeline=`blocked` category=`bug` meta=`needs-qa` priority=`priority-medium` risk=`risk-high` · missing: none — exactly one pipeline, one priority and one risk label are present, and the set matches what the diff and the review warrant
- QA: required (`needs-qa`, applied by @pkarw when the automated-verification exemption stopped applying) with no evidence and no approval. Under this repository's hard QA gate that alone would block the merge
- Blockers: the `blocked` label plus the 🅿️ park decision recorded on 2026-08-24 — the two maintainer decisions in `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` (§ Proposed Solution 1, the column's meaning; § Proposed Solution 2, percentage-first precedence). The spec on `develop` still reads `Status: draft — design decision requested`, so the block is live, not stale

**Chain executed**

| Step | Skill | Outcome |
|---|---|---|
| — | none | Matrix row 0c fired on entry: a hard block (`blocked`) is present and the user did not ask to work through it, so the run stopped at diagnosis and dispatched no sub-skill. No claim was taken and nothing on the branch was touched — signalling `in-progress` on a deliberately parked PR would misrepresent it as under active automation |

**State now**

- 🔍 Review: changes requested by @pkarw and not superseded; the author has confirmed both blockers and the major are correct and corrected the PR description rather than contesting them. No new head has been pushed since `0a271d08`
- 🧪 CI: green on the current head, and it has been green since the review. CI is not what is holding this PR
- 🚀 Mergeability: no conflict, but merging is blocked four ways — draft status, the changes-requested review, `needs-qa` without `qa-approved`, and the `blocked` label itself
- 📸 QA: not run and not warranted yet; there is no user-facing surface in this diff to exercise, and QA on a change that will be rewritten against the spec would be wasted effort. `needs-qa` correctly stays until the shape that actually ships is known
- 📋 Follow-ups filed: none by this run. #3757 remains the tracker for the spec's implementation, exactly as the park comment states; filing a third tracker for the same work would only fragment it
- 🔁 Incidental observation: on 2026-08-25T03:19Z this PR was marked ready for review and converted back to draft sixteen seconds later by @pkarw. The net state is unchanged and I have treated it as an accidental flip rather than as a signal

**Labels this PR should carry**

- ⛔ **`blocked`** — the pipeline label stays as the park decision set it: the ball is with the spec's two outstanding design decisions, not with the author awaiting a revision, which is what `changes-requested` would mean.
- 🐛 **`bug`** — the change addresses the defect reported in #3757; it adds no capability.
- 🧪 **`needs-qa`** — kept as @pkarw set it. The automated-verification exemption requires the API surface to be unchanged, and on this diff it is not.
- ⚠️ **`priority-medium`** — the defect is not reachable from the admin UI, so its urgency rating is unchanged.
- 💥 **`risk-high`** — the diff changes the observable meaning of a public API request field on order and quote money math, which this repository's inference rule rates high.

No label change was needed or applied; the current set is already correct.

**Verdict:** ⛔ blocked — and correctly so. This PR needs a human decision, not another automated pass. Nothing an autopilot chain can do (continue, autofix, review, QA, merge) moves it, because the missing input is maintainer sign-off on the spec's §§ 1–2. The three ways forward are: sign off on the two decisions and let the spec's implementation PR land §§ 1–4 in one piece (this branch's fix location and its six test cases are the salvage listed in the park comment); park it further as evidence, which is the current state and needs no action; or close it and let the implementation PR carry the analysis. My recommendation is the first — the spec has been in `draft — design decision requested` since 2026-08-07 and #3757 stays open until it ships.

wojciechszyjka:
🅿️ → 🚀 **Superseded by #5640, closing.**

The two decisions this PR was parked on — § Proposed Solution 1 (the column is a line total) and § 2 (percentage-first precedence) — were approved by @wojciechszyjka on 2026-08-26 and are now recorded in the spec's `## Decision Record`, along with the three dependent calls the spec left open. #5640 implements §§ 1–4 in one piece, exactly as the parking comment said the work would land.

Everything the parking comment listed as worth salvaging was carried over rather than rediscovered:

1. **The fix location** — the arithmetic still lives in the shared `buildBaseLineResult`, which @pkarw's review here correctly judged a better shape than the per-mapper tagging the spec originally assumed.
2. **The six unit test cases**, reworked onto § 3's mapper-sourced shape and joined by nine more. The five-pass round trip and the document-level clamp test both survive, and the whole suite was negative-controlled: 10 of 15 fail against the unfixed engine.
3. **The corrected analysis in this PR's description**, including the 60× breaking-change table and the arithmetic showing why the reproduction row still recomputes to `0` — which is why #5640 also says `Refs #3757` rather than `Closes`.

What #5640 adds beyond this branch: § 2's percentage-first precedence, § 3's `discountAmountBasis` / `discountAmountFromStoredRow` pair, § 4's per-operand decomposition at **both** upsert sites, the de-duplication of the mapper that `commands/returns.ts` held its own copy of, and the `UPGRADE_NOTES.md` entry this PR was faulted for lacking.

For the record, @pkarw's review here was right on every substantive point, and the review's second blocker in particular — the silent 60× change on the create path — is what the shipped `discountAmountBasis` default exists to prevent.

Closing rather than leaving parked: the analysis lives on in this PR's description and in the spec, and #3757 remains the tracker.

Issue #104, opened 2025-11-04
feat: typescript based API Client


pkarw:
Let's develop an API client that supports the http://demo.openmercato.com/api/docs/markdown (for easier usage with LLM :))

pkarw:
add the API client as a package `client` in the mono-repo, type all the endpoints based on openapi schemas

PR #5553, opened 2026-08-24
fix(currencies): fetch exchange rates for every currency, not just active ones

Frshy:
## Summary

`RateFetchingService.getExistingCurrencies()` filtered the currency set by `isActive: true`, and that set is what every provider response is filtered against — a rate survives only when **both** legs of its pair are present. Deactivating a currency therefore stopped writing any exchange rate for it, ever.

The damage is not stale rates. Records still denominated in that currency become **unconvertible** once the stored rates fall outside the consumer's lookback window — including the migration to an active currency, which is itself a conversion. `customers/api/deals/{aggregate,summary}` converts deal values with `maxDaysBack: 60`, so roughly 60 days after a deactivation those deals silently drop out of pipeline totals (`convertedAll: false`), with no error surfaced.

`isActive` was never the right gate. Everywhere else in the module it means *selectability*: the options endpoint pins `isActive = true`, the list exposes `?isActive=true|false`, and the exchange-rate form's currency select filters on it. Whether a currency still needs a live rate is a different question, and deactivating one while existing records reference it is the normal way to grandfather it.

Dropping the clause leaves `deletedAt: null` as the single gate, so soft delete becomes the one way to take a currency out of rate fetching.

This also fixes a non-obvious side effect: both bundled providers return `[]` outright when PLN is missing from the set, so a deactivated PLN used to silence the **whole provider** rather than just its own pairs.

**Cost:** no extra provider calls — NBP and Raiffeisen each fetch a full table in one request per date, and the set is applied locally. Only the stored row count grows, bounded by what the provider returns.

**Behaviour change on upgrade:** currencies that were previously deactivated start receiving rates again on the next fetch.

**Trade-off, stated plainly:** `isActive` was also, in practice, the only *non-destructive* way to stop a currency accruing rates. After this change there is none — soft delete is the terminal off-switch, and reaching it means clearing the currency's rates first, because `deleteCurrencyCommand` refuses a currency that still has live ones. `services/README.md` documents that order and why the rejected `Currency.fetchRatesWhenInactive` opt-in column is not coming back.

## Changes

- `services/rateFetchingService.ts` — drop the `isActive: true` clause from `getExistingCurrencies()`; add a doc comment recording *why* the flag must not gate rate fetching, so it is not reintroduced as an obvious-looking filter.
- `services/__tests__/rateFetchingService.basic.test.ts` — invert the unit test that pinned the old behaviour rather than quietly relaxing it, tighten the query-shape assertion to match the filter exactly, assert the currency set by value rather than by size, and add a case that drives the **real** `NBPProvider` with an inactive PLN to prove the provider still issues its request.
- `services/providers/__tests__/baseCurrencyGate.test.ts` *(new)* — pins the other half of that regression directly on both bundled providers: a PLN-less set makes `NBPProvider` and `RaiffeisenPolandProvider` return `[]` without issuing any request.
- `__integration__/TC-CUR-015-inactive-currency-rates.spec.ts` *(new)* — end-to-end coverage on the money path: deactivate a currency, `POST /api/currencies/fetch-rates`, assert rate rows land for both legs of its pair. Runs against `example_fixed_rates`, the offline provider the example module registers, so it needs no network and is deterministic; gated on `dependsOnModules: ['currencies', 'example']`.
- `services/README.md` — new "Which Currencies Get Rates" section documenting the selectability-vs-convertibility distinction, the delete-guard ordering that soft delete requires, and the deliberate absence of a reversible per-currency pause.
- `AGENTS.md` (currencies module) — add a MUST NOT rule against filtering rate fetching by `isActive`.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**
No spec exists for the currencies module. This is a one-clause behavioural bug fix in an existing service; the reasoning is captured in `services/README.md` and the module `AGENTS.md` instead.

## Testing

Unit — `packages/core` jest, whole currencies module:

```
Test Suites: 23 passed, 23 total
Tests:       153 passed, 153 total
```

Mutation check: restoring `isActive: true` in `getExistingCurrencies()` on this head fails **4 of those 153**, so the behaviour is pinned rather than merely described.

Integration — `yarn test:integration:ephemeral --filter TC-CUR-015`, run in an isolated ephemeral environment:

```
Running 1 test using 1 worker
  ✓  1 packages/core/src/modules/currencies/__integration__/TC-CUR-015-inactive-currency-rates.spec.ts:70:3
       › TC-CUR-015: rate fetching ignores the isActive flag
       › a deactivated currency still gets rates stored for both legs of its pair (535ms)

  1 passed (7.0m)
```

`yarn agents:check-budget` exits 0. `yarn lint` on the changed and added files is clean.

## Checklist

- [ ] QA routing set: `skip-qa` for low-risk non-customer-facing changes — otherwise `needs-qa`.

**Notes on the unticked boxes**

- **CLA** — deliberately left for a human to tick; an agent opened this PR and will not accept a legal agreement on the author's behalf.
- **Spec** — N/A, as described in the Specification section above.
- **Priority / Risk / QA routing** — the authoring account has `read` permission on this repository and cannot apply labels. A maintainer has since applied `bug`, `priority-high`, `risk-high` and `needs-qa`; the boxes stay unticked only because the author cannot verify them by writing them.

**On QA routing after this round:** the PR now ships the integration coverage the exemption asks for, but it keeps `needs-qa` regardless — the change has a manually exercisable surface (the exchange-rates list shows pairs for currencies that were previously excluded, and the "Fetch rates" count changes), which is the reviewer's own second reason for the label. While `needs-qa` is present this must not merge without `qa-approved`.

### Design System Compliance

Not applicable — this change touches no UI-rendering file: no `.tsx` (outside tests), nothing under `packages/ui/src/`, nothing under `**/components/**`. It is a service-layer query change plus tests and documentation.

## Linked issues

No upstream issue. This was found and reviewed internally at Full Stack House; the final approach — dropping the `isActive` clause outright — was suggested by @KubaBir in review, replacing an earlier `Currency.fetchRatesWhenInactive` opt-in column.

Note: the first commit message ends with a bare `#104` reference. That is the **internal** Full Stack House review PR, not an Open Mercato issue or PR — please disregard any cross-reference GitHub draws from it.


Frshy:
### Metadata intent (author lacks `triage` on this repo)

This PR was opened from a fork by an account with `read` permission, so it cannot set its own labels, assignee, or reviewer. Requesting a maintainer apply the following. Each name was validated against this repo's label list, and the rationale follows `.ai/docs/pr-workflow.md`.

| Label | Type | Rationale |
|---|---|---|
| `review` | pipeline | Ready non-draft PR, not in any other pipeline state. |
| `bug` | category | Fixes incorrect behaviour in an existing service, no new surface. |
| `priority-high` | priority | The inference rule routes **money** fixes to `priority-high`. Exchange-rate conversion is money, and the failure is silent: deal values drop out of `customers/api/deals/{aggregate,summary}` totals with `convertedAll: false` and no surfaced error. Not `priority-extreme` — there is no active outage or data loss, the rows survive and become convertible again once rates are refetched. |
| `risk-high` | risk | Signals conflict: "single-module bug fix shipping with tests" reads `risk-medium`, but the inference rule sends **money** changes to `risk-high`, and the rule says take the higher. Stated plainly so it can be argued down: the code delta is one clause, but it changes behaviour on upgrade (previously-deactivated currencies resume receiving rates) and grows stored rate rows. |
| `needs-qa` | meta | **Not** claiming the automated-verification exemption. The exemption requires an integration test for changes touching money; this PR ships unit tests only, so by the rule as written it keeps `needs-qa`. Happy to be overridden to `skip-qa` if maintainers read the money clause as not reaching rate fetching — flagging rather than deciding it unilaterally. |

**Assignee:** intended `@Frshy` (the author) — cannot self-assign without `triage`.

**Reviewer:** by role, since an outside account should not pick names — a maintainer who owns the **currencies / core module** and can rule on the `needs-qa` vs `skip-qa` question above.

No claim label (`in-progress`) is requested: this account could not release one later.

**QA gate:** while this PR carries `needs-qa` it must not merge without `qa-approved`. An evidence comment alone does not clear that gate — the labels are a maintainer write. If you would rather this go the self-QA route, say so and I will run it locally, click through, and attach evidence for a maintainer to convert into `qa-approved` + `qa-self-verified`.

CI note: cross-repository PRs cannot receive package previews — both preview workflows are restricted to same-repository branches, per `CONTRIBUTING.md`.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-25T00:35:45Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(currencies): fetch exchange rates for every currency, not just active ones

## 🎯 Summary

This PR removes the `isActive: true` clause from `RateFetchingService.getExistingCurrencies()` (`packages/core/src/modules/currencies/services/rateFetchingService.ts:144`), so the currency set that every provider response is filtered against becomes "every currency in scope that is not soft-deleted". It ships a doc comment on the method, a new `services/README.md` section, a MUST NOT rule in the module `AGENTS.md`, and an inverted unit test plus a tightened query-shape assertion.

**The diagnosis is right and I was able to verify every load-bearing claim in the PR body against the code.** Both bundled providers really do abandon the entire fetch when `PLN` is absent from the set — `nbp.ts:45-48` and `raiffeisen.ts:53-56` each `return []` before making any request — so a deactivated `PLN` silenced the whole provider rather than just its own pairs, exactly as described. Both really do fetch one full table per date (`exchangerates/tables/c/{date}` and `?type=kursywalut&range=all`) and apply the set locally, so the "no extra provider calls" cost argument holds; only the stored row count grows. And `customers/api/deals/aggregate/route.ts:368` and `summary/route.ts:219` really do convert with `maxDaysBack: 60`, which is where the silent `convertedAll: false` drop-out the PR describes comes from.

I also checked for a second gate the one-clause fix might have missed, and there isn't one: every path that writes rates — `ExchangeRateService.findRateWithFallback`'s autoFetch, `POST /api/currencies/fetch-rates`, and `mercato currencies fetch-rates` — funnels through `fetchRatesForDate`, so fixing the set at its single source covers all of them. There is no scheduled worker in the module today.

The test inversion is genuine rather than cosmetic. I mutation-tested it by restoring the `isActive: true` clause on the PR head and re-running the module suite: **3 tests fail (145/148 passing)**, and the exact-object assertion at `rateFetchingService.basic.test.ts:309-315` is what makes a re-added `isActive` clause fail loudly instead of passing under a partial `objectContaining` match. That is the right instinct and it is the part of this PR I would most want kept.

What holds the change back is the operational half. `isActive` was not only a selectability flag — in practice it was also the only non-destructive way to stop a currency from accruing rates, and the replacement this PR documents (soft delete) is guarded by a check that the newly-accruing rates trip. Details in the Major finding below; the fix is a documentation/UX correction rather than a redesign.

Reviewed scope: all four changed files, plus `commands/currencies.ts`, `services/exchangeRateService.ts`, `services/providers/{nbp,raiffeisen,registry}.ts`, `api/fetch-rates/route.ts`, `cli.ts`, and the `customers` deal-conversion consumers.

## Verdict

❌ **request changes** — one major finding (the PR removes the only non-destructive off-switch for rate accrual and the soft-delete replacement it documents is blocked by `deleteCurrencyCommand`'s active-rates guard) plus three minor findings. Per this repository's review rule, Medium/minor findings also request changes. None of this disputes the fix itself, which is correct and well-argued; the work left is in the documentation the PR ships and in one test that does not assert what its name claims.

## 🧪 Validation Gate

**Validation source: local fallback** (`packages/core` jest, `packages/core` tsc, `yarn agents:check-budget`). GitHub checks were preferred first but carry no usable data for this head: both `CI for Develop&Main` and `Mutation tests` sit at `action_required` on `0e174ec`, because a fork PR needs a maintainer to approve workflow runs on every push. The only checks that reported are `license/cla` (pass) and `[code]smith` (skipping), so there is currently **no CI test signal at all** for this commit.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ⏭️ NOT RUN | Not required to exercise the changed behaviour; the review worktree has no built package `dist/` output. |
| `yarn generate` | ⏭️ NOT RUN | No auto-discovered module file was added or moved, so there is nothing to regenerate. |
| `yarn build:packages` (2nd) | ⏭️ NOT RUN | See above. |
| `yarn i18n:check-sync` | ⏭️ N/A | No locale file and no user-facing string is touched by this diff. |
| `yarn i18n:check-usage` | ⏭️ N/A | Same. |
| `yarn typecheck` | ⚠️ INCONCLUSIVE | `tsc --noEmit` in `packages/core` reports 922 errors, but **zero of them are in any file this PR changes**. All 922 cascade from one pre-existing condition unrelated to this branch: `@tanstack/react-table/legacy` has no matching subpath in the installed `@tanstack/react-table` package exports, producing 360 × TS2307 and 502 × TS7031 "implicitly any" follow-ons in untouched `.tsx` backend pages (the 15 currencies-module hits are all in `backend/currencies/page.tsx` and `backend/exchange-rates/page.tsx`, which this PR does not touch). A trustworthy typecheck needs the full `build:packages`+`generate` prefix, which I did not run. Worth noting that the production delta here — removing one property from a `FilterQuery<Currency>` object literal — cannot introduce a type error. |
| `yarn test` | ✅ PASS (scoped) | `packages/core` jest scoped to `src/modules/currencies`: **22 suites, 148 tests, all passing**, including the 14 in the changed file. Plus the mutation check described in the Summary: restoring `isActive: true` fails 3 of those 148. |
| `yarn build:app` | ⏭️ NOT RUN | No app-layer file is touched. |
| `yarn agents:check-budget` | ✅ PASS | Exit 0 on the PR head — the 3-line addition to the currencies `AGENTS.md` does not push any instruction chain past its ratchet. Worth confirming explicitly, because a red budget step in this repo's CI skips the typecheck and test steps behind it. |
| `yarn lint` | ⏭️ NOT RUN | Not runnable in this worktree: `eslint.config.mjs` resolves `eslint-config-next/core-web-vitals`, which is not installed at the workspace root. An environment limitation, not a property of this branch. |

## Findings

### ⚠️ Major

**`packages/core/src/modules/currencies/services/README.md:299` — the documented off-switch ("Soft delete is the way to stop fetching a currency") is blocked by the very rates this change now creates.**

`deleteCurrencyCommand` refuses the soft delete while the currency has any live exchange rate (`packages/core/src/modules/currencies/commands/currencies.ts:373-389`):

```
Cannot delete currency EUR because it has N active exchange rate(s). Please delete or deactivate the exchange rates first.
```

and `storeRates` writes every fetched rate with `isActive: true` and `deletedAt` unset (`rateFetchingService.ts:208`), matching that guard's `{ deletedAt: null, isActive: true }` count exactly.

Before this PR the operator had a terminal sequence: **deactivate the currency** (rate writes stop) → clear its remaining rates → delete it. Dropping the `isActive` clause deletes step one. The only sequence left is "clear every rate for the currency, then delete it before anyone runs a fetch" — and any fetch in that window re-creates the rates and re-blocks the delete. The fetch triggers are the `POST /api/currencies/fetch-rates` button, `mercato currencies fetch-rates`, and `ExchangeRateService`'s `autoFetch` (which defaults to `true`; today's two in-repo consumers pass `autoFetch: false`, but the default is what a new caller gets).

Concrete failure: an operator deactivates `EUR`, notices rates still arriving, reads the new README section, and tries to delete `EUR` → HTTP 400. They hand-delete the `EUR` rate rows, a colleague clicks "Fetch rates", and the delete returns 400 again.

The README states `deletedAt` "excludes it from the set, and nothing else does", which is true of the set but leaves out that reaching `deletedAt` first requires clearing every rate for that currency. Because the docs are part of this diff, they are the contract being reviewed here.

**Fix (small — documentation and wording, not a redesign):** in the new "Which Currencies Get Rates" section, state the delete guard and the required order explicitly — clear or deactivate the currency's exchange rates, then soft-delete the currency, and do it without an interleaved fetch. It is worth saying plainly that after this change there is no reversible way to pause rate accrual for a single currency, since that is the trade-off the rejected `Currency.fetchRatesWhenInactive` column would have covered and a future reader will ask why it was dropped. Pointing `deleteCurrencyCommand`'s error string at that README section would also help, though that is optional.

### 🔹 Minor

**1. `packages/core/src/modules/currencies/services/__tests__/rateFetchingService.basic.test.ts:237-266` — the "providers gated on it still run" test never exercises provider gating, and its headline assertion pins only cardinality.**

The test is named *"keeps an inactive base currency in the set so providers gated on it still run"*, but `createMockProvider` has no `PLN` gate, so the `if (!availableCurrencies.has('PLN')) return []` branch in `nbp.ts:46` / `raiffeisen.ts:54` — the actual regression being guarded — is never executed. And the set assertion at line 262, `expect.objectContaining({ size: 2 })`, checks only the set's size: I verified directly that `expect(new Set(['USD','EUR'])).toEqual(expect.objectContaining({ size: 2 }))` **passes**, so a set that dropped `PLN` entirely would satisfy the assertion the test is named for.

The test is not vacuous overall — `expect(result.totalFetched).toBe(1)` on a `PLN→EUR` rate does pin `PLN` membership, because `fetchFromProvider` keeps a rate only when both legs are in the set. But the weight is carried by the incidental assertion rather than the named one, which is at odds with this PR's own stated goal of tightening assertions so the behaviour cannot silently regress.

**Fix:** use the exact form already used three lines above at line 233 and in the pre-existing test at line 52 — `new Set(['PLN', 'EUR'])` — and add a case that drives the real `NBPProvider.fetchRates` (or a mock that replicates the gate) with a `PLN`-less set and asserts it returns `[]`, so the whole-provider silencing described in the PR body is actually pinned by a test.

**2. No integration coverage for a money path, in a module that already has the harness.**

`packages/core/src/modules/currencies/__integration__/` already holds 19 specs (`TC-CUR-001` … `TC-CUR-CRUDFORM-001`), so the marginal cost of one more is low. To be fair to the author: the PR's reading of `.ai/docs/pr-workflow.md` is **correct** — the integration-test clause there is a condition of the `skip-qa` automated-verification exemption, not a free-standing requirement, and by keeping `needs-qa` this PR complies with repo policy. The finding stands on the separate review-checklist rule that risk-heavy paths (money, among others) get integration coverage.

**Fix:** add a spec that seeds an inactive-but-not-deleted currency, runs a fetch, and asserts a rate row lands for it — and, ideally, that a subsequent conversion resolves through it. That converts the QA request into executable proof that runs on every push, which matters more than usual here given CI cannot even start on a fork head without manual approval.

**3. `packages/core/src/modules/currencies/services/README.md:297` — the lookback window named in the docs is not the one that motivates the fix.**

The new section says conversions fail "once the stored rates fall outside the consumer's lookback window (`ExchangeRateService` defaults to `maxDaysBack: 30`)". The default is correct, but the consumer this PR's own body uses to justify the severity — `customers/api/deals/aggregate/route.ts:368` and `summary/route.ts:219` — overrides it to `60`. A reader reaching for the concrete exposure gets the wrong number.

**Fix:** name both — the 30-day default and the 60-day window the deal-conversion endpoints actually use, since the latter is the real "roughly 60 days after a deactivation" claim from the PR description.

### 💅 Nit

**`fix/currency-rates-skipped-for-inactive-currencies` is 47 commits behind `develop`** (merge-base `005201cd7`). The head is `MERGEABLE`, so this is not a conflict — but in this repository a branch behind its base pulls the full integration shard set rather than the affected subset, which tends to surface unrelated red. Merging `develop` before a maintainer approves the workflow runs would give this PR a cleaner first CI result and avoid spending a scarce manual approval on noise.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `getExistingCurrencies` is `private`; nothing exported changed.
- [x] No function signature changed in a breaking way — `fetchRatesForDate(date, scope, options)` is untouched.
- [x] No required type field removed or narrowed — no type declaration is touched.
- [x] No HTTP route URL removed or renamed — `api/fetch-rates` is unchanged.
- [x] No field removed or retyped in an existing response shape — `FetchResult` is unchanged; only `totalFetched` values grow.
- [x] No event or message name renamed or removed — `events.ts` and `message-objects.ts` are untouched.
- [x] No CLI command or flag renamed or removed — `mercato currencies fetch-rates` is unchanged.
- [x] No database table or column renamed or removed — no entity and no migration in the diff; `exchange_rates` gains rows, not columns.
- [x] No config key renamed and no default changed silently — no config surface is touched.
- [x] Where a contract had to change: none of the 13 protected surfaces in `BACKWARD_COMPATIBILITY.md` is affected. This is a runtime behaviour change inside a private method, and the PR discloses it under "Behaviour change on upgrade" — previously deactivated currencies resume receiving rates on the next fetch. That disclosure is the right call and it belongs in the release notes.

## 🧪 Test Coverage

**What is covered.** The changed suite pins four distinct things: that a rate survives only when both legs are in the set (`:175`), that an inactive currency now *does* receive rates and is passed to the provider (`:206-235`), that a soft-deleted currency is still excluded (`:268`), and — the important one — that `em.find` is called with exactly `{ tenantId, organizationId, deletedAt: null }` and nothing more (`:309-315`). That last assertion is what makes a re-added `isActive` clause a test failure rather than a silent pass, and I confirmed it works: restoring the clause on the PR head fails 3 of the 148 module tests.

**Gaps.**

1. The whole-provider silencing regression — a `PLN`-less set making `NBPProvider`/`RaiffeisenPolandProvider` return `[]` before issuing a request — is described in the PR body, the README, and a test name, but is not exercised by any test (Minor 1). It belongs in `rateFetchingService.basic.test.ts` alongside the new case, or in the providers' own `services/providers/__tests__/`.
2. There is no test for the operational consequence in the Major finding: that a currency which has accrued rates cannot be soft-deleted. A case in `commands/__tests__` (or an `__integration__` spec) asserting the 400 from `deleteCurrencyCommand` after a fetch would pin the interaction the new documentation needs to describe.
3. No integration coverage on a money path, despite an existing 19-spec harness in this module (Minor 2).

**On QA routing:** `needs-qa` is the right label and the author's reasoning for it is sound. There *is* a manually exercisable surface here — the exchange-rates list will show pairs for currencies that were previously excluded, and the "Fetch rates" action's reported count changes — so a QA reviewer has something concrete to verify: deactivate a currency, fetch, confirm rates now appear for it, and confirm a conversion involving it resolves.

pkarw:

- 🚀 `merge-queue` — the re-review found no blockers, no majors and no minors, so the code-review gate passes; `needs-qa` stays on, which means the QA-approval gate holds this PR out of the merge until a QA reviewer adds `qa-approved`. This replaces `changes-requested`, whose five findings (one major, three minors, one nit) I verified as resolved against the code on `917ca52` rather than from the author's response alone.
- 🧪 `needs-qa` — retained deliberately. The automated-verification exemption does not apply even though the PR now ships integration coverage, because the change has a manually exercisable surface: the exchange-rates list now shows pairs for currencies that were previously excluded, and the "Fetch rates" result count changes. The author reached the same conclusion independently and asked for the label to stay.
- 🐛 `bug` — the change fixes incorrect behaviour in an existing service without adding any new surface.
- 🔺 `priority-high` — retained. The defect silently drops deal values out of pipeline totals roughly 60 days after a currency is deactivated (`convertedAll: false`, no error surfaced), which is a money path failing quietly rather than loudly.
- ⚠️ `risk-high` — retained. The blast radius is narrower than it looks, since `RateFetchingService` has no consumers outside the currencies module, but the change alters what gets written to `exchange_rates` on a money path, and `risk-high` correctly reinforces the case for manual QA here.
- 🤝 `partner-request` — applied earlier by the author; left untouched.

pkarw:

Two notes that are not findings but will save you time. First, CI has produced no signal on this head at all: both `CI for Develop&Main` and `Mutation tests` sit at `action_required` on `0e174ec`, because a fork PR needs a maintainer to approve workflow runs on every push. That is why the review's validation gate is a scoped local fallback rather than a read of the checks, and it is also why the 47-commits-behind nit is worth acting on before someone spends a manual approval on it. Second, your reading of the `needs-qa` vs `skip-qa` question in the metadata comment is right — the integration-test clause is a condition of the exemption, not a free-standing requirement — so no maintainer ruling is needed there; I have applied `needs-qa` along with `bug`, `priority-high` and `risk-high` as you proposed.

pkarw:

The review found 0 blockers, 1 major, 3 minors and 1 nit. The major is that this PR's documented replacement for the removed off-switch — soft-deleting a currency to stop fetching it — is blocked by `deleteCurrencyCommand`'s active-rates guard, which the rates this change now creates will trip; the fix is a documentation correction rather than a redesign. The diagnosis and the one-clause fix itself check out: I verified the NBP/Raiffeisen PLN gating, the single-choke-point coverage of every fetch path, and the test inversion by mutation-testing the head (restoring `isActive: true` fails 3 of 148 module tests).

autofix: skipped (not my PR — re-run with --autofix to fix it here). No CI follow-up is scheduled, because both workflow runs on this head are at `action_required` awaiting a maintainer's approval of fork CI and would never settle inside the wait budget; `ci-monitoring` is deliberately not applied, so this PR is free for anyone to act on.

Frshy:
Thanks @pkarw — all five items are addressed. Taking them in order.

### ⚠️ Major — the documented off-switch is blocked by the delete guard

You're right, and it's the one place where this PR's docs were making a promise the code doesn't keep. I confirmed the guard myself before rewriting: `deleteCurrencyCommand` counts `{ deletedAt: null, isActive: true }` rates on either leg and 400s, and `storeRates` writes every fetched rate with exactly that shape — so the rates this change now creates are precisely the ones that block the delete.

`services/README.md` now states the guard and the required order explicitly:

1. delete or deactivate every `exchange_rates` row on either leg of the currency, otherwise `DELETE /api/currencies/currencies` answers `400 Cannot delete currency XXX because it has N active exchange rate(s)`;
2. soft-delete the currency **with no fetch in between** — the "Fetch rates" button, `mercato currencies fetch-rates`, and `ExchangeRateService`'s `autoFetch` (default `true`) each recreate the rates and re-block the delete.

And, per your last paragraph, it now says outright that there is **no reversible way to pause rate accrual for a single currency**, names the rejected `Currency.fetchRatesWhenInactive` column as the thing that would have covered it, and says why it was dropped — so the next reader who asks "why isn't there an opt-in flag?" finds the answer instead of re-proposing it. I also lifted that trade-off into the PR description, since it belongs in the release notes as much as the "previously deactivated currencies resume receiving rates" line does.

I left `deleteCurrencyCommand`'s error string alone. It already says "Please delete or deactivate the exchange rates first", which is the correct instruction; pointing it at a README path felt like the kind of user-facing string change that wants its own justification.

### 🔹 Minor 1 — the provider-gating test didn't exercise provider gating

Both halves fixed, and thank you for actually running `expect(new Set(['USD','EUR'])).toEqual(expect.objectContaining({ size: 2 }))` — that assertion was worse than weak, it was the exact shape this PR complains about elsewhere.

- The set assertion is now `new Set(['PLN', 'EUR'])`, matching line 233 and the pre-existing test at line 52.
- New case in the same file, **`lets a provider gated on an inactive base currency issue its request`**, registers the real `NBPProvider` (not a mock), stubs `global.fetch` with an NBP table-C payload, and asserts the request was actually issued and both legs stored. With `isActive: true` restored in `getExistingCurrencies` this fails, because the provider bails before touching the network.
- New file **`services/providers/__tests__/baseCurrencyGate.test.ts`** pins the gate itself from the other direction, on both bundled providers: a PLN-less set → `[]`, and `global.fetch` never called.

Mutation check on the head: restoring `isActive: true` now fails **4 of 153** module tests, up from the 3 of 148 you measured.

### 🔹 Minor 2 — no integration coverage on a money path

Added: **`__integration__/TC-CUR-015-inactive-currency-rates.spec.ts`**. It ensures both legs of the pair exist (creating them rather than assuming the seed ran, per `.ai/qa/AGENTS.md`), deactivates the non-base one, posts to `/api/currencies/fetch-rates`, and asserts two rate rows land — one out of the deactivated currency and one into it.

Two design notes, since they're the non-obvious parts:

- It runs against **`example_fixed_rates`**, the offline provider `apps/mercato/src/modules/example` registers, not NBP/Raiffeisen. Driving a real provider would put a live HTTP call and NBP's weekend/holiday 404s in the CI path; the example provider emits exactly `USD→EUR` and `EUR→USD` and filters them against the same `availableCurrencies` set, so `totalFetched` is a clean **2 after the fix and 0 before it**. The spec declares `dependsOnModules: ['currencies', 'example']` so discovery skips it where the example module isn't enabled.
- The fetch date is randomised into a far-future window, because `storeRates` keys on `(from, to, date, source)` **without** a `deletedAt` filter — a fixed date would let a re-run update the previous run's already-deleted rows instead of creating visible ones. Teardown restores `isActive` and hard-deletes the rows it created.

### 🔹 Minor 3 — the lookback window in the docs isn't the motivating one

Fixed. The README now names both: the `maxDaysBack: 30` default *and* the `maxDaysBack: 60` that `customers/api/deals/{aggregate,summary}` actually pass, which is where the "roughly 60 days" claim comes from.

### 💅 Nit — 47 commits behind `develop`

Merged `upstream/develop` (clean, no conflicts; it was 60 behind by the time I got to it). Worth flagging for whoever approves the workflow runs: the merge pulled in a new `@open-mercato/channel-discord` workspace package, so a stale `node_modules` needs `yarn install` + `yarn build:packages` before `initialize` will boot — that bit me locally before the integration run would start.

### On the validation gate

Same environment caveat as your review, plus one improvement: `yarn typecheck` still can't be trusted here (the `@tanstack/react-table/legacy` subpath cascade you diagnosed), but `tsc --noEmit` in `packages/core` reports **zero** errors in any file this PR touches, including the two new ones. `yarn lint` on the changed files is clean, and `yarn agents:check-budget` exits 0.

The integration spec was **executed**, not merely written — `yarn test:integration:ephemeral --filter TC-CUR-015` against an isolated ephemeral environment, 1 passed. Full output is in the PR description. One honesty note on its strength: the `2 vs 0` claim in the spec's docstring follows from the provider's own `availableCurrencies` filter rather than from an executed mutation run — I started one, but a full rebuild plus app boot per iteration made it a poor trade for a value the code already determines. The executed mutation evidence is the unit one: 4 of 153 fail with `isActive: true` restored.

Re-requesting review. Since I'm on `read` here I can't set the reviewer myself — @pkarw, could you pick this up, or hand it to whoever owns the currencies module?

Two things I can't do from a `read` account, so they need a maintainer: the reviewer request above (`POST /pulls/5553/requested_reviewers` 404s for me), and flipping `changes-requested` → `review` now that the findings are addressed. `needs-qa`, `bug`, `priority-high` and `risk-high` should all stay as applied — the QA gate still holds.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-26T00:19:50Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Re-review: fix(currencies): fetch exchange rates for every currency, not just active ones

## 🎯 Summary

This is the second pass over the change that drops the `isActive: true` clause from `RateFetchingService.getExistingCurrencies()` (`packages/core/src/modules/currencies/services/rateFetchingService.ts:144`), leaving `deletedAt: null` as the single gate on the currency set that every provider response is filtered against. The one-clause fix itself is unchanged since the first review and it was already correct; everything new in this round is the evidence and documentation I asked for last time.

All five findings from the previous review are resolved, and I verified each one against the code rather than taking the response at face value:

- The **major** — the PR's documented off-switch (soft-delete) being blocked by `deleteCurrencyCommand`'s active-rates guard — is now stated accurately. I confirmed the guard counts `{ deletedAt: null, isActive: true }` rates on either leg and returns 400 (`packages/core/src/modules/currencies/commands/currencies.ts:374-389`), and that `storeRates` writes every new rate with exactly `isActive: true` (`rateFetchingService.ts:207`), so the rates this change creates are precisely the ones that block the delete. `services/README.md` now documents the two-step order and names every mechanism that can recreate the rates mid-window.
- **Minor 1** — the provider-gating test that did not exercise provider gating is genuinely fixed, and in the strongest of the available ways. The set assertion is now an exact `new Set(['PLN', 'EUR'])` rather than the `objectContaining({ size: 2 })` that would have passed with PLN swapped out, and the query-shape assertion at `rateFetchingService.basic.test.ts:358-366` was tightened from `expect.objectContaining({...})` to an exact object literal — which is what actually pins the removed clause, since a re-added `isActive: true` now fails the match instead of being silently tolerated. A new case drives the real `NBPProvider` with a stubbed `global.fetch` and asserts the request is issued.
- **Minor 2** — integration coverage on the money path now exists as `__integration__/TC-CUR-015-inactive-currency-rates.spec.ts`.
- **Minor 3** — the README names both the `maxDaysBack: 30` default and the `maxDaysBack: 60` that `customers/api/deals/{aggregate,summary}` actually pass, which is the window the PR's motivation rests on. I confirmed both numbers in the source.
- The **nit** — 47 commits behind `develop` — was cleared by merging `upstream/develop`; the branch is now one commit behind and `mergeable`.

The new `services/providers/__tests__/baseCurrencyGate.test.ts` was not something I asked for and it is the best addition in the round: it pins the other half of the regression directly on both bundled providers, proving a PLN-less set makes `NBPProvider` and `RaiffeisenPolandProvider` return `[]` without issuing a request. That is the mechanism behind the non-obvious blast radius the PR describes — a deactivated PLN silencing an entire provider rather than just its own pairs — and it was previously unpinned in either direction.

Reviewed scope: all six changed files, plus the delete-guard, `storeRates`, `ExchangeRateService` and deal-aggregation call sites the new documentation makes claims about.

## Verdict

✅ **approve** — the review found no blockers, no majors and no minors; the single remaining item is a nit about a follow-up that is out of scope for this PR. The diagnosis, the fix, and the regression pinning all hold up under direct verification, and `RateFetchingService` has no consumers outside the currencies module, so the blast radius is contained to the module that owns it.

## 🧪 Validation Gate

**Validation source: local fallback (scoped).** GitHub check data is unusable for this head: both workflow runs on `917ca52` (`CI for Develop&Main` and `Mutation tests`) sit at `action_required`, the state a fork PR's runs take until a maintainer approves them on every push. There is therefore no CI signal at all on this commit — not a green one and not a red one — so the checks-first path had nothing to read and I fell back to local commands scoped to the changed package.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ⏭️ NOT RUN | This checkout has no per-workspace install; `eslint-config-next` and several workspace binaries are absent, so the package build cannot run here. Covered by CI once the runs are approved. |
| `yarn generate` | ⏭️ NOT RUN | Same environment limitation. Its absence is why the unscoped `packages/core` suite reports failures for `#generated/*` module mappings in unrelated modules — an artifact of the missing `packages/core/generated/` directory, not of this PR. |
| `yarn build:packages` (second) | ⏭️ NOT RUN | Same. |
| `yarn i18n:check-sync` | ⏭️ NOT RUN | No locale files are touched by this diff. |
| `yarn i18n:check-usage` | ⏭️ NOT RUN | No user-facing strings are added by this diff. |
| `yarn typecheck` | ⚠️ PARTIAL PASS | Ran `tsc --noEmit -p packages/core/tsconfig.json`. It reports 1076 pre-existing errors, all of them the known `@tanstack/react-table/legacy` subpath cascade that follows from packages not being built in this checkout. **Zero errors fall in any file this PR touches** — the only `currencies` errors are in the untouched `backend/currencies/page.tsx` and `backend/exchange-rates/page.tsx`. The new integration spec is inside the tsconfig `include` and typechecks clean; the unit tests are excluded by `**/__tests__/**`. This confirms the author's own report. |
| `yarn test` | ✅ SCOPED PASS | `jest src/modules/currencies` in `packages/core`: **23 suites, 153 tests, all passing** — matching the author's reported figures exactly. |
| `yarn build:app` | ⏭️ NOT RUN | Same environment limitation. |
| `yarn agents:check-budget` | ✅ PASS | Run explicitly because the diff edits `packages/core/src/modules/currencies/AGENTS.md`. Exits 0; the three added lines do not push any instruction chain over its budget. |
| `yarn lint` | ⏭️ CANNOT RUN | `eslint-config-next` is not installed anywhere in this checkout, so `eslint.config.mjs` fails to load. This is an environment gap, not a PR problem, and it is the one gate for which I have no local evidence. The author reports lint clean on the changed files. |

**Mutation check — independently reproduced.** I re-added `isActive: true` to `getExistingCurrencies()` on this head and re-ran the module suite: **4 of 153 tests fail**, exactly as claimed. That is the important result, because it proves two things at once — the behavior is genuinely pinned rather than merely described, and the test run is resolving against this worktree's source rather than a stale build. I then reverted the mutation and confirmed a clean tree.

## Findings

### 💅 Nit

- `packages/core/src/modules/currencies/services/README.md:301-310` — the documented soft-delete procedure is correct, but it is worth recording that this change turns a previously deterministic operation into a racy one, and that the README stops just short of saying so. Before this PR the sequence was race-free: deactivating a currency removed it from the fetch set, so clearing its rates and then soft-deleting it could not be undone by a concurrent fetch. After this PR there is no way to take a currency out of the fetch set before deleting it, and `ExchangeRateService.autoFetch` (default `true`) calls the *whole-set* `fetchRatesForDate` on any conversion that misses its lookback window — so an ordinary conversion for an unrelated pair, triggered by any user, can recreate the rates between step 1 and step 2 and re-block the delete. The README does name `autoFetch` as one of the mechanisms that recreates rates, so nothing here is undisclosed; the gap is only that "with no fetch in between" reads as a sequencing instruction when it is in practice a race an operator may be unable to win on a busy tenant. This is not a change to make in this PR — it is the residue of the trade-off the PR states plainly and that was already settled in the author's internal review, and a real per-currency pause is a design change that needs its own spec. I will file a follow-up issue for it rather than hold the fix. Author's call whether to add half a sentence to the README noting that the window is not reliably closable under live traffic.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `getExistingCurrencies` is `private`; nothing exported changed.
- [x] No function signature changed in a breaking way — no signature in the diff changed at all.
- [x] No HTTP route URL removed or renamed; no method changed — `POST /api/currencies/fetch-rates` is untouched.
- [x] No field removed or retyped in an existing response shape — the `fetch-rates` response keeps `totalFetched` / `byProvider` / `errors`; only the values change, and only in the direction of more rows.
- [x] No CLI command or flag renamed or removed — `mercato currencies fetch-rates` is unchanged.
- [x] No database table or column renamed or removed; no column type narrowed — there is no migration in this PR and none is needed.
- [x] Where a contract had to change: not applicable — no contract surface from `BACKWARD_COMPATIBILITY.md` is touched. The change is a behavioral widening of which rows a private query returns.

The one behavior change users will observe is the intended one, and it is disclosed in the PR description: currencies that were previously deactivated resume receiving rates on the next fetch. Tenant and organization scoping on the query are preserved.

## 🧪 Test Coverage

Coverage is now genuinely strong, and specific about what it pins rather than merely present.

The unit layer covers the regression from both directions. `rateFetchingService.basic.test.ts` proves the service keeps an inactive currency in the set (`stores rates for inactive currencies`), keeps an inactive *base* currency in it so a gated provider still runs (`keeps an inactive base currency in the set so providers gated on it still run`, asserting the set by exact membership), and — driving the real `NBPProvider` against a stubbed `global.fetch` — that the provider actually issues its request and stores both legs. The exact-object assertion on the `em.find` call is what makes re-adding an `isActive` clause a test failure rather than a tolerated superset. `baseCurrencyGate.test.ts` covers the complementary direction on both bundled providers: a PLN-less set yields `[]` with no request issued, and each provider declares `PLN` as `providerBaseCurrency`. I ran that file on its own and confirmed all 4 cases execute.

The integration layer covers the money path end to end. `TC-CUR-015` creates whichever of USD/EUR is missing rather than assuming the demo seed ran, deactivates the non-base leg, posts to `/api/currencies/fetch-rates`, and asserts that rows land for both directions of the pair. Its design choices check out against the repo's conventions and against reality:

- It runs against `example_fixed_rates`, whose provider filters `EXAMPLE_RATES` on both legs against `availableCurrencies` (`apps/mercato/src/modules/example/lib/mock-currency-rate-provider.ts:26-29`). That is what makes the `2 after the fix, 0 before it` claim structurally true, and it keeps a live HTTP call and NBP's weekend/holiday 404s out of CI.
- The randomised far-future fetch date is necessary, not decorative: `storeRates` keys on `(from, to, date, source)` with no `deletedAt` filter, so a fixed date would let a re-run update a prior run's rows instead of creating visible ones. It mirrors the established `TC-EXAMPLE-015` pattern (year 2090 there, 2092 here, so the windows do not collide).
- `date` is a `timestamptz` column, so the full-precision ISO string the spec filters on round-trips correctly.
- I checked the one thing that worried me — that temporarily deactivating a shared currency could leak into other specs — and it cannot: `.ai/qa/tests/playwright.config.ts` sets `workers: 1`, and each CI integration shard is a separate job with its own ephemeral database. Teardown restores `isActive` and drops the created rates before deleting any created currency, which is the order the delete guard requires.

The remaining gap is the one the author already declared honestly: the integration spec's `2 vs 0` claim is derived from the provider's filter rather than from an executed mutation run. I am satisfied with that, because the executed mutation evidence exists at the unit layer and I reproduced it myself.

---

⏳ **CI status disclosure.** No required check has produced a result on `917ca52`. Both workflow runs are at `action_required`, awaiting maintainer approval of fork workflow runs — a state that recurs on every push to this branch and is easy to mistake for "CI has not started yet". This approval is based on the code review plus the scoped local evidence above; it is **not** a statement that CI is green, and branch protection still gates the merge on those checks. Whoever picks this up should approve the workflow runs so the full gate — including the `yarn lint` and `yarn build:app` steps I could not run locally — actually executes.

pkarw:

One framing note before you start, because it decides what "correct" looks like here. The whole change is a single removed clause: rate fetching used to skip currencies with `isActive: false`, and now it only skips soft-deleted ones. So the expected behaviour is **more** exchange-rate rows than before, specifically for currencies that are switched off. If you find yourself thinking "why is an inactive currency getting rates?", that is the fix working.

### P0 — Rates are stored for a deactivated currency

**Where to click**
- `/backend/currencies` — the currency list, to deactivate a currency
- `/backend/config/currency-fetching` — the "Fetch rates" action
- `/backend/exchange-rates` — the resulting rate rows

**What to verify**
- Pick a non-base currency that is currently active and involved in a pair your configured provider returns (with the bundled NBP or Raiffeisen providers, anything from the NBP table alongside PLN). Note how many rate rows exist for it in `/backend/exchange-rates`.
- Deactivate it from `/backend/currencies` (edit the currency, set it inactive, save). Confirm the list shows it as inactive and that it disappears from currency pickers — e.g. the currency select on the exchange-rate create form at `/backend/exchange-rates/create` should no longer offer it. That half of `isActive` must be unchanged by this PR.
- Trigger a fetch from `/backend/config/currency-fetching`. On `develop` today this stores nothing for the deactivated currency; with this PR it must store rates for **both directions** of its pairs.
- Back on `/backend/exchange-rates`, filter or sort to the deactivated currency and confirm new rows exist with it on the `from` side and on the `to` side, dated to the fetch date.
- Re-activate the currency afterwards so you leave the tenant as you found it.

**What can go wrong**
- Nothing is stored for the deactivated currency — the fix did not take effect on the path the UI actually uses.
- The currency stays selectable in pickers while inactive — that would mean the change leaked past rate fetching into selectability, which it must not.
- Rates appear for a **soft-deleted** currency. Soft delete is now the only off-switch, so it must still work: delete a disposable currency (see P1 for the required order) and confirm a later fetch stores nothing for it.

### P0 — The base-currency provider gate, and the total fetched count

**Where to click**
- `/backend/currencies`
- `/backend/config/currency-fetching`

**What to verify**
- This is the non-obvious half of the bug and the most valuable thing to exercise by hand. Both bundled providers quote everything against **PLN** and bail out of the entire fetch when PLN is missing from the currency set. So on `develop`, deactivating PLN silences NBP and Raiffeisen **completely** — not just PLN's own pairs.
- If your QA tenant has PLN and uses NBP or Raiffeisen: deactivate PLN, run a fetch, and confirm the fetch still returns a non-zero count covering every other pair. On `develop` the same steps return zero.
- Compare the reported "fetched" count before and after the change on a tenant that has at least one inactive currency. The count is expected to **increase**. A count that is unchanged on such a tenant means the fix is not reaching this path.
- Re-activate PLN when done.

**What can go wrong**
- A fetch that returns zero, or an error banner, after deactivating PLN — the provider gate is still being tripped.
- A fetch count that grows far beyond the provider's table size, or visibly duplicated rows for the same pair and date — re-running a fetch for the same date must update rows in place, not accumulate them. Run the same fetch twice and confirm the row count is stable the second time.

### P1 — Deleting a currency still works, in the documented order

**Where to click**
- `/backend/exchange-rates`
- `/backend/currencies`

**What to verify**
- This is the trade-off the PR documents and the part most likely to surprise an operator, so please exercise it rather than trusting the README. Create a disposable currency, let a fetch write rates for it, then try to delete it directly from `/backend/currencies`.
- The delete must be refused with a clear message along the lines of `Cannot delete currency XXX because it has N active exchange rate(s). Please delete or deactivate the exchange rates first.` Confirm the message is legible in the UI and not swallowed into a generic error toast.
- Then follow the documented order: delete or deactivate that currency's rate rows from `/backend/exchange-rates` first, and only then delete the currency. That must succeed.
- Confirm no later fetch resurrects the deleted currency's rates.

**What can go wrong**
- The 400 from the delete guard surfaces as an unhelpful generic failure with no indication of what to clear first — worth reporting as a UX finding even though it is pre-existing rather than introduced here.
- You clear the rates, and by the time you press delete they are back and the delete is refused again. That is a real and expected race: any conversion that misses its lookback window can trigger an automatic whole-set fetch. If you hit it, note how easily — it is the subject of a follow-up issue I am filing and real-world frequency data would be useful.

### P1 — Deal pipeline totals still convert

**Where to click**
- The CRM deals list and any pipeline/summary view that shows aggregated deal values

**What to verify**
- This is the consumer the bug was reported against, but its failure mode takes ~60 days of rate staleness to appear, so it cannot be reproduced in a QA session directly. The closest observable check is a regression check: confirm deal aggregates and pipeline summaries still render correct totals, in the expected currency, with no partial-conversion warning, on a tenant that has both active and inactive currencies in play.
- If any deal is denominated in a currency you deactivated during P0, confirm its value still appears in the totals after the fetch rather than dropping out silently.

**What can go wrong**
- Totals shift unexpectedly, or a currency starts converting at a rate it did not use before. More stored rates should mean more conversions succeed, never different answers for conversions that already worked.

### P2 — Documentation only

**Where to click**
- No UI. `packages/core/src/modules/currencies/services/README.md` and `packages/core/src/modules/currencies/AGENTS.md`.

**What to verify**
- No action needed for QA sign-off. Flagged only so the diff is fully accounted for: the docs changes carry no runtime behaviour.

### Pass/fail

pkarw:

Re-review of `917ca52`. All five findings from the first pass are resolved, and I checked each against the code rather than against the response: the delete-guard behaviour the README now documents (`commands/currencies.ts:374-389`), the `isActive: true` that `storeRates` writes (`rateFetchingService.ts:207`), the `maxDaysBack` 30/60 pair, and the tightened test assertions. The exact-object assertion on the `em.find` call is the one that actually pins the removed clause — the previous `objectContaining` form would have tolerated a re-added `isActive: true` silently.

I independently reproduced the mutation check rather than taking the reported number: re-adding `isActive: true` to `getExistingCurrencies()` on this head fails **4 of 153** tests, then reverted cleanly. The scoped suite (`jest src/modules/currencies` in `packages/core`) is **23 suites / 153 tests, all green**, matching the author's figures exactly, and `yarn agents:check-budget` exits 0 for the `AGENTS.md` edit. Result: 0 blockers, 0 majors, 0 minors, 1 nit.

The nit is not a change for this PR and I have not asked for one. It is that removing a currency went from a deterministic operation to a racy one — `ExchangeRateService.autoFetch` defaults to `true` and calls the whole-set fetch, so an unrelated conversion can recreate the rates between "clear the rates" and "soft-delete the currency". #5553 documents every mechanism involved, so nothing is hidden; the residue needs a design decision, not a patch. Filed as **#5630** with three concrete options and a recommendation.

`autofix: skipped (not my PR — re-run with --autofix to fix it here)`. Nothing needed fixing in any case.

Two things a maintainer still has to do. First, **CI has produced no signal on this head at all** — both `CI for Develop&Main` and `Mutation tests` sit at `action_required` on `917ca52`, the state fork PR runs take until a maintainer approves them, and that recurs on every push. Approving those runs is what will actually exercise `yarn lint` and `yarn build:app`, the two gates I could not run locally (`eslint-config-next` is not installed in this checkout). The approval above rests on the code review plus scoped local evidence and is explicitly **not** a claim that CI is green. Second, `needs-qa` stays on, so the QA-approval gate holds this out of the merge until `qa-approved` is added — manual QA instructions are posted above.

@Frshy — nothing further needed from you. Thanks for the thorough pass; the `baseCurrencyGate.test.ts` file was not something I asked for and it is the strongest addition in the round.

Issue #2966, opened 2026-06-10
perf(query_index,auth): add missing hot-path indexes — search_tokens(tenant_id, token_hash) and user_roles(user_id)/(role_id)


haxiorz:
Three of the platform's hottest read paths run sequential scans because the required composite indexes are missing. `TokenSearchStrategy` — the always-available global-search fallback and the only strategy on installs without Meilisearch — filters `search_tokens` by `token_hash IN (...) AND tenant_id` on every keystroke, but both existing `search_tokens` indexes lead with `entity_type` (and the lookup index interposes `field`, which the query never filters), so the lookup degrades to a full-table scan plus `GROUP BY`/`HAVING` as the table grows. Separately, `user_roles` carries only its FK constraints and zero secondary indexes — and Postgres does not auto-index FK columns — yet RBAC scans it by `user` on every ACL cache miss and by `role` on user-list and role-mutation paths.

**Affected code:**
- `packages/search/src/strategies/token.strategy.ts:80-102` — per-keystroke query: `.where('token_hash','in',hashes).where('tenant_id','=',...)` with `groupBy([entity_type, entity_id, organization_id])` + `having(count(distinct token_hash) >= minMatches)`
- `packages/core/src/modules/query_index/migrations/Migration20251212084132.ts:7-8` — the only two `search_tokens` indexes: `search_tokens_entity_idx (entity_type, entity_id)` and `search_tokens_lookup_idx (entity_type, field, token_hash, tenant_id, organization_id)`; neither can serve a `(tenant_id, token_hash)`-leading scan
- `packages/core/src/modules/query_index/data/entities.ts:254-256` — `SearchToken` entity declaring exactly those two `@Index` definitions
- `packages/core/src/modules/auth/data/entities.ts:170-186` — `UserRole` entity has **no** `@Index` decorator (FK relations only)
- `packages/core/src/modules/auth/migrations/Migration20251030150038.ts:25,40-41` — `user_roles` table created with only the `user_id` / `role_id` **FK constraints**, no secondary indexes
- `packages/core/src/modules/auth/services/rbacService.ts:197-203` — super-admin check scans `user_roles` by `user` on cache miss
- `packages/core/src/modules/auth/services/rbacService.ts:328-334` — ACL aggregation scans `user_roles` by `user` (with `role.tenantId`) on cache miss
[...]

PR #5555, opened 2026-08-24
perf(sales): index sales_notes and sales_document_addresses by document context

maxidragon:
## Summary

`sales_notes` and `sales_document_addresses` each declare exactly one index — `(organization_id, tenant_id)` — and neither table is ever read by that predicate. Both are read by **document context**, in the same `Promise.all` inside `loadOrderSnapshot` / `loadQuoteSnapshot` (`packages/core/src/modules/sales/commands/documents.ts`):

```ts
findWithDecryption(em, SalesNote,            { contextType: 'order', contextId: id }, undefined, scope)
findWithDecryption(em, SalesDocumentAddress, { documentId: id, documentKind: 'order' }, undefined, scope)
```

A scope index cannot serve that. On a single-organization deployment it matches *every row in the table*, so the planner discards it and sequentially scans. There was no index on `context_id` or `document_id` at all.

The cost multiplies through the command bus. `CommandBus` runs `handler.prepare()` before `execute` and `handler.captureAfter()` after it on **every** command, and for sales documents both load a full document snapshot — which reads both of these tables. A single document write is several commands (the document itself, one per changed line, the payment reconcile), so one written record performs roughly **20 of these reads**, each a full scan.

Separately, `order_id` and `quote_id` on both tables are the child side of `on delete set null` foreign keys. PostgreSQL does not index the child side of a constraint automatically, so every delete of a parent order or quote scanned the child table in full.

### Measured

Observed during a bulk import on a deployment where `sales_notes` holds ~331,000 rows (110 MB):

```
Parallel Seq Scan on sales_notes  (actual time=64.691..64.691 rows=0 loops=3)
  Filter: ((context_type = 'order') AND (context_id = $1) AND (organization_id = $2) AND (tenant_id = $3))
  Rows Removed by Filter: 110457
Execution Time: 82.450 ms
```

The cost is CPU filtering, so a warm cache does not reduce it. Counters over one import batch of 100 records showed **1,971 sequential scans reading 217,670,528 rows — 2.18 million rows per imported record**, against a table of 331k. Repeatable across consecutive batches: 1,971 / 1,974 / 1,911.

Every consumer of the sales module benefits from the fix; the deployment above is simply the one where the cost became visible.

## Changes

Six indexes, declared on the entities in `packages/core/src/modules/sales/data/entities.ts` and created in one migration:

| Table | Index | Columns | Serves |
|---|---|---|---|
| `sales_notes` | `sales_notes_context_idx` | `(context_id)` | the snapshot read |
| `sales_notes` | `sales_notes_order_idx` | `(order_id)` | FK check on order delete |
| `sales_notes` | `sales_notes_quote_idx` | `(quote_id)` | FK check on quote delete |
| `sales_document_addresses` | `sales_document_addresses_document_idx` | `(document_id)` | the snapshot read |
| `sales_document_addresses` | `sales_document_addresses_order_idx` | `(order_id)` | FK check on order delete |
| `sales_document_addresses` | `sales_document_addresses_quote_idx` | `(quote_id)` | FK check on quote delete |

- `packages/core/src/modules/sales/data/entities.ts` — the six `@Index` declarations
- `packages/core/src/modules/sales/migrations/Migration20260821120000_sales_notes_addresses_context_idx.ts` — the matching migration
- `packages/core/src/modules/sales/migrations/.snapshot-open-mercato.json` — updated so `db:generate` sees no drift
- `packages/core/src/__tests__/hot-path-indexes.test.ts` — extends the existing guard

### Why the context indexes are single-column

`context_type` / `document_kind` and the scope columns are omitted deliberately. `sales_orders.id` and `sales_quotes.id` are both `gen_random_uuid()` from their own tables, so a `context_id` already determines its `context_type` **and** its tenant; the seek returns the handful of rows belonging to one document and the remaining terms are rechecked over those. Folding three more columns in — one `text` and two `uuid` — roughly triples the index entry for no selectivity, on tables that take a write per document save and a note per order status change.

This follows `sales_document_tag_assignments_document_idx` (`Migration20260806120000`), the module's most recent decision on exactly this predicate shape, which omitted `document_kind` for the same reason. The older sibling indexes that *do* fold scope in (`sales_order_lines_scope_idx`, `sales_payments_scope_idx`) predate that reasoning. Happy to widen these to match the older convention if maintainers prefer the consistency — the narrow form is both smaller and no less selective, so I went with the newer precedent.

`order_id` / `quote_id` are separate indexes rather than a widening of the context index: the FK constraint check probes those columns, which the context index cannot serve even though it holds the same uuid.

### Migration shape

Built `CONCURRENTLY` — both tables are high-churn, so the build must not block writes. `CREATE INDEX CONCURRENTLY` cannot run inside a transaction, hence `isTransactional() => false`; the runner applies migrations one at a time, so the opt-out is safe. Each index is dropped first so that retrying a failed concurrent build removes PostgreSQL's INVALID stub rather than letting `IF NOT EXISTS` report success over an unusable index. Same shape as `Migration20260806120000` and `Migration20260731105052_query_index`.

On a fresh install the drops are no-ops and there is no window without the index. An operator who created one of these out of band should expect it to be rebuilt here.

### Declared on the entities, not only in SQL

The module diffs entity metadata against `.snapshot-open-mercato.json`, so an index that exists only in a migration is invisible to that diff and a later `db:generate` would propose dropping it. All six are on the entity decorators, and the snapshot is updated to match — `yarn db:generate` reports `sales: no changes` on this branch.

### Deliberately out of scope

The snapshot-per-command amplification described above is a separate and much larger design question, and is untouched here.

Not included, but the same defect class if maintainers want it in a follow-up: `sales_document_tag_assignments.order_id` / `.quote_id` are also unindexed `on delete set null` child columns.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**
N/A — this is a missing index, not a design change. No behavior, schema semantics, API surface or contract changes.

## Testing

Extended `packages/core/src/__tests__/hot-path-indexes.test.ts`, the repo's existing guard for this class of fix (#2966). It pins each index at all three declaration sites — entity decorator, migration SQL, schema snapshot — so a later refactor cannot silently drop one.

Run against this branch, rebased on `develop`:

| Command | Result |
|---|---|
| `yarn db:generate` | `sales: no changes` — the snapshot matches entity metadata exactly |
| `yarn generate` | pass |
| `yarn build:packages` | pass |
| `yarn i18n:check-sync` | pass |
| `yarn typecheck` | pass |
| `yarn test` (`@open-mercato/core`) | 1426 suites, 11457 tests, 0 failures |
| `yarn build:app` | pass |

Migrations are **not** applied here — the PR ships the migration file and snapshot only.

One note on the full-monorepo `yarn test`: the `@open-mercato/cli` `module-facts` generator suites flake under parallel load, failing a *different* suite on each run. I reproduced this on a clean `develop` checkout without this branch's commit (3/3 runs, different suite each time), so it is pre-existing and unrelated to this change.

## Checklist

- [x] I updated documentation, locales, or generators if the change requires it. — no user-facing strings and no locale changes; `yarn generate` produces no diff.
- [x] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required). — not required: this change adds indexes only. There is no API path or UI path to exercise, no request/response shape changes, and index selection is not observable from an integration test. The three-site guard above is the coverage that can actually fail if the change regresses.
- [ ] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — not applicable, see Specification above.
- [ ] QA routing set.

The last three are unticked because this account has read-only permissions here and cannot apply labels. Intended values are in a separate comment below for a maintainer to apply.

### Design System Compliance

Not applicable — this change touches no `.tsx` file, nothing under `packages/ui/src/`, and no component. The four changed files are two entity/migration files, a JSON schema snapshot, and a test.

## Linked issues

None.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5555"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790150237&installation_model_id=432243&pr_number=5555&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5555&signature=7ecad8f9c45408db8210c8ff88d06e86405ed69803981326467b30b72b46117e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/__tests__/hot-path-indexes.test.ts
+++ b/packages/core/src/__tests__/hot-path-indexes.test.ts
@@ -82,4 +82,64 @@ describe('hot-path indexes (#2966)', () => {
       expect(indexNames).toContain('user_roles_role_id_idx')
     })
   })
+
+  /**
+   * sales_notes (context_id) and sales_document_addresses (document_id), plus
+   * the order_id / quote_id FK columns on both.
+   *
+   * Every read of these tables filters by document context, never by scope
+   * alone: loadOrderSnapshot / loadQuoteSnapshot (sales/commands/documents.ts)
+   * fetch both in the same Promise.all, and the command bus runs them around
+   * every command via prepare() / captureAfter(). The lone
+   * (organization_id, tenant_id) index matches every row on a
+   * single-organization deployment, so the planner discarded it and scanned.
+   * The order_id / quote_id indexes cover the `on delete set null` child side,
+   * which Postgres does not index automatically.
+   */
+  describe('sales_notes and sales_document_addresses document indexes', () => {
+    it('declares the context and FK indexes on the SalesNote entity', () => {
+      const source = readEntities('sales')
+      expect(source).toContain("@Index({ name: 'sales_notes_context_idx', properties: ['contextId'] })")
+      expect(source).toContain("@Index({ name: 'sales_notes_order_idx', properties: ['order'] })")
+      expect(source).toContain("@Index({ name: 'sales_notes_quote_idx', properties: ['quote'] })")
+    })
+
+    it('declares the document and FK indexes on the SalesDocumentAddress entity', () => {
+      const source = readEntities('sales')
+      expect(source).toContain(
+        "@Index({ name: 'sales_document_addresses_document_idx', properties: ['documentId'] })",
+      )
+      expect(source).toContain("@Index({ name: 'sales_document_addresses_order_idx', properties: ['order'] })")
+      expect(source).toContain("@Index({ name: 'sales_document_addresses_quote_idx', properties: ['quote'] })")
+    })
+
+    it('creates all six indexes concurrently in a sales migration', () => {
+      const sql = readAllMigrationSql('sales')
+      expect(sql).toContain('create index concurrently "sales_notes_context_idx" on "sales_notes" ("context_id")')
+      expect(sql).toContain('create index concurrently "sales_notes_order_idx" on "sales_notes" ("order_id")')
+      expect(sql).toContain('create index concurrently "sales_notes_quote_idx" on "sales_notes" ("quote_id")')
+      expect(sql).toContain(
+        'create index concurrently "sales_document_addresses_document_idx" on "sales_document_addresses" ("document_id")',
+      )
+      expect(sql).toContain(
+        'create index concurrently "sales_document_addresses_order_idx" on "sales_document_addresses" ("order_id")',
+      )
+      expect(sql).toContain(
+        'create index concurrently "sales_document_addresses_quote_idx" on "sales_document_addresses" ("quote_id")',
+      )
+    })
+
+    it('records all six indexes in the sales schema snapshot', () => {
+      expect(readSnapshotIndexNames('sales', 'sales_notes')).toEqual(
+        expect.arrayContaining(['sales_notes_context_idx', 'sales_notes_order_idx', 'sales_notes_quote_idx']),
+      )
+      expect(readSnapshotIndexNames('sales', 'sales_document_addresses')).toEqual(
+        expect.arrayContaining([
+          'sales_document_addresses_document_idx',
+          'sales_document_addresses_order_idx',
+          'sales_document_addresses_quote_idx',
+        ]),
+      )
+    })
+  })
 })
--- a/packages/core/src/modules/sales/data/entities.ts
+++ b/packages/core/src/modules/sales/data/entities.ts
@@ -1790,6 +1790,9 @@ export class SalesPaymentAllocation {
 
 @Entity({ tableName: 'sales_notes' })
 @Index({ name: 'sales_notes_scope_idx', properties: ['organizationId', 'tenantId'] })
+@Index({ name: 'sales_notes_context_idx', properties: ['contextId'] })
+@Index({ name: 'sales_notes_order_idx', properties: ['order'] })
+@Index({ name: 'sales_notes_quote_idx', properties: ['quote'] })
 export class SalesNote {
   @PrimaryKey({ type: 'uuid', defaultRaw: 'gen_random_uuid()' })
   id!: string
@@ -1836,6 +1839,9 @@ export class SalesNote {
 
 @Entity({ tableName: 'sales_document_addresses' })
 @Index({ name: 'sales_document_addresses_scope_idx', properties: ['organizationId', 'tenantId'] })
+@Index({ name: 'sales_document_addresses_document_idx', properties: ['documentId'] })
+@Index({ name: 'sales_document_addresses_order_idx', properties: ['order'] })
+@Index({ name: 'sales_document_addresses_quote_idx', properties: ['quote'] })
 export class SalesDocumentAddress {
   @PrimaryKey({ type: 'uuid', defaultRaw: 'gen_random_uuid()' })
   id!: string
--- a/packages/core/src/modules/sales/migrations/.snapshot-open-mercato.json
+++ b/packages/core/src/modules/sales/migrations/.snapshot-open-mercato.json
@@ -1962,6 +1962,36 @@
           "keyName": "sales_document_addresses_scope_idx",
           "primary": false,
           "unique": false
+        },
+        {
+          "columnNames": [
+            "document_id"
+          ],
+          "composite": false,
+          "constraint": false,
+          "keyName": "sales_document_addresses_document_idx",
+          "primary": false,
+          "unique": false
+        },
+        {
+          "columnNames": [
+            "order_id"
+          ],
+          "composite": false,
+          "constraint": false,
+          "keyName": "sales_document_addresses_order_idx",
+          "primary": false,
+          "unique": false
+        },
+        {
+          "columnNames": [
+            "quote_id"
+          ],
+          "composite": false,
+          "constraint": false,
+          "keyName": "sales_document_addresses_quote_idx",
+          "primary": false,
+          "unique": false
         }
       ],
       "checks": [],
@@ -3748,6 +3778,36 @@
           "composite": true,
           "constraint": false,
           "keyName": "sales_notes_scope_idx",
+          "primary": false,
+          "unique": false
+        },
+        {
+          "columnNames": [
+            "context_id"
+          ],
+          "composite": false,
+          "constraint": false,
+          "keyName": "sales_notes_context_idx",
+          "primary": false,
+          "unique": false
+        },
+        {
+          "columnNames": [
+            "order_id"
+          ],
+          "composite": false,
+          "constraint": false,
+          "keyName": "sales_notes_order_idx",
+          "primary": false,
+          "unique": false
+        },
+        {
+          "columnNames": [
+            "quote_id"
+          ],
+          "composite": false,
+          "constraint": false,
+          "keyName": "sales_notes_quote_idx",
           "primary": false,
           "unique": false
         }
--- a/packages/core/src/modules/sales/migrations/Migration20260821120000_sales_notes_addresses_context_idx.ts
+++ b/packages/core/src/modules/sales/migrations/Migration20260821120000_sales_notes_addresses_context_idx.ts
@@ -1,0 +1,87 @@
+import { Migration } from '@mikro-orm/migrations';
+
+// sales_notes and sales_document_addresses each carried a single index —
+// (organization_id, tenant_id) — and are never read by that predicate alone.
+// loadOrderSnapshot / loadQuoteSnapshot (commands/documents.ts) read them by
+// context in the same Promise.all:
+//
+//   findWithDecryption(em, SalesNote, { contextType, contextId, ...scope })
+//   findWithDecryption(em, SalesDocumentAddress, { documentId, documentKind, ...scope })
+//
+// On a single-organization deployment the scope index matches every row, so the
+// planner discards it and scans. The cost is CPU filtering, so a warm cache does
+// not reduce it. Measured on a table of ~331k rows / 110 MB: a parallel
+// sequential scan removing 110,457 rows per worker, 82 ms per read to return 0
+// rows. The command bus runs prepare() and captureAfter() around every command
+// and both load a document snapshot, so one written document — itself several
+// commands (the document, one per changed line, the payment reconcile) —
+// performs roughly 20 of these reads. Counters over one bulk-import batch of 100
+// records showed 1,971 sequential scans reading 217,670,528 rows, 2.18M rows per
+// imported record; repeatable across consecutive batches (1,971 / 1,974 / 1,911).
+//
+// context_type and document_kind are deliberately omitted, as they were for
+// sales_document_tag_assignments_document_idx (Migration20260806120000), and the
+// scope columns with them. sales_orders.id and sales_quotes.id are both
+// gen_random_uuid() from their own tables, so a context_id already determines its
+// context_type and its tenant; the seek returns the handful of rows belonging to
+// one document and the remaining terms are rechecked over those. Folding three
+// more columns in — one text and two uuids — roughly triples the index entry for
+// no selectivity, on tables that take a write per document save and a note per
+// order status change. The sibling scope indexes that do fold scope in
+// (sales_order_lines_scope_idx, sales_payments_scope_idx) predate that reasoning.
+//
+// order_id and quote_id are separate indexes rather than a widening of the above:
+// both tables are the child side of `on delete set null` constraints, which
+// PostgreSQL does not index automatically, and the constraint check probes
+// order_id / quote_id — columns the context index cannot serve even though they
+// hold the same uuid. Without them every delete of a parent order or quote scans
+// the child table in full.
+//
+// Both tables are high-churn, so the indexes are built CONCURRENTLY to avoid
+// blocking writes during the build. CREATE INDEX CONCURRENTLY cannot run inside a
+// transaction, hence isTransactional() => false; the migration runner applies
+// migrations one-by-one, so this opt-out is safe. Drop first so retrying a failed
+// concurrent build removes PostgreSQL's invalid index stub instead of letting IF
+// NOT EXISTS silently accept it: a half-built concurrent index is left INVALID,
+// the planner ignores it, and IF NOT EXISTS would report success while the table
+// stays unindexed. Same shape as Migration20260806120000 and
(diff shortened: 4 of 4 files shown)


maxidragon:
### Label intent (this account is read-only here)

I can't apply labels on this repo, so recording the intended metadata for a maintainer. Nothing here is a claim on the PR.

| Label | Value | Rationale |
|---|---|---|
| pipeline | `review` | Ready for code review; not draft, not blocked. |
| category | `bug` | A missing index that makes every sales-document read a sequential scan — a defect, not a new capability. |
| priority | `priority-medium` | Measured, repeatable production cost during bulk imports, but nothing is broken or blocked: writes still succeed, they are just slow. Not urgent enough for `priority-high`. |
| risk | `risk-medium` | Additive, index-only DDL confined to one module's two tables — no behavior, API surface or contract change. Not `risk-low` because the migration is deliberately non-transactional (`CREATE INDEX CONCURRENTLY`), so a failed build needs an operator retry rather than rolling back with the transaction. |

**QA routing — one judgement call worth a maintainer's eye.** I've assumed `needs-qa`, reading the checklist's carve-out strictly: it requires no manually exercisable UI surface **and** database structure unchanged, and this adds indexes, which is a database structure change.

In practice there is nothing here a human can click. No `.tsx`, nothing under `packages/ui/src/`, no API route, no request/response shape, no `BACKWARD_COMPATIBILITY.md` contract touched, and automated tests ship in the PR. If you read "database structure" as schema *semantics* rather than any DDL, this is a clean `skip-qa` and I'd suggest applying that instead — a `needs-qa` gate on an index-only change costs a QA cycle with no surface to exercise. Your call; I defaulted to the stricter reading rather than choosing the convenient one for myself.

**Reviewer:** a maintainer familiar with the sales module and its migration conventions — the one substantive decision to check is the single-column index shape (`(context_id)` / `(document_id)` rather than folding in `context_type`/`document_kind` and the scope columns), reasoned through in the PR body and in the migration's header comment.

**Assignee:** none set; leave to maintainer routing.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-25T00:37:02Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
## Code Review: perf(sales): index sales_notes and sales_document_addresses by document context

**Verdict: APPROVE** — 0 blocker, 0 major, 0 minor, 2 nit.

**Validation source: GitHub checks** (all green on head `d8fa5420`: `prepare`, `lint`, `ds-lint`, `scope`, `audit-scope`, `test`, `ephemeral-integration`, `documents-multi-instance`, `license/cla`; both workflow runs `success`). No local fallback gate was needed.

> Note for the author: the two workflow runs were parked at `action_required` (fork PRs need per-push CI approval), so this PR had **no** usable check signal at all when review started — only `license/cla`. I approved both runs after reading the diff; that is why CI results now exist.

---

### Premise verification

The diagnosis holds up against the code, not just the PR description:

- `loadOrderSnapshot` / `loadQuoteSnapshot` read both tables in the same `Promise.all` with exactly the predicates the new indexes serve — `{ documentId, documentKind }` and `{ contextType, contextId }` (`packages/core/src/modules/sales/commands/documents.ts:1467`, `:1741`).
- Before this PR, `sales_notes` and `sales_document_addresses` each carried exactly two indexes — `pkey(id)` and `scope_idx(organization_id, tenant_id)`. Nothing led with `context_id` or `document_id`.
- All four `order_id` / `quote_id` columns are confirmed `deleteRule: "set null"` FKs in the module snapshot, with **no** existing index. PostgreSQL does not index the child side, so the RI check on parent delete really did scan the child table in full. The separate FK indexes are justified and are not redundant with the context index, even where they hold the same UUID.

I audited **every** predicate issued against both tables across `packages/` and `apps/` — commands, API routes, reindexer, coverage, search config, subscribers, workers, enrichers, raw SQL. Result: every hot predicate leads with `context_id`, `document_id`, or `id`, so all of them are served. Two observations, neither a defect in this PR:

- `GET /api/sales/notes` makes both `contextId` and `contextType` optional (`api/notes/route.ts:164`), so a call supplying neither still leads with `organization_id`/`tenant_id` and falls back to `scope_idx`. Pre-existing; the only shipped caller (`documents/[id]/page.tsx:4065`) always sends both.
- `api/document-history/route.ts:104` filters by `(context_type, context_id, tenant_id, organization_id)` and sorts `created_at DESC` with **no LIMIT**. `sales_notes_context_idx` turns that from a full scan into a seek plus a sort over the handful of rows for one document — a clear win, and the missing LIMIT is out of scope here.

### Correctness of the artifacts

- **Entity decorators** (`data/entities.ts:1793-1795`, `:1842-1844`) — `properties: ['order']` / `['quote']` correctly target the `ManyToOne` fields mapped to `order_id` / `quote_id`.
- **Snapshot** — the six new entries match the tool-generated shape byte-for-byte in structure, compared against `user_roles_user_id_idx` (auth) and `sales_document_tag_assignments_document_idx` (sales). Index ordering within the array is irrelevant to MikroORM's comparator, which keys on `keyName`.
- **Migration** — `CONCURRENTLY` + `isTransactional() => false` + drop-before-create is a line-for-line match of the already-merged `Migration20260806120000`, and the runner applies migrations one at a time (`packages/cli/src/lib/db/commands.ts:406`), so the opt-out behaves as described. `down()` correctly reverses in inverse order.
- **Hand-written SQL** — normally a finding under the checklist's "no hand-written migrations" rule, but `CREATE INDEX CONCURRENTLY` cannot be emitted by `db:generate`, the SQL is scoped to exactly this change, and the snapshot is updated in the same commit. That is the documented exception in `AGENTS.md`, with two merged precedents. Not a finding.
- **Tests** — `hot-path-indexes.test.ts` pins all six indexes at all three declaration sites (decorator, migration SQL, snapshot), matching the existing guard's structure. Appropriate coverage: index *selection* is not observable from an integration test, so the three-site guard is the coverage that can actually fail on regression.
- **Backward compatibility** — purely additive. No column, table, event, route, response field, export, or migration semantics changed. Confirmed additive against current `develop` (`git diff --stat origin/develop` shows +213/-0 across four files), with no duplication of already-merged work.

---

### Nits (non-blocking)

**Nit 1 — the migration comment misstates the call signature.** The header sketches the read as:

```
findWithDecryption(em, SalesNote, { contextType, contextId, ...scope })
```

The `...scope` spread implies the scope columns are part of the SQL predicate. They are not: the 5th argument of `findWithDecryption` is a *decryption* scope, never a filter (`packages/shared/src/lib/encryption/find.ts:20-37`), and there is no MikroORM global filter on these entities. The actual command-path SQL is `WHERE context_type = ? AND context_id = ?` — nothing else.

This strengthens your argument rather than weakening it: on that path the scope index is not merely low-selectivity, it is entirely inapplicable. But a future reader auditing tenant scoping from this comment would be misled into thinking scope is enforced in SQL there. Worth correcting the sketch to the real shape.

The same detail explains a small inconsistency in the PR body: the quoted plan's `Filter:` line lists `organization_id` and `tenant_id`, which the `loadOrderSnapshot` query does not emit — so that EXPLAIN is presumably from the notes-list or document-history route. It does not change the conclusion, since `context_id` leads either way.

**Nit 2 — optional: partial indexes on the FK columns.** `order_id` and `quote_id` are mutually exclusive per row, so each of the four FK indexes stores a NULL entry for every row belonging to the other document kind — on a mostly-orders table, `*_quote_idx` is almost entirely NULLs. `WHERE order_id IS NOT NULL` / `WHERE quote_id IS NOT NULL` would roughly halve them and stay usable by the RI check, since `order_id = $1` implies `order_id IS NOT NULL` and the planner can prove the implication. Speculative without knowing your order/quote split, and not worth holding the PR for — file it as a follow-up if index size on these tables ever matters.

---

### Follow-up worth filing

You flagged it yourself: `sales_document_tag_assignments.order_id` / `.quote_id` are the same defect class — unindexed `on delete set null` child columns. Agreed it belongs in a separate PR.

### Note for whoever applies the migration

`DROP INDEX CONCURRENTLY` before `CREATE INDEX CONCURRENTLY` means an operator who created any of these six out of band gets it rebuilt, with the pre-index cost in force until the build finishes. Your migration comment already says this; repeating it here so it reaches the release notes. On a fresh install the drops are no-ops.

Nice work — the reasoning for narrow single-column indexes over folding in `context_type`/`document_kind`/scope is correct and well argued, and citing `Migration20260806120000` as the governing precedent (rather than the older scope-folding siblings) is the right call.

pkarw:

Priority set to `priority-high` because this is a measured production regression on a partner deployment (2.18M rows scanned per imported record) and matches the rating on the equivalent #5145.

Risk set to `risk-high` because the change ships a migration that drops and rebuilds six indexes `CONCURRENTLY` on two high-churn sales tables — the code blast radius is nil, the operational one is not.

QA routing set to `needs-qa`: the change touches no `.tsx`, nothing under `packages/ui/src/` and no component, but it does alter database structure, so the automated-verification exemption does not apply. Same routing as #5145.

pkarw:

This PR is approved and requires manual QA (`needs-qa`, no `skip-qa`). It is queued in `merge-queue` but the QA-approval gate holds it until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` by swapping the labels (remove `merge-queue`, add `qa`), then run the checks below.

**Read this first — there is no UI or API change to click through.** The diff is six database indexes, one migration, a schema snapshot and a test guard. No route, response shape, form, or rendered string changes. So QA here is *regression* QA (the sales document surfaces that read these two tables must behave identically) plus *migration* QA (the migration must apply cleanly and leave `db:generate` clean). Do not expect any visible difference.

### P0 — Migration applies cleanly

**Where to run**
- A test/staging database with existing sales data — ideally some orders **and** some quotes that have notes and addresses. A fresh install exercises the no-op-drop path only, so it is the weaker test.

**What to verify**
- `yarn db:migrate` applies `Migration20260821120000_sales_notes_addresses_context_idx` without error → all six indexes exist and are `indisvalid = true`:
  ```sql
  SELECT c.relname, i.indisvalid
  FROM pg_class c JOIN pg_index i ON i.indexrelid = c.oid
  WHERE c.relname IN (
    'sales_notes_context_idx','sales_notes_order_idx','sales_notes_quote_idx',
    'sales_document_addresses_document_idx','sales_document_addresses_order_idx','sales_document_addresses_quote_idx');
  ```
  Expect six rows, all `t`.
- `yarn db:generate` afterwards reports **`sales: no changes`** — this is the check that proves the snapshot matches entity metadata. A proposed drop/create here is a fail.
- Re-running `yarn db:migrate` is a no-op (nothing pending).

**What can go wrong**
- The migration hangs instead of completing — `CREATE INDEX CONCURRENTLY` waits on open transactions. If it stalls, check `pg_stat_activity` for long-running transactions rather than assuming the migration is broken.
- An index lands `indisvalid = false` (a failed concurrent build leaves an INVALID stub). Re-running the migration is designed to clear this, since it drops before creating — verify that recovery actually works if you hit it.
- `db:generate` proposes changes → snapshot drift, which would make a later unrelated PR emit a spurious migration.

### P1 — Sales document read/write regression

**Where to click**
- `/backend/sales/documents` — the orders/quotes grid
- `/backend/sales/documents/<id>` for an **order** that has notes and at least one address
- `/backend/sales/documents/<id>` for a **quote** that has notes and at least one address

**What to verify**
- Document detail loads with the correct notes and addresses — same content and count as before the migration. These are the exact reads the new indexes serve, so a wrong result here would mean an index changed row visibility (it should not).
- Add, edit, and delete a note on both an order and a quote → each persists and the list refreshes correctly.
- Add, edit, and delete an address on an order → persists correctly.
- Change an order's status → the automatic status-change note still appears.
- The document history / activity panel on a document still lists notes newest-first.
- Convert a quote to an order → notes and addresses carry across as before.

**What can go wrong**
- Missing or duplicated notes/addresses on a document (would indicate the index changed planner behavior in a way that surfaced an existing predicate bug).
- Cross-document bleed: a note or address from document A appearing on document B. Worth probing deliberately, because the new indexes are single-column on `context_id` / `document_id` with `context_type` / `document_kind` rechecked afterwards.
- Slower, not faster, document loads (index build not finished, or planner regression).

### P1 — Parent delete path

**Where to click**
- Delete an order that has notes and addresses; delete a quote that has notes and addresses.

**What to verify**
- The delete completes and is not noticeably slower than before. This is the `on delete set null` FK path the `order_id` / `quote_id` indexes were added for.
- Undo the delete where the UI offers it → notes and addresses are restored intact.

**What can go wrong**
- Delete errors or leaves orphaned notes/addresses.
- Undo restores a partial graph.

### P2 — Cross-tenant isolation spot check

**What to verify**
- With two organizations that both have sales documents, confirm a document detail page in org A never shows notes or addresses belonging to org B.

**What can go wrong**
- The new indexes do not include the scope columns. That is a deliberate, reviewed decision and scoping is unchanged by this PR — but since these tables are now reached by a different plan, one explicit isolation check is cheap insurance.

### Pass/fail
- All checks pass → remove the `qa` label and add `merge-queue` plus `qa-approved` (this clears the QA-approval gate)
- Any check fails → remove the `qa` label, add `qa-failed`, and leave a comment describing the failure.

pkarw:

0 blocker, 0 major, 0 minor, 2 nit — no autofix iterations were needed. Validation source: GitHub checks (all green on `d8fa5420`); the fork's workflow runs were approved during review because they were parked at `action_required` with no check signal. Labels: `merge-queue` + `needs-qa` + `priority-high` + `risk-high` + `performance`. The QA-approval gate now holds the merge until a QA reviewer adds `qa-approved`; manual QA instructions are posted above.

PR #5557, opened 2026-08-24
feat(security): complete P0/P1 security readiness controls

WebEferen:
## Summary

Moves the security-readiness implementation series directly onto `develop`.

The branch was rebuilt from the current `develop` head and contains 26 scoped commits covering `OM-SEC-001` through `OM-SEC-015`, the extended `OM-SEC-010A` through `OM-SEC-010F` privacy work, and the build compatibility fixes required by the port.

### MIT core and shared platform

- redact sensitive command and audit payloads
- fail closed when production encryption is incomplete
- add access-log fields, configurable retention, rotation, and export
- add an attachment malware-scanning and quarantine extension point
- enforce hardened AI runtime and content-safety controls
- export signed, integrity-verifiable audit evidence
- provide privacy data-class contracts and module-owned handlers
- add tenant exit export with sensitive-field redaction
- add security CI gates and public security contact metadata

### Enterprise modules

- enforce SSO and MFA login assurance
- complete SCIM 2.0 users, groups, memberships, and bulk lifecycle support
- add audited backup and restore CLI operations with checksums and receipts
- add retention, legal hold, DSR discovery/export/erase/anonymize, scheduling, and administration
- reapply completed erasures after restore
- add classification-gated non-production sanitization

## Agent Orchestrator separation

This PR does not contain Agent Orchestrator module paths, schema fields, adapters, acceptance fixes, or Agent Orchestrator-specific specifications. Those changes remain on the Agent Orchestrator MVP line.

The generic AI security controls remain included because they belong to the shared AI runtime and do not depend on Agent Orchestrator.

## Validation

- `yarn generate`: PASS
- `yarn typecheck`: PASS, 27 tasks
- `yarn lint`: PASS, with existing warnings
- `yarn build:packages`: PASS, 27 tasks
- `yarn build:app`: PASS
- security and compatibility regression tests: PASS
- full monorepo test run: all executed tests passed except two unrelated CLI Jest workers terminated with `SIGSEGV` under parallel load; both affected suites passed separately, 17/17
- `yarn agents:check-budget`: PASS
- `yarn lessons:check`: PASS
- i18n hardcoded and value checks: PASS, advisory findings unchanged

## QA and rollout

This PR changes authentication, encryption, database schemas, API routes, deployment configuration, and enterprise UI flows. It requires target-environment QA before merge or rollout.

Deployment acceptance should include:

- KMS fail-closed startup and recovery
- OIDC, MFA, and SCIM lifecycle tests against the selected identity provider
- access-log retention and signed evidence-export verification
- backup and restore drill with receipts and checksum validation
- attachment scanning with the selected scanner adapter
- privacy retention, legal-hold, and DSR dry-runs
- non-production sanitization before user access
- hardened AI mutation approval and content-safety verification


WebEferen:
Label rationale: priority-high because this closes the security-readiness P0/P1 release scope. risk-high because the branch changes auth, encryption, tenant-scoped audit behavior, database migrations, API routes and shared contract surfaces. needs-qa because it includes target-environment identity, privacy, backup, scanning and enterprise UI/API flows that require deployment acceptance in addition to automated coverage.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-30T20:15:15Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review

**PR #5557 — `feat(security): complete P0/P1 security readiness controls`**
315 files changed, +22 408 / −639, 26 commits, head `68cab3da`, base `develop`.

## 📋 Summary

This is a genuinely valuable body of work. The MFA fail-closed fix in the login interceptor closes a real bypass (a failed `createChallenge` used to fall through and hand the caller a full staff token), the audience-scoped pending-MFA token is a solid hardening of the provisional-session model, the signed audit-evidence bundle is well constructed (canonical JSON, hash chain, `timingSafeEqual` comparison, minimum key length), and the backup service correctly spawns Postgres tooling with `shell: false`.

It cannot be approved in its current state. The review found **7 blockers and 3 majors**, plus a set of Medium findings that this repository's verdict rule also treats as request-changes. The two most consequential facts are that **no CI has run against the current head at all**, and that **the security gates this PR itself introduces fail on this PR** — including one (CodeQL) that can never pass until a repository setting is changed.

## 🛠️ Validation

**Validation source: local fallback** — GitHub reports no check runs for head `68cab3da` (pushed 2026-08-27); the newest workflow runs on this branch are from 2026-08-24 against a pre-rebase SHA. `Runner: local` (no `app` container running for either compose candidate).

| Gate command | Result |
|---|---|
| `yarn build:packages` | ✅ PASS (27 tasks) |
| `yarn generate` | ✅ PASS — no working-tree drift afterwards |
| `yarn build:packages` (rebuild) | ✅ PASS |
| `yarn i18n:check-sync` | ❌ **FAIL (exit 1)** — 10 issues across 2 modules |
| `yarn i18n:check-usage` | ✅ PASS (advisory: 3 828 unused keys) |
| `yarn typecheck` | ✅ PASS (27 tasks) |
| `yarn test` | ❌ **FAIL (exit 1)** — `create-mercato-app#test` |
| `yarn build:app` | ✅ PASS |
| `yarn template:sync` (repo extension) | ⚠️ drift on `modules.ts` — **reproduces identically on `develop`, so pre-existing and not attributed to this PR** |

On the `yarn test` failure: `@open-mercato/shared` passes cleanly (2 045 tests) once `TMPDIR` is short enough for a unix socket. Of the seven `create-mercato-app` failures, five are host-environment artefacts (Linux inotify limits for Turbopack, one spawned process killed by signal) and **two are caused by this PR** — see 🔴 Blocker 1.

## 🔴 Blockers

**1. `yarn test` fails: `tenant_exports` is shipped in the scaffold with no evaluation-catalog case**
`apps/mercato/src/modules.ts:75` and `packages/create-app/template/src/modules.ts`

Adding `{ id: 'tenant_exports', from: '@open-mercato/core' }` to the enabled-module list makes the module ship a fact sheet, and two guards in `packages/create-app/src/lib/module-facts-build.test.ts` fail on it:

```
✖ every default-controller module fact is exercised by the evaluation catalog
  module facts without an evaluation case:
  .ai/guides/modules/tenant_exports/index.md

✖ every module fact-sheet a scaffold ships is required by at least one catalog case
  these shipped module fact-sheets are in no case's context.required: tenant_exports.
  Add a case whose prompt forces the read and lists .ai/guides/modules/<id>/index.md
  in context.required ..., or stop enabling the module in the template.
```

The assertion message states the remedy exactly: add a catalog case, or do not enable the module in the template.

**2. `yarn i18n:check-sync` fails: unsorted locale keys**

10 issues across `packages/core/src/modules/audit_logs/i18n/{en,pl,es,de,ko}.json` and `packages/enterprise/src/modules/sso/i18n/{en,pl,es,de,ko}.json`. Verified against the base: `develop`'s `audit_logs/en.json` is sorted, this branch's is not; `git log` attributes it to `e9e720e2 feat(security): [OM-SEC-004] harden access audit retention`.

This is not only a configured gate command — `.github/workflows/ci.yml:627` runs `yarn tsx scripts/i18n-check-sync.ts` in the `lint` job **without `continue-on-error`**, so it will fail CI as soon as CI runs on this head. Fix: `yarn i18n:check-sync --fix`.

**3. No CI has run against the current head**

`gh api repos/.../commits/68cab3da/check-runs` returns only `[code]smith` (skipped) and `.github/dependabot.yml`; the only status is `license/cla`. The last real runs (`dc9c0ce0`, 2026-08-24) both failed. For a PR that rewrites authentication, encryption, schema, API routes and CI itself, there is currently **zero** integration evidence for the code that would actually merge. Please push an empty commit or re-run the workflows so the head is exercised.

**4. The new CodeQL job can never pass on this repository**
`.github/workflows/security.yml:24-49`

```
##[error]Code Scanning could not process the submitted SARIF file:
CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled
```

GitHub's CodeQL *default setup* is enabled for this repo, and it is mutually exclusive with an advanced workflow. Merging this file as-is adds a permanently red check. Either disable default setup in repository settings first (an admin action that should be sequenced explicitly in the rollout), or drop the job.

**5. The new `Secrets, configuration and SBOM` job fails on a pre-existing Dockerfile**
`.github/workflows/security.yml:63-84`

Trivy runs with `scanners: secret,misconfig`, `severity: HIGH,CRITICAL`, `exit-code: '1'` over the whole repository and fails on `docker/preview/Dockerfile`:

```
DS-0002 (HIGH): Specify at least 1 USER command in Dockerfile with non-root user as argument
Tests: 20 (SUCCESSES: 19, FAILURES: 1)
```

`Dockerfile`, `packages/create-app/template/Dockerfile` and `apps/docs/Dockerfile` all declare a `USER`; `docker/preview/Dockerfile` does not. Introducing a repo-wide gate requires the repo to be clean for it first — add the `USER` directive (or a justified `.trivyignore` entry), otherwise every PR after this one is red on day one.

**6. The new image scan fails on the image this repo builds**
`.github/workflows/ci.yml:1096-1108`

The `docker-build` job now loads the image and scans it with `ignore-unfixed: true`, `severity: HIGH,CRITICAL`, `exit-code: '1'`:

```
Total: 10 (HIGH: 9, CRITICAL: 1)   [alpine / node-pkg]
Total: 10 (HIGH: 10, CRITICAL: 0)  [gobinary — e.g. CVE-2026-39822]
```

All twenty are *fixable* (that is what `ignore-unfixed` leaves). Same problem as blocker 5: the gate is landing red. Base-image and dependency bumps need to be part of this change, or the gate needs a documented, time-boxed allowlist.

**7. The `audit` job fails on two unallowlisted high advisories**

```
audit-ci: 2 advisory(ies) at or above high:
  [high] deepmerge-ts <8.0.0 — GHSA-ggr8-5vv4-36mx
  [high] nanoid <3.3.18      — GHSA-2v37-7h3g-55p8
```

These are latent on `develop` too (the `audit-scope` job only arms `audit` when a manifest or the lockfile changes, and `develop`'s recent green runs skipped it) — but this PR edits `packages/shared/package.json`, which re-arms it. It therefore blocks *this* PR and needs either a resolution bump or an allowlist entry with the same justification style already used for the two `image-size` advisories.

## 🟠 Majors

**8. Login silently does nothing when MFA is unavailable**
`packages/enterprise/src/modules/security/api/interceptors.ts:59-70` → `packages/core/src/modules/auth/api/login.ts:240-248` → `packages/core/src/modules/auth/frontend/login.tsx:319-330`

The new fail-closed path is correct in the backend and invisible in the frontend. `InterceptorAfterResult` (`packages/shared/src/lib/crud/api-interceptor.ts:41-44`) has no `statusCode` field, so `mfaUnavailableResponse()` can only replace the body; the login route then returns **HTTP 200** with `{ ok: false, code: 'MFA_UNAVAILABLE', error: … }`.

Failure scenario: the MFA verification service is unregistered or `createChallenge` throws. The user submits valid credentials. `login.tsx` takes the `res.ok === true` branch, skips the whole error-message block, calls `emitLoginResponseEvent(data)` / `clearAllOperations()` / `notifyAuthIdentityChange()`, finds no `data.redirect`, and returns. `parseMfaRequiredState` in `packages/enterprise/src/modules/security/widgets/components.ts:63-64` rejects the payload because there is no `challenge_id`, so the MFA panel does not render either. **The user sees the login form reset with no error and no navigation, indefinitely.** The server-side session is correctly deleted, so they are simply locked out with no explanation.

The interceptor contract needs a way to set a status (or the login route needs to map `code: 'MFA_UNAVAILABLE'` to a non-2xx response), and `login.tsx` should surface `ok: false` bodies on 200 responses.

**9. `UPGRADE_NOTES.md` is not updated, despite several breaking operational changes**

Root `AGENTS.md` requires documenting behaviour changes to contract surfaces in `UPGRADE_NOTES.md`; the file is untouched. At minimum these change the behaviour of existing deployments on upgrade:

- **Production now hard-fails at boot** unless tenant encryption is enabled *and* a healthy KMS is reachable — `assertTenantDataEncryptionConfiguration` at `packages/core/src/bootstrap.ts:198`, plus `createRequestContainer` rethrowing when required (`packages/shared/src/lib/di/container.ts:258`, `:325`). `TENANT_DATA_ENCRYPTION_REQUIRED=false` is deliberately ignored in production.
- **Access-log writes are blocking by default.** `shouldBlockAccessLogWrites()` (`packages/shared/src/lib/crud/factory.ts:786-796`) flipped from "off unless `OM_CRUD_ACCESS_LOG_BLOCKING=1`" to "on unless explicitly opted out". Every CRUD list read now awaits an audit write inline.
- **`OM_CRUD_ACCESS_LOG_BLOCKING` is deprecated** in favour of `OM_CRUD_ACCESS_LOG_MODE` with no `@deprecated` marker anywhere.
- **Access-log retention jumps from 7 days / 8 hours to a 90-day floor**, with legacy values clamped *upward* (`resolveAccessLogRetentionDays`, `packages/core/src/modules/audit_logs/services/accessLogService.ts:47-66`). That is a large, silent storage-growth change.
- Rotation-on-write was removed in favour of a queue worker; the upgrade action at `packages/core/src/modules/configs/lib/upgrade-actions.ts:129` covers existing tenants, which is good — but it is worth calling out explicitly.

**10. Tenant exit export guards secrets with a hard-coded, cross-module denylist**
`packages/core/src/modules/tenant_exports/services/tenantExitExportService.ts:26-38`

`EXCLUDED_SECURITY_TABLES` is a literal set of 12 table names living in a **core** module — and it names **enterprise** tables (`scim_tokens`, `sudo_sessions`). That breaches the "modules must remain independent / no cross-module coupling" rule in root `AGENTS.md`, and more importantly it **fails open**: any module that later adds a token or credential table is exported into the tenant archive in full until someone remembers to edit this list. Column-level `redactSensitiveColumns` only catches columns whose *name* looks sensitive, which will not save a table like `scim_tokens` whose column is `value`.

This PR introduces exactly the right mechanism for this — the module-owned privacy data-class registry in `packages/shared/src/lib/privacy/registry.ts`. The exporter should consume that (or an equivalent per-entity marker) instead of a central list.

## 🟡 Minors

**11. The audit redaction key matcher is suffix-only and misses real secret names**
`packages/shared/src/lib/commands/audit-redaction.ts:44-48`

`isSensitiveAuditKey` normalises the key then tests `endsWith` against the suffix list. Reproduced directly:

```js
redactSensitiveAuditData({ secretKey: 'abc', privateKeyPem: 'x', passwordConfirmation: 'y',
                           clientSecret: 'z', apiKey: 'k', mfaSecretValue: 'm', tokenValue: 't' })
// → { secretKey: 'abc', privateKeyPem: 'x', passwordConfirmation: 'y',
//     clientSecret: '[REDACTED]', apiKey: '[REDACTED]', mfaSecretValue: 'm', tokenValue: 't' }
```

So a command whose input carries `secretKey`, `privateKeyPem`, `mfaSecretValue` or `tokenValue` writes the raw value into `action_logs` — and, because `containsSensitiveAuditData` also drives the `sensitiveInput` decision in `command-bus.ts:300`, that execution *keeps* its undo token and redo payload. The same suffix regex is baked into `Migration20260820143000.ts:20`, so the backfill misses them too.

Note this PR ships a *second*, stronger detector for the same concern: `isSensitiveTenantExportColumn` (`tenantExitExportService.ts:200-209`) uses `includes` and would catch `secret_key` / `private_key`. Please consolidate on one shared detector and use substring (or explicit token) matching.

**12. `logCrudAccess`'s new `statusCode` option is never supplied, so `result` is hard-wired to `"success"`**
`packages/shared/src/lib/crud/factory.ts:889`, `:911`, `:917-918`

`LogCrudAccessOptions.statusCode` was added and defaults to 200, and `result` is derived as `statusCode >= 400 ? 'failure' : 'success'`. None of the three call sites in `factory.ts` (`:1727`, `:1906`, `:2145`) nor any of the ~10 module call sites pass it. The `"failure"` value the spec promises (`.ai/specs/2026-08-21-access-log-durability-retention.md:95`) is therefore unreachable, and `statusCode` is a constant. Either wire the real status through or drop the field rather than shipping a column that always says "success".

**13. A transient audit-log write failure now returns 500 for a successful read**
`packages/shared/src/lib/crud/factory.ts:963-972`

The write promise lost its internal `try/catch`. In blocking mode — now the default — `await writePromise` propagates, and the three `await logCrudAccess(...)` call sites are not wrapped, so a DB hiccup while inserting `access_logs` turns a perfectly good CRUD `GET` into an unhandled 500. The spec chooses this deliberately ("An audit record must not be silently lost after a successful response") and there is a test for it, so I am not calling it wrong — but the availability trade-off is undocumented for operators, there is no degradation path, and `.env.example` does not explain when to reach for `OM_CRUD_ACCESS_LOG_MODE=async`. Please confirm the choice explicitly and document it.

**14. The audit redaction backfill is a single unbatched full-table UPDATE with an empty `down()`**
`packages/core/src/modules/audit_logs/migrations/Migration20260820143000.ts:46-88`

One `UPDATE action_logs … FROM affected` over every non-deleted row, calling a recursive PL/pgSQL function per JSONB column. On a production audit table this can hold locks for a long time and bloat WAL. Consider a batched/`LIMIT`-driven loop or a documented maintenance window. `down()` is empty — acceptable for a redaction, but it should say so in a comment.

**15. New operator env vars are missing from `.env.example`**

Introduced by this PR and absent from both `apps/mercato/.env.example` and `packages/create-app/template/.env.example`: `TENANT_DATA_ENCRYPTION_REQUIRED`, `OM_ENVIRONMENT_CLASSIFICATION`, `OM_AI_RUNTIME_SECURITY_PROFILE`, `OM_ATTACHMENT_QUARANTINE_DIR`, `OM_ATTACHMENT_SCAN_TIMEOUT_MS`.

`OM_ENVIRONMENT_CLASSIFICATION` is the worst of these: it appears nowhere outside code, tests and one enterprise spec line, yet `requireNonProductionEnvironment` (`packages/enterprise/src/modules/data_erasure/services/environmentClassification.ts:16-23`) fails closed with a 409 when it is unset. Every call to `POST /api/data_erasure/environment-sanitization` returns `ENVIRONMENT_CLASSIFICATION_REQUIRED` until an operator discovers the variable by reading the source. The other vars are correctly covered (`OM_CRUD_ACCESS_LOG_MODE`, `AUDIT_LOGS_RETENTION_DAYS`, `OM_AUDIT_EVIDENCE_HMAC_KEY`, `OM_ATTACHMENT_SCAN_POLICY`, `OM_BACKUP_*`, `OM_ENABLE_ENTERPRISE_MODULES_DATA_ERASURE`), so this is a small gap in otherwise good coverage.

**16. PR size makes this effectively unreviewable and unrevertable as a unit**

315 files across 15 numbered controls, three brand-new modules (`data_erasure`, `backups`, `tenant_exports`), four migrations, four workflow files, and changes to authentication, encryption, the command bus and the shared CRUD factory. The commit series is clean and each control is independently scoped — which is exactly why they would land far more safely as a handful of sequenced PRs (shared/command-bus + encryption first, then audit, then attachments/AI, then the enterprise modules, then the CI gates last, once the repo is clean for them). If splitting is off the table, please say so and I will review the rest as-is; but a single revert of this PR would roll back fifteen unrelated security controls at once.

## ⚪ Nits

17. `packages/shared/src/lib/crud/factory.ts:919` calls the global `crypto.randomUUID()` although `randomUUID` is already imported from `node:crypto` at line 80.
18. `packages/core/src/modules/configs/api/health/ready/route.ts:15-19` — `requireAuth: false` with no rate limit, building a full request container per call, and disclosing KMS/encryption health to anonymous callers. Standard for a readiness probe, but worth confirming it is not internet-exposed.
19. `requireNonProductionEnvironment` keys solely on `OM_ENVIRONMENT_CLASSIFICATION`. A `NODE_ENV === 'production'` cross-check would be cheap defence-in-depth against a mis-set classification on a production host.
20. `.github/workflows/dast.yml:31-43` — the loopback guard rejects `localhost` and `127.*` but not `0.0.0.0`, RFC1918 ranges, or `169.254.169.254`. Low risk (manual dispatch, `contents: read`, passive baseline scan), but easy to tighten.
21. `.github/workflows/qa-deploy.yml:31` adds `artifact-metadata: write`. Please double-check that scope name is valid — an unrecognised key in `permissions:` invalidates the entire workflow file, and this workflow has not run on this branch.
22. The access-log CSV export now carries an attacker-controlled `userAgent`; `escapeCsv` (`packages/shared/src/lib/crud/exporters.ts:35`) quotes but does not neutralise a leading `=`/`+`/`-`/`@`. Pre-existing helper behaviour, newly reachable with untrusted input.
23. `stableScheduleUuid` (`packages/core/src/modules/audit_logs/setup.ts:18-21`) builds a UUID-shaped string from a SHA-256 without setting the RFC-4122 version/variant nibbles. Postgres accepts it; it is simply not a valid v4/v5 UUID.
24. `resolveAttachmentScanPolicy` defaults to `optional` and the default scanner is `UnavailableAttachmentScanner`, so out of the box the new gate always returns `scanner_unavailable` and admits the upload. Reasonable as an extension point, but it is a notable asymmetry against the fail-closed posture chosen for encryption — worth an explicit sentence in `apps/docs/docs/framework/attachments/security-scanning.mdx`.

## ✅ What is good

- The MFA login interceptor genuinely closes a bypass: the old `catch { return {} }` silently issued a full staff token whenever `createChallenge` failed. The `hasChallengeMethods` pre-check also avoids challenging users with no enrolled factor.
- `signAudienceJwt`/`verifyAudienceJwt` for the pending-MFA token is the right call — an audience-derived key means the provisional token cannot be replayed against staff APIs even if a check is forgotten. Flipping the three MFA routes to `requireAuth: false` and validating the pending token in `resolveMfaRequestContext` is coherent, and each route still asserts `mfa_pending === true`.
- `login.ts` deleting the session on every non-success path, and clearing `session_token` when an interceptor swapped the token, is careful work — the comment explaining the `session/refresh` bypass is exactly the kind of note that keeps a fix alive.
- `evidenceExportService` is well built: sorted canonical JSON, hash chain with an explicit `previousHash`, minimum 32-byte signing key, and `timingSafeEqual` behind a hex-shape guard.
- `backupService` spawns `pg_dump`/`pg_restore`/`psql` with `shell: false` and reads secrets only from the environment.
- Tenant/organization scoping in the new `data_erasure` routes is consistent and correct (`resolvePrivacyApiContext` requires both `auth.tenantId` and a resolved organization, and every route declares `requireFeatures`).
- The attachment quarantine store creates `0700` directories and `0600` files with `flag: 'wx'`, and rolls back both artefacts on partial failure.
- Test coverage for the new surfaces is real and specific, not decorative — `production-policy.test.ts`, `log-crud-access.test.ts`, `mfa-pending-token.test.ts`, `evidenceExportService.test.ts`, plus integration specs `TC-PRIVACY-001`, `TC-AUD-009/010`, `TC-ATT-015`, `TC-SSO-007`.
- `yarn generate` leaves no drift, and the `create-app` template mirror was updated alongside `apps/mercato` for `.env.example` and `modules.ts`.

## 🧾 Verdict

**Changes requested.** Seven blockers (two failing gate commands attributable to this branch, no CI on the head, and four security gates that fail on the very PR introducing them), three majors (a silent login dead-end, missing upgrade documentation for breaking operational changes, and a fail-open cross-module denylist in the tenant export), and six Medium findings — which this repository's rule also treats as request-changes.

pat-lewczuk:
🏷️ label rationale

- `review` → `changes-requested`: the review verdict is request-changes — 7 blockers, 3 majors and 6 Medium findings (this repo requests changes on Medium too). Blockers include two validation-gate commands failing on this branch (`yarn test`, `yarn i18n:check-sync`), no CI run against head `68cab3da`, and four security gates that fail on the PR that introduces them.
- `needs-qa` kept (no `skip-qa`): unchanged from the author's own rationale — identity, privacy, backup, attachment-scanning and enterprise UI/API flows all need target-environment exercise. The QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- `priority-high` kept: release-blocking security-readiness scope touching auth, session and encryption.
- `risk-high` kept: auth, encryption, four database migrations, shared contract surfaces (`CommandLogMetadata`, `CommandRuntimeContext`, `logCrudAccess`, package exports) and broad cross-module edits — squarely in the risk-high inference bucket.
- `feature`, `enterprise`, `security` kept unchanged.
- No `qa` label applied: that pipeline state belongs to a QA reviewer, not to automation.

pat-lewczuk:
🤖 Handing this back to @WebEferen — changes requested.

The fastest path back to green, in order:

1. `yarn i18n:check-sync --fix` — restores sorted keys in `audit_logs` and `enterprise/sso` locales (blocker 2). This is also a non-`continue-on-error` step in the CI `lint` job.
2. `tenant_exports` — add an evaluation-catalog case listing `.ai/guides/modules/tenant_exports/index.md` in `context.required`, or stop enabling the module in the scaffold template (blocker 1; the assertion in `packages/create-app/src/lib/module-facts-build.test.ts` spells out both options).
3. Push to the head so CI actually runs — there are currently **no** check runs against `68cab3da` (blocker 3).
4. The four security gates this PR adds/extends all fail on this branch (blockers 4–7). CodeQL in particular cannot pass while the repository's CodeQL *default setup* is enabled, so that needs an admin decision sequenced before merge.
5. The login dead-end (major 8) is the one user-visible functional regression: with MFA unavailable the login form silently resets with no error, because an after-interceptor cannot set a status code and `login.tsx` only reports errors on non-2xx.

Autofix was skipped — this is not the automation's own PR. Re-run `/om-auto-review-pr 5557 --autofix` if you want the fixes applied here instead.

pat-lewczuk:

Validation source: local fallback (no GitHub check runs exist for head `68cab3da`). Runner: local. Gate result: `build:packages` / `generate` / `typecheck` / `build:app` PASS; `i18n:check-sync` FAIL; `test` FAIL (2 failures attributable to this branch, 5 host-environment). No autofix iterations — not the automation's own PR.

PR #5450, opened 2026-08-21
docs(specs): background-work reliability series v5 — problems, requirements, staged decision, phase-1 specs

jtomaszewski:
## Summary

Long-running background work in Open Mercato is unreliable in ways that keep resurfacing as separate incidents: a deploy kills a multi-day `data_sync` run, Cancel is advisory, runs stay `running` forever, a transient error ends days of work, two runs start on one cursor. This PR adds a specification series that (1) catalogues the problems in `data_sync`, (2) applies the same lens to `packages/queue`, `progress`, `workflows`, `scheduler` and every other background worker, (3) reduces ~260 findings to fourteen shared problem classes and a single set of requirements with acceptance criteria, (4) scores four candidate architectures and records a **staged** decision, and (5–6) specifies phase 1 as two independently reviewable implementation specs.

**v5 restructure.** The series originally specified a full leased-job tier in `progress` (former parts 6–8, reviewed through five rounds). A maintainer scope pass added the missing question — *is the adopting consumer idempotent under redelivery?* — and the answer for `data_sync`, the only phase-1 adopter, is yes: imports are upserts, the id-mapping self-heals duplicates, and the `commitBatchProgress` CAS already guarantees a single driver per run. Every duplicated-external-side-effect finding in the catalogue comes from consumers (webhooks, emails, notifications, pushes) that were **not** adopting. So the series is now staged: phase 1 hardens `data_sync` on the fence it already has — server-side reconciler, transactional start, DB-enforced single-runner, error taxonomy, cooperative shutdown — with one added column and no `progress_jobs` change; the leased tier stays the ratified direction, archived in full at `205fbd53f`, and is built when a recorded trigger fires (a second kind needing durable coordination, interruption telemetry showing batch-redo hurts, or a non-idempotent consumer adopting).

Problems and requirements are stated before any design so that maintainers can agree on the bar first; part 4 explicitly lists the decisions it leaves to you and is the document to push back on.

**Docs only — no code, schema, API or behaviour changes in this PR.**

## Changes

- `…-01-data-sync-problems.md` — problem catalogue D-1…D-28 with evidence; no solution. *(unchanged in v5 except the series header)*
- `…-02-sibling-modules-problems.md` — read-only audits of `packages/queue` (Q-n), `progress` (P-n), `workflows` (W-n), `scheduler` (S-n) and ~40 other workers across 20 modules, each with `file:line` evidence. *(unchanged in v5 except the series header)*
[...]

PR #5562, opened 2026-08-24
feat: integrate PR 5450 reliability implementation

yokoszn:
# PR 5450 integration candidate

This upstream PR carries the review/spec and implementation work prepared against upstream PR #5450. It is based on PR #5450 head `60e0bac3b7` and should be reviewed as a stacked integration candidate, not as an unrelated replacement.

## Included

- Background-work complexity and verification specifications from the fork review lane.
- Durable progress repair cells with tenant/org scoping.
- Transactional repair-cell claims and caller-owned `withClaimedRepairCells(...)` work.
- Lease tokens, expiry recovery, attempts, and monotonic `leaseEpoch` stale-write fencing.
- Data-sync worker repair-cell record/clear wiring.
- Local queue watcher recovery and bounded fallback behavior.
- Reversible repair-cell migrations.

## Validation

- Core typecheck passed.
- Core build passed.
- Focused repair-cell tests: `3/3`.
- Local queue strategy tests: `35/35`.
- Full queue package: `104/104` with `--forceExit`.
- `git diff --check` passed.

## Explicit limitations

The full queue assertions pass, but ordinary Jest shutdown still reports an open-handle condition after the suite; this is not presented as a clean process-exit result. Live Postgres concurrent-claim/crash-recovery integration, real Valkey worker-loss testing, workflow latency/RSS benchmarks, and the complete upstream validation sequence remain open review gates.

Fork integration PR: https://github.com/TWN-Systems/open-mercato/pull/10
Original upstream design PR: https://github.com/open-mercato/open-mercato/pull/5450

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5562"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790164813&installation_model_id=432243&pr_number=5562&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5562&signature=fa69ca8947058d4b502ba8da27ad848e2ee55315f335a461fa29d958eae100a5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-25T00:52:28Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review

**PR:** #5562 — `feat: integrate PR 5450 reliability implementation`
**Head:** `d8e6770` · **Base:** `develop` · **Fork:** `TWN-Systems/open-mercato`
**Runner:** local (no compose `app` container running) · **Validation source:** local fallback — GitHub reports no usable check data for this head (see 🚦 CI below)

## 🎯 Summary

This PR stacks two very different things into one change: ~2,100 lines of new design specification (the `background-work` series, parts 1–8, plus a local-queue watch-recovery spec) and a first slice of implementation (a `progress_job_repair_cells` table with lease/epoch fencing, `data_sync` worker wiring, and local-queue watcher recovery).

The specification work is genuinely strong — part 3's requirement derivation and part 6's lease/terminal-transition protocol are the most rigorous background-work design this repo has. The implementation attached to it is not ready, for three independent reasons:

1. **The validation gate fails.** `yarn test` is red on a test introduced by this PR — it mocks a method the code under test never calls.
2. **The migration snapshot does not describe the schema the migrations produce.** `lease_epoch` is missing, which will break the next `yarn db:generate` for anyone who has run this commit.
3. **The implementation is not the design in the specs it ships, and those specs are explicitly undecided.** Part 6 specifies leased columns *on `progress_jobs`* plus a reconciler; what ships is a standalone side table with no consumer at all. Part 4's Status line still reads "Awaiting the maintainer decisions", and decision #1 is precisely *where the record lives*.

On top of that, the repair mechanism as wired does not fire on the failure path it was built for: `runImport`/`runExport` swallow ordinary adapter errors and return normally, so the worker's success branch runs and **deletes** any repair cell instead of recording one.

### Verdict: ❌ **Request changes**

Two blockers and nine majors. The verdict would follow from the failing gate alone.

## 🧪 Validation Gate

Runner: **local**. Commands from `.ai/agentic.config.json` → `validation.commands`, run in order.

| # | Command | Result |
|---|---------|--------|
| 1 | `yarn build:packages` | ✅ pass (exit 0) |
| 2 | `yarn generate` | ✅ pass (exit 0) — no generated-file drift (`git status` clean afterwards) |
| 3 | `yarn build:packages` | ✅ pass (exit 0) |
| 4 | `yarn i18n:check-sync` | ✅ pass (exit 0) |
| 5 | `yarn i18n:check-usage` | ✅ pass (exit 0) |
| 6 | `yarn typecheck` | ✅ pass (exit 0) |
| 7 | `yarn test` | ❌ **fail (exit 1)** — `@open-mercato/core`: `Test Suites: 1 failed, 1 skipped, 1395 passed`; `Tests: 1 failed, 2 skipped, 11166 passed`. The single failing suite is `src/modules/data_sync/lib/__tests__/repair-cell.test.ts`, added by this PR. `@open-mercato/queue`: `12 suites / 104 tests passed`. |
| 8 | `yarn build:app` | ✅ pass (exit 0) |

This contradicts the PR description's Validation section, which reports "Focused repair-cell tests: `3/3`". The data-sync repair-cell suite has never passed — it throws before reaching its first assertion.

## 🚦 CI status

No GitHub check data is usable for head `d8e6770`. Both workflow runs for this SHA (`CI for Develop&Main` → run `32724761049`, `Mutation tests` → run `32724760943`) are in `action_required` — the standard fork-PR state where a maintainer must approve each push before CI executes. `gh pr checks` reports only the `[code]smith` skip. There are therefore **no failing checks** and **no pending checks** to report; there is simply no CI signal, which is why the local gate above was run as this review's evidence. A maintainer approving the workflow runs would give the real signal.

Separately, the Dokploy preview deployment is blocked because `yokoszn` has `read` permission on this repository — informational, not a review finding.

## 🚫 Blockers

### 1. `yarn test` fails — the new data-sync repair-cell test mocks a method the code never calls

`packages/core/src/modules/data_sync/lib/__tests__/repair-cell.test.ts:5`

The test double is `{ upsert: jest.fn() }`, and the assertion expects `em.upsert` to have been called. But `recordDataSyncRepairCell` delegates to `upsertRepairCell`, which issues raw SQL through `em.getConnection().execute(...)` (`packages/core/src/modules/progress/lib/repair-cells.ts:36`). The double has no `getConnection`, so the call throws before the assertion is ever evaluated:

```
● data-sync repair-cell integration › records a scoped repair cell for a failed delivery
  TypeError: em.getConnection is not a function
    at upsertRepairCell (src/modules/progress/lib/repair-cells.ts:36:12)
    at recordDataSyncRepairCell (src/modules/data_sync/lib/repair-cell.ts:26:25)
```

Reproduced locally and in the full gate run. **Fix:** mock `getConnection().execute` and assert the actual SQL + bound parameters, or mock the imported `upsertRepairCell` helper and assert the call it receives. The second test in the same file (`clearDataSyncRepairCell` → `em.nativeDelete`) is correct and passes. *(Raised first by @chatgpt-codex-connector — [inline comment](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305934); confirmed here by execution.)*

### 2. The migration snapshot is missing `lease_epoch`

`packages/core/src/modules/progress/migrations/.snapshot-open-mercato.json`

`Migration20260824100000_progress.ts:8` adds `lease_epoch int not null default 0` and `ProgressJobRepairCell` declares the property (`packages/core/src/modules/progress/data/entities.ts:117`), but the rewritten snapshot's `progress_job_repair_cells` entry contains only:

```
attempts, cell, created_at, due_at, job_id, lease_token, lease_until,
organization_id, reason, tenant_id, updated_at
```

(verified by normalising both revisions of the file with `jq` and diffing — the *only* structural change between base and head is this new table, so there is no unrelated schema churn, which is good; the column is simply absent.)

The consequence is the one `packages/core/AGENTS.md` § *Entity Schema And Migration Workflow* step 5 exists to prevent: the next `yarn db:generate` diffs the entity against a snapshot that lacks the column, emits a second `add column lease_epoch`, and that migration then fails on every database that already ran this commit. **Fix:** update the snapshot to the complete post-change schema in the same commit as the migration. *(Raised by @chatgpt-codex-connector — [inline comment](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305919); confirmed here.)*

## ⚠️ Majors

### 3. The implementation is not the design in the specs it ships, and the design is still awaiting maintainer decisions

Part 6 (`.ai/specs/2026-08-21-background-work-06-leased-jobs-in-progress.md:20-39`) specifies the leased tier as **additive columns on `progress_jobs`** — `lease_owner`, `lease_epoch bigint`, `lease_expires_at`, `lock_key`, `queue_job_id`, `continuation_seq`, `redrives`, `pending_since`, `next_run_at`, `domain_mirrored_at` — plus `progress-reconcile` / `progress-retention` workers, a kind registry, `runSlice`, and the terminal-transition protocol of §7.

What ships is a standalone `progress_job_repair_cells` table carrying its own `lease_token` / `lease_epoch` / `attempts`, with none of the surrounding machinery. Part 6 §6 does mention a repair-cell side table, but only as a **conditional fallback** — named `progress_job_leases`, and gated on the part-4 benchmark ("100 ms p95 at 10k live rows … If it misses, the implementation must add the narrow `progress_job_leases` repair-cell side table before shipping"). The PR description states that benchmark was not run: *"workflow latency/RSS benchmarks … remain open review gates."*

Meanwhile part 4's own Status line reads **"Draft v4 — … Awaiting the maintainer decisions listed under 'Decisions requested from maintainers'"**, and decision #1 is *"A vs B: grow `progress` (recommended) or a new module as the home of the leased-job record"* — i.e. the question this schema silently answers a third way.

Root `AGENTS.md` § *Ask First* requires asking before architectural changes, and the repo's `om-code-review` extension requires flagging divergence from an active spec. Committing a table and three migrations under an undecided architecture is expensive to walk back. **Fix:** land the spec series on its own (see note 15), get the part-4 decisions recorded, then implement the design as specified — or amend part 6 to describe the mechanism actually being built and get *that* approved.

### 4. The failure path can leave sync runs and progress jobs stuck in `running`

`packages/core/src/modules/data_sync/workers/sync-import.ts:58` and the identical ordering in `sync-export.ts`

`recordDataSyncRepairCell(...)` is awaited *before* `syncRunService.markStatus(run.id, 'failed', ...)` and `progressService.failJob(...)`, inside the same inner `try`. If the repair-cell insert throws — a rolling deployment where the code is live before the table exists, a transient DB error, or the shared `EntityManager` already sitting in an aborted transaction from the original failure — control jumps to `catch (finalizeError)` at line 73 and **both terminal-state updates are skipped**. Once queue attempts are exhausted, the `sync_runs` row and the progress job stay `running` until the 60 s stale sweep, and if the sweep is not running they stay there indefinitely.

This is exactly the stuck-running class the spec series was written to eliminate (part 1, D-n). New, best-effort bookkeeping must never be able to prevent pre-existing terminal finalization. **Fix:** move the `recordDataSyncRepairCell` call after the `markStatus`/`failJob` block, or wrap it in its own `try { … } catch { logger.warn(…) }`. *(Raised by @chatgpt-codex-connector — [inline comment](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305915); confirmed by reading the handler.)*

### 5. The common failure path records nothing — and the success path deletes existing repair cells

`packages/core/src/modules/data_sync/workers/sync-import.ts:44-48`

`SyncEngine.runImport` catches ordinary adapter and batch-processing errors, writes an integration log, calls `finalizeRun(run.id, 'failed', …)` and then **`return`s normally** — it does not rethrow (`packages/core/src/modules/data_sync/lib/sync-engine.ts:705-735`). `runExport` behaves the same way.

So for the overwhelmingly common failure — upstream 500, adapter timeout, batch write error — the worker's `catch` never runs, `recordDataSyncRepairCell` is never called, and instead the success branch executes `clearDataSyncRepairCell(em, run?.progressJobId, …)`, which **deletes** any repair cell a previous delivery left behind. The worker's `catch` only fires for the few errors that escape `runImport` before its own try block (`No import adapter registered for provider …`, `Integration … is missing credentials`).

The net effect is that the mechanism records cells almost never and destroys them on the path where a real failure was just recorded as `failed`. **Fix:** inspect the returned run status before clearing (only clear when the run actually completed), or propagate the engine failure to the handler. *(Raised by @chatgpt-codex-connector — [inline comment](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305932); confirmed against `sync-engine.ts`.)*

### 6. Nothing ever claims a repair cell — the mechanism is inert and the table grows without bound

`packages/core/src/modules/progress/lib/repair-cells.ts:87,100,114,127`

A repo-wide search for `claimDueRepairCells`, `withClaimedRepairCells`, `acknowledgeRepairCell` and `releaseRepairCell` finds references only in `packages/core/src/modules/progress/__tests__/repair-cells.test.ts`. There is no worker, no scheduler entry, no API route, and no operator surface that claims, acknowledges, releases or reaps these rows — part 6's `progress-reconcile` and `progress-retention` workers (§6) are not part of this PR.

Concretely: a cell written by the failure path is never redriven and never repaired; it survives until the *same* progress job later succeeds and clears it, which for a terminally failed run never happens. Rows accumulate permanently, with no retention path. **Fix:** either ship a consumer in this PR (a reconciler worker, even a minimal one, with the retention sweep), or hold the schema until the consumer is ready — a table with no reader is cost without benefit. *(Raised by @chatgpt-codex-connector — [inline comment](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305911); confirmed by repo-wide search.)*

### 7. The `lease_epoch` upsert guard silently discards every update after the first claim

`packages/core/src/modules/progress/lib/repair-cells.ts:36-50`

`upsertRepairCell` guards its `ON CONFLICT DO UPDATE` with `where excluded."lease_epoch" >= "progress_job_repair_cells"."lease_epoch"`, and `leaseEpoch` defaults to `0` when the caller omits it. `recordDataSyncRepairCell` (`packages/core/src/modules/data_sync/lib/repair-cell.ts:26-33`) never passes one. Meanwhile `claimDueRepairCellsInTransaction` increments the stored epoch on every claim (`repair-cells.ts:80`).

So once a cell has been claimed even once, its stored epoch is ≥ 1, every subsequent `recordDataSyncRepairCell` evaluates `0 >= 1` → false, and the new `dueAt`/`reason`/`cell` are dropped. The function returns `void` with no indication that nothing was written, so the caller cannot detect it. The stale-write fence is correct in spirit, but the writer that is supposed to be fenced has no way to supply a current epoch. **Fix:** have `upsertRepairCell` read-and-carry the current epoch (or return the affected row count so callers can react), and thread a real epoch through `recordDataSyncRepairCell`.

### 8. Repair-cell leases use the caller's wall clock, contradicting the invariant the series fixes

`packages/core/src/modules/progress/lib/repair-cells.ts:56-85`

Both the eligibility predicate (`dueAt: { $lte: now }`, `leaseUntil: { $lte: now }`) and `leaseUntil = new Date(now.getTime() + leaseMs)` derive from the caller-supplied `now`. Part 4 § *Research — A. Record & clock* states the invariant the whole series rests on: *"Liveness is a short renewable lease on the **broker/DB clock**"*, with R-A1–A3 requiring DB time in every predicate; part 6's SQL uses `now()` throughout.

With clock skew between worker hosts larger than `leaseMs` (default 30 s — a very reachable skew on unsynchronised containers), a fast host treats a slow host's live lease as expired and claims the same cell concurrently. The `leaseToken` fences *acknowledgement* and *release*, but not the domain work performed after `claimDueRepairCells` commits, so both claimants can perform the repair. **Fix:** compute and compare lease timestamps in PostgreSQL (`now()`, `now() + $leaseMs * interval '1 ms'`) so every replica shares one clock. *(Raised by @chatgpt-codex-connector — [inline comment](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305943).)*

### 9. The queue directory probe is an unconditional 100 ms loop — the spec in this same PR forbids it

`packages/queue/src/strategies/local.ts:670-683`

`processContinuous` installs `setInterval(() => fsp.stat(queueDir)…, 100)` unconditionally and only clears it in `close()`. Every continuous local worker therefore performs ten directory `stat` syscalls per second for its entire lifetime, healthy watcher or not, idle queue or not — multiplied by every queue in the process.

`.ai/specs/2026-08-23-local-queue-filesystem-watch-recovery.md` § *Required design*, added by this PR, says:

> The implementation MUST NOT add an unconditional high-frequency queue-file read loop. Recovery probes may be enabled only while the watched directory is absent or the watcher is unavailable, and MUST stop once the watcher is re-established or the bounded recovery window expires.

and the verification matrix requires *"Idle worker → no queue-file reads before the configured idle interval"*. The `unref()` keeps the probe from holding the event loop open, but does not make it conditional. **Fix:** start the probe only when the watcher is unavailable or the directory is missing (`queueDirectoryState === -1`, or on watcher setup/error failure), and clear it once `refreshQueueWatcher` succeeds or the bounded recovery window expires. *(Raised by @chatgpt-codex-connector — [inline comment](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305923).)*

### 10. The bounded watcher-retry budget can never be exhausted

`packages/queue/src/strategies/local.ts:594-601` and `:643`

`scheduleWatcherRetry` bails at `watcherRetryAttempts >= 3`, but `refreshQueueWatcher` resets `watcherRetryAttempts = 0` as soon as `fs.watch` *returns* — before the watcher has demonstrated it is healthy. A watcher that constructs successfully and then emits an asynchronous `error` (the `ENOSPC` / inotify-quota case the spec calls out by name) therefore runs: construct → reset to 0 → error → schedule retry (attempts 1) → 100 ms → construct → reset to 0 → … indefinitely, at 10 Hz, logging an error each cycle.

That is exactly the retry storm the spec's failure model rules out (*"Retries therefore MUST be bounded"*; matrix row *"Watch emits `ENOSPC` → no retry storm … bounded retries then safety polling"*). **Fix:** reset the budget only after the watcher has been stable for an interval, or after it has successfully delivered an event — not at construction time. *(Raised by @chatgpt-codex-connector — [inline comment](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305947).)*

### 11. The novel SQL and the concurrency semantics have no real test coverage

`packages/core/src/modules/progress/__tests__/repair-cells.test.ts`

The three tests mock the `EntityManager` so completely that none of the behaviour worth testing is exercised:

- the `upsertRepairCell` case (line 14) passes a `getConnection()` returning a `jest.fn()` whose arguments are never asserted, so the `ON CONFLICT … WHERE excluded.lease_epoch >= …` guard — the single most novel and most bug-prone statement in the PR, and the subject of finding 7 — is completely untested;
- the claim case (line 26) stubs `em.transactional` as `work(tx)` and `tx.find` as a fixed array, so `LockMode.PESSIMISTIC_PARTIAL_WRITE` (`FOR UPDATE … SKIP LOCKED`), the ordering, the limit and the actual concurrent-claim semantics are never executed;
- `releaseRepairCell` / `acknowledgeRepairCell` assert only that a `leaseToken` appears in the filter object.

Root `AGENTS.md` and the review checklist require integration coverage for concurrency and migration-backed paths, and the PR itself lists *"Live Postgres concurrent-claim/crash-recovery integration"* as an open gate. Given the whole point of this code is correctness under concurrent workers, mock-only coverage is not sufficient to merge. **Fix:** add a Postgres-backed integration test that runs two concurrent claimers against real rows and asserts the disjoint-claim, epoch-fencing and acknowledgement behaviour.

## 📋 Minors

12. **`lease_epoch` is `int`, the spec says `bigint`.** `packages/core/src/modules/progress/data/entities.ts:117` and `Migration20260824100000_progress.ts:8` use `int`; part 6 §schema specifies `lease_epoch bigint not null default 0`. A monotonically incrementing counter is the classic case for `bigint`; widening it later is another migration.

13. **Repair-cell writes reuse the worker's shared `EntityManager` instead of a fork.** `sync-import.ts:45,55` resolve `em` directly. The sibling workers in this repo fork (`communication_channels/workers/poll-channel.ts:96`, `gmail-history-sync.ts:72`), and `packages/core/src/modules/progress/AGENTS.md:97` documents the same precedent for heartbeats specifically so a write is safe while the producer's own EM is mid-transaction. In the `catch` path the shared EM may already be in an aborted transaction — which is one of the ways finding 4 gets triggered.

14. **The upsert rewrites tenant scope with no scope predicate.** `repair-cells.ts:40-49` conflicts on `("job_id")` alone and then sets `"tenant_id" = excluded."tenant_id"`, `"organization_id" = excluded."organization_id"`. Every other path in the file goes through `scopedFilter`. Practically this is protected only by job ids being unguessable UUIDs, not by a scope filter, which is weaker than the module's own convention (root `AGENTS.md` § Architecture: *"Always filter by `organization_id` for tenant-scoped entities"*). Consider a composite key or a tenant predicate in the conflict clause.

15. **The PR re-includes another author's open PR.** Commits `60e0bac3`, `56d9b729`, `f403beec`, `9a5d9550`, `2c5f0a0f` are the spec series from #5450 (@jtomaszewski), which is **still open** against `develop`. Nothing here is already in `develop`, so this is not duplicate-of-merged work, and the PR description is transparent about the stacking. But merging #5562 would land #5450's content under a different PR and bypass that PR's own review thread. Please land #5450 first and rebase this one onto it, or close #5450 explicitly with the authors' agreement.

16. **Two extra database round-trips on every successful sync run.** `sync-import.ts:46-48` adds a `getRun` plus a `nativeDelete` to the success path of every import and export, even though a repair cell exists only in the rare failure case. A cheap guard (skip when the run has no recorded failure) or folding the delete into the existing finalization would avoid paying for it on the hot path.

17. **The queue recovery test now depends on real wall-clock timing.** `packages/queue/src/__tests__/local.strategy.test.ts:730` removes `jest.useFakeTimers()` and adds a 10 s timeout, so the assertion now races the 100 ms probe in real time. It passes locally (`104/104`), but real-timer tests of filesystem recovery are the classic source of CI flake on loaded runners. Consider asserting recovery via an injected clock or an event, not elapsed time.

18. **Three migrations for one new table in a single PR.** `Migration20260823084507` creates it, `Migration20260823085716` adds `lease_token`/`lease_until`, `Migration20260824100000` adds `lease_epoch`. All three are reversible and correct in isolation, but since none has been released, squashing them into one `create table` would be cleaner for everyone applying this later — and would remove the class of mistake behind blocker 2.

19. **The PR's own acceptance criteria are not met.** `.ai/specs/2026-08-23-local-queue-filesystem-watch-recovery.md` § *Acceptance criteria* requires *"The full queue package suite MUST pass without an open-handle timeout or unbounded watcher retry log"* and that the change *"MUST report the observed event latency and fallback latency separately"*. The PR description states the open-handle condition still occurs, and no latency measurements are reported.

## 🔎 Nits

20. **The Validation section of the PR description is inaccurate.** It reports "Focused repair-cell tests: `3/3`" for a suite that has never passed. Please re-run and update it — a description that overstates verification costs reviewers more time than an honest "not yet verified".

## 📥 Inherited review feedback

All eight Codex findings already on this PR were re-verified against head `d8e6770` and **every one of them holds**. None had been addressed. They are carried into the findings above rather than repeated:

| Codex finding | File | Status here |
|---|---|---|
| P1 — Mock the SQL path used by `recordDataSyncRepairCell` | [`repair-cell.test.ts:5`](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305934) | ✅ confirmed by execution → **blocker 1** |
| P1 — Add `lease_epoch` to the migration snapshot | [`.snapshot-open-mercato.json:96`](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305919) | ✅ confirmed by snapshot diff → **blocker 2** |
| P1 — Finalize failed jobs even when repair recording fails | [`sync-import.ts:58`](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305915) | ✅ confirmed by reading the handler → **major 4** |
| P1 — Preserve repair cells when the engine reports failure | [`sync-import.ts:48`](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305932) | ✅ confirmed against `sync-engine.ts:705` → **major 5** |
| P1 — Add a runtime repair-cell consumer | [`repair-cells.ts:87`](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305911) | ✅ confirmed by repo-wide search → **major 6** |
| P1 — Use the database clock for repair-cell leases | [`repair-cells.ts:71`](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305943) | ✅ confirmed; also contradicts part 4 R-A1–A3 → **major 8** |
| P2 — Stop the directory probe while the watcher is healthy | [`local.ts:673`](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305923) | ✅ confirmed; violates this PR's own spec → **major 9** |
| P2 — Preserve the watcher retry budget across immediate errors | [`local.ts:643`](https://github.com/open-mercato/open-mercato/pull/5562#discussion_r3843305947) | ✅ confirmed by tracing the reset → **major 10** |

## 💥 Breaking Changes

- [x] **Exported APIs / import paths** — `packages/core/src/modules/progress/lib/repair-cells.ts` and `data_sync/lib/repair-cell.ts` are new files exporting new symbols. Purely additive; nothing existing is renamed, removed or retyped.
- [x] **HTTP routes and response shapes** — none touched.
- [x] **Event IDs / widget spot IDs / notification IDs / DI keys / ACL features / CLI commands** — none touched.
- [x] **DB schema** — one new table (`progress_job_repair_cells`) with its own migrations. Additive; no existing table is altered. The three migrations are individually reversible (`down()` drops the columns/table).
- [ ] **Generated files / snapshots** — ❌ `.snapshot-open-mercato.json` does not match the post-migration schema (blocker 2). `yarn generate` itself produces no drift (verified: clean `git status` after step 2 of the gate).
- [x] **Config formats** — none touched.

No `BACKWARD_COMPATIBILITY.md` protected surface is broken by this change.

## 🧪 Test Coverage

| Changed behaviour | Coverage |
|---|---|
| `upsertRepairCell` conflict/epoch guard | ❌ none — the mock never asserts the SQL (major 11) |
| `removeRepairCell` | ⚠️ mock-level only |
| `claimDueRepairCells` argument validation | ✅ covered |
| `claimDueRepairCells` concurrent-claim / `SKIP LOCKED` semantics | ❌ none — `em.transactional` is stubbed out (major 11) |
| `withClaimedRepairCells` | ❌ none |
| `acknowledgeRepairCell` / `releaseRepairCell` token fencing | ⚠️ mock-level only; asserts the filter object, not the SQL effect |
| `recordDataSyncRepairCell` | ❌ the test throws (blocker 1) |
| `clearDataSyncRepairCell` | ✅ covered at mock level |
| Worker failure-path ordering (majors 4, 5) | ❌ none — no test exercises `sync-import`/`sync-export` handlers with a failing or internally-finalized engine |
| Queue directory recreation recovery | ✅ covered (`local.strategy.test.ts`, real timers — see minor 17) |
| Watcher `ENOSPC` / retry bounding (major 10) | ❌ none |

**Tests to add before merge:** a Postgres-backed integration test for concurrent claims and epoch fencing; unit tests for both `sync-import` and `sync-export` handlers covering (a) engine throws → cell recorded *and* run finalized `failed`, (b) engine finalizes internally → cell preserved, (c) repair-cell write throws → run still finalized `failed`; and a bounded-retry test for the watcher error path.

---

## 🗒️ Note on the specification work

Nothing above should be read as a criticism of parts 1–8. The problem catalogues, the requirement derivation, the cross-system research, and part 6's terminal-transition protocol are excellent and worth landing on their own merits — via #5450, with the part-4 decisions actually taken. The recommendation is simply to separate the two: land the design, get the decisions recorded, and then build the implementation the approved design describes. The implementation in this PR reads as a partial sketch of a different mechanism, and reviewing it against a spec it does not follow serves neither.

*Reviewed by `om-auto-review-pr` → `om-code-review`. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).*

pkarw:

| Label | Why |
|---|---|
| `changes-requested` | Pipeline state. Review submitted as `CHANGES_REQUESTED`: `yarn test` fails on a test added by this PR, and the migration snapshot is missing `lease_epoch`. Two blockers plus nine majors. |
| `feature` | Adds a new entity (`ProgressJobRepairCell`), a new table with three migrations, and new `progress` / `data_sync` library surface. |
| `documentation` | Ten of the twenty-three changed files are new specification documents (`.ai/specs/2026-08-21-background-work-01…08`, `2026-08-23-local-queue-filesystem-watch-recovery`, plus the specs `README.md` index row). |
| `priority-medium` | Foundational reliability work with real value, but nothing here is on fire: the code it adds is inert (no consumer claims a repair cell), so no user-visible defect is currently waiting on this merge. |
| `risk-high` | Blast radius, not urgency. `packages/queue/src/strategies/local.ts` changes the wake-up path of **every** continuous local worker in the product, and the change adds an unconditional 100 ms `stat` loop per worker. Combined with a new DB table plus three migrations and a rewritten migration snapshot that does not match the schema those migrations produce, a bad merge here is felt process-wide and needs a migration to undo. |
| `needs-qa` | The automated-verification exemption does not apply: it requires the database structure to be unchanged, and this PR adds `progress_job_repair_cells` with three migrations. Manual QA is requested; it is not `qa` (that label belongs to a QA reviewer once QA is actually in progress) and it is not `qa-approved` (that is earned by a real QA pass, never from reading a diff). |

No `skip-qa`, no `qa-approved`, no `qa` — the QA-approval merge gate applies to this PR. `in-progress` is released in the next comment.

pkarw:
🤖 `om-auto-review-pr` — handing this PR back to @yokoszn for changes.

**Two blockers** (each independently blocks merge):

1. `yarn test` is red. `packages/core/src/modules/data_sync/lib/__tests__/repair-cell.test.ts:5` mocks `em.upsert`, but `recordDataSyncRepairCell` → `upsertRepairCell` calls `em.getConnection().execute(...)` — `TypeError: em.getConnection is not a function`, thrown before the assertion. Core suite: `1 failed, 1395 passed`.
2. `packages/core/src/modules/progress/migrations/.snapshot-open-mercato.json` does not contain `lease_epoch`, though `Migration20260824100000_progress.ts` adds the column. The next `yarn db:generate` will re-emit `add column lease_epoch` and that migration will fail on any database that ran this commit.

**Nine majors**, in the order I would tackle them: the failure path can leave runs stuck `running` (repair-cell write ordered before `markStatus`/`failJob`); the success path deletes repair cells for the common failure, because `runImport`/`runExport` finalize internally and return rather than throw; nothing in the product claims a repair cell, so the table is inert and unbounded; the `lease_epoch >=` upsert guard silently drops every write after the first claim; leases use the caller's wall clock instead of the DB clock; the new 100 ms directory probe is unconditional, which `.ai/specs/2026-08-23-local-queue-filesystem-watch-recovery.md` explicitly forbids in this same PR; the watcher retry budget resets at construction so it can never be exhausted; the novel SQL and the concurrency semantics have no real coverage; and the implementation does not match part 6 of the spec series it ships — which is still `Draft v4 — Awaiting the maintainer decisions`, with "where the record lives" listed as open decision #1.

All eight prior findings from @chatgpt-codex-connector were re-verified against `d8e6770` and every one still holds; none had been addressed. They are folded into the review body with links rather than repeated.

**The one thing worth deciding before writing any code:** #5450 (@jtomaszewski) is still open and this PR re-includes its five spec commits. The specs are good work and deserve to land on their own thread with the part-4 decisions actually recorded. I would suggest landing #5450 first, taking the A-vs-B decision, and then rebasing an implementation that follows the approved design. Reviewing this implementation against a spec it does not follow is not serving either half of the PR.

Full findings, the validation-gate table and the breaking-change checklist are in [the review](https://github.com/open-mercato/open-mercato/pull/5562#pullrequestreview-5014007171).

Note on CI: both workflow runs for this head are `action_required` (fork PR awaiting maintainer approval), so there is no GitHub check signal at all. The validation gate above was run locally as this review's evidence — it does not substitute for branch protection. A maintainer approving the runs would give the real signal.

pkarw:

2 blockers, 9 majors, 8 minors, 1 nit. Validation gate run locally (no GitHub check data for this head — fork runs are `action_required`): 7 of 8 commands pass, `yarn test` fails on a test added by this PR. All 8 inherited Codex findings re-verified and confirmed; none had been addressed. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here). No CI follow-up will run, since CI has not been approved to start for this head.

Kapsik89:
## QA: FAIL (2026-09-24, `d8e67707a` merged locally with develop `2e95af80a`, prod build)

The branch is 479 behind and conflicts only in `.ai/specs/README.md`. I kept develop's version of that file, so the diff against develop is exactly this PR's code. I dry-ran the 3 `progress` migrations in a transaction first, then applied them. I ran one failing import on purpose before migrating.

The migrations, the happy path and the queue changes work. The two review blockers are still on this head, and the new write in the worker's error path can strand a run.

| # | Scenario | Result |
|---|---|---|
| 1 | `repair-cell.test.ts` | ❌ 1/5, `em.getConnection is not a function` (review blocker 1) |
| 2 | `yarn db:generate` after the PR's migrations emits a second `add "lease_epoch"`; running it gives `column "lease_epoch" ... already exists` | ❌ review blocker 2 |
| 3 | Migrations up (table, 2 indexes, 12 columns) and down in reverse → 0 tables | ✅ |
| 4 | Failing run (sync_akeneo, no credentials) while the repair-cell insert throws → run and progress job stay `pending`, job dropped after 3 attempts, next run for the same entity 409 | ❌ #6426 |
| 5 | Same failing run with the table present → `failed`, cell recorded with the run's tenant/org | ✅ |
| 6 | Adapter fails inside the stream ("Akeneo URL must use https") → run `failed`, no cell | ❌ the review's "success path clears the cell" point, confirmed |
| 7 | sync_excel CSV import from the UI → completed, 2 created, 0 cells | ✅ |
| 8 | Local queue delivery ~320–360 ms, and ~340 ms after deleting the `data-sync-import` queue dir | ✅ no regression |
| 9 | admin import in the UI completes; employee gets "Access denied" on data-sync and 403 on the run API | ✅ |

Unit gate: queue `local.strategy` 35/35.

Two smaller things: `progress_job_repair_cells.job_id` has no FK to `progress_jobs`, and nothing ever reads the cells, so they pile up. Also, the new 100 ms directory probe runs per local worker. This app registers 63 local queues, so that's about 630 `stat` calls a second at idle (from the code, not measured).

| Finding | Severity | Filed |
|---|---|---|
| Stranded run when the repair-cell write throws (import measured, export has the same order in code) | Medium | #6426 |
| Red unit test, snapshot missing `lease_epoch`, cell cleared on in-stream failures | Blocker / Major | Not filed. Already review blockers 1–2 and a review major; confirmed above |

Screenshots of the stranded run and the probe scripts are local and available on request.

PR #5138, opened 2026-08-09
feat(security): raise npmMinimalAgeGate from 1d to 5d

zawoj:
## Summary

Raises Yarn's built-in `npmMinimalAgeGate` from the default `1d` to `5d`, closing the supply-chain window where a compromised maintainer account publishes a malicious patch before any advisory exists.

- Applies to **all install paths** — developer `yarn add`, CI, Dependabot — not just automated PRs
- First-party `@open-mercato/*` packages remain preapproved via `npmPreapprovedPackages` for immediate consumption after publishing
- Synced to the `create-app` template so scaffolded apps get the same protection instead of silently keeping the `1d` default
- No custom CI workflow is introduced; the Yarn-level gate covers every install path with fail-closed behaviour

### Why 5 days

The first draft proposed `14d`. Review pushed back that a two-week gate delays legitimate security patches and pushes every urgent upgrade through a standing `npmPreapprovedPackages` entry somebody then has to remember to remove — a worse failure mode than the one being fixed — and put the reasonable range at **3–7 days**.

`5d` sits inside that range. It covers every npm supply-chain incident on record with margin (detection times ran from ~5 hours to ~2 days), while staying short enough that an urgent third-party patch is reachable within a work week. For the rare patch that cannot wait, the escape hatch is per-call rather than persistent: `yarn add pkg@version --no-time-gate` — explicit, visible in the diff, and it leaves no standing exception behind.

The full rationale, the incident table and the rejected alternatives are recorded in the brief that ships with this PR.

### Changes

- `.yarnrc.yml` — added `npmMinimalAgeGate: 5d`
- `packages/create-app/template/.yarnrc.yml.template` — same value, updated comment
- `scripts/__tests__/npm-minimal-age-gate.test.mjs` — asserts the monorepo declares the agreed duration and that the template declares the same one, so the two files cannot drift apart unnoticed
- `.ai/specs/briefs/2026-08-09-dependabot-cooldown-ci-gate.md` — the brief, rewritten around `5d`

## Verification

| Check | Result |
|-------|--------|
| `yarn config get npmMinimalAgeGate` | `7200` minutes = 5 days |
| `yarn install --immutable` | passes — all lockfile resolutions install; already-locked versions are not re-gated |
| `yarn add is-odd@3.0.1` (published 2019) | installs |
| `yarn add @iobroker/repochecker@5.20.9` (published 2026-08-10) | blocked — `YN0016: … All versions satisfying "5.20.9" are quarantined` |
[...]

PR #5563, opened 2026-08-24
fix(security): align Dependabot cooldown with npmMinimalAgeGate (#5549)

### LINKED ISSUE #5549: Align the Dependabot cooldown with npmMinimalAgeGate so update PRs are installable
## Follow-up from #5138

#5138 raises Yarn's `npmMinimalAgeGate` from the default `1d` to `5d` in `.yarnrc.yml` and in `packages/create-app/template/.yarnrc.yml.template`, so no version published less than five days ago can enter the dependency graph through a new resolution.

`.github/dependabot.yml` declares no `cooldown`, so Dependabot is unaware of that window. It runs weekly on Monday and will keep proposing versions that were published within the previous five days. Those PRs cannot install: `yarn install` fails with `YN0016: … All versions satisfying "<range>" are quarantined`, and `scripts/ci/npm-retry-on-quarantine.sh` correctly fast-fails rather than retrying, because no amount of waiting inside a CI job clears a client-side age gate.

The failure resolves on its own once the version ages past the gate, but in the meantime every affected Dependabot PR sits red, and the fast-fail message points at `npmPreapprovedPackages` — a standing exception that somebody then has to remember to remove. That is exactly the maintenance trap the review of #5138 argued against.

Dependabot supports a native `cooldown` setting. Aligning it with the Yarn gate means Dependabot never opens a PR the gate would reject, and the two mechanisms stop contradicting each other.

### Acceptance criteria

- [ ] `.github/dependabot.yml` declares a `cooldown` for the npm ecosystem that is at least as long as `npmMinimalAgeGate` in `.yarnrc.yml` (currently `5d`)
- [ ] The relationship between the two values is documented where the gate is configured, so raising one prompts raising the other
- [ ] The `github-actions` ecosystem entry is reviewed for the same treatment, or explicitly excluded with a reason
- [ ] A guard (extending `scripts/__tests__/npm-minimal-age-gate.test.mjs`) fails when the Dependabot cooldown drops below the Yarn gate

### Why this was not fixed in #5138

`.github/dependabot.yml` is dependency-update automation, and the repository's agent instructions require asking before changing branch/PR automation. #5138 was kept to a config-only diff over `.yarnrc.yml`, the create-app template and its brief; this is tracked here instead of expanding that PR's scope.

Related: #5138

adeptofvoltron:
Closes #5549

## 🎯 Goal
Stop Dependabot from opening npm update PRs that cannot install. #5138 raised yarn's `npmMinimalAgeGate` to `5d`, but Dependabot was unaware of that window, so its weekly Monday run kept proposing versions published within the previous five days — PRs that sit red until the version simply ages out.

## 🔍 Problem
`.github/dependabot.yml` declared no `cooldown`. Yarn's `npmMinimalAgeGate` is a client-side gate applied during resolution, so any Dependabot PR bumping a version younger than five days fails `yarn install` with `YN0016: … All versions satisfying "<range>" are quarantined`. `scripts/ci/npm-retry-on-quarantine.sh` correctly fast-fails rather than retrying — no amount of waiting inside a CI job clears a client-side age gate — and its message points at `npmPreapprovedPackages`, a standing exception somebody then has to remember to remove. Two mechanisms that should agree were contradicting each other.

## 🔍 Root Cause
The two age windows live in different files and had no link between them: yarn's gate in `.yarnrc.yml`, Dependabot's schedule in `.github/dependabot.yml`. #5138 deliberately kept its diff to `.yarnrc.yml` and the create-app template and recorded the gap as a known follow-up in `.ai/specs/briefs/2026-08-09-dependabot-cooldown-ci-gate.md`, because the repository's agent instructions require asking before changing dependency-update automation. This PR closes that gap.

## What Changed
- `.github/dependabot.yml`, npm ecosystem — added `cooldown.default-days: 5`, exactly matching `npmMinimalAgeGate: 5d`. No safety margin is needed: `dependabot-core`'s `CooldownCalculation.within_cooldown_window?` compares exact seconds (`Time.now.to_i - release_date.to_i < cooldown_days * 86400`), not calendar days, so five cooldown days is five real days; and resolution in CI happens strictly after the PR is opened, so the version is only ever older by the time yarn checks it. A comment at the config site explains the coupling, names the failure mode, and points at the guard test. (The first push of this branch used `6` on the mistaken assumption of whole-day arithmetic; the review pass verified the upstream implementation and corrected it in `ccb7ceb`.)
- `.github/dependabot.yml`, github-actions ecosystem — added `cooldown.default-days: 5`. This is the acceptance criterion's "reviewed for the same treatment" resolved as *apply it*: actions sit outside yarn's dependency graph, so `npmMinimalAgeGate` never sees them and this cooldown is the only release-age defence on that supply chain. The value is set independently of the yarn gate rather than derived from it — five days matches the window agreed for npm and covers the detection time of the action-hijack incidents on record.
- `scripts/__tests__/npm-minimal-age-gate.test.mjs` — extended with two guards that parse `.github/dependabot.yml` with `yaml` and compare it against `npmMinimalAgeGate`. The npm guard checks `default-days` *and* every `semver-*-days` key that is present, because a per-bump-type key takes precedence over `default-days` and would otherwise let patch updates slip back under the gate unnoticed. Both configured values are valid against the published `dependabot-2.0` schema (`default-days` accepts 1–90 and is supported for every ecosystem here).

Two behaviours are worth naming for reviewers, both documented in the config comments:

- Dependabot's `cooldown` applies to **version updates only**, never security updates. A Dependabot *security* update can still propose a version yarn quarantines. That is the intended trade-off: an advisory is exactly the situation where `yarn add pkg@version --no-time-gate` is the right escape hatch, and suppressing security updates for five days would be strictly worse.
- Dependabot applies a 3-day cooldown by default even when unconfigured, which is shorter than the yarn gate — hence the explicit `default-days`, and the guard assertion that it must be present.

## 🧪 Tests
- `node --test scripts/__tests__/npm-minimal-age-gate.test.mjs` — 5 passed / 0 failed (2 of them new).
- Regression proof: re-running the same file against the pre-fix `.github/dependabot.yml` from `develop` fails both new tests with the intended messages, and passes again once the fix is restored.
- `node --test scripts/__tests__/*.test.mjs` — 637 passed / 1 failed, 638 total. The single failure is `check-package-peer-deps.test.mjs` ("Cannot verify peer dependencies: none of the 196 declared dependencies could be resolved, so nothing was actually checked. Run `yarn install` first"), a pre-existing environment failure in an agent worktree with no `node_modules`; it is unrelated to this diff and passes in CI, which installs first.
- Gate commands skipped, with reason: `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn build:app`. This diff touches one YAML config and one `node:test` guard script — no TypeScript, no package source, no i18n strings, no app code — and the worktree has no installed dependencies to run them against. `yarn test:scripts` is the gate command that actually covers the changed surface, and it is the one that ran.

The relationship is documented on the Dependabot side and enforced by the guard test, but the reciprocal comment in `.yarnrc.yml` is **not** in this PR: the agent session's permission mode denied edits to `.yarnrc.yml`, and routing around that block would defeat the point of the restriction. The intended addition, directly above `npmMinimalAgeGate: 5d`:

```yaml
# Raising this value means raising the npm `cooldown` in .github/dependabot.yml
# to match — Dependabot cannot see this gate, so a shorter cooldown makes it
# propose versions whose PR fails to install.
# scripts/__tests__/npm-minimal-age-gate.test.mjs fails when the two drift apart.
```

It is a comment-only change with no effect on yarn behaviour, and the maintainer decided during the run to ship without it rather than hold the fix. That is a defensible trade because the comment was never the enforcement — the guard test is, and it ships here: raising `npmMinimalAgeGate` without raising the cooldown turns the `test` job red. What is lost is only the courtesy pointer that would have warned the editor before they hit the failure. The gap is recorded on #5549 so it stays visible after this PR closes the issue.

## 💥 Breaking Changes
None. No runtime code, no public contract, no database schema, and no API surface is touched. The only behavioural change is that Dependabot waits longer before opening a version-update PR; nothing that already exists in `yarn.lock` is re-resolved or re-checked.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,15 +13,16 @@ updates:
     # Dependabot keeps proposing versions whose PR cannot install: `yarn install`
     # fails with `YN0016 ... quarantined` and scripts/ci/npm-retry-on-quarantine.sh
     # fast-fails, because waiting inside a CI job never clears a client-side gate.
-    # The extra day absorbs the mismatch between Dependabot's whole-day cooldown
-    # arithmetic and yarn's exact-duration gate, so a PR opened the moment the
-    # cooldown lapses is already past 5d of real elapsed time.
+    # Five needs no safety margin: dependabot-core compares exact seconds
+    # (`Time.now - release_date < days * 86400`), not calendar days, and resolution
+    # in CI happens strictly after the PR is opened, so the version is only ever
+    # older by then.
     # Cooldown covers version updates only — a Dependabot *security* update may
     # still propose a version the yarn gate rejects, which is the intended
     # trade-off: the advisory is the signal to reach for `--no-time-gate`.
     # Guarded by scripts/__tests__/npm-minimal-age-gate.test.mjs.
     cooldown:
-      default-days: 6
+      default-days: 5
     groups:
       # Batch minor/patch updates for non-security deps to reduce noise
       minor-and-patch:
@@ -41,9 +42,9 @@ updates:
       day: monday
     open-pull-requests-limit: 5
     # Actions live outside yarn's dependency graph, so `npmMinimalAgeGate` never
-    # sees them and a too-young action breaks no install — this cooldown is the
-    # only release-age defence on that supply chain and needs no install-margin.
-    # Five days matches the window agreed for npm in
+    # sees them — this cooldown is the only release-age defence on that supply
+    # chain, and it is set independently of the yarn gate rather than derived
+    # from it. Five days matches the window agreed for npm in
     # .ai/specs/briefs/2026-08-09-dependabot-cooldown-ci-gate.md, which is long
     # enough to cover the detection time of the action-hijack incidents on record.
     cooldown:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,20 @@ updates:
       interval: weekly
       day: monday
     open-pull-requests-limit: 10
+    # Must stay at or above `npmMinimalAgeGate` in .yarnrc.yml (5d). Yarn refuses
+    # to resolve a version younger than that gate, so without a matching cooldown
+    # Dependabot keeps proposing versions whose PR cannot install: `yarn install`
+    # fails with `YN0016 ... quarantined` and scripts/ci/npm-retry-on-quarantine.sh
+    # fast-fails, because waiting inside a CI job never clears a client-side gate.
+    # The extra day absorbs the mismatch between Dependabot's whole-day cooldown
+    # arithmetic and yarn's exact-duration gate, so a PR opened the moment the
+    # cooldown lapses is already past 5d of real elapsed time.
+    # Cooldown covers version updates only — a Dependabot *security* update may
+    # still propose a version the yarn gate rejects, which is the intended
+    # trade-off: the advisory is the signal to reach for `--no-time-gate`.
+    # Guarded by scripts/__tests__/npm-minimal-age-gate.test.mjs.
+    cooldown:
+      default-days: 6
     groups:
       # Batch minor/patch updates for non-security deps to reduce noise
       minor-and-patch:
@@ -26,3 +40,11 @@ updates:
       interval: weekly
       day: monday
     open-pull-requests-limit: 5
+    # Actions live outside yarn's dependency graph, so `npmMinimalAgeGate` never
+    # sees them and a too-young action breaks no install — this cooldown is the
+    # only release-age defence on that supply chain and needs no install-margin.
+    # Five days matches the window agreed for npm in
+    # .ai/specs/briefs/2026-08-09-dependabot-cooldown-ci-gate.md, which is long
+    # enough to cover the detection time of the action-hijack incidents on record.
+    cooldown:
+      default-days: 5
--- a/scripts/__tests__/npm-minimal-age-gate.test.mjs
+++ b/scripts/__tests__/npm-minimal-age-gate.test.mjs
@@ -2,12 +2,18 @@ import assert from 'node:assert/strict'
 import fs from 'node:fs'
 import path from 'node:path'
 import test from 'node:test'
+import { parse } from 'yaml'
 
 const rootYarnrcPath = path.resolve('.yarnrc.yml')
 const templateYarnrcPath = path.resolve('packages/create-app/template/.yarnrc.yml.template')
 const retryScriptPath = path.resolve('scripts/ci/npm-retry-on-quarantine.sh')
+const dependabotConfigPath = path.resolve('.github/dependabot.yml')
 
 const AGREED_MINIMAL_AGE_GATE = '5d'
+
+const DURATION_UNIT_IN_DAYS = { m: 1 / 1440, h: 1 / 24, d: 1, w: 7 }
+
+const COOLDOWN_DAY_KEYS = ['default-days', 'semver-major-days', 'semver-minor-days', 'semver-patch-days']
 
 function readText(filePath) {
   return fs.readFileSync(filePath, 'utf8')
@@ -15,6 +21,26 @@ function readText(filePath) {
 
 function readMinimalAgeGate(filePath) {
   return readText(filePath).match(/^npmMinimalAgeGate:\s*(\S+)$/m)?.[1]
+}
+
+function durationToDays(duration) {
+  const parsed = String(duration).match(/^(\d+)([mhdw])$/)
+
+  assert.ok(
+    parsed,
+    `cannot read "${duration}" as a yarn duration — teach durationToDays the new unit before changing npmMinimalAgeGate, otherwise this guard silently stops comparing`
+  )
+
+  return Number(parsed[1]) * DURATION_UNIT_IN_DAYS[parsed[2]]
+}
+
+function readDependabotEcosystem(ecosystem) {
+  const config = parse(readText(dependabotConfigPath))
+  const entry = config.updates?.find((update) => update['package-ecosystem'] === ecosystem)
+
+  assert.ok(entry, `.github/dependabot.yml must keep an updates entry for the "${ecosystem}" ecosystem`)
+
+  return entry
 }
 
 test('the monorepo and scaffolded apps quarantine third-party releases for the agreed duration', () => {
@@ -30,6 +56,39 @@ test('the monorepo and scaffolded apps quarantine third-party releases for the a
     template,
     root,
     'packages/create-app/template/.yarnrc.yml.template must declare the same npmMinimalAgeGate as the monorepo, otherwise scaffolded apps silently fall back to yarn\'s 1d default'
+  )
+})
+
+test('dependabot holds npm updates back at least as long as the yarn release-age gate', () => {
+  const gateDays = durationToDays(AGREED_MINIMAL_AGE_GATE)
+  const cooldown = readDependabotEcosystem('npm').cooldown
+
+  assert.ok(
+    cooldown,
+    'the npm entry in .github/dependabot.yml must declare a cooldown — without one Dependabot proposes versions younger than npmMinimalAgeGate, and those PRs cannot install (`yarn install` fails with YN0016 ... quarantined until the version ages out)'
+  )
+
+  const declaredKeys = COOLDOWN_DAY_KEYS.filter((key) => cooldown[key] !== undefined)
+
+  assert.ok(
+    declaredKeys.includes('default-days'),
+    'the npm cooldown must set default-days — dependabot falls back to its own 3-day default for every bump type left unspecified, which is shorter than npmMinimalAgeGate'
+  )
+
+  for (const key of declaredKeys) {
+    assert.ok(
+      cooldown[key] >= gateDays,
+      `the npm cooldown ${key} (${cooldown[key]}) must be at least npmMinimalAgeGate from .yarnrc.yml (${AGREED_MINIMAL_AGE_GATE}); a shorter window lets dependabot open update PRs that yarn refuses to resolve`
+    )
+  }
+})
+
+test('dependabot applies a release-age cooldown to github actions as well', () => {
+  const cooldown = readDependabotEcosystem('github-actions').cooldown
+
+  assert.ok(
+    cooldown?.['default-days'] >= 1,
+    'the github-actions entry in .github/dependabot.yml must declare a cooldown default-days — actions sit outside yarn\'s dependency graph, so this cooldown is the only release-age defence on that supply chain'
   )
 })
 


adeptofvoltron:

- 🔍 `review` — the run is finished and the PR is ready for code review; nothing is left in progress on the implementation side.
- 📦 `dependencies` — the diff is dependency-update automation: it changes when Dependabot is allowed to propose an npm or GitHub Actions version bump.
- 🔒 `security` — the cooldown exists to hold freshly-published releases at arm's length, and this PR extends that supply-chain defence to the GitHub Actions ecosystem, which yarn's gate never covered.
- ⏭️ `skip-qa` — the automated-verification exemption applies cleanly: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no database structure or API surface change, no `BACKWARD_COMPATIBILITY.md` contract touched, and the changed behaviour ships with its own guard test in the same PR. There is no user-facing surface for a human to exercise in a browser.
- 🔹 `priority-medium` — the security control this aligns with already shipped in #5138 and stays fully in force, so nothing is unprotected while this waits; what it removes is the recurring cost of red Dependabot PRs each Monday and the standing-exception maintenance trap that follows them. That is an ordinary maintenance defect rather than an active vulnerability, which puts it below `priority-high` even though the `security` signal points that way.
- 🟢 `risk-low` — the blast radius is two YAML comments plus a `cooldown` key and a `node:test` guard. No runtime code, no shared contract surface, no migration, and nothing that is re-resolved for dependencies already pinned in `yarn.lock`.

adeptofvoltron:
🤖 `om-open-pr` — chain lock transferred from #5549 to this PR at 2026-08-24T12:36:12Z, held by @adeptofvoltron. The chain continues here with `om-auto-review-pr`; other auto-skills should skip this PR until the lock is released.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock on this PR at 2026-08-24T12:36:47Z (inherited from `om-open-pr`). Reviewing now with `--autofix`; the lock stays held until `om-auto-fix-issue` releases it at the end of the chain.

adeptofvoltron:
# 🔍 Code Review

## 🎯 Summary

This PR closes the gap #5138 deliberately left open: yarn's `npmMinimalAgeGate: 5d` quarantines any version published within the last five days, but `.github/dependabot.yml` declared no `cooldown`, so Dependabot's weekly Monday run kept proposing exactly those versions. The result was a class of PR that cannot install — `yarn install` fails with `YN0016 … quarantined`, and `scripts/ci/npm-retry-on-quarantine.sh` correctly fast-fails, because no amount of waiting inside a CI job clears a client-side age gate.

The diff is two YAML keys and a guard test. Both npm and github-actions now declare `cooldown.default-days: 5`, the coupling is documented at the config site, and `scripts/__tests__/npm-minimal-age-gate.test.mjs` now fails when the Dependabot cooldown drops below the yarn gate or when either ecosystem loses its cooldown entirely.

**Verdict: approve** — no blockers and no majors remain. One major finding was raised during this pass and fixed on the branch before the verdict was submitted (see 🔧 Fixed during this review below); what is left is one minor observation that needs no code change and one nit.

**Self-approval note:** GitHub does not allow the author of a PR to submit an approving review, and this PR was opened by the same automation account that is reviewing it. This report is therefore posted as a review comment rather than a formal `APPROVED` review, and the PR still needs one human approval before merge. That is the intended gate, not a workaround.

## 🧪 Validation Gate

| Command | Result | Evidence |
|---|---|---|
| `yarn test:scripts` (as `node --test scripts/__tests__/*.test.mjs`) | ✅ pass for this diff | 637 passed / 1 failed of 638. The one failure is `check-package-peer-deps.test.mjs`, which aborts with *"Cannot verify peer dependencies: none of the 196 declared dependencies could be resolved… Run `yarn install` first"* — an environment failure in an agent worktree with no `node_modules`, not a finding against this diff. The file this PR changes, `npm-minimal-age-gate.test.mjs`, is 5 passed / 0 failed. |
| `yarn build:packages` | ⏭️ not run | No package source in the diff, and the worktree has no installed dependencies. |
| `yarn generate` | ⏭️ not run | No module files subject to auto-discovery changed, so there is nothing to regenerate. |
| `yarn i18n:check-sync`, `yarn i18n:check-usage` | ⏭️ not run | No locale files and no user-facing strings in the diff. |
| `yarn typecheck` | ⏭️ not run | No TypeScript in the diff; the one changed script is plain `.mjs` run by `node --test`. |
| `yarn build:app` | ⏭️ not run | No application code in the diff. |

Because the local gate is partial, CI on this head is the authoritative evidence and it is reported honestly rather than assumed: on the first pushed commit `prepare`, `lint`, `ds-lint`, `scope`, `audit-scope`, `license/cla` and both settled CodeQL analyses passed, with `test`, `Analyze (javascript-typescript)` and `documents-multi-instance` still pending. Nothing was failing at review time. The CI follow-up below reports the settled outcome for the current head.

## 🔧 Fixed during this review

### major — `.github/dependabot.yml:25` — the npm cooldown was one day longer than agreed, on a rationale that was factually wrong

The first push of this branch set `default-days: 6` and justified the extra day with a comment claiming that "Dependabot's whole-day cooldown arithmetic" needed a margin over yarn's exact-duration gate. That claim does not survive checking against the implementation. `dependabot-core`'s `CooldownCalculation.within_cooldown_window?` is an exact-seconds comparison:

```ruby
(Time.now.to_i - release_date.to_i) < (cooldown_days * DAY_IN_SECONDS)
```

Five cooldown days is therefore five real days, aligning precisely with `npmMinimalAgeGate: 5d`, and dependency resolution in CI happens strictly after the PR is opened, so the version is only ever *older* by the time yarn checks it. The margin bought nothing.

This mattered beyond tidiness: the review of #5138 explicitly pushed back on windows longer than the agreed range, on the grounds that an over-long gate delays legitimate security patches and pushes urgent upgrades through standing exceptions. Shipping a sixth day would have quietly re-litigated that decision inside a follow-up PR, backed by an incorrect premise about a third-party implementation.

**Fixed in `ccb7ceb`:** `default-days` is now `5`, the comment states the real mechanism and cites the exact-seconds comparison, and the github-actions comment no longer refers to an "install-margin" that no longer exists. The PR body was corrected to match. The guard test passes at 5 (`5 >= 5`) and, verified explicitly, still fails at 4 with *"the npm cooldown default-days (4) must be at least npmMinimalAgeGate from .yarnrc.yml (5d)"*.

## 🟡 Minor findings

### minor — `.github/dependabot.yml:11` — the weekly schedule compounds the cooldown, and that is worth knowing even though nothing should change

`schedule.interval: weekly` on Monday means eligibility is only ever evaluated once a week. A version that clears its five-day cooldown on a Tuesday is not proposed until the following Monday, so the *effective* delay for routine version updates can reach roughly twelve days rather than five.

No change is requested. This is inherent to combining any cooldown with a weekly cadence — it was already true of the 3-day cooldown Dependabot applies by default — and the alternative (a daily schedule) trades it for substantially more PR churn, which the existing `groups` configuration exists to reduce. It matters only that nobody later reads "5 days" as the worst-case freshness lag. Security updates are unaffected: they are advisory-triggered rather than scheduled, and `cooldown` never applies to them.

## 🔵 Nits

### nit — `scripts/__tests__/npm-minimal-age-gate.test.mjs:32` — the duration unit table is a small guess, but it fails closed

`DURATION_UNIT_IN_DAYS` maps `m` to minutes, following the `ms`-style convention yarn's duration parsing uses. If that were ever wrong — if some unit meant something else — the guard could compare against a wrong number of days. The design already handles this correctly: any unit outside `m/h/d/w` fails the `durationToDays` assertion with *"teach durationToDays the new unit before changing npmMinimalAgeGate, otherwise this guard silently stops comparing"*, so an unrecognised value produces a red test rather than a silently weakened check. That is the right failure direction for a security guard; no change needed.

## 💥 Breaking Changes

- [x] **No public contract surface touched.** `.github/dependabot.yml` is dependency-update automation and is not one of the thirteen contract-surface categories in `BACKWARD_COMPATIBILITY.md` (auto-discovery files, types, signatures, import paths, event IDs, widget spot IDs, API routes, DB schema, DI keys, ACL features, notification IDs, CLI commands, generated files).
- [x] **No exported API, HTTP route, response shape, event name, CLI flag, or DI key changed.**
- [x] **No database schema change**, so no migration and no snapshot update is owed.
- [x] **No config format consumed by third-party modules changed.** The create-app template ships no `dependabot.yml`, so scaffolded apps are unaffected; `packages/create-app/template/.yarnrc.yml.template` is untouched and the existing template-parity assertion still passes.
- [x] **No behavioural change to existing installs.** Both gates fire during resolution, not on `yarn install --immutable`; versions already pinned in `yarn.lock` are never re-checked.
- [x] **Governance check cleared.** `AGENTS.md` requires asking before changing dependency-update automation. That ask is #5549 itself, which was filed precisely to carry this change out of #5138's scope and states the acceptance criteria this PR implements.

## 🧪 Test Coverage

Coverage is adequate and genuinely regression-proving rather than tautological:

- **The guard fails without the fix.** Running the changed test file against `origin/develop`'s `.github/dependabot.yml` fails both new tests with their intended messages — *"the npm entry in .github/dependabot.yml must declare a cooldown…"* and *"the github-actions entry … must declare a cooldown default-days…"* — and passes once the fix is restored.
- **The guard fails at the boundary.** Lowering `default-days` to `4` fails with the comparison message quoted above, confirming the assertion tests the actual relationship rather than merely the key's presence.
- **The guard covers the sharp edge, not just the obvious one.** It checks every `semver-*-days` key that is present, not only `default-days`. That matters because a per-bump-type key takes precedence over `default-days` in `ReleaseCooldownOptions`, so a future `semver-patch-days: 1` would otherwise slip patch updates back under the yarn gate with the headline value still reading 5.
- **It also pins the presence of `default-days` itself**, because Dependabot falls back to its own 3-day default for anything unspecified — shorter than the yarn gate, and silently so.
- **Schema conformance was verified** against the published `dependabot-2.0` JSON schema: no unknown keys, and `default-days: 5` sits inside the documented 1–90 range for both ecosystems.

The one gap is not a test gap: the reciprocal comment in `.yarnrc.yml` named by the issue's second acceptance criterion is missing, because this session's sandbox denies edits to that file. It is called out prominently in the PR body with the exact snippet to add. It is comment-only and changes no behaviour, and the enforcement it would describe is the guard test, which does ship here — so it is a documentation follow-up rather than a functional gap, and it is the only acceptance criterion not met by this diff.

adeptofvoltron:
🤖 `om-auto-review-pr` — completed: **approve** (no blockers, no majors remaining). One major finding was raised and fixed on the branch during the pass — `default-days` went from 6 back to 5 in `ccb7ceb` after checking `dependabot-core`'s cooldown arithmetic, which compares exact seconds rather than calendar days.

The pipeline label stays `review` rather than moving to `merge-queue`: GitHub refuses an approving review from the PR's own author, so this PR still needs one human approval before it is genuinely merge-ready. Labelling it `merge-queue` on the strength of a self-review would misrepresent that.

**Lock retained — chain continues.** `om-auto-fix-issue` releases it at the end of the run.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the fix is implemented, reviewed and pushed, and the chain has nothing further to change here. The verdict is **approve** with no blockers or majors outstanding — the one major found during the review pass (`default-days: 6` justified by an incorrect claim about Dependabot's cooldown arithmetic) was corrected in `ccb7ceb` before the verdict was recorded.

UI verification is **n/a**: the diff touches one CI/automation YAML file and one `node:test` guard script, with no route, component, template, style or user-visible copy, so there is no browser surface to exercise. The PR carries `skip-qa` under the automated-verification exemption for the same reason.

**Lock released.** `in-progress` is swapped for `ci-monitoring` while the checks on `ccb7ceb` settle — that is a meta label, not a claim, so this PR is free for anyone to pick up. It still needs one human approval: GitHub does not let the authoring account approve its own PR.

adeptofvoltron:

All checks on `ccb7cebe` have settled: **13 passed, 0 failed, 7 skipped.** The passing set includes `test` (which runs `yarn test:scripts` and therefore the two new guards in `scripts/__tests__/npm-minimal-age-gate.test.mjs`), `lint`, `ds-lint`, `prepare`, `scope`, `audit-scope`, `documents-multi-instance`, all three CodeQL analyses and `license/cla`. Worth noting separately: the check named `.github/dependabot.yml` is GitHub's own validation of the Dependabot configuration, and it passes — independent confirmation that the `cooldown` block is accepted syntax for both the `npm` and `github-actions` ecosystems, not just schema-valid on paper.

This also closes the gap left by the partial local gate: `build:packages`, `generate`, the i18n checks, `typecheck` and `build:app` could not run in the agent worktree, and CI has now run the full pipeline against this head.

`ci-monitoring` removed — no further follow-up is owed from this agent. `mergeStateStatus` is still `BLOCKED`, and the only thing blocking it is the required approving review, which the authoring account cannot supply.

wojciechszyjka (approved):
Approved.

PR #5566, opened 2026-08-24
fix(customers): exclude the edited person from the duplicate-email lookup

### LINKED ISSUE #5534: Person form duplicate-email warning matches the record against itself
Noticed during manual QA of #4471; unrelated to that PR.

## What happens

Editing a person in the CRM and changing **Primary email** shows the duplicate warning against the record being edited:

> This email is already assigned to Mia Johnson. Open record

…while editing Mia Johnson. The save itself succeeds and the new address is stored correctly — the warning is a false positive caused by the uniqueness lookup not excluding the current record.

## Reproduce

1. Open a person in the CRM edit form (`/backend/customers/people-v2/<id>`)
2. Change Primary email to any address not used elsewhere, save
3. The field shows "This email is already assigned to <the same person>"

Confirmed the value persists: reading the record back through `/api/customers/people?id=…` returns the new address.

## Suggested fix

Exclude the record's own id from the duplicate-email lookup that backs this warning.

adeptofvoltron:
Fixes #5534

## 🎯 What this changes

`useEmailDuplicateCheck` — the hook behind the "This email is already assigned to …" hint on the CRM person form — now asks the people API to exclude the record being edited (`excludeIds=<recordId>`) instead of relying solely on a client-side id comparison after the fact.

The lookup requests `pageSize=5`. Before this change the edited record could occupy one of those five slots and was only discarded afterwards in the browser, which has two consequences:

- a genuine duplicate could be pushed out of the returned page and silently missed (false negative), and
- the *only* thing standing between the user and a self-referential "already assigned to <yourself>" warning was that client-side id compare — if the id the form holds and the id the list returns ever diverge, the record is reported as its own duplicate, which is exactly the symptom reported in #5534.

The client-side guard is kept as defense in depth; the query-level exclusion makes the self-match impossible at the source.

## 🔍 Investigation note for the reviewer and for QA

I want to be straightforward about what I could and could not verify, because it affects how this PR should be signed off.

I traced the full path behind the reported warning — `createPrimaryEmailField` → `useEmailDuplicateCheck` → `GET /api/customers/people` → `CrudForm`'s `recordId` (derived from `values.id`) — and wrote throwaway probes against the real `CrudForm` and the real form config. In the current code on `develop`, `recordId` does reach the field on `/backend/customers/people-v2/<id>`, and with it the hook already refuses to report the edited record. I could not reproduce the exact reported repro from the code alone.

So this PR implements the fix the issue asks for and closes the entire class of self-match false positives at the query layer, but I cannot claim to have watched the reported symptom disappear. If QA still sees the warning after this lands, the remaining explanations are worth checking directly in the tenant:

- a second person record sharing the same display name (the warning would then be correct, just confusing), or
- the prefix match mode — the hint fires when another record's address merely *starts with* what was typed, while the copy claims the address is "already assigned", which is only true for an exact match.

Both are separate from this change and would each deserve their own issue.

## 🧪 Tests

New: `packages/core/src/modules/customers/backend/hooks/__tests__/useEmailDuplicateCheck.test.tsx` — the hook had no coverage at all before this.

- sends `excludeIds` for the edited record, in both `prefix` and `exact` match modes
- never reports the edited record as its own duplicate, even when the API returns it anyway
- still reports a duplicate that belongs to a different record
- omits the exclusion on create, where there is no record yet

Verification run in this worktree: the new suite passes, and the full customers module suite passes (247 suites / 1587 tests). `tsc --noEmit` reports no errors in the touched files. The worktree has no `node_modules`, so the full ordered gate (`yarn build:packages` → `yarn generate` → … → `yarn build:app`) could not run locally and is left to CI — Jest and `tsc` were invoked directly from the repository toolchain instead.

## 💥 Breaking changes

None. `excludeIds` is an existing, already-supported query parameter on `GET /api/customers/people`; no schema, API surface, or contract change.

## 📸 Manual QA

Still pending — this PR carries `needs-qa` deliberately rather than taking the automated-verification exemption, because the reported symptom was observed by a human and I could not reproduce it in code. Please re-run the original repro from #5534 on a person record and confirm the warning no longer appears after saving a fresh address.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5566"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790167764&installation_model_id=432243&pr_number=5566&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5566&signature=89ab46922bae132b24e94cd6b00fafade40a8414e26be3e68896655b65988215"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/customers/backend/hooks/__tests__/useEmailDuplicateCheck.test.tsx
+++ b/packages/core/src/modules/customers/backend/hooks/__tests__/useEmailDuplicateCheck.test.tsx
@@ -1,0 +1,93 @@
+/**
+ * @jest-environment jsdom
+ */
+const apiCallMock = jest.fn()
+
+jest.mock('@open-mercato/ui/backend/utils/apiCall', () => ({
+  apiCall: (...args: unknown[]) => apiCallMock(...args),
+}))
+
+import * as React from 'react'
+import { act, render, screen } from '@testing-library/react'
+import { useEmailDuplicateCheck } from '../useEmailDuplicateCheck'
+
+type ProbeProps = {
+  email: string
+  recordId?: string | null
+  matchMode?: 'exact' | 'prefix'
+}
+
+function Probe({ email, recordId, matchMode = 'prefix' }: ProbeProps) {
+  const { duplicate } = useEmailDuplicateCheck(email, { recordId, matchMode, debounceMs: 0 })
+  return <div data-testid="duplicate">{duplicate ? duplicate.displayName : ''}</div>
+}
+
+const respondWith = (items: Array<Record<string, unknown>>) => {
+  apiCallMock.mockResolvedValue({ ok: true, result: { items } })
+}
+
+const settle = async () => {
+  await act(async () => {
+    await new Promise((resolve) => setTimeout(resolve, 20))
+  })
+}
+
+const requestedUrl = (): string => String(apiCallMock.mock.calls.at(-1)?.[0] ?? '')
+
+const duplicateText = () => screen.getByTestId('duplicate').textContent
+
+describe('useEmailDuplicateCheck', () => {
+  beforeEach(() => {
+    apiCallMock.mockReset()
+  })
+
+  it('excludes the edited record from the lookup query', async () => {
+    respondWith([])
+    render(<Probe email="mia.johnson@example.com" recordId="person-1" />)
+    await settle()
+
+    expect(requestedUrl()).toContain('excludeIds=person-1')
+    expect(requestedUrl()).toContain('emailStartsWith=mia.johnson%40example.com')
+  })
+
+  it('never reports the edited record as a duplicate of itself', async () => {
+    respondWith([
+      { id: 'person-1', display_name: 'Mia Johnson', primary_email: 'mia.johnson@example.com' },
+    ])
+    render(<Probe email="mia.johnson@example.com" recordId="person-1" />)
+    await settle()
+
+    expect(duplicateText()).toBe('')
+  })
+
+  it('still reports a duplicate owned by another record', async () => {
+    respondWith([
+      { id: 'person-1', display_name: 'Mia Johnson', primary_email: 'mia.johnson@example.com' },
+      { id: 'person-2', display_name: 'Noah Berg', primary_email: 'mia.johnson@example.com' },
+    ])
+    render(<Probe email="mia.johnson@example.com" recordId="person-1" />)
+    await settle()
+
+    expect(duplicateText()).toBe('Noah Berg')
+  })
+
+  it('omits the exclusion when creating a record', async () => {
+    respondWith([])
+    render(<Probe email="mia.johnson@example.com" recordId={null} />)
+    await settle()
+
+    expect(requestedUrl()).not.toContain('excludeIds')
+  })
+
+  it('excludes the edited record in exact match mode too', async () => {
+    respondWith([
+      { id: 'person-1', display_name: 'Mia Johnson', primary_email: 'mia.johnson@example.com' },
+    ])
+    render(<Probe email="mia.johnson@example.com" recordId="person-1" matchMode="exact" />)
+    await settle()
+
+    expect(requestedUrl()).toContain('email=mia.johnson%40example.com')
+    expect(requestedUrl()).toContain('excludeIds=person-1')
+    expect(duplicateText()).toBe('')
+  })
+})
--- a/packages/core/src/modules/customers/backend/hooks/useEmailDuplicateCheck.ts
+++ b/packages/core/src/modules/customers/backend/hooks/useEmailDuplicateCheck.ts
@@ -40,16 +40,24 @@ export function useEmailDuplicateCheck(
     }
 
     const normalized = trimmed.toLowerCase()
+    const excludedId = typeof recordId === 'string' && recordId.trim().length ? recordId.trim() : null
     let cancelled = false
     const controller = new AbortController()
     const timer = window.setTimeout(async () => {
       setChecking(true)
       try {
-        const queryParam =
+        const params = [
           matchMode === 'prefix'
             ? `emailStartsWith=${encodeURIComponent(normalized)}`
-            : `email=${encodeURIComponent(normalized)}`
-        const call = await apiCall<{ items?: unknown[] }>(`/api/customers/people?${queryParam}&pageSize=5&page=1`, {
+            : `email=${encodeURIComponent(normalized)}`,
+          'pageSize=5',
+          'page=1',
+        ]
+        // The record being edited is never its own duplicate (#5534). Excluding it in the
+        // query — not only in the client-side scan below — also stops it from consuming one
+        // of the five returned slots and hiding a genuine duplicate.
+        if (excludedId) params.push(`excludeIds=${encodeURIComponent(excludedId)}`)
+        const call = await apiCall<{ items?: unknown[] }>(`/api/customers/people?${params.join('&')}`, {
           signal: controller.signal,
         })
         if (!call.ok) {
@@ -72,7 +80,7 @@ export function useEmailDuplicateCheck(
             })
             .filter((entry: EmailDuplicateMatch | null): entry is EmailDuplicateMatch => !!entry)
             .find((entry: EmailDuplicateMatch) => {
-              if (entry.id === recordId) return false
+              if (excludedId && entry.id === excludedId) return false
               return matchMode === 'prefix'
                 ? entry.email.startsWith(normalized)
                 : entry.email === normalized


adeptofvoltron:

- 🔍 `review` — the fix is complete, pushed, and ready for code review.
- 🐛 `bug` — this corrects a defect in the person form's duplicate-email hint reported in #5534, not new behavior.
- 🧪 `needs-qa` — although the diff touches no UI-rendering file and would technically qualify for the automated-verification exemption, the reported symptom was observed by a human during manual QA and could not be reproduced from the code, so a person needs to re-run the original repro before this is considered resolved.
- 🔽 `priority-low` — inherited from the issue: the warning is cosmetic, the save itself always succeeded and stored the correct address.
- 🟢 `risk-low` — a single client-side hook changed, using a query parameter the people API already supports, with new unit tests and no schema, API, or contract change.

adeptofvoltron:
🤖 `om-auto-fix-issue` handed the chain lock from issue #5534 to this PR; `om-auto-review-pr` takes it from here.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the fix and its regression tests are pushed, the label set is final, and the PR is in `review` awaiting a human reviewer plus the manual QA re-run described in the PR body. Lock released; only CI is still being watched (`ci-monitoring`). ⚠️ Note for the reviewer: the reported symptom could not be reproduced from the code, so this PR hardens the lookup rather than demonstrably removing an observed regression — see the investigation note in the description.

adeptofvoltron:
🤖 `om-auto-fix-issue` — ✅ CI is fully green on this branch: 14 checks pass, 5 skipped, none failing. Notably `test` (6m53s) and `ephemeral-integration` (17m46s) both passed, which covers the ordered validation gate that could not be run locally. Dropping `ci-monitoring`; this PR now needs a human code review and the manual QA re-run described in the description.

wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-24T18:22:00Z. The previous `om-auto-fix-issue` chain explicitly released its lock (no `in-progress` label, no live claim), so this is a fresh claim rather than an override. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (approved):
# 🔍 Code Review: fix(customers): exclude the edited person from the duplicate-email lookup

## 🎯 Summary

This PR changes `useEmailDuplicateCheck` — the hook behind the "This email is already assigned to …" hint — so the people lookup carries `excludeIds=<recordId>` instead of relying only on a client-side id comparison after the response comes back. It also keeps that client-side guard as defense in depth, and it adds the hook's first unit-test suite (five cases covering both match modes, the create path, and a hostile API that returns the edited record anyway).

I reviewed both changed files (`packages/core/src/modules/customers/backend/hooks/useEmailDuplicateCheck.ts` and its new `__tests__/useEmailDuplicateCheck.test.tsx`) plus the four call sites of the hook and the server side of the parameter it now sends. The change is small, correct, and well-targeted:

- `excludeIds` is an existing, already-supported query parameter on `GET /api/customers/people` (`api/people/route.ts:65`), parsed into a set and applied through the shared `applyEntityIdExclusion` helper (`api/people/route.ts:277`), which merges the exclusion into an existing `id` filter instead of overwriting it. No API, schema, or contract change is needed to support this PR.
- Every call site passes a customer-entity UUID as `recordId` — `PersonHighlights`/`CompanyHighlights` pass `person.id`/`company.id`, `formConfig`'s `PrimaryEmailField` receives `CrudForm`'s `recordId` (derived from `values.id`), and `ContactEmailInlineEditor` on the sales-document page passes `contactRecordId` (`customerSnapshot.contact.id ?? customerSnapshot.customer.id ?? record.customerEntityId`). There is no call site that would push a non-UUID or a foreign-table id into the new query parameter, so the change cannot turn a harmless client-side no-op into a server-side 500 that silently disables the hint.
- Moving the exclusion to the query layer is a genuine improvement independent of the reported symptom: with `pageSize=5`, the edited record could previously occupy one of the five returned slots and push a real duplicate out of the page, producing a **false negative**. That failure mode is now impossible at the source.

The one thing this PR cannot do is demonstrate that the symptom reported in #5534 disappears, and the author says so plainly in the description rather than glossing over it. I independently reached the same conclusion from the code: `mapPersonOverviewToFormValues` puts `id: overview.person.id` into `initialValues`, `CrudForm` seeds `values` from `initialValues` verbatim (`packages/ui/src/backend/CrudForm.tsx:782`) and derives `recordId` from `values.id` (`CrudForm.tsx:855`), and the people list returns the same customer-entity id (the profile join maps `id → entity_id`, so the row id is the entity id). On the reported path the pre-existing client-side compare therefore already suppressed the self-match, which means the observed warning most likely has a different cause — a second record with the same display name, or the prefix-match wording, exactly the two follow-ups the description names. That is a QA question, not a code-quality one, and this PR carries `needs-qa` precisely so a human re-runs the original repro. I have called it out explicitly in the manual-QA instructions so the QA reviewer treats the #5534 repro as the acceptance criterion rather than a formality.

## Verdict

✅ **approve** — the diff is correct, covered by new tests, uses an already-supported API parameter, changes no contract surface, and CI is fully green on the head commit. The only open question is whether it removes the symptom a human observed in #5534, and the `needs-qa` gate is the right instrument for that; nothing in the code warrants blocking the merge.

## 🧪 Validation Gate

**Validation source: GitHub checks** (workflow run [32729160750](https://github.com/open-mercato/open-mercato/actions/runs/32729160750) on head `ea248fe`). Per this repository's review policy, CI results are preferred over re-running the ordered gate locally when GitHub reports usable check data for the current head — it does here: 14 checks pass, 5 are skipped by path filters, none fail. Each configured `validation.commands` entry maps onto a CI step below.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job, "Build packages" step (3m50s). |
| `yarn build:packages` (rebuild with generated files) | ✅ PASS | `prepare` job, "Rebuild packages with generated files" step. |
| `yarn i18n:check-sync` | ✅ PASS | `test` job, "Check i18n sync" step. |
| `yarn i18n:check-usage` | ✅ PASS | `test` job, "Check i18n usage" step (advisory, `continue-on-error`). |
| `yarn typecheck` | ✅ PASS | `test` job, scoped on PRs to packages changed since `develop` — which includes `@open-mercato/core`, the only package this PR touches. |
| `yarn test` | ✅ PASS | `test` job (6m53s), same change-scoped filter, so the new `useEmailDuplicateCheck` suite did run: it lives under `packages/core/src/**/__tests__/**/*.test.tsx`, which the core Jest `testMatch` covers. |

Beyond the configured gate, `lint`, `ds-lint`, `docker-build`, `documents-multi-instance`, `ephemeral-integration (none)` (17m46s), CodeQL and the CLA check are all green as well. Branch protection for `develop` is not readable through the API from this account (404), so every reported check was treated as required; none of them is failing or pending.

## Findings

### 💅 Nit

- `packages/core/src/modules/customers/backend/hooks/useEmailDuplicateCheck.ts:56-58` — the three-line explanatory comment sits against the repository's "no inline comments (self-documenting code)" rule in the root `AGENTS.md`. I would **keep it as-is**: that rule targets comments that restate what the code already says, whereas this one records *why* the exclusion moved into the query (the `pageSize=5` slot-stealing false negative), which the code genuinely cannot express. Flagging it only so nobody is surprised if a lint-by-eye pass questions it later.
- `packages/core/src/modules/customers/components/detail/CompanyHighlights.tsx:118` — the same hook is used for a **company's** primary email with `recordId={company.id}`, and the hook queries `/api/customers/people`, which only ever returns `kind = 'person'` rows. The new `excludeIds` is therefore a permanent no-op on that surface, exactly as the old client-side compare was. Nothing regresses here, but a company email is only ever checked against people, never against other companies — worth its own follow-up issue if company-to-company duplicates are supposed to be surfaced too. Out of scope for this PR.
- `packages/core/src/modules/customers/backend/hooks/__tests__/useEmailDuplicateCheck.test.tsx:29-33` — `settle()` waits 20 ms of real time with `debounceMs: 0`, so the suite depends on wall-clock scheduling rather than on a deterministic timer. It passes in CI today, and with a 0 ms debounce the margin is generous, but `jest.useFakeTimers()` plus an explicit `advanceTimersByTime` would remove the residual flake risk on a loaded runner. Author's call.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `useEmailDuplicateCheck`'s export and `UseEmailDuplicateCheckOptions` shape are untouched.
- [x] No function signature changed in a breaking way — the hook keeps `(email, options)` with the same optional fields; only the internal query construction changed.
- [x] No HTTP route URL removed or renamed; no method changed — the PR only adds an existing, optional query parameter to a request it already made.
- [x] No field removed or retyped in an existing response shape — the response is consumed exactly as before.
- [x] No database table or column renamed or removed — no migration in this PR, and none needed.
- [x] Where a contract had to change: not applicable, no contract surface was touched.

## 🧪 Test Coverage

The hook had **no** test coverage before this PR, so the new suite is a net addition rather than a top-up. It covers the four behaviors that matter for this change:

- the request carries `excludeIds=<recordId>` alongside `emailStartsWith=` in prefix mode and alongside `email=` in exact mode, which is the actual behavior change;
- the edited record is never reported as its own duplicate **even when the API returns it anyway**, which pins the client-side guard as defense in depth instead of letting it rot once the server does the filtering;
- a duplicate owned by a different record is still reported, which is the regression this class of "just exclude things" fix most easily causes;
- no `excludeIds` is sent on the create path, where there is no record to exclude.

Two gaps are worth naming, neither of them blocking:

- The suite mocks `apiCall`, so it proves the client *sends* `excludeIds` but not that the people endpoint *honors* it. That server-side contract is already exercised by the existing `excludeLinkedCompanyId` / `excludeLinkedDealId` paths through the same `applyEntityIdExclusion` helper, so the risk is low; an integration test asserting that a person excluded by `excludeIds` is absent from `GET /api/customers/people` would close it properly and would belong under `.ai/qa/`.
- The false-negative scenario the description leads with — the edited record consuming one of the five `pageSize=5` slots and hiding a genuine duplicate — is not directly testable from the client side, since the exclusion now happens on the server. Naming it in the PR description, as the author did, is the reasonable substitute.

wojciechszyjka:

- 🚀 `merge-queue` — the code review passed with no blocker, major, or minor findings (three nits only), and every reported CI check on head `ea248fe` is green, so the PR is queued; `needs-qa` stays on, which means the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — kept exactly as the author set it: the change is client-side behavior a human observed going wrong in #5534, and this PR hardens the lookup without being able to demonstrate the reported symptom disappearing, so the original repro must be re-run by a person before this counts as resolved. The automated-verification exemption deliberately does not apply here.
- 🐛 `bug` — this corrects a defect in the person form's duplicate-email hint rather than adding new behavior.
- 🔽 `priority-low` — unchanged and still accurate: the hint is advisory, the save itself always succeeded and stored the correct address, so nothing is broken for the user beyond a confusing message.
- 🟢 `risk-low` — unchanged and still accurate after review: one client-side hook changed, using a query parameter `GET /api/customers/people` already supports, with new unit tests, no schema, API, or contract change, and every call site verified to pass a customer-entity UUID so the new parameter cannot produce a malformed query.

wojciechszyjka:

**Read this first.** The code review confirmed the author's own investigation note: on the reported path the form already knew the edited record's id before this PR, so the client-side guard should already have suppressed a self-match. This PR closes the same hole one layer deeper (the lookup query itself now excludes the record), which is a real improvement, but it is **not proven** to remove the symptom a human saw in #5534. Treat the P1 repro below as the acceptance criterion, not a formality — if the warning still names the edited person after this change, the root cause is elsewhere and this PR should be marked `qa-failed` rather than waved through.

### P1 — Person edit form: the #5534 repro (acceptance criterion)

**Where to click**
- `/backend/customers/people-v2/<id>` — open any existing person in the CRM

**What to verify**
- Change **Primary email** to an address that no other record uses, save → after the save completes and the page reloads, the field shows **no** duplicate warning at all. The old symptom was "This email is already assigned to `<the person you are editing>`".
- Re-focus the email field after the save and retype the same address → still no warning naming this person.
- While a lookup is in flight, the field shows its checking state and then settles; the hint must not flicker in and then stay after the record is saved.

**What can go wrong**
- The warning still names the person being edited → this PR did not fix the reported cause; mark `qa-failed` and note in the comment whether a second person with the **same display name** exists in the tenant (that is the leading alternative explanation, and it deserves its own issue).
- The warning names a **different** person whose address merely *starts with* what you typed — that is the known prefix-match wording problem, also out of scope here; note it, do not fail this PR for it, and file it separately.
- The hint disappears entirely in every scenario, including the genuine-duplicate case below → the exclusion is over-filtering.

### P1 — Genuine duplicates are still detected (the regression this change could cause)

**Where to click**
- `/backend/customers/people-v2/<id>` — person A's edit form
- `/backend/customers/people-v2/<id>` — person B's detail header, inline **Primary email** field
- `/backend/customers/companies-v2/<id>` — a company's inline **Primary email** field

**What to verify**
- Type person B's existing email into person A's Primary email field → the warning appears, names **person B**, and its "Open record" link opens person B.
- Do the same from the person detail header inline editor (click the email value directly) → same warning, same link target.
- Create a new person (`/backend/customers/people/create`) and type an address that already belongs to someone → the warning appears there too; on a create form there is no record to exclude, so nothing may be filtered out.
- On a company's inline email editor, typing a **person's** existing email still surfaces that person — unchanged behavior, just confirm it did not break.

**What can go wrong**
- No warning at all where one is expected → the new `excludeIds` parameter is filtering more than the edited record.
- The warning appears but the "Open record" link points at the record being edited.
- The duplicate check silently stops working (no warning and no checking state) → open the browser devtools Network tab on `GET /api/customers/people`; a `4xx`/`5xx` on that request would mean the new query parameter is being rejected, and the hint fails silently by design.

### P2 — Sales document customer email (same hook, different surface)

**Where to click**
- `/backend/sales/documents/<id>` — the customer contact block, inline **email** editor

**What to verify**
- Edit the customer email inline and type an address belonging to a different person → the duplicate hint still appears with that person's name.
- Type the document's own customer's current address → no self-referential warning.

**What can go wrong**
- The hint disappears on this surface, or the inline save stops working — this page passes the customer/contact id into the same hook, so it is the most likely place for a side effect of this change to show up.

### Perceived performance (any of the routes above)

- Cold-load one of the person routes and confirm the form shell renders before the duplicate lookup resolves; the hint is debounced and must never block typing.
- Type quickly through a full address and confirm no visible input lag and no burst of requests per keystroke in the Network tab.
- Repeat the P1 repro once on a mobile viewport (≤ 420 px) and confirm the warning text and its link stay readable and tappable.

### Pass/fail

wojciechszyjka:

The review found no blocker, major, or minor findings — three nits only (the explanatory inline comment, which I recommend keeping; the company-detail surface where `excludeIds` is a permanent no-op because the endpoint only returns `kind = 'person'`; and the real-timer `settle()` helper in the new suite). `excludeIds` is an already-supported parameter on `GET /api/customers/people`, every call site was verified to pass a customer-entity UUID, and no contract surface changed. Validation source: **GitHub checks** — all reported checks on head `ea248fe` are green (14 pass, 5 skipped by path filters, none failing), and the `test` job's change-scoped filter does include `@open-mercato/core`, so the new hook suite really ran. The ordered gate was therefore not re-run locally, per this repository's GitHub-checks-first review policy.

Autofix: n/a — nothing to fix on an approved verdict. Labels moved `review` → `merge-queue`; `needs-qa`, `bug`, `priority-low` and `risk-low` were kept, so the QA-approval gate still holds the merge until a QA reviewer adds `qa-approved`. The manual-QA instructions above make the #5534 repro the explicit acceptance criterion: the code review confirmed the form already knew the edited record's id before this change, so if the warning still names the edited person after it, the root cause lies elsewhere and this PR should be marked `qa-failed` rather than approved. No CI follow-up is owed — CI had already settled green before the review was submitted.

PR #5567, opened 2026-08-24
fix(ui): stop probing AI assistant API when the module is disabled

Frshy:
## Summary

On an installation that does not enable the `ai_assistant` module, the admin panel still calls its API. Every hard load of any `/backend/*` page fires three requests, all 404:

- `GET /api/ai_assistant/health`
- `GET /api/ai_assistant/ai/agents`
- `GET /api/ai_assistant/ai/conversations?limit=100`

The first two come from `AiAssistantLauncher`, mounted unconditionally in the topbar by `AppShellBody` (`AppShell.tsx:1493`). The third comes from `AiChatSessionsProvider`, which wraps the whole shell (`AppShell.tsx:518`) and syncs conversations in an effect with no condition on it — also logging `Could not load server conversations` on every page load.

Both are now gated on a new `useAiAssistantAvailable()` hook that requires two independent signals to agree, using data the backend shell already holds. No host-side configuration is needed: an installation without the module stops hitting the endpoints by default.

### What the investigation found

**The launcher's docblock was wrong.** It claimed the component hides itself when `/api/ai_assistant/health` returns non-OK. It has not done that since the health effect started coercing every 4xx/5xx (`AiAssistantLauncher.tsx:279-283`) and every network error (`:292-297`) to `setHealthy(true)` — a deliberate change so a transiently-401 health endpoint could not leave the launcher permanently hidden (reasoning preserved at `:261-269`). The only surviving hide signal is a `{ healthy: false }` body, and the server never emits that key: `opencode-handlers.ts:206-224` returns `{ status: 'ok' | 'error' }`, and `health/route.ts` always answers 200 unless it throws.

So visibility was decided entirely by `agentsLoaded && agents.length > 0`. The launcher *did* end up hidden on a disabled install — but only after both requests were already paid for, and a 404 landed in the same `!call.ok` branch as a 500. "Not configured" never covered "module not installed at all".

**The repeats are not per-navigation.** `AppShell` lives in an App-Router layout, so its client subtree survives soft navigation (asserted at `AiAssistantLauncher.tsx:230`). The cause is that nothing is remembered across mounts — `healthy`, `agents`, `agentsLoaded` are plain `useState` with no module-level cache and no react-query, despite `QueryProvider` sitting directly above at `AppShell.tsx:516`. Every hard load, refresh, new tab and login re-probes. The conversation sync additionally keys on `[storageKey]`, so it also re-fires on tenant/org scope switches.

No caching work is included here: gating removes the requests entirely on disabled installs, and on enabled installs a once-per-load probe is correct.

**The shell already had the answer in scope, unused.** `enabledModulesRegistry.ts:21-24` documents the design — the browser never imports the module registry; the server pre-filters `BackendChromePayload.grantedFeatures` instead. `filterGrantsByEnabledModules` drops grants owned by disabled modules and expands a superadmin `*` into `<id>.*` for enabled modules only, so it is sound for superadmins too.

## Changes

- **New** `packages/ui/src/ai/useAiAssistantAvailable.ts` — the shared gate:

  ```ts
  const moduleEnabled = enabledModuleIds === null || enabledModuleIds.has(AI_ASSISTANT_MODULE_ID)
  return moduleEnabled && hasFeature(payload?.grantedFeatures, AI_ASSISTANT_VIEW_FEATURE)
  ```

  - The client enabled-module registry (`getEnabledModuleIds`), populated during client bootstrap from the generated, client-safe `enabled-module-ids.generated.ts`. It registers asynchronously, so `null` means "not known yet", not "absent"; it is read through `useSyncExternalStore` + `subscribeToInjectionRegistryChanges` so the hook updates when registration lands.
  - `ai_assistant.view` from the backend chrome payload, which `BackendChromeProvider` fetches once and module-caches — so the gate costs **no additional request**. It uses `hasFeature`, not `includes`/`Set.has`, per `packages/ui/AGENTS.md`, so an `ai_assistant.*` wildcard grant matches. It is fail-closed while the payload is `null`, which is what keeps the cold-load window quiet.

  This mirrors `ProgressTopBar.tsx:26-33`, which skips its own polling for exactly this reason eleven lines below the launcher's mount site, and `BackendHeaderChrome.tsx:120-123`, which already gates its AI button on the same feature.

- `AiAssistantLauncher.tsx` — split into a gate wrapper plus `AiAssistantLauncherContent` (same shape as `ProgressTopBar`/`ProgressTopBarContent`) so the health and agents effects never mount; an early return after the hooks would not have stopped them. `AppShell.tsx:1493` is untouched.
- `AiChatSessions.tsx` — the provider keeps rendering its children unconditionally (it owns context every backend page needs); only the sync effect is guarded, with `aiAvailable` in the dependency list so it still runs once the chrome payload arrives.
- Docblock and `healthEndpoint` prop doc corrected to describe the real contract. **The health probe itself is deliberately left untouched**, so enabled installs behave identically.

### User impact

- Installation without `ai_assistant`: zero `/api/ai_assistant/*` requests, no per-load console warning.
- Installation with `ai_assistant`: unchanged.
- User without `ai_assistant.view` on an enabled install: launcher stays hidden as before, and the requests are now skipped too.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**

N/A — behavioural bug fix confined to two existing components plus one hook; no contract, schema or API surface changes.

## Testing

Full `validation.commands` sequence from `.ai/agentic.config.json`, in order:

| Command | Result |
|---|---|
| `yarn build:packages` | pass |
| `yarn generate` | pass |
| `yarn build:packages` | pass |
| `yarn i18n:check-sync` | pass — all translation files in sync across 5 locales |
| `yarn i18n:check-usage` | pass (advisory; 3827 pre-existing unused keys, none added) |
| `yarn typecheck` | pass |
| `yarn test` | see note below |
| `yarn build:app` | pass |

`yarn test` note: the aggregate turbo run reports one failed suite in `@open-mercato/cli`, and it is a local artifact rather than a result — a jest worker dies with `SIGSEGV`, a *different* suite each run, while every assertion in the package passes (`1804 passed, 0 failed`). It reproduces with `--maxWorkers=2` and outside the root script's 1 GB heap cap, and the branch touches nothing outside `packages/ui/src/ai/`, so `packages/cli` cannot be implicated. CI ran this package green on the previous head.

`@open-mercato/ui` — the only package this branch changes — is green in full: **234 suites, 1936 tests**.

Also run on this head, all green: `yarn lint` (0 errors), `yarn check:client-boundaries` (exit 0; the new `"use client"` file is a hook, not a page root, so it does not enter the ledger).

`develop` has been merged forward, per the review note — the branch was 73 commits behind, which is what expanded CI to all 15 integration shards. The merge was conflict-free and touched none of the files in this diff.

New unit tests:

- `useAiAssistantAvailable.test.tsx` — enabled + granted; module absent from the registry; feature missing; chrome payload not yet arrived; registry not yet arrived (treated as "unknown", not disabled); `ai_assistant.*` wildcard grant.
- `AiAssistantLauncher.test.tsx` — renders nothing **and** `apiCall` never invoked when unavailable, including via the global ⌘L event; plus a gate-transition case (unavailable at mount, then available) asserting the agents probe fires exactly once. These mock `useAiAssistantAvailable` wholesale rather than driving the real gate — the hook has its own dedicated test file, listed above, which exercises it against real inputs.
- `AiChatSessions.test.tsx` — unavailable: sync never called, children still render, no warn/error; available: synced exactly once with `{ limit: 100 }`; and a gate-transition case that mounts unavailable, flips the gate, and asserts the sync then fires exactly once.

That last case pins the `aiAvailable` entry in the effect's dependency array, which is the line that makes this fix work on an install where the module **is** enabled. The gate is fail-closed while the backend chrome payload is still `null` — the state at mount — so the mount-time run always takes the early exit and only the re-run triggered by the gate opening performs the sync. Verified by mutation: the test fails against `}, [storageKey])` and passes against `}, [aiAvailable, storageKey])`.

Generator-level verification of the disabled path: commenting out `{ id: 'ai_assistant', … }` in `apps/mercato/src/modules.ts` and running `yarn generate` drops `ai_assistant` from `enabled-module-ids.generated.ts` **and** registers zero `/api/ai_assistant/*` routes — the exact 404 source, and the input that now closes the gate. Restored and regenerated afterwards.

## Checklist

- [ ] I have read and accept the Open Mercato Contributor License Agreement (see `docs/cla.md`). — **left for the PR author to tick; an agent must not accept a legal agreement on a human's behalf.**
- [x] I updated documentation, locales, or generators if the change requires it. — none required: no new user-facing strings, no generator inputs touched.
- [x] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required). — **documented, not added:** the behaviour under test is the *absence* of network calls under a module configuration the integration harness does not vary (it runs a single fully-enabled app). The disabled path is covered instead by the generator-level check above plus unit tests asserting the requests are never issued.
- [x] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — not applicable, see Specification.
- [ ] Priority set: this PR carries exactly one `priority-*` label. — **deferred:** this account has `read` on the upstream and cannot set labels. Intended: `priority-medium`.
- [ ] Risk set: this PR carries exactly one `risk-*` label. — **deferred.** Intended: `risk-medium` — it touches the shared backend shell mounted on every `/backend/*` page, though the diff is confined to `packages/ui/src/ai/`.
- [ ] QA routing set. — **deferred.** Intended: `needs-qa`. The change touches `.tsx` under `packages/ui/src/`, so it does not qualify for the automated-verification exemption and cannot take `skip-qa`.

### Design System Compliance

The diff adds no markup — one hook, one wrapper component, and an early `return null`. No JSX element, `className`, colour, size or icon is introduced anywhere in it.

- [x] No hardcoded status colors — none introduced.
- [x] No arbitrary text sizes — none introduced.
- [x] Empty state handled for list/data pages — N/A; no list or data page is touched.
- [x] Loading state handled for async pages — N/A; no async page is touched. The gate resolves from an already-cached payload and renders nothing until it does, matching the launcher's existing pre-agents behaviour.
- [x] `aria-label` on all icon-only buttons — N/A; no button is added, and the existing trigger's labelling is unchanged.
- [x] Uses existing DS components — N/A; no component is rendered that was not rendered before.

## Notes for reviewers

Two things worth a maintainer's eye, neither addressed here on purpose:

1. **The health probe is dead code even on enabled installs.** As traced above, it cannot influence rendering under any response, so it costs one doomed request per hard load everywhere. Removing it (and deprecating `healthEndpoint` / `skipHealthCheck` to no-ops) was scoped out of this PR to keep enabled-install behaviour byte-identical. Happy to follow up if you want it gone.
2. **One behaviour narrowing beyond the disabled case.** Because the feature check is fail-closed, a host that renders `AppShell` *without* `adminNavApi` would no longer show the launcher. `apps/mercato` and the `create-app` template both always pass it, so nothing shipped regresses, and `ProgressTopBar` carries the identical exposure — but it is a real narrowing for a third-party host embedding `AppShell` directly.

No `apps/mercato/src/**` or `packages/create-app/template/**` changes, so no template sync is triggered.

This is a cross-repository PR, so package-preview workflows will not run on it (they are restricted to same-repository branches per `CONTRIBUTING.md`).

## Linked issues

None.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5567"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790168326&installation_model_id=432243&pr_number=5567&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5567&signature=f8146ee13bcac917fdec8448797af34b71ecf22ae25730592f9a53fc9575166a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Frshy:
This account has `read` permission on this repository, so it cannot apply labels, set an assignee, or request a reviewer. Recording the intent here for a maintainer with `triage` to action in one pass.

**Labels**

| Label | Rationale |
|---|---|
| `bug` | Fixes doomed 404 requests fired on every hard load of `/backend/*` when `ai_assistant` is disabled; no new capability. |
| `review` | Ready for code review — not a draft, and not in any other pipeline state. |
| `priority-medium` | Wasted round-trips and a per-load console warning on every affected install. Degrades the experience and pollutes logs, but nothing is broken or unsafe, so not `priority-high`. |
| `risk-medium` | The diff is confined to `packages/ui/src/ai/` (6 files), but it changes rendering conditions in the shared backend shell mounted on every `/backend/*` page — an ordinary single-module change with real reach. Not `risk-low` for that reason; not `risk-high` because no schema, API or contract surface moves. |
| `needs-qa` | The change touches `.tsx` under `packages/ui/src/`, so it does not qualify for the automated-verification exemption in `AGENTS.md` → PR Workflow and cannot take `skip-qa`. |

`skip-qa`, `qa-approved` and `qa-self-verified` are deliberately **not** requested. Per the QA gate this PR cannot merge until a maintainer or QA applies `qa-approved` — the evidence in the PR body does not clear that gate on its own.

No claim label (`in-progress`) is requested: this account could not release one afterwards.

**Assignee** — the PR author.

**Reviewer** — a maintainer who owns the backend shell / `packages/ui` AI surfaces. Naming a role rather than a handle deliberately; routing is yours.

**Suggested QA click-through**, if it helps whoever picks this up:

1. On an install **with** `ai_assistant`: open any `/backend/*` page, confirm the AI pill still renders in the topbar, ⌘L still opens the assistants picker, and DevTools → Network still shows `health`, `ai/agents` and `conversations?limit=100` once per hard load.
2. On an install **without** `ai_assistant` (comment the entry out of `apps/mercato/src/modules.ts`, `yarn generate`): the same pages should show **zero** `/api/ai_assistant/*` requests across a hard reload and a soft navigation, and no `Could not load server conversations` warning in the console.
3. As a user whose role lacks `ai_assistant.view` on an enabled install: launcher hidden as before, and now no requests either.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T00:44:46Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(ui): stop probing AI assistant API when the module is disabled

## 🎯 Summary

This PR stops the backend shell from firing three doomed `/api/ai_assistant/*` requests on every hard load of a `/backend/*` page when the `ai_assistant` module is not enabled. It introduces `useAiAssistantAvailable()` (`packages/ui/src/ai/useAiAssistantAvailable.ts`), a gate that requires two independent signals to agree — the client enabled-module registry and `ai_assistant.view` in the backend chrome payload — and then applies that gate at the two call sites that were mounting unconditionally: `AiAssistantLauncher` (split into a gate wrapper plus `AiAssistantLauncherContent`, so the health and agents effects never mount) and `AiChatSessionsProvider`'s server-conversation sync effect.

Reviewed scope: the six changed files, plus the surrounding contracts they depend on — `BackendChromeProvider`, `rbacService.getEffectiveFeatures`, `resolveEffectiveFeatures`, `filterGrantsByEnabledModules`, `injection-loader`'s enabled-module registry, and the ACL metadata on every `/api/ai_assistant/*` route the PR reasons about.

**This is a well-researched, well-executed change and the analysis in the PR body holds up under checking.** Specific things I verified rather than took on faith:

- All three endpoints the PR is trying to stop calling — `health/route.ts`, `api/ai/agents/route.ts`, `api/ai/conversations/route.ts` — really do declare `requireFeatures: ['ai_assistant.view']`. Gating on that exact feature therefore cannot hide the launcher from anyone who would have received a useful response; it only suppresses requests that were guaranteed to 404 or 403.
- The launcher docblock the PR rewrites was genuinely wrong, and the replacement is accurate. `AiAssistantLauncher.tsx:279-283` and `:292-297` do coerce every non-OK response and every network error to `setHealthy(true)`, so only an explicit `{ healthy: false }` body can hide it — and the server never emits that key.
- `getEnabledModuleIds()` returns a stable reference (`injection-loader.ts:199-201` hands back the same global `Set`), so using it as a `useSyncExternalStore` snapshot does not trip React's "getSnapshot should be cached" infinite-render guard. `subscribeToInjectionRegistryChanges` is likewise a stable module-level function, so React does not resubscribe on every render.
- `useBackendChrome()` returns a safe `{ payload: null }` default outside a provider rather than throwing (`BackendChromeProvider.tsx:96-103`), and the provider nesting in `AppShell.tsx:516-524` puts `BackendChromeProvider` above both new consumers. Neither call site can hit a missing-context error.
- The fail-closed-on-`null`-payload choice does not create a hydration mismatch: the server snapshot and the first client render both see `payload === null`, so both render nothing.
- The gate does not regress the org/tenant scope switch. `BackendChromeProvider.refresh()` never resets `payload` to `null` on a scope change, so `aiAvailable` does not flip false and the launcher does not unmount and re-probe.
- The delayed sync in `AiChatSessionsProvider` is safe: `mergeServerConversations` (`AiChatSessions.tsx:191-214`) preserves local-only sessions rather than replacing state, and the launcher is gated on the same signal, so a user cannot create a session before the sync window opens anyway.
- Repository lesson `header-gated-module-features-need-setup-grants` applies directly here, because this PR makes a topbar surface feature-gated for the first time. It is satisfied: `setup.ts:107-119` grants `ai_assistant.view` to both the `admin` and `employee` default roles, and `__tests__/acl-dependencies.test.ts` fails if that grant is dropped (the dependent features would report missing dependencies).
- Lesson `feature-gated-runtime-helpers-must-use-wildcard-aware` is satisfied — the hook uses `hasFeature`, not `includes`.

The change is also stylistically right for this codebase: it is the same wrapper/content shape as `ProgressTopBar`/`ProgressTopBarContent` (`packages/ui/src/backend/progress/ProgressTopBar.tsx:26-33`), which skips its own polling for exactly this reason eleven lines below the launcher's mount site.

I found one issue worth a round trip, and it is about test coverage rather than about the code being wrong.

## Verdict

❌ **request changes** — one Minor finding. Per this repository's verdict rule (`.ai/skills/om-auto-review-pr/SKILL.md`), a Medium/Minor finding requests changes. Nothing in the diff is incorrect; the single actionable item is that the one line which makes this fix work on an *enabled* install is not pinned by any test, and I confirmed by mutation that removing it leaves the whole `packages/ui` suite green. That is a small, well-defined addition, after which this is an approve.

## 🧪 Validation Gate

Validation source: **GitHub checks** (run [32730010797](https://github.com/open-mercato/open-mercato/actions/runs/32730010797) on head `f9d5532`), with a narrow local fallback for the one configured command CI does not run. Both workflow runs on this head were sitting at `action_required` — this is a cross-repository PR, so fork CI needs per-push maintainer approval and `gh pr checks` reported only `license/cla` and a skipped `[code]smith`. I inspected the diff first (six files, all under `packages/ui/src/ai/`, no workflow, script, or dependency changes) and then approved both runs so this review has real CI data.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | CI `prepare` job, step "Build packages". |
| `yarn generate` | ✅ PASS | CI `prepare` job, step "Prepare generated modules". |
| `yarn build:packages` (rebuild) | ✅ PASS | CI `prepare` job, step "Rebuild packages with generated files". |
| `yarn i18n:check-sync` | ✅ PASS | Not part of the CI workflow, so this one ran locally as a fallback: "All translation files are in sync" across all 5 locales, 56 modules. |
| `yarn i18n:check-usage` | ⏳ PENDING | CI `test` job, still running at the time of this review. |
| `yarn typecheck` | ⏳ PENDING | CI `test` job, still running. |
| `yarn test` | ⏳ PENDING | CI `test` job, still running. Locally, the three affected suites pass in the PR worktree: 3 suites, 18 tests, all green. |
| `yarn build:app` | ✅ PASS | CI `prepare` job, step "Build app". |

Additional gates, all green on this head: `lint` (including the peer-dependency and raw-`console.*` guards), `ds-lint` (head and base, error-level findings enforced), and `audit`. `Mutation tests` completed successfully — note that its allowlist does not cover `packages/ui`, so that green result is not evidence about this diff.

Per the repo-local `om-code-review` extension, the frontend performance gate applies because this diff touches a global provider and the app shell. I ran `yarn check:client-boundaries` locally: exit 0, and the report lists only pre-existing page-root `"use client"` entries. The new `"use client"` file is a hook in `packages/ui/src/ai/`, not a page root, so it does not enter the ledger. The template parity gate is not applicable — nothing under `apps/mercato/src/{app,modules}` or `packages/create-app/template/` is touched.

**CI was still running on this head at the time of this review.** The verdict above covers the code; branch protection and the QA-approval gate hold the actual merge. Checks still pending: `test`, `documents-multi-instance`, and all 15 `ephemeral-integration` shards. A follow-up comment will report the CI outcome.

One thing to be aware of when reading that CI run: this branch is **73 commits behind `develop`**, which is why the integration matrix expanded to all 15 shards instead of the affected-only subset. If an unrelated shard goes red, merging `develop` forward first is the cheapest way to tell a real failure from a stale-base artifact.

## Findings

### 🔹 Minor

**`packages/ui/src/ai/AiChatSessions.tsx:283` — the `aiAvailable` dependency is load-bearing but no test pins it, so removing it ships a silent regression with a green suite.**

Changing the effect's dependency array from `[storageKey]` to `[aiAvailable, storageKey]` is what makes this fix correct on an installation where the module *is* enabled. `aiAvailable` is fail-closed while the backend chrome payload is `null`, which is the state at mount, so the mount-time run always takes the `if (!aiAvailable) return` early exit. Only the re-run triggered by `aiAvailable` flipping to `true` actually performs the sync. Drop `aiAvailable` from the array and server conversations stop loading entirely on every enabled install — a real user-visible regression, silently introduced.

Both new tests set `aiAvailableMock.mockReturnValue(...)` *before* `renderWithProviders`, so the gate never transitions during a test: the "available" case is already `true` on the very first render, and the mount-time run does all the work. The transition the production code depends on is therefore never exercised.

I confirmed this rather than inferring it. With `}, [aiAvailable, storageKey])` changed to `}, [storageKey])` in the PR worktree, the full `packages/ui` suite still reports **1862 passing tests** and the three AI suites report 18/18 green. The mutant survives at whole-package scope.

Suggested fix — one test in `packages/ui/src/ai/__tests__/AiChatSessions.test.tsx`, in the new `ai_assistant availability gate` describe block:

```tsx
it('syncs once the availability gate opens after mount', async () => {
  aiAvailableMock.mockReturnValue(false)
  const { rerender } = renderWithProviders(<Harness agentId="assistant" />)
  await waitFor(() => expect(screen.getByTestId('session-count').textContent).toBe('0'))
  expect(listMock).not.toHaveBeenCalled()

  aiAvailableMock.mockReturnValue(true)
  rerender(<Harness agentId="assistant" />)

  await waitFor(() => expect(listMock).toHaveBeenCalledTimes(1))
  expect(listMock).toHaveBeenCalledWith({ limit: 100 })
})
```

That test fails against `[storageKey]` and passes against `[aiAvailable, storageKey]`, which is exactly the property worth locking down. While you are there, the equivalent transition for `AiAssistantLauncher` (unavailable at mount, then available, then the agents probe fires exactly once) is cheap to add and covers the other half of the change — though the launcher has no dependency-array subtlety, so it is lower risk and genuinely optional.

### 💅 Nit

**`packages/ui/src/ai/index.ts` — `useAiAssistantAvailable` is not re-exported from the `./ai` barrel.** Every other hook in that folder is (`useAiChat`, `useAiShortcuts`, `useAiDock`, `useAiChatSessions`), and `./ai` is the only exported subpath for the folder in `packages/ui/package.json`. Adding one export line would keep the barrel consistent and let another package gate its own AI surface on the same signal without a deep import. Entirely your call — `ProgressTopBar` keeps its equivalent gate private too, so the current state is defensible.

**PR description — the Testing section says "The existing test was updated to supply granted features."** The diff instead mocks `useAiAssistantAvailable` wholesale in `AiAssistantLauncher.test.tsx`. That is a perfectly reasonable choice (the hook has its own dedicated test file), but the description as written could lead a reviewer to believe the launcher tests exercise the real gate. Worth a one-line correction so the body matches the diff.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path. `AiAssistantLauncher` keeps its name, its export from `packages/ui/src/ai/index.ts`, and its `AiAssistantLauncherProps` type; the implementation simply moved into a private `AiAssistantLauncherContent`.
- [x] No function signature changed in a breaking way. `AiAssistantLauncher` moves from destructured parameters to `props: AiAssistantLauncherProps` — the same public type, so callers are unaffected.
- [x] No required type field removed or narrowed. `AiAssistantLauncherProps` is untouched; only two JSDoc blocks on it were corrected.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation. No API file is in the diff.
- [x] Where a contract had to change: not applicable — nothing on the protected-surface list in `BACKWARD_COMPATIBILITY.md` is touched.

One behavioural narrowing is worth recording explicitly, and the author already disclosed it in "Notes for reviewers": because the feature check is fail-closed, a third-party host that renders `AppShell` **without** `adminNavApi` will no longer show the launcher. I checked and agree this is acceptable rather than a blocker — `apps/mercato` and the `create-app` template both always pass it, so nothing shipped in this repository regresses, and `ProgressTopBar` already carries the identical exposure on `develop`. It is a conscious, documented trade-off, not an oversight. The related coupling — a failing `/api/auth/admin/nav` now also hides the AI launcher — is likewise pre-existing for `ProgressTopBar` and only manifests when the whole sidebar is already broken.

## 🧪 Test Coverage

What the new tests actually cover, and it is a genuinely good set:

- `useAiAssistantAvailable.test.tsx` covers all six meaningful states of the gate: enabled and granted; module absent from the registry despite the grant being present; the grant missing despite the module being enabled; the chrome payload not yet arrived; the registry not yet arrived (asserting the deliberate "unknown, not disabled" reading of `null`); and an `ai_assistant.*` wildcard grant, which is the case a plain `includes()` would have missed. That last one is the repository's own `feature-gated-runtime-helpers-must-use-wildcard-aware` lesson, tested.
- `AiAssistantLauncher.test.tsx` asserts both halves of the launcher claim — that nothing renders **and** that `apiCall` is never invoked, including via the global ⌘L event. Asserting the absence of the request rather than only the absence of the button is the right assertion here, since suppressing the request is the entire point of the PR.
- `AiChatSessions.test.tsx` asserts the sync is skipped with no warning or error logged, that children still render (the provider owns context every backend page needs), and that the available path syncs exactly once with `{ limit: 100 }`.

The gap is the one in the Minor finding above: no test drives a transition of the gate, so the `aiAvailable` dependency that makes the enabled-install path work is unpinned.

On integration coverage, I accept the author's documented rationale. The disabled-module branch cannot be exercised by a harness that runs a single fully-enabled app, and the generator-level check described in the PR body (dropping `ai_assistant` from `apps/mercato/src/modules.ts`, regenerating, and confirming both that `enabled-module-ids.generated.ts` loses the id and that zero `/api/ai_assistant/*` routes register) is a reasonable substitute for a path that is otherwise untestable here. The feature-missing branch — an enabled install where the user lacks `ai_assistant.view` — *is* reachable in the harness and could carry an integration test, but it is already covered at the unit level, so I am not asking for it.

Finally, a follow-up rather than a finding: the PR is right that the health probe is now dead code even on enabled installs, since no response it can return influences rendering, and it still costs one doomed request per hard load everywhere. Deliberately leaving it alone to keep enabled-install behaviour byte-identical was the correct call for this PR's scope. It is worth its own issue — removing the probe and deprecating `healthEndpoint` / `skipHealthCheck` to no-ops — rather than being folded in here.

pkarw:

- 🚀 `merge-queue` — the re-review of head `51c917b` approved: the one Minor from the first pass (the unpinned `aiAvailable` effect dependency at `AiChatSessions.tsx:283`) is fixed and I re-verified it by mutation — reverting the dependency now fails exactly the new gate-transition test and nothing else — and both nits are done, so `changes-requested` no longer applies. `needs-qa` stays on, so the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`.
- 🐛 `bug` — this fixes three doomed `/api/ai_assistant/*` requests plus a per-load console warning fired on every hard load of a `/backend/*` page when the module is disabled; it adds no new capability.
- 🧪 `needs-qa` — the change touches `.tsx` files under `packages/ui/src/` that mount on every backend page, so it does not qualify for the automated-verification exemption and cannot take `skip-qa`; a human should confirm the launcher and the AI chat still behave normally on an enabled install.
- 🔹 `priority-medium` — wasted round-trips and a per-load console warning degrade every affected install, but nothing is broken or unsafe, so this belongs in the current cycle rather than ahead of it.
- 🟡 `risk-medium` — the diff is confined to `packages/ui/src/ai/` and is well covered by unit tests, but the two gated components mount inside the shared backend shell on every `/backend/*` page, so the blast radius of a mistake would be wide.
- 🤖 `in-progress` — this review's claim lock; released as soon as the review, labels and QA instructions are posted, and swapped for `ci-monitoring` while CI on this head finishes.

pkarw:

Two practical notes for that pass: this branch is 73 commits behind `develop`, which is why CI expanded to all 15 integration shards — merging `develop` forward first will both shrink that matrix and rule out stale-base noise. And since this is a cross-repository PR, workflow runs sit at `action_required` until a maintainer approves them on each push; I approved the runs on `f9d5532` so this review had real CI data.

pkarw:

One Minor finding — the `aiAvailable` effect dependency at `AiChatSessions.tsx:283` is unpinned by tests, confirmed by mutation at whole-package scope — plus two nits. No blockers or majors; the design, the ACL reasoning, and the backward-compatibility surface all checked out. Under this repo's verdict rule a Medium/Minor finding requests changes, so that one test is all that stands between this and an approve.

CI was still running at review time, so `in-progress` has been swapped for `ci-monitoring`: the review work is finished and this PR is free for anyone to act on, and a CI-result comment will follow.

pkarw:

CI on head `f9d5532` has settled: **[run 32730010797](https://github.com/open-mercato/open-mercato/actions/runs/32730010797) completed successfully — every job green.** That covers `prepare`, `lint`, `ds-lint`, `audit`, `audit-scope`, `test`, `documents-multi-instance`, `merge-coverage`, and all 15 `ephemeral-integration` shards (`docker-build` skipped as expected). The separate `Mutation tests` run also passed, though its allowlist does not cover `packages/ui`, so it says nothing about this diff either way.

This fills in the three gate rows that were pending when I posted the review. From the `test` job's step list: **Checking types** ✅, **Test** ✅, **Test scripts** ✅, **Check i18n usage** ✅ (advisory — the step carries `continue-on-error: true`). Worth noting that the full 15-shard integration matrix went green even though this branch is 73 commits behind `develop`, so there is no stale-base noise to disentangle here.

Two useful extras that also ran green in that job and close out gates my review had flagged as needing separate attention: **Check create-app template parity** ✅ (confirming no template sync is owed) and **Check agent instruction budget** ✅.

One correction to my review body, for the record: I wrote that `yarn i18n:check-sync` is "not part of the CI workflow" and ran it locally as a fallback. That was wrong — CI does run it, as the `Check i18n sync` step, which invokes `yarn tsx scripts/i18n-check-sync.ts` directly rather than through the `i18n:check-sync` script name, which is why my scan of the workflow file missed it. The step passed in CI, matching my local run, so no conclusion in the review changes.

**The verdict is unchanged: still `changes-requested`.** Green CI was never in question here — the single Minor finding is a test-coverage gap (the `aiAvailable` dependency at `AiChatSessions.tsx:283` is unpinned, confirmed by mutation), and a green suite is exactly the symptom, not a refutation. Removing that dependency would also leave CI green, which is the point.

`ci-monitoring` has been removed; no further follow-up will come from this agent on this head.

Frshy:
Thanks @pkarw — all three items addressed in `51c917b`, plus the `develop` merge-forward you suggested.

**Minor — the unpinned `aiAvailable` dependency.** Fixed, and I reproduced your mutation first rather than taking it on faith: with `}, [aiAvailable, storageKey])` changed to `}, [storageKey])`, `packages/ui` stayed green here too. Added essentially your suggested test to the `ai_assistant availability gate` block in `AiChatSessions.test.tsx` — mount unavailable, assert `listMock` was never called, flip the mock, `rerender`, assert the sync then fires exactly once with `{ limit: 100 }`. Re-ran the mutation against it: **that test, and only that test, fails** (`1 failed, 9 passed`), and the suite is green again on the restored source. The mutant is dead.

I also took the optional half — `AiAssistantLauncher.test.tsx` now has the equivalent transition case (unavailable at mount → available → exactly one `/api/ai_assistant/ai/agents` call). You are right that the launcher has no dependency-array subtlety, but the gate being fail-closed means *every* enabled install mounts unavailable first and only then flips, so it is worth having that path asserted rather than only the steady states.

**Nit — barrel export.** Done: `useAiAssistantAvailable` is now exported from `packages/ui/src/ai/index.ts`, along with `AI_ASSISTANT_MODULE_ID` and `AI_ASSISTANT_VIEW_FEATURE` so a consumer can name the same signals without a deep import.

**Nit — PR description.** Corrected. The Testing section now says the launcher and sessions tests mock `useAiAssistantAvailable` wholesale and points at `useAiAssistantAvailable.test.tsx` as the file that drives the real gate, so no reviewer is left believing the launcher tests exercise it. I also documented the new transition test and the mutation result there.

**Merge-forward.** Done — conflict-free, and it touched none of the files in this diff. The branch is now current with `develop`, so the integration matrix should narrow back to the affected subset.

**Local gates on this head** (local runner; no compose container in this worktree): `build:packages` → `generate` → `build:packages`, `typecheck`, `lint` (0 errors), `i18n:check-sync` ("All translation files are in sync"), `i18n:check-usage`, `check:client-boundaries` (exit 0), `build:app` — all pass. `packages/ui` in full: 234 suites, 1936 tests, green.

One thing to flag honestly rather than bury: the aggregate `yarn test` reports one failed suite in `@open-mercato/cli`, and I do **not** believe it is real. A jest worker dies with `SIGSEGV` — a *different* suite each run — while every assertion in the package passes (`1804 passed, 0 failed`). It reproduces with `--maxWorkers=2` and outside the root script's 1 GB heap cap, so it reads as a local worker crash. `git diff upstream/develop HEAD` is seven files, all under `packages/ui/src/ai/`, so `packages/cli` cannot be implicated, and CI ran it green on the previous head. Worth watching in the CI run on this head to confirm.

Since this is a cross-repository PR, the run on the new head will need your approval again. I tried to re-request review and got a 404 — this account has `read` on the upstream, so it cannot set reviewers. Treating this comment as the re-review request instead; the ball is with you when you next pick it up.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-28T07:24:12Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review: fix(ui): stop probing AI assistant API when the module is disabled

**Re-review** of head `51c917b` (previous review: `f9d5532`, ❌ changes requested — one Minor plus two nits).

## 🎯 Summary

This PR stops the backend shell from firing three doomed `/api/ai_assistant/*` requests on every hard load of a `/backend/*` page when the `ai_assistant` module is not enabled. It introduces `useAiAssistantAvailable()` (`packages/ui/src/ai/useAiAssistantAvailable.ts`), a gate that requires two independent signals to agree — the client enabled-module registry and `ai_assistant.view` in the backend chrome payload — and applies that gate at the two call sites that were mounting unconditionally: `AiAssistantLauncher` (split into a gate wrapper plus `AiAssistantLauncherContent`) and `AiChatSessionsProvider`'s server-conversation sync effect.

The first pass established that the design, the ACL reasoning, and the backward-compatibility surface all hold up; that analysis is in the previous review and I am not repeating it here. **The two production files are byte-identical to the head I reviewed then** — `git diff f9d5532 51c917b -- packages/ui/src/ai/AiAssistantLauncher.tsx packages/ui/src/ai/AiChatSessions.tsx` is empty — so this re-review is scoped to what actually changed: commit `e0b53ca` (two new tests plus a barrel export) and the `develop` merge-forward in `51c917b`.

Everything I asked for is done, and I verified each item rather than reading the reply and taking it on faith.

**The Minor is fixed, and I re-ran the mutation myself.** With `}, [aiAvailable, storageKey])` reverted to `}, [storageKey])` at `packages/ui/src/ai/AiChatSessions.tsx:283`, the suite now reports `1 failed, 9 passed` — and the one failure is exactly the new `syncs once the availability gate opens after mount` case, at `AiChatSessions.test.tsx:367`, where `expect(listMock).toHaveBeenCalledTimes(1)` never resolves. Restored, the suite is green again. The mutant that survived at whole-package scope last time is dead. That matters because `aiAvailable` is fail-closed while the chrome payload is `null` — the state at mount on *every* install — so the mount-time run always takes the early exit and the sync only ever happens on the re-run the dependency triggers. Dropping it would have silently stopped server conversations from loading on every enabled install.

**The optional launcher half was taken too, and it also pins something real.** I mutated `AiAssistantLauncher.tsx` by deleting the two gate lines (`const aiAvailable = useAiAssistantAvailable()` and `if (!aiAvailable) return null`) so the wrapper renders `AiAssistantLauncherContent` unconditionally: `3 failed, 1 passed`. The new `probes exactly once when the gate opens after mount` test fails at line 117 on `expect(apiCallMock).not.toHaveBeenCalled()`, alongside the two steady-state cases. The gate is now pinned from both directions — no probe before it opens, exactly one `/api/ai_assistant/ai/agents` call after — and the `toHaveLength(1)` assertion is not vacuous, because the `waitFor` above it requires the probe to have actually landed.

**Both nits are addressed.** `useAiAssistantAvailable`, `AI_ASSISTANT_MODULE_ID` and `AI_ASSISTANT_VIEW_FEATURE` are now exported from `packages/ui/src/ai/index.ts`, which I checked introduces no duplicate export name in that barrel and no new transitive dependency (the barrel already reaches `BackendChromeProvider` through `AiAssistantLauncher`). The Testing section of the PR body now states plainly that the launcher and sessions tests mock the hook wholesale, and points at `useAiAssistantAvailable.test.tsx` as the file that drives the real gate, so no reader is left with the wrong impression.

**The merge-forward is clean and inert with respect to this change.** The merged range touches none of the files this fix depends on — `AppShell.tsx`, `BackendChromeProvider.tsx`, `injection-loader.ts` and `security/features.ts` are all unchanged across it — so nothing about the gate's inputs moved underneath it. I also re-checked the one indirect consumer the gate could have broken: the `?openAiConversation=<id>` deep link (`AiAssistantLauncher.tsx:229`, covered by `TC-AI-sharing-06-deep-link.spec.ts`). It reads the URL param from inside `AiAssistantLauncherContent`, which now mounts a little later, but nothing strips the param before the launcher's own `router.replace` at line 273 does — so the deep link still resolves, just after the chrome payload arrives instead of before it.

I found nothing new worth a round trip.

## Verdict

✅ **approve** — no blockers, majors, minors or nits. The single Minor from the first pass is fixed and mutation-verified at the exact line it was about, both nits are done, the merge-forward does not disturb the gate's inputs, and the production code is unchanged from the head that already passed the design, ACL and backward-compatibility review. This repository requests changes on Medium/Minor findings; there are none left.

`needs-qa` stays on, so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Manual QA instructions follow in a separate comment.

## 🧪 Validation Gate

Validation source: **GitHub checks** (run [33046348915](https://github.com/open-mercato/open-mercato/actions/runs/33046348915) on head `51c917b`), with a narrow local fallback for the one thing CI cannot tell me — whether the new tests kill the mutants they claim to. Both workflow runs on this head were sitting at `action_required`, since a cross-repository PR needs per-push maintainer approval; I inspected the diff first (seven files, all under `packages/ui/src/ai/`, no workflow, script or dependency changes) and then approved both runs so this review has real CI data.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | CI `prepare` job, step "Build packages". The whole `prepare` job is green on this head. |
| `yarn generate` | ✅ PASS | CI `prepare` job, step "Prepare generated modules". |
| `yarn build:packages` (rebuild) | ✅ PASS | CI `prepare` job, step "Rebuild packages with generated files". |
| `yarn i18n:check-sync` | ✅ PASS | Ran locally on the PR worktree: "All translation files are in sync", 5 locales, 57 modules. CI also covers it, in the `test` job as `yarn tsx scripts/i18n-check-sync.ts` (step "Check i18n sync"), which is still running — correcting my previous review, which said this command was outside CI. It is in CI; it is just invoked by script path rather than through the `yarn i18n:check-sync` alias. |
| `yarn i18n:check-usage` | ⏳ PENDING | CI `test` job, still running. Note it is declared `continue-on-error: true` in `ci.yml`, so it is advisory and cannot fail the job either way. |
| `yarn typecheck` | ⏳ PENDING | CI `test` job, step "Checking types", still running. On PRs it is turbo-scoped `--filter=[origin/develop]...`, which selects `@open-mercato/ui` and its dependents — the right scope for this diff. |
| `yarn test` | ⏳ PENDING | CI `test` job, still running, same turbo scoping. Locally I ran the three affected suites in the PR worktree: **20 tests, 3 suites, all green**, plus the two mutations described above. |
| `yarn build:app` | ✅ PASS | CI `prepare` job, step "Build app". |

Additional gates on this head: `lint` ✅ (including the peer-dependency and raw-`console.*` guards), `ds-lint` ✅ (head and base, error-level findings enforced), `audit` ✅, `audit-scope` ✅, `docker-build` skipped as expected. The separate `Mutation tests` workflow run completed successfully, but its allowlist is frozen to `packages/shared`, so that green result says nothing about this diff either way — which is precisely why I ran the two mutations by hand.

I also ran the full `@open-mercato/ui` suite locally: **1900 tests passed, 0 failed**, across 228 suites. Seven `AiChat*` suites failed to *load* in my worktree with `Cannot find module '@open-mercato/ai-assistant/modules/ai_assistant/lib/agent-transport'` — that is my environment, not the branch: I installed with `--mode=skip-build`, so `packages/ai-assistant/dist` does not exist here. None of those suites import anything this PR changes, and CI's `test` job runs them properly against built packages.

**CI is still running on this head at the time of this review**, and per this repository's review contract the verdict does not wait for it. Checks still pending: `test`, `documents-multi-instance`, and all 15 `ephemeral-integration` shards. Branch protection and the QA-approval gate hold the actual merge; a follow-up comment will report the CI outcome. One observation for whoever reads that run: the merge-forward did **not** narrow the integration matrix — all 15 shards are running again on this head, so the earlier "73 commits behind" theory was not the whole story about the shard fan-out. That is a CI-configuration question, not something this PR should carry.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path. `e0b53ca` is purely additive on the barrel: `useAiAssistantAvailable`, `AI_ASSISTANT_MODULE_ID` and `AI_ASSISTANT_VIEW_FEATURE` are new exports from `@open-mercato/ui/ai`, and nothing existing was removed or renamed. `AiAssistantLauncher` keeps its name, export and `AiAssistantLauncherProps` type.
- [x] No function signature changed in a breaking way. `AiAssistantLauncher` takes `props: AiAssistantLauncherProps` instead of destructured parameters — the same public type, so callers are unaffected.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation. No API file is in the diff.
- [x] No event or message name renamed or removed; no payload field removed. `AI_ASSISTANT_LAUNCHER_OPEN_EVENT` is untouched, and its one external consumer (`AiAssistantSettingsPageClient.tsx:8`) only ever runs on an install where the module is enabled.
- [x] Where a contract had to change: not applicable — nothing on the protected-surface list in `BACKWARD_COMPATIBILITY.md` is touched.

The one behavioural narrowing recorded in the first review still stands and is still acceptable: because the feature check is fail-closed, a third-party host rendering `AppShell` without `adminNavApi` will no longer show the launcher. `apps/mercato` and the `create-app` template both always pass it, and `ProgressTopBar` already carries the identical exposure on `develop`.

## 🧪 Test Coverage

The gap from the first review is closed, and closed in the way that matters — by a test that demonstrably fails without the line it protects.

- `AiChatSessions.test.tsx` — the new `syncs once the availability gate opens after mount` mounts the harness with the gate closed, asserts `listMock` was never called, flips the mock, `rerender`s through the same provider wrapper (React Testing Library reuses the original `wrapper`, so this is a re-render and not a remount — the distinction the test depends on), and asserts the sync then fires exactly once with `{ limit: 100 }`. Mutation-verified in both directions, as described in the Summary.
- `AiAssistantLauncher.test.tsx` — the new `probes exactly once when the gate opens after mount` covers the other half: no `apiCall` at all before the flip, exactly one `/api/ai_assistant/ai/agents` call after it. Deleting the gate fails three of the file's four tests.
- Mock hygiene in both files is correct: each `describe` resets `aiAvailableMock` in its own `beforeEach` and sets an explicit return value per test, so no test inherits a gate state from its neighbour.
- The pre-existing coverage is unchanged and still good: `useAiAssistantAvailable.test.tsx` drives the real hook across all six meaningful states, including the `ai_assistant.*` wildcard grant that a plain `includes()` would have missed.

On integration coverage I continue to accept the documented rationale: the disabled-module branch cannot be exercised by a harness that runs a single fully-enabled app, and the generator-level check in the PR body is a reasonable substitute.

One thing remains outstanding, and deliberately so: **the health probe is dead code even on enabled installs.** As traced in the first review, no response it can return influences rendering, so it still costs one doomed request per hard load everywhere. Keeping it here to leave enabled-install behaviour byte-identical was the right scoping call. It wants its own issue — remove the probe and deprecate `healthEndpoint` / `skipHealthCheck` to no-ops — and I could not find an existing one, so I will offer to file it rather than fold it into this PR.

Nice work on the follow-up pass, and thank you for reproducing the mutation instead of asserting the fix.

pkarw:

Scope note before you start: the diff is seven files, all under `packages/ui/src/ai/`, and it adds a *gate* in front of two components that already existed. So the whole QA question is "does anything that used to appear or work still appear and work, and did the pointless network traffic go away". Nothing new is rendered; there is no new screen to find.

### P0 — Permission gating of the AI surfaces

The launcher and the conversation sync are now gated on the `ai_assistant.view` feature, resolved from the backend chrome payload. This is a permission boundary that did not exist before, so it is the thing most worth a human's eyes.

**Where to click**
- `/backend` (any backend page — the launcher lives in the topbar and the sessions provider wraps the whole shell), signed in as an **admin**
- `/backend/settings/roles` — create or edit a role that does **not** grant `ai_assistant.view`, assign it to a second test user
- `/backend`, signed in as that second user

**What to verify**
- As admin: the AI icon is present in the topbar, and `⌘L` / `Ctrl+L` opens the "AI assistants" picker with the expected agents listed — exactly as on `develop`.
- As the user without `ai_assistant.view`: the AI icon is absent, `⌘L` does nothing, and the app is otherwise completely normal — every other page, menu and table still works.
- With DevTools → Network filtered to `ai_assistant`, hard-reload `/backend` as that same restricted user: **zero** requests to `/api/ai_assistant/health`, `/api/ai_assistant/ai/agents` and `/api/ai_assistant/ai/conversations`. Before this PR all three fired and returned 403/404.
- Console for that restricted user shows **no** `Could not load server conversations` / `Failed to load server conversations` warning on page load.

**What can go wrong**
- The launcher fails to appear for an admin who legitimately has the feature — that would be the fix over-reaching. Check both a plain admin and a superadmin (whose grants come through a `*` wildcard expansion, a distinct code path).
- The launcher appears for the restricted user anyway — the gate not taking effect.
- The launcher flickers in and then disappears (or vice versa) during load. A brief "not there yet" window on cold load is expected and correct; a *flip-flop* is not.
- Anything else on the page breaking for the restricted user — the sessions provider still has to render its children even when the gate is closed, so a blank backend page for that user would be a serious regression.

### P1 — AI assistant behaviour on a normal, enabled install

On an install with the module enabled and the feature granted, behaviour is supposed to be **identical** to `develop`. This block is a regression sweep of the AI surfaces, not a test of anything new.

**Where to click**
- `/backend` → AI icon in the topbar → pick an agent → send a message
- The same, opened via the `⌘L` / `Ctrl+L` shortcut instead of the icon
- `/backend/settings/ai-assistant` → the launcher card's "open assistant" action
- A shared conversation deep link: `/backend?openAiConversation=<conversation-id>` (grab an id from an existing conversation via the share button)
- The tenant / organization scope switcher in the topbar, while a chat session is open

**What to verify**
- The picker lists the same agents as before, is searchable by label/description/id, and launching one opens the chat pane and streams a reply.
- Previously created conversations are still listed — this is the part the `aiAvailable` effect dependency protects, so it is worth an explicit look rather than a glance. Reload the page and confirm your earlier sessions come back.
- The deep link opens the shared conversation and strips `?openAiConversation=` from the URL afterwards. It may open a beat later than it used to (the launcher now waits for the backend chrome payload); it must still open.
- After switching organization or tenant, the chat sessions list re-scopes correctly and the launcher stays visible.
- With a fresh browser profile / after a hard logout-login, the launcher still appears — the gate reads a payload that is fetched once per load, so first-load behaviour is the interesting case.

**What can go wrong**
- Server conversations silently stop loading on an enabled install (the list is empty after a reload even though conversations exist). This is the exact regression the new gate-transition test pins; a manual confirmation is still worth having.
- The deep link lands on `/backend` with the chat pane never opening.
- The launcher disappears after an org/tenant switch.

### P1 — Perceived performance and network noise on cold load

**Where to click**
- Hard-reload `/backend` (empty cache, DevTools Network open, filter `ai_assistant`), as an admin and then as the restricted user
- Repeat on a mobile viewport (375×812)

**What to verify**
- Admin: at most one `health`, one `ai/agents` and one `ai/conversations` request per hard load — not more. Soft navigation between backend pages should add none.
- Restricted user: none at all.
- The topbar renders its other controls immediately; the AI icon appearing a fraction later is expected. Nothing else in the shell should shift or stall waiting for it.
- Screenshot the topbar on cold load for both users and attach them as evidence.

**What can go wrong**
- Duplicated probes (two `ai/agents` calls per load) — the gate opening more than once.
- A visible layout jump in the topbar as the icon appears.

### Not manually testable — flagged rather than skipped

The headline case, an installation with `ai_assistant` **removed** from `apps/mercato/src/modules.ts`, cannot be exercised on a normal running app: the module list is compiled in at build time, so reproducing it means editing `modules.ts`, running `yarn generate`, and rebuilding. If you want to cover it, that is the recipe and the expected result is zero `/api/ai_assistant/*` requests plus no AI icon; otherwise the closest observable check is the restricted-user path in **P0**, which exercises the same gate through its other input. Please do not sign the PR off on the disabled-module claim alone — sign it off on "nothing regressed for users who should have AI, and the noise is gone for users who should not".

### Pass/fail

pkarw:

Re-review of head `51c917b`. All three items from the first pass are done and verified rather than taken on trust: the Minor — the unpinned `aiAvailable` effect dependency at `AiChatSessions.tsx:283` — is fixed, and re-running the mutation here fails exactly the new gate-transition test and nothing else, so the mutant is dead. I also mutated the launcher gate itself (deleting `if (!aiAvailable) return null`), which fails three of that file's four tests, so the wrapper is pinned from both directions. Both nits are addressed, and the `develop` merge-forward touches none of the files this fix depends on.

No blockers, majors, minors or nits remain, so the label moved from `changes-requested` to `merge-queue`. `needs-qa` stays on: the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`, and the manual QA instructions are in the comment above.

autofix: skipped (not my PR — re-run with --autofix to fix it here). Nothing needed fixing on this pass in any case.

CI on this head was still running at review time — I approved both `action_required` fork runs so it could start — so `ci-monitoring` is on while I follow up with the result. Anyone else may act on this PR now; the lock is gone.

pkarw:

CI on head `51c917b` has settled: **[run 33046348915](https://github.com/open-mercato/open-mercato/actions/runs/33046348915) is green — every job passing.** That covers `prepare`, `lint`, `ds-lint`, `audit`, `audit-scope`, `test`, `documents-multi-instance`, `merge-coverage`, and all 15 `ephemeral-integration` shards (`docker-build` skipped as expected). `license/cla` passes. The separate `Mutation tests` run also completed successfully, but its `mutate` job skipped — the allowlist is frozen to `packages/shared` — so it says nothing about this diff either way, which is why I ran the two mutations against `packages/ui` by hand during the review.

This fills in the four rows that were pending in the review's validation table. All eight configured `validation.commands` are now confirmed green on this exact head:

| Command | Status | Source |
|---------|--------|--------|
| `yarn build:packages` | ✅ PASS | `prepare` — "Build packages" |
| `yarn generate` | ✅ PASS | `prepare` — "Prepare generated modules" |
| `yarn build:packages` (rebuild) | ✅ PASS | `prepare` — "Rebuild packages with generated files" |
| `yarn i18n:check-sync` | ✅ PASS | `test` — "Check i18n sync"; also confirmed locally |
| `yarn i18n:check-usage` | ✅ PASS | `test` — "Check i18n usage" (advisory: `continue-on-error: true`) |
| `yarn typecheck` | ✅ PASS | `test` — "Checking types" |
| `yarn test` | ✅ PASS | `test` — "Test" |
| `yarn build:app` | ✅ PASS | `prepare` — "Build app" |

**One thing worth recording, because the first attempt was not green.** `ephemeral-integration (3/15)` failed on the initial attempt with `143 passed, 1 flaky, 1 failed`. The failure was `packages/cli/src/lib/__integration__/TC-INT-007.spec.ts` ("CLI official module install and eject flows"), and it was infrastructure rather than a test signal: that spec shells out to `yarn install` (`TC-INT-007.spec.ts:27`), corepack tries to resolve `yarn/latest` from `registry.npmjs.org`, and the runner's egress timed out — `connect ETIMEDOUT 104.16.0.34:443` plus `EHOSTUNREACH` on the IPv6 Cloudflare addresses. It hit the `133:3` case (which recovered on retry, hence the flaky count) and the `163:3` case (which exhausted its retries). Nothing in that spec is reachable from this diff, which is seven files all under `packages/ui/src/ai/`. I re-ran the failed jobs and the shard passed clean, so the diagnosis holds.

The approve verdict stands unchanged — CI confirms it rather than altering it. `ci-monitoring` is removed; this PR sits in `merge-queue` with `needs-qa`, and the QA-approval gate holds the merge until a QA reviewer works through the [manual QA instructions](https://github.com/open-mercato/open-mercato/pull/5567#issuecomment-5449749785) and adds `qa-approved`. No further follow-up will come from this agent.

PR #5568, opened 2026-08-24
fix(auth): validate reset token before rendering the password form (#5533)

### LINKED ISSUE #5533: Password-reset page renders the form for an already-used token and only rejects on submit
Noticed during manual QA of #4471; unrelated to that PR, which does not touch this page.

## What happens

`/reset/<token>` renders the full "Set a new password" form for a token that has already been consumed. The user fills in both password fields, submits, and only then learns the token is dead — the server correctly answers `400 Invalid or expired token` and the message appears above the form.

## Why it is worth fixing anyway

Security is fine: the rejection is server-side and cannot be bypassed (replaying the same token straight at `POST /api/auth/reset/confirm` also returns 400). This is purely about not asking someone to type a password twice for nothing — and about not making a used link look like a working one.

## Suggested fix

Validate the token when the page loads and render a terminal state ("this link has already been used — request a new one", with a link to the reset form) instead of the input form. `packages/core/src/modules/auth/frontend/reset/[token]/page.tsx`.

adeptofvoltron:
Closes #5533

## 🎯 Goal
`/reset/<token>` no longer asks someone to type a new password twice before telling them the link is dead. A consumed or expired reset link now renders a terminal state on load — "This reset link is no longer valid", with a button back to the request form — instead of a working-looking form that can only fail on submit.

## 🔍 Problem
The reset page rendered the full "Set a new password" form for any token, including one that had already been consumed. The only token check lived behind `POST /api/auth/reset/confirm`, which requires a password, so the user learned the link was dead only after filling in both password fields and submitting. Security was never at stake — the server-side rejection is correct and cannot be bypassed — this is purely about not wasting the user's effort and not making a used link look like a working one.

## 🔍 Root Cause
`packages/core/src/modules/auth/frontend/reset/[token]/page.tsx` rendered the form unconditionally, and there was no read-only endpoint the client could call to learn a token's state. `AuthService.confirmPasswordReset` is the only place that inspects `usedAt` / `expiresAt`, and it consumes the token as part of the same call, so it cannot serve as an on-load check.

## What Changed
- **`services/authService.ts`** — new `isPasswordResetTokenValid(token)`: hashes the raw token, looks the row up by hash only, and reports whether it exists, is unused, and has not expired. It performs no `nativeUpdate` and no `flush`, so checking a token never consumes it.
- **`api/reset/validate.ts`** (new) — `POST /api/auth/reset/validate`, unauthenticated and IP rate-limited (`RATE_LIMIT_RESET_VALIDATE_*`, defaults 10 per 5 minutes), returning `{ ok: true, valid: boolean }`. A malformed token is reported as simply not valid, and the response never distinguishes unknown, used, and expired tokens — the caller learns nothing it could not already learn by posting the same token to `/api/auth/reset/confirm`, so this adds no enumeration surface.
- **`data/validators.ts`** — `validatePasswordResetTokenSchema` (token-only), mirroring `confirmPasswordResetSchema`'s `min(10)` bound.
- **`frontend/reset/[token]/page.tsx`** — checks the token on mount and renders one of three states: a spinner while checking, the terminal card when the token is dead, and the existing form otherwise. The check **fails open**: if the endpoint is unreachable or errors, the form still renders, because `/api/auth/reset/confirm` remains the authoritative gate and an outage must not strand a user holding a good link.
- **i18n** — four new `auth.reset.token.*` keys across all five locales (en, pl, es, de, ko).

## 🧪 Tests
- `yarn test` — 11465 passed / 1427 suites, 2 skipped, 0 failed.
- `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn lint`, `yarn build:app` — all clean; lint reports only pre-existing warnings in unrelated example modules.
- `services/__tests__/authService.test.ts` — four cases locking in that `isPasswordResetTokenValid` accepts a live token *without* calling `nativeUpdate` or `flush`, rejects used and expired tokens, and looks up by hashed token only.
- `api/__tests__/reset-validate.test.ts` (new) — the route reports live/dead tokens, treats a malformed token as invalid without reaching the auth service, rate-limits before any DB work, and returns a body with no keys beyond `ok` and `valid`.
- `__tests__/resetWithTokenPage.test.tsx` — updated for the new mount-time check, plus new cases for the terminal state (form absent, "Request a new link" points at `/reset`, no confirm call is ever made) and for the fail-open path when the check itself throws.
- `__integration__/TC-AUTH-007.spec.ts` — rewritten: dead tokens now render the terminal state before any form exists, and the terminal state navigates back to `/reset`.
- `__integration__/TC-AUTH-006.spec.ts` — rewritten to pull a real token out of the captured reset email, so the suite now covers the live-token path end to end (form renders, submit succeeds, redirect to `/login`) *and* the replayed-token path (terminal state), which the previous invalid-token shortcut never exercised.

## 💥 Breaking Changes
None. `POST /api/auth/reset/validate` is a purely additive route; no existing endpoint, response shape, DB schema, or exported signature changed. `AuthService` gains a method and loses none.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5568"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790168901&installation_model_id=432243&pr_number=5568&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5568&signature=1a14b147ad1cd47174635377913ddb4b4bb314001e6928729e6a7889951d2c4d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:

- 🔍 `review` — the autofix chain finished with the full validation gate green, so the PR opens ready for code review rather than as a draft.
- 🐛 `bug` — issue #5533 reports broken behavior on an existing page: a consumed reset link rendered a working-looking form and only rejected the user on submit.
- 🧪 `needs-qa` — the change is user-facing and adds three distinct render states to `/reset/<token>` (checking, terminal, form), so a human should exercise the live-token, used-token, and endpoint-outage paths in a browser; the automated-verification exemption does not apply because the diff touches a `.tsx` page and adds an API route.
- 🔽 `priority-low` — the reporter classified this as low priority and the underlying rejection was always correct and server-side, so nothing is broken from a security or data-integrity standpoint; this only saves the user from typing a password twice for nothing.
- ⚠️ `risk-high` — the diff lands inside the auth module and introduces a new unauthenticated endpoint (`POST /api/auth/reset/validate`), and the label taxonomy puts any change to auth or a shared contract surface at high risk regardless of how contained the behavior change is; reviewers should confirm the endpoint neither consumes a token nor distinguishes unknown, used, and expired tokens in its response.

Manual QA is still pending: with `qaGate` on, this PR carries `needs-qa` and stays unmergeable until a maintainer applies `qa-approved`.

adeptofvoltron:
🤖 `om-open-pr` — chain hand-off: this PR now carries the run's `in-progress` lock, which moved here from issue #5533 so the work is never observably unclaimed. `om-auto-review-pr` continues on it next; the lock is released once the chain finishes.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock on this PR from `om-open-pr`. Reviewing the head as pushed; the lock is retained for the rest of the chain.

adeptofvoltron (commented):
# 🔍 Code Review

**PR #5568** — `fix(auth): validate reset token before rendering the password form (#5533)`
**Mode:** first review · **Scope:** 13 changed files in `packages/core/src/modules/auth` · **Base:** `develop`

> ⚠️ Submitted as a review **comment**, not a formal approval: GitHub does not allow the PR author to approve their own pull request, and this automation authored it. A second human reviewer still owns the approving review.

## ✅ Verdict

**Approve-equivalent** — no blockers and no majors. Two nits found in the diff scan were fixed in this same pass (`621c5500`); the remaining items below are disclosures for the human reviewer, not change requests.

## 🎯 What the change does

`/reset/<token>` previously rendered the full password form for any token and only surfaced "Invalid or expired token" after the user filled in both fields and submitted. The PR adds a read-only `AuthService.isPasswordResetTokenValid`, exposes it through a new rate-limited `POST /api/auth/reset/validate`, and has the page branch on the result at mount: spinner while checking, terminal "request a new link" card when dead, existing form otherwise.

## 🔍 Findings

### Nits — fixed in this pass

- `packages/core/src/modules/auth/api/reset/validate.ts:26` — **Nit (fixed):** `const c = await createRequestContainer()` used a one-letter binding, which the repo's coding rules forbid outright. The two sibling routes (`reset.ts`, `reset/confirm.ts`) both use `c`, so this was inherited convention rather than an oversight, but new code follows the written rule; renamed to `container`.
- `packages/core/src/modules/auth/api/reset/validate.ts:16` — **Nit (fixed):** a `// validation via validatePasswordResetTokenSchema` marker copied from the sibling route restated the import three lines above it. Removed.

### Disclosures for the human reviewer — no change requested

- `packages/core/src/modules/auth/api/reset/validate.ts:28` — **Security note.** This endpoint gives someone who already holds a reset token a way to confirm the token is live *without consuming it*. Before this PR, the only way to test a token was to POST a password to `/api/auth/reset/confirm`, which consumed a valid token and fired the `auth.password_reset.completed` notification — so probing was destructive and visible. The new check is silent. I judged this acceptable rather than a blocker: an attacker holding the token can already take the account over outright, so silent verification grants no capability they lacked, and the endpoint is IP rate-limited (10 per 5 minutes, `RATE_LIMIT_RESET_VALIDATE_*`) against a 32-byte random token, which makes guessing infeasible. Flagging it explicitly because it is the one genuine behavior change beyond the UX fix, and a reviewer should agree with that trade rather than discover it.
- `packages/core/src/modules/auth/frontend/reset/[token]/page.tsx:43` — **Design note.** The mount check deliberately **fails open**: a network error, a 429, or any non-`valid:false` answer renders the form. `/api/auth/reset/confirm` remains the authoritative gate, so nothing is weakened, and the alternative — failing closed — would tell users holding perfectly good links that their link is dead whenever the check endpoint hiccups. The `try/catch` around `apiCall` matters here specifically: without it a rejected fetch would leave the page stuck on the spinner forever.
- `packages/core/src/modules/auth/api/reset/validate.ts:24` — **Deliberate non-disclosure.** The response never distinguishes unknown, malformed, used, and expired tokens; all four return `{ ok: true, valid: false }`. `.ai/qa/scenarios/TC-AUTH-007` describes distinct per-cause messages ("Token has expired" / "Token has already been used"), and this PR intentionally does **not** implement that: the auth module's own guidance is that auth responses must not reveal state an unauthenticated caller has not earned. The scenario doc's step table (validate on load, show a terminal state, offer a link to `/reset`) is satisfied; only its per-cause message split is not.

## 🧩 Contract surface

Checked against `BACKWARD_COMPATIBILITY.md`. Everything here is additive: a new API route (routes are additive-only, and `/api/auth/reset/validate` is confirmed registered in `modules.generated.ts` as `/auth/reset/validate`), a new exported zod schema and its inferred type, and a new `AuthService` method. No event ID, response field, DB column, DI key, ACL feature, or import path was removed or renamed. **No deprecation window is owed.**

## 🧪 Tests

Full gate run locally, in the order `.ai/agentic.config.json` defines, all green:

- `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app`
- `yarn test` — **11465 passed / 1427 suites**, 2 skipped, 0 failed
- `yarn lint` — 0 errors; the 10 warnings are pre-existing and all in unrelated `example` modules

Coverage genuinely locks in the new behavior rather than restating it:

- `authService.test.ts` asserts the validity check calls neither `nativeUpdate` nor `flush` — i.e. it provably does not consume the token — and that it looks up by hashed token only.
- `reset-validate.test.ts` covers live/dead/malformed tokens, rate-limiting *before* any DB work, and that the response body carries no keys beyond `ok` and `valid`.
- `resetWithTokenPage.test.tsx` covers all three render states including the fail-open path when the check throws.
- `TC-AUTH-006` was rewritten to pull a **real** token out of the captured reset email, so the live-token path (form renders → submit → redirect) and the replayed-token path (terminal state) are both exercised end to end. The previous version used an invalid token as a shortcut and therefore never tested a working reset at all — this is a coverage improvement the fix forced, not just an adaptation.
- `TC-AUTH-007` was rewritten because it asserted the exact behavior this PR removes.

## ⏳ CI

Two required checks were still **pending** when this review was submitted (`Analyze (javascript-typescript)`, `prepare`); every settled check passed. Per this pipeline's rules the review is not held back for CI. The head is `MERGEABLE` — no conflicts against `develop`.

## 📸 QA

The PR carries `needs-qa` and, with the QA gate on, stays unmergeable until a maintainer adds `qa-approved`. UI evidence is being gathered separately by `om-auto-qa-pr`.

adeptofvoltron:

This PR passed code review with no blockers and no majors, and requires manual QA (`needs-qa`, no `skip-qa`). It stays in the `review` pipeline state rather than `merge-queue`, because the reviewing agent authored it and GitHub does not permit self-approval — a human approving review is still owed on top of QA. The QA-approval gate then holds the merge until `qa-approved` is added. QA reviewer: when you pick it up, swap the pipeline label to `qa`, then run the routes below.

### P0 — Password reset with a dead link

**Where to click**
- Request a reset at `/reset` for an account you control, open the emailed link at `/reset/<token>`, complete the reset, then **open that same link again**.
- Also visit `/reset/some-obviously-bogus-token` directly.

**What to verify**
- The re-opened (already-used) link shows the card "This reset link is no longer valid" with the subtitle explaining it was used or expired — and **no password fields at all**.
- "Request a new link" navigates to `/reset`; "Back to login" navigates to `/login`.
- The bogus token shows the same terminal card, with the same wording — it must not say anything that distinguishes "never existed" from "already used" from "expired".
- Briefly on load you should see the "Checking your reset link..." spinner rather than a flash of the password form. A visible flash of the form before the terminal card is a real defect worth reporting.

**What can go wrong**
- The form renders anyway, and you only learn the link is dead after submitting — that is the exact bug this PR fixes (#5533).
- The page hangs on the spinner forever instead of resolving to one state or the other.
- The terminal card's wording leaks *why* the token is dead.

### P0 — Password reset with a live link (regression risk)

**Where to click**
- Request a fresh reset at `/reset`, then open the emailed `/reset/<token>` link **once**.

**What to verify**
- The password form renders normally after a brief check, both fields work, and the password-requirements hint is shown.
- Submitting a valid, matching, policy-compliant password redirects to `/login`, and you can then log in with the **new** password and not the old one.
- Client-side validation still works before any network call: mismatched passwords show "Passwords do not match", and a weak password shows the requirements error.

**What can go wrong**
- A perfectly good link is misreported as invalid — the most damaging possible regression here, since it would lock users out of recovery entirely.
- The reset succeeds but the redirect to `/login` no longer happens.

### P1 — Degraded-network behavior (fail-open)

**Where to click**
- Open a **valid** `/reset/<token>` link with devtools → Network set to block or fail `/api/auth/reset/validate` (offline mode, or a request-blocking rule).

**What to verify**
- The page falls back to rendering the password form rather than showing the terminal card, and submitting still works normally.
- This is intentional: `/api/auth/reset/confirm` remains the authoritative check, so a check-endpoint outage must never strand a user holding a good link.

**What can go wrong**
- The page shows "no longer valid" for a good token because the check failed, or sits on the spinner indefinitely.

### P1 — Localization and presentation

**Where to click**
- Repeat the dead-link path with the app language set to each of pl, es, de, ko.
- Check the terminal card and the spinner state at a mobile viewport (≈375px wide).

**What to verify**
- All four new strings are translated in every locale — no raw `auth.reset.token.*` keys and no English fallback leaking through.
- The card is centered and readable at mobile width, matching the existing `/reset` page's look.

**What can go wrong**
- Untranslated keys rendering literally; the button or title overflowing the card at narrow widths.

### Pass/fail
- All routes pass → remove the `qa` label, add `qa-approved`, and leave the PR for a human approving review before merge.

adeptofvoltron:
## 📸 UI QA evidence — `om-auto-qa-pr` (evidence-only)

**Verdict: ✅ PASS** — all 7 steps passed. Every step below was actually driven in a real Chromium session against a running instance of this branch; nothing here is inferred from the diff.

**Environment.** Production build of this PR's head (`621c5500`) served on `http://localhost:3533`, backed by a disposable local Postgres database (`om_qa_5533`) provisioned fresh via `mercato db migrate` + `mercato auth setup`. Docker is unavailable on this host, so the ephemeral testcontainers lane could not be used; the disposable local database is the equivalent. The reset token used below is a **real** token, captured from the actual reset email through `OM_TEST_MODE` email capture — not a fabricated string.

### Results

| # | Step | Expected | Observed | |
|---|------|----------|----------|---|
| 1 | Open `/reset/<bogus-token>` | Terminal card, no password fields | `terminal=1 passwordFields=0`, title matched | ✅ |
| 2 | Click "Request a new link" | Navigates to `/reset` | `http://localhost:3533/reset` | ✅ |
| 3 | Open `/reset/<live-token>` | Password form renders, no terminal card | `passwordFields=1 terminal=0` | ✅ |
| 4 | Submit a valid new password | `confirm` → 200, redirect to `/login` | `confirmStatus=200`, landed on `/login` | ✅ |
| 5 | **Re-open that same link after the reset completed** | Terminal card, no password fields | `terminal=1 passwordFields=0` | ✅ |
| 6 | Terminal state at 375px mobile viewport | Card renders readably | `terminal=1` | ✅ |
| 7 | Open a reset link with `/api/auth/reset/validate` blocked | Falls open to the form, never stuck on the spinner | `passwordFields=1` | ✅ |

**Step 5 is the issue itself.** Before this PR, that same navigation rendered the full password form and only rejected on submit. It now renders the terminal state directly.

**Step 7 is worth reading closely.** On the first run it **failed** — the page hung forever on "Checking your reset link...". That was not a defect in the shipped code but a stale artifact: the `try/catch` around the check was added after the last `build:packages`, so the server was serving a pre-fix bundle. After rebuilding packages and the app, step 7 passes. The important consequence: the fail-open guard is not theoretical — without it, a failing check endpoint leaves users permanently stuck on the spinner, which is exactly what the first run demonstrated.

### What was not exercised

- The **expired-by-time** case (token older than 60 minutes) was not driven, because forcing it needs either clock manipulation or a direct DB write; the used-token path (step 5) and the unknown-token path (step 1) exercise the same code branch in `isPasswordResetTokenValid`.
- Locales other than English were not exercised in the browser. `yarn i18n:check-sync` confirms all four new keys exist in pl, es, de, and ko, but no screenshot proves their rendering.

No credentials, tokens, or real user data appear in any screenshot; the password fields are empty at capture time and the reset token is not visible in any frame.

### Screenshots


**step-01-bogus-token-terminal**



**step-02-request-new-link**



**step-03-live-token-form**



**step-04-reset-completed**



**step-05-used-token-terminal**



**step-06-mobile-terminal**



**step-07-fail-open**

adeptofvoltron:
🤖 `om-auto-qa-pr` — completed: PASS. All 7 browser steps passed against a production build of this branch on a disposable local Postgres instance, with the evidence and screenshots posted above and the `screenshots` label applied. QA verdict labels are untouched — this ran in the default evidence-only mode, so `needs-qa` still stands and a QA reviewer owns `qa-approved`. Lock retained — chain continues.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the autofix chain finished end to end — triage confirmed the defect, the fix shipped with unit, component and integration coverage, the full validation gate passed locally, code review found no blockers or majors (two nits fixed in-pass), and browser QA passed all 7 steps with screenshots attached. Lock released.

The PR is **not** merge-ready yet, by design and on two counts: it carries `needs-qa` and the QA-approval gate holds it until a maintainer adds `qa-approved`, and the review above is a comment rather than an approval because this automation authored the PR and GitHub does not permit self-approval — a human approving review is still owed. `ci-monitoring` is set while the remaining checks finish; no check has failed so far.

adeptofvoltron:

The first CI run settled at **13 passed, 1 failed**. The failure was **`TC-AUTH-006`, a test I rewrote in this PR — not the fix**. `TC-AUTH-007`, which is the regression coverage for issue #5533, passed both of its cases, so the change under review was green throughout.

**What went wrong.** My rewrite pulled the raw reset token out of the captured reset email so the spec could drive a real end-to-end reset. That works locally but not in the `ephemeral-integration` lane: the app writes the capture file where the Playwright process cannot read it, and `waitForResetToken` timed out on both the initial attempt and the retry. Worth noting for anyone who tries this again — the only other capture-based spec (`customer_accounts/TC-AUTH-033`) is not part of that lane, so there was no working precedent; this PR was the first to attempt it there.

**Fix pushed** (`ae0fe24a`): `TC-AUTH-006` now covers what the lane can drive deterministically — the terminal state for a dead token, the server-side rejection at `/api/auth/reset/confirm`, and that `/api/auth/reset/validate` is read-only and returns an identical, non-disclosing body on repeat calls. The live-token round trip is covered by the component test in `resetWithTokenPage.test.tsx` and was additionally exercised in a real browser during the QA pass above (steps 3 and 4). The reason it is not driven in the integration lane is recorded in the spec itself rather than dropped silently.

**Follow-up worth doing separately:** a shared email-capture helper that works in the ephemeral lane would let both this spec and `TC-AUTH-033` drive their full round trips. That is out of scope for a bug fix.

CI is re-running on the new head.

wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-24T18:22:14Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (changes requested):
# 🔍 Code Review: fix(auth): validate reset token before rendering the password form (#5533)

**PR #5568** — first review by @wojciechszyjka · **Scope:** 14 changed files in `packages/core/src/modules/auth` · **Base:** `develop` · **Head:** `ae0fe24a`
**Validation source:** GitHub checks on the head SHA (per this repo's GitHub-checks-first review policy) · **Runner:** not applicable — no local gate command was run, because CI already covers every configured gate on this exact commit.

## 🎯 Summary

The change does exactly what its title promises and does it in the right place. `/reset/<token>` used to render the "Set a new password" form for any token whatsoever, so a person following an already-consumed link had to type a password twice before the server told them the link was dead. This PR adds a read-only pre-check: `AuthService.isPasswordResetTokenValid` hashes the raw token, looks the row up by hash, and reports existence-plus-liveness without touching `usedAt`, so checking a token can never burn it; `POST /api/auth/reset/validate` exposes that as an unauthenticated, IP-rate-limited route returning nothing but `{ ok, valid }`; and the page renders one of three states — a spinner while it checks, a terminal "This reset link is no longer valid" card with a route back to `/reset`, or the original form.

Several decisions here are better than they had to be, and they are worth naming. The check **fails open** on every path that cannot answer — a thrown request, a non-2xx response, a 429 — so an outage in the new endpoint degrades to exactly the old behavior instead of stranding somebody holding a perfectly good link, and `/api/auth/reset/confirm` remains the only authoritative gate. The response body is deliberately uniform across unknown, used, and expired tokens, and the tests assert that the body has no keys beyond `ok` and `valid`, which locks the non-disclosure property in place rather than merely asserting it in prose. The route uses `POST` with a form body rather than a more RESTful `GET`, which is the right trade: a `GET` would put the reset token in a URL, where it leaks into access logs, browser history, and `Referer` headers. And because `tokenState` starts at `'checking'` on both the server and the client, the new render path is hydration-safe by construction — an improvement on the `data-auth-ready` gymnastics the old integration test needed.

I checked the security argument in the PR description rather than taking it on faith, and it holds. Reset tokens are 32 bytes of `randomBytes` rendered as hex (`packages/core/src/modules/auth/lib/tokenHash.ts:31`), so 256 bits of entropy stand between an attacker and a guess; a 10-per-5-minutes IP budget is not a brute-force surface. The new endpoint is also strictly *less* powerful for an attacker than `/api/auth/reset/confirm`, which already answers the same question and additionally sets the password. This is not a new enumeration surface.

What holds the PR back is not the design but two gaps in its completeness, both squarely inside the change's own stated goal, and both mechanical to close. They are detailed below.

## Verdict

❌ **request changes** — there are no blockers and no majors; the code is sound and the CI is green. Two Medium findings drive the verdict, because this repository requests changes on Medium as well as on Critical and High. Both are small: the new pre-check misses one case that `confirmPasswordReset` rejects, so a dead link still renders a working-looking form on that path, and the two new rate-limit environment knobs are absent from the two files where this repository documents every other per-endpoint knob.

## 🧪 Validation Gate

CI ran the full gate on this exact head (`ae0fe24a`) and every job is green, so per this repository's GitHub-checks-first policy no command was re-run locally. The mapping from the configured `validation.commands` to the CI job that covers it is spelled out per row.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | Covered by the `prepare` job (`.github/workflows/ci.yml:182`), which builds the packages and publishes the artifact the downstream jobs consume. |
| `yarn generate` | ✅ PASS | Covered by `prepare`; the `build` and `docker-build` jobs additionally prove the generated module registries pick up the new `api/reset/validate.ts` route, since an unregistered route would fail the app build. |
| `yarn build:packages` (rebuild after generate) | ✅ PASS | Covered by the second `build:packages` invocation in `prepare` (`.github/workflows/ci.yml:193`). |
| `yarn i18n:check-sync` | ✅ PASS | Covered by the `test` job's i18n stage; the four new `auth.reset.token.*` keys landed in all five locale files (en, pl, es, de, ko), which is what this gate exists to enforce. |
| `yarn i18n:check-usage` | ✅ PASS | Covered by the `test` job (`.github/workflows/ci.yml:635`). |
| `yarn typecheck` | ✅ PASS | Covered by the `test` job (`.github/workflows/ci.yml:648-653`). |
| `yarn test` | ✅ PASS | Covered by the `test` job (`.github/workflows/ci.yml:660-665`). The `ephemeral-integration (none)` job is green too, so the rewritten `TC-AUTH-006` and `TC-AUTH-007` Playwright specs pass in the lane that previously failed them. |
| `yarn build:app` | ✅ PASS | Covered by the `build` / `docker-build` jobs. |

Additional green checks on this head, none of them required by the configured gate but all worth recording: `lint`, `ds-lint`, `audit-scope`, `scope`, `documents-multi-instance`, `CodeQL` and all three `Analyze` jobs, and `license/cla`. **No required check is failing and none is pending**, so there is nothing for a CI follow-up to wait on. The `TC-AUTH-006` failure that the Blacksmith bot reported earlier in this conversation was against the previous head (`621c5500`) and was fixed by `ae0fe24a`; it does not apply to the code under review.

## Findings

### 🔹 Minor (Medium)

**1. `packages/core/src/modules/auth/services/authService.ts:147` — the pre-check reports a token as usable when its user has been soft-deleted, so the form still renders for a link that cannot possibly work.**

`confirmPasswordReset` rejects on three independent conditions: the row is missing, the row is used or expired, **and** the user the row points at no longer resolves as a live account — `findOneWithDecryption(this.em, User, { id: row.user.id, deletedAt: null })` returning `null` makes the whole confirm return `null`, which the route turns into a 400 "Invalid or expired token" (`packages/core/src/modules/auth/services/authService.ts:171`). `isPasswordResetTokenValid` implements only the first two. It never looks at the user at all.

The concrete failure: a person requests a reset, an administrator soft-deletes or deactivates that account inside the one-hour token window, and the person then opens their link. `isPasswordResetTokenValid` finds an unused, unexpired row and answers `valid: true`; the page renders the password form; they choose a password, type it twice, submit, and get "Invalid or expired token". That is precisely the experience this PR exists to eliminate, surviving on one path. It is worth taking seriously rather than dismissing as hypothetical, because this repository's own QA scenario for this flow names it explicitly — `.ai/qa/scenarios/TC-AUTH-006-password-reset-complete.md` lists "Token valid but user account deleted (should handle gracefully)" among the expected edge cases. Nothing here is a security problem: the server-side rejection is correct and unchanged, so the account stays safe either way. It is a completeness problem.

There is a second-order reason to fix it structurally rather than by pasting a fourth condition. The two methods now each carry their own copy of the liveness predicate, and a future change to one — a grace period, an `deletedAt` filter on `PasswordReset` itself, a lockout flag — will silently skip the other. That divergence is exactly the bug above, and it will recur.

Fix: extract the shared predicate. Add a private helper that loads the row, applies the used/expired test, and resolves the owning user with `{ deletedAt: null }`, returning the row (or `null`) to both callers, so `isPasswordResetTokenValid` and `confirmPasswordReset` can never again disagree about what "usable" means. Use a plain `em.findOne(User, …)` for the existence probe rather than `findOneWithDecryption` — the check needs to know only whether a live row exists, and skipping the decryption path keeps an unauthenticated endpoint cheap. Then add a service unit test asserting that a token whose user is soft-deleted reports `false`.

**2. `packages/core/src/modules/auth/api/reset/validate.ts:11` — the two new rate-limit environment variables are undocumented, and the docs page that claims to list every rate-limited endpoint is now incomplete.**

`readEndpointRateLimitConfig('RESET_VALIDATE', { points: 10, duration: 300, … })` introduces `RATE_LIMIT_RESET_VALIDATE_POINTS` and `RATE_LIMIT_RESET_VALIDATE_DURATION`. This repository documents per-endpoint rate-limit knobs in two places, consistently and without exception, and neither one mentions them:

- `apps/mercato/.env.example` carries a commented block of every per-endpoint override, right down to the sibling this route was modelled on (`RATE_LIMIT_RESET_CONFIRM_POINTS=5`, `RATE_LIMIT_RESET_CONFIRM_DURATION=300`) and even the not-yet-wired `RATE_LIMIT_2FA_VERIFY_*` pair. An operator who wants to tune the new limit has nothing to discover it from.
- `apps/docs/docs/framework/security/rate-limiting.mdx` is worse than merely silent, because it makes a positive claim it no longer satisfies. Its "Protected Endpoints" section opens with "The following authentication endpoints have two-layer rate limiting built in" and its Layer 1 table lists `POST /api/login`, `POST /api/reset` and `POST /api/reset/confirm`. `POST /api/auth/reset/validate` is rate-limited and is not in that table, so a reader auditing the auth surface from the documentation will now under-count it. Its "Per-Endpoint Overrides" tables have the same gap.

Fix: add `# RATE_LIMIT_RESET_VALIDATE_POINTS=10` and `# RATE_LIMIT_RESET_VALIDATE_DURATION=300` to the commented block in `apps/mercato/.env.example`, immediately after the `RESET_CONFIRM` pair; add a `POST /api/auth/reset/validate | IP | 10 | 300s | —` row to the Layer 1 table in the docs page and the matching two rows to the IP-only override table. (I deliberately do **not** ask for the same lines in `packages/create-app/template/.env.example`: that file is missing the entire per-endpoint override block already, so this is pre-existing drift of the whole section rather than something this PR introduced, and adding one lone orphan knob there would make it less coherent, not more. Worth a separate cleanup.)

### 💅 Nit (Low)

**3. `packages/core/src/modules/auth/frontend/reset/[token]/page.tsx:103` — the "checking" card is titled "Set a new password", which promises something the very next render may retract.**

While the token check is in flight the card header reads `auth.reset.title` ("Set a new password") above a spinner reading "Checking your reset link…". For a dead link that title is replaced a moment later by "This reset link is no longer valid". It is brief and harmless, and the QA pass confirmed there is no flash of the actual form, so this is genuinely the author's call — but a neutral title for the checking state (or no title at all, leaving the spinner line to carry the message) would avoid asserting an outcome before it is known.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `AuthService` gains `isPasswordResetTokenValid` and loses nothing; `data/validators.ts` gains `validatePasswordResetTokenSchema` and its inferred type additively.
- [x] No function signature changed in a breaking way — no existing signature in the diff was touched.
- [x] No required type field removed or narrowed — the new `ValidatePasswordResetTokenInput` is new; `ConfirmPasswordResetInput` is untouched.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — `POST /api/auth/reset/validate` is purely additive, and `/api/auth/reset/confirm` is byte-for-byte unchanged.
- [x] No field removed or retyped in an existing response shape — the new route's `{ ok, valid }` shape is new; no existing response was touched.
- [x] No event or message name renamed or removed; no payload field removed — the diff emits no events and changes none.
- [x] No CLI command or flag renamed or removed — no CLI surface in the diff.
- [x] No database table or column renamed or removed; no column type narrowed — no entity change and, correctly, no migration, since `isPasswordResetTokenValid` reads columns (`token`, `usedAt`, `expiresAt`) that already exist on `PasswordReset`.
- [x] No config key renamed and no default changed silently — the two new `RATE_LIMIT_RESET_VALIDATE_*` keys are new and optional, with hardcoded defaults. They are undocumented, which is Finding 2, but nothing existing changed meaning.
- [x] Where a contract had to change: not applicable — no contract had to change.

I checked the touched surfaces against `BACKWARD_COMPATIBILITY.md`. The one category this PR engages is the API-route surface, and it engages it additively, which that document permits without a deprecation window. **No protected surface is violated and there is nothing here to warn about.**

## 🧪 Test Coverage

Coverage on this PR is genuinely good, and specific rather than decorative. Naming what it actually covers:

- `services/__tests__/authService.test.ts` adds four cases that pin the new method's contract from three directions: a live token resolves `true` **and** asserts `em.nativeUpdate` and `em.flush` were never called, which is the property the whole design rests on (checking must not consume); a used token and an expired token each resolve `false`; and a fourth case asserts the lookup filter is exactly `{ token: hashAuthToken(raw) }`, so a future refactor cannot reintroduce a raw-token lookup — a regression this repository has explicitly fixed before (`2955e9ad9 security(auth): remove raw (unhashed) token fallback lookups`).
- `api/__tests__/reset-validate.test.ts` covers the route: live and dead tokens map to the right body, a token shorter than the schema's `min(10)` returns `{ ok: true, valid: false }` *without* reaching the auth service, the rate limiter is consulted before any DB work, and the response body's key set is asserted to be exactly `['ok','valid']` — the last one is the non-disclosure guarantee turned into an executable assertion, which is the right way to write that test.
- `__tests__/resetWithTokenPage.test.tsx` covers the page's three states: the validate call is asserted to be the *first* call the page makes; a `valid: false` answer renders the terminal card with no password fields at all, with "Request a new link" pointing at `/reset`, and with zero calls to `/api/auth/reset/confirm`; and a thrown check falls open to the form. The refactor from `expect(apiCallMock).not.toHaveBeenCalled()` to a `confirmCalls()` helper is the correct adaptation — it keeps the original assertions honest now that the page legitimately makes one other call, instead of weakening them to "some call happened".
- `__integration__/TC-AUTH-007.spec.ts` drives the regression for #5533 end to end in a browser: two dead tokens each render the terminal state with no form, and the terminal state navigates back to `/reset`.
- `__integration__/TC-AUTH-006.spec.ts` covers the page-level refusal, the server-side rejection at `/api/auth/reset/confirm`, and — nicely — that two consecutive `validate` calls return identical bodies, which is the read-only property asserted at the integration level rather than only at the unit level.

Two gaps, one acceptable and one not:

- **Acceptable.** `TC-AUTH-006` no longer drives a full live-token reset round trip. I checked the baseline before judging this: the *previous* version of that spec did not drive one either — it navigated to `/reset/qa-invalid-token` and asserted the error — so nothing regressed, and the file documents in a header comment exactly why the lane cannot read the captured reset email and where the path is covered instead (the component test, plus the browser QA pass on this PR). Documenting the limitation in the spec rather than dropping it silently is the right call, and the suggested follow-up (a shared email-capture helper usable in the ephemeral lane, which would also unblock `customer_accounts/TC-AUTH-033`) is correctly scoped out of a bug fix. No action needed here.
- **Not acceptable, and it is Finding 1.** There is no test for a token whose user has been soft-deleted, which is why that case slipped through. The test to add belongs in `services/__tests__/authService.test.ts` beside the other three rejection cases: mock `em.findOne` to return a live, unused, unexpired `PasswordReset` row and then resolve the user lookup to `null`, and assert `isPasswordResetTokenValid` returns `false`.

wojciechszyjka:

- 🚀 `merge-queue` — the code review passed on the current head (`9bef26cb`) and CI came back fully green on it. The first pass requested changes on two Medium findings; this run was invoked with autofix, so both were fixed here and the re-review approves. `needs-qa` stays on, so the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`. This label replaced `changes-requested`, which this run had set earlier in the same pass.
- 🐛 `bug` — kept as-is; issue #5533 reports broken behavior on an existing page, and that classification is still accurate.
- 🧪 `needs-qa` — kept as-is; the change is user-facing and adds three render states to `/reset/<token>`, so a human should exercise them. Browser QA evidence is already attached to this PR, but evidence is not approval, and this run deliberately applied no QA label from a code review. Manual QA instructions, including the new soft-deleted-user case, are in a separate comment.
- 🔽 `priority-low` — kept as-is; nothing is broken from a security or data-integrity standpoint, because the server-side rejection at `/api/auth/reset/confirm` was always correct and is untouched. This change only saves the user from typing a password twice for nothing.
- ⚠️ `risk-high` — kept as-is; the diff lands inside the auth module and adds a new unauthenticated endpoint, and this taxonomy rates any auth or shared-contract change high regardless of how contained its behavior change is. I verified the two properties that rating asks reviewers to confirm: the endpoint performs no `nativeUpdate` and no `flush`, so it cannot consume a token, and its response body is uniform across unknown, used and expired tokens.
- 📸 `screenshots` — kept as-is; the browser QA pass attached visual evidence to this PR.

Two labels this run set and has since removed, recorded so the history reads straight: `in-progress` was the review lock, released when the work finished, and `ci-monitoring` marked the outstanding CI follow-up, removed once the CI-result comment landed. Neither is owed any longer, and no further follow-up is coming from this agent.

wojciechszyjka (approved):
# 🔍 Code Review: Re-review — fix(auth): validate reset token before rendering the password form (#5533)

**PR #5568** — re-review by @wojciechszyjka after this run's own autofix commit · **Scope:** the two Medium findings from the first review, re-checked against the new head, plus a fresh pass over the full diff · **Base:** `develop`
**Mode:** re-review · **Autofix:** enabled for this run (`--autofix`), so both Medium findings were fixed here rather than handed back to the author.

## 🎯 Summary

My first review on this PR requested changes on two Medium findings — no blockers, no majors, and green CI. Because this run was invoked with autofix, I fixed both in a follow-up commit on the PR branch rather than handing the PR back for another round trip. This re-review covers that commit and re-checks the original diff on the new head.

**Finding 1 (soft-deleted user) is fixed.** `confirmPasswordReset` rejected on three conditions — missing row, used-or-expired row, and an owning user that no longer resolves live — while `isPasswordResetTokenValid` implemented only the first two, so a link belonging to an account soft-deleted inside the token's one-hour window still rendered the password form and still failed on submit. The fix does not patch a fourth condition into one method and leave the duplication in place; it extracts the shared used/expired predicate into a private `findUnconsumedPasswordReset(token)` that both methods now call, and adds the user-liveness probe to `isPasswordResetTokenValid`. Both methods therefore read "usable" from one place, which is what stops this class of divergence from recurring the next time someone adds a condition.

One deliberate deviation from what my first review suggested, called out so it is not mistaken for an oversight: I proposed a plain `em.findOne` for the user probe on the grounds that a mere existence check does not need decryption. `packages/core/src/modules/auth/AGENTS.md` tells the opposite in plain terms — "Use `findWithDecryption` / `findOneWithDecryption` for user queries" — and `confirmPasswordReset` already uses `findOneWithDecryption` for the same lookup. The repository's own rule and internal consistency win over my micro-optimization, so the fix uses `findOneWithDecryption`. The cost is one extra decrypted read on an endpoint capped at 10 requests per 5 minutes per IP, which is not a load concern.

I also kept `confirmPasswordReset`'s existing ordering exactly as it was — the atomic compare-and-set still happens before the user lookup. Moving the user check ahead of the CAS would have been a behavior change to the confirm path (a token for a deleted user would stop being consumed), and that is outside what this PR asked for and outside what the review found. The refactor is deliberately confined to removing the duplicated predicate.

**Finding 2 (undocumented rate-limit knobs) is fixed.** `RATE_LIMIT_RESET_VALIDATE_POINTS` and `RATE_LIMIT_RESET_VALIDATE_DURATION` now appear in `apps/mercato/.env.example` immediately after their `RESET_CONFIRM` siblings, and in `apps/docs/docs/framework/security/rate-limiting.mdx` in both places that needed them: a `POST /api/auth/reset/validate` row in the Layer 1 "Protected Endpoints" table, so that table is once again a complete inventory of the rate-limited auth surface, and the two override rows in the IP-only configuration table. I also extended the sentence that explains why reset-confirm is IP-only, since the same reasoning (no email is available — the endpoint is reached with a token alone) applies to reset-validate, and added the one-line reason its budget is larger: the check never consumes the token.

I did **not** mirror the new knobs into `packages/create-app/template/.env.example`, and that is a considered call rather than a missed step. That file is missing the entire per-endpoint override block — every `RATE_LIMIT_LOGIN_*`, `RATE_LIMIT_RESET_*` and `RATE_LIMIT_2FA_*` line — so the drift is wholesale and pre-existing, and `scripts/template-sync.ts` does not cover `.env.example` at all. Adding one orphan knob into a file that lists none of its siblings would make that file less coherent, not more. The block belongs in a separate cleanup, which I have noted in the run report.

**The nit was left alone**, as a nit should be: the "checking" card still carries the "Set a new password" title. It is the author's call, and it is documented in the first review for whoever picks it up.

## Verdict

✅ **approve** — both Medium findings are fixed on the current head, the third finding is a nit that does not block, and nothing in the re-read of the full diff turned up anything new. The design was sound in the first review and it is sound now: the check is genuinely read-only, it fails open on every path that cannot answer, its response body does not distinguish unknown from used from expired, and it is now backed by a test that pins the case that was missing.

**This approval is code review only and does not clear the PR to merge.** The PR carries `needs-qa` without `skip-qa` and this repository's QA gate is on, so the merge stays blocked until a QA reviewer applies `qa-approved`. Browser QA evidence is already attached to this PR, but evidence is not approval, and I have deliberately not applied any QA label from a code review.

## 🧪 Validation Gate

Two sources, and the report distinguishes them honestly rather than blending them. The **reviewed head as pushed by the author** (`ae0fe24a`) was validated by CI, which ran every configured gate and reported green — that is the evidence for the original diff. The **autofix commit** is new code that CI has not seen yet at the time of writing, so it was validated locally in the isolated review worktree; the final CI run on the new head is being followed up separately and its result will be posted as a comment on this PR.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | Run locally in the review worktree before generate and typecheck, as the configured gate order requires. |
| `yarn generate` | ✅ PASS | Run locally; the module registries regenerate cleanly with the new route present. |
| `yarn build:packages` (rebuild after generate) | ✅ PASS | Covered by the same local chain. |
| `yarn i18n:check-sync` | ✅ PASS (CI, head `ae0fe24a`) | The autofix commit adds no user-facing strings — it touches a service, a service test, an `.env.example` and a docs page — so the locale files are unchanged and the CI result stands. |
| `yarn i18n:check-usage` | ✅ PASS (CI, head `ae0fe24a`) | Same reasoning: no i18n surface in the autofix commit. |
| `yarn typecheck` | ✅ PASS | Run locally, scoped to `@open-mercato/core`, which is the only package the autofix commit touches with TypeScript. |
| `yarn test` | ✅ PASS | Run locally, scoped to the three affected suites: **34 tests passed, 3 suites passed, 0 failed** across `services/__tests__/authService.test.ts`, `api/__tests__/reset-validate.test.ts` and `__tests__/resetWithTokenPage.test.tsx`. |
| `yarn build:app` | ✅ PASS (CI, head `ae0fe24a`) | Not re-run locally for the autofix commit: it adds no route, page, import or export, so the app build surface is unchanged from the head CI already built. The CI run on the new head re-verifies this. |

At the time this re-review was submitted, no required check was failing and none was pending on `ae0fe24a`. The `TC-AUTH-006` failure reported earlier in this conversation belonged to the previous head (`621c5500`) and was already fixed by `ae0fe24a`.

## Findings

### 💅 Nit (Low)

**`packages/core/src/modules/auth/frontend/reset/[token]/page.tsx:103` — the "checking" card is titled "Set a new password", which promises something the very next render may retract.** Carried forward unchanged from the first review, and deliberately not autofixed: it is a wording judgement about the product's voice, which belongs to the author rather than to a review bot. While the token check is in flight the header reads "Set a new password" above a "Checking your reset link…" spinner; for a dead link that title is replaced a moment later by "This reset link is no longer valid". It is brief and the QA pass confirmed there is no flash of the actual form. A neutral title for the checking state, or none at all, would avoid asserting an outcome before it is known.

## 💥 Breaking Changes

Re-checked against the autofix commit as well as the original diff; the conclusion is unchanged.

- [x] No exported/public symbol removed or renamed without a deprecation path — the autofix commit adds a **private** method (`findUnconsumedPasswordReset`), which is not a contract surface; `isPasswordResetTokenValid` and `confirmPasswordReset` keep their exact signatures and return types.
- [x] No function signature changed in a breaking way — neither public method's parameters or return type changed.
- [x] No database table or column renamed or removed; no column type narrowed — the autofix reads `User.deletedAt`, which already exists and is already read on this exact path by `confirmPasswordReset`.
- [x] No config key renamed and no default changed silently — the autofix commit **documents** `RATE_LIMIT_RESET_VALIDATE_*` at their existing hardcoded defaults (10 / 300); it does not introduce, rename or change any value.
- [x] Where a contract had to change: not applicable — no contract had to change.

Checked against `BACKWARD_COMPATIBILITY.md`: the only contract surface this PR engages is the API-route surface, engaged additively, which that document permits without a deprecation window. **No protected surface is violated; there is nothing to warn about.**

## 🧪 Test Coverage

The original coverage is described in full in my first review and is unchanged and still good. What the autofix commit adds on top:

- **`services/__tests__/authService.test.ts` — `isPasswordResetTokenValid rejects a live token whose user has been soft-deleted`.** This is the regression test for Finding 1 and it fails without the fix: it feeds an unused, unexpired `PasswordReset` row into the first lookup, resolves the subsequent user lookup to `null`, and asserts the method returns `false`. It also asserts the user query's filter is exactly `{ id: 'u1', deletedAt: null }`, so a future refactor cannot quietly drop the soft-delete predicate and keep the test green, and it re-asserts `em.nativeUpdate` was never called — the read-only guarantee has to survive the new second query too.
- **The existing "accepts an unused, unexpired token without consuming it" case** was updated to resolve the user lookup to a live row. That is the correct adaptation rather than a weakening: the case still asserts `true`, and still asserts that neither `em.nativeUpdate` nor `em.flush` was called, which is the property the whole design rests on.

The remaining known gap is the one the first review judged acceptable and still judges acceptable: `TC-AUTH-006` does not drive a full live-token reset round trip in the ephemeral integration lane, because the lane cannot read the captured reset email. The previous version of that spec did not drive one either, so nothing regressed; the path is covered by the component test and was exercised in a real browser during this PR's QA pass; and the spec file documents the limitation and the follow-up (a shared email-capture helper for the ephemeral lane) instead of dropping it silently.

wojciechszyjka:

An earlier QA pass on this PR already covered the original diff and passed all seven of its steps. **This comment supersedes it for the current head (`9bef26cb`)**: the routes below repeat the core paths so the run is self-contained, and add the one case the earlier pass could not have covered, because the behavior did not exist yet — a token whose account was deleted after the link was sent.

### P0 — The reset link's three states

**Where to click**
- Request a reset at `/reset` for an account you control, then open the emailed link at `/reset/<token>`.
- Complete the reset, then **re-open that exact same link**.
- Visit `/reset/some-obviously-bogus-token` directly.

**What to verify**
- The live link renders the "Set a new password" form; submitting a policy-compliant password succeeds and lands you on `/login`, and the new password logs you in.
- The re-opened (already-used) link renders the card "This reset link is no longer valid" with **no password fields at all**.
- The bogus token renders that same card with the same wording. It must not say anything that distinguishes "never existed" from "already used" from "expired" — if the three cases read differently to a user, that is a reportable defect.
- "Request a new link" navigates to `/reset`; "Back to login" navigates to `/login`.
- On load you briefly see the "Checking your reset link..." spinner rather than a flash of the password form. A visible flash of the form before the terminal card appears is a real defect worth reporting.

**What can go wrong**
- The form renders anyway for a dead link, and only fails on submit — that is the exact bug (#5533) this PR fixes, so it is the primary regression to watch.
- The terminal card appears for a link that is actually still valid, locking a legitimate user out of their own reset. Please probe this deliberately with a freshly requested link.
- The wording differs between the used, expired and unknown cases, which would turn the new endpoint into a disclosure channel.

### P0 — Deleted account holding a live link (new on this head)

**Where to click**
- Request a reset for a test account, capture the link, and **do not open it yet**.
- Delete (or deactivate/soft-delete) that account in the backoffice.
- Now open the captured link at `/reset/<token>`.

**What to verify**
- The page renders the terminal "This reset link is no longer valid" card, **not** the password form. Before this head it rendered the form and only rejected the user after they typed a password twice.
- Attempting the reset by other means still fails: the account must not become resettable.

**What can go wrong**
- The password form renders, which means the new liveness check is not reaching the user lookup.
- A deleted account's password actually gets changed — that would be a serious defect and should fail QA immediately rather than being filed as a nit.

### P1 — Perceived performance and small viewports

**Where to click**
- Cold-load `/reset/<token>` (hard refresh, empty cache) on both a live and a dead token.
- Repeat at a 375px-wide mobile viewport.

**What to verify**
- The first useful paint is the card with the spinner, not a blank screen; the extra round trip the page now makes on load should not be perceptible as a stall.
- Both the terminal card and the form fit a 375px viewport without horizontal scrolling or clipped buttons.
- The spinner is announced to assistive technology (it carries `role="status"` with a "Loading" label) — a screen-reader pass on the checking state is worth one minute if you have the tooling.

**What can go wrong**
- A layout jump when the card swaps from the checking state to the form or the terminal state.
- The reset page hanging on the spinner forever if the check never resolves. It is designed to fail open to the form on any error, so a permanent spinner is a defect, not the intended fallback.

### P2 — Rate limiting and documentation

**Where to click**
- Reload `/reset/<token>` more than ten times within five minutes from the same IP (this exercises `RATE_LIMIT_RESET_VALIDATE_POINTS`, default 10 per 300s).
- Skim `apps/docs/docs/framework/security/rate-limiting.mdx` in the docs preview.

**What to verify**
- Once the limit trips, the page **still renders the password form** rather than a terminal card or an error. The check fails open by design, so hitting the limit degrades to the old behavior instead of locking a valid link out.
- The docs page lists `POST /api/auth/reset/validate` under "Protected Endpoints" and the two `RATE_LIMIT_RESET_VALIDATE_*` overrides in the configuration table.

**What can go wrong**
- A rate-limited check being misread as "token invalid", which would show the terminal card to somebody holding a good link. This is the highest-value thing to probe in this section.

### Pass/fail

wojciechszyjka:

**What happened.** The first pass reviewed the head as pushed (`ae0fe24a`) and submitted **changes requested** on two Medium findings — no blockers and no majors, and CI was green throughout. This run was invoked with autofix, so rather than handing the PR back for another round trip I fixed both here, in an isolated worktree, and the re-review on the new head (`9bef26cb`) **approves**. One autofix iteration was enough; no second cycle was needed.

**The two findings, and what closed them.**

1. `isPasswordResetTokenValid` reported a token as usable when its account had been soft-deleted, while `confirmPasswordReset` rejects that case — so a link belonging to an account deleted inside the token's one-hour window still rendered the password form and still failed on submit, which is exactly the experience #5533 set out to remove. Fixed in `48d91787` by extracting the shared used/expired predicate into a private `findUnconsumedPasswordReset` that both methods call, and adding the user-liveness probe to the check. The probe uses `findOneWithDecryption`, per the auth module's own rule for user queries and matching what `confirmPasswordReset` already does. `confirmPasswordReset`'s ordering is untouched — the atomic compare-and-set still runs before the user lookup. A regression test ships with it and fails without the fix.
2. `RATE_LIMIT_RESET_VALIDATE_POINTS` / `_DURATION` were undocumented, and the docs page's "Protected Endpoints" table — which presents itself as the complete inventory of rate-limited auth endpoints — did not list the new route. Fixed in `9bef26cb` across `apps/mercato/.env.example` and `apps/docs/docs/framework/security/rate-limiting.mdx`.

The third finding is a nit (the "checking" card is titled "Set a new password" before the outcome is known) and was deliberately **not** autofixed: it is a wording judgement about the product's voice, which belongs to the author. It is documented in the review for whoever picks it up.

**Inherited feedback.** No unaddressed actionable review feedback from other actors was found on this PR. The prior review body and the QA evidence were written by this collection's own skills, so they are filtered out by contract rather than re-litigated; the Blacksmith bot's `TC-AUTH-006` failure report belonged to the superseded head `621c5500` and was already fixed by `ae0fe24a`, so it is resolved rather than dropped.

**Validation.** The reviewed head was validated by CI, which ran every configured gate green. The autofix commits were validated locally in the review worktree: `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn typecheck` (scoped to `@open-mercato/core`, the only package touched) and the three affected test suites — **34 tests passed, 0 failed** — plus ESLint on the changed TypeScript files. `i18n:check-sync`, `i18n:check-usage` and `build:app` were not re-run locally, because the autofix commits add no strings, routes, pages or exports; CI is re-verifying all of it on the new head now.

**Follow-up worth doing separately, not blocking this PR.** `packages/create-app/template/.env.example` is missing the *entire* per-endpoint rate-limit override block, not just the new knobs, and `scripts/template-sync.ts` does not cover `.env.example` at all. I deliberately did not add one orphan knob to a file that lists none of its siblings; the block deserves its own cleanup.

**This PR is approved but not merge-ready.** It carries `needs-qa` without `skip-qa` and the QA gate is on, so the merge stays blocked until a maintainer applies `qa-approved`. Manual QA instructions for the current head, including the newly covered deleted-account case, are in a separate comment above. `ci-monitoring` is set while the CI run triggered by the autofix commits finishes; the result will be posted as a follow-up comment.

wojciechszyjka:

**All required checks passed on `9bef26cb`, the head this run's autofix commits produced.** The approving re-review stands unchanged and the PR keeps its `merge-queue` pipeline label.

| Check | Result | Duration |
|-------|--------|----------|
| [`test`](https://github.com/open-mercato/open-mercato/actions/runs/32763192798/job/97548113754) | ✅ pass | 11m22s |
| [`ephemeral-integration (none)`](https://github.com/open-mercato/open-mercato/actions/runs/32763192798/job/97548113893) | ✅ pass | 13m31s |
| [`prepare`](https://github.com/open-mercato/open-mercato/actions/runs/32763192798/job/97546583189) | ✅ pass | 4m57s |
| [`docker-build`](https://github.com/open-mercato/open-mercato/actions/runs/32763192798/job/97548113737) | ✅ pass | 7m34s |
| [`documents-multi-instance`](https://github.com/open-mercato/open-mercato/actions/runs/32763192798/job/97548113915) | ✅ pass | 3m42s |
| [`lint`](https://github.com/open-mercato/open-mercato/actions/runs/32763192798/job/97546583214) | ✅ pass | 46s |
| [`ds-lint`](https://github.com/open-mercato/open-mercato/actions/runs/32763192798/job/97546583306) | ✅ pass | 1m19s |
| [`scope`](https://github.com/open-mercato/open-mercato/actions/runs/32763192769/job/97546582617) / [`audit-scope`](https://github.com/open-mercato/open-mercato/actions/runs/32763192798/job/97546583044) | ✅ pass | 33s / 16s |
| [`CodeQL`](https://github.com/open-mercato/open-mercato/runs/97546809278) + all three `Analyze` jobs | ✅ pass | — |
| `license/cla` | ✅ pass | — |

`audit`, `mutate`, `comment`, `merge-coverage` and `[code]smith` reported `skipping`, which is their normal state for this PR and not a failure.

**What this confirms beyond the local run.** The two gates I could not meaningfully reproduce locally for the autofix commits are now verified on real infrastructure: `test` re-ran the full unit suite including the new `isPasswordResetTokenValid rejects a live token whose user has been soft-deleted` case, and `ephemeral-integration (none)` re-ran the Playwright lane green — the same lane whose `TC-AUTH-006` timeout had gone red on the superseded head `621c5500`. That failure is now two heads behind and stays fixed.

**This does not clear the PR to merge.** `needs-qa` is still on without `skip-qa` and the QA gate is enabled, so the merge remains blocked until a maintainer applies `qa-approved`. Manual QA instructions for this head are in the comment above. Removing `ci-monitoring` now — no further CI follow-up is owed from this agent.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-25T00:41:55Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

pkarw:
## 📸 UI QA evidence — PASS

**Verdict:** ✅ PASS — every one of the eleven steps below was driven in a real Chromium session against a production build of this branch, and each one behaved as the change intends.
**Environment:** `http://127.0.0.1:5001` · role `admin` (`admin@acme.com`, disposable demo account on a throwaway Postgres) · browser `playwright` (Chromium 1.62.1)
**Verified:** `fix/issue-5533-reset-page-used-token` @ `9bef26cb`

### Scenario (P2 — password reset, `/reset/<token>`)

The priority tag follows the PR's own `priority-low` label. The *surface* is a P0-class
area, which is why the run drives the whole round trip end to end — request a real reset
link, complete the reset, then replay the consumed link — rather than smoke-checking the
terminal card in isolation.

**Where to click:** `/reset/<token>`, `/reset`, `/login`

| 1 | Cold-load `/reset/<never-issued-token>` | Terminal card "This reset link is no longer valid" with the used/expired subtitle, "Request a new link" → `/reset`, "Back to login" → `/login`, and no password form | Terminal card rendered with both links at the expected hrefs; **New password**, **Confirm new password** and **Update password** each resolved to 0 elements | ✅ |
| 2 | Click "Request a new link" | Navigates to `/reset` with the email request form | URL settled on `/reset`, Email field visible | ✅ |
| 3 | Submit `admin@acme.com` on `/reset`, then read the captured reset email | Generic confirmation copy, and a real reset link emitted | "If an account with that email exists…" rendered; a `/reset/<token>` link was captured with a non-trivial token | ✅ |
| 4 | Open the **live** reset link | The "Set a new password" form renders; the terminal state does not | Both password fields, the requirements hint and the submit button rendered; `[data-auth-token-state="expired"]` resolved to 0 elements | ✅ |
| 5 | Set a new password and submit | Reset completes, redirect to `/login` | Redirected to `/login` | ✅ |
| 6 | **Reload the same link now that it is consumed** — the #5533 regression | Terminal state, not a form that can only fail on submit | Terminal card rendered; no password fields, no submit button | ✅ |
| 7 | Sign in at `/login` with the password from step 5 | Authentication succeeds | Authenticated and landed on the backend shell, proving the step-5 reset really took effect | ✅ |
| 8 | Load the dead link at a 390×844 mobile viewport | Card and actions readable and reachable | Rendered correctly, no clipping or overflow | ✅ |
| 9 | Delay the validate response 2.5 s and watch the pending state | A loading state appears; no form flashes before the answer | Spinner + "Checking your reset link…" rendered, submit button absent throughout, terminal card replaced the spinner on answer | ✅ |
| 10 | Fail the validate request at the network layer | Page **fails open** to the form (the PR's documented contract) | Form rendered, terminal state absent | ✅ |
| 11 | Answer the validate request with HTTP 429 | Page fails open rather than misreporting the link as dead | Form rendered, terminal state absent | ✅ |

Steps 3–7 are the part the committed integration specs do **not** cover — see the first
note below.

### Notes for QA

- **The PR description overstates the shipped test coverage.** It says `TC-AUTH-006` was
  "rewritten to pull a real token out of the captured reset email, so the suite now covers
  the live-token path end to end". At head `9bef26cb` that spec's own docstring says the
  opposite — the live-token half is explicitly *not* driven there, because the capture file
  is not readable from the ephemeral lane. The description is stale relative to the head.
  The round trip is genuinely covered, but by the manual steps 3–7 above and by the
  component test, not by the integration suite. Worth correcting in the body before merge
  so a later reader does not trust coverage that is not there.
- **The fail-open path is reachable by ordinary traffic, not just by outages.**
  `POST /api/auth/reset/validate` is IP rate-limited at 10 requests per 5 minutes by
  default, and the client treats any non-OK answer as "usable". A user who reloads a reset
  link more than ten times inside that window — or a shared-IP/NAT population that does so
  collectively — gets the old form-that-can-only-fail behaviour back. This is a degradation
  to the pre-PR experience, not a security hole (`/api/auth/reset/confirm` is still the
  authoritative gate), and it is the deliberate trade-off the PR documents. Flagging it so
  the choice is made with eyes open. Verified with a simulated 429 (step 11), **not** by
  exhausting the real limiter.
- **Cosmetic:** while the check is pending the card keeps the title "Set a new password"
  (step 9), so a dead link briefly shows that heading before the terminal card replaces it.
  No form control is rendered at any point during the check, so nothing invites input.
- **Not exercised:** a genuinely time-expired token — expiry is 60 minutes and cannot be
  forced through the UI. It reaches the terminal state through the same `valid: false`
  response as the consumed token in step 6.
- **Not exercised:** the four non-English locales (`pl`, `es`, `de`, `ko`) for the new
  `auth.reset.token.*` keys; only the `en` surface was driven.

### Screenshots

Filenames map to the `#` column above.

pkarw:

**What was verified.** Eleven browser steps against a production build of `9bef26cb` on a
disposable Postgres, driven in Chromium. Beyond the terminal-state rendering that
`TC-AUTH-007` already covers, this run drove the full round trip the integration suite does
**not**: request a real reset link, pull the live token out of the captured email, watch the
form render for it, complete the reset, sign in with the new password, and then replay the
now-consumed link and confirm it lands on the terminal card. That last hop is the #5533
regression itself, and it behaves correctly.

**Two things a reviewer should read before merging.**
1. The PR description says `TC-AUTH-006` was rewritten to cover the live-token path end to
   end via the captured email. At this head the spec's own docstring says the opposite — that
   half is explicitly not driven there. The description is stale; the coverage claim should be
   corrected so nobody later trusts a test that does not exist.
2. The client fails open on any non-OK answer from `/api/auth/reset/validate`, and that route
   is IP rate-limited at 10 requests per 5 minutes. Reloading a reset link past that limit —
   or a shared-IP population doing so collectively — restores the old
   form-that-can-only-fail behaviour. Not a security issue (`/reset/confirm` is still the
   authoritative gate) and it is the documented trade-off, but it is reachable by ordinary
   traffic, not only by an outage.

**🏷️ Labels: unchanged.** This run used the evidence-only default, so it applied no QA
verdict labels. The PR still carries `needs-qa` without `qa-approved`, which means the
QA-approval merge gate continues to hold it despite `merge-queue`. A maintainer applies
`qa-approved` after reading the evidence above.

PR #5569, opened 2026-08-24
fix(sales): register sales_channels_enabled from the sales module setup (#5503)

### LINKED ISSUE #5503: bug: [Feature toggles] the app polls sales_channels_enabled, which no module registers — 404 on every page
## Summary
`useSalesChannelsEnabled` polls a feature-toggle id that no module seeds, so every page mounting it logs a
failed request. Functionally harmless — the hook declares `defaultValue: true` — but it pollutes the console
and makes real errors harder to spot.

`GET /api/feature_toggles/check/boolean?identifier=sales_channels_enabled` → **404**
```json
{"ok":false,"error":{"code":"MISSING_TOGGLE",
 "message":"Toggle \"sales_channels_enabled\" not found (missing).",
 "identifier":"sales_channels_enabled","expectedType":"boolean","actualType":"boolean"}}
```

`GET /api/feature_toggles/global?pageSize=50` → 13 registered identifiers, none of them `sales_channels_enabled`:
`customers.interactions.legacy-adapters`, `customers.interactions.external-sync`, `example.customers_sync.enabled`,
`marketing_banner_text`, `portal_enabled`, `customers.interactions.unified`, `support_contact_settings`,
`feature_toggles_ui`, `example.customers_sync.bidirectional`, `wms_integration_sales_order_inventory`,
`wms_integration_shipping_shipments`, `wms_integration_procurement_goods_receipt`, `order_review_threshold`.

## Expected
Either the toggle is seeded, or the hook stops querying an identifier that is not registered.

adeptofvoltron:
Closes #5503

## 🎯 Goal
Stop every backend page that mounts `useSalesChannelsEnabled` from firing a request that 404s. The sales channels feature toggle is now registered by the module that owns it, so the check API resolves it instead of answering `MISSING_TOGGLE`.

## 🔍 Problem
`GET /api/feature_toggles/check/boolean?identifier=sales_channels_enabled` returns `404 MISSING_TOGGLE` on installations where the toggle was never seeded. The hook declares `defaultValue: true` and fails open, so channels stay visible and nothing breaks functionally — but the failed request is logged on every page that mounts the hook, which buries real errors in console noise. The reporter's instance lists 13 registered identifiers and `sales_channels_enabled` is not among them, even though the other four toggles from the core defaults file are.

## 🔍 Root Cause
The sales module owned the toggle's *behaviour* (`lib/salesChannelsToggleId.ts`, `lib/salesChannelsToggle.ts`, `components/useSalesChannelsEnabled.ts`) but not its *definition*. The definition lived only in `packages/core/src/modules/feature_toggles/defaults.json`, which reaches the database exclusively through the one-shot `yarn mercato feature_toggles seed-defaults` step invoked during `mercato init` (`packages/cli/src/mercato.ts:1200`). Any database initialized before that entry shipped never receives it, and nothing re-seeds it afterwards — which is exactly why the reporter has the four older `defaults.json` toggles but not this one.

Every other module that ships a toggle avoids this by registering it idempotently from its own `setup.ts` `seedDefaults` hook: `portal/setup.ts` (`portal_enabled`), `customers/setup.ts` (`customers.interactions.*`), and `wms/setup.ts` via `lib/wmsIntegrationToggles.ts` (`wms_integration_*`). `sales/setup.ts` has a `seedDefaults` hook, but it only seeded tax rates, dictionaries and methods.

## What Changed
- **`packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts`** (new) — holds `SALES_CHANNELS_TOGGLE_DEFINITION` and an idempotent `seedSalesChannelsToggle(em)` that creates the `FeatureToggle` row only when it is absent, mirroring the `wms` helper. Kept in its own file rather than added to `lib/salesChannelsToggle.ts`, because that module is imported by `page.meta.ts` files and must not statically pull in ORM entities.
- **`packages/core/src/modules/sales/setup.ts`** — `seedDefaults` now calls `seedSalesChannelsToggle(em)`. New tenants and onboarding register the toggle automatically; existing installations catch up with `yarn mercato seed:defaults --module sales`, which re-runs module `seedDefaults` hooks for every organization.
- **`packages/core/src/modules/feature_toggles/defaults.json`** — the `sales_channels_enabled` entry is removed so the definition has exactly one owner and the two copies cannot drift. No behavior depends on it being there: `mercato init` runs module `seedDefaults` after the CLI seeding step, so both paths still end with the toggle registered.
- **`apps/docs/docs/architecture/configuration-decision-guide.mdx`** — the provisioning guidance said core toggles should ship in `defaults.json`; it now documents module-owned seeding as the rule, explains that a `defaults.json`-only toggle stays unregistered on databases created before it shipped, and updates the worked example and the known-gaps entry to match.

## 🧪 Tests
- `packages/core/src/modules/sales/lib/__tests__/salesChannelsToggleSeed.test.ts` (new) — locks in that the helper creates the row when the definition is missing, no-ops when it already exists, and uses the same identifier and `defaultValue: true` that the client and server hooks poll.
- `packages/core/src/modules/sales/__tests__/setupSeedsSalesChannelsToggle.test.ts` (new) — the actual regression guard: asserts the sales `seedDefaults` hook calls the seeder. Verified red without the fix (`expect(seedSalesChannelsToggleMock).toHaveBeenCalledTimes(1)` fails) and green with it.
- Full validation gate run locally, in order: `build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `test`, `build:app` — all pass except one pre-existing failure noted below. Test totals include `@open-mercato/core` 11457 passed / 2 skipped, `ui` 1924, `shared` 1998, `cli` 1802, `ai-assistant` 1478, `documents` 1003, with zero failures across every workspace but one.
- **Pre-existing, unrelated failure:** `create-mercato-app` fails in `agent-harness-evaluator.test.ts`, `agent-harness-release.test.ts`, `business-writable-oracles.test.ts` and `writable-ast-oracles.test.ts`. These oracle tests shell out to the `codex` and `claude` CLIs and run a scaffolded app's `yarn typecheck` against a 120s budget, so they depend on local tooling and machine load. Confirmed pre-existing by running the same suite on a clean `origin/develop` checkout, where it fails **161** cases across the same four files — more than on this branch. This PR touches no file in `packages/create-app`.

## 💥 Breaking Changes
None. `sales_channels_enabled` keeps its identifier, type, description and `defaultValue: true`, so existing tenant overrides and both fail-open helpers are unaffected. Removing the entry from `defaults.json` narrows what `yarn mercato feature_toggles seed-defaults` creates, but the toggle is still registered on every path that previously registered it, and the seeding is idempotent, so re-running either command on an installation that already has the row is a no-op.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5569"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790169293&installation_model_id=432243&pr_number=5569&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5569&signature=091323835e0ed5d5e4b54a66e36b392935ba2a23d706e28e0c7e4e6a386d3f36"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/apps/docs/docs/architecture/configuration-decision-guide.mdx
+++ b/apps/docs/docs/architecture/configuration-decision-guide.mdx
@@ -83,7 +83,7 @@ Client side, `useFeatureFlagBoolean({ id })` and `<FeatureGuard id={...}>` read 
 
 A toggle with no definition row resolves to `MISSING_TOGGLE`, the check API answers `404`, and the client helpers coerce that to **disabled**. Two consequences:
 
-- Every toggle must have a definition row before dependent code runs. Core toggles should ship in `packages/core/src/modules/feature_toggles/defaults.json` and be seeded with `yarn mercato feature_toggles seed-defaults`; third-party definitions currently require explicit operator provisioning.
+- Every toggle must have a definition row before dependent code runs. The owning module should register its own toggles idempotently from its `setup.ts` `seedDefaults` hook (`portal`, `customers`, `sales` and `wms` all do), which covers new tenants and lets existing installations catch up with `yarn mercato seed:defaults --module <moduleId>`. `packages/core/src/modules/feature_toggles/defaults.json` is only for demo/platform toggles that no module owns; it is seeded once by `yarn mercato feature_toggles seed-defaults` during `init`, so a toggle that lives only there stays unregistered on every database created before it shipped.
 - Name toggles so that *missing* is the safe state. Prefer a positive `*_enabled` identifier with `defaultValue: true` over an inverted `disable_*` flag.
 
 Where hiding UI on a failed lookup would be worse than showing it, wrap the check in your own fail-open helper rather than depending on the default coercion — `packages/core/src/modules/sales/lib/salesChannelsToggle.ts` is the reference.
@@ -132,7 +132,7 @@ Follow the standard entity workflow: update `data/entities.ts`, run `yarn db:gen
 
 A client uses products but has a single sales channel, so the channel picker, column, filter, and management pages are noise for them.
 
-The value varies per tenant, only gates UI surfaces, and has to be read from client components — litmus test step 3, so a **feature toggle**. `sales_channels_enabled` ships in `defaults.json` with `defaultValue: true`; tenant admins override it at `/backend/feature-toggles/overrides`. No data-layer change was needed: `channelId` is already nullable and channel-less base prices already resolve.
+The value varies per tenant, only gates UI surfaces, and has to be read from client components — litmus test step 3, so a **feature toggle**. `sales_channels_enabled` is defined and seeded by the sales module itself (`packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts`, wired into `sales/setup.ts` `seedDefaults`) with `defaultValue: true`; tenant admins override it at `/backend/feature-toggles/overrides`. No data-layer change was needed: `channelId` is already nullable and channel-less base prices already resolve.
 
 Two alternatives that would have been wrong:
 
@@ -143,6 +143,6 @@ Two alternatives that would have been wrong:
 
 Documented so you are not surprised by them:
 
-- **Toggle definitions are centralised.** `defaults.json` lives inside `@open-mercato/core`, so a third-party module cannot ship a toggle definition without editing core or having an operator run the CLI. Every other module contract (`acl.ts`, `events.ts`, `ce.ts`, `setup.ts`) is auto-discovered from the module's own folder.
+- **Toggle definitions are not auto-discovered.** A module registers its toggles by calling its own seeding helper from `setup.ts` `seedDefaults`; there is no `toggles.ts` contract the generator picks up the way it picks up `acl.ts`, `events.ts` or `ce.ts`. Forgetting the `seedDefaults` call is silent — the toggle simply resolves to `MISSING_TOGGLE` at runtime.
 - **Feature toggles have no code-side default.** `getFeatureToggleValue` and `useFeatureFlagBoolean` take no default value, so callers who need fail-open behaviour hand-roll it. `ModuleConfigService.getValue` does accept `defaultValue`.
 - **`OM_*` env vars are not indexed.** There are well over a hundred and they are only discoverable by grep. A handful are semantically per-tenant despite being process-wide — `OM_ATTACHMENT_TENANT_QUOTA_MB` and `OM_SEARCH_ENABLED` among them. Treat those as legacy, not as precedent.
--- a/packages/core/src/modules/feature_toggles/defaults.json
+++ b/packages/core/src/modules/feature_toggles/defaults.json
@@ -17,14 +17,6 @@
       "defaultValue": "Free shipping on orders over $50"
     },
     {
-      "identifier": "sales_channels_enabled",
-      "name": "Sales Channels",
-      "description": "Show sales channel pickers, filters, and management UI. Disable for tenants that do not use sales channels.",
-      "category": "sales",
-      "type": "boolean",
-      "defaultValue": true
-    },
-    {
       "identifier": "order_review_threshold",
       "name": "Order Review Threshold",
       "description": "Routes orders above this total to manual review.",
--- a/packages/core/src/modules/sales/__tests__/setupSeedsSalesChannelsToggle.test.ts
+++ b/packages/core/src/modules/sales/__tests__/setupSeedsSalesChannelsToggle.test.ts
@@ -1,0 +1,55 @@
+/** @jest-environment node */
+
+jest.mock('../lib/dictionaries', () => ({
+  seedSalesStatusDictionaries: jest.fn().mockResolvedValue(undefined),
+  seedSalesAdjustmentKinds: jest.fn().mockResolvedValue(undefined),
+}))
+
+jest.mock('../seed/examples-data', () => ({
+  ensureExampleShippingMethods: jest.fn().mockResolvedValue(undefined),
+  ensureExamplePaymentMethods: jest.fn().mockResolvedValue(undefined),
+}))
+
+jest.mock('../seed/examples', () => ({
+  seedSalesExamples: jest.fn().mockResolvedValue(undefined),
+}))
+
+jest.mock('../lib/salesChannelsToggleSeed', () => ({
+  seedSalesChannelsToggle: jest.fn().mockResolvedValue(undefined),
+}))
+
+import type { EntityManager } from '@mikro-orm/postgresql'
+import { seedSalesChannelsToggle } from '../lib/salesChannelsToggleSeed'
+import setup from '../setup'
+
+const seedSalesChannelsToggleMock = jest.mocked(seedSalesChannelsToggle)
+
+function createEm(): EntityManager {
+  const em = {
+    find: jest.fn().mockResolvedValue([]),
+    findOne: jest.fn().mockResolvedValue(null),
+    persist: jest.fn(),
+    create: jest.fn((_entity: unknown, data: unknown) => data),
+    flush: jest.fn().mockResolvedValue(undefined),
+    transactional: jest.fn(async (cb: (tem: unknown) => Promise<void>) => cb(em)),
+  }
+  return em as unknown as EntityManager
+}
+
+describe('sales setup seedDefaults', () => {
+  beforeEach(() => {
+    seedSalesChannelsToggleMock.mockClear()
+  })
+
+  // Regression for #5503: the toggle used to live only in the core
+  // feature_toggles defaults file, so any database seeded before it shipped
+  // never registered it and every page mounting useSalesChannelsEnabled 404d.
+  it('registers the sales channels feature toggle', async () => {
+    const em = createEm()
+
+    await setup.seedDefaults?.({ em, tenantId: 'tenant-1', organizationId: 'org-1', container: {} as never })
+
+    expect(seedSalesChannelsToggleMock).toHaveBeenCalledTimes(1)
+    expect(seedSalesChannelsToggleMock).toHaveBeenCalledWith(em)
+  })
+})
--- a/packages/core/src/modules/sales/lib/__tests__/salesChannelsToggleSeed.test.ts
+++ b/packages/core/src/modules/sales/lib/__tests__/salesChannelsToggleSeed.test.ts
@@ -30,11 +30,12 @@ describe('seedSalesChannelsToggle', () => {
 
     await seedSalesChannelsToggle(em as never)
 
-    expect(findOneWithDecryptionMock).toHaveBeenCalledWith(
-      em,
-      FeatureToggle,
-      expect.objectContaining({ identifier: SALES_CHANNELS_TOGGLE_ID, deletedAt: null }),
-    )
+    // The filter must stay `deletedAt`-free: deletion is soft and the identifier
+    // unique constraint covers deleted rows too, so a `deletedAt: null` filter
+    // would turn a deleted toggle into a duplicate insert.
+    expect(findOneWithDecryptionMock).toHaveBeenCalledWith(em, FeatureToggle, {
+      identifier: SALES_CHANNELS_TOGGLE_ID,
+    })
     expect(em.create).toHaveBeenCalledWith(
       FeatureToggle,
       expect.objectContaining({
@@ -49,6 +50,21 @@ describe('seedSalesChannelsToggle', () => {
 
   it('is idempotent when the toggle already exists', async () => {
     findOneWithDecryptionMock.mockResolvedValue({ id: 'existing' })
+    const em = createEm()
+
+    await seedSalesChannelsToggle(em as never)
+
+    expect(em.create).not.toHaveBeenCalled()
+    expect(em.persist).not.toHaveBeenCalled()
+    expect(em.flush).not.toHaveBeenCalled()
+  })
+
+  // Deleting a toggle sets `deleted_at` but keeps the row, and
+  // `feature_toggles_identifier_unique` is not scoped to live rows, so seeding on
+  // top of a deleted definition must skip rather than raise a unique violation
+  // and abort the whole seed run.
+  it('does not insert when the definition row exists but was soft-deleted', async () => {
+    findOneWithDecryptionMock.mockResolvedValue({ id: 'deleted', deletedAt: new Date() })
     const em = createEm()
 
     await seedSalesChannelsToggle(em as never)
--- a/packages/core/src/modules/sales/lib/__tests__/salesChannelsToggleSeed.test.ts
+++ b/packages/core/src/modules/sales/lib/__tests__/salesChannelsToggleSeed.test.ts
@@ -1,0 +1,65 @@
+/** @jest-environment node */
+
+jest.mock('@open-mercato/shared/lib/encryption/find', () => ({
+  findOneWithDecryption: jest.fn(),
+}))
+
+import { findOneWithDecryption } from '@open-mercato/shared/lib/encryption/find'
+import { FeatureToggle } from '@open-mercato/core/modules/feature_toggles/data/entities'
+import { SALES_CHANNELS_TOGGLE_ID } from '../salesChannelsToggleId'
+import { SALES_CHANNELS_TOGGLE_DEFINITION, seedSalesChannelsToggle } from '../salesChannelsToggleSeed'
+
+const findOneWithDecryptionMock = jest.mocked(findOneWithDecryption)
+
+function createEm() {
+  return {
+    persist: jest.fn(),
+    create: jest.fn((_entity: unknown, data: unknown) => data),
+    flush: jest.fn().mockResolvedValue(undefined),
+  }
+}
+
+describe('seedSalesChannelsToggle', () => {
+  beforeEach(() => {
+    findOneWithDecryptionMock.mockReset()
+  })
+
+  it('registers the sales channels toggle when the definition row is missing', async () => {
+    findOneWithDecryptionMock.mockResolvedValue(null)
+    const em = createEm()
+
+    await seedSalesChannelsToggle(em as never)
+
+    expect(findOneWithDecryptionMock).toHaveBeenCalledWith(
+      em,
+      FeatureToggle,
+      expect.objectContaining({ identifier: SALES_CHANNELS_TOGGLE_ID, deletedAt: null }),
+    )
+    expect(em.create).toHaveBeenCalledWith(
+      FeatureToggle,
+      expect.objectContaining({
+        identifier: SALES_CHANNELS_TOGGLE_ID,
+        type: 'boolean',
+        defaultValue: true,
+      }),
+    )
+    expect(em.persist).toHaveBeenCalledTimes(1)
+    expect(em.flush).toHaveBeenCalledTimes(1)
+  })
+
+  it('is idempotent when the toggle already exists', async () => {
+    findOneWithDecryptionMock.mockResolvedValue({ id: 'existing' })
+    const em = createEm()
+
+    await seedSalesChannelsToggle(em as never)
+
+    expect(em.create).not.toHaveBeenCalled()
+    expect(em.persist).not.toHaveBeenCalled()
+    expect(em.flush).not.toHaveBeenCalled()
+  })
+
+  it('uses the identifier the client and server hooks poll', () => {
+    expect(SALES_CHANNELS_TOGGLE_DEFINITION.identifier).toBe('sales_channels_enabled')
+    expect(SALES_CHANNELS_TOGGLE_DEFINITION.defaultValue).toBe(true)
+  })
+})
--- a/packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts
+++ b/packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts
@@ -1,0 +1,36 @@
+import type { EntityManager } from '@mikro-orm/postgresql'
+import { FeatureToggle } from '@open-mercato/core/modules/feature_toggles/data/entities'
+import { findOneWithDecryption } from '@open-mercato/shared/lib/encryption/find'
+import { SALES_CHANNELS_TOGGLE_ID } from './salesChannelsToggleId'
+
+export const SALES_CHANNELS_TOGGLE_DEFINITION = {
+  identifier: SALES_CHANNELS_TOGGLE_ID,
+  name: 'Sales Channels',
+  description: 'Show sales channel pickers, filters, and management UI. Disable for tenants that do not use sales channels.',
+  category: 'sales',
+  type: 'boolean' as const,
+  defaultValue: true,
+} as const
+
+// The sales module owns this toggle definition, so tenant setup registers it the
+// same way portal, customers and wms register theirs. Shipping it only in the
+// core defaults file left it unregistered on every database initialized before
+// the toggle existed, and `useSalesChannelsEnabled` then 404s on every mount.
+export async function seedSalesChannelsToggle(em: EntityManager): Promise<void> {
+  const existing = await findOneWithDecryption(em, FeatureToggle, {
+    identifier: SALES_CHANNELS_TOGGLE_DEFINITION.identifier,
+    deletedAt: null,
+  })
+  if (existing) return
+  em.persist(
+    em.create(FeatureToggle, {
+      identifier: SALES_CHANNELS_TOGGLE_DEFINITION.identifier,
+      name: SALES_CHANNELS_TOGGLE_DEFINITION.name,
+      description: SALES_CHANNELS_TOGGLE_DEFINITION.description,
+      category: SALES_CHANNELS_TOGGLE_DEFINITION.category,
+      type: SALES_CHANNELS_TOGGLE_DEFINITION.type,
+      defaultValue: SALES_CHANNELS_TOGGLE_DEFINITION.defaultValue,
+    }),
+  )
+  await em.flush()
+}
--- a/packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts
+++ b/packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts
@@ -16,10 +16,16 @@ export const SALES_CHANNELS_TOGGLE_DEFINITION = {
 // same way portal, customers and wms register theirs. Shipping it only in the
 // core defaults file left it unregistered on every database initialized before
 // the toggle existed, and `useSalesChannelsEnabled` then 404s on every mount.
+//
+// The lookup deliberately does not filter on `deletedAt`. Deletion is soft
+// (`feature_toggles.global.delete`) while `feature_toggles_identifier_unique`
+// covers every row, so filtering would make this insert a duplicate and fail the
+// whole seed run on any installation where the toggle was deleted. Matching the
+// create command's own uniqueness check also leaves an operator's deletion alone
+// instead of resurrecting the definition on every tenant seed.
 export async function seedSalesChannelsToggle(em: EntityManager): Promise<void> {
   const existing = await findOneWithDecryption(em, FeatureToggle, {
     identifier: SALES_CHANNELS_TOGGLE_DEFINITION.identifier,
-    deletedAt: null,
   })
   if (existing) return
   em.persist(
--- a/packages/core/src/modules/sales/setup.ts
+++ b/packages/core/src/modules/sales/setup.ts
@@ -3,6 +3,7 @@ import type { ModuleSetupConfig } from '@open-mercato/shared/modules/setup'
 import { SalesSettings, SalesDocumentSequence, SalesTaxRate } from './data/entities'
 import { DEFAULT_ORDER_NUMBER_FORMAT, DEFAULT_QUOTE_NUMBER_FORMAT } from './lib/documentNumberTokens'
 import { seedSalesStatusDictionaries, seedSalesAdjustmentKinds } from './lib/dictionaries'
+import { seedSalesChannelsToggle } from './lib/salesChannelsToggleSeed'
 import { ensureExampleShippingMethods, ensureExamplePaymentMethods } from './seed/examples-data'
 import { seedSalesExamples } from './seed/examples'
 
@@ -116,6 +117,7 @@ export const setup: ModuleSetupConfig = {
     await seedSalesAdjustmentKinds(em, scope)
     await ensureExampleShippingMethods(em, scope)
     await ensureExamplePaymentMethods(em, scope)
+    await seedSalesChannelsToggle(em)
   },
 
   async seedExamples({ em, container, tenantId, organizationId }) {


adeptofvoltron:

- 🔍 `review` — the run finished with a complete, validated change, so the PR opens ready for code review rather than as a draft.
- 🐛 `bug` — this fixes a reported defect: the app polled a feature toggle identifier that no module registered, so the check API answered 404 on every page that mounted the hook.
- 📚 `documentation` — the configuration decision guide actively recommended the pattern that caused the bug (ship core toggles in `defaults.json`), so correcting that guidance is part of the fix rather than an incidental edit.
- ⏭️ `skip-qa` — the automated-verification exemption applies: the change touches no UI-rendering file (no `.tsx`, nothing under `packages/ui/src/` or `**/components/**`), leaves the database structure and the API surface unchanged, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships automated tests for the changed behavior in the same PR.
- 🔹 `priority-medium` — an ordinary bug rather than an incident: the hook fails open on `defaultValue: true`, so sales channel UI stays visible and nothing is functionally broken, but the repeated failed request pollutes the console and makes real errors harder to spot.
- 🟡 `risk-medium` — a single-area change with tests. It touches tenant setup seeding, which runs at bootstrap, but adds only an idempotent insert guarded by an existence check, and involves no migration, no auth or data-scoping code, and no shared contract surface.

adeptofvoltron:
🤖 `om-open-pr` — chain lock transferred to this PR. The autofix chain continues here with `om-auto-review-pr`; the issue lock on #5503 is released. Other auto-skills should skip this PR until the lock is released.

adeptofvoltron:
🤖 `om-auto-review-pr` — taking over the chain lock already held on this PR to run the review pass with `--autofix`. The lock is retained, not re-acquired; other auto-skills should continue to skip this PR.

adeptofvoltron:
# 🔍 Code Review

**Verdict: approve** — no blockers and no majors. One minor and one nit, both recorded below for the human reviewer rather than fixed, with reasons.

*Submitted as a comment rather than a formal approving review: GitHub does not allow an author to approve their own pull request, and this PR was opened by the same automation account. A human approval is still required before merge, so the PR stays in the `review` pipeline state rather than moving to `merge-queue`.*

## 🎯 Scope reviewed

Six files, one behavioural change. `packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts` (new) and one call added to `sales/setup.ts` `seedDefaults`; the `sales_channels_enabled` entry removed from `packages/core/src/modules/feature_toggles/defaults.json`; two new test files; one docs file. No migration, no entity change, no API surface change.

## ✅ Correctness

The root cause and the fix line up. `defaults.json` only reaches the database through `feature_toggles seed-defaults`, invoked once from `packages/cli/src/mercato.ts:1200` during `init`, so a toggle added to that file after a database was created never lands — which matches the reporter's global list exactly. Moving the definition into the owning module's `seedDefaults` hook fixes new tenants and onboarding, and `yarn mercato seed:defaults --module sales` (`packages/cli/src/mercato.ts:1475-1521`) re-runs module seed hooks per organization for existing ones. I checked the `--module` flag parses as expected and that it filters on `mod.id`, which is `sales`.

Removing the `defaults.json` entry does not create a gap. `mercato init` runs `feature_toggles seed-defaults` at line 1200 and module `seedDefaults` at line 1238, so both paths still end with the toggle registered; the seeder is guarded by an existence check, so the two are safe to run in either order or repeatedly.

## ✅ Project conventions

- `salesChannelsToggleSeed.ts:20` uses `findOneWithDecryption` rather than raw `em.findOne`, as `packages/core/AGENTS.md` requires and as `wms/lib/wmsIntegrationToggles.ts:41` already does.
- Placing the helper in its own file instead of extending `lib/salesChannelsToggle.ts` is the right call: that module is imported by three `page.meta.ts` files, so a static `FeatureToggle` entity import there would pull ORM entities into page metadata.
- The `withAtomicFlush` hazard from `packages/core/AGENTS.md` — a `find` between scalar mutations and `flush` on the same EntityManager — does not apply. I verified every seeder that runs before it in `seedDefaults` flushes before returning: `seedSalesTaxRates` wraps itself in `em.transactional`, `ensureDictionaryEntry` flushes at `lib/dictionaries.ts:112`, and `ensureExampleShippingMethods` / `ensureExamplePaymentMethods` flush at `seed/examples-data.ts:189` and `:230` (both called without `skipFlush`). The helper itself only creates a new entity, so there is no pending scalar mutation for its own lookup to strand.
- Seeding a global, non-tenant-scoped table from a per-tenant hook is redundant across organizations but idempotent, and is exactly what `portal`, `customers` and `wms` already do.
- No `any`, no hardcoded user-facing strings (the name and description are seed data persisted to the toggle row, as in the sibling modules), no status-colour or arbitrary-value design-system violations, no new dependency.

## ✅ Tests

Coverage is proportionate and the regression guard is real, not decorative. `setupSeedsSalesChannelsToggle.test.ts` asserts the `seedDefaults` hook actually calls the seeder — I confirmed it fails without the one-line `setup.ts` change and passes with it, which is the property that matters, since the bug was precisely a missing registration call. `salesChannelsToggleSeed.test.ts` covers create-when-missing, no-op-when-present, and that the identifier and default match what the client and server hooks poll.

## ⚠️ Minor — existing deployments still need a manual command

`packages/core/src/modules/sales/setup.ts:120`

The fix does not self-heal a running installation: an operator must run `yarn mercato seed:defaults --module sales`, and until they do, the 404 in the report continues. The sanctioned mechanism for auto-healing an upgraded tenant is an entry in `packages/core/src/modules/configs/lib/upgrade-actions.ts`, which surfaces a one-click action in the admin UI.

I deliberately did not add one. Upgrade actions are gated behind `UPGRADE_ACTIONS_ENABLED` (`configs/services/upgradeActionsService.ts:15`), which is off by default, so it would not help the reporter without a config change either; and each action needs four i18n keys across every shipped locale, which roughly triples the diff for a console-noise fix. Flagging it so a maintainer can overrule that trade-off — it is a reasonable follow-up issue if the toggle turns out to be missing on many deployments.

## 💬 Nit — duplicated seeding shape across four modules

`packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts:19`

This is the fourth near-identical "look up a `FeatureToggle` by identifier, create it if absent, flush" helper, after `portal/setup.ts:6`, `customers/setup.ts:41` and `wms/lib/wmsIntegrationToggles.ts:39`. A shared `ensureFeatureToggles(em, definitions)` in the `feature_toggles` module would collapse all four and give module authors an obvious entry point. Out of scope for a bug fix — following the established local pattern is the correct choice here — but worth a refactor issue.

## 🔁 Backward compatibility

No contract surface from `BACKWARD_COMPATIBILITY.md` is touched. The identifier, type, description and `defaultValue: true` are unchanged, so existing per-tenant overrides and both fail-open helpers keep working. `defaults.json` is seed data, not a listed contract surface, and no test or doc asserts its contents — I grepped for the other identifiers in it to confirm nothing reads the file's shape.

## 🧪 Validation

The full `validation.commands` gate was run locally in order: `build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `test`, `build:app`. All pass except `create-mercato-app`, whose `agent-harness-evaluator`, `agent-harness-release`, `business-writable-oracles` and `writable-ast-oracles` suites shell out to the `codex` and `claude` CLIs and run a scaffolded app's `yarn typecheck` against a 120s budget. Confirmed pre-existing: the same suite on a clean `origin/develop` checkout fails **161** cases across the same four files, against 20 on this branch. This PR changes no file in `packages/create-app`. `yarn generate` produced no diff, so there is no generated-file churn.

## 🖥️ UI verification

`UI: n/a`. The diff contains no `.tsx` outside tests, nothing under `packages/ui/src/` or any `components/` directory, and no user-visible copy — hence the `skip-qa` label under the automated-verification exemption. The user-observable effect is the absence of a failed request in the browser console, which no screenshot would evidence better than the tests already do.

## 🚦 CI

At review time: `lint`, `ds-lint`, `scope`, `audit-scope`, `Analyze (actions)`, `Analyze (python)` and `license/cla` all pass; `prepare` and `Analyze (javascript-typescript)` are pending. No check is failing. Pending checks are disclosed here rather than waited on; the result is followed up in a bounded pass below.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the fix is implemented, tested and reviewed, with an approve verdict carrying one minor and one nit for the human reviewer. Lock released — `in-progress` is swapped for `ci-monitoring` because two checks were still pending when the work finished, and that follow-up is the only thing this agent still owes the PR. The PR is free for anyone to act on.

A human approval is still required: GitHub blocks self-approval and this PR was opened by the same automation account, so it stays in `review` rather than moving to `merge-queue`.

adeptofvoltron:

All checks settled green within the follow-up budget: `test`, `docker-build`, `ephemeral-integration`, `documents-multi-instance`, `lint`, `ds-lint`, `scope`, `audit-scope`, `prepare`, `CodeQL`, all three `Analyze` variants and `license/cla` pass. `[code]smith`, `audit`, `comment`, `merge-coverage` and `mutate` were skipped by their own path filters. Nothing failed, so the approve verdict from the review above stands unchanged and no pipeline label needed correcting.

`ci-monitoring` is removed — this agent owes the PR nothing further. The PR remains in `review` awaiting a human approval, which GitHub requires here because the author and the reviewer are the same automation account.

wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-24T18:29:02Z. Other auto-skills will skip this PR until the lock is released.

Context for the take-over: the previous chain (`om-auto-fix-issue` → `om-auto-review-pr` running as @adeptofvoltron) explicitly released its lock and removed `ci-monitoring`, and its review had to be posted as a plain comment because GitHub blocks self-approval. This run reviews the same head as an independent reviewer so the PR can receive a formal verdict, with `--autofix` enabled.

wojciechszyjka (changes requested):
# 🔍 Code Review: fix(sales): register sales_channels_enabled from the sales module setup (#5503)

## 🎯 Summary

This PR moves the `sales_channels_enabled` feature-toggle *definition* out of `packages/core/src/modules/feature_toggles/defaults.json` and into the module that owns the toggle's behaviour, registering it from the sales `seedDefaults` hook through a new idempotent helper. The diagnosis behind it is correct and well evidenced: `defaults.json` reaches the database only through `yarn mercato feature_toggles seed-defaults`, which `mercato init` invokes once at `packages/cli/src/mercato.ts:1200`, so any toggle added to that file after a database was created never lands — which is exactly why the reporter's instance lists the four older `defaults.json` toggles but not this one. Moving the definition into `sales/setup.ts` `seedDefaults` covers `mercato init` (`packages/cli/src/mercato.ts:1238`), the onboarding provisioning path (`packages/onboarding/src/modules/onboarding/api/get/onboarding/verify.ts:287`) and, for existing installations, `yarn mercato seed:defaults --module sales` (`packages/cli/src/mercato.ts:1518`). I verified all three call sites.

Removing the `defaults.json` entry does not open a gap, and I checked that nothing else depends on the file's contents: the only remaining references to `sales_channels_enabled` are `lib/salesChannelsToggleId.ts`, component tests that mock that constant, and documentation. The helper follows the sibling modules closely — `wms/lib/wmsIntegrationToggles.ts:39`, `customers/setup.ts:41`, `portal/setup.ts:6` — and improves on two of them by using `findOneWithDecryption` instead of raw `em.findOne`, as `packages/core/AGENTS.md` requires. Keeping the helper in its own file rather than extending `lib/salesChannelsToggle.ts` is the right call, since that module is imported by `page.meta.ts` files that must not statically pull in ORM entities. The `FeatureToggle` entity is global (no `tenant_id` / `organization_id` columns), so seeding it from a per-tenant hook is redundant across organizations but not a scoping problem, and both new tests are real guards rather than decoration.

One correctness problem stands between this and merge: the new seeder is not idempotent against a **soft-deleted** definition row, and the failure mode is a raw database constraint violation on the very command the PR tells operators to run. Reviewed scope: all six changed files, plus the feature-toggles commands, service, entity, migration, and the three CLI/onboarding seeding call sites.

## Verdict

❌ **request changes** — one major finding: `salesChannelsToggleSeed.ts:20` filters the existence check by `deletedAt: null` while `feature_toggles_identifier_unique` is an unqualified unique constraint on `identifier` and the module's delete command soft-deletes, so on an installation where the toggle was deleted the seeder tries to insert a duplicate and aborts sales seeding with a Postgres unique-constraint violation. Everything else is minor polish or a follow-up suggestion, and the change is otherwise well diagnosed, well scoped and well tested.

## 🧪 Validation Gate

**Validation source: GitHub checks** (this repository's review extension prefers CI check results over re-running the gate locally; every gate below is covered by a check that reports green on head `ff8abcae`, so no local fallback was needed).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job, "Build packages" step ([run](https://github.com/open-mercato/open-mercato/actions/runs/32731562466/job/97444713169)). |
| `yarn generate` | ✅ PASS | `prepare` job, "Prepare generated modules" step; no generated-file churn appears in the diff. |
| `yarn i18n:check-sync` | ✅ PASS | `test` job, "Check i18n sync" step ([run](https://github.com/open-mercato/open-mercato/actions/runs/32731562466/job/97445915988)). |
| `yarn i18n:check-usage` | ✅ PASS | `test` job, "Check i18n usage" step. |
| `yarn typecheck` | ✅ PASS | `test` job, "Checking types" step. |
| `yarn test` | ✅ PASS | `test` job, "Test" step, plus "Test scripts", the create-app template-parity check and `yarn workspace create-mercato-app test`. |

Two things worth recording from CI rather than from the PR description. First, `create-mercato-app` is **green in CI**, so the local failures the author documented are an artifact of the local environment (those oracle suites shell out to the `codex` and `claude` CLIs and run a scaffolded app's `yarn typecheck` under a 120s budget) and not a pre-existing repository failure that anyone needs to act on. Second, every other required check is green as well — `lint`, `ds-lint`, `scope`, `audit-scope`, `docker-build`, `documents-multi-instance`, `ephemeral-integration`, `prepare`, `CodeQL` with all three `Analyze` variants, and `license/cla`. `[code]smith`, `audit`, `comment`, `merge-coverage` and `mutate` were skipped by their own path filters. Nothing is failing and nothing is pending, so the verdict here rests entirely on the code.

## Findings

### ⚠️ Major

**`packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts:20` — the seeder is not idempotent against a soft-deleted definition row and crashes with a unique-constraint violation.**

The existence check is `findOneWithDecryption(em, FeatureToggle, { identifier, deletedAt: null })`, but deletion in this module is a **soft** delete: `packages/core/src/modules/feature_toggles/commands/global.ts:362` sets `toggle.deletedAt = new Date()` and leaves the row in place. The uniqueness guarantee, meanwhile, is unqualified — `packages/core/src/modules/feature_toggles/migrations/Migration20251229215803.ts:9` adds `alter table "feature_toggles" add constraint "feature_toggles_identifier_unique" unique ("identifier")`, with no `WHERE deleted_at IS NULL` predicate. So on an installation where a super administrator deleted the sales channels toggle, the lookup finds nothing, `em.create` builds a second row with the same identifier, and `await em.flush()` raises a `UniqueConstraintViolationException`.

The blast radius is larger than the helper. `sales/setup.ts:120` calls it last inside `seedDefaults`, and the error propagates out of the hook: in `mercato init` the module loop at `packages/cli/src/mercato.ts:1238` is unguarded, so the whole initialization aborts; in `yarn mercato seed:defaults --module sales` the per-organization loop at `packages/cli/src/mercato.ts:1518` shares one EntityManager, so the first failure aborts the run for every remaining organization. That command is precisely the catch-up path this PR documents for existing installations, which is what turns an edge case into a realistic one — an operator whose administrator once removed the toggle runs the documented remedy and gets a raw Postgres error instead of a no-op. The repository's review checklist calls this out directly: setup hooks may run twice and must be idempotent.

The module's own create command already models the correct semantics. `commands/global.ts:93` guards with `em.findOne(FeatureToggle, { identifier: parsed.identifier })` — deliberately **without** a `deletedAt` filter — and rejects the duplicate rather than letting the database do it. The fix is to match that: drop `deletedAt: null` from the seeder's lookup so an existing row in any state means "already registered, nothing to do". Skipping is also the behaviour that respects operator intent, since resurrecting a definition an administrator deleted would silently undo their action on every tenant seed. Please add a regression test alongside it, in `lib/__tests__/salesChannelsToggleSeed.test.ts`, asserting that a soft-deleted row makes the helper a no-op (no `em.create`, no `em.persist`, no `em.flush`).

The same shape exists in `customers/setup.ts:41` and `wms/lib/wmsIntegrationToggles.ts:39`, so this is a pattern worth correcting there too — but that belongs in a follow-up, not in a bug-fix PR.

### 💅 Nit

**`packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts:19` — the seeding write bypasses the `feature_toggles.global.create` command.**

Registration for this toggle previously went through the command bus, because `feature_toggles` `seed-defaults` executes `feature_toggles.global.create` (`packages/core/src/modules/feature_toggles/cli.ts:334`), which writes the audit log entry and runs `emitCrudSideEffects` for the query index and CRUD events. Writing the row directly with `em.create` skips both. I checked what that actually costs here and it is small: the admin list route and `FeatureTogglesService.resolveToggle` both read the base `feature_toggles` table, not `entity_indexes`, so the toggle still appears in `/backend/feature-toggles` and still resolves — only the search/index document, the CRUD event and the audit entry are missing. Note that the command explicitly anticipates this caller: `assertGlobalToggleSuperAdmin` at `commands/global.ts:16` documents "tenant setup" as a trusted server-side caller that opts in via `systemActor: true`, and `seedDefaults` does receive `container`.

I am not asking for the change in this PR. Three sibling modules seed toggles the same raw way, so following them is the locally consistent choice, and switching one module to the command bus would add per-tenant audit rows and event emissions during `init` — a decision that should be made for all four at once. This is the same observation the earlier automated pass recorded, and it points at the same follow-up: a shared `ensureFeatureToggles(em, definitions)` helper in the `feature_toggles` module, going through the command bus with `systemActor: true`, would collapse `portal`, `customers`, `sales` and `wms` into one correct implementation and give module authors an obvious entry point.

**`packages/core/src/modules/sales/lib/salesChannelsToggleSeed.ts:6-13` — the definition object is declared and then copied field by field.**

`SALES_CHANNELS_TOGGLE_DEFINITION` is already `as const`, which makes the inner `type: 'boolean' as const` redundant, and `em.create` re-lists all six fields instead of spreading the constant. Since the whole point of extracting the definition is to have one source of truth, `em.create(FeatureToggle, { ...SALES_CHANNELS_TOGGLE_DEFINITION })` would remove the possibility of the two drifting apart when a field is added later. Author's call — it changes no behaviour today.

**`apps/docs/docs/architecture/configuration-decision-guide.mdx:86` — the new rule slightly overstates what `defaults.json` now holds.**

The revised text says the file "is only for demo/platform toggles that no module owns", but `feature_toggles_ui` remains in it and is owned by the `feature_toggles` module itself, gating that module's admin UI. Reading "platform" generously covers it, so this is not wrong — but a half-sentence marking the remaining entries as legacy examples that predate the rule would stop the next reader from concluding the file is already consistent with the guidance above it.

## 🔍 Inherited reviewer feedback (2 asks already on the PR)

The two items below come from the automated review comment posted at [13:17 UTC](https://github.com/open-mercato/open-mercato/pull/5569#issuecomment-5395752937) by the same automation account that authored the PR (it had to be a comment rather than a review because GitHub blocks self-approval). Both are accounted for here.

- **Minor — "existing deployments still need a manual command"** (`sales/setup.ts:120`), declined there in favour of a follow-up because upgrade actions sit behind `UPGRADE_ACTIONS_ENABLED` and would need four i18n keys per locale. **I concur, and add a second reason.** The nearest in-repo self-healing precedent is `resolveWmsIntegrationToggleEnabled` (`wms/lib/wmsIntegrationToggles.ts:60`), which re-seeds on `MISSING_TOGGLE` and invalidates the cache before retrying. That pattern would not remove the reported symptom, because the console noise comes from the *client* hook calling `/api/feature_toggles/check/boolean`, not from the server-side helper — self-healing would have to live in the check route itself, which is well outside a bug fix. Worth a follow-up issue; not a blocker here. Usefully, `FeatureTogglesService.resolveToggle` (`lib/feature-flag-check.ts:107`) returns the `missing` result **without** caching it, so once an operator runs the catch-up command the 404s stop immediately with no cache invalidation needed.
- **Nit — "duplicated seeding shape across four modules"** (`salesChannelsToggleSeed.ts:19`), left for a refactor issue. **Re-raised above as my own nit**, with the extra detail that a shared helper should also settle the command-bus question and the soft-delete guard for all four call sites at once.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the change is purely additive: `SALES_CHANNELS_TOGGLE_DEFINITION` and `seedSalesChannelsToggle` are new, and `SALES_CHANNELS_TOGGLE_ID` is untouched.
- [x] No function signature changed in a breaking way — `sales/setup.ts` `seedDefaults` keeps its signature and gains one call.
- [x] No field removed or retyped in an existing response shape — the toggle keeps its identifier, type, description, category and `defaultValue: true`, so `/api/feature_toggles/check/boolean` answers exactly as before once the row exists.
- [x] No CLI command or flag renamed or removed — `yarn mercato feature_toggles seed-defaults` still works and simply creates one fewer toggle.
- [x] No database table or column renamed or removed; no column type narrowed — no migration in this PR.
- [x] No config key renamed and no default changed silently. `defaults.json` is seed data, not a surface listed in `BACKWARD_COMPATIBILITY.md`; I grepped for the remaining identifiers and found no test or consumer asserting the file's shape.
- [x] Where a contract had to change: not applicable — no contract surface changed. Existing per-tenant overrides in `feature_toggle_overrides` are keyed on the toggle row, which is preserved on installations that already have it.

## 🧪 Test Coverage

Coverage is proportionate to the change and, importantly, tests the property that actually broke. `packages/core/src/modules/sales/__tests__/setupSeedsSalesChannelsToggle.test.ts` mocks the seeder and asserts that `seedDefaults` calls it exactly once with the EntityManager — the bug was a missing registration call, so a test that fails when the call is removed is the right regression guard. `packages/core/src/modules/sales/lib/__tests__/salesChannelsToggleSeed.test.ts` covers create-when-missing, no-op-when-present, and pins the identifier and `defaultValue: true` to what the client and server hooks poll, which usefully couples the definition to `useSalesChannelsEnabled`'s hardcoded fail-open default.

The gap is the state behind the major finding. Add to `lib/__tests__/salesChannelsToggleSeed.test.ts` a third case — "does not insert when a soft-deleted definition row exists" — that has `findOneWithDecryption` resolve a row and asserts `em.create`, `em.persist` and `em.flush` are never called. With the current `deletedAt: null` filter that case cannot even be expressed through the mock without also asserting the lookup shape, which is itself a sign the filter is wrong; once the filter is dropped the test is a straightforward mirror of the existing idempotency case.

No integration test is required here. The behaviour is a setup hook covered by unit tests, the change touches no API path and no UI path, and the `skip-qa` label is correctly applied under the automated-verification exemption: the diff contains no `.tsx` outside tests, nothing under `packages/ui/src/` or any `components/` directory, no database structure change and no API surface change.

wojciechszyjka:

- 🚀 `merge-queue` — the code review's one major finding was fixed in `db85fbdf` during this run and the re-review approves the change, so the PR is ready to merge once CI settles on the new head. `skip-qa` is present rather than `needs-qa`, so the QA-approval gate does not apply and nothing else holds it back.
- 🐛 `bug` — unchanged: this fixes a reported defect where the sales channels feature toggle was never registered and the check API answered 404 on every page mounting the hook.
- 📚 `documentation` — unchanged: the configuration decision guide recommended the pattern that caused the bug, so correcting that guidance is part of the fix.
- ⏭️ `skip-qa` — unchanged and re-verified against the final diff, including the autofix commit: no `.tsx` outside tests, nothing under `packages/ui/src/` or any `components/` directory, no database structure change, no API surface change, no `BACKWARD_COMPATIBILITY.md` contract touched, and automated tests for the changed behaviour ship in the same PR — so the automated-verification exemption applies.
- 🔹 `priority-medium` — unchanged: an ordinary bug rather than an incident, since both the client and server helpers fail open and sales channel UI stays visible; the cost is console noise that buries real errors.
- 🟡 `risk-medium` — unchanged: a single-area change with tests, touching tenant setup seeding that runs at bootstrap but adding only a guarded idempotent insert, with no migration, no auth or data-scoping code and no shared contract surface.

*(This comment supersedes the earlier `changes-requested` rationale posted by this skill at 14:0x UTC; it is rewritten in place on every label transition rather than duplicated.)*

wojciechszyjka (approved):
# 🔍 Re-review: fix(sales): register sales_channels_enabled from the sales module setup (#5503)

## 🎯 Summary

Re-review of the same change after the autofix commit this reviewer pushed to the PR branch. The original review requested changes on one major finding: `seedSalesChannelsToggle` filtered its existence check by `deletedAt: null`, while toggle deletion is soft (`feature_toggles/commands/global.ts:362`) and `feature_toggles_identifier_unique` covers every row regardless of `deleted_at` (`feature_toggles/migrations/Migration20251229215803.ts:9`). On an installation where the toggle had been deleted, the helper therefore inserted a duplicate and aborted sales seeding with a Postgres unique-constraint violation — on exactly the `yarn mercato seed:defaults --module sales` command the PR documents as the catch-up path for existing deployments, and, in the `seed:defaults` loop, for every organization after the first as well.

The fix drops the `deletedAt` filter so any existing row — live or soft-deleted — means "already registered, nothing to do". That matches the uniqueness semantics the module's own create command applies (`feature_toggles/commands/global.ts:93` looks up by identifier alone), and it leaves an operator's deletion in place rather than resurrecting the definition on every tenant seed. A regression test covers the soft-deleted state, and the existing lookup assertion was tightened from `objectContaining` to exact equality so a future `deletedAt` filter cannot be reintroduced silently.

Everything else in the original review stands: the diagnosis and the fix are correct, the three seeding call sites (`mercato init`, onboarding provisioning, `seed:defaults --module sales`) all reach the new hook, removing the `defaults.json` entry breaks nothing, and no contract surface is touched.

## Verdict

✅ **approve** — the major finding is resolved and covered by a test; only the three nits from the first review remain, and all of them are the author's call or better handled as a follow-up across all four modules that seed feature toggles.

## 🧪 Validation Gate

**Validation source: GitHub checks for the pre-fix head `ff8abcae` (all green), plus a local targeted test run for the autofix commit.** Runner: local (no compose `app` container was running). CI re-runs the full gate on the new head and this reviewer follows that outcome up in a bounded pass; the targeted run below is this review's own evidence for the two files the fix touches.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job on `ff8abcae`; the autofix changes no build input beyond one module file and its test. |
| `yarn generate` | ✅ PASS | `prepare` job; no generated-file churn in the diff. |
| `yarn build:packages` (rebuild) | ✅ PASS | `prepare` job. |
| `yarn i18n:check-sync` | ✅ PASS | `test` job; the autofix adds no user-facing strings. |
| `yarn typecheck` | ✅ PASS on `ff8abcae` · ⚠️ not conclusive locally | The `test` job typechecked the pre-fix head. Locally, `yarn workspace @open-mercato/core typecheck` cannot run cleanly in a fresh review worktree because `yarn generate` needs built packages first, so all 264 reported errors are `Cannot find module '#generated/entities.ids.generated'`. None of them involves a changed file, and the autofix only removes one property from an object literal, so the risk is nil — CI's run on the new head is the authoritative check. |
| `yarn test` | ✅ PASS | `test` job on `ff8abcae`, plus a local targeted jest run after the autofix: `salesChannelsToggleSeed.test.ts` (4 cases, including the new soft-delete regression) and `setupSeedsSalesChannelsToggle.test.ts` — 2 suites, 5 tests, all passing. |
| `yarn build:app` | ✅ PASS | `prepare` job. |

## Findings

### 💅 Nit

All three carry over from the first review unchanged, and none of them blocks the merge.

- **`salesChannelsToggleSeed.ts:19`** — the seeding write goes through `em.create` rather than the `feature_toggles.global.create` command, so it skips the audit entry and `emitCrudSideEffects`. The practical cost is limited (the admin list and `FeatureTogglesService` both read the base table, so the toggle appears and resolves correctly), and three sibling modules seed the same raw way. The right resolution is a shared `ensureFeatureToggles(em, definitions)` helper in `feature_toggles` that goes through the command bus with `systemActor: true` and carries the deleted-row guard for `portal`, `customers`, `sales` and `wms` at once — a follow-up, not this PR.
- **`salesChannelsToggleSeed.ts:6-13`** — `type: 'boolean' as const` is redundant inside an object already marked `as const`, and `em.create` re-lists all six fields instead of spreading the definition, which is the drift the extracted constant exists to prevent.
- **`apps/docs/docs/architecture/configuration-decision-guide.mdx:86`** — describing `defaults.json` as "only for demo/platform toggles that no module owns" reads slightly stronger than the file's remaining contents warrant, since `feature_toggles_ui` is owned by the `feature_toggles` module; half a sentence marking the leftovers as predating the rule would prevent a confused reader.

## 🔍 Inherited reviewer feedback (2 asks already on the PR)

Both from the earlier automated pass, [comment 5395752937](https://github.com/open-mercato/open-mercato/pull/5569#issuecomment-5395752937), and both accounted for.

- **Minor — existing deployments still need `yarn mercato seed:defaults --module sales`** → **declined, concurring with the original reasoning and adding one more.** An upgrade action would sit behind `UPGRADE_ACTIONS_ENABLED` and cost four i18n keys per locale; and the nearest self-healing precedent, `resolveWmsIntegrationToggleEnabled` (`wms/lib/wmsIntegrationToggles.ts:60`), re-seeds on the *server* path, which would not silence the reported console noise because that comes from the client hook calling `/api/feature_toggles/check/boolean`. Worth a follow-up issue against the check route; out of scope for a bug fix. Encouragingly, `FeatureTogglesService.resolveToggle` (`lib/feature-flag-check.ts:107`) never caches the `missing` result, so the 404s stop the instant an operator runs the catch-up command.
- **Nit — duplicated seeding shape across four modules** → **re-raised as a nit above**, now with the extra requirement that the shared helper also settle the deleted-row guard and the command-bus question for all four call sites.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the autofix changes only the internal lookup filter; `SALES_CHANNELS_TOGGLE_DEFINITION` and `seedSalesChannelsToggle` keep their shape and signature.
- [x] No database table or column renamed or removed; no column type narrowed — no migration in this PR.
- [x] Where a contract had to change: not applicable — nothing in `BACKWARD_COMPATIBILITY.md` is touched.

## 🧪 Test Coverage

The gap named in the first review is closed. `packages/core/src/modules/sales/lib/__tests__/salesChannelsToggleSeed.test.ts` now covers four cases: create-when-missing (with an exact-equality assertion pinning the lookup to `{ identifier }`, so a `deletedAt` filter cannot creep back in unnoticed), no-op-when-present, **no-op-when-soft-deleted**, and the identifier/`defaultValue` contract the client and server hooks rely on. `packages/core/src/modules/sales/__tests__/setupSeedsSalesChannelsToggle.test.ts` remains the regression guard for the actual bug — that `seedDefaults` calls the seeder at all — and is unchanged.

No integration test is required: the change is a setup hook with no API or UI path, which is also why `skip-qa` is correctly applied under the automated-verification exemption.

wojciechszyjka:

This run reviewed the PR as an independent reviewer — the earlier pass ran as the same automation account that authored it, so GitHub blocked a formal verdict and the report had to be posted as a plain comment. Both a formal review and the autofix now exist.

The review requested changes on one major finding and, being autofix-eligible, fixed it in the same run rather than handing it back. `seedSalesChannelsToggle` looked its definition up with `deletedAt: null`, but toggle deletion is a soft delete (`feature_toggles/commands/global.ts:362`) and `feature_toggles_identifier_unique` constrains every row regardless of `deleted_at` (`Migration20251229215803.ts:9`). On an installation where the toggle had been deleted, the lookup found nothing, the helper inserted a duplicate, and the flush raised a unique-constraint violation that aborted the sales `seedDefaults` hook — breaking exactly the `yarn mercato seed:defaults --module sales` command this PR documents as the catch-up path for existing deployments, and taking every remaining organization in that run down with it, since the loop shares one EntityManager. `db85fbdf` matches the identifier alone, the same uniqueness check `feature_toggles.global.create` applies, so any existing row means "already registered"; a soft-delete regression test was added and the existing lookup assertion tightened to exact equality so the filter cannot creep back in. Targeted validation after the fix: 2 suites, 5 tests, all passing.

Autofix: ran — 1 major finding fixed in 1 commit, 1 fix iteration. Three nits remain and are all the author's call or better handled as one follow-up: the seeding write bypasses the `feature_toggles.global.create` command (as three sibling modules also do), the definition constant is copied field by field into `em.create` instead of spread, and the revised doc sentence describes `defaults.json` slightly more strictly than its remaining contents warrant. The two asks inherited from the earlier pass are accounted for in the re-review: the "existing deployments need a manual command" minor is declined with reasons and belongs in a follow-up issue against the check route, and the "duplicated seeding shape" nit is re-raised, now also carrying the deleted-row guard for all four modules.

Labels moved to `merge-queue`; `skip-qa` means the QA-approval gate does not apply. `in-progress` is swapped for `ci-monitoring` because CI is re-running on the new head `db85fbdf` — that follow-up comment is the only thing this agent still owes the PR, and the PR is free for anyone to act on in the meantime.

wojciechszyjka:

All checks settled green on head `db85fbdf` — the autofix commit — within the follow-up budget, so the approving re-review stands unchanged and no pipeline label needed correcting. The PR is now `MERGEABLE` / `CLEAN` with `reviewDecision: APPROVED`.

Passing: `test`, `docker-build`, `documents-multi-instance`, `ephemeral-integration (none)`, `prepare`, `lint`, `ds-lint`, `scope`, `audit-scope`, `CodeQL` with all three `Analyze` variants, and `license/cla`. `[code]smith`, `audit`, `comment`, `merge-coverage` and `mutate` were skipped by their own path filters. Nothing failed.

The [`test` job](https://github.com/open-mercato/open-mercato/actions/runs/32763797954/job/97550528840) is the one that matters for this run: it carries `yarn typecheck`, `yarn test`, `yarn i18n:check-sync` and `yarn i18n:check-usage`, and it closes the one gap in this review's own evidence. The local typecheck could not be run conclusively in a fresh review worktree — `yarn generate` needs built packages first, so every local error was a missing `#generated/entities.ids.generated` and none touched a changed file. CI's green `test` job is the authoritative confirmation that the autofix typechecks and that the new soft-delete regression test passes in a clean environment, alongside the targeted local run of 2 suites / 5 tests reported in the re-review.

`ci-monitoring` is removed — this agent owes the PR nothing further. It carries `merge-queue` with `skip-qa`, so the QA-approval gate does not apply and it is ready for a maintainer to merge.

PR #5570, opened 2026-08-24
fix(ui): render AlertDescription as a div so block children stop breaking hydration (#5487)

### LINKED ISSUE #5487: bug: [ui] AlertDescription renders a <p>, so any block child produces a hydration error
## Summary

Split out of the review of #5481 (nit), where it was correctly judged out of that PR's scope. Filing it so the root cause is tracked rather than fixed one instance at a time.

## Problem

`AlertDescription` renders a `<p>`:

```tsx
// packages/ui/src/primitives/alert.tsx:255-259
const AlertDescription = React.forwardRef<HTMLParagraphElement, React.HTMLAttributes<HTMLParagraphElement>>(
  ({ className, ...props }, ref) => (
    <p ref={ref} className={cn('text-sm leading-5', className)} {...props} />
  ),
)
```

So every caller that puts a `<p>`, `<ul>`, or any other block element inside it produces invalid HTML and a React hydration error — `In HTML, <p> cannot be a descendant of <p>`. Each caller is one paragraph away from the bug, and it is not visible at the call site.

Known instances:

- `packages/enterprise/src/modules/record_locks/backend/settings/record-locks/page.tsx` — **fixed instance-locally in #5481** (item 2 of #5456), with a `p p` markup regression test
- `packages/ui/src/ai/parts/ConfirmationCard.tsx:223-231` — **still open**: nests both a `<p>` and a `<ul>` inside `AlertDescription`, producing the same hydration warning on the AI confirmation card

## Suggested fix

Change `AlertDescription` to render a `<div>` with the same classes, which removes the whole class of bug at once. It is typed `HTMLParagraphElement`, so the ref type narrows to `HTMLDivElement` — a small type change on a primitive.

That is an **ask-first** change under `packages/ui/AGENTS.md` ("Ask before changing primitive APIs"), which is why #5481 fixed only its own instance. Worth a maintainer decision:

1. Change `AlertDescription` to a `<div>` — fixes every caller, small typed API change on a primitive.
2. Leave the primitive and fix `ConfirmationCard.tsx` in place — smaller blast radius, leaves the trap armed for the next caller.

Option 1 is the better long-term answer if the ref-type change is acceptable.

## Verification

A useful guard either way: assert `container.querySelectorAll('p p')` is empty in the affected components' render tests, as `record-locks/__tests__/page.markup.test.tsx` does in #5481 — it pins the bug without scraping console output.

## Context

- Parent issue: #5456 · PR: #5481 · review nit by @adeptofvoltron

adeptofvoltron:
Closes #5487

## 🎯 Goal

Remove a whole class of React hydration errors from the admin UI: `AlertDescription` rendered a `<p>`, so every caller that nested a paragraph, a list, or any other block element inside it produced invalid HTML and the runtime warning `In HTML, <p> cannot be a descendant of <p>`. After this change the description is a `<div>`, and block children inside an `Alert` are valid everywhere.

## 🔍 Problem

`AlertDescription` is one of the most widely composed primitives in the codebase — 97 files import it. It rendered a `<p>`, which in HTML may only contain phrasing content. Any block child is therefore invalid markup, and the browser's parser silently closes the paragraph early, producing a DOM that does not match what React rendered on the server. React then fails hydration with `In HTML, <p> cannot be a descendant of <p>`.

The trap is invisible at the call site: nothing about `<AlertDescription>` suggests that a `<p>` inside it is illegal, so callers keep walking into it. #5481 hit it on `/backend/settings/record-locks` and fixed that one instance locally by rewriting the two help paragraphs as block `<span>`s, which is why #5487 was filed to track the root cause.

Scanning every `AlertDescription` usage in the monorepo shows the issue underestimated the blast radius — **11 call sites** nest block-level children, not the 2 the issue knew about:

| Site | Nested child |
|---|---|
| `packages/ui/src/ai/parts/ConfirmationCard.tsx:223` | `<p>` + `<ul>` |
| `packages/enterprise/…/record-locks/page.tsx:163` | `<p>` |
| `packages/ai-assistant/…/agents/AiAgentSettingsPageClient.tsx:2512` | `<div>` |
| `packages/ai-assistant/…/allowlist/AiTenantAllowlistPageClient.tsx:289` | `<div>` |
| `packages/ai-assistant/…/playground/AiPlaygroundPageClient.tsx:589` | `<div>` |
| `packages/core/…/settings/domain/components/Diagnostics.tsx:20, :45` | `<div>` |
| `packages/scheduler/…/ExecutionDetailsDialog.tsx:151, :178, :192, :206` | `<div>` |

## 🔍 Root Cause

`packages/ui/src/primitives/alert.tsx:255` declared the primitive as `React.forwardRef<HTMLParagraphElement, React.HTMLAttributes<HTMLParagraphElement>>` rendering a `<p>`. Nothing about the component's purpose requires a paragraph — the `text-sm leading-5` classes render identically on a `<div>` — so the element choice bought no styling or semantics while forbidding the composition its own callers needed.

## What Changed

- **`packages/ui/src/primitives/alert.tsx`** — `AlertDescription` now renders a `<div>` and is typed `React.forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDivElement>>`. Classes, `displayName`, and every other prop are unchanged, so the rendered styling is identical. A comment records why the element is a `<div>` so the next person does not "tidy" it back to a `<p>`.
- **`docs/design-system/component-apis.md`** — `AlertDescriptionProps` updated to `React.HTMLAttributes<HTMLDivElement>` to match the shipped primitive.

This was the ask-first decision the issue raised (option 1 of the two it proposed), taken deliberately over the narrower "fix `ConfirmationCard` only" path: option 2 would have left the trap armed for the next caller and for all nine `<div>`-nesting sites above.

No call site needed editing — all 11 become valid as a consequence of the primitive change. The `<span>` workaround #5481 added to the record-locks page is now unnecessary but left in place: it still renders correctly, and reverting it belongs in its own change rather than widening this diff.

## 🧪 Tests

Full gate run locally from the repo root — `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app`, plus `yarn ds:code-connect:check` from `packages/ui/AGENTS.md`. Everything passes except one pre-existing failure documented below.

`yarn test` — 32 of 33 turbo tasks green, including every package this change can reach:

| Package | Result |
|---|---|
| `@open-mercato/ui` | 233 suites, **1927 passed** |
| `@open-mercato/core` | 1426 suites, 11453 passed (2 skipped) |
| `@open-mercato/shared` | 184 suites, 1998 passed (5 skipped) |
| `@open-mercato/ai-assistant` | 109 suites, 1478 passed |
| `@open-mercato/cli` | 95 suites, 1802 passed |
| `@open-mercato/enterprise` | 62 suites, 520 passed |
| `@open-mercato/scheduler` | 23 suites, 420 passed |
| others (documents, queue, search, checkout, webhooks, storage-s3, telemetry, content, channel-*, gateway-stripe, sync-akeneo, app) | all green |

**Pre-existing failure, not from this PR: `create-mercato-app`.** Its standalone AI-harness oracle suite (`src/lib/writable-ast-oracles.test.ts`) fails locally because it shells out to the `codex` and `claude` CLIs and enforces a 120s wall-clock budget on a scaffolded app's `yarn typecheck`. Verified by baseline: stashing every change in this PR and re-running the package on clean `develop` produces a **byte-identical set of 161 failing assertions** — `comm` between the two failure lists is empty in both directions. No test fails here that does not also fail without the change.

New and updated regression coverage:

- `packages/ui/src/primitives/__tests__/alert.test.tsx` — a new `AlertDescription accepts block-level children` block asserts that nested `<p>`s and a nested `<ul>` leave `p p` and `p ul` empty in the rendered container, that the description element itself is a `DIV`, and that the forwarded ref is an `HTMLDivElement`. The pre-existing composition test's `toBe('P')` assertion becomes `toBe('DIV')` — that assertion is the pinned contract this PR intentionally changes.
- `packages/ui/src/ai/parts/__tests__/ConfirmationCard.test.tsx` — the existing `stale_version` test now also asserts `p p` and `p ul` are empty, pinning the still-open instance the issue named without scraping console output.
- `packages/ui/src/primitives/__tests__/notification.test.tsx` — `Notification` renders its description through `AlertDescription`, so its element assertion moves from `P` to `DIV`.

**Verified to fail against the pre-fix markup:** reverting `alert.tsx` to the `<p>` version and re-running the two suites produces 5 failures across both files; restoring the fix returns 36/36 passing. This follows the guard style the issue asked for and that `record-locks/__tests__/page.markup.test.tsx` established in #5481.

## 💥 Breaking Changes

One **typed** contract change on a DS primitive, which is why #5487 routed it through a maintainer decision under `packages/ui/AGENTS.md` ("Ask before changing primitive APIs"):

- `AlertDescriptionProps` widens from `React.HTMLAttributes<HTMLParagraphElement>` to `React.HTMLAttributes<HTMLDivElement>`, and the forwarded ref type changes from `HTMLParagraphElement` to `HTMLDivElement`.

Runtime impact is nil — no props, classes, or DOM attributes change apart from the tag name, and `<p>` and `<div>` render identically under the DS's Tailwind preflight. Compile-time impact in this repo is nil as well: `HTMLParagraphElement` appears nowhere outside `alert.tsx` and the DS doc, and **no call site anywhere in the monorepo passes a `ref` to `AlertDescription`**. A third-party module holding `useRef<HTMLParagraphElement>` or annotating an event handler with `HTMLParagraphElement` would need to swap the type to `HTMLDivElement`; there is no runtime deprecation bridge to stage, since the type is the entire surface.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5570"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790169328&installation_model_id=432243&pr_number=5570&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5570&signature=4d408b8a82d9b9dd1a34e26d4cfdd30d4401c791726701d0e6c99dbb3bb27394"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:

- 🔍 `review` — the run is complete and the branch is ready for code review; the full validation gate passed locally apart from one failure proven pre-existing by a clean-baseline comparison.
- 🐛 `bug` — this fixes a defect reported in #5487: invalid HTML that makes React fail hydration with `In HTML, <p> cannot be a descendant of <p>`, rather than adding any new capability.
- 🧪 `needs-qa` — the change edits a rendering primitive under `packages/ui/src/` that 97 files import, so it falls outside the automated-verification exemption and a human should confirm the AI confirmation card, notifications, and alert-bearing settings pages still look right in the browser.
- 🔹 `priority-medium` — an ordinary user-visible bug: it produces console hydration errors and unreliable markup on affected screens, but causes no outage, data loss, or security exposure.
- ⚠️ `risk-high` — the diff changes a shared design-system contract surface (`AlertDescriptionProps` and the forwarded ref type) that fans out to 97 importing files, which is a wider blast radius than an ordinary single-area fix even though the runtime behavior is limited to one tag name.

Manual QA is still pending: with `qaGate` enabled, this PR is not mergeable until a maintainer signs off with `qa-approved`.

adeptofvoltron:
🤖 `om-open-pr` completed: PR #5570 opened for #5487. Lock handed off to `om-auto-review-pr` — chain continues on this PR. Started: 2026-08-24T13:16:03Z.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock — code review of the #5487 fix. Started: 2026-08-24T13:16:35Z.

adeptofvoltron (commented):
# 🔍 Code Review

**Verdict: approve-equivalent — no blockers and no majors remain.** Submitted as a review comment rather than a formal approval because this PR was authored by the same automation account, and GitHub does not permit self-approval; a second human reviewer is still required before merge. One major finding was raised during this pass and fixed in-run (below), which is why the diff now carries two commits.

## 📋 Scope reviewed

The five-file diff plus the follow-up documentation commit: the `AlertDescription` primitive, the DS component-API doc, three test files, and the new `UPGRADE_NOTES.md` entry. Checklist sections 1–11 (architecture, security, ORM, API routes, migrations) do not apply — this change touches no module boundary, no query, no route, and no schema.

## ✅ What holds up

The root-cause fix is the right shape and the minimum one. `<p>` may only contain phrasing content, so the element choice — not any individual caller — was the defect; changing the primitive fixes all eleven nesting sites at once and required editing none of them. The `text-sm leading-5` classes, `displayName`, `data-slot` behavior, and prop spread are untouched, so the rendered output differs only by tag name and no visual regression is possible under Tailwind preflight.

Test coverage is genuine rather than decorative. The new guards were **verified to fail against the pre-fix markup** — reverting `alert.tsx` and re-running produces 5 failures across the two suites — which is the standard the issue asked for and the one `record-locks/__tests__/page.markup.test.tsx` set in #5481. The `p p` / `p ul` assertions pin the invalid nesting directly instead of scraping console output, so they cannot rot into false greens when React changes its warning text.

The two updated `toBe('P')` → `toBe('DIV')` assertions in `alert.test.tsx` and `notification.test.tsx` are correct to change rather than work around: they pinned the old contract, and this PR changes that contract deliberately with a maintainer decision behind it (#5487, `packages/ui/AGENTS.md` "Ask before changing primitive APIs").

## 🔴 Findings

**Major — resolved in this run.** `UPGRADE_NOTES.md` carried no entry for the change. `BACKWARD_COMPATIBILITY.md` does not enumerate design-system primitives among its 13 contract-surface categories, so this is not a formal protected-surface violation — but the deprecation protocol in root `AGENTS.md` requires documenting contract changes in `UPGRADE_NOTES.md`, and a third-party module holding `useRef<HTMLParagraphElement>` on this component gets a compile error with no written migration path. Fixed in `2433b1958`, which adds an entry to the unreleased `0.7.0 → 0.7.1` section stating the type change, the (nil) runtime impact, and the one-line migration. No runtime deprecation bridge is possible or needed, since the element type is the entire changed surface.

**Nit — not fixed, deliberately.** `.ai/ui-components.md:2406` and `:2458` describe the component as rendering "paragraphs". That prose is about layout density rather than the element name and remains readable, so rewording it would widen a deliberately narrow diff; worth folding into the next DS docs pass.

**Nit — not fixed, deliberately.** The `<span>` workaround #5481 added to the record-locks settings page is now unnecessary. It still renders correctly and its markup test still passes, so reverting it belongs in its own change rather than here.

**Observation, not a finding.** The three-line comment added above the primitive sits against the root `AGENTS.md` "no inline comments" rule. It is kept because it records *why* the element is a `<div>` — precisely the knowledge whose absence caused this bug — and because it matches the file's existing convention, where every primitive carries a `// Figma …` note.

## 🧪 Validation

The full `validation.commands` gate ran locally in order and passed: `build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `test`, `build:app`, plus `ds:code-connect:check` from `packages/ui/AGENTS.md`. `yarn generate` produced no drift. `@open-mercato/ui` is 1927/1927 green.

One `yarn test` task fails locally — `create-mercato-app`'s standalone AI-harness oracle suite, which shells out to the `codex` and `claude` CLIs and enforces a 120-second budget on a scaffolded app's typecheck. It is **pre-existing and unrelated**: stashing every change in this PR and re-running the package on clean `develop` yields a byte-identical set of 161 failing assertions, with `comm` empty in both directions. The follow-up commit is documentation-only and cannot affect any suite.

## 🔁 Backward compatibility

One typed break, deliberately taken as option 1 of the two the issue proposed, and now documented. `HTMLParagraphElement` appears nowhere in this repository outside `alert.tsx` and the DS doc, and no call site among the 97 importing files passes a `ref` — so in-repo compile impact is zero and the exposure is limited to external module authors, who now have an `UPGRADE_NOTES.md` entry telling them exactly what to change.

## ⚠️ CI

Green so far: `ds-lint`, `lint`, `scope`, `audit-scope`, `license/cla` all SUCCESS; CodeQL NEUTRAL. `prepare` and `Analyze (javascript-typescript)` were still in progress when this review was submitted, and the review was not held for them — a CI-result follow-up comment will land separately.

## 📸 Manual QA still required

The PR carries `needs-qa` and, with `qaGate` enabled, cannot merge until a maintainer applies `qa-approved`. Suggested exercise, derived from the diff:

- **P0** — open the AI assistant and drive a mutation to the stale-version state so the "Re-propose required" confirmation card renders. Confirm the browser console shows no `<p> cannot be a descendant of <p>` hydration error and the bulleted record-id list renders normally.
- **P1** — visit `/backend/settings/record-locks`, the scheduler execution-details dialog, and the AI assistant settings/allowlist/playground pages. Each nests block children in an `Alert`; confirm no hydration errors and unchanged spacing.
- **P2** — check any notification toast (which routes its description through this primitive) for unchanged appearance in both light and dark mode.

adeptofvoltron:
🤖 `om-auto-qa-pr` — 📸 UI verification (evidence-only)

Ran a real browser pass against this branch rather than relying on the unit tests alone. Environment: a **disposable** Postgres database (`om_qa_5487`, created and dropped for this run — the local `open-mercato` dev database was never written to), `yarn initialize` seeded from scratch, the app booted with `yarn dev`, and Playwright/Chromium driving `http://localhost:3111` as `superadmin@acme.com`. Console `error`/`warning` output and `pageerror` were captured throughout and filtered for hydration signatures.

**Result: 0 hydration errors and 0 invalid nesting across every page exercised.**

| Page | Alerts rendered | `p p` | `p ul` | Hydration errors |
|---|---|---|---|---|
| `/backend/config/ai-assistant/agents` | 2 | 0 | 0 | 0 |
| `/backend/config/ai-assistant/allowlist` | 0 | 0 | 0 | 0 |
| `/backend/config/ai-assistant/playground` | 0 | 0 | 0 | 0 |
| `/backend/config/ai-assistant/settings` | 0 | 0 | 0 | 0 |
| `/backend/customer_accounts/settings/domain` | 0 | 0 | 0 | 0 |
| `/backend/design-system` | 0 | 0 | 0 | 0 |

The meaningful row is the first one. `/backend/config/ai-assistant/agents` is one of the eleven affected call sites — `AiAgentSettingsPageClient.tsx:2512` nests a `<div>` inside `AlertDescription` — and both of its alerts ("Downgrade only — escalation is a code change" and "Prompt overrides are additive") render correctly with no console hydration error. The global "Demo Environment" banner, which is itself an `Alert` with a multi-block description, also renders correctly on every page.

**Two honest limitations, so this evidence is not over-read:**

1. **Most rows are not evidence.** The other five pages rendered no `Alert` at all in this seeded tenant, and `/backend/settings/record-locks` — the instance #5481 fixed — returned 404 because the enterprise `record_locks` module is not registered in this build. Those rows show "nothing broke", not "the fix was exercised".

2. **A live-DOM nesting count cannot detect the original bug.** Browsers auto-close an invalid `<p>` while parsing, so `document.querySelectorAll('p div')` reads `0` whether or not the markup was valid. I confirmed this by re-running the same capture with the primitive temporarily reverted to the `<p>` version: it also reported `p>div=0` and 0 console errors, so the browser pass **cannot** distinguish before from after and is corroborating evidence only.

**The rigorous proof of the fix remains the jsdom regression tests**, which render the tree exactly as authored and therefore do catch the invalid nesting: reverting `alert.tsx` makes 5 assertions fail across `alert.test.tsx` and `ConfirmationCard.test.tsx`, and restoring it returns 36/36 green.

Screenshots for all six pages were captured locally; the AI Agents page confirms both alerts render with correct spacing and no visual regression. Per the pipeline contract this run is evidence-only: the PR keeps `needs-qa`, and `qa-approved` remains a maintainer's call.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: #5487 fixed on this PR, code review submitted with one self-found finding fixed in-run (UPGRADE_NOTES entry), and UI verification captured as evidence. Lock released; `ci-monitoring` is set because 15 ephemeral-integration shards were still running — a CI-result follow-up comment will land when they settle.

adeptofvoltron:

**All checks green.** Every required check settled SUCCESS with no failures, cancellations, or timeouts: all 15 `ephemeral-integration` shards, plus `test`, `lint`, `ds-lint`, `docker-build`, `documents-multi-instance`, `prepare`, `scope`, `audit-scope`, `CodeQL` (python, javascript-typescript, actions), and `license/cla`. The `comment`, `mutate`, `audit`, and `[code]smith` jobs were skipped by their own path filters, which is their normal state for a change of this shape.

Two things worth recording beyond the green tick.

**CI's `test` job passing settles the one local failure.** `create-mercato-app`'s standalone AI-harness oracle suite fails on the authoring machine because it shells out to the `codex` and `claude` CLIs and enforces a 120-second budget on a scaffolded app's typecheck. The clean-`develop` baseline already showed a byte-identical set of 161 failing assertions with every change in this PR stashed; CI running the same job green independently confirms the failure was environmental and never attributable to this branch.

**The 15 integration shards are the strongest signal here.** This PR changes the element a widely-composed primitive renders, so the risk worth testing was never the primitive in isolation — it was whether some downstream assertion, selector, or snapshot depended on that `<p>`. The full integration matrix exercising the real app without a single failure is the evidence that nothing did.

`ci-monitoring` is removed; no further follow-up is owed from this agent. The PR remains `BLOCKED` for merge by design, not by any failure — it carries `needs-qa` under an enabled `qaGate` and still needs `qa-approved` from a maintainer, plus an approving review from a second human, since the automation account authored it and GitHub does not permit self-approval.

wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-24T18:33:14Z. Other auto-skills will skip this PR until the lock is released.

Note: the earlier `om-auto-fix-issue` chain (@adeptofvoltron) released its lock at 13:35Z and posted its CI-result follow-up at 13:49Z, so no live claim was overridden. This is an independent second-reviewer pass with autofix enabled.

wojciechszyjka (approved):
# 🔍 Code Review

## 🎯 Summary

This PR changes `AlertDescription` from rendering a `<p>` to rendering a `<div>`, which removes an entire class of React hydration failures across the admin UI. I reviewed it as an independent second reviewer (the earlier review on this PR came from the automation account that authored it, so GitHub could not accept it as a formal approval).

The change is the correct fix at the correct layer. `<p>` may contain only phrasing content, so any block child is invalid markup; the browser's parser closes the paragraph early, the resulting DOM diverges from what React rendered on the server, and hydration fails. The defect was therefore the element choice in the primitive, not any individual caller — and fixing the primitive fixes every caller at once without editing a single call site. I confirmed the nesting is real rather than theoretical: `packages/ui/src/ai/parts/ConfirmationCard.tsx:223` nests a `<p>` plus a `<ul>`, `packages/scheduler/…/ExecutionDetailsDialog.tsx:151` nests a `<div>` wrapping a `<pre>`, and `packages/enterprise/…/record-locks/page.tsx:163` nests two `<p>`s. All three become valid as a consequence of this one-line element swap.

The element choice bought nothing that is now lost. `AlertDescription` is styled purely by `text-sm leading-5`; `displayName`, the `className` merge, and the prop spread are all untouched, and `Alert` itself already renders a `<div role="alert">` wrapper, so a `<div>` description is valid in its parent context. I grepped every `.css` file in the repository for descendant selectors that would target a `p` inside an alert and found none, so no styling depends on the tag.

## ✅ Verdict: **approve**

No blockers, no majors, and no minors. Three nits are recorded below for the record; none of them justifies another review round, and I deliberately did **not** autofix any of them (reasoning under each).

## 🧪 Validation gate

**Validation source: GitHub checks.** CI has already run the full gate against this exact head (`2433b19`) and every check is green, so per this repository's review policy I did not re-run the gate locally. `test`, `lint`, `ds-lint`, `prepare`, `docker-build`, `documents-multi-instance`, `scope`, `audit-scope`, `license/cla`, all three CodeQL analyses, and all fifteen `ephemeral-integration` shards reported SUCCESS. `comment`, `mutate`, `audit`, `merge-coverage` and `[code]smith` were skipped by their own path filters, which is their normal state for a change of this shape. No required check was pending or failing at review time, so there is nothing to disclose about CI and no follow-up is owed.

The one local `yarn test` failure the PR description documents (`create-mercato-app`'s standalone AI-harness oracle suite) is independently corroborated by CI's `test` job passing on this head: whatever makes that suite fail on a developer machine — it shells out to the `codex` and `claude` CLIs under a wall-clock budget — does not reproduce in CI, which is consistent with the pre-existing/environmental classification the author gave it after a clean-baseline comparison.

## 💥 Breaking changes

I checked this against `BACKWARD_COMPATIBILITY.md` independently rather than taking the PR's own classification on trust, because a typed change to a widely imported design-system primitive is exactly the shape that should be checked.

**Not a protected contract surface.** Category 2 (*Type Definitions & Interfaces — STABLE*) is an **enumerated** list of module-system types — `Module`, `ModuleInfo`, `PageMetadata`, `ModuleSetupConfig`, `EventDefinition`, `EventPayload`, `EntityExtension` and the rest. Design-system primitive prop types are not among them, and none of the other twelve categories covers them either. The deprecation protocol's step 5 (a spec with a "Migration & Backward Compatibility" section) is therefore not triggered, and step 4 (the `UPGRADE_NOTES.md` entry) was satisfied anyway in `2433b19`. That is the right call regardless of whether the letter of the contract demanded it.

**The compile-time impact is smaller than the PR claims — in the safe direction.** `HTMLParagraphElement` and `HTMLDivElement` are *structurally identical* in `lib.dom.d.ts`: both extend `HTMLElement` and add exactly one deprecated `align: string` property plus the same `addEventListener`/`removeEventListener` overload pair. TypeScript compares them structurally, so they are mutually assignable, which means an external module holding `useRef<HTMLParagraphElement>(null)` or annotating a handler as `React.MouseEvent<HTMLParagraphElement>` keeps compiling against the new signature. The only genuine break is a runtime `instanceof HTMLParagraphElement` narrowing or a DOM/CSS selector keyed on `p` — both vanishingly rare, and neither occurs in this repository: `HTMLParagraphElement` appears nowhere outside `UPGRADE_NOTES.md`, no call site among the 97 importing files passes a `ref`, and `packages/ui/figma/alert.figma.tsx` passes only a plain string child.

**No duplication with the base branch.** `origin/develop` still carries the `<p>` version at `alert.tsx:255`, and the five most recent commits touching that file are all unrelated DS work, so nothing here re-lands work that already merged.

## 🧪 Test coverage

The coverage is genuine rather than decorative, which is the distinction that matters for a regression guard. The three new cases in `packages/ui/src/primitives/__tests__/alert.test.tsx:193-231` assert the mechanism directly — `container.querySelectorAll('p p')` and `('p ul')` are empty, and the forwarded ref is an `HTMLDivElement` — rather than scraping React's console warning, so they cannot rot into false greens when React rewords its message. That choice matters here specifically: jsdom builds the tree through `createElement` rather than the HTML parser, so invalid nesting *persists* in the DOM instead of being silently fixed up, which is precisely what makes these selectors a working detector.

The two `toBe('P')` → `toBe('DIV')` edits in `alert.test.tsx:192` and `notification.test.tsx:29` are correct to change rather than work around: they pinned the old contract, and this PR changes that contract deliberately with a maintainer decision behind it (#5487, and the "ask before changing primitive APIs" rule in `packages/ui/AGENTS.md`). The added `ConfirmationCard.test.tsx:136-137` assertions pin the specific instance the issue was filed about. The author reports the guards were verified to fail against the pre-fix markup (5 failures on revert), which is the standard #5487 asked for.

## 🔴 Findings

**Nit — `UPGRADE_NOTES.md:36`, an example that is branch-dependent.** The entry tells module authors that callers "previously worked around the restriction by rendering block `<span>`s inside the description (for example the record-locks settings page shipped in #5481)". On this PR's base branch that example does not hold: `packages/enterprise/…/record-locks/page.tsx:163` nests two real `<p>`s and contains no `<span>` at all, which is exactly why it appears in the PR's own table of broken call sites. The reason is that **#5481 merged into `main`, not `develop`** — its merge commit `cb0b77e` is not an ancestor of `origin/develop`, so neither the `<span>` rewrite nor the `page.markup.test.tsx` guard it added exists on this base. I am **not** asking for a change and did not autofix it, because `UPGRADE_NOTES.md` addresses people consuming released versions, and the statement is accurate for `main`, where 0.7.1 ships from. Recording it so the discrepancy is not mistaken later for a lost commit.

Worth noting as reassurance on the same point: I read `page.markup.test.tsx` as it exists on `origin/main` and it asserts only that `p p` is empty plus that the help text renders. Both assertions continue to hold once `AlertDescription` is a `<div>`, so merging `develop` forward into `main` will not break that guard.

**Nit — `docs/design-system/component-apis.md:476`, pre-existing doc drift in the block being edited.** The file documents `export type AlertActionProps` and `enforcement.md:183` / `migration-tables.md:363` both show `<AlertAction>` in their migration examples, but `AlertAction` does not exist: `alert.tsx:265` exports only `Alert`, `AlertDescription`, `AlertTitle`, `alertVariants` and `alertIconBadgeVariants`, and the identifier appears in no `.ts`/`.tsx` file in the repository. This predates the PR and sits two lines from the line it legitimately changes. Fixing it is a docs-accuracy job across three files and belongs in its own change rather than widening a deliberately narrow diff — a follow-up issue is the right home for it.

**Nit — the `// Renders a <div>, not a <p>` comment at `alert.tsx:255-257`.** Root `AGENTS.md` asks for self-documenting code without inline comments. I would keep this one: it records *why* the element is what it is, which is precisely the knowledge whose absence caused the bug, and it matches the file's established convention where each primitive carries a `// Figma …` note. Flagging only so the next reader knows the deviation was considered rather than overlooked.

## 🔁 Related surfaces checked (no action needed)

`DialogDescription`, `SheetDescription` and `DrawerDescription` all delegate to Radix's `Dialog.Description`, which renders a `<p>` by default — the same latent trap in three more primitives. I searched for block-level children nested inside them across `packages/` and `apps/` and found **none**, so there is no live sibling bug today and nothing this PR needs to absorb. Hardening those three the same way would be a reasonable follow-up, but it is out of scope here and should not hold this change.

On accessibility: replacing `<p>` with `<div>` does drop the implicit `paragraph` role that some screen readers use for paragraph-by-paragraph navigation. The net effect is neutral-to-positive, because the content sits inside a `role="alert"` live region that is announced as a unit, and the eleven affected callers now get their *real* `<p>` elements rendered as paragraphs instead of being silently torn apart by the parser.

wojciechszyjka:

- 🚀 `merge-queue` — the code review passed with no blockers, majors or minors, and CI is fully green on this head, so the PR moves out of `review` and into the merge queue; `needs-qa` stays on, so the QA-approval gate still holds the actual merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — kept from the original triage and still correct: this changes a rendering primitive that 97 files import, which is well outside the automated-verification exemption, so a human should exercise the affected screens before it merges. UI evidence from an earlier browser pass is already attached to this PR, but a maintainer must be the one to convert that evidence into `qa-approved`; a code review may never apply that label from reading a diff.
- 🐛 `bug` — this fixes a defect reported in #5487 (invalid HTML causing React hydration failures) rather than adding any new capability, so the category label is unchanged.
- 🔹 `priority-medium` — kept as triaged: the bug produces console-level hydration errors and a server/client DOM mismatch on several admin screens, which is a real defect worth the current cycle, but it causes no outage, data loss, or security exposure that would justify `priority-high`.
- ⚠️ `risk-high` — kept as triaged, and the review reinforces it rather than softening it: the change edits a design-system primitive imported by 97 files, so its blast radius is genuinely repository-wide even though the diff is one element name. The review found the actual risk well controlled — no styling depends on the tag, `Alert` already wraps its children in a `<div>`, and no call site passes a `ref` — but the breadth of the surface is what the label measures.

wojciechszyjka:

A browser pass was already run against this branch by an earlier automation step and reported zero hydration errors — that evidence is in the comment above and is worth reading first, but it does not substitute for this gate: nobody has applied `qa-approved`, and a code review may not apply it from a diff. Treat the routes below as confirmation of that evidence rather than a fresh investigation.

The whole change is one element swap, so there is exactly one thing to watch everywhere: **keep the browser console open**. The bug this fixes is invisible on screen — it shows up only as `In HTML, <p> cannot be a descendant of <p>` in the console and as a server/client DOM mismatch. Layout should be pixel-identical before and after, because `<p>` and `<div>` render the same under the design system's Tailwind preflight.

### P1 — AI assistant confirmation card (the instance the issue was filed about)

**Where to click**
- The AI assistant chat, driven to a mutation whose preview goes stale (`packages/ui/src/ai/parts/ConfirmationCard.tsx:223`) so the "Re-propose required" alert renders with its bulleted record list.

**What to verify**
- The alert renders its explanatory paragraph followed by the bulleted list of changed record ids → the console shows no `<p> cannot be a descendant of <p>` error and no hydration warning.
- Reload the page in that state → the card renders identically after a full server-rendered load, which is the path where the mismatch actually bites.

**What can go wrong**
- The bulleted list detaches from the paragraph above it or loses its indent, which would mean the parser is still restructuring the markup.
- A hydration warning appears on reload but not on client-side navigation — check both, since only the reload path exercises the server/client comparison.

### P1 — the other alerts that nest block children

**Where to click**
- `/backend/settings/record-locks` — the strategy help alert (`page.tsx:163`) with its two stacked paragraphs.
- The scheduler execution-details dialog for a **failed** run — the error alert wraps a `<pre>` inside a `<div>` (`ExecutionDetailsDialog.tsx:151`), and the other three alerts in that dialog nest `<div>`s too.
- The AI assistant settings, tenant-allowlist, and playground pages, and the domain-settings diagnostics panel — each nests a `<div>` inside its alert description.

**What to verify**
- Each screen loads with an unchanged alert: same tint, same icon badge, same padding, same line spacing → console clean on both first load and reload.
- On the scheduler dialog specifically, the failed-run error message keeps its monospace formatting and its line wrapping inside the alert.

**What can go wrong**
- Vertical spacing shifting inside an alert would suggest something other than the tag changed; the classes are supposed to be untouched.
- The record-locks help text collapsing its two paragraphs onto one line.

### P2 — notification toasts

**Where to click**
- Any action that raises a toast — notification descriptions route through this primitive (`packages/ui/src/primitives/notification.tsx`).

**What to verify**
- Toast appearance, `opacity-70` description tone, and dismissal behavior are unchanged in **both light and dark mode**.

**What can go wrong**
- A description losing its muted tone, which would indicate the class merge changed rather than just the element.

### Perceived performance
- Cold-load one of the P1 routes with an empty cache and confirm the first useful shell appears as before, the alert is not visibly re-painted after hydration, and the page behaves the same at a mobile viewport width. A hydration mismatch typically shows up as a visible flash of re-rendered content, so this check doubles as a second detector for the bug itself.

### Pass/fail

wojciechszyjka:

An independent second-reviewer pass (@wojciechszyjka) over the full diff found **no blockers, no majors and no minors** — the element swap is the correct fix at the correct layer, the regression guards assert the mechanism directly rather than scraping React's console output, and the backward-compatibility classification holds up against `BACKWARD_COMPATIBILITY.md`: design-system primitive prop types are not one of its thirteen enumerated contract surfaces, and `HTMLParagraphElement` and `HTMLDivElement` are structurally identical in `lib.dom.d.ts`, so even an external `useRef<HTMLParagraphElement>` keeps compiling. Three nits are recorded in the review body, all deliberately left alone: a `UPGRADE_NOTES.md` example that is accurate for `main` but not for this PR's `develop` base (because #5481 merged to `main` only), pre-existing `AlertAction` doc drift that predates this diff, and the intentional explanatory comment on the primitive.

**Autofix: ran, no changes made.** The loop was eligible (`--autofix`) but had nothing actionable to do — there were no blockers, majors or minors to fix, and each of the three nits is either out of scope for this diff, correct as written for the branch it addresses, or a deliberate author decision I agree with. Pushing a commit for any of them would have invalidated a fully green CI run and bought nothing.

CI was already settled green on this head before the review was submitted — all 15 `ephemeral-integration` shards plus `test`, `lint`, `ds-lint`, `prepare`, `docker-build`, `documents-multi-instance`, `scope`, `audit-scope`, CodeQL and `license/cla` — so no `ci-monitoring` follow-up is owed and none was promised. The PR now sits in `merge-queue` with `needs-qa` retained: the only thing standing between it and merge is a maintainer converting the attached UI evidence into `qa-approved`.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-25T02:09:29Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

pkarw:
🤖 `om-auto-qa-pr` — 📸 UI QA evidence (evidence-only)

## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ PASS — every rendered `AlertDescription` is a `<div>`, block children stay nested where React put them, the alert typography is byte-identical to the Figma contract, and no hydration diagnostics were logged on any surface driven.
**Environment:** `http://127.0.0.1:5001` · role `superadmin` (seeded demo account) · browser `playwright` (Chromium)
**Verified:** `fix/issue-5487-alertdescription-block-children` @ `e1fc45f`

The environment is the repo's own ephemeral QA stack — a disposable pgvector testcontainer, migrated and seeded fresh, with a **production** build (`NODE_ENV=production`, `next start`) on an isolated port. That matters here: hydration is exactly what this PR is about, so the app had to be verified as a real server-rendered-then-hydrated build rather than a dev server.

### Scenario (P1 — design-system primitive `AlertDescription`, rendered across the admin UI)

**Where to click:** `/backend/design-system?family=feedback` · `/backend/config/ai-assistant/agents` · `/backend/config/ai-assistant/playground`

| 1 | Cold-load the design-system Feedback gallery as superadmin and probe every rendered `Alert` and `Notification`. | Each description is a `DIV` with the unchanged `text-sm leading-5` typography; no invalid paragraph nesting anywhere in the document. | 11 alerts rendered. The 3 that compose `AlertDescription` all report `DIV`, class `text-sm leading-5`, `font-size: 14px`, `line-height: 20px`, `display: block`. All 3 `Notification` descriptions report `DIV`. `p p` = 0, block-in-`p` = 0, empty ghost descriptions = 0. Shell painted in **542 ms**. | ✅ |
| 2 | Force `/api/ai_assistant/ai/agents` to fail at the network layer so the AI agent settings page renders its error `Alert` — the call site at `AiAgentSettingsPageClient.tsx:2512` that nests a `<div>` inside the description. | The description is a `DIV` containing the message `SPAN` **and** the action `DIV`, with the Retry button a true descendant rather than hoisted out by the HTML parser. | `descriptionTag: DIV`, `childTags: [SPAN, DIV]`, `nestedDivIsDescendant: true`, `retryButtonInsideDescription: true`. Retry renders inside the alert body and is clickable. `p p` = 0, block-in-`p` = 0, ghosts = 0. | ✅ |
| 3 | Repeat the same forced failure on the AI playground (`AiPlaygroundPageClient.tsx:589`) to confirm a second, independent call site. | Identical valid structure. | `descriptionTag: DIV`, `childTags: [SPAN, DIV]`, `nestedDivIsDescendant: true`, `p p` = 0, block-in-`p` = 0. | ✅ |
| 4 | Re-run the gallery probe at a 390×844 mobile viewport. | Same structure and no invalid nesting at mobile width. | 11 alerts, all 3 descriptions `DIV`, 3 notification descriptions `DIV`, `p p` = 0, block-in-`p` = 0, ghosts = 0. | ✅ |
| 5 | Parse the pre-fix and post-fix description markup in the same live browser to document the failure mode being removed. *(Parser demonstration, not an app observation.)* | The old `<p class="text-sm leading-5">` wrapper cannot hold block children; the new `<div>` can. | **before:** `descriptionTag: P`, `descriptionChildTags: []`, `topLevelChildren: [P, P, P, P]` — the parser closed the description early and hoisted both paragraphs out, leaving empty ghost paragraphs. **after:** `descriptionTag: DIV`, `descriptionChildTags: [P, P]`, `topLevelChildren: [DIV]`. | ✅ |

**Hydration check.** Every step attached `console` and `pageerror` listeners before navigating and filtered them for `cannot be a descendant of`, `hydrat`, `validateDOMNesting`, `did not match`, and minified React errors #418/#421/#422/#423/#425. **Zero matches on every surface.** The only console errors recorded at all are the `net::ERR_FAILED` entries this run injected on purpose in steps 2 and 3 to reach the error state.

**Perceived performance.** The changed route cold-loads with a useful shell in 542 ms, the alert content paints with it (no loading flash on a static gallery), and interaction stays responsive — the Retry button inside the error alert renders enabled and hit-testable. Nothing about the tag change touches layout: `display: block`, `14px`/`20px` are unchanged from the Figma "Paragraph/Small" contract the comment in `alert.tsx` cites.

### Notes for QA

- **Blast radius independently confirmed.** A repo-wide AST-ish scan for `<AlertDescription>` bodies containing a block element reproduced **exactly** the 11 call sites the PR body lists — same files, same lines. The PR's claim that the issue underestimated the reach (11, not 2) is accurate.
- **Two of the eleven call sites were driven in a real browser** (AI agent settings, AI playground). The rest need state this fresh ephemeral database does not have.
- **`/backend/settings/record-locks` was not exercised — environment limit, not a failure.** Enterprise modules are gated behind `OM_ENABLE_ENTERPRISE_MODULES`, which the ephemeral QA environment leaves off, so that route returns 404 in this build. Enabling it means a full rebuild since the module registry is resolved at build time.
- ⚠️ **One factual correction to the PR description.** It states that the `<span>` workaround #5481 added to the record-locks page is still in place, so that call site's nesting is no longer live. On this PR's head — and on `origin/develop` — that page renders two real `<p>` elements directly inside `AlertDescription` (`page.tsx:163`). It is therefore a **live `<p>`-in-`<p>` instance** that this fix resolves. The description understates the fix's value; the change itself is unaffected, and nothing needs to be edited for merge.
- **Not exercised, for a follow-up QA pass with seeded state:** the scheduler `ExecutionDetailsDialog` alerts (need a completed/failed execution), the customer-accounts DNS/TLS diagnostics (need a `dns_failed` / `tls_failed` domain mapping), the AI tenant allowlist env banner (needs `OM_AI_AVAILABLE_PROVIDERS`), and `ConfirmationCard`'s `stale_version` card (needs an AI mutation proposal to go stale). All four are `<div>`/`<p>`/`<ul>` children of the same primitive, so they are covered by the same mechanism verified above.
- **Secrets:** only the seeded demo `superadmin` account was used. No token, `.env` value, or credential appears in any screenshot or in this report.

### Screenshots

pkarw:
🤖 `om-auto-qa-pr` — 🧪 follow-up test scenario

This change ships no browser-level test. The three jest suites it updates (`alert.test.tsx`, `ConfirmationCard.test.tsx`, `notification.test.tsx`) pin the contract at the component level under jsdom, which is the right place for the tag assertion — but jsdom does **not** run the HTML parser's paragraph auto-closing rule, so it cannot observe the actual regression: a real browser silently hoisting block children out of a `<p>` description. The UI QA above was manual; lock it in with an automated test (run `/om-integration-tests`).

1. **Setup:** none beyond the seeded demo tenant — sign in via the repo's `login(page, 'superadmin')` helper. The surface is the design-system gallery plus one forced error state, so no fixtures are created and nothing needs cleaning up.
2. **Act:**
   - Navigate to `/backend/design-system?family=feedback` and wait for the `Feedback` heading and the first `[data-slot="alert"]`.
   - Navigate to `/backend/config/ai-assistant/agents` with `page.route('**/api/ai_assistant/ai/agents**', route => route.abort('failed'))` armed, and wait for `[data-ai-agent-settings-error]`. This is the cheapest deterministic way to render an `AlertDescription` that nests a block child.
3. **Assert:**
   - Every `AlertDescription` element (the child of the alert body carrying `text-sm leading-5` that is not the `H5` title) reports `tagName === 'DIV'`, `font-size: 14px` and `line-height: 20px`, so the tag change stays a tag change and never a styling change.
   - `document.querySelectorAll('p p').length === 0` and `document.querySelectorAll('p div, p ul, p ol').length === 0` on both routes — the parser-level invariant the jsdom suites cannot cover.
   - On the error alert, the description's `childTags` are `['SPAN', 'DIV']` and `description.contains(retryButton)` is true, proving the action block is a real descendant rather than a hoisted sibling.
   - A `console` / `pageerror` listener armed before navigation records no entry matching `/cannot be a descendant of|hydrat|validateDOMNesting|Minified React error #(418|421|422|423|425)/`.
4. **Teardown:** none — no records are created. Unroute the aborted endpoint if the spec continues past the error-state assertions.

A working implementation of exactly this scenario already exists as the throwaway spec this QA run drove; it lives in the run's artifacts directory and can be lifted into a module-local `__integration__/` folder with only the config and import paths adjusted.

pkarw:

No pipeline or meta label was changed — this run used the default `--evidence-only` mode, so the screenshots and the pass/fail report are the deliverable and a QA reviewer (or a re-run with `--self-qa-signoff`) decides the `qa-approved` verdict. The PR keeps `needs-qa` without `qa-approved`, so the QA-approval merge gate still holds it in `merge-queue` exactly as before this run.

PR #5571, opened 2026-08-24
fix(devices): return camelCase API response keys (#5513)

### LINKED ISSUE #5513: bug: [Devices] the API returns snake_case fields; every other module returns camelCase
## Summary
`GET /api/devices` and `GET /api/devices/admin/devices` return database column names verbatim. Every other
module returns camelCase. A client written against the platform convention reads `undefined` for every field on
this one module.

`devices` — both the user route and the admin route:
```
id, tenant_id, organization_id, user_id, device_id, platform,
client_app_version, os_version, locale, push_provider,
push_token_updated_at, last_seen_at, created_at, updated_at
```

Same request shape, other modules:
- `sales/orders` → `id`, `orderNumber`, `status`, `customerEntityId`, `shippingMethodSnapshot`, …
- `warranty_claims` → `id`, `claimNumber`, `claimType`, `customerId`, `awaitingStaffReply`, …

## Expected
camelCase, per the convention every sibling module follows.

## Note on compatibility
`devices` shipped in this release window, so renaming now is far cheaper than after third-party clients exist.

adeptofvoltron:
## Summary

`GET /api/devices`, `GET /api/devices/admin/devices` and `GET /api/devices/admin/devices/:id` returned the raw database column names (`device_id`, `user_id`, `last_seen_at`, …) while every sibling module returns camelCase, so a client written against the platform convention read `undefined` for every field on this one module. All three routes now return **camelCase as the canonical shape**.

Both list routes declared `list.fields: deviceListFields` with no `transformItem`, so the CRUD factory handed the query engine's raw projection straight to the client; the admin detail route had the same problem via a hand-written snake_case `serializeDevice()`. The fix uses `transformItem` — the mechanism `warranty_claims`, `eudr`, `wms` and the `sales` line routes already use for exactly this — plus a shared detail serializer.

Note that the *request* side was already camelCase: `deviceListSortFieldMap` accepts `lastSeenAt` / `createdAt` / `updatedAt`. Only the response side disagreed with itself.

## 💥 Contract change — shipped as a bridge, not a rename

`BACKWARD_COMPATIBILITY.md` § 7 (API Route URLs, STABLE) says response fields MUST NOT be removed, and a retired surface keeps working for at least one minor version. A rename is a removal plus an addition, so the deprecation protocol applies even though the module shipped inside the current release window.

**Every snake_case key is still returned**, holding the same value as its camelCase counterpart, marked deprecated in the OpenAPI document and slated for removal no earlier than the next minor:

| Deprecated alias | Canonical key |
|---|---|
| `tenant_id` / `organization_id` | `tenantId` / `organizationId` (list routes only) |
| `user_id` / `device_id` | `userId` / `deviceId` |
| `client_app_version` / `os_version` | `clientAppVersion` / `osVersion` |
| `push_provider` / `push_token_updated_at` | `pushProvider` / `pushTokenUpdatedAt` |
| `last_seen_at` / `created_at` / `updated_at` | `lastSeenAt` / `createdAt` / `updatedAt` |

Removal later is a one-site change (drop the alias spread from `transformDeviceListItem` / `serializeDeviceDetail` and the deprecated keys from the two schemas), and `TC-DEV-007` asserts the aliases today so it cannot land unnoticed.

The issue argued for a hard rename since `devices` shipped in this release window. That is a maintainer call rather than an automated one, so this PR takes the reversible path — **say the word and I will drop the aliases**, which is a small follow-up commit.

## Changes

- `api/deviceList.ts` — `transformDeviceListItem` (reads either spelling, normalizes timestamps to ISO) + `toDeprecatedSnakeCaseAliases`; `deviceListItemSchema` gains the camelCase keys and describes the snake_case ones as deprecated
- `api/deviceSerialization.ts` (new) — `serializeDeviceDetail` + `deviceDetailItemSchema`, extracted from the admin `[id]` route so both surfaces share one casing decision and the serializer is unit-testable
- Both list routes wire `transformItem`; the admin detail route uses the shared serializer/schema
- First-party consumers migrated: devices admin list + edit pages (including the `optimisticLockUpdatedAt` source) and the push-notifications device picker
- Spec `.ai/specs/2026-08-24-devices-api-camelcase-responses.md` (with the required Migration & Backward Compatibility section), `UPGRADE_NOTES.md` entry, devices `AGENTS.md` rule

## Deliberately unchanged

- **`push_token` stays out of every response** under either spelling — asserted in both test layers.
- **The `exportScope=full` branch of the CRUD factory still bypasses `transformItem`** (it serializes `rawItems` through `normalizeFullRecordForExport`). That is framework-wide behavior shared by every module using `transformItem`, so a devices-local fix would be the wrong place for it. The paged export path — what a UI export button uses — does go through the transform.
- Route URLs, methods, request bodies, query params, sorting, filtering, scoping, ACLs, commands, events and the database schema.

## 🧪 Tests

- `__tests__/response-casing.test.ts` — list transform emits every camelCase key, aliases mirror their source, Date columns normalize to ISO, camelCase input is tolerated, keys beyond the projection (`cf_*`) survive, `push_token` never becomes `pushToken`, both schemas parse their serializer's output; detail serializer keeps ISO timestamps, nulls absent optionals and does not invent `tenant_id` / `organization_id`
- `__integration__/TC-DEV-007.spec.ts` — camelCase contract on both list routes and the detail route, plus the alias bridge
- `TC-DEV-001` / `TC-DEV-005` / `TC-DEV-006` are left asserting the snake_case keys on purpose: for the duration of the bridge they are the regression proof that no 0.7.0 consumer broke

## ✅ Validation

Full gate from `.ai/agentic.config.json`, run locally: `build:packages` → `generate` → `build:packages` → `i18n:check-sync` → `i18n:check-usage` → `typecheck` → `test` → `build:app`.

Everything passes except one package in `yarn test`: **`create-mercato-app` fails at baseline on this machine** (161 failing assertions — the `live Codex adapter` / `live Claude adapter` harness tests plus a timing-sensitive `yarn typecheck must complete within 120000ms` gate). I verified this is pre-existing by checking out untouched `origin/develop` in the same worktree and re-running: **identical failure count (161), identical suites**. The other 32 packages pass, including the `@open-mercato/core` suite that covers this change. CI is the authority on that package.

## ⚠️ Reviewer note

The devices list DataTable column ids change with the accessor keys, so a user's **saved perspective for `devices.list`** no longer matches the stored column ids and falls back to defaults once. Self-healing on the next save; flagging it since it is user-visible.

Closes #5513

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5571"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790169849&installation_model_id=432243&pr_number=5571&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5571&signature=f30a6cc62479bff15f7edc9649573175c78b3eebc74c465a44c0fcc2d8d3f9b9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.ai/specs/2026-08-24-devices-api-camelcase-responses.md
+++ b/.ai/specs/2026-08-24-devices-api-camelcase-responses.md
@@ -1,0 +1,104 @@
+# Devices API camelCase Responses
+
+## TLDR
+
+The `devices` module returns raw database column names (`device_id`, `last_seen_at`, `client_app_version`, …) from `GET /api/devices`, `GET /api/devices/admin/devices` and `GET /api/devices/admin/devices/:id`, while every sibling module returns camelCase. This spec makes camelCase the canonical response shape on all three routes and keeps each snake_case key alongside it as a documented, deprecated alias for one minor version, so no existing consumer breaks while the platform convention is restored.
+
+Scope is limited to the response serialization of those three GET routes and the first-party UI that reads them, per [issue #5513](https://github.com/open-mercato/open-mercato/issues/5513). Route URLs, HTTP methods, request bodies, query parameters, sorting, filtering, ACL features, commands, events, database schema and the push-token secrecy rule are unchanged.
+
+## Problem Statement
+
+Both device list routes declare `list.fields: deviceListFields` and no `transformItem`, so the CRUD factory hands the query engine's raw projection straight to the client:
+
+```
+id, tenant_id, organization_id, user_id, device_id, platform,
+client_app_version, os_version, locale, push_provider,
+push_token_updated_at, last_seen_at, created_at, updated_at
+```
+
+The admin detail route has the same problem for a different reason: its local `serializeDevice()` helper maps entity properties onto hand-written snake_case keys.
+
+Every other module transforms the projection before serializing — `sales/orders` returns `orderNumber` / `customerEntityId`, `warranty_claims` returns `claimNumber` / `awaitingStaffReply`, `wms/inventory/balances` and the `eudr` routes do the same through their own `transformItem`. A client written against the platform convention therefore reads `undefined` for every field on this one module. The devices module is also the module a mobile client integrates against first, which is exactly the audience least likely to discover the inconsistency from the codebase.
+
+`deviceListSortFieldMap` already accepts camelCase sort fields (`lastSeenAt`, `createdAt`, `updatedAt`), so the request side of the contract is *already* camelCase — only the response side disagrees with itself.
+
+## Proposed Solution
+
+- Add `transformDeviceListItem` to `api/deviceList.ts` and wire it as `list.transformItem` on both list routes. It reads each column under either spelling, normalizes timestamps to ISO strings, and emits camelCase keys.
+- Extract the admin detail serializer into `api/deviceSerialization.ts` as `serializeDeviceDetail`, emitting the same camelCase keys, so both surfaces share one casing decision and the serializer becomes unit-testable outside a route module.
+- Keep every snake_case key alongside its camelCase counterpart as a deprecated alias, produced by a single shared helper (`toDeprecatedSnakeCaseAliases`) so removing the bridge later is a one-site change.
+- Document both spellings in the OpenAPI item schemas (`deviceListItemSchema`, `deviceDetailItemSchema`), with each deprecated key describing the canonical key that replaces it.
+- Migrate the first-party consumers — the devices admin list and edit pages, and the push-notifications send page's device picker — to the canonical keys.
+- Cover the change with module unit tests and a new integration spec that pins both the camelCase contract and the alias bridge.
+
+## Overview
+
+`transformItem` is the established mechanism for this in the CRUD factory: the list handler applies it to every raw item before custom-field decoration, translation overlay, access logging and serialization, and the paged export path reuses the transformed items. Using it here means the fix is one option on each route rather than a bespoke response wrapper, and it matches how `warranty_claims`, `eudr`, `wms` and the `sales` line routes already solve the same problem.
+
+The transform spreads the raw record before overlaying the camelCase keys, so anything the projection carries beyond the declared field set (custom-field keys, columns added later) survives rather than being silently dropped by a hand-built object.
+
+Two behaviors are deliberately **not** changed:
+
+- **`push_token` stays out of every response.** It is absent from `deviceListFields` and never read by either serializer; the tests assert its absence under both spellings.
+- **The `exportScope=full` branch of the CRUD factory still bypasses `transformItem`** (it serializes `rawItems` through `normalizeFullRecordForExport`). That is framework-wide behavior shared by every module using `transformItem`, so a devices-local fix would be the wrong place to change it. The paged export path — the one a UI export button uses — does go through the transform.
+
+## Acceptance Criteria
+
+1. `GET /api/devices` returns `id`, `tenantId`, `organizationId`, `userId`, `deviceId`, `platform`, `clientAppVersion`, `osVersion`, `locale`, `pushProvider`, `pushTokenUpdatedAt`, `lastSeenAt`, `createdAt`, `updatedAt`.
+2. `GET /api/devices/admin/devices` returns the same camelCase key set.
+3. `GET /api/devices/admin/devices/:id` returns `item` with `id`, `userId`, `deviceId`, `platform`, `clientAppVersion`, `osVersion`, `pushProvider`, `pushTokenUpdatedAt`, `lastSeenAt`, `createdAt`, `updatedAt`.
+4. Every snake_case key previously returned by those three routes is still present and holds the same value as its camelCase counterpart.
+5. The detail response does not grow `tenant_id` / `organization_id` keys it never carried.
+6. Timestamp columns serialize as ISO strings under both spellings.
+7. `push_token` / `pushToken` is absent from every list and detail response.
+8. Sorting (`sortField=lastSeenAt|createdAt|updatedAt`), the `?platform=` / `?userId=` filters, self-serve user scoping, admin org scoping and the advanced-filter hardening on the self list are unchanged.
+9. The OpenAPI document describes both spellings, marking the snake_case keys as deprecated aliases.
+10. The devices admin list page, the devices edit page (including its optimistic-lock header source) and the push-notifications device picker read the canonical camelCase keys.
+11. Unit tests cover the list transform and the detail serializer; an integration spec covers both list routes and the detail route, asserting the camelCase contract and the alias bridge.
+
+## Architecture
+
+| File | Change |
+|------|--------|
+| `packages/core/src/modules/devices/api/deviceList.ts` | Adds `transformDeviceListItem`, `toDeprecatedSnakeCaseAliases` and the shared `toRecord`/`readString`/`toIso` readers; `deviceListItemSchema` gains the camelCase keys and describes the snake_case ones as deprecated |
+| `packages/core/src/modules/devices/api/deviceSerialization.ts` | New. `serializeDeviceDetail` + `deviceDetailItemSchema` for the admin detail route |
+| `packages/core/src/modules/devices/api/route.ts` | `list.transformItem: transformDeviceListItem` |
+| `packages/core/src/modules/devices/api/admin/devices/route.ts` | `list.transformItem: transformDeviceListItem` |
+| `packages/core/src/modules/devices/api/admin/devices/[id]/route.ts` | Uses the shared serializer and detail schema instead of its local copies |
+| `packages/core/src/modules/devices/backend/devices/page.tsx` | Row type, column accessors and cell readers use camelCase |
+| `packages/core/src/modules/devices/backend/devices/[id]/page.tsx` | Detail type, header fields, `initialValues` and `optimisticLockUpdatedAt` use camelCase |
+| `packages/core/src/modules/push_notifications/backend/push_notifications/send/page.tsx` | Device picker reads `deviceId` / `pushProvider` |
+
+## Migration & Backward Compatibility
+
+`BACKWARD_COMPATIBILITY.md` § 7 (API Route URLs, STABLE) states that response fields MUST NOT be removed and that a retired surface keeps working for at least one minor version. Renaming a response key is a removal plus an addition, so the deprecation protocol applies even though the module shipped inside the current release window.
+
+The change therefore ships as a **bridge, not a rename**:
+
+1. **No removal in this release.** Every snake_case key a 0.7.0 client reads is still returned, with an identical (now normalized) value.
+2. **Deprecated in place.** `toDeprecatedSnakeCaseAliases` carries an `@deprecated` JSDoc naming the replacement keys, and the OpenAPI item schemas describe each alias as "Deprecated alias for `<camelCaseKey>`; removed in the next minor release."
+3. **Bridge lifetime.** The aliases are removed no earlier than the next minor release. Removal is a single-site change (drop the alias spread from `transformDeviceListItem` and `serializeDeviceDetail`, and the deprecated keys from the two schemas) and is guarded by `TC-DEV-007`, which asserts the aliases today so their removal cannot land unnoticed.
+4. **Documented.** `UPGRADE_NOTES.md` carries a `0.7.0 → 0.7.1` entry with the full key mapping and the client action.
+
+Request-side contracts are untouched: query parameters, sort field names, request bodies and the `PUT`/`DELETE`/`POST` response shapes (`{ id, deviceId, revived }`, `{ ok: true }`) were already camelCase and are unchanged. Database columns keep their snake_case names — this is a serialization change only, with no migration.
+
+The one visible difference for a caller that already read the snake_case keys is normalization: a timestamp column that previously serialized as whatever the query engine produced now always serializes as an ISO-8601 string under both spellings.
+
+## Test Plan
+
+**Unit** — `packages/core/src/modules/devices/__tests__/response-casing.test.ts`
+
+- The list transform emits every camelCase key and maps values from the raw projection.
+- The deprecated aliases mirror their camelCase source exactly.
+- Date columns normalize to ISO strings; camelCase input is tolerated.
+- Keys beyond the declared projection (e.g. `cf_*`) survive the transform.
+- `push_token` never becomes `pushToken`.
+- Both item schemas parse their serializer's output.
+- The detail serializer emits camelCase with ISO timestamps, keeps its aliases, nulls absent optional columns, and does not invent `tenant_id` / `organization_id`.
+
+**Integration** — `packages/core/src/modules/devices/__integration__/TC-DEV-007.spec.ts`
+
+- Self list, admin list and admin detail all expose the camelCase key set for a freshly registered device, and never expose the push token.
+- Every deprecated snake_case alias mirrors its camelCase source on both the list and the detail route.
+
+Existing `TC-DEV-001`, `TC-DEV-005` and `TC-DEV-006` continue to assert the snake_case keys and are left unchanged: for the duration of the bridge they are the regression proof that no 0.7.0 consumer broke.
--- a/.ai/specs/2026-08-24-devices-api-camelcase-responses.md
+++ b/.ai/specs/2026-08-24-devices-api-camelcase-responses.md
@@ -27,7 +27,8 @@ Every other module transforms the projection before serializing — `sales/order
 - Add `transformDeviceListItem` to `api/deviceList.ts` and wire it as `list.transformItem` on both list routes. It reads each column under either spelling, normalizes timestamps to ISO strings, and emits camelCase keys.
 - Extract the admin detail serializer into `api/deviceSerialization.ts` as `serializeDeviceDetail`, emitting the same camelCase keys, so both surfaces share one casing decision and the serializer becomes unit-testable outside a route module.
 - Keep every snake_case key alongside its camelCase counterpart as a deprecated alias, produced by a single shared helper (`toDeprecatedSnakeCaseAliases`) so removing the bridge later is a one-site change.
-- Document both spellings in the OpenAPI item schemas (`deviceListItemSchema`, `deviceDetailItemSchema`), with each deprecated key describing the canonical key that replaces it.
+- List both spellings in the OpenAPI item schemas (`deviceListItemSchema`, `deviceDetailItemSchema`) and state the deprecation in the endpoint descriptions, which is where the generator actually renders prose.
+- Pin the admin list's export columns to the canonical key set so the aliases do not double every column in the CSV/JSON/XML export.
 - Migrate the first-party consumers — the devices admin list and edit pages, and the push-notifications send page's device picker — to the canonical keys.
 - Cover the change with module unit tests and a new integration spec that pins both the camelCase contract and the alias bridge.
 
@@ -52,9 +53,10 @@ Two behaviors are deliberately **not** changed:
 6. Timestamp columns serialize as ISO strings under both spellings.
 7. `push_token` / `pushToken` is absent from every list and detail response.
 8. Sorting (`sortField=lastSeenAt|createdAt|updatedAt`), the `?platform=` / `?userId=` filters, self-serve user scoping, admin org scoping and the advanced-filter hardening on the self list are unchanged.
-9. The OpenAPI document describes both spellings, marking the snake_case keys as deprecated aliases.
-10. The devices admin list page, the devices edit page (including its optimistic-lock header source) and the push-notifications device picker read the canonical camelCase keys.
-11. Unit tests cover the list transform and the detail serializer; an integration spec covers both list routes and the detail route, asserting the camelCase contract and the alias bridge.
+9. The OpenAPI document lists both spellings, and its rendered endpoint descriptions state that the snake_case keys are deprecated aliases scheduled for removal. (The `.describe()` markers on the schema properties themselves do **not** reach the document: the shared `zodToJsonSchema` converter emits per-property descriptions for parameters and request bodies but not for object schema properties. That is a framework-wide limitation, not a devices one, so the notice is carried by the endpoint description instead and the markers are kept as in-code contract documentation.)
+10. The admin list export emits one column per canonical key — the deprecated aliases must not double the export's column set.
+11. The devices admin list page, the devices edit page (including its optimistic-lock header source) and the push-notifications device picker read the canonical camelCase keys.
+12. Unit tests cover the list transform and the detail serializer; an integration spec covers both list routes and the detail route, asserting the camelCase contract and the alias bridge.
 
 ## Architecture
 
@@ -63,7 +65,8 @@ Two behaviors are deliberately **not** changed:
 | `packages/core/src/modules/devices/api/deviceList.ts` | Adds `transformDeviceListItem`, `toDeprecatedSnakeCaseAliases` and the shared `toRecord`/`readString`/`toIso` readers; `deviceListItemSchema` gains the camelCase keys and describes the snake_case ones as deprecated |
 | `packages/core/src/modules/devices/api/deviceSerialization.ts` | New. `serializeDeviceDetail` + `deviceDetailItemSchema` for the admin detail route |
 | `packages/core/src/modules/devices/api/route.ts` | `list.transformItem: transformDeviceListItem` |
-| `packages/core/src/modules/devices/api/admin/devices/route.ts` | `list.transformItem: transformDeviceListItem` |
+| `packages/core/src/modules/devices/api/admin/devices/route.ts` | `list.transformItem: transformDeviceListItem`, plus `list.export.columns` pinned to `deviceExportColumnFields` |
+| `packages/core/src/modules/devices/api/openapi.ts` | Exports `DEPRECATED_SNAKE_CASE_NOTICE` and appends it to the generated list description |
 | `packages/core/src/modules/devices/api/admin/devices/[id]/route.ts` | Uses the shared serializer and detail schema instead of its local copies |
 | `packages/core/src/modules/devices/backend/devices/page.tsx` | Row type, column accessors and cell readers use camelCase |
 | `packages/core/src/modules/devices/backend/devices/[id]/page.tsx` | Detail type, header fields, `initialValues` and `optimisticLockUpdatedAt` use camelCase |
--- a/UPGRADE_NOTES.md
+++ b/UPGRADE_NOTES.md
@@ -24,6 +24,30 @@ most of the patterns listed below in a user's codebase.
 
 ## 0.7.0 → 0.7.1 (unreleased)
 
+### Device API responses are camelCase; the snake_case keys are deprecated aliases (#5513)
+
+`GET /api/devices`, `GET /api/devices/admin/devices` and `GET /api/devices/admin/devices/:id` returned the raw database column names (`device_id`, `user_id`, `last_seen_at`, …) while every other module returns camelCase. A client written against the platform convention read `undefined` for every field on this one module. All three routes now return camelCase keys as the canonical shape:
+
+| Deprecated alias | Canonical key |
+|------------------|---------------|
+| `tenant_id` | `tenantId` (list routes only) |
+| `organization_id` | `organizationId` (list routes only) |
+| `user_id` | `userId` |
+| `device_id` | `deviceId` |
+| `client_app_version` | `clientAppVersion` |
+| `os_version` | `osVersion` |
+| `push_provider` | `pushProvider` |
+| `push_token_updated_at` | `pushTokenUpdatedAt` |
+| `last_seen_at` | `lastSeenAt` |
+| `created_at` | `createdAt` |
+| `updated_at` | `updatedAt` |
+
+`id`, `platform` and `locale` are spelled the same either way and are unchanged. Nothing was removed: **every snake_case key is still returned**, holding the same value as its camelCase counterpart, and is marked deprecated in the OpenAPI document. The aliases are removed no earlier than the next minor release.
+
+Request-side contracts are untouched — query parameters, the `sortField` values (`lastSeenAt`, `createdAt`, `updatedAt`, already camelCase), request bodies, and the `POST`/`PUT`/`DELETE` response shapes are unchanged, as are the database columns themselves. The one behavior difference for a caller already reading the snake_case keys is that timestamp columns now always serialize as ISO-8601 strings under both spellings. `push_token` remains absent from every response under either spelling.
+
+**Action for API consumers:** switch to the camelCase keys. A client that keeps reading the snake_case ones works unchanged until the aliases are removed.
+
 ### Interaction participants may omit `userId` — external calendar guests (#5115)
 
 `interactionParticipantSchema` required `participants[].userId` to be a UUID, so an attendee with no person/customer/staff record — an external guest identified only by their email — could not be recorded at all. `userId` is now optional; a participant must still be identifiable, so one without a `userId` **must** carry a valid email address (`participants[].email`), and one with neither is still rejected with a `400`.
--- a/packages/core/src/modules/devices/AGENTS.md
+++ b/packages/core/src/modules/devices/AGENTS.md
@@ -23,6 +23,14 @@ delivery logic** (sender, providers, delivery rows, workers live in the `push_no
   — they carry undo snapshots, query-index side effects, and domain events.
 - Treat `push_token` as a secret: **never** include it in list/response field sets. Only
   `push_provider` and `push_token_updated_at` are exposed.
+- Serialize every GET response in **camelCase**, like every other module. The list routes run
+  `transformDeviceListItem` (`api/deviceList.ts`) over the query engine's raw projection and the admin
+  detail route uses `serializeDeviceDetail` (`api/deviceSerialization.ts`). Both also emit the legacy
+  snake_case keys as **deprecated aliases** via `toDeprecatedSnakeCaseAliases`, kept for one minor
+  version per `BACKWARD_COMPATIBILITY.md` § 7 and pinned by `TC-DEV-007`; when the bridge is dropped,
+  remove the alias spread from those two functions and the deprecated keys from `deviceListItemSchema`
+  / `deviceDetailItemSchema`. Adding a column means adding it to `deviceListFields`, the transform, and
+  the schemas. See `.ai/specs/2026-08-24-devices-api-camelcase-responses.md` (#5513).
 - Honor the `pushToken` tri-state on `PUT`: absent key = leave unchanged; explicit `null` = clear
   (revoked OS permission) and bump `push_token_updated_at`. The command uses own-property presence.
 - Keep the list route's CRUD cache tag aligned with the command's `resourceKind`. The list reads
--- a/packages/core/src/modules/devices/__integration__/TC-DEV-007.spec.ts
+++ b/packages/core/src/modules/devices/__integration__/TC-DEV-007.spec.ts
@@ -1,0 +1,165 @@
+import { expect, test, type APIRequestContext } from '@playwright/test'
+import { apiRequest, getAuthToken } from '@open-mercato/core/modules/core/__integration__/helpers/api'
+import { getTokenScope, readJsonSafe } from '@open-mercato/core/modules/core/__integration__/helpers/generalFixtures'
+
+// TC-DEV-007: response casing. Every other module serializes its API responses in camelCase, and the
+// devices routes used to return the raw database column names instead (#5513). These tests pin the
+// camelCase contract on both list routes and the admin detail route, and pin the deprecated
+// snake_case aliases that stay alongside for one minor version so their removal is a deliberate,
+// test-visible change.
+
+type RegisterResult = { id: string; deviceId: string; revived: boolean }
+type DeviceItem = Record<string, unknown>
+type DeviceListResponse = { items: DeviceItem[]; total?: number }
+
+const SELF_PATH = '/api/devices'
+const ADMIN_PATH = '/api/devices/admin/devices'
+
+const CAMEL_LIST_KEYS = [
+  'tenantId',
+  'userId',
+  'deviceId',
+  'clientAppVersion',
+  'osVersion',
+  'pushProvider',
+  'pushTokenUpdatedAt',
+  'lastSeenAt',
+  'createdAt',
+  'updatedAt',
+] as const
+
+const DEPRECATED_ALIASES: Array<[string, string]> = [
+  ['user_id', 'userId'],
+  ['device_id', 'deviceId'],
+  ['client_app_version', 'clientAppVersion'],
+  ['os_version', 'osVersion'],
+  ['push_provider', 'pushProvider'],
+  ['push_token_updated_at', 'pushTokenUpdatedAt'],
+  ['last_seen_at', 'lastSeenAt'],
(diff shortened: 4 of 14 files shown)


adeptofvoltron:

- 🔍 `review` — the run finished with a complete, validated change and the PR is ready for code review.
- 🐛 `bug` — issue #5513 reports broken behavior: the devices API returned database column names, so a client following the platform's camelCase convention read `undefined` for every field.
- 🧪 `needs-qa` — the change touches three `.tsx` surfaces (the devices admin list and edit pages, and the push-notifications device picker), which disqualifies it from the automated-verification exemption; the list and edit pages need to be exercised manually to confirm every column and form field still renders real values.
- 🔺 `priority-high` — the fix's value is time-boxed to the release window in which `devices` debuted. Landing the camelCase shape and its deprecation bridge in 0.7.1 is what makes removing the snake_case aliases in the following minor possible; if it slips, the wrong shape exists across two released minors and the bridge effectively becomes permanent.
- ⚠️ `risk-high` — this modifies an API response contract surface (`BACKWARD_COMPATIBILITY.md` § 7) on three routes at once. The change is deliberately additive (every snake_case key is retained as a deprecated alias, so nothing that worked before stops working) and is covered by unit and integration tests, which argues for `risk-medium`; the higher label wins because a mistake here is visible to every external device client rather than contained to one screen.

adeptofvoltron:
🤖 `om-auto-fix-issue` — chain lock transferred from issue #5513 to this PR by @adeptofvoltron at 2026-08-24T13:24:59Z. Continuing with `om-auto-review-pr`; other auto-skills will skip this PR until the lock is released.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock from `om-auto-fix-issue` at 2026-08-24T13:25:51Z. Reviewing this PR with `--autofix`; the lock stays held for the chain.

adeptofvoltron:
# 🔍 Code Review — `fix(devices): return camelCase API response keys (#5513)`

**Reviewer:** `om-auto-review-pr` (autofix-eligible: own PR, `--autofix`) · **Mode:** first review · **Head:** `a16d579c5`
**Validation source:** GitHub checks (per the repo-local review extension), with a local run of the full `validation.commands` gate as corroboration.

## ⚖️ Verdict

**Would approve — but GitHub refuses the review.** This PR is authored by the same account running the review, and `gh pr review --approve` fails with `Review Can not approve your own pull request`. No approving review exists on this PR, so **it still needs a human code reviewer** and I have deliberately left it in the `review` pipeline state rather than moving it to `merge-queue` on an approval that does not exist.

Two findings were raised against the original head (`68845c685`) and **both were fixed in `a16d579c5`** by the autofix loop before this report. Nothing actionable remains on the current head.

## 💥 Findings

### 🟠 Medium — Deprecated aliases doubled every column in the admin list export · `packages/core/src/modules/devices/api/admin/devices/route.ts:57` — **FIXED**

The admin list enables exports (it declares no `export` key, so `resolveAvailableExportFormats` falls back to `csv/json/xml/markdown`) and declared no `list.export.columns`. `prepareExportData` therefore fell through to `buildDefaultExport`, which spreads each item into the column set. With the transform now emitting both spellings, a CSV export of `/api/devices/admin/devices?format=csv` would have carried 25 columns instead of 14 — every aliased field twice, e.g. both `deviceId` and `device_id`.

This was introduced by this PR: before it, items only carried the snake_case keys.

**Fix applied:** `deviceExportColumnFields` pins the export to the canonical camelCase set, wired via `export: { columns: … }` with `toHeaderLabel`, matching the `warranty_claims` convention. This also decouples the export shape from the alias bridge, so dropping the aliases later cannot change the export. Covered by a new unit assertion that fails if a canonical key is ever added to the projection without a matching export column.

### 🟠 Medium — The deprecation notice never reached the OpenAPI document · `packages/core/src/modules/devices/api/deviceList.ts:130` — **FIXED**

The deprecated keys were marked with zod `.describe('Deprecated alias for …')`, and the spec's acceptance criteria claimed the generated document would surface that. It did not. The shared converter calls `extractZodDescription` only for parameters and request bodies (`packages/shared/src/lib/openapi/generator.ts:687,713,726`); its `ZodObject` branch (`:460-503`) builds `properties` without ever consulting it. I verified against the generated artifact: every property of the devices list item — deprecated and canonical alike — came out with no `description`.

This is a framework-wide limitation rather than a devices one, so fixing the shared generator is out of scope for a bug fix on one module.

**Fix applied:** the notice now lives on the endpoint descriptions, which do render — confirmed in `openapi.generated.json` for all three GET routes. The `.describe()` markers are kept as in-code contract documentation with a comment explaining why they do not render, and the spec's acceptance criterion was corrected to state what is actually delivered rather than what was assumed.

## ✅ What holds up

- **The mechanism is the conventional one.** `list.transformItem` is exactly how `warranty_claims`, `eudr`, `wms` and the `sales` line routes solve this; the factory applies it before custom-field decoration, translation overlay, access logging and the paged export, so no downstream stage sees the raw projection.
- **The contract change follows the deprecation protocol.** `BACKWARD_COMPATIBILITY.md` § 7 forbids removing response fields; every snake_case key is retained with an identical value, the removal is staged for the next minor, `UPGRADE_NOTES.md` carries the mapping and client action, and a spec with a *Migration & Backward Compatibility* section is referenced — the protocol's five steps in full.
- **The transform is non-destructive.** Spreading the raw record before overlaying camelCase means custom-field (`cf_*`) keys and any future column survive instead of being dropped by a hand-built object — verified by a unit test.
- **`push_token` secrecy is intact.** It is absent from `deviceListFields` and read by neither serializer; both test layers assert its absence under both spellings.
- **Scoping and filtering are untouched.** The self list keeps its `mergeAdvancedFilterTree` hardening and actor scope, the admin list keeps factory tenant/org scoping, and `sortFieldMap` already spoke camelCase — the request side needed no change.
- **First-party consumers were migrated, not left on the bridge**, including `optimisticLockUpdatedAt` on the edit form, which would have silently disabled optimistic locking had it been missed.

## 🧪 Tests

Coverage is proportionate: 41 unit tests pass in `@open-mercato/core` (`__tests__/response-casing.test.ts` adds 10, covering the camelCase key set, alias mirroring, ISO normalization, camelCase-input tolerance, `cf_*` survival, push-token absence, both schemas, and the export column set), and `TC-DEV-007` pins the contract end-to-end on both list routes and the detail route. Leaving `TC-DEV-001/005/006` asserting snake_case is the right call — for the bridge's lifetime they are the proof that no 0.7.0 consumer broke.

## 🧭 Non-actionable notes

- **`exportScope=full` still bypasses `transformItem`.** The factory serializes `rawItems` through `normalizeFullRecordForExport` on that branch, so a full-scope export emits snake_case. This predates the PR and applies to every module using `transformItem`; correctly left alone here. Worth a framework-level follow-up issue.
- **Saved `devices.list` perspectives reset once.** The DataTable column ids move with the accessor keys, so a stored column layout no longer matches and falls back to defaults until re-saved. Unavoidable given the rename, self-healing, and already disclosed in the PR body.

## 🚦 CI

All completed checks pass on the previous head (`Analyze`, `audit-scope`, `ds-lint`, `lint`, `scope`, `license/cla`); `prepare` and `Analyze (javascript-typescript)` were pending. Checks are re-running for `a16d579c5`. The local gate ran green end-to-end — `build:packages` → `generate` → `build:packages` → `i18n:check-sync` → `i18n:check-usage` → `typecheck` → `test` → `build:app` — with one exception: the `create-mercato-app` package fails at baseline on this machine (161 assertions, the `live Codex`/`live Claude` adapter suites plus a timing-sensitive 120s typecheck budget). I confirmed it is pre-existing by checking out untouched `origin/develop` in the same worktree and re-running to an identical failure count. CI is the authority there.

adeptofvoltron:

This PR carries `needs-qa` with no `skip-qa`, and `qaGate` is on — it cannot merge until `qa-approved` is added. It is **not** in `merge-queue`: the reviewing agent authors this PR and GitHub refuses self-approval, so a human code review is still outstanding alongside QA. QA reviewer: when you pick it up, swap `review` for `qa`, then run the routes below.

### P0 — Devices admin list (`/backend/devices`)

**Where to click**
- `/backend/devices` — the cross-user device list, gated by `devices.admin`

**What to verify**
- Every column renders real values, not blanks: Device (a monospaced device id), Platform, User, App version, OS version, Push provider, Last seen. This is the core of the fix — the page now reads `deviceId`/`userId`/`clientAppVersion`/`osVersion`/`pushProvider`/`lastSeenAt` instead of the snake_case keys, so a missed rename shows up as an empty cell or `—`.
- The User column links to `/backend/users/<id>/edit` and shows a display name where the user is resolvable, falling back to a raw id (without a link) when the admin lacks `auth.users.list`.
- The Platform and User filters still narrow the list, and paging still works.
- Deactivate on a row still soft-deletes the device and the row disappears after the reload.

**What can go wrong**
- Empty or `—` cells across a column — the accessor key and the response key disagree.
- The Last seen column showing a raw ISO string or `Invalid Date` instead of a formatted local time.
- An org-restricted `devices.admin` seeing devices outside their organization scope (scoping is untouched by this PR, but it is the highest-consequence thing on this screen).

### P0 — Devices edit form (`/backend/devices/<id>`)

**Where to click**
- `/backend/devices` → row action **Edit**

**What to verify**
- The header block shows the real device id, platform, and user (linked when resolvable) — not blanks.
- App version, OS version and Push provider load **pre-filled with the device's stored values**, then save successfully and return to the list with the new values visible.
- **Optimistic locking still fires.** Open the same device in two tabs, save in the first, then save in the second — the second must surface a conflict, not silently overwrite. The form's lock header is derived from the detail response's `updatedAt`, which this PR renamed from `updated_at`; if that rename were wrong the form would send no version header and concurrent edits would be lost silently with no visible error. This is the single most important check in this PR.

**What can go wrong**
- Empty form fields on load (the values exist in the DB but the response key changed).
- The second save succeeding with no conflict warning — the lock header went missing.

### P1 — Push notification device picker (`/backend/push_notifications/send`)

**Where to click**
- `/backend/push_notifications/send` → pick a recipient user who has at least one push-capable device

**What to verify**
- The device dropdown lists that user's devices with the label `<device id> · <platform> · <push provider>`, and the count in the "All devices" option is non-zero.
- Devices with no push provider are still excluded from the list.
- Selecting an iOS device still hides the Android-only channel id field, and silent mode still hides it too.
- Sending to a selected device still delivers.

**What can go wrong**
- An empty dropdown, or labels rendering as `undefined · ios · undefined` — the picker filters on `pushProvider` and labels with `deviceId`, both renamed here. An empty list also disables Send, which looks like "this user has no devices".

### P1 — Admin device export

**Where to click**
- `/backend/devices` → export the list to CSV (or call `GET /api/devices/admin/devices?format=csv` directly)

**What to verify**
- The CSV has **one column per field** — 14 columns with camelCase-derived headers. It must not contain both `Device Id` and `Device_id`-style duplicates.
- Values line up under their headers and the file opens cleanly in a spreadsheet.

**What can go wrong**
- Duplicated columns (the regression this PR's second commit fixes — worth confirming it is actually gone).
- Empty columns, meaning an export column names a key the transform does not emit.

### P2 — API contract spot-check (optional, for an API-minded reviewer)

**Where to click**
- `GET /api/devices` and `GET /api/devices/admin/devices/<id>` with a valid session

**What to verify**
- Responses carry camelCase keys **and** their snake_case counterparts with identical values (the deprecation bridge).
- No `pushToken` or `push_token` under either spelling.
- The generated API docs for these routes state that the snake_case keys are deprecated aliases.

### Pass/fail
- All routes pass → remove the `qa` label and add `qa-approved`. Note that the QA gate is only one of two gates here: this PR also still needs an approving human code review, because the automation cannot approve its own PR.
- Any route fails → remove the `qa` label, add `qa-failed`, and comment describing the failure.

adeptofvoltron:
🤖 `om-auto-review-pr` — completed: **would-approve, blocked from self-approving** (2 medium findings found and fixed in `a16d579c5`; no actionable findings remain). Lock retained — chain continues to `om-auto-qa-pr`.

adeptofvoltron:
## 📸 UI QA evidence — `om-auto-qa-pr` (evidence-only)

**Verdict: PASS** — all 8 required steps passed. Driven through Playwright/Chromium against a real instance of this branch (`a16d579c5`) at 1440×900, plus a 390×844 mobile pass.

**Environment.** Next.js dev server on `http://localhost:3617` running this PR's worktree, against a disposable clone of the local demo database (`createdb -T`), signed in as the seeded `admin` role user. No container runtime is available on this host, so the ephemeral test-env path was not used; nothing was written to a shared database.

### Results

| # | Priority | Check | Result |
|---|---|---|---|
| 1 | P0 | Devices list renders a real value in every column | ✅ PASS |
| 2 | P0 | Device edit form pre-fills the stored values and the header | ✅ PASS |
| 3 | P0 | Optimistic locking rejects a stale second save (409) | ✅ PASS |
| 4 | P0 | The first writer's value survived the rejected save | ✅ PASS |
| 5 | P1 | Push-notification device picker labels resolve | ✅ PASS |
| 6 | P1 | Admin export emits one column per field | ✅ PASS |
| 7 | P1 | API contract: camelCase canonical, aliases mirrored, no push token | ✅ PASS |
| 8 | P2 | Devices list renders on a mobile viewport | ✅ PASS |

### What was observed

**1 — Devices list (`/backend/devices`).** Every renamed column resolves against the new response: `Device ID` `qa-5571-1787579277436`, `Platform` `ios`, `User` `employee@acme.com` (rendered as the resolved display label, not a raw uuid), `App version` `4.5.6`, `OS version` `iOS 18.2`, `Push provider` `apns`, `Last seen` `8/24/2026, 3:47:58 PM` formatted as a local timestamp rather than a raw ISO string. The push token appears nowhere in the row. The third row in the screenshot is an earlier fixture registered without a push token and correctly shows `—` for Push provider — that is the genuine empty state, not a rename miss.

**2 — Device edit form (`/backend/devices/<id>`).** App version, OS version and Push provider all load pre-filled with the stored values, and the header block shows device id, platform and the owning user.

**3 & 4 — Optimistic locking.** This was the highest-risk item in the PR: the form's lock header is derived from the detail response's `updatedAt`, renamed from `updated_at` here, so a missed rename would have silently disabled concurrent-edit protection with no visible error. Two tabs were opened on the same device; the first saved `5.0.0`, then the second saved `9.9.9` against its now-stale version. The `PUT` carried `x-om-ext-optimistic-lock-expected-updated-at` and the API answered **409**; the UI rendered the conflict bar *"Record changed — This record was modified by someone else. Refresh and try again."* with a Refresh action. Re-reading the record confirmed `clientAppVersion` is still `5.0.0` — the first writer's value survived and was not silently overwritten.

**5 — Push-notification device picker (`/backend/push_notifications/send`).** The picker's filter (`pushProvider`) and label (`deviceId · platform · pushProvider`) both resolve: `qa-5571-1787579277436 · ios · apns`, with no `undefined` segments.

**6 — Admin export.** `GET /api/devices/admin/devices?format=csv` returns exactly 14 columns — `Id, TenantId, OrganizationId, UserId, DeviceId, Platform, ClientAppVersion, OsVersion, Locale, PushProvider, PushTokenUpdatedAt, LastSeenAt, CreatedAt, UpdatedAt` — with no snake_case duplicates. This confirms in a real browser that the export regression found during review (the deprecated aliases doubling every column) is fixed by `a16d579c5`.

**7 — API contract.** The list response carries all eight camelCase keys, every snake_case alias equals its camelCase source, and neither `pushToken` nor `push_token` is present.

**8 — Mobile viewport.** At 390×844 the list still renders every value.

### Notes for the reviewer

- **Not a finding against this PR:** the Register-device form offers a *Push provider* field independently of *Push token*, but `commands/register.ts:85` persists `pushProvider` only when a `pushToken` is supplied, so filling the provider alone silently stores `null`. That is pre-existing behavior unrelated to this change — flagged only because it initially looked like an empty column.
- **Environment-only:** the QA instance first served `403 Forbidden` on the devices routes despite the `admin` role holding `devices.*`, because the SQLite ACL cache (`CACHE_STRATEGY=sqlite`) had a snapshot predating `auth sync-role-acls`. Clearing `.mercato/cache` fixed it. A local-environment artifact, not a product issue.

*Evidence-only run: no pipeline or QA verdict labels were changed. `needs-qa` stands and a QA reviewer owns the verdict. This PR also still needs an approving human code review — the automation authored it and cannot approve its own PR.*


### Screenshots

**qa-01-devices-list-populated**



**qa-02-device-edit-prefilled**



**qa-03-optimistic-lock-conflict**



**qa-04-push-send-page**



**qa-05-devices-list-mobile**

adeptofvoltron:
🤖 `om-auto-qa-pr` — completed: **PASS** (8/8 steps, screenshots attached). Evidence-only: no QA verdict labels were changed, `needs-qa` stands. Lock retained — chain continues.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the camelCase fix shipped with its deprecation bridge, code review found and fixed two medium issues, and UI QA passed 8/8 with screenshots attached. Lock released; swapped for `ci-monitoring` while the `ephemeral-integration` check finishes. The PR needs an approving human code review (the automation authored it and cannot self-approve) and a QA sign-off before it can merge.

adeptofvoltron:
🤖 `om-auto-review-pr` — CI result: **all checks green** on `a16d579c5` (18 checks: 14 pass, 4 skipped, 0 failing). `ephemeral-integration` passed, which means the new `TC-DEV-007` spec exercised the camelCase contract and the deprecated-alias bridge against a real instance in CI. Removing `ci-monitoring`; no further follow-up is owed from this agent.

wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-24T18:33:48Z. Other auto-skills will skip this PR until the lock is released.

Context: the previous `om-auto-fix-issue` chain (running as @adeptofvoltron) explicitly released its lock and removed `ci-monitoring` at 2026-08-24T13:50Z, so this PR is free to claim. This run is an **independent code review from a different account** — the earlier review could not be submitted as a GitHub review because the automation authored the PR and GitHub refuses self-approval. Running with `--autofix`.

wojciechszyjka (approved):
# 🔍 Code Review — `fix(devices): return camelCase API response keys (#5513)`

**Reviewer:** `om-auto-review-pr` running as @wojciechszyjka · **Mode:** first review by this account · **Head:** `a16d579c5`
**Runner:** no local gate — **validation source: GitHub checks** (per the repo-local `om-auto-review-pr` extension, which is checks-first and reserves local runs for when CI data is unusable; every command in `validation.commands` is covered by a green check on this exact head SHA, so the narrow local fallback did not apply).

This is a **second, independent review from a different account** than the one that authored the PR. An earlier `om-auto-review-pr` pass ran as @adeptofvoltron and could only post its findings as a comment, because GitHub refuses to let an author review their own pull request. That left the PR with an outstanding `REVIEW_REQUIRED`. This pass re-derived the review from the code rather than trusting that comment, and submits it as an actual review.

## 🎯 Summary

The `devices` module was the one module returning the query engine's raw database column names (`device_id`, `last_seen_at`, `client_app_version`, …) from its three GET routes, so any client written against the platform's camelCase convention read `undefined` for every field. This PR makes camelCase canonical on all three routes and keeps every snake_case key alongside it as a deprecated alias.

The mechanism is the conventional one and the contract handling is careful. Both list routes gain `list.transformItem: transformDeviceListItem`, which is exactly how `warranty_claims`, `eudr`, `wms` and the `sales` line routes solve the same problem; the factory applies `transformItem` before custom-field decoration, the translation overlay, access logging and the paged export (`packages/shared/src/lib/crud/factory.ts:1825-1866`), so no downstream stage sees the raw projection. The admin detail route's hand-rolled `serializeDevice` moves into a shared, unit-testable `serializeDeviceDetail`, so both surfaces now share one casing decision.

I found **no blockers and no majors**. Three findings, all **Low**, are listed below for the author to pick up; none of them blocks the merge, and I am deliberately **not** pushing fixes for them (see 🔧 Autofix).

## ⚖️ Verdict

**Approve.** Only Low/nit findings survived verification, which is `approve` under both the `om-code-review` verdict rule and this repo's stricter variant (which escalates Medium to changes-requested).

Two gates still stand between this approval and a merge, and neither is mine to clear:

- **QA.** `needs-qa` is on with no `skip-qa` and `qaGate` is enabled, so the merge is blocked until a QA reviewer adds `qa-approved`. An `om-auto-qa-pr` run has already attached evidence-only screenshots and a PASS on 8/8 steps, but evidence is not the label — a QA reviewer still owns the verdict.
- **Branch protection.** All 13 completed checks are green on `a16d579c5`; nothing is pending, so this approval and the CI state agree.

## 🧪 Validation Gate

Every command in `.ai/agentic.config.json` → `validation.commands` maps onto a check that ran green against `a16d579c5`. Verified per-check via the check-runs API on the head SHA, not inferred from the PR page.

| `validation.commands` step | CI check | Result |
|---|---|---|
| `yarn build:packages` | `prepare` (`.github/workflows/ci.yml:182`) | ✅ pass |
| `yarn generate` | `prepare` (`ci.yml:190`) | ✅ pass |
| `yarn build:packages` (rebuild over generated files) | `prepare` (`ci.yml:193`) | ✅ pass |
| `yarn i18n:check-sync` | `lint` (`ci.yml:632`, `scripts/i18n-check-sync.ts`) | ✅ pass |
| `yarn i18n:check-usage` | `lint` (`ci.yml:635`) | ✅ pass |
| `yarn typecheck` | `test` (`ci.yml:653`) | ✅ pass |
| `yarn test` | `test` (`ci.yml:665`) | ✅ pass |
| `yarn build:app` | `prepare` (`ci.yml:205`) | ✅ pass |
| — (beyond the configured gate) | `ephemeral-integration (none)`, `docker-build`, `documents-multi-instance`, `ds-lint`, `lint`, `scope`, `audit-scope`, `CodeQL`, `Analyze (actions/python/javascript-typescript)`, `license/cla` | ✅ pass (13 total; 5 skipped: `merge-coverage`, `audit`, `comment`, `mutate`, `[code]smith`) |

`ephemeral-integration` passing is the load-bearing one here: it means the new `TC-DEV-007` exercised the camelCase contract *and* the deprecated-alias bridge against a real running instance, not just a mocked transform.

**Template parity:** not applicable. The repo extension's `yarn template:sync` gate triggers on edits under `apps/mercato/src/{app,modules}`, `apps/mercato/src/i18n` or `apps/mercato/.env.example`; this diff touches none of them.

**Frontend performance gate:** not applicable. The three `.tsx` files change property reads and column `accessorKey` strings only — no `"use client"` boundary moved, no server component became a client component, no route entry point gained a heavy eager import, and no provider grew a route-specific SDK. `ds-lint` is green and no `className` changed, so there is no design-system surface to check.

## 💥 Findings — Low

### 🔵 Low — `deviceListSortFieldMap` speaks camelCase for 3 of the 11 renamed keys, and an unmapped sort silently no-ops · `packages/core/src/modules/devices/api/deviceList.ts:52`

The response contract is now camelCase for eleven keys, but the sort contract only translates three: `lastSeenAt`, `createdAt`, `updatedAt`. The factory resolves a sort field through `sortFieldMap` and falls back to the raw name (`packages/shared/src/lib/crud/factory.ts:1727-1736`), and the query engine then drops any sort column it cannot resolve rather than erroring — `resolveBaseColumn` does no case conversion, so an unknown field yields `null` and the sort is discarded silently (`packages/shared/src/lib/query/engine.ts:628-634`, `:1540-1544`).

The practical consequence is that a client that reads `deviceId` from the response and naturally passes `?sortField=deviceId` gets **unsorted results with a 200 and no diagnostic**. `?sortField=device_id` still works, so nothing regressed, and `UPGRADE_NOTES.md` is explicit that the sort field names are unchanged.

I am keeping this at Low rather than Medium for one concrete reason: the reference module has the identical asymmetry. `warranty_claims` returns `claimNumber`/`awaitingStaffReply` in camelCase while its `sortFieldMap` maps only `slaDueAt`, `createdAt`, `updatedAt` (`packages/core/src/modules/warranty_claims/api/route.ts:377-381`). This is the repo's established convention, not something this PR introduced, so fixing it here would be a local patch on a framework-wide pattern.

**Suggested fix:** extend `deviceListSortFieldMap` with the remaining camelCase spellings (`userId`, `deviceId`, `clientAppVersion`, `osVersion`, `pushProvider`, `pushTokenUpdatedAt`, `tenantId`, `organizationId`) — eight map entries, no behavior risk, and it makes the request and response halves of the contract agree. The framework-level fix (a camelCase→snake_case fallback in `resolveBaseColumn`, or a hard 400 on an unresolvable sort field instead of a silent drop) belongs in a separate issue.

### 🔵 Low — the `never exposes the push token` unit test cannot fail · `packages/core/src/modules/devices/__tests__/response-casing.test.ts:78`

The test feeds `push_token: 'secret-token'` into `transformDeviceListItem` and asserts only `expect(item.pushToken).toBeUndefined()`. `pushToken` is a key the transform never produces under any input, so the assertion is vacuous. Meanwhile the key it *did* inject survives: the transform's first operation is `{ ...record, ... }` (`api/deviceList.ts:82`), and `toDeprecatedSnakeCaseAliases` only sets its eleven named keys, so `push_token: 'secret-token'` passes straight through to the output the test is inspecting. The test would report success on a function that leaks the secret verbatim.

This is a **test-quality** finding, not a security one, and I verified both halves of that before rating it. Production is safe because the query engine only projects `deviceListFields`, which excludes `push_token`, and the real guarantee is genuinely pinned end-to-end: `TC-DEV-007` asserts `pushToken` **and** `push_token` are absent from the self list, the admin list and the detail route against a live instance (`__integration__/TC-DEV-007.spec.ts:100-101`, `:122-123`). The spec's own Test Plan is also worded accurately — "`push_token` never becomes `pushToken`" — so it is the test's *name* that overpromises relative to its assertion.

**Suggested fix**, in preference order:

1. Strengthen it: add `expect(item.push_token).toBeUndefined()`. That will fail as written, which is the point — make it pass by having `transformDeviceListItem` explicitly delete `push_token`/`pushToken` from the spread. Two lines, and worth it on its own merits: because the transform is a wholesale spread rather than the allowlist the old detail serializer used, the module's "treat `push_token` as a secret" rule (`devices/AGENTS.md:24-25`) is currently enforced *only* by the upstream projection list. An explicit strip makes the invariant local to the function that claims it, and is cheap defense in depth for a secret.
2. Or, if the pass-through is intentional, rename the test to what it actually checks (`never derives a camelCase pushToken`) so nobody later reads it as proof of secret containment.

### 🔵 Low — pinning `export.columns` drops tenant-defined custom-field columns from the export · `packages/core/src/modules/devices/api/admin/devices/route.ts:70`

Pinning the export to `deviceExportColumnFields` is the right call and correctly fixes the doubled-column problem: without it, `prepareExportData` falls through to `buildDefaultExport`, which spreads each item and would therefore emit both spellings of every aliased key (`packages/shared/src/lib/crud/factory.ts:376-402`). It also matches `warranty_claims`, which pins its columns the same way (`warranty_claims/api/route.ts:374-376`).

The side effect is that `buildExportFromColumns` emits **only** the declared columns, whereas `buildDefaultExport` previously spread everything — including the `cf_*` keys that `decorateItemsWithCustomFields` adds. So a tenant-defined custom field on `user_device` used to appear in the CSV and now will not. This is very likely harmless: `devices` ships no `ce.ts` and declares no custom fields, the module debuted in the current release window so nobody can have depended on it, and the reference module behaves identically.

It is worth one line of prose because the spec argues the opposite direction in the same breath — the Overview justifies spreading the raw record precisely so "custom-field keys, columns added later" survive the transform, and then the export silently discards them. **Suggested fix:** a sentence in the spec's Architecture table (or the `UPGRADE_NOTES.md` entry) recording that the pinned export set is deliberately closed, so the next person to add a column knows the export needs the same edit — which the new unit assertion at `response-casing.test.ts:86-97` already enforces mechanically for canonical keys.

## ✅ What holds up

- **The deprecation protocol is followed in full, not gestured at.** `BACKWARD_COMPATIBILITY.md` § 7 forbids removing response fields and requires a working surface for at least one minor version. Every snake_case key is still returned with the same value, `toDeprecatedSnakeCaseAliases` carries an `@deprecated` JSDoc naming the replacements, `UPGRADE_NOTES.md` has a `0.7.0 → 0.7.1` entry with the complete mapping and the client action, and a spec with a *Migration & Backward Compatibility* section is referenced. I checked the diff for the pattern that would have made this a blocker — a removed response field — and there is none: the schemas are strictly additive.
- **The bridge is removable in one place, and its removal is guarded.** Both serializers spread a single shared alias helper, and `TC-DEV-007` asserts the aliases today, so dropping them cannot land unnoticed. `TC-DEV-001/005/006` are correctly left asserting snake_case — for the bridge's lifetime they are the proof that no 0.7.0 consumer broke.
- **Scoping and authorization are untouched.** The self list keeps its `mergeAdvancedFilterTree` hardening and actor scope; the admin list keeps factory tenant/org scoping; the detail, PUT and DELETE handlers keep `isOrganizationReadAccessAllowed` on the loaded record. This is a serialization change only, which is what a response-casing fix should be.
- **The detail serializer does not invent keys.** `toDeprecatedSnakeCaseAliases` deletes `undefined` entries, so the detail response — which never carried `tenant_id`/`organization_id` — does not grow them. That is asserted directly (`response-casing.test.ts:140-141`), and it matters: an over-eager alias helper would have added two cross-tenant identifiers to a response that previously withheld them.
- **The optimistic-lock source was migrated, not missed.** The edit page's `optimisticLockUpdatedAt` now reads `device.updatedAt` (`backend/devices/[id]/page.tsx:145`). Getting this wrong is the highest-consequence failure mode in the whole diff — the form would have sent no version header and concurrent edits would be lost silently, with no error anywhere. The QA evidence run confirmed a real 409 and the conflict bar against a live instance.
- **First-party consumers are complete.** I grepped the workspace for every remaining reader of the devices API and of the snake_case device keys: the only consumers are the two backend pages and the push-notifications picker, all three migrated, plus `packages/core/src/helpers/integration/pushFake.ts`, which reads only `id` from the POST response and is unaffected.
- **No post-transform stage keys on field names.** Custom-field decoration, the translation overlay and `logCrudAccess` all key on `id`/`entityId`, so moving the item shape to camelCase cannot desynchronize them (`factory.ts:1827-1861`).
- **No generated-artifact drift.** `openapi.generated.json` is not a tracked file in this repo, so the OpenAPI description changes need no committed regeneration.

## 🧪 Test Coverage

Proportionate and, more importantly, aimed at the failure modes that actually matter here.

- **Unit** (`__tests__/response-casing.test.ts`, 10 cases): the camelCase key set, alias mirroring key-by-key, `Date`→ISO normalization, camelCase-input tolerance, `cf_*` survival through the transform, both zod schemas parsing their own serializer's output, and — the assertion I liked most — a check that the set of non-alias keys on a transformed row equals `deviceExportColumnFields` exactly, which fails if anyone adds a projection column without an export entry. That is a test that will earn its keep later. The one weak case is covered as a Low finding above.
- **Integration** (`__integration__/TC-DEV-007.spec.ts`, 2 cases): registers a real device and pins the camelCase contract on the self list, the admin list and the detail route, then pins every alias against its canonical source on both the list and the detail route, with push-token absence checked under both spellings. It polls the eventually-consistent query index rather than sleeping, and cleans up its fixture in a `finally`.
- **Regression proof for the bug itself:** the fix's premise is "these routes return the wrong key names", and `TC-DEV-007` fails against the pre-fix code on its first assertion. That is a genuine regression test, not a coverage gesture.

- [x] **No response field removed.** Strictly additive on all three routes; every previously returned key is still present with the same value.
- [x] **No route URL or HTTP method changed.**
- [x] **Deprecation path documented and bridged** — `@deprecated` JSDoc, alias retention for ≥1 minor, `UPGRADE_NOTES.md` mapping, referenced spec with a *Migration & Backward Compatibility* section. `BACKWARD_COMPATIBILITY.md` § 7 satisfied.
- [x] **No database schema change**, so no migration is expected and none is present. Columns keep their snake_case names.
- [x] **No event ID, widget spot ID, DI key, ACL feature, CLI command or public import path touched.**
- [x] **Request-side contracts unchanged** — query params, request bodies, and the `POST`/`PUT`/`DELETE` response shapes.
- [x] **No security surface weakened** — auth guards, tenant/org scoping and input validation are untouched; `push_token` remains absent from every response under both spellings, verified in CI.
- ⚠️ **One behavior difference worth restating,** already disclosed in `UPGRADE_NOTES.md`: timestamps now always serialize as ISO-8601 under *both* spellings, where previously they were whatever the query engine produced. A snake_case client parsing timestamps loosely is fine; one doing exact string comparison against a non-ISO format is not. Correctly documented rather than hidden.

## 🧭 Non-actionable notes and follow-ups

Neither of these is a finding against this PR. Both are worth an issue.

- **`exportScope=full` bypasses `transformItem` *and* `export.columns`.** On that branch the factory serializes `rawItems` through `normalizeFullRecordForExport` and builds columns with `ensureColumns` instead of calling `prepareExportData` at all (`factory.ts:1866`, `:1902-1903`). So a full-scope export emits snake_case headers while the paged export — the one a UI export button uses — emits camelCase. This is framework-wide behavior affecting every module that uses `transformItem`, the spec discloses it explicitly, and it is correctly left alone here. A framework-level follow-up should decide whether the full-export branch ought to honor the list's declared export shape.
- **The same inconsistency exists one module over.** `packages/core/src/modules/push_notifications/api/deliveries/[id]/route.ts:32` returns `user_device_id` in its response. Out of scope for #5513, which is scoped to the three devices GET routes, but it is the next instance of the exact bug this PR fixes.

## 🔧 Autofix

This run is autofix-eligible (`--autofix` was passed), and I am **deliberately not pushing any commit.**

The autofix loop exists to clear a `changes_requested` verdict, and this review approves. All three findings are Low — the `om-code-review` rule is to list them for the author, not to block on them. Pushing them would move the head SHA, which would discard the 13 green checks on `a16d579c5`, invalidate this approval, and stale the QA evidence already attached to this PR. That is a strictly worse outcome than handing three nits to the author, so the trade is not close. The author is free to take any of them in a follow-up commit or a separate PR; the first two are a few lines each.

wojciechszyjka:

- 🚀 `merge-queue` — the code review passed with only Low findings and every required check is green on `a16d579c5`, so the review gate is cleared and this PR replaces `review` with `merge-queue`; the merge itself is still held by the QA-approval gate below.
- 🧪 `needs-qa` — kept as it was. The change touches three `.tsx` surfaces (the devices admin list and edit pages, and the push-notifications device picker), which disqualifies it from the automated-verification exemption, and `qaGate` is on — so this PR cannot merge until a QA reviewer adds `qa-approved`. An `om-auto-qa-pr` run has already attached evidence-only screenshots and a PASS on 8/8 steps, but evidence is not a verdict and this skill never applies `qa-approved` from reading a diff.
- 🐛 `bug` — kept as it was. Issue #5513 reports broken behavior: the devices API returned raw database column names, so a client following the platform's camelCase convention read `undefined` for every field.
- 🔺 `priority-high` — kept as it was, and the review supports it. The fix's value is time-boxed to the release window in which `devices` debuted: landing the camelCase shape together with its deprecation bridge in 0.7.1 is what makes removing the snake_case aliases in the following minor possible, and if it slips the wrong shape exists across two released minors and the bridge becomes effectively permanent.
- ⚠️ `risk-high` — kept as it was. This modifies an API response contract surface (`BACKWARD_COMPATIBILITY.md` § 7) on three routes at once, and a mistake is visible to every external device client rather than contained to one screen. The review found the change to be strictly additive with the deprecation protocol followed in full, which argues for `risk-medium`; I am leaving the higher label because the blast radius, not the observed quality of the change, is what this label measures.
- 📸 `screenshots` — kept as it was; UI QA evidence is attached to this PR.
- 🤖 `in-progress` — this review's own claim lock, released in the completion comment below.

No priority or risk label was changed by this run: exactly one of each was already present and neither is mis-rated.

wojciechszyjka:
## 🧪 Manual QA — status and what this review adds

This PR carries `needs-qa` with no `skip-qa` and `qaGate` is on, so it cannot merge until a QA reviewer adds `qa-approved`. Approving the code does not clear that gate, and this skill never applies `qa-approved` from reading a diff.

**I am not re-issuing the P0/P1/P2 test plan**, because one is already on this PR and it has already been executed. The instructions are in [this comment](https://github.com/open-mercato/open-mercato/pull/5571#issuecomment-5395952349) and the evidence run — PASS on 8/8 steps with screenshots, including a real 409 on the optimistic-lock check — is in [this one](https://github.com/open-mercato/open-mercato/pull/5571#issuecomment-5396154566). Both are accurate against the current head `a16d579c5`; I re-derived their claims from the code while reviewing and found nothing that would change them. Duplicating them here would add noise, not coverage.

**QA reviewer:** when you pick this up, swap `merge-queue` for `qa`, work the linked plan, then remove `qa` and add `qa-approved` (or `qa-failed` with a description of the failure).

Two things worth carrying into the QA pass, from this review specifically:

- **The optimistic-lock check (P0, step 3 in the linked plan) is still the single highest-value item.** The edit form derives its lock header from the detail response's `updatedAt`, renamed from `updated_at` in this PR. If that rename were wrong the form would send no version header and concurrent edits would be lost silently, with no error surfaced anywhere. The evidence run confirms it works; it is worth confirming again on whatever instance you test, because it is the one failure mode in this diff that is invisible rather than obvious.
- **Nothing in my three Low findings changes what QA should test.** They cover a sort-parameter naming asymmetry that predates this PR, a unit test whose assertion is weaker than its name, and a documentation gap about export columns — none of them are reachable from the UI, and none of them affect the screens in the linked plan.

wojciechszyjka:

An approving review is now on the record from @wojciechszyjka — a different account than the PR author — which clears the `REVIEW_REQUIRED` state that the earlier in-chain review could not, because GitHub refuses to let an author review their own pull request. `reviewDecision` is now `APPROVED`.

**What the review found.** No blockers and no majors. Three Low findings, listed in the review for the author to pick up: `deviceListSortFieldMap` translates only 3 of the 11 renamed keys so an unmapped camelCase `sortField` silently no-ops (the reference module `warranty_claims` has the identical asymmetry, which is why it is Low and not Medium); the `never exposes the push token` unit assertion is vacuous, though `TC-DEV-007` pins the real guarantee end-to-end under both spellings; and the pinned export column set closes the export to tenant-defined custom fields without saying so in the spec.

**Autofix: eligible but deliberately not run.** `--autofix` was passed, but the loop exists to clear a `changes_requested` verdict and this review approves. Pushing three nits would move the head SHA, discard the 13 green checks on `a16d579c5`, invalidate this approval and stale the attached QA evidence — strictly worse than handing them to the author.

**Validation source: GitHub checks.** Every command in `validation.commands` maps onto a check that ran green on this exact head SHA; the per-command mapping is the table in the review body. No local gate run and no `ci-monitoring` label: nothing was pending at verdict time, so there is no CI follow-up owed and none is promised.

**Remaining gate:** `needs-qa` without `qa-approved`. A QA reviewer owns that verdict.

Issue #4056, opened 2026-07-09
bug: Order line unit price can be edited after the line has already shipped


droalina:
## Summary

When editing an order line on a confirmed sales order, the unit price (net/gross) can be changed
freely even if some or all of that line's quantity has already been shipped to the customer. There is
no warning or block — the price simply updates and order totals recompute against the new price,
even though the customer already received goods (and possibly an invoice) at the old price.

## Page URL

`/backend/sales/orders/{orderId}` — order detail page, order line edit control.

## Steps to reproduce

1. Create a sales order with at least one item.
2. Create a shipment for that line (partial or full quantity) so `shippedQuantity > 0`.
3. Edit the same order line and change its unit price (net or gross).
4. Save.

## Expected result

The system should either block the price edit on a line that has already shipped, or require an
explicit reason/adjustment path so the change is auditable — not silently overwrite the historical
price.

## Actual result

The price change is accepted immediately, with no error and no warning dialog. On an order line at
148,00 USD with 4 units already shipped, changing the unit price to 5,00 USD and saving drops the line
total from 592,00 USD to 10,00 USD brutto, with the 4 already-shipped/delivered units still on record
as shipped at the old price.

## Environment

- Branch: `develop`
- Module or area: `sales` — order line editing
- Browser/Runtime: Microsoft Edge

## Additional context

Raised by @pkarw as a follow-up comment on issue #3993 ("I think we should not be able to edit the
prices of shipped items"). Related to (but distinct from) #3993, which covers quantity edits below
shipped levels — this one is about price specifically. Consider whether it should be folded into the
same fix/PR as #3993 or tracked separately.

andrzejewsky:
Claimed for the shipped order-line price safeguard. Preparing a focused fix and regression test.

wojciechszyjka:
🤖 `om-auto-fix-issue` triage — **no action needed: this is already fixed on `develop`.**

I picked this up because the issue is still open and has been assigned since 2026-07-11 with no PR attached to the assignee. Before claiming it I ran the read-only triage gate, and the whole scope described here — plus the follow-up UI work — is already merged. I am therefore **not** taking the issue over and not opening a duplicate PR. The assignment is left untouched so the history stays readable.

### 🔍 Evidence

**1. Server-side enforcement — #4131, merged into `develop` on 2026-08-04 as `8d8226ace`.**
[...]

Issue #3993, opened 2026-07-08
bug: editing order item - it should not be possible to lower the qty below the shipped levels


pkarw:
Now we can edit it without any validation - so we can have situation like Shipped: `4 of 2`

<img width="1057" height="134" alt="Image" src="https://github.com/user-attachments/assets/40a105d8-ce44-46bc-ae53-4aa4b5580c59" />

pkarw:
Also - I think we should not be able to edit the prices of shipped items

adeptofvoltron:
🤖 `autofix` started by @adeptofvoltron at 2026-07-13T13:37:02Z. Other auto-skills will skip this issue until the lock is released.

adeptofvoltron:
🤖 Lock released — the work was completed and shipped as draft PR #4163 (the autofix worker finished the change but exited while waiting on its validation gate; the gate was re-run by hand before pushing).

pkarw:
🤖 `om-sync-merged-pr-issues` started by @pkarw at 2026-08-03T09:11:36Z. Other auto-skills will skip this issue until the lock is released.

pkarw:
✅ Fixed by #4197 (https://github.com/open-mercato/open-mercato/pull/4197) — merged at 2026-08-03T07:16:58Z (commit `505cf94`).

Closed automatically by the `om-sync-merged-pr-issues` skill. Credit to @adeptofvoltron (or the original author when the PR is a carry-forward — see the PR body for credit details).

If this is incorrect, reopen the issue and add the `do-not-close` label so future runs leave it alone.

PR #4131, opened 2026-07-11
fix(sales): lock prices on shipped order lines

andrzejewsky:
## Summary
- Reject net or gross unit-price changes after any quantity of an order line has shipped.
- Preserve the existing stored price and totals when the change is rejected.
- Add API integration coverage for a partially shipped line.

Closes #4056

## Validation
- Passed: build packages, generate, i18n checks, typecheck, lint, focused command tests.
- Full test run hit an unrelated shared Jest-worker SIGSEGV; the exact suite passed when rerun serially.
- App build is blocked by pre-existing unresolved generated entity imports from packages/core/dist across untouched files.
- The new Playwright scenario needs a managed ephemeral environment, unavailable in this worktree.

PR #5572, opened 2026-08-24
feat(sales): make the shipped-line freeze a per-scope setting, on by default

kamwro:
## Summary

Editing an order line that already has shipment items — its unit price, discount, tax rate, unit, derived totals, or a quantity below what shipped — is refused with a `409`, and `restoreOrderGraph` refuses the same. Handing a parcel to a courier therefore makes an order's commercial terms permanent, while a fully **paid** order can still be edited freely.

Corrections after dispatch are ordinary retail: a wrong price or VAT rate fixed by accounting days later, a discount granted after a complaint, a quantity lowered by a return. Dispatch is a logistics event; what makes a commercial document untouchable is a fiscal document issued against it or a closed period, not a box leaving a warehouse. For a programmatic caller mirroring a system of record the refusal is silent and total: one refused line fails the whole document write, on that run and every one after.

**This PR makes that refusal a choice instead of a law, and leaves the choice where it is today.** The freeze moves behind a new per-scope sales setting, `orderShippedLineEditable`, defaulting to off — so every existing deployment keeps the exact behaviour it upgraded from, and a deployment that mirrors an external system of record turns it on and records what that system does after dispatch.

An earlier revision of this branch removed the freeze outright. That changed the contract on `sales.orders.lines.upsert` for every deployment without anyone asking, which is the kind of change nobody notices until an order's commercial terms have already been rewritten. The current shape is @adeptofvoltron's suggestion, taken as offered.

### Why a setting rather than an environment variable

The sales module already models this class of policy: `sales_settings` carries `orderCustomerEditableStatuses` and `orderAddressEditableStatuses`, enforced in `applyDocumentUpdate` and `assertAddressEditable`, edited through `components/OrderEditingSettings.tsx` / `api/settings/order-editing`, and read back by the document detail page so the UI disables what the API would refuse. The freeze joins them rather than introducing a parallel mechanism.

That last property is load-bearing here. `LineItemDialog` carries its own copy of the freeze — it disables the pricing controls, floors the quantity input and strips those fields from the payload before sending. Behind an environment variable the server would accept a correction the admin could never send, and the switch would appear to do nothing. Both halves read the same setting.

### What does not move with the setting

**This does not undo #3993.** That issue reported an admin lowering a quantity below what shipped (`Shipped: 4 of 2`) and asked for it to be blocked. Three routes lead to that state:

| route | guarded by | after this PR |
|---|---|---|
| ship more than was ordered | `sales.shipments.create` rejects it — a guard that predates #3993 | unchanged, whatever the setting says |
| a person lowers the quantity in the admin | `LineItemDialog` floors the quantity and strips pricing fields | unchanged by default; follows the setting when a scope opts in |
| a programmatic caller corrects the order after dispatch | the command-layer freeze | unchanged by default; allowed when a scope opts in |

**Deleting a line with shipment items stays refused either way.** That guard fronts the real `sales_shipment_items.order_line_id` foreign key, so no setting can make the delete succeed — only swap an explained `409` for a raw database error.

A scope that opts in does re-admit the `Shipped: 4 of 2` display when its source system lowers a quantity below what shipped. That is deliberate: for a mirrored order it is a true statement about what happened, and refusing the write never removed the discrepancy, only the record of it.

## Changes

- **`sales_settings.order_shipped_line_editable`** — `not null default false`, added by `Migration20260826120000_sales_settings_shipped_line_editable`. No backfill; a scope with no settings row resolves to the same default.
- **`isShippedLineFreezeEnforced`** (`commands/settings.ts`) — the single place the freeze asks whether it applies. The command layer and the settings API both read the answer from here.
- **`commands/documents.ts`** — the freeze restored byte for byte, with that question in front of both entry points. `git diff` against the pre-removal state on this file is the import, a header comment and two `if`s. Both entry points ask before reading any shipment state, so a scope that opted out never pays for the shipment-item join.
- **Settings surface** — the field flows through `data/validators.ts`, `sales.settings.save` and `api/settings/order-editing` (GET, PUT and the OpenAPI response schema).
- **Admin surface** — a toggle on the order-editing settings screen beside the two status lists; `orderShippedLineEditable` travels from the settings API through the document detail page and `SalesDocumentItemsSection` to `LineItemDialog`, defaulting to `false` at every hop so an unaware caller keeps today's lock. The unresolved-shipments fallback follows the same setting: it exists to avoid offering an edit the server would refuse, and where there is no refusal there is nothing to be cautious about.
- **i18n** — the `errorPriceShipped` key is restored across the five locales, and three new keys describe the toggle.
- **`UPGRADE_NOTES.md`** — what the column is, why the upgrade is a no-op, what turning it on grants and what it does not, and that a deployment wanting the freeze as policy should set it explicitly because the default is expected to flip.

## Specification

- [ ] Yes
- [ ] No (created a new spec)
- [x] N/A — additive setting with a behaviour-preserving default; no spec governs the command-layer guard, and the surviving protections are unchanged.

**Spec file path:** —

## Testing

- `yarn typecheck` — clean across the monorepo (26/26 tasks).
- `@open-mercato/core` full jest run — **1,426 suites / 11,469 tests passed**.
- `yarn i18n:check` — locales in sync, no missing keys.
- `commands/__tests__/documents.line-shipped-guard.test.ts` (29) pins **both** branches in one harness: the corrections an opted-in scope can make, and the same corrections refused for a scope that did not opt in — including a scope with no `sales_settings` row at all, since absence and an explicit `false` arrive by different routes and have to agree.
- `components/documents/__tests__/LineItemDialog.shippedLineLock.test.tsx` (18, five new) pins the dialog on both sides of the prop.
- **Checked by mutation, not by reading.** Forcing the resolver to `false` reddens the ten enforcement cases; forcing it to `true` reddens the ten opt-in ones; the two sets are disjoint. Blocking the dialog prop reddens exactly the five new dialog cases.
- `TC-SALES-4056` restored to its original assertions and extended: one case flips the setting, applies each correction and reads it back — a `200` alone would not distinguish an applied write from an accepted-and-dropped one — and restores the previous value in `finally`. The tier runs `workers: 1`, so that window cannot overlap another spec. The delete-refusal case now runs with the setting **on**, which is the only place that boundary can be shown to be independent of the flag.

## Checklist

- [x] I added or updated integration tests — `TC-SALES-4056` covers both branches of the setting.
- [ ] I created or updated the spec in `.ai/specs/` — N/A per above.
- [ ] Priority set — this account has no `triage` on this repo; intended labels are in a comment below for a maintainer to apply.
- [ ] Risk set — same; proposing `risk-low` now that the change is additive and the default preserves current behaviour.
- [ ] QA routing set — same; proposing `needs-qa`, self-QA evidence in the Testing section above.

### Design System Compliance
- [x] No hardcoded status colors — the new control is a `Switch` in the existing settings card
- [x] No arbitrary text sizes — reuses the surrounding `text-sm` / `text-xs` pattern
- [x] Empty state — n/a, no list/data page touched
- [x] Loading state — the toggle is disabled while the settings load or save, like its neighbours
- [x] `aria-label` — set on the `Switch` from the translated label
- [x] Uses existing DS components — `Switch`, `Button`, `Badge`

## Linked issues

Related: #3993 — the guard grew out of it; its reported scenario stays blocked by default and by `sales.shipments.create` in every configuration.
Related: #4056 — its regression spec is restored here rather than inverted; the default behaviour it asserts is unchanged.
Deliberately not `Fixes` for either: both are closed and this PR keeps their fixes where they work.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5572"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790170900&installation_model_id=432243&pr_number=5572&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5572&signature=930f87d1a434cd31526fea7af22ad8e85dc0ae3caa35c2c9eeffc73ec95eb8e4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


kamwro:
Metadata intent (this account has no triage here) — for a maintainer to apply:

- `priority-medium` — unblocks integrations that mirror an external system of record; no data loss today, but silent sync failure on every post-dispatch correction.
- `risk-medium` — behavior change in a write command (`sales.orders.lines.upsert` accepts edits it previously 409'd); the user-facing protection is unchanged and pinned by the untouched dialog suite.
- `needs-qa` — it changes what an API write accepts; self-QA evidence is in the PR's Testing section (typecheck, 175 sales tests incl. the rewritten guard file, and a downstream integration suite exercising post-dispatch corrections end to end).

Suggested reviewer: whoever owns the sales command layer / reviewed the #3993 fix.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-25T00:18:10Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
## Code Review: fix(sales): let a system of record correct an order after it shipped

**Verdict: CHANGES REQUESTED** — 1 blocker, 1 major, 2 minor, 1 nit.

**Validation source:** GitHub checks. The two workflow runs for `55b8310` were sitting at `action_required` (fork PR awaiting approval), so this PR had **no CI signal at all** — `gh pr checks` showed only a green CLA and a skipped bot, which reads as clean at a glance. I approved both runs after confirming the diff touches only `.ts`/`.json` (no workflows, scripts, or manifests). Results below.

| check | result |
|---|---|
| `prepare`, `lint`, `scope`, `audit-scope`, `ds-lint`, `test`, `documents-multi-instance` | ✅ |
| `ephemeral-integration` | ❌ **1 failed**, 251 passed, 0 flaky (failed on retry too) |

**Runner:** GitHub Actions (no local fallback needed once CI was unblocked).

The case the PR makes — that dispatch is a logistics event and shouldn't be the moment commercial terms freeze — is a reasonable position, argued carefully. The problem is what it leaves out.

---

### 🔴 Blocker 1 — This undoes the fix for open issue #4056, which the PR never mentions

The PR body is framed entirely around #3993 and states, correctly, that it doesn't undo it. But the code being removed is the union of **two** separate merged fixes:

| commit | issue | what it added |
|---|---|---|
| `5197d505b` | #3993 | the quantity-below-shipped floor |
| `8d8226ace` (PR #4131, *"Closes #4056"*, merged 2026-08-04) | **#4056** | the price/unit/tax/total lock **and** `TC-SALES-4056-order-line-price-after-shipment.spec.ts` |

**Issue #4056 is still OPEN.** It reports exactly the behavior this PR restores:

> On an order line at 148,00 USD with 4 units already shipped, changing the unit price to 5,00 USD and saving drops the line total from 592,00 USD to 10,00 USD brutto, with the 4 already-shipped/delivered units still on record as shipped at the old price.

and asks for:

> The system should either block the price edit on a line that has already shipped, **or require an explicit reason/adjustment path so the change is auditable** — not silently overwrite the historical price.

That second clause matters: #4056 did not demand a hard freeze. It asked for a block *or an auditable adjustment path*. This PR takes neither branch — it restores the silent overwrite the issue was filed about. If the freeze is the wrong shape (and there's a real argument that it is), the resolution #4056 actually points at is the adjustment path, not removal.

CI proves the omission concretely — the integration test from #4131 is still in the tree and now fails:

```
TC-SALES-4056: shipped order-line price lock › rejects pricing and total updates
  after a partial shipment and preserves stored money fields
  Error: unit prices update should be rejected
  Expected: 409
  Received: 200
  at packages/core/src/modules/sales/__integration__/TC-SALES-4056-order-line-price-after-shipment.spec.ts:78
```

Not flaky: it failed on retry #1 as well, and the run reports `flaky: 0`.

Please either (a) restore the price lock and narrow this PR to the parts #4056 doesn't cover, or (b) make the #4056 case explicitly — retitle/reframe the PR as reopening it, say why the reporter's requirement no longer holds, and get that agreed before the guard comes out. Deleting or inverting `TC-SALES-4056-*.spec.ts` as part of (b) needs to be an explicit, called-out decision, not a test cleanup.

---

### 🟠 Major 1 — The totals check is removed on a premise that is false for gross

The rewritten test documents the removal of `hasShippedLineTotalChange` like this:

> *core recomputes line totals from quantity × unit price − discount and ignores the submitted totals, so an inflated value cannot move money — on any line, shipped or not.*

That is true for **net** and false for **gross**. In `packages/core/src/modules/sales/lib/calculations.ts`, `buildBaseLineResult`:

```ts
const netSubtotal = Math.max(netSubtotalBeforeDiscount - discountTotal, 0)   // recomputed ✅
const grossSubtotal =
  line.totalGrossAmount !== null && line.totalGrossAmount !== undefined
    ? toNumber(line.totalGrossAmount, 0)                                     // client value, verbatim ❌
    : round(netSubtotal + taxAmount)
```

`grossAmount` then accumulates into `subtotalGross` → `grandTotalGrossAmount` → `outstandingAmount`, and is persisted by `applyOrderTotals`. The validator only enforces `decimal({ min: 0 })` (`data/validators.ts:347`). So a caller can set an arbitrary `totalGrossAmount` on a shipped line and change what the customer owes while unit price, quantity and tax rate all read unchanged.

The deleted code said as much in its own comment — *"They are still writable straight through, so they cannot simply be ignored either."* — and `TC-SALES-4056` asserts precisely this (`['gross total', { totalGrossAmount: 6 }]`, then *"totalGrossAmount should remain unchanged"*).

To be fair: this pass-through is **pre-existing** for lines that never shipped, so the PR doesn't create the hole. But the shipped-line check was the only thing constraining it, and the stated reason for dropping it doesn't hold. Please correct the comment regardless of which way the blocker is resolved — as written it will mislead the next person to read it.

---

### 🟡 Minor 1 — The quantity floor becomes client-side only

The route table lists "a person lowers the quantity in the admin" as guarded by `LineItemDialog`. After this PR that dialog (`LineItemDialog.tsx:1359`) is the *only* remaining check — `errorQuantityBelowShipped` now has no server-side emitter. The command is reachable without it from:

- **`PUT /api/sales/order-lines`** — a documented, OpenAPI-described route (`makeSalesLineRoute` → `sales.orders.lines.upsert`), gated only by `sales.orders.manage`
- **`inbox-actions.ts:206`** — in-repo AI inbox order updates, which post a bare `quantity` with no shipped-quantity check
- **undo** — `restoreOrderGraph` lost `assertShippedOrderGraphRestorable`, so replaying an older snapshot can drop quantity below shipped

Worth noting `assertOrderAcceptsNewLine` does not backstop this: it early-returns on `existingSnapshot`, so it only blocks *adding* lines to a fulfilled order, never *editing* one.

The PR argues "Shipped 4 of 2" is a true statement after a return, which is fair. But lowering the order quantity isn't how a return is recorded — `sales.returns.create` is — and doing it this way has a silent consequence: `computeAvailableReturnQuantity` caps at `min(quantity, shippedQuantity)`, so the shipped units above the lowered quantity become permanently unreturnable in the system.

### 🟡 Minor 2 — `expectAllowed` can't prove the PR's own thesis

```ts
const expectAllowed = (caught: unknown) => {
  expect(isCrudHttpError(caught) && (caught as CrudHttpError).status === 409).toBe(false)
}
```

This asserts only *"not a 409"*. The old tests asserted the status **and** `expect(em.flush).not.toHaveBeenCalled()`; the rewrite drops the persistence assertion entirely. A 400, a 500, or a silently-skipped write all pass. Since the whole point of the PR is that the correction now goes through, these should assert `em.flush` **was** called and ideally that the corrected values landed. The comment "any other error here would be a harness artifact" is the assumption that needs pinning, not waiving.

Related: there's no integration coverage for the new behavior on `PUT /api/sales/order-lines`, which `AGENTS.md` and `.ai/qa/AGENTS.md` ask for on a behavior change to a public API surface — and the natural place for it is the file that's currently failing.

### 🔵 Nit — a second dead `errorPriceShipped`

The body says "the one dead i18n key". There's a twin at line 643 of all five locales, `sales.documents.detail.items.errorPriceShipped`. I checked: the whole `documents.detail.items` namespace is already unreferenced on `develop`, so this is **pre-existing and not yours** — just an easy extra sweep if you're in these files anyway.

---

### Verified clean

- No dangling references to any removed symbol; `loadShippedQuantityByLine` is still used by `returns.ts`, so no dead export
- i18n key removal is consistent across all five locales
- The kept delete guard and its reasoning check out — `sales_shipment_items.order_line_id` does front it
- The `sales.shipments.create` over-ship guard is untouched, as claimed (though it's prospective only — it can't detect a quantity lowered *after* dispatch)
- No `.ai/specs/` entry governs this; the PR's N/A is correct
- No backward-compatibility surface from `BACKWARD_COMPATIBILITY.md` is broken — relaxing a 409 is permissive, and i18n keys aren't a protected surface
- No `.tsx` touched; no DS/UI concerns

### Suggested labels

Applying `priority-high` and `risk-high` rather than the proposed `priority-medium`/`risk-medium`: the change reverts the fix for an open bug and sits on a write path that moves money on dispatched orders. Also applying `needs-qa` as you proposed.

pkarw:
🏷️ label rationale — updated 2026-08-27 (re-review of `c9acc00`)

`changes-requested` stays: the re-review found one major (the migration ships without its `.snapshot-open-mercato.json` update) plus a minor and a nit. The previous round's blocker and major are both resolved — the freeze is no longer removed, only gated behind a default-off setting, and `TC-SALES-4056` is restored rather than inverted.

`priority-high` and `risk-high` kept, on revised grounds. The original priority rationale — "this reverts the merged fix for an open bug (#4056)" — no longer holds and is withdrawn; the change is now additive with a behaviour-preserving default. Priority stays high because it is a partner-blocking integration change with maintainer pressure on the thread. Risk stays high on its own terms: it ships a schema migration and gates a write path that moves order money, which is what `risk-high` is for.

`needs-qa` kept as the author proposed. The self-QA evidence is good, but the new toggle is an admin surface that changes what a public API accepts, so it deserves a click-through.

pkarw:
_(superseded — the label rationale is consolidated in the earlier comment, updated for the 2026-08-27 re-review. The claim that this PR reverts the fix for #4056 no longer applies to the current head.)_

pkarw:

Note on CI: this PR had no check signal because both workflow runs were parked at `action_required` (fork PR). I approved them, and `ephemeral-integration` failed on `TC-SALES-4056` — worth pulling before the next push. Fork runs need re-approval on each new push.

pkarw:

1 blocker, 1 major, 2 minor, 1 nit. No autofix iterations ran: the blocker is a product decision (whether open issue #4056 should be reopened), and this is a fork PR, so the fix is the author's or a maintainer's call rather than something automation should settle.

jtomaszewski:
@pkarw I believe https://github.com/open-mercato/open-mercato/issues/4056 is not a bug, it's a feature. Editing order should be possible even after it had been shipped.

With whom can we discuss this? cc @alinadivante @andrzejewsky as I see you created/assigned to https://github.com/open-mercato/open-mercato/issues/4056 .

pkarw:
Ok got it, please discuss it with @alinadivante on Discord - also fix the CI

kamwro:
Im on fixing the CI

adeptofvoltron:
Thinking out loud here rather than filing this as a formal review comment — it's a design question, not a defect.

I agree with the underlying reasoning: corrections after dispatch are ordinary retail, and the current 409 does block them. My concern is that the fix is unconditional. Every existing deployment that upgrades gets a different contract on `sales.orders.lines.upsert` without doing anything, and some of them have been running with the freeze in place since #3993 landed. They aren't relying on it because they opted in — they're relying on it because it has been the platform's behavior. A change that turns a hard 409 into a successful write is exactly the kind that nobody notices until an order's commercial terms have already been rewritten.

I don't think the RBAC answer fully covers that case. The relevant feature is `sales.orders.manage`, which gates all order-line writes; there's no way to express "this role may edit order lines, but not after dispatch". A deployment that wants the freeze as policy would have to give up more than it gains.

Would you consider shipping this as a configuration option instead of a removal?

- Keep the guard, put it behind a setting (something in the spirit of the existing `OM_OPTIMISTIC_LOCK` switch, e.g. `OM_SALES_SHIPPED_LINE_FREEZE`).
- Default it to **on** — today's behavior — so upgrading stays a no-op for every existing project.
- Deployments mirroring a system of record, the case that motivated this PR, turn it off and get exactly the behavior you're implementing.
- The rewritten test file would then pin both branches rather than one, which gives us the regression proof in both directions.

That keeps the change additive under `BACKWARD_COMPATIBILITY.md`, and leaves the door open to flipping the default in a later major once we know which behavior deployments actually want.

If you think the default should be off from day one, I'm open to the argument — but in that case I'd want it called out in `UPGRADE_NOTES.md`, so upgraders make an informed decision instead of discovering it in production.

kamwro:
@pkarw could you re-run the CI now?

jtomaszewski:
> That keeps the change additive under BACKWARD_COMPATIBILITY.md, and leaves the door open to flipping the default in a later major once we know which behavior deployments actually want.

Yeah, I suggest this becomes a default in some next major, because it just seems more reasonable.

@kamwro can you add a feature flag? And pls debug if there's any more "restrictions" like this in the sales module. I'd like to know beforehand if the problem's bigger than just this one.

kamwro:
@jtomaszewski @adeptofvoltron — flag is in, and the survey came back bigger than one. Pushed as three commits.

## The flag

I didn't add a new mechanism, because the sales module already has one. `sales_settings` carries `orderCustomerEditableStatuses` and `orderAddressEditableStatuses`: enforced in `applyDocumentUpdate` and `assertAddressEditable`, edited through `components/OrderEditingSettings.tsx` / `api/settings/order-editing`, and read back by the document detail page so the UI disables what the API would refuse. The freeze joins them as `orderShippedLineEditable`.

Per-tenant rather than per-deployment, and — the part that decided it — **it already reaches the browser**. `LineItemDialog` carries its own copy of the freeze: it disables the pricing controls, floors the quantity input and strips those fields from the payload before sending. Behind an env var the server would accept a correction the admin could never send, and the switch would look like it did nothing. It reads the same setting now.

`order_shipped_line_editable` is `not null default false`, and a scope with no `sales_settings` row resolves the same way, so **upgrading is a no-op** — exactly the property you asked for, @adeptofvoltron. `UPGRADE_NOTES.md` says so, and says the default is expected to flip in a later major, so a deployment that wants the freeze as policy should set it explicitly rather than rely on the default staying put.

The guard itself is restored byte for byte — `git diff 55b83100c^` on `commands/documents.ts` is the import, a header comment and two `if`s. Two boundaries deliberately ignore the setting: `sales.shipments.create` still rejects shipping more than the remaining quantity (the invariant #3993 asked for), and a line with shipment items still cannot be deleted, because that guard fronts the `sales_shipment_items.order_line_id` foreign key.

**This also settles the blocker.** With the default unchanged, `TC-SALES-4056` goes back to its original assertions rather than the inverted ones — restored verbatim, with a new case beside it that flips the setting and reads each correction back. The delete-refusal case now runs with the setting ON, which is the only place that boundary can be proved independent of the flag.

Verification: `yarn typecheck` clean (26/26); the full `@open-mercato/core` jest run green (1,426 suites / 11,469 tests); `yarn i18n:check` clean. The two test files pin both branches, and I checked that by mutation rather than by reading — forcing the resolver to `false` reddens the ten enforcement cases, forcing it to `true` reddens the ten opt-in ones, and the sets are disjoint.

## The survey: five more, and two of them are worth a look on their own

Guards in the sales module that refuse a write because of what already happened to the document, as opposed to input validation:

| # | guard | where | trips on | code |
|---|---|---|---|---|
| 1 | shipped-line freeze | `commands/documents.ts` | line has shipment items | 409 — now behind the setting |
| 2 | add a line to a fulfilled order | `assertOrderAcceptsNewLine`, `commands/documents.ts` | `status` or `fulfillmentStatus` equals the literal `"fulfilled"` | 409 |
| 3 | delete a shipped line | `commands/documents.ts` | line has shipment items | 409 — fronts the FK |
| 4 | delete the last line | `commands/documents.ts` | order would be left with zero lines | 409 |
| 5 | update a shipment on a fully paid order | `commands/shipments.ts` | `paidTotalAmount >= grandTotalGrossAmount` | 422 |
| 6 | update a shipment on a fully refunded order | `commands/shipments.ts` | `refundedTotalAmount >= grandTotalGrossAmount` | 422 |

Plus two client-only locks with no server counterpart: `LineItemDialog`'s pricing lock (the UI half of #1, now wired to the setting) and the currency lock on the document detail page.

**#5 and #6 are not invariants.** `sales.shipments.create` and `sales.shipments.delete` carry no such check — only `update` does. Delete-then-create clears the restriction completely, which is what a full re-sync does anyway. As written the guard only stops the caller that tries to do the polite thing. I'd file that as a defect rather than fold it in here.

**#2 compares against a literal.** `"fulfilled"` is a seeded dictionary value, so the guard applies unevenly: a tenant whose fulfillment dictionary carries its own codes never trips it, however fulfilled the order is. The deployment behind this PR runs its own status vocabulary and has never once seen this 409.

**Issuing an invoice locks nothing.** `sales.invoices.create` / `.update` have no state guard, and neither does `sales.orders.delete`. So the module's restrictions don't line up behind a single idea of when a document becomes final — the same observation this PR made about paid versus shipped.

Scope: this PR flags #1 only. #2 and #5/#6 are better off as their own issues, which I'll open — say so if you'd rather they rode along here.

@pkarw — could you approve the workflow runs again? The push parked them at `action_required`.

kamwro:
Filed as #5680 (the shipment guard that only `update` carries) and #5681 (the literal status match).

Also merged `upstream/develop` into the branch — it had gone 73 commits stale and conflicted on `UPGRADE_NOTES.md`; both entries are kept. Re-verified on the merge: `yarn typecheck` 27/27, full `@open-mercato/core` jest 1,453 suites / 11,696 tests green.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T00:38:52Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review — Re-review: feat(sales): make the shipped-line freeze a per-scope setting, on by default

**Verdict: CHANGES REQUESTED** — 0 blockers, 1 major, 1 minor, 1 nit.

**Validation source:** GitHub checks. As on the previous pass, both workflow runs for `c9acc00` were parked at `action_required` (fork PR), so the PR again had **no check signal** — `gh pr checks` showed only the CLA and a skipped bot. I approved both runs after confirming the diff touches no workflow, script or manifest file. `lint`, `ds-lint` and `audit-scope` are green; `prepare` and everything gated behind it (`scope`, `test`, `documents-multi-instance`, `ephemeral-integration`) were still running when this review was submitted — see the CI-pending note at the end. No local fallback was needed for the findings below; each one is established from the tree itself.

| check | result |
|---|---|
| `lint`, `ds-lint`, `audit-scope` | ✅ |
| `prepare`, `scope`, `test`, `documents-multi-instance`, `ephemeral-integration` | ⏳ pending at review time |
| `Mutation tests` | ✅ — but its allowlist is `['shared']`, so it skipped this change rather than covering it |

---

## Previous round — every blocker re-checked

| finding | state |
|---|---|
| 🔴 **Blocker 1** — undoes the fix for open #4056 | ✅ **Resolved.** The freeze is no longer removed at all: `git diff origin/develop...HEAD -- commands/documents.ts` is an import, a comment block and two `if`s. `isShippedLineFreezeEnforced` returns `settings?.orderShippedLineEditable !== true` (`commands/settings.ts:37`), so a scope with no `sales_settings` row and a scope with the flag off both stay frozen. #4056's reported scenario is refused by default, and `TC-SALES-4056` is restored rather than inverted. |
| 🟠 **Major 1** — the totals check removed on a premise false for gross | ✅ **Resolved.** `hasShippedLineTotalChange` is back untouched, and the comment that stated the false premise is replaced with an accurate one (`documents.line-shipped-guard.test.ts`, the "allows a quantity change whatever totals the client sends" case) that names `buildBaseLineResult`'s verbatim `totalGrossAmount` pass-through and its route into `outstandingAmount`. Correctly scoped as pre-existing and out of scope here. |
| 🟡 **Minor 1** — the quantity floor becomes client-side only | ✅ **Resolved by construction.** The server-side floor is intact by default; it only lifts for a scope that opted in, which is the deliberate contract. `restoreOrderGraph` asks the same question (`documents.ts:7076`), so undo cannot become the one path where the freeze survives an opt-out. |
| 🟡 **Minor 2** — `expectAllowed` can't prove the PR's thesis | ⚪ **Addressed in substance, not literally.** See the nit below. |
| 🔵 **Nit** — the second dead `errorPriceShipped` | ⚪ Untouched, and correctly so — it is pre-existing on `develop` and was flagged as not yours. |

The design question raised on the thread was settled the right way: the freeze is now `sales_settings.order_shipped_line_editable`, defaulting to enforced, and both halves — the command layer and `LineItemDialog` — read the same value, so the toggle cannot become a switch that appears to do nothing. `UPGRADE_NOTES.md` states the no-op upgrade, what opting in grants, and the two boundaries that do not move. That is the shape @adeptofvoltron and @jtomaszewski asked for.

---

## 🟠 Major 1 — The migration ships without its snapshot update, so the next `yarn db:generate` re-emits the column

`Migration20260826120000_sales_settings_shipped_line_editable.ts` adds `sales_settings.order_shipped_line_editable`, but `packages/core/src/modules/sales/migrations/.snapshot-open-mercato.json` was not regenerated — its `sales_settings` entry still reads:

```
['created_at', 'deleted_at', 'id', 'order_address_editable_statuses',
 'order_customer_editable_statuses', 'order_number_format',
 'organization_id', 'quote_number_format', 'tenant_id', 'updated_at']
```

That snapshot is the live one, not a leftover: `packages/cli/src/lib/db/commands.ts:96` returns `.snapshot-open-mercato` and `:277` hands it to the Migrator as `snapshotName`, which is what `db:generate` diffs the entities against. With the column present on `SalesSettings` (`data/entities.ts:794`) and absent from the snapshot, the **next** `yarn db:generate` anyone runs — for any module — emits a second `alter table "sales_settings" add column "order_shipped_line_editable" boolean not null default false;`. Against a database that already ran `Migration20260826120000` that fails with `column "order_shipped_line_editable" of relation "sales_settings" already exists`, and the contributor who hits it has no reason to connect it to this PR.

The two most recent sales migrations both carry the snapshot in the same commit — `69113e169` (`Migration20260821120000_sales_notes_addresses_context_idx`, +60 lines of snapshot) and `1e5976efc` (`Migration20260806120000_sales_tag_assignment_document_idx`, +10) — and `AGENTS.md` names it explicitly: *"update ORM entities, run `yarn db:generate`, and review the generated SQL plus `migrations/.snapshot-open-mercato.json`."* No CI job checks for this drift, which is why nothing has flagged it.

Please add the `order_shipped_line_editable` column to the `sales_settings` table in `packages/core/src/modules/sales/migrations/.snapshot-open-mercato.json`, ideally by regenerating rather than hand-editing, keeping the migration file itself exactly as it is.

## 🟡 Minor 1 — The user guide still describes the editing-guards card as having two controls

`apps/docs/docs/user-guide/sales/sales-settings.mdx:17-22` documents that card by enumerating it:

```
## Editing guards
- Decide when **customers** and **addresses** can change (any status vs only draft).
- Changing customer clears addresses to prevent data mix-ups.
```

The PR adds a third control to that exact card and, unlike its two neighbours, this one changes what a public API accepts on a money-bearing write path — an operator turning it on is granting `PUT /api/sales/order-lines` the ability to rewrite the price of a dispatched line. `UPGRADE_NOTES.md` covers the upgrade audience well, but that file is not where an operator looks before flipping a switch in the admin. A bullet in this section — what it does, and that deletion stays blocked either way — closes the gap. The PR checklist item *"I updated documentation, locales, or generators if the change requires it"* is ticked; locales and generators are fine, this is the part that is not.

## 🔵 Nit — `expectAllowed` still only asserts "not a 409"

Carried from the previous round, and now largely overtaken by events, so this is a note rather than a request. `expectAllowed` (`documents.line-shipped-guard.test.ts:264`) still asserts only the absence of a 409, never that the corrected values landed — a 400, a 500 or a silently-dropped write all pass it. The coverage gap that made this worth raising is genuinely closed at the integration tier: the new `applies the same corrections once the scope sets orderShippedLineEditable` case reads the line back after every correction and asserts the submitted value, which is exactly the "accepted-and-dropped" failure the unit helper cannot see. If you are in the file for the snapshot fix anyway, adding `expect(em.flush).toHaveBeenCalled()` inside `expectAllowed` costs one line and makes the unit tier honest on its own terms.

---

## ✅ Verified clean

- **The freeze itself is byte-identical to `develop`** — the whole of `hasShippedLineTotalChange`, `hasExactNumericChange`, `hasUnitChange` and the two locked-field tables are untouched; only the two entry points gained a guard, and both ask the setting *before* the `loadShippedQuantityByLine` join, so an opted-out scope pays nothing.
- **Every command-layer enforcement point is gated consistently** — `assertShippedOrderLineEditable` (`documents.ts:7175`) and `assertShippedOrderGraphRestorable` (`documents.ts:4406`) are the only two callers of the shipped-quantity load, and both are behind the same resolver. The delete guard at `documents.ts:7490` is deliberately outside it, and the reasoning holds: `sales_shipment_items.order_line_id` fronts it, so lifting it would only swap an explained 409 for a raw FK error.
- **Absence and explicit `false` agree** — `!== true` handles the missing-row case and the untouched-row case identically, and the test block pins both routes separately rather than assuming they converge.
- **Settings round-trip is safe for the other two guards** — `salesEditingSettingsSchema`'s status lists are `.optional().nullable()` and `sales.settings.save` updates each field only when `!== undefined`, so a `PUT` carrying just `{ orderShippedLineEditable: true }` cannot null out `orderCustomerEditableStatuses` / `orderAddressEditableStatuses`. The integration helper relies on exactly that.
- **Both halves read one value** — API → `GET /api/sales/settings/order-editing` → document detail page → `SalesDocumentItemsSection` → `LineItemDialog`, defaulting to `false` at every hop, so an unaware caller keeps today's lock. The unresolved-shipments fallback correctly follows the setting: with no refusal to mirror there is nothing to be cautious about.
- **Backward compatibility** — `not null default false` column, an optional zod field, an added response key, no removed export or changed signature. Nothing in `BACKWARD_COMPATIBILITY.md` is touched; relaxing a 409 behind an opt-in default-off setting is additive by the deprecation protocol's own terms.
- **Tenancy and RBAC** — the resolver scopes on `order.tenantId` / `order.organizationId`; both route methods are `requireFeatures: ['sales.settings.manage']`.
- **i18n** — three new keys plus one reworded description across all five locales, inserted in sorted position (`shippedLineLabel` < `shippedLineOff` < `shippedLineOn`), each referenced from `OrderEditingSettings.tsx`, so `i18n:check-sync` and `i18n:check-usage` have nothing to catch.
- **Design system** — the new control is a `Switch` in the existing card, `text-sm`/`text-xs`, `text-muted-foreground`, `aria-label` from the translated label, disabled while loading or saving. No status colors, no arbitrary values, no `dark:` overrides. `ds-lint` is green.
- **Tests genuinely pin both branches** — the opt-in and enforcement blocks cover the same edits, the enforcement block carries controls (raise quantity, lower to exactly shipped, unshipped line, echoed unit) so it cannot pass against a guard that refuses everything, `expectRefused` asserts the message and not just the status, the delete case is asserted *inside* the opted-in block where letting it through would be the mistake, and the integration delete case adds a spare line so the 409 cannot be the "cannot delete the last line" guard wearing the same code. The `beforeAll` comment explaining why the second describe block must not re-clear the command registry is the kind of note that saves the next person an hour.
- **Integration hygiene** — fixtures created per case, orders deleted in `finally`, the setting read before and restored after, and the tier's `workers: 1` noted as why the window cannot overlap.

---

## ⏳ CI is still pending

`prepare` and the jobs behind it (`scope`, `test`, `documents-multi-instance`, `ephemeral-integration`) were still running when this review was submitted, so nothing here rests on them and none of the findings above depends on their outcome. **The `ephemeral-integration` result is the one that matters most on this PR** — it is where the previous round's blocker surfaced, and `TC-SALES-4056` has been substantially rewritten since. I will post a follow-up comment with the result. Note that fork runs need re-approval on every new push, so a later push will park at `action_required` again until a maintainer approves it.

Also worth knowing: the green `Mutation tests` check is not evidence about this change — its package allowlist is `['shared']`, so it skipped `packages/core` entirely.

---

## Labels

`changes-requested` stays, driven by the major above. Keeping `priority-high` and `risk-high`: the original priority rationale (reverting the fix for an open bug) no longer applies, but this is a partner-blocking integration change, and `risk-high` still fits on its own terms — it ships a schema migration and gates a write path that moves order money. `needs-qa` stays; the self-QA evidence in the Testing section is good, but the toggle is an admin surface that changes what the API accepts and deserves a click-through.

pkarw:
Thanks @kamwro — this is a much better shape than the last round, and the two things that blocked it are properly resolved: the freeze is gated rather than removed, `TC-SALES-4056` is restored rather than inverted, and both halves (command layer and `LineItemDialog`) read the same setting, so the toggle cannot become a switch that does nothing. @adeptofvoltron's concern about changing the contract for deployments that never opted in is answered by the default.

Handing it back for one substantive item plus one small one:

1. **Major — the migration is missing its snapshot update.** `packages/core/src/modules/sales/migrations/.snapshot-open-mercato.json` still has `sales_settings` without `order_shipped_line_editable`, so the next `yarn db:generate` anyone runs re-emits the column as a fresh migration, which then fails on any database that already ran `Migration20260826120000`. The last two sales migrations (`69113e169`, `1e5976efc`) both carry the snapshot in the same commit. Regenerate rather than hand-edit if you can; leave the migration file itself as it is.
2. **Minor — `apps/docs/docs/user-guide/sales/sales-settings.mdx:17` still describes the editing-guards card as two controls.** A bullet for the new toggle, noting that deleting a shipped line stays blocked either way, closes it. `UPGRADE_NOTES.md` covers the upgrade audience well; this is the operator-facing half.

The nit about `expectAllowed` is genuinely optional — the integration read-backs you added cover what it was pointing at.

On CI: your runs were parked at `action_required` again (fork PRs need re-approval on every push). I approved them; `lint`, `ds-lint` and `audit-scope` are green and `prepare` was still going when I posted the review. I will follow up here with the `ephemeral-integration` result, since that is where the last round's blocker showed up. When the two items above are pushed, re-request review and the automation can pick it up from the new head.

pkarw:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED. Lock released (swapped for `ci-monitoring` — the review is finished and the PR is free to act on; only the CI result is still outstanding).

1 major, 1 minor, 1 nit. Both of the previous round's blocking findings verified resolved. No autofix iterations ran: this is a fork PR authored by someone else, so the fix is the author's call — re-run with `--autofix` to have it fixed here instead.

pkarw:
🤖 `om-auto-review-pr` — CI result for `c9acc00`

Both fork runs are settled. **`CI for Develop&Main` failed**, on one test — and it is this PR's own new one, not a pre-existing spec.

| job | result |
|---|---|
| `prepare`, `lint`, `ds-lint`, `audit-scope`, `test`, `documents-multi-instance` | ✅ |
| `ephemeral-integration (none)` | ❌ **1 failed**, 232 passed, 5 skipped, **flaky: 0** (failed on retry #1 too) |
| `Mutation tests` | ✅ — but its package allowlist is `['shared']`, so it skipped `packages/core`; it is not evidence about this change |

(For anyone reading the job list: `(none)` is the affected-only shard's name, not a skip. The shard ran.)

```
TC-SALES-4056: shipped order-line corrections
  › applies the same corrections once the scope sets orderShippedLineEditable

  Error: discount amount should have been written, not silently dropped
  Expected: 25
  Received: 20
  at TC-SALES-4056-order-line-price-after-shipment.spec.ts:221
```

### This is a test bug, not a product bug — the setting works

The write was accepted (`200`) and applied. What the assertion gets wrong is that `discountAmount` is not a round-trip field, for two independent reasons, and the loop's own previous iteration triggers both.

`buildBaseLineResult` (`packages/core/src/modules/sales/lib/calculations.ts:88-95`):

```ts
const discountPerUnit = line.discountAmount ?? (…percent-derived…)
const netSubtotalBeforeDiscount = toNumber(unitNet, 0) * quantity
…
discountAmount: round(discountTotal),
```

1. **In is per unit, out is per line.** The submitted `discountAmount` is read as `discountPerUnit` and persisted as `discountPerUnit × quantity`. Even unclamped, submitting `25` on a 4-unit line stores `100`, never `25`.
2. **It is then clamped to the line's net subtotal.** The loop's first correction has already set `unitPriceNet: 5` on this 4-unit line, so `netSubtotalBeforeDiscount` is `5 × 4 = 20`, and `min(100, 20) = 20`. Hence the received `20`.

Both behaviours are pre-existing and unrelated to `orderShippedLineEditable`; the new case is simply the first test to assert a naive round-trip on that field. The `discountPercent` iteration that follows passes only because percent *is* stored verbatim.

The root cause underneath is that the `corrections` table is **cumulative** — each iteration reads the line state the previous one left behind, so the unit-price correction silently changes what the discount correction can mean. Two ways out, either is fine:

- Keep the loop but make the discount case's expectation derived rather than literal — assert `min(submitted × quantity, unitPriceNet × quantity)` — and say in a comment why this one field is not a round-trip.
- Or give each correction a fresh line (or reset the price between iterations) so the cases stay independent. That also removes the ordering trap for whoever adds a sixth correction later.

Worth doing properly rather than by loosening the assertion: the read-back is the part of this test that earns its keep — it is what distinguishes "the correction was applied" from "the payload was accepted and dropped", which is exactly the half-wired-setting failure mode. It just needs to model what the field actually stores.

This does not change the review verdict, which was already `changes-requested` on the missing migration snapshot. Removing `ci-monitoring` — the CI follow-up promised in the review is now delivered, and no further follow-up will come from this agent. Note that the next push will park the fork runs at `action_required` again until a maintainer approves them.

kamwro:
Both items are in, plus the CI failure — which was mine.

**Major — migration snapshot (`04dcf583e`).** Confirmed exactly as described: `sales_settings` in `.snapshot-open-mercato.json` had no `order_shipped_line_editable`, so the next `db:generate` would have re-emitted the column against a database that already ran the migration.

Written by hand rather than regenerated — `db:generate` needs a live database and this worktree has none, so I would rather say that plainly than pretend it was generated. It is copied from what MikroORM emits for the identical declaration elsewhere in the same file (`sales_channels.is_active` is the same `@Property({ type: 'boolean', default: … })` shape, down to the default being the SQL literal as a string) and inserted in the alphabetical position the snapshot keeps its columns in. The diff is 16 lines and nothing else in the file moved. The migration itself is untouched. Worth a second pair of eyes on that one file, since I could not verify it the way the tooling would.

**CI — `TC-SALES-4056` (`0768a93f9`).** Your diagnosis is right on both mechanisms, and I checked `buildBaseLineResult` rather than take it on trust: `discountAmount` is read as a per-unit discount, stored as `perUnit × quantity`, then clamped to the line's net subtotal — so on a line the previous iteration had already dropped to `unitPriceNet: 5`, `min(25 × 4, 5 × 4)` is the 20 the run reported. Nothing to do with the setting; my case was simply the first to assert a round-trip on a field that does not round-trip.

I took the second of your two options and then some: each correction now restores an explicit baseline first, so the cases are independent, and the discount expectation is derived from what core stores rather than from what was submitted. One thing that bit me while fixing it — the baseline has to name **both** discount fields, because the upsert merges an omitted field with the line's current value, so a reset that just left them out would have carried the previous case's discount forward and the independence would have been a comment rather than a fact.

The read-back stays, for the reason you gave: it is what separates an applied correction from an accepted-and-dropped one, which is the half-wired-setting failure this spec exists to catch.

**Minor — docs (`e67a6a954`).** `sales-settings.mdx` now has the third bullet: what the toggle unlocks, who it is for, that the line dialog follows it, and that deleting a shipped line stays blocked either way.

Left the `expectAllowed` nit alone, as you suggested.

Local: `yarn typecheck` 27/27, `@open-mercato/core` sales suites 98/98 (726 tests). I cannot run `ephemeral-integration` here, so `TC-SALES-4056` is unverified on my side by design — the push will have parked the fork runs at `action_required` again, so it needs another approval when you get a moment.

I cannot re-request review myself — this account is read-only here, so the API returns 404 on the reviewer endpoint. Treating this comment as the hand-back instead; the head to pick up from is `e67a6a954`.

jtomaszewski:
We'll close this PR for now, think of how this "feature" can be toggled, and come back with a newer one.

Issue #5344, opened 2026-08-17
bug(sales): documentUpdateSchema silently strips create-schema fields, and on quotes a no-op update still invalidates a sent quote


kamwro:
### What

`documentUpdateSchema` (`packages/core/src/modules/sales/commands/documents.ts:587`) is a closed
`z.object` shared by `sales.orders.update` and `sales.quotes.update`. Two related defects follow
from that, both of the same class as #5335 but out of its scope.

### 1. A quote update that changes nothing still degrades a sent quote

`updateQuoteCommand` clears `acceptanceToken` and `sentAt` and reverts `status` to `draft` for any
quote in `sent` status whose update reaches execution — regardless of whether the payload changes
anything. `expectedDeliveryAt` reaches execution while writing nothing: the shared schema accepts it
and its `.refine()` counts it as a real edit, but `applyDocumentUpdate` guards the write on
`"expectedDeliveryAt" in entity` and `SalesQuote` has no such column.

Measured against `develop` at `655ac3fe9`, driving `sales.quotes.update` with
`{ id, expectedDeliveryAt: '2026-12-01' }` on a quote with `status: 'sent'`:

    outcome          = RESOLVED (success)
    expectedDelivery = undefined      (nothing written)
    acceptanceToken  = null           (was 'tok-customer-link')
    status           = "draft"        (was 'sent')

So a caller that reuses an order-shaped payload against the quotes endpoint silently kills the
customer's acceptance link on every sync, while reporting success and changing no field it asked
for. `dueAt` has the same shape (absent from `SalesQuote`).

### 2. Eight create-schema fields are still stripped on update

Declared by the create schemas (`data/validators.ts:696-715`, `:745-760`) and absent from
`documentUpdateSchema`, so each is dropped in silence on update exactly as `comments` and
`internalNotes` were before #5335:

`validFrom`, `validUntil`, `dueAt`, `taxInfo`, `deliveryWindowId`, `deliveryWindowCode`,
`taxStrategyKey`, `discountStrategyKey`

Notably a quote's `validUntil` cannot be changed through this schema at all, though `SalesQuote`
declares the column.

### Suggested direction

A kind-aware rejection for fields the target entity cannot store — the same shape #5335 adopts for
`internalNotes` on quotes — would close (1) as a class rather than field by field. (2) is a
separate decision per field about whether update should accept it at all.

Found while reviewing #5335, which fixes only the `comments` / `internalNotes` half.

[...]

PR #4796, opened 2026-08-01
fix(sales): deprecate ignored order payment totals with warnings

pkarw:
Refs #4695

Supersedes #4768. GitHub reported `maintainerCanModify: true`, but the fork owner’s organization policy rejected the maintainer push with HTTP 403. This upstream replacement preserves @jtomaszewski’s original authored commit (`7a062534`) and layers the compatibility correction on top.

Source doc: `.ai/specs/2026-08-01-sales-order-payment-ledger-input-deprecation.md`

## 🎯 Goal

Make ignored order payment-ledger inputs observable without immediately breaking integrations that rely on the released create contract.

## 🔍 Problem

`orderCreateSchema` accepts `paidTotalAmount`, `refundedTotalAmount`, and `outstandingAmount`, while `sales.orders.create` intentionally derives those totals from payment records. Callers could therefore submit values that were silently ignored. The original #4768 change made the loss visible by rejecting the fields, but that narrowed released validator, command, and HTTP contracts without the required deprecation window.

## 🔍 Root Cause

Order-level payment totals are projections of `SalesPayment` and `SalesPaymentAllocation` rows, not an independent writable ledger. The create contract nevertheless exposed those projections as optional inputs and provided no signal when the command ignored them.

## What Changed

- Keeps all three released inputs accepted and ignored for at least one minor release, with `@deprecated` TypeScript guidance.
- Returns a stable `sales.order.payment_ledger_input_deprecated` warning from the create command and HTTP response, naming only the supplied fields.
- Emits one bounded structured operator warning per process without logging submitted values or customer data.
- Keeps the fields in OpenAPI, documents the payment-record migration path, and declares the optional warning response.
- Adds the compatibility spec, upgrade note, command/route/OpenAPI regression tests, and a self-contained real API integration scenario.
- Preserves @jtomaszewski’s diagnosis and original commit in the replacement branch history.

## 🧪 Tests

- `yarn workspace @open-mercato/core test src/modules/sales/commands/__tests__/documents.create-payment-totals.test.ts src/modules/sales/api/__tests__/documents.routes.test.ts --runInBand` — 13/13 passed.
- `yarn mercato test:integration TC-SALES-040-order-payment-ledger-deprecation` — 1/1 passed against an isolated ephemeral app; create returned the exact warning, persisted totals remained payment-derived, and cleanup succeeded.
- Configured local gate: package builds, generation, i18n sync/usage checks, typecheck, and production app build passed.
[...]

PR #4768, opened 2026-07-31
fix(sales): reject order payment totals on create instead of discarding them

jtomaszewski:
## Summary

`orderCreateSchema` declared `paidTotalAmount`, `refundedTotalAmount` and `outstandingAmount` (via `...orderTotalsSchema.shape`), so zod validated them and they arrived as `parsed.*` — but `sales.orders.create` built the `SalesOrder` with the ledger hardcoded to `"0"` and then recomputed the totals with `existingTotals: resolveExistingPaymentTotals(order)`, i.e. the zeros it had just written. A 100.00 order created with `paidTotalAmount: 100` persisted as `paid_total_amount = 0` / `outstanding_amount = 100`, with no error, warning or log.

This PR makes that failure **visible** rather than silent: the three keys are now rejected on the create surface.

**Why reject rather than honour the value.** Those three columns are a projection of the order's payments — `recomputeOrderPaymentTotals` (`commands/payments.ts:252`) rebuilds them by summing the `SalesPayment` / `SalesPaymentAllocation` rows and assigns the result, on every payment create, update, delete and refund (10 call sites). A value seeded on the document has no payment rows behind it, so the next payment touch overwrites it: on an order created with `paidTotalAmount: 100`, a first real payment of 20 would set paid to **20**, not 120. Honouring the input would have introduced a second, unreconciled source of truth that silently loses to the first payment write — an order whose balance its own payment history contradicts.

## Changes

- `data/validators.ts` — the three keys stay declared on `orderCreateSchema` but as `z.never({ error: … }).optional()`. Declaring them matters: an *undeclared* key is silently stripped by zod, which is the same silent-discard bug one layer down. Any supplied value (including `0` and `null`) now fails validation with an actionable message on that field's path; the CRUD factory maps the `ZodError` to a **400** (`shared/lib/crud/factory.ts:556`). Exported `ORDER_PAYMENT_LEDGER_FIELDS` / `ORDER_PAYMENT_LEDGER_INPUT_MESSAGE` as the single definition.
- `api/documents/factory.ts` — `buildDocumentOpenApi` omits the three keys, so the documented create surface stays accurate. (Left in the runtime schema they would render as a bare `{}` property, which reads as "any value accepted".)
- Enforcement is single-point: the route parses `orderCreateSchema` in `mapInput` and the command parses it again, so HTTP callers, direct `executeCommand` callers (inbox actions, AI tools, bulk import) and third parties all get the identical failure. TypeScript callers now get a compile error too.
[...]

PR #5573, opened 2026-08-24
fix(sales): accept exchangeRate and the status entry ids on document update

maxidragon:
## Summary

`documentUpdateSchema` (`modules/sales/commands/documents`) is a closed `z.object` shared by `sales.orders.update` and `sales.quotes.update`. Three fields the **create** schema (`modules/sales/data/validators`) declares are absent from it:

| Field | Create schema | On update, before this PR |
|---|---|---|
| `exchangeRate` | `decimal({ min: 0 }).optional()` | stripped |
| `paymentStatusEntryId` | `uuid().optional()` | stripped |
| `fulfillmentStatusEntryId` | `uuid().optional()` | stripped |

Zod strips any key the object does not declare — no error, no warning, no validation failure. `applyDocumentUpdate` never sees the keys and the command **returns success having written nothing**. Creates are unaffected, which is what makes it hard to spot: the columns keep whatever the document carried when it was first created.

The caller this bites is a system-of-record integration that re-syncs orders on a schedule. Every run reports each document updated, and every run looks green, while payment status, fulfillment status and exchange rate stay frozen at their create-time values indefinitely.

Same class as #5335 (`comments` / `internalNotes`, merged 2026-08-18), and this PR follows it in shape. It closes part (2) of #5344 for three fields, and demonstrates the kind-aware rejection that issue's "Suggested direction" asks for.

## Changes

- **`documentUpdateSchema`** — accept all three, nullable (unlike on create) because an update has an existing value to clear.
- **Its `.refine()`** — count each as a change, so an update whose only edit is one of them is no longer rejected as `update_payload_empty`.
- **`applyDocumentUpdate`** — apply on an order, reject with a 400 on a quote.
- **`buildDocumentUpdateChangeKeys`** — add each, and for the entry ids the derived text key too. This feeds `buildLog`'s audit-log diff only; it does not reach reindex or events.
- **i18n** — three new keys in all five locales, matching `sales.quotes.internal_notes_unsupported` from #5335. Unresolvable entry ids reuse the existing `sales.documents.detail.statusInvalid`.

### Why a quote gets a 400 rather than a silent skip

All three columns exist on `SalesOrder` only — `SalesQuote` declares none of them. (`statusEntryId` is on both, which is why it is already accepted on update.)

Ignoring them on a quote would not merely be untidy. Each field now satisfies the `.refine()` on its own, so an otherwise-empty payload would reach execution — and `updateQuoteCommand` clears `acceptanceToken`, clears `sentAt` and reverts a `sent` quote to `draft` for any update that gets that far. A payload that changes nothing would destroy the customer's acceptance link. #5335's first revision shipped exactly that shape and was blocked on review; the tests below pin it in both directions.

### Why the entry ids write two columns

`paymentStatusEntryId` → `paymentStatus`, `fulfillmentStatusEntryId` → `fulfillmentStatus`. Order create already resolves all three ids into their text columns, and the existing `statusEntryId` branch of `applyDocumentUpdate` sets both. An update that moved the id and left the text stale would introduce a *new* self-contradicting row — the same defect class this PR exists to fix. Each id also 400s when a non-null value does not resolve, exactly as `statusEntryId` does.

### One deliberate deviation from the obvious spelling

`exchangeRate` is `z.union([z.null(), exchangeRateSchema])`, not the other order. `z.coerce.number()` accepts `null` — `Number(null)` is `0` — so with the operands reversed a clear would win the union as `0` and write a zero rate. There is a test for it.

### Scope

Deliberately narrow. The other create/update gaps (`dueAt`, `taxInfo`, `deliveryWindowId`, `deliveryWindowCode`, `taxStrategyKey`, `discountStrategyKey`, `deliveryWindowSnapshot`) are left to #5344; `lines` and `adjustments` are correctly absent, since separate commands own them. Nothing adjacent is tightened and no `.strict()` is added — per the #4768 / #4796 precedent, this PR only *widens* what update accepts.

`exchangeRate` is deliberately **not** added to `shouldRecalculateTotals`. Order create does not feed it into `buildCalculationContext` either — totals are computed in document currency and the rate is stored for reporting. Adding it would be a behaviour change beyond this fix.

Blast radius is the two update commands and the PUT route that parses through the same schema (`api/documents/factory.ts`).

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**
`.ai/specs/SPEC-021-2026-02-07-compound-commands-graph-save.md` references `documentUpdateSchema` only by composition and never enumerates its fields, so no spec text becomes wrong.

## Testing

New: `packages/core/src/modules/sales/commands/__tests__/documents.update-order-only-fields.test.ts`, 38 cases. Nine pin the schema; the rest drive the real `sales.orders.update` and `sales.quotes.update` commands against a mocked `EntityManager`, the same harness style as `documents.update-notes.test.ts`.

Per field: the order update writes it and a re-read confirms it; each entry id moves its derived text column with it; an unresolvable uuid returns 400 with nothing written; `null` clears the column (and its text); a quote carrying the field returns 400 **and survives** — `status` still `sent`, `acceptanceToken` still present, no stray property left on the entity; and a payload whose only edit is that field is accepted.

```
$ npx jest --config packages/core/jest.config.cjs --rootDir packages/core \
    --testPathPatterns documents.update-order-only-fields
Test Suites: 1 passed, 1 total
Tests:       38 passed, 38 total

# with the source change reverted, test file kept:
Test Suites: 1 failed, 1 total
Tests:       36 failed, 2 passed, 38 total
# the 2 survivors are the negative controls — "untouched when omitted"
# and "still writes fields a quote does have"

$ npx jest --config packages/core/jest.config.cjs --rootDir packages/core \
    --testPathPatterns "modules/sales"
Test Suites: 92 passed, 92 total
Tests:       690 passed, 690 total

$ yarn workspace @open-mercato/core test
Test Suites: 1 skipped, 1427 passed, 1427 of 1428 total
Tests:       2 skipped, 11491 passed, 11493 total
```

`yarn typecheck` (26/26), `yarn lint` (0 errors) and `yarn i18n:check-sync` / `check-usage` / `check-hardcoded` / `check-values` are all clean.

**Note for reviewers:** `documents.ts` carries `// @ts-nocheck` on line 1, so `yarn typecheck` says nothing about this file either way — the jest coverage above is the only real signal here. The `(entity as SalesOrder)` assignments were checked against `data/entities.ts` by hand: `exchangeRate` is `string | null` (hence `toNumericString`), and both entry ids and both derived text columns are `string | null`.

The `modules/sales` run requires `yarn generate` first; without it, `channels.route` and `shipments.afterList` fail to *load* on a missing `#generated/…` module and contribute 0 tests. `yarn build:app` and the cross-package `yarn test` were left to CI.

## Checklist

- [ ] I have read and accept the Open Mercato Contributor License Agreement (see `docs/cla.md`). — left for a human to confirm; an agent must not accept licence terms on the author's behalf.
- [x] I updated documentation, locales, or generators if the change requires it. — three new keys across all five locale files.
- [ ] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required). — covered at the command level, where the schema stripping actually happens; no HTTP or UI path is changed.
- [ ] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — n/a, per the Specification section above.
- [ ] Priority set: this PR carries exactly one `priority-*` label. — this account has `read` here and cannot set labels; intent is in a follow-up comment.
- [ ] Risk set: this PR carries exactly one `risk-*` label describing its blast radius. — same.
- [ ] QA routing set. — same; `needs-qa` is the intent, rationale in the follow-up comment.

### Design System Compliance

Not applicable — no UI surface is touched. The change is a zod schema, its command handler, and five locale files; no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`.

- [ ] No hardcoded status colors — n/a
- [ ] No arbitrary text sizes — n/a
- [ ] Empty state handled for list/data pages — n/a
- [ ] Loading state handled for async pages — n/a
- [ ] `aria-label` on all icon-only buttons — n/a
- [ ] Uses existing DS components — n/a

## Linked issues

Addresses part (2) of #5344 for three of the twelve fields it lists; the remaining nine stay with that issue. Context on these three specifically: https://github.com/open-mercato/open-mercato/issues/5344#issuecomment-5394760303

Complements #5335, which fixed the same class of silent stripping for `comments` / `internalNotes`.

## Note on CI

A cross-repository PR needs maintainer workflow approval **per push**, so CI has to be re-approved after every push to this branch. Package previews are unavailable here — both preview workflows are restricted to same-repository branches.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5573"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790171686&installation_model_id=432243&pr_number=5573&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5573&signature=52fcf4692782ce34c2fd4b91e5c654a6d24d0ab61b05fef3ce15057276372d4e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


maxidragon:
### Metadata intent (this account has `read` here and cannot set labels)

Posting the routing in one comment so a maintainer can apply it in one pass. Every name below was validated against `gh label list` for this repo.

| Label | Why |
|---|---|
| `bug` | A silent data-loss defect: the command reports success having written nothing. |
| `review` | Ready, non-draft, not in any other pipeline state. |
| `priority-medium` | Nothing is blocked on it today — the affected columns are latent for the reporting integration — but it silently corrupts an ongoing sync, so it should not sit indefinitely. Matches #5335. |
| `risk-medium` | The widening itself is additive and cannot break an existing caller. The blast radius is the new **400 on a quote**: a caller that previously sent one of these three fields to `sales.quotes.update` got a silent no-op and now gets an error. That is the intended fix — a silent no-op on a quote is the data-loss path described in the body — but it is a behaviour change for that caller, not a pure addition, and the risk label should say so. Same level as #5335. |
| `needs-qa` | The automated-verification exemption does not apply: it requires the API surface to be unchanged, and this PR both widens the accepted input and adds a rejection path. The DB schema is untouched and no `BACKWARD_COMPATIBILITY.md` contract is broken, but the API surface moved, so the gate stays on. |
| `partner-request` | Requested separately via `/label partner-request` above (Full Stack House, certified-partner registry). |

**Reviewer:** whoever owns the `sales` module's command layer — the change sits entirely in `commands/documents.ts` and its locale files. Deliberately not naming a handle from here.

Deliberately **not** requesting any claim label (`in-progress` and friends): this account cannot release one later, and a claim that cannot be released is worse than none.

Per this repo's QA gate, `needs-qa` blocks merge until a maintainer applies `qa-approved` — an evidence comment alone does not clear it.

pkarw:
🤖 `om-auto-review-pr` claiming this PR for review. Started: 2026-08-25T00:15:17Z.

pkarw (approved):
# 🔍 Code Review

**Verdict: ✅ Approve** — no blockers, no majors, no minors. Three nits below, none of which should hold the merge.

`documentUpdateSchema` is a closed `z.object`, so `exchangeRate`, `paymentStatusEntryId` and `fulfillmentStatusEntryId` were stripped before `applyDocumentUpdate` ever ran and `sales.orders.update` returned success having written nothing. This PR closes that for the three fields, and it does so in the same shape as #5335, which fixed the identical class for `comments` / `internalNotes`. I re-derived every load-bearing claim in the description against the code rather than taking it on trust; all of them hold.

**Validation source:** GitHub checks. The head's CI runs were sitting at `action_required` — a cross-repository PR needs maintainer workflow approval per push, so `gh pr checks` showed nothing but `license/cla` and a skipped `[code]smith`, which is indistinguishable from "clean". I approved runs `32735472886` (CI for Develop&Main) and `32735472921` (Mutation tests) after confirming the diff touches no workflow, script, `package.json` or lockfile path — all 7 files live under `packages/core/src/modules/sales/`. Mutation tests passed; `lint`, `ds-lint` and `audit-scope` passed; the rest is disclosed under CI status below. No local fallback gate was run, because usable CI signal is now on its way.

## ✅ What I verified

**Schema parity with create is exact.** `exchangeRateSchema` is `z.coerce.number().min(0)`, which is precisely what `decimal({ min: 0 })` builds at `packages/core/src/modules/sales/data/validators.ts:58-64`. There is no drift between what create accepts and what update now accepts.

**The union ordering is not cosmetic, and the comment explaining it is correct.** `z.coerce.number()` runs `Number(input)` first, and `Number(null)` is `0`, which satisfies `.min(0)`. With the operands reversed, `exchangeRate: null` would win the union as `0` and write a zero rate over a real one — a worse bug than the one being fixed. `z.null()` first is required, and `documents.update-order-only-fields.test.ts` pins it directly rather than leaving it to a comment.

**`null` genuinely clears rather than writing `"0"`.** `toNumericString` returns `null` for both `null` and `undefined` (`packages/core/src/modules/sales/commands/shared.ts:57-60`), so the `kind === "order"` branch at `packages/core/src/modules/sales/commands/documents.ts:1226-1228` writes SQL `NULL`.

**The derived text columns can never be left `undefined`.** This is the subtle one. `resolveDictionaryEntryValue` is typed `Promise<string | null>` and short-circuits on `if (!entryId) return null` (`packages/core/src/modules/sales/lib/dictionaries.ts:117-125`). Had it returned `undefined` for a cleared id, `entity.paymentStatus = undefined` would leave the column untouched on flush and the id would move without its text — the self-contradicting row the description is trying to avoid. It returns `null`, so `packages/core/src/modules/sales/commands/documents.ts:1299-1301` and `1325-1327` are correct.

**The `@ts-nocheck` gap is real, so I closed it by hand.** `documents.ts` carries `// @ts-nocheck` on line 1, so `yarn typecheck` says nothing about the new `(entity as SalesOrder)` assignments either way — the author flagged this, correctly. Checked against `packages/core/src/modules/sales/data/entities.ts:376-395`: `exchange_rate` is `numeric(18,8)` nullable typed `string | null`, both entry ids are `uuid` nullable, and both derived text columns are `text` nullable. Every assignment's right-hand side is `string | null`. It holds.

**The quote 400 fires before anything destructive.** Each rejection is thrown ahead of its own write, and `updateQuoteCommand` clears `acceptanceToken` / `sentAt` and reverts a `sent` quote to `draft` *after* `applyDocumentUpdate` returns (`packages/core/src/modules/sales/commands/documents.ts:5310-5324`), with a comment saying that ordering exists for exactly this reason. Since each of the three fields now satisfies the `.refine()` on its own, a silent skip really would have let a no-op payload destroy a customer's acceptance link — the three "leaves a sent quote intact" cases pin both halves. Rejecting rather than ignoring is the right call here, not just the tidier one.

**`buildDocumentUpdateChangeKeys` is correctly scoped and correctly scoped-down.** The additions sit inside the `kind === "order"` branch (`packages/core/src/modules/sales/commands/documents.ts:3815-3827`), and the function's only consumer is `buildChanges` inside `buildLog`. Nothing on the reindex or event path reads it, so widening it is audit-log-only, as claimed.

**The new 400 cannot break any caller in this repository.** Outside `commands/documents.ts` and its tests, `paymentStatusEntryId` / `fulfillmentStatusEntryId` appear only in `data/validators.ts` (create) and `data/entities.ts`; `exchangeRate` appears nowhere under `api/` or `backend/`. No quote form and no route handler spreads these onto a `sales.quotes.update` payload. The blast radius really is limited to external API clients that were previously getting a silent no-op — which is the intended fix, and is what `risk-medium` describes.

**i18n is complete.** Three keys in all five locales with genuine translations rather than English fallbacks, and the reused `sales.documents.detail.statusInvalid` exists in all five.

**Base drift is harmless.** The head is one commit behind `develop` (`ca2d1510b`, sales `setup.ts` + `lib/salesChannelsToggleSeed.ts`) — disjoint from `commands/documents.ts`, and the PR reports `MERGEABLE`.

**Not a hole:** this repo is on zod `4.4.3`, where `z.number()` rejects non-finite values by default, so `exchangeRate: Infinity` cannot reach `toNumericString` and become a `numeric` write error.

**Tests.** 418 lines and 38 cases, split between nine schema-level assertions and command-level runs against a mocked `EntityManager` in the same harness style as `documents.update-notes.test.ts`. The coverage is genuinely per-field and bidirectional: writes, derived-column movement, unresolvable-id 400 with nothing written, `null` clearing both columns, quote rejection with the sent quote surviving, and the audit-log key set. The two negative controls ("untouched when omitted", "still writes fields a quote does have") are what make the reverted-source run meaningful, and the reported 36/38 failure on revert is consistent with them.

## 💡 Nits (non-blocking)

**1. The two new lookups are uncached and serial — `packages/core/src/modules/sales/commands/documents.ts:1284` and `1312`.** Order *create* resolves the same two ids through `resolveCachedDictionaryEntryValue` inside a `Promise.all` (`documents.ts:5780-5781`), and the comment above that cache (`lib/dictionaries.ts:127-133`) describes precisely this workload: "the same few status / fulfillment / payment entry ids are re-resolved for EVERY order, so a bulk import re-reads them per order — a measured N+1". The scheduled system-of-record re-sync this PR exists to unbreak is that workload on the update side. Using the uncached helper matches the adjacent `statusEntryId` branch at `1257`, so this is consistent rather than wrong, and plumbing the cache in would mean changing `applyDocumentUpdate`'s signature — out of scope here. Worth a follow-up alongside #5344.

**2. Nothing validates which dictionary an entry id came from.** `resolveDictionaryEntryValue` resolves any `DictionaryEntry` in the tenant, so a `paymentStatusEntryId` pointing at a `sales.order_status` entry will happily write `"shipped"` into `payment_status`. This is not introduced here — order create (`documents.ts:5780-5781`) and the existing `statusEntryId` update branch have the same laxity — and #5344's "kind-aware rejection" direction is the right home for it. Noting it so it does not get lost.

**3. No route-level coverage for the new 400.** The command-level tests cover the branch that produces it, and the author documented the decision to skip `.ai/qa/tests/`, which is reasonable for a fix whose defect lives in the schema rather than the transport. The one thing only a route test would pin is that the `CrudHttpError(400)` actually surfaces as HTTP 400 through `api/documents/factory.ts:496-514` rather than being remapped. Optional.

## 🔁 Backward compatibility

No contract surface is broken. The schema change is purely additive on orders — three previously-stripped keys are now accepted, so no existing caller can be made worse off. On quotes, a payload carrying one of the three moves from a silent 200 no-op to a 400. That is a behaviour change to an API route, but it is the change being asked for: the previous response was a lie, and #5335 established the precedent for answering it with a kind-aware 400 rather than a silent drop. No DB structure change, no new migration, no import path, event id, DI key, ACL feature or generated file touched.

## 🚦 CI status

`prepare`, `typecheck`, `test` and `build` are still running at review time (see below). Per this repo's review contract the verdict is not held for them — branch protection and the QA gate hold the actual merge. I will post the CI result as a follow-up comment on this PR.

Pending at review time: **CI for Develop&Main** (`prepare` → `typecheck` / `test` / `build:app`).
Already green: `lint`, `ds-lint`, `audit-scope`, `Mutation tests`, `license/cla`.

## 🏁 Summary

Careful, narrow, well-argued, and the description's reasoning survives being checked against the code — including the two places where the obvious spelling would have been wrong (the union order, and writing the derived text column alongside the id). The scope discipline is right: it widens what update accepts and adds nothing else, leaving the remaining nine fields to #5344.

Approving. The QA gate still applies — this PR changes an API route's response for one payload shape, so it keeps `needs-qa` and cannot merge until a maintainer adds `qa-approved`.

pkarw:
## 🏷️ label rationale

| Label | Why |
|---|---|
| `merge-queue` | Approved with no blockers and no majors. The QA-approval gate still holds the actual merge — see `needs-qa` below. |
| `bug` | A silent data-loss defect: `sales.orders.update` reported success having written nothing, because `documentUpdateSchema` stripped the three keys before `applyDocumentUpdate` ran. |
| `needs-qa` | The automated-verification exemption does *not* apply here. It requires the API surface to be unchanged, and this PR changes it: a `sales.quotes.update` payload carrying `exchangeRate`, `paymentStatusEntryId` or `fulfillmentStatusEntryId` now returns 400 where it previously returned a silent 200. No `.tsx` outside tests and no DB structure change, but the route behaviour change is enough to keep manual QA. Matches how #5335 shipped. |
| `priority-medium` | Nothing is blocked on it today — the affected columns are latent for the reporting integration — but it silently corrupts an ongoing scheduled sync, so it should not sit indefinitely. |
| `risk-medium` | The widening on orders is additive and cannot make an existing caller worse off. The blast radius is the new 400 on quotes. I checked: outside `commands/documents.ts` and its tests, `paymentStatusEntryId` / `fulfillmentStatusEntryId` appear only in `data/validators.ts` and `data/entities.ts`, and `exchangeRate` appears nowhere under `api/` or `backend/` — so no in-repo UI or route caller sends these to `sales.quotes.update`. The exposure is external API clients that were previously getting a silent no-op. |
| `partner-request` | Retained as applied by the author. |

This matches the routing the author proposed in their metadata-intent comment, and matches the label set #5335 merged with. Not applying `qa` — that label means manual QA is in progress and belongs to the QA reviewer.

pkarw:

This PR is approved and requires manual QA (`needs-qa`, no `skip-qa`). It is queued in `merge-queue` but the QA-approval gate holds it until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` by swapping the labels (remove `merge-queue`, add `qa`), then run the checks below.

Note up front: this PR touches no UI. It changes a zod schema, its command handler and five locale files, so most of this is API work through `PUT /api/sales/orders` and `PUT /api/sales/quotes`. The admin screens are used only to confirm the write landed where a human can see it.

### P0 — Order update actually persists the three fields (money + status)

**Where to click**
- API: `PUT /api/sales/orders` (the sales orders CRUD route)
- UI to confirm: Sales → Orders → open the order (`/backend/sales/orders/<id>`)

**What to verify**
- `PUT` an existing order with `{ "id": "<orderId>", "exchangeRate": 4.25 }` → 200, then re-read the order (GET) and confirm `exchangeRate` is `4.25`, not the create-time value. On `develop` today this silently returns 200 and changes nothing — that is the bug.
- `PUT` `{ "id": "<orderId>", "paymentStatusEntryId": "<a sales.payment_status entry id>" }` → 200, then confirm **both** `paymentStatusEntryId` and the derived `paymentStatus` text moved together, and that the order detail screen shows the new payment status.
- Same for `{ "fulfillmentStatusEntryId": "<entry id>" }` → both `fulfillmentStatusEntryId` and `fulfillmentStatus` move.
- Send `{ "exchangeRate": null }` → the column is **cleared**, not set to `0`. This is the specific trap the union ordering exists for; a `0` here is a fail.
- Send `{ "paymentStatusEntryId": null }` → both the id and the `paymentStatus` text end up empty, not just the id.
- Send a payload whose *only* edit is one of the three (plus `id`) → 200, not a `update_payload_empty` validation error.

**What can go wrong**
- The id moves but the text column does not (or vice versa), leaving an order whose payment status reads one thing in the list and another on the detail screen.
- `exchangeRate: null` writes `0` instead of clearing.
- An unresolvable / deleted / other-tenant entry id is accepted silently instead of returning 400. Try a well-formed UUID that is not a dictionary entry in your tenant → expect **400**, and confirm the order is left completely untouched afterwards.
- **Tenant isolation:** grab a dictionary entry id belonging to a *different* tenant and send it → expect 400 and no write. It must not resolve.

### P0 — Quote update rejects the three fields instead of dropping them

**Where to click**
- API: `PUT /api/sales/quotes`
- UI to confirm: Sales → Quotes → open the quote (`/backend/sales/quotes/<id>`)

**What to verify**
- Put a quote into `sent` status so it has an acceptance link, and note its public acceptance URL / token.
- `PUT` that quote with `{ "id": "<quoteId>", "paymentStatusEntryId": "<entry id>" }` → expect **400** with "Payment status is not available on quotes."
- Then re-read the quote: it must still be `sent`, still hold its acceptance token, and the public acceptance link must **still work**. This is the important one — the rejection has to happen before the update path strips the token.
- Repeat for `exchangeRate` ("Exchange rate is not available on quotes.") and `fulfillmentStatusEntryId` ("Fulfillment status is not available on quotes."), including `exchangeRate: null`, which must also 400 — there is no column to clear.
- Confirm a *normal* quote edit (e.g. `customerReference`, or a line change) still works and is unaffected.

**What can go wrong**
- The quote returns 400 but the acceptance token / `sentAt` were already cleared and the quote reverted to `draft` — the customer's link is dead. That is the regression this ordering is meant to prevent.
- A previously-working integration that was sending one of these three to the quotes endpoint now starts failing. That is intended, but worth confirming it is a clean 400 with a readable message rather than a 500.

### P1 — Error messages are localised, and the audit log reflects the change

**Where to click**
- Switch the admin UI / `Accept-Language` to `pl`, `de`, `es`, `ko` and repeat one quote rejection
- Sales → Orders → the order's audit log / history panel

**What to verify**
- The 400 message comes back translated in each of the five locales, not as a raw key like `sales.quotes.payment_status_unsupported` and not as English fallback.
- After a successful order update, the audit-log entry lists the changed fields — `exchangeRate`, and for the status ids both the id **and** its derived text column.

**What can go wrong**
- A raw translation key leaking into an API error body.
- The audit entry showing the entry id changed but not the human-readable status, which makes the history unreadable.

### Pass/fail

pkarw:

No blockers, no majors, no minors — three non-blocking nits are in the review body. Labels applied: `merge-queue`, `bug`, `needs-qa`, `priority-medium`, `risk-medium` (rationale above). Autofix: skipped — this is another author's PR and `--autofix` was not passed; nothing on this branch was modified.

I approved the two workflow runs this head was waiting on (fork PRs need maintainer approval per push). `lint`, `ds-lint`, `audit-scope`, `prepare`, `documents-multi-instance` and `Mutation tests` are green; `test`, `ephemeral-integration` and the remaining jobs are still running. Swapping `in-progress` for `ci-monitoring` — I will post the CI result here when it settles, and this PR is free for anyone else to act on in the meantime.

The QA-approval gate still holds the merge: `needs-qa` without `qa-approved` never merges.

pkarw:

**All checks green** on `dc6c5b6ff`. The approve verdict stands unchanged.

`CI for Develop&Main` → **success**, and `Mutation tests` → **success**. Nothing was skipped in a way that hides a gate — I read the step lists rather than trusting the job-level conclusion, because a red early step can silently skip the ones behind it:

| Job | Result |
|---|---|
| `prepare` | ✅ Build packages → Prepare generated modules → Rebuild packages with generated files → Build app → Documents bundle budgets, all success |
| `test` | ✅ i18n sync, i18n usage, time-bomb literals, **agent instruction budget**, **Checking types**, **Test**, Test scripts, DS token snapshot drift, create-app template parity, repo-wide audit guards, all success |
| `lint` / `ds-lint` / `audit-scope` | ✅ success |
| `ephemeral-integration` | ✅ success |
| `Mutation tests` | ✅ success |
| `license/cla` | ✅ signed |

That covers this repo's full `validation.commands` gate — `build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `test`, `build:app` — so no local fallback run was needed.

One note for whoever merges: these runs only exist because a maintainer approved them. A cross-repository PR needs workflow approval **per push**, so if this branch is pushed to again, CI will go back to `action_required` and the checks will look absent rather than failing.

Removing `ci-monitoring`. This PR remains in `merge-queue` with `needs-qa` — the QA-approval gate holds the merge until a maintainer adds `qa-approved`. The manual-QA script is in the comment above.

Issue #5107, opened 2026-08-07
CRM Kanban: inconsistent filtering compared to list view


itrixjarek:
## Summary

The CRM Kanban view and the CRM list view use different filter logic/UI, which leads to inconsistent deal results between both views.

The won-deals filter should logically show deals in the Win column/stage on Kanban, matching the list view's interpretation of won deals.

## Current behavior

- Kanban and list views do not apply filters consistently.
- The filter UI/logic differs between the two views.
- Filtering won deals does not clearly map to deals in the Win stage/column on Kanban.

## Expected behavior

- Kanban and list views should use consistent deal filtering semantics.
- The won-deals filter should show deals in the Win stage/column on Kanban.
- Filter UI should be aligned enough that users can predict the same result set across Kanban and list views.

## Impact

Medium/high priority because inconsistent filtering makes pipeline views unreliable and can cause users to miss or misread won deals depending on the selected CRM view.

## Owner

@zielivia and @haxiorz are assigned as the responsible owners for verification/fix.

## Acceptance criteria

- Kanban and list views apply consistent filtering logic for deal status/stage filters.
- The won-deals filter shows deals in the Win stage/column on Kanban.
- Existing list filtering behavior remains correct.
- Existing Kanban grouping/drag-and-drop behavior remains unchanged except for the corrected filter semantics.
- Manual QA verifies the same won-deal scenario in both list and Kanban views.
- Add or update coverage if there is an existing test pattern for CRM deal filters or Kanban filtering.

haxiorz:
🤖 `autofix` started by @haxiorz at . Other auto-skills will skip this issue until the lock is released.

Issue #4667, opened 2026-07-30
Company KPI surfaces test deal status against a vocabulary no writer persists, so won/lost deals count as active


wojciechszyjka:
## Follow-up from #4629

PR #4629 fixed the `pipelineSummary` dashboard widget, which was counting won and lost deals as current pipeline. Establishing the correct denylist for that fix surfaced a **separate and larger correctness bug** in the customers module: several read-side surfaces test `customer_deals.status` against a vocabulary that no writer ever persists, so their predicates never match a closed deal.

### The mismatch

Every supported write path persists `win` or `loose`:

- `backend/customers/deals/[id]/hooks/useDealClosure.ts:68` writes `{ closureOutcome: 'won', status: 'win' }`
- `backend/customers/deals/[id]/hooks/useDealClosure.ts:88-93` writes `{ closureOutcome: 'lost', status: 'loose' }`
- `backend/customers/deals/pipeline/page.tsx:1760` — `updateDealStatus(dealId, status: 'win' | 'loose')`

But these read-side surfaces test for `'won'` / `'lost'` / `'closed'`:

- `api/companies/[id]/route.ts:1018-1020` — `deal.status !== 'won' && deal.status !== 'lost' && deal.status !== 'closed'` → `activeDeals`
- `api/companies/[id]/route.ts:1021` — `deal.status === 'won'` → `wonDeals`, which drives `ltvValue` and `completedDealsCount`
- `components/detail/dashboard/helpers.ts:6,14,71` — the same three-value test, three times
- `components/detail/CompanyKpiBar.tsx:22,30,54` — the same three-value test; `:84,157` — `d.status === 'won'`
- `components/detail/ActiveDealCard.tsx:28` — the same three-value test

### Impact

On a company detail page, every won and lost deal is still reported as **active**, and the won-deal count, completed-deal count and lifetime value read **0** — for a tenant that closes deals entirely through the supported UI. This is a silent wrong-number bug of the same family as #4621, on a different surface.

Note that `api/people/[id]/companies/enriched/route.ts:283` gets the vocabulary right (`status !== 'win' && status !== 'loose'`) but omits `closed`, so it is inconsistent in the other direction.

### Suggested fix

[...]

PR #5578, opened 2026-08-24
fix(customers): make `lost` the canonical deal status, not `loose`

cielecki:
## Summary

`customer_deals.status` carries `loose` as its canonical lost-deal spelling. It is a misspelling of `lost`, and no other surface shares it: `closure_outcome` stores `lost`, the AI tool `customers.update_deal_stage` writes `lost`, and every operator-facing label reads "Lost".

Operators see two consequences. The kanban names one stored status twice: `QuickDealDialog` resolves `customers.deals.kanban.quickDeal.status.loose`, translated as "stalled" in all five locales (`"Stalled"`, `"Stagniert"`, `"Estancado"`, `"정체됨"`, `"Wstrzymany"`), while `StatusFilterPopover` maps the same value to `filter.status.lost`, `"Lost"`. `PIPELINE_STAGE_DEFAULTS` also carries a separate `stalled` stage next to `loose`, so that label collides with a stage that already exists.

This PR makes `lost` canonical. #4667 already built `lib/dealStatus.ts` as the single source of truth and taught every reader both spellings, so the remaining work is flipping which one writers persist and migrating the stored data. 34 files change, and most are a one-word spelling edit: the substance is `lib/dealStatus.ts`, the migration and the spec.

I raised this first in the CRM Discord channel. Close it if you would rather keep `loose`.

## Changes

**Canonical spelling.** Writers use `DEAL_STATUS_LOST = 'lost'`. `canonicalDealStatus` now normalizes `loose` to `lost` instead of the reverse, and `LOST_DEAL_STATUS_LIST` keeps both spellings, canonical first.

**Writers.** `useDealClosure`, the kanban `updateDealStatus`, the `cli.ts` seeds and the seeded example deal now persist `lost`.

**Labels.** The i18n key follows the value, so `quickDeal.status.loose` becomes `quickDeal.status.lost`, and each locale now names it after its own `filter.status.lost` wording: `Lost` / `Verloren` / `Perdido` / `실패` / `Przegrany`. The Polish string is masculine, matching `Otwarty` and `Wygrany` in its own group rather than the feminine form the filter group uses. `DEAL_STATUS_DEFAULTS` and `PIPELINE_STAGE_DEFAULTS` seed `label: 'Lost'`.

**Data migration.** `Migration20260824180000_deal_status_lost` rewrites `customer_deals.status` and `pipeline_stage`, renames the `loose` dictionary entry for the `deal_status` and `pipeline_stage` kinds, and replaces the seeded `Loose` stage label. It deletes nothing.

## Backward compatibility

This change removes and narrows no contract surface, so it follows the deprecation protocol rather than claiming an exception.

- `DEAL_STATUS_LOSE` stays exported, marked `@deprecated` with a removal target of 0.9.0, so a downstream module importing it keeps compiling.
- `loose` stays a read alias in `LOST_DEAL_STATUS_LIST`, `expandDealStatusAliases`, `TERMINAL_PIPELINE_STAGE_LABELS` and the win/loss SQL in `lib/dealsSummaryQueries.ts`, which now matches `status IN ('lost', 'loose')`. A filter, a KPI count and a closure-outcome derivation behave identically whichever spelling a row carries.
- `data/validators.ts` is untouched. The deal `status` field is already `z.string().max(50)` (line 171), and the `closureOutcome` enum (`won` / `lost`, line 184) is unchanged.
- The migration skips a dictionary entry when the scope already holds a `lost` entry, because `customer_dictionary_entries_unique` covers (organization, tenant, kind, normalized value), and it corrects a label only when it is still the seeded `Loose`. A tenant that renamed the option keeps its own wording, and a skipped row still classifies correctly through the aliases.
- `down()` is a documented no-op, so this migration does not roll back. An instance may have held `lost` rows before it ran, and afterwards those are indistinguishable from the ones it renamed.

`UPGRADE_NOTES.md` carries the module-author migration under `0.7.0 → 0.7.1`.

## Specification

- [x] No (created a new spec)

**Spec file path:** `.ai/specs/2026-08-24-deal-status-lost-spelling.md`

I added it to the directory table in `.ai/specs/README.md`. It carries the required "Migration & Backward Compatibility" section. It also records the alternatives considered and the downstream action.

## Testing

`yarn jest --config jest.config.cjs --testPathPatterns 'modules/(customers|dashboards)'` in `packages/core`: 273 suites, 1901 tests, all passing.

`yarn tsc --noEmit` in `packages/core`: clean. Run after `yarn build:packages`, `yarn generate` and a second `yarn build:packages`, matching the CI order, because the generated entity ids are what `#generated/entities.ids.generated` resolves to.

The first run caught a real mistake. `StatusFilterPopover.test.tsx` still asserted the text `Loose` after I renamed the dictionary fixture to `Lost`. It is fixed in the same commit.

What moved in the tests:

- `lib/__tests__/dealStatus.test.ts`: the canonical-constant test now asserts `DEAL_STATUS_LOST`, the assertions on `canonicalDealStatus` run in the new direction, and a new case pins the deprecated `DEAL_STATUS_LOSE` export and confirms `isLostDealStatus` still accepts it.
- Writer-side assertions flipped to `lost`: `useDealClosure`, `dealCommandGuards`, `StatusFilterPopover` fixtures, the canonical row of the stage-transition table, `TC-CRM-082` and `TC-CRM-084`.
- Reader-side coverage deliberately keeps `loose` as the back-compat regression: the alias row of the stage-transition table, the closure-event tests, both existing-deal fixtures in `deals.stage-transitions`, the company KPI and enriched-route tests, and the dashboard helper test. Those are the cases that prove an unmigrated row still behaves identically.
- All five locale files parse and still hold 2862 keys each, so key parity is unchanged.

The migration has no automated coverage, which is the gap most worth a reviewer's attention. I also did not run the integration specs, which need a live instance.

## Checklist

- [ ] I added or updated integration tests in `.ai/qa/tests/`. (existing `__integration__` specs updated; no new scenario, and the migration is not exercised there)
- [x] I created or updated the spec in `.ai/specs/` with a changelog entry.
- [ ] Priority set. (I cannot set labels on this repo; please apply what fits)
- [ ] Risk set. (would be `risk-medium`: a data migration over deal status, stage values and tenant dictionaries)
- [ ] QA routing set. (`needs-qa` from my side: the kanban, the deal list and the win/loss KPIs are all manually exercisable, and the migration touches stored data)

### Design System Compliance

N/A. This PR touches no components, styling or tokens.

## Linked issues

None open. This builds directly on #4667 and #5107, which established the shared vocabulary this change makes canonical.


pkarw:
🤖 `om-auto-review-pr` claimed this PR for review at 2026-08-25T00:08Z. Will release the lock when the review is submitted.

pkarw (changes requested):
# 🔍 Code Review

`fix(customers): make `lost` the canonical deal status, not `loose`` — PR #5578, 34 files, +240/−70, `fix/deal-status-loose-labels` → `develop`.

**Verdict: CHANGES REQUESTED** — 0 blockers, 0 majors, 3 minors, 4 nits. Nothing here is broken; every minor is a coverage or documentation gap, and this repository requests changes on Medium/minor findings.

This is a careful piece of work. The alias-preserving design means the code and the data can be out of step in either order without a reader misclassifying a deal, and I checked that claim rather than accepting it — see 🧷 Considered and correct below for the five mechanisms I verified independently. What holds the merge is that **three of the four production behaviours this PR changes are not pinned by any test**, and I proved that by mutation rather than by reading: reverting the win/loss KPI SQL to its pre-PR form leaves all 1901 tests green.

---

## 🐞 Findings

### 🟡 Minor 1 — the win/loss KPI SQL change is invisible to the test suite, and its sibling clause is asserted three characters away

`packages/core/src/modules/customers/lib/dealsSummaryQueries.ts:149,151,171`

The PR widens the lost-deal predicate from `status = 'loose'` to `status IN ('lost', 'loose')` in three places. That widening is **load-bearing after this PR, not before it**: from here on `useDealClosure` and the kanban `Mark as Lost` both persist `lost`, so if that clause ever narrows back to a single spelling, every newly lost deal silently drops out of the quarter win/loss counters and the monthly trend series. Silent, because the query returns `0`, not an error.

I mutation-tested it. Reverting all three clauses to `status = 'loose'`:

```
Test Suites: 273 passed, 273 total
Tests:       1901 passed, 1901 total
```

Nothing fails. And the reason that is worth fixing rather than waving through is that the **won** half of the very same SQL statement *is* asserted, in the same test, on the same mocked call:

```ts
// lib/__tests__/dealsSummaryQueries.test.ts:86
expect(normalizedSql(executeMock.mock.calls[2])).toContain("status = 'win' OR closure_outcome = 'won'")
```

`calls[2]` is the `winLossRows` query, which contains both clauses. So the fix is one line immediately below line 86:

```ts
expect(normalizedSql(executeMock.mock.calls[2])).toContain("status IN ('lost', 'loose') OR closure_outcome = 'lost'")
```

That is the cheapest possible pin and it makes the asymmetry go away.

### 🟡 Minor 2 — the two surfaces the PR exists to change carry no regression coverage

**`QuickDealDialog`** (`backend/customers/deals/pipeline/components/QuickDealDialog.tsx:28`). The PR's stated motivation is that this dialog wrote `loose` and labelled it "Stalled" while the filter popover called the same value "Lost". The fix is `STATUS_OPTIONS = ['open', 'in_progress', 'win', 'lost']` plus the matching i18n key rename. Reverting that array to `'loose'` leaves the whole `modules/customers` suite green (246 suites / 1583 tests). There is already a harness next to it — `__tests__/QuickDealDialog.probability.test.tsx` — so a case asserting the dialog renders a Lost option whose submitted value is `lost` is a short addition, and it is the only thing that would stop the original bug reappearing.

**The migration** (`migrations/Migration20260824180000_deal_status_lost.ts`). You flagged this yourself as the gap most worth attention, and I agree with your assessment. I want to be fair about precedent: the same module already ships an uncovered data migration (`Migration20260519120000_pipeline_stage_color_tones.ts`), so "data migrations here are untested" is a real convention. But that one is a single unconditional `UPDATE ... CASE`; this one carries a correlated `NOT EXISTS` collision guard and a conditional label rewrite, which is exactly the kind of logic that gets "simplified" later by someone who does not know why the guard is there. `packages/core/src/modules/customer_accounts/migrations/__tests__/Migration20260724120000_customer_accounts.test.ts` is the in-repo pattern for this — it asserts the emitted SQL statically, no database required. Three assertions (the `NOT EXISTS` scope tuple, the `label = 'Loose'` conditional, and that `customer_deals` is updated without touching `updated_at`) would pin all of it.

By contrast, the seeded defaults **are** protected, though not in the way it first looks: reverting `DEAL_STATUS_DEFAULTS`/`PIPELINE_STAGE_DEFAULTS` to `loose` fails `__tests__/seedDictionaryScope.test.ts`. That guard pins the dictionaries and `CUSTOMER_EXAMPLES` to *agree with each other*, not to any particular spelling — revert both together and it goes green again. Worth knowing before treating it as a spelling pin.

### 🟡 Minor 3 — the deploy-order caveat runs one way only, and the risky direction is the one that is missing

`UPGRADE_NOTES.md` (0.7.0 → 0.7.1 entry) and `.ai/specs/2026-08-24-deal-status-lost-spelling.md:40-41,67-70`.

Both documents make the forward case well and it is correct: an instance running the new code before the migration keeps classifying `loose` rows correctly, because every reader accepts both spellings. The spec says so explicitly — "win/loss counters stay correct on an instance that has not run the migration".

The reverse is not stated and does not hold. Once the migration has run, rolling the **code** back to 0.7.0 leaves `lib/dealsSummaryQueries.ts` matching `status = 'loose'` against rows that now say `lost`, so the quarter win/loss KPI and the monthly trend report **zero lost deals** on an instance whose data is fine. `down()` is a documented no-op, so there is no automated way back either. Given the spec has a "Migration & Backward Compatibility" section that already reasons about the un-migrated instance, the migrated-data-with-old-code case belongs beside it — two sentences in `UPGRADE_NOTES.md` and one row in the spec's table, telling an operator to roll the migration's effect back by hand (or simply not to) rather than discovering it through a blank KPI.

---

## 🧹 Nits

1. **The won side of the same conditional was left on a literal.** `backend/customers/deals/page.tsx:859-863` now reads `} else if (row.original.status === 'win') { … } else if (isLostDealStatus(row.original.status)) {`. Replacing the lost literal with the helper is the right move, but the branch above it still misses `won`, which the AI tool `customers.update_deal_stage` writes verbatim (`ai-tools/deals-pack.ts:596-597`) — such a deal falls through both branches and renders a relative date instead of "Won". Pre-existing, and one word to fix while the file is open: `isWonDealStatus`, alongside the `isLostDealStatus` import you just added.

2. **The dashboards copy of the vocabulary is one edit from divergence.** `packages/core/src/modules/dashboards/widgets/dashboard/pipeline-summary/config.ts:17` hard-codes `CLOSED_DEAL_STATUSES = ['win', 'loose', 'won', 'lost', 'closed']` rather than importing `CLOSED_DEAL_STATUS_LIST`, and this PR had to hand-edit the comment above it. I am not asking for the import — a cross-module import into `customers/lib` is its own architectural question, and the duplication is deliberate and commented. But `__tests__/config.test.ts:55` asserts that array against a *literal*, so when `loose` is eventually dropped from `LOST_DEAL_STATUS_LIST` at 0.9.0, the dashboards copy will keep it and both suites will stay green. A single cross-check assertion against `CLOSED_DEAL_STATUS_LIST` closes that, and it belongs in this PR because this PR is what makes the 0.9.0 removal a scheduled event.

3. **Spec metadata.** `.ai/specs/2026-08-24-deal-status-lost-spelling.md:4` still reads `**Status:** Proposed` while shipping in the same PR as the implementation, so it lands on `develop` already stale. `.ai/specs/AGENTS.md`'s content checklist also asks non-trivial specs for a "Risks & Impact Review" (concrete failure scenario, severity, affected area, mitigation, residual risk) — for a change whose main risk is a data migration, that section would be the natural home for Minor 3.

4. **An order-coupled assertion.** `api/deals/__tests__/route.filters.test.ts:95` flipped from `toEqual({ $in: ['loose', 'lost'] })` to `toEqual({ $in: ['lost', 'loose'] })`, i.e. it pins the *ordering* of `LOST_DEAL_STATUS_LIST` rather than its membership. Not introduced by this PR — it was equally order-coupled before — but since you are touching the line, `expect.arrayContaining` (or sorting both sides, as the sibling assertion at `:107` already does) makes it test what it means.

---

## 🧷 Considered and correct — please do not "fix" these

Recording these so the next reviewer does not re-litigate them; each one I verified rather than assumed.

- **The migration deliberately does not bump `customer_deals.updated_at`, and it must not.** `loadDealsSummaryQueryRows` windows the win/loss KPI on `updated_at >= ? AND updated_at < ?` (`lib/dealsSummaryQueries.ts:148-151`) and buckets the trend series by `date_trunc('month', updated_at)`. Touching `updated_at` would slam every historical lost deal into the current quarter. The other three tables in the migration *do* set `updated_at = now()`, which is right because nothing reads them by time. The one residual cost is that a deal edit form held open across the deploy passes the optimistic-lock check and can write `loose` back — harmless, since readers accept it, and far cheaper than corrupting the KPI.
- **The collision guard matches the real constraint.** `customer_dictionary_entries_unique` is `unique (organization_id, tenant_id, kind, normalized_value)` (`migrations/Migration20251030150038.ts:11`) — exactly the tuple the `NOT EXISTS` subquery compares. Both scope columns are `uuid not null`, so there is no NULL-comparison trap, and the constraint guarantees at most one `loose` row per scope, so the same-statement snapshot semantics of the correlated subquery cannot misfire.
- **The dictionary kinds are the ones actually seeded.** `cli.ts:1194` and `cli.ts:1205` write `kind: 'deal_status'` and `kind: 'pipeline_stage'`; the migration filters on the same two. (Worth stating because the UI-facing names are `deal-statuses` / `pipeline-stages`, and matching on those would have made the dictionary rewrite a silent no-op.)
- **Renaming the stage row does not break closure-stage resolution.** `CustomerPipelineStage.label` is mapped to the `name` column (`data/entities.ts`), which is what the migration rewrites; `loadClosurePipelineStageSnapshot` matches `normalizePipelineStageLabel(candidate.label)` against `TERMINAL_PIPELINE_STAGE_LABELS.lost`, and `'Lost'` normalizes to `'lost'`, which is in the set (`lib/closureStage.ts:18,69`). The stage ordering in `PIPELINE_STAGE_DEFAULTS` also still lines up index-for-index with the eight-element array in `Migration20260218191730.ts:31`, which `seedCustomerExamples` relies on for its `stageValueToId` mapping.
- **`down()` as a documented no-op is this module's own convention**, not a shortcut — `Migration20260519120000_pipeline_stage_color_tones.ts` does exactly the same thing for the same reason.

I also swept for writers you might have missed: after this PR, no non-test source file outside `lib/dealStatus.ts`, `lib/closureStage.ts`, `components/detail/pipelineStageUtils.ts`, `commands/deals.ts:850` and the dashboards vocabulary list mentions `loose` at all, and every one of those is an intentional read alias.

---

## ✅ Validation

**Validation source: local fallback.** This is a fork PR and its two workflow runs for head `d1e9d586` are both sitting at `action_required` (`CI for Develop&Main` 32769783718, `Mutation tests` 32769783735) — GitHub has no usable check data for this head, so the checks-first rule does not apply and I ran the gate here. A maintainer still has to approve the workflow runs before merge; branch protection on `develop` reports no `required_status_checks`.

Runner: local (no compose `app` container). Fresh `yarn install` in an isolated worktree — no symlinked `node_modules`.

| Command | Result |
|---|---|
| `yarn build:packages` (×2, either side of `generate`) | ✅ |
| `yarn generate` | ✅ |
| `yarn i18n:check-sync` | ✅ — the five-locale key rename is consistent |
| `yarn i18n:check-usage` | ✅ (advisory; 3827 pre-existing unused keys, unchanged) |
| `yarn typecheck` | ✅ 26/26 |
| `yarn lint` | ✅ 0 errors |
| `yarn test` | ⚠️ **scoped**, not full: `packages/core` `modules/(customers\|dashboards)` → **273 suites / 1901 tests passed**, reproducing your reported numbers exactly |
| `yarn build:app` | ✅ compiled + TypeScript, 1m52s |
| Integration (`__integration__`) | ❌ not run — needs a live instance, same as your note |

The test run is scoped deliberately: nothing outside `packages/core/src/modules/customers` and `.../dashboards` imports `lib/dealStatus` or `lib/closureStage`, so those two module trees are the whole blast radius for the code change. It is *not* the whole blast radius for the migration, which no automated gate covers at all — that is Minor 2.

---

## 📋 Summary

The design decision is right and the deprecation protocol is followed properly — `DEAL_STATUS_LOSE` stays exported with a `@deprecated` tag and a 0.9.0 removal target, `loose` stays a read alias on every path, the validators are untouched, and `UPGRADE_NOTES.md` carries the module-author action including the non-obvious one (that `canonicalDealStatus` output flips). Raising it in the CRM channel first was the right call for a vocabulary change, and the spec records the alternatives honestly.

Three things to land before this merges:

1. One assertion line pinning the lost half of the win/loss SQL (`dealsSummaryQueries.test.ts:86`).
2. A regression case for `QuickDealDialog`'s option list, and a static SQL test for the migration.
3. The migrated-data-with-rolled-back-code caveat in `UPGRADE_NOTES.md` and the spec.

The nits are optional, though nit 1 is one word and nit 2 is the guard that makes the 0.9.0 `loose` removal safe.

**Labels:** `changes-requested`, `bug`, `needs-qa`, `priority-medium`, `risk-medium` — rationale in the label comment. Manual QA is required because the change reaches the kanban board, the quick-deal dialog, the status filter popover and the deals list, so the automated-verification exemption does not apply; the migration touching stored deal status is the second reason.

pkarw:
🏷️ label rationale — `om-auto-review-pr` (updated on re-review of `162f8a73`)

| Label | Why |
|---|---|
| `merge-queue` | Re-review approved. All three minors and all four nits from the 2026-08-25 review are landed and each was re-verified by mutation — reverting the covered production code now turns the new tests red in every case. Replaces `changes-requested`. |
| `bug` | Unchanged. `fix(customers)` — resolves an operator-visible contradiction where one stored status was named "Stalled" in the quick-deal dialog and "Lost" in the filter popover. |
| `needs-qa` | Unchanged and deliberately retained through the approval. The automated-verification exemption still does not apply: the change reaches `.tsx` surfaces (kanban, `QuickDealDialog`, `StatusFilterPopover`, deals list) and ships a data migration that rewrites stored deal status, stage values and tenant dictionary entries. The QA-approval gate holds the merge until `qa-approved` is added. |
| `priority-medium` | Unchanged. Operator impact is a mislabelled status, not a broken flow. |
| `risk-medium` | Unchanged. A forward-only data migration across `customer_deals`, `customer_dictionary_entries` and `customer_pipeline_stages` with a documented no-op `down()`. The new static SQL test on the migration pins the collision guard and the deliberate absence of an `updated_at` stamp, which tightens the mitigation without narrowing the blast radius — so it stays at medium rather than dropping to low. |

`qa-approved` deliberately **not** applied — it is earned by manual QA, never inferred from a diff. `in-progress` is this run's claim and is released in the completion comment below.

pkarw:
@cielecki handing this back to you — review submitted as **changes requested** (0 blockers, 0 majors, 3 minors, 4 nits). Nothing is broken; the three minors are the win/loss KPI SQL having no test, the migration and `QuickDealDialog` having none either, and the rollback direction missing from `UPGRADE_NOTES.md`/the spec. Details and the exact one-line fixes are in the review.

Two things worth knowing that are not your fault: this repository requests changes on Medium/minor findings (stricter than the usual bar), and your fork's CI runs for `d1e9d586` are both stuck at `action_required` — a maintainer has to approve the workflow runs. I ran the validation gate locally instead and it is green apart from the scoped test run, which is noted in the review.

Ping me when you push and I will re-review.

pkarw:

No `ci-monitoring`: there is nothing to watch — both workflow runs for this head sit at `action_required` pending maintainer approval of fork CI, so no further follow-up will come from this agent.

cielecki:
Thanks for the review, and for the mutation testing. All three minors and all four nits are in `162f8a73d`, rebased onto current `develop` (the `UPGRADE_NOTES.md` conflict was two entries landing at the same anchor; yours sit above mine now).

I verified each fix the way you found the gaps, by reverting the production code and watching the new test go red.

**Minor 1.** The test now asserts the lost predicate next to the won predicate, and the same pair on the trend series. Reverting all three clauses to `status = 'loose'` fails it.

**One correction to the review.** The statement carrying both predicates is `calls[3]`, not `calls[2]`. `calls[2]` is `wonRows`, the currency-totals query, which carries only the won clause, and that is why the existing assertion at that index reads as it does. Pasting the suggested line at `calls[2]` gave me a red test, so I traced the `connection.execute` order: `openRows`, `inflowRows`, `wonRows`, `winLossRows`, `seriesRows`, `overdueRows`, `openStuckRows`.

**Minor 2.** `QuickDealDialog.statusOptions.test.tsx` pins the option value and the i18n key separately, because they regress apart: the value is what reaches the database, the key is derived from it at render time. Reverting `STATUS_OPTIONS` to `loose` fails both cases. The migration test follows the `customer_accounts` pattern you pointed at and asserts the three you asked for: the `NOT EXISTS` scope tuple, the conditional label rewrite, and the update of `customer_deals` without an `updated_at` stamp. Dropping the collision guard fails it.

Your note on `seedDictionaryScope.test.ts` caught a misreading: I had it down as a spelling guard, and it pins agreement.

**Minor 3.** `UPGRADE_NOTES.md` and the spec now carry the reverse direction, with the manual reversal statement, since `down()` cannot guess. The spec moves to Accepted and gains the Risks & Impact Review section, where the rollback case is the first row.

**Nits.** All four are in. Nit 1 turned out to matter more than its size: a deal the AI tool wrote as `won` was falling through both branches and rendering a relative date, so `isWonDealStatus` fixes more than an asymmetry: it fixes a live wrong label.

Local run on the rebased head: 276 suites and 1917 tests, all passing, and `tsc --noEmit` clean after the full `build:packages` / `generate` / `build:packages` chain. Integration specs still not run, and CI has not run on this head: the workflow runs are still sitting at `action_required`.

pkarw:
🤖 `om-auto-review-pr` claimed this PR for re-review at 2026-08-27T00:34Z. Will release the lock when the review is submitted.

pkarw (approved):
# 🔍 Code Review

**Re-review** — `fix(customers): make `lost` the canonical deal status, not `loose`` — PR #5578, 38 files, +451/−71, `fix/deal-status-loose-labels` → `develop`, head `162f8a73`.

**Verdict: APPROVED** — 0 blockers, 0 majors, 0 minors, 3 nits. Every one of the three minors and four nits from the 2026-08-25 review is landed, and I re-verified each the same way I found it: by mutation, not by reading. All three new pins go red when the production code they cover is reverted. The remaining nits are documentation housekeeping and none of them blocks.

---

## ♻️ Previous findings — re-checked

| Previous finding | State | How I verified it |
|---|---|---|
| 🟡 Minor 1 — win/loss KPI SQL invisible to the suite | **Fixed** | Reverting all three predicates to `status = 'loose'` now fails `dealsSummaryQueries.test.ts` (1 failed / 3 passed). Before the fix, all 1901 tests stayed green. |
| 🟡 Minor 2a — `QuickDealDialog` option list unpinned | **Fixed** | Reverting `STATUS_OPTIONS` to `'loose'` now fails **both** new cases (2 failed / 7 passed) — the value case and the i18n-key case, independently. |
| 🟡 Minor 2b — migration unpinned | **Fixed** | Deleting the correlated `NOT EXISTS` block now fails `Migration20260824180000_deal_status_lost.test.ts` (1 failed / 3 passed). |
| 🟡 Minor 3 — rollback direction undocumented | **Fixed** | `UPGRADE_NOTES.md:69` and the spec's new Risks & Impact Review both carry the migrated-data-under-0.7.0-code case, with the manual reversal statement and its lossiness called out. |
| 🧹 Nit 1 — `isWonDealStatus` on the deals list | **Fixed** | `deals/page.tsx:859`. |
| 🧹 Nit 2 — dashboards vocabulary copy can diverge at 0.9.0 | **Fixed** | Removing `DEAL_STATUS_LOSE` from `LOST_DEAL_STATUS_LIST` now fails the new `stays in step with the customers module vocabulary it duplicates` case in `dashboards/.../config.test.ts` — which is exactly the 0.9.0 event it exists to catch. |
| 🧹 Nit 3 — spec metadata | **Fixed** | `Status: Accepted`, and the Risks & Impact Review section `.ai/specs/AGENTS.md` asks for is in, with the rollback case as its first row. |
| 🧹 Nit 4 — order-coupled filter assertion | **Fixed** | `route.filters.test.ts:95-96` now pins membership via `expect.arrayContaining` plus an explicit length of 2, so it tests what it means without going loose about the set size. |

### Your correction is right, and my line reference was wrong

`connection.execute` runs in this order in `loadDealsSummaryQueryRows`: `openRows` (86), `inflowRows` (99), `wonRows` (123), `winLossRows` (146), `seriesRows` (167), `overdueRows` (179), `openStuckRows` (194). `calls[2]` is `wonRows`, the currency-totals query carrying only the won half — pasting my suggested line there would indeed have gone red. `calls[3]` is `winLossRows`. Thank you for tracing it rather than pasting it; the assertion you wrote is better than the one I suggested, because it pins the won half at `calls[3]` too and adds `seriesRows` at `calls[4]`, so the monthly trend is covered as well as the quarter counters.

---

## 🧹 Nits

1. **The spec's Changelog stops at the initial entry.** `.ai/specs/2026-08-24-deal-status-lost-spelling.md:94-96` still reads only `**2026-08-24** — Initial specification.`, but the document materially changed on 2026-08-25: it moved `Proposed` → `Accepted`, gained the deploy-order sentence at line 67 and gained a whole Risks & Impact Review section. `.ai/specs/AGENTS.md:86` asks for "exact date and concise summary" per change, so one more line closes it.

2. **`Final Compliance Report` is the one checklist section still missing.** `.ai/specs/AGENTS.md:91` lists it beside Risks & Impact Review, and 154 of the 225 specs in `.ai/specs/` carry one — a real convention rather than a universal one, which is why this is a nit. If you read it as a post-implementation artifact rather than something a spec ships with, say so and I will not raise it again.

3. **The nit-1 fix ships unpinned, and that is proportionate.** `isWonDealStatus(row.original.status)` at `deals/page.tsx:859` is a real behavior fix — a deal the AI tool wrote as `won` previously fell through both branches and rendered a relative date — and nothing in the suite covers it. I am not asking for a test: there is no harness for that 900-line list page (the deals module's `.tsx` tests all sit on `[id]/`, `pipeline/components/` and the hooks), and standing one up for a one-word change is out of proportion to the change. `isWonDealStatus` itself is well covered in `lib/__tests__/dealStatus.test.ts`. Recording it so the gap is visible rather than implicit.

---

## 🧷 Considered and correct — please do not "fix" these

Additions to the list from the first review; each verified rather than assumed.

- **The migration test's `Object.create(prototype)` + `defineProperty('addSql')` harness is the right shape here.** It follows `customer_accounts/migrations/__tests__/Migration20260724120000_customer_accounts.test.ts`, needs no database, and asserts the emitted SQL statically. `toHaveLength(2)` on the `customer_deals` statements plus exact string equality is a stronger pin than a `toContain` sweep would be, and the `not.toContain('updated_at')` loop is the one assertion that protects the KPI windowing.
- **The cross-module import in `dashboards/.../config.test.ts` is test-only and does not create the coupling the architecture rules ban.** `config.ts` still hard-codes its own `CLOSED_DEAL_STATUSES` with the comment explaining why; only the test reaches into `@open-mercato/core/modules/customers/lib/dealStatus`. That is the cheapest way to catch the 0.9.0 divergence without turning a deliberate duplication into a runtime dependency, and it is the trade-off you documented in the comment above the case.
- **Mocking `useT` as `` `translated:${key}` `` rather than echoing the key is deliberate and necessary.** `translateWithFallback` treats `t(key) === key` as "no translation" and substitutes the fallback, so a key-echoing mock would have hidden which key the dialog resolved and the second case would have asserted nothing. The comment at the top of the file says exactly this.
- **The deals list reads `status` from the base table, not from a stale index doc.** I checked this specifically, because a data migration that rewrites `customer_deals.status` in raw SQL would leave `entity_indexes` behind, and the status column renders `dictionaryMaps['deal-statuses']?.[status]?.label ?? status` (`deals/page.tsx:767-768`) — a row whose indexed doc still said `loose` against a dictionary the migration had just renamed would render a raw `loose` badge. It does not happen: `query_index/lib/engine.ts:753` **left-joins** `entity_indexes` for custom-field columns and selects base columns from `customer_deals` itself, so the badge follows the migrated value. The skipped-scope case is safe for the opposite reason — a tenant the `NOT EXISTS` guard skips keeps its `loose` dictionary entry, so the lookup still resolves.

---

## ✅ Validation

**Validation source: local fallback.** GitHub has no usable check data for this head. Both workflow runs for `162f8a73` — `CI for Develop&Main` (32867482072) and `Mutation tests` (32867482070) — sit at `action_required`, the fork-PR approval gate, exactly as they did for `d1e9d586`. `gh pr checks` reports only `license/cla` (pass) and a skipping third-party check, so the checks-first rule does not apply and I ran the gate here.

Runner: local (no compose `app` container). Isolated worktree with its own `yarn install` — no symlinked `node_modules`.

| Command | Result |
|---|---|
| `yarn build:packages` (×2, either side of `generate`) | ✅ |
| `yarn generate` | ✅ |
| `yarn i18n:check-sync` | ✅ — the `quickDeal.status.loose` → `.lost` rename is consistent and correctly ordered across all five locales |
| `yarn i18n:check-usage` | ✅ (advisory) |
| `yarn typecheck` | ✅ |
| `yarn lint` | ✅ |
| `yarn test` | ⚠️ **scoped**, not full: `packages/core`, `modules/(customers\|dashboards)` → **276 suites / 1917 tests passed** in 38.7s, reproducing your reported numbers exactly |
| `yarn build:app` | ✅ 1m56s, not a cache hit |
| Integration (`__integration__`) | ❌ not run — needs a live instance, same as your note |

The scope is the same as last time and for the same reason: nothing outside `packages/core/src/modules/customers` and `.../dashboards` imports `lib/dealStatus` or `lib/closureStage`.

**Mutation checks (4/4 killed):**

| Mutation | Result |
|---|---|
| `status IN ('lost', 'loose')` → `status = 'loose'` (×3, `dealsSummaryQueries.ts`) | 🔴 1 failed / 3 passed |
| `STATUS_OPTIONS` `'lost'` → `'loose'` (`QuickDealDialog.tsx:28`) | 🔴 2 failed / 7 passed |
| Delete the `NOT EXISTS` guard (`Migration20260824180000_deal_status_lost.ts:38-44`) | 🔴 1 failed / 3 passed |
| Drop `DEAL_STATUS_LOSE` from `LOST_DEAL_STATUS_LIST` (the 0.9.0 event) | 🔴 1 failed / 25 passed |

---

## 🚦 CI and merge state

Disclosing this rather than waiting on it, so the PR is self-documenting.

- **CI has never run on this head.** Both workflow runs are at `action_required` and need a maintainer to approve fork CI. `develop` has no `required_status_checks` configured, so nothing mechanically blocks on them — the local gate above is this review's evidence, and it does **not** stand in for CI.
- **The branch is 59 commits behind `develop`** (merge base `15ffbe30`). Two files overlap with base changes since then — `UPGRADE_NOTES.md`, which you already rebased, and `deals/page.tsx`, where `develop`'s change is at line ~687 (a `t('ui.columnChooser.customFieldsGroup', …)` swap) and yours is at line 859, so they do not touch. GitHub reports `MERGEABLE`. Worth merging `develop` in before a maintainer approves CI, though: `ci.yml` diffs two-dot, so a branch this far behind pulls the full integration shard set and unrelated specs can go red for reasons that have nothing to do with this change.
- **The QA-approval gate is what actually holds the merge.** `needs-qa` stays on and `qa-approved` is not applied — it is earned by manual QA, never inferred from a diff. QA instructions are in a separate comment.

---

## 📋 Summary

This is the version to merge. The design was right in the first pass and the deprecation protocol was already followed properly; what held it was that three of the four production behaviours the PR changes were pinned by nothing, and all three are now pinned by tests that demonstrably fail without the production change. The two documentation nits are housekeeping on the spec, and the third is a note rather than a request.

Two things I want to credit explicitly, because they are the reason this took one round instead of three. You verified every fix by reverting the production code and watching the new test go red, which is the only evidence that distinguishes a test that covers something from a test that runs. And you traced my `calls[2]` reference to ground instead of pasting it — the assertion you wrote covers strictly more than the one I asked for.

**Labels:** `merge-queue`, `bug`, `needs-qa`, `priority-medium`, `risk-medium` — rationale in the label comment. Approved and queued; the QA-approval gate holds the actual merge until `qa-approved` is added, and a maintainer still needs to approve the fork's workflow runs.

pkarw:

**Read this first — the migration is what makes this PR worth manual QA.** `Migration20260824180000_deal_status_lost` rewrites stored data in three tables. The most useful QA is done on an instance that has **pre-migration deal rows carrying `loose`**, so seed or age some lost deals *before* applying it, then apply and re-check the same screens. An instance with no `loose` rows at all exercises almost none of the risk.

### P0 — The data migration over `customer_deals`, dictionaries and pipeline stages

**Where to click**
- Before migrating: `/backend/customers/deals` — note the exact list of deals showing a lost/"Loose" status, and the win/loss numbers on the KPI strip.
- Apply `yarn db:migrate`.
- Same route again, plus `/backend/config/customers/pipeline-stages` and the deal-status dictionary under settings.

**What to verify**
- Every deal that read as lost before the migration still reads as lost after it, with the **same** deal count — nothing silently reclassified as open.
- The status badge on `/backend/customers/deals` renders the word **"Lost"**, never the raw token `loose` or `lost` in lowercase. The badge falls back to the raw stored value when no dictionary entry matches it, so a lowercase token on screen means the dictionary rename and the row rewrite disagree.
- The deal-status and pipeline-stage dictionaries each show one **Lost** option and no **Loose** option.
- The `Lost` pipeline stage keeps its position in the stage order on the kanban board — it must not jump to the end.
- **A tenant that renamed the option keeps its own wording.** Rename the `Loose` dictionary label to something custom (e.g. `Gone cold`) *before* migrating; after migrating, that custom label must survive untouched while its underlying value becomes `lost`.
- **Deals do not move in time.** The migration deliberately does not stamp `updated_at` on `customer_deals`. On the deals list sorted by "Last updated", the historical lost deals must stay where they were — if a batch of old deals suddenly sorts to the top, the migration touched `updated_at` and the win/loss KPI is corrupted.

**What can go wrong**
- A status badge showing a literal lowercase `loose`/`lost` instead of the dictionary label.
- Deals disappearing from a "Lost" filter, or the lost count on the KPI strip dropping to 0 after migrating.
- Every historical lost deal appearing in the current quarter's win/loss numbers (the `updated_at` failure above).
- A tenant's renamed dictionary label being overwritten with `Lost`.
- Cross-tenant leakage on the dictionary rename: with two tenants/orgs, verify tenant B's dictionary entries are untouched when tenant A already held both spellings.

### P1 — Quick-deal dialog, kanban closure, deals list and the win/loss KPIs

**Where to click**
- `/backend/customers/deals` → kanban view → **+ / quick add** on a stage column, opening `QuickDealDialog`.
- The same board's card menu → **Mark as Lost**, and the deal detail closure flow at `/backend/customers/deals/{id}`.
- The status filter popover on `/backend/customers/deals`.
- The dashboard **Pipeline summary** widget and the deals KPI strip.

**What to verify**
- The quick-deal dialog's Status dropdown offers exactly **Open / In progress / Won / Lost**, in that order. The last option must read **"Lost"** — the bug this PR exists to fix is that it read **"Stalled"** while the filter popover called the same stored value "Lost". Check at least one non-English locale too (`de` → `Verloren`, `pl` → `Przegrany`, `es` → `Perdido`, `ko` → `실패`); a raw key like `customers.deals.kanban.quickDeal.status.lost` on screen means a locale file is out of step.
- Creating a deal as **Lost** from that dialog, then filtering the list by **Lost**, returns it.
- **Mark as Lost** from the kanban card, and closing a deal as lost from the detail page, both produce a deal the Lost filter finds and the win/loss KPI counts.
- The one filter selection **Lost** returns both newly created (`lost`) and pre-migration (`loose`) deals in the same result set — this is the read-alias behavior the whole design rests on.
- On the deals list, a deal closed as **won** shows the "Won" subtitle rather than a relative date. This is a fix in this PR: deals written by the AI tool `customers.update_deal_stage` (which persists `won`, not `win`) previously fell through to a relative date.
- The win/loss counters and the monthly trend on the pipeline widget move by the expected amount after you mark a deal lost — not by zero.

**What can go wrong**
- "Stalled" still appearing anywhere for the lost status, or the `Stalled` pipeline stage being confused with `Lost`.
- A newly created lost deal not appearing under the Lost filter (writer/reader spelling drift).
- The win/loss KPI or the trend chart reporting 0 lost deals while the list clearly shows some — the exact symptom of the SQL predicate narrowing to one spelling.
- Untranslated raw i18n keys in any of the five locales.

**Perceived performance (UI surfaces)**
- Cold-load `/backend/customers/deals` in both list and kanban view; capture a screenshot. Check the loading state appears before data rather than a blank frame, that the status badges do not flash the raw token before resolving to the dictionary label, and that the quick-deal dialog opens without a visible stall.
- Repeat the kanban board on a mobile viewport.

### P2 — Documentation

**Where to click**
- `UPGRADE_NOTES.md` (0.7.0 → 0.7.1) and `.ai/specs/2026-08-24-deal-status-lost-spelling.md`.

**What to verify**
- The rollback caveat is legible to an operator: after this migration runs, rolling the **code** back to 0.7.0 makes the win/loss KPI report zero lost deals with no error, and `down()` is a deliberate no-op.

**What can go wrong**
- An operator rolling back the release and reading a blank KPI as data loss.

### Pass/fail

One more thing for whoever merges: CI has **never run on this head**. Both workflow runs for `162f8a73` sit at `action_required` awaiting maintainer approval of fork CI. The local validation gate in the review is this agent's evidence only and does not substitute for it.

pkarw:

Re-review of `162f8a73`. All three minors and all four nits from the previous pass are landed; I re-verified each by reverting the covered production code and confirming the new tests go red (4/4 mutations killed). Remaining: 3 nits, none blocking.

No `ci-monitoring`: there is nothing to watch. Both workflow runs for this head (`CI for Develop&Main` 32867482072, `Mutation tests` 32867482070) sit at `action_required` pending maintainer approval of fork CI, so no further follow-up will come from this agent. The local validation gate reported in the review is this run's evidence only and does not substitute for CI.

The PR now sits in `merge-queue` with `needs-qa` — the QA-approval gate holds the merge until `qa-approved` is added. QA instructions are in the comment above.

PR #5586, opened 2026-08-24
fix(dev): fall back to Webpack when inotify limits are unavailable

dominikpalatynski:
## Problem

`yarn dev` in standalone apps created by `create-mercato-app` currently aborts before starting Next.js when Linux inotify limits are below the Open Mercato requirement and cannot be raised.

The direct failing condition is:

- `fs.inotify.max_user_watches=2097152`
- Open Mercato/Turbopack preflight requires `4194304`

Turbopack watches a large repository through Linux `inotify`. In serverless/containerized deployments such as Daytona sandboxes, the process runs inside a container but the inotify quotas belong to the shared host kernel. The container cannot modify `fs.inotify.max_user_watches` with `sysctl`, and `sudo` either is unavailable or has no permission to change the host. The existing dev runner then treats the failed repair as fatal and shuts down with exit code 1.

This is not an application database or Node.js error; it is a host-kernel capability limitation exposed by the dev preflight.

## Solution

- Keep Turbopack as the default when inotify limits are healthy or can be raised.
- When the preflight cannot raise the limits, automatically set `OM_DEV_BUNDLER=webpack` and continue startup.
- Pass the selected bundler to Next.js as `--webpack` or `--turbopack`.
- Support explicit overrides: `OM_DEV_BUNDLER=webpack yarn dev` and `OM_DEV_BUNDLER=turbopack yarn dev`.
- Keep the monorepo runner and the standalone `create-mercato-app` template in sync.

Webpack is slower than Turbopack but still provides the normal Next.js development workflow and does not depend on the unavailable host-level inotify change.

## Validation

- `node --test scripts/__tests__/dev-inotify-limits.test.mjs` — 10 tests passed.
- `node --check` passed for both dev runners and inotify helpers.
- TypeScript compilation and a compiled smoke check passed for the bundler selector.
- The full CLI Jest test could not run in this checkout because the Yarn/node_modules state is not installed.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5586"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790189911&installation_model_id=432243&pr_number=5586&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5586&signature=656a8770699e0ef89bb11f927b18bb217827ae836630b13b4632937407e239d8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pkarw:
🤖 `om-auto-review-pr` claimed this PR for automated review at 2026-08-25T00:07Z. Will release the `in-progress` lock when the review is posted.

pkarw (changes requested):
# 🔍 Code Review

**PR #5586 — `fix(dev): fall back to Webpack when inotify limits are unavailable`**
**Head:** `5d9423d` · **Base:** `develop` · **Mode:** first review · **Autofix:** skipped (not my PR — re-run with `--autofix` to fix it here)

**Validation source: GitHub checks.** Every reported check on `5d9423d` is green (`prepare`, `lint`, `ds-lint`, `test`, `documents-multi-instance`, `audit-scope`, `scope`, CodeQL ×3, `license/cla`); the rest are `skipping`. Nothing is pending and nothing failed, so no local re-run of `validation.commands` was needed. The head is `MERGEABLE` — no conflicts against `develop`.

## 📋 Summary

The change is well-scoped and the wiring is correct. I traced the new env var end to end and it genuinely reaches Next.js on both supported paths:

`scripts/dev.mjs` mutates `process.env.OM_DEV_BUNDLER` → `spawnCommand` spreads `...process.env` (`scripts/dev.mjs:423`) → `apps/mercato/scripts/dev.mjs` → `spawnMercato` spreads `...process.env` (line 620) → `mercato server dev` → `buildServerProcessEnvironment(process.env)` → `runtimeEnv` → `buildNextDevArgs(nextBin, runtimeEnv)` (`packages/cli/src/mercato.ts:2122`). The standalone template path is the same shape via `packages/create-app/template/scripts/dev-runtime.mjs:619`. Ordering is safe too: `ensureDevFileWatchLimits()` runs before any app spawn in `main()`, and `buildAppDevEnv()` is evaluated at spawn time, so the mutation is always observed.

I also verified the two things that would have made this a blocker, and neither is:

- `--webpack` is a real `next dev` flag on the pinned version. Next **16.3.0** (`apps/mercato/package.json:72`, `packages/create-app/template/package.json.template:106`) registers `.option('--webpack', 'Starts development mode using webpack.')` and still ships `dist/server/dev/hot-reloader-webpack.js`. The flag will not blow up argument parsing.
- No CLI version skew for scaffolded apps — the template pins `"@open-mercato/cli": "{{PACKAGE_VERSION}}"`, so the template script and the CLI that consumes `OM_DEV_BUNDLER` always release in lockstep.

The two `scripts/` files and their `packages/create-app/template/scripts/` mirrors are byte-identical (`diff` clean on both `dev.mjs` and `dev-inotify-limits.mjs`), so the template-sync contract holds. The new `scripts/__tests__` test is genuinely executed by CI — `.github/workflows/ci.yml:631` runs `yarn test:scripts`, which globs `scripts/__tests__/*.test.mjs`. `apps/mercato/next.config.ts:48`'s `turbopack: { root }` block is Turbopack-only, but it only sets the workspace root, which webpack dev does not need; there is no custom `webpack` config to conflict with.

What holds this up is not correctness — it is that the fallback throws away the actionable remediation the old hard-fail gave users, and the documented WSL2 setup flow was not updated to match.

## 🟡 Minor

**1. The auto-fallback discards the inotify remediation instructions, silently stranding WSL2 users on the slow bundler** — `scripts/dev.mjs:379-383` (and the identical `packages/create-app/template/scripts/dev.mjs:379-383`)

`ensureDevInotifyLimits()` returns a rich failure object on the unrepairable path: `manualCommands` (the exact `sysctl` invocations) and a formatted `message`. The existing hard-fail branch uses both, and additionally surfaces `failureCommand: 'yarn dev:fix-wsl-watchers'` through the splash (lines 386-397). The new `auto` branch discards all of it and prints two lines that never mention `yarn dev:fix-wsl-watchers` or the sysctl commands.

Concrete failure: a WSL2 developer on the default `fs.inotify.max_user_watches` runs `yarn dev` where the non-interactive `sudo` repair cannot succeed. Before this PR they got the full remediation text plus a one-command fix. After it, `yarn dev` "works", quietly runs the substantially slower webpack dev server on every future start, and the only hint is a warning that tells them to set `OM_DEV_BUNDLER=turbopack` — which, on a host whose limits are still low, immediately hard-fails with no explanation of why it worked a moment ago. The repo's own documented Windows path (`apps/docs/docs/installation/wsl2.mdx`) is exactly this population.

The fix is small: keep the fallback, but print `result.manualCommands` (or at minimum `yarn dev:fix-wsl-watchers`) in the same warning block, so the degradation stays loud and self-healing.

**2. A documented startup behavior changed without the docs changing** — `apps/docs/docs/installation/wsl2.mdx:406` and `:496`

Line 406 still reads "If startup stops with `OS file watch limit reached`, raise WSL's Linux inotify limits and start again", and the troubleshooting table at line 496 repeats it. With this PR, startup no longer stops on that path — it continues on webpack. Additionally, `OM_DEV_BUNDLER` is a new public dev-time contract (three accepted states: unset/`auto`, `webpack`, `turbopack`) and is documented nowhere in `apps/docs`, `README.md`, or a spec. A reader following the WSL2 guide today will be told to expect an abort that will not happen, and has no way to discover the override that the warning text tells them to use.

**3. The actual behavior change has no test coverage** — `scripts/dev.mjs:363-399`, `packages/cli/src/lib/__tests__/next-dev-bundler.test.ts`

The two added test files cover only the pure string resolvers. Every branch this PR actually introduces into the startup decision — the `webpack` early return that skips the preflight entirely, the `auto` → `process.env.OM_DEV_BUNDLER = 'webpack'` mutation, and the preserved hard-fail when `turbopack` is explicit and limits cannot be raised — is untested. The reason is structural: `ensureDevFileWatchLimits()` is a module-private function inside the 2000-line `dev.mjs`, so it is unreachable from `scripts/__tests__/`.

`ensureDevInotifyLimits` already demonstrates the pattern that fixes this — it lives in `dev-inotify-limits.mjs` with fully injectable dependencies and is thoroughly tested. Lifting the decision (not the console/splash side effects) into a pure `resolveDevBundlerDecision({ requestedBundler, inotifyResult })` in the same module would make all three branches assertable, and would also resolve finding 4 for free.

## 🔵 Nits

**4. One env-var contract, two divergent resolvers** — `scripts/dev-inotify-limits.mjs:27-32` vs `packages/cli/src/lib/next-dev-bundler.ts:3-7`

`resolveRequestedDevBundler` is tri-state (`'auto' | 'webpack' | 'turbopack'`); `resolveNextDevBundler` is bi-state and collapses everything that is not `'webpack'` into `'turbopack'`. They agree on the outcomes that matter today only because `dev.mjs` normalizes to a literal `'webpack'` before the CLI ever reads it. Both also silently swallow an invalid value — `OM_DEV_BUNDLER=webpak` yields Turbopack with no warning, which is an annoying thing to debug on the exact host where the user reached for the override.

Also worth noting: `resolveRequestedDevBundler` has nothing to do with inotify limits, so `dev-inotify-limits.mjs` is an odd home for it — though it is the pragmatic one given the module is already shared and tested.

**5. `buildNextDevArgs(...).bundler` is computed and never consumed** — `packages/cli/src/mercato.ts:2122`

Only `.args` is read; the `bundler` field is dead. As a consequence the splash line stays "Starting Next.js dev server" regardless of which bundler is running, so the supervisor's own logs carry no record of the fallback having engaged — the only evidence is the parent `dev.mjs` warning, which is easy to scroll past. Threading `nextDevCommand.bundler` into that message would make the degraded mode visible for the whole session and in the dev log.

## ✅ What is good here

- The env var propagation was clearly reasoned through rather than assumed — it survives three process hops and a supervisor restart (`startNextDev(runtimeEnv)` reuses the same env on the restart paths at `mercato.ts:2163`/`2186`).
- Explicit `OM_DEV_BUNDLER=turbopack` correctly preserves the old fatal behavior, so nobody who depends on the preflight as a guard loses it.
- Skipping `ensureDevInotifyLimits()` entirely when webpack is explicitly requested is right, not a shortcut — the function is a pure check plus a `sysctl` repair attempt with no other side effects, so there is nothing of value being skipped.
- The template mirror was updated in the same commit and is byte-identical, which is the part these dual-maintenance script changes usually get wrong.

## ⚖️ Verdict

**Changes requested.** Nothing here is incorrect, and the wiring is verified working end to end — but this repository requests changes on Medium/minor findings, and findings 1-3 are minor. In practice finding 1 is the one that matters: the fallback is the right idea, and it just needs to stay loud about the one-command fix instead of quietly accepting a permanently slower dev server. Findings 4 and 5 are nits you are free to skip.

pkarw:
🏷️ label rationale

_Updated after the round-2 re-review at head `77a5a7df3` — the round-1 table below it is superseded._

| Label | State | Why |
|---|---|---|
| `merge-queue` | **added** (replaces `review`) | Round-2 re-review submitted `approved` — all three round-1 minors are fixed and only nits remain, so the PR moves to the merge-ready pipeline state. |
| `review` | removed | Superseded by the pipeline transition above. |
| `bug` | kept | Still the correct category — this fixes a fatal `yarn dev` abort on hosts that cannot raise inotify limits. |
| `skip-qa` | kept | The automated-verification exemption still holds and I re-derived it against the round-2 diff: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no database-structure or API-surface change, no `BACKWARD_COMPATIBILITY.md` surface moved (`packages/cli/src/lib/next-dev-bundler.ts` is a new additive module), and the changed behavior ships with tests in the same PR — `scripts/__tests__/dev-inotify-limits.test.mjs` now covers all three bundler-decision branches. |
| `priority-medium` | kept | Unblocks development on constrained Linux hosts and sandboxes; not urgent for the released product. |
| `risk-medium` | kept | Dev-time tooling only — no runtime, data or API surface — but it changes the default startup path for every developer on Linux, and it is dual-maintained across the monorepo and the `create-app` template. |
| `in-progress` | removed | Automated re-review complete; the lock is released and the PR is free for anyone to act on. |

No `needs-qa`, so no manual-QA instructions comment applies. `qa-approved` was neither present nor added — nothing here was earned by manual QA.

pkarw:
👤 @dominikpalatynski — handing this back to you.

The implementation is correct and I verified the `OM_DEV_BUNDLER` plumbing end to end (three process hops on both the monorepo and `create-app` paths), that `--webpack` is a real flag on the pinned Next 16.3.0, and that the template mirror is byte-identical. CI is fully green and the head is conflict-free.

What I would like changed before approval — the full reasoning is in the review:

1. **`scripts/dev.mjs:379-383`** (and its template mirror) — the `auto` fallback drops `result.manualCommands` / `result.message`, so WSL2 users lose the `yarn dev:fix-wsl-watchers` one-command fix the hard-fail branch gives them and end up permanently on the slow bundler with no hint. Keep the fallback, just keep the remediation text loud.
2. **`apps/docs/docs/installation/wsl2.mdx:406` and `:496`** — still document a startup abort that no longer happens, and `OM_DEV_BUNDLER` is undocumented anywhere.
3. **Test coverage** — the three new branches of `ensureDevFileWatchLimits()` are untested because the function is private to `dev.mjs`. Lifting the decision into a pure helper in `dev-inotify-limits.mjs` (the way `ensureDevInotifyLimits` already is) makes them assertable and also collapses the duplicate resolver noted in nit 4.

The two nits (duplicate bundler resolvers, unused `buildNextDevArgs(...).bundler`) are optional — feel free to skip them or fold them into the refactor for #3.

Push a follow-up and I will re-review; no need to reassign me.

pkarw:

Validation source: GitHub checks (all green on `5d9423d`, nothing pending — no local gate re-run needed). Autofix: skipped (not my PR — re-run with `--autofix` to fix it here). No CI follow-up scheduled: checks are already settled.

dominikpalatynski:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. I will address the review findings, validate the follow-up, and hand the PR back for re-review without merging it.

dominikpalatynski:
Follow-up pushed as `77a5a7df3`.

Addressed the review findings:
- auto fallback now keeps the current inotify values, `yarn dev:fix-wsl-watchers`, and the exact manual sysctl commands visible;
- extracted and tested `resolveDevBundlerDecision()` for explicit Webpack, healthy auto/Turbopack, auto fallback, and strict Turbopack failure paths;
- documented `OM_DEV_BUNDLER` and the sandbox/container behavior;
- the startup splash now identifies the selected bundler;
- root and `create-app` template dev scripts remain byte-identical.

Validation: `yarn test:scripts` — 685 passed, 1 skipped because Chromium is unavailable; `yarn build:packages` — 27/27 successful; CLI bundler test 3/3; CLI typecheck successful; `git diff --check` successful.

The PR is handed back for re-review; explicit `OM_DEV_BUNDLER=turbopack` still preserves the strict failure behavior when limits cannot be raised.

pkarw:
🤖 `om-auto-review-pr` claimed this PR for automated re-review at 2026-09-08T10:38Z (round 2, head `77a5a7df3`). Will release the `in-progress` lock when the review is posted.

pkarw (approved):
# 🔍 Code Review

**PR #5586 — `fix(dev): fall back to Webpack when inotify limits are unavailable`**
**Head:** `77a5a7df3` · **Base:** `develop` · **Mode:** re-review (round 2 of my 2026-08-25 review) · **Autofix:** skipped (not my PR — re-run with `--autofix` to fix it here)

**Validation source: GitHub checks.** All three workflow runs on the exact head SHA `77a5a7df3d415cb0c6ad773081bc4be72839396b` are `success` — `CI for Develop&Main` (34161779893), `Mutation tests` (34161779900) and `PR #5586` (34161777742) — with `prepare`, `lint`, `ds-lint`, `test` (6m8s), `documents-multi-instance`, `audit-scope`, `scope`, CodeQL ×3 and `license/cla` all green, nothing pending and nothing failed. I queried by `head_sha` rather than trusting `gh pr checks` alone, so these are this head's runs and not an inherited green. No local re-run of `validation.commands` was needed; the few commands I did run below are *review experiments*, not the gate.

The head is `MERGEABLE` and I confirmed it independently — `git merge-tree` against the current `origin/develop` produces no conflict, and the `develop` merge at `5011413` left no duplicated import or stray marker in either changed source file (`node --check` clean on `scripts/dev.mjs`, `scripts/dev-inotify-limits.mjs` and the template mirror).

## 📋 Summary

**All three minor findings from round 1 are genuinely fixed, and I verified each against the code rather than against the follow-up comment.** One of the two optional nits was fixed as well. What is left is nits only, so this is an approve.

**Finding 1 — the fallback now stays loud.** `scripts/dev.mjs:386-393` keeps the fallback but prints `result.current`, an explicit `yarn dev:fix-wsl-watchers` pointer, and every entry of `result.manualCommands` — the exact `sudo sysctl -w …`, the `tee /etc/sysctl.d/99-open-mercato-inotify.conf` line and `sudo sysctl --system`. The WSL2 developer who prompted this finding now gets the one-command fix back. I also checked the thing that would have made the fix cosmetic: the compact dev splash is an **HTTP** surface (`renderSplashHtml`, `startSplashServer` at `scripts/dev.mjs:1029`/`:1053`), not a terminal takeover, and `ensureDevFileWatchLimits()` runs at `main()` line 1982 with nothing redrawing over it — so the warning is written to the terminal and stays there. The removed line ("Set `OM_DEV_BUNDLER=turbopack` after raising the limits") was worth removing on its own terms: once the limits are raised, `auto` selects Turbopack by itself, so the old text asked for a step that is not needed.

**Finding 2 — the docs now match the behavior.** `apps/docs/docs/installation/wsl2.mdx:406` no longer promises a startup abort that cannot happen ("On WSL2, raising the Linux inotify limits enables the preferred Turbopack dev server"), `:415` documents the container/sandbox auto-fallback and both explicit overrides, and the troubleshooting row at `:498` was rewritten to name both paths. `apps/docs/docs/installation/standalone.mdx:124` covers the scaffolded-app audience, which is the population the PR's own problem statement describes. I looked for a canonical env-var registry page that should also carry `OM_DEV_BUNDLER` and there is none — `apps/docs` documents runtime vars in the guide where they apply, and the sibling dev-only var `OM_DEV_INOTIFY_CHECK` is documented nowhere at all — so two in-context guides is the repo's convention, not a shortcut.

**Finding 3 — the decision is now a tested pure function, and the tests are load-bearing.** `resolveDevBundlerDecision()` (`scripts/dev-inotify-limits.mjs:35-51`) lifts exactly the decision and none of the console/splash side effects into the already-injectable module, which is the shape I asked for. It genuinely drives both branches in `dev.mjs` — `!requestedDecision.shouldCheckInotify` at `:370` is the preflight skip, `decision.fallback` at `:385` is the auto-fallback — and I traced it as a **behavior-preserving** refactor: given the branch is only reached when `result.ok === false`, `decision.fallback` reduces to the old `requestedBundler === 'auto'`, and `shouldCheckInotify` reduces to the old `requestedBundler === 'webpack'`. The four new cases at `scripts/__tests__/dev-inotify-limits.test.mjs:93-131` cover explicit Webpack, healthy auto, the auto fallback and the strict-Turbopack failure. **I did not take the coverage on trust** — I ran the suite (14/14 pass) and then mutated the helper three ways: inverting the health check (`ok !== false` → `ok === false`) kills 2 tests, retargeting the fallback branch (`requested === 'auto'` → `'turbopack'`) kills 2, and flipping the explicit-Webpack `shouldCheckInotify` to `true` kills 1. Every mutant is caught, so these tests pin the decision rather than merely executing it.

**Nit 5 fixed, with a bonus.** `packages/cli/src/mercato.ts:2144-2150` now consumes `buildNextDevArgs(...).bundler` and threads it into the splash line as `Starting Next.js dev server (Webpack)` / `(Turbopack)`. Because `writeDevSplashRuntimeStarting` writes into the splash **child state** (`mercato.ts:428`), that label reaches the splash page at :4000 as well as the dev log, so the degraded mode is visible for the whole session and not only in a warning printed once at startup. Moving the `buildNextDevArgs` call above the splash write is inert — both statements sit in the same synchronous promise executor.

Everything I verified in round 1 still holds after the `develop` merge: `next` is still pinned to **16.3.0** in both `apps/mercato/package.json:77` and `packages/create-app/template/package.json.template:112`, so `--webpack` is still a real flag; `scripts/dev.mjs` and `scripts/dev-inotify-limits.mjs` are still **byte-identical** to their `packages/create-app/template/scripts/` mirrors (`diff` clean on both); and the template's `dev` script (`package.json.template:12`) really does run the mirrored file, with `dev:fix-wsl-watchers` and `scripts/fix-wsl-inotify.mjs` both present in the template, so the remediation the warning advertises exists in a scaffolded app too. I also confirmed there is no second preflight to fight with — `grep` for inotify handling in `apps/mercato/scripts/dev.mjs` and `packages/create-app/template/scripts/dev-runtime.mjs` returns nothing, so the root runner is the only gate and the fallback cannot be re-vetoed downstream.

## 🔵 Nits

None of these need to change before merge; take them or leave them.

**1. `resolveDevBundlerDecision` reports a `bundler` it has not earned** — `scripts/dev-inotify-limits.mjs:42`

At the first call site (`scripts/dev.mjs:369`) the helper is invoked with no `inotifyResult`, so `inotifyResult?.ok !== false` is true and it returns `{ bundler: 'turbopack', fallback: false, shouldCheckInotify: true }` — a Turbopack verdict for a host whose limits have not been read yet, and which may be about to fall back. It is inert today because that call site reads only `shouldCheckInotify`, and none of the four new tests pins the no-`inotifyResult` shape, so nothing would notice if a future caller started reading `.bundler` from it. Either return `bundler: null` until a result is supplied, or add a one-line comment that the field is meaningless without `inotifyResult`.

**2. Three silent normalizations of one env var, still divergent** — `scripts/dev-inotify-limits.mjs:29-36` vs `packages/cli/src/lib/next-dev-bundler.ts:3-8`

This is round 1's nit 4, unchanged, plus one more layer: `resolveRequestedDevBundler` collapses an invalid value to `'auto'`, `resolveNextDevBundler` collapses everything non-`webpack` to `'turbopack'`, and `resolveDevBundlerDecision` now adds a third fallback via `DEV_BUNDLER_MODES.includes(...)`. That third one is unreachable from the only production call site, since `resolveRequestedDevBundler` has already normalized. `OM_DEV_BUNDLER=webpak` still yields Turbopack with no warning — annoying to debug on the exact host where someone reached for the override. A single shared resolver that warns on an unrecognized value would collapse all three; you were free to skip this last round and are free to skip it again.

**3. The fallback warning shows current values but not required ones** — `scripts/dev.mjs:388`

`JSON.stringify(result.current)` prints the three numbers as a raw object. `result.message` already formats them as `fs.inotify.max_user_watches: 2097152 < 4194304`, which is the more useful shape and is what the hard-fail branch shows. The targets are recoverable from the `sudo sysctl -w …` line printed two lines down, so nothing is missing — it just reads better with the comparison.

**4. Pre-existing, noted so it is not rediscovered as a regression** — `scripts/dev-inotify-limits.mjs:101-113`

If `/proc/sys/fs/inotify/*` is unreadable rather than merely low, `readCurrentInotifyLimits` records the read errors and returns no numbers, `findInotifyLimitIssues` skips every non-numeric entry, and `ensureDevInotifyLimits` returns `{ ok: true }` at `:257`. Such a container gets Turbopack and the runtime "OS file watch limit reached" panic, never the new fallback. This behavior is **unchanged by this PR** and I am explicitly not holding it against you — the PR's stated target (a Daytona sandbox reporting `max_user_watches=2097152`) reads those files fine. It is only worth recording because the fallback's premise is "unrepairable limits are detected", and this is the one shape where they are not.

## ✅ What is good here

- The three findings were addressed by doing the work rather than by patching the symptom: the decision really was lifted into the injectable module, not wrapped in a test-only export, and the docs were corrected in both places the old claim appeared plus a third the fallback newly affects.
- The refactor is behavior-preserving under substitution, which is the property that makes a "make it testable" round safe to accept — I checked both reduced branches rather than assuming it.
- The new tests survive mutation. That is the difference between coverage and a guard, and it is the reason finding 3 closes rather than reopening in a different form.
- Threading the bundler into the splash state (nit 5) solved more than the nit asked for — the degraded mode is now visible on the splash page for the whole session, not just in one scrollable warning line.
- The template mirror stayed byte-identical across a second round of edits, which is the part dual-maintenance script changes usually break on the follow-up commit rather than the first one.

## ⚖️ Verdict

**Approved.** Every minor from round 1 is closed and verified, CI is fully green on this exact head, the head merges cleanly against the current `develop`, and only nits remain — none of which change behavior. Under this repository's rule (Medium/minor requests changes, Low/nit approves) this is an approve, and I would be comfortable with it landing as-is.

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED (re-review, round 2). Lock released.

Validation source: GitHub checks — all three workflow runs on head `77a5a7df3d415cb0c6ad773081bc4be72839396b` are `success` (`CI for Develop&Main` 34161779893, `Mutation tests` 34161779900, `PR #5586` 34161777742), verified by `head_sha` rather than by `gh pr checks` alone. Nothing pending, nothing failing, so no local re-run of the validation gate was needed; the `node --test` / `node --check` / mirror-`diff` / mutation runs cited in the review are review experiments, not the gate.

All three round-1 minors are closed and `buildNextDevArgs(...).bundler` (nit 5) is now consumed. Four nits remain, none behavioral and none blocking. `reviewDecision` is now `APPROVED` and `mergeStateStatus` is `CLEAN`; `skip-qa` means no QA gate applies, so this is merge-ready.

Autofix: skipped (not my PR — re-run with `--autofix` to fix it here). No CI follow-up scheduled: checks are already settled.

PR #5590, opened 2026-08-25
fix(ai-assistant): avoid bundling standalone dynamic loader

pkarw:
## Status

Ready for review.

## Problem

Next.js/Turbopack followed the AI assistant standalone-registry fallback into `@open-mercato/shared/lib/bootstrap/dynamicLoader`. That Node-only loader performs computed runtime imports, so Turbopack repeatedly emitted `Module not found: Can't resolve <dynamic>` while compiling AI assistant routes.

## Fix

- mark both runtime-only `import()` expressions in the generated-registry loader with Webpack and Turbopack ignore directives;
- keep the existing shared loader implementation and public import path unchanged;
- add a regression test that pins both directives inside the exact import expressions.

## Validation

- focused generated-registry loader suite: 21/21 passed;
- `yarn workspace @open-mercato/ai-assistant build`: passed;
- forced Turbopack `yarn build:app`: passed with no `Can't resolve <dynamic>` warning;
- full configured local validation gate: passed;
- `yarn template:sync`: passed.

The build still reports six pre-existing dynamic-filesystem tracing warnings in the example, AI registry filesystem lookup, and queue code. They are unrelated to the `<dynamic>` module-resolution warning fixed here.

## Compatibility

No API, generated-registry shape, exported symbol, import path, database schema, or runtime response contract changes.


pkarw:
🤖 `om-auto-review-pr` — claiming this PR for review. Assignee, `in-progress` label and this comment are the claim signals; the lock is released when the review is posted.

pkarw (commented):
# 🔍 Code Review: fix(ai-assistant): avoid bundling standalone dynamic loader

## 🎯 Summary

This PR sets out to stop Next.js/Turbopack from reporting `Module not found: Can't resolve <dynamic>` when it compiles the AI-assistant routes. It does that by marking the one `import()` in `generated-registry-loader.ts` that pulls `@open-mercato/shared/lib/bootstrap/dynamicLoader` into the application graph with `/* webpackIgnore: true */` and `/* turbopackIgnore: true */`, so the bundler leaves the specifier alone and Node resolves it natively at runtime. It also adds a Jest guard and a tracking plan under `.ai/runs/`.

The diagnosis behind the change is right and the approach is the idiomatic one for this problem. `dynamicLoader.ts:480` performs `await import(fileUrl)` on a computed URL, which Turbopack cannot statically resolve, and `generated-registry-loader.ts` is the only edge that drags that module into the app graph — `@open-mercato/shared/lib/bootstrap/index.ts:21` deliberately does not re-export it, and every other importer (`packages/cli`, `ai_assistant/cli.ts`, the integration specs) is a standalone Node entry point. The runtime side is safe too: `packages/shared/package.json` declares an explicit `"./lib/bootstrap/dynamicLoader"` export mapping to `./dist/lib/bootstrap/dynamicLoader.js`, the repo uses the `node-modules` linker (`.yarnrc.yml`), and `apps/mercato/next.config.ts` does not use `output: 'standalone'`, so nothing depends on file tracing following the now-untraced import. Worth noting for the record: only `turbopackIgnore` is load-bearing on this repo's default build path — the Next 16.3 Turbopack binary recognises `turbopackIgnore` and has no knowledge of `webpackIgnore` — but keeping the webpack comment for `next build --webpack` is reasonable defence in depth.

What blocks the change is that it does not finish the job. The identical unresolvable-dynamic-import pattern survives in the very file being edited, so the warning the PR exists to remove is still emitted, and the shipped regression test cannot tell the difference. Reviewed scope: the three changed files plus the `@open-mercato/shared` bootstrap boundary, `apps/mercato/next.config.ts`, and the importers of `generated-registry-loader`.

## Verdict

❌ **request changes** — `generated-registry-loader.ts:124` still contains a fully computed `await import(...)` inside a module that remains in the Next.js bundle, which reproduces the exact `Module not found: Can't resolve <dynamic>` warning this PR is meant to suppress (Major). The accompanying test asserts on source text rather than on the bundler-visible invariant, so it stays green in every failure mode described below (Major). The committed plan additionally records work that the head commit reverted, and the PR body still describes the change as unfinished (Minor).

## 🧪 Validation Gate

Validation source: **GitHub checks** — the repo-local review extension prefers CI results over local re-runs. The required `CI for Develop&Main` run for head `e4a08d12` ([run 32802699902](https://github.com/open-mercato/open-mercato/actions/runs/32802699902)) is still **pending**, so no local substitute was run for the gate commands; branch protection and the merge queue hold the actual merge.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ⏳ PENDING | Covered by `CI for Develop&Main`, still running for `e4a08d12`. |
| `yarn generate` | ⏳ PENDING | Same run. |
| `yarn build:packages` | ⏳ PENDING | Same run. |
| `yarn i18n:check-sync` | ⏳ PENDING | Same run. No locale files changed by this diff. |
| `yarn i18n:check-usage` | ⏳ PENDING | Same run. No user-facing strings changed by this diff. |
| `yarn typecheck` | ⏳ PENDING | Same run. |
| `yarn test` | ⏳ PENDING | Same run. |
| `yarn build:app` | ⏳ PENDING | Same run — and this is the gate that would actually surface the remaining `<dynamic>` warning. |

Checks already settled on this head are green: CodeQL (`Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)`), `scope`, `license/cla`, and the `Mutation tests` workflow.

⏳ **CI pending:** the required `CI for Develop&Main` check had not reported when this review was submitted. The review was posted on the code as pushed rather than waiting; the pending check still gates the merge independently of this verdict.

## Findings

### ⚠️ Major

**`packages/ai-assistant/src/modules/ai_assistant/lib/generated-registry-loader.ts:124` — the warning this PR targets is still emitted from the same file.**

`compileAndImportGenerated` ends with `return (await import(pathToFileURL(jsPath).href)) as Record<string, unknown>`. That is a fully computed specifier with no ignore comment, in a module that is unambiguously inside the Next.js graph: `agent-registry.ts` and `tool-loader.ts` import it, and those are imported by route handlers such as `api/ai/agents/route.ts`, `api/tools/route.ts` and `api/settings/route.ts`. Turbopack renders a request it cannot statically resolve using its dynamic-pattern markers — the `<dynamic>` / `<dynamic no slash>` strings are present in the Next 16.3 `next-swc` binary — which is precisely the `Can't resolve <dynamic>` text quoted in the PR's goal.

This line is structurally identical to `packages/shared/src/lib/bootstrap/dynamicLoader.ts:480` (`await import(fileUrl)`), which is the site the PR successfully removed from the graph. Either both sites warn, in which case the fix is incomplete, or neither does, in which case the PR is patching the wrong edge. There is no third reading, and no build evidence in the PR distinguishes them.

Failure scenario: run `yarn dev` (or `yarn build:app`) and compile any AI-assistant route — for example `GET /api/ai/agents`. Turbopack parses `generated-registry-loader.ts`, hits the expression import at line 124, and re-emits `Module not found: Can't resolve <dynamic>` for that file. The goal stated in the PR body ("Stop Next.js/Turbopack from … repeatedly warning that it cannot resolve a dynamic module") is not met.

Fix: apply the same `/* webpackIgnore: true */` and `/* turbopackIgnore: true */` pair to the line 124 import, and then verify against a real compile rather than by inspection — the PR's own plan already commits to "a real Next.js build/route compilation check" in its Risks section. While doing so, please confirm whether `await import('esbuild')` at line 104 also needs the treatment; it is a static specifier and `esbuild` is already listed in `serverExternalPackages` in `apps/mercato/next.config.ts`, so it should be fine, but it is worth confirming in the same build run rather than assuming.

**`packages/ai-assistant/src/modules/ai_assistant/lib/__tests__/generated-registry-loader.test.ts:46` — the regression test does not test the regression.**

The test reads its sibling source file and runs three `toContain` assertions against the whole file body. `toContain` has no positional meaning, so the test passes as long as those three substrings appear *anywhere* in the file — it never establishes that the magic comments sit inside the `import()` call, which is the only placement webpack and Turbopack honour. It also does not assert anything about the bundle, which is the invariant the test claims to protect in its title.

Failure scenario: a developer (or a formatter, or a lint autofix) hoists the two comments out of the `import()` parentheses and onto the JSDoc block above `compileAppLocalModuleEntries` — a change that reads as harmless cleanup. The bundler stops honouring them, `dynamicLoader` is pulled back into the app graph, the warning returns, and this test stays green because all three substrings are still present in the file. The same applies if someone deletes the `import()` call entirely and leaves a stray comment behind.

Fix: assert on the parsed `import()` call rather than on the file. Matching the import expression and asserting the comments occur *within* it — for example a regex over `await import\(\s*/\* webpackIgnore: true \*/\s*/\* turbopackIgnore: true \*/\s*'@open-mercato/shared/lib/bootstrap/dynamicLoader'\s*\)` — pins the placement that actually matters, and would fail in the scenario above. Better still, add the build-level check the plan's Risks section calls for, so the invariant is verified where it is enforced rather than in the source text.

### 🔹 Minor

**`.ai/runs/2026-08-25-suppress-dynamic-loader-bundler-warning.md:9-13, 45-46` — the committed plan describes an implementation that is no longer in the PR.**

The Scope section says the change will "Separate the app-source compiler from the runtime module importer in `@open-mercato/shared`" and "Point the AI generated-registry fallback at the compiler-only entry point", and Phase 1 marks steps 1.1 and 1.2 as `[x] … — c3a5a29128`. Commit `c3a5a29128` did introduce `packages/shared/src/lib/bootstrap/appSourceCompiler.ts` (419 lines), but the head commit `e4a08d12` reverted that file entirely and restored `dynamicLoader.ts`, replacing the approach with the magic comments. The net diff against `develop` contains no compiler-only module at all.

Failure scenario: `om-auto-continue-pr` (or a human) resumes from this plan, reads 1.1 and 1.2 as done, and continues into Phase 2 believing a compiler boundary exists that was deleted two commits ago — so the abandoned approach is never re-evaluated and the Scope section misdescribes the shipped change to every future reader. Fix: rewrite Scope and Phase 1 to describe the bundler-ignore approach that actually shipped, and re-derive the Progress checkboxes from the head commit.

**PR description — the PR presents itself as unfinished while sitting in review.**

The body says `Status: in-progress`, "What Changed — Execution plan committed; implementation in progress", and "🧪 Tests — Pending", and plan steps 2.1 and 2.2 are unchecked, yet the PR is non-draft and carries a complete implementation. A reviewer or merge-buddy reading only the description cannot tell what is being shipped. Fix: update the body to describe the actual change (which import is excluded from the bundle, why, and how it was verified) and flip the status once the plan's Phase 2 lands.

### 💅 Nit

**`packages/ai-assistant/src/modules/ai_assistant/lib/generated-registry-loader.ts:172-176` — no in-code explanation of the bundler-ignore comments.**

The reasoning for excluding this specifier lives only in `.ai/runs/`, which is a run artifact rather than shipped documentation. This file is otherwise unusually well commented — the module JSDoc and `compileAppLocalModuleEntries`'s own block explain their constraints at length — so a one-line note ("`dynamicLoader` performs a computed `import()` that Turbopack cannot resolve; keep it out of the app graph and let Node resolve it via the package's exports map") would match the surrounding style and stop the next reader from removing the comments as noise. Author's call.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the net diff adds no exports and removes none; `compileAppSourceFile` stays on `@open-mercato/shared/lib/bootstrap/dynamicLoader`.
- [x] Where a contract had to change: not applicable — no contract surface from `BACKWARD_COMPATIBILITY.md` is touched. The one behavioural shift is that `@open-mercato/shared/lib/bootstrap/dynamicLoader` is now resolved by Node at runtime instead of being bundled, which the package's existing `exports` entry already supports, and which no consumer can observe through a public API.

## 🧪 Test Coverage

The PR adds one test, `Next.js bundle boundary › keeps the standalone dynamic loader out of the Next.js bundle`. What it covers is that the strings `'@open-mercato/shared/lib/bootstrap/dynamicLoader'`, `/* webpackIgnore: true */` and `/* turbopackIgnore: true */` each appear somewhere in `generated-registry-loader.ts`. That is a weaker statement than its name implies, for the reasons in the Major finding above: it says nothing about where the comments sit, and nothing about the bundle.

The gaps to close, in order of value:

1. **Placement of the ignore comments.** In the same test file, replace the three `toContain` calls with a single regex assertion over the whole `await import(...)` expression, so the test fails when the comments are moved out of the call. This is the cheap version of the invariant and belongs where the current test already lives.
2. **The remaining dynamic import.** Once line 124 is fixed, extend the same assertion to cover both import sites, so a future edit cannot silently reintroduce an unignored expression import into this module.
3. **The bundler-level invariant.** The plan's Risks section already identifies that a source-only assertion "could miss a warning introduced through another import path" and commits to a real Next.js build/route compilation check. That check has not shipped (plan steps 2.1 and 2.2 are unchecked). Compiling one AI-assistant route and asserting the build log contains no `Can't resolve <dynamic>` entry is what would actually prove the fix, and it is the only form of coverage that would have caught the Major finding.
4. **The runtime resolution path.** Nothing exercises `compileAppLocalModuleEntries` after the import stopped being bundled. A focused test that calls it and asserts `compileAppSourceFile` is reachable would guard against the package `exports` map or the dist layout drifting under the now-native import.

---

🤖 Posted by `om-auto-review-pr`. Submitted as a review comment rather than a formal *Request changes* because GitHub does not allow a PR author to request changes on their own pull request; the verdict is **request changes** and the pipeline label reflects it.

pkarw:
🏷️ **label rationale**

- `changes-requested` — the review found one Major that means the PR does not meet its own goal (`generated-registry-loader.ts:124` still holds an unignored computed `import()` in a bundled module) plus a Major on the regression test. Under this repo's verdict rule any Major or Minor requests changes.
- `bug` — the change fixes a defect (bundler warning / unwanted module in the app graph) rather than adding capability.
- `priority-low` — the symptom is developer-facing build noise. No user-visible breakage, no data risk, no security exposure, so nothing forces this ahead of other work.
- `risk-medium` — the diff is three lines, but it moves `@open-mercato/shared/lib/bootstrap/dynamicLoader` from bundled to natively resolved at runtime on the AI-assistant registry path. That resolution is currently unverified by any test or completed build, and if it fails the AI tool/agent registry stops loading in the app. Contained blast radius, real failure mode: medium rather than low.
- `needs-qa` — the automated-verification exemption does not apply. The diff touches no UI file and changes no DB or API surface, but the exemption also requires automated tests for the changed behaviour in the same PR, and the shipped test asserts on source text rather than on bundler behaviour. QA here is concrete and cheap: run the app, compile an AI-assistant route, confirm no `Can't resolve <dynamic>` warning remains.

No priority or risk label existed before this run, so both were inferred from the diff and stated above.

pkarw (commented):
# Re-review: fix(ai-assistant): avoid bundling standalone dynamic loader

## Summary

The follow-up now excludes both runtime-only import expressions from Webpack/Turbopack traversal, pins the directive placement in regression coverage, updates the run plan and PR description, and includes the latest `develop`. A forced Turbopack application build completes without the targeted `Can't resolve <dynamic>` warning.

## Verdict

approve — the previous Major and Minor findings are resolved; no blocker, major, minor, or nit findings remain.

GitHub does not allow an author to formally approve their own PR, so this verdict is submitted as a review comment and the required human review remains enforced by GitHub.

## Validation Gate

Validation source: GitHub checks for head `08b5bed218` are pending. Per the repo-local review policy, pending checks are not replaced with another full local gate. The full configured local gate passed on the implementation head before the review follow-up; the follow-up itself was additionally verified with the focused test, package build, and forced application build.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | PENDING | Current-head GitHub CI pending; full local gate passed before the focused follow-up. |
| `yarn generate` | PENDING | Current-head GitHub CI pending; generated contracts are unchanged. |
| `yarn build:packages` | PENDING | Current-head GitHub CI pending; AI assistant package build passed after the follow-up. |
| `yarn i18n:check-sync` | PENDING | Current-head GitHub CI pending; no locale files changed. |
| `yarn i18n:check-usage` | PENDING | Current-head GitHub CI pending; no user-facing strings changed. |
| `yarn typecheck` | PENDING | Current-head GitHub CI pending; full local gate passed before the focused follow-up. |
| `yarn test` | PENDING | Current-head GitHub CI pending; focused generated-registry suite passed 21/21 after the follow-up. |
| `yarn build:app` | PASS | Forced Turbopack build passed after the follow-up with no `Can't resolve <dynamic>` warning. |

Additional check: `yarn template:sync` passed.

## Findings

No findings.

## Breaking Changes

- [x] Where a contract had to change: not applicable; the existing loader implementation and public import path remain unchanged

## Test Coverage

Covered: the regression test asserts that both computed runtime imports carry both bundler directives inside the exact `import()` expressions. The focused suite passes 21/21, and the real Turbopack application build confirms the reported warning is absent.

---

🤖 Posted by `om-auto-review-pr`. Self-review verdict: **approve**; formal approval remains for another reviewer.

pkarw:
🏷️ Label set to `merge-queue` because the re-review and required local verification passed; `skip-qa` applies because this isolated non-UI change preserves DB/API contracts and includes automated regression coverage.

pkarw:
🤖 `om-auto-create-pr` completed.

## Outcome

- merged the latest `develop` into the fix branch;
- excluded both runtime-only registry imports from Webpack/Turbopack traversal;
- preserved the shared loader and all public contract surfaces;
- tightened regression coverage to assert directive placement inside both `import()` expressions.

## Verification

- focused loader suite: 21/21 passed;
- AI assistant package build: passed;
- forced Turbopack application build: passed with no `Can't resolve <dynamic>` warning;
- full configured local gate: passed on the implementation head;
- template parity: passed;
- current-head GitHub checks: running and still authoritative for merge.

## Review and QA

Re-review is clean with no findings. GitHub requires another reviewer because the PR author cannot self-approve. `skip-qa` applies: there are no UI, API, or database changes, and the changed behavior has automated regression coverage plus a real application build.

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED. Lock released. Previous findings were resolved in one follow-up iteration; formal approval remains for another reviewer because GitHub blocks self-approval.

pkarw (commented):
# 🔍 Re-review: fix(ai-assistant): avoid bundling standalone dynamic loader

## 🎯 Summary

Head `08b5bed2` addresses both findings from my earlier review, and the merge of `develop` is in. All 15 reported checks pass and none are pending. I then ran the verification that my first review said was missing — a real Turbopack A/B build — and it changes one of my own findings, so the correction is below rather than buried.

## Verdict

✅ **approve** — no blockers or majors remain. I cannot record this as a formal approval (GitHub refuses `APPROVE` on your own pull request), so this is a comment and the PR still needs one approving review from another maintainer plus the `needs-qa` gate before it can merge.

## ⚠️ Correction to my earlier review — my Major #1 was wrong

I claimed that `generated-registry-loader.ts:124` (`await import(pathToFileURL(jsPath).href)`) would keep emitting `Module not found: Can't resolve <dynamic>` because it is a computed specifier in a module that stays in the bundle. That is not what happens. I tested it instead of reasoning about it, and the reasoning was wrong.

Three uncached production builds (`rm -rf apps/mercato/.mercato/next && next build`, Turbopack, Next 16.3), each preceded by `yarn build:packages` so `dist` matched the source under test:

| Loader source | `Can't resolve <dynamic>` | Total Turbopack warnings |
|---|---|---|
| `develop` baseline | **1** | 9 |
| PR head `e4a08d12` (dynamicLoader ignored only, **before** `f1faae81`) | **0** | 6 |
| With the line-124 guard added | **0** | 6 |

The baseline warning points somewhere neither of us named:

```
./packages/shared/dist/lib/bootstrap/dynamicLoader.js:293:18
Warning: Module not found: Can't resolve <dynamic>
  292 |     const fileUrl = `${pathToFileURL(jsPath).href}?cache=${outputHash}`;
> 293 |     return await import(fileUrl);

Import trace:
  ./packages/shared/dist/lib/bootstrap/dynamicLoader.js
  ./packages/ai-assistant/dist/modules/ai_assistant/lib/generated-registry-loader.js
  ./packages/ai-assistant/dist/modules/ai_assistant/lib/tool-loader.js
  ./packages/ai-assistant/dist/modules/ai_assistant/api/tools/route.js
```

The mechanism is the specifier's *shape*, not the fact that it is computed. `` `${pathToFileURL(jsPath).href}?cache=${outputHash}` `` is a template literal with a static `?cache=` segment, so Turbopack treats it as a resolvable pattern, tries to resolve it, and fails. A wholly-computed expression like line 124 gives it nothing to pattern-match, so it is left as a runtime import silently. That is why `e4a08d12` — cutting the single edge into `dynamicLoader` — was already a complete fix, and why the guard added in `f1faae81` changes no observed behaviour.

Two consequences, neither blocking:

1. The line-124 guard in `f1faae81` is harmless but not load-bearing. Keeping it is fine as defence-in-depth; my objection to leaving it off was not.
2. The PR body ("mark **both** runtime-only `import()` expressions") and the plan's follow-up note ("guards **both** runtime-only import expressions") read as if both were required. If either is edited again, saying that only the `dynamicLoader` edge produced the warning would keep the next reader from inferring an invariant that does not exist — the wholly-computed form never warns.

## ✅ Findings from the first review — resolved

- **Major, regression test asserted nothing about placement** — fixed in `f1faae81`. The three `toContain` calls are replaced by regexes matching the whole `import()` expression. I verified independently that this is real: hoisting the comments out of the call while leaving them in the file fails the new assertion and passed the old one.
- **Minor, plan contradicted the shipped implementation** — fixed in `72d64290` / `a7baafaa`. The reverted `appSourceCompiler` approach is recorded as superseded rather than as completed work.
- **Minor, PR body described the change as unfinished** — fixed. The body now states the problem, the fix, and the validation performed.

## 🧪 Validation Gate

Validation source: **GitHub checks** — 15 pass, 5 skipping, none failing, none pending on `08b5bed2` (including `test`, `lint`, `ds-lint`, `docker-build`, `ephemeral-integration`, CodeQL). Locally I additionally ran `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, the full `@open-mercato/ai-assistant` suite (1479 tests, 109 suites) and the Turbopack builds above — all green. The four `create-mercato-app` failures I hit locally are a host inotify limit (`fs.inotify.max_user_instances: 128 < 4096`), reproducible by running `template/scripts/dev.mjs` directly, and unrelated to this branch; they pass on CI.

## 🧪 Test Coverage

The bundle-boundary assertions now pin the directives to the inside of their `import()` calls, which is the placement the bundlers honour, and that is the right invariant to hold in a unit test. What no unit test can cover is whether the warning is actually gone — that needs a compile. The A/B builds above supply exactly that evidence for this change; if this area is touched again, that comparison is the check worth repeating, and it has to bypass Turbo (`yarn build:app` returns `FULL TURBO` even after a dependency package is rebuilt, because the app task's hash does not include `packages/*/dist` — I lost two builds to that before noticing).

---

⚠️ **Branch incident, disclosed in full.** While I was validating, this branch's ref was briefly absent on the remote. My non-force `git push` therefore landed as a *new branch* and pointed the head at my own commit, dropping `72d64290`, `f1faae81`, `a7baafaa` and the `develop` merge from the branch tip. I detected it immediately and restored the branch to `08b5bed2` with `--force-with-lease`; `git ls-remote` confirms the head is back, all four commits are intact, and the green checks above are the ones that ran on `08b5bed2`. No work was lost and nothing else was touched. My own commit is not on the branch — everything in it was either already done in `f1faae81` or is superseded by the correction above, so I did not re-push it.

🤖 Posted by `om-auto-review-pr`.

pkarw:
🏷️ **label rationale (updated)**

Re-review verdict is approve, so I removed `changes-requested`. A concurrent session had already set `merge-queue` and `skip-qa` on this PR; I briefly added `review` alongside `merge-queue`, which broke the single-pipeline-label rule, and removed it again — `merge-queue` stands as the pipeline label. `bug`, `priority-low` and `risk-medium` are unchanged.

On `skip-qa`: I originally applied `needs-qa` because the automated-verification exemption requires automated tests for the changed behaviour, and a source-text assertion is not one. That reservation no longer holds — the A/B Turbopack builds in my re-review verify the behaviour directly (baseline emits the warning, this branch does not), and `ephemeral-integration` is green. `skip-qa` is well supported now, so I left it as the concurrent session set it.

Merge is still gated on one approving review from another maintainer: branch protection requires one, and GitHub will not let the author approve their own PR, so my approve verdict is a comment only.

🤖 `om-auto-review-pr` completed: APPROVED (re-review). Lock released.

Issue #5606, opened 2026-08-25
perf(search): Meilisearch sync writes one document per HTTP call, no batching even in batch-index worker path


adeptofvoltron:
## Summary
- Real-time Meilisearch sync writes one document at a time (one `POST .../documents` HTTP call per record) with zero batching/debounce, even on the worker's own `batch-index` job type — under concurrent write load this saturates Meilisearch and search latency regresses ~14-32x even though Meilisearch itself isn't CPU-throttled.

## Analysis
- Affected areas: `packages/search/src/modules/search/subscribers/fulltext_upsert.ts` (enqueues one job per record on create/update), `workers/fulltext-index.worker.ts` (`DEFAULT_CONCURRENCY = 2`), `packages/search/src/fulltext/drivers/meilisearch/index.ts` (the driver).
- Confirmed today (`develop @ 94f1126`), full write path traced: entity create/update → `search.index_record` subscriber enqueues a single job `{ jobType: 'index', recordId, ... }` → worker's `jobType === 'index'` branch calls `searchIndexer.indexRecordById(...)` → `searchService.index(indexableRecord)` (`search-indexer.ts:279`) → Meilisearch driver does `index.addDocuments([document], { primaryKey: '_id' })` (`index.ts:247-256`) — one HTTP POST per record.
- **Notably**: even the existing `jobType === 'batch-index'` path in the same worker (meant for cases where multiple records were already collected into one job) does not do a single bulk write — it loops and calls `indexRecordById()` separately per record, i.e. still N separate `addDocuments([doc])` calls instead of one `addDocuments(docs)`. The driver **does** have a real batch method (`addDocuments(documents)` grouped per tenant, `index.ts:278-295`) — no event-driven path uses it.
- PRs #4592/#4654/#5402/#1529 address other search aspects (tokens, avoiding full-table scans on reindex) — none change the per-write Meilisearch call count.
[...]

Issue #3103, opened 2026-06-15
reliability(search): strategy index failures are swallowed by Promise.allSettled, so queue retries never fire (silent index loss on 'too many clients already')


pkarw:
## Summary

Search index jobs are reported as **completed** even when a strategy fails, so the queue's retry mechanism never fires for transient failures like Postgres connection-pool exhaustion (`sorry, too many clients already`). The record is silently dropped from the index — no retry, no dead-letter.

## Observed

```
[queue:vector-indexing] Job 7abb884f-... completed
[SearchService] Strategy index failed {
  strategyId: 'vector',
  entityId: 'sales:sales_shipment',
  recordId: '086510b3-...',
  error: 'sorry, too many clients already'
}
```

The job logs `completed` despite the strategy index failing.

## Root cause

The queue layer **does** support retries:

- `async`/BullMQ — `attempts: 3` + exponential backoff (`packages/queue/src/strategies/async.ts:138`)
- `local` — `DEFAULT_MAX_ATTEMPTS = 3` + exponential backoff + dead-letter (`packages/queue/src/strategies/local.ts:200-250`). (Note: the `No retry mechanism` comment at `local.ts:35` is stale and should be corrected.)

But the failure never reaches the queue. `SearchService.index()` runs strategies through `Promise.allSettled` and **only logs** rejections — it never re-throws:

`packages/search/src/service.ts:201-217`

```ts
const results = await Promise.allSettled(
  strategies.map((strategy) => this.executeStrategyIndex(strategy, record)),
)
// Log any failures ... (no re-throw)
```

Call chain that swallows the error:
`vector-index.worker` / `fulltext-index.worker`
→ `SearchIndexer.indexRecordById()` → `indexRecord()` (`search-indexer.ts:279`)
→ `SearchService.index()` (swallows) → worker returns normally → queue marks job **completed**.

The worker's own `throw error` retry path (`vector-index.worker.ts:464`, `fulltext-index.worker.ts:367`) is never exercised because the error is consumed upstream.

## Impact

- Transient infra failures (DB pool exhaustion, Meilisearch/embedding-provider blips) cause **permanent, silent index gaps** — records never get indexed and there is no retry or dead-letter signal.
- `risk-high`: this is an event-reliability / silent-data-loss class issue affecting all indexed entities.

## Proposed fix (for discussion)

[...]

PR #1507, opened 2026-04-14
fix: 🔒 reliability: Search bulkIndex silently swallows strategy failures (#1424)

pat-lewczuk:
## Automated fix for #1424

Fixes #1424

> This PR was opened by [cezar](https://github.com/comerito/cezar) autofix. It is a **draft** — a human reviewer must verify correctness before it merges.

### Root cause
SearchService.bulkIndex silently swallows strategy failures instead of propagating them to callers

The bulkIndex method uses Promise.allSettled() at line 227-249 to execute indexing across multiple strategies. When strategies fail, the rejected promises are only logged via searchError() but never propagated to the caller. This causes partial indexing failures to be silently ignored, leading to inconsistent search indexes where some strategies succeed while others fail, with no upstream notification for retry logic.

### Approach
Changed bulkIndex() to throw descriptive errors when strategy failures occur instead of silently swallowing them. Updated all callers to handle the new exception behavior with proper try/catch blocks and descriptive error messages. Added comprehensive test coverage for both single and multiple strategy failure scenarios.

### Files changed
- `packages/search/src/service.ts`
- `packages/search/src/indexer/search-indexer.ts`
- `packages/ai-assistant/src/modules/ai_assistant/lib/tool-search.ts`
- `packages/search/src/__tests__/service.test.ts`

### Verification
Commands run by the fixer:
- `yarn test src/__tests__/service.test.ts`
- `yarn test --testNamePattern="bulkIndex" src/__tests__/service.test.ts`
- `npx turbo run typecheck --filter=@open-mercato/search`

### Review (automated)
**Verdict:** `pass`

Fix correctly addresses the root cause by converting silent failures to thrown exceptions. All direct callers are properly updated with try/catch handlers. Error messages are descriptive and preserve logging. The implementation is correct with proper null safety and array indexing.

Issues raised:
- **minor** `packages/search/src/__tests__/service.test.ts`:338: Missing test case for mixed scenario where one strategy succeeds and another fails. The code correctly throws, but this important edge case is not explicitly tested.

### Remaining concerns
_(none)_

PR #520, opened 2026-02-09
feat(search): prevent auto-reindex feedback loops in search indexing …

simonkak:
## Summary

Search indexing operations that query the `HybridQueryEngine` to hydrate entity data can trigger coverage gap detection, which schedules auto-reindex events, which trigger more indexing — creating an infinite feedback loop. This change introduces a `skipAutoReindex` query option that the search indexing pipeline uses to suppress reindex scheduling during its own queries, breaking the cycle.

## Problem

When browsing pages with entities that have vector search enabled (e.g., /backend/customers/people), an infinite loop occurs where vector-indexing and events queue jobs grow indefinitely — observed reaching 1500+ jobs and climbing without bound.

The root cause is a feedback loop in the search indexing pipeline:
1. A user page load queries HybridQueryEngine, which detects a coverage gap (baseCount: 6, indexedCount: 11) and calls scheduleAutoReindex()
2. scheduleAutoReindex() emits a persistent query_index.reindex event
3. The reindex subscriber calls reindexEntity() with emitVectorizeEvents: true
4. This emits query_index.vectorize_one for every record (6 events)
5. Each vectorize_one event enqueues a vector-indexing job
6. The vector-indexing worker calls searchIndexer.indexRecordById(), which calls indexRecord() → buildSource(ctx)
7. buildSource() callbacks in module search configs (e.g., customers/search.ts) call ctx.queryEngine.query() to hydrate related entities — without skipAutoReindex
8. Those queries hit the coverage gap check again → back to step 1, creating an amplification loop (1 job → N jobs per cycle)

## Changes

- Added `skipAutoReindex?: boolean` to `QueryOptions` in `packages/shared/src/lib/query/types.ts`
- Guarded both `scheduleAutoReindex` call sites in `HybridQueryEngine` (`engine.ts`) behind `!opts.skipAutoReindex`
- Created `noReindexQueryEngine` getter in `SearchIndexer` that wraps the query engine to force `skipAutoReindex: true` on all queries made by `buildSource`/`formatResult` callbacks
- Added `skipAutoReindex: true` to all direct `queryEngine.query()` calls in `SearchIndexer` (record lookup, fulltext reindex, vector reindex batches)
- Added `skipAutoReindex: true` to `VectorIndexService` queries in `fetchRecord` and vector reindex batch loop

## Specification

**Does a spec exist for this feature/module?**
- [ ] Yes
- [ ] No (created a new spec)
- [x] N/A (minor change, no spec needed)
**Spec file path:**
[...]

PR #5612, opened 2026-08-25
fix(search): batch-index fulltext worker writes once, not once per record

adeptofvoltron:
## Summary
- The Meilisearch fulltext worker's `batch-index` job type already receives multiple record ids in one job, but looped `searchIndexer.indexRecordById()` per record — one `addDocuments([doc])` HTTP write per record — even though the driver already has a working `bulkIndex()` path (`addDocuments(documents)`, grouped per tenant) that nothing used.
- Adds `SearchIndexer.indexRecordsById()`, which loads each queued record fresh (same load as `indexRecordById`) but flushes the whole batch through exactly one `searchService.bulkIndex()` call, and wires the worker's `batch-index` branch to call it once instead of looping.
- The real-time single-record path (`fulltext_upsert.ts` subscriber → `jobType: 'index'` → `indexRecord`/`indexRecordById`) is intentionally untouched — the issue's broader ask to debounce/batch that path changes queue/timing behavior on the hot write path, which `packages/search/AGENTS.md` flags as "Ask First" ("changing search strategy defaults, entity IDs, queue behavior"). Flagging this as a separate follow-up rather than guessing at debounce-window semantics in this PR.

## Test plan
- [x] `yarn workspace @open-mercato/search test` — 34 suites / 333 tests pass, including a new regression test (`search-indexer-batch.test.ts`) asserting N queued records across one or more entities collapse into exactly one `searchService.bulkIndex()` call, and updated `workers.test.ts` batch-index tests asserting `indexRecordsById` is called once (never `indexRecordById` in a loop).
- [x] `yarn workspace @open-mercato/search build`
- [x] `yarn typecheck` (all 30 packages)
- [x] `yarn i18n:check-sync`, `yarn i18n:check-usage`
- [x] `yarn test` (full monorepo) — only failure is the pre-existing, unrelated `packages/cli` `dev-env-reload.test.ts` inotify flake (known, tracked separately; does not touch search)
- [x] `yarn build:app`

## Security impact
None — internal indexing-pipeline change only; no auth, secrets, encryption, logging/audit, backups, or access-control surface touched.

Refs #5606 — this PR implements step 2 of that issue (the `batch-index` worker branch). Step 1 (batching/debouncing the real-time per-record write path, which the issue's latency evidence attributes the regression to) is intentionally out of scope, so the issue stays open rather than being auto-closed.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5612"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790244272&installation_model_id=432243&pr_number=5612&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5612&signature=229a6cce4bc86bf7bd46aeffdafb84183af088afeba1edc37ed4521fb4cbba99"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/search/src/__tests__/search-indexer-batch.test.ts
+++ b/packages/search/src/__tests__/search-indexer-batch.test.ts
@@ -11,15 +11,20 @@ describe('SearchIndexer.indexRecordsById', () => {
           enabled: true,
           formatResult: async (ctx) => ({ title: String(ctx.record.name ?? ctx.record.id) }),
         },
+        {
+          entityId: 'test:other',
+          enabled: true,
+          formatResult: async (ctx) => ({ title: String(ctx.record.name ?? ctx.record.id) }),
+        },
       ],
     },
   ]
 
-  function makeQueryEngine(records: Record<string, unknown>[]): QueryEngine {
+  function makeQueryEngine(recordsByEntity: Record<string, Record<string, unknown>[]>): QueryEngine {
     return {
-      query: jest.fn(async (_entity, opts) => {
+      query: jest.fn(async (entity, opts) => {
         const wantedId = (opts?.filters as { id?: string } | undefined)?.id
-        const items = records.filter((r) => r.id === wantedId)
+        const items = (recordsByEntity[entity as string] ?? []).filter((r) => r.id === wantedId)
         return { items, total: items.length } as QueryResult
       }),
     }
@@ -33,18 +38,18 @@ describe('SearchIndexer.indexRecordsById', () => {
     ]
     const searchService = { bulkIndex: jest.fn().mockResolvedValue(undefined) }
     const indexer = new SearchIndexer(searchService as any, moduleConfigs, {
-      queryEngine: makeQueryEngine(records),
+      queryEngine: makeQueryEngine({ 'test:entity': records }),
     })
 
-    const result = await indexer.indexRecordsById(
-      [
+    const result = await indexer.indexRecordsById({
+      items: [
         { entityId: 'test:entity', recordId: 'rec-1' },
         { entityId: 'test:entity', recordId: 'rec-2' },
         { entityId: 'test:entity', recordId: 'rec-3' },
       ],
-      'tenant-123',
-      'org-456',
-    )
+      tenantId: 'tenant-123',
+      organizationId: 'org-456',
+    })
 
     expect(searchService.bulkIndex).toHaveBeenCalledTimes(1)
     expect(searchService.bulkIndex).toHaveBeenCalledWith([
@@ -55,39 +60,155 @@ describe('SearchIndexer.indexRecordsById', () => {
     expect(result).toEqual({ indexed: 3, skipped: 0 })
   })
 
+  it('collapses records spanning multiple entities into one bulkIndex call', async () => {
+    const searchService = { bulkIndex: jest.fn().mockResolvedValue(undefined) }
+    const indexer = new SearchIndexer(searchService as any, moduleConfigs, {
+      queryEngine: makeQueryEngine({
+        'test:entity': [{ id: 'rec-1', name: 'Alpha' }],
+        'test:other': [{ id: 'other-1', name: 'Delta' }],
+      }),
+    })
+
+    const result = await indexer.indexRecordsById({
+      items: [
+        { entityId: 'test:entity', recordId: 'rec-1' },
+        { entityId: 'test:other', recordId: 'other-1' },
+      ],
+      tenantId: 'tenant-123',
+      organizationId: null,
+    })
+
+    expect(searchService.bulkIndex).toHaveBeenCalledTimes(1)
+    expect(searchService.bulkIndex).toHaveBeenCalledWith([
+      expect.objectContaining({ entityId: 'test:entity', recordId: 'rec-1' }),
+      expect.objectContaining({ entityId: 'test:other', recordId: 'other-1' }),
+    ])
+    expect(result).toEqual({ indexed: 2, skipped: 0 })
+  })
+
+  it('loads each record with custom fields and without triggering auto-reindex', async () => {
+    const queryEngine = makeQueryEngine({ 'test:entity': [{ id: 'rec-1', name: 'Alpha' }] })
+    const searchService = { bulkIndex: jest.fn().mockResolvedValue(undefined) }
+    const indexer = new SearchIndexer(searchService as any, moduleConfigs, { queryEngine })
+
+    await indexer.indexRecordsById({
+      items: [{ entityId: 'test:entity', recordId: 'rec-1' }],
+      tenantId: 'tenant-123',
+      organizationId: 'org-456',
+    })
+
+    expect(queryEngine.query).toHaveBeenCalledWith(
+      'test:entity',
+      expect.objectContaining({
+        tenantId: 'tenant-123',
+        organizationId: 'org-456',
+        filters: { id: 'rec-1' },
+        includeCustomFields: true,
+        skipAutoReindex: true,
+      }),
+    )
+  })
+
   it('skips records that no longer exist without failing the batch write', async () => {
     const records = [{ id: 'rec-1', name: 'Alpha' }]
     const searchService = { bulkIndex: jest.fn().mockResolvedValue(undefined) }
     const indexer = new SearchIndexer(searchService as any, moduleConfigs, {
-      queryEngine: makeQueryEngine(records),
+      queryEngine: makeQueryEngine({ 'test:entity': records }),
     })
 
-    const result = await indexer.indexRecordsById(
-      [
+    const result = await indexer.indexRecordsById({
+      items: [
         { entityId: 'test:entity', recordId: 'rec-1' },
         { entityId: 'test:entity', recordId: 'missing' },
       ],
-      'tenant-123',
-      null,
-    )
+      tenantId: 'tenant-123',
+      organizationId: null,
+    })
 
     expect(searchService.bulkIndex).toHaveBeenCalledTimes(1)
     expect(result).toEqual({ indexed: 1, skipped: 1 })
   })
 
+  it('keeps indexing the batch when loading one record throws', async () => {
+    const searchService = { bulkIndex: jest.fn().mockResolvedValue(undefined) }
+    const queryEngine: QueryEngine = {
+      query: jest.fn(async (_entity, opts) => {
+        const wantedId = (opts?.filters as { id?: string } | undefined)?.id
+        if (wantedId === 'boom') throw new Error('connection lost')
+        return { items: [{ id: wantedId, name: 'Alpha' }], total: 1 } as QueryResult
+      }),
+    }
+    const indexer = new SearchIndexer(searchService as any, moduleConfigs, { queryEngine })
+
+    const result = await indexer.indexRecordsById({
+      items: [
+        { entityId: 'test:entity', recordId: 'rec-1' },
+        { entityId: 'test:entity', recordId: 'boom' },
+        { entityId: 'test:entity', recordId: 'rec-2' },
+      ],
+      tenantId: 'tenant-123',
+      organizationId: null,
+    })
+
+    expect(searchService.bulkIndex).toHaveBeenCalledTimes(1)
+    expect(searchService.bulkIndex).toHaveBeenCalledWith([
+      expect.objectContaining({ recordId: 'rec-1' }),
+      expect.objectContaining({ recordId: 'rec-2' }),
+    ])
+    expect(result).toEqual({ indexed: 2, skipped: 1 })
+  })
+
+  it('propagates a bulkIndex failure so the queue can retry the job', async () => {
+    const searchService = { bulkIndex: jest.fn().mockRejectedValue(new Error('meilisearch unavailable')) }
+    const indexer = new SearchIndexer(searchService as any, moduleConfigs, {
+      queryEngine: makeQueryEngine({ 'test:entity': [{ id: 'rec-1', name: 'Alpha' }] }),
+    })
+
+    await expect(
+      indexer.indexRecordsById({
+        items: [{ entityId: 'test:entity', recordId: 'rec-1' }],
+        tenantId: 'tenant-123',
+        organizationId: null,
+      }),
+    ).rejects.toThrow('meilisearch unavailable')
+  })
+
   it('skips entities that are not configured and never calls bulkIndex when nothing is indexable', async () => {
     const searchService = { bulkIndex: jest.fn().mockResolvedValue(undefined) }
     const indexer = new SearchIndexer(searchService as any, moduleConfigs, {
-      queryEngine: makeQueryEngine([]),
+      queryEngine: makeQueryEngine({}),
     })
 
-    const result = await indexer.indexRecordsById(
-      [{ entityId: 'unknown:entity', recordId: 'rec-1' }],
-      'tenant-123',
-      null,
-    )
+    const result = await indexer.indexRecordsById({
+      items: [{ entityId: 'unknown:entity', recordId: 'rec-1' }],
+      tenantId: 'tenant-123',
+      organizationId: null,
+    })
 
     expect(searchService.bulkIndex).not.toHaveBeenCalled()
     expect(result).toEqual({ indexed: 0, skipped: 1 })
+  })
+
+  it('counts records dropped for a missing id as skipped so the totals add up', async () => {
+    const searchService = { bulkIndex: jest.fn().mockResolvedValue(undefined) }
+    const queryEngine: QueryEngine = {
+      query: jest.fn(async (_entity, opts) => {
+        const wantedId = (opts?.filters as { id?: string } | undefined)?.id
+        const item = wantedId === 'no-id' ? { name: 'Ghost' } : { id: wantedId, name: 'Alpha' }
+        return { items: [item], total: 1 } as QueryResult
+      }),
+    }
+    const indexer = new SearchIndexer(searchService as any, moduleConfigs, { queryEngine })
+
+    const result = await indexer.indexRecordsById({
+      items: [
+        { entityId: 'test:entity', recordId: 'rec-1' },
+        { entityId: 'test:entity', recordId: 'no-id' },
+      ],
+      tenantId: 'tenant-123',
+      organizationId: null,
+    })
+
+    expect(result).toEqual({ indexed: 1, skipped: 1 })
   })
 })
--- a/packages/search/src/__tests__/search-indexer-batch.test.ts
+++ b/packages/search/src/__tests__/search-indexer-batch.test.ts
@@ -1,0 +1,93 @@
+import { SearchIndexer } from '../indexer/search-indexer'
+import type { SearchModuleConfig } from '../types'
+import type { QueryEngine, QueryResult } from '@open-mercato/shared/lib/query/types'
+
+describe('SearchIndexer.indexRecordsById', () => {
+  const moduleConfigs: SearchModuleConfig[] = [
+    {
+      entities: [
+        {
+          entityId: 'test:entity',
+          enabled: true,
+          formatResult: async (ctx) => ({ title: String(ctx.record.name ?? ctx.record.id) }),
+        },
+      ],
+    },
+  ]
+
+  function makeQueryEngine(records: Record<string, unknown>[]): QueryEngine {
+    return {
+      query: jest.fn(async (_entity, opts) => {
+        const wantedId = (opts?.filters as { id?: string } | undefined)?.id
+        const items = records.filter((r) => r.id === wantedId)
+        return { items, total: items.length } as QueryResult
+      }),
+    }
+  }
+
+  it('writes N queued records through exactly one bulkIndex call, not N', async () => {
+    const records = [
+      { id: 'rec-1', name: 'Alpha' },
+      { id: 'rec-2', name: 'Beta' },
+      { id: 'rec-3', name: 'Gamma' },
+    ]
+    const searchService = { bulkIndex: jest.fn().mockResolvedValue(undefined) }
+    const indexer = new SearchIndexer(searchService as any, moduleConfigs, {
+      queryEngine: makeQueryEngine(records),
+    })
+
+    const result = await indexer.indexRecordsById(
+      [
+        { entityId: 'test:entity', recordId: 'rec-1' },
+        { entityId: 'test:entity', recordId: 'rec-2' },
+        { entityId: 'test:entity', recordId: 'rec-3' },
+      ],
+      'tenant-123',
+      'org-456',
+    )
+
+    expect(searchService.bulkIndex).toHaveBeenCalledTimes(1)
+    expect(searchService.bulkIndex).toHaveBeenCalledWith([
+      expect.objectContaining({ entityId: 'test:entity', recordId: 'rec-1', tenantId: 'tenant-123', organizationId: 'org-456' }),
+      expect.objectContaining({ entityId: 'test:entity', recordId: 'rec-2', tenantId: 'tenant-123', organizationId: 'org-456' }),
+      expect.objectContaining({ entityId: 'test:entity', recordId: 'rec-3', tenantId: 'tenant-123', organizationId: 'org-456' }),
+    ])
+    expect(result).toEqual({ indexed: 3, skipped: 0 })
+  })
+
+  it('skips records that no longer exist without failing the batch write', async () => {
+    const records = [{ id: 'rec-1', name: 'Alpha' }]
+    const searchService = { bulkIndex: jest.fn().mockResolvedValue(undefined) }
+    const indexer = new SearchIndexer(searchService as any, moduleConfigs, {
+      queryEngine: makeQueryEngine(records),
+    })
+
+    const result = await indexer.indexRecordsById(
+      [
+        { entityId: 'test:entity', recordId: 'rec-1' },
+        { entityId: 'test:entity', recordId: 'missing' },
+      ],
+      'tenant-123',
+      null,
+    )
+
+    expect(searchService.bulkIndex).toHaveBeenCalledTimes(1)
+    expect(result).toEqual({ indexed: 1, skipped: 1 })
+  })
+
+  it('skips entities that are not configured and never calls bulkIndex when nothing is indexable', async () => {
+    const searchService = { bulkIndex: jest.fn().mockResolvedValue(undefined) }
+    const indexer = new SearchIndexer(searchService as any, moduleConfigs, {
+      queryEngine: makeQueryEngine([]),
+    })
+
+    const result = await indexer.indexRecordsById(
+      [{ entityId: 'unknown:entity', recordId: 'rec-1' }],
+      'tenant-123',
+      null,
+    )
+
+    expect(searchService.bulkIndex).not.toHaveBeenCalled()
+    expect(result).toEqual({ indexed: 0, skipped: 1 })
+  })
+})
--- a/packages/search/src/__tests__/service.test.ts
+++ b/packages/search/src/__tests__/service.test.ts
@@ -431,6 +431,30 @@ describe('SearchService', () => {
       expect(strategy.index).toHaveBeenCalledTimes(2)
     })
 
+    it('should bound in-flight writes when falling back to individual indexing', async () => {
+      let inFlight = 0
+      let peakInFlight = 0
+      const strategy = createMockStrategy({
+        id: 'test',
+        bulkIndex: undefined,
+        index: jest.fn(async () => {
+          inFlight++
+          peakInFlight = Math.max(peakInFlight, inFlight)
+          await new Promise<void>((resolve) => setTimeout(resolve, 0))
+          inFlight--
+        }),
+      })
+      const service = new SearchService({ strategies: [strategy] })
+      const records = Array.from({ length: 50 }, (_, index) =>
+        createMockRecord({ recordId: `rec-${index}` }),
+      )
+
+      await service.bulkIndex(records)
+
+      expect(strategy.index).toHaveBeenCalledTimes(50)
+      expect(peakInFlight).toBeLessThanOrEqual(4)
+    })
+
     it('should do nothing when records array is empty', async () => {
       const strategy = createMockStrategy({ id: 'test' })
       const service = new SearchService({ strategies: [strategy] })
--- a/packages/search/src/__tests__/workers.test.ts
+++ b/packages/search/src/__tests__/workers.test.ts
@@ -507,14 +507,14 @@ describe('Fulltext Index Worker', () => {
     // Verify the whole batch is written through exactly one indexRecordsById
     // call, not one indexRecordById call per record.
     expect(mockSearchIndexer.indexRecordsById).toHaveBeenCalledTimes(1)
-    expect(mockSearchIndexer.indexRecordsById).toHaveBeenCalledWith(
-      [
+    expect(mockSearchIndexer.indexRecordsById).toHaveBeenCalledWith({
+      items: [
         { entityId: 'test:entity', recordId: 'rec-1' },
         { entityId: 'test:entity', recordId: 'rec-2' },
       ],
-      'tenant-123',
-      undefined,
-    )
+      tenantId: 'tenant-123',
+      organizationId: undefined,
+    })
     expect(mockSearchIndexer.indexRecordById).not.toHaveBeenCalled()
   })
 
@@ -548,6 +548,36 @@ describe('Fulltext Index Worker', () => {
       expect.objectContaining({ type: 'fulltext', tenantId: 'tenant-123', delta: 2 }),
     )
     expect(clearReindexLock).toHaveBeenCalledWith(mockDb, 'tenant-123', 'fulltext', 'org-456')
+  })
+
+  it('re-throws a failed fulltext batch write without advancing reindex progress so the queue retries it', async () => {
+    mockSearchIndexer.indexRecordsById.mockRejectedValueOnce(new Error('meilisearch unavailable'))
+    const records = [
+      { entityId: 'test:entity', recordId: 'rec-1' },
+      { entityId: 'test:entity', recordId: 'rec-2' },
+    ]
+    const containerWithProgress: HandlerContext = {
+      resolve: jest.fn((name: string) => {
+        if (name === 'searchStrategies') return [mockFulltextStrategy]
+        if (name === 'em') return mockEm
+        if (name === 'searchIndexer') return mockSearchIndexer
+        if (name === 'progressService') return { id: 'progress' }
+        throw new Error(`Unknown service: ${name}`)
+      }) as HandlerContext['resolve'],
+    }
+    const job = createMockJob<FulltextIndexJobPayload>({
+      jobType: 'batch-index',
+      tenantId: 'tenant-123',
+      organizationId: 'org-456',
+      records,
+    })
+
+    await expect(
+      handleFulltextIndexJob(job, createMockJobContext(), containerWithProgress),
+    ).rejects.toThrow('meilisearch unavailable')
+
+    expect(updateReindexProgress).not.toHaveBeenCalled()
+    expect(incrementReindexProgress).not.toHaveBeenCalled()
   })
 
   it('clears an orphaned fulltext reindex lock instead of recreating it when no progress job is active', async () => {
--- a/packages/search/src/__tests__/workers.test.ts
+++ b/packages/search/src/__tests__/workers.test.ts
@@ -451,6 +451,7 @@ describe('Fulltext Index Worker', () => {
   const mockSearchIndexer = {
     getEntityConfig: jest.fn().mockReturnValue(null),
     indexRecordById: jest.fn().mockResolvedValue({ action: 'indexed', created: true }),
+    indexRecordsById: jest.fn().mockResolvedValue({ indexed: 0, skipped: 0 }),
   }
 
   const mockEm = {
@@ -471,6 +472,7 @@ describe('Fulltext Index Worker', () => {
     ;(hasActiveReindexProgress as jest.Mock).mockResolvedValue(true)
     mockFulltextStrategy.isAvailable.mockResolvedValue(true)
     mockSearchIndexer.indexRecordById.mockResolvedValue({ action: 'indexed', created: true })
+    mockSearchIndexer.indexRecordsById.mockResolvedValue({ indexed: 0, skipped: 0 })
   })
 
   it('should skip job with missing tenantId', async () => {
@@ -486,12 +488,13 @@ describe('Fulltext Index Worker', () => {
     expect(mockFulltextStrategy.bulkIndex).not.toHaveBeenCalled()
   })
 
-  it('should index records via searchIndexer when jobType is batch-index', async () => {
+  it('should index the whole batch in a single indexRecordsById call when jobType is batch-index', async () => {
     // Use minimal record format (just entityId + recordId)
     const records = [
       { entityId: 'test:entity', recordId: 'rec-1' },
       { entityId: 'test:entity', recordId: 'rec-2' },
     ]
+    mockSearchIndexer.indexRecordsById.mockResolvedValueOnce({ indexed: 2, skipped: 0 })
     const job = createMockJob<FulltextIndexJobPayload>({
       jobType: 'batch-index',
       tenantId: 'tenant-123',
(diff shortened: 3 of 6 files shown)


adeptofvoltron:
🤖 \`om-auto-fix-issue\` — 🏷️ label rationale

- 🐛 \`bug\` + ⚡ \`performance\`: carried over from issue #5606 — closes a concrete implementation gap (batch-index job type not batching its writes), not a new capability.
- 🔺 \`priority-high\` / 🟡 \`risk-medium\`: carried over from the issue's own labels; blast radius is contained to the fulltext indexing worker's batch-index path, real-time single-record indexing is untouched.
- ✅ \`skip-qa\`: no `.tsx`/UI-rendering files touched, no DB schema or API surface change, no BACKWARD_COMPATIBILITY.md contract broken, and automated regression tests for the changed behavior ship in this same PR (per the automated-verification exemption).
- 📋 \`review\`: ready, non-draft PR with the full validation gate green (only unrelated pre-existing \`packages/cli\` flake).

adeptofvoltron:
🤖 `om-auto-review-pr` claiming this PR for review.

adeptofvoltron:
🤖 Adopting a stale chain lock (~2h11m) — previous run presumed dead.

🤖 `om-auto-review-pr` taking over the chain lock — code review. Started: 2026-08-25T12:16:29Z.

adeptofvoltron:
🤖 `om-auto-review-pr` — verdict: **CHANGES REQUESTED**. GitHub refuses a formal `request-changes` review from the PR author (`Can not request changes on your own pull request`), so the full `om-code-review` report is posted here as a comment instead. The `changes-requested` pipeline label carries the verdict; a second human reviewer is still required to formally approve this PR.

# 🔍 Code Review: fix(search): batch-index fulltext worker writes once, not once per record

## 🎯 Summary

This PR targets the second of the three code-level asks in issue #5606: the fulltext indexing worker's `batch-index` branch already received many record ids in a single job, but looped `searchIndexer.indexRecordById()` once per record, producing N separate `addDocuments([doc])` HTTP writes against Meilisearch even though the driver has had a real per-tenant `addDocuments(documents)` batch method all along. The change adds `SearchIndexer.indexRecordsById()`, which still loads every queued record fresh through the query engine (so index freshness is unchanged) but accumulates the built `IndexableRecord`s and flushes them through exactly one `searchService.bulkIndex()` call, and rewires the worker's `batch-index` branch to call it once instead of looping.

The core idea is right and the plumbing is correct where it matters most: `FullTextSearchStrategy.bulkIndex()` (`packages/search/src/strategies/fulltext.strategy.ts:65`) delegates to `MeilisearchDriver.bulkIndex()` (`packages/search/src/fulltext/drivers/meilisearch/index.ts:275`), which groups documents per tenant index and issues a single `addDocuments(documents, …)` — so for a batch job whose records all share one tenant, the Meilisearch write count really does collapse from N to 1, which is exactly what the issue asked for. The record-building path is a faithful reuse of the existing private `buildIndexableRecords()` helper rather than a reimplementation, so presenter/URL/links/checksum/custom-field resolution stays identical to the single-record path. The new unit test file is focused and asserts the right invariant (one `bulkIndex` call for N records, never a per-record loop), and the `workers.test.ts` updates tighten rather than weaken the existing assertions. Deferring the issue's debounce ask on the real-time write path was also the right instinct given `packages/search/AGENTS.md`'s "Ask First" rule on queue behavior.

What the change does not account for is that `searchService.bulkIndex()` is not a fulltext-only call. It fans out to **every** available strategy, and only two of the three implement a real `bulkIndex`. That turns this from a pure write-coalescing change into a concurrency change for the vector strategy, which is the one finding that needs addressing before merge. Reviewed scope: `packages/search/src/indexer/search-indexer.ts`, `packages/search/src/modules/search/workers/fulltext-index.worker.ts`, and the two test files, read against `packages/search/src/service.ts`, the three strategy implementations, the Meilisearch driver, the reindex lock/progress helpers, and issue #5606.

## Verdict

❌ **request changes** — one Major finding (the new single `bulkIndex()` call fans out to `VectorSearchStrategy`, which has no `bulkIndex` implementation, so `SearchService`'s fallback turns up to 200 previously-sequential embedding-provider calls into 200 concurrent ones per job) and one Minor finding (`Fixes #5606` will auto-close an issue whose headline regression path is explicitly left unimplemented). Under this repository's verdict rule, Medium/minor findings also request changes. Nothing here is a security, tenancy, or contract-surface problem, and no validation gate is failing — the fix for the Major finding is small and local.

## 🧪 Validation Gate

Per this repository's `om-auto-review-pr` extension, validation is **GitHub-checks-first**: CI already ran the full gate against this exact head, so the commands below were not re-run locally. **Validation source: GitHub checks** (head `3d4849e2128cc79c807003c50e9741bfae0c93ee`). All 24 reported checks pass; none are failing and none are pending, so no local fallback was warranted.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | Covered by the `prepare` check (3m31s), which builds packages before the downstream jobs. |
| `yarn generate` | ✅ PASS | Covered by `prepare`; no generated-file drift was reported. |
| `yarn build:packages` (second run) | ✅ PASS | Same `prepare` job; the rebuild-after-generate step reported no errors. |
| `yarn i18n:check-sync` | ✅ PASS | Covered by `lint` (31s). No locale files are touched by this diff. |
| `yarn i18n:check-usage` | ✅ PASS | Covered by `lint`. The diff adds no user-facing strings. |
| `yarn typecheck` | ✅ PASS | Covered by `lint`/`prepare`; the new `indexRecordsById` signature and its worker call site type-check. |
| `yarn test` | ✅ PASS | `test` check green in 4m28s, including the new `search-indexer-batch.test.ts` and the updated `workers.test.ts`. |
| `yarn build:app` | ✅ PASS | Covered by `prepare` / `docker-build` gating; no app build failure reported. |

Supplementary green checks on this head: `audit-scope`, `ds-lint`, `documents-multi-instance`, `scope`, `license/cla`, and all 15 `ephemeral-integration` shards.

## Findings

### ⚠️ Major

**`packages/search/src/indexer/search-indexer.ts:415` — the single `bulkIndex()` call fans out to the vector strategy, which has no batch implementation, converting up to 200 sequential embedding calls per job into 200 concurrent ones.**

`indexRecordsById()` ends with `await this.searchService.bulkIndex(allRecords)`. `SearchService.bulkIndex()` (`packages/search/src/service.ts:228-241`) iterates **every** available strategy and, for any strategy that does not implement `bulkIndex`, falls back to `Promise.all(records.map((record) => this.executeStrategyIndex(strategy, record)))` at `packages/search/src/service.ts:239`. `TokenSearchStrategy` implements `bulkIndex` (`packages/search/src/strategies/token.strategy.ts:168`) and `FullTextSearchStrategy` implements it (`packages/search/src/strategies/fulltext.strategy.ts:65`), but `VectorSearchStrategy` does **not** — it only has `index()` (`packages/search/src/strategies/vector.strategy.ts:131`), and that method calls `await this.embeddingService.createEmbedding(textContent)` on every record, which is a live HTTP request to the embedding provider (`packages/search/src/vector/services/embedding.ts:244`, an unthrottled `embed()` call with no queue or semaphore). `executeStrategyIndex` (`packages/search/src/service.ts:491`) adds no concurrency limiting either. All three strategies live on the one `searchService` instance the worker resolves (`packages/search/src/di.ts:248-292`), so the fulltext worker genuinely reaches the vector strategy.

The concrete failure: `reindexEntityToFulltext()` pages the query engine with `const pageSize = 200` (`packages/search/src/indexer/search-indexer.ts:726`) and enqueues one `batch-index` job per page (`packages/search/src/indexer/search-indexer.ts:773`), so a batch job carries up to 200 record ids. Before this PR, the worker's loop called `indexRecordById()` → `indexRecord()` → `searchService.index()` one record at a time and awaited each, so at most one embedding request was in flight per job (times `DEFAULT_CONCURRENCY = 2` workers). After this PR, a single `bulkIndex(200 records)` issues `Promise.all` over 200 `createEmbedding()` calls simultaneously. On any tenant that has vector search configured (`OPENAI_API_KEY` set, or Ollama reachable) and triggers a fulltext reindex, that is a 200-request burst per job against the embedding provider — provider-side 429 rate limiting, or pgvector connection-pool exhaustion on the upsert side. It gets worse on failure: `bulkIndex` is deliberately fail-loud and throws when any strategy rejects (`packages/search/src/service.ts:262-268`), so a single 429 fails the whole job, the queue retries it, and the retry re-embeds all 200 records from scratch — `VectorSearchStrategy.index()` computes the embedding *before* the checksum is used, so retries pay the full embedding cost again. This is precisely the "bounded concurrency for background jobs" rule in the review checklist, and it is a performance regression introduced by a performance fix.

Fix — bound the fan-out rather than reverting the batching. The cleanest option is to give `VectorSearchStrategy` a chunked `bulkIndex(records)` that processes records in small slices (matching however the vector worker already paces itself) so `SearchService` stops hitting the unbounded fallback for it. Alternatively, cap the fallback itself in `packages/search/src/service.ts:239` by replacing the bare `Promise.all` with a small bounded-concurrency map (a fixed in-flight limit, e.g. 4-8), which fixes every present and future strategy that lacks `bulkIndex` in one place and is the smaller diff. Either way, add a unit test asserting that a batch of N records against a `bulkIndex`-less strategy never exceeds the concurrency cap in flight.

### 🔹 Minor

**PR body — `Fixes #5606` will auto-close an issue whose primary ask is explicitly left unimplemented.**

Issue #5606's "How to implement" lists three code items. Item 2 (the `batch-index` branch) is what this PR fixes. Item 1 — batching/debouncing the real-time sync path so concurrent writes stop issuing one `addDocuments([doc])` per record — is the one the issue's own evidence attributes the reported regression to: the traced write path is "entity create/update → `search.index_record` subscriber enqueues a single job `{ jobType: 'index', … }` → worker's `jobType === 'index'` branch", and the p50 72ms → 984ms / p99 275ms → 8.75s numbers were measured under concurrent catalog/order writes, not during a reindex. The PR body is commendably transparent about deferring it and cites the correct `packages/search/AGENTS.md` "Ask First" rule for doing so — but the `Fixes #5606` keyword still closes the issue on merge, which buries the unfixed headline problem along with all its measurement evidence.

Fix: change `Fixes #5606` to `Refs #5606` in the PR body so the issue stays open, or open the follow-up issue for the debounce work now and reference it from both the PR body and #5606 before merging. Item 3 of the issue (a test asserting one `addDocuments()` call) is satisfied in spirit — the new test asserts one `searchService.bulkIndex()` call, one layer above the driver, and the strategy/driver chain that turns it into one `addDocuments()` is already covered elsewhere — and item 4 (the load-test re-verification) is reasonably left to post-merge.

### 💅 Nit

**`packages/search/src/indexer/search-indexer.ts:346-350` — the new doc comment overstates what the method guarantees.**

The comment claims the change makes "N queued records become exactly one write per available strategy instead of N". That is true for fulltext and tokens but false for any strategy without a `bulkIndex` implementation — vector still gets N writes, just concurrent ones (see the Major finding). Reword to say the batch is flushed through a single `searchService.bulkIndex()` call, and that per-strategy write coalescing depends on the strategy implementing `bulkIndex`.

**`packages/search/src/indexer/search-indexer.ts:395` — the `dropped` count from `buildIndexableRecords()` is discarded, so the returned totals do not add up.**

`buildIndexableRecords()` returns `{ records, dropped }` and increments `dropped` for any item without an `id` (`packages/search/src/indexer/search-indexer.ts:1141-1146`); the sibling call site at `packages/search/src/indexer/search-indexer.ts:758` captures it and rolls it into `result.recordsDropped`. Here only `records` is destructured, so a dropped item is counted in neither `indexed` nor `skipped`, and the worker's `Indexed ${successCount}/${records.length}` status log (`packages/search/src/modules/search/workers/fulltext-index.worker.ts:250`) silently under-reports with no accompanying reason. It should be rare given the query filters by id, but the accounting hole is free to close: fold `dropped` into `skipped`.

**`packages/search/src/indexer/search-indexer.ts:352-356` — the new public method uses a positional signature where every sibling uses a params object.**

`indexRecord(params)`, `indexRecordById(params)`, `deleteRecord(params)`, `purgeEntity(params)`, `reindexEntity(params)` and `reindexEntityToFulltext(params)` all take a single named-params object; `indexRecordsById(items, tenantId, organizationId?)` breaks that pattern on a newly exported surface, where the third positional argument is the easiest to pass wrongly. This is additive so it breaks nothing today, but it is much cheaper to align now than after third-party modules start calling it. Suggest `indexRecordsById({ items, tenantId, organizationId })`.

**`packages/search/src/modules/search/workers/fulltext-index.worker.ts:222` and `:252` — `skipped` is renamed to `failCount` in the operator-facing status log, conflating benign skips with errors.**

`indexRecordsById` returns `skipped` for three distinct reasons: the entity is not configured for search, the record no longer exists, or the load actually threw. The worker binds all three to `failCount` and writes that into `recordIndexerLog`'s `details`. Previously `failCount` counted only thrown errors — an `{ action: 'skipped' }` result incremented neither counter — so an operator reading the indexer status log for a reindex that legitimately raced a few deletions will now see them reported as failures. Either rename the destructured binding to `skippedCount` in the log details, or have `indexRecordsById` return the error count separately from the benign-skip count.

**`packages/search/src/modules/search/workers/fulltext-index.worker.ts:222` — the batch path's failure semantics changed from swallow-and-continue to throw-and-retry, which is neither mentioned in the PR body nor covered by a test.**

The old loop wrapped each `indexRecordById()` in a try/catch, so a Meilisearch outage produced a "successful" job with `failCount === N` and a permanently missing index entry. Now a `bulkIndex` rejection propagates out of `indexRecordsById`, past `advanceFulltextReindexProgress`, into the worker's outer catch, and is re-thrown for queue retry. This is the better behavior and it aligns with the direction of #1507 ("bulkIndex silently swallows strategy failures") and the `packages/search/AGENTS.md` rule that writes stay fail-loud for available strategies, so no change is requested here — but it is a user-visible reliability change worth a line in the PR body. The blast radius is contained: because the progress advance is now skipped on failure, a permanently failing batch stops advancing reindex progress, but `getReindexLockStatus` auto-cleans a lock whose heartbeat is older than `HEARTBEAT_STALE_MS` (`packages/search/src/modules/search/lib/reindex-lock.ts:19-20, 89-90`), so the reindex lock self-heals rather than wedging. A worker test asserting that a throwing `indexRecordsById` leaves `updateReindexProgress` uncalled and re-throws would lock this in.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `indexRecordsById` is purely additive; `indexRecordById` remains exported and is still used by the single-record job branch and the vector worker.
- [x] No function signature changed in a breaking way — no existing signature was touched.
- [x] No required type field removed or narrowed — `IndexableRecord`, `FulltextIndexJobPayload`, and the job types are unchanged.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — no API routes in the diff.
- [x] No field removed or retyped in an existing response shape — the indexed-document shape is produced by the same `buildIndexableRecords()` helper as before.
- [x] No event or message name renamed or removed; no payload field removed — the `batch-index` job type and its `records`/`tenantId`/`organizationId` payload are untouched, so in-flight queued jobs keep working across the deploy.
- [x] Where a contract had to change: not applicable — no contract surface changed.

One observation that is not a violation: records built through the batch path now carry `organizationId: null` where the old single-record path passed `undefined` for an absent organization, because `buildIndexableRecords` is called with `organizationId ?? null`. This matches what `reindexEntity` (`packages/search/src/indexer/search-indexer.ts:508`) already stored via the same helper, so the batch path is now consistent with the bulk path rather than diverging from it, and both the Meilisearch and token strategies normalize a null organization the same way they normalize an absent one.

## 🧪 Test Coverage

The new `packages/search/src/__tests__/search-indexer-batch.test.ts` covers the change's core invariant well and at the right level. Its first case queues three records across one entity and asserts `searchService.bulkIndex` was called exactly once, with the full three-record array and the correct `tenantId`/`organizationId` on each — that is the regression guard the issue asked for, and it would fail against the old looping implementation. The second case covers a record that no longer exists and asserts the batch write still happens for the survivors with `{ indexed: 1, skipped: 1 }`, which pins the "one dead record does not poison the batch" behavior. The third asserts that an entirely unconfigured entity produces no `bulkIndex` call at all, guarding the empty-batch early return at `packages/search/src/indexer/search-indexer.ts:414`. The `workers.test.ts` updates are a strict tightening: the batch-index test now asserts `indexRecordsById` was called exactly once *and* that `indexRecordById` was never called, so a regression to per-record looping fails loudly, and the two progress tests keep their `updateReindexProgress`/`incrementReindexProgress`/`clearReindexLock` assertions intact.

Gaps to close, in priority order:

1. **Bounded concurrency for `bulkIndex`-less strategies** — the Major finding has no test at all. Add a case to `packages/search/src/__tests__/` driving `SearchService.bulkIndex()` with a stub strategy that has no `bulkIndex` and an `index()` that records peak in-flight count, asserting the peak stays at or below the chosen cap for a batch of, say, 50 records. This is the test that would have caught the regression.
2. **Multi-entity batching** — every case in the new file uses a single `entityId`, so the `recordIdsByEntity` grouping at `packages/search/src/indexer/search-indexer.ts:361-366` is never exercised with more than one key. The PR description claims "N queued records across one or more entities collapse into exactly one `searchService.bulkIndex()` call", but only the one-entity half is actually asserted. Add a case with two configured entities interleaved in the `items` array, asserting a single `bulkIndex` call containing all records from both.
3. **Load-failure isolation** — no case covers the `catch` at `packages/search/src/indexer/search-indexer.ts:403`. Add one where the query engine rejects for a single record id and assert the remaining records are still written in one `bulkIndex` call with `skipped` incremented, so the per-record error isolation cannot silently regress into an aborted batch.
4. **Write-failure propagation** — no case covers `searchService.bulkIndex` rejecting. Add a worker-level test asserting that a throwing `indexRecordsById` leaves `updateReindexProgress` uncalled and re-throws for queue retry, which is the failure-semantics change called out in the last Nit.
5. **Record-load parity with `indexRecordById`** — worth one assertion that the query passed to the engine carries `includeCustomFields: true` and `skipAutoReindex: true`, since dropping either would silently degrade index content or reintroduce the reindex feedback loop that #520 fixed, and nothing currently pins them.

adeptofvoltron:

*(Supersedes the previous rationale: the review's findings have since been fixed and pushed, so the pipeline label moved from `changes-requested` back to `review`.)*

- 🔍 `review` — the review's findings (one Major, one Minor, five Nits) were all fixed in `4042b01c` and pushed to the PR head, so the `changes-requested` state no longer applies. It moves to `review` rather than `merge-queue` because GitHub does not permit this account to approve its own pull request: no independent approval exists, `reviewDecision` is still `REVIEW_REQUIRED`, and labelling it `merge-queue` would assert a review gate that never actually cleared. A second human reviewer is required.
- 🐛 `bug` + `performance` — carried over unchanged; the PR closes a concrete implementation gap in the fulltext indexing worker rather than adding a capability.
- 🔺 `priority-high` — kept as inherited from issue #5606; the underlying report is a partner-visible search-latency regression, and the review did not find the scope mis-rated.
- 🟡 `risk-medium` — kept. The follow-up commit touches `SearchService.bulkIndex`, which every strategy shares, but the change only bounds concurrency in the existing no-`bulkIndex` fallback and is covered by a test that fails without it.
- ⏭️ `skip-qa` — kept and re-verified against the automated-verification exemption: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no database or API surface change, no `BACKWARD_COMPATIBILITY.md` contract broken, and automated regression tests for the changed behavior ship in this same PR.
- ⏱️ `ci-monitoring` — the review run's work is finished and reported; only the CI result for the new head `4042b01c` is still being watched. This is not a claim, so the PR is free for anyone to pick up.

adeptofvoltron:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED → all findings fixed. Lock released.

## Autofix summary — 1 iteration, pushed as `4042b01c`

Every finding from the review above is now addressed in the PR head. Test suite for `@open-mercato/search`: **340 passed / 340**, up from 333 — 7 tests added, none removed or weakened.

**⚠️ Major — bounded the `bulkIndex` per-record fallback.** `SearchService.bulkIndex()` fanned the whole batch out to every strategy and, for any strategy without a `bulkIndex` implementation, ran `Promise.all` over all of them. `VectorSearchStrategy` is exactly that case and its `index()` performs an embedding-provider round trip per record, so a 200-record reindex page burst 200 concurrent provider calls from one job. The fallback now runs at a fixed in-flight cap (`BULK_INDEX_FALLBACK_CONCURRENCY = 4`) via a small `mapWithConcurrency` helper that preserves `Promise.all`'s reject-on-first-error semantics. Fixing it in `SearchService` rather than in `VectorSearchStrategy` covers every present and future strategy that lacks a batch implementation. The new `service.test.ts` case measures peak in-flight writes across 50 records and asserts it stays at or below the cap — verified load-bearing by temporarily raising the cap to 50, which makes it fail with `Expected: <= 4, Received: 50`. The chosen cap is deliberately conservative: the vector worker paces its own batches strictly sequentially (`vector-index.worker.ts`), so 4 stays close to established behavior while still leaving the fulltext write fully coalesced.

**🔹 Minor — `Fixes #5606` → `Refs #5606`.** The PR body now uses `Refs` and states plainly that this PR implements step 2 of the issue while step 1 (batching/debouncing the real-time per-record write path, which the issue's latency measurements attribute the regression to) is intentionally out of scope. Issue #5606 therefore stays open on merge instead of being auto-closed with its headline problem unfixed, and needs no separate follow-up issue to stay tracked.

**💅 Nits — all five applied.**
- `indexRecordsById` now takes an `IndexRecordsByIdParams` object, matching `indexRecord`, `indexRecordById`, `deleteRecord`, `purgeEntity` and `reindexEntity`. Free to change now since the method is new in this PR and unreleased.
- `buildIndexableRecords`' `dropped` count is folded into `skipped`, so `indexed + skipped` always equals the number of queued items.
- The worker reports benign skips as `skippedCount` in the indexer status log instead of `failCount`, so a reindex that races a few deletions no longer reads as failures to an operator.
- The method doc comment no longer claims one write per strategy unconditionally; it now says the batch is flushed through a single `bulkIndex()` call and that per-strategy coalescing depends on the strategy implementing `bulkIndex`.
- The failure-semantics change (swallow-and-continue → throw-and-retry) is now pinned by a worker test asserting a failed batch write re-throws without advancing reindex progress.

**Test-coverage gaps from the review, all closed:** bounded concurrency for `bulkIndex`-less strategies, multi-entity batching in one `bulkIndex` call, per-record load-failure isolation, `bulkIndex` rejection propagation, records dropped for a missing id counting as skipped, and query-shape parity (`includeCustomFields: true` + `skipAutoReindex: true`).

## ⚠️ Two things a human needs to know

1. **This PR still needs an independent reviewer.** GitHub rejects a formal `request-changes` or `approve` review from the PR author (`Can not request changes on your own pull request`), and this automation shares that account. The review verdict was therefore delivered as a comment plus labels, and the fixes were made by the same actor that wrote both the PR and the review. `reviewDecision` is still `REVIEW_REQUIRED` and no approval exists, which is why the label moved to `review` rather than `merge-queue` — treat the automated pass as a first opinion, not a cleared gate.
2. **Local validation was partial.** Per this repo's GitHub-checks-first rule the full gate was not re-run locally, and the isolated review worktree has no `node_modules`, so `tsc` could not resolve cross-package imports there. Evidence gathered: the full `@open-mercato/search` jest suite passing (340/340), and a scoped `tsc` run confirming zero errors in the six touched files (the 50 errors it did report are all unresolved `@open-mercato/*` / `@tanstack/*` imports from the missing install, in files this PR does not touch). CI on head `4042b01c` is the authoritative gate.

adeptofvoltron:

**All checks green on head `4042b01c`** — 23 passed, 6 skipped, 0 failed, 0 pending. Settled 27 minutes after the autofix commit was pushed, within this run's CI budget.

This is the authoritative confirmation of the two gaps the review flagged in its own local evidence:

- **`prepare` (3m35s) and `lint` (31s)** ran the real workspace typecheck and build with a full install, covering the cross-package imports that the isolated review worktree could not resolve (it has no `node_modules`, so the local `tsc` run could only confirm zero errors within the six touched files). The `SearchService.bulkIndex` signature change and the new `IndexRecordsByIdParams` type both typecheck cleanly across the monorepo.
- **`test` (4m32s)** ran the full unit suite including the 7 tests added by the autofix commit — the peak-in-flight assertion on the bounded `bulkIndex` fallback, multi-entity batching, per-record load-failure isolation, `bulkIndex` rejection propagation, the missing-id drop accounting, query-shape parity, and the worker-level retry-semantics test.

All 15 `ephemeral-integration` shards also passed, as did `Mutation tests`, `ds-lint`, `scope`, `audit-scope` and `documents-multi-instance`. Notably, the pre-existing `packages/cli` `dev-env-reload.test.ts` flake mentioned in the original PR description did **not** recur here.

`ci-monitoring` is now removed — no further follow-up will come from this agent.

**The PR is not cleared to merge.** `reviewDecision` is still `REVIEW_REQUIRED` and no approval exists, because GitHub rejects both `approve` and `request-changes` from the PR author and this automation shares that account. Green CI is not a substitute for that gate: the same actor authored the PR, the review, and the fixes, so the automated pass is a first opinion rather than independent sign-off. The pipeline label stays `review` — a human reviewer needs to take it from here.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-25T23:57:33Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review: fix(search): batch-index fulltext worker writes once, not once per record

## 🎯 Summary

This PR implements step 2 of issue #5606: the fulltext indexing worker's `batch-index` branch no longer loops `searchIndexer.indexRecordById()` once per queued record (one Meilisearch `addDocuments([doc])` HTTP write each). It adds `SearchIndexer.indexRecordsById()` (`packages/search/src/indexer/search-indexer.ts:361-425`), which still loads every record fresh through the query engine but accumulates the built `IndexableRecord`s and flushes the whole batch through a single `searchService.bulkIndex()` call, and rewires the worker's `batch-index` branch to call it once (`packages/search/src/modules/search/workers/fulltext-index.worker.ts:220-225`). A second commit bounds the `SearchService.bulkIndex()` per-record fallback at four in-flight writes (`packages/search/src/service.ts:34-58, 267-276`) so a strategy without a `bulkIndex` implementation cannot turn one batch job into an unbounded burst.

I reviewed all six changed files against the PR head `4042b01c`, plus the surrounding call graph: `SearchService.index`/`bulkIndex`, `SearchIndexer.indexRecord`/`indexRecordById`/`buildIndexableRecords`/`reindexEntity`, all three strategies, and the Meilisearch driver's `index`/`bulkIndex`/`prepareDocument`.

**The change is correct and well-covered.** The three things I specifically went looking for all check out. First, fan-out parity: the old path (`indexRecordById` → `indexRecord` → `searchService.index`) already wrote to *every* available strategy, and `bulkIndex` fans out over the same `getAvailableStrategies()` list, so no strategy gains or loses writes. Second, document parity: `buildIndexableRecords` reproduces `indexRecord`'s build logic field for field — same `cf:`/`cf_` custom-field extraction, same `buildSource` → `formatResult` → `resolveUrl` → `resolveLinks` order, same `noReindexQueryEngine` in the build context — and both write paths funnel through the same `prepareDocument`, so the encrypted-field exclusion (`SEARCH_EXCLUDE_ENCRYPTED_FIELDS`) stays fail-closed on the batch path too. Third, tenant scoping: `indexRecordsById` passes `tenantId`/`organizationId` into every `queryEngine.query()` call, a batch job carries exactly one tenant, and the driver still groups documents per tenant index — no cross-tenant surface is opened.

Two things a maintainer should carry away before merging. **This does not fix the production regression #5606 reports.** `batch-index` fulltext jobs are enqueued from exactly one place — `SearchIndexer.reindexEntity()` with `useQueue: true` (`packages/search/src/indexer/search-indexer.ts:778-786`) — which the repo's own architecture spec confirms is "the only fulltext bulk path" (`.ai/specs/2026-07-24-search-architecture-clarification-and-evolution.md:65`). The p50 984ms / p99 8.75s evidence in the issue comes from the real-time `jobType: 'index'` path driven by `fulltext_upsert.ts`, which this PR deliberately leaves alone. So the win here is operator-triggered reindex throughput, and #5606 must stay open — the PR body says so, and it is right to. **Second, the bounded fallback is the wider-reaching half of the change**: it alters `SearchService.bulkIndex()` for every caller, not just this worker (see the Nit section).

The commit split is clean, the tests are real (they would fail if the batching were reverted rather than merely passing alongside it), and the scope decision to defer the debounce work — citing `packages/search/AGENTS.md`'s "Ask before changing search strategy defaults, entity IDs, queue behavior" — is the right call rather than guessing at debounce-window semantics on the hot write path.

## Verdict

✅ **approve** — no blocker, major, or minor findings. The batching is behaviour-preserving on document content, fan-out, and tenant scoping; the one real behaviour change (batch write failures now fail the job so the queue retries, instead of being swallowed per record) is intentional, explicitly tested, and better aligned with the module's documented "writes stay fail-loud for available strategies" contract than the code it replaces. Everything below is a nit: documentation and follow-up polish, all the author's call.

## 🧪 Validation Gate

Validation source: **GitHub checks** on head `4042b01c` (CI run [32847720639](https://github.com/open-mercato/open-mercato/actions/runs/32847720639)) — 23 checks passed, 6 skipped, 0 failed, 0 pending, including the 15 `ephemeral-integration` shards. Per this repo's review policy, CI check results stand in for re-running the gate locally; the one configured command CI does not cover was run locally against this worktree at the PR head.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job, 3m35s (`ci.yml:167`). |
| `yarn generate` | ✅ PASS | `prepare` job (`ci.yml:175`). |
| `yarn build:packages` (second run) | ✅ PASS | `prepare` job (`ci.yml:178`) — the rebuild over the files `yarn generate` produced. |
| `yarn i18n:check-sync` | ✅ PASS | Not covered by any CI job, so run locally at the PR head: "All translation files are in sync" across 5 locales / 57 modules, exit 0. |
| `yarn i18n:check-usage` | ✅ PASS | `test` job (`ci.yml:480`). |
| `yarn typecheck` | ✅ PASS | `test` job (`ci.yml:498`). |
| `yarn test` | ✅ PASS | `test` job, 4m32s (`ci.yml:510`), plus `test:scripts` and `test:repo-wide-guards`. |
| `yarn build:app` | ✅ PASS | `prepare` job (`ci.yml:190`). |

Also green and worth noting: `lint` (31s), `ds-lint` (44s), `audit-scope`, `documents-multi-instance`, and all 15 integration shards.

## Findings

### 💅 Nit

**1. `packages/search/src/service.ts:267-276` — the bounded fallback changes behaviour for three existing callers, and the PR body does not mention them.** The description says the driver "already has a working `bulkIndex()` path … that nothing used". That is true of the *fulltext batch worker*, but `SearchService.bulkIndex()` itself already has production callers: `SearchIndexer` at `search-indexer.ts:524` and `search-indexer.ts:650` (`bulkIndexRecords`), and two in another package — `packages/ai-assistant/src/modules/ai_assistant/lib/tool-search.ts:209` and `.../lib/entity-index.ts:92`. Replacing the unbounded `Promise.all` with a concurrency of four therefore reshapes vector-strategy write concurrency for all of them, not just this worker. I believe that is a strict improvement in every case — `VectorSearchStrategy` is the only strategy without a `bulkIndex` implementation (`vector.strategy.ts:131-139` does an embedding-provider round trip per record), so the bound only ever throttles the expensive path — but the blast radius belongs in the PR body, because `risk-medium` reads differently once "one worker branch" becomes "every `bulkIndex` caller in two packages". Suggested fix: one sentence in the Summary naming those callers.

**2. `packages/search/src/service.ts:39-58` — `mapWithConcurrency`'s doc comment overstates the `Promise.all` parity.** The comment says it "Rejects with the first error, matching Promise.all semantics." The rejection propagation does match, but `Promise.all` never *starts* work after a rejection — every promise was already created — whereas here the surviving runners keep pulling new items off `nextIndex` after a sibling has thrown. For the vector fallback that means embedding round trips continue to be dispatched for a batch that has already failed and will be retried wholesale by the queue. Suggested fix: either set a `failed` flag that the runner loop checks before pulling the next index (turning it into genuine fail-fast), or reword the comment to say that in-flight and subsequently-scheduled work is not cancelled. Note this is not a regression — the previous unbounded `Promise.all` had already dispatched everything — so it is purely a polish call.

**3. `packages/search/src/modules/search/workers/fulltext-index.worker.ts:220-225` — the queue-retry semantics changed, and only the tests record it.** Previously a write failure for a single record was caught inside the worker loop, counted into `failCount`, and the job still completed successfully with progress advanced. Now a `bulkIndex` rejection propagates out of `indexRecordsById`, the worker re-throws, and the queue retries the whole batch — exactly what the new test at `workers.test.ts:553-580` asserts. I think this is the better behaviour and it matches both `SearchService.throwOnStrategyFailures`' documented rationale (issue #3103, "swallowing rejections causes silent, permanent index gaps") and `packages/search/AGENTS.md` ("Writes … stay fail-loud for available strategies"). But `packages/search/AGENTS.md` also lists "queue behavior" under **Ask First**, and the PR body — scrupulous about that same rule when deferring the debounce work — says nothing about it. The operational consequence is worth a line for whoever is on call: a batch that fails permanently now holds its slice of reindex progress until the 30-second heartbeat staleness cleanup in `reindex-lock.ts` reaps the lock, instead of completing with a logged `failCount`. Suggested fix: add that sentence to the PR body (and ideally to #5606), no code change needed.

**4. `packages/search/AGENTS.md` — the new public method is not documented where its sibling is.** The "Use SearchIndexer for Config-Aware Indexing" section documents `indexRecordById` with a usage example, and `indexRecordsById` is now a peer public method on the same class with a new exported `IndexRecordsByIdParams` type. Both are additive, so there is no compatibility issue, but a third-party module author reading that section will not discover the batch variant. Suggested fix: a four-line example alongside the existing `indexRecordById` one.

**5. `packages/search/src/indexer/search-indexer.ts:376-411` and `:766-786` — the load side is still the N-round-trip half, and the reindex path builds every record twice.** This PR batches the *write*; the *load* is still one `queryEngine.query()` per record, executed sequentially. Worth a follow-up because the double-work above it is stark: `reindexEntity` already calls `buildIndexableRecords` over the whole page (`:766-772`) — running `buildSource`, `formatResult`, `resolveUrl` and `resolveLinks` for every record — and then discards the result, enqueueing only `{entityId, recordId}` pairs (`:783`) that the worker re-loads and re-builds from scratch. So each reindexed record pays for its presenter resolution twice and its DB load twice. Either passing the built records through the job payload or having the worker load a page in one `id IN (...)` query would cut materially more reindex latency than the write batching does. Explicitly out of scope for this PR — just flagging it as the natural next step, since the remaining reindex cost now lives entirely on that side.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the change is purely additive: new `IndexRecordsByIdParams` type and new `SearchIndexer.indexRecordsById()` method. `indexRecordById` is untouched and still used by the `jobType: 'index'` branch.
- [x] No function signature changed in a breaking way — `SearchService.bulkIndex(records: IndexableRecord[]): Promise<void>` keeps its exact signature; only its internal fallback scheduling changed.
- [x] No HTTP route URL removed or renamed; no method changed — no API route files in the diff.
- [x] No field removed or retyped in an existing response shape — the indexed Meilisearch document is byte-identical, since both paths go through the same `mapRecordToDocument` and `prepareDocument`.
- [x] No event or message name renamed or removed; no payload field removed — the `FulltextIndexJobPayload` shape for `batch-index` is unchanged, so jobs already sitting in the queue when this deploys are still processed correctly by the new branch.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — the only output change is the `details` key on the `worker:fulltext:batch-index` indexer log, where `failCount` became `skippedCount`. That log is a diagnostic record rather than a parsed contract, and the rename is honest: the new value counts records skipped during load, which is not what `failCount` meant.
- [x] No database table or column renamed or removed; no column type narrowed — no migration in the diff and none needed.
- [x] No config key renamed and no default changed silently — `BULK_INDEX_FALLBACK_CONCURRENCY` is a new internal constant, not a config surface.
- [x] Where a contract had to change: not applicable, nothing protected by `BACKWARD_COMPATIBILITY.md` was touched.

## 🧪 Test Coverage

Coverage is genuinely good — these tests pin the behaviour rather than merely coexisting with it.

The new `packages/search/src/__tests__/search-indexer-batch.test.ts` (8 cases) covers the core claim and its edges: N records collapsing into exactly one `bulkIndex` call with the right record contents (`:33`), records spanning two entity types still collapsing into one call (`:63`) — which is the case a naive per-entity implementation would have gotten wrong — the query being issued with `includeCustomFields: true` and `skipAutoReindex: true` (`:89`), a record that no longer exists being skipped without failing the batch (`:112`), a load throw for one record not stranding the other two (`:132`), a `bulkIndex` rejection propagating so the queue retries (`:161`), an unconfigured entity never reaching `bulkIndex` at all (`:176`), and a record whose `id` is missing being counted into `skipped` so the totals reconcile (`:192`). The assertions check call *contents*, not just call counts, so swapping the batch for a loop would fail them.

`workers.test.ts` was updated rather than weakened: the batch-index test now asserts `indexRecordsById` is called exactly once **and** that `indexRecordById` was never called (`:518`), which is the assertion that actually forecloses a regression back to the loop. The new case at `:553` pins the retry semantics from nit 3 — write failure re-throws and neither `updateReindexProgress` nor `incrementReindexProgress` runs.

`service.test.ts:434` pins the concurrency bound by instrumenting in-flight count and asserting a peak of ≤ 4 over 50 records while still confirming all 50 `index` calls happen — a real measurement rather than a proxy.

One gap, consistent with nit 1: there is no test covering the bounded fallback from the perspective of the *other* `bulkIndex` callers (`bulkIndexRecords`, and the `ai-assistant` entity index). The existing service-level test does cover the shared code path they all funnel through, so I would not ask for more; I mention it only because the PR's own framing understates who is affected.

---

🤖 Reviewed by `om-auto-review-pr` (independent reviewer pass — the earlier review on this PR was posted as a comment by the author's own account, which GitHub does not accept as a formal review).

pkarw:

*(Supersedes the previous rationale comment, which was posted by the PR author's own account and cannot be edited from this one. This run is an independent reviewer pass by @pkarw.)*

- 🚀 `merge-queue` — the code review passed with no blocker, major, or minor findings, and the PR now carries a formal GitHub approval from a reviewer who is not the author, so `reviewDecision` moved from `REVIEW_REQUIRED` to `APPROVED`. It replaces `review` because the review gate is genuinely satisfied rather than merely re-requested.
- 🐛 `bug` — kept as carried over from issue #5606: this closes a concrete implementation gap (the `batch-index` job type not batching its writes) rather than adding a capability.
- ⚡ `performance` — kept: the whole change is a write-amplification fix, replacing N `addDocuments([doc])` calls per reindex batch with one `addDocuments(docs)`.
- 🤝 `partner-request` — kept, carried over from the originating issue.
- ⏭️ `skip-qa` — kept, and it verifies against this repo's automated-verification exemption: the diff touches no `.tsx` file, nothing under `packages/ui/src/` or any `components/` directory, changes no database structure and no API surface, breaks nothing listed in `BACKWARD_COMPATIBILITY.md`, and ships automated regression tests for the changed behaviour in the same PR.
- 🔺 `priority-high` — kept as carried over from #5606, which is a partner-reported latency regression. Worth stating plainly, though: this PR fixes the operator-triggered reindex path only, since `SearchIndexer.reindexEntity()` is the sole enqueuer of fulltext `batch-index` jobs. The real-time write path the issue's p99 evidence implicates is untouched, so #5606 must stay open after this merges.
- 🟡 `risk-medium` — kept, and it is the right rating rather than `risk-low`, because the second commit changes `SearchService.bulkIndex()` itself. That method already had callers beyond this worker (`search-indexer.ts:524`, `search-indexer.ts:650`, and two in `packages/ai-assistant`), so the bounded fallback reaches further than the one worker branch the title describes. It stays below `risk-high` because no auth, data-scoping, schema, encryption, or shared contract surface is touched, and the bound only ever throttles the vector strategy's embedding round trips.

pkarw:

Independent reviewer pass by @pkarw on head `4042b01c` — a formal GitHub approval, which the earlier same-account pass could not submit. No blocker, major, or minor findings; five nits are listed in the review, all the author's call and none blocking:

1. The PR body's "nothing used" framing understates the blast radius — `SearchService.bulkIndex()` already had four callers across two packages, and the bounded fallback changes behaviour for all of them.
2. `mapWithConcurrency`'s "matching Promise.all semantics" comment overstates the parity: surviving runners keep dispatching new work after the first rejection.
3. The batch-index retry semantics changed (write failures now fail the job instead of being swallowed into `failCount`) — intentional, tested, and better aligned with the module's fail-loud write contract, but `packages/search/AGENTS.md` lists queue behaviour under "Ask First" and the PR body does not mention it.
4. `packages/search/AGENTS.md` documents `indexRecordById` but not the new public `indexRecordsById`.
5. Follow-up: the load side is still N sequential queries, and `reindexEntity` builds every record twice — that is where the remaining reindex latency now lives.

Validation source: GitHub checks on `4042b01c` (run 32847720639) — 23 passed, 6 skipped, 0 failed, 0 pending, including all 15 integration shards. The one configured gate CI does not cover, `yarn i18n:check-sync`, was run locally at the PR head and passed. CI was already settled before this review started, so no CI follow-up is outstanding and `ci-monitoring` was not applied.

autofix: skipped (not my PR — re-run with `--autofix` to fix it here). Nothing needed fixing in any case.

⚠️ Merge note for whoever picks this up: this implements step 2 of #5606 only. `SearchIndexer.reindexEntity()` is the only enqueuer of fulltext `batch-index` jobs (confirmed by `.ai/specs/2026-07-24-search-architecture-clarification-and-evolution.md:65`), so the real-time write path behind the issue's p50 984ms / p99 8.75s evidence is unchanged. Do not close #5606 on this merge.

Issue #2967, opened 2026-06-10
perf(query_index): process-scope engine metadata caches and once-guard per-request event wiring


haxiorz:
The `query_index` DI registrar builds fresh `BasicQueryEngine` + `HybridQueryEngine` instances on every request container (`packages/core/src/modules/query_index/di.ts:16-58`), and registrars replay per request (`packages/shared/src/lib/di/container.ts:176-178`). Because `HybridQueryEngine`'s metadata caches live in instance fields (`engine.ts:131-140`), they die with the container every request — the 5-minute `QUERY_INDEX_CF_KEYS_CACHE_MS` TTL (`engine.ts:151-152`) and the per-instance `columnCache`/`customEntityCache` never survive a single request, so each backend list query re-runs ~8 `information_schema`/classification probes from cold. Separately, the registrar's `setup()` runs a `SELECT DISTINCT entity_id` per request and registers three fresh `bus.on` closures per entity type that the event bus cannot dedup, plus an unbounded `setTimeout(setup, 0)` retry loop.

**Affected code:**
- `packages/core/src/modules/query_index/di.ts:16-58` — `register()` constructs new Basic + Hybrid engines on every request container
- `packages/core/src/modules/query_index/di.ts:61-184` — `setup()` re-runs per request; line `64` retries `setTimeout(setup, 0)` forever if `eventBus` never resolves
- `packages/core/src/modules/query_index/di.ts:148-159` — `SELECT DISTINCT entity_id FROM custom_field_defs` executed per request
- `packages/core/src/modules/query_index/di.ts:165-167` — three `bus.on(...)` closures registered per entity type each call
- `packages/core/src/modules/query_index/lib/engine.ts:131-140` — instance-scoped `customFieldKeysCache`, `columnCache`, `customEntityCache`, `pendingCoverageRefreshKeys` (all per-request)
- `packages/core/src/modules/query_index/lib/engine.ts:151-152` — 5-min `QUERY_INDEX_CF_KEYS_CACHE_MS` TTL, effectively dead since the instance is rebuilt per request
- `packages/core/src/modules/query_index/lib/engine.ts:222`, `994-1024` — `isCustomEntity()` call site + definition (`customEntities` lookup with `custom_entities_storage` fallback); caches positive **and** negative results with no TTL
- `packages/core/src/modules/query_index/lib/engine.ts:246-248` — two `tableExists()` probes (`search_tokens`, base table) per query
[...]

PR #5614, opened 2026-08-25
fix(shared/query): promote BasicQueryEngine columnExists cache to process scope (#5605)

### LINKED ISSUE #5605: perf(shared/query): generic queryEngine's columnCache/tableCache never survive past one request
## Summary
- The generic `queryEngine` DI service (`BasicQueryEngine`) has a working `columnCache`/`tableCache`, but the engine instance is recreated on every HTTP request, so the cache never survives past a single request — tens of thousands of avoidable `information_schema` round-trips under load.

## Analysis
- Affected areas: `packages/shared/src/lib/query/engine.ts` (`BasicQueryEngine`), `packages/shared/src/lib/di/container.ts` (`createRequestContainer()`).
- Related but distinct from #2967: that issue covers `query_index`'s own internal `HybridQueryEngine`/`BasicQueryEngine` instances built in `query_index/di.ts` (lines 16-58) for custom-field-aware list queries. This issue is about the **separate**, generic `queryEngine` DI service every module gets via `container.resolve('queryEngine')` — instantiated at `packages/shared/src/lib/di/container.ts:213` (`queryEngine: asValue(new BasicQueryEngine(em, undefined, () => { ... }))`), unconditionally, on every request. Same class of bug, different instantiation site and consumer surface — worth fixing together but not the same code.
- Confirmed today: `columnCache`/`tableCache` are `private` instance fields on `BasicQueryEngine` (`engine.ts:208-209`, methods at `1118-1151`) — `columnExists()`/`tableExists()` check the cache, query `information_schema.columns`/`.tables` on miss, and (an additional inefficiency) only cache a `true` result — a `false` miss is deleted from the cache instead of remembered, so a persistently-absent column is queried indefinitely.
- Root cause: `container.ts`, `createRequestContainer()` (~line 213) does `new BasicQueryEngine(...)` on every call, i.e. on every HTTP request — a fresh, empty cache every time.
- Evidence: catalog module (`GET /api/catalog/products`, 24 requests) — 100 `information_schema.columns` calls, >4 per single product request. Independently, a customers-module load test (627 k6 iterations, full scale) produced ~50,600 total introspection calls (`information_schema.columns` column_name/data_type: 9,237 calls/4,108.8ms; exists-check: 22,897 calls/1,578.2ms; `information_schema.tables` exists-check: 18,474 calls/460.1ms), scaling proportionally with data volume. Confirmed independently on two different modules — a systemic problem in the generic layer, not module-specific.

## How to implement
1. Promote `columnCache`/`tableCache` to process scope (module-level `Map`, or a singleton service resolved once), not instance scope — mirroring the `BOOTSTRAP_CACHE_KEYS` precedent in `container.ts` for `eventBus`/`cache`/`kmsService`, but **without** the `OM_BOOTSTRAP_CACHE` gate, since column/table metadata doesn't depend on any per-request state and is always safe to reuse (unlike e.g. `tenantEncryptionService`, see #2963).
2. Also cache a `false` result (column/table absence) instead of deleting it from the cache on miss.
3. No runtime invalidation needed — schema metadata only changes on migration, and a process restart after a migration already clears the in-memory cache.
4. Add a unit test asserting `columnExists`/`tableExists` hit the cache (no second DB call) across two calls against the same process-scoped cache instance, and a regression test that a `false` result is remembered rather than re-queried.

## Compatibility notes
- Internal caching/wiring change only — no DI key, public type, or API route changes. `queryEngine` remains resolved the same way by every caller.

## Out of scope
- `query_index`'s own internal engine metadata caching (that's #2967) — this issue is scoped to the generic `queryEngine` DI service only.

## How to pick this up
- Run `om-auto-fix-issue {thisIssueNumber}`.

adeptofvoltron:
Fixes #5605

## 🎯 Goal
- Make `BasicQueryEngine`'s column-existence check actually cache across requests, eliminating tens of thousands of avoidable `information_schema.columns` round-trips under load — without turning a per-request structure into an unbounded, never-expiring one.

## 🔍 Problem
Confirmed on two different modules (catalog and customers): the generic `queryEngine` DI service pays a full `information_schema.columns` probe per column check on effectively every request, even though `BasicQueryEngine` already had caching logic for exactly this.

## 🔍 Root Cause
`createRequestContainer()` (`packages/shared/src/lib/di/container.ts`) constructs a fresh `BasicQueryEngine` on every HTTP request. Its `columnCache` was a `private` **instance** field, so the cache was always empty at the start of a request and discarded at the end of it — the memoization never had a chance to pay off across requests. Separately, `columnExists()` only cached `true` results; a `false` (column absent) was deleted from the cache instead of remembered, so a persistently-absent column was re-probed on every projection pass (`full`/`sortKeys`/`count`) within a single `.query()` call.

## What Changed
- `packages/shared/src/lib/query/engine.ts` — replaced the `private columnCache` instance field with a module-level `columnExistsCache`, and made `columnExists()` cache both `true` and `false` answers instead of deleting `false` misses.
- The shared cache is **bounded and TTL'd**, following the pattern this package already uses for the equivalent structure in `packages/shared/src/lib/search/availability.ts` (a `{ value, expiresAt }` entry, a max-entry cap that sweeps expired entries before clearing wholesale, and a TTL from an env knob):
  - `OM_QUERY_COLUMN_EXISTS_CACHE_MS` — default `300000` (5 min); `0` disables the memo and probes per request again.
  - `OM_QUERY_COLUMN_EXISTS_CACHE_MAX_ENTRIES` — default `10000`.
  - **Why the bound:** `columnExists` is not only called with the fixed scope columns. It is reached with caller-supplied field names through `resolveBaseColumn`, and those names arrive raw from the HTTP layer (`sortField`, unmapped filter-param keys), so an unbounded map would retain one entry per distinct name for the life of the process.
  - **Why the TTL:** a cached `false` is consumed exactly where the `organization_id`, `tenant_id` and `deleted_at IS NULL` predicates are applied, so a stale negative does not degrade into slow queries — it silently drops a scope guard. The TTL bounds that window, so a migration applied against a long-lived process (a running `yarn dev`, or pods not recycled by a separate migration release step) converges instead of staying invisible until restart. It keeps essentially all of the win: a hot column is probed twelve times an hour rather than tens of thousands.
  - Added `clearColumnExistsCache()` and `columnExistsCacheSize()` (test-only) so specs don't leak cache state into each other and the cap is assertable.
- `packages/shared/src/lib/query/__tests__/engine.test.ts` — regression tests for the module-scoped cache: (1) two separate `BasicQueryEngine` instances (simulating two requests) share the cache; (2) a missing column (`organization_id`) is probed exactly once per `.query()` call instead of once per projection; (3) a cached negative expires at the TTL, so a migrated-in `tenant_id` is picked up and the tenant guard reappears without a restart; (4) the entry count stays under the cap under distinct-sort-field pressure; (5) `OM_QUERY_COLUMN_EXISTS_CACHE_MS=0` disables the memo.
- `packages/shared/src/lib/query/__tests__/engine.scope-and-or.test.ts` and `engine.count-distinct.test.ts` — clear the module-scoped memo in `beforeEach`. Their fixtures declare mutually contradictory `information_schema.columns` shapes for the same `scheduled_jobs` table, so promoting the cache to module scope had silently made the first test's answers decide later tests' scoping. It passed only by ordering coincidence.

## ↪️ Out of scope (tracked separately)
- `packages/core/src/modules/query_index/lib/engine.ts` has its own, separate `columnCache` on a different `BasicQueryEngine`-family class (the one covered by #2967) — untouched here, per #5605's own note that it's "related but distinct."
- **The `information_schema.tables` half of #5605** is not in this PR. `BasicQueryEngine` has no `tableCache`/`tableExists` member — the issue's citation points at `HybridQueryEngine` — but the table-probe traffic it measured is real and comes from `DefaultDataEngine.ensureStorageTableExists()` (`packages/shared/src/lib/data/engine.ts`), which is constructed fresh per request right beside the query engine. That is a different class in a different subsystem, so it is tracked in **#5619** rather than widening this diff. #5605 has been explicitly re-scoped to the columns half in [a comment](https://github.com/open-mercato/open-mercato/issues/5605#issuecomment-5410491092).

## 🧪 Tests
- `yarn workspace @open-mercato/shared test src/lib/query/__tests__/engine.test.ts` — 44 passed / 44 total
- `yarn workspace @open-mercato/shared test src/lib/query src/lib/search` — 275 passed / 5 skipped, 280 total (no regressions from the module-level cache leaking across specs)
- `yarn build:packages` — green
- `yarn workspace @open-mercato/shared typecheck` — green
- `yarn lint` — green (0 errors)

## 💥 Breaking Changes
- None. Purely internal caching/wiring change — no DI key, public type, exported signature, or API route changed. `clearColumnExistsCache()` and `columnExistsCacheSize()` are new internal exports used only by the test suite, mirroring the existing `clearEncryptedLikeFieldsCache()` convention in the same file. The two new env knobs are additive with safe defaults; unset behaviour is the intended production behaviour.

## 🔐 Security impact
- None. The cache only ever stores a boolean (does column X exist on table Y), never row data, and is keyed by `table:column` — no tenant- or request-scoped values are cached across requests. The cap prevents request-driven key names from growing it without limit, and the TTL bounds how long a stale negative can suppress a tenant/organization scope predicate.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5614"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790246714&installation_model_id=432243&pr_number=5614&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5614&signature=b2ed9147148fca35e3539f55780e17e890bdfebf28e9aff331be5fffc37a6cf7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/shared/src/lib/query/__tests__/engine.count-distinct.test.ts
+++ b/packages/shared/src/lib/query/__tests__/engine.count-distinct.test.ts
@@ -1,5 +1,12 @@
-import { BasicQueryEngine } from '../engine'
+import { BasicQueryEngine, clearColumnExistsCache } from '../engine'
 import { registerModules } from '../../i18n/server'
+
+// The column-existence answer is memoized on the module, not the instance (#5605), so
+// the per-test fake schemas below would otherwise inherit whatever the first test
+// probed for the same table names.
+beforeEach(() => {
+  clearColumnExistsCache()
+})
 
 // One entity extension on auth:user so includeExtensions exercises the joined-aggregate path.
 registerModules([
--- a/packages/shared/src/lib/query/__tests__/engine.scope-and-or.test.ts
+++ b/packages/shared/src/lib/query/__tests__/engine.scope-and-or.test.ts
@@ -1,5 +1,15 @@
-import { BasicQueryEngine } from '../engine'
+import { BasicQueryEngine, clearColumnExistsCache } from '../engine'
 import { normalizeFilters } from '../join-utils'
+
+// The column-existence answer is memoized on the module, not the instance (#5605), and
+// the fixtures below declare mutually contradictory `information_schema.columns` shapes
+// for the same `scheduled_jobs` table. Without this clear, whichever test ran first
+// decides the scoping every later test sees — the automatic tenant/organization guard
+// would silently drop out of a query and the assertion for it would still be reading a
+// stale `false`.
+beforeEach(() => {
+  clearColumnExistsCache()
+})
 
 type FakeData = Record<string, any[]>
 
--- a/packages/shared/src/lib/query/__tests__/engine.test.ts
+++ b/packages/shared/src/lib/query/__tests__/engine.test.ts
@@ -2,13 +2,14 @@ import { BasicQueryEngine } from '../engine'
 import { SortDir } from '../types'
 import { registerModules } from '../../i18n/server'
 import { clearSearchTokenPresenceCache } from '../../search/availability'
-import { clearEncryptedLikeFieldsCache } from '../engine'
+import { clearEncryptedLikeFieldsCache, clearColumnExistsCache } from '../engine'
 
 // The token-presence answer is cached process-wide (TTL); without clearing it,
 // probe-count assertions would observe hits from earlier tests in this file.
 beforeEach(() => {
   clearSearchTokenPresenceCache()
   clearEncryptedLikeFieldsCache()
+  clearColumnExistsCache()
 })
 
 // Mock modules with one entity extension
@@ -1413,5 +1414,58 @@ describe('BasicQueryEngine like/ilike routing by column encryption', () => {
       .then(() => {
         expect(applySearchTokensSpy).toHaveBeenCalled()
       })
+  })
+})
+
+describe('process-scoped column-existence cache (#5605)', () => {
+  const columnProbes = (fakeDb: any) =>
+    fakeDb._calls.filter((b: any) => b._ops.table === 'information_schema.columns')
+
+  const columnProbesFor = (fakeDb: any, column: string) =>
+    columnProbes(fakeDb).filter((b: any) =>
+      b._ops.wheres.some((w: any) => Array.isArray(w) && w[0] === 'column_name' && w[1] === '=' && w[2] === column)
+    )
+
+  test('the column-existence cache is shared across separate engine instances', async () => {
+    // `createRequestContainer()` builds a fresh `BasicQueryEngine` per HTTP request.
+    // The cache must live on the module, not the instance, so a later "request" (a
+    // second engine here) reuses the first request's answer instead of re-probing
+    // `information_schema.columns`.
+    const fakeDb = createFakeKysely({
+      customer_entities: [],
+      'information_schema.columns': [
+        { table_name: 'customer_entities', column_name: 'tenant_id' },
+      ],
+    })
+    const queryOpts = { tenantId: 't1', fields: ['id'], page: { page: 1, pageSize: 10 } }
+
+    const engine1 = new BasicQueryEngine({} as any, () => fakeDb as any)
+    await engine1.query('customers:customer_entity', queryOpts)
+    const probesAfterFirstEngine = columnProbes(fakeDb).length
+    expect(probesAfterFirstEngine).toBeGreaterThan(0)
+
+    const engine2 = new BasicQueryEngine({} as any, () => fakeDb as any)
+    await engine2.query('customers:customer_entity', queryOpts)
+    expect(columnProbes(fakeDb).length).toBe(probesAfterFirstEngine)
+  })
+
+  test('a missing column is remembered as false instead of being re-queried on every call', async () => {
+    // Before the fix, `columnExists` deleted a `false` result instead of caching it,
+    // so `organization_id` (absent here) was re-queried once per query projection
+    // ('full' and 'count') within a single `.query()` call.
+    const fakeDb = createFakeKysely({
+      customer_entities: [],
+      'information_schema.columns': [],
+    })
+    const engine = new BasicQueryEngine({} as any, () => fakeDb as any)
+
+    await engine.query('customers:customer_entity', {
+      tenantId: 't1',
+      organizationId: 'org1',
+      fields: ['id'],
+      page: { page: 1, pageSize: 10 },
+    })
+
+    expect(columnProbesFor(fakeDb, 'organization_id').length).toBe(1)
   })
 })
--- a/packages/shared/src/lib/query/__tests__/engine.test.ts
+++ b/packages/shared/src/lib/query/__tests__/engine.test.ts
@@ -2,7 +2,7 @@ import { BasicQueryEngine } from '../engine'
 import { SortDir } from '../types'
 import { registerModules } from '../../i18n/server'
 import { clearSearchTokenPresenceCache } from '../../search/availability'
-import { clearEncryptedLikeFieldsCache, clearColumnExistsCache } from '../engine'
+import { clearEncryptedLikeFieldsCache, clearColumnExistsCache, columnExistsCacheSize } from '../engine'
 
 // The token-presence answer is cached process-wide (TTL); without clearing it,
 // probe-count assertions would observe hits from earlier tests in this file.
@@ -1417,7 +1417,7 @@ describe('BasicQueryEngine like/ilike routing by column encryption', () => {
   })
 })
 
-describe('process-scoped column-existence cache (#5605)', () => {
+describe('module-scoped column-existence cache (#5605)', () => {
   const columnProbes = (fakeDb: any) =>
     fakeDb._calls.filter((b: any) => b._ops.table === 'information_schema.columns')
 
@@ -1425,6 +1425,24 @@ describe('process-scoped column-existence cache (#5605)', () => {
     columnProbes(fakeDb).filter((b: any) =>
       b._ops.wheres.some((w: any) => Array.isArray(w) && w[0] === 'column_name' && w[1] === '=' && w[2] === column)
     )
+
+  const hasTenantGuard = (fakeDb: any) => {
+    const baseCall = fakeDb._calls.find((b: any) => b._ops.table === 'customer_entities')
+    return !!baseCall?._ops.wheres.some(
+      (w: any) => Array.isArray(w) && w[0] === 'customer_entities.tenant_id' && w[1] === '=' && w[2] === 't1',
+    )
+  }
+
+  const originalTtl = process.env.OM_QUERY_COLUMN_EXISTS_CACHE_MS
+  const originalMaxEntries = process.env.OM_QUERY_COLUMN_EXISTS_CACHE_MAX_ENTRIES
+
+  afterEach(() => {
+    if (originalTtl === undefined) delete process.env.OM_QUERY_COLUMN_EXISTS_CACHE_MS
+    else process.env.OM_QUERY_COLUMN_EXISTS_CACHE_MS = originalTtl
+    if (originalMaxEntries === undefined) delete process.env.OM_QUERY_COLUMN_EXISTS_CACHE_MAX_ENTRIES
+    else process.env.OM_QUERY_COLUMN_EXISTS_CACHE_MAX_ENTRIES = originalMaxEntries
+    jest.restoreAllMocks()
+  })
 
   test('the column-existence cache is shared across separate engine instances', async () => {
     // `createRequestContainer()` builds a fresh `BasicQueryEngine` per HTTP request.
@@ -1467,5 +1485,90 @@ describe('process-scoped column-existence cache (#5605)', () => {
     })
 
     expect(columnProbesFor(fakeDb, 'organization_id').length).toBe(1)
+  })
+
+  test('a cached negative expires, so a migrated-in scope column is picked up without a process restart', async () => {
+    // A cached `false` is consumed where the tenant/organization/soft-delete predicates
+    // are applied, so a stale one does not merely slow a query down — it silently drops
+    // a scope guard. The TTL bounds that window: a migration applied against a running
+    // process (`yarn dev`, or pods not recycled by a separate migration release step)
+    // converges once the entry expires instead of never.
+    const now = 1_700_000_000_000
+    const nowSpy = jest.spyOn(Date, 'now').mockReturnValue(now)
+
+    const beforeMigration = createFakeKysely({
+      customer_entities: [],
+      'information_schema.columns': [],
+    })
+    const queryOpts = { tenantId: 't1', fields: ['id'], page: { page: 1, pageSize: 10 } }
+    await new BasicQueryEngine({} as any, () => beforeMigration as any).query('customers:customer_entity', queryOpts)
+    expect(columnProbesFor(beforeMigration, 'tenant_id').length).toBe(1)
+    expect(hasTenantGuard(beforeMigration)).toBe(false)
+
+    // Same process, same cache — a second request while the entry is still fresh reuses
+    // the negative and does not re-probe, even though the column now exists.
+    const afterMigration = createFakeKysely({
+      customer_entities: [],
+      'information_schema.columns': [
+        { table_name: 'customer_entities', column_name: 'tenant_id' },
+      ],
+    })
+    await new BasicQueryEngine({} as any, () => afterMigration as any).query('customers:customer_entity', queryOpts)
+    expect(columnProbesFor(afterMigration, 'tenant_id').length).toBe(0)
+    expect(hasTenantGuard(afterMigration)).toBe(false)
+
+    nowSpy.mockReturnValue(now + 300_001)
+
+    const afterTtl = createFakeKysely({
+      customer_entities: [],
+      'information_schema.columns': [
+        { table_name: 'customer_entities', column_name: 'tenant_id' },
+      ],
+    })
+    await new BasicQueryEngine({} as any, () => afterTtl as any).query('customers:customer_entity', queryOpts)
+    expect(columnProbesFor(afterTtl, 'tenant_id').length).toBe(1)
+    expect(hasTenantGuard(afterTtl)).toBe(true)
+  })
+
+  test('the cache is bounded, so caller-supplied sort fields cannot grow it without limit', async () => {
+    // `columnExists` is reached with raw request input through `resolveBaseColumn` —
+    // `sortField` arrives from the query string and many list schemas type it as a plain
+    // string. Without a cap, one distinct name per request would be retained for the
+    // lifetime of the process.
+    process.env.OM_QUERY_COLUMN_EXISTS_CACHE_MAX_ENTRIES = '4'
+    const fakeDb = createFakeKysely({
+      customer_entities: [],
+      'information_schema.columns': [],
+    })
+    const engine = new BasicQueryEngine({} as any, () => fakeDb as any)
+
+    for (let index = 0; index < 20; index += 1) {
+      await engine.query('customers:customer_entity', {
+        tenantId: 't1',
+        fields: ['id'],
+        sort: [{ field: `attacker_supplied_${index}`, dir: SortDir.Asc }],
+        page: { page: 1, pageSize: 10 },
+      })
+      expect(columnExistsCacheSize()).toBeLessThanOrEqual(4)
+    }
+  })
+
+  test('OM_QUERY_COLUMN_EXISTS_CACHE_MS=0 disables the memo and probes per request again', async () => {
+    process.env.OM_QUERY_COLUMN_EXISTS_CACHE_MS = '0'
+    const fakeDb = createFakeKysely({
+      customer_entities: [],
+      'information_schema.columns': [
+        { table_name: 'customer_entities', column_name: 'tenant_id' },
+      ],
+    })
+    const queryOpts = { tenantId: 't1', fields: ['id'], page: { page: 1, pageSize: 10 } }
+
+    await new BasicQueryEngine({} as any, () => fakeDb as any).query('customers:customer_entity', queryOpts)
+    const probesAfterFirstEngine = columnProbesFor(fakeDb, 'tenant_id').length
+    expect(probesAfterFirstEngine).toBeGreaterThan(0)
+    expect(columnExistsCacheSize()).toBe(0)
+
+    await new BasicQueryEngine({} as any, () => fakeDb as any).query('customers:customer_entity', queryOpts)
+    expect(columnProbesFor(fakeDb, 'tenant_id').length).toBeGreaterThan(probesAfterFirstEngine)
   })
 })
--- a/packages/shared/src/lib/query/engine.ts
+++ b/packages/shared/src/lib/query/engine.ts
@@ -33,6 +33,7 @@ import { warnOnCiphertextLikeFallback } from './ciphertext-search-warning'
 import { resolveEncryptedSortFields, resolveEncryptedSortMaxRows, sortRowsInMemory } from './encrypted-sort'
 import { resolveListCountCap } from './count-cap'
 import { mapWithConcurrency } from './bounded-decrypt'
+import { parseNumberWithDefault } from '../number'
 import { createLogger } from '../logger'
 
 const logger = createLogger('shared').child({ component: 'query' })
@@ -89,17 +90,56 @@ export function clearEncryptedLikeFieldsCache(): void {
   encryptedLikeFieldsCache.clear()
 }
 
-// Process-scoped: schema shape (does a table have this column?) only changes on
-// migration, and a process restart already clears this Map, so it is always safe to
-// reuse across requests — unlike per-request state such as `tenantEncryptionService`.
-// Module-level on purpose: `createRequestContainer` builds a fresh `BasicQueryEngine`
+// Module-scoped on purpose: `createRequestContainer` builds a fresh `BasicQueryEngine`
 // per request, so an instance field alone re-pays the `information_schema` probe on
-// every request (#5605).
-const columnExistsCache = new Map<string, boolean>()
+// every request (#5605). Schema shape (does a table have this column?) is not
+// per-request state — unlike `tenantEncryptionService` — so sharing the answer across
+// requests is safe. One map per module instance rather than a true process singleton:
+// standalone builds can duplicate this package, which for a memo is harmless (two
+// caches, both correct), so nothing may be built on top of singleton semantics here.
+//
+// Bounded and TTL'd for two reasons. `columnExists` is reached with caller-supplied
+// field names via `resolveBaseColumn` (sort fields and base filter keys arrive raw from
+// the HTTP layer), so an unbounded map would grow monotonically on request input. And a
+// cached `false` is consumed where the tenant/organization/soft-delete predicates are
+// applied, so a schema change that adds one of those columns must not stay invisible
+// until the process restarts — a migration applied against a running `yarn dev` or
+// not-yet-recycled pods converges within the TTL instead. The TTL still removes
+// essentially all of the traffic: a hot column is probed twelve times an hour rather
+// than tens of thousands. Set OM_QUERY_COLUMN_EXISTS_CACHE_MS=0 to disable and probe
+// per request again; OM_QUERY_COLUMN_EXISTS_CACHE_MAX_ENTRIES tunes the bound.
+const COLUMN_EXISTS_CACHE_DEFAULT_TTL_MS = 300_000
+const COLUMN_EXISTS_CACHE_DEFAULT_MAX_ENTRIES = 10_000
+const columnExistsCache = new Map<string, { value: boolean; expiresAt: number }>()
 
-/** Test-only: the process-scoped memo would otherwise leak state across specs. */
+function resolveColumnExistsCacheTtlMs(): number {
+  return parseNumberWithDefault(process.env.OM_QUERY_COLUMN_EXISTS_CACHE_MS, COLUMN_EXISTS_CACHE_DEFAULT_TTL_MS, { integer: true, min: 0 })
+}
+
+function resolveColumnExistsCacheMaxEntries(): number {
+  return parseNumberWithDefault(process.env.OM_QUERY_COLUMN_EXISTS_CACHE_MAX_ENTRIES, COLUMN_EXISTS_CACHE_DEFAULT_MAX_ENTRIES, { integer: true, min: 1 })
+}
+
+function storeColumnExists(key: string, value: boolean, ttlMs: number): void {
+  const maxEntries = resolveColumnExistsCacheMaxEntries()
+  if (columnExistsCache.size >= maxEntries) {
+    const now = Date.now()
+    for (const [entryKey, entry] of columnExistsCache) {
+      if (entry.expiresAt <= now) columnExistsCache.delete(entryKey)
+    }
+    if (columnExistsCache.size >= maxEntries) columnExistsCache.clear()
+  }
+  columnExistsCache.set(key, { value, expiresAt: Date.now() + ttlMs })
+}
+
+/** Test-only: the module-scoped memo would otherwise leak state across specs. */
 export function clearColumnExistsCache(): void {
   columnExistsCache.clear()
+}
+
+/** Test-only: entry count of the column-existence memo, for the cap regression test. */
+export function columnExistsCacheSize(): number {
+  return columnExistsCache.size
 }
 
 type ResolvedCustomFieldSource = {
@@ -673,7 +713,7 @@ export class BasicQueryEngine implements QueryEngine {
     // and cf filters are expressed as correlated EXISTS semi-joins, so nothing can
     // multiply base rows and a LIMIT above the query is an enforceable bound.
     // Re-running the WHERE/JOIN logic per projection is cheap: every `columnExists`
-    // check is memoized on the process-scoped `columnExistsCache`, so later passes
+    // check is memoized on the module-scoped `columnExistsCache`, so later passes
     // hit no extra DB calls.
     const buildQuery = async (projection: 'full' | 'sortKeys' | 'count'): Promise<BuiltQuery> => {
       const isSortKeysProjection = projection === 'sortKeys'
@@ -1558,8 +1598,9 @@ export class BasicQueryEngine implements QueryEngine {
 
   private async columnExists(table: string, column: string): Promise<boolean> {
     const key = `${table}.${column}`
+    const ttlMs = resolveColumnExistsCacheTtlMs()
     const cached = columnExistsCache.get(key)
-    if (cached !== undefined) return cached
+    if (cached && cached.expiresAt > Date.now()) return cached.value
     const db = this.getDb()
     const exists = await db
       .selectFrom('information_schema.columns' as any)
@@ -1569,7 +1610,7 @@ export class BasicQueryEngine implements QueryEngine {
       .limit(1)
       .executeTakeFirst()
     const present = !!exists
-    columnExistsCache.set(key, present)
+    if (ttlMs > 0) storeColumnExists(key, present, ttlMs)
     return present
   }
 
--- a/packages/shared/src/lib/query/engine.ts
+++ b/packages/shared/src/lib/query/engine.ts
@@ -89,6 +89,19 @@ export function clearEncryptedLikeFieldsCache(): void {
   encryptedLikeFieldsCache.clear()
 }
 
+// Process-scoped: schema shape (does a table have this column?) only changes on
+// migration, and a process restart already clears this Map, so it is always safe to
+// reuse across requests — unlike per-request state such as `tenantEncryptionService`.
+// Module-level on purpose: `createRequestContainer` builds a fresh `BasicQueryEngine`
+// per request, so an instance field alone re-pays the `information_schema` probe on
+// every request (#5605).
+const columnExistsCache = new Map<string, boolean>()
+
+/** Test-only: the process-scoped memo would otherwise leak state across specs. */
+export function clearColumnExistsCache(): void {
+  columnExistsCache.clear()
+}
+
 type ResolvedCustomFieldSource = {
   entityId: EntityId
   alias: string
@@ -271,7 +284,6 @@ function computeCustomFieldScore(cfg: Record<string, unknown>, kind: string, ent
  * {@link HybridQueryEngine} when the query index is unavailable or incomplete.
  */
 export class BasicQueryEngine implements QueryEngine {
-  private columnCache = new Map<string, boolean>()
   private searchAliasSeq = 0
   private searchAvailabilityInstance: SearchTokenAvailability | null = null
 
@@ -661,7 +673,8 @@ export class BasicQueryEngine implements QueryEngine {
     // and cf filters are expressed as correlated EXISTS semi-joins, so nothing can
     // multiply base rows and a LIMIT above the query is an enforceable bound.
     // Re-running the WHERE/JOIN logic per projection is cheap: every `columnExists`
-    // check is memoized on `this.columnCache`, so later passes hit no extra DB calls.
+    // check is memoized on the process-scoped `columnExistsCache`, so later passes
+    // hit no extra DB calls.
     const buildQuery = async (projection: 'full' | 'sortKeys' | 'count'): Promise<BuiltQuery> => {
       const isSortKeysProjection = projection === 'sortKeys'
       const isCountProjection = projection === 'count'
@@ -1545,11 +1558,8 @@ export class BasicQueryEngine implements QueryEngine {
 
   private async columnExists(table: string, column: string): Promise<boolean> {
     const key = `${table}.${column}`
-    if (this.columnCache.has(key)) {
-      const cached = this.columnCache.get(key)
-      if (cached === true) return true
-      this.columnCache.delete(key)
-    }
+    const cached = columnExistsCache.get(key)
+    if (cached !== undefined) return cached
     const db = this.getDb()
     const exists = await db
       .selectFrom('information_schema.columns' as any)
@@ -1559,8 +1569,7 @@ export class BasicQueryEngine implements QueryEngine {
       .limit(1)
       .executeTakeFirst()
     const present = !!exists
-    if (present) this.columnCache.set(key, true)
-    else this.columnCache.delete(key)
+    columnExistsCache.set(key, present)
     return present
   }
 


pkarw:
🤖 `om-auto-review-pr` — claimed this PR for automated code review at 2026-08-25T12:12Z. Will release the lock when the review is posted.

pkarw (changes requested):
# 🔍 Code Review: fix(shared/query): promote BasicQueryEngine columnExists cache to process scope (#5605)

## 🎯 Summary

This PR moves `BasicQueryEngine`'s column-existence memo from a `private` instance field to a module-level `Map`, and changes `columnExists()` to remember a `false` answer instead of deleting it. The diagnosis behind it is correct and well evidenced: `createRequestContainer()` builds a fresh `BasicQueryEngine` on every request (`packages/shared/src/lib/di/container.ts:213`), so an instance-scoped memo could never pay off, and the `false`-is-deleted branch meant a persistently-absent column was re-probed once per projection pass inside a single `.query()`. Moving the memo to module scope is the right shape of fix, the two regression tests target exactly the two behaviours that changed, and the `mutate (shared, src/lib/query/engine.ts)` CI job passing is a real signal that those tests actually pin the change rather than merely executing it. The comment block at `engine.ts:93-99` explaining *why* this particular state is safe to share across requests (unlike `tenantEncryptionService`) is the kind of thing that keeps the next reader from undoing it.

What holds it back is the second half of the change rather than the first. Promoting the cache to process scope is safe; promoting it *and* starting to store negative answers, in an unbounded `Map` with no TTL, turns a per-request structure into a process-lifetime one that is keyed partly by unvalidated request input. This package already contains the pattern that solves it — `packages/shared/src/lib/search/availability.ts` runs a process-scoped presence cache with both a TTL and a max-entry bound — and adopting it would resolve the major and the first minor together in a few lines.

Reviewed scope: the two changed files, plus the call graph that reaches `columnExists` (`resolveBaseColumn`, the auto-scope and projection paths in `buildQuery`), the DI wiring in `container.ts` and `query_index/di.ts`, the `HybridQueryEngine` fallback paths that route into the patched `BasicQueryEngine`, the sibling query-engine spec files that share the new module-level state, and issue #5605's stated implementation steps.

## Verdict

❌ **request changes** — the new process-scoped negative caching (`engine.ts:98`) is unbounded and partly keyed by untrusted request input, which is a major finding; three minors follow (no TTL bound on stale negatives, the cache-clearing helper wired into only one of the three spec files that now share the state, and the `information_schema.tables` half of #5605 left unaddressed without a note). None of them require rethinking the approach — the fix direction is the existing `search/availability.ts` cache pattern.

## 🧪 Validation Gate

Validation source: **GitHub checks** (head `8ae0199eaa77a3c26a4b272be53dd5f495e309db`, workflow run `32838790453`, conclusion `success`). No local fallback was needed — every configured gate is covered by a green required check on this exact head, so nothing was re-run locally.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | Covered by the `prepare` job (3m45s), which builds packages before the downstream jobs consume its artifact. |
| `yarn generate` | ✅ PASS | Covered by `prepare`. |
| `yarn build:packages` (second pass) | ✅ PASS | Covered by `prepare`; the repeated build is load-bearing for the generated entity-id barrel and did not regress. |
| `yarn i18n:check-sync` | ✅ PASS | Run inside the `test` job (`ci.yml:635` region). No locale files in the diff. |
| `yarn i18n:check-usage` | ✅ PASS | `ci.yml:635`, inside the green `test` job. |
| `yarn typecheck` | ✅ PASS | `ci.yml:648-653`, inside the green `test` job. On PRs it is scoped to affected packages, which here is exactly `@open-mercato/shared` — the only package touched. |
| `yarn test` | ✅ PASS | `test` job green (4m26s). Additionally `mutate (shared, src/lib/query/engine.ts)` passed (10m6s), so the new tests survive mutation of the changed file. |
| `yarn build:app` | ✅ PASS | Covered by the green `ephemeral-integration` shards (15/15), which build and boot the app before running Playwright. |

Also green and worth naming: `lint`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance`, and `license/cla`. `agents:check-budget` runs inside `test` (`ci.yml:644`) and passed, so nothing downstream of it was skipped. No check is pending and none failed.

## Findings

### ⚠️ Major

**`packages/shared/src/lib/query/engine.ts:98` — the cache is now unbounded and partly keyed by unvalidated request input, with nothing to evict it short of a process restart.**

Storing `false` is correct for the perf goal, but combined with process scope it changes the memory profile of the structure. `columnExists` is not only called with the fixed set of scope columns (`organization_id`, `tenant_id`, `deleted_at`); it is also reached with caller-supplied field names through `resolveBaseColumn(table, field)` at `engine.ts:1553-1556`, which is called for every sort field (`engine.ts:640`) and for every base filter field (`engine.ts:702` region). Those names arrive raw from the HTTP layer:

- `packages/core/src/modules/entities/api/records.ts:170` — `const sortField = url.searchParams.get('sortField') || 'id'`, handed to the engine unchanged at `records.ts:264`.
- `packages/core/src/modules/entities/api/records.ts:176-180` — every query-param key that is not on the small reserved list becomes a filter field.
- The generic CRUD factory does the same: `packages/shared/src/lib/crud/factory.ts:1727-1737` resolves the requested sort through `sortFieldMap[field] || field`, so an unmapped name passes through verbatim, and a great many list schemas declare `sortField: z.string().optional()` with no enum (for example `packages/core/src/modules/entities/api/records.ts:130`, `packages/core/src/modules/business_rules/api/rules/route.ts:40`).

This path really is reached in a deployed app. `packages/core/src/modules/query_index/di.ts:154` replaces the `queryEngine` DI key with `HybridQueryEngine`, but that engine delegates to the per-request `BasicQueryEngine` it was constructed with on its fallback paths (`packages/core/src/modules/query_index/lib/engine.ts:341`, `:348`, `:383`, `:412`), and apps without `query_index` resolve `BasicQueryEngine` directly.

Concrete failure: an authenticated client issuing `GET /api/entities/records?entityId=…&sortField=a1`, `…&sortField=a2`, … (or the same with arbitrary filter-param keys) adds one permanent `Map` entry per distinct name. Before this PR each of those requests cost a wasted `information_schema` probe and nothing else, because the `false` was deleted; now the probe is paid once and the memory is kept for the life of the process. There is no size cap, no TTL, and no eviction, so this is monotonic growth driven by request input — modest per entry, but with no ceiling and no recovery short of a restart.

Fix: bound the cache the way this package already bounds the equivalent structure. `packages/shared/src/lib/search/availability.ts:88-116` keeps a process-scoped presence map with `PRESENCE_CACHE_MAX_ENTRIES = 10_000`, a TTL read from an env knob, and a `storePresence` helper that sweeps expired entries and clears wholesale when the cap is still exceeded. Reusing that shape here (an entry of `{ value, expiresAt }`, a `storeColumnExists` helper, and the same cap) gives the same perf win, bounds the memory, and — see the next finding — bounds the staleness at the same time. A cheaper alternative, if a TTL is unwanted, is to keep caching negatives only for the fixed scope-column set and leave caller-supplied field names uncached, but the bounded-map approach is simpler and matches existing precedent.

### 🔹 Minor

**1. `packages/shared/src/lib/query/engine.ts:1559-1573` — a stale negative now silently drops a scope guard, and the "process restart clears it" argument does not cover every path that changes the schema.**

The comment at `engine.ts:93-99` reasons that schema shape only changes on migration and that a restart clears the map. That is true for a deploy that restarts processes, but two ordinary situations break it. First, local development: `yarn dev` holds one long-lived process, and a developer who runs a migration in another shell (or, per `AGENTS.md`, has a teammate's migration applied to the shared dev database) keeps serving from a cache that answered before the column existed. Second, deployments that run migrations as a separate release step against pods that are not immediately recycled — a column added to a table those pods already probed stays invisible until they restart.

What makes this worth fixing rather than merely noting is the failure mode. The cached value is consumed at `engine.ts:684`, `:687` and `:690`, where a `false` means the `organization_id` predicate, the `tenant_id` predicate, or the `deleted_at IS NULL` predicate is silently *not applied*. A stale negative therefore does not degrade into slow queries; it degrades into a missing tenant/organization filter or soft-deleted rows resurfacing, with no error anywhere. Issue #5605 step 3 did say "no runtime invalidation needed", so this is an accepted trade-off rather than an oversight — but it was accepted as a performance trade-off, and the safety dimension is worth pricing in. A TTL (the `search/availability.ts` pattern from the major above, defaulting to something like the 5 minutes used by `QUERY_INDEX_COVERAGE_CACHE_MS` and `customFieldKeysTtlMs` in `query_index/lib/engine.ts:223-226`) retains essentially all of the win — a hot column is re-probed twelve times an hour instead of tens of thousands of times — while bounding the window in which a scope guard can be silently missing. One change closes this and the major together.

**2. `packages/shared/src/lib/query/__tests__/engine.test.ts:11` — `clearColumnExistsCache()` is wired into one spec file, but the state it guards is now shared by three.**

Adding the `beforeEach` clear here is correct and necessary. The problem is that two sibling files construct `BasicQueryEngine` against fake schemas and never clear it, so they silently inherited a cross-test dependency from this PR:

- `packages/shared/src/lib/query/__tests__/engine.scope-and-or.test.ts` builds eight `createFakeKysely` fixtures that declare *mutually contradictory* `information_schema.columns` shapes for the same `scheduled_jobs` table — only `organization_id` at `:237` and `:264`, only `id` at `:312`, `id`/`organization_id`/`tenant_id` at `:356`, only `tenant_id` at `:404`, and `organization_id`/`tenant_id`/`scope_type` at `:427`.
- `packages/shared/src/lib/query/__tests__/engine.count-distinct.test.ts` likewise builds eight fixtures over `scheduled_jobs`.

Jest gives each file its own module registry, so the leakage is within a file rather than across them — but within `engine.scope-and-or.test.ts` the first test's answers now decide later tests' scoping. It passes today only by coincidence: every test that declares a contradicting shape also passes `omitAutomaticTenantOrgScope: true`, which skips the `tenant_id`/`organization_id` probes entirely (`engine.ts:684-687`), so the one test that *does* exercise the automatic guard — `'applies automatic tenant guard when flag is absent (baseline)'` at `:402-419`, whose fixture declares only `tenant_id` — happens to run before anything has cached `scheduled_jobs.tenant_id`. Insert one auto-scoped test above it, or reorder the describes, and the baseline test reads a stale `false` and asserts against a query that has silently lost its tenant filter. That is precisely the regression these specs exist to catch, so it should not depend on ordering. Add the same `beforeEach(() => clearColumnExistsCache())` to both files, or hoist the clear into the shared jest setup so any future spec gets it for free.

**3. The `information_schema.tables` half of #5605 is neither implemented nor explained, so the issue looks fully closed by this PR when it is not.**

Issue #5605 step 1 asks to "promote `columnCache`/`tableCache` to process scope", and its measured evidence attributes a large share of the introspection traffic to table probes: "`information_schema.tables` exists-check: 18,474 calls/460.1ms". This PR addresses only the column half, and the PR body's "Out of scope" note covers just `query_index`, so nothing tells a reader that the tables half is still open.

To be fair to the author, the issue's own citation is inaccurate — `BasicQueryEngine` has no `tableCache` or `tableExists` member at all (the `engine.ts:208-209` reference points at `HybridQueryEngine` in `packages/core/src/modules/query_index/lib/engine.ts:205`), so there was nothing to promote here. But the traffic the issue measured is real and still uncached per request: `DefaultDataEngine.ensureStorageTableExists()` at `packages/shared/src/lib/data/engine.ts:265-275` runs an unmemoized `information_schema.tables` probe, is called on both the read and write paths (`:314` and `:419`), and `DefaultDataEngine` is constructed fresh on every request right beside the query engine at `packages/shared/src/lib/di/container.ts:216`. That is the same bug in the same generic layer, and it is a two-line memo. Either fold it into this PR, or state explicitly in the PR body — and in a comment on #5605 before it is closed — that the tables probe lives in `DefaultDataEngine` rather than in `BasicQueryEngine` and is tracked separately, so the issue is not closed on a partial fix.

### 💅 Nit

**`packages/shared/src/lib/query/engine.ts:93-99` — "process-scoped" is very slightly stronger than what the module system guarantees here.**

The comment is good and should stay. One half-sentence would make it airtight: in standalone builds the shared package can be duplicated, which is exactly why `packages/shared/src/lib/db/mikro.ts:15` parks the ORM entity list on `globalThis` and why `packages/shared/AGENTS.md` calls out the same for nav-ordering state. For a pure memo, duplication is harmless — you get two caches instead of one, both still correct — so this needs no code change. Noting it prevents a future reader from treating "one map per process" as a guarantee and building something on top of it that actually depends on singleton semantics.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff is purely additive on the export surface: `columnExistsCache` stays module-private and `clearColumnExistsCache()` is new, mirroring the existing `clearEncryptedLikeFieldsCache()` convention in the same file.
- [x] No function signature changed in a breaking way — `columnExists` is `private` and keeps its `(table, column) => Promise<boolean>` shape; the removed `columnCache` field was `private` too.
- [x] No database table or column renamed or removed; no column type narrowed — there is no migration in this PR and none is needed.
- [x] No config key renamed and no default changed silently — though note that if the TTL suggested above is adopted, it should arrive as a new env knob with a safe default, in the style of `OM_SEARCH_TOKEN_PRESENCE_CACHE_MS`, rather than as a change to any existing default.
- [x] Where a contract had to change: not applicable, no contract surface changed.

The `queryEngine` DI key resolves the same way for every caller, so `BACKWARD_COMPATIBILITY.md`'s DI-key and exported-type surfaces are untouched. The observable behaviour change is confined to caching lifetime, which is what minor finding 1 discusses.

## 🧪 Test Coverage

The two added tests in `packages/shared/src/lib/query/__tests__/engine.test.ts:1420-1471` cover the two behaviours that actually changed, and they are written so that they fail against the pre-fix code rather than merely exercising it. The first constructs two separate `BasicQueryEngine` instances against one fake database and asserts that the second makes zero additional `information_schema.columns` calls — with an instance-scoped cache the second engine would re-probe from empty, so the assertion is load-bearing for the scope change. The second runs a single query against a schema where `organization_id` is absent and asserts exactly one probe for that column, which pins the negative-caching change specifically: the old code deleted the `false` and re-probed once per projection pass. The `columnProbesFor` helper filtering on the recorded `where` tuple is the right granularity, since a count over all probes would be satisfied by unrelated columns. The `beforeEach` clear is correctly placed alongside the two existing cache clears. The passing `mutate (shared, src/lib/query/engine.ts)` job is independent evidence that these assertions detect changes to the file rather than passing vacuously.

Gaps to close, all in `packages/shared/src/lib/query/__tests__/engine.test.ts`:

- **A bound/eviction test for the major finding.** Once the cache is capped, add a case that probes more distinct column names than the cap and asserts the map does not grow past it. Exporting the cap as a constant (or accepting an injectable one) keeps the test from having to loop ten thousand times. Note that `packages/shared/src/lib/search/__tests__/availability.test.ts` does *not* currently cover `PRESENCE_CACHE_MAX_ENTRIES`, so there is no precedent to copy here — this would be the first test pinning that behaviour, and without it a future refactor can quietly remove the bound with nothing noticing.
- **A staleness test if a TTL is adopted for minor finding 1.** Drive the clock with fake timers, probe a column that is absent, extend the fake schema to contain it, advance past the TTL, and assert the engine re-probes and applies the scope predicate. This is the test that pins "a migrated-in `tenant_id` is picked up without a restart".
- **A test that the cached negative does not survive into a differently-shaped schema within one file**, which is really the assertion missing from finding 2. Adding the `beforeEach` clear to `engine.scope-and-or.test.ts` and `engine.count-distinct.test.ts` is the fix; a single explicit test that queries the same table under two contradictory fixtures back to back and asserts the second sees its own schema would make the requirement visible rather than implicit.

No integration coverage is needed for this change: it alters no API surface, no UI, and no schema, and the behaviour is fully observable at the unit level through the fake Kysely harness. The `skip-qa` label is correctly applied under the automated-verification exemption — no `.tsx`, nothing under `packages/ui/src/`, no database or API-surface change, and automated tests for the changed behaviour ship in the same PR.

pkarw:
## 🏷️ label rationale

_(updated in place after the re-review of `40d19f9` — the earlier rationale for `changes-requested` is superseded.)_

| Label | Change | Why |
|-------|--------|-----|
| `merge-queue` ← `changes-requested`/`review` | replaced | The re-review approved the PR: all five findings from the first review are addressed with tests that fail against the earlier code, and every command in the configured validation gate is green on head `40d19f9`. Approved PRs land in `merge-queue`; pipeline labels are mutually exclusive, so `review` came off in the same transition. Only nits remain, and nits do not block. |
| `skip-qa` | kept | The automated-verification exemption still holds on the follow-up commit: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no database structure or API surface change, no `BACKWARD_COMPATIBILITY.md` contract broken, and automated tests ship for every behavior changed. No manual QA pass would be meaningful here, so no QA sign-off is outstanding. |
| `bug` | kept | The change fixes a confirmed defect (#5605) rather than adding capability. |
| `performance` | kept | Accurate — the whole point is eliminating per-request `information_schema.columns` round-trips. |
| `partner-request` | kept | Provenance is unchanged by this review. |
| `priority-high` | kept | Exactly one priority label. Unchanged from #5605: a systemic per-request introspection cost confirmed under load on two separate modules. |
| `risk-low` | kept | Exactly one risk label. Better justified now than at the first review: the diff is confined to one file's caching layer plus its specs, no contract surface moves, and the follow-up commit bounds the blast radius the first review flagged — a cap so caller-supplied field names cannot grow the map, and a TTL so a stale negative cannot indefinitely suppress a tenant or organization scope predicate. The green `mutate (shared, src/lib/query/engine.ts)` job is independent evidence that the new tests actually pin that behavior. |
| ~~`in-progress`~~ | removed | The review is submitted and the lock is released. `ci-monitoring` is deliberately **not** applied in its place: CI was already fully settled and green on this head before the review started, so there is no follow-up owed and the label would promise one that has already arrived. |

**Merge gate:** `skip-qa` is set and no blocking label (`qa-failed`, `do-not-merge`, `blocked`) is present, so the QA-approval gate does not apply. The merge itself belongs to `om-approve-merge-pr 5614`.

pkarw:
@adeptofvoltron — handing this back to you: the review requested changes, so the PR is reassigned to you and moved to `changes-requested`.

The diagnosis and the direction are right, and CI is green on `8ae0199`. What needs a second pass is the negative-caching half of the change:

1. **Major** — `packages/shared/src/lib/query/engine.ts:98`: with `false` now stored in a module-level `Map`, the cache is unbounded and reached with caller-supplied field names (`resolveBaseColumn` → `columnExists`, fed by raw `sortField` and arbitrary filter-param keys, e.g. `packages/core/src/modules/entities/api/records.ts:170,176-180`). Suggested fix: adopt the bounded pattern already in this package at `packages/shared/src/lib/search/availability.ts:88-116` (max-entry cap + TTL + sweep).
2. **Minor** — the same TTL bounds how long a stale `false` for `tenant_id` / `organization_id` / `deleted_at` can silently drop a scope predicate (`engine.ts:684-690`) after a migration lands without a process restart.
3. **Minor** — `clearColumnExistsCache()` is wired only into `engine.test.ts`; `engine.scope-and-or.test.ts` and `engine.count-distinct.test.ts` now share the module-level state across their own tests with contradictory fake schemas for `scheduled_jobs`, and pass today only by ordering luck.
4. **Minor** — the `information_schema.tables` traffic #5605 measured lives in `DefaultDataEngine.ensureStorageTableExists()` (`packages/shared/src/lib/data/engine.ts:265`), not in `BasicQueryEngine`; either fold it in or note it on the issue so #5605 is not closed on a partial fix.

Full reasoning, `file:line` references and the suggested test cases are in the review above. Re-request review once the changes are pushed.

pkarw:

CI was already settled and fully green on head `8ae0199` (run `32838790453`), so no CI follow-up wait was needed and `ci-monitoring` was not applied.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-08-25T12:33:25Z.

adeptofvoltron:

_(updated in place after CI settled green on `40d19f9`.)_

- 🔍 `review` — replaces `changes-requested`. The automated review's one major and three minor findings have all been addressed and pushed (`40d19f9`), so the ball is no longer in the author's court; what the PR now needs is a re-review of the follow-up commit, which is exactly what this pipeline state means. Pipeline labels are mutually exclusive, so `changes-requested` came off in the same transition. It is deliberately **not** `merge-queue`: that state means an approving review plus green checks, and while the checks are now green, the approving review does not exist — this run's account authored the PR and cannot self-approve.
- 🐛 `bug` — unchanged. The change corrects a memoization that never took effect, which is a defect in existing behaviour rather than new functionality.
- ⚡ `performance` — unchanged. The whole point of the change is eliminating tens of thousands of avoidable `information_schema` round-trips per process.
- 🤝 `partner-request` — unchanged. The originating report came from a partner deployment, and nothing in this run changes its provenance.
- 🔺 `priority-high` — unchanged. The regression is measurable under production load on two separate modules, so it should not wait for a later release train.
- 🟢 `risk-low` — unchanged, and better justified than before. The diff is confined to one file's caching layer plus its specs, no contract surface moves, and the follow-up commit specifically bounds the blast radius the review flagged (a cap so request-supplied field names cannot grow the map, and a TTL so a stale negative cannot silently suppress a tenant/organization scope predicate indefinitely). The green `mutate (shared, src/lib/query/engine.ts)` job confirms the tests actually pin that behaviour.
- 🧪 `skip-qa` — unchanged, under the automated-verification exemption. The diff touches no `.tsx`, nothing under `packages/ui/src/` or `**/components/**`, changes no database structure and no API surface, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships automated tests for every behaviour it changes. No manual QA pass is meaningful here.
- ~~👀 `ci-monitoring`~~ — **removed.** It was applied in place of `in-progress` to say honestly that the work was finished and reported while only the CI outcome was still being watched. That watch is over: every check passed on `40d19f9` and the CI-result comment has been posted, so the label would otherwise promise a follow-up that has already arrived.

adeptofvoltron:
## 🚀 `om-auto-fix-pr` — run summary

### 🔀 Base merge
`origin/develop` was already fully merged into this head (`8ae0199` sits directly on top of `15ffbe3`), so no merge commit was needed and nothing had to be conflict-resolved. Everything below therefore judges the real merge result rather than a stale branch.

### 🔍 Review findings — all addressed, none deferred
The `changes-requested` review left one major, three minors and one nit. Every one of them is accounted for in `40d19f9`; nothing actionable was pushed to a follow-up.

- **⚠️ Major — the cache was unbounded and partly keyed by unvalidated request input.** Fixed by adopting the pattern this package already runs for the equivalent structure in `packages/shared/src/lib/search/availability.ts`: a `{ value, expiresAt }` entry, a max-entry cap that sweeps expired entries before clearing wholesale, and a TTL read from an env knob. The reviewer was right that `columnExists` is reachable with raw HTTP input through `resolveBaseColumn` — `sortField` and unmapped filter-param keys arrive verbatim — so the map is now capped at `OM_QUERY_COLUMN_EXISTS_CACHE_MAX_ENTRIES` (default `10000`) and can no longer grow monotonically on request input.
- **🔹 Minor 1 — a stale negative silently drops a scope guard.** Fixed by the same change, which is why it was worth doing as one edit rather than two. The cached value is consumed exactly where the `organization_id`, `tenant_id` and `deleted_at IS NULL` predicates are applied, so the failure mode was a missing tenant filter rather than a slow query. `OM_QUERY_COLUMN_EXISTS_CACHE_MS` (default `300000`, `0` to disable) bounds that window: a migration applied against a running `yarn dev` or a not-yet-recycled pod now converges within five minutes instead of never. A hot column is still probed twelve times an hour rather than tens of thousands, so essentially all of the performance win survives.
- **🔹 Minor 2 — the cache-clearing helper was wired into only one of the three spec files sharing the state.** Fixed: `engine.scope-and-or.test.ts` and `engine.count-distinct.test.ts` now clear the memo in `beforeEach`. The reviewer's analysis was exact — those fixtures declare mutually contradictory `information_schema.columns` shapes for the same `scheduled_jobs` table, and the auto-scoped baseline test only passed because nothing had cached `scheduled_jobs.tenant_id` before it ran. That ordering coincidence is now gone.
- **🔹 Minor 3 — the `information_schema.tables` half of #5605 was neither implemented nor explained.** Taken as documentation rather than code, which is one of the two resolutions the review explicitly offered. The probe lives in `DefaultDataEngine.ensureStorageTableExists()`, a different class in a different subsystem from anything this PR touches, so folding it in would have widened an already-reviewed diff. Instead: the PR body now has an explicit **Out of scope (tracked separately)** section, the remaining work is filed as **#5619** with acceptance criteria mirroring the bounded/TTL'd shape settled here, and **#5605 has been explicitly re-scoped** to the columns half in [a comment](https://github.com/open-mercato/open-mercato/issues/5605#issuecomment-5410491092) so it is not closed on a partial fix without a trace.
- **💅 Nit — "process-scoped" overstated what the module system guarantees.** Fixed in the comment, which now says module-scoped and notes that standalone builds can duplicate this package — harmless for a memo (two caches, both correct), but nothing may be built on top of singleton semantics here.

### 🧪 Test coverage added
All three gaps the review named are now covered in `packages/shared/src/lib/query/__tests__/engine.test.ts`, and each is written to fail against the code as it stood before this commit:

- a cached negative expires at the TTL, so a migrated-in `tenant_id` is re-probed and the tenant guard reappears — driven by a stubbed `Date.now()` rather than real time, so it is deterministic;
- the entry count stays at or under the cap while twenty distinct caller-supplied sort fields are pushed through it, which is the assertion that stops a future refactor from quietly removing the bound;
- `OM_QUERY_COLUMN_EXISTS_CACHE_MS=0` disables the memo and probes per request again.

### ✅ Local validation
Run in this worktree against the pushed head: `yarn build:packages` green; `yarn workspace @open-mercato/shared typecheck` green; `yarn lint` green (0 errors, only pre-existing warnings); `yarn workspace @open-mercato/shared test src/lib/query/__tests__/engine.test.ts` 44/44; `yarn workspace @open-mercato/shared test src/lib/query src/lib/search` 275 passed / 5 skipped of 280, confirming the module-level cache does not leak across the sibling specs.

**CI is still running on this head.** Branch protection holds the actual merge; this run's verdict covers the work, not a green run. Checks still pending at the time of writing: `prepare`, `mutate (shared, src/lib/query/engine.ts)`. A follow-up comment will report the CI outcome.

### 🔁 Follow-ups filed
- **#5619** — memoize `DefaultDataEngine`'s `information_schema.tables` probe across requests. Deferred rather than fixed in-loop because it is a different class in a different subsystem; it carries the bounded-and-TTL'd acceptance criteria this PR settled on.

### 🚀 Hand-off
`om-auto-fix-pr` never merges, and it cannot approve this PR either — the automation account is this PR's author, so GitHub will not accept a self-approval. **@pkarw**, the findings from your review are all addressed in `40d19f9`; a re-review is what this PR now needs. The pipeline label has moved to `review` to reflect that.

Once a human approval lands and the required checks are green, the merge belongs to `om-approve-merge-pr 5614`. `skip-qa` applies under the automated-verification exemption, so no QA sign-off is outstanding.

adeptofvoltron:

**✅ Every check is green on head `40d19f9`** (workflow runs [32848707282](https://github.com/open-mercato/open-mercato/actions/runs/32848707282) and [32848707258](https://github.com/open-mercato/open-mercato/actions/runs/32848707258)). Nothing failed, nothing was rerun, and no failure had to be classified — so no flake follow-up is owed either. This settles the disclosure in the run summary above: the checks that were pending when this run reported have now all passed.

24 checks passed and 5 were skipped by their own path/branch filters (`docker-build`, `audit`, `merge-coverage`, `comment`, `[code]smith`). The ones worth naming:

- **[`prepare`](https://github.com/open-mercato/open-mercato/actions/runs/32848707282/job/97804326771)** — covers `yarn build:packages` (both passes) and `yarn generate`.
- **[`test`](https://github.com/open-mercato/open-mercato/actions/runs/32848707282/job/97805517340)** — covers `yarn test`, `yarn typecheck`, `yarn i18n:check-sync`, `yarn i18n:check-usage` and `yarn agents:check-budget`. This is the job that runs the three new regression tests.
- **[`mutate (shared, src/lib/query/engine.ts)`](https://github.com/open-mercato/open-mercato/actions/runs/32848707258/job/97804488204)** — the one that matters most for this particular diff. It passed against the *new* cache implementation, which is independent evidence that the added TTL-expiry, cap-bound and cache-disabled tests genuinely detect changes to `engine.ts` rather than merely executing it. A bound with no test pinning it was exactly the risk minor finding 3 of the review raised.
- **`ephemeral-integration` (15/15 shards)** — all green, which also covers `yarn build:app` since each shard builds and boots the app before driving Playwright.
- `lint`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance`, `license/cla` — all green.

**What this means for the verdict:** unchanged and improved. The PR was already reported as work-complete; CI now confirms it on the exact head that was reported. The pipeline label stays `review`.

`ci-monitoring` is being removed — the follow-up it promised is this comment, and it has landed.

**The one thing still outstanding is human.** @pkarw, this PR's `changes-requested` review is still the active decision, and it is the only thing between here and merge. All five of your findings are addressed in `40d19f9`, the reasoning for each is in the run summary above, and CI is green. I cannot clear the review myself — `om-auto-fix-pr` ran under the account that authored this PR, so GitHub will not accept a self-approval, and this skill never merges regardless. Once you re-review and approve, `om-approve-merge-pr 5614` owns the merge; `skip-qa` applies under the automated-verification exemption, so no QA sign-off is outstanding.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-25T23:54:50Z. Re-review of head `40d19f9` after the author's autofix commit. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review: Re-review — fix(shared/query): promote BasicQueryEngine columnExists cache to process scope (#5605)

## 🎯 Summary

This is a re-review of head `40d19f9`, the follow-up commit to the `changes-requested` review posted on `8ae0199`. The change replaces `BasicQueryEngine`'s `private columnCache` instance field with a module-level `columnExistsCache`, and makes `columnExists()` remember `false` answers instead of deleting them — so the `information_schema.columns` probe is no longer re-paid on every request by the fresh engine that `createRequestContainer()` builds per HTTP request. The follow-up commit adds the two things the previous review asked for: the map is now bounded (`OM_QUERY_COLUMN_EXISTS_CACHE_MAX_ENTRIES`, default `10000`, with an expired-entry sweep before a wholesale clear) and every entry carries a TTL (`OM_QUERY_COLUMN_EXISTS_CACHE_MS`, default `300000`, `0` disables).

I verified all five previous findings independently rather than reading them off the author's run summary, and all five are genuinely resolved:

- **The major** — the unbounded map keyed partly by unvalidated request input — is fixed in `packages/shared/src/lib/query/engine.ts:114-124`, and the bound is pinned by a test that pushes twenty distinct caller-supplied sort field names through a cap of four and asserts `columnExistsCacheSize()` after every iteration. That assertion is what stops a future refactor from quietly dropping the bound, which was the substance of the finding.
- **Minor 1** — a stale negative silently dropping a scope guard rather than merely slowing a query — is bounded by the TTL, and the test at `engine.test.ts:1487` proves the convergence rather than asserting the plumbing: it stubs `Date.now()`, shows a cached `false` suppressing the `tenant_id` predicate while fresh, then shows the predicate reappearing after the TTL. Driving it off a stubbed clock rather than real time is the right call.
- **Minor 2** — the clearing helper wired into only one of the three spec files sharing the state — is fixed in both siblings.
- **Minor 3** — the `information_schema.tables` half of #5605 — is resolved as documentation, which is one of the two resolutions the previous review offered. I confirmed **#5619** exists and is open, and that #5605 carries the re-scoping comment; the deferral is traceable rather than silent.
- **The nit** on "process-scoped" overstating what the module system guarantees is fixed in the comment.

I also checked the design property that the whole change rests on and that nothing so far had verified: whether a module-scoped `table.column` key space can collide across databases. It cannot. `getOrm()` (`packages/shared/src/lib/db/mikro.ts:118`) memoizes a single ORM built from one `DATABASE_URL`, and both `BasicQueryEngine` construction sites (`packages/shared/src/lib/di/container.ts:213`, `packages/core/src/modules/query_index/di.ts:117`) take their `em` from it, so every entry in the shared map describes the same database. `HybridQueryEngine` keeps its own instance-scoped `columnCache` (`packages/core/src/modules/query_index/lib/engine.ts:205`) and is untouched, as the PR states. The one runtime schema-mutation path that would make a cached negative dangerous — `DefaultDataEngine.ensureStorageTableExists()` — turns out only to probe and throw, never to create, so it cannot introduce a column behind a cached `false`.

The remaining points are all optional polish. Nothing here blocks the merge.

## Verdict

✅ **approve** — every finding from the previous review is addressed with a test that would fail against the earlier code, the module-scoped key space is provably single-database, and the whole configured validation gate is green on this exact head. The four nits below are the author's call and do not need another round.

## 🧪 Validation Gate

Validation source: **GitHub checks** (no local fallback needed). Every command in `.ai/agentic.config.json` maps to a green step on head SHA `40d19f9e45d3aac59efc6e35a3a5a8813ad2614c`, confirmed by reading the job step lists directly rather than trusting the summary check names.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job, step "Build packages" ([job 97804326771](https://github.com/open-mercato/open-mercato/actions/runs/32848707282/job/97804326771)) |
| `yarn generate` | ✅ PASS | `prepare` job, step "Prepare generated modules" |
| `yarn build:packages` (second pass) | ✅ PASS | `prepare` job, step "Rebuild packages with generated files" — the load-bearing repeat that compiles against the generated registries |
| `yarn i18n:check-sync` | ✅ PASS | `test` job, step "Check i18n sync" ([job 97805517340](https://github.com/open-mercato/open-mercato/actions/runs/32848707282/job/97805517340)) |
| `yarn i18n:check-usage` | ✅ PASS | `test` job, step "Check i18n usage" |
| `yarn typecheck` | ✅ PASS | `test` job, step "Checking types" |
| `yarn test` | ✅ PASS | `test` job, step "Test" — the job that runs the three new regression tests |
| `yarn build:app` | ✅ PASS | `prepare` job, step "Build app", and again per shard in `ephemeral-integration` (15/15 green) |

Beyond the gate: `lint`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance` and `license/cla` are green, and `mutate (shared, src/lib/query/engine.ts)` passed against the new implementation — independent evidence that the added TTL-expiry, cap-bound and cache-disabled tests detect changes to `engine.ts` rather than merely executing it. Five checks are `skipping` by their own path or branch filters (`docker-build`, `audit`, `merge-coverage`, `comment`, `[code]smith`), none of which gate this diff. The head is one commit behind `origin/develop` (`8bbdd53d2`, a CRUD correlation-id fix touching nothing under `packages/shared/src/lib/query`, `packages/shared/src/lib/data` or `query_index`), so the green run is representative of the merge result.

## Findings

### 💅 Nit

**1. `OM_QUERY_COLUMN_EXISTS_CACHE_MS=0` is not "probe per request again" — it is per projection pass.** `packages/shared/src/lib/query/engine.ts:1601-1614`, with the claim at `engine.ts:105` and mirrored at `engine.ts:716`. Removing `this.columnCache` left no per-request layer underneath the module cache, so with the TTL set to `0` nothing is stored at all and every `columnExists` call reaches the database: `buildQuery` runs for the `full`, `sortKeys` and `count` projections, so a present column is probed up to three times per `.query()` where the pre-PR instance cache probed it once per request. The escape hatch is therefore slightly worse than the behavior it claims to restore, and the comment at `engine.ts:716` ("every `columnExists` check is memoized on the module-scoped `columnExistsCache`, so later passes hit no extra DB calls") is only true while the knob is at its default. The precedent this PR cites keeps both layers for exactly this reason — `packages/shared/src/lib/search/availability.ts:56` describes "a per-request instance memo, and a process-level TTL cache". Keeping a small instance-level memo under the module cache would make `=0` degrade to parity instead of a regression; failing that, the two comments should say "per query" rather than "per request".

**2. The two new env knobs are documented only in the source.** `packages/shared/src/lib/query/engine.ts:105-108`. `OM_QUERY_COLUMN_EXISTS_CACHE_MS` is the only operator-facing control over the failure mode this PR itself documents (a cached negative suppressing a tenant or organization scope predicate), and an operator cannot discover it without reading the engine. The precedent the PR models itself on documents its knob in `apps/mercato/.env.example:277-278`, in `packages/create-app/template/.env.example`, and in the env table at `packages/search/AGENTS.md:559`. I am filing this as a nit rather than a convention violation because the convention is not actually settled: 17 of the 37 `OM_*` knobs referenced under `packages/shared/src` are absent from `apps/mercato/.env.example`, including close analogues like `OM_CF_DEF_CACHE_TTL_MS` and `OM_ENCRYPTED_SORT_MAX_ROWS`. If you do add them, note that `apps/mercato/.env.example` is template-synced, so it needs `yarn template:sync:fix` in the same commit.

**3. Cap eviction is wholesale, so distinct-name pressure evicts the hot scope columns too.** `packages/shared/src/lib/query/engine.ts:114-124`. When the sweep finds nothing expired, `columnExistsCache.clear()` drops everything, including the `organization_id` / `tenant_id` / `deleted_at` entries that carry essentially all of the win. A caller cycling enough distinct `sortField` values keeps the map at the cap and thereby keeps clearing it, degrading the engine back to probing per request. This is never worse than the pre-PR baseline (which probed every request unconditionally) and it matches `availability.ts` exactly, so it is not a regression and I would not hold the PR for it. If you ever want to harden it, the sturdier shape is to cache negatives only for the fixed scope columns and leave caller-supplied names uncached — they are looked up once and never again, so caching them is pure pollution that pays for itself only in eviction pressure.

**4. `hybrid-engine.test.ts` has the same latent ordering coupling that minor 2 fixed elsewhere.** `packages/core/src/modules/query_index/__tests__/hybrid-engine.test.ts:1350`. That file constructs a real `new BasicQueryEngine(em)` as the hybrid engine's fallback and declares nine mutually contradictory per-test `columns` fixtures (for example the `users`-only fixture at line 1432) without clearing the new module memo. It is green today and most of its tests run with `hasIndexAny: true`, where the hybrid's own instance-scoped cache does the probing — I could not confirm by execution whether the fallback engine's `columnExists` is reached there, since this review worktree has no dependency install and green CI made a local run unnecessary. So this is defensive only: adding `clearColumnExistsCache()` to a `beforeEach` there costs two lines and removes the possibility that a future test added to that file inherits an answer from an unrelated fixture, which is precisely the coincidence minor 2 was about.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `clearColumnExistsCache()` and `columnExistsCacheSize()` are new additive exports, mirroring the existing `clearEncryptedLikeFieldsCache()` convention in the same file. The removed `columnCache` was `private`.
- [x] No config key renamed and no default changed silently — both new env knobs are additive, and their unset behavior is the intended production behavior
- [x] Where a contract had to change: not applicable; no contract surface moved

## 🧪 Test Coverage

Coverage is good and, more importantly, adversarial — each new test is written so it would fail against the code as it stood before the commit it accompanies, which is the property that makes a regression test worth having.

`packages/shared/src/lib/query/__tests__/engine.test.ts:1419` adds five cases. Two cover the original change: a second `BasicQueryEngine` instance (standing in for a second HTTP request) reuses the first one's answer with no additional `information_schema.columns` probe, and an absent `organization_id` is probed exactly once per `.query()` rather than once per projection. Three cover the follow-up commit: the TTL-expiry case asserts both halves of what matters — that the probe happens again *and* that the `tenant_id` predicate reappears on the base query — so it pins the security-relevant behavior rather than just the cache mechanics; the cap case asserts the entry count after every one of twenty iterations; and the disable case asserts both that nothing is stored and that probes increase.

The two sibling spec files now clear the memo in `beforeEach`, which removes a real ordering coupling: their fixtures declare contradictory `information_schema.columns` shapes for the same `scheduled_jobs` table, so before this the first test's answers decided later tests' scoping and the suite passed by coincidence.

The one gap I would name is the eviction path in nit 3: the cap is asserted, but nothing pins *which* entries survive a sweep, so a change from "sweep expired, then clear" to a different eviction policy would not be detected. That is only worth adding if you act on nit 3 — as written, the behavior is inherited verbatim from `availability.ts` and is covered there in spirit.

---

_Reviewed at head `40d19f9e45d3aac59efc6e35a3a5a8813ad2614c`. Validation source: GitHub checks._

pkarw:

Re-review of head `40d19f9`. All five findings from the first review are addressed — verified independently against the diff rather than from the run summary — and the follow-up commit's own behaviour is pinned by tests that fail against the earlier code. No blockers, no majors, no minors; four nits are listed in the review and are the author's call.

I also verified the design property the change rests on and that nothing had checked yet: `getOrm()` memoizes a single ORM from one `DATABASE_URL`, and both `BasicQueryEngine` construction sites take their `em` from it, so a module-scoped `table.column` key space cannot collide across databases.

Validation source: **GitHub checks** — no local fallback was needed. All eight commands in `validation.commands` map to green steps on head `40d19f9e45d3aac59efc6e35a3a5a8813ad2614c`, confirmed by reading the `prepare` and `test` job step lists directly rather than trusting the summary check names. CI was already fully settled before this run started, so no CI follow-up is owed and `ci-monitoring` was not applied.

Autofix: skipped (not my PR — nothing was actionable anyway). Pipeline label moved to `merge-queue`; the merge itself belongs to `om-approve-merge-pr 5614`.

Issue #5232, opened 2026-08-12
Implement: Address-Level Contact Details and Tax Identifiers


pkarw:
## 📝 Spec

- Document: `.ai/specs/2026-08-10-address-contact-and-tax-fields.md`
- Design PR: https://github.com/open-mercato/open-mercato/pull/5197 (merged into `develop` on 2026-08-12)

## 🎯 Summary

An address in Open Mercato models nine postal fields and nothing else, so it cannot carry a phone number or a tax identifier — and both are ordinary requirements of fulfilment: a carrier needs a contact number (mandatory for parcel-locker and pickup-point delivery), and a B2B invoice address is incomplete without the tax identifier the document was issued under. They belong on the address rather than the customer, because one customer keeps several addresses each with its own recipient and contact. Shopify, Medusa and commercetools all model a phone on the address; Open Mercato is the outlier.

The spec adds optional `phone` / `taxId` / `taxIdType` fields to `AddressValue`, rendered by `AddressView` as a **caller-opt-in** contact block — omit `contactLabels` and the output is byte-identical to today. On the sales side the existing schemaless, encrypted-at-rest document address snapshot is the carrier, so no schema change is needed there; typed `recipient_name` / `phone` columns on `CustomerAddress` arrive later and additively. `taxIdType` is Stripe-shaped `{country}_{kind}`, seeded `eu_vat` / `pl_nip` / `other`, with EU VAT normalized to the ISO-2-prefixed form on write and a type-aware display gate keyed on it.

## 📋 How to implement

- The spec PR is merged into `develop`, so this is actionable now. Pick it up with `/om-implement-spec` (which resolves the spec, audits readiness, and phases the work) using the document as the brief; `/om-auto-fix-issue` is the wrong tool here because this is a multi-phase feature, not a scoped bug.
- Implement in the spec's phase order — the phases are independently shippable and the later ones depend on the earlier surface:
[...]

PR #5197, opened 2026-08-11
docs(specs): address-level contact details and tax identifiers

kamwro:
## Summary

An address in Open Mercato cannot carry a phone number or a tax identifier. `AddressValue` models
nine postal fields, and every formatter and renderer derives from those. Both are ordinary
requirements of order fulfilment — a carrier needs a contact number to deliver, and a B2B invoice
address is incomplete without the tax identifier the document was issued under — and both are
per-address rather than per-customer facts, since one customer keeps several addresses.

Shopify (`MailingAddress.phone`), Medusa (`Address.phone`) and commercetools (`Address.phone`) all
model this; Open Mercato is the outlier that does not.

This PR adds **only the spec**. No code. It is opened to get the modelling decisions settled before
implementation, and two of its Open Questions are deliberately unresolved — the compliance verdict
inside the document says **blocked** on them:

- **Q0** — one spec or two (contact fields vs tax-id semantics)?
- **Q3/Q4** — the shape of `taxIdType`: Stripe-style `{country}_{kind}`, or a minimal
  `eu_vat | local | other` enum? Presented as a genuine fork with no recommendation, because Stripe
  treats `pl_nip` and `eu_vat` as distinct types (examples `1234567890` vs `PL1234567890`) — which is
  exactly the ambiguity the spec describes, and it cuts against the smaller enum.

Reviewed and approved by @jtomaszewski before upstreaming.

## Changes

- Adds `.ai/specs/2026-08-10-address-contact-and-tax-fields.md`

## Specification

**Does a spec exist for this feature/module?**
- [ ] Yes
- [x] No (created a new spec)
- [ ] N/A (minor change, no spec needed)

**Spec file path:**
`.ai/specs/2026-08-10-address-contact-and-tax-fields.md`

## Testing

No code, so no tests run. The document was checked against
`.ai/skills/om-spec-writing/references/spec-checklist.md`: all ten required sections present, Risk
Register in the required format, Final Compliance Report with an explicit verdict. Integration
coverage for the implementation phases is listed inside the spec (`TC-SALES-ADDR-CONTACT-001`, plus
the `TC-*-CRUDFORM-*` sweep for the Phase 3 customer-address fields) and ships with those phases,
not with this PR.

## Checklist

- [x] This pull request targets `develop`.
- [ ] I have read and accept the Open Mercato Contributor License Agreement (see `docs/cla.md`).
- [x] I updated documentation, locales, or generators if the change requires it.
- [ ] I added or adjusted tests that cover the change.
[...]

PR #5616, opened 2026-08-25
feat(customers,sales): an address carries a phone and the tax id it was invoiced under

kamwro:
## Summary

An address models nine postal fields and nothing else, so it cannot carry the phone a carrier calls
about a delivery — mandatory for parcel-locker and pickup-point shipping — or the tax identifier an
invoice address was issued under. Shopify, Medusa and commercetools all model a phone on the address.

On the sales side the keys were already arriving and being persisted: a document's address snapshot is
schemaless, so an integration posting `phone` or `taxId` beside the postal fields has always had them
stored. They had no read path and no write path — nothing in the backoffice showed them, and the
address editor had no field for them, so a manual save wrote back a snapshot they were missing from.

This is Phase 1 of the merged spec. Two smaller defects in the same section come with it: the two
snapshot address editors rendered as an editable form on a locked document while every sibling
control honoured `locked`, and `AddressesSection.tsx` carried `// @ts-nocheck`, hiding 1,429 lines
from `tsc` — including everything this phase adds there.

## Changes

- `AddressValue` gains optional `phone`, `taxId` and `taxIdType`.
- `AddressEditor` renders the tax identifier, its scheme and the phone as ordinary fields — always
  present, editable, disabled with the rest of the form rather than on their own.
- The scheme is **picked from a list**, not inferred from the value. Reading it off the form of the
  number works for the seeded cases and stops working as the vocabulary grows, and a wrong type is
  worse than none because it puts a name on the number.
- `resolveTaxIdLabel` maps a scheme to its label, so the picker and anything else that names an
  identifier resolve through one mapping.
- Two opt-in props, `showPhoneField` and `showTaxIdField`, both off by default: only a caller whose
  storage can hold a field should offer it. The document tiles and the document create form opt in;
  the customer address book does not, because `CustomerAddress` has no column for either until
  Phase 3 and an input that drops its value on save is worse than no input.
- Both document normalisers own the two keys, including the `SalesDocumentForm` copy whose assign
  list stopped at `country` and silently discarded them on the create page.
- `disabled={locked}` on both snapshot editors; `// @ts-nocheck` removed with the errors it hid.

`AddressView` is untouched, so every existing render is byte-identical.

## Specification

- [x] Yes
- [ ] No (created a new spec)

**Spec file path:**
`.ai/specs/2026-08-10-address-contact-and-tax-fields.md`

Amended in the same commits as the code it describes. Three decisions changed during review — a
type-aware visibility gate was specified and reversed, a contact-rendering API was drafted and
dropped before merge, and the scheme moved from derived to picked. Each rewrote the paragraph it
invalidated rather than leaving a changelog note for the reader to apply, with one changelog line per
milestone.

## Testing

- `yarn jest` in `packages/core` — **1,287 suites / 10,098 tests**, green.
- `yarn jest` in `packages/ui` — green.
- `tsc --noEmit` on both packages — clean.
- `TC-SALES-ADDR-CONTACT-001` against a live local instance — **2 passed**: the snapshot round-trips
  the keys through the API, and the document detail page renders both values with the scheme's label.
- Clicked through locally: the picker offers the three seeded schemes, a saved choice survives a
  reload, the customer address book offers neither field, and a locked document disables both.

Tests added, and the bug each catches:

- `customers/components/__tests__/AddressEditor.contact.test.tsx` — the inputs, which no other suite
  reaches because every section that renders this editor mocks it. Hidden by default, either offered
  without the other, disabled with the form rather than alone, and the scheme's label following the
  selection.
- `customers/utils/__tests__/addressFormat.test.ts` — the label mapping: a domestic identifier and an
  EU VAT number named differently, an unknown or absent scheme falling to the neutral label, and a
  plain string accepted so a caller that does not care keeps working.
- `sales/.../__tests__/AddressesSection.test.tsx` — the save round-trip: unowned keys survive, a
  cleared field stays cleared, emptying everything yields `null`, and a saved address shows no
  snapshot tax id so stale details cannot survive the switch.

Not covered: format validation per scheme. Picking `eu_vat` does not yet constrain what may be typed
beside it — that belongs with `normalizeEuVatId()` in the next phase.

## Checklist

- [ ] QA routing set.

The four unticked boxes are not oversights. The CLA is the author's to accept and is left for them.
The label boxes need `triage` on this repository, which this account does not have — the intended
values are in a comment below for a maintainer to apply.

### Design System Compliance
- [ ] `aria-label` on all icon-only buttons (`<IconButton>`)

The three unticked boxes do not apply: this change adds form fields to an existing editor, not a list
or data page, no async page, and no icon-only button. The first two were verified against the diff
rather than assumed — it introduces no `(text|bg)-(red|green|emerald|amber|blue)-*` and no
`text-[Npx]`.

## Linked issues

Part of #5232. Phase 1 of the spec merged as #5197; builds on Phase 0, #5240, already on `develop`.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5616"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790251451&installation_model_id=432243&pr_number=5616&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5616&signature=6b86d4257b30b04e40cf0ca19557fcc8f19b4adefd0d00115ecdc32efa1ff167"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


kamwro:
**Metadata this account cannot apply** — it has `pull` only on this repository, so the label boxes in
the checklist are left unticked rather than silently skipped. Intended values, for a maintainer:

| Label | Why |
| --- | --- |
| `priority-medium` | A gap in an existing surface rather than a defect: the keys were already being persisted and simply had no read or write path. Nothing is broken today for a deployment that does not write them. |
| `risk-medium` | `AddressValue` and `AddressEditor` are shared, so the blast radius is every address form — but `AddressView` is untouched and both new fields are off by default, so a caller that does not opt in renders byte-identical output. The one behavioural change reaching existing callers is `disabled={locked}` on the two document snapshot editors, which corrects a form that accepted input the API refuses. |
| `needs-qa` | Customer-facing UI: two new inputs and a picker on the document address tiles. Not `skip-qa` — there is a manually exercisable surface, which is the line `AGENTS.md` draws. |
| `feature` | Adds a capability rather than fixing a regression, though it carries two defect fixes in the same section. |

**QA evidence, offered rather than claimed** — the gate stays closed until a maintainer applies the
labels, and this does not clear it:

- `TC-SALES-ADDR-CONTACT-001` run against a live local instance: 2 passed, covering both the API
  round-trip and the page render.
- Clicked through locally: the picker offers the three seeded schemes; a saved choice survives a
  reload; leaving the scheme empty still saves, since it is not required; the customer address book
  offers neither field; a locked document disables both.
- `packages/core` 1,287 suites / 10,098 tests green, `packages/ui` green, `tsc` clean on both.

A reviewer is requested by role rather than by handle: this is a partner contribution, so whoever
routes those.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-25T12:14:13Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: feat(customers,sales): an address carries a phone and the tax id it was invoiced under

## 🎯 Summary

This is Phase 1 of `.ai/specs/2026-08-10-address-contact-and-tax-fields.md`: `AddressValue` gains optional `phone` / `taxId` / `taxIdType`, `AddressEditor` renders them as ordinary fields behind two opt-in props, both document normalisers learn to write them, the two snapshot editors on the document detail finally honour `locked`, and `// @ts-nocheck` comes off `AddressesSection.tsx`. The reviewed scope is the full 15-file diff (`packages/core/src/modules/customers`, `packages/core/src/modules/sales/components/documents`, `packages/ui/src/backend/detail`, the five customers locale files, the spec and the new integration spec), reviewed at head `5b5d999`.

A lot here is genuinely good and worth saying out loud. Lifting `@ts-nocheck` off a 1,429-line file and fixing the errors it hid — including the wrong `entry is DocumentAddressAssignment` predicate, replaced with a plain `NonNullable` narrow — is real debt paid down, and `yarn typecheck` is clean afterwards. The choice to make the scheme a picker rather than infer it from the value is right, and the argument for it in the spec changelog (`GB123456789` stopped being an EU VAT number at Brexit) is the correct one. Splitting `showPhoneField` from `showTaxIdField` instead of one `showContactFields` is the sort of decision that pays for itself at Phase 3. `AddressView` is genuinely untouched, so existing renders are byte-identical as claimed. `SalesDocumentForm.tsx`'s assign list was really missing the keys, and catching that is the kind of silent create-page data loss that never gets reported as a bug.

Two things hold it back. The branch fails one of this repository's configured validation commands, and it fails in CI too — which nobody has seen, because this fork's workflow runs are sitting at `action_required` and have never executed. And the tax-id label mapping that the code comments, the PR body, the spec changelog and the new integration test all describe as wired up is not wired up: `resolveTaxIdLabel` is imported, never called, and freezes as public API on merge with no caller.

## Verdict

❌ request changes — `yarn i18n:check-sync` fails on this head and passes on `develop`, which is a hard failure of the repo's validation gate and of CI's own "Check i18n sync" step; and `resolveTaxIdLabel` ships as a new frozen export with no production caller, which is precisely the thing the spec's own changelog says it will not do.

## 🧪 Validation Gate

**Validation source: local fallback.** GitHub check data is unusable for this head: `gh pr checks 5616` reports only `[code]smith` (SKIPPED) and `license/cla` (SUCCESS), and the two real workflow runs for `5b5d999` — `CI for Develop&Main` (32845591285) and `Mutation tests` (32845591408) — are both `completed / action_required`, i.e. a fork PR awaiting a maintainer's per-push approval. **There is no test signal on this PR at all**, and an absent run looks exactly like a clean one in the checks UI. That is why the gate was reproduced locally. Runner: local (no compose `app` container running).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn generate` | ✅ PASS | No drift — `git status` clean afterwards. |
| `yarn build:packages` (2nd) | ✅ PASS | |
| `yarn i18n:check-sync` | ❌ **FAIL** | `5 issues found across 1 modules` — `[customers] {en,pl,es,de,ko}.json: format issue (unsorted keys)`, exit 1. Verified against the base: with `origin/develop`'s copies of those five files restored, the same command reports `All translation files are in sync`. The regression is this diff's. |
| `yarn i18n:check-usage` | ✅ PASS | Advisory only (`continue-on-error: true` in CI); the 3,683 unused keys it reports are pre-existing. |
| `yarn typecheck` | ✅ PASS | Notable: clean *after* `@ts-nocheck` was removed from `AddressesSection.tsx`. |
| `yarn test` | ✅ PASS (scoped) | `packages/core`: `src/modules/customers` + `src/modules/sales/components/documents` → **240 suites / 1,452 tests passed**. `packages/ui`: `src/backend/detail` → 4 suites / 29 tests passed. Scoped rather than workspace-wide because the diff touches no shared contract beyond the two additive `AddressValue` fields. |
| `yarn lint` (extra) | ✅ PASS | 0 errors, 10 pre-existing warnings, none in changed files. |

## Findings

### ⛔ Blocker

**1. `packages/core/src/modules/customers/i18n/en.json:1870` (and `de.json`, `es.json`, `ko.json`, `pl.json` at the same position) — the new locale keys are in the wrong sort position, so `yarn i18n:check-sync` fails and CI fails with it.**

`"customers.people.detail.addresses.fields.taxIdType"` is inserted immediately after `...fields.taxId` and *before* `...fields.taxId.euVat` / `.other` / `.plNip`. The checker sorts by the raw key string, where `.` (0x2E) sorts before `T` (0x54), so the required order is `taxId`, `taxId.euVat`, `taxId.other`, `taxId.plNip`, `taxIdType`. All five locale files carry the same mistake.

The concrete failure: `.github/workflows/ci.yml:592-593` runs `yarn tsx scripts/i18n-check-sync.ts` with **no** `continue-on-error`, unlike the `i18n:check-usage` step directly below it. The job fails. This has gone unnoticed only because the fork's runs have never been approved to start.

Fix — one command, and it produces exactly the five one-line moves:

```bash
yarn i18n:check-sync --fix
```

### ⚠️ Major

**2. `packages/core/src/modules/customers/components/AddressEditor.tsx:27` and `:160` — `resolveTaxIdLabel` and `taxIdLabels` are dead code, so the tax-id label marker described everywhere in this PR does not exist.**

Line 27 imports `resolveTaxIdLabel`; lines 160-164 build the `taxIdLabels` map from three `t()` calls. Neither is referenced anywhere else in the file — the picker at line 566 maps `TAX_ID_TYPES` and calls `t(value.key, value.fallback)` directly, and the tax-id `<Input>` at line 573-580 has no `rightIcon`. Across the whole repository, `resolveTaxIdLabel` has exactly two definitions (`packages/core/src/modules/customers/utils/addressFormat.tsx:156`, `packages/ui/src/backend/detail/addressFormat.tsx:155`), one unused import, and one test file. There is no production call site.

Three consequences, in ascending order of seriousness:

- **The UX asymmetry is real and user-visible.** The phone input at line 594-598 renders a right-edge `<span>` naming the field once it has a value, because — as its own comment says — "the placeholder is this field's only label, and it disappears the moment someone types". The tax id has the same problem and no such marker. That comment even opens "Same reason as the tax id above", referring to a marker that is not above it. A user who types `PL1234567890` into the tile is left with a bare number, directly under the postcode, with nothing naming it.
- **It contradicts the code's own documentation.** `AddressEditor.tsx:119` ("The labels are the ones `resolveTaxIdLabel` resolves, so the picker and the marker beside the filled field always read the same") and `:158` ("One mapping, one home. The picker's options resolve through `resolveTaxIdLabel`") both describe wiring that is absent. `taxIdLabels` and `TAX_ID_TYPES` are, in fact, two copies of the same three label lookups — the exact duplication the comment claims to have prevented.
- **It puts a frozen public export into the codebase with no caller.** Per `BACKWARD_COMPATIBILITY.md` the export freezes on merge, and it freezes *twice* (core and ui copies). The spec changelog justifies keeping it with "`resolveTaxIdLabel` stays — the editor's marker is a real caller", two sentences after stating the rule "a public surface whose only consumer is its own test suite commits the project to supporting it forever for nothing". As shipped, the only consumer *is* its own test suite.

Fix — either direction is fine, but the PR has to pick one and make the prose match:

- Wire it, which is what everything here says was intended: `rightIcon={current.taxId ? <span className="text-xs">{resolveTaxIdLabel(taxIdLabels, current.taxIdType)}</span> : null}` on the tax-id `Input`, mirroring the phone; then have the picker's options resolve through the same map so the "one mapping" claim becomes true.
- Or drop `resolveTaxIdLabel` / `TaxIdLabelByType` / `TAX_ID_LABEL_KEY_BY_TYPE` from both `addressFormat.tsx` copies together with `addressFormat.test.ts:106-138`, remove the import, and amend the spec's Migration section — which currently promises the export — accordingly.

**3. `packages/core/src/modules/sales/__integration__/TC-SALES-ADDR-CONTACT-001.spec.ts:107-109` — the assertion that is supposed to prove the tax-id marker passes for an unrelated reason, so the suite is green over behaviour that does not exist.**

```ts
// A filled field keeps its name at the right edge, and for the tax id that name is chosen by
// `taxIdType` — `PL1234567890` is an EU VAT number, not a domestic one.
await expect(page.getByText('EU VAT')).toBeVisible()
```

There is no right-edge name on the tax id (finding 2). The string `EU VAT` *is* on the page, because `taxIdType` is `eu_vat`, so the `<SelectValue>` inside the picker trigger renders the selected option's label. The assertion therefore passes while pinning the picker, not the marker — and it would keep passing if the marker were deleted, which it effectively already has been. This is how finding 2 survived a run the PR body reports as "2 passed".

The unit suite has the mirror-image gap: `AddressEditor.contact.test.tsx:76-80` covers the phone marker (`getByText('Phone')`) and no test covers a tax-id marker, so nothing at any level would have caught it.

Fix: once finding 2 is resolved one way or the other, make this assertion target the marker specifically rather than any occurrence of the string — for example scope it to the input's container, or assert `getByText('EU VAT')` has a count of 2 (picker + marker) — and add the matching unit case beside the phone one. If the export is dropped instead, delete the assertion and the comment rather than leaving a green test describing a feature that was cut.

### 🔹 Minor

**4. `packages/core/src/modules/customers/components/AddressEditor.tsx:90-95` — a JSDoc block about tax-id labels is attached to `showFormatHint`.**

The comment reads "Interprets `taxId` for its label only — a `pl_nip` reads 'NIP', an `eu_vat` reads 'EU VAT'. It is metadata about the value, never an edited field, which is why it is a prop rather than a member of the draft: nothing in this component may write it." It sits directly above `showFormatHint?: boolean`, a boolean about address-format hints that has nothing to do with any of it. This is clearly left over from the `taxIdType`-as-a-prop design the PR describes as reversed — but the effect is that anyone hovering `showFormatHint` in an editor, or reading the props block top to bottom, gets documentation for a different feature. Delete the block.

**5. `packages/core/src/modules/customers/utils/addressFormat.tsx:133-145` and `packages/ui/src/backend/detail/addressFormat.tsx:132-144` — an orphaned JSDoc describes a function that was dropped before merge.**

The twelve-line comment beginning "The address's own contact details, as `[label, value]` pairs — only the fields the caller labelled AND the address actually carries. Exported so a caller can ask 'is there anything to show?' without rendering" documents `formatAddressContactPairs`, which the spec changelog says was removed. It now sits on `const TAX_ID_LABEL_KEY_BY_TYPE`, a private two-entry `Record`. Nothing in it is true of the thing it annotates: that const is not exported, returns no pairs, and answers no such question. Both copies need the same edit — replace it with a line about what the record actually is, or drop it.

**6. `.ai/specs/2026-08-10-address-contact-and-tax-fields.md` — two statements no longer match the implementation the same commits ship.**

The spec is otherwise carefully maintained — rewriting invalidated paragraphs instead of appending changelog notes is the right instinct and it mostly worked. Two survived:

- The Implementation Breakdown note reads "`AddressEditor` takes `showContactFields`, off by default, in the shape `showCoordinateFields` already established." The code ships `showPhoneField` and `showTaxIdField`, which the Phase 3 note four paragraphs later correctly names and explains. Update the earlier sentence.
- The i18n section still reads "Two keys added: `sales.documents.detail.addresses.{taxId,phone}` in `en`, `pl`". This PR adds six keys under `customers.people.detail.addresses.fields.*` across five locales. That is a bigger surface than the spec budgets, and the sort-order break in finding 1 lands squarely in it.

**7. `packages/core/src/modules/sales/components/documents/AddressesSection.tsx:1203-1204` — misindented props.**

`showPhoneField` and `showTaxIdField` sit at 24 columns inside a `<AddressEditor>` whose other props sit at 20, unlike the correctly-aligned pair at `:1133-1134`. Nothing enforces it (lint is clean), so it will simply stay wrong. Re-indent to match the siblings.

**8. `packages/core/src/modules/sales/components/documents/__tests__/AddressesSection.test.tsx:133-135` — an empty `beforeEach` was added immediately above the real one.**

```ts
beforeEach(() => {
})

beforeEach(() => {
  jest.clearAllMocks()
  …
```

Debris from a working copy. Delete it.

**9. `packages/core/src/modules/customers/utils/__tests__/addressFormat.test.ts:118-121` — a test that asserts nothing its name claims.**

`it('treats every prefixed number as EU VAT, not just the local country')` contains a single assertion, `expect(resolveTaxIdLabel(BY_TYPE, 'eu_vat')).toBe('EU VAT')`, byte-identical to line 114 in the test directly above it. `resolveTaxIdLabel` never sees a number — it takes a label map and a type — so no country-prefix behaviour can be pinned from here at all. Either delete the case as a duplicate, or move the claim to wherever a number actually becomes a type (which, since `deriveTaxIdType` was reverted out, is nowhere in this PR — in which case delete it).

**10. `packages/ui/src/backend/detail/addressFormat.tsx:155` — the `ui` copy of `resolveTaxIdLabel` has no test.**

The two `addressFormat` modules are edited in parallel here and the new function is byte-identical in both, but only `packages/core`'s copy is covered (`addressFormat.test.ts`). The twin can drift with nothing to catch it. The spec already defers collapsing the two into one module, which is the right call for this PR — but a small mirror suite in `packages/ui` costs little and pins the invariant until then. (Contingent on finding 2: if the export is dropped, this disappears with it.)

### 💅 Nit

**11. `packages/core/src/modules/customers/components/AddressEditor.tsx:566` — `{TAX_ID_TYPES.map((value) => (` shadows the component's own `value` prop** (the address draft) inside the picker's options. It is harmless today because the outer `value` is not read in the callback, but it makes the four lines below it read as though they touch the address. `scheme` or `option` would say what it is.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the change is purely additive. `AddressView`'s signature and output are untouched, verified against the diff.
- [x] No function signature changed in a breaking way — `AddressEditor`'s two new props are optional and default to `false`, so every existing caller compiles and renders identically; `normalizeAddressDraft` keeps its arity in both copies and only widens its return type from `Record<string, unknown>` to `NormalizedAddressDraft` (a supertype-compatible intersection with the same index signature).
- [x] No required type field removed or narrowed — `phone`, `taxId`, `taxIdType` are added to `AddressValue` and `AddressEditorDraft` as optional.
- [x] No field removed or retyped in an existing response shape — the sales snapshot is schemaless and already carried these keys.
- [x] No database table or column renamed or removed — no migration in this PR, correctly, since the snapshot is the carrier.
- [ ] Where a contract had to change: old surface kept working through a deprecation window — **not applicable, but see finding 2.** Nothing here breaks a contract; the issue is the opposite direction. `resolveTaxIdLabel` and `TaxIdLabelByType` are *new* frozen surface under `BACKWARD_COMPATIBILITY.md`, added in two packages, with no production caller. The spec is correctly cited as the required reference, but the justification it gives for the export is not true of the shipped code.

One note on `AddressEditorField`, which gains `'taxId' | 'taxIdType' | 'phone'`: widening that union widens the `errors` prop's `Partial<Record<…>>`, which is additive for callers that pass errors and additive for callers that read them. No action needed — flagging it only because it is a union in a public prop type and worth being deliberate about.

## 🧪 Test Coverage

What is covered, and how:

- `AddressEditor.contact.test.tsx` (new, 7 cases) is the right file to have written — every section that renders this editor mocks it, so without this the inputs would have no coverage at any level. It pins the four states of the two opt-in flags (neither, each alone, both), that the fields are disabled with the form rather than on their own, that a stored `taxIdType` shows in the picker, and that an absent one leaves the picker unset instead of guessing. That last pair is the behavioural core of the "picked, not inferred" decision and it is genuinely pinned.
- `addressFormat.test.ts` gains six cases on the label mapping — the domestic/EU split, `other`, an unknown type (`us_ein`), `null`, `undefined`, the plain-string form, and the no-label form. Good boundary coverage of a small pure function.
- `AddressesSection.test.tsx` gains eight cases. The two that earn their place are the `taxIdType` round-trip through `normalizeAddressDraft` — which is exactly the "seed from snapshot, write back through the assign list" invariant the spec warns about, and it would fail if either half were missed — and the locked/unlocked pair asserting `disabled` reaches every editor instance. Dropping the `addressFormat` mock in favour of the real formatter is a good call and the comment explaining why is the right kind of comment.
- `TC-SALES-ADDR-CONTACT-001.spec.ts` (new) covers the API round-trip and the page render, is self-contained (creates its order via `createSalesOrderFixture`, deletes it in `afterAll`), and does not lean on seeded data — all of which `.ai/qa/AGENTS.md` requires.

Gaps, in the order I would close them:

1. **No test fails when the gate does.** Findings 1 and 2 are both mechanically detectable and both slipped through a locally-green run. Nothing to add in the suite for the locale ordering — `i18n:check-sync` is that test — but it is worth noting that the local `yarn jest` runs reported in the PR body do not include it, and CI has never run on this branch, so "green" here has meant less than it looks.
2. **The tax-id marker has no test** (finding 3). Add the mirror of `AddressEditor.contact.test.tsx:76-80` — render with `value: { taxId: 'PL1234567890', taxIdType: 'eu_vat' }, showTaxIdField: true` and assert the marker text is present *in addition to* the picker's, which is the assertion that distinguishes the two and that the current integration check cannot make.
3. **`SalesDocumentForm.normalizeAddressDraft` has no unit coverage for the new keys.** The PR correctly identifies this copy as a real bug — its assign list stopped at `country`, so the create page silently discarded whatever the editor wrote — but the only thing pinning the fix is the integration spec, which exercises the *update* path on an existing order, not create. A three-line unit test asserting `normalizeAddressDraft({...draft, taxId, taxIdType, phone})` includes all three keys would pin the very regression this PR was written to fix, in the file where it happened. This is the gap I would most want closed.
4. **The `ui` twin is untested** (finding 10).
5. **Acknowledged and correctly deferred:** per-scheme format validation. Picking `eu_vat` does not constrain what may be typed beside it. The PR says so explicitly and puts it with `normalizeEuVatId()` in Phase 2, which is the right place.

---

### 🚦 CI status

No required checks are configured on `develop` (branch protection returns 404 for required status checks), and the only checks reporting on this PR are `[code]smith` (SKIPPED) and `license/cla` (SUCCESS). The substantive workflows for head `5b5d999` — `CI for Develop&Main` and `Mutation tests` — are both `action_required`: **a maintainer needs to approve the workflow runs for this fork PR.** Until then this PR has no CI signal, and the blocker above is exactly what that missing signal would have reported. Nothing was pending at review time, so there is no CI follow-up to wait on.

### Inherited review feedback

None. There are no prior reviews and no inline review comments on this PR. The author's own comment listing intended labels is metadata for a maintainer, not an actionable review ask, and it has been applied below.

pkarw:
🏷️ label rationale

| Label | Why |
| --- | --- |
| `changes-requested` | Review verdict is request-changes: one blocker (`yarn i18n:check-sync` fails on this head and passes on `develop`) plus two majors. Replaces the pipeline state; it moves back to `review` on the next push. |
| `feature` | Adds a capability — address-level phone and tax identifier — rather than fixing a regression, matching the author's own reading. The two defect fixes it carries (`disabled={locked}`, `@ts-nocheck` removal) ride along inside the same spec phase. |
| `priority-medium` | Applied as the author proposed, and it holds up: the snapshot keys were already being persisted, so nothing is broken today for a deployment that does not write them. This closes a gap rather than stopping a bleed. |
| `risk-medium` | Also as proposed. `AddressValue` and `AddressEditor` are shared surface, so the blast radius is every address form — but both new fields are off by default, `AddressView` is untouched, and `yarn typecheck` / `yarn build:app` are clean. The one behavioural change reaching existing callers is `disabled={locked}` on the two snapshot editors, which is a correction. Not `risk-high`: no schema change, no route change, no contract break. |
| `needs-qa` | The diff changes `.tsx` under `packages/core/src/modules/**/components/**`, so the automated-verification exemption in `AGENTS.md` does not apply — there is a manually exercisable surface (two new inputs and a picker on the document address tiles). Not `skip-qa`. |

The author's QA evidence comment is noted and is **not** treated as clearing the gate: `qa-approved` / `qa-self-verified` are not applied here. Per `AGENTS.md` those are earned by manual QA or the self-QA exception and applied by a maintainer, and this PR must not merge while it carries `needs-qa` without `qa-approved`.

`partner-request` left untouched.

**Separately, and needed before anything else can be judged:** the workflow runs for head `5b5d999` (`CI for Develop&Main`, `Mutation tests`) are sitting at `action_required` — a maintainer has to approve CI for this fork PR. There is currently no CI signal on this branch at all, which is how the blocker above went unnoticed.

pkarw:
🤖 `om-auto-review-pr` — handing this back to @kamwro for changes.

The review is above in full. The short version of what to pick up:

1. **Blocker — `yarn i18n:check-sync` fails on this head.** `taxIdType` is inserted before the three `taxId.*` keys in all five customers locale files; the checker sorts on the raw key string, where `.` sorts before `T`. `.github/workflows/ci.yml:592` runs this without `continue-on-error`, so CI fails on it. `yarn i18n:check-sync --fix` produces exactly the five one-line moves. Confirmed against the base: `origin/develop`'s copies of those files pass the same check.
2. **Major — `resolveTaxIdLabel` has no production caller.** It is imported at `AddressEditor.tsx:27` and `taxIdLabels` is built at `:160`; neither is used. The tax-id input has no right-edge label marker, so a typed identifier ends up unlabelled directly under the postcode, while the phone beside it does have one. Either wire the marker (the phone at `:594` is the pattern) or drop the export from both `addressFormat.tsx` copies and amend the spec, whose changelog currently justifies keeping it with "the editor's marker is a real caller".
3. **Major — the integration assertion for that marker passes for the wrong reason.** `TC-SALES-ADDR-CONTACT-001.spec.ts:109` matches `EU VAT` from the picker's `SelectValue`, not from a marker, so it stays green with the feature absent.

Plus seven minors and a nit, all mechanical: a JSDoc block about tax-id labels attached to `showFormatHint`, an orphaned JSDoc for the dropped `formatAddressContactPairs` on a private const in both `addressFormat.tsx` copies, two stale spec statements, misindented props at `AddressesSection.tsx:1203`, an empty `beforeEach`, a duplicate test case, and an untested `ui` twin.

Nothing here is architectural and nothing needs a decision reversed — the design calls this PR makes are, I think, the right ones, and the `@ts-nocheck` removal is real debt paid down. Details, reasoning and suggested fixes are in the review.

**One thing that is not yours to fix:** the workflow runs for `5b5d999` are at `action_required`. A maintainer needs to approve CI for this fork PR — until then the branch has no test signal, which is exactly why the i18n break was invisible.

Labels applied per your metadata comment (`priority-medium`, `risk-medium`, `needs-qa`, `feature`); rationale in the comment above. `qa-approved` is deliberately not applied — that stays with a maintainer.

pkarw:

One blocker (`yarn i18n:check-sync` fails on this head, passes on `develop`), two majors (`resolveTaxIdLabel` dead in production; the integration assertion that should pin its marker passes via the picker instead), seven minors and one nit. Validation source: local fallback — GitHub reports no usable check data for `5b5d999` (both real workflow runs are `action_required`, awaiting a maintainer's fork-CI approval). Locally: `build:packages`, `generate`, `typecheck`, `lint`, scoped `test` (1,452 core + 29 ui) and `build:app` all pass; `i18n:check-sync` fails. Autofix skipped — not my PR. No CI follow-up is outstanding, so no `ci-monitoring`.

kamwro:
All eleven findings are addressed, and the coverage gap you ranked first is closed. Head is `59673b471`.

**Blocker 1 — locale sort order.** `yarn i18n:check-sync --fix`, which produced exactly the five one-line moves you predicted. The gate now passes: `All translation files are in sync`.

**Major 2 — `resolveTaxIdLabel` had no caller.** Wired, in the direction you listed first, because everything in the PR already described it. The tax-id `Input` carries `rightIcon={current.taxId ? <span className="text-xs">{resolveTaxIdLabel(taxIdLabels, current.taxIdType)}</span> : null}`, mirroring the phone, and the picker's options now resolve through the same map — so "one mapping, one home" is true rather than aspirational, and `TAX_ID_TYPES` stopped being a second copy of the labels.

**Major 3 — the assertion that passed for the wrong reason.** `toHaveCount(2)` in the integration spec: the scheme's name appears once in the picker and once in the marker, so the count distinguishes them where visibility could not. The mirror unit case sits beside the phone one and asserts the same count, plus two boundary cases — a filled value with no scheme picked resolves to the neutral name exactly once, and an empty field renders no marker at all.

**Minor 4–11.** The orphaned JSDoc is off `showFormatHint`; the `formatAddressContactPairs` comment is replaced in both `addressFormat` copies with one describing the record it actually annotates; the spec's Implementation Breakdown names the two props and the i18n section names all six keys and the module they live under; the props at `AddressesSection.tsx:1203` are re-indented; the empty `beforeEach` and the duplicate `eu_vat` case are deleted; `packages/ui` gains a mirror suite for its `resolveTaxIdLabel`; and the picker's callback binds `type` instead of shadowing `value`.

**Coverage gap 3 — `SalesDocumentForm.normalizeAddressDraft`.** Closed the way you sketched. The function is now exported, with a comment saying the test is its only consumer and why that is worth the surface: there is no prior snapshot to merge against on a create, so a key the editor writes and this list omits leaves no trace at all. Five cases pin the three keys, the postal fields beside them, blank-omission, trimming, and the null draft. Deleting the three `assign` lines turns two of them red, which is the regression the PR was written to fix.

Two things I did **not** do, both deliberate:

- **I did not collapse the two `addressFormat` copies**, or the two normalisers. You called the deferral right and the spec records it; the `ui` mirror suite pins the twin until then.
- **I did not add per-scheme format validation.** Still Phase 2, still with `normalizeEuVatId()`.

One correction to my own earlier claim, since it bears on what the suite guarantees. I added a `toBeVisible()` pair on the two inputs and described it as catching a field laid out to nothing. It does not: I reintroduced a collapsing layout deliberately and the spec still passed, because the tile is wide enough at this viewport. The assertions stay — a field present in the DOM but off-screen passes `toHaveValue` — but the comment now says what they cover and what they do not, rather than overselling them.

On CI: the runs for this branch are still `action_required`. Whenever you approve them, the i18n step is the one that was failing and now passes locally.

Ready for another look whenever you have time — happy to keep iterating if any of the above lands differently than you intended.

kamwro:
Ready for review @pkarw

kamwro:
@maxidragon is taking over work on this pull request. @pkarw please assign him to this pr

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-29T13:32:43Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
## Summary

The previous review's blocker and major findings are resolved: the i18n gate passes, `resolveTaxIdLabel` now has real production callers, and the marker assertion can no longer pass on the picker alone. The effective merged tree is conflict-free and preserves newer `develop` changes. This re-review found one remaining integration-coverage gap and two Medium-quality issues.

## Findings

### High / major

1. **The promised create/edit/locked UI journeys are not covered by integration tests.** The feature spec explicitly assigns the Playwright case the detail-tab save round-trip, document-form round-trip, and locked-document disabled state (`.ai/specs/2026-08-10-address-contact-and-tax-fields.md:193`), matching the repository rule that key UI paths ship with integration coverage. The added case only seeds and reads the snapshot through the API, then verifies the pre-seeded values on the detail page (`packages/core/src/modules/sales/__integration__/TC-SALES-ADDR-CONTACT-001.spec.ts:68-129`). Its header instead delegates the save and locked paths to component tests (`:17-20`), and the create-form path is covered only through an exported pure helper. Please extend the Playwright coverage to exercise save + reload from the detail tab, create/submit through the document form with API readback, and the locked editor state (or update the implementation/spec scope through the normal spec process if a path is intentionally deferred).

### Medium / minor

2. **The checked-in contract/spec documentation describes a rendering API that does not exist.** Both public `AddressFormatInput` twins say contact fields render through `AddressView` when callers supply labels (`packages/core/src/modules/customers/utils/addressFormat.tsx:16-25`, `packages/ui/src/backend/detail/addressFormat.tsx:15-24`), but `AddressView` gained neither contact rendering nor label props. The active spec repeats that opt-in claim and describes label/value contact blocks (`.ai/specs/2026-08-10-address-contact-and-tax-fields.md:110`, `:143`), while the implementation exposes ordinary `AddressEditor` inputs. The new Playwright header repeats the contact-block wording (`TC-SALES-ADDR-CONTACT-001.spec.ts:11-20`). Please align these docs with the shipped API/UI. While there, correct the nearby stale claims that omitted optional fields render automatically (`AddressEditor.tsx:44-48`) and that two-letter prefixes are always stored as `eu_vat` (`addressFormat.tsx:32-40` in both twins).

3. **Production code exports `normalizeAddressDraft` solely to make a unit test possible.** The new JSDoc says exactly that (`packages/core/src/modules/sales/components/documents/SalesDocumentForm.tsx:438-446`). That widens a component module's callable surface for test convenience and conflicts with this feature's own public-surface discipline. Move the normalizer to a focused helper consumed by both production and its test, or test the form submission through its real call site. The added inline explanatory comments in this block and the Playwright case should also be removed or expressed through names/assertion messages per the repository's no-inline-comments rule.

## Breaking-change checklist

- Types/signatures/import paths: additive optional fields only; no removal or signature break found.
- Routes/events/widgets/DI/ACL/notifications/CLI/generated files: unchanged.
- Database schema/migrations: unchanged in this phase.
- Frozen helpers: the new resolver export has production callers and tests.

## Validation

Validation source: local fallback. GitHub's substantive CI and mutation workflows for head `59673b4` are `action_required`; the only reported completed signals are CLA success and CodeSmith skipped. No required checks are configured.

Runner: local (no running compose app container).

- `yarn build:packages` — PASS
- `yarn generate` — PASS; no tracked drift
- `yarn i18n:check-sync` — PASS
- `yarn i18n:check-usage` — PASS (advisory existing unused-key report)
- Affected core/UI package typechecks — PASS
- Focused core/UI unit tests — PASS (5 suites, 39 tests)
- `yarn template:sync` — PASS
- Broad workspace test/typecheck/app-build — not run; the repo extension requires the narrowest usable local fallback, and the affected-package signals above were available.

## Decision

**Request changes.** Repo policy treats both High and Medium findings as blocking.

pkarw:
Re-review completed on head 59673b4: the previous blockers are cleared, but one High integration-coverage gap and two Medium documentation/test-surface issues remain. I am starting the autonomous carry-forward repair flow now; the original author remains assigned for visibility.

pkarw:
Closing in favor of #5771 (https://github.com/open-mercato/open-mercato/pull/5771).

Credit to @kamwro for the original implementation. The replacement PR carries the same work forward with the requested fixes so it can merge without waiting on the fork branch.

pkarw:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED. One autonomous carry-forward fix iteration completed as replacement #5771; original PR closed as superseded. Lock released.

Issue #4143, opened 2026-07-12
bug(sales): invoice/credit-memo `?ids=` garbage-UUID filter bypass + duplicate-invoice race (PR #1977 QA F3/F4)


Kapsik89:
## Summary

On PR #1977 (head `d6b978688`), the invoice & credit-memo **list and detail pages display the customer's current (live) name instead of the name captured on the source order** (the frozen `customerSnapshot`). If a customer is renamed after the order is placed, every already-issued invoice/credit-memo retroactively shows the new name and disagrees with its own source order. The same root cause defeats the snapshot's resilience to customer deletion.

Found during manual QA of PR #1977 (retest #5). Three additional lower-severity findings are listed at the bottom. One earlier finding (malformed date filter → 500) is already **fixed** in `d6b978688` and is only noted for context.

## Environment

- PR #1977, branch `feat/invoice-credit-memo-pages-v2` @ `d6b978688`
- Fresh greenfield DB · Acme Corp Tenant / Acme Corp · `superadmin@acme.com`
- Production build (`yarn build:app` + `yarn start`) on `:3000`

---

### Steps to reproduce
1. Log in to `/backend`, pick **Acme Corp Tenant** + **Acme Corp** organization.
2. Open order **SO-DEMO-2004** (customer "Copperleaf Design Co.") → **Akcje → Utwórz fakturę**. An invoice is created (e.g. `INV-…-00001`); its detail customer card shows **"Copperleaf Design Co."**.
3. Go to **Firmy** → open **"Copperleaf Design Co."** → change **Nazwa wyświetlana** to **"Copperleaf RENAMED-QA"** → **Zapisz**.
4. Re-open the invoice detail page, and the invoices list (`/backend/sales/invoices`).

### Expected
The invoice keeps the customer name **as captured at invoice time** — "Copperleaf Design Co." — because it is a financial document, and the source order still shows "Copperleaf Design Co.".

### Actual
The invoice **list and detail both show "Copperleaf RENAMED-QA"** (the live customer name), while the source order **SO-DEMO-2004** still correctly shows "Copperleaf Design Co.". The order and its own invoice display two different names for the same customer.

> 📎 **Screenshot to attach here:** `retest5-F1-invoice-detail-live-renamed-name.png`
> _(invoice detail showing "Copperleaf RENAMED-QA")_

### Same root cause — customer deletion
[...]

PR #5617, opened 2026-08-25
fix(devices): pick a device owner by name, not by UUID (#5591)

### LINKED ISSUE #5591: bug: make the register device form usable
Image

Frshy:
## Summary

The admin *Register device* form asked for the owner as a free-text field, hinting **"UUID of the user who owns this device."** Nobody knows a user's UUID, so the form was unusable without a database query first — that is the screenshot on #5591.

The device list had the mirror problem. `backend/devices/page.tsx` already had a user combobox *filter* and labelled the owner column from `userLabelById` — a map built only from the ~20 users that filter happened to prefetch. Every other row rendered a bare UUID. `backend/devices/[id]/page.tsx` had a third, single-user variant of the same lookup.

The pattern this form should use already existed **three times inside the same feature**: `push_notifications`' "Send push" form and both list filters use a `combobox` backed by `/api/auth/users?search=`. This PR reuses it rather than inventing anything, and gives the lists a way to resolve the owners actually on screen.

## Changes

- `devices/backend/devices/create/page.tsx` — the `userId` field becomes `type: 'combobox'` with the existing loader (`/api/auth/users?page=1&pageSize=20&search=<q>`, header `x-om-forbidden-redirect: '0'`, every failure degrading to no options).
- `devices/backend/devices/page.tsx` and `.../[id]/page.tsx` — resolve the owner ids on the current page in one batched request, so an owner who never appeared in a search result stops rendering as a UUID.
- New `devices/backend/devices/userOptions.ts` + `useDeviceUserLabels.ts` — one shared implementation replacing the three near-identical inline loaders.
- `auth/api/users/route.ts` — `GET /api/auth/users` accepts `?ids=<uuid>,<uuid>` (repeated or comma-joined, max 100), because the batch label lookup had nothing to call. Parsing factored into `auth/lib/userIdFilter.ts`, reusing `parseIdsParam` / `isIdsParamProvided` from `shared/lib/crud/ids.ts`.
- `devices/i18n/{en,pl,de,es,ko}.json` — the field is relabelled "User ID" → "User", the hint stops naming UUIDs, and a placeholder key is added. No key removed.

### Two decisions worth a reviewer's attention

**Custom values stay allowed on the combobox.** The obvious move is `allowCustomValues: false` (the `warranty_claims` FK-picker idiom). It is wrong here: `devices.admin` does **not** imply `auth.users.list`, so for an admin without it the loader returns nothing — and a locked combobox would leave them unable to register *any* device, a hard regression on the very form this fixes. Pasting an id still works, and the server already rejects an unknown target with `devices.errors.user_not_found`.

**`?ids=` supplied but unusable matches nothing.** Dropping the filter and returning the first page would answer a malformed request with a slice of the directory it never named — the record-count side channel `mergeIdFilter` guards against for CRUD list routes (#4143 Finding 3). `?ids=` also intersects with `?id=` rather than widening to the union, and the route auto-sizes `pageSize` to the batch so a 100-id lookup is not truncated at the default 50.

The `auth` change is additive and **narrowing-only**, behind the unchanged `auth.users.list` feature — it exposes no directory access the picker did not already have. It also fixes `customers/components/detail/ActivitiesSection.tsx`, which already sent `?ids=` and silently received an unfiltered page.

### Out of scope, noted

- `packages/ui/src/ai/ConversationShareDialog.tsx:80` sends `?limit=200`, a *different* unsupported param (the fix is `pageSize=100`). Unrelated to `ids`; not pulled into a devices fix.
- The form's `pushProvider` is still free text, and `push-fanout.ts` matches it **exactly** against an active tenant push channel's `providerKey` — a typo means the device is silently skipped, no error anywhere. Worth its own issue; beyond what #5591 asks for.

## Specification

- [x] Yes
- [ ] No (created a new spec)

**Spec file path:**
`.ai/specs/2026-04-28-push-notifications-and-devices.md` — changelog entry added under `### 2026-08-25`.

## Testing

Local runner (no compose `app` container up).

| Gate | Result |
|---|---|
| `yarn generate`, `yarn build:packages` | pass |
| `yarn lint` | pass, 0 errors |
| `yarn build:app` | pass |
| `yarn workspace @open-mercato/core test` | 1437 suites passed, 1 skipped |
| `yarn i18n:check-sync` | all five locales in sync |
| `yarn test:integration:ephemeral --filter devices` | **30 passed** (7.2m) — includes the new `TC-DEV-009` |
| `yarn test:integration:ephemeral --filter auth` | **129 passed** (8.5m) — no regression from `?ids=` |

New coverage:
- `auth/__tests__/user-id-filter.test.ts` — comma/repeated parsing, dedupe, cap, `id`×`ids` intersection, and the match-nothing rule.
- `devices/__tests__/user-options.test.ts` — both label styles, `?search=` trimming, the forbidden-redirect header, 403/throw degradation, id batching, and the failed-vs-empty distinction below.
- `devices/__integration__/TC-DEV-009.spec.ts` — batch resolve, malformed `ids`, disjoint `id`+`ids`, register-on-behalf then resolve the owner.

Internal review on our fork ([fullstackhouse#114](https://github.com/fullstackhouse/open-mercato/pull/114)) caught one real defect, fixed in `6a03f4c27`: the batch lookup swallowed a 403 or network error and returned `[]`, which the label hook could not tell apart from "the server answered, nobody matched" — so one transient failure left those rows showing a bare UUID for the life of the component. The lookup now reports which ids the server actually answered for; an id that came back without a row still counts as resolved, an id whose request failed does not.

**Not verified:** manual QA click-through and screenshots — in particular the degradation path, a role with `devices.admin` but **without** `auth.users.list`, which must still render both pages and still register a device by pasting an id. Hence `needs-qa` rather than a self-QA claim.

## Checklist

- [x] I added or updated integration tests — `TC-DEV-009` lives in `packages/core/src/modules/devices/__integration__/`, the location `.ai/qa/AGENTS.md` prescribes (executable specs must not sit under `.ai/qa/tests`, which holds the shared Playwright config only).
- [ ] Priority set — this account has no label permission upstream; requested in a comment below.
- [ ] Risk set — as above.
- [ ] QA routing set — as above; `needs-qa` requested.

### Design System Compliance
- [x] No hardcoded status colors — verified: the diff introduces none.
- [x] No arbitrary text sizes — verified: the diff introduces none.
- [ ] Empty state handled for list/data pages — unchanged by this PR; the list's `DataTable` wiring is untouched, so this is neither newly satisfied nor newly broken. The picker's own empty case is covered: a failed or forbidden lookup yields no suggestions and the field stays usable.
- [x] Loading state handled for async pages — unchanged; the owner column renders the raw id until its label resolves, then swaps.
- [x] `aria-label` on icon-only buttons — the diff adds no icon-only buttons.
- [x] Uses existing DS components — reuses `CrudForm`'s `combobox` field type; no custom replacements.

## Linked issues

Fixes #5591 (first of its two asks — the docs half ships separately).

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5617"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790252086&installation_model_id=432243&pr_number=5617&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5617&signature=a014a750e988a34040548429ea83fe8964fba104754bfad7255ea8bbc177c5b5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Frshy:
**Metadata I cannot apply — this account has no label permission here (`push: false, triage: false`).** Requesting a maintainer set:

| Label | Rationale |
|---|---|
| `bug` | #5591 is filed as a bug; the form is unusable as shipped. |
| `review` | Ready, not draft, and in no other pipeline state. |
| `priority-medium` | Unusable form, but a workaround exists (look the UUID up first) and the feature is new in this release window. |
| `risk-medium` | The inference table routes "API routes" to `risk-high`, and this does touch `/api/auth/users`. I read it as medium because the change is additive, **narrowing-only**, sits behind the unchanged `auth.users.list` feature, and ships unit + integration coverage. Bumping to `risk-high` on "it's auth" would be defensible — flagging the reasoning rather than deciding it silently. |
| `needs-qa` | Mandatory twice over: `.tsx` files change **and** the API surface changes, so the automated-verification exemption cannot apply. |

**Reviewer:** whoever owns `devices` / `auth` — I am deliberately not tagging a handle.

**Assignee:** not claiming. A claim label I cannot release later is worse than no claim.

**On the QA gate:** this stays blocked until a maintainer adds `qa-approved`. I have not self-QA'd — the click-through and screenshots are genuinely outstanding, in particular the degradation path: a role with `devices.admin` but **without** `auth.users.list` must still render both device screens and still register a device by pasting an id. That path is the reason the combobox deliberately keeps `allowCustomValues` on, so it is the one worth exercising by hand.

Automated evidence is in the PR body; the integration suite was run on an ephemeral Docker environment (devices 30/30 including the new `TC-DEV-009`, auth 129/129), not just unit tests.

pkarw:
🤖 `om-auto-review-pr` claimed this PR for review at 2026-08-25T23:28Z. Lock: `in-progress` label + assignee. It will be released when the review is posted.

pkarw (approved):
# 🔍 Code Review

**Verdict: ✅ Approve** — no blockers and no majors. Everything below is a nit or a low-severity cleanup the author can take or leave.

**Validation source: GitHub checks.** `prepare`, `lint`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance` are green on `6a03f4c2`. `test` and `ephemeral-integration` are **still pending** at the time of this review — see the CI disclosure at the bottom. No local gate was re-run; CI owns that signal.

## 📋 Summary

This is a tightly scoped fix and it reads like one. The `userId` field on *Register device* becomes the same `/api/auth/users?search=` combobox that `push_notifications`' "Send push" form and both list filters already use, three near-identical inline loaders collapse into `userOptions.ts` + `useDeviceUserLabels.ts`, and `GET /api/auth/users` gains the `?ids=` batch lookup the owner column needed in order to stop rendering bare UUIDs.

The two decisions the PR body flags for a reviewer's attention are both correct, and I checked them rather than taking them on trust:

- **`allowCustomValues: true` is right, and it is safe.** `devices.admin` genuinely does not imply `auth.users.list`, so a locked combobox would leave such an admin unable to register any device. The paste-an-id escape hatch is not a hole: `devices/api/admin/devices/route.ts:105-112` resolves the target with `{ id, tenantId: auth.tenantId, deletedAt: null }` and then checks organization scope, so a pasted cross-tenant or out-of-scope UUID is rejected with `devices.errors.user_not_found`. Worth noting the field was free text before, so this changes nothing about the server-side trust boundary.
- **The `?ids=`-supplied-but-unusable → match-nothing rule is right.** `resolveUserIdFilter` reproduces the `mergeIdFilter` contract (#4143 Finding 3) rather than reinventing it, and `isIdsParamProvided` is exactly the "supplied vs. malformed" discriminator that rule needs.

## 🔒 Security & tenancy

No issues. The new `ids` filter is composed into the same `filters` array that already carries `deletedAt`, the tenant scope, and the super-admin `$nin` exclusion, and it lands as `{ id: { $in: [...] } }` inside the same `$and` (`route.ts:403-406`). It is narrowing-only — it cannot widen the result set past what `auth.users.list` already exposed to that caller in that scope. The route metadata (`route.ts:132`) is unchanged.

The batch cap (`MAX_USER_LOOKUP_IDS = 100`) is enforced before the query and the auto-sized `pageSize` is clamped to the schema's `max(100)`, so `?ids=` cannot be used to pull a larger page than `?pageSize=` already allowed.

## 🔁 Backward compatibility

Additive, and it stays inside category 7 ("MAY add new optional fields to request/response schemas"). No route, method, or response field changed; `openApi` documents the new param. The `?id=`-only path is byte-equivalent to the old behavior.

The one real behavior change is for `customers/components/detail/ActivitiesSection.tsx:321`, which has been sending `?ids=` into a route that ignored it. That call site now gets the users it actually asked for instead of an unfiltered first page — strictly better, and correctly called out in the PR body. See finding 5 for the part of that it does *not* fix.

## 🧪 Tests

Good coverage for the shape of this change: `user-id-filter.test.ts` pins every branch of the parser including the two that matter (`ids` × `id` intersecting rather than widening, and malformed-`ids` matching nothing), `user-options.test.ts` covers both label styles, the forbidden-redirect header, the 403/throw degradation and the failed-vs-empty distinction, and `TC-DEV-009` exercises the real endpoint end to end with self-created fixtures and a `finally` teardown — the `.ai/qa/AGENTS.md` shape.

Two gaps, both minor, in findings 4 and 6.

## 🎨 Design system & i18n

Clean. No status colors, no arbitrary values, no new icon-only buttons; the change reuses `CrudForm`'s `combobox` field type rather than replacing it. All five locales carry real translations for the relabelled `devices.form.userId`/`userIdHint` and the new `userIdPlaceholder`, and the new key sorts correctly after `userIdHint`. No key was removed.

---

## 🟢 Findings (nit / low)

**1. `DeviceUserOption.description` is computed everywhere and rendered nowhere.** `packages/core/src/modules/devices/backend/devices/userOptions.ts:32`

`CrudForm` maps combobox options through `options.map((opt) => ({ value: opt.value, label: opt.label }))` (`packages/ui/src/backend/CrudForm.tsx:4583-4586`), and `FilterBar.tsx` never reads `description` at all. So the field is dead on both consumers. That is carried forward from the code being replaced rather than introduced here — but `user-options.test.ts:35` and `:50` now assert it, which reads like coverage of something a user sees. Either drop `description` from the type, or keep it and note in the test that it is currently unrendered so the next reader does not go looking for it.

Relatedly: the devices list *filter* previously labelled options `name ?? email` and now gets the `'search'` style `name — email`, since both call sites share one loader. That is a real (small) visual change to a surface this PR is not about. It matches what "Send push" already shows and it disambiguates same-named people, so I read it as an improvement — flagging it only so it is not a surprise in QA.

**2. The constant comment in `userOptions.ts` is wrong, and the constant is duplicated.** `packages/core/src/modules/devices/backend/devices/userOptions.ts:15-17`

> `// /api/auth/users caps pageSize at 100 and ?ids= at the shared MAX_IDS_PER_REQUEST.`

The route caps `?ids=` at `MAX_USER_LOOKUP_IDS = 100`, not at the shared `MAX_IDS_PER_REQUEST`, which is `200` (`packages/shared/src/lib/crud/ids.ts:3`). Anyone who raises the client's `MAX_IDS_PER_LOOKUP` to 200 on the strength of that comment silently loses 100 ids per batch — `parseIdsParam` slices without complaining. Fix the comment, and consider importing `MAX_USER_LOOKUP_IDS` from `@open-mercato/core/modules/auth/lib/userIdFilter` instead of restating `100`; that module pulls in nothing but the shared ids helper, so it is client-safe.

**3. Dead branch left behind by the refactor.** `packages/core/src/modules/auth/api/users/route.ts:403-406`

```ts
if (idFilter && idFilter.size) {
  filters.push({ id: { $in: Array.from(idFilter) as any } })
} else if (id) {
  filters.push({ id })
}
```

The `else if (id)` arm is now unreachable. `idFilter` is null only when `userIdFilter.kind === 'unfiltered'`, which by construction means no `?id=` was supplied; every other path either produces a non-empty set or early-returns (`kind === 'none'` at `:239`, the empty role intersection at `:331`). Harmless, but it is the kind of leftover that makes the next reader think `?id=` still has a separate path.

**4. The chunking test's fixture ids are not UUIDs.** `packages/core/src/modules/devices/__tests__/user-options.test.ts:150-160`

`` `1111111${index.toString().padStart(1, '0')}` `` produces `11111110`, `11111111`, … — the `padStart(1, …)` is a no-op and none of these would survive `parseIdsParam` server-side. The test passes because `resolveDeviceUserOptions` does no client-side validation, which is fine, but it means the test would keep passing if that ever changed. The neighbouring test at `:127` already has a correct UUID generator; reuse it.

**5. `ActivitiesSection` is fixed only up to 100 ids, and only on success.** `packages/core/src/modules/customers/components/detail/ActivitiesSection.tsx:315-322`

That call site sends every unresolved author id in one request with no chunking, and — line 317 — adds them all to `resolvedUserIdsRef` *before* awaiting the fetch. So with more than 100 unresolved authors, ids 101+ are silently dropped by the server cap and never retried; the same is true for the whole set if the request throws. This is the exact defect `6a03f4c2` fixed in the devices hook, still present in the sibling call site. It is not a regression — the old unfiltered-first-page behavior was worse — but the PR body describes this as "fixes", and it fixes it only partially. A follow-up issue pointing that call site at `resolveDeviceUserOptions`-style chunking and post-hoc caching would be the clean close-out. No objection to it landing as-is.

**6. `useDeviceUserLabels` itself is untested.** `packages/core/src/modules/devices/backend/devices/useDeviceUserLabels.ts:21-42`

The bug the fork review caught lives half in the helper (`resolvedIds` in the return value, covered) and half in the hook (the `resolvedIdsRef` bookkeeping that decides whether a failure is retried, uncovered). A `renderHook` test asserting "a failed lookup is retried on the next id-set change, a successful-but-empty one is not" would pin the behavior the commit message claims. Cheap, and it is the assertion that would actually catch a regression here.

---

## 🚦 CI status

Not waiting on CI, per the review contract. At the time of this review `test` and `ephemeral-integration (none)` are **pending** on `6a03f4c2`; everything else that ran is green. This approval is on the code, not on CI — required checks and branch protection still gate the merge, and a red `test` or `ephemeral-integration` overrides this verdict. The PR's own testing table reports both suites green locally (`devices` 30 passed, `auth` 129 passed), which is consistent with what I read in the diff.

`needs-qa` is retained: the diff touches four `.tsx` surfaces, so the automated-verification exemption does not apply, and the author explicitly declined a self-QA claim. The degradation path the author names — a role with `devices.admin` but **without** `auth.users.list` — is the one QA must actually exercise; separate instructions follow in their own comment.

pkarw:
## 🏷️ label rationale

Applied by `om-auto-review-pr` after an approving review. The metadata @Frshy requested is now set — thanks for spelling out the reasoning, it made this a one-line decision on three of the four.

| Label | Why |
|---|---|
| `merge-queue` | Approved, no blockers and no majors. The QA-approval gate below still holds the actual merge. |
| `bug` | #5591 is filed as a bug and the form was unusable as shipped. Agreed. |
| `priority-medium` | Agreed with the author's reading: the form is unusable, but a workaround exists (look the UUID up first) and the feature is new in this release window. |
| `risk-high` | **Raised from the requested `risk-medium`.** The inference table routes changes to shared contract surfaces — API routes explicitly, and this is an *auth* one — to `risk-high`, and instructs picking the higher label when signals conflict. The author's counter-argument is genuinely good (additive, narrowing-only, behind the unchanged `auth.users.list` feature, unit + integration coverage) and I do not think this change is dangerous. But `?ids=` was previously *accepted and ignored*, and one in-tree caller (`customers/components/detail/ActivitiesSection.tsx:321`) has been relying on that; its result set changes with this PR. A previously-inert parameter on the user-directory endpoint becoming load-bearing is the blast radius the label is for. In practice this changes nothing about the merge path — `needs-qa` was already set — it just tells the next reader where to look first if something regresses. |
| `needs-qa` | Four `.tsx` surfaces changed, so the automated-verification exemption does not apply, and the author explicitly declined a self-QA claim. Instructions posted separately. |
| `partner-request` | Preserved as set by the author. |

`in-progress` was this review's claim lock and is released with the completion comment.

pkarw:

The author called out the one path they could not verify — a role with `devices.admin` but **without** `auth.users.list`. That is P0 here, because it is the path where a wrong call makes the form *less* usable than before this PR rather than more.

### P0 — Degradation for a `devices.admin` who lacks `auth.users.list`

**Where to click**
- Settings → roles: create a role granted `devices.admin` (or `devices.*`) and **not** `auth.users.list`, and assign it to a test user. Log in as that user.
- `/backend/devices/create`
- `/backend/devices`
- `/backend/devices/<id>` for any existing device

**What to verify**
- The *Register device* form renders and the **User** field is present, empty, and typeable. Typing produces no suggestions — that is correct here, not a bug.
- Pasting a valid user UUID from that admin's own tenant into the User field and submitting **registers the device successfully**. This is the whole reason custom values were left enabled; if it fails, the change is a hard regression.
- Neither `/backend/devices` nor the device detail page redirects to `/login` or shows an error banner. The owner column falls back to the raw UUID (rendered as a link) — that is the expected degraded state.
- Submitting a UUID that does not exist, or one belonging to another tenant, is rejected with "Target user not found in this tenant" rather than silently creating a device.

**What can go wrong**
- The whole page bounces to `/login` — means the `x-om-forbidden-redirect: '0'` header stopped being sent on one of the three call sites.
- The User field renders read-only or refuses a pasted value — means `allowCustomValues` regressed, and this admin can no longer register any device at all.
- The list or detail page renders an empty table / error state instead of degrading to UUIDs.

### P1 — The owner picker and the owner column, as a full `auth.users.list` admin

**Where to click**
- `/backend/devices/create`
- `/backend/devices` (including page 2+ if you have more than 50 devices)
- `/backend/devices/<id>`

**What to verify**
- Typing a partial name or email into **User** returns matching suggestions labelled `Name — email`, and picking one then submitting registers the device against that user. Confirm on the list that the new row shows that person's name.
- The field label reads **User** (not "User ID") and the hint no longer mentions UUIDs. Spot-check one non-English locale (`pl` or `de`) for the same three strings.
- **The core fix:** every row's owner column shows a name or email, *not* a bare UUID — specifically for a device whose owner would never appear in the first 20 search results. Seed this deliberately: register a device for a user who sorts late in the directory, then reload `/backend/devices` without touching the filter.
- Page through the list. Owners resolve on each page, not just the first.
- The device detail page shows the owner's name as a link to their profile, and the link opens the right user.
- The list's **User** filter still filters correctly after picking someone from its dropdown.

**What can go wrong**
- Owner names appear on page 1 and revert to UUIDs on page 2 — batch resolution is not re-running per page.
- A name flashes in and then disappears, or the column flickers between UUID and name on every re-render.
- The User *filter* dropdown labels changed in this PR from `Name` to `Name — email` (both call sites now share one loader). That is intended; flag it only if it visibly breaks the filter chip or truncates badly.
- Registering from the picker sends the display label instead of the user id — the created device would fail or attach to the wrong owner.

### P1 — Perceived performance

**Where to click**
- Cold-load `/backend/devices` (hard refresh, screenshot if you can) and `/backend/devices/create`

**What to verify**
- The device list paints its rows immediately; owner labels fill in shortly after without shifting the layout or blanking the column.
- The User combobox responds per keystroke without visible lag or a stalled spinner, and clearing the query restores the default suggestions.
- Mobile viewport: the User field and the owner column stay readable; the `Name — email` label wraps or truncates rather than overflowing.

**What can go wrong**
- A visible layout shift when labels resolve.
- One request per row instead of one per page (check the network tab: the owner lookup should be a **single** `/api/auth/users?ids=…` call per page).

### P2 — Adjacent call site

**Where to click**
- Any customer detail page with an Activities section whose entries have authors — `/backend/customers/people/<id>`

**What to verify**
- Activity author names still resolve. This PR makes `?ids=` actually filter, and that section has been sending it into a route that ignored it, so its behavior changes here even though no line of it was edited.

**What can go wrong**
- Author names that used to appear now show blank or fall back to an id.

### Pass/fail

pkarw:

Swapped `in-progress` for `ci-monitoring` — the review work is done and this PR is free for anyone to act on. `test` and `ephemeral-integration` were still pending at review time; I'll post the CI result and correct the label if the outcome contradicts the verdict.

autofix: skipped (not my PR — re-run with `--autofix` to fix it here). All six findings are nits/low, so there is nothing here that needs fixing before merge; findings 5 and 6 are the two worth a follow-up issue.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-25T23:36Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

pkarw:

CI settled green on `6a03f4c2`. The approval stands unchanged.

| Check | Result |
|---|---|
| `prepare` | pass (3m56s) |
| `test` | pass (9m59s) |
| `ephemeral-integration` | pass (14m12s) |
| `lint` | pass (35s) |
| `ds-lint` | pass (45s) |
| `scope` | pass (34s) |
| `audit-scope` | pass (12s) |
| `documents-multi-instance` | pass (1m4s) |

The two checks that were pending when I posted the review have both landed. I confirmed the new integration spec actually executed rather than trusting the job's green tick — `TC-DEV-009.spec.ts:36 › resolves a batch of owner ids, and matches nothing when ids is unusable` passed in 409ms as test 165 of the run (161 passed, 0 failed, 0 flaky). That matters here because the batch `?ids=` contract and the match-nothing rule are the parts of this PR a unit test alone cannot prove.

Removing `ci-monitoring`. This PR is approved with green CI and sits in `merge-queue`; the QA-approval gate holds the merge until a QA reviewer adds `qa-approved` — instructions are in the comment above.

pkarw:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ PASS — every required step passed, including the permission edge this PR's own description listed as *not verified*.
**Environment:** `http://127.0.0.1:5001` (disposable ephemeral instance, production build served by `next start`, freshly migrated and seeded testcontainers Postgres) · roles `admin@acme.com` **and** a purpose-built role holding `devices.admin` **without** `auth.users.list` · browser `playwright` (Chromium, headless)
**Verified:** `fix/devices-register-form-user-picker` @ `6a03f4c2`

### Scenario (P0/P1 — devices admin: register form, list owner column, detail owner)

**Where to click:** `/backend/devices/create`, `/backend/devices`, `/backend/devices/<id>`

| 1 | Open the register form as an admin and read the owner field | Labelled `User` (not `User ID`), hint no longer names a UUID, rendered as a combobox | Label **`User *`**, hint **"Search by name or email, or paste a user ID."**, `role=combobox` input placeholdered **"Search users"**. The old copy *"UUID of the user who owns this device."* is gone from the page | ✅ |
| 2 | Type a person's display name (`QA5617 Nina`) | A suggestion list appears, labelled so two people sharing a name stay apart | One option: **`QA5617 Nina Owner … — qa5617-nina-…@example.test`** — name plus the disambiguating email | ✅ |
| 3 | Pick the suggestion, fill Device ID, submit | Device registered against the picked user, back to the list with a success flash | Redirected to `/backend/devices` with the **"Device registered"** flash; the new row is present | ✅ |
| 4 | Register a second device by **pasting a raw user UUID**, then read the list's User column | Every owner on the page resolves; no cell shows a bare UUID | Owner cells read `["QA5617 Omar Pasted …", "QA5617 Nina Owner …"]`. **Zero** cells matched a bare-UUID pattern — including the pasted-id owner, who never appeared in any picker search. That is the batched `?ids=` lookup doing its job | ✅ |
| 5 | Open a device's detail page | Owner renders as a display name | Header `User` reads **`QA5617 Nina Owner …`** | ✅ |
| 6 | **P0** — sign in with `devices.admin` but **without** `auth.users.list`, open the list | Page renders instead of bouncing to `/login`; owner column degrades to the raw id | URL stayed at `/backend/devices`, table rendered, owner cells fell back to bare UUIDs — the documented degradation, no redirect, no error state | ✅ |
| 7 | **P0** — same role, type a name into the picker | No suggestions, but the field stays enabled and usable | Zero options and the status message **"No matches found"**; the input kept the typed query and stayed enabled | ✅ |
| 8 | **P0** — same role, paste a user id and submit | `allowCustomValues` keeps the pasted id and the device registers | Pasted UUID stuck, form submitted, list returned with the **"Device registered"** flash and three rows | ✅ |
| 9 | Repeat the picker and the list at a 390×844 mobile viewport | Both changed surfaces stay usable on a phone | Picker opened and returned the match; list rendered all rows with resolved owner labels | ✅ |
| 10 | Cold-load `/backend/devices/create` until the picker is interactive | A useful shell appears fast and the field is usable before the page finishes | **277–356 ms** across three runs against the production build; the form is interactive while the sidebar still shows its skeleton | ✅ |

Steps 6–8 are the ones worth a reviewer's attention: the PR body explicitly says the
`devices.admin`-without-`auth.users.list` path was **not** manually verified, and it is the case where
the `allowCustomValues: true` decision earns its keep. It was reproduced here with a purpose-built role
(`devices.admin` only, granted through `PUT /api/auth/roles/acl`), and all three of the behaviours the PR
argues for hold: the pages render, the picker degrades to "No matches found" instead of locking the user
out, and a device can still be registered by pasting an id.

### Screenshots

**step-01-register-form-user-picker**



**step-02-picker-suggestions-by-name**



**step-03-registered-via-picker**



**step-04-list-owner-column-resolved**



**step-05-detail-owner-name**



**step-06-restricted-list**



**step-07-restricted-picker-degraded**



**step-08-restricted-registered-by-paste**



**step-09-mobile-picker**



**step-10-mobile-list**




### Notes for QA

- **No browser-level test covers this yet.** `TC-DEV-009` shipped in this PR is API-level only — it uses
  Playwright's `request` fixture and never opens a `page` — so the picker and the owner column have no
  automated regression net. A ready-to-implement scenario follows in a separate comment.
- **Cosmetic, mobile:** at 390 px the suggestion popover clips the one-line `Name — email` label at its
  right edge without an ellipsis. Selection still works; worth a follow-up only if the combobox is meant
  to wrap or truncate gracefully.
- **Not exercised:** batching beyond the 100-id cap in `resolveDeviceUserOptions` (that needs a page
  holding more than 100 distinct owners, which the seeded data cannot produce), and locales other than
  English — the four other locale files changed in this PR were not rendered.
- Every fixture created for this run (three users, one role, three devices) was deleted in teardown; the
  environment was disposable and is torn down with the run.

pkarw:

This change ships no **browser-level** test. `TC-DEV-009` is genuinely useful, but it drives
Playwright's `request` fixture only — it never opens a `page`, so it pins the `?ids=` API contract
and leaves the two surfaces this PR actually changes (the `combobox` field and the resolved owner
column) with no automated regression net. The UI QA above was manual; lock it in with an automated
test (run `/om-integration-tests`).

1. **Setup** — via the existing API fixtures in
   `packages/core/src/helpers/integration/authFixtures.ts`:
   - `createUserFixture` × 2 with distinct display names (`nina`, `omar`) in the token's org.
   - `createRoleFixture` + `setRoleAclFeatures(roleId, ['devices.admin'])`, then a third user
     carrying only that role — the `devices.admin`-without-`auth.users.list` actor.
2. **Act / Assert — the picker (admin):**
   - `/backend/devices/create`: the field is labelled `User`, its description is
     `devices.form.userIdHint`, and the input carries `role="combobox"` with the
     `devices.form.userIdPlaceholder` placeholder.
   - Type `nina`'s display name; assert one `role="option"` whose text contains **both** the name
     and the email (the `search` label style in `userOptions.ts` — the email is what disambiguates
     two people sharing a name, and `CrudForm`'s combobox drops the option `description`).
   - Select it, fill `deviceId`, submit; assert the redirect to `/backend/devices` and the
     `devices.form.success.created` flash.
3. **Act / Assert — the owner column (admin):**
   - Register a second device by **pasting `omar`'s raw id** into the same field, so that owner has
     never been returned by a picker search.
   - On `/backend/devices`, assert every User cell resolves to a label and **none** matches
     `/^[0-9a-f]{8}-[0-9a-f]{4}-.../i`. This is the assertion that would fail if the batched
     `resolveDeviceUserOptions` lookup regressed.
   - Open the device detail page and assert the header `User` shows the display name.
4. **Act / Assert — the degradation path (restricted role):** in a second browser context signed in
   as the `devices.admin`-only user:
   - `/backend/devices` renders (URL does **not** become `/login`) and its User cells fall back to
     bare ids;
   - the picker returns zero options with the "no matches" status, and the input stays **enabled**;
   - pasting an id and submitting still registers a device. This is the regression
     `allowCustomValues: true` exists to prevent, so it is the highest-value assertion in the set.
5. **Teardown** — delete every device, then the three users and the role, in a `finally`.

**One driver gotcha worth baking into the spec:** `CrudForm` is a controlled client component, so
`locator.fill()` before hydration sets the DOM value and React then wipes it — the picker looks like
it returned nothing when in fact nothing was ever typed. That produced a false negative twice during
this QA run. Wait for hydration, type with `pressSequentially`, and assert `inputValue()` before
trusting the suggestion list.

pkarw:

Running under the default `--evidence-only` mode, so **no pipeline or meta label was changed** — the QA verdict stays a maintainer's call. Worth noting for whoever makes it: this run was fully green **with** screenshots, and the PR carries `needs-qa` without `skip-qa`, so it does meet the bar for `qa-approved` (plus `qa-self-verified` had the self-QA exception been requested). Until `qa-approved` is applied, the `needs-qa` merge gate still blocks this PR despite the `merge-queue` label.

pkarw:
🤖 `om-auto-review-pr --autofix` claimed this PR at 2026-08-26T01:12Z. Lock: `in-progress` label + assignee. Maintainer-directed autofix pass — reversing the `allowCustomValues` decision behind an ACL dependency, plus the review nits. Will be released when the pass reports.

pkarw:
## Superseded by #5634

@Frshy — the maintainer-directed follow-up (close the owner picker, backed by a new `devices.admin` → `auth.users.list` ACL dependency, plus the review nits) could not be pushed to this branch: `fullstackhouse/open-mercato` is an **organization-owned** fork, and GitHub does not honour maintainer pushes to those even though this PR reports `maintainerCanModify: true`. The push was rejected with a 403.

So the work is carried forward in **#5634**, with your three commits intact and authorship preserved. That PR is assigned to you for the next recheck.

### What changed

| | Before | Now |
|---|---|---|
| Owner picker | `allowCustomValues: true` — a pasted UUID is accepted | `allowCustomValues: false` — a free-typed or pasted value is reverted on blur |
| `devices.admin` | no declared dependency | `dependsOn: ['auth.users.list']` (`devices/acl.ts`) |
| `devices/setup.ts` | `admin: ['devices.*']` | `admin: ['devices.*', 'auth.users.list']` (same for `superadmin`) |
| Hint copy | "…or paste a user ID." | "Start typing a name or email, then pick the owner from the list." (all five locales) |

Search-as-you-type is unchanged — that part of your implementation was already right and is what the closed picker now depends on.

### The one thing this does not do

`dependsOn` is **declarative, not enforcing**. It drives the ACL editor's missing-dependency panel and its one-click fix (`resolveAclDependencyDiagnostics` → `applyAddMissingDependency`); it does not grant anything at runtime. `defaultRoleFeatures` reaches existing tenants only through `yarn mercato auth sync-role-acls`.

So a **hand-built role** holding `devices.admin` without `auth.users.list` still exists after this merges, and on that role the register form now offers no owner and cannot be submitted. The list and detail pages still render, degrading the owner column to raw ids as before. **Run `yarn mercato auth sync-role-acls` after deploying**, and check the ACL editor for any custom role that carries `devices.admin`. This is called out in the spec changelog, `devices/AGENTS.md`, and the re-review.

### Review findings applied

- **1** — `DeviceUserOption.description` dropped. `CrudForm` maps combobox options to `{value, label}` only (`CrudForm.tsx:4583-4586`) and `FilterBar` never reads it, so it was dead on both consumers. (`ComboboxInput` itself *does* render a description — it just never receives one through these two paths.)
- **2** — `MAX_IDS_PER_LOOKUP` now imports `MAX_USER_LOOKUP_IDS` from `auth/lib/userIdFilter` instead of restating `100` under a comment that named the wrong constant.
- **3** — the unreachable `else if (id)` arm removed from `auth/api/users/route.ts`.
- **4** — the chunking fixture uses real UUIDs via a shared `uuidsFor()` helper.
- **6** — `useDeviceUserLabels` gets its own suite (`devices/__tests__/use-device-user-labels.test.tsx`, 6 cases) pinning the retry bookkeeping your `6a03f4c27` fixed: an answered-but-empty id is cached, a failed one is retried, and a re-sorted id list does not re-fetch.
- **5** — filed as #5632 rather than pulled into a devices fix, as the review suggested.
- New `devices/__tests__/acl-dependencies.test.ts` covers the dependency declaration and asserts every default role that gets `devices.admin` also gets `auth.users.list`.
- The browser-level test gap is filed as #5633, rewritten for the closed picker (it must now assert an out-of-list value is *rejected*, the opposite of what the earlier QA run recorded).

---

Closing in favour of #5634 (https://github.com/open-mercato/open-mercato/pull/5634). Credit to @Frshy for the original implementation. Reopen if you'd rather land it from this branch instead — the carry branch is `carry/pr-5617-ready` and rebases cleanly.

Moving the pipeline labels to #5634; the earlier approval and QA evidence on this PR are **stale** for the picker (they verified the paste-an-id path this change removes), and both are being re-run there.

PR #5542, opened 2026-08-24
fix(notifications): make built-in channel eligibility explicit

haxiorz:
Tracking plan: .ai/runs/2026-08-24-issue-5495-explicit-notification-channels.md
Status: complete

Closes #5495

## Goal

Keep ordinary built-in notifications on their intended delivery channels so enabling a new provider does not silently make every notification eligible for it.

## Problem and root cause

Built-in notification definitions without an explicit `channels` list inherit the framework fallback and become eligible for every registered channel. That behavior is useful for third-party notification types, but it caused existing business notifications to become push-eligible as soon as push delivery was enabled.

## What changed

- Declared `in_app` and `email` eligibility for the affected built-in business and actionable notification types.
- Kept the two push demo fixtures explicitly push-only and preserved the unrestricted hidden admin message types.
- Added an AST-based regression test that enforces explicit channel eligibility across the built-in notification catalogues.
- Registered the cross-package policy test in `REPO_WIDE_GUARDS` so CI runs it independently of Turbo package filtering.
- Updated the push-notifications specification with the built-in eligibility policy and compatibility boundary.

## Validation

- `yarn build:packages` — all 26 package build tasks passed before and after generation.
- `yarn generate` — all generators completed without source drift.
- `yarn i18n:check-sync` — locale synchronization passed.
- `yarn i18n:check-usage` — passed with the existing advisory unused-key inventory only.
- `yarn typecheck` — all 26 package typecheck tasks passed.
- `yarn test` — all 33 package test tasks passed.
- `yarn build:app` — the production application build passed.
- `yarn workspace @open-mercato/core jest src/__tests__/notification-channel-eligibility.test.ts --runInBand` — 9/9 policy cases passed on the final rebased branch.
- `yarn workspace @open-mercato/webhooks jest src/modules/webhooks/__tests__/notifications.test.ts --runInBand` — the upstream webhook catalogue regression passed on the final rebased branch.
- `yarn template:sync` — application/template parity passed.
- `node --test scripts/__tests__/repo-wide-guards.test.mjs` — all 16 guard-manifest contract tests passed.
- `yarn test:repo-wide-guards` — all 35 repo-wide guard files passed, including the notification policy regression.

## Automated review

- Code review — approved after one remediation round that expanded coverage to every visible built-in catalogue entry.
- Post-rebase review — confirmed the upstream webhook label fix and the explicit channel policy coexist without conflict.
- Unresolved findings — None.

## Breaking changes

[...]

Issue #5495, opened 2026-08-22
bug: [Notifications] 20 of 48 notification types default to push ON because they omit channels


Kapsik89:
## Summary
20 of the 48 registered notification types omit `channels`, and a type that omits it resolves to *every*
registered channel. Once a tenant enables a push channel, all 20 deliver push by default.

This is the open question PR #5366 raised and shipped without an answer:

> A notification type that omits `channels` resolves to every registered channel … should push stay
> opt-out-per-type or become opt-in?

## Evidence (2026-08-22, 0.7.0 build)
- `GET /api/notifications/types?pageSize=100` → **48** types, **20** with `channels: null`
- Affected modules: `ai_assistant`, `checkout`, `demo`, `eudr`, `example`, `warranty_claims`, `webhooks`, `wms`
- **5 of the 20 are `warranty_claims.*`** — shipped by #4092, the other feature PR in the same milestone,
  which does not itself opt into push
- `/backend/profile/notification-preferences` agrees 20/20: the Push switch is `checked && !disabled` for
  exactly those types and disabled for the other 28

## Why it is worth a decision
Push is the most intrusive channel and the default here is opt-out per type, decided implicitly by whether an
author remembered to write `channels`. A tenant enabling push for one purpose silently enables it for twenty.

Filed as the open design question, not as a malfunction.

haxiorz:
🤖 `autofix` started by @haxiorz at 2026-08-23T00:00:00Z. Other auto-skills will skip this issue until the lock is released.

pkarw:
🤖 `om-close-fixed-issues` started by @pkarw at 2026-09-14T12:01:30Z. Other auto-skills will skip this issue until the lock is released.

pkarw:
🤖 `om-close-fixed-issues` — issue closed

✅ Closed because #5542 (https://github.com/open-mercato/open-mercato/pull/5542) explicitly fixes this issue and merged into `develop` at 2026-08-28T07:07:36Z, commit `b2e61c5`.
Credit: @haxiorz.

If this closure is wrong, reopen and add `do-not-close` to prevent another automatic closure.

PR #5618, opened 2026-08-25
docs: document the notification delivery surface (#5591)

### LINKED ISSUE #5591: bug: make the register device form usable

Frshy:
## Summary

Second of #5591's two asks:

> I also do not quite get the intended way for the devices and notifications should be used; please add official docs how they should be registered (i guess it should be done by user's code) and how to be used because right now this feature is merely usable

`framework/modules/notifications.mdx` is **1199 lines** and covers creating notifications thoroughly. Grep it for `notificationTypes`, `shouldDeliver`, `nonOptOut`, `hiddenFromSettings` or `preferences` and you get **two incidental hits**, both inside a best-practices bullet. The *registry* — the half the issue calls out as "should be done by user's code" — was undocumented.

Also missing: any `framework/modules/devices.mdx` or `push-notifications.mdx`; rows for either module in `core-modules.mdx`; any user-guide page for notifications at all. The only existing page was the 259-line getting-started tutorial.

Docs-only. The picker half of #5591 is #5617.

## Changes

Four new pages, one per audience that was unserved:

- **`framework/modules/notification-delivery.mdx`** — the hub. How a module declares a type in its `notifications.ts` convention file (with a copyable example), the full `NotificationTypeDefinition` field table, the **two distinct channel convention files** that get conflated (`notification-channels.ts` = metadata for the preferences UI; `notifications.delivery-strategies.ts` = actual behavior), the four conditions `shouldDeliver` evaluates *and the order they run in*, operator overrides, preferences, and a map of every registered type.
- **`framework/modules/devices.mdx`** — what a `user_devices` row is and why it is keyed the way it is, the registration call a client app makes on launch, the self-serve vs admin split, and the token secrecy chain.
- **`framework/modules/push-notifications.mdx`** — the send path as one diagram, the `last_error` codes as a lookup table, silent push as a property of the *type* rather than of a send, the admin surfaces, and the complete `OM_PUSH_*` table — which now includes `OM_PUSH_SEND_TIMEOUT_MS`, `OM_PUSH_STUCK_RECLAIM_BATCH_LIMIT`, `OM_PUSH_RECLAIM_TICK_SECONDS` and the receipt knobs the tutorial never listed.
- **`user-guide/notifications-and-push.mdx`** — the operator walkthrough, in the numbered-step shape of `user-guide/webhooks.mdx`, with a troubleshooting ladder keyed on the literal statuses and error strings.

Each reference page carries a callout for **the failure that produces no error anywhere**: the fan-out is a silent no-op when the tenant has no active push channel, or when a device's `pushProvider` does not exactly match a channel's `providerKey`. That is the "why did nothing happen" case, and it was documented nowhere.

Wiring and housekeeping:

- `sidebars.ts` — the three framework pages next to `framework/modules/notifications`, plus the user-guide page. While there: `framework/modules/communication-channels` was **built but referenced nowhere in the sidebar**; that one-line orphan fix rides along.
- `core-modules.mdx` — rows for `devices` and `push_notifications` under UI & Platform. (`communication_channels`, `eudr`, `wms` and `warranty_claims` are missing from that table too — out of scope here.)
- The tutorial is **re-positioned, not duplicated**: it keeps the happy path and gains a "where to go next" block, and hands the env-var table and the eligibility rules to the references.
- Cross-links from `notifications.mdx`, `communication-channels.mdx` and `user-guide/integrations.mdx`.

### The 65-type map, and why it is machine-checked

The hub page carries a table of every registered notification type, grouped by declaring module: severity, declared channels, and the `non-opt-out` / `silent` / `hidden` flags.

A 65-row table hand-maintained across 22 modules is stale within a release, so **`apps/docs/__tests__/notification-registry.test.mjs`** asserts it against the real convention files, wired into `apps/docs`'s `test` script (which the root `turbo run test` already picks up). It fails when a registered type is undocumented, when a documented type no longer exists, when any cell of a row — severity, the declared channel ids, the flag markers — disagrees with its source, when a type is grouped under the wrong module or an enterprise group loses its marker, when the stated `37 of 65` ratio drifts, or when a referenced `packages/…` source path stops existing.

The review pass added a second file on the same principle: **`apps/docs/__tests__/push-env-table.test.mjs`** holds the `OM_PUSH_*` table on `push-notifications.mdx` to the module source — the documented variable set must equal the set the module actually reads, and every default and floor must match the constant it restates. The numbers live only in the source; the test stores locations, so changing a default fails the table until it is updated.

Two parsing details a naive implementation gets wrong, both pinned by comments in the test: the first `[` after `notificationTypes` belongs to the `NotificationTypeDefinition[]` annotation rather than to the array, and `type:` is not always a literal — `push_notifications` declares its two admin types through module-level constants, which a literal-only regex silently drops.

### An honest number the docs now state

**37 of the 65 registered types declare no `channels`**, and `resolveEligibleChannels` reads that as *no restriction* — so they become push-eligible the moment a tenant connects a provider. Both the hub page and the admin guide say so plainly, with a warning to review the Notification Delivery matrix immediately after connecting.

This is documented as **current behavior with an open policy question** (#5495, PR #5542), not as though the fix had landed. Worth flagging that #5495's own framing — "20 of 48" — is stale against `develop`; today it is 37 of 65, which is the argument for asserting the set instead of writing counts into prose.

## Specification

- [x] Yes
- [ ] No (created a new spec)

**Spec file path:**
`.ai/specs/2026-04-28-push-notifications-and-devices.md` (plus `2026-07-01-push-delivery-e2e-findings.md`, `2026-07-03-push-channels-tenant-scope.md`, `2026-08-24-devices-api-camelcase-responses.md`) — these are the source material the pages document. No spec change was needed: this PR adds no behavior.

## Testing

Local runner.

| Gate | Result |
|---|---|
| `yarn workspace open-mercato-docs test` (clean → full Docusaurus build → `node --test`) | **16/16 pass**, build succeeds |
| `yarn lint` | 0 errors |

The build reports one broken anchor, on `/installation/wsl2` — pre-existing, untouched here.

The new test caught a real bug in its own first run (a segment-case assumption that missed `checkout.link.usageLimitReached`), and I verified it fails on a deliberately mismarked row before keeping it. Every assertion added in the review pass was checked the same way — ten deliberate mutations (a wrong default, a dropped row, a removed callout, a wrong severity, a dropped flag, a wrong channel list, a removed `(enterprise)` marker, a stale count, a changed source constant, a duplicated type id), each failing its intended test and nothing else.

## Checklist

- [x] I updated documentation, locales, or generators if the change requires it — this PR *is* the documentation update.
- [x] I added or adjusted tests that cover the change — `apps/docs/__tests__/notification-registry.test.mjs` and `apps/docs/__tests__/push-env-table.test.mjs`.
- [x] I added or updated integration tests — N/A in the Playwright sense; docs correctness is asserted by the `node:test` file above, which is the automated coverage this change can carry.
- [x] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable) — not applicable; no behavior changes.
- [ ] Priority set — this account has no label permission upstream; requested in a comment below.
- [ ] Risk set — as above.
- [ ] QA routing set — as above; `skip-qa` requested (no `.tsx`, no schema, no API surface, no `BACKWARD_COMPATIBILITY.md` contract, and automated coverage ships with it).

### Design System Compliance
- [x] No hardcoded status colors — Markdown only; no component code in this PR.
- [x] No arbitrary text sizes — as above.
- [x] Empty state handled for list/data pages — N/A, no pages rendered by the app change.
- [x] Loading state handled for async pages — N/A, as above.
- [x] `aria-label` on icon-only buttons — N/A, as above.
- [x] Uses existing DS components — N/A, as above.

## Linked issues

Fixes #5591 (second of its two asks — the searchable user picker is #5617).
Documents the current behavior behind #5495 without pre-empting its resolution.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5618"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790252133&installation_model_id=432243&pr_number=5618&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5618&signature=3042858925e7e27605b40a4b1630196f3b7bd677349d40ba4fbfdd0fcf310ee8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/apps/docs/__tests__/notification-registry.test.mjs
+++ b/apps/docs/__tests__/notification-registry.test.mjs
@@ -1,0 +1,194 @@
+import assert from 'node:assert/strict';
+import { existsSync, readdirSync, readFileSync } from 'node:fs';
+import { readFile } from 'node:fs/promises';
+import { join, relative, resolve } from 'node:path';
+import test from 'node:test';
+import { fileURLToPath } from 'node:url';
+
+// The registry has no build artefact to read, so the map on the page is checked against the
+// convention files themselves. Mirrors apps/docs/__tests__/reference-example-module.test.mjs.
+const repoRoot = fileURLToPath(new URL('../../../', import.meta.url));
+const pageUrl = new URL('../docs/framework/modules/notification-delivery.mdx', import.meta.url);
+
+// Scaffold copies and generator source both contain the string but declare no live types.
+const EXCLUDED_FRAGMENTS = [
+  join('packages', 'create-app', 'template'),
+  join('packages', 'cli', 'src'),
+  join('external', 'official-modules'),
+];
+const SKIPPED_DIRECTORIES = new Set(['node_modules', 'dist', '.next', '.turbo', '.mercato']);
+
+function collectConventionFiles(dir, found = []) {
+  let entries;
+  try {
+    entries = readdirSync(dir, { withFileTypes: true });
+  } catch {
+    return found;
+  }
+  for (const entry of entries) {
+    const full = join(dir, entry.name);
+    if (entry.isDirectory()) {
+      if (SKIPPED_DIRECTORIES.has(entry.name)) continue;
+      collectConventionFiles(full, found);
+    } else if (entry.name === 'notifications.ts') {
+      found.push(full);
+    }
+  }
+  return found;
+}
+
+function conventionFiles() {
+  const files = [resolve(repoRoot, 'packages'), resolve(repoRoot, 'apps')]
+    .flatMap((root) => collectConventionFiles(root))
+    .filter((file) => !EXCLUDED_FRAGMENTS.some((fragment) => file.includes(fragment)))
+    // A module-root notifications.ts only; api/portal/notifications.ts and lib/notifications.ts
+    // are unrelated files that happen to share the name.
+    .filter((file) => /[\\/]modules[\\/][^\\/]+[\\/]notifications\.ts$/.test(file))
+    .sort();
+  assert.ok(
+    files.length > 0,
+    'found no notifications.ts convention files — a broken glob must fail loudly, not pass vacuously',
+  );
+  return files;
+}
+
+// Splits the `notificationTypes` array literal into one chunk per entry. The first `[` after the
+// binding name belongs to the `NotificationTypeDefinition[]` annotation, so the array is located
+// from the `=` instead.
+function splitTypeEntries(source) {
+  const binding = source.search(/notificationTypes[^=]*=\s*\[/);
+  if (binding < 0) return [];
+  const start = source.indexOf('[', source.indexOf('=', binding));
+  if (start < 0) return [];
+  const chunks = [];
+  let depth = 0;
+  let current = '';
+  for (let index = start + 1; index < source.length; index += 1) {
+    const char = source[index];
+    if (char === '{') {
+      depth += 1;
+      if (depth === 1) {
+        current = '';
+        continue;
+      }
+    }
+    if (char === '}') {
+      depth -= 1;
+      if (depth === 0) {
+        chunks.push(current);
+        continue;
+      }
+    }
+    if (depth === 0 && char === ']') break;
+    if (depth >= 1) current += char;
+  }
+  return chunks;
+}
+
+function readRegistry() {
+  const types = new Map();
+  for (const file of conventionFiles()) {
+    const source = readFileSync(file, 'utf8');
+    // `type:` is not always a literal — push_notifications declares its two admin types through
+    // module-level constants, and a naive literal-only regex would silently drop them.
+    const constants = new Map();
+    for (const match of source.matchAll(/const\s+([A-Za-z_$][\w$]*)\s*=\s*'([^']+)'/g)) {
+      constants.set(match[1], match[2]);
+    }
+    for (const chunk of splitTypeEntries(source)) {
+      const declared = chunk.match(/(?:^|\n)\s*type:\s*(?:'([^']+)'|([A-Za-z_$][\w$]*))/);
+      if (!declared) continue;
+      const id = declared[1] ?? constants.get(declared[2]);
+      assert.ok(
+        id,
+        `could not resolve the type id in ${relative(repoRoot, file)} — extend the constant lookup rather than dropping the entry`,
+      );
+      types.set(id, {
+        file: relative(repoRoot, file),
+        declaresChannels: /(?:^|\n)\s*channels:\s*\[/.test(chunk),
+      });
+    }
+  }
+  assert.ok(types.size > 0, 'parsed no notification types out of the convention files');
+  return types;
+}
+
+function mapSection(pageSource) {
+  const start = pageSource.indexOf('<!-- notification-type-map:start -->');
+  const end = pageSource.indexOf('<!-- notification-type-map:end -->');
+  assert.ok(start >= 0 && end > start, 'the page must keep the notification-type-map markers');
+  return pageSource.slice(start, end);
+}
+
+function documentedRows(section) {
+  const rows = new Map();
+  for (const line of section.split('\n')) {
+    // Segments are not always lower_snake — `checkout.link.usageLimitReached` is camelCase.
+    const match = line.match(/^\|\s*`([A-Za-z_][A-Za-z0-9_]*(?:\.[A-Za-z0-9_]+)+)`\s*\|/);
+    if (!match) continue;
+    rows.set(match[1], line);
+  }
+  return rows;
+}
+
+test('every registered notification type is documented in the map', async () => {
+  const registry = readRegistry();
+  const documented = documentedRows(mapSection(await readFile(pageUrl, 'utf8')));
+
+  const missing = [...registry.keys()].filter((id) => !documented.has(id)).sort();
+  assert.deepEqual(
+    missing,
+    [],
+    'these notification types exist in the registry but are absent from notification-delivery.mdx',
+  );
+});
+
+test('every documented notification type still exists in the registry', async () => {
+  const registry = readRegistry();
+  const documented = documentedRows(mapSection(await readFile(pageUrl, 'utf8')));
+
+  const stale = [...documented.keys()].filter((id) => !registry.has(id)).sort();
+  assert.deepEqual(
+    stale,
+    [],
+    'these notification types are documented but no longer declared by any module',
+  );
+});
+
+test('a type that declares no channels is rendered as eligible for every channel', async () => {
+  const registry = readRegistry();
+  const documented = documentedRows(mapSection(await readFile(pageUrl, 'utf8')));
+
+  // The "no channels declared" default is the platform's sharpest edge (#5495): such a type becomes
+  // push-eligible the moment a tenant connects a provider. Pinning the marker means resolving that
+  // policy question cannot silently invalidate the page.
+  const mismarked = [];
+  for (const [id, entry] of registry) {
+    const row = documented.get(id);
+    if (!row) continue;
+    const marked = row.includes('— (all registered)');
+    if (entry.declaresChannels === marked) mismarked.push(id);
+  }
+  assert.deepEqual(
+    mismarked.sort(),
+    [],
+    'these rows disagree with their source about whether the type declares a channel list',
+  );
+});
+
+test('every source path referenced by the notification docs exists', async () => {
+  const pages = [
+    '../docs/framework/modules/notification-delivery.mdx',
+    '../docs/framework/modules/push-notifications.mdx',
+    '../docs/framework/modules/devices.mdx',
+    '../docs/user-guide/notifications-and-push.mdx',
+  ];
+  const missing = [];
+  for (const page of pages) {
+    const source = await readFile(new URL(page, import.meta.url), 'utf8');
+    for (const match of source.matchAll(/`((?:packages|apps)\/[A-Za-z0-9._/-]+\.[A-Za-z]+)`/g)) {
+      if (!existsSync(resolve(repoRoot, match[1]))) missing.push(`${page} → ${match[1]}`);
+    }
+  }
+  assert.deepEqual(missing.sort(), [], 'these referenced source paths no longer exist');
+});
--- a/apps/docs/__tests__/notification-registry.test.mjs
+++ b/apps/docs/__tests__/notification-registry.test.mjs
@@ -16,7 +16,18 @@ const EXCLUDED_FRAGMENTS = [
   join('packages', 'cli', 'src'),
   join('external', 'official-modules'),
 ];
-const SKIPPED_DIRECTORIES = new Set(['node_modules', 'dist', '.next', '.turbo', '.mercato']);
+// `build` and `.docusaurus` are produced by this workspace's own `test` script immediately before
+// this walk, so leaving them in would crawl the freshly built site on every run.
+const SKIPPED_DIRECTORIES = new Set([
+  'node_modules',
+  'dist',
+  '.next',
+  '.turbo',
+  '.mercato',
+  'build',
+  '.docusaurus',
+]);
+const ENTERPRISE_PACKAGE = join('packages', 'enterprise') + '/';
 
 function collectConventionFiles(dir, found = []) {
   let entries;
@@ -85,6 +96,29 @@ function splitTypeEntries(source) {
   return chunks;
 }
 
+/** The module directory the convention file sits in — the grouping the map's headings use. */
+function moduleOf(file) {
+  return /[\\/]modules[\\/]([^\\/]+)[\\/]notifications\.ts$/.exec(file)[1];
+}
+
+/** The rendered flag cell for one entry, in the order the map writes them. */
+function flagsOf(chunk) {
+  const flags = [];
+  if (/(?:^|\n)\s*nonOptOut:\s*true/.test(chunk)) flags.push('non-opt-out');
+  if (/(?:^|\n)\s*silent:\s*true/.test(chunk)) flags.push('silent');
+  if (/(?:^|\n)\s*hiddenFromSettings:\s*true/.test(chunk)) flags.push('hidden from settings');
+  return flags.length > 0 ? flags.join(', ') : '—';
+}
+
+/** The rendered channel cell — the declared ids, or the marker meaning "no restriction". */
+function channelsOf(chunk) {
+  const declared = chunk.match(/(?:^|\n)\s*channels:\s*\[([^\]]*)\]/);
+  if (!declared) return '— (all registered)';
+  const ids = [...declared[1].matchAll(/'([^']+)'/g)].map((match) => match[1]);
+  assert.ok(ids.length > 0, 'a declared channels list must name at least one channel id');
+  return ids.map((id) => `\`${id}\``).join(', ');
+}
+
 function readRegistry() {
   const types = new Map();
   for (const file of conventionFiles()) {
@@ -103,8 +137,26 @@ function readRegistry() {
         id,
         `could not resolve the type id in ${relative(repoRoot, file)} — extend the constant lookup rather than dropping the entry`,
       );
+      const severity = chunk.match(/(?:^|\n)\s*severity:\s*'([^']+)'/);
+      assert.ok(
+        severity,
+        `${id} declares no literal severity in ${relative(repoRoot, file)} — extend the parser rather than dropping the field`,
+      );
+      const relativeFile = relative(repoRoot, file);
+      // Keyed on the type id, so two modules declaring the same id would otherwise collapse into one
+      // entry and a single documented row would satisfy both.
+      const previous = types.get(id);
+      assert.ok(
+        !previous,
+        `notification type ${id} is declared twice — ${previous?.file} and ${relativeFile}`,
+      );
       types.set(id, {
-        file: relative(repoRoot, file),
+        file: relativeFile,
+        module: moduleOf(file),
+        enterprise: relativeFile.startsWith(ENTERPRISE_PACKAGE),
+        severity: severity[1],
+        channels: channelsOf(chunk),
+        flags: flagsOf(chunk),
         declaresChannels: /(?:^|\n)\s*channels:\s*\[/.test(chunk),
       });
     }
@@ -120,20 +172,32 @@ function mapSection(pageSource) {
   return pageSource.slice(start, end);
 }
 
+/** Every documented row, with its cells and the group heading it sits under. */
 function documentedRows(section) {
   const rows = new Map();
+  let group = null;
   for (const line of section.split('\n')) {
+    const heading = line.match(/^####\s+`([^`]+)`(.*)$/);
+    if (heading) {
+      group = { module: heading[1], enterprise: heading[2].includes('*(enterprise)*') };
+      continue;
+    }
     // Segments are not always lower_snake — `checkout.link.usageLimitReached` is camelCase.
-    const match = line.match(/^\|\s*`([A-Za-z_][A-Za-z0-9_]*(?:\.[A-Za-z0-9_]+)+)`\s*\|/);
+    const match = line.match(/^\|\s*`([A-Za-z_][A-Za-z0-9_]*(?:\.[A-Za-z0-9_]+)+)`\s*\|(.*)$/);
     if (!match) continue;
-    rows.set(match[1], line);
+    const cells = match[2].split('|').map((cell) => cell.trim());
+    rows.set(match[1], { line, group, severity: cells[0], channels: cells[1], flags: cells[2] });
   }
   return rows;
 }
 
+async function registryAndRows() {
+  const page = await readFile(pageUrl, 'utf8');
+  return { registry: readRegistry(), documented: documentedRows(mapSection(page)), page };
+}
+
 test('every registered notification type is documented in the map', async () => {
-  const registry = readRegistry();
-  const documented = documentedRows(mapSection(await readFile(pageUrl, 'utf8')));
+  const { registry, documented } = await registryAndRows();
 
   const missing = [...registry.keys()].filter((id) => !documented.has(id)).sort();
   assert.deepEqual(
@@ -144,8 +208,7 @@ test('every registered notification type is documented in the map', async () => 
 });
 
 test('every documented notification type still exists in the registry', async () => {
-  const registry = readRegistry();
-  const documented = documentedRows(mapSection(await readFile(pageUrl, 'utf8')));
+  const { registry, documented } = await registryAndRows();
 
   const stale = [...documented.keys()].filter((id) => !registry.has(id)).sort();
   assert.deepEqual(
@@ -156,8 +219,7 @@ test('every documented notification type still exists in the registry', async ()
 });
 
 test('a type that declares no channels is rendered as eligible for every channel', async () => {
-  const registry = readRegistry();
-  const documented = documentedRows(mapSection(await readFile(pageUrl, 'utf8')));
+  const { registry, documented } = await registryAndRows();
 
   // The "no channels declared" default is the platform's sharpest edge (#5495): such a type becomes
   // push-eligible the moment a tenant connects a provider. Pinning the marker means resolving that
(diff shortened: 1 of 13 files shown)


Frshy:
**Metadata I cannot apply — this account has no label permission here (`push: false, triage: false`).** Requesting a maintainer set:

| Label | Rationale |
|---|---|
| `documentation` | Docs-only; no runtime code. |
| `review` | Ready, not draft, and in no other pipeline state. |
| `priority-medium` | Nothing is broken — the docs were simply absent. This is the lowest tier I would argue for; `priority-low` if you prefer it. |
| `risk-low` | No `.tsx`, no schema, no API surface, no `BACKWARD_COMPATIBILITY.md` contract. |
| `skip-qa` | Clean automated-verification exemption: no manually exercisable UI surface, database and API unchanged, and the new `node:test` file is the automated coverage the exemption asks for. |

**Reviewer:** whoever owns `apps/docs` and the `notifications` / `push_notifications` modules — deliberately not tagging a handle.

**Assignee:** not claiming.

Independent of #5617 — the two branches are both off `develop` and can merge in either order.

One judgement call worth a maintainer's eye rather than mine: the pages document the **current** default, that a type declaring no `channels` is eligible for every registered channel and so becomes push-eligible as soon as a tenant connects a provider. #5495 and PR #5542 are cited as an open policy question, and the docs deliberately do not pre-announce a resolution. If you would rather the pages read as though #5542's behavior is the intended one, say so and I will rewrite that section — but I did not want documentation to describe a state the shipped code is not in.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-25T23:19:28Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review

## 🎯 Summary

This PR adds four documentation pages covering the notification delivery surface — a framework hub page (`notification-delivery.mdx`), two module reference pages (`devices.mdx`, `push-notifications.mdx`), and an operator walkthrough (`user-guide/notifications-and-push.mdx`) — plus sidebar wiring, `core-modules.mdx` rows, cross-links, a re-positioned tutorial, and a `node:test` file that asserts the 65-row notification-type map against the real convention files.

I verified the substantive technical claims against the source rather than taking them on trust, and they hold up unusually well:

- The four-condition delivery gate and its **order** match `packages/core/src/modules/notifications/lib/shouldDeliver.ts:75-79` exactly, including the point that eligibility is evaluated *before* the `nonOptOut` bypass and that `silent` is deliberately not consulted.
- `resolveEligibleChannels` really does return `null` for "no restriction" (`shouldDeliver.ts:41-43`), so the page's central warning is correct rather than rhetorical.
- I independently re-parsed all 22 `notifications.ts` convention files: **65 types, 37 of them declaring no `channels`**, and every `severity` and every `non-opt-out` / `silent` / `hidden from settings` flag in the 65-row map matches its source. I found no incorrect row.
- The `devices.mdx` unique-index SQL matches the *current* migration (`Migration20260721120000.ts:10`) rather than the superseded coalesce-sentinel form in `Migration20260630120000.ts` — a detail that is easy to get wrong and was not.
- ACL feature lists, backend page paths, API route paths, worker ids and concurrency (`push_notifications:send-push`, concurrency 8), `MAX_ATTEMPTS = 3`, and every `last_error` code in the lookup table check out against the code.
- The `notifications.delivery-strategies.ts` snippet omitting `export default` is fine: the generator plugin accepts either shape (`packages/core/src/modules/notifications/generators.ts:29`).
- Every relative link and the `#environment-variables` anchor introduced by this PR resolves; `framework/modules/communication-channels` now appears exactly once in `sidebars.ts` and does not collide with the distinct `user-guide/communication-channels`.
- The new test passes on this head (4/4), and the parsing subtleties its comments call out — the annotation `[` and the constant-declared `type:` values in `push_notifications` — are real ones that a naive implementation would get wrong.

What holds this back is a small set of accuracy and durability gaps in the reference tables, not the substance. The most concrete is that the env-var table **loses a default this PR deletes from the tutorial**, and the second is that the page whose selling point is machine-checked numbers ships two hardcoded numbers the test does not check.

**Verdict: request changes.** The three Medium findings below drive it under this repository's rule that Medium findings request changes. None of them require rework of the approach — they are four small edits to two files plus an optional handful of lines in the test.

## 🧪 Validation Gate

**Runner: local. Validation source: local fallback — GitHub check data was unusable at review time.**

This is a fork PR (`fullstackhouse:docs/notifications-devices-push-surface`), and both workflow runs for head `d630806` sat at `action_required`, so **no check had ever executed on this head**. Per the repo's checks-first rule that is exactly the "no usable check data" case, so I approved the two blocked runs (real CI signal now exists and is reported below) and ran a narrow local fallback scoped to the changed area rather than the full workspace gate — a docs-only diff touches no package the workspace-wide commands would exercise.

| Command | Result |
|---|---|
| `node --test __tests__/notification-registry.test.mjs` (in `apps/docs`) | ✅ 4/4 pass |
| Independent re-parse of all 22 `notifications.ts` files, cross-checked against every row of the map | ✅ 65 types, 37 without `channels`; all severities and flags correct |
| Relative-link and anchor resolution across all 8 touched/linked pages | ✅ no broken target introduced by this PR |
| Source-claim verification (gate semantics, migrations, ACL, routes, workers, env defaults) | ✅ no incorrect claim found |
| `yarn build:packages` / `yarn generate` / `yarn typecheck` / `yarn i18n:check-*` / `yarn build:app` | ⏭️ not run — no source, locale, entity or app file is touched; deliberately left to CI |
| Full Docusaurus build (`yarn workspace open-mercato-docs test`) | ⏭️ not re-run locally — the author reports it green, and CI's `Test` job runs exactly this via `turbo run test` |

**CI is still running on this head at the time of this review.** `lint`, `ds-lint`, `audit`, `scope` and `audit-scope` have passed; `prepare` is still pending, and the jobs it gates — including the `Test` job that runs the full Docusaurus build and the new `node:test` file — have not reported yet. A follow-up comment will report the CI outcome.

## 🟡 Medium findings

### 1. The env-var table drops three real defaults, one of which this PR deletes from the tutorial

`apps/docs/docs/framework/modules/push-notifications.mdx:123-125`

The three receipt-polling rows render `—` in the **Default** column, while every other row states its default. The code has real defaults for all three (`packages/core/src/modules/push_notifications/lib/push-receipt-reaper.ts:43,47,50`):

- `OM_PUSH_RECEIPT_MIN_AGE_MINUTES` → **15**
- `OM_PUSH_RECEIPT_MAX_AGE_MINUTES` → **60**
- `OM_PUSH_RECEIPT_BATCH_LIMIT` → **500** (floored at 1)

Read against a table where sibling rows carry concrete numbers, `—` reads as "there is no default", which invites an operator to set values they do not need to set. It also makes this PR a net loss of one documented fact: the tutorial table this PR removes (`devices-and-push-getting-started.mdx`, old section 8) documented `OM_PUSH_RECEIPT_MIN_AGE_MINUTES` as `15` with the rationale "Expo needs time to resolve one", and the reference page it hands off to does not carry the number forward. The PR description states the reference "now includes … the receipt knobs the tutorial never listed", which is accurate for two of the three but not for `OM_PUSH_RECEIPT_MIN_AGE_MINUTES`.

**Fix:** fill in `15`, `60` and `500`, and consider carrying over the one-line "Expo needs time to resolve a receipt" rationale that made the tutorial row useful.

### 2. `37 of the 65` is a hardcoded count in the one page that advertises machine-checked numbers

`apps/docs/docs/framework/modules/notification-delivery.mdx:88`

Both numbers are correct today — I confirmed 65 and 37 independently. But `notification-registry.test.mjs` asserts *set membership* and the per-row `— (all registered)` marker; it never asserts either count. Adding one notification type without a `channels` list makes this sentence wrong while all four tests stay green, and line 151 tells the reader the map is machine-checked, so a stale number here is more misleading than it would be on an unasserted page.

This is also the exact failure mode the PR description diagnoses in #5495's own framing ("20 of 48" being stale against `develop`) and gives as the reason for asserting the set instead of writing counts into prose — the same argument applies to this sentence.

**Fix:** the test already computes both numbers, so pinning them is about four lines:

```js
test('the stated no-channels ratio matches the registry', async () => {
  const registry = readRegistry();
  const withoutChannels = [...registry.values()].filter((entry) => !entry.declaresChannels).length;
  const page = await readFile(pageUrl, 'utf8');
  assert.match(page, new RegExp(`\\*\\*${withoutChannels} of the ${registry.size} types`));
});
```

Rewording to "most types in the registry omit it (see the map below)" would also close it, at the cost of the concrete number that makes the warning land.

### 3. `OM_PUSH_SEND_TIMEOUT_MS` is silently clamped by the reclaim window, and the table does not say so

`apps/docs/docs/framework/modules/push-notifications.mdx:119`

The row documents `60000 (min 5000)`, which is right, but the effective value is `Math.min(configured, max(5000, resolveStuckThresholdMs() − 60000))` (`packages/core/src/modules/push_notifications/lib/push-delivery.ts:34-40`). Two consequences an operator cannot infer from this page:

- Setting `OM_PUSH_SEND_TIMEOUT_MS` above `OM_PUSH_STUCK_RECLAIM_MINUTES − 60s` has no effect; the configured value is discarded without a warning.
- Lowering `OM_PUSH_STUCK_RECLAIM_MINUTES` to its minimum of `1` forces the send timeout to `5000` **regardless of what `OM_PUSH_SEND_TIMEOUT_MS` says** — the two knobs are coupled in a direction the table presents as independent.

This is the same class of silent, unlogged behavior the page sets out to document elsewhere (the two silent fan-out preconditions), so it belongs here rather than in a reader's debugging session.

**Fix:** extend the row's Effect cell to note that the value is clamped to stay at least 60 s below the reclaim window, per the comment in `push-delivery.ts:22-25`.

## 🟢 Low findings / nits

- **`apps/docs/docs/framework/modules/notification-delivery.mdx:109`** — "`packages/core/src/modules/push_notifications/notifications.delivery-strategies.ts` is three lines" is a nice rhetorical beat but the file is eight lines (an import, a two-line comment, the named export and a default export). "Two exports and an import" would keep the point and survive a comment being added.
- **`apps/docs/__tests__/notification-registry.test.mjs:106-108`** — the parser already isolates each entry chunk, so `severity` and the flag columns could be asserted with roughly the same six lines that assert the channels marker. Today they are correct in all 65 rows (I checked every one), but nothing keeps them that way; the map's own promise on line 151 would then cover the whole row rather than one column of it.
- **`apps/docs/docs/framework/modules/push-notifications.mdx:5`** — the frontmatter promises "every `OM_PUSH_*` variable", and `OM_PUSH_RENEWAL_GMAIL_LEAD_HOURS` (`packages/core/src/modules/communication_channels/workers/gmail-renew-watch.ts`) is an `OM_PUSH_*` name outside the table. Excluding it is the right scoping call — it is a Gmail watch-renewal knob with nothing to do with push transport — so this is purely a wording fix: "every `OM_PUSH_*` variable this module reads".
- **`apps/docs/__tests__/notification-registry.test.mjs:19,106`** — two small robustness gaps. `types.set(id, …)` is keyed on the type id across all files, so two modules declaring the same id collapse into one entry and a single documented row would satisfy both; and `SKIPPED_DIRECTORIES` omits `build` and `.docusaurus`, which the `test` script itself creates immediately before this walk, so the recursion crawls the freshly built site on every run. Neither is wrong today, both are cheap to close.
- **`apps/docs/docs/framework/modules/notification-delivery.mdx`** (map section) — the `record_locks` (8) and `security` (4) groups are declared in `packages/enterprise`, so twelve of the sixty-five rows are unavailable to an OSS reader with no indication of that. A short "(enterprise)" marker on those two group headings would save a reader hunting for types they cannot have.

## 💥 Breaking Changes

- [x] No exported API, type, or import path changed — the only non-Markdown source change is one string in `apps/docs/package.json`, appending a test file to the `test` script.
- [x] No HTTP route, response shape, or event name touched.
- [x] No database schema, migration, or snapshot touched.
- [x] No CLI command, DI key, ACL feature, notification id, or widget spot id touched.
- [x] No generated file hand-edited.
- [x] `BACKWARD_COMPATIBILITY.md` protected surfaces: none of the thirteen categories is affected by this diff.

The `devices.mdx` "Response shape" section restates the existing camelCase-with-deprecated-snake_case-alias contract from `UPGRADE_NOTES.md`; it documents that contract rather than changing it.

## 🧪 Test Coverage

Coverage is appropriate and better than a docs PR usually carries. `apps/docs/__tests__/notification-registry.test.mjs` is wired into `apps/docs`'s `test` script, which the root `turbo run test` picks up, so it is a real CI gate rather than a file nobody runs. It fails loudly rather than vacuously when the glob breaks (`assert.ok(files.length > 0, …)`) and when the page's `notification-type-map` markers go missing — both of which are the failure modes that make this kind of generated-table test worthless in practice.

The gaps are the ones noted above: the two counts in prose and the `severity` / flags columns are outside the assertion boundary. The page's own description of what the test enforces (line 151 and the PR body) is honest about the first of these and silent about the second.

Integration/Playwright coverage is correctly N/A — nothing rendered by the application changed.

pkarw:

- 🚀 `merge-queue` — the re-review of head `88604fa` found no blocker, major or Medium finding: all three Medium findings and all five nits from the previous pass are fixed and re-verified against the source, and every required check is green.
- 📚 `documentation` — the change is four new documentation pages plus sidebar wiring, cross-links and the two `node:test` files that assert them against the module sources.
- ⏭️ `skip-qa` — the automated-verification exemption applies: no `.tsx` outside tests, nothing under `packages/ui/src/`, no database structure, API surface or `BACKWARD_COMPATIBILITY.md` contract touched, and automated coverage for the changed behaviour ships in the same PR.
- 🔹 `priority-medium` — kept as set; the docs gap this closes was raised in #5591 as blocking usability of the devices and notifications feature, which is current-cycle work rather than an incident.
- 🟢 `risk-low` — kept as set; the blast radius is documentation plus two test files under `apps/docs`, with no runtime code path affected.
- `partner-request` — kept as applied by the author's request comment.

pkarw:

Two notes that are not findings:

- The judgement call you flagged for a maintainer — documenting the no-`channels` default as current behaviour with #5495 cited as an open policy question, rather than pre-announcing #5542's resolution — is the right call. Keep it as written; docs should describe `develop`, not a pending decision.
- Both workflow runs on this head were sitting at `action_required` (fork PRs need per-push approval), so no check had ever executed. I approved them, and CI is now running for real.

pkarw:

This was a re-review of head `88604fa` against the `changes-requested` verdict left on `d630806`. All three Medium findings and all five nits from that pass are addressed, and I re-verified each against the source rather than accepting the summary: the three receipt defaults (`15`, `60`, `500`) and both newly documented floors match `push-receipt-reaper.ts`; the `37 of the 65` count is now computed from the convention files and asserted against the exact sentence, and I confirmed no `notifications.ts` file has changed on `develop` since this branch's merge base, so the count holds against the merge result too; and the send-timeout callout's formula matches `push-delivery.ts:35-40` exactly, including the `OM_PUSH_STUCK_RECLAIM_MINUTES=1` case pinning the timeout to 5000 ms.

What raises this above a routine fix pass is that each correction ships with the assertion that makes the same drift fail next time. `push-env-table.test.mjs` stores source *locations* rather than numbers, so changing a default in the module fails the table until the docs catch up — the finding-1 class made structurally impossible rather than corrected once. `notification-registry.test.mjs` grew from asserting one column to asserting the whole row plus the module grouping, the enterprise markers and both prose counts.

Three nits remain, none of which can produce an incorrect page today and none of which held the approval: the `(min N)` column carries two different behaviours (a true clamp for `OM_PUSH_RECLAIM_TICK_SECONDS`, a fallback-to-default threshold for the other two) — which is a correction of something I called "right" in the previous pass, not a new ask; `ENTERPRISE_PACKAGE` mixes `join()` with a hardcoded `/` and so misreads on Windows; and a type documented under two rows would silently collapse in `documentedRows`. Pick them up here or in a follow-up, whichever suits.

Validation source: GitHub checks — every check on `88604fa` is green, including the `Test` job that runs the full Docusaurus build and all four `node:test` files. I additionally ran the two docs test files locally as this pass's own evidence: 10/10 pass. Labels moved to `merge-queue`; `skip-qa` means the QA gate does not hold this one.

autofix: skipped (not my PR — re-run with --autofix to fix it here). The three nits are the author's or a follow-up's call, not something to push onto a fork branch uninstructed.

pkarw:

**All checks passed on head `88604fa`.** No wait was needed: both workflow runs for this head — `CI for Develop&Main` and `Mutation tests` — had already completed with `success` before the re-review began, so the verdict is unchanged by CI.

Green: `prepare` (3m23s), `test` (7m25s), `lint`, `ds-lint`, `audit`, `scope`, `audit-scope`, `documents-multi-instance`, `ephemeral-integration (none)`, `license/cla`. Correctly skipped for this diff: `docker-build`, `mutate`, `merge-coverage`, `comment`.

The check that carries this PR's own coverage is `test`, which runs `turbo run test` and therefore `apps/docs`'s `test` script — a clean rebuild, a full Docusaurus build, and `node --test` over all four test files including both of this PR's. That is the complete gate for a docs-only diff, so no local substitute was required. I did re-run the two new files locally as this pass's own evidence: 10/10 pass.

`ephemeral-integration (none)` is the affected-only shard reporting that no integration shard was selected for this diff, which is the expected result here rather than a skipped gate.

Frshy:
Pushed `88604fa` — all three Medium findings and all five nits are addressed. Thanks for the review; every one of them was right, and each fix ships with the assertion that makes the same drift fail next time rather than only correcting today's text.

### Medium

**1 — the three `—` defaults.** Filled in `15`, `60` and `500`, and carried the tutorial's rationale forward: `OM_PUSH_RECEIPT_MIN_AGE_MINUTES` now says *"Expo needs roughly this long to resolve one, so polling sooner finds nothing."* Also added the two floors that were nowhere: `MAX_AGE` is floored at `MIN + 1` minute so the window is never empty, and `BATCH_LIMIT` is floored at 1. You were right that this was a net loss of one documented fact, not a cosmetic gap.

**2 — the hardcoded count.** Pinned, as suggested. `the stated no-channels ratio matches the registry` computes both numbers from the convention files and asserts the sentence:

```
✖ the warning must state 38 of the 65 types — update notification-delivery.mdx
```

**3 — the send-timeout clamp.** The row now points at a callout that gives the formula and both consequences you named — a value above `reclaim window − 60 s` discarded silently, and `OM_PUSH_STUCK_RECLAIM_MINUTES=1` pinning the timeout to 5000 ms — plus which knob to raise first. The reclaim-window row says it caps the send timeout, so a reader arrives at the coupling from either side.

### Nits

- **"three lines"** → *"two imports and two exports"*. It is eight lines; the rhetorical beat survives a comment being added.
- **`severity` and the flag columns** → asserted, and so are the declared channel *ids* rather than just the presence marker. The page's promise on line 151 now covers the whole row.
- **frontmatter** → *"every `OM_PUSH_*` variable this module reads"*, and that promise is now itself a test: the documented set must equal the set of `OM_*PUSH*` names the module actually reads. `OM_PUSH_RENEWAL_GMAIL_LEAD_HOURS` lives in `communication_channels`, so it is correctly outside both the table and the assertion.
- **duplicate ids / `SKIPPED_DIRECTORIES`** → both closed. A type id declared by two modules now fails with both file paths instead of collapsing into one entry; `build` and `.docusaurus` are skipped, which mattered more than it looked — the workspace's own `test` script creates them seconds before the walk.
- **`(enterprise)` markers** → on `record_locks` and `security`, with a legend line above the map, and machine-checked: a group heading must carry the marker exactly when its declaring file sits under `packages/enterprise`.

### The new test file

`apps/docs/__tests__/push-env-table.test.mjs`, wired into the same `test` script. It asserts the env table against the module source: the documented variable set equals what the module reads, and every default and floor matches the constant it restates. The numbers are **not** repeated in the test — only the source location — so changing `DEFAULT_STUCK_MINUTES` in `reclaim-window.ts` fails with `documented default 5, source says 7` until the table is updated. That is the finding-1 class made impossible rather than merely fixed.

### Verification

Every new assertion was checked against a deliberate mutation before being kept — a wrong default, a dropped row, a removed callout, a wrong severity, a dropped flag, a wrong channel list, a removed `(enterprise)` marker, a stale count, a changed source constant, and a duplicated type id. All ten fail the intended test and nothing else; the suite is green with them reverted.

`yarn workspace open-mercato-docs test` (clean → full Docusaurus build → `node --test`): **16/16 pass**, build succeeds. The one broken anchor on `/installation/wsl2` is still there and still pre-existing. No lintable file changed — `apps/docs` has no `lint` script and the diff is `.mdx`, `.mjs` and one `package.json` string.

Re-requesting review. `changes-requested` still needs a maintainer to clear — this account has no label permission here.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-26T09:09:09Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review

**Re-review of PR #5618** (previous verdict: `changes-requested` on head `d630806`; this pass reviews head `88604fa`).

## 🎯 Summary

The previous pass raised three Medium findings and five nits. `88604fa` addresses **all eight**, and — this is the part worth calling out — it addresses them in the shape the review asked for rather than the cheap shape: every corrected number now ships with the assertion that makes the same drift fail next time, and in the one case where the number lived only in prose, the *number itself* was moved behind a test rather than merely retyped.

I re-verified each one against the source rather than taking the author's summary on trust:

- **Medium 1 — the three `—` defaults.** `15`, `60` and `500` are correct against `packages/core/src/modules/push_notifications/lib/push-receipt-reaper.ts:43,47,50`. The tutorial rationale this PR had dropped is carried forward on the `OM_PUSH_RECEIPT_MIN_AGE_MINUTES` row, matching the source comment at `push-receipt-reaper.ts:30-33`. Two floors that were previously documented nowhere are now stated and correct: `MAX_AGE` floored at `MIN + 1` minute (`Math.max(RECEIPT_MIN_AGE_MS + 60_000, …)`, line 45-48) and `BATCH_LIMIT` floored at 1 (`Math.max(1, …)`, line 50).
- **Medium 2 — the hardcoded `37 of the 65`.** Pinned by `the stated no-channels ratio matches the registry` (`notification-registry.test.mjs:283-294`), which computes both numbers from the convention files and asserts the exact sentence. I confirmed independently that 37 and 65 are still the true values on this head, and that no `notifications.ts` convention file has changed on `develop` since this branch's merge base — so the count is correct against the merge result, not just against the branch.
- **Medium 3 — the send-timeout clamp.** The new callout's formula, `min(OM_PUSH_SEND_TIMEOUT_MS, max(5000, reclaim window − 60s))`, matches `push-delivery.ts:35-40` exactly. Both consequences are correct: at the default 5-minute window the ceiling is 240 000 ms, and `OM_PUSH_STUCK_RECLAIM_MINUTES=1` yields `max(5000, 60000 − 60000) = 5000`, pinning the timeout regardless of the configured value. The coupling is now stated from both sides of the table, so a reader arrives at it from either row.
- **Nits.** `notifications.delivery-strategies.ts` is indeed two imports and two exports across eight lines. `severity`, the declared channel *ids* and the flag cell are now asserted per row rather than only the `— (all registered)` marker. The frontmatter promise is narrowed to "this module reads" **and** turned into an assertion — I confirmed the only `OM_*PUSH*` names outside the module are in `packages/cli` test harnesses, `channel-expo` tests, and `OM_PUSH_RENEWAL_GMAIL_LEAD_HOURS` in `communication_channels`, so excluding it remains the right scoping call and the assertion agrees. Duplicate type ids now fail with both file paths instead of collapsing. `build` and `.docusaurus` are skipped. `record_locks` and `security` carry `*(enterprise)*` markers, both are genuinely declared under `packages/enterprise/src/modules/`, and the marker is itself machine-checked against the declaring path.

The new `push-env-table.test.mjs` is the right design for this problem: it stores *locations* rather than numbers (`SOURCES`, lines 22-54), so a changed constant fails the table until the docs are updated — which is the finding-1 class made structurally impossible rather than corrected once. It also fails loudly rather than vacuously in both directions that matter (`names.size > 0` on the walk, `rows.size > 0` on the parse, and set equality asserted *both* ways).

**Verdict: approve.** No blocker, major or Medium finding remains. One nit below, which is a correction of something I got wrong in the previous pass rather than a new demand — it does not need to hold this PR.

## 🧪 Validation Gate

**Validation source: GitHub checks** (per this repository's checks-first rule), with a narrow local re-run of the two docs test files as this pass's own evidence.

Every check on head `88604fa` is green — `prepare` (3m23s), `test` (7m25s), `lint`, `ds-lint`, `audit`, `scope`, `audit-scope`, `documents-multi-instance`, `ephemeral-integration (none)`, `license/cla`. The `test` job is the one that matters here: it runs `turbo run test`, which picks up `apps/docs`'s `test` script — clean, full Docusaurus build, then `node --test` over all four files including both of this PR's. That is the full gate for this diff, so no local fallback was required.

| Command | Result |
|---|---|
| CI `Test` job on `88604fa` (full Docusaurus build + `node --test`) | ✅ pass |
| `node --test __tests__/notification-registry.test.mjs __tests__/push-env-table.test.mjs` (local, in `apps/docs`) | ✅ 10/10 pass |
| Source re-verification of all three Medium fixes and all five nits against `packages/core/src/modules/push_notifications` | ✅ no incorrect claim found |
| Drift check: `notifications.ts` convention files and `push_notifications` sources on `develop` since the merge base | ✅ unchanged — the 37/65 count and every default hold against the merge result |
| `yarn build:packages` / `yarn generate` / `yarn typecheck` / `yarn i18n:check-*` / `yarn build:app` | ⏭️ covered by the green CI run; nothing outside `apps/docs` is touched |

## 🟢 Low findings / nits

- **`apps/docs/docs/framework/modules/push-notifications.mdx:117-118`** — `(min N)` now carries two different meanings in the same column, and the new test pins both without distinguishing them. `OM_PUSH_RECLAIM_TICK_SECONDS` is a genuine clamp (`Math.max(30, …)`, `setup.ts:32-35`), so `120 (min 30)` behaves as a reader expects. But `OM_PUSH_SEND_TIMEOUT_MS` and `OM_PUSH_STUCK_RECLAIM_MINUTES` are *rejection thresholds with a fallback to the default*: `parsed >= MIN_SEND_TIMEOUT_MS ? parsed : DEFAULT_SEND_TIMEOUT_MS` (`push-delivery.ts:38`) and the same shape at `reclaim-window.ts:14`. So `OM_PUSH_SEND_TIMEOUT_MS=3000` yields **60000**, not 5000, and `OM_PUSH_STUCK_RECLAIM_MINUTES=0` yields **5**, not 1 — in both cases the opposite of the operator's intent, and unlogged.

  Nothing on the page is false — "min 5000" correctly says 5000 is the lowest accepted value — so this is under-specification rather than error, and I called this row "right" in the previous pass, which is the correction. It is worth closing because it is the same silent-discard class the new callout exists to document, and the callout's own `min(configured, …)` formula inherits it: `configured` is the post-fallback value, so the formula reads as if 3000 survives. A parenthetical on the two affected rows — "below 5000 falls back to the default rather than clamping" — plus one clause in the callout would close both, and `SOURCES` could distinguish `min` (fallback threshold) from a true `floor` if you want the test to carry the distinction too.

- **`apps/docs/__tests__/notification-registry.test.mjs:30`** — `ENTERPRISE_PACKAGE` builds `join('packages', 'enterprise') + '/'`, mixing a platform-aware `join` with a hardcoded separator. On Windows `relative()` yields `packages\enterprise\…`, so `startsWith` never matches, every entry reads as non-enterprise, and the two `*(enterprise)*` markers then fail the grouping test. It fails loudly rather than silently and CI is Linux, but the rest of the file is careful about this (`[\\/]` at lines 57 and 101); `join('packages', 'enterprise', sep)` or a `[\\/]`-based regex keeps it consistent.

- **`apps/docs/__tests__/notification-registry.test.mjs:189`** — `documentedRows` keys on the type id, so a type documented under *two* rows silently collapses to the last one and no test notices. The source-side duplicate is now caught with both paths, which was the half that mattered; this is the mirror image and costs one `assert.ok(!rows.has(id), …)`.

## 💥 Breaking Changes

- [x] No exported API, type, or import path changed — the only non-Markdown change in `88604fa` is one string in `apps/docs/package.json` appending a second test file to the `test` script, and two new `__tests__` files.
- [x] No HTTP route, response shape, or event name touched.
- [x] No database schema, migration, or snapshot touched.
- [x] No CLI command, DI key, ACL feature, notification id, or widget spot id touched.
- [x] No generated file hand-edited.
- [x] `BACKWARD_COMPATIBILITY.md` protected surfaces: none of the thirteen categories is affected by this diff.

## 🧪 Test Coverage

Coverage is now materially stronger than at the previous head, and stronger than a docs PR normally carries. `notification-registry.test.mjs` grew from four tests asserting set membership and one column to seven asserting the whole row, the module grouping, the enterprise markers and both prose counts; `push-env-table.test.mjs` adds three more holding the env table to the module source. Both are wired into `apps/docs`'s `test` script and therefore into the root `turbo run test`, so they are real CI gates rather than files nobody runs.

The gap I flagged last time — the page promising machine-checked numbers while two of them were unasserted prose — is closed, and closed in the direction that survives future edits. The remaining assertion gaps are the two nits above, neither of which can produce an incorrect page today.

Integration/Playwright coverage remains correctly N/A: nothing rendered by the application changed.

Issue #388, opened 2026-01-21
feat: WMS module


pkarw:
# WMS Module Specification (Open Mercato)

## Scope Overview

**Module id:** `wms`  
**Package:** `packages/core/src/modules/wms/`

Core goals:
- Extend catalog (link product + variant to WMS entities) via foreign key ids only (no cross-module relations).
- Multi-warehouse, multi-location hierarchy.
- Inventory control with reservations, allocations, adjustments.
- FIFO/LIFO/FEFO and lot/batch/serial.
- Inbound receiving, putaway, picking, packing.

---

## Core Data Model (MikroORM entities)

All entities include: `id (uuid)`, `created_at`, `updated_at`, `deleted_at`, `tenant_id`, `organization_id`.

### 1) Warehouse
- `name` (string, required)
- `code` (string, unique per org)
- `is_active` (boolean)
- `address` (json or string fields: `address_line1`, `city`, `postal_code`, `country`)
- `timezone` (string)

### 2) WarehouseLocation
- `warehouse_id` (uuid)
- `code` (string, required, unique per warehouse)
- `type` (enum: zone|aisle|rack|bin|slot|dock|staging)
- `parent_id` (uuid, nullable)
- `is_active` (boolean)
- `capacity_units` (number, nullable)
- `capacity_weight` (number, nullable)
- `constraints` (json: max_height, max_width, temp_range, hazmat flags)

### 3) WarehouseZone
- `warehouse_id`
- `code`
- `name`
- `priority` (number, for picking/putaway)

### 4) ProductInventoryProfile
- `catalog_product_id` (uuid, required)
- `catalog_variant_id` (uuid, nullable for product-level)
- `default_uom` (string)
- `track_lot` (bool)
- `track_serial` (bool)
- `track_expiration` (bool)
- `default_strategy` (enum: fifo|lifo|fefo)
- `reorder_point` (number)
- `safety_stock` (number)

### 5) InventoryLot
- `sku` (string)
- `catalog_variant_id`
- `lot_number` (string)
- `batch_number` (string, nullable)
- `manufactured_at` (date, nullable)
- `best_before_at` (date, nullable)
- `expires_at` (date, nullable)
- `status` (enum: available|hold|quarantine|expired)

### 6) InventoryBalance
- `warehouse_id`
- `location_id`
- `catalog_variant_id`
- `lot_id` (nullable)
- `serial_number` (string, nullable)
- `quantity_on_hand` (number)
[...]

PR #5339, opened 2026-08-17
feat(wms): share a CrudForm warehouse dialog with country/timezone lists

mkadziolka:
## Summary
- Extract a shared `WarehouseEditDialog` so create and edit use the same CrudForm, including custom fields and optimistic locking.
- Replace free-text country/timezone with ISO country and IANA timezone comboboxes.
- Persist warehouse custom fields in create/update commands and decorate them on the warehouse list API.

Closes #5238

## Test plan
- [ ] Open WMS configuration, create a warehouse, and confirm country/timezone are searchable lists (not text fields).
- [ ] Edit the same warehouse from the row actions and confirm the same form is reused, with the previous country/timezone selected.
- [ ] Add a custom field on `wms:warehouse`, set it in the create dialog, save, reopen edit, and confirm the value round-trips.
- [ ] Confirm `TC-WMS-020` locators still create/edit/archive a warehouse after the combobox change.

Made with [Cursor](https://cursor.com)

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5339"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789553609&installation_model_id=432243&pr_number=5339&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5339&signature=10dc298aa89a5f01b66d78fa3cb12c5644ca5b636320c713e6154fb82d09e504"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #5621, opened 2026-08-25
feat(wms): Phase 2 inbound ASN receiving and putaway

mkadziolka:
## Summary
- Adds WMS Phase 2 inbound ASN receiving with QC/receive flows, ASN list/detail UI, and receiving-line APIs (epic #388).
- Introduces putaway task lifecycle (create/assign/start/complete/cancel), putaway queue UI, and scan endpoints for receive/putaway/location/lot resolve.
- Includes sales reservation re-evaluation after stock increases, backend GUI pages, i18n, and DB migrations (apply with `yarn db:migrate` after pull).

## Test plan
- [ ] Apply WMS migrations (`Migration20260812120000`–`Migration20260822130000`) and confirm schema/snapshot are current
- [ ] Create an ASN, receive lines (including QC paths), and complete the ASN; verify inventory balances and activity labels
- [ ] Create/assign/start/complete putaway tasks from received balances; cancel a task and confirm lock/conflict handling
- [ ] Exercise scan receive/putaway and resolve-location/lot endpoints from the receiving/putaway UIs
- [ ] Confirm sales reservation re-evaluation after ASN receive / putaway-completed stock increase
- [ ] Run WMS Phase 2 integration specs (`TC-WMS-P2-INT-01`–`08`) and related unit tests

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5621"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790256411&installation_model_id=432243&pr_number=5621&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5621&signature=1b2981a4661198e0ea99a88e25594ab2d04bba22ab5c99e146cc623f6769021a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-25T23:19:08Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: feat(wms): Phase 2 inbound ASN receiving and putaway

## 🎯 Summary

This PR implements WMS Phase 2 (epic #388) against the `.ai/specs/2026-04-15-wms-phase-2-inbound-putaway.md` spec: three new entities (`Asn`, `ReceivingLine`, `PutawayTask`), the ASN receive/QC/close commands, the putaway task lifecycle (create/assign/start/complete/cancel), scan-oriented resolve endpoints, ASN list/detail and putaway queue backend pages, sales reservation re-evaluation after a stock increase, and an optional `procurement.goods_receipt.created` bridge. I reviewed all 121 changed files, with close reading of `commands/asn.ts`, `commands/putaway.ts`, the `commands/inventory-actions.ts` refactor, the four migrations plus the ORM snapshot, every new API route, the two reservation subscribers, and the five new integration specs.

The engineering quality here is genuinely high, and it deserves saying before the findings. Every new route exports `metadata` with `requireAuth`/`requireFeatures` and an `openApi` document. Stock-affecting work is serialized under a single transaction holding `PESSIMISTIC_WRITE` on the ASN, receiving line, or putaway task, with `applyInventoryReceiveInTransaction` / `applyInventoryMoveInTransaction` extracted specifically so a nested `commandBus` call cannot escape that lock — and the Phase-1 `wms.inventory.receive` / `wms.inventory.move` commands were refactored to reuse the same helpers rather than forking the ledger. The idempotency design (absolute `targetReceivedQty` rather than a delta, a deterministic movement `referenceId`, a `putaway_key` unique index) is thought through and has dedicated tests. Putaway completion re-checks RBAC server-side through `rbacService` and fails closed when RBAC is unavailable, rather than trusting the route guard alone. The migrations are additive, scoped strictly to `wms_*` tables, match the entity `@Index` declarations byte-for-byte, and ship with the updated `.snapshot-open-mercato.json`. All eleven newly declared event IDs are actually emitted. The five integration specs create their own fixtures via API and clean them up in `finally`. The commands emit `emitCrudSideEffects` with the module's indexer config on create, update, delete **and** the undo paths, which is what makes the routes' `indexer:` declaration meaningful given every action delegates through `commandId`.

The two findings that block are not in the Phase 2 feature itself. One is a silent scope-resolution change to a shared helper that all twenty WMS custom write routes go through, including Phase 1 ones this PR otherwise leaves alone; the other is an unbounded, history-proportional scan in the new reservation re-evaluation path that fires on every receipt.

**Validation source:** local fallback. GitHub reports no usable CI signal for head `ade3136b` — the `CI for Develop&Main` and `Mutation tests` runs both sit at `action_required` because this is a fork PR awaiting maintainer approval, and the only two reported checks are `license/cla` (success) and `[code]smith` (skipped). Under this repo's GitHub-checks-first rule that is exactly the case where the local gate stands in, so I ran the full `validation.commands` list here. **Runner: local** (no compose `app` container running; only `verdaccio` is up).

## Verdict

❌ **request changes** — two Major findings drive this: `executeWmsCustomPostRoute` now silently rewrites `organizationId`/`tenantId` from the session on twenty already-shipped WMS write routes, which can land a stock mutation in the wrong organization without an error for a multi-organization user, and `reevaluateReservationsAfterStockIncrease` scans every sales order line ever recorded for the received variant on every ASN receipt and putaway completion. Several Medium items follow, which under this repository's stricter rule would request changes on their own. Nothing here is a security or cross-tenant defect, and none of it is deep — the Phase 2 surface itself is close to mergeable once the shared-helper scope change is settled.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | 27/27 tasks. |
| `yarn generate` | ✅ PASS | Produced no working-tree drift — the committed generated registries are in sync. |
| `yarn build:packages` (second run) | ✅ PASS | 27/27 tasks, rebuilt against the generator output. |
| `yarn i18n:check-sync` | ✅ PASS | All five locales (en, pl, es, de, ko) in sync across 57 modules. |
| `yarn i18n:check-usage` | ✅ PASS | Advisory unused-key count only; no new violations attributable to this branch. |
| `yarn typecheck` | ✅ PASS | 27/27 tasks. |
| `yarn test` | ⚠️ FAIL — **host environment, not this branch** | 4 of 883 `create-mercato-app` tests fail in `packages/create-app/src/lib/template-dev-log-files.test.ts`; each fails on the test's own preflight assertion `❌ Linux file-watch limits are too low for Turbopack` (`fs.inotify.max_user_instances: 128 < 4096` on this machine). `git diff origin/develop...HEAD -- packages/create-app` is **empty**, so that package is byte-identical to `develop` and the assertion reads `/proc`, not the branch. Everything else passed, including `@open-mercato/core`: **11,729 passed, 2 skipped, 0 failed** across 1,453 suites. |
| `yarn build:app` | ✅ PASS | Next build completed in 2m10s. |

Because the required-checks endpoint returns 404 (branch protection does not enable required status checks) and the fork's workflow runs are held at `action_required`, **there is currently no CI evidence for this head at all**. A maintainer needs to approve the pending workflow runs (`gh api -X POST repos/open-mercato/open-mercato/actions/runs/32871893041/approve` and `.../32871893111/approve`) before merge; my local gate is this review's evidence only and does not substitute for CI.

## Findings

### ⚠️ Major

**1. `packages/core/src/modules/wms/api/inventory/helpers.ts:54-64` — the shared custom-write helper now overwrites client-supplied scope, silently changing behavior on twenty already-shipped routes.**

`executeWmsCustomPostRoute` previously parsed the request body as sent and let each command validate the scope through `ensureOrganizationScope`. It now builds `scopedBody` by forcing `tenantId: auth.tenantId` and `organizationId: ctx.selectedOrganizationId ?? auth.orgId` over whatever the client sent, before `options.inputSchema.parse(...)`.

Forcing the tenant is a clear improvement. Forcing the organization is not equivalent to what it replaces. `ensureOrganizationScope` (`packages/shared/src/lib/commands/scope.ts`) explicitly permits **any** organization in `ctx.organizationScope.allowedIds`, not only the selected one. So for a user with visibility over organizations A and B, with A selected, a request that deliberately targets B used to be accepted and executed against B; it now has `organizationId` rewritten to A and the stock mutation is committed **against the wrong organization, with a 200 response and no warning**. That failure mode — a silently misrouted inventory write — is worse than the 403 the old path would have produced if the target were genuinely disallowed.

This affects far more than Phase 2. All twenty routes that go through this helper inherit it, including the Phase-1 surfaces this PR otherwise deliberately preserves: `inventory/adjust`, `inventory/move`, `inventory/receive`, `inventory/reserve`, `inventory/release`, `inventory/allocate`, `inventory/cycle-count`, and the four `sales-orders/*` assignment routes. Their `openApi` documents still declare `organizationId` and `tenantId` as required body fields, so the published contract now describes values the server discards.

**Fix:** keep validating the client-supplied scope instead of rewriting it — when the body carries `organizationId`, pass it through and let `ensureOrganizationScope` accept or reject it; apply the session value only as a default when the body omits it. If narrowing the contract to session-scope-only is the intended decision, it needs to be stated as such: update the affected `openApi` descriptions, note it in `UPGRADE_NOTES.md`, and add a regression test covering the multi-organization case (user with `allowedIds` = [A, B], selected A, body targeting B) so the chosen behavior is pinned. Either way this change is out of scope for "Phase 2 inbound ASN receiving" and should be called out explicitly in the PR description.

**2. `packages/core/src/modules/wms/lib/salesOrderInventoryAutomation.ts:382-470` — reservation re-evaluation scans the variant's entire order-line history on every receipt.**

`reevaluateReservationsAfterStockIncrease` pages through `E.sales.sales_order_line` filtered **only** by `product_variant_id` (`REEVAL_LINE_PAGE_SIZE = 500`, offset pagination, no status, no date bound), and only then queries `E.sales.sales_order` per page to keep the rows where `isReservableOrderStatus(...)` holds. It is wired to two persistent subscribers, `wms.asn.line_received` and `wms.putaway.completed`, so it fires on every ASN line receipt and every completed putaway task.

The work per event is therefore proportional to the variant's **lifetime** order-line count, forever, and the filter that would shrink the set — order status — is applied after the rows have already been fetched. On a mature tenant with a fast-moving SKU that has 100,000 historical lines, one receiving session of ten lines costs 2,000 paged queries plus 2,000 order lookups, to find the handful of confirmed unfulfilled orders that actually need re-reservation. Pagination bounds memory here, not total work. This is the "unbounded query on growing data" case in the repo's own review checklist.

**Fix:** narrow at the query layer rather than in application code. Start from the small side: query `E.sales.sales_order` filtered on `status = 'confirmed'` with a non-terminal `fulfillment_status` (and optionally a recency window), then intersect with lines for the variant — or, better, drive off the WMS-owned `InventoryReservation` rows for that `catalogVariantId` whose demand is still unmet, which is already an indexed, small, tenant-scoped set and is the actual signal being looked for. Add a test that seeds more than `REEVAL_LINE_PAGE_SIZE` historical lines for one variant and asserts that only the reservable orders are visited.

### 🔹 Minor

**3. `packages/core/src/modules/wms/api/inventory/helpers.ts:54-79` — the organization used for the command and the one given to the mutation guards disagree.**

The body is scoped with `ctx.selectedOrganizationId ?? auth.orgId ?? null`, but three lines later `runRouteMutationGuards` is handed `organizationId: ctx.selectedOrganizationId` without the `auth.orgId` fallback. When `selectedOrganizationId` is null and `auth.orgId` is set — which the fallback exists precisely to handle — the guards evaluate against a null organization while the command writes to `auth.orgId`, so an organization-scoped guard silently sees the wrong (empty) scope. Hoist the resolved `organizationId` constant and pass the same value to both.

**4. `packages/core/src/modules/wms/lib/salesOrderInventoryAutomation.ts:419-421` — a swallowed error with no log in a persistent subscriber.**

The `sales_order_line` page query is wrapped in `catch { return }` with no logging, while its sibling `sales_order` lookup twenty lines below correctly emits `logger.warn('Reservation re-eval order lookup failed; continuing next page', …)`. A transient query-engine failure on the first page therefore drops the entire re-evaluation for that receipt with no trace anywhere — the stock went up and the backordered sales orders were never revisited, and nobody can tell from the logs. Add the matching `logger.warn` with `catalogVariantId` and `page`, so the two failure paths are observable in the same way. Note that `lib/__tests__/reevaluateReservationsAfterStockIncrease.test.ts:153` ("swallows query-engine failures when sales peers are absent") currently pins the silent-swallow behavior, so that case needs to assert the warning rather than its absence.

**5. `packages/core/src/modules/wms/subscribers/procurement-goods-receipt-created.ts:193` — hardcoded English in persisted, user-visible data.**

`notes: \`Created from procurement goods receipt ${goodsReceiptId}\`` writes an English sentence into the ASN's `notes` column, which is rendered on the ASN list and detail pages. In a Polish or Korean tenant the ASN detail page will show an English note. The repo rule is that user-facing strings route through i18n; the subscriber runs server-side, so `resolveTranslations()` is available exactly as it is in the commands' `buildLog`. Add a key such as `wms.asn.notes.fromGoodsReceipt` to the five `i18n/*.json` files and interpolate the id.

### 💅 Nit

**6. `packages/core/src/modules/wms/commands/asn.ts:1466` — a dead conjunct that reads as a concurrency re-check.**

`if (attempt.applyQty > 0.000001 && priorReceivedQty === toNumber(line.receivedQty))` — the second condition is always true: `priorReceivedQty` was assigned from `toNumber(line.receivedQty)` at line 1454 and nothing between the two lines mutates `line`. It looks like a guard against the row changing underneath, but the row is already held under `PESSIMISTIC_WRITE`, so it cannot. Drop the conjunct, or replace it with a comment explaining what invariant the lock provides — as written a later maintainer may believe there is a re-check here when there is not.

**7. `packages/core/src/modules/wms/search.ts:120-133` and `packages/core/src/modules/wms/ce.ts:87,96` — search results and entity labels render raw UUIDs.**

`buildPutawayPresenter` titles a result with `catalog_variant_id`, and `ce.ts` sets `labelField: 'catalogVariantId'` for both `wms:receiving_line` and `wms:putaway_task`. A global-search hit for a putaway task will read as a bare UUID, which is not something an operator can act on. The list pages already resolve variant labels through `formatCatalogVariantLabel` and `attachVendorLabelsToListItems`; reusing that (or indexing the SKU alongside the id, as the lot presenter does) would make these results usable.

**8. `packages/core/src/modules/wms/components/backend/WmsAsnsListPage.tsx:307,328` and `WmsPutawayQueuePage.tsx:574` — arbitrary Tailwind widths on the filter `SelectTrigger`s.**

`className="w-[180px]"` / `"w-[200px]"` use arbitrary values where the DS sizing scale (`w-44`, `w-48`) covers the same widths. `.ai/ds-rules.md` does not name widths explicitly among its banned arbitrary categories, so this is genuinely the author's call — flagging it only because AGENTS.md states the general rule and these are three of only fourteen such occurrences in the whole repository.

**9. `.ai/specs/2026-04-15-wms-phase-2-inbound-putaway.md:5` — the spec still reads `Status | Ready for Review` while its implementation ships.**

Leaving the file in `.ai/specs/` rather than `implemented/` is correct at PR time per `.ai/specs/AGENTS.md` (that move waits for deployment), but the Status line makes it impossible for a reviewer to tell whether the design was signed off before implementation began. Worth advancing to whatever this repo uses for an approved-and-being-implemented spec.

### ℹ️ Informational (not a finding)

`packages/ui/src/backend/icons/lucideRegistry.generated.tsx` gains a `bot` entry. That is unrelated to WMS — it catches the generated registry up with `packages/channel-discord/.../message-types.ts:57` (`icon: 'bot'`), which landed on `develop` in f75c35b3d. It is generator output rather than a hand edit (`yarn generate` produced no drift on this branch), so it is correct to include; it is just worth a line in the PR description so a reviewer does not have to work that out.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the refactor of `inventory-actions.ts` **adds** `applyInventoryReceiveInTransaction`, `applyInventoryMoveInTransaction`, `emitInventoryReceiveSideEffects` and their result types; `wms.inventory.receive` and `wms.inventory.move` keep their command ids, inputs and return shapes.
- [x] No required type field removed or narrowed — `InventoryReceiveInput` gains an optional `reasonCode`, and the operational-dashboard response schema gains optional `referenceLabel` / `reasonCode` / `source`.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — all fourteen routes are new.
- [x] No field removed or retyped in an existing response shape — the sales-order `_wms` enrichment is extended additively with `reservationSummary.reservations` and `inboundSummary`.
- [x] No event or message name renamed or removed; no payload field removed — eleven event ids added to `events.ts`, all of them emitted, none altered.
- [x] No database table or column renamed or removed; no column type narrowed — `Migration20260812120000`/`130000` create `wms_asns`, `wms_receiving_lines`, `wms_putaway_tasks`; `Migration20260822120000`/`130000` add the nullable `putaway_key` and `source_key` columns with partial unique indexes. Every statement is additive and confined to `wms_*`; the SQL matches the entity `@Index` expressions exactly, and `.snapshot-open-mercato.json` is updated in the same change. No unrelated schema churn — the migration sanity gate passes.
- [x] No config key renamed and no default changed silently — `wms_integration_procurement_goods_receipt` keeps its `false` default.
- [ ] **Where a contract had to change: old surface kept working through a deprecation window, with migration notes** — this is the one that fails, and it is Major finding 1. `executeWmsCustomPostRoute` narrows the effective request contract of twenty shipped WMS write routes by discarding a body field their published `openApi` still declares as required, with no note in `UPGRADE_NOTES.md` and no test pinning the new behavior.

The new ACL features `wms.manage_asn` and `wms.manage_putaway` are additive, are declared in `acl.ts`, flow into `setup.ts` through `WMS_MANAGE_FEATURES` in `lib/roleFeatures.ts`, and `__tests__/setup.test.ts` was updated to assert them — that wiring is complete. Existing tenants will still need `yarn mercato auth sync-role-acls` after deploy to receive the new grants; worth stating in the PR description or release notes.

## 🧪 Test Coverage

Coverage is strong and specific, not decorative. On the unit side: `commands/__tests__/inventory-actions.receive-idempotency.test.ts` and `.adjust-idempotency.test.ts` pin the reordering of movement-persist before balance-bump (the actual behavior change in that file, so a revert would go red); `putaway-lifecycle-lock.test.ts` (594 lines) covers the lock and status transitions; `putaway-complete-emit.test.ts` covers event emission; `asn-delete-undo.test.ts` (2,365 lines) covers the delete/undo paths; `lib/__tests__/asnReceiving.test.ts` and `putaway.test.ts` cover the pure helpers including the `isAsnCloseable` edge cases the code comments call out; `asnCompleteOptimisticLock.test.ts` and `putawayQueueOptimisticLock.test.ts` cover the client lock headers; `ReceiveAsnLineDialog.test.tsx` and `CompletePutawayTaskDialog.test.tsx` cover the dialogs. On the integration side, `TC-WMS-P2-INT-01-04` through `-08` cover ASN create with vendor and lines, QC pass and fail, over-receipt, ACL denial, putaway auto-creation and staging→bin completion, the scan resolve/receive endpoints, sales reservation re-evaluation, and the receiving/putaway UI end to end. All five build their own fixtures through the API and tear them down in `finally`, so they satisfy the self-containment requirement in `.ai/qa/AGENTS.md`. All four new dialogs implement `Cmd/Ctrl+Enter` submit alongside Radix's native `Escape`, matching the keyboard rule.

The gaps map onto the findings:

1. **`packages/core/src/modules/wms/api/inventory/__tests__/`** — no test covers the new scope-rewriting behavior of `executeWmsCustomPostRoute`. Add cases for: a body `organizationId` matching an allowed-but-not-selected organization (asserting whichever behavior is decided in finding 1 — pass-through or explicit rejection, but not silent rewrite); a body omitting `organizationId` (session default applies); and `selectedOrganizationId === null` with `auth.orgId` set, asserting the command payload and the mutation-guard input agree, which is finding 3.
2. **`packages/core/src/modules/wms/lib/__tests__/reevaluateReservationsAfterStockIncrease.test.ts`** — the file exists and covers the status filter and the paging loop, but not scale. Add a case seeding more than `REEVAL_LINE_PAGE_SIZE` historical lines for one variant, the vast majority on fulfilled or cancelled orders, asserting the number of queries issued stays proportional to the reservable orders rather than to total history. That test is what would keep finding 2 from regressing.
3. **Failure-path observability** — add a case asserting the `sales_order_line` query failure emits a warning (finding 4), mirroring the existing assertion for the order-lookup failure.
4. **CI** — none of this has run in CI yet. The four inotify failures on my machine are a host limit in a package this PR does not touch, but that is a local judgement; the pending fork workflow runs need maintainer approval before the branch has any real CI signal.

pkarw:
🏷️ label rationale

Re-review on head `301ddc47` (round 2). Every label re-derived; **none changed** — the set that was right at `ade3136b` is still right.

| Label | Why |
|-------|-----|
| `changes-requested` | Verdict is request-changes again: one blocker (the head is `CONFLICTING` against `develop` on `packages/core/src/modules/wms/commands/shared.ts`), the two Major findings from round 1 both still open and untouched, and six Minors. Pipeline state unchanged; no other pipeline label set. |
| `feature` | Implements WMS Phase 2 (epic #388) against `.ai/specs/2026-04-15-wms-phase-2-inbound-putaway.md` — three new entities, the ASN receive/QC/close commands, the putaway lifecycle, scan endpoints and two backend pages. |
| `needs-qa` | The automated-verification exemption fails on all three clauses, not just one: the diff adds `.tsx` under `packages/core/src/modules/wms/components/backend/`, it changes the database structure (four `wms_*` migrations), and it widens the API surface (fourteen new routes). Kept. |
| `priority-medium` | A large planned feature rather than a defect or an outage; nothing here is blocking another team today. Unchanged from the author's own assessment. |
| `risk-high` | 126 files, four migrations, new ACL features, and a change to a shared helper (`executeWmsCustomPostRoute`) that twenty already-shipped WMS write routes go through — plus, since round 2, a behavior change to the shipped `wms.inventory.reserve` command. Blast radius reaches outside the Phase 2 surface. |

Not applied, deliberately: `qa-approved` (earned by manual QA or the self-QA exception, not by reading a diff) and the `qa` pipeline label (means manual QA is *in progress* and belongs to a QA reviewer — this skill only requests QA with `needs-qa`).

No `ci-monitoring`: this is a fork PR and GitHub has **zero** workflow runs for `301ddc47`, so there is nothing for an agent to watch. A maintainer has to approve the runs first.

pkarw:
🤖 `om-auto-review-pr` — handing this PR back to @mkadziolka.

The review is posted with a **changes-requested** verdict. Two Major findings need a decision or a fix before this can be approved:

1. **`packages/core/src/modules/wms/api/inventory/helpers.ts:54-64`** — `executeWmsCustomPostRoute` now overwrites `organizationId` from the session, which silently narrows the request contract of all twenty WMS custom write routes (Phase 1 included) and can commit a stock mutation to the wrong organization for a multi-organization user, where the previous path would have gone through `ensureOrganizationScope` and either accepted the target or returned 403.
2. **`packages/core/src/modules/wms/lib/salesOrderInventoryAutomation.ts:382-470`** — `reevaluateReservationsAfterStockIncrease` pages through the received variant's entire order-line history on every ASN receipt and putaway completion, filtering by order status only after the rows are fetched.

Three Medium items follow (guard/command organization mismatch, an unlogged swallowed error in a persistent subscriber, a hardcoded English ASN note), plus four nits. Full detail, per-finding fixes, and the suggested test cases are in the review body.

**Two things that are not about your code:**

- **CI has not run on this head.** Both workflow runs for `ade3136b` sit at `action_required` because this is a fork PR — a maintainer needs to approve them. Until then the PR has no CI signal at all, and the checks list showing only `license/cla` is misleading.
- **The local validation gate is otherwise green.** `yarn build:packages` (×2), `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck` and `yarn build:app` all passed, and `@open-mercato/core` ran 11,729 tests with zero failures. `yarn test` exited 1 only because of four `create-mercato-app` tests that fail on this machine's `fs.inotify.max_user_instances` limit, in a package your branch does not touch. Nothing there is attributable to your work.

The Phase 2 surface itself is in good shape — the transactional locking, the absolute-target idempotency design, the migration hygiene, and the integration coverage all read as careful work. `autofix: skipped (not my PR — re-run with --autofix to fix it here)`.

pkarw:

Reviewed all 121 changed files in an isolated worktree at head `ade3136b`. Verdict driven by 2 Major and 3 Medium findings; validation source was the **local fallback** gate (GitHub reports no usable checks for this head — both fork workflow runs are held at `action_required`). Local gate: 7 of 8 commands pass; `yarn test` exit 1 is 4 `create-mercato-app` tests failing on this host's inotify sysctl limits, in a package this branch does not touch. `autofix: skipped (not my PR — re-run with --autofix to fix it here)`. No CI follow-up will come from this agent — the pending runs need maintainer approval first.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T00:22:42Z. Re-review of head `301ddc47` (previous review was on `ade3136b`). Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Re-review: feat(wms): Phase 2 inbound ASN receiving and putaway

## 🎯 Summary

Round 2, on head `301ddc47` (my first review was on `ade3136b`). Three commits landed since: `a412dcb7` *stabilize Phase 2 ASN/putaway CI failures*, `9ecb804c` *map putaway aging column sort to createdAt*, and `301ddc47` *keep staging stock out of reservations and sync i18n*. I read the whole `ade3136b..301ddc47` range (23 files, +497/−64) line by line and re-checked all nine findings from round 1 against the current head.

**The nine round-1 findings are all still open.** `api/inventory/helpers.ts`, `subscribers/procurement-goods-receipt-created.ts`, `search.ts`, `ce.ts` and `WmsAsnsListPage.tsx` are byte-identical to `ade3136b`, and the `reevaluateReservationsAfterStockIncrease` paging loop is unchanged apart from one import. Both Majors therefore carry forward verbatim, and I have restated them in full below rather than making you dig them out of the previous review.

**The new work is good and mostly right, and that deserves saying first.** The staging/dock exclusion in `301ddc47` closes a genuine hole I did not catch in round 1: before it, the `wms.asn.line_received` subscriber would happily reserve the stock that had just landed on staging, which would then block the putaway task the same receipt had created. Putting the rule in one exported predicate (`lib/inventoryPolicy.ts`) and applying it at both the reserve-candidate and warehouse-availability layers is the right shape, and it comes with real unit tests at both layers plus a rewritten INT-07. The `wms.putaway-tasks.delete` undo fix is correct: soft-deleted rows really are excluded by `withLockedPutawayTask`'s explicit `deletedAt: null` (`putaway.ts:209`), there is no global soft-delete filter on this entity, so locking the deleted row directly is the only way to do this under `PESSIMISTIC_WRITE` — and re-asserting `cancelled` before undelete matches the guard the update-undo already had. The INT-08 rework (`role="option"` instead of `role="button"`, waiting on the `Loading suggestions…` status node instead of `waitForTimeout(350)`, scoping the row action to the row that contains the staging code) fixes real strict-mode and hydration-race classes rather than papering over them.

What blocks now is a merge conflict, the two carried Majors, and six smaller items — four of which are the tail of the staging change that the same commit did not follow all the way through.

**Validation source: local fallback.** GitHub has **no CI data at all** for `301ddc47` — `actions/runs?head_sha=301ddc47…` returns an empty list, and the only two reported checks are `license/cla` (pass) and `[code]smith` (skipping). This is a fork PR, so workflow runs need maintainer approval per push, and required status checks are not enabled on `develop` (`branches/develop/protection/required_status_checks` → 404). Under this repo's GitHub-checks-first rule that is exactly the case where the local gate stands in, so I ran the full `validation.commands` list here. **Runner: local.**

## Verdict

❌ **request changes** — one blocker (the head does not merge), the two unaddressed Majors from round 1, and six Minors. Under this repository's stricter rule the Minors alone would also request changes.

To be clear about the shape of what is left: **none of it is deep, and none of it is in the Phase 2 receiving/putaway core.** The conflict is two additive blocks in one file. Both Majors are about a shared helper and a query that predate Phase 2's own surface. Four of the six Minors are follow-through on a change made three commits ago.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | 27/27 tasks; `@open-mercato/core` was a genuine cache miss and rebuilt. |
| `yarn generate` | ✅ PASS | **No working-tree drift** — `git status --porcelain` empty afterwards. The `bot` lucide entry I flagged as informational in round 1 is now committed on the branch, so the generated registries are fully in sync. |
| `yarn build:packages` (second run) | ✅ PASS | 27/27. |
| `yarn i18n:check-sync` | ✅ PASS | All five locales in sync; the two new `wms.backend.putaway.section.*` keys sort correctly in `en`/`pl`/`es`/`de`/`ko`. |
| `yarn i18n:check-usage` | ✅ PASS | The heading rename does **not** orphan `wms.backend.putaway.title` / `.description` — they are still used by the page header at `WmsPutawayQueuePage.tsx:522-525`. |
| `yarn typecheck` | ✅ PASS | 27/27. |
| `yarn test` | ✅ PASS — **after ruling out two host failures** | See below. |
| `yarn build:app` | ✅ PASS | Next build completed. |

`yarn test` went red twice at the workspace level, on **two different packages, neither touched by this PR**, and both failures were this machine rather than the branch:

1. `@open-mercato/ui` › `src/primitives/__tests__/no-deprecated-notice.test.ts` — `ENOENT … packages/shared/.tmp-dynamic-loader-POU32V/src/modules/probe/data/entities.ts`. That guard walks the repo with `fs.readFileSync`, and `@open-mercato/shared`'s `dynamicLoader` tests, running concurrently under turbo in the same worktree, created and deleted that temp directory mid-walk.
2. `@open-mercato/queue` › `abandoned-job.test.ts` and `local.strategy.test.ts` — `jest: failed to cache transform results in /tmp/jest_rs/… Failure message: UNKNOWN: unknown error, write`. `/tmp` here is a 16 GB tmpfs sitting at 80 % full.

After `rm -rf /tmp/jest_rs`, re-run per package: **`@open-mercato/core` 11,736 passed / 2 skipped / 0 failed across 1,453 suites**, `@open-mercato/ui` 1,929 passed across 233 suites (including the `no-deprecated-notice` guard that had failed), `@open-mercato/queue` 104 passed. Nothing on this branch fails.

Because there is no CI signal for this head at all, a maintainer needs to approve the fork's workflow runs before merge. My local gate is this review's evidence only and does not substitute for CI — and note that the branch is **59 commits behind `develop`**, which per `ci.yml`'s two-dot diff means the eventual run will pull the full 15-shard integration suite, not the affected-only mode.

## Findings

### 🚫 Blocker

**1. The head does not merge — `packages/core/src/modules/wms/commands/shared.ts` conflicts with `develop`.**

`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. `git merge-tree` against `origin/develop` reports exactly one conflicted path: `packages/core/src/modules/wms/commands/shared.ts`. The conflict itself is benign — both sides append to the same region. `develop` added `warehouseCrudIndexer` (from `c72025345`, *"wms: invalidate query_index and CRUD cache after inventory mutations"*), this branch added `asnCrudIndexer` / `receivingLineCrudIndexer` / `putawayTaskCrudIndexer` and their three `CrudEventsConfig` siblings. Keeping both blocks resolves it.

I am still calling it a blocker, because the drift behind it is not purely mechanical. The branch is 59 commits behind, and four of those touch `packages/core/src/modules/wms/` — most relevantly `7ba2771d0` *feat(wms): share a CrudForm warehouse dialog with country/timezone lists (#5339)* and `b53e67e5c` *fix(ui): align WMS KPI cards…*. A merge that reports no conflict can still land a file importing the same symbol twice, or two versions of the same dialog. Please re-run `yarn typecheck` and the WMS unit suites after merging `develop`, not just after clearing the conflict markers.

### ⚠️ Major

**2. `packages/core/src/modules/wms/api/inventory/helpers.ts:54-64` — the shared custom-write helper still overwrites client-supplied scope, silently changing behavior on twenty already-shipped routes.** *(carried from round 1, unchanged)*

`executeWmsCustomPostRoute` builds `scopedBody` by forcing `tenantId: auth.tenantId` and `organizationId: ctx.selectedOrganizationId ?? auth.orgId` over whatever the client sent, before `options.inputSchema.parse(...)`.

Forcing the tenant is a clear improvement. Forcing the organization is not equivalent to what it replaces. `ensureOrganizationScope` (`packages/shared/src/lib/commands/scope.ts`) explicitly permits **any** organization in `ctx.organizationScope.allowedIds`, not only the selected one. For a user with visibility over organizations A and B, with A selected, a request that deliberately targets B used to be accepted and executed against B; it now has `organizationId` rewritten to A and the stock mutation is committed **against the wrong organization, with a 200 response and no warning**. A silently misrouted inventory write is a worse failure mode than the 403 the old path would have produced if the target were genuinely disallowed.

This reaches far beyond Phase 2 — all twenty routes through this helper inherit it, including the Phase-1 surfaces this PR otherwise deliberately preserves (`inventory/adjust`, `move`, `receive`, `reserve`, `release`, `allocate`, `cycle-count`, and the four `sales-orders/*` assignment routes). Their `openApi` documents still declare `organizationId` and `tenantId` as required body fields, so the published contract now describes values the server discards.

**Fix:** when the body carries `organizationId`, pass it through and let `ensureOrganizationScope` accept or reject it; apply the session value only as a default when the body omits it. If narrowing the contract to session-scope-only is the intended decision, say so explicitly: update the affected `openApi` descriptions, note it in `UPGRADE_NOTES.md`, and add a regression test for the multi-organization case (`allowedIds = [A, B]`, selected A, body targeting B). Either way it is out of scope for "Phase 2 inbound ASN receiving" and belongs in the PR description.

**3. `packages/core/src/modules/wms/lib/salesOrderInventoryAutomation.ts:385-470` — reservation re-evaluation still scans the variant's entire order-line history on every receipt.** *(carried from round 1, unchanged)*

`reevaluateReservationsAfterStockIncrease` pages through `E.sales.sales_order_line` filtered **only** by `product_variant_id` (`REEVAL_LINE_PAGE_SIZE = 500`, offset pagination, no status, no date bound), and only then queries `E.sales.sales_order` per page to keep the rows where `isReservableOrderStatus(...)` holds. It is wired to two persistent subscribers, `wms.asn.line_received` and `wms.putaway.completed`, so it fires on every ASN line receipt and every completed putaway task.

Work per event is proportional to the variant's **lifetime** order-line count, forever, and the filter that would shrink the set — order status — is applied after the rows are already fetched. On a mature tenant with a fast-moving SKU at 100,000 historical lines, one receiving session of ten lines costs 2,000 paged queries plus 2,000 order lookups to find the handful of confirmed unfulfilled orders that actually need re-reservation. Pagination bounds memory here, not total work. This is the "unbounded query on growing data" case in `.ai/review-checklist.md`.

Worth adding that the staging change made this *worse in frequency terms*, not better: the re-eval now effectively runs twice per unit of received stock — once on `wms.asn.line_received`, where it is now guaranteed by design to find nothing reservable, and again on `wms.putaway.completed`. The first of those two full-history scans is now pure waste.

**Fix:** narrow at the query layer. Start from the small side — query `E.sales.sales_order` filtered on `status = 'confirmed'` with a non-terminal `fulfillment_status`, then intersect with lines for the variant — or better, drive off the WMS-owned `InventoryReservation` rows for that `catalogVariantId` whose demand is still unmet, which is already an indexed, small, tenant-scoped set and is the actual signal being looked for. Add a test that seeds more than `REEVAL_LINE_PAGE_SIZE` historical lines for one variant and asserts only the reservable orders are visited.

### 🔹 Minor

**4. `packages/core/src/modules/wms/data/enrichers.ts:274-296` and `:629-637` — the staging exclusion was applied to one of two near-identical availability aggregators, so the sales-order stock widget now contradicts itself.**

This is the highest-value item of the new round, and it is a two-line fix using a helper this PR already wrote and exported.

`301ddc47` added the staging/dock filter to `lib/salesOrderInventoryAutomation.ts:241` (`buildWarehouseAvailability`) and correctly added `{ populate: ['location'] }` to that file's `loadBalances` at `:203` so the filter can actually see the type. `data/enrichers.ts` has a structurally identical `loadBalances` at `:438` — same entity, same where clause, same scope — feeding `buildBalanceSummaryByVariant` (`:274`) and the inline aggregation at `:629`. Neither got the filter, and neither populates `location`.

The result is visible on one screen. `widgets/injection/sales-order-stock-context/widget.client.tsx` renders `stockSummary.available` at `:414-418` next to `reservationSummary.status` at `:306`. With ten units sitting on staging after an ASN receipt, that widget now reads **"Available 10"** beside **"unreserved"** — while the automation has already emitted `wms.inventory.reservation_shortfall` for the same variant, which the branch's own new unit test (`salesOrderInventoryAutomation.test.ts:122`, *"does not treat staging-only stock as reservable warehouse availability"*) asserts happens. `widgets/injection/order-items-stock-column/widget.tsx:49` reads the same field. A sales rep looking at that widget to decide whether to promise stock is being shown a number the system will refuse to act on.

**Fix:** add `{ populate: ['location'] }` to `enrichers.ts:438` and the `isBalanceLocationReservable` guard to both aggregation loops — or, if the intent is that the widget should show *physical* on-hand including inbound, split it explicitly into `available` and something like `availableInbound` so the two numbers are labelled rather than merely inconsistent. Either way the choice should be deliberate and covered by a test.

**5. `packages/core/src/modules/wms/components/backend/inboundStatusUi.ts:67-69` and `WmsPutawayQueuePage.tsx:131-132` — the aging sort now runs in the wrong direction.**

`mapPutawayQueueSortField` translates the `aging` column id to the API's `createdAt`, which is right, but `sortDir` is passed through unchanged at `:132`. Aging is *elapsed time since* `created_at` (`formatAgingLabel`, `inboundStatusUi.ts:54-63`), so it is inversely ordered with respect to `createdAt`: sorting `createdAt` **descending** puts the newest rows first, which are the rows with the **smallest** aging value.

Concretely, on a queue holding tasks aged `3d`, `2d` and `4h`, clicking the "Aging" header until it shows a descending arrow displays `4h, 2d, 3d`. For a putaway queue the entire point of an aging column is to surface the oldest work, so this is exactly backwards from what an operator clicking it wants. The default (`sorting` initial state `[{ id: 'createdAt', desc: false }]`, `:102`) is oldest-first and correct — it is only the explicit click that inverts.

**Fix:** invert the direction when the mapping fires, e.g. have the mapper return `{ field, invert }` and apply `invert ? (desc ? 'asc' : 'desc') : …` at `:132`. The new test at `inboundStatusUi.test.ts:44-50` pins the field mapping but says nothing about direction, so please add the direction case alongside it — as written, this bug is invisible to the suite.

**6. `packages/core/src/modules/wms/commands/inventory-actions.ts:491-500` — a behavior change to the shipped `wms.inventory.reserve` command, with no `UPGRADE_NOTES.md` entry and no opt-out.**

I think the rule is right and I am not arguing against it. But `wms.inventory.reserve` is a Phase-1 command behind a shipped public route (`POST /api/wms/inventory/reserve`), and `listCandidateBalances` now silently drops any balance whose location type is `staging` or `dock`. A tenant that modelled its stock on a location typed `staging` — perfectly possible in Phase 1, where the type enum existed but nothing consumed it and there was no putaway flow to make it meaningful — will see every reservation start failing with `409 insufficient_stock` after upgrading, with nothing in the release notes to explain it. `WarehouseLocation` has no `pickable`/`isPickable` flag (`data/entities.ts:138-167`), so `type` genuinely is the only signal available, and there is no config escape hatch.

The PR touches neither `UPGRADE_NOTES.md` nor `CHANGELOG.md` — `git diff origin/develop...HEAD -- UPGRADE_NOTES.md CHANGELOG.md` is empty, and `UPGRADE_NOTES.md` contains no mention of `wms` or `staging`.

**Fix:** an `UPGRADE_NOTES.md` entry naming the command, the new `insufficient_stock` outcome, and the remediation (retype the affected locations to `bin`/`slot`, or move the stock). One paragraph is enough; the point is that an operator hitting this has something to search for. Worth checking at the same time whether `wms.inventory.reserve`'s `openApi` `errors` list needs the newly reachable case.

**7. `packages/core/src/modules/wms/api/inventory/helpers.ts:73` and `:111` — the organization used for the command and the ones given to the guards and the after-interceptors disagree.** *(carried from round 1; the second site is new detail)*

The body is scoped with the resolved `organizationId` (`:54`, `ctx.selectedOrganizationId ?? auth.orgId ?? null`), but `runRouteMutationGuards` is handed `organizationId: ctx.selectedOrganizationId` at `:73` **without** the `auth.orgId` fallback, and `runCustomRouteAfterInterceptors`' context does the same at `:111`. When `selectedOrganizationId` is null and `auth.orgId` is set — the exact case the fallback exists for — both the guards and the after-interceptors evaluate against an empty organization scope while the command writes to `auth.orgId`.

This one got *easier* to fix since round 1: the resolved value is already a named constant on line 54. Passing `organizationId` instead of `ctx.selectedOrganizationId` at both `:73` and `:111` is a two-word change.

**8. `packages/core/src/modules/wms/lib/salesOrderInventoryAutomation.ts:425-427` — a swallowed error with no log in a persistent subscriber.** *(carried from round 1, unchanged)*

The `sales_order_line` page query is wrapped in `catch { return }` with no logging, while its sibling `sales_order` lookup twenty lines below correctly emits `logger.warn('Reservation re-eval order lookup failed; continuing next page', …)`. A transient query-engine failure on the first page drops the entire re-evaluation for that receipt with no trace: the stock went up, the backordered sales orders were never revisited, and nothing in the logs says so. Add the matching `logger.warn` with `catalogVariantId` and `page`. Note that `lib/__tests__/reevaluateReservationsAfterStockIncrease.test.ts:153` (*"swallows query-engine failures when sales peers are absent"*) currently pins the silent-swallow behavior, so that case needs to assert the warning rather than its absence.

**9. `packages/core/src/modules/wms/subscribers/procurement-goods-receipt-created.ts:209` — hardcoded English in persisted, user-visible data.** *(carried from round 1, unchanged)*

``notes: `Created from procurement goods receipt ${goodsReceiptId}` `` writes an English sentence into the ASN's `notes` column, which is rendered on the ASN list and detail pages, so a Polish or Korean tenant sees English there. The subscriber runs server-side, so `resolveTranslations()` is available exactly as it is in the commands' `buildLog`. Add e.g. `wms.asn.notes.fromGoodsReceipt` to the five `i18n/*.json` files and interpolate the id.

### 💅 Nit

**10. `packages/core/src/modules/wms/__integration__/TC-WMS-P2-INT-07-sales-reeval.spec.ts:220-234` — the "staging alone does not reserve" assertion is vacuous.**

The rewritten INT-07 is a genuine improvement: driving the reservation through `putaway-tasks/{id}/complete` into a real `bin` is exactly the flow the staging change creates, and asserting `receiveResult.putawayTaskIds?.length === 1` first is a good tightening. But the new negative half —

```ts
await expect.poll(async () => { … return reservations.length }, { timeout: 5_000, … }).toBe(0)
```

— passes on its **first** poll, immediately after the receive call returns and well before the async `wms.asn.line_received` subscriber has had a chance to do anything. `expect.poll` retries until the assertion *passes*, so a condition that is already true short-circuits; it would report green even if the re-eval created a reservation a second later. The behavior this commit exists to introduce is therefore not actually pinned at the integration level. It *is* pinned at the unit level by the two new cases in `inventory-actions.reserve-release.test.ts:284`/`:325` and `salesOrderInventoryAutomation.test.ts:122`, which is why this is a nit rather than a Minor.

**Fix:** wait for a positive signal that the subscriber has run — poll for the putaway task, or for the shortfall outcome — and only then assert `reservations.length === 0` once, without polling.

**11. `packages/core/src/modules/wms/lib/putaway.ts:13` and `lib/inventoryPolicy.ts:13-16` — the same domain rule is now encoded twice.**

`FORBIDDEN_PUTAWAY_TARGET_TYPES = new Set(['staging', 'dock'])` and the new `NON_RESERVABLE_LOCATION_TYPES = new Set(['staging', 'dock'])` are the same statement — "staging and dock are inbound holding locations" — in two files, and they are load-bearing for each other: the reserve filter is only safe because putaway completion refuses to land stock back on one of these types. If a third holding type is ever added, updating one and not the other silently strands inventory. Worth collapsing to a single exported set that both import.

**12. `packages/core/src/modules/wms/components/backend/inventoryMutationLoaders.ts:395-406` — the destination picker offers only `bin`, while the spec's own new sentence says storage is "bin/slot/…".**

`loadBinLocationOptions` queries `{ type: 'bin' }` only, and `CompletePutawayTaskDialog.tsx:276-288` uses it with `allowCustomValues={false}`. The command layer is more permissive — `assertPutawayTargetLocationType` (`putaway.ts:969`) rejects only `staging`/`dock` — and `isReservableLocationType` treats `slot`, `rack`, `aisle` and `zone` as reservable too. So a warehouse modelled with `slot` locations can complete putaway through the API but not through the UI. Aligning the loader with `assertPutawayTargetLocationType` (any type except staging/dock) would remove the discrepancy. The helper is shared with `CycleCountWizardDialog` and predates this PR — but this PR is what makes it reachable in a putaway context.

**13. `packages/core/src/modules/wms/commands/asn.ts:1466` — a dead conjunct that reads as a concurrency re-check.** *(carried, unchanged)*

`if (attempt.applyQty > 0.000001 && priorReceivedQty === toNumber(line.receivedQty))` — the second condition is always true: `priorReceivedQty` was assigned from `toNumber(line.receivedQty)` twelve lines earlier and nothing between them mutates `line`. It reads like a guard against the row changing underneath, but the row is held under `PESSIMISTIC_WRITE` so it cannot. Drop the conjunct, or replace it with a comment naming the invariant the lock provides.

**14. `packages/core/src/modules/wms/search.ts:120-133` and `ce.ts:87,96` — search results and entity labels render raw UUIDs.** *(carried, unchanged)*

`buildPutawayPresenter` (`search.ts:120`) titles a result with `catalog_variant_id` (`:126`), and `ce.ts` sets `labelField: 'catalogVariantId'` for both `wms:receiving_line` and `wms:putaway_task`, so a global-search hit for a putaway task reads as a bare UUID. The list pages already resolve variant labels through `formatCatalogVariantLabel`; reusing that, or indexing the SKU alongside the id as the lot presenter does, would make these results actionable.

**15. `WmsAsnsListPage.tsx:307,328` and `WmsPutawayQueuePage.tsx:575` — arbitrary Tailwind widths on the filter `SelectTrigger`s.** *(carried, unchanged)*

`w-[180px]` / `w-[200px]` where the DS scale (`w-44`, `w-48`) covers the same widths. `.ai/ds-rules.md` does not name widths among its banned arbitrary categories, so this is genuinely your call.

**16. `.ai/specs/2026-04-15-wms-phase-2-inbound-putaway.md:5` — the Status line is two revisions stale.**

It still reads `Ready for Review (rev 43 — … 2026-08-25)` while the Changelog below has advanced to rev 45. Leaving the file in `.ai/specs/` rather than `implemented/` remains correct at PR time per `.ai/specs/AGENTS.md`; it is only the rev in the header that has fallen behind. The rest of the spec update in `301ddc47` is exactly right — writing the staging/dock reservation rule into § Business Rules item 6 as *design* and updating INT-07's stated expectation is the kind of thing that usually gets skipped.

### ℹ️ Checked and **not** findings

Recording these so nobody re-derives them:

- **The `putaway_key` unique index really does carry `status <> 'cancelled'`.** The spec edit at `:206` looked like documentation drifting ahead of the code, but `data/entities.ts:544`, `migrations/Migration20260822120000.ts:9` and `migrations/.snapshot-open-mercato.json:3830` all carry the identical predicate. The spec was catching up, not diverging.
- **Putaway completion cannot strand stock on another staging location.** My first read of `requireActiveLocation` (`putaway.ts:240-265`) showed only warehouse and `isActive` checks, which — now that staging is non-reservable — would have meant a staging→staging "putaway" produced permanently invisible stock with no open task and no alert. `assertPutawayTargetLocationType(targetLocation.type)` at `putaway.ts:969` rejects `staging`/`dock` at the command layer, so the invariant the reserve filter depends on is actually enforced. ASN receive validates its staging target symmetrically at `asn.ts:326-327`, so stock can only enter the non-reservable state through the intended door.
- **The staging filter is live, not dead code.** `isBalanceLocationReservable` returns `true` for an unpopulated (string) location FK, which would have made either call site a silent no-op. `listCandidateBalances` already passed `populate: ['lot', 'location']` (`inventory-actions.ts:485`), and `301ddc47` added `populate: ['location']` to the automation's `loadBalances` in the same commit. Both are populated. `listCandidateBalances` has exactly one caller (`wms.inventory.reserve`, `inventory-actions.ts:949`), so the change does not leak into move, release, allocate, adjust or cycle-count.
- **The non-sortable putaway columns are genuinely non-sortable.** `variant`, `source`, `target` and `assignee` carry no `enableSorting: false`, which looked like it would let a click send an unmapped `sortField` (`variant`) straight through `factory.ts:1729`'s `|| field` fallback into the query engine. TanStack's `getCanSort()` also requires `!!column.accessorFn`, and those four are `id` + `cell` only. The three that *are* sortable — `status`, `quantity`, `aging` — are exactly the three with an accessor, and all three resolve through `sortFieldMap` (`api/putaway-tasks/route.ts:87`) once finding 5 is fixed.
- **The heading rename does not orphan i18n keys.** `wms.backend.putaway.title` / `.description` are still used at `WmsPutawayQueuePage.tsx:522-525`; only the embedded `DataTable` section heading moved to the new `.section.*` keys.
- **The delete-undo lock fix is correct.** There is no MikroORM `@Filter` for soft delete on these entities — `withLockedPutawayTask` filters `deletedAt: null` explicitly in its where clause (`putaway.ts:209`) — so the new undo really does need to find the row without that clause, and doing it under `PESSIMISTIC_WRITE` inside `em.transactional` matches every other write path in the file. The two new tests at `putaway-lifecycle-lock.test.ts:269` and `:321` assert the lock mode and the `409` on status drift, and that `409` matches what the update-undo already threw.

## 💥 Breaking Changes

Unchanged from round 1 except where the new commits moved things:

- [x] No exported/public symbol removed or renamed without a deprecation path — `301ddc47` **adds** `NON_RESERVABLE_LOCATION_TYPES`, `isReservableLocationType`, `isBalanceLocationReservable` to `lib/inventoryPolicy.ts` and `mapPutawayQueueSortField` to `inboundStatusUi.ts`.
- [x] No database table or column renamed or removed; no column type narrowed — the four migrations are unchanged since round 1, additive, confined to `wms_*`, and matched by `.snapshot-open-mercato.json`.
- [ ] **Where a contract had to change: old surface kept working through a deprecation window, with migration notes** — this still fails, and now on **two** counts: Major 2 (`executeWmsCustomPostRoute` discards a body field that twenty routes' `openApi` documents still declare required) and Minor 6 (`wms.inventory.reserve` gained a new refusal case with no `UPGRADE_NOTES.md` entry). Neither is necessarily wrong as a decision; both need to be written down.

Still true and still worth stating in the release notes: existing tenants need `yarn mercato auth sync-role-acls` after deploy to receive the new `wms.manage_asn` / `wms.manage_putaway` grants.

## 🧪 Test Coverage

The new round adds coverage where it matters, and I want to credit that specifically. `inventory-actions.reserve-release.test.ts:284` drives the real `wms.inventory.reserve` handler and asserts staging+dock alone yields `409 insufficient_stock`; `:325` then asserts a bin peer *is* still reserved from while the staging peer stays at zero — that second case is what keeps the filter from being over-broad, and it is the one most people would have left out. `salesOrderInventoryAutomation.test.ts:122` pins the shortfall path end to end through `reserveInventoryForConfirmedOrder`. `inventoryPolicy.test.ts:108` covers the unpopulated-FK and missing-location edges that decide whether the filter is live at all. `putaway-lifecycle-lock.test.ts:269`/`:321` cover both the happy undelete and the status-drift refusal.

Gaps, mapping onto the findings:

1. **`api/inventory/__tests__/`** — still no test for the scope-rewriting behavior of `executeWmsCustomPostRoute` (findings 2 and 7). Needed: a body `organizationId` naming an allowed-but-not-selected organization (asserting whichever behavior you decide — pass-through or explicit rejection, but not silent rewrite); a body omitting `organizationId`; and `selectedOrganizationId === null` with `auth.orgId` set, asserting the command payload, the guard input and the interceptor context all agree.
2. **`lib/__tests__/reevaluateReservationsAfterStockIncrease.test.ts`** — still covers the status filter and paging but not scale (finding 3). Seed more than `REEVAL_LINE_PAGE_SIZE` historical lines for one variant, the vast majority on fulfilled or cancelled orders, and assert the query count stays proportional to the reservable orders rather than to total history.
3. **`data/enrichers.ts`** — nothing asserts what the enricher reports for a staging-only balance (finding 4). The unit test on the automation side is a ready-made template.
4. **`inboundStatusUi.test.ts`** — the new case pins the sort *field* but not the *direction* (finding 5), which is why the inversion shipped green.
5. **Failure-path observability** — still no case asserting the `sales_order_line` query failure emits a warning (finding 8).
6. **CI** — none of this has run in CI on any head. The two `yarn test` failures I hit locally are host artefacts in packages this PR does not touch (detailed in the gate section above), but that is a local judgement; the fork's workflow runs need maintainer approval before the branch has real CI signal, and the 59-commit lag means that run will be the full 15-shard integration suite.

pkarw:
🤖 `om-auto-review-pr` — handing this PR back to @mkadziolka.

Round-2 review posted on head `301ddc47` with a **changes-requested** verdict. Short version of what to do next:

**1. Merge `develop` first (blocker).** The head is `CONFLICTING` — one file, `packages/core/src/modules/wms/commands/shared.ts`, where `develop` added `warehouseCrudIndexer` and this branch added the three Phase-2 indexers plus their events configs. Keeping both blocks resolves it. The branch is 59 commits behind and four of those touch `packages/core/src/modules/wms/` (notably #5339's shared warehouse dialog), so please re-run `yarn typecheck` and the WMS suites after the merge, not just after clearing the markers.

**2. The two Majors from round 1 are still open and untouched.** `api/inventory/helpers.ts` and the `reevaluateReservationsAfterStockIncrease` paging loop are byte-identical to `ade3136b`. Both need either a fix or an explicit, documented decision — they are restated in full in the review body so you do not have to scroll back:
- `helpers.ts:54-64` — `executeWmsCustomPostRoute` rewrites `organizationId` from the session over whatever the client sent, so a multi-organization user targeting an allowed-but-not-selected org gets the stock mutation committed against the wrong org with a `200`. Twenty already-shipped routes inherit this, and their `openApi` still declares the field required.
- `salesOrderInventoryAutomation.ts:385-470` — the re-eval pages every `sales_order_line` ever recorded for the variant, filtering by order status only after fetching. It now fires twice per unit of received stock, and the first of those two full-history scans is guaranteed to find nothing.

**3. Four of the six new Minors are follow-through on `301ddc47`.** The staging/dock rule itself is right — I said so in the review, and the two unit tests behind it are the good kind. The gaps are: `data/enrichers.ts:274/629` is the twin of the function you fixed and still counts staging toward `available`, so the sales-order stock widget now shows "Available 10" beside "unreserved"; the aging sort direction is inverted (`inboundStatusUi.ts:67-69`); `wms.inventory.reserve` is a shipped command that gained a new refusal case with no `UPGRADE_NOTES.md` entry; and INT-07's new `expect.poll(...).toBe(0)` passes on its first tick, so the "staging alone does not reserve" behavior is not actually pinned at the integration level.

Fixing the blocker and the two Majors is what flips this; the Minors are small and mostly mechanical.

**Validation:** local fallback again — GitHub has **no workflow runs at all** for `301ddc47` (fork PRs need per-push maintainer approval, and required status checks are not enabled on `develop`). The full `validation.commands` gate passes locally: `build:packages` ×2, `generate` (no drift), `i18n:check-sync`, `i18n:check-usage`, `typecheck` and `build:app` all green, and `@open-mercato/core` is **11,736 passed / 2 skipped / 0 failed** across 1,453 suites. The two `yarn test` reds I hit were this machine, not your branch — one is a repo-walking guard in `@open-mercato/ui` racing a sibling package's temp directory, the other is jest failing to write its transform cache to a `/tmp` tmpfs at 80 % full; both pass when run per package after clearing `/tmp/jest_rs`. Details are in the review's gate section.

A maintainer still needs to approve the fork's workflow runs before this can merge — my local gate is evidence for the review only, not a substitute for CI. Reassigning to you.

pkarw:

Round-2 re-review of head `301ddc47` (round 1 was `ade3136b`). Read the full `ade3136b..301ddc47` range (23 files, +497/−64) and re-checked all nine round-1 findings against the current head — all nine are still open. Verdict driven by 1 blocker (merge conflict on `commands/shared.ts`), 2 carried Majors and 6 Minors.

Validation source: **local fallback** — GitHub has zero workflow runs for this head (fork PR awaiting per-push maintainer approval; `develop` has no required status checks). Full `validation.commands` gate run locally, `Runner: local`; every command green, with `@open-mercato/core` at 11,736 passed / 2 skipped / 0 failed. The two `yarn test` workspace reds were host artefacts in untouched packages and are documented in the review.

Labels unchanged (`changes-requested`, `feature`, `needs-qa`, `priority-medium`, `risk-high`) — each re-derived, rationale updated in place. No `ci-monitoring`: there is no CI run to watch until a maintainer approves the fork's workflows. autofix: skipped (not my PR — re-run with --autofix to fix it here).

mkadziolka:
@pkarw — response to your CHANGES_REQUESTED review (round 2, 2026-08-27). Addressing each item from the audit:

### Blocker
- **`shared.ts` merge conflict** — Resolved. PR is **MERGEABLE**.

### Major
- **Org overwrite in `executeWmsCustomPostRoute`** — Fixed: request `organizationId` is now passed through with `ensureOrganizationScope`, plus coverage and `UPGRADE_NOTES` (24206eba4 / d8e5d2820).
- **Unbounded re-evaluation** — Fixed: confirmed orders are processed first, with a scale test covering the path.

### Minors
- **#7 / #8 / #16** — Fixed in earlier commits on this branch.
- **#4 enrichers staging, #5 aging invert, #6 reserve `UPGRADE_NOTES`, #9 i18n ASN notes** — Fixed in 099398f93.

### Nits (#10–#15)
Acknowledged; deferred as optional follow-ups (non-blocking for this PR).

### CI
Green on head **099398f93** (lint / test / ephemeral).

Please re-review when you have a moment. Thanks.

PR #5537, opened 2026-08-23
fix(auth): preserve partial user ACL updates

haxiorz:
Tracking plan: .ai/runs/2026-08-24-issue-5493-preserve-partial-acl.md
Status: complete

Closes #5493

## Goal

Keep per-user ACL restrictions intact when an administrator updates only one ACL dimension, without silently widening or removing the user's effective permissions.

## Problem and root cause

The user ACL PUT route normalized omitted `features` to an empty list and omitted `organizations` to `null`. An organization-only request was therefore classified as an empty override and deleted the user's ACL row, falling back to the broader role ACL; a features-only request similarly erased the stored organization restriction.

Because a persisted `UserAcl` is an absolute override of role-derived permissions, retaining an organization-only row with no features would instead silently remove every role-granted feature. The route rejects that merged state rather than persisting it.

## What changed

- Merge omitted ACL dimensions from the existing per-user ACL before authorization and persistence.
- Reject an organization-scoped, non-super-admin ACL when the merged feature list is empty, returning an actionable translated `400` before dispatching the write command.
- Treat `organizations: []` as an empty scope: all-empty input clears the override, while a remaining feature override persists with the empty organization list.
- Preserve omitted `isSuperAdmin`, allow an explicit `false` to revoke it, and retain explicit all-dimension clearing.
- Document the partial-update semantics in the PUT OpenAPI description and explain why untouched stored features bypass re-sanitization.
- Extend route regression coverage and the existing user-ACL integration scenario to verify that an organizations-only PUT preserves the stored feature grant and its effective permission.
- Merge the current `develop` head and resolve its locale and lesson-catalog changes without dropping either side.
- Authorize partial updates against the merged ACL snapshot, with a real-guard `403` regression and an explicit OpenAPI refusal contract.

## Validation

- Local configured gate on code head `837aef769`: both `yarn build:packages` runs, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, and `yarn build:app` passed.
- `yarn test`: all changed auth/core tests passed; the monorepo command failed only in unchanged `create-mercato-app` macOS sandbox-harness cases because sandboxed Homebrew Node could not load `libuv.1.dylib`, with two nested harness typechecks also timing out.
- Focused user ACL route suites: 18/18 tests passed.
- `yarn i18n:check-values`: passed with no missing translations.
- `yarn lessons:check` and `git diff --check`: passed on final head.
[...]

PR #5623, opened 2026-08-25
fix(auth): preserve partial user ACL updates

### LINKED ISSUE #5493: bug: an admin scope edit silently deletes a user's ACL and widens their permissions
## Summary

`PUT /api/auth/users/acl` accepts an `organizations` array, but whether the request persists anything depends
entirely on whether `features` is non-empty in the **same** request. A request carrying only `organizations`
answers `200 {"ok":true,"sanitized":false}` and **deletes the user's entire custom ACL**.

Because a per-user ACL is the mechanism for **narrowing** a role, deleting it drops the user back to the full
role. The failure direction is therefore **fail-open**: an administrator adjusting an organization scope, in
good faith, silently removes the restriction they previously applied — and the API confirms success.

## Steps to reproduce

Disposable user on the `employee` role (the role alone grants 102 features), given a deliberately narrowing
per-user ACL:

| # | Actor | Request | Result |
|---|---|---|---|
| 1 | admin | `PUT /api/auth/users/acl {userId, features:['dashboards.view'], organizations:[ACME]}` | stored — `hasCustomAcl: true`, 1 feature |
| 2 | the user | `GET customers/people` · `customers/companies` · `sales/orders` · `catalog/products` | **403 · 403 · 403 · 403** — correctly restricted |
| 3 | admin | `PUT /api/auth/users/acl {userId, organizations:[ACME]}` — no `features` key | **`200 {"ok":true,"sanitized":false}`** |
| — | | `GET /api/auth/users/acl?userId=…` | `hasCustomAcl: false`, `features: []`, `organizations: null` — **the ACL is gone** |
| 4 | the user | the same four requests, fresh login | **200 · 200 · 200 · 200** — 9 people, 14 companies, 12 orders, 2 products |

## Expected

Either the request persists the organization scope it was given, or it is rejected. It must not report success
while deleting the record — and it must never widen a user's effective permissions as a side effect.

## The full behaviour matrix

Against a disposable user, deterministic, 4 of 4 steps:

| # | Request | Response | State afterwards |
|---|---|---|---|
| 0 | — | — | `hasCustomAcl:false, features:[], organizations:null` |
| 1 | `PUT {userId, organizations:[X]}` | `200 {"ok":true,"sanitized":false}` | unchanged — **nothing stored** |
| 2 | `PUT {userId, features:['customers.people.view'], organizations:[X]}` | `200` | stored correctly ✅ |
| 3 | `PUT {userId, features:[…]}` — no `organizations` | `200` | features updated, **`organizations: null`** — scope silently cleared |
| 4 | `PUT {userId, organizations:[X,Y]}` | `200 {"ok":true,"sanitized":false}` | **entire ACL deleted** |

The role endpoint `PUT /api/auth/roles/acl` behaves identically.

## Impact

No attacker is required. The trigger is a normal administrative action, the response says it worked, and
neither the response nor the ACL screen afterwards indicates that a restriction was removed. A user who was
explicitly denied access to customers, companies, orders and products regains all of it.

## Environment

- `https://demo.openmercato.com`, footer and OpenAPI both report **0.7.0**
- Verified 2026-08-22 on a freshly created user, with the escalation observed from the user's own session
  rather than inferred from the ACL record
- Reproducibility: **always**
- Pre-existing — first recorded 2026-08-20, still present after the 0.7.0 version bump
- Test user deleted afterwards

patzick:
Closes #5493

Ports the user-ACL partial-update fix to the `main` release line. The same
correction shipped to `develop` via #5537; `main` has diverged (35 commits `develop`
does not carry), so the change is applied to `main`'s own version of the route rather
than cherry-picked.

## Problem and root cause

`PUT /api/auth/users/acl` normalized an omitted `features` to `[]` and an omitted
`organizations` to `null`. An organization-only request was therefore classified as an
empty override, so the route **deleted the user's ACL row** and answered
`200 {"ok":true,"sanitized":false}`. Because a per-user ACL is the mechanism for
*narrowing* a role, deleting it drops the user back to the full role — the failure
direction is fail-open. A features-only request similarly erased the stored
organization restriction.

Simply retaining an organization-only row with no features is not a fix either: a
persisted `UserAcl` is an absolute override, so that state would silently revoke every
role-granted feature. The route must reject it rather than persist it.

## What changed

- Merge omitted `features`, `organizations`, and `isSuperAdmin` from the stored ACL
  before authorization and persistence, so a single-dimension edit no longer clears
  the dimensions it did not touch.
- Reject an organization-scoped, non-super-admin override whose merged feature list is
  empty with an actionable `400`, reusing the existing `auth.acl.organizationWarning`
  message (already present in every locale on `main`).
- Skip re-sanitizing an omitted feature list — it is already stored and in effect, and
  re-sanitizing during an unrelated organization edit would revoke untouched grants.
- Preserve an omitted `isSuperAdmin`, while an explicit `false` still revokes it, and
  explicit all-dimension clearing still removes the row.
- Document the partial-update semantics in the PUT OpenAPI description.

The role endpoint (`PUT /api/auth/roles/acl`) referenced in the issue **already**
implements these merge semantics on `main`, so it needed no change.

## Tests

- New `packages/core/src/modules/auth/api/users/acl/__tests__/partial-update.test.ts`
  covers five cases: organizations-only preserves features; features-only preserves the
  organization scope; organization scope with no feature grant returns `400` and writes
  nothing; explicit all-dimension clearing still removes the row; an omitted super-admin
  flag is preserved.
- `TC-AUTH-043-user-acl-override.spec.ts` extended so the integration scenario issues a
  follow-up organizations-only PUT and asserts the feature grant and the organization
  restriction both survive.
- Verified the new unit cases fail against the unfixed route (4 of 5 fail; the
  explicit-clear case passes both ways by design, guarding against over-correction).

## Validation

- `yarn workspace @open-mercato/core test src/modules/auth/api/users/acl src/modules/auth/lib/__tests__ --runInBand` — 18 suites, 145 tests passed.
- `yarn build:packages`, `yarn generate`, `yarn workspace @open-mercato/core typecheck` — passed.
- `eslint` on the three changed files — clean.
- Per the run's instructions, only the related tests were run locally; the full gate
  (`yarn test`, `yarn build:app`, i18n checks, integration) is delegated to CI.

## Breaking changes

None. No route, schema, response shape, database, event, or exported API contract was
removed or renamed. The one intentional behavior correction is that an unsafe
organization-scoped zero-feature override now returns `400` instead of silently
deleting the ACL.

## Labels and delivery

- `review` — ready for maintainer review.
- `bug` — corrects a destructive partial-update defect in an existing endpoint.
- `security` — the defect silently widens a user's effective permissions.
- `priority-high` — a routine admin scope edit removes an access restriction and reports success.
- `risk-medium` — confined to one auth route, but it affects authorization state.
- `skip-qa` — no UI-rendering or database surface changed; route and integration
  coverage for the corrected API behavior ships in this PR.

## Review notes — deliberate decisions

Three points came out of review that are behavior choices rather than defects, recorded
here so a reviewer can overrule them:

- **An omitted `isSuperAdmin` is preserved** (`?? existingIsSuperAdmin`). Previously a
  features-only PUT silently *demoted* a super admin; now it leaves the flag alone, and an
  explicit `false` still revokes it. This matches `PUT /api/auth/roles/acl`, which already
  behaves this way on `main`, and matches the develop-track fix. A non-super-admin actor
  still cannot grant the flag.
- **`AclEditor` can still submit the rejected state.** The editor warns about
  "organizations selected, no features" but does not block submit, and its `onSubmit`
  commits the `auth/users` write before the ACL write — so a 400 here leaves the profile
  saved and the retry hitting a 409 until the page reloads. The escape hatch is the
  "Allow all organizations" button. Closing this properly is a UI change (block submit, or
  refresh `initialUser` after the first write) and is deliberately not bundled into this
  backend fix; it affects the `develop` track identically.
- **Removing an override now requires clearing both dimensions** —
  `{features: [], organizations: null}`. Documented in `UPGRADE_NOTES.md` and the API
  reference rather than special-cased in the route.

`TC-WF-029` called `setUserAclVisibility` with organizations only, which previously
"succeeded" by storing nothing — so the scope filter it set up never existed. It now
restates the role's `workflows.*` grant. **The same latent breakage exists on `develop`**
(identical fixture and call site). CI scopes the integration suite to the modules a PR
touches, so #5537 — which changed only `auth` — ran 122 auth-scoped tests and never
selected the `workflows` spec. This PR touches both modules, so its run covered 206 tests
including `TC-WF-029` and the extended `TC-AUTH-043`; both pass.

### DIFF
--- a/UPGRADE_NOTES.md
+++ b/UPGRADE_NOTES.md
@@ -24,6 +24,32 @@ most of the patterns listed below in a user's codebase.
 
 ## 0.6.7 → 0.7.0 (2026-08-26)
 
+### `PUT /api/auth/users/acl` merges omitted fields instead of clearing them (#5493)
+
+The route used to treat every omitted field as a cleared one: an omitted `features`
+became `[]` and an omitted `organizations` became `null`. A request carrying only
+`organizations` was therefore classified as an empty override, so the route **deleted the
+user's ACL row** and answered `200 {"ok":true}`. Because a per-user ACL is how a role gets
+*narrowed*, deleting it dropped the user back to their full role — the failure direction
+was fail-open, triggered by an ordinary administrative scope edit.
+
+Omitted `features`, `organizations`, and `isSuperAdmin` now keep their stored values, so a
+single-dimension edit no longer clears the dimensions it did not touch. Two consequences
+for callers that relied on the old shape:
+
+- **Removing an override now needs every dimension cleared explicitly.** Send
+  `{ userId, features: [], organizations: null }`. A bare `{ userId, features: [] }` against
+  a row that carries an organization restriction is now rejected (see below) rather than
+  deleting the row.
+- **An organization-scoped override with no feature grant returns `400`.** A `UserAcl` is an
+  absolute override, so persisting that state would revoke every role-granted feature
+  instead of narrowing the role. Restate the grant alongside the scope —
+  `{ userId, organizations: [orgId], features: ['module.*'] }`. Test fixtures and scripts
+  that set a scope with an organizations-only call need the same restatement; previously
+  such a call reported success while storing nothing.
+
+`PUT /api/auth/roles/acl` already behaved this way and is unchanged.
+
 ### Passkey MFA verification requires a real WebAuthn assertion (#3852)
 
 `PasskeyProvider.verify()` used to accept a second payload shape — `{ credentialId, challenge }` — beside the genuine `{ response }` assertion, and approved it by string comparison. Both compared values are public: `prepareChallenge()` returns the credential id and the challenge to the caller, and `GET /api/security/mfa/methods` discloses `providerMetadata.credentialId`. A third shape needed even less: with no prepared challenge at all, only the disclosed credential id was compared. Anyone who could reach the verify step for a session therefore passed the passkey second factor with no authenticator private key and no signature, in both login-time MFA and passkey-as-sudo step-up.
--- a/apps/docs/docs/api/auth.mdx
+++ b/apps/docs/docs/api/auth.mdx
@@ -208,6 +208,15 @@ Response:
 { "ok": true }
 ```
 
+Omitted fields keep their stored value: send only `features` and the organization
+scope is preserved, send only `organizations` and the feature grant is preserved.
+To remove an override entirely, clear every dimension explicitly
+(`{ "userId": "…", "features": [], "organizations": null }`).
+
+A per-user ACL is an absolute override rather than an addition to the role, so an
+organization-scoped override with no feature grant would revoke every role-granted
+feature. That combination is rejected with `400` instead of being persisted.
+
 ## Roles
 
 ### List Roles — `GET /auth/roles`
--- a/packages/core/src/modules/auth/__integration__/TC-AUTH-043-user-acl-override.spec.ts
+++ b/packages/core/src/modules/auth/__integration__/TC-AUTH-043-user-acl-override.spec.ts
@@ -77,7 +77,32 @@ test.describe('TC-AUTH-043: user ACL override grants a feature (#2464)', () => {
       expect(aclBody?.hasCustomAcl, 'user should now have a custom ACL').toBe(true);
       expect(aclBody?.features ?? [], 'granted feature should be listed').toContain(GRANTED_FEATURE);
 
-      // The grant takes effect immediately (PUT invalidates the rbac cache for the user).
+      // A follow-up PUT that carries only the organization scope must keep the feature
+      // grant it did not touch, instead of clearing the override (#5493).
+      const scopePutRes = await apiRequest(request, 'PUT', '/api/auth/users/acl', {
+        token: superadminToken,
+        data: { userId, organizations: [organizationId] },
+      });
+      expect(scopePutRes.status(), 'partial organization PUT should return 200').toBe(200);
+      const scopePutBody = await readJsonSafe<{ ok?: boolean }>(scopePutRes);
+      expect(scopePutBody?.ok, 'partial organization PUT should report ok=true').toBe(true);
+
+      const scopedAclRes = await apiRequest(
+        request,
+        'GET',
+        `/api/auth/users/acl?userId=${encodeURIComponent(userId)}`,
+        { token: superadminToken },
+      );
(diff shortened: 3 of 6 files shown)


patzick:
🤖 `om-auto-fix-issue` claimed this PR. Ports the #5493 user-ACL partial-update fix to the `main` release line (the `develop` track shipped in #5537; `main` has diverged, so this is an adapted fix rather than a cherry-pick). Other auto-skills will skip this PR until the lock is released.

patzick:
🤖 `om-auto-fix-issue` run complete: fix ported to `main`, three review rounds applied (empty-organization-list guard, TC-WF-029 fixture restatement, contract docs). Lock released; CI is the remaining gate.

patzick:
🤖 `om-auto-review-pr --autofix` taking over this PR (re-entering the chain lock). Validation source: GitHub checks (main CI gate still pending; no local substitutes run per the repo-local checks-first rule).

patzick:
🤖 `om-auto-review-pr --autofix` complete. **Verdict: approved** — no unresolved Critical/High/Medium findings.

**Validation source: GitHub checks** (no local CI re-run, per the repo-local checks-first rule).
CI run [32872708371](https://github.com/open-mercato/open-mercato/actions/runs/32872708371) — success:
`test` ✅ · `lint` ✅ · `ds-lint` ✅ · `prepare` ✅ · `docker-build` ✅ · `ephemeral-integration` ✅ · CodeQL ✅ · `audit` skipped (`audit-scope` ✅).

The integration suite genuinely ran (206 tests, 199 passed / 0 failed / 7 skipped) and covered both specs this PR touches — `TC-AUTH-043-user-acl-override` ✓ and `TC-WF-029` ✓ — so the fixture restatement is verified rather than assumed.

Review rounds and their outcomes are recorded in the PR description. One item is deliberately **not** fixed here: `AclEditor` can still submit the state the new 400 rejects, and because `onSubmit` commits the `auth/users` write first, a failed ACL save leaves a 409 retry trap until reload. That trap is pre-existing — it fires for any ACL save failure, including the existing 403 — and closing it is a UI change outside this backend fix. Worth a follow-up issue.

Labels: `review`/`ci-monitoring` → `merge-queue`. `needs-qa` was not present; the PR carries `skip-qa` (no UI-rendering or DB surface changed, with route + integration coverage shipped). Lock released.

Issue #5633, opened 2026-08-26
test(devices): add browser-level coverage for the device owner picker and owner column


pkarw:
## Context

Split out of the review and UI QA of #5617 (see [this comment](https://github.com/open-mercato/open-mercato/pull/5617#issuecomment-5418609832)), updated for the picker semantics that PR now ships.

`TC-DEV-009` is genuinely useful but is **API-level only** — it drives Playwright's `request` fixture and never opens a `page`. The two surfaces #5617 actually changes (the `combobox` owner field on *Register device*, and the batched owner-column resolution on the list and detail pages) have no automated regression net; both were verified by hand only.

The picker's contract also changed late in that PR: `allowCustomValues: false`, backed by `devices.admin` declaring `dependsOn: ['auth.users.list']`. The paste-an-id escape hatch the earlier QA exercised is **gone**, so this spec must assert the opposite of what that QA run recorded.

## Scenario

1. **Setup** — via `packages/core/src/modules/core/__integration__/helpers/authFixtures.ts`: two users with distinct display names in the token's org; a role granted `devices.admin` **without** `auth.users.list`, plus a third user carrying only that role.
2. **Picker (full admin)** — `/backend/devices/create`: the field is labelled `devices.form.userId`, described by `devices.form.userIdHint`, and carries `role="combobox"`. Type a partial name, assert one `role="option"` containing **both** name and email (the `'search'` label style — the email is what disambiguates same-named people). Select it, fill `deviceId`, submit, assert the redirect and the `devices.form.success.created` flash.
3. **Out-of-list values are rejected** — the assertion that pins the new contract: type a raw UUID that no search returned, blur, and assert the input does **not** retain it (`ComboboxInput.confirmSelection` reverts when `allowCustomValues` is false) and the form cannot be submitted with it.
4. **Owner column** — with a device whose owner never appeared in a picker search (register it via the API, not the form), assert every User cell on `/backend/devices` resolves to a label and **none** matches `/^[0-9a-f]{8}-[0-9a-f]{4}-/i`. This is the assertion that fails if the batched `resolveDeviceUserOptions` lookup regresses. Repeat past page 1. Then assert the detail page header shows the display name.
[...]

Issue #5632, opened 2026-08-26
fix(customers): ActivitiesSection author lookup drops ids past 100 and caches its own failures


pkarw:
## Context

Split out of the review of #5617 (finding 5) — noted there as "no objection to it landing as-is", so it is filed here rather than pulled into a devices fix.

#5617 makes `GET /api/auth/users?ids=` actually filter (it was previously accepted and ignored). `packages/core/src/modules/customers/components/detail/ActivitiesSection.tsx` has been sending `?ids=` into that route all along, so its behaviour improves — but only partially.

## The two defects

`ActivitiesSection.tsx:308-322`:

1. **No chunking.** Every unresolved author id goes into one request. The route caps `?ids=` at `MAX_USER_LOOKUP_IDS = 100` and `parseIdsParam` **slices silently**, so with more than 100 unresolved authors on screen, ids 101+ are dropped by the server without any error.

2. **The failure is cached.** Line 317 adds every id to `resolvedUserIdsRef` *before* awaiting the fetch:

   ```ts
   for (const uid of unresolvedIds) resolvedUserIdsRef.current.add(uid)
   const controller = new AbortController()
   readApiResultOrThrow(...)
   ```

   Combined with (1), ids 101+ are marked resolved and never retried. If the request itself throws (403, network), the whole set is marked resolved and every one of those authors renders blank for the life of the component.

This is the same defect `6a03f4c27` fixed in the devices hook (`useDeviceUserLabels`), still present in this sibling call site.

## Suggested fix

Point the call site at the shape `packages/core/src/modules/devices/backend/devices/userOptions.ts` already uses:

- chunk at `MAX_USER_LOOKUP_IDS` (import it from `@open-mercato/core/modules/auth/lib/userIdFilter`, do not restate it);
- return which ids the server actually answered for, and cache **only those** — an id that came back without a row counts as resolved, an id whose request failed does not;
- mark ids resolved *after* the response, not before.

Whether to extract a shared helper or keep two call sites is an implementation call; the devices version is deliberately devices-local today.

## Coverage

A unit test in `packages/core/src/modules/customers/components/detail/__tests__/` asserting (a) a 150-id set produces two requests, and (b) a failed request leaves the ids unresolved so the next id-set change retries them. `packages/core/src/modules/devices/__tests__/use-device-user-labels.test.tsx` is the model.

adeptofvoltron:
[...]

PR #5544, opened 2026-08-24
fix(auth): align feature checks with selected organization

haxiorz:
Tracking plan: .ai/runs/2026-08-25-issue-5498-feature-check-org-scope.md
Status: complete

Closes #5498

## Goal

Make client-side feature visibility agree with API and page guards when a user selects a non-home organization.

## Problem and root cause

`POST /api/auth/feature-check` evaluated permissions with the organization and tenant embedded in the JWT. Route guards already resolve the request-selected organization, so users with a scoped grant in another allowed organization could pass the guarded route while the feature-check endpoint reported that the same feature was denied.

## What changed

- Resolve feature-check authorization through the established request organization-scope helper.
- Reuse the same resolved tenant and organization for aggregate and per-feature checks.
- Add unit coverage for the selected scope and a self-contained integration regression that compares the endpoint with a guarded route.
- Correct stale specification statements while retaining server-built navigation as the single authoritative navigation snapshot.

## Validation

- `yarn build:packages` — passed before and after generation (26 package tasks)
- `yarn generate` — passed; no unintended tracked output
- `yarn i18n:check-sync` — passed
- `yarn i18n:check-usage` — passed with the existing advisory unused-key inventory
- `yarn typecheck` — passed (26 package tasks)
- `yarn test` — passed (33/33 package tasks)
- `yarn build:app` — passed
- `yarn workspace @open-mercato/core test --runTestsByPath src/modules/auth/api/__tests__/feature-check.test.ts --runInBand` — passed (11/11)
- `yarn test:integration:ephemeral --no-reuse-env --no-screenshots TC-AUTH-063-feature-check-org-scope` — passed (1/1)
- `yarn template:sync` — passed

## Automated review

- `om-code-review` — approved with no findings and no remediation required
- Standard-profile direct review — approved with zero findings
- Unresolved findings — none

## Breaking changes

None. Authentication, request and response schemas, endpoint URL, status codes, wildcard behavior, and the empty-feature fast path are unchanged. There are no database, event, import-path, or other contract-surface changes.

## Labels and delivery

- `review`
- `bug`
- `skip-qa`
- `priority-medium`
- `risk-medium`

## Manual QA

Manual UI QA is not required: this changes no UI-rendering file, database structure, or API shape, and the changed behavior is covered by focused unit tests plus a fresh isolated integration scenario.

---
[...]

PR #5634, opened 2026-08-26
fix(devices): pick a device owner by name, with a closed picker and an ACL dependency (#5591)

### LINKED ISSUE #5591: bug: make the register device form usable

pkarw:
Supersedes #5617.

**Credit: original implementation by @Frshy** ([#5617](https://github.com/open-mercato/open-mercato/pull/5617), three commits carried forward verbatim). This PR exists only because the fork branch lives under an **organization** account, which GitHub excludes from maintainer pushes even though the PR reports `maintainerCanModify: true` — so the requested follow-up fixes could not land on the original branch.

## What was carried forward

Everything from #5617, unchanged: the *Register device* owner field becomes a searchable `combobox` (`/api/auth/users?search=`), the list and detail owner columns resolve the ids actually on screen in one batched request instead of reusing whichever ~20 users the filter prefetched, and `GET /api/auth/users` gains the `?ids=<uuid>,<uuid>` batch lookup that made the batched resolution possible.

### 1. The owner picker is closed, behind an ACL dependency

#5617 deliberately kept `allowCustomValues: true`, arguing that `devices.admin` does not imply `auth.users.list`, so a locked combobox would leave such an admin unable to register any device. That reasoning was right about the picker and wrong about the ACL. This PR fixes the ACL instead:

| | #5617 | Here |
|---|---|---|
| Owner picker | `allowCustomValues: true` — a pasted UUID is accepted | `allowCustomValues: false` — a free-typed or pasted value is reverted on blur |
| `devices.admin` | no declared dependency | `dependsOn: ['auth.users.list']` (`devices/acl.ts`) |
| `devices/setup.ts` | `admin: ['devices.*']` | `admin: ['devices.*', 'auth.users.list']` (same for `superadmin`) |
| Hint copy | "…or paste a user ID." | "Start typing a name or email, then pick the owner from the list." (five locales) |

Search-as-you-type is unchanged — that is what the closed picker now depends on. `auth.users.list` is granted from `devices/setup.ts` rather than left to the auth module's own `admin: ['auth.*']`, so the dependency holds even where that grant was narrowed.

### ⚠️ Deploy note — `dependsOn` diagnoses, it does not grant

`resolveAclDependencyDiagnostics` powers the ACL editor's missing-dependency panel and its one-click fix; nothing expands the dependency at runtime, and `defaultRoleFeatures` reaches existing tenants only through **`yarn mercato auth sync-role-acls`**.

So a **hand-built role** holding `devices.admin` without `auth.users.list` still exists after this merges, and on that role the register form now offers no owner and cannot be submitted. The list and detail pages still render, degrading the owner column to raw ids exactly as before. Run the sync after deploying and check the ACL editor for any custom role carrying `devices.admin`. This is recorded in the spec changelog and `devices/AGENTS.md`.

### 2. Review findings from #5617 applied

- **1** — `DeviceUserOption.description` dropped. `CrudForm` maps combobox options to `{value, label}` only (`CrudForm.tsx:4583-4586`) and `FilterBar` never reads it, so it was dead on both consumers. (`ComboboxInput` itself does render a description — it just never receives one through these two paths.)
- **2** — `MAX_IDS_PER_LOOKUP` imports `MAX_USER_LOOKUP_IDS` from `auth/lib/userIdFilter` instead of restating `100` under a comment that named the wrong constant (`MAX_IDS_PER_REQUEST`, which is 200 — raising the client cap on that comment's strength would have silently lost 100 ids per batch).
- **3** — the unreachable `else if (id)` arm removed from `auth/api/users/route.ts`.
- **4** — the chunking fixture uses real UUIDs via a shared `uuidsFor()` helper; the old `padStart(1, '0')` was a no-op producing 8-character ids.
- **6** — `useDeviceUserLabels` gets its own suite pinning the retry bookkeeping `6a03f4c27` fixed: an answered-but-empty id is cached, a failed one is retried, and a re-sorted id list does not re-fetch.
- **5** — filed as #5632 rather than pulled into a devices fix, as the review suggested.

New `devices/__tests__/acl-dependencies.test.ts` covers the dependency declaration and asserts every default role granted `devices.admin` also gets `auth.users.list`.

## Specification

`.ai/specs/2026-04-28-push-notifications-and-devices.md` — a `### 2026-08-26` changelog entry records the reversal, the `dependsOn` caveat, and the review follow-ups; the superseded `2026-08-25` bullet is marked as such.

## Testing

Local runner (no compose `app` container up), full `validation.commands` gate:

| Gate | Result |
|---|---|
| `yarn build:packages` → `yarn generate` → `yarn build:packages` | pass |
| `yarn i18n:check-sync` | pass — all five locales in sync |
| `yarn lint` | pass |
| `yarn test` | core **11582 passed** / 2 skipped, app 554, ai-assistant 1479, and every other workspace green |
| `yarn build:app` | pass |

**One known-unrelated failure:** four `create-mercato-app` "standalone template dev wrapper" tests fail on this host with `❌ Linux file-watch limits are too low for Turbopack` (`fs.inotify.max_user_watches: 253174 < 4194304`). That is a host sysctl limit, not a branch defect — this PR touches nothing under `packages/create-app`. CI runners are not subject to it.

New unit coverage: `devices/__tests__/acl-dependencies.test.ts`, `devices/__tests__/use-device-user-labels.test.tsx`, plus the corrected `devices/__tests__/user-options.test.ts`. Carried forward from #5617: `auth/__tests__/user-id-filter.test.ts` and the API-level `TC-DEV-009`.

Browser-level QA runs on this branch and its evidence is posted below. The missing *automated* browser coverage is tracked in #5633, rewritten for the closed picker (it must now assert an out-of-list value is rejected — the opposite of what #5617's QA run recorded).

## Linked issues

Fixes #5591 (first of its two asks — the docs half ships separately).

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5634"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790299938&installation_model_id=432243&pr_number=5634&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5634&signature=2c3d8ce3fbb922c03d334dfbfbfeb16877e4712f0c33fe19fce30f408ea1b6de"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pkarw:
Thanks @Frshy — this replacement PR carries your original work forward with the requested fixes applied, and assigns it to you so you can do the next recheck from the carried-forward branch.

To be explicit about **why** this is a new PR rather than commits on your branch: `fullstackhouse/open-mercato` is an **organization-owned** fork, and GitHub does not honour maintainer pushes to those even when the PR reports `maintainerCanModify: true`. The push was rejected with a 403 before this route was taken. Your three commits are here verbatim, authorship intact.

The one decision worth your attention is that this **reverses** the `allowCustomValues: true` call your PR body argues for. Your reasoning was correct *given the ACL as it stood* — the fix is to declare the dependency (`devices.admin` → `auth.users.list`) rather than to leave the picker open. The residual risk is spelled out under the deploy note in the description: `dependsOn` only diagnoses, so a hand-built `devices.admin`-only role needs `yarn mercato auth sync-role-acls` before it can complete the register form.

pkarw:
# 🔍 Code Review

**Verdict: ✅ Approve — posted as a comment, not a review, because GitHub refuses `Can not approve your own pull request`.** This PR carries @Frshy's work forward under my account for the fork-permission reason in the description, so treat this as the reviewer's record rather than a formal approval; a second maintainer's tick is still needed to satisfy branch protection.

No blockers, no majors. Two minors and a nit below.

**Validation source: local fallback** (`build:packages` → `generate` → `build:packages` → `i18n:check-sync` → `i18n:check-usage` → `typecheck` → `lint` → `test` → `build:app`, the full `validation.commands` gate plus `lint`). GitHub checks on this head are still landing — `lint`, `ds-lint`, `scope`, `audit-scope` and the CodeQL analyses are green; `prepare`, `test` and `ephemeral-integration` were pending when this was written. The local gate is this run's evidence only; required checks still gate the merge.

## 📋 Summary

Commits 1–3 are #5617 verbatim and were [reviewed and approved there](https://github.com/open-mercato/open-mercato/pull/5617#pullrequestreview-3689563012) — searchable owner picker, batched owner-column resolution, and the additive `?ids=` filter on `GET /api/auth/users`. That analysis stands unchanged and is not repeated; the security, tenancy and backward-compatibility findings there still hold, and nothing in commit 4 touches the `?ids=` contract.

Commit 4 does two things: it **reverses** #5617's `allowCustomValues: true` decision, and it makes that reversal safe by declaring the ACL dependency the old decision worked around. Reviewing it as a fresh change:

**The reversal is correctly scoped.** `ComboboxInput.confirmSelection` (`packages/ui/src/backend/inputs/ComboboxInput.tsx:299-326`) reverts on both blur and Enter-without-a-highlighted-option when `allowCustomValues` is false, and it reverts to the *known* label rather than baking a raw id back in — so the closed picker cannot freeze a UUID into the visible input while eager resolution is pending. The devices list filter was already `allowCustomValues={false}` (`FilterOverlay.tsx:344`), so the two comboboxes on this feature now agree instead of disagreeing.

**The ACL half uses the mechanism the repo already has.** `dependsOn` on a cross-module target is established practice — `auth.users.create` already depends on `directory.organizations.view` (`auth/acl.ts:8`) — and granting another module's feature from `setup.ts` has precedent too (`example` grants `payment_gateways.*`, `dashboards` grants `analytics.view`, `search` grants `vector.*`). Granting it from `devices/setup.ts` rather than leaning on auth's own `admin: ['auth.*']` is the right call: it makes the dependency hold where that grant was narrowed.

## 🔒 Security & tenancy

No new exposure. `auth.users.list` was already reachable from these screens before this PR — the picker and both owner columns have been calling `/api/auth/users` since commit 2, and the route's own guard is unchanged. What changes is that roles the platform *intends* to have that access are now declared and seeded with it, rather than silently degrading. The grant is narrow (`auth.users.list`, not `auth.*`), and `auth.users.list` returns only the caller's own tenant/org scope.

Removing the `else if (id)` arm (`auth/api/users/route.ts:403-406`) is behaviour-preserving, and I checked rather than assumed: `resolveUserIdFilter(rawIds, url.searchParams.get('id'))` at `:211` folds `?id=` into `{kind: 'ids'}` whenever it is present, `kind: 'none'` early-returns at `:240`, and the role-intersection branch early-returns on an empty set at `:332`. `idFilter` is therefore null only when neither param was supplied, which is exactly when the removed arm could not have fired. The `?id=`-only path still compiles to `{ id: { $in: [id] } }`, which is equivalent to the old `{ id }`.

## 🔁 Backward compatibility

Nothing on a frozen surface. `acl.ts` gains an **optional** field on an existing descriptor (`dependsOn` is `readonly string[] | undefined` in `FeatureDescriptor`) — additive, category 2/10. No feature id is added, renamed or removed, so no ACL rows change meaning. `defaultRoleFeatures` widens two role grants, which is additive by construction.

The one real behaviour change for an existing installation is the one the description calls out, and it is worth a reviewer holding still for: **a hand-built role with `devices.admin` and no `auth.users.list` can no longer complete the register form.** Before this PR it could, by pasting an id. That is a deliberate trade, it is documented in three places (PR body, spec changelog, `devices/AGENTS.md`), and `yarn mercato auth sync-role-acls` closes it for roles derived from the defaults — but it is not automatic for custom roles, and it is the thing to check first if someone reports "I can't register a device any more". See finding 1.

`DeviceUserOption.description` is removed from an exported type. That type is module-internal (`devices/backend/devices/userOptions.ts`, three in-module consumers, no re-export from a package entry point), so it is not a third-party contract surface.

## 🧪 Tests

Good, and the new suites pin the right things rather than the easy things.

`use-device-user-labels.test.tsx` covers the half of the fork-review defect that `6a03f4c27` fixed but left untested — the `resolvedIdsRef` bookkeeping. The case that earns its keep is "retries an id whose lookup failed, instead of remembering the failure": it fails if anyone reverts to caching on the requested set instead of the answered set. The "answered but matched no row still counts as resolved" case pins the other side, so the two together actually constrain the behaviour rather than just exercising it. The re-sort case guards the `idsKey` memo.

`acl-dependencies.test.ts` follows the shape the other modules use, and the assertion that matters is "every default role that gets `devices.admin` also gets its dependency" — that one fails if someone drops `auth.users.list` from `setup.ts` while leaving `dependsOn` in place.

`user-options.test.ts` fixture fix is real: `padStart(1, '0')` was a no-op producing 8-character ids that `parseIdsParam` would reject server-side.

Gaps, both filed rather than fixed here: no browser-level coverage of the picker or the owner column (#5633, rewritten so it asserts an out-of-list value is *rejected*), and `ActivitiesSection`'s own `?ids=` chunking (#5632).

## 🎨 Design system & i18n

Clean. No status colours, no arbitrary values, no `dark:` overrides, no new icon-only buttons; `yarn ds-lint` is green on this head. All five locales carry a real translation of the reworded `devices.form.userIdHint` — no key added, none removed, `yarn i18n:check-sync` and `yarn i18n:check-usage` both pass.

---

## 🟡 Findings (minor / nit)

**1. `dependsOn` cannot enforce what the closed picker now assumes, and only a runbook step closes the gap.** `packages/core/src/modules/devices/acl.ts:4-13`

This is the finding I would most want a second reader to push back on. `resolveAclDependencyDiagnostics` (`packages/shared/src/security/aclDependencies.ts:60`) is consumed by exactly two files — `AclEditor.tsx` and `AclDependencyDiagnosticsPanel.tsx` — both editor UI. Nothing expands a dependency at grant time, at login, or in `authorizeFeatures`. So the guarantee backing `allowCustomValues: false` is "an admin will eventually notice the panel, or someone will run `sync-role-acls`", not an invariant.

The alternative considered and deliberately not taken was adding `auth.users.list` to `requireFeatures` on `backend/devices/create/page.meta.ts`. That would make the dependency real — a role without it gets an honest 403 instead of a form it cannot submit — but it is a page-guard contract change beyond what was asked here, and it would also lock out a role that could otherwise still *view* the form. Worth deciding explicitly rather than by default; I lean toward adding it in a follow-up, because "the form renders and silently cannot be completed" is a worse failure mode than a 403.

Either way the deploy step is load-bearing: **run `yarn mercato auth sync-role-acls` when this ships.**

**2. The client now imports a constant from another module's server-side lib.** `packages/core/src/modules/devices/backend/devices/userOptions.ts:2`

`MAX_USER_LOOKUP_IDS` comes from `auth/lib/userIdFilter`, which is fine today — that file's only import is `@open-mercato/shared/lib/crud/ids`, which is pure (one type import, no runtime deps), so nothing server-only reaches the browser bundle and `build:app` confirms it. Cross-module lib imports inside `core` are established (`api_keys`, `directory`, `customer_accounts` all import from `auth/lib/*`).

The risk is drift rather than the current state: if `userIdFilter.ts` ever gains a server-only import, this client module breaks at runtime, not at build time. It is a one-line file with a stated purpose, so the risk is low — but a reviewer touching it later should know a browser bundle depends on it. A comment on the export, or a home in `shared/lib/crud/ids.ts` next to `MAX_IDS_PER_REQUEST`, would remove the trap. Not worth churning this PR for.

**3. The register form is now the only device surface with no escape hatch, and the error copy still assumes one.** `packages/core/src/modules/devices/api/admin/devices/route.ts` (`devices.errors.user_not_found`)

The server still rejects an unknown target with "Target user not found in this tenant". With the picker closed, a user can no longer reach that message through the form — only through the API. That is not a defect (the API path is real and the message is right for it), but the message is now unreachable from the UI it was written for. Nit; flagging so nobody reads its absence from QA as a regression.

---

## 🚦 CI status

Not waiting on CI, per the review contract. `lint`, `ds-lint`, `scope`, `audit-scope`, `license/cla` and the CodeQL analyses are green on this head; `prepare`, `test` and `ephemeral-integration` were **pending** when this was written. This assessment is on the code — a red `test` or `ephemeral-integration` overrides it.

One local-gate caveat, stated so it is not mistaken for a clean sweep: four `create-mercato-app` "standalone template dev wrapper" tests fail on this machine with `❌ Linux file-watch limits are too low for Turbopack` (`fs.inotify.max_user_watches: 253174 < 4194304`). That is a host sysctl limit — this PR touches nothing under `packages/create-app`, and the same four fail on an unmodified checkout. CI runners are not subject to it.

`needs-qa` is retained: four `.tsx`/UI surfaces change and the API surface changed in the carried-forward commits, so the automated-verification exemption does not apply. The QA that matters is the *reversal* — #5617's QA evidence verified the paste-an-id path this PR removes and is therefore stale. Fresh browser evidence follows in its own comment.

pkarw:
## 📸 UI QA evidence — ✅ PASS (7/7)

**Verdict:** ✅ PASS — all seven scenarios passed, including the two this change exists for: an unresolvable owner is rejected, and the ACL editor surfaces the new dependency.
**Environment:** `http://127.0.0.1:5001` (disposable ephemeral instance — production build served by `next start`, freshly migrated and seeded testcontainers Postgres) · roles `admin@acme.com` **and** a purpose-built role holding `devices.admin` **without** `auth.users.list` · browser `playwright` (Chromium, headless)
**Verified:** `carry/pr-5617-ready` @ `bbbd72f9b`

> ⚠️ The [UI QA on #5617](https://github.com/open-mercato/open-mercato/pull/5617#issuecomment-5418606553) is **stale** — it verified the paste-an-id path this PR removes. This run replaces it.

| 1 | Open *Register device* and read the owner field | Labelled `User`, hint no longer offers pasting an id, rendered as a combobox | Label **`User *`**, hint **"Start typing a name or email, then pick the owner from the list."**, `role=combobox` placeholdered **"Search users"**. Neither *"paste a user ID"* nor the old *"UUID of the user who owns this device."* appears | ✅ |
| 2 | Type a person's display name | Suggestions appear, labelled so two people sharing a name stay apart | One option: **`QA5617 Nina Owner … — qa5617-nina-…@example.test`** — name plus the disambiguating email | ✅ |
| 3 | Pick the suggestion, fill Device ID, submit | Device registered against the picked user | Redirected to `/backend/devices` with the **"Device registered"** flash; the row is present | ✅ |
| 4 | **P0** — type a well-formed UUID belonging to **nobody**, then blur | The field rejects it | Zero options offered; after blur the input is **`""`** — reverted, not kept | ✅ |
| 5 | **P0** — type arbitrary free text (`definitely not a user`), then blur | Same rejection | After blur the input is **`""`** | ✅ |
| 6 | **P0** — with no owner resolved, fill Device ID and submit | No device is created | URL stayed at **`/backend/devices/create`**; the required guard blocked the submit | ✅ |
| 7 | Register a device **via the API** for a user no picker search ever returned, then read the list's User column | Every owner resolves; no cell shows a bare UUID | Owner cells read `["QA5617 Nina Owner …", "QA5617 Omar ApiOnly …"]`. **Zero** cells matched a bare-UUID pattern — including the API-registered owner. That is the batched `?ids=` lookup doing its job | ✅ |
| 8 | Open a device's detail page | Owner renders as a display name | Header `User` reads **`QA5617 Nina Owner …`** | ✅ |
| 9 | **P0** — sign in as `devices.admin` **without** `auth.users.list`, open the list | Page renders instead of bouncing to `/login`; owner column degrades to raw ids | URL stayed at `/backend/devices`, table rendered, owner cells fell back to bare UUIDs — the documented degradation, no redirect, no error state | ✅ |
| 10 | **P0** — same role, type a name into the picker | No suggestions, field stays enabled | Zero options, status message **"No matches found"**, input enabled | ✅ |
| 11 | **P0** — same role, try to register anyway | The documented consequence: it cannot be completed | The typed id reverted to **`""`** and submit left the URL at `/backend/devices/create` — **no device created** | ✅ |
| 12 | **P1** — open that role in the ACL editor | The new dependency is surfaced with a one-click fix | The panel reads **`"Manage devices across users" needs: List users`** with an **`Add "List users"`** button | ✅ |
| 13 | **P2** — repeat the picker and the list at 390×844 | Both changed surfaces stay usable on a phone | Picker opened and returned the match; list rendered all rows with resolved owner labels | ✅ |
| 14 | Cold-load `/backend/devices/create` until the picker is interactive | Fast enough to type into immediately | **279 ms** against the production build | ✅ |

Steps 9–11 are the ones worth a reviewer's attention: they measure the **cost** of closing the picker rather than assuming it. A `devices.admin` role without `auth.users.list` keeps both read surfaces and loses only the register form — which is exactly the trade the PR describes, now demonstrated rather than argued.

Step 12 matters because the ACL editor is the *only* place the new dependency surfaces at runtime. It works, and the one-click fix is present.

### ⚠️ One correction this QA forced, already pushed as `e68b8dd2c`

My own description of the change was **too absolute**, and probing the real component rather than trusting the code comment is what caught it.

`allowCustomValues: false` does **not** mean "a pasted id never works". `ComboboxInput.findOptionForInput` matches on `option.value` as well as on label, and `CrudForm` stores the unfiltered first page (20 rows) as `suggestions` for the form's lifetime (`CrudForm.tsx:3130-3141` — `storeDynamicOptions` only writes for `query === undefined`). So a raw id belonging to a user on that cached first page still resolves — to that person, rendered by name.

Measured directly with a probe:

| Typed | Options offered | After blur |
|---|---|---|
| id of a user on the cached first page | 0 (the search returns nothing for a UUID) | resolves to **`Probe Person … — probe-…@example.test`** |
| id belonging to nobody | 0 | **`""`** |
| free text | 0 | **`""`** |

That is not a hole — the accepted value is a real user already visible to the caller, and the server still scope-checks it. But "free-typed or pasted values are reverted" was wrong as written, so the field comment, the spec changelog and `devices/AGENTS.md` now state the actual guarantee: **a value that resolves to nobody is rejected**. The P0 assertions above are written against that guarantee rather than the one I originally claimed.

### Screenshots

**step-01 — register form, owner picker**

step-01

**step-02 — suggestions by name**

step-02

**step-03 — registered via the picker**

step-03

**step-04 — an unresolvable id is rejected**

step-04

**step-05 — submit blocked without a resolved owner**

step-05

**step-06 — list owner column fully resolved**

step-06

**step-07 — detail page owner name**

step-07

**step-08 — restricted role: list still renders, degraded to ids**

step-08

**step-09 — restricted role: picker degraded to "No matches found"**

step-09

**step-10 — restricted role: cannot register**

step-10

**step-11 — ACL editor flags the missing dependency**

step-11

**step-12 — mobile picker**

step-12

**step-13 — mobile list**

step-13

### Notes for QA

- **Still no automated browser coverage.** This run was scripted but lives in a QA artifacts folder, not in the repo's integration suite. #5633 tracks landing it, already rewritten for the closed picker.
- **Not exercised:** batching past the 100-id cap in `resolveDeviceUserOptions` (needs a page with >100 distinct owners, which seeded data cannot produce), and locales other than English — the four other locale files changed here were not rendered.
- Every fixture created for this run (three users, one role, three devices) was deleted in teardown; the environment is disposable.
- The screenshots above are on the `qa-evidence-pr-5634` branch; delete it after merge.

pkarw:
## 🤖 CI result — all green on `e68b8dd2c`

Every check settled green. The code-review verdict above stands unchanged.

| Check | Result |
|---|---|
| `prepare` | pass (3m36s) |
| `test` | pass (7m4s) |
| `ephemeral-integration (none)` | pass (10m39s) |
| `docker-build` | pass (4m2s) |
| `documents-multi-instance` | pass (50s) |
| `lint` | pass (31s) |
| `ds-lint` | pass (43s) |
| `scope` | pass (29s) |
| `audit-scope` | pass (13s) |
| `CodeQL` + `Analyze (actions/javascript-typescript/python)` | pass |
| `license/cla` | pass |

I confirmed the **new** suites actually executed rather than trusting the green tick:

```
PASS src/modules/devices/__tests__/acl-dependencies.test.ts
PASS src/modules/devices/__tests__/use-device-user-labels.test.tsx
PASS src/modules/devices/__tests__/user-options.test.ts
PASS src/modules/auth/__tests__/user-id-filter.test.ts
```

and in the integration shard:

```
✓ 166 …/devices/__integration__/TC-DEV-009.spec.ts:36:3 › resolves a batch of owner ids,
      and matches nothing when ids is unusable (456ms)
162 passed, 0 failed, 0 flaky
```

That matters here because the `?ids=` batch contract and the match-nothing rule are the parts a unit test alone cannot prove.

Note the four `create-mercato-app` "standalone template dev wrapper" failures I reported from my **local** gate did **not** occur here — they were this machine's `fs.inotify` sysctl limit, as diagnosed. CI is unaffected.

---

### Where this PR stands

- ✅ Code review: approve, no blockers, 3 minor/nit findings (posted as a comment — GitHub refuses self-approval)
- ✅ UI QA: 7/7 pass with screenshots, including the closed-picker rejection and the `devices.admin`-without-`auth.users.list` degradation
- ✅ CI: fully green, new tests verified as executed

**Two things this still needs, neither of which I can do:**

1. **A second maintainer's approving review.** My review is a comment, so branch protection is not satisfied.
2. **The QA gate.** This PR carries `needs-qa` without `skip-qa`, so it cannot merge until `qa-approved` is applied. The evidence above meets the bar (real browser, production build, screenshots, the permission edge exercised) and the self-QA exception would also cover `qa-self-verified` — but I authored this PR, so clearing my own QA gate is a maintainer's call, not mine. Labels left untouched, pipeline stays at `review`.

**Deploy step, easy to miss:** run `yarn mercato auth sync-role-acls` after this ships, so existing tenants' roles pick up `auth.users.list` alongside `devices.admin`. Without it a custom `devices.admin`-only role gets a register form it cannot complete — see finding 1 in the review.

Housekeeping: delete the `qa-evidence-pr-5634` branch (screenshot host) after merge.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-26T05:59:32Z. Other auto-skills will skip this PR until the lock is released.

pkarw (commented):
# 🔍 Code Review

**Re-review of PR #5634** — *fix(devices): pick a device owner by name, not by UUID* — at head `e68b8dd2c`, base `develop`.

**Verdict: ✅ Approve — posted as a comment rather than a review, because GitHub refuses `Can not approve your own pull request`.** This PR carries @Frshy's work forward under my account for the fork-permission reason in the description, so this is the reviewer's record, not a formal approval. Branch protection still needs a second maintainer's tick, and the QA-approval gate still needs `qa-approved`.

## 🎯 Summary

This is the second pass over this branch. The first one (comment at 01:35 UTC) reviewed `bbbd72f9b`; one commit has landed since — `e68b8dd2c docs(devices): state precisely what the closed owner picker rejects` — so this re-review re-checks the whole diff and, in particular, the areas the first pass called out.

The change does three things. `GET /api/auth/users` gains an optional `?ids=<uuid>,<uuid>` batch lookup (max 100) that intersects with `?id=` and `?roleId=` and auto-sizes the page so a full batch is not truncated; the devices admin screens stop rendering bare UUIDs by resolving the owner ids actually on screen through that endpoint; and the *Register device* owner field becomes a closed `combobox` backed by `devices.admin → dependsOn: ['auth.users.list']` plus a matching grant in `devices/setup.ts`.

I found no blockers and no majors. The three items below are nits — worth picking up, none of them a reason to hold the merge.

The one thing every reviewer of this PR should read before it ships is the deploy note in the description: `dependsOn` **diagnoses, it does not grant**, so a hand-built role holding `devices.admin` without `auth.users.list` keeps existing after this merges, and on that role the register form offers no owner and cannot be submitted. `yarn mercato auth sync-role-acls` is the fix and it is recorded in the spec changelog, `devices/AGENTS.md`, and the PR body. That is a disclosed, maintainer-directed trade-off rather than a defect, so I have not filed it as a finding — but it is a real operational step, not boilerplate.

## 🧪 Validation Gate

**Validation source: GitHub checks** (per this repo's checks-first review extension — CI already ran the full gate on this exact head, so nothing was re-run locally).

Every check on `e68b8dd2c` settled green:

| Check | Result | Covers |
|---|---|---|
| `prepare` | ✅ pass (3m36s) | `yarn build:packages` → `yarn generate` → `yarn build:packages` → `yarn build:app` |
| `test` | ✅ pass (7m4s) | `yarn check:dep-versions`, `yarn tsx scripts/i18n-check-sync.ts`, `yarn i18n:check-usage`, `yarn check:time-bombs:fail`, `yarn agents:check-budget`, `yarn turbo run typecheck`, `yarn turbo run test`, `yarn test:scripts` |
| `lint` | ✅ pass (31s) | `yarn lint:check-graph`, `yarn lint`, `yarn logger:check-console:ci` |
| `ephemeral-integration (none)` | ✅ pass (10m39s) | Playwright integration shard for the affected scope (`none` is the affected-only shard name, not a skip) |
| `ds-lint` | ✅ pass (43s) | Design-system lint report (see Finding 1 — the job is a report, not a gate) |
| `docker-build`, `documents-multi-instance`, `scope`, `audit-scope`, `CodeQL`, `Analyze (js-ts/python/actions)`, `license/cla` | ✅ pass | — |

So the whole `validation.commands` list is accounted for by CI on this head; I did not run a local fallback. The typecheck and test steps are turbo-scoped to packages changed since the base on PRs, and `@open-mercato/core` is in that set, so the changed code was genuinely typechecked and tested.

One advisory note, not a finding: the head is 5 commits behind `develop` (`8bbdd53d2`, `700c5c8d9`, `15164f163`, `465fbf8f8`, `97319f09f`). GitHub reports `mergeable: MERGEABLE` and none of those commits touch a file this PR touches — the closest, `15164f163 fix(auth): align feature checks with selected organization (#5544)`, only edits `auth/api/feature-check.ts`, which this PR does not — so the green run stands.

## ✅ What I verified

- **No data-scoping breach on the new `?ids=` path.** `idFilter` is ANDed into the same `filters` array that already carries `{ tenantId: effectiveTenantId }`, the organization narrowing, and the super-admin `$nin`, and it is only appended at `route.ts:405` after all of them. A caller cannot name an id outside their tenant and read it back.
- **The removed `else if (id)` arm is genuinely unreachable, and removing it does not widen anything.** This was the first pass's Finding 3, so I re-derived it rather than trusting it. `idFilter` is non-null only when `userIdFilter.kind === 'ids'`, which `resolveUserIdFilter` returns only with `ids.length > 0`; and the `roleIds` block at `route.ts:310-333` returns an empty page at line 321 and line 332 rather than falling through with an empty set. So at line 405, `idFilter` non-null implies `idFilter.size > 0`, and the `else` branch could never have fired. The empty-intersection widening I went looking for is not there.
- **`?id=` + `?ids=` intersect rather than union**, including the disjoint case — covered by unit test (`user-id-filter.test.ts:32`) and end-to-end by `TC-DEV-009` (`?ids=bob&id=alice` → zero items).
- **A supplied-but-unusable `ids` value matches nothing** instead of falling back to the full first page, matching the rule `mergeIdFilter` already enforces for CRUD list routes. Both the unit suite and `TC-DEV-009` pin it.
- **`allowCustomValues: false` actually reaches the input.** `CrudForm.tsx:4593` forwards `builtin?.allowCustomValues ?? true` to `ComboboxInput`, whose revert-on-blur behavior is itself covered by `packages/ui/src/backend/inputs/__tests__/ComboboxInput.test.tsx`. The PR's claim that a raw id which *is* a known option still resolves is also correct — `findOptionForInput` matches on `option.value` — and the spec and `devices/AGENTS.md` now say so precisely, which is what commit `e68b8dd2c` added.
- **The ACL test is not vacuous.** `resolveAclDependencyDiagnostics` gates on `hasFeature(grantedList, featureId)` (`aclDependencies.ts:81`), which is wildcard-aware, so `devices.*` really does match `devices.admin` and really does force the `auth.users.list` check. Dropping the grant from `setup.ts` would fail `acl-dependencies.test.ts:43`. Same for the `unknownReferences` assertion — `dependsOn` targets are checked against the catalog.
- **The chunking fixture is real now.** `uuidsFor()` emits `44444444-4444-4444-8444-<12 hex>`, which satisfies the strict version/variant regex `parseIdsParam` applies; the old `padStart(1, '0')` fixture was an 8-character no-op that would have been dropped server-side. This was the first pass's Finding 4 and it is properly fixed.
- **The `useDeviceUserLabels` suite pins the retry bookkeeping `6a03f4c27` fixed** — answered-but-empty is cached, failed is retried, a re-sorted list does not re-fetch — with real assertions on the ids asked for, not just call counts.
- **No contract surface was broken.** The `ids` query param and the `dependsOn` field are additive; the ACL grant is additive; `DeviceUserOption.description` was dropped from a type that never shipped (it was introduced on the unmerged #5617 branch); no DB schema change, so no migration is expected; the OpenAPI `GET` description documents the new param.
- **i18n is complete and genuinely translated** across all five locales, with the new `devices.form.userIdPlaceholder` key present in each, in sorted position. No hardcoded user-facing strings were introduced.
- **`TC-DEV-009` is self-contained** — it creates its own users and device and tears them down in a `finally`, with no reliance on seeded data.

## 🟢 Findings — Nits

**1. The PR adds one new DS-lint finding (44 → 45 on `om-ds/require-loading-state`)** — `packages/core/src/modules/devices/backend/devices/userOptions.ts:48`.

This is a false positive: `require-loading-state` fires on any file that calls `apiCall`/`apiCallOrThrow`/`readApiResultOrThrow` and contains no `isLoading`/`loading` identifier or loading component (`packages/eslint-plugin-ds/rules/require-loading-state.js:39`), with no page/component check. `userOptions.ts` renders nothing — it is a plain data module. The finding only exists because the `apiCall` moved out of `devices/backend/devices/page.tsx`, which does pass `isLoading` to its DataTable, and into the new helper. The `ds-lint` job passed, so nothing is blocked; but the report tracks a net delta against `develop` and this pushes it up by one, on a change that made the code better. Worth a targeted `eslint-disable-next-line om-ds/require-loading-state` with the one-line rationale, so the delta stays honest.

**2. Automated browser coverage for the closed picker is deferred to #5633.** `.ai/qa/AGENTS.md` asks for integration coverage of key UI paths in the same change, and the register form's owner field is one. I am filing this as a nit rather than a minor because the behavior that would be asserted — revert-an-unknown-value-on-blur — is owned and unit-tested by `ComboboxInput` itself, the devices page contributes only the one declarative prop, the API surface underneath is covered by `TC-DEV-009`, the manual QA evidence on this PR walks the exact rejection path (7/7 PASS), and #5633 is already rewritten for the closed picker. Please make sure #5633 does get picked up — it must now assert the *opposite* of what #5617's QA run recorded.

**3. Follow-up, adjacent rather than in-scope: `ActivitiesSection` still has the bug this PR fixed in `useDeviceUserLabels`.** At `packages/core/src/modules/customers/components/detail/ActivitiesSection.tsx:317` the ids are added to `resolvedUserIdsRef` *before* the request is issued, so a failed lookup is remembered as resolved for the life of the component — exactly the failure mode `6a03f4c27` fixed on the devices side and that `resolvedIds` was introduced to prevent. This PR does not touch that file, and it is strictly better off after this merge than before (it was already sending `?ids=` and silently receiving an unfiltered page, which is what this PR fixes server-side). Worth its own issue rather than scope creep here.

## 💥 Breaking Changes

Checked against `BACKWARD_COMPATIBILITY.md`. **None.**

| Protected surface | Touched? | Assessment |
|---|---|---|
| API routes / response shapes | `GET /api/auth/users` | Additive only — a new optional `ids` query param; no response field added or removed; OpenAPI description updated |
| ACL features | `devices.admin` | Additive — gains a `dependsOn` array; no feature id renamed or removed |
| DB schema | no | No entity change, so no migration is expected |
| Public exports / import paths | `auth/lib/userIdFilter` (new) | New module; nothing removed. `DeviceUserOption.description` was dropped from a type that only ever existed on the unmerged #5617 branch |
| Event ids, widget spot ids, DI keys, notification ids, CLI commands, generated files | no | Untouched |

The one behavioral change that reaches existing installs is the ACL one, and it is a *grant widening* plus a *form tightening*: see the deploy note above. It does not break a documented contract, but it does need `yarn mercato auth sync-role-acls` after deploy.

## 🧪 Test Coverage

Coverage ships with the change and it is real coverage, not shape-checking:

- `auth/__tests__/user-id-filter.test.ts` — 7 cases over `resolveUserIdFilter`: unfiltered, single-id contract, comma batch with non-UUIDs and duplicates dropped, intersection with `?id=`, the supplied-but-unusable → match-nothing rule, the disjoint case, the cap, and whitespace-only.
- `devices/__tests__/user-options.test.ts` — both label styles, query trimming, the `x-om-forbidden-redirect: 0` header, graceful degradation on 403 and on a thrown request, unusable ids dropped, the resolved-vs-failed distinction, batch-level partial failure, dedup with an exactly-sized page, and 150-id chunking into 100 + 50.
- `devices/__tests__/use-device-user-labels.test.tsx` — the retry bookkeeping, asserted on the ids actually requested.
- `devices/__tests__/acl-dependencies.test.ts` — the dependency declaration, the target being registered, the own-device features staying clean, and every default role granted `devices.admin` also getting the dependency.
- `devices/__integration__/TC-DEV-009.spec.ts` — the API path end to end, self-contained and cleaned up.

The gap is the browser-level path, tracked in #5633 (Finding 2).

## 📋 What still stands between this and a merge

Neither is a review finding; both are process gates this review cannot clear on its own:

1. **A second maintainer's approving review.** GitHub will not let me approve my own PR, so branch protection is unsatisfied until someone else ticks it.
2. **`qa-approved`.** The PR carries `needs-qa`, and this repo's hard rule is that `needs-qa` without `qa-approved` never merges. The self-QA evidence is already posted on this PR (7/7 PASS, with screenshots); a maintainer applying `qa-approved` + `qa-self-verified` on the strength of that comment is what clears the gate. I will not apply either from reading a diff.

pkarw:
🏷️ **label rationale** — `om-auto-review-pr` re-review at head `e68b8dd2c`

No label changed. Here is why each one is where it is:

- **`review` stays; the PR does not move to `merge-queue`.** The code-review verdict is approve, but GitHub refuses `Can not approve your own pull request`, so no approving review exists and branch protection is still unsatisfied. Moving this out of `review` would take it off reviewers' queues while it genuinely still needs a second maintainer's tick — so `review` is the honest state. Once someone else approves, this belongs in `merge-queue`.
- **`needs-qa` stays, and there is no `qa-approved`.** This repo's hard rule is that `needs-qa` without `qa-approved` never merges. Self-QA evidence is already posted on this PR (7/7 PASS, screenshots attached); applying `qa-approved` + `qa-self-verified` on the strength of that comment is a maintainer's call, not something this skill will infer from a diff.
- **`priority-medium`** — unchanged. A devices admin currently identifies owners by raw UUID; it is a real usability defect on a partner-facing surface, but nothing is broken or exposed, so it is not high.
- **`risk-high`** — unchanged, and correct. The change widens `defaultRoleFeatures` and adds a hard dependency on `auth.users.list` for a form that previously accepted free text, so a hand-built role that skipped the sync command loses the ability to register a device until `yarn mercato auth sync-role-acls` runs. Blast radius beyond the devices module is what earns the label.
- **`bug`, `screenshots`, `partner-request`** — unchanged, all still accurate.

pkarw:

Re-review of head `e68b8dd2c` (the first pass reviewed `bbbd72f9b`). No blockers, no majors — three nits, listed in the review above: the +1 `om-ds/require-loading-state` DS-lint delta on `userOptions.ts:48`, the browser-level coverage deferred to #5633, and an adjacent follow-up in `ActivitiesSection.tsx:317` that is not this PR's to fix.

autofix: not run — the verdict is approve and nothing actionable remains. Pushing a nit fix would invalidate the green CI and the QA evidence already recorded against this head, which is not a trade worth making for a lint false positive.

Validation source: GitHub checks (all green on this head); no local gate was re-run.

Labels unchanged — `review` stays rather than moving to `merge-queue`, because GitHub refuses a self-approval and branch protection still needs a second maintainer. `needs-qa` also still needs `qa-approved` before this can merge; the self-QA evidence is already posted.

Issue #5528, opened 2026-08-23
bug: send-as-user and the new-conversation composer still require an email recipient on non-email channels


wojciechszyjka:
Split out of #4976 while fixing its test-send half in #5483.

## What is still broken

`POST /api/communication_channels/send-as-user` is the hub's programmatic send facade, and every field of its schema is a mail construct (`api/post/send-as-user/route.ts:20-31`):

```js
to: z.array(z.string().email()).min(1),
cc: z.array(z.string().email()).optional(),
bcc: z.array(z.string().email()).optional(),
subject: z.string().min(1).max(500).regex(/^[^\r\n]*$/),   // required; chat messages have no subject
body: z.object({ plain: …, html: … }),                      // MIME alternatives
inReplyTo: …, references: …,                                // RFC 5322 threading headers
```

A Discord recipient is a channel snowflake, chat messages have no subject, and `inReplyTo`/`references` have no chat equivalent. So an in-process caller (or any module using the DI-resolvable `communicationChannelsSendAsUser` facade) cannot send on a non-email channel.

The same is true of the backend composer for *starting a new* public conversation: `useComposeSendOperation` and `useSendDraftOperation` (`packages/ui/src/backend/messages/useMessageComposeOperations.ts:156`, `:286`) gate every public compose on a client-side `isValidEmailAddress`, and `buildComposePayload` cannot express "public message, no address".

## What is NOT broken, and why this is not urgent

The operator's actual reply path already works without an address. `messages.messages.reply` (`messages/commands/messages.ts:690`) does not go through `composeMessageSchema` — it creates the `Message` directly, copies `externalEmail` from the original (`null` for a chat channel) and emits `messages.message.sent`, which `communication_channels/subscribers/outbound-bridge.ts` picks up and queues for delivery. `useReplySubmitOperation` performs no address validation at all.

So what is blocked is *starting a new* public conversation from the hub — which on a chat channel is not a meaningful product action anyway, because there is no way to name the target beyond the channel itself.

## Why it was deferred rather than fixed in #5483

Relaxing `to` in `test-send` is a validation change: the channel already knows whether its recipients are addresses. Generalising `send-as-user` is a **design** task — it needs a decision about what the chat-shaped variant of that contract looks like (does `subject` become optional, or does the chat path get its own entry point? what replaces `inReplyTo`? do `cc`/`bcc` simply not exist?). Guessing at that inside a bug fix would have been the same mistake as inventing a synthetic address for #4975.

[...]

PR #5261, opened 2026-08-13
fix(communication_channels): validate test-send recipients per provider (#4976)

wojciechszyjka:
Refs #4976
Tracking plan: .ai/runs/2026-08-13-issue-4976-outbound-recipient-validation.md
Source doc: .ai/specs/2026-06-19-discord-communication-channel-integration.md
Status: complete

## 🎯 Goal

Give a channel provider whose recipients are *identifiers* rather than *addresses* a working outbound path: `POST /api/communication_channels/channels/{id}/test-send` now accepts a provider-native recipient, so the smoke test the Discord spec documents (§ 6 "Test it") can actually run.

## 🔍 Problem

QA of #4391 against a real Discord bot found there is no product path to send a Discord message. The adapter posts fine — replaying its exact REST contract returns HTTP 200 and the message lands in the channel — but nothing in the product can ask it to run. `test-send` typed `to` as `z.string().email()`, so a Discord channel snowflake (`1534331920463433771`) was rejected with a 422 before any adapter code executed, which also made the `Default channel ID` setting unreachable.

## 🔍 Root Cause

The hub assumed *outbound recipient ⇒ email address* and encoded that assumption in the route's own zod schema, where no adapter can influence it. Every provider therefore inherited email validation whether or not it addresses anything by email.

## What Changed

- **`lib/adapter.ts`** — new **optional** `ChannelCapabilities.recipientFormat: 'email' | 'provider-native'`. Absent means `'email'`, so every adapter written before this field keeps its exact behavior.
- **`lib/outbound-recipient.ts`** (new) — `validateOutboundRecipient(recipient, capabilities)`, the single place both shapes are checked. The `'email'` branch is the previous `z.string().email()` rule verbatim.
- **`api/post/channels/[id]/test-send/route.ts`** — the body schema now accepts any non-empty string up to 320 chars, and the recipient is validated against the *resolved adapter's* capabilities, still returning 422 on rejection.
- **`lib/email-capabilities.ts`** — the shared email baseline states `recipientFormat: 'email'` explicitly rather than relying on the default.
- **Docs** — a "Recipient shape" section in the communication-channels provider guide, so provider authors know the field exists and that declaring `'provider-native'` makes the format check theirs.
- **Spec** — § Shared prerequisite records what landed and what did not; changelog entry added.

### Security note — the guard was tightened, not relaxed

[...]

PR #5635, opened 2026-08-26
fix(communication_channels): let test-send fall back to the provider's own target (#4976)

### LINKED ISSUE #4976: bug: no product path to send a Discord message — both hub outbound endpoints require an email recipient

wojciechszyjka:
Refs #4976

## 🎯 Goal

An operator can finally run the outbound smoke test the Discord spec documents: `POST /api/communication_channels/channels/{id}/test-send` with **no recipient**, so the bot posts to the `Default channel ID` the connect dialog asked for. Until now that request shape did not exist, and the credential field was written by the dialog and read by no product path unless an inbound thread already happened to exist.

## 🔍 Problem

#4976 reports that there is no product path to send a Discord message. #5261 fixed one half of it — the recipient's **format**, so a channel snowflake is no longer rejected as "not an email". What it did not fix is the recipient's **presence**. The route still declared:

```js
// api/post/channels/[id]/test-send/route.ts:48
to: z.string().min(1).max(MAX_OUTBOUND_RECIPIENT_LENGTH).regex(/^[^\r\n]*$/),
```

…while the smoke test § 6 *"Test it"* of `.ai/specs/2026-06-19-discord-communication-channel-integration.md` documents is the one with **no** `to` at all:

> **Outbound smoke test**: `POST /api/communication_channels/channels/{id}/test-send` … The bot posts to `defaultChannelId`; verify the message appears in Discord.

So an operator could smoke-test by pasting the channel id in by hand, but the `Default channel ID` field — whose help text promised *"default text channel for outbound sends and the test-send smoke test"* — was unreachable. This is the gap the issue's own Impact section names, and the one @wojciechszyjka left open on 2026-08-23 with "I have not opened a PR for this — after #5261 the remaining delta is a few lines".

## 🔍 Root Cause

Presence and format were conflated. `validateOutboundRecipient` branches on `capabilities.recipientFormat` for the *shape* of the recipient, but its first guard — `typeof recipient !== 'string'` → `Recipient is required` — applied to every provider regardless, and the route's schema made the field mandatory before the adapter was even resolved. Meanwhile the adapter side had already implemented the fallback the spec describes:

```js
// packages/channel-discord/.../lib/adapter.ts:339 — resolveTargetChannelId
discordChannelId → conversationId → metadata.to → defaultChannelId
```

Nothing in the product could produce a message that reached that last step.

## What Changed

- **`packages/core/src/modules/communication_channels/lib/outbound-recipient.ts`** — `validateOutboundRecipient` gains one leading branch: an **omitted** recipient is accepted when, and only when, the adapter declares `recipientFormat: 'provider-native'`, because those adapters carry their own configured outbound target. Everything after that branch is untouched, so the CR/LF rejection, the character allowlist and the `..` guard the spec marks MUST-survive are byte-identical — an absent value carries nothing to filter.
- **`.../api/post/channels/[id]/test-send/route.ts`** — `to` becomes `.optional()`, and the route omits the `to` key from `metadata` entirely rather than passing `undefined`, so an adapter reading `metadata.to` sees no key at all and falls through to its own default.
- **`packages/channel-discord/.../integration.ts`** — the `Default channel ID` help text now tells the operator how to exercise the field ("call the hub test-send endpoint with no recipient to verify it"), replacing copy that described a capability the endpoint did not have.
- **`.ai/specs/2026-06-19-discord-communication-channel-integration.md`** — § *Shared prerequisite* status updated and a changelog entry added, including the two things this still does not fix.

**The optionality is provider-derived, not caller-derived**, and that is the whole safety argument. An email provider has no default address to fall back to, so `'email'` — the format every existing provider resolves to, declared or defaulted — still answers `Recipient is required`. And omission means `undefined` alone: `null` and `''` stay 422 on **every** provider, because a caller who meant to address someone and mistyped must not be silently rerouted to the provider's default channel.

## 🧪 Tests

- `yarn workspace @open-mercato/core test` (communication_channels scope) — **734 passed / 79 suites**.
- Full gate, local mode, all 8 commands from `.ai/agentic.config.json` in order — `build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `test` (**34/34 turbo tasks**), `build:app` — all green.
- **Verified the new tests are real regressions**: stashing only the two production files and re-running makes 4 of them fail, and restoring the files makes them pass.

New coverage:

- `outbound-recipient.test.ts` — an omitted recipient is accepted on a provider-native provider; it is still `Recipient is required` on an email provider across all five capability shapes (`undefined`, `null`, `{}`, explicit `'email'`, `baseEmailCapabilities`); and the allowance does **not** extend to `null` or `''`.
- `test-send/__tests__/route.test.ts` — a body naming no recipient reaches a provider-native adapter with **no `to` key in `metadata` at all** (asserted with `'to' in metadata`, not just `toBeUndefined`, so a regression that reintroduces the key is caught); the same body is still refused with 422 on an email channel, declared or defaulted, before `sendMessage` is called.
- `TC-CHANNEL-EMAIL-HUB-001.spec.ts` — an integration case on a seeded connected email channel asserting an omitted recipient is still a 422 from the recipient validator, so the optionality cannot be made unconditional by accident. (Gated on `OM_ENABLE_TEST_CHANNEL_SEEDING`, like its siblings.)

## Scope, and why this does not close the issue

Two things named in #4976 are still not done, and the PR body says so rather than letting the linkage imply otherwise:

1. **`send-as-user` and the conditional `subject`** — tracked separately in **#5528**. `lib/send-as-user.ts:134` feeds `to[0]` into `externalEmail` on `messages.messages.compose`, so widening that schema alone would move the 422 one layer deeper and let the endpoint advertise a capability it does not have. It lands with the #4975 variant decision.
2. **The spec's "(or click *Test send* in the channel detail UI)"** — no such control exists in any `.tsx`. Noted in the spec changelog rather than silently carried; the API is currently the only smoke-test path.

@wojciechszyjka asked on the issue not to close #4976 on a partial PR, so this uses `Refs`. Closing is a maintainer call once #5528 is judged to carry the rest.

## 💥 Breaking Changes

None. A required request field became optional — **additive under `BACKWARD_COMPATIBILITY.md` Cat. 7**; every caller that sends `to` today behaves byte-identically, and every existing provider resolves to `recipientFormat: 'email'`, which this change does not touch.

## Note on an unrelated diff

`yarn generate` refreshes `packages/ui/src/backend/icons/lucideRegistry.generated.tsx` with a `bot` entry sourced from `channel-discord/message-types.ts`, already merged on `develop`. That is pre-existing generated drift on the base branch, not a product of this change, so it is reverted here to keep the diff scoped — worth a separate cleanup.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5635"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790313652&installation_model_id=432243&pr_number=5635&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5635&signature=fc07cdf50a3b5890e7be5f0072dfc89691e7e61ec6a1e099f91d88d9ceaa7e4d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.ai/specs/2026-06-19-discord-communication-channel-integration.md
+++ b/.ai/specs/2026-06-19-discord-communication-channel-integration.md
@@ -1017,6 +1017,17 @@ not solved by it**: touch-point 2 is #4976, touch-point 3 is #4977.
   feeds `input.to[0]` into `externalEmail` on `messages.messages.compose`, i.e. straight into the
   touch-point 1 validator this section is still deciding. Widening its schema alone would move the
   422 one layer deeper. It lands with the variant decision — as does the conditional `subject`.
+  **Status (2026-08-26): `test-send` complete.** The 2026-08-13 change fixed the recipient's
+  *format* but left its *presence* mandatory, so the smoke test § 6 actually documents — the one
+  with no recipient, posting to `defaultChannelId` — still had no request shape. `to` is now
+  optional at the schema, and `validateOutboundRecipient` accepts an omitted recipient when (and
+  only when) the adapter declares `'provider-native'`, because those adapters carry their own
+  configured target. The email path is unchanged: no email adapter has a default address, so
+  `'email'` — the format every existing provider resolves to — still requires a recipient.
+  Omission means `undefined` alone; `null` and `''` remain 422 on every provider, so a caller who
+  meant to name a recipient and got it wrong is never silently rerouted to the provider default.
+  The CR/LF and allowlist guards are untouched — an absent value carries no payload to filter.
+  `send-as-user` remains out, now tracked separately as **#5528**.
 - **Channel identity** (`connect-credential-channel.ts:161-169`) — tracked as **#4977** (the measured
   consequence: reconnecting Discord creates a duplicate channel, because the duplicate-mailbox guard
   stops applying once `externalIdentifier` is `NULL`): let the adapter supply its own
@@ -1287,6 +1298,31 @@ restored — in the same change that makes the feature real, not before it.
   (which is what makes a Discord sender resolve to a person rather than a display name) remains
   deliberately deferred.
   
+### 2026-08-26 — Touch-point 2: the `test-send` half completed (#4976)
+
+- The smoke test § 6 "Test it" documents — `POST /channels/{id}/test-send` with **no** recipient,
+  so the bot posts to `defaultChannelId` — now has a request shape. The 2026-08-13 change fixed
+  the recipient's format but left `to` mandatory, so the credential field the connect dialog asks
+  for was still read by no product path unless an inbound thread already existed. `to` is now
+  `.optional()` and `validateOutboundRecipient` accepts an omitted recipient for a
+  `'provider-native'` adapter, whose own `resolveTargetChannelId` fallback then applies.
+- **The optionality is provider-derived, not caller-derived**, and that is the whole safety
+  argument: an email provider has no default address to fall back to, so `'email'` — the format
+  every existing provider resolves to, declared or defaulted — still answers `Recipient is
+  required`. An integration case on a seeded connected channel pins this so the optionality
+  cannot be made unconditional by accident.
+- Omission is `undefined` only. `null` and `''` stay 422 on **every** provider: a caller that sent
+  an explicit empty recipient meant to address someone and got it wrong, and rerouting that to the
+  provider default would deliver the message somewhere the caller never named.
+- The CR/LF, allowlist and `..` guards § Shared prerequisite marks MUST-survive are untouched —
+  they sit after the omission branch and an absent value carries nothing to filter.
+- Still open on this touch-point: `send-as-user` and the conditional `subject`, now tracked as
+  **#5528**. Also unimplemented, and noted here rather than silently carried: § 6's "(or click
+  *Test send* in the channel detail UI)" — no such control exists in any `.tsx`, so the API is
+  currently the only smoke-test path.
+- Additive under Cat. 7 (a required request field became optional; no caller that sent `to` is
+  affected); no deprecation protocol required.
+
 ### 2026-08-13 — Touch-point 2 partially implemented (#4976)
 
 - The outbound smoke test is reachable again for a non-email provider:
--- a/.ai/specs/2026-06-19-discord-communication-channel-integration.md
+++ b/.ai/specs/2026-06-19-discord-communication-channel-integration.md
@@ -1310,7 +1310,13 @@ restored — in the same change that makes the feature real, not before it.
   argument: an email provider has no default address to fall back to, so `'email'` — the format
   every existing provider resolves to, declared or defaulted — still answers `Recipient is
   required`. An integration case on a seeded connected channel pins this so the optionality
-  cannot be made unconditional by accident.
+  cannot be made unconditional by accident — but it runs only where
+  `OM_ENABLE_TEST_CHANNEL_SEEDING` is enabled, and that flag is set nowhere in the repository
+  today, so it skips in CI. **The everywhere-guarantee is therefore the unit coverage**
+  (`lib/__tests__/outbound-recipient.test.ts` and the `test-send` route test, which assert the
+  email path across all five capability shapes and run in every environment); the integration
+  case is confirmation against a real seeded channel, not the primary guard. Enabling the flag
+  in the integration harness is proposed separately in #5662.
 - Omission is `undefined` only. `null` and `''` stay 422 on **every** provider: a caller that sent
   an explicit empty recipient meant to address someone and got it wrong, and rerouting that to the
   provider default would deliver the message somewhere the caller never named.
--- a/packages/channel-discord/src/modules/channel_discord/integration.ts
+++ b/packages/channel-discord/src/modules/channel_discord/integration.ts
@@ -85,8 +85,8 @@ export const integration: IntegrationDefinition = {
         required: false,
         placeholder: '123456789012345678',
         helpText:
-          'Default text channel the adapter posts to when an outbound message names no Discord channel. The hub'
-          + ' test-send endpoint accepts a channel snowflake as its recipient, so you can use it to verify this end to end.',
+          'Default text channel the adapter posts to when an outbound message names no Discord channel. Call the hub'
+          + ' test-send endpoint with no recipient to verify it, or pass a channel snowflake as the recipient to post somewhere else.',
       },
     ],
   },
--- a/packages/channel-discord/src/modules/channel_discord/integration.ts
+++ b/packages/channel-discord/src/modules/channel_discord/integration.ts
@@ -85,8 +85,9 @@ export const integration: IntegrationDefinition = {
         required: false,
         placeholder: '123456789012345678',
         helpText:
-          'Default text channel the adapter posts to when an outbound message names no Discord channel. Call the hub'
-          + ' test-send endpoint with no recipient to verify it, or pass a channel snowflake as the recipient to post somewhere else.',
+          'Default text channel the adapter posts to when an outbound message names no Discord channel. To verify it, call the'
+          + ' hub test-send endpoint with an empty JSON body ({}) so no recipient is named, or pass a channel snowflake as the'
+          + ' recipient to post somewhere else.',
       },
     ],
   },
--- a/packages/channel-discord/src/modules/channel_discord/lib/__tests__/adapter.test.ts
+++ b/packages/channel-discord/src/modules/channel_discord/lib/__tests__/adapter.test.ts
@@ -91,6 +91,29 @@ describe('DiscordChannelAdapter.sendMessage', () => {
     expect(captured.createMessage?.channelId).toBe('default-chan')
   })
 
+  it('explains itself to the operator when nothing names a channel', async () => {
+    // Reachable from the product since #4976 made test-send's recipient
+    // optional: an operator smoke-tests a connection whose `Default channel ID`
+    // was left blank. The route hands `result.error` straight back as
+    // `providerError`, so this string is read by a person, not a log.
+    const { captured } = stubRest()
+    const adapter = getDiscordChannelAdapter()
+
+    const result = await adapter.sendMessage({
+      content: { text: 'x' },
+      credentials,
+      scope: { organizationId: 'o', tenantId: 't' },
+    })
+
+    expect(result.status).toBe('failed')
+    expect(result.error).toBe(
+      'No Discord channel to post to — set "Default channel ID" on the connection, or pass a channel id as the recipient.',
+    )
+    expect(result.error).not.toContain('[internal]')
+    // Nothing is posted when there is no target.
+    expect(captured.createMessage).toBeUndefined()
+  })
+
   it('posts to the recipient the hub resolved, instead of silently using the default', async () => {
     // QA of #4391 against a live bot: the hub's test-send route puts the
     // operator's recipient in `metadata.to`, nothing read it, and every send
--- a/packages/channel-discord/src/modules/channel_discord/lib/adapter.ts
+++ b/packages/channel-discord/src/modules/channel_discord/lib/adapter.ts
@@ -80,10 +80,15 @@ class DiscordChannelAdapter implements ChannelAdapter {
 
     const targetChannelId = resolveTargetChannelId(input, credentials.defaultChannelId)
     if (!targetChannelId) {
+      // Operator-facing, not `[internal]`: since #4976 made the hub's test-send
+      // recipient optional, this is the response an operator gets for the first
+      // thing they are told to try — a smoke test on a connection whose
+      // `Default channel ID` was left blank. It surfaces verbatim as
+      // `providerError`, so it names the two ways out rather than the internals.
       return {
         externalMessageId: '',
         status: 'failed',
-        error: '[internal] Discord send requires a target channel id (conversationId or defaultChannelId)',
+        error: 'No Discord channel to post to — set "Default channel ID" on the connection, or pass a channel id as the recipient.',
       }
     }
 
--- a/packages/core/src/modules/communication_channels/__integration__/TC-CHANNEL-EMAIL-HUB-001.spec.ts
+++ b/packages/core/src/modules/communication_channels/__integration__/TC-CHANNEL-EMAIL-HUB-001.spec.ts
@@ -148,6 +148,47 @@ test.describe('TC-CHANNEL-EMAIL-HUB-001: per-user channel API contract', () => {
     }
   })
 
+  // `to` became optional so a provider that carries its own outbound target can run
+  // the documented smoke test without naming a recipient (#4976). That must not leak
+  // into the email path: no email adapter has a default address to fall back to, so
+  // an omitted recipient here has to stay a 422 rather than become a silent no-target
+  // send. This is the case that would fail if the optionality were made unconditional.
+  test('POST test-send on a connected email channel still requires a recipient', async ({
+    request,
+  }) => {
+    test.slow()
+    let token: string | null = null
+    let channelId: string | null = null
+    try {
+      token = await getAuthToken(request, 'admin')
+      const seedingAvailable = await isChannelSeedingAvailable(request, token)
+      test.skip(
+        !seedingAvailable,
+        'OM_ENABLE_TEST_CHANNEL_SEEDING is not enabled in this environment; cannot seed a connected channel.',
+      )
+
+      const stamp = Date.now()
+      channelId = await seedConnectedChannel(request, token, {
+        displayName: `TC-CHANNEL-EMAIL-HUB-001 no-recipient ${stamp}`,
+        externalIdentifier: `hub-001-no-recipient-${stamp}@test-seed.local`,
+      })
+
+      const response = await apiRequest(
+        request,
+        'POST',
+        `/api/communication_channels/channels/${encodeURIComponent(channelId)}/test-send`,
+        { token, data: { body: 'no recipient named' } },
+      )
+      expect(response.status(), 'an email-format channel refuses an omitted recipient').toBe(422)
+      const body = await readJsonSafe<{ error?: string }>(response)
+      expect(body?.error, 'the rejection comes from the recipient validator').toBe(
+        'Recipient is required',
+      )
+    } finally {
+      await deleteChannelIfExists(request, token, channelId)
+    }
+  })
+
   test('POST test-send on a connected channel rejects a CR/LF header injection with 422', async ({
     request,
   }) => {
--- a/packages/core/src/modules/communication_channels/__integration__/TC-CHANNEL-EMAIL-HUB-001.spec.ts
+++ b/packages/core/src/modules/communication_channels/__integration__/TC-CHANNEL-EMAIL-HUB-001.spec.ts
@@ -189,6 +189,82 @@ test.describe('TC-CHANNEL-EMAIL-HUB-001: per-user channel API contract', () => {
     }
   })
 
+  // The other side of the same contract, on a channel that is NOT email-shaped.
+  // `providerFlavor: 'chat'` connects the stub declaring
+  // `recipientFormat: 'provider-native'`, so these cases exercise the branch a
+  // real Discord channel takes — the one that was previously reachable only with
+  // a live bot, and therefore never covered by CI.
+  test('POST test-send on a connected provider-native channel accepts a provider id and an omitted recipient', async ({
+    request,
+  }) => {
+    test.slow()
+    let token: string | null = null
+    let channelId: string | null = null
+    try {
+      token = await getAuthToken(request, 'admin')
+      const seedingAvailable = await isChannelSeedingAvailable(request, token)
+      test.skip(
+        !seedingAvailable,
+        'OM_ENABLE_TEST_CHANNEL_SEEDING is not enabled in this environment; cannot seed a connected channel.',
+      )
+
+      const stamp = Date.now()
+      channelId = await seedConnectedChannel(request, token, {
+        displayName: `TC-CHANNEL-EMAIL-HUB-001 provider-native ${stamp}`,
+        providerFlavor: 'chat',
+      })
+      const path = `/api/communication_channels/channels/${encodeURIComponent(channelId)}/test-send`
+
+      // A provider-issued id: rejected outright on an email channel, accepted here.
+      const withProviderId = await apiRequest(request, 'POST', path, {
+        token,
+        data: { to: '1534331920463433771', body: 'provider-native recipient' },
+      })
+      expect(
+        withProviderId.status(),
+        'a provider-native channel accepts a provider-issued recipient id',
+      ).toBe(200)
+
+      // The case this test exists for: the documented smoke test shape, with no
+      // recipient at all, so the adapter's own configured target applies.
+      const withoutRecipient = await apiRequest(request, 'POST', path, {
+        token,
+        data: { body: 'omitted recipient' },
+      })
+      expect(
+        withoutRecipient.status(),
+        'a provider-native channel accepts a request that names no recipient',
+      ).toBe(200)
+      const sent = await readJsonSafe<{ status?: string }>(withoutRecipient)
+      expect(sent?.status, 'the omitted-recipient send reaches the adapter').toBe('sent')
+
+      // Widening presence must not weaken the transport-safety guards: an
+      // explicit empty or null recipient is a caller that meant to address
+      // someone and got it wrong, and must never fall back to the default.
+      for (const badRecipient of ['', null]) {
+        const response = await apiRequest(request, 'POST', path, {
+          token,
+          data: { to: badRecipient, body: 'x' },
+        })
+        expect(
+          response.status(),
+          `an explicit ${JSON.stringify(badRecipient)} recipient is refused, not treated as omitted`,
+        ).toBe(422)
+      }
+
+      // The path-steering allowlist still applies on this branch.
+      for (const steering of ['C123/messages', 'C123%2Fmessages', '..']) {
+        const response = await apiRequest(request, 'POST', path, {
+          token,
+          data: { to: steering, body: 'x' },
+        })
+        expect(response.status(), `a path-steering recipient (${steering}) is refused`).toBe(422)
+      }
+    } finally {
+      await deleteChannelIfExists(request, token, channelId)
+    }
+  })
+
   test('POST test-send on a connected channel rejects a CR/LF header injection with 422', async ({
     request,
   }) => {
--- a/packages/core/src/modules/communication_channels/api/post/channels/[id]/test-send/__tests__/route.test.ts
+++ b/packages/core/src/modules/communication_channels/api/post/channels/[id]/test-send/__tests__/route.test.ts
@@ -83,6 +83,17 @@ function invoke(to: unknown) {
   return POST(request, { params: { id: CHANNEL_ID } })
 }
 
+// Distinct from `invoke(undefined)` on purpose: this is the request shape the
+// Discord spec's outbound smoke test documents — a body that names no recipient
+// at all, so the adapter posts to its configured default channel.
+function invokeWithoutRecipient() {
+  const request = new Request(
+    `http://localhost/api/communication_channels/channels/${CHANNEL_ID}/test-send`,
+    { method: 'POST', body: JSON.stringify({ body: 'Ping from QA' }) },
+  )
+  return POST(request, { params: { id: CHANNEL_ID } })
+}
+
 describe('POST /api/communication_channels/channels/[id]/test-send — recipient validation', () => {
   beforeEach(() => {
     jest.clearAllMocks()
@@ -161,6 +172,39 @@ describe('POST /api/communication_channels/channels/[id]/test-send — recipient
       await expect(response.json()).resolves.toEqual({
         error: 'Recipient may only contain letters, digits, and the characters . _ : @ + -',
       })
+      expect(adapter.sendMessage).not.toHaveBeenCalled()
+      expect(afterSuccessMock).not.toHaveBeenCalled()
+    })
+  })
+
+  // The half of #4976 that #5261 did not reach: the route made `to` mandatory,
+  // so the smoke test the spec documents — no recipient, adapter posts to
+  // `defaultChannelId` — had no request shape and 422'd at the schema.
+  describe('a request that names no recipient', () => {
+    it('reaches a provider-native adapter with no `metadata.to` at all', async () => {
+      const adapter = buildAdapter('provider-native')
+      getChannelAdapterMock.mockReturnValue(adapter)
+
+      const response = await invokeWithoutRecipient()
+
+      expect(response.status).toBe(200)
+      expect(adapter.sendMessage).toHaveBeenCalledTimes(1)
+      const metadata = adapter.sendMessage.mock.calls[0][0].metadata as Record<string, unknown>
+      expect('to' in metadata).toBe(false)
+      expect(afterSuccessMock).toHaveBeenCalledTimes(1)
+    })
+
+    it.each([
+      ['declared explicitly', 'email' as const],
+      ['left to the default', undefined],
+    ])('is still refused with 422 when recipientFormat is %s', async (_label, format) => {
+      const adapter = buildAdapter(format)
+      getChannelAdapterMock.mockReturnValue(adapter)
+
+      const response = await invokeWithoutRecipient()
+
+      expect(response.status).toBe(422)
+      await expect(response.json()).resolves.toEqual({ error: 'Recipient is required' })
       expect(adapter.sendMessage).not.toHaveBeenCalled()
       expect(afterSuccessMock).not.toHaveBeenCalled()
     })
--- a/packages/core/src/modules/communication_channels/api/post/channels/[id]/test-send/route.ts
+++ b/packages/core/src/modules/communication_channels/api/post/channels/[id]/test-send/route.ts
@@ -37,15 +37,20 @@ export const metadata = {
   },
 }
 
-// `to` is deliberately NOT typed as an email here. The recipient shape depends
-// on the provider (an email address for Gmail/IMAP, a channel snowflake for
-// Discord), and the adapter is only known once the channel is loaded — so the
-// schema accepts any non-empty single-line string and `validateOutboundRecipient`
-// applies the provider-appropriate rules below, defaulting to email. The CR/LF
-// rejection stays at the schema level so a header-injection attempt is refused
-// before the request reaches the channel lookup or the mutation guard.
+// `to` is deliberately NOT typed as an email here, and deliberately optional.
+// The recipient shape depends on the provider (an email address for Gmail/IMAP,
+// a channel snowflake for Discord), and the adapter is only known once the
+// channel is loaded — so the schema accepts any non-empty single-line string and
+// `validateOutboundRecipient` applies the provider-appropriate rules below,
+// defaulting to email. Whether the recipient may be left out is provider-derived
+// too: an adapter that carries its own outbound target (Discord's
+// `defaultChannelId`) accepts a body with no `to` and posts there, which is the
+// smoke test the Discord spec documents; an email provider has no such default
+// and `validateOutboundRecipient` still refuses the request. The CR/LF rejection
+// stays at the schema level so a header-injection attempt is refused before the
+// request reaches the channel lookup or the mutation guard.
 const bodySchema = z.object({
-  to: z.string().min(1).max(MAX_OUTBOUND_RECIPIENT_LENGTH).regex(/^[^\r\n]*$/),
+  to: z.string().min(1).max(MAX_OUTBOUND_RECIPIENT_LENGTH).regex(/^[^\r\n]*$/).optional(),
   subject: z.string().min(1).max(500).optional(),
   body: z.string().max(50_000).optional(),
 })
@@ -232,7 +237,14 @@ export async function POST(req: Request, context: RouteContext): Promise<Respons
         tenantId: auth.tenantId as string,
         organizationId: channelCredentialsOrg,
       },
-      metadata: { to: body.to, subject: body.subject ?? 'Test send', testSend: true },
+      // `to` is omitted from the metadata rather than passed as `undefined` when
+      // the caller left it out, so an adapter reading `metadata.to` sees no key
+      // at all and falls through to its own configured target.
+      metadata: {
+        ...(body.to !== undefined ? { to: body.to } : {}),
+        subject: body.subject ?? 'Test send',
+        testSend: true,
+      },
     })
     await guard.afterSuccess()
     return NextResponse.json({
--- a/packages/core/src/modules/communication_channels/lib/__tests__/outbound-recipient.test.ts
+++ b/packages/core/src/modules/communication_channels/lib/__tests__/outbound-recipient.test.ts
@@ -74,12 +74,46 @@ describe('validateOutboundRecipient', () => {
     })
   })
 
+  // The second half of #4976: a provider-native adapter is configured with its
+  // own outbound target (Discord's `defaultChannelId`), so omitting the
+  // recipient is how the documented smoke test addresses it. Before this, `to`
+  // was mandatory on every provider and the credential field was written by the
+  // connect dialog but read by no product path.
+  describe('an omitted recipient', () => {
+    it('is accepted on a provider-native provider, so the adapter default applies', () => {
+      expect(validateOutboundRecipient(undefined, providerNative)).toEqual({ ok: true })
+    })
+
+    it.each([undefined, null, {}, { recipientFormat: 'email' as const }, baseEmailCapabilities])(
+      'is still required on an email provider for capabilities %#',
+      (capabilities) => {
+        expect(validateOutboundRecipient(undefined, capabilities)).toEqual({
+          ok: false,
+          error: 'Recipient is required',
+        })
+      },
+    )
+
+    // Omission means `undefined` and nothing else. A caller that sent an
+    // explicit empty or null recipient meant to address someone and got it
+    // wrong; silently rerouting that to the provider default would deliver a
+    // message somewhere the caller never named.
+    it.each([
+      ['an empty string', ''],
+      ['null', null],
+    ])('does not extend to %s on a provider-native provider', (_label, recipient) => {
+      expect(validateOutboundRecipient(recipient, providerNative)).toEqual({
+        ok: false,
+        error: 'Recipient is required',
+      })
+    })
+  })
+
   describe('shape guards, both formats', () => {
     it.each([
       ['an empty string', ''],
       ['a non-string', 42],
       ['null', null],
-      ['undefined', undefined],
     ])('rejects %s', (_label, recipient) => {
       expect(validateOutboundRecipient(recipient, providerNative)).toEqual({
         ok: false,
--- a/packages/core/src/modules/communication_channels/lib/adapter.ts
+++ b/packages/core/src/modules/communication_channels/lib/adapter.ts
@@ -60,6 +60,17 @@ export interface ChannelCapabilities {
    * applies transport-safety checks only (see `validateOutboundRecipient`) and
    * the adapter owns the provider-specific format — it MUST treat the value as
    * untrusted input.
+   *
+   * **`'provider-native'` also opts the provider into presence-optionality, not
+   * only format.** The hub accepts an outbound request that names no recipient
+   * at all and forwards it with no `metadata.to` key, on the understanding that
+   * the adapter resolves its own configured target (Discord: `defaultChannelId`).
+   * An adapter declaring this therefore MUST resolve such a target, and MUST
+   * fail legibly — an operator-readable `SendMessageResult.error`, not a throw —
+   * when it has none, because that message is surfaced verbatim to whoever ran
+   * the send. Declare `'email'` (or omit the field) if your provider has no
+   * default destination: the hub then keeps answering `Recipient is required`,
+   * which is the correct outcome for a request with nowhere to go.
    */
   recipientFormat?: 'email' | 'provider-native'
 }
--- a/packages/core/src/modules/communication_channels/lib/outbound-recipient.ts
+++ b/packages/core/src/modules/communication_channels/lib/outbound-recipient.ts
@@ -37,11 +37,24 @@ export type OutboundRecipientCheck = { ok: true } | { ok: false; error: string }
  * path to send at all (#4976). Validation now follows the adapter's
  * `capabilities.recipientFormat`, defaulting to `'email'` so every existing
  * provider keeps byte-identical behavior.
+ *
+ * A `'provider-native'` provider may also be addressed by **omitting** the
+ * recipient entirely: those adapters are configured with their own outbound
+ * target (Discord's `defaultChannelId`) and fall back to it when the message
+ * names no channel, which is the smoke test the Discord spec documents in
+ * § 6 "Test it". An email provider has no such default, so `'email'` — the
+ * format every existing provider resolves to — still requires a recipient.
+ * Omission means `undefined` only: `null` or `''` is a caller that meant to
+ * supply an address and got it wrong, and stays a 422 on every provider.
  */
 export function validateOutboundRecipient(
   recipient: unknown,
   capabilities: Pick<ChannelCapabilities, 'recipientFormat'> | null | undefined,
 ): OutboundRecipientCheck {
+  const isProviderNative = capabilities?.recipientFormat === 'provider-native'
+  if (recipient === undefined && isProviderNative) {
+    return { ok: true }
+  }
   if (typeof recipient !== 'string' || recipient.length === 0) {
     return { ok: false, error: 'Recipient is required' }
   }
@@ -51,7 +64,7 @@ export function validateOutboundRecipient(
       error: `Recipient must be at most ${MAX_OUTBOUND_RECIPIENT_LENGTH} characters`,
     }
   }
-  if (capabilities?.recipientFormat !== 'provider-native') {
+  if (!isProviderNative) {
     return emailRecipientSchema.safeParse(recipient).success
       ? { ok: true }
       : { ok: false, error: 'Recipient must be a valid email address' }
--- a/packages/core/src/modules/communication_channels/lib/test-seed.ts
+++ b/packages/core/src/modules/communication_channels/lib/test-seed.ts
@@ -137,6 +137,24 @@ class TestSeedChannelAdapter implements ChannelAdapter {
 }
 
 /**
+ * Capabilities for the chat stub. Identical to the email stub's except for the
+ * recipient shape: a chat channel is addressed by a provider-issued identifier
+ * (a Discord snowflake), not an email address.
+ *
+ * Without this override the chat stub inherited `baseEmailCapabilities`, so it
+ * claimed `channelType: 'discord'` while validating recipients as email
+ * addresses — a channel that could not be addressed the way the provider it
+ * imitates actually is. That left the hub's `'provider-native'` branch
+ * (#4976: per-provider recipient validation, and the omitted recipient that
+ * falls back to the adapter's own target) reachable only with a live Discord
+ * bot, so no integration test could cover it and CI never exercised it.
+ */
+const testSeedChatCapabilities: ChannelCapabilities = {
+  ...testSeedCapabilities,
+  recipientFormat: 'provider-native',
+}
+
+/**
  * Chat-flavoured twin of {@link TestSeedChannelAdapter}: same network-free
  * behaviour, but it declares a non-email `channelType`, so a channel connected
  * through it is shaped like a real chat channel — including an
@@ -145,6 +163,7 @@ class TestSeedChannelAdapter implements ChannelAdapter {
 class TestSeedChatChannelAdapter extends TestSeedChannelAdapter {
   readonly providerKey: string = TEST_SEED_CHAT_PROVIDER_KEY
   readonly channelType: string = 'discord'
+  readonly capabilities = testSeedChatCapabilities
 
   async normalizeInbound(raw: InboundMessage): Promise<NormalizedInboundMessage> {
     // Unlike the email stub, this one is reachable: the test-seed ingest action


wojciechszyjka:

*(Updated after the `om-auto-review-pr` pass — this comment always describes the current label set.)*

- 🔍 `review` — **kept, not swapped for `merge-queue`.** The review pass found no blockers and no majors, which would normally route the PR onward, but the verdict was submitted as a *comment*: GitHub does not permit approving one's own PR, and self-approval is not this account's to give. The PR therefore has zero approving reviews and genuinely still needs a human reviewer, so `review` is the accurate pipeline state and `merge-queue` would overstate it.
- 🐛 `bug` — this closes a wiring gap reported in #4976: a documented smoke test that had no request shape, leaving a credential field the connect dialog asks for unreadable by any product path.
- 🧪 `needs-qa` — the automated-verification exemption does not apply, because this widens a public API request schema (`to` on `test-send` becomes optional). No `.tsx` is touched, but the API surface changes, so the exemption's "API surface unchanged" condition fails and manual QA against a real Discord bot is warranted — exactly as on the sibling PR #5261. Detailed QA routes are posted in a separate comment.
- 🔺 `priority-high` — it matches the priority already carried by #4976 itself, and the gap blocks the outbound half of the Discord provider from being demonstrable as documented, which is release-relevant for #4391.
- ⚠️ `risk-high` — the diff is small and well-tested, which on its own would read as medium, but it changes a **shared hub contract surface** consumed by every channel provider, not one module's internals. Per the PR-workflow rule, when the blast-radius signals conflict the higher one wins: the recipient-validation helper is the single choke point for outbound recipients across providers, so a mistake here is not contained to Discord. The mitigating evidence is recorded rather than used to argue the label down — the email path is provably unchanged (asserted across all five capability shapes plus an integration case), the transport-safety guards sit after the new branch untouched, and the new cases were verified to fail without the production change.
- 🤖 `in-progress` — retained, not released: this is the `om-auto-fix-issue` chain lock, handed over from the issue, and the chain continues with UI verification.

wojciechszyjka:
🤖 Chain lock moved from #4976 onto this PR — `om-auto-fix-issue` continues here with `om-auto-review-pr`. Other auto-skills should skip this PR until `in-progress` is removed.

wojciechszyjka:
🤖 `om-auto-review-pr` taking over the chain lock on this PR (inherited from `om-open-pr --handoff`). Running the code review now; the lock is retained, not re-acquired.

wojciechszyjka (commented):
# 🔍 Code Review

**PR:** #5635 — `fix(communication_channels): let test-send fall back to the provider's own target (#4976)`
**Base:** `develop` · **Head:** `022770238` · **Mode:** first review, autofix-eligible (own PR, `--autofix`)

## 🎯 Summary

The PR closes the half of #4976 that #5261 left behind: the recipient's **presence**, as opposed to its format. `to` on `POST /communication_channels/channels/{id}/test-send` becomes optional, and `validateOutboundRecipient` accepts an omitted recipient only for an adapter declaring `recipientFormat: 'provider-native'`, so Discord's `resolveTargetChannelId` can fall through to the configured `defaultChannelId` — the smoke test § 6 of the Discord spec documents, which until now had no request shape at all.

The design decision worth naming is that the optionality is **provider-derived, not caller-derived**. It would have been a line shorter to make `to` optional and let every provider inherit that, and it would have been wrong: no email adapter has a default address, so an omitted recipient there is a request with no destination, not a request that means "use the default". The implementation keeps that asymmetry in one place — a single leading branch in the helper — rather than distributing it across the route and each adapter, and pins it from both sides (accepted on provider-native, still `Recipient is required` across all five email-capability shapes, plus an integration case on a seeded connected channel).

Two boundary choices are also right and non-obvious. Omission is `undefined` **only**: `null` and `''` stay 422 on every provider, so a caller who meant to address someone and mistyped is never silently rerouted to the provider's default channel — a failure mode that would be very hard to notice, since it returns `200 sent`. And the route omits the `to` key from `metadata` entirely rather than passing `undefined`, so an adapter that does `'to' in metadata` behaves the same as one that does a truthiness check.

**Security posture is unchanged, which was the main thing to verify.** The CR/LF rejection, the `PROVIDER_NATIVE_ALLOWED` character allowlist and the `..` traversal guard — flagged MUST-survive in the spec because this recipient is interpolated into `/channels/{recipient}/messages` — all sit *after* the new branch and are untouched. The new branch is reachable only when the value is `undefined`, which carries no payload to filter, so the widening cannot be used to smuggle anything past them.

## ✅ Verdict: **APPROVE**

No blockers and no majors. One minor was found during this review and **fixed in `022770238`** rather than deferred; details below.

*Submitted as a review comment rather than a formal approval: GitHub does not permit approving one's own PR, and self-approval is not this account's call regardless. A maintainer approval is still required before merge, in addition to the `qa-approved` gate.*

## 🧪 Validation Gate

Run locally (no compose `app` container running → local mode), all eight commands from `.ai/agentic.config.json` in order, on the final head `022770238`:

| # | Command | Result |
|---|---------|--------|
| 1 | `yarn build:packages` | ✅ pass — 27/27 tasks |
| 2 | `yarn generate` | ✅ pass |
| 3 | `yarn build:packages` | ✅ pass — 27/27 tasks |
| 4 | `yarn i18n:check-sync` | ✅ pass — all 5 locales in sync |
| 5 | `yarn i18n:check-usage` | ✅ pass — advisory unused-key report only (3828, pre-existing, unchanged by this PR) |
| 6 | `yarn typecheck` | ✅ pass — 27/27 tasks |
| 7 | `yarn test` | ✅ pass — 34/34 turbo tasks; `@open-mercato/core` 734/734 in `communication_channels`, `@open-mercato/channel-discord` 259/259 |
| 8 | `yarn build:app` | ✅ pass |

The gate was run twice: once on `2a0423bc4` and again in full on `022770238` after the autofix, so the reported results describe the head as pushed.

## 🟡 Minor Findings

### 1. `packages/channel-discord/.../lib/adapter.ts:86` — an `[internal]` error string became operator-facing — **FIXED in `022770238`**

Making the recipient optional turns a previously unreachable branch into the response an operator gets for the very first thing the documentation tells them to try. Before this PR, `test-send` always supplied `metadata.to`, so `resolveTargetChannelId` always returned a target and the no-target branch could only be reached through `deliver-outbound-message`. Now an operator who smoke-tests a Discord connection with `Default channel ID` left blank lands on it directly.

That branch returned:

```
'[internal] Discord send requires a target channel id (conversationId or defaultChannelId)'
```

The route hands `result.error` back verbatim as `providerError` with HTTP 200, so this reaches a person's screen. `[internal]` is the repo's marker for strings deliberately opted out of i18n *because they are not operator-facing* (root `AGENTS.md` → UI & HTTP), so leaving it there would both mislabel the string and show an operator a developer sentence naming two internal identifiers, neither of which is a thing they can act on.

**Fix applied:** reworded to name the two ways out — `No Discord channel to post to — set "Default channel ID" on the connection, or pass a channel id as the recipient.` — with a test asserting the message, the absence of `[internal]`, and that nothing is posted when there is no target.

- [x] **API route URLs (Cat. 7, STABLE)** — no route renamed, no method changed, no response field removed. A **required request field became optional**, which is the permissive direction: every caller that sends `to` today is byte-identical. `BACKWARD_COMPATIBILITY.md` § 7 explicitly permits widening request schemas.
- [x] **Public types and signatures (Cat. 3)** — `validateOutboundRecipient`'s signature, `OutboundRecipientCheck`, `MAX_OUTBOUND_RECIPIENT_LENGTH` and `ChannelCapabilities` are all unchanged. The helper's *behavior* widens for one input combination (`undefined` + `'provider-native'`), accepting more than before rather than less; it has exactly one call site in the repo and is not re-exported through a package index.
- [x] **Event IDs, widget spot IDs, DI keys, ACL features, notification IDs, CLI commands** — untouched.
- [x] **Database schema (Cat. 8)** — no entity, migration or snapshot change; none required.
- [x] **Generated files** — none hand-edited. `yarn generate` produces an unrelated one-icon refresh of `packages/ui/src/backend/icons/lucideRegistry.generated.tsx` (a `bot` entry sourced from `channel-discord/message-types.ts`, already merged on `develop`); that is pre-existing drift on the base branch and was deliberately excluded from this diff.
- [x] **Spec requirement** — `.ai/specs/2026-06-19-discord-communication-channel-integration.md` § *Shared prerequisite* is updated and a dated changelog entry records the classification.

No deprecation protocol is required.

## 🧪 Test Coverage

Coverage is proportionate and, more importantly, **verified to be regression coverage rather than coverage-shaped assertions**: stashing only the two production files and re-running makes 4 of the new cases fail; restoring them makes all pass.

- `lib/__tests__/outbound-recipient.test.ts` — omitted recipient accepted on provider-native; still `Recipient is required` on email across all five capability shapes (`undefined`, `null`, `{}`, explicit `'email'`, `baseEmailCapabilities`); the allowance does not extend to `null` or `''`.
- `test-send/__tests__/route.test.ts` — a body naming no recipient reaches a provider-native adapter with **no `to` key in `metadata`**, asserted with `'to' in metadata` rather than `toBeUndefined()`, so a regression that reintroduces the key as `undefined` is caught rather than passing silently; the same body is still 422 on an email channel before `sendMessage` is called.
- `channel-discord/lib/__tests__/adapter.test.ts` — the operator-facing no-target message, and that nothing is posted when there is no target.
- `__integration__/TC-CHANNEL-EMAIL-HUB-001.spec.ts` — an omitted recipient on a seeded connected **email** channel is still 422 from the recipient validator, so the optionality cannot be made unconditional by accident.

**One coverage caveat, stated rather than glossed:** the integration case is gated on `OM_ENABLE_TEST_CHANNEL_SEEDING`, like its siblings in that file, so it skips in CI. The email-path guarantee is therefore enforced everywhere by the route and helper unit tests, with the integration case as confirmation against a real seeded channel wherever seeding is enabled.

## 📋 Checklist Notes

- **API routes** — `openApi` and `metadata` exports preserved; auth guard, RBAC feature check, channel scoping and mutation guard all untouched and still run before the recipient check. The mutation guard receives `body` as its payload and may now see it without a `to` key; the guard contract takes an opaque `Record<string, unknown>`, and no guard in the repo reads that field.
- **Tenant scoping** — unchanged; the channel lookup still filters on `tenantId` + org scope and uses `findOneWithDecryption`.
- **No `any` introduced**, no raw `fetch`, no hand-rolled crypto, no new hardcoded user-facing string routed around i18n. The `helpText` edit in `integration.ts` follows the existing convention for integration credential fields, which are plain strings in every provider package (`channel-apns`, `channel-discord`, …); there is no `helpTextKey` in that contract, so this is not a deviation introduced here.

## ⚠️ Scope Note for the Reviewer

The PR uses `Refs #4976`, not `Closes`, because two things the issue names remain open: `send-as-user` / the conditional `subject` (tracked in #5528) and the spec's *"click Test send in the channel detail UI"*, which exists in no `.tsx`. Both are stated in the PR body and the spec changelog rather than left implied. Closing #4976 is a maintainer call once #5528 is judged to carry the remainder.

wojciechszyjka:

This PR passed review with no blockers and requires manual QA. It is **kept in `review`, not moved to `merge-queue`**, deliberately: the review was submitted as a comment because GitHub does not allow approving one's own PR, so this PR still has **zero approving reviews** and needs a human maintainer, not just QA. Moving it to `merge-queue` would advertise a readiness it has not reached. QA reviewer: when you pick it up, swap `review` for `qa`, run the routes below, and the QA-approval gate holds the merge until `qa-approved` is added.

**What this changes in one line:** `POST /api/communication_channels/channels/{id}/test-send` now accepts a body with **no `to`**, and a provider whose adapter declares `recipientFormat: 'provider-native'` (today: Discord only) posts to its configured `Default channel ID`. Nothing else about the endpoint changed.

### P0 — Email channels must be unaffected

This is the regression that would matter most, because it is silent: an email channel is a shared-contract consumer of the same helper, and the whole safety argument of this PR is that the widening does not reach it.

**Where to click**
- Backend → Settings → Communication channels → any connected **Gmail or IMAP** channel.
- Or hit the API directly: `POST /api/communication_channels/channels/{emailChannelId}/test-send`.

**What to verify**
- A body with **no** `to` (e.g. `{"body":"hello"}`) → **422** with `{"error":"Recipient is required"}`. It must not send, and must not reach the adapter.
- `{"to":"","body":"hello"}` and `{"to":null,"body":"hello"}` → **422** as well. Empty and null are *not* the same as omitted.
- `{"to":"qa@example.com","body":"hello"}` → sends exactly as it did before this PR, to that address.
- `{"to":"1534331920463433771"}` (a snowflake on an email channel) → **422**, `Recipient must be a valid email address`.

**What can go wrong**
- An omitted recipient silently succeeding on an email channel — meaning a send with no destination, or a send to a wrong default.
- The existing email path changing its error text or status code, which downstream callers may key on.

### P0 — The Discord smoke test, end to end

**Where to click**
- Connect a real Discord bot with **`Default channel ID`** filled in (the help text on that field was updated by this PR — check it now reads sensibly).
- `POST /api/communication_channels/channels/{discordChannelId}/test-send`.

**What to verify**
- `{"body":"omitted-recipient smoke test"}` (no `to`) → `200 {"status":"sent","externalMessageId":"…"}`, **and the message physically appears in the `Default channel ID` channel**. Please fetch it back from Discord rather than trusting the `200` — that is exactly the failure mode QA caught on #4391, where a validated recipient was discarded and every send went to the default while the endpoint answered `200 sent`.
- `{"to":"<a different real channel snowflake>","body":"…"}` → still posts to **that** channel, not to the default. The explicit recipient must keep outranking the default.
- With `Default channel ID` set to a **nonexistent** id and no `to` → `status: "failed"` with Discord's `Unknown Channel`, and nothing posted anywhere.

**What can go wrong**
- A recipient being accepted and then discarded (the #4391 symptom): check the *destination*, not just the status.
- A reply inside an existing conversation being redirected — send a reply in a Discord thread and confirm it stays in that thread; `conversationId` must outrank both `metadata.to` and the default.

### P1 — The blank-default error message

This path is newly reachable from the product and is the message an operator sees on their first attempt if they skipped the optional field.

**Where to click**
- Connect a Discord bot and leave **`Default channel ID` blank**.
- `POST /api/communication_channels/channels/{id}/test-send` with `{"body":"hello"}` and no `to`.

**What to verify**
- Response is `200` with `status: "failed"` and `providerError` reading: *No Discord channel to post to — set "Default channel ID" on the connection, or pass a channel id as the recipient.*
- It must **not** contain the string `[internal]`, and must not name internal identifiers like `conversationId`.

**What can go wrong**
- A developer-facing string reaching the operator, or an error that does not tell them which field to fill in.

### P1 — Transport-safety guards must still hold

The recipient is interpolated into `/channels/{recipient}/messages`, so these are the cases that must stay refused.

**Where to click**
- `POST .../test-send` on the **Discord** channel.

**What to verify**
- `{"to":"qa@example.com\r\nBcc: attacker@example.com"}` → **422** (refused at the schema, before the channel is even looked up).
- `{"to":"C123/messages"}`, `{"to":"C123%2Fmessages"}`, `{"to":".."}` → **422**, `Recipient may only contain letters, digits, and the characters . _ : @ + -`.

**What can go wrong**
- Any of these reaching the adapter would be a path-steering hole, not a cosmetic validation miss. All four are covered by unit tests, but they are cheap to confirm against the live route.

### P2 — Permissions and scoping (unchanged, worth a spot check)

**Where to click**
- `POST .../test-send` as a user **without** `communication_channels.manage` on a shared/tenant-wide channel, and against a channel belonging to **another tenant**.

**What to verify**
- Non-manager on a shared channel → `404`/`403` as before; cross-tenant channel id → `404`. Neither path changed in this PR, but the recipient check now runs after them and it is worth confirming the order was not disturbed.

### Pass/fail
- All routes pass → post your evidence, remove `qa`, and add `qa-approved`. Note that this PR **also** still needs a maintainer's approving review; `qa-approved` alone does not make it mergeable.
- Anything fails → add `qa-failed` with the failing route, the request body you sent, and the actual response.

**No screenshots requested** — this PR touches no rendering surface. The one user-visible string change is the `Default channel ID` help text in the Discord connect dialog; a single screenshot of that dialog would be welcome but is not required.

wojciechszyjka:
🤖 `om-auto-review-pr` — completed: **APPROVE** (no blockers, no majors; one minor found and fixed in `022770238` rather than deferred). Full report in the review above; manual-QA routes in the comment below it.

Autofix ran: the review's own finding — an `[internal]`-prefixed adapter error that this PR made operator-reachable — was fixed and pinned with a test, and the entire eight-command validation gate was re-run green on the amended head rather than assumed from the earlier run.

Pipeline label deliberately left at `review` instead of `merge-queue`; the reasoning is in the label-rationale comment.

**Lock retained — chain continues.** `om-auto-fix-issue` still owns `in-progress` on this PR and releases it once it finishes.

wojciechszyjka:
🤖 `om-auto-fix-issue` run complete: the `test-send` half of #4976 is implemented, reviewed (no blockers), and the eight-command validation gate is green on head `022770238`.

**UI verification: n/a.** The diff touches no rendering surface — no `.tsx`, nothing under `packages/ui/src/`, no components. The single user-visible string is the `Default channel ID` help text in the Discord connect dialog, which is integration-definition data rather than a UI change and needs a live connected bot to observe; it is covered in the manual-QA comment instead of a browser sweep, since the QA reviewer has the bot and this agent does not.

**Lock released.** `in-progress` swapped for `ci-monitoring` — the work is finished and reported, and the PR is free for anyone to act on; only CI is still being watched. Remaining gates are a maintainer's approving review and `qa-approved`.

wojciechszyjka:
🤖 `om-auto-review-pr` — CI result: ✅ **all green on `022770238`**, zero failures.

Every check settled within the follow-up budget: `test` (7m1s), `ephemeral-integration`, `docker-build` (4m9s), `prepare` (3m26s), `documents-multi-instance`, `scope`, `lint`, `ds-lint`, `audit-scope`, `CodeQL` and all three `Analyze` jobs pass; `audit`, `mutate`, `comment`, `merge-coverage` and `[code]smith` skipped by their own conditions. This matches the local gate, so the review's verdict stands unchanged — no re-review needed.

`ci-monitoring` removed; no further follow-up will come from this agent.

`mergeStateStatus` is still `BLOCKED`, which is correct and expected: the PR is `MERGEABLE` against `develop` with no conflicts, and what holds it is the two human gates — a maintainer's approving review (this PR has none; self-approval is not available) and `qa-approved` behind the `needs-qa` gate. QA routes are in the manual-QA comment above.

wojciechszyjka:
## 📸 QA evidence — run against a live app (`e2868b62e`)

Ran the QA routes from the instructions above against a real running instance (production build, `OM_ENABLE_TEST_CHANNEL_SEEDING=true`, real Postgres, bound to localhost) rather than only in unit tests. Below is what was actually measured, verbatim.

### ✅ P0 — email channels are unaffected

Against a **seeded connected email channel** on the live app:

```
POST /channels/{id}/test-send  {"body":"no recipient named"}
→ 422 {"error":"Recipient is required"}

POST /channels/{id}/test-send  {"to":"not-an-email"}
→ 422 {"error":"Recipient must be a valid email address"}

POST /channels/{id}/test-send  {"to":"qa@example.com","body":"…"}
→ 200 sent
```

This is the assertion the whole change rests on, and it holds end to end: the widening does not reach the email path.

### ✅ P0 — the provider-native path, and a harness gap found while proving it

I could not verify the Discord half the way the instructions describe, because the local dev database has **no connected Discord channel** (`select … from communication_channels where provider_key='discord'` → 0 rows) and the bot credentials are not on this machine. So I went at it through the test harness instead — and found that **the harness could not do it either**.

The test-seed `chat` stub advertises itself as *"the shape a real Discord/Slack channel has"* and declares `channelType: 'discord'`, but it inherited `baseEmailCapabilities`, i.e. `recipientFormat: 'email'`. Measured on a seeded chat channel **before** any change:

```
POST /channels/{chatChannelId}/test-send  {"to":"1534331920463433771"}
→ 422 {"error":"Recipient must be a valid email address"}

POST /channels/{chatChannelId}/test-send  {"body":"…"}          ← no recipient
→ 422 {"error":"Recipient is required"}
```

A channel imitating Discord could not be addressed the way Discord is. That means **neither half of #4976 had integration coverage, and neither could get any** — not the provider-issued recipient from #5261, nor the omitted recipient from this PR. Both were reachable only with a live bot, so CI has never exercised either.

Fixed in `e2868b62e`: the chat stub now declares `recipientFormat: 'provider-native'`. Re-measured on the same seeded channel:

```
POST /channels/{chatChannelId}/test-send  {"body":"QA 4976 omitted recipient"}
→ 200 {"status":"sent","externalMessageId":"test-seed-…","providerError":null}

POST /channels/{chatChannelId}/test-send  {"to":"1534331920463433771","body":"QA"}
→ 200 {"status":"sent",…}

POST /channels/{chatChannelId}/test-send  {"to":"qa@example.com","body":"QA"}
→ 200 {"status":"sent",…}          ← widening never narrows
```

### ✅ P1 — the widening did not cost the guards anything

Same provider-native channel, live:

| request | result |
|---|---|
| `{"to":""}` | `422` — refused at the schema, **not** treated as omitted |
| `{"to":null}` | `422` — refused at the schema, **not** treated as omitted |
| `{"to":"C123/messages"}` | `422` `Recipient may only contain letters, digits, and the characters . _ : @ + -` |
| `{"to":"C123%2Fmessages"}` | `422` — same, so the percent-encoded variant does not slip through |
| `{"to":".."}` | `422` `Recipient must not contain ".."` |
| `{"to":"qa@example.com\r\nBcc: attacker@example.com"}` | `422` at the schema regex, before the channel is looked up |

The empty/null result is the one worth dwelling on: it is the difference between "the operator chose the default" and "the operator meant to name a recipient and got it wrong". Only the second would be dangerous, because a `200 sent` to the wrong channel is very hard to notice, and both are refused.

### 🧪 Suites

`TC-CHANNEL-EMAIL-HUB-001` against the live app: **10/10 passed**, including both new cases — and, because seeding was enabled, the email-path case **actually ran** rather than skipping as it does in CI today.

```
✓ POST test-send on a connected email channel still requires a recipient (807ms)
✓ POST test-send on a connected provider-native channel accepts a provider id and an omitted recipient (2.0s)
```

Full gate re-run green on `e2868b62e`: `build:packages`, `generate`, `typecheck`, `i18n:check-sync`, `test` (34/34 turbo tasks), `build:app`.

### ⚠️ What this evidence does NOT cover

Being explicit, because the gap matters and the label should reflect a measurement rather than an expectation:

1. **The actual Discord REST post.** Everything above proves the *hub wiring* — routing, validation, and that the recipient reaches the adapter. It does not prove the bot posts into `#ogólny`, because no live bot was reachable from this machine. The `resolveTargetChannelId` fallback itself is unit-tested, but the end-to-end "the message is physically in the channel" check from the P0 instructions still needs a real connection.
2. **The `Default channel ID` blank-connection message** (P1) — the reworded operator-facing error. Its content is unit-tested; the live path was not exercised for the same reason.

So `needs-qa` stays. What is left for a human with the bot connected is exactly two things: run the smoke test with no recipient and **fetch the message back from Discord** rather than trusting the `200`, then blank the `Default channel ID` and confirm the error reads sensibly. Test A/B from @Kapsik89's #4391 re-check (valid `to` with a bogus default, and vice versa) remains the right shape for confirming the recipient is used rather than discarded.

*No production data was touched: every channel used here was seeded through the env-gated test fixture and deleted afterwards (`204` on both).*

wojciechszyjka:
## ✅ CI green on `e2868b62e` — plus one correction to my QA comment

**CI:** all checks pass on the head with the harness fix, zero failures — `test` (7m34s), `ephemeral-integration`, `docker-build` (4m13s), `prepare` (3m33s), `documents-multi-instance`, `scope`, `lint`, `ds-lint`, `audit-scope`, `CodeQL` and all three `Analyze` jobs.

### ⚠️ Correction: the new integration cases still skip in CI

My QA comment above implied the harness fix restores CI coverage for the `'provider-native'` branch. **It does not, and I should have checked before implying it.** `OM_ENABLE_TEST_CHANNEL_SEEDING` is set nowhere in this repository — not in any workflow, script, compose file, or the ephemeral-integration runner:

```
$ grep -rn "OM_ENABLE_TEST_CHANNEL_SEEDING" .github/ scripts/ packages/cli/src/
(no matches outside the specs and the module's own gate)
```

So both cases in `TC-CHANNEL-EMAIL-HUB-001` — the email one and the new provider-native one — still hit `test.skip()` in `ephemeral-integration`, exactly like their pre-existing siblings in that file. The green tick above is real, but it is not evidence that these two ran.

**What the harness fix actually changed** is narrower than "CI now covers it", and still worth having: the `'provider-native'` branch went from *impossible to cover at all* to *covered by a spec that runs wherever the flag is on*. Before `e2868b62e`, no amount of enabling flags would have helped — the chat stub validated recipients as email addresses, so the branch was unreachable without a live Discord bot. Now it is one environment variable away, and it demonstrably passes when that variable is set (10/10 locally, output in the comment above).

**What I deliberately did not do:** turn the flag on in CI. That would suddenly un-skip every seeded-channel spec across the repo — `TC-CRM-EMAIL-001` through `-005`, `TC-CHANNEL-IDENTITY-001`, `TC-CHANNEL-DISCORD-003/004/006/007` and others — which is a scope and stability decision for a maintainer, not something to slip into a bug fix. If you want it, it belongs in its own PR where a red result can be read as information rather than as this PR breaking CI.

The accurate summary of coverage on this PR is therefore:

| layer | covers | runs in CI |
|---|---|---|
| unit (`outbound-recipient`, `test-send` route, `channel-discord` adapter) | both providers, both presence cases, the guards, the operator-facing error | ✅ yes |
| integration (`TC-CHANNEL-EMAIL-HUB-001`) | the same contract against a seeded connected channel | ⏭️ only with `OM_ENABLE_TEST_CHANNEL_SEEDING=true` |
| live app (this run) | both providers end to end, guards re-verified over HTTP | n/a — evidence posted above |
| live Discord bot | the actual REST post into the channel | ❌ still outstanding, needs the bot |

The review verdict is unchanged: the behavior itself is pinned by unit tests that run everywhere and were verified to fail without the production change.

pkarw:
🤖 `om-auto-review-pr` claiming this PR for a code review (reviewer: @pkarw). Other auto-skills should skip it until `in-progress` is removed.

pkarw (changes requested):
# 🔍 Code Review

**PR:** #5635 — `fix(communication_channels): let test-send fall back to the provider's own target (#4976)`
**Base:** `develop` · **Head:** `e2868b62e` · **Reviewer:** @pkarw · **Mode:** first review by this reviewer (the earlier `# 🔍 Code Review` on this PR was posted by the author's own automation as a comment, since GitHub does not allow self-approval — this is the first review from a different account, and therefore the first one that can carry a formal verdict)

## 🎯 Summary

The change is small, correct, and argued well. `to` on `POST /communication_channels/channels/{id}/test-send` becomes `.optional()`, `validateOutboundRecipient` gains one leading branch that accepts an omitted recipient when the adapter declares `recipientFormat: 'provider-native'`, and the route omits the `to` key from `metadata` rather than passing `undefined`, so Discord's `resolveTargetChannelId` falls through to the configured `defaultChannelId`. That is exactly the smoke test § 6 of the Discord spec documents and which had no request shape until now.

I independently verified the two things the safety argument rests on:

- **`validateOutboundRecipient` has exactly one production call site** — `api/post/channels/[id]/test-send/route.ts:189`. Nothing else in the repo (including `send-as-user` and `deliver-outbound-message`) routes through it, so the widening cannot leak into another outbound path.
- **`recipientFormat` is read by nothing except that helper.** Grepping the whole workspace, the only declarations are `email-capabilities.ts:63` (`'email'`), `channel-discord/lib/capabilities.ts:73` and `test-seed.ts:154`. So the `test-seed.ts` change is contained to test-send on a chat-flavoured seeded channel, and the email path is provably untouched.

The boundary choices are right and the reasons are non-obvious enough to be worth naming. Omission is `undefined` **only** — `''` fails `.min(1)` and `null` fails `.optional()` at the schema, so both stay 422 on every provider, and a caller who meant to address someone and mistyped is never silently rerouted to a default channel while getting a `200 sent` back. And `...(body.to !== undefined ? { to: body.to } : {})` means an adapter doing `'to' in metadata` behaves identically to one doing a truthiness check.

**Security posture is unchanged.** The CR/LF schema regex, the `PROVIDER_NATIVE_ALLOWED` allowlist and the `..` traversal guard all sit *after* the new branch, and the branch is reachable only when the value is `undefined`, which carries no payload to filter. Auth, RBAC feature load, `assertCanManageChannel`, tenant/org scoping and the mutation guard all still run before the recipient check, in the same order.

## ✅ Verdict: **CHANGES REQUESTED**

No blockers and no majors — the behavior in this diff is sound and I would happily merge the code as written. The verdict is driven by **one Medium finding**, which this repository's review rules escalate to `changes_requested` (`.ai/skills/om-auto-review-pr/SKILL.md` § Verdict rule). It is a JSDoc paragraph on a contract surface, not a code change; I would expect it to take about five minutes.

## 🧪 Validation Gate

**Validation source: GitHub checks.** No local fallback was needed and none was run.

Every check on the exact head SHA `e2868b62e` (queried via `repos/.../commits/e2868b62e.../check-runs`, not the PR summary, so this is the head as pushed):

| Check | Result |
|---|---|
| `test` | ✅ success |
| `lint` | ✅ success |
| `ds-lint` | ✅ success |
| `prepare` | ✅ success |
| `ephemeral-integration (none)` | ✅ success |
| `docker-build` | ✅ success |
| `scope`, `audit-scope` | ✅ success |
| `CodeQL`, `Analyze (actions / javascript-typescript / python)` | ✅ success |
| `audit`, `mutate`, `comment`, `merge-coverage`, `[code]smith` | ⏭️ skipped by their own conditions |

Zero failures, zero pending. `mergeable: MERGEABLE`; `mergeStateStatus: BLOCKED` reflects the human gates (an approving review and `qa-approved`), not a conflict. The branch is 7 commits behind `develop`, but none of those commits touch `communication_channels` or `channel-discord`, so there is no hidden merge risk to flag.

## 🟠 Medium Findings

### 1. `packages/core/src/modules/communication_channels/lib/adapter.ts:52` — `recipientFormat` gained a second meaning that its own contract doc does not mention

This PR's root-cause section says it plainly: *"Presence and format were conflated."* The fix un-conflates them in the helper — and then re-conflates them in the flag, by making one enum value (`'provider-native'`) carry two independent claims:

1. *"my recipients are provider-issued identifiers, not email addresses"* (format — what the flag has always meant), and
2. *"I carry my own configured outbound target, so the hub may accept a request that names no recipient at all"* (presence — new in this PR).

The new semantics are documented thoroughly, but only at the *consumer* (`outbound-recipient.ts:41-48`), in the route comment, and in a Discord-specific spec. The place a provider author actually reads is the JSDoc on the capability itself, and it is unchanged:

```ts
// packages/core/src/modules/communication_channels/lib/adapter.ts:56-63
* Declare 'provider-native' when recipients are provider-issued identifiers
* rather than email addresses (e.g. a Discord channel snowflake). The hub then
* applies transport-safety checks only (see validateOutboundRecipient) and
* the adapter owns the provider-specific format — it MUST treat the value as
* untrusted input.
```

This is a cross-package contract, not a module internal: `channel-discord/lib/capabilities.ts:1` imports it as `@open-mercato/core/modules/communication_channels/lib/adapter`, and root `AGENTS.md` directs every external provider into its own workspace package, so this JSDoc *is* the provider-author-facing contract for `ChannelCapabilities`.

**The concrete consequence.** The next provider-native adapter — Slack, Telegram, Teams — declares the flag for reason (1), because that is the only reason the doc gives, and silently inherits (2). If that adapter has no configured default target, the hub stops answering `422 Recipient is required` and instead lets a destination-less send reach `sendMessage`, where the best case is Discord's outcome (`200` with `status: 'failed'` and a provider string) and the worst case is an adapter that throws, or one with some other implicit default. Today only Discord declares the flag and Discord handles it gracefully, which is why this is Medium and not High — it is a forward-looking hazard on a shared hub contract, not a live defect.

**Suggested fix**, either of:

- *(cheap, and enough)* Extend the `recipientFormat` JSDoc to state the second consequence and the obligation it implies — something like: *"Declaring `'provider-native'` also opts the provider into presence-optionality: the hub accepts an outbound request that names no recipient and the adapter MUST resolve its own configured target (and fail legibly when it has none). Declare `'email'` if your provider has no such default."*
- *(cleaner, if you'd rather not overload the flag at all)* a separate optional capability, e.g. `hasDefaultOutboundTarget?: boolean`, that `validateOutboundRecipient` checks for the omission branch instead. That keeps presence and format genuinely separate, which is the principle the PR argues for. This is a larger change and I am not asking for it in this PR — the JSDoc is sufficient.

## 🔵 Low Findings / Nits

### 2. `api/post/channels/[id]/test-send/route.ts:95` — the new help text points operators at a request that 422s with a Zod dump

`integration.ts:88` now tells the operator, in the `Default channel ID` field itself: *"Call the hub test-send endpoint with no recipient to verify it."* Read literally — `curl -X POST .../test-send` with no body at all — that fails:

`readJsonSafe(req, null)` returns `null` for an empty body (`readJsonSafe.ts:6`: `if (!raw) return fallback`), and `bodySchema.parse(null)` throws, so the route answers 422 with `err.message` — the stringified Zod issue array (`[{"code":"invalid_type","expected":"object","received":"null",…}]`). The smoke test only works if you remember to send `{}`.

This is pre-existing behavior, but the copy this PR adds is what makes it the first thing an operator hits, so it is worth one of two lines: either default the parse (`readJsonSafe(req, {})`, which is now safe precisely because every field is optional), or have the help text name the body to send. Not blocking either way.

### 3. `.ai/specs/2026-06-19-discord-communication-channel-integration.md:1311` — the changelog cites the integration case without noting it skips in CI

The entry says *"An integration case on a seeded connected channel pins this so the optionality cannot be made unconditional by accident."* You disclosed in a PR comment that `TC-CHANNEL-EMAIL-HUB-001` is gated on `OM_ENABLE_TEST_CHANNEL_SEEDING`, which is set nowhere in this repo, so both cases hit `test.skip()` in `ephemeral-integration` today — and I confirmed `ephemeral-integration (none)` passing is the affected-only shard, not evidence those cases ran.

The comment is the right disclosure, but comments are not the durable artifact; the spec is. One clause in that bullet — *"…wherever `OM_ENABLE_TEST_CHANNEL_SEEDING` is enabled; it skips in CI today, so the everywhere-guarantee is the unit coverage"* — keeps a future reader from over-reading it. The unit tests genuinely do carry the guarantee everywhere, so nothing about the coverage story is wrong, only the spec's phrasing of it.

- [x] **API routes (Cat. 7, STABLE)** — no route, method or response field changed. A required request field became optional, which is the permissive direction; every caller that sends `to` today is byte-identical. The `openApi` block declares no `requestBody`, so there is no generated-doc drift, and `422` was already a documented response.
- [x] **Public types and signatures (Cat. 3)** — `validateOutboundRecipient`'s signature, `OutboundRecipientCheck`, `MAX_OUTBOUND_RECIPIENT_LENGTH` and the `ChannelCapabilities` shape are all unchanged. The *behavior* widens for exactly one input combination (`undefined` + `'provider-native'`), accepting more rather than less. Finding 1 is about this category's **documentation**, not a violation of it — no deprecation bridge is owed for a widening.
- [x] **DB schema (Cat. 8)** — no entity, migration or snapshot change; none required.
- [x] **Event IDs, widget spot IDs, DI keys, ACL features, notification IDs, CLI commands** — untouched.
- [x] **Generated files** — none hand-edited.
- [x] **Spec requirement** — the touching spec is updated with a dated changelog entry (see nit 3 for the one phrasing tweak).

No deprecation protocol required.

## 🧪 Test Coverage

Proportionate, and — checked by reading them rather than trusting the summary — the assertions actually pin behavior rather than shape:

- `outbound-recipient.test.ts` — omitted accepted on provider-native; still `Recipient is required` on email across all five capability shapes (`undefined`, `null`, `{}`, explicit `'email'`, `baseEmailCapabilities`); the allowance explicitly does not extend to `''` or `null`. The `['undefined', undefined]` row was correctly *removed* from the shared shape-guard table rather than left contradicting the new branch.
- `test-send/__tests__/route.test.ts` — the metadata assertion is `expect('to' in metadata).toBe(false)`, not `toBeUndefined()`, so a regression reintroducing the key as `undefined` fails rather than passing silently. That distinction is the whole point of the route-side change and it is the right way to pin it.
- `channel-discord/lib/__tests__/adapter.test.ts` — asserts the operator-facing string, the absence of `[internal]`, **and** `captured.createMessage === undefined`, so "explains itself" and "posts nothing" are pinned separately.
- `__integration__/TC-CHANNEL-EMAIL-HUB-001.spec.ts` — both new cases clean up via `deleteChannelIfExists` in `finally` and seed their own fixtures, per `.ai/qa/AGENTS.md`. See nit 3 on how they are described in the spec.

## 📋 Checklist Notes

- **API route contract** — `openApi` and `metadata` exports preserved; the auth guard, RBAC feature load, `assertCanManageChannel`, the `isActive`/`status === 'connected'` check and the mutation guard all still run before the recipient check, unchanged and in the same order.
- **Tenant scoping** — untouched; the channel lookup still filters `tenantId` + `channelOrgScopeWhere(organizationId)` + `deletedAt: null` through `findOneWithDecryption`, and credentials still resolve at `channel.organizationId ?? tenantId`.
- **Production safety of the `test-seed.ts` change** — `ensureTestSeedAdapterRegistered` returns early unless `isTestChannelSeedingEnabled()`, and `test-seed.test.ts:43` pins that both stubs stay unregistered with the gate off. The subclass field override is safe: nothing reads `this.capabilities` during construction.
- **The de-`[internal]`-ed adapter string is not routed through `t()`, and that is correct here** — `sendMessage` has no i18n context, and every sibling adapter error (`'Outbound conversion failed'`, `'Discord send failed'`) is a plain string surfaced the same way. Not a finding; noting it so it does not get raised as one later.
- **`helpText` as a plain string** matches every other provider (`channel-gmail`, `channel-imap` — none use a key), so the copy change is not a deviation introduced here.
- **No `any` introduced**, no raw `fetch`, no hand-rolled crypto, no `pageSize` surface, no DS tokens or `.tsx` touched.

## ⚠️ Scope Note

`Refs #4976` rather than `Closes` is the right call and is justified in the body: `send-as-user` / the conditional `subject` is tracked in #5528, and § 6's *"click Test send in the channel detail UI"* control exists in no `.tsx` — I confirmed that independently (`grep test-send --include=*.tsx` returns nothing). Closing #4976 stays a maintainer decision once #5528 lands.

---

*Re-review welcome as soon as finding 1 is addressed — the code itself needs no change and I do not expect a second round beyond confirming the doc.*

pkarw:
🏷️ label rationale — @pkarw review pass

*(Updated in place after the re-review of `71a586d29`. This comment always describes the current label set; it supersedes the earlier rationale posted by the author's own automation where they differ.)*

- 🚀 `merge-queue` — **swapped in for `review`.** The re-review of `71a586d29` found zero findings at any severity: the Medium that drove my previous `changes-requested` (the `recipientFormat` JSDoc) is resolved, both Lows are resolved, and I verified each against the code rather than the reply. An approving review is now on the PR, which clears branch protection's `required_approving_review_count: 1` — the gate this PR could not clear before, since the author's own automation cannot self-approve. `merge-queue` is the honest state now, where it would have overstated readiness before.
- 🧪 `needs-qa` — **kept, and it is what still holds the merge.** The automated-verification exemption still does not apply: the API request schema widens (`to` becomes optional on `test-send`), so the "API surface unchanged" condition fails even though no `.tsx` is touched. The author's live-app evidence covers the hub wiring convincingly, but explicitly *not* the live Discord REST post, which needs a real bot. With `qaGate` on, this label blocks the merge until a QA reviewer adds `qa-approved`. Routes are in the manual-QA comment below.
- 🐛 `bug` — **kept.** Accurate: this closes a wiring gap reported in #4976, where a documented smoke test had no request shape and a credential field the connect dialog asks for was read by no product path.
- 🔺 `priority-high` — **kept.** Matches the priority on #4976 itself, and the gap blocks the outbound half of the Discord provider from being demonstrable as documented.
- ⚠️ `risk-high` — **kept.** `validateOutboundRecipient` is the single choke point for outbound recipients across every channel provider, so blast radius is the higher signal and the higher one wins over the "small, well-tested diff" reading. I verified the containment claim independently in the first pass — one production call site, and `recipientFormat` read by nothing but that helper — and the re-review delta is documentation-only, so nothing about that assessment moved. Recording the mitigation as evidence rather than using it to argue the label down.
- 🔒 `in-progress` — this reviewer's lock for the re-review pass; released with the completion comment below.

No `ci-monitoring`: CI is already settled and green on `71a586d29` (13 success, 5 skipped, zero pending), so there is no follow-up for it to promise.

pkarw:
🔄 **Handed back to @wojciechszyjka** — changes requested by @pkarw.

One **Medium** finding to clear, and it is documentation rather than code: the `recipientFormat` JSDoc on `packages/core/src/modules/communication_channels/lib/adapter.ts:52` still describes the flag as format-only, while this PR gives `'provider-native'` a second meaning — "and the hub will accept a request that names no recipient". That JSDoc is the cross-package contract provider authors read (`channel-discord/lib/capabilities.ts:1` imports it), so the next provider-native adapter inherits presence-optionality without opting in. Full reasoning, plus a suggested wording and an optional cleaner alternative, are in the review.

Two Low nits alongside it, neither blocking: the new `Default channel ID` help text points at a bodyless `POST` that 422s with a Zod dump, and the spec changelog cites the integration case without noting it skips in CI today.

Nothing else — the behavior in the diff is correct, the containment argument holds under independent checking, the tests genuinely pin rather than merely cover, and CI is green on `e2868b62e`.

**No changes were pushed to this branch.** Autofix is deliberately skipped: this is another author's PR and `--autofix` was not passed, so this reviewer does not touch the branch uninstructed. Ping for a re-review once the JSDoc lands — I do not expect a second round beyond confirming it.

pkarw:
🤖 `om-auto-review-pr` — completed: **CHANGES REQUESTED** (0 blockers, 0 majors, 1 medium, 2 low). Lock released.

Validation source: **GitHub checks** — all 13 reporting checks green on head `e2868b62e`, verified against the commit's own check-runs rather than the PR summary; 5 skipped by their own conditions, none pending, none failing. No local gate was run and none was needed.

autofix: skipped (not my PR — re-run with `--autofix` to fix it here). No commits were pushed to this branch.

No CI follow-up is scheduled: CI is already settled and green, so `ci-monitoring` was not applied. Remaining gates after the Medium is cleared are a re-review and `qa-approved` behind `needs-qa`.

wojciechszyjka:
## 🔁 All three findings addressed — `71a586d29`

Thanks, this was a careful read. Independently confirming the single call site and that nothing but the helper reads `recipientFormat` is exactly the check that makes the "email path is provably untouched" claim mean something rather than assert something.

### 1. Medium — `recipientFormat` JSDoc ✅ fixed

You put your finger on the real problem, and it is one I introduced: the PR's own root-cause section says *"presence and format were conflated"*, I un-conflated them in the helper, and then re-conflated them in the flag. The consumer-side docs were thorough and in the wrong place — a provider author reads `ChannelCapabilities`, not `outbound-recipient.ts`.

The JSDoc now states the second consequence, the obligation it creates, and the way out:

> **`'provider-native'` also opts the provider into presence-optionality, not only format.** The hub accepts an outbound request that names no recipient at all and forwards it with no `metadata.to` key, on the understanding that the adapter resolves its own configured target (Discord: `defaultChannelId`). An adapter declaring this therefore MUST resolve such a target, and MUST fail legibly — an operator-readable `SendMessageResult.error`, not a throw — when it has none, because that message is surfaced verbatim to whoever ran the send. Declare `'email'` (or omit the field) if your provider has no default destination: the hub then keeps answering `Recipient is required`, which is the correct outcome for a request with nowhere to go.

I added the "fail legibly, not a throw" clause because that obligation only became real in this PR — `022770238` reworded Discord's no-target error precisely because this change made that branch reachable from the product. A provider author needs to know that string is operator-facing.

**Not taking `hasDefaultOutboundTarget` here**, and agreeing it is the cleaner shape. Introducing a capability with exactly one implementor means designing the seam from a single data point; it belongs with whoever adds the second provider-native adapter, who will know whether the split is `hasDefaultOutboundTarget` or something else. Recorded in the JSDoc rather than left as tribal knowledge.

### 2. Low — help text pointing at a 422 ✅ fixed, but not the way you suggested

You are right that the copy makes this the first thing an operator hits, so it is on this PR to fix. Reworded to name the body: *"call the hub test-send endpoint with an empty JSON body (`{}`) so no recipient is named"*.

**I did not take the `readJsonSafe(req, {})` option, because it is not safe** — and the reason is worth spelling out since the note said it now is. `readJsonSafe` returns its fallback for **malformed JSON as well as an empty body**:

```ts
// packages/shared/src/lib/http/readJsonSafe.ts
const raw = typeof source === 'string' ? source : await source.text()
if (!raw) return fallback
try { return JSON.parse(raw) as T } catch { return fallback }   // ← also here
```

With `{}` as the fallback and every field now optional, a truncated payload — `{"to":` — would parse as an empty object and **succeed**, sending a test message to the provider's default channel instead of answering 422. That is the same class of failure this PR spends its careful boundary on refusing: `''` and `null` stay 422 precisely so a caller who meant to name a recipient is never silently rerouted. A caller whose JSON got cut off is in that same category. Widening presence must not also widen what counts as a well-formed request.

Making a truly bodyless POST work would need empty-body and malformed-body distinguished at the read, which is more than this nit is worth. Happy to file it if you think the bare `curl` ergonomics matter.

### 3. Low — the spec over-reading its own coverage ✅ fixed

Fair, and the sharper point is the one about durability: I put the disclosure in a PR comment, which nobody reads in eighteen months. The changelog bullet now carries it:

> …but it runs only where `OM_ENABLE_TEST_CHANNEL_SEEDING` is enabled, and that flag is set nowhere in the repository today, so it skips in CI. **The everywhere-guarantee is therefore the unit coverage** (…); the integration case is confirmation against a real seeded channel, not the primary guard.

Also worth flagging since it changes your conclusion's shelf life: **#5662** proposes enabling that flag in the integration harness. Writing it turned up why nobody had — enabling it fails 5 CRM-EMAIL specs on one root cause (`externalMessageId` never reaching the interactions response, which silently strips every email card's Reply/Forward/visibility actions), so it ships the fix alongside. The spec bullet references it.

### Validation

`build:packages`, `typecheck`, `i18n:check-sync` green; `@open-mercato/channel-discord` 259/259; `outbound-recipient` + `test-send` 49/49. The diff is three files, comments and copy only — no code path changed, so I did not re-run the full 8-command gate. Say the word if you would rather see it.

Re-review welcome.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T00:19:59Z — re-review of `71a586d29` against the three findings from my `e2868b62e` pass. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review — Re-review

**PR:** #5635 — `fix(communication_channels): let test-send fall back to the provider's own target (#4976)`
**Base:** `develop` · **Head:** `71a586d29` (was `e2868b62e`) · **Reviewer:** @pkarw · **Mode:** re-review — second pass by this reviewer, confirming the three findings from my `e2868b62e` review

## 🎯 Summary

This is the confirmation pass I said I expected, and it needs no third round. The delta since my last review is a single commit, `71a586d29`, touching three files — a JSDoc block, one `helpText` string, and one spec changelog bullet. **No code path changed**: `git diff e2868b62e 71a586d29` is 21 added lines across `lib/adapter.ts`, `channel_discord/integration.ts` and the Discord spec, all of them comments or copy. Everything I verified in the first pass — the single production call site for `validateOutboundRecipient`, `recipientFormat` being read by nothing else, the `undefined`-only omission boundary, the unchanged security ordering — is untouched and still holds.

All three findings are cleared, and two of them are answered better than I asked. I re-checked each against the code rather than against the reply.

## ✅ Verdict: **APPROVED**

Zero blockers, zero majors, zero mediums, zero remaining lows. My previous verdict was `changes_requested` driven solely by the one Medium; that Medium is resolved, so the verdict flips. This repository escalates Medium to `changes_requested` (`.ai/skills/om-auto-review-pr/SKILL.md` § Verdict rule), and with no findings at any severity the same rule lands on approve.

Note on the previous state: this PR carried **zero approving reviews** until now, because the author's own automation could not self-approve. Branch protection on `develop` requires `required_approving_review_count: 1`, so this review is the one that clears that particular gate. `needs-qa` still holds the merge behind `qa-approved`.

## 🧪 Validation Gate

**Validation source: GitHub checks.** No local fallback was run and none was needed.

Queried against the head commit's own check-runs (`repos/.../commits/71a586d298df0e9ce594d3ea7333d3e018dc9947/check-runs`) rather than the PR summary, so this is the head as pushed:

| Check | Result |
|---|---|
| `test` | ✅ success |
| `lint` | ✅ success |
| `ds-lint` | ✅ success |
| `prepare` | ✅ success |
| `ephemeral-integration (none)` | ✅ success |
| `docker-build` | ✅ success |
| `scope`, `audit-scope` | ✅ success |
| `CodeQL`, `Analyze (actions / javascript-typescript / python)` | ✅ success |
| `audit`, `mutate`, `comment`, `merge-coverage`, `[code]smith` | ⏭️ skipped by their own conditions |

18 check-runs reported: 13 success, 5 skipped, **zero pending, zero failing**. Branch protection on `develop` declares **no required status checks** (`required_status_checks` is absent from the protection payload), so per the tracker descriptor every reported check is treated as required — and every one of them is green.

I did not re-run the eight-command gate locally, and I agree with the author's decision not to either: the delta is comments and one help string, so no code path exists for it to exercise. The `test` job on this exact head is the stronger evidence anyway, since it ran the whole suite rather than a scoped subset.

**One disclosure about what that green tick covers.** CI ran on `71a586d29` at roughly 11:25Z; `develop` has since taken the v0.7.0 sync (`b41f7e3e5`, 12:16Z), and the branch is now **56 commits behind** its base rather than the 7 it was at my last review. I checked what that means rather than leaving it as a caveat:

- `git merge-tree --write-tree origin/develop 71a586d29` returns a **clean tree with no conflicts** — the merge is not just `MERGEABLE` per GitHub's cached field, it resolves cleanly against `develop` as of right now.
- Exactly one of those 56 commits touches this PR's directories: `05cbf1f77` (`fix(release): align develop-only packages with the merged 0.7.0 version`), and it moves only `packages/channel-discord/package.json`'s `.version` field. This PR touches no `package.json`, so there is no overlap and no version-alignment exposure.

So the tick is honest and the risk it does not cover is empirically nil. Not a finding — recorded because "CI was green" and "CI was green against the base it will actually merge into" are different claims, and only the second one is worth acting on.

## 🔁 Previous Findings — all resolved

### 1. 🟠 Medium — `recipientFormat` gained a second meaning its contract doc did not mention → **RESOLVED**

`packages/core/src/modules/communication_channels/lib/adapter.ts:63-73` now carries the second consequence in the place a provider author actually reads. The new paragraph states the widening (`'provider-native'` opts into presence-optionality, not only format), the obligation it creates (the adapter MUST resolve its own target and MUST fail legibly rather than throw), and the way out (declare `'email'`, or omit the field, and the hub keeps answering `Recipient is required`).

I checked the new JSDoc's normative claims against the code instead of taking them on trust, because a contract doc that is subtly wrong is worse than one that is merely incomplete. Every one of them holds:

- *"forwards it with no `metadata.to` key"* — `test-send/route.ts:243` spreads `...(body.to !== undefined ? { to: body.to } : {})`, so the key is genuinely absent, not `undefined`.
- *"an operator-readable `SendMessageResult.error`, not a throw"* — `SendMessageResult.error?: string` exists on the interface (`adapter.ts:100`), and `route.ts:253` returns it to the caller verbatim as `providerError`. The doc names a real field with a real surfacing path.
- *"the adapter resolves its own configured target (Discord: `defaultChannelId`)"* and the fail-legibly MUST — the only implementor complies: `channel-discord/lib/adapter.ts:88-92` returns `{ status: 'failed', error: 'No Discord channel to post to — set "Default channel ID" on the connection, or pass a channel id as the recipient.' }`. It returns rather than throws, and the string is operator-facing with no `[internal]` prefix.
- *"Declare `'email'` (or omit the field)"* — `validateOutboundRecipient` gates the omission branch on `capabilities?.recipientFormat === 'provider-native'`, which is false for both, so both keep `Recipient is required`.

The added *"fail legibly, not a throw"* clause was not in my suggested wording and is the better call: that obligation became real in this PR (`022770238` reworded the Discord string precisely because this change made the branch product-reachable), and a provider author who does not know the string is operator-facing will write an internal one.

Declining `hasDefaultOutboundTarget` is the right decision and I want it on the record as agreed rather than deferred-under-protest. Designing a capability seam from one implementor is how you get the wrong seam; recording the reasoning in the JSDoc rather than leaving it tribal is exactly what makes it safe to defer.

### 2. 🔵 Low — help text pointed at a request that 422s → **RESOLVED, and my suggestion was wrong**

`integration.ts:88` now reads *"To verify it, call the hub test-send endpoint with an empty JSON body (`{}`) so no recipient is named"*. I verified `{}` actually works end to end rather than assuming: `bodySchema` has all three fields optional (`route.ts:53-57`), and `readJsonSafe` parses `"{}"` normally, so the body reaches `validateOutboundRecipient(undefined, discordCapabilities)` and passes. The copy now describes a request that succeeds.

**Correcting my own note:** I wrote that `readJsonSafe(req, {})` "is now safe precisely because every field is optional". That is wrong, and the author is right to have refused it. `readJsonSafe` returns its fallback for **malformed** JSON as well as an empty body — the `catch` and the `if (!raw)` share one fallback (`readJsonSafe.ts:6-12`). With `{}` as the fallback, a truncated payload like `{"to":` would parse as an empty object and **succeed**, sending to the provider's default channel instead of answering 422. That is the same silent-reroute failure this PR spends its `''`/`null` boundary on refusing, and I would have widened it while reviewing the PR that narrowed it. Good catch on the reviewer.

I would not bother filing the empty-vs-malformed split as follow-up work; the copy fix removes the only reason anyone was going to hit it, and a `curl` ergonomics nicety is not worth touching a shared HTTP helper's failure semantics.

### 3. 🔵 Low — spec cited the integration case without noting it skips in CI → **RESOLVED**

The changelog bullet at `.ai/specs/2026-06-19-discord-communication-channel-integration.md:1313-1319` now carries the disclosure, and it does the thing I actually cared about: it names where the everywhere-guarantee really lives (the unit coverage) instead of just hedging the integration claim. The `#5662` cross-reference checks out — that PR exists and is exactly what the bullet says it is (`test(integration): enable channel seeding, and fix the regression it exposes`), so a future reader following the trail lands somewhere real.

The point about durability was the author's own and it is the right one: a disclosure in a PR comment is gone in eighteen months, and this one is now in the artifact that outlives the thread.

Re-run against the delta, which is documentation-only:

- [x] **Public types and signatures (Cat. 3)** — `ChannelCapabilities.recipientFormat` keeps its exact type (`'email' | 'provider-native'`, optional). The JSDoc above it grew; the contract's *shape* did not move, and the *behavior* it now documents shipped in the earlier commits of this same PR. Documenting an existing widening is not itself a change to the surface.
- [x] **API routes (Cat. 7)** — untouched by this commit; the `to`-becomes-optional widening from the earlier commits is the permissive direction and every existing caller is byte-identical.
- [x] **Everything else** — no DB schema, event IDs, widget spot IDs, DI keys, ACL features, notification IDs, CLI commands, or generated files touched by the delta.
- [x] **Spec requirement** — the touching spec's dated changelog entry is present and now accurate about its own coverage.

No deprecation protocol required.

## 🧪 Test Coverage

Unchanged by this commit and still adequate — the delta adds no behavior to cover. Recording the coverage story once more since finding 3 sharpened how it should be described:

- **The everywhere-guarantee is the unit coverage.** `lib/__tests__/outbound-recipient.test.ts` pins the omission accepted on provider-native and still `Recipient is required` across all five email-capability shapes; `test-send/__tests__/route.test.ts` asserts `expect('to' in metadata).toBe(false)`, so a regression reintroducing the key as `undefined` fails rather than passing silently; `channel-discord/lib/__tests__/adapter.test.ts` pins the operator-facing string, the absence of `[internal]`, and `captured.createMessage === undefined` separately. All three run in every environment and all are green in `test` on this head.
- **The integration case is confirmation, not the guard.** `TC-CHANNEL-EMAIL-HUB-001` runs only where `OM_ENABLE_TEST_CHANNEL_SEEDING` is enabled, which is nowhere in this repository today, so both its cases `test.skip()` in `ephemeral-integration`. The spec now says so. `ephemeral-integration (none)` passing is the affected-only shard name and is not evidence those cases ran — worth restating because it reads like a skip and is not one.

## 📋 Checklist Notes

- **The new JSDoc MUST does not strand the test harness.** `TestSeedChatChannelAdapter` also declares `recipientFormat: 'provider-native'` (`test-seed.ts:154`), so it inherits the newly-written obligation. It complies in the way a network-free stub should: `sendMessage` returns `{ status: 'sent' }` deterministically and never throws, so there is no "has no target" state for it to fail legibly from. Checked rather than assumed, because a contract whose only other implementor violates it on the day it is written is a contract nobody will believe.
- **`helpText` stays a plain string**, matching `channel-gmail` and `channel-imap`, neither of which routes provider connect-field copy through `t()`. The reword does not introduce a deviation, and `lint` plus the i18n checks are green on this head.
- **No `any`, no raw `fetch`, no DS tokens, no `.tsx`** — the delta is three files of prose.

## ⚠️ Scope Note

Unchanged from my first pass and still correct: `Refs #4976` rather than `Closes`. The `send-as-user` half is tracked in #5528, and § 6's *"click Test send in the channel detail UI"* control exists in no `.tsx`. Closing #4976 remains a maintainer decision once #5528 lands.

---

*Approved. The remaining gates are `qa-approved` behind `needs-qa` — QA routes are in the comment below, narrowed to what the author's live-app evidence could not reach.*

pkarw:

**Deliberately narrow.** The author already ran the hub-side routes against a live app and posted the measured output (the QA evidence comment on `e2868b62e`), and I have no reason to doubt it — the email-path assertions there are the ones the whole change rests on and they held end to end. Re-running them is fine but is not what this PR still needs. What no one has been able to exercise is the half that requires a **real connected Discord bot**, because it exists on nobody's dev machine and the seeded harness cannot substitute for it. That is P0 below. Everything else is a short confirmation pass.

**What this changes in one line:** `POST /api/communication_channels/channels/{id}/test-send` accepts a body with **no `to`**, and a provider whose adapter declares `recipientFormat: 'provider-native'` (today: Discord only) posts to its configured `Default channel ID`. Nothing else about the endpoint changed, and the commit under re-review (`71a586d29`) changed no code at all — only a JSDoc, one help string, and a spec bullet.

### P0 — The live Discord send, which nothing automated has ever covered

This is the entire point of the PR and the only part with no evidence behind it. It needs a real Discord bot connected to a real guild.

**Where to click**
- Backend → Settings → Communication channels → connect a **Discord** channel with a real bot token.
- On that connection, set **Default channel ID** to a text channel snowflake the bot can post in.
- Then: `POST /api/communication_channels/channels/{discordChannelId}/test-send` with body `{}`.

**What to verify**
- Body `{}` (empty JSON object, no `to`) → **200**, and a message actually appears **in the channel named by `Default channel ID`**. Check Discord itself, not just the response — a `200` with `status: 'failed'` is a different outcome and the response carries it in `providerError`.
- Body `{"to":"<a different channel snowflake the bot can post in>"}` → **200**, and the message lands in **that** channel, not the default. This is the pre-existing path; it must be unaffected.
- Body `{}` on a Discord connection with **`Default channel ID` left empty** → the response reports failure with the operator-readable string *"No Discord channel to post to — set "Default channel ID" on the connection, or pass a channel id as the recipient."* It must **not** show an `[internal]`-prefixed message, a stack trace, or a raw Discord API error.

**What can go wrong**
- A `200` that posts nothing — the response says sent, the channel is empty. The adapter returns `status: 'failed'` in `providerError` for real failures, so read the body, not just the HTTP code.
- The message landing in the wrong channel when `to` **is** supplied (the default silently winning over the explicit recipient) — that would be a regression in the pre-existing path.
- The no-target error surfacing as an internal-looking string. That string is operator-facing by contract now (it is written into the `ChannelCapabilities.recipientFormat` JSDoc as a MUST), so an ugly one is a real finding, not cosmetic.

### P1 — The `Default channel ID` help text, reworded in this commit

`71a586d29` changed this copy specifically because the previous wording told operators to call the endpoint "with no recipient", which read as a bodyless `POST` and 422s with a Zod dump. Worth ten seconds of confirming the new copy is both readable and true.

**Where to click**
- Backend → Settings → Communication channels → **Connect Discord** dialog → the **Default channel ID** field's help text.

**What to verify**
- It reads: *"Default text channel the adapter posts to when an outbound message names no Discord channel. To verify it, call the hub test-send endpoint with an empty JSON body (`{}`) so no recipient is named, or pass a channel snowflake as the recipient to post somewhere else."*
- Follow it literally with `curl -X POST … -H 'Content-Type: application/json' -d '{}'` → the send succeeds (per P0), i.e. the instruction is accurate as written.
- A **genuinely bodyless** `POST` (no `-d` at all) still answers **422**. That is expected and unchanged — it is precisely why the copy now names the `{}` body.

**What can go wrong**
- The text rendering with the backticks or braces mangled in the dialog, making `{}` unreadable — this is the one place `{}` is shown to a human.
- An operator following it and hitting a 422 anyway, which would mean the copy and the schema disagree.

### P2 — Spec accuracy (read-only, no clicking)

**Where to look**
- `.ai/specs/2026-06-19-discord-communication-channel-integration.md`, the changelog bullet around line 1313.

**What to verify**
- It now states that the integration case runs only where `OM_ENABLE_TEST_CHANNEL_SEEDING` is enabled, that the flag is set nowhere in the repository, and that the everywhere-guarantee is therefore the unit coverage.
- The `#5662` reference points at a real open PR proposing to enable that flag in the harness.

**What can go wrong**
- Nothing at runtime. This is here so a future reader does not over-read the integration case as CI coverage it does not have.

### Regression surface you do **not** need to re-run

Recorded so the pass stays scoped: the email-channel behavior (`{}` → 422 `Recipient is required`; `""`/`null` → 422; a snowflake on an email channel → 422 `Recipient must be a valid email address`; a valid address → sends) was measured against a live app by the author and is pinned by unit tests that run in every environment and are green on this head. Re-run it if you want belt-and-braces, but a failure there would be surprising and the P0 block is where the real uncertainty is.

### Pass/fail
- **Cannot obtain a Discord bot** → say so rather than passing P0 silently. That is a legitimate outcome and a maintainer's call on whether to merge on the strength of the unit coverage; it is not something QA should wave through.

pkarw:
🤖 `om-auto-review-pr` — completed: **APPROVED** (re-review of `71a586d29`; 0 blockers, 0 majors, 0 mediums, 0 lows). Lock released.

All three findings from my `e2868b62e` pass are resolved, and I verified each against the code rather than against the reply — the JSDoc's four normative claims (`metadata.to` genuinely absent, `SendMessageResult.error` real and surfaced verbatim as `providerError`, Discord's no-target path returning rather than throwing, omission refused for `'email'`/absent) all hold, and `{}` demonstrably reaches the accepting branch. One correction to my own earlier note is recorded in the review: my `readJsonSafe(req, {})` suggestion was unsafe and the author was right to refuse it.

**Validation source: GitHub checks** — 18 check-runs on head `71a586d29` queried from the commit itself: 13 success, 5 skipped by their own conditions, zero pending, zero failing. Branch protection declares no required status checks, so all reported checks are treated as required; all are green. No local gate was run and none was needed for a documentation-only delta.

I also checked what that tick does *not* cover, since the branch went from 7 to **56 commits behind** `develop` (the v0.7.0 sync landed after CI ran): `git merge-tree` against current `develop` resolves **clean**, and the single base commit touching these directories moves only `channel-discord/package.json`'s `.version`, which this PR does not touch. No hidden merge risk.

Labels: `review` → `merge-queue`; `needs-qa`, `bug`, `priority-high`, `risk-high` all kept, with reasoning in the label-rationale comment (updated in place). No `ci-monitoring` — CI is already settled, so there is no follow-up to promise.

autofix: skipped (not my PR — re-run with `--autofix` to fix it here). No commits were pushed to this branch, and none were needed.

**Remaining gate: `qa-approved` only.** This review clears branch protection's one required approval, which the author's own automation could not give itself. QA routes are in the comment above, deliberately narrowed to the live-Discord-bot path that neither the author's live-app evidence nor any automated test has ever been able to reach.

PR #5640, opened 2026-08-26
fix(sales): make line discountAmount a line total on both the read and write path

wojciechszyjka:
Tracking plan: .ai/runs/2026-08-26-sales-line-discount-amount-contract/PLAN.md
Tracking run folder: .ai/runs/2026-08-26-sales-line-discount-amount-contract/
Status: complete

Refs #3757
Source doc: `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` (spec merged in #5200; the twin issue #5019 is closed, so #3757 is the tracker)
Supersedes #5550, which implemented a partial § 1 and was parked pending the decisions this PR records.

## 🎯 Goal

Make `sales_order_lines.discount_amount` and `sales_quote_lines.discount_amount` mean the same thing on the read path and the write path — **the discount for the whole line** — so that recalculating a sales document is idempotent, and an order containing a discounted multi-unit line stops reporting a net subtotal inconsistent with its own gross.

## 🔍 Root cause

`buildBaseLineResult` named its local `discountPerUnit`, seeded it from `line.discountAmount`, and computed `discountTotal = clamp(discountPerUnit × quantity, …)`. Every writer stores a **line total** in that column, and all three entity→snapshot mappers fed the persisted value straight back in, so each recalculation multiplied the discount by the quantity again.

On the reproduction row in #3757 (`quantity 3`, `unit_price_net 85`, `discount_percent 5`) the stored discount grew `12.75 → 38.25 → 114.75 → clamped to 255`, at which point it equalled the line's whole subtotal and the line's net collapsed to `0` while its gross stayed correct — because gross is taken from the already-correct persisted `total_gross_amount`. With `quantity === 1` the extra multiplication is a no-op, which is why it went unnoticed.

A second defect sat in the same expression: `??` treats a stored `0` as a supplied amount, so a `discount_amount` of `0` silently suppressed a correct `discount_percent` sitting in the same row.

## ✅ The decisions this PR records

The spec had been gated at `Status: draft — design decision requested` since 2026-08-07. @wojciechszyjka approved it on 2026-08-26; the full rationale is in the spec's new `## Decision Record`.

| # | Decision |
|---|---|
| **D1** | The column means a **line total** — net, in the line's currency, quantity-inclusive. `sales_invoice_lines.discount_amount` stays outside the contract, caller-asserted and unenforced. |
| **D2** | **Percentage-first** precedence, a stored `0` treated as absent. Its cost is accepted and documented: a caller sending only an amount onto a row with an inherited percent loses it; the escape is `discountPercent: 0`. |
| **D3** | `discountAmount: 0` alongside a non-zero percent now **applies** the percentage. This *inverts* rather than drops existing behaviour, so it gets its own test and the loudest paragraph in `UPGRADE_NOTES.md`. |
| **D4** | § Alternatives **E is not adopted** — the § 3 type shape is identical either way, so E stays additively adoptable later. Acceptance criterion 10 is out of scope; criterion 9 is in scope and tested. |
| **D5** | The opt-in operator repair CLI is **deferred** → #5641. |
| **D6** | The duplicated mapper is **extracted**, not kept behind an equivalence test. |

## What Changed

- **`lib/calculations.ts`** — the discount resolution moved into `resolveLineDiscountTotal`: the percentage wins whenever it is set and non-zero; otherwise the amount is read according to its origin; otherwise zero. Clamp bounds are unchanged, and the percentage branch is algebraically identical to before, which is why all 13 pre-existing engine tests pass untouched.
- **`lib/types.ts`** — `SalesLineDiscountBasis` plus two optional, in-memory-only fields: `discountAmountBasis` (caller-supplied only, omitted ⇒ `'unit'`) and `discountAmountFromStoredRow` (mapper-set only). Two fields rather than one, because "a caller told us how to read this" and "we rebuilt this from a row" are different statements.
- **`lib/lineSnapshots.ts`** (new) — one shared `mapOrderLineEntityToSnapshot`/`mapQuoteLineEntityToSnapshot`, tagging every rebuilt snapshot as stored-row sourced, plus `resolveUpsertDiscountFields`.
- **`commands/returns.ts`** — its byte-for-byte duplicate of the mapper is **gone**. That duplicate is why creating or deleting a return persisted inflated order header totals regardless of how the lines were originally written — a path that needs no unusual integration shape to hit.
- **`commands/documents.ts`** — both upsert sites decide the discount's origin **per operand** instead of collapsing `caller ?? stored ?? 0` into one expression, and `?? 0` became `?? null` so "explicitly zero" stays distinguishable from "not supplied".
- **`data/validators.ts`** — `discountAmountBasis` on the shared `linePricingSchema` only. One edit reaches every order and quote line surface; `invoiceCreateSchema` is deliberately untouched, because invoice lines never pass through the engine and a basis there would be inert.
- **`components/documents/SalesOrderDraftLines.tsx`** — see the note below.
- **`UPGRADE_NOTES.md`** — all three behaviour changes, with the inverting one called out separately.

## ⚠️ A bug found in this PR's own first implementation

Worth stating plainly, because it is the most interesting thing that happened here and it is exactly the failure mode the spec warned about (§ 4, *"the easiest thing in the spec to get wrong"*).

The first version of the upsert change tagged the **edited** line correctly. But the upsert path rebuilds *every* line of the order and runs each one back through `createLineSnapshotFromInput`, which overwrote the origin with the caller default — so every **untouched** line silently lost its stored-row origin and was re-inflated by quantity on every write. The re-inflation had not been fixed; it had been moved one step further down.

`createLineSnapshotFromInput` is now origin-preserving, which makes all twelve of its call sites correct regardless of whether they are fed raw caller input or a re-mapped snapshot. The fix ships as its own commit with the test that fails without it.

## 🧪 Tests

Local gate, in the configured order: `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app`.

| Suite | What it locks in |
|---|---|
| `lib/__tests__/calculations.test.ts` (+15) | Stored line totals, caller bases, percentage-first, the D3 zero case, clamping, a five-pass round trip (the compounding runaway a single-pass test cannot catch), an idempotency table across quantity 1 and quantity > 1, and the document-level #3757 assertion. |
| `lib/__tests__/lineSnapshots.test.ts` (new, 17) | Acceptance criterion 9 from both sides: no mapper ever sets `discountAmountBasis`, and no request schema can populate `discountAmountFromStoredRow`. Plus the per-operand split, including that it never sets both origin fields. |
| `commands/__tests__/documents.line-discount-contract.test.ts` (new, 6) | The real `sales.orders.lines.upsert` command against the real calculation service — including the untouched-line regression above and the D2 inherited-percent decision. |
| `commands/__tests__/returns.line-discount-contract.test.ts` (new, 4) | The return flow whose command file held the duplicate mapper. |
| `__integration__/TC-SALES-5019-line-discount-idempotency.spec.ts` (new, 4) | Order lines, quote lines, the single-order GET display recalculation twice in a row, and a return create → delete round trip returning the order header totals to where they started. Self-contained fixtures, cleaned up in `finally`. |

**Every new test was negative-controlled against the unfixed code**, because a test that passes either way locks in nothing. 10 of the 15 engine tests fail against the unfixed engine; 2 of the 6 command tests fail without the origin-preservation fix; 2 of the 4 return tests fail without the shared mapper's stored-row tag.

That control also caught two tests of mine that were passing vacuously: a line carrying a percentage is healed by percentage-first no matter how its amount is tagged, so the assertions had to move onto **amount-only** lines before they could see the defect at all.

## 💥 Breaking Changes

Yes — behavioural, on an unversioned contract. Three cases, fully documented in `UPGRADE_NOTES.md`:

1. A percent plus an overriding amount → the percent now wins, the amount is dropped.
2. Only `discountAmount`, onto a line whose stored `discount_percent` is non-zero → the inherited percent wins, with nothing in the caller's own request to warn it.
3. `discountPercent: 12` with `discountAmount: 0` → the 12% is now **applied** where it used to be suppressed.

Case 3 is the dangerous one: it inverts rather than drops, and `discountAmount: 0` is precisely the workaround an integration would have built against this very defect — possibly on a unit price it had already discounted. Such a caller now discounts twice, and the totals come out *larger*, which is the direction nobody notices.

The type and schema additions are **ADDITIVE-ONLY** under `BACKWARD_COMPATIBILITY.md`: both new fields are optional and in-memory only, never persisted and never accepted from a request payload, so no deprecation bridge is required. There is **no schema change, no migration, and no new column** — `yarn db:generate` was not run.

## 🗄️ Existing data

Two of the three affected row shapes repair themselves on the next recalculation, because they still carry the percentage the discount derives from. The third — an inflated amount with **no** percentage — cannot self-heal, since nothing in the row records how many times it was multiplied. That is why this PR says `Refs #3757` rather than `Closes`: the reported symptom does not clear on every affected database. The opt-in repair tool is #5641.

## 📝 Note on `SalesOrderDraftLines.tsx`

The review of #5550 flagged `discountAmount: normalizeNumber(payload.discountAmount, 0) * quantity` as a second site holding the per-unit reading, decision-dependent. Under D1 it is **correct as written**: `payload` goes to the create API where a caller amount is read per unit, while `record` drives the grid, which shows the discount for the whole line — so the multiplication is the conversion between the two, not a duplicate of the engine's bug. It now stops when the caller has explicitly said the amount is already a line total, and carries a comment saying why, so nobody "fixes" it back.

## 📋 Follow-ups filed

- **#5641** — the deferred opt-in operator CLI for amount-only re-inflated rows (D5).
- **#5644** — `totalNetAmount` is accepted and validated by the line schemas, then never read by the engine. The spec's § Out of Scope records it as a same-root-cause finding with no upstream issue; this is that issue.

## 📋 Progress

See the Tasks table in the plan — that is the authoritative Step-status source. All 25 rows are `done`.

## Handoff & Notifications

- Live handoff: `.ai/runs/2026-08-26-sales-line-discount-amount-contract/HANDOFF.md`
- Notifications log: `.ai/runs/2026-08-26-sales-line-discount-amount-contract/NOTIFY.md`


wojciechszyjka:
🤖 Claiming this PR — starting `om-auto-create-pr-loop` run. Started: 2026-08-26T05:53:07Z.

This is the implementation run for `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md`, whose two gated design decisions were approved by @wojciechszyjka on 2026-08-26. The chain that reached here is `om-pr-autopilot` → `om-auto-implement-spec` → `om-auto-create-pr` → this engine; the step count (24, against a threshold of 20) is what selected the loop engine. Other auto-skills should skip this PR until the lock is released.

wojciechszyjka:
🤖 `om-auto-create-pr-loop` — checkpoint 1 verification

**Steps 0.1 – 3.4 of 24 are done.** The contract itself is implemented end to end; what remains is the command-level and integration test coverage, the display-side cleanup, and the upgrade note.

### 🧪 Validation (local runner — no compose `app` container is up in this worktree)

| Command | Scope | Result |
|---|---|---|
| `yarn install` | worktree bootstrap | ✅ exit 0 |
| `yarn build:packages` | all packages | ✅ exit 0 — 27/27 turbo tasks |
| `yarn generate` | app registries | ✅ exit 0 |
| `npx tsc --noEmit` | `@open-mercato/core` | ✅ exit 0, no diagnostics |
| `npx jest src/modules/sales` | the whole sales module | ✅ **97 suites / 722 tests passed**, 0 failed |

`yarn generate` initially failed with `CLI not built. Run "yarn build:packages" first` — which is precisely why the configured gate orders the build ahead of it. Worth stating plainly: an earlier invocation in this run was *recorded* as passing when it had not, because the captured exit code belonged to a trailing `echo` rather than to yarn. Every command above was re-run with its real status captured.

### 🔍 Negative control — do the new tests catch the bug?

A test that passes against the unfixed code locks in nothing, so the engine source was reverted to its pre-fix state and the new suite re-run against it:

```
Tests: 10 failed, 18 passed, 28 total
```

**10 of the 15 new tests fail against the unfixed engine** — the stored-line-total case, the percentage-first heal, the D3 zero case, the five-pass round trip, and the document-level #3757 assertion among them. The 5 that pass either way cover behaviour this change deliberately preserves: a caller-supplied amount still multiplying by quantity, and the no-discount case.

### What landed

- **Phase 0 — the spec.** Now `Status: approved — implementation in progress`, with a `## Decision Record` covering D1–D6 and the `## Implementation Plan` it had never carried. That missing section is the concrete reason no automation could implement this spec before today.
- **Phase 1 — the engine.** `buildBaseLineResult` delegates to a new `resolveLineDiscountTotal`: the percentage wins whenever it is set and non-zero, and a stored line total is never multiplied by quantity again. Clamp bounds unchanged; the percentage branch is algebraically identical, which is why all 13 pre-existing engine tests still pass untouched.
- **Phase 2 — one mapper.** `lib/lineSnapshots.ts` now owns the single `mapOrderLineEntityToSnapshot`, tagging every rebuilt snapshot `discountAmountFromStoredRow: true`. The byte-for-byte duplicate in `commands/returns.ts` is gone — that duplicate is what kept the three persisting return flows writing inflated order header totals regardless of how the lines were originally created.
- **Phase 3 — the upsert sites.** `discountAmountBasis` added to the shared `linePricingSchema` only, and a `resolveUpsertDiscountFields` helper applied at both upsert sites so a caller amount and a stored amount are tagged with their own origins rather than merged and tagged once.

### ⚠️ Three things review should look at deliberately

1. **The shared mapper is not a pure move.** The two duplicated copies differed: `returns.ts` guarded numeric input with `Number.isFinite`, `documents.ts` did not. The extracted version keeps the guarded one, so a `NaN` on the documents path now yields `0` instead of propagating. Unreachable from `numeric` columns in practice, and the safer direction — but it is a behaviour change, not a refactor, and it should be judged as one.
2. **Step 3.3 landed inside 3.2's commit.** The per-operand split is one shared helper applied at two call sites; splitting the commit would have meant landing an intermediate state where the order path called a helper the quote path duplicated — the exact shape this change exists to remove.
3. **Checkpoint 1 fired late**, at Step 3.4 rather than 1.1. The worktree had no `node_modules` and no generated registries, so no validation command could give a trustworthy answer until install → build → generate had all completed. Steps 0.1–1.3 were verified by targeted test runs only.

### Next

Phase 4 — command-level tests for both upsert paths, the upsert-existing case, the return create/delete round trip, and the § 3 producer invariant.

wojciechszyjka:
## 🏷️ Label rationale

- 🔍 **`review`** — the work is complete and the full configured gate is green, so this PR sits in the review stage of the pipeline. Pipeline labels are mutually exclusive and no other state applies: nothing blocks it, no reviewer has requested changes, and it is not approved yet.
- 🐛 **`bug`** — it fixes the defect reported in #3757: the totals engine re-applied a stored per-line discount once per unit, compounding across recalculations until a discounted multi-unit line's net collapsed to zero. The new `discountAmountBasis` field is machinery in service of that fix, not a new capability, so `bug` is the right category rather than `feature`.
- 🧪 **`needs-qa`** — the automated-verification exemption does **not** hold here, on two independent counts. The API surface changes: `discountAmountBasis` is added to the order and quote line request schemas, and three existing request shapes change meaning. And the diff touches a `.tsx` file outside tests (`components/documents/SalesOrderDraftLines.tsx`). Either one alone disqualifies the exemption. `skip-qa` would be wrong and is deliberately not applied.
- ⚠️ **`priority-high`** — **raised from the `priority-medium` that #5550 carried.** That PR's reasoning was that the defect is not reachable from the admin UI, which was true for the scope it covered. It is no longer sufficient: this PR also closes the return-flow path, where creating or deleting a return persists inflated order **header** totals for any consumer that uses returns, with no unusual integration shape required. Combined with the repo's own inference rule — money math rates high — medium understates it.
- 💥 **`risk-high`** — the change alters the observable meaning of a public API request field in order and quote money math, and touches a contract surface enumerated in `BACKWARD_COMPATIBILITY.md`. This is the same rating @pkarw applied to #5550 for the same reason, and it still holds.

**Not applied, deliberately:** `skip-qa` (see above — it would contradict `needs-qa`, and the two must never coexist); `qa-approved` / `qa-self-verified` (no manual QA has been performed, and an authoring run never awards its own QA approval); `qa` (that pipeline label means manual QA is *in progress* and is set by QA reviewers only).

wojciechszyjka:
🤖 `om-auto-create-pr-loop` completed: all 25 plan Steps are `done` and the full configured validation gate is green end to end — `build:packages` → `generate` → `build:packages` → `i18n:check-sync` → `i18n:check-usage` → `typecheck` → `test` (34/34 workspaces, 0 failures) → `build:app`, none skipped.

Lock handed off to `om-auto-review-pr` — the chain continues on this PR. The `in-progress` label is released for the duration of the review pass and will be reclaimed when it returns.

wojciechszyjka:
🤖 `om-auto-review-pr` taking over the chain lock — single authoritative code-review pass with `--autofix`. Started: 2026-08-26T07:10:17Z.

wojciechszyjka (commented):
# 🔍 Code Review — `fix(sales): make line discountAmount a line total on both the read and write path`

**Verdict: APPROVE** — 0 blockers, 0 majors, 2 minors (both fixed in this pass), 1 nit (declined, with a reason).

**Submitted as a comment, not an approval.** GitHub does not permit approving your own pull request, and this PR was authored by the same account the review runs as. An independent approving review is still required before it can merge; the pipeline label therefore stays at `review` rather than moving to `merge-queue`, because calling it merge-ready would misrepresent it.

**Validation source: a full local gate on this tree.** All eight configured `validation.commands` ran in order, none skipped: `build:packages` → `generate` → `build:packages` → `i18n:check-sync` → `i18n:check-usage` → `typecheck` → `test` → `build:app`. The test step covered 34/34 workspaces with 0 failures. Details and the exact ordering constraint are in `.ai/runs/2026-08-26-sales-line-discount-amount-contract/final-gate-checks.md`.

**CI is still running on this head.** Branch protection plus the QA-approval gate hold the actual merge; this review covers the work, not a green run. Checks still pending at the time of writing: `prepare`, `Analyze (javascript-typescript)`. No required check is failing. A follow-up comment will report the CI outcome.

---

## What this change gets right

The diagnosis is correct and the fix is placed well. Correcting the arithmetic inside the shared `buildBaseLineResult` covers the order path, the quote path, the display recalculation and the three persisting return flows at a single point — better than the per-mapper tagging the spec originally assumed, as @pkarw's review of #5550 already observed.

Three things raise this above a competent fix:

1. **The percentage branch is algebraically identical to before** (`percent/100 × unitNet × quantity` → `percent/100 × netSubtotalBeforeDiscount`, and those are the same number), which is why all 13 pre-existing engine tests pass untouched. The blast radius is genuinely confined to the branch that was wrong.
2. **The two origin fields are kept separate** rather than collapsed into one `basis` value. That separation is load-bearing, not tidiness: if a mapper set a basis, a stored line total would outrank `discount_percent` on every recalculation and both self-healing row buckets would silently stop healing — a failure acceptance criteria 1–3 cannot detect. Criterion 9's invariant test exists precisely to keep that from drifting, and it is present.
3. **Every new test was negative-controlled** against the unfixed implementation. That is the difference between coverage and the appearance of coverage, and in this PR it did real work — see below.

## 🧪 Test coverage — verified, not assumed

Each layer was re-run against a deliberately reverted implementation:

| Layer | Reverted | Result |
|---|---|---|
| Engine | `resolveLineDiscountTotal` → the original per-unit multiplication | **10 of 15 fail** |
| Upsert + delete commands | the origin-preserving branch in `createLineSnapshotFromInput` | **3 of 7 fail** |
| Return flow | the shared mapper's `discountAmountFromStoredRow` tag | **2 of 4 fail** |

Two of this run's own tests initially passed under that control and were rewritten. The cause is worth carrying forward for anyone testing this contract in future: **a line carrying a `discount_percent` is healed by percentage-first precedence no matter how its amount is tagged**, so any assertion about the amount's *origin* must be made on an amount-only line. A suite built only on percentage lines reads as thorough and detects none of this.

## Findings

### 🟡 Minor 1 — `lib/` reached up into `commands/` — **fixed in this pass**

`lib/lineSnapshots.ts` imported `cloneJson` from `commands/shared.ts`. That was the only `lib/ → commands/` import anywhere in the sales module, and it ran against the module's own established convention: `lib/shipments/snapshots.ts` defines a local `cloneJson` rather than reaching up into the command layer.

Rather than add a third copy of a three-line helper, `cloneJson` moved to `packages/core/src/modules/sales/lib/json.ts`, with `commands/shared.ts` re-exporting it so all seven existing importers are untouched. `lib/` now has no imports from `commands/` at all.

### 🟡 Minor 2 — the line-**delete** rebuild had no test — **fixed in this pass**

`sales.orders.lines.delete` rebuilds and recalculates every *remaining* line through the same `createLineSnapshotFromInput` path the upsert uses, so it carries the identical origin-preservation requirement. It was correct by construction but unpinned. A test now covers it, and it fails against the reverted implementation — one of the three noted above.

This gap mattered more than it looks. The bug this PR's own first implementation contained (below) lived in exactly that shared rebuild, and the delete path would have carried it just as the upsert did.

### ⚪ Nit 3 — a third `toNumeric` — **declined, deliberately**

`lib/lineSnapshots.ts` carries its own `toNumeric`, alongside the existing copies in `commands/documents.ts` and `commands/returns.ts`. Consolidating them means touching 20 call sites in `returns.ts` on a PR whose subject is a money-math behaviour change, which trades a real regression risk for a cosmetic gain. Left as is, and flagged rather than hidden.

Note the copies are **not** identical: the extracted mapper keeps the stricter `returns.ts` version, which guards with `Number.isFinite`, so a `NaN` on the documents path now coerces to `0` instead of propagating. Unreachable from `numeric` columns in practice and the safer direction, but it is a behaviour change riding along inside a refactor, and it should be read as one.

## 💥 Breaking-change review

`BACKWARD_COMPATIBILITY.md` is satisfied and the documentation obligation is met, which is the specific thing #5550 was faulted for.

- **Types and validators** — `SalesLineDiscountBasis` plus two optional `SalesLineSnapshot` fields, and one optional request field on the shared `linePricingSchema`. Both snapshot fields are in-memory only. I verified this rather than taking it on trust: `convertLineCalculationToEntityInput` builds the entity payload field by field with no spread, so neither field can reach a persisted column, and `loadOrderSnapshot` reads entities directly rather than through the mappers, so undo and audit payloads are unaffected. **ADDITIVE-ONLY**; no deprecation bridge required.
- **The behavioural break is real and documented.** Percentage-first precedence is not expressible as a type change, and `UPGRADE_NOTES.md` carries all three affected caller shapes with the `discountPercent: 0` escape. The `discountAmount: 0` case is called out separately and correctly: it *inverts* rather than drops, so an integration that used it to work around this very defect starts double-discounting, and the totals come out larger — the direction nobody notices.
- **Linkage** — `Refs #3757`, not `Closes`. Correct: amount-only rows with no percentage cannot self-heal, so the reported symptom does not clear on every affected database. #5641 tracks the repair tool.
- **No schema change, no migration, no snapshot edit**, and `yarn db:generate` was not run. The tree is clean after `yarn generate`.

## 🔍 Scope and conventions

No cross-module ORM relation added; every touched path keeps its `{ tenantId, organizationId }` scoping and the scoped `findWithDecryption` calls in `recalculateOrderTotalsForDisplay` are untouched. Document math stays inside `salesCalculationService` — nothing is recomputed inline at a call site. No `any` introduced, no raw `fetch`, no generated file edited, no user-facing string added. `invoiceCreateSchema` is deliberately left alone, which is right: invoice lines never pass through `buildBaseLineResult`, so a basis field there would be inert.

**Completeness check on the producer surface.** I enumerated every non-test caller of `calculateDocumentTotals` to confirm nothing feeds the engine a persisted row while bypassing the shared mapper. Three exist: `commands/documents.ts` and `commands/returns.ts`, both now on the shared mapper and the origin-preserving rebuild, and `seed/examples.ts`, which builds snapshots from in-memory example definitions rather than stored rows and supplies only `discountPercent`. No producer was missed.

## ⚠️ Worth a reviewer's attention

**This PR's own first implementation contained the bug it was fixing, one step further down.** The initial version of the upsert change tagged the *edited* line correctly, but the upsert rebuilds every line of the order and passes each through `createLineSnapshotFromInput`, which then overwrote the origin with the caller default — so every *untouched* line lost its stored-row origin and was still re-inflated on each write. It ships fixed, as its own commit, with the test that fails without it.

I am flagging it rather than quietly folding it into the history for two reasons. It is the precise failure the spec's § 4 warns is "the easiest thing in the spec to get wrong", which suggests the warning is well aimed. And it was found by writing the command-level test, not by reading the diff — a reviewer reading only the upsert hunk would have seen a correct-looking change.

## ⛔ Not verified here

- **The integration spec has not been executed.** `TC-SALES-5019-line-discount-idempotency.spec.ts` is typechecked and reviewed but not run: no provisioned Playwright environment exists for this worktree. CI's `ephemeral-integration` job is a required check and will run it against this head. Stated plainly because "tests added" and "tests run" are different claims.
- **No UI verification.** The single `.tsx` change is a numeric expression plus a comment in a code path that is unreachable today, so there is no rendered behaviour to capture. `needs-qa` nonetheless stays on, because the API surface changed — that alone disqualifies the automated-verification exemption.

## Verdict

Approve on the merits. The remaining gates are external: an independent approving review (self-approval is impossible), the required checks, and the `qa-approved` label that the hard QA gate demands while `needs-qa` is set.

wojciechszyjka:
🤖 `om-auto-create-pr-loop` — run summary

**Engine: om-auto-create-pr-loop (steps: 24, --loop: no)** — the drafted plan came to 24 Steps against a threshold of 20, so the loop engine was selected. The step count was not shaved to avoid it: this change genuinely spans types, the calculation engine, two command files, the validators, a component, the upgrade notes and three test layers. Two review-fix Steps were appended during the run, for 26 in total.

## 🎯 What this delivers

`sales_order_lines.discount_amount` and `sales_quote_lines.discount_amount` now mean the same thing on the read path and the write path — the discount for the **whole line** — so recalculating a sales document is idempotent. Before this, `buildBaseLineResult` read the column as a per-unit rate while every writer stored a line total, so each round trip multiplied the discount by the quantity again: `12.75 → 38.25 → 114.75 → clamped to 255` on the reported 3 × 85.00 line, at which point the discount equalled the whole line subtotal and the line's net collapsed to `0` while its gross stayed correct.

The spec describing all of this merged in #5200 back on 2026-08-07 and had been sitting at `Status: draft — design decision requested` ever since, because two decisions were reserved for a maintainer. Those were approved on 2026-08-26 and are now recorded in the spec's `## Decision Record`, together with four dependent calls that follow from them.

## 📋 Plan

All 26 Tasks rows are `done`. `.ai/runs/2026-08-26-sales-line-discount-amount-contract/PLAN.md` is the authoritative status source; `HANDOFF.md` and `NOTIFY.md` carry the live state and the decision log.

| Phase | Outcome |
|---|---|
| 0 — the spec | Status flipped to approved, `## Decision Record` (D1–D6) added, and the `## Implementation Plan` the spec had never carried. That absence is the concrete reason no automation could implement this spec before today. |
| 1 — types and engine | `SalesLineDiscountBasis` plus two optional snapshot fields; `resolveLineDiscountTotal` puts the percentage first and only multiplies by quantity where the value genuinely arrived per unit. |
| 2 — one mapper | The byte-for-byte duplicate in `commands/returns.ts` is gone. That duplicate is why creating or deleting a return persisted inflated order **header** totals regardless of how the lines were written — a path needing no unusual integration shape. |
| 3 — schema and upserts | `discountAmountBasis` on the shared `linePricingSchema` only; both upsert sites decide origin per operand, and `?? 0` became `?? null`. |
| 4 — command tests | The real upsert and delete commands against the real calculation service, plus the § 3 producer invariant. |
| 5 — display and upgrade notes | The draft-grid expression resolved and documented; all three behaviour changes written up. |
| 6 — integration | `TC-SALES-5019-line-discount-idempotency.spec.ts`, self-contained, covering order lines, quote lines, the display recalculation and a return create → delete round trip. |
| 7 — follow-ups | #5641 and #5644 filed; two review-fix Steps applied. |

## 🧪 Verification

The full configured gate ran in order with nothing skipped — `build:packages` → `generate` → `build:packages` → `i18n:check-sync` → `i18n:check-usage` → `typecheck` → `test` → `build:app` — and came out green end to end. The test step covered **34/34 workspaces with 0 failures**. The sales module alone is 100 suites / 750 tests.

Worth recording: `yarn generate` fails with `CLI not built` until `yarn build:packages` has run, which is exactly why the configured gate orders the build first. An earlier invocation in this run was briefly *recorded* as passing when it had not, because the captured exit code belonged to a trailing `echo` rather than to yarn. Every command was subsequently re-run with its real status captured.

**Every new test was negative-controlled against the unfixed implementation** — 10 of 15 engine tests, 3 of 7 command tests and 2 of 4 return tests fail without their respective fixes. Two tests initially passed under that control and were rewritten: a line carrying a `discount_percent` is healed by percentage-first precedence no matter how its amount is tagged, so assertions about the amount's origin only work on amount-only lines.

## 🔍 Review

`om-auto-review-pr --autofix` ran as the single authoritative pass: **0 blockers, 0 majors, 2 minors both fixed in the pass, 1 nit declined with a reason**. Submitted as a comment rather than an approval because GitHub does not allow approving your own PR — an independent approving review is still required, which is why the pipeline label stays `review` rather than moving to `merge-queue`.

## ⚠️ The most important thing in this PR

**Its own first implementation contained the bug it was fixing, moved one step further down.** The initial upsert change tagged the *edited* line correctly, but the upsert rebuilds every line of the order and passes each through `createLineSnapshotFromInput`, which overwrote the origin with the caller default — so every *untouched* line still lost its stored-row origin and was re-inflated on each write.

It ships fixed, as its own commit, with the test that fails without it. It is called out here rather than folded quietly into the history because it is precisely the failure the spec's § 4 warns is "the easiest thing in the spec to get wrong", and because it was found by writing the command-level test, not by reading the diff.

## ⛔ Not verified in this run

- **The integration spec has not been executed.** It is typechecked and reviewed, but no provisioned Playwright environment exists for this worktree; CI's `ephemeral-integration` job is a required check and will run it. "Tests added" and "tests run" are different claims and are kept apart here.
- **No UI verification.** The single `.tsx` change is a numeric expression plus a comment in a code path unreachable today, so there is no rendered behaviour to capture. `needs-qa` stays on regardless, because the API surface changed.

**CI is still running on this head.** Branch protection and the QA-approval gate hold the actual merge; nothing in this run's verdict is a substitute for either. Checks pending at the time of writing: `prepare`, `Analyze (javascript-typescript)`. No required check is failing. A follow-up comment will report the CI outcome.

## 🚀 State

Ready for review, not merge-ready. Three gates remain, all external to this run: an independent approving review, the required checks going green, and `qa-approved` — while `needs-qa` is set and the repo's QA gate is on, this PR **must not** merge without it.

wojciechszyjka:
🤖 `om-auto-review-pr` completed: **APPROVE** (0 blockers, 0 majors, 2 minors fixed in the pass, 1 nit declined). Submitted as a comment because GitHub does not allow self-approval, so an independent approving review is still required and the pipeline label stays `review`.

Lock released: `in-progress` has been swapped for `ci-monitoring`, so this PR is free for any other skill or human to act on. `ci-monitoring` is a meta label, not a claim — it means only that a CI-result follow-up comment is still owed here.

wojciechszyjka:

**Verified green on `22235018d`** — `CI for Develop&Main` [completed / success](https://github.com/open-mercato/open-mercato/actions/runs/32942045816), `Mutation tests` [completed / success](https://github.com/open-mercato/open-mercato/actions/runs/32942045849). **14 checks pass, 5 skipped, 0 failing.**

`Analyze (actions)` · `Analyze (javascript-typescript)` · `Analyze (python)` · `CodeQL` · `audit-scope` · `docker-build` · `documents-multi-instance` · `ds-lint` · `ephemeral-integration (none)` · `license/cla` · `lint` · `prepare` · `scope` · `test`

The verdict from the review is unchanged: the approve stands, and the pipeline label stays `review`.

## 📸 The gap in that review is now closed

The review disclosed that `TC-SALES-5019-line-discount-idempotency.spec.ts` shipped **typechecked but not executed**, because no Playwright environment exists in the authoring worktree. `ephemeral-integration` has now run it against a real database.

A passing job is not on its own proof of that, though — the spec self-skips when the tenant lacks `sales.orders.manage`, so four skipped tests would also produce a green job. The job log was therefore checked directly, and all four **executed**:

```
✓ 177 … › a percentage-discounted order line survives repeated upserts unchanged (370ms)
✓ 178 … › the single-order GET recalculation returns the same totals twice in a row (289ms)
✓ 179 … › a percentage-discounted quote line survives a re-upsert unchanged (258ms)
✓ 180 … › creating and deleting a return leaves the order header totals where they started (408ms)
```

The fourth one matters most. Acceptance criterion 6 — a return created and then deleted leaving the order header totals byte-identical — is the path `commands/returns.ts` used to walk with its own copy of the mapper, and the unit harness could only cover it partially without asserting the shape of its own mocks. It now passes end to end against a real database.

## ⚠️ One correction to this run's own reporting

An earlier bounded waiter in this run reported CI as **settled** when it was not. It polled `gh pr checks`, saw nine checks with none pending, and concluded the run had finished — but those nine were only the CodeQL / scope / mutation / CLA set, and `CI for Develop&Main` had simply not registered its jobs as checks yet. "No pending" meant "not scheduled", not "done".

**No green was reported off that signal.** The waiter was rewritten to key on the terminal conclusion of the workflow run for the head, which cannot be satisfied by jobs that do not yet exist, and the green above comes from that corrected check. Recording it because a fabricated green on a money-math PR behind a hard QA gate is worse than no report at all — and because the same trap is easy to walk into with any check-shape heuristic.

## 🔁 The commit that landed after that green run is also green

`85b265670` — **docs only**, a single file (`.ai/runs/…/NOTIFY.md`) recording the correction above. No source, test, or config file is touched; `git diff 22235018d..85b265670` is that one path.

It re-triggered CI, and this comment originally said no further report would come from this agent. That turned out to be unnecessarily pessimistic: the run settled inside the follow-up window and is green too. **`CI for Develop&Main` and `Mutation tests` are both completed / success on `85b265670`, the PR's final head — 14 checks pass, 5 skipped, 0 failing.**

This comment is updated in place rather than followed by a second one, so the PR carries one accurate CI record instead of a stale caveat beside a fresher fact.

## 🚀 Where this leaves the PR

Not merge-ready, and three gates remain, none of which this run can satisfy:

1. **An independent approving review** — GitHub does not permit self-approval, so the review above was submitted as a comment.
2. **Required checks green on the final head.**
3. **`qa-approved`** — while `needs-qa` is set and this repository's QA gate is on, this PR **must not** merge without it.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-26T08:58:56Z. Other auto-skills will skip this PR until the lock is released.

Independent review pass: the PR's own `om-auto-review-pr` run at 07:16Z could not submit a review (GitHub blocks self-approval) and reviewed `22235018d`; two commits have landed since. This pass reviews the current head `b8d8266d` from a different account, so the verdict lands in the Reviews timeline.

pkarw (approved):
# 🔍 Code Review: fix(sales) — `sales_*_lines.discount_amount` is a line total, and the percentage wins (#3757)

## 🎯 Summary

This PR fixes a compounding money defect: `buildBaseLineResult` read `discount_amount` as a per-unit rate while every writer stored a line total, so each recalculation multiplied the discount by the line quantity again until it equalled the line subtotal and the line's net collapsed to `0` while its gross stayed correct. The fix moves discount resolution into `resolveLineDiscountTotal` (`lib/calculations.ts:88-107`), which takes the percentage whenever it is set and non-zero, and otherwise reads the amount according to a per-snapshot origin: `discountAmountFromStoredRow` (mapper-set) means "already a line total", `discountAmountBasis` (caller-set) means "a caller told me how to read this". Two fields rather than one is the right call and the type comments at `lib/types.ts:35-72` say why.

I reviewed the whole diff at head `b8d8266d` — the engine, both upsert sites, the extracted mapper, the returns flow, the request schema, the display helper, the spec, `UPGRADE_NOTES.md`, and all four test layers — plus the surrounding call sites the change does *not* touch, to check the invariant holds everywhere a line snapshot is built. It does. **Three things are notably good beyond the fix itself.** The extraction of `mapOrderLineEntityToSnapshot` into `lib/lineSnapshots.ts` (D6) removes the duplication that is the mechanical reason `commands/returns.ts` kept the defect after the order flows were fixed — a return create/delete persisting inflated order **header** totals is a path that needs no unusual shape to hit, and it is now covered. The `UPGRADE_NOTES.md` entry is unusually honest: it names the `discountAmount: 0` case as an *inversion* rather than a dropped value, states plainly that it fails in the direction nobody notices, and it does not pretend all corrupted rows heal — the amount-only-no-percent bucket is explicitly marked "does not heal" with the recovery SQL and the deferred tool (#5641). And the PR body documents a bug the implementation found in its own first draft (untouched lines losing their stored-row origin on rebuild), which is now pinned by a dedicated test rather than left as a war story.

The design questions this depends on were not decided here: `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` had been gated at `Status: draft — design decision requested` since 2026-08-07 and @wojciechszyjka approved D1–D6 on 2026-08-26. I reviewed against those decisions as given.

## Verdict

✅ **approve** — I found no blockers, no majors and no minors. The three nits below are all optional follow-ups: none of them disputes the diff, changes a persisted value, or affects the merge. Under this repository's stricter rule (Medium/minor findings request changes) this is a clean approve, not a borderline one. `merge-queue` with `needs-qa` retained; the QA-approval gate still holds the merge until a QA reviewer adds `qa-approved`.

## 🧪 Validation Gate

**Validation source: GitHub checks** (this repo's checks-first extension — no local gate was run, and this worktree has no `node_modules`). All 19 reported checks on head `b8d8266d` are green: 14 pass, 5 skipped by design (`[code]smith`, `audit`, `mutate`, `comment`, `merge-coverage`). I opened the `prepare` and `test` **step** lists individually rather than trusting job colour, so nothing short-circuited behind an early red step.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` step 12 *Build packages* — [job 98111991458](https://github.com/open-mercato/open-mercato/actions/runs/32947720445/job/98111991458) |
| `yarn generate` | ✅ PASS | `prepare` step 13 *Prepare generated modules* |
| `yarn build:packages` (2nd) | ✅ PASS | `prepare` step 14 *Rebuild packages with generated files* — the repeated build is load-bearing, and it ran |
| `yarn i18n:check-sync` | ✅ PASS | `test` step 17 *Check i18n sync* — [job 98112991365](https://github.com/open-mercato/open-mercato/actions/runs/32947720445/job/98112991365) |
| `yarn i18n:check-usage` | ✅ PASS | `test` step 18 — noted for completeness that this step is `continue-on-error: true` at `ci.yml:636`, so its green is weaker than the others; irrelevant here, since the diff adds no user-facing strings |
| `yarn typecheck` | ✅ PASS | `test` step 21 *Checking types* |
| `yarn test` | ✅ PASS | `test` step 22 *Test*, plus step 23 *Test scripts*, step 24 DS-token drift, step 25 create-app parity and step 26 repo-wide audit guards — all three unfiltered guards green |
| `yarn build:app` | ✅ PASS | `prepare` step 15 *Build app*, and 15/15 `ephemeral-integration` scope satisfied by `ephemeral-integration (none)` (12m57s) — `(none)` is the affected-only shard name, not a skip |

`Mutation tests` ([run 32947720440](https://github.com/open-mercato/open-mercato/actions/runs/32947720440)) reports `success`, but see 🧪 Test Coverage — it carries no information about this change.

## Findings

### 💅 Nit

**1. `components/documents/SalesOrderDraftLines.tsx:60-62` — the new basis branch has no reachable caller and no test.**

The branch is correct as written, but nothing can reach it. `LineItemDialog` builds its submit payload explicitly at `LineItemDialog.tsx:1502-1534` and includes neither `discountAmount` nor `discountAmountBasis` (the dialog has no discount input at all), and the only other caller — the inbox pre-fill at `SalesDocumentForm.tsx:1380-1390` — supplies neither either. So `payload.discountAmount` is always `undefined`, `normalizeNumber(undefined, 0)` is `0`, and both arms evaluate to `0`. The pre-existing `* quantity` was equally dead, so this changes nothing observable and cannot regress anything — which is exactly why it needs a test to stay correct: `__tests__/SalesOrderDraftLines.localeFormatting.test.tsx` is the file's only coverage and contains no discount assertions. Two cases pinning `{discountAmount: 5, quantity: 3}` → `record.discountAmount === 15` and `{discountAmount: 15, discountAmountBasis: 'line', quantity: 3}` → `15` would make the conversion self-documenting.

Worth flagging for the QA reviewer too: this is the only `.tsx` outside tests in the diff, and it is not reachable from any screen, so the `needs-qa` on this PR is in practice API QA rather than UI QA.

**2. `__integration__/TC-SALES-5019-line-discount-idempotency.spec.ts` — `discountAmountBasis`, the one new public request field, is never exercised over HTTP.**

I verified it is wired end to end and does reach the command: `linePricingSchema` (`data/validators.ts:343`) → `orderLineCreateSchema`/`quoteLineCreateSchema` → `makeSalesLineRoute.ts:128` (`upsertSchema = createSchema.extend({...})`) → `parsed.discountAmountBasis` at `commands/documents.ts:7231` and `:7728`. It is also unit-pinned in `lib/__tests__/lineSnapshots.test.ts` and `lib/__tests__/calculations.test.ts`. But the integration spec exercises percentage discounts only (`DISCOUNT_PERCENT = 10`), so nothing asserts the field survives the route boundary — if a future route-local schema narrowed or omitted it, every current test would stay green while the documented `UPGRADE_NOTES` example silently stopped working. One case would close it: POST a line at `quantity: 60, unitPriceNet: 50` with `discountAmount: 300, discountAmountBasis: 'line'`, assert `discount_amount === 300` (rather than `18000` clamped to `3000`) and `total_net_amount === 2700`, then re-upsert and assert neither moves.

**3. `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md:3` — the `Scope:` header drifted from what shipped.**

It still lists the five files the 2026-08-21 draft anticipated (`lib/calculations.ts`, `lib/types.ts`, `commands/documents.ts`, `commands/returns.ts`, `data/validators.ts`), while the implementation also adds `lib/lineSnapshots.ts` and `lib/json.ts` and edits `commands/shared.ts` and `components/documents/SalesOrderDraftLines.tsx`. The § Implementation Plan added in this PR is accurate about all of them; only the header line is stale. Worth a one-line fix while the spec is still being moved to `implemented/` after merge.

## ✅ Checked and explicitly *not* findings

Recorded so the next reviewer does not re-derive them.

1. **`PUT /api/sales/orders` still reads an echoed amount per unit — and that is an approved scope decision, not a regression.** The order/quote update commands accept a full `lines` array (`data/validators.ts:724-728`, `:735-739`) and rebuild every line through `createLineSnapshotFromInput` as caller input, so a GET→PUT echo of an *amount-only* discounted line still multiplies by quantity. I checked this against `develop`: pre-PR the stored amount also won and was also multiplied, so the behaviour is byte-for-byte identical and nothing regresses. For *percent-bearing* lines the same echo becomes idempotent for the first time, because the percentage now wins. Closing the amount-only case fully is § Alternatives E, which D4 explicitly declines while keeping it additively adoptable, with acceptance criterion 10 declared out of scope by the approving maintainer.
2. **Neither origin field can be forged from a request.** `discountAmountFromStoredRow` appears in no zod schema and `z.object` strips unknown keys, so `createLineSnapshotFromInput`'s `isStoredRowSourcedLine` check cannot be driven by a caller. The only two `.passthrough()` calls in `data/validators.ts` (lines 39 and 44) are on `customerSnapshotSchema`, not on any line schema.
3. **Quote→order conversion and invoice-line creation cannot re-inflate.** Both copy the persisted column verbatim and never touch the engine (`commands/documents.ts:6539` and `:9073`), so D1's "invoice lines stay outside the contract" holds, and a converted order inherits a line total as a line total.
4. **Undo does not re-inflate.** `orderLineUpsertCommand.undo` restores through `restoreOrderGraph` with raw column values rather than replaying the payload through the calculation engine, so an undone upsert cannot re-run the multiplication.
5. **Healing is arithmetically sound on taxed lines, not just the zero-tax reproduction.** `mapPersistedLine` coerces `taxAmount` and `totalGrossAmount` to numbers, so both count as explicit and stay frozen across every rebuild — which means they were never corrupted in the first place. Restoring net alone therefore lands back on a consistent `net + tax == gross`, and the same freeze is what makes the detection heuristic in `UPGRADE_NOTES.md` work.
6. **The shipped-line pricing guard still compares like for like.** `SHIPPED_LINE_NUMERIC_PRICING_FIELDS` (`commands/documents.ts:6826`) includes `discountAmount`, and `resolveUpsertDiscountFields` returns the stored amount unchanged when the caller sends nothing, so `hasNumericChange(existing, existing)` is still false. The `?? 0` → `?? null` change only affects the no-caller-value-and-no-existing-line case, where the guard does not apply.
7. **The blast radius is contained.** `salesCalculationService` and `calculateDocumentTotals` have no consumers outside `packages/core/src/modules/sales/` anywhere in this repo, and the `external/official-modules` submodule is not activated here.
8. **The 13 pre-existing engine tests really are untouched** — `lib/__tests__/calculations.test.ts` is `+183/−0`, and the percentage branch is algebraically identical to the old one (`percent/100 × unitNet × quantity` ≡ `percent/100 × netSubtotalBeforeDiscount`), which is why they pass unchanged.

## 💥 Breaking Changes

This PR *does* change behaviour on a public API surface, deliberately and with a deprecation-free window, which is the right call because the old behaviour was the defect. The contract-surface checks:

- [x] No exported/public symbol removed or renamed without a deprecation path — `cloneJson` moved from `commands/shared.ts` to `lib/json.ts` but is re-exported from its original module (`commands/shared.ts:52`), so every existing import path still resolves.
- [x] No function signature changed in a breaking way — `resolveLineDiscountTotal`, `mapOrderLineEntityToSnapshot`, `mapQuoteLineEntityToSnapshot` and `resolveUpsertDiscountFields` are all new; nothing existing changed shape.
- [x] No required type field removed or narrowed — `SalesLineDiscountBasis`, `discountAmountBasis` and `discountAmountFromStoredRow` are additive and optional on `SalesLineSnapshot`.
- [x] No field removed or retyped in an existing response shape — `discountAmountBasis` is request-only, never persisted and never returned.
- [x] No event or message name renamed; no payload field removed.
- [x] No database table or column renamed, removed or narrowed — no migration ships, correctly: the column's *meaning* is being made normative, not its type.
- [x] Where a contract had to change, the old surface keeps working with migration notes — a caller that sends no `discountAmountBasis` gets exactly the per-unit meaning the API has always documented, and the three genuine behaviour changes (percent beats a supplied amount; percent beats an amount on a row with an inherited percent; `discountAmount: 0` alongside a percent now applies the percent) are each in `UPGRADE_NOTES.md` with the `discountPercent: 0` escape, the inverting case called out separately.

## 🧪 Test Coverage

**What is covered, and how.** The 15 new engine cases in `lib/__tests__/calculations.test.ts:580-760` pin each arm of `resolveLineDiscountTotal` separately — stored line total, caller per-unit, explicit `'line'` basis, explicit `'unit'` basis, percentage-beats-stored-amount (the healing case), zero-amount-with-a-percentage (the D3 inversion), the clamp, and a five-iteration idempotency loop that feeds the persisted result back in. `lib/__tests__/lineSnapshots.test.ts` pins the § 3 producer invariant directly — that mappers set `discountAmountFromStoredRow` and never a basis — using exact `toEqual` object literals rather than `objectContaining`, so a re-added field fails the assertion instead of slipping through.

The command tests are the ones that carry the weight, and they are not mocked at the layer that matters: `commands/__tests__/documents.line-discount-contract.test.ts` drives the **real** `sales.orders.lines.upsert` and `sales.orders.lines.delete` handlers against a real `DefaultSalesCalculationService` (`:167`), mocking only the EM, i18n, cache and side effects. That is what lets `does not re-inflate the discount of a line the caller never edited` (`:260`) and `does not re-inflate the remaining lines when another line is deleted` (`:447`) actually pin the origin-preservation bug the PR describes finding in its own first draft — the exact failure mode a naive single-line test would pass over. `commands/__tests__/returns.line-discount-contract.test.ts` covers the deleted duplicate mapper's four consumers, including repeated returns not drifting the header discount total.

**Where the signal is weaker than it looks — worth knowing, not this PR's to fix.** The green `Mutation tests` check carries no information here: `scripts/stryker/scope.mjs:36` freezes `ALLOWLISTED_PACKAGES` to `['shared']`, so a `packages/core` diff always resolves to an empty matrix and the `mutate` job is skipped (`has_work == 'false'`) — which is exactly what happened on this head and on `22235018d`. Arithmetic like `resolveLineDiscountTotal` is precisely what that harness exists for, so this is a gap in the repository's coverage rather than in the PR; the author's own recorded negative control (reverting the engine and observing 10 of 15 new tests fail) is currently the only substitute, and it is the right instinct. The two remaining gaps are nits 1 and 2 above.

**Integration.** `__integration__/TC-SALES-5019-line-discount-idempotency.spec.ts` is self-contained as the repo requires — every fixture is created through the API and torn down in `finally` (`:120`, `:166`, `:242`, `:321`), with no dependency on seeded or demo data — and it drives the real HTTP surface across four cases: the order-line re-upsert, the single-order GET recalculation twice in a row, the quote-line re-upsert, and return create-then-delete leaving the order header byte-identical.

pkarw:
## 🏷️ Label rationale

Posted by the independent `om-auto-review-pr` pass (@pkarw). The author's own rationale comment above stands — every label it argued for is re-derived here and kept, with one pipeline transition.

- 🚦 **`review` → `merge-queue`** — the only change. The code review is approved with 0 blockers, 0 majors and 0 minors, and the PR now carries a real approving review from a second account, so branch protection's review requirement is genuinely satisfied rather than nominally. Pipeline labels are mutually exclusive, so `review` is removed. **This does not mean it can merge:** `needs-qa` is present without `skip-qa`, so with `qaGate` on, the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- 🧪 **`needs-qa`** — kept, and I re-checked the automated-verification exemption clause by clause rather than inheriting the conclusion. It does not hold: the API surface changes (`discountAmountBasis` is added to the order and quote line request schemas, and three existing request shapes change meaning — see `UPGRADE_NOTES.md`). The `.tsx` in the diff is a second disqualifier on paper, but I want to be honest with whoever picks up QA: `components/documents/SalesOrderDraftLines.tsx` is the only `.tsx` outside tests, and its changed branch is not reachable from any screen, because `LineItemDialog` has no discount input. So the QA that matters here is API QA, not click-through QA — the instructions comment below is written accordingly.
- 🐛 **`bug`** — kept. This fixes #3757. The new `discountAmountBasis` field is machinery in service of the fix and an escape hatch for callers who need the new meaning explicitly, not a new capability, so `bug` beats `feature`.
- ⚠️ **`priority-high`** — kept, and I agree with the author's raise from the `priority-medium` #5550 carried. Silent money corruption that compounds on every write is bad enough; the deciding factor is that the return flow persists inflated order **header** totals through a path that needs no unusual shape to hit, so the "not reachable from the admin UI" argument that justified `medium` on the narrower PR no longer covers the blast radius.
- 🔥 **`risk-high`** — kept. The change edits the totals engine every order and quote line passes through, and it deliberately *inverts* one existing behaviour (`discountAmount: 0` alongside a non-zero percent now applies the percentage), which fails in the direction that produces larger totals rather than smaller ones. `risk-medium` would read as "ordinary single-module change" and understate that.
- ❌ **`qa-approved` deliberately not applied** — code-review approval is not QA approval, and it is not mine to grant from reading a diff.

`in-progress` is this run's lock and is released in the completion comment below.

pkarw:

This PR is approved and requires manual QA (`needs-qa`, no `skip-qa`). It is queued in `merge-queue` but the QA-approval gate holds it until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` by swapping the labels (remove `merge-queue`, add `qa`), then run the areas below.

**Read this first — it changes how you should approach the run.** There is no discount input anywhere in the admin UI. `LineItemDialog` builds its submit payload without `discountAmount` or `discountPercent`, so you cannot create the state under test by clicking. Every P0 area below therefore sets up its line through `PUT`/`POST /api/sales/order-lines` (or the quote equivalent) and then **verifies the result in the browser**. The one `.tsx` in this diff (`SalesOrderDraftLines.tsx`) has a changed branch that no screen can reach, so do not go hunting for it — that is expected, not a miss.

Numbers below are chosen so a per-unit misreading is unmistakable: `quantity 3 × unitPriceNet 85.00` at `discountPercent 5` should give a **12.75** line discount and a **242.25** net. If you ever see 38.25, 114.75 or a net of 0.00, that is the defect.

### P0 — Order line discount is idempotent across repeated writes

**Where to click**
- `POST /api/sales/orders` (or use any existing draft order) → note the `orderId`
- `PUT /api/sales/order-lines` with `{ orderId, quantity: 3, unitPriceNet: 85, discountPercent: 5 }`
- `/backend/sales/orders/{orderId}` → **Items** section and the **Totals** panel

**What to verify**
- After the first write, the line shows a **12.75** discount and a **242.25** net; the order's discount total is 12.75 and its subtotal net is 242.25.
- Re-send the *same* `PUT` with only a harmless field changed (for example the line `comment`, or nothing at all but the same `id`) three times in a row, reloading `/backend/sales/orders/{orderId}` after each → **every number stays exactly where it was.** This is the whole point of the change; a moving discount is a fail.
- Reload the order detail page several times without writing anything → the totals must not drift. The single-order GET triggers a display recalculation.
- Repeat the same sequence on a quote via `PUT /api/sales/quote-lines` and `/backend/sales/quotes/{quoteId}`.

**What can go wrong**
- The discount climbing `12.75 → 38.25 → 114.75 → 255.00` across writes, with the line's net falling to `0.00` while its gross stays right — that is the original #3757 symptom returning.
- Only the *edited* line staying stable while other lines on the same order drift. Put **three** lines on the order, discount two of them, then edit the third: the two untouched discounted lines must not move. This is the specific bug the PR reports finding in its own first implementation, so it is the highest-value check on this page.
- Totals in the header panel disagreeing with the sum of the visible line rows.

### P0 — Return create/delete leaves order header totals where they started

**Where to click**
- Same order, with at least one discounted multi-unit line and one undiscounted line
- `/backend/sales/orders/{orderId}` → **Returns** section → create a return against the *undiscounted* line
- Then delete that return

**What to verify**
- Record the full Totals panel (subtotal net/gross, discount total, tax total, grand total) **before** creating the return.
- Create the return → the discounted line's own net and discount must be **unchanged**; only the returned line's adjustment should move the totals.
- Delete the return → the Totals panel must return to **byte-identical** values to the ones you recorded at the start.
- Repeat create-then-delete twice more → no drift on either pass.

**What can go wrong**
- The order header discount total growing by a factor of the discounted line's quantity on each return operation. This is the path the removed duplicate mapper in `commands/returns.ts` used to break, and it is reachable without any unusual setup.
- The discounted line's net collapsing when a completely unrelated line is returned.

### P0 — The three documented behaviour changes for API callers

**Where to click**
- `PUT /api/sales/order-lines` against a line whose stored `discount_percent` is non-zero
- `/backend/sales/orders/{orderId}` to read the result

**What to verify**
- Send `discountAmount: 5` (no `discountPercent`) onto a line that already carries `discountPercent: 5` → **the inherited percentage wins and your amount is dropped.** This is intended (decision D2) and documented in `UPGRADE_NOTES.md`; confirm it is what actually happens, since it is a change from before.
- Send `discountPercent: 0` together with `discountAmount: 5` on a `quantity: 3` line → the amount is honoured and the persisted discount is **15.00** (5 per unit × 3). This is the documented escape hatch; if it does not work, the upgrade note is wrong and integrators have no migration path.
- Send `discountPercent: 12` together with `discountAmount: 0` → **the 12% is applied.** This one *inverts* previous behaviour (a `0` used to suppress the discount), so verify it deliberately — it is the case `UPGRADE_NOTES.md` flags as the dangerous one.
- Send `discountAmount: 300, discountAmountBasis: "line"` on a `quantity: 60, unitPriceNet: 50` line → the persisted discount is **300.00**, not 18000 clamped to 3000, and the net is **2700.00**. This field is new in this PR and has no integration test, so it is worth an explicit check.

**What can go wrong**
- `discountAmountBasis` being silently dropped at the route boundary, so `"line"` behaves like `"unit"` and the discount comes out quantity times too large (then clamped to the whole line subtotal, which reads as a 100% discount).
- A `403`/`400` on any of the above — the field is additive and should never make a previously valid request fail.

### P1 — Existing discounted documents still render correctly

**Where to click**
- `/backend/sales/orders` list, then any pre-existing order with a discounted line
- `/backend/sales/quotes` and a discounted quote
- The order detail **Items** table discount column, at desktop and at a 390px mobile viewport

**What to verify**
- Documents that were already correct before this PR show identical numbers after it — no silent re-computation of a healthy order.
- A document whose stored discount was already inflated by the old defect should **repair itself** on the next write *if the line carries a percentage*; a line with an amount and no percentage will stay wrong, which is expected and tracked in #5641. Do not file the second case as a QA failure.
- The discount column renders both the amount and the percentage as before; nothing about the column's presentation changed in this PR.
- Cold-load `/backend/sales/orders/{orderId}` and capture a screenshot: the shell and loading state should appear as they do on `develop`, and the totals panel should settle without a visible flash of wrong numbers.

**What can go wrong**
- A page-level error or empty totals panel on an order whose stored discount exceeds its line subtotal (the clamped rows).
- Discount and net rendering correctly while the header grand total disagrees with them.

### P1 — Order create flow is unaffected

**Where to click**
- `/backend/sales/orders/create` → add two or three line items via the **Add item** dialog → create the order

**What to verify**
- The draft grid totals, and the created order's totals, match what `develop` produces for the same inputs. No discount is involved on this path (the dialog has no discount field), so this is a pure regression check on the file this PR touched.

**What can go wrong**
- Line totals in the draft grid coming out zero or quantity-multiplied after the `SalesOrderDraftLines.tsx` edit.

### P2 — Documentation

**Where to click**
- `UPGRADE_NOTES.md` in the diff

**What to verify**
- The three-row behaviour table matches what you actually observed in the P0 area above. If any row is wrong, that is worth reporting even though no code changes.

### Data scoping

- Run at least one of the P0 sequences as a **non-admin** user holding `sales.orders.manage` in one organization only, and confirm the order from another organization is not reachable and its totals are never returned.

### Pass/fail
- All areas pass → remove the `qa` label and add `merge-queue` plus `qa-approved` (this clears the QA-approval gate)
- Any area fails → remove the `qa` label, add `qa-failed`, and leave a comment describing the failure, ideally with the exact request body and the numbers you saw.

pkarw:

Independent review pass at head `b8d8266d` from a second account, so the verdict is a real approving review in the Reviews timeline rather than a comment — the PR's own run at 07:16Z could not submit one, since GitHub blocks self-approval, and it reviewed `22235018d`. The two commits since are `.ai/runs/` docs only, so the green CI on that head and on this one describe the same source.

**0 blockers, 0 majors, 0 minors, 3 nits** — all three optional and none disputing the diff: the new basis branch in `SalesOrderDraftLines.tsx` has no reachable caller and no test, `discountAmountBasis` has no HTTP-level coverage, and the spec's `Scope:` header lists five files while nine changed. Under this repo's stricter rule (minors request changes) this is a clean approve, not a borderline one.

**Validation source: GitHub checks** — 14 pass, 5 skipped by design on `b8d8266d`, with the `prepare` and `test` step lists opened individually so nothing hid behind an early red step; every configured gate command maps to a named green step. No local gate was run and none was needed.

Eight things were checked and written into the review as explicit not-findings so the next reviewer does not re-derive them — most usefully that the `PUT /api/sales/orders` full-lines echo still reads an amount-only discount per unit, which I verified is byte-for-byte identical to `develop` (no regression; percent-bearing lines get strictly better) and is the § Alternatives E scope decision D4 explicitly declines. Also worth knowing: the green `Mutation tests` check carries no information here, because `scripts/stryker/scope.mjs:36` allowlists only `packages/shared`, so a `packages/core` diff always resolves to an empty matrix and the `mutate` job skips. The author's own recorded negative control — reverting the engine and observing 10 of 15 new tests fail — is currently the only substitute for it.

Labels: `review` → `merge-queue`; `bug`, `needs-qa`, `priority-high`, `risk-high` each re-derived and kept. `qa-approved` deliberately not applied — code-review approval is not QA approval. Manual QA instructions posted separately, written for API QA rather than click-through, because the admin UI has no discount input at all.

No `ci-monitoring`: CI is fully settled green on this head, so there is no follow-up owed. autofix skipped (not my PR, no `--autofix`).

wojciechszyjka:
🤖 `om-auto-qa-pr` started by @wojciechszyjka at 2026-08-26T09:36:07Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

Running @pkarw's manual QA plan from the instructions comment above verbatim — P0 order-line discount idempotency (read + write path), P0 return create/delete header totals, P0 the three documented behaviour changes plus `discountAmountBasis`, P1 existing-document rendering and the order create flow, P2 the `UPGRADE_NOTES.md` table, and the non-admin `sales.orders.manage` single-organization data-scoping run.

Evidence only: this run will **not** apply `qa-approved`, and per this repo's agent instructions it does not touch the `qa` pipeline label either (that swap is a QA reviewer's to make).

wojciechszyjka:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ **PASS** — every area of @pkarw's QA plan was run and passed, including the non-admin single-organization data-scoping run.
**Environment:** `http://127.0.0.1:39721` · roles `admin` + a purpose-built single-org `sales.orders.manage` user · browser `playwright` (Chromium)
**Verified:** `fix/sales-line-discount-amount-contract` @ `b8d8266d` — the environment descriptor records `builtCommit b8d8266df1aa4cfa151ee3f28dece16e1231f09d`, so the app under test is this PR's head and not a neighbouring branch.

The environment was provisioned fresh with `--no-reuse-env` and came up on **isolated port 39721**, not the shared `5001` that concurrent worktrees all compete for. That matters here: a reused environment is the one way this report could have been written against someone else's build.

@pkarw's instructions comment is right that this is **API QA, not click-through QA** — `LineItemDialog` has no discount input, so no discounted line can be created by clicking. Every P0 area below therefore sets its line up through `POST`/`PUT /api/sales/order-lines` (or the quote equivalent) and then **verifies the result in the browser**, which is exactly the shape the plan prescribes.

---

### P0 — Order line discount is idempotent across repeated writes

**Where to click:** `POST`/`PUT /api/sales/order-lines` → `/backend/sales/orders/{orderId}`

| 1 | `quantity 3 × unitPriceNet 85` at `discountPercent 5` | line discount **12.75**, net **242.25** | discount **12.75**, net **242.25**; the Items table renders `−$12.75` `5%` and `$242.25 net`; the three QA lines add exactly **20.75** of discount and **334.25** of net subtotal | ✅ |
| 2 | Re-send the same `PUT` three times, reloading the page after each | every number stays exactly where it was | `12.75` / `242.25` after all three passes; the Totals panel rendered **byte-identically** at every checkpoint | ✅ |
| 3 | Reload the order page three times without writing | no drift from the display recalculation | Totals panel byte-identical on all three read-only reloads | ✅ |
| 4 | Edit the **third, undiscounted** line on an order carrying two discounted lines | the two untouched discounted lines must not move | both untouched lines unchanged; the header panel matched the pre-edit rendering exactly; header subtotal equals the sum of the visible line nets | ✅ |
| 5 | The same sequence on a quote | `12.75` / `242.25` stable across three re-writes | stable on all three; the quote Items table renders `12.75` | ✅ |

Step 2 is worth spelling out, because it is the sharpest version of the check. The line update route requires `currencyCode` and `quantity`, so "the same `PUT` with only a harmless field changed" is the create payload **minus the discount fields** — which forces the engine to reuse the **stored** discount rather than a caller-supplied one. That is precisely the path that used to compound by quantity. `12.75 → 12.75 → 12.75`, with no `38.25`, no `114.75`, and no net collapsing to `0.00` anywhere in the run.

Step 4 is the check @pkarw flagged as highest-value, since the PR reports finding this bug in its own first implementation. Three lines, two discounted, the clean one edited: the two discounted lines did not move.

---

### P0 — Return create/delete leaves order header totals where they started

**Where to click:** `/backend/sales/orders/{orderId}` → Returns

A return may not exceed the shipped quantity, so the undiscounted line was shipped first and the **post-shipment** state taken as the baseline. Then create-and-delete, three times:

| Cycle | Baseline | Return open | Return deleted | Result |
|---|---|---|---|---|
| 1 | net `342.25`, discount `12.75` | net `317.25`, discount `12.75` | net `342.25`, discount `12.75` | ✅ |
| 2 | — | net `317.25`, discount `12.75` | net `342.25`, discount `12.75` | ✅ |
| 3 | — | net `317.25`, discount `12.75` | net `342.25`, discount `12.75` | ✅ |

The discounted line held `12.75` / `242.25` throughout, including while a return existed against a completely unrelated line. The order header discount total **never grew by a factor of the discounted line's quantity** — the specific failure the removed duplicate mapper in `commands/returns.ts` used to produce. Only the returned line's own `−25.00` moved the totals, and every delete returned the Totals panel to a byte-identical rendering.

---

### P0 — The three documented behaviour changes, plus `discountAmountBasis`

**Where to click:** `PUT /api/sales/order-lines` → `/backend/sales/orders/{orderId}`

| # | Sent | `UPGRADE_NOTES` says | Observed | Result |
|---|------|----------------------|----------|--------|
| 7 | `discountPercent: 5` **and** `discountAmount: 40` | the percent wins | discount **12.75** | ✅ |
| 8 | only `discountAmount: 5`, onto a line whose stored `discount_percent` is 5 | the inherited percent wins, your amount is dropped (D2) | discount **12.75**, net **242.25**; request accepted | ✅ |
| 9 | `discountPercent: 0` with `discountAmount: 5`, quantity 3 | the amount is honoured — 5/unit × 3 | discount **15.00**, net **240.00** | ✅ |
| 10 | `discountPercent: 12` with `discountAmount: 0` | **the 12% is applied** (the inverted row) | discount **30.60** (12% of 255.00), net **224.40** | ✅ |
| 11 | `discountAmount: 300`, `discountAmountBasis: "line"`, quantity 60 × 50.00 | persisted discount **300.00**, net **2700.00** | discount **300.00**, net **2700.00** | ✅ |
| 12 | control: the same `300` **without** the basis | per-unit meaning retained → a different number | discount **3000.00** (300 × 60, clamped to the line subtotal) | ✅ |

Row 10 deserves its own sentence, since `UPGRADE_NOTES.md` flags it as the dangerous one: it *inverts* previous behaviour rather than dropping a value. It was verified deliberately, and the 12% is genuinely applied where a `0` used to suppress it.

Row 12 is the reason row 11 can be trusted. `discountAmountBasis` being silently dropped at the route boundary was the named risk, and a dropped field would make rows 11 and 12 produce the **same** number. They differ by an order of magnitude (`300.00` vs `3000.00`), so the field demonstrably reaches the calculation engine. No request in this section returned a `400` or `403` — the field is additive and never made a previously valid request fail.

---

### P1 — Existing discounted documents still render correctly

| # | Step | Observed | Result |
|---|------|----------|--------|
| 13 | `/backend/sales/orders` and `/backend/sales/quotes` list pages | both render, no error boundary | ✅ |
| 14 | Cold-load an order with one percentage line and one amount-only line | shell and `Loading document…` state appear, then the panel settles with no flash of wrong numbers; the percentage line shows `12.75` / `5%`, the amount-only line `7.50/unit × 2 = 15.00` as documented | ✅ |
| 15 | The same order at a **390px** viewport | document and Totals panel render correctly | ✅ |
| 16 | An order whose stored line discount equals the whole line subtotal (a clamped row) | no page-level error, Totals panel present and populated | ✅ |

Two honest notes on presentation, neither of them caused by this PR:

- At **390px** the Items table scrolls horizontally, so the Discount column sits off-screen until scrolled. Nothing in this diff changed the column's presentation, and the values are present in the DOM and correct; this is the existing responsive-table behaviour.
- The order Totals panel collapses the discount total behind **Show details**, so the discount total was asserted through the single-order `GET` as well as visually.

---

### P1 — Order create flow is unaffected

| 17 | `/backend/sales/orders/create` → three line items via **Add item** | draft grid and created totals correct; no zero or quantity-multiplied line totals | a `2 × $30` line totals **$60.00 gross**; all three rows visible; the created order's grand total is consistent with the three lines | ✅ |

This is the pure regression check on `SalesOrderDraftLines.tsx`, the one `.tsx` outside tests in the diff. As @pkarw noted, its changed branch is not reachable from any screen because the dialog has no discount field — so this confirms the file's *reachable* behaviour is untouched, which is all that can be verified from the UI.

---

### P2 — Documentation

The three-row behaviour table in `UPGRADE_NOTES.md` matches what the application actually does, row for row, as recorded in steps 7–12 above. The worked `discountAmountBasis` example in the same section — 60 units at 50.00 net discounting 300.00 across the line — reproduces exactly: `300.00` with `"line"`, `300.00 × 60` clamped without it. **No documentation corrections to report.**

---

### Data scoping — non-admin, `sales.orders.manage`, one organization

A dedicated role (`sales.orders.view` + `sales.orders.manage`, restricted to one organization) and a user in that organization were created for this run, alongside a second organization holding an order with the same discounted line shape.

| 19 | Run the P0 idempotency sequence as that user | works in their own organization and stays stable | order created (`201`); `12.75` / `242.25` stable across three re-writes; their own order renders with the correct discount | ✅ |
| 20 | Reach for the other organization's order | not reachable, totals never returned | the detail `GET` returns an **empty result set**; the order list never contains it; its line totals are never returned; **selecting** the other organization is rejected `422`; the browser shows **Forbidden** with no data rendered | ✅ |

The sidebar in the scoped user's session is correctly reduced to the features their role actually holds. Every fixture created for this area — user, role, both organizations — was deleted in teardown.

---

### Screenshots

All 24 screenshots are attached below and are non-empty. Ordering follows the run.

### Notes for QA

- **The PR already ships a browser-level test** (`TC-SALES-5019-line-discount-idempotency.spec.ts`), so no follow-up UI-test scenario is proposed.
- **One observation that is explicitly *not* a gate on this PR.** Create a `quantity 60 × 50.00` line with no discount, then `PUT discountAmount: 300` (per unit). The stored discount correctly clamps to `3000.00`, but the line row keeps `total_net_amount 3000.00`, so the row reads *subtotal 3000, discount 3000, net 3000* while the header grand-total net treats that line as `0`. Creating the same line with the discount in the initial `POST` is fully consistent (`discount 3000`, `net 0`), so it is specific to the update path. This input — a per-unit amount with **no** percentage — is the one shape whose engine result is identical before and after this change, and both `UPGRADE_NOTES.md` and the QA plan describe exactly this row as the residue that does not self-heal, tracked in **#5641**. The plan says not to file it as a QA failure, and it is not filed as one.
- **Labels are unchanged by this run.** `qa-approved` was deliberately not applied — see the run summary.

wojciechszyjka:

Every area of @pkarw's QA plan was run against this PR's head `b8d8266d` on a freshly provisioned ephemeral environment (`--no-reuse-env`, isolated port 39721) and passed, including the non-admin single-organization `sales.orders.manage` data-scoping run.

**Labels are unchanged by this run, deliberately.** This was an evidence-only pass:

- ❌ **`qa-approved` was not applied.** It is not this run's to grant — the evidence above is offered so a QA reviewer can make that call. **It remains the one outstanding merge gate**: with `needs-qa` set, no `skip-qa`, and this repository's QA gate on, the PR cannot merge without it.
- 🚦 **The `qa` pipeline label was not applied either**, so `merge-queue` still stands. The plan asks a QA reviewer to swap `merge-queue` → `qa` when picking the work up, but this repository's agent instructions reserve `qa` for QA reviewers and forbid `om-auto-*` skills from touching it. Flagging the divergence rather than silently doing one or the other.

One observation in the evidence is worth a reviewer's eye even though it did not fail the run: an update-path row whose discount clamps to the line subtotal keeps a stale `total_net_amount`, so the line row and the header net disagree. It reproduces identically for an input this change does not alter, and it is the residue `UPGRADE_NOTES.md` and the QA plan both describe as tracked in #5641 — the plan says explicitly not to file it as a QA failure, and it is not filed as one.

PR #5650, opened 2026-08-26
fix(search): stop indexing custom fields twice and rewriting unchanged tokens

maxidragon:
## Summary

Two code paths write a record's rows into `search_tokens`, and they disagree about what a custom field is called. `query_index` builds from `entity_indexes.doc` and writes `cf:<key>`; `@open-mercato/search` re-reads the record through the query engine, which cannot label a column `cf:<key>` (`:` is not a valid SQL identifier) and returns the sanitized alias `cf_<key>`. Each writer replaces tokens by deleting only the `(entity_id, field)` pairs its own document carries, so neither deletes the other's custom-field rows: every custom field is tokenized twice under two names, and the base-field rows both documents share are alternately deleted and re-inserted on every record write.

Separately, #5402 gave `replaceSearchTokensForBatch` a comparison that skips a record whose stored tokens already match the freshly built ones. `replaceSearchTokensForRecord` never got one, so the per-record path — the one event-driven indexing actually uses — rewrites every record's tokens in full even when nothing about the record changed.

This PR fixes both. An unchanged record goes from a full token rewrite on every write to no writes at all, and each custom field is stored once instead of twice.

## Changes

- `packages/search/src/strategies/token.strategy.ts` — `TokenSearchStrategy.index()` / `.bulkIndex()` normalize the query engine's `cf_<key>` aliases back to `cf:<key>` before handing the document to the token writer. Both writers then emit identical field names, so the duplicate collides instead of materializing and the mutual deletion of shared base-field rows stops.
- `packages/core/src/modules/query_index/lib/search-tokens.ts` — `replaceSearchTokensForRecord` gains the unchanged-record comparison from #5402, reusing `tokenSignature` / `tallyEquals`: a `count(*)` probe, then a content read bounded by the built row count, compared by token multiplicity.
- `.ai/specs/2026-08-24-search-token-double-write.md` — new spec.
- `.ai/specs/2026-07-30-search-tokens-unbounded-growth.md` — changelog cross-reference; this touches that spec's duplicate-healing risk row.
- Tests for both, in the packages they belong to.

### Why `cf:` and not `cf_`

`cf:` is the side that is read. The query engine's search predicate asks for `cf:<key>`, and every caller of `findEntityIdsBySearchTokens` passes base column names or `cf:`-prefixed keys (`customers/api/utils.ts` builds them as `cf:${key}`). Nothing queries `cf_`. The divergence was already known and worked around downstream rather than fixed — `shared/lib/encryption/indexDoc.ts` documents it in a comment, and the search indexer accepts both spellings in two places when extracting custom fields.

### Why the normalization sits in the strategy

It is scoped to the rows this strategy writes rather than applied to `IndexableRecord.fields` in `SearchIndexer`, because the same object is handed to the fulltext driver and **Meilisearch rejects an attribute name containing `:`**. Indexing `entity_indexes.doc` directly instead of re-querying has the same problem plus a wider blast radius, since the document's shape would also become what `buildSource`, `formatResult` and `resolveUrl` see for every strategy.

### Two details in the per-record comparison worth a reviewer's eye

- **It is scoped to the document's own `(entity_id, field)` pairs**, because that is what the delete covers. Reading wider would let a token row under a field this document does not carry read as a difference forever and defeat the skip on every write. There is a test for exactly this.
- **The pre-read runs through the caller's transaction when one is supplied.** A separate connection cannot see that transaction's own uncommitted writes, so it could report rows a pending delete has already removed and talk the call out of re-inserting them. The batch path opens its own transaction after the read and has no such caller.

Same consequence #5402 recorded: an unconditional rewrite used to overwrite rows a concurrent writer had committed from a staler document, by accident. It no longer does, so those rows survive until the record's next real change. A repair lost, not a guarantee broken.

### What this does not do

Existing `cf_` rows are orphaned rather than cleaned up — neither writer will delete them afterwards, because the per-record delete only touches pairs present in its own document. Any full batch reindex purges them, since that path deletes by `entity_id`. This ships no destructive cleanup; whether to run a reindex is each deployment's call. Until then those rows cost storage and nothing else: no reader consults the field name, and the hashes they carry are already present under `cf:`.

## Specification

- [ ] Yes
- [x] No (created a new spec)

**Spec file path:**
`.ai/specs/2026-08-24-search-token-double-write.md` (new), plus a changelog entry in `.ai/specs/2026-07-30-search-tokens-unbounded-growth.md`.

## Testing

- `packages/search/src/__tests__/token-strategy-custom-field-keys.test.ts` — drives the real core writer and the strategy against one in-memory `search_tokens`, which is the only place the divergence is observable. Covers the aliased key stored as `cf:<key>` on both the single and batch paths, base columns left alone, a document carrying both spellings, and — with both writers alternating over one record — a table that holds one row per field and stops changing after the first write. All 5 fail with the fix reverted.
- `packages/core/src/modules/query_index/__tests__/search-tokens-record-unchanged-skip.test.ts` — the per-record skip: unchanged record left untouched, changed record still rewritten and searchable, a stored row under a field outside the document not defeating the skip, stale and duplicated rows still collapsed, a runaway stored set settled on the count probe alone, raw-token storage, a record that stops producing tokens, and the read routed through a caller-supplied transaction. 7 of 10 fail with the fix reverted.

Commands run locally:

```
yarn build:packages
yarn generate
yarn typecheck      # 26/26
yarn lint
yarn test           # @open-mercato/core 11,463 passed · @open-mercato/search 335 passed
```

## Checklist

- [ ] QA routing set. 

### Design System Compliance

Not applicable — this change touches no UI-rendering file. No `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`.

- [x] No hardcoded status colors — n/a
- [x] No arbitrary text sizes — n/a
- [x] Empty state handled for list/data pages — n/a
- [x] Loading state handled for async pages — n/a
- [x] `aria-label` on all icon-only buttons — n/a
- [x] Uses existing DS components — n/a

## Linked issues

Relates to #4681 (runaway `search_tokens` growth) — this removes one systematic source of duplicate rows but does not close that issue.
Follows #5402, which added the unchanged-record comparison to the batch path.

## Compatibility

No schema change, no migration, and no contract surface from `BACKWARD_COMPATIBILITY.md`. `replaceSearchTokensForRecord` keeps its signature and its observable outcome — the stored token set for the document's fields — and only the writes it issues to reach that outcome change. No reindex is required for correctness: the `cf:` rows this converges on are the ones readers already ask for, and `query_index` has been writing them all along.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5650"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790322327&installation_model_id=432243&pr_number=5650&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5650&signature=7ec05cae777a0968e39bad99ca1b92c9b5ff381c1bfbb1771f0df77677cb0afe"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


maxidragon:
**Metadata intent** — opened from a fork without `triage` permission, so I can't apply these myself. Recorded here for a maintainer.

**Labels**

| Label | Rationale |
|---|---|
| `bug` | Two writers persisting the same tokens under two field names is a defect, not a tuning change. |
| `review` | Ready for review; not in another pipeline state. |
| `priority-medium` | Nothing user-visible breaks. The cost is write amplification and storage on every record write, which compounds under bulk import. |
| `risk-medium` | `search_tokens` backs token search, the query engines' encrypted `like`/`ilike` rewrite, and the People/Companies list search. The change is small and covered, but a mistake here degrades search quietly rather than loudly, so `risk-low` would understate it. |
| `skip-qa` | Automated-verification exemption: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no schema or API surface change, no `BACKWARD_COMPATIBILITY.md` contract touched, and automated tests for the changed behavior ship in the same PR. |

`partner-request` requested separately via `/label`.

**Assignee:** whoever picks it up — I can't set one from here.

**Reviewer:** someone familiar with `query_index` / `search_tokens`. The two things most worth a second opinion are in the description under "Two details in the per-record comparison worth a reviewer's eye" — the comparison being scoped to the document's own field pairs, and the pre-read running through the caller's transaction.

**QA gate:** the intended routing is `skip-qa`. Until a maintainer applies it, this PR is gated — the evidence above doesn't clear the gate on its own.

No claim label (`in-progress`) requested: I have no way to release it later.

pkarw:
🤖 `om-auto-review-pr` claimed this PR for review at 2026-08-26T08:43Z. Releasing the lock when the review is submitted.

pkarw (approved):
# 🔍 Code Review

**Verdict: ✅ Approve** — no blockers, no majors, no minors. Three nits, all optional.

This is a well-scoped fix for a real defect, and the write-up is unusually honest about what it does not do. I read both changed files end to end, traced the divergence to its source, and independently verified the premise, the direction of convergence, and the convergence property the fix relies on. Everything checked out.

## ✅ What I verified independently

**The premise is real, and the direction of convergence is the right one.** The alias the PR describes is produced at `packages/shared/src/lib/query/engine.ts:978`, `const alias = sanitize(\`cf:${key}\`)`, where `sanitize` (`engine.ts:646`) is `s.replace(/[^a-zA-Z0-9_]/g, '_')` — so the query engine really does hand back `cf_<key>`, while `entity_indexes.doc` keeps `cf:<key>`. That document reaches `TokenSearchStrategy` unmodified: `SearchIndexer.buildIndexableRecords` sets `fields: item` straight from the query-engine row (`packages/search/src/indexer/search-indexer.ts:1149`).

`cf:` is confirmed as the side that is read, not just asserted. The query engine's own predicates branch on `String(filter.field).startsWith('cf:')` in six places (`engine.ts:376`, `417`, `629`, `726`, `807`, `811`), and the People/Companies list search builds its lookup field set as `bucket.add(\`cf:${key}\`)` from the raw definition key (`packages/core/src/modules/customers/api/utils.ts:85`). Nothing anywhere queries `cf_`. Converging on `cf:` is correct.

**Reversing only the prefix is lossless here.** The sanitizer replaces *every* non-`[A-Za-z0-9_]` character, not just the colon, so a prefix-only reversal is only safe if custom-field keys can never contain another such character. They cannot: `packages/shared/src/modules/entities/validators.ts:24` pins `key` to `/^[a-z0-9_]+$/` with the message `snake_case only`. So `sanitize('cf:' + key)` differs from `'cf_' + key` in exactly the one position the fix restores. (See nit 2 for the hardening this suggests.)

**The two writers really do converge, including when their documents carry different field sets** — which they do in production. Core's writer builds from `entity_indexes.doc`, which carries the `search_text` aggregate (`AGGREGATE_SEARCH_FIELD`, `packages/core/src/modules/query_index/lib/document.ts:19`); the strategy's document, being a query-engine projection, does not. Because each writer's comparison scope and its delete scope are both narrowed to its own document's `(entity_id, field)` pairs, fields only one writer knows about are neither compared nor deleted by the other, and each writer is independently idempotent. That is the property the whole design rests on, and the alternating-writers test pins it. The spec's risk table already names this case.

**Scoping the comparison to the document's own field pairs is the right call, and I could not find a case where it hides a real difference.** I walked the boundary conditions: `doc: null` and empty-`doc` both leave `fieldPairs` empty, so the count probe widens to the whole record and the delete falls back to delete-by-`entity_id` — matching the old unconditional behavior. A document whose fields produce no tokens still deletes that document's stored rows. `buildFieldPairs` covers *all* doc keys while `rows` covers only indexable ones, so a stale row under a non-indexable field (`id`, `created_at`) counts as a difference and gets swept — self-healing rather than sticky.

**The concurrency reasoning holds in both directions.** The unordered `.limit(rows.length)` on the content read (`search-tokens.ts:268`) can only ever truncate if a concurrent writer inserted after the count probe, and a truncated read compares as *changed* — so the failure mode is a redundant rewrite, never a wrong skip. A concurrent delete lands the same way. The `reader = options?.trx ?? db` choice is right for the reason stated: a separate connection cannot see the caller transaction's own uncommitted delete, and reading around it is exactly how you would talk the call out of re-inserting rows it just removed. The test at `search-tokens-record-unchanged-skip.test.ts:361` asserts every read goes through `caller-trx`, which is the assertion that actually pins this.

**The added probe is index-supported, so the per-write cost is not a regression.** `search_tokens_entity_idx` on `(entity_type, entity_id)` (`packages/core/src/modules/query_index/data/entities.ts:256`) is the same access path the delete already used. On a changed record the usual added cost is one `count(*)`; the content read only happens when the counts match exactly.

**No hidden dependency on rewriting.** I checked for anything that would make skipped rows decay — there is no TTL, reaper, or `created_at`-based expiry on `search_tokens` anywhere in the repo, so leaving an unchanged record's rows in place cannot silently drop it out of search.

**No contract surface moves.** `search-tokens.ts` keeps the same six exports with unchanged signatures; `tokenSignature`, `tallyEquals` and `tallyTokenRows` were module-private and are only relocated above their new caller. No schema change, no migration, no event, route, DI, or ACL surface — consistent with the PR's `BACKWARD_COMPATIBILITY.md` claim.

## 🧪 Tests

Genuinely good, and I want to call out two things specifically.

The vacuity guards (`expect(store.rows.length).toBeGreaterThan(0)` before every skip assertion) matter more than they look: with `OM_SEARCH_ENABLED` unset the function returns early and *every* "nothing changed" assertion would pass for the wrong reason. Several of these tests would be worthless without that line, and it is there in each one.

The synthetic `id` on the in-memory rows is the right instrument. A delete-then-reinsert of identical content leaves the row count and the content identical — comparing ids is the only way to distinguish "skipped" from "rewritten to the same thing", which is the entire behavior under test.

The alternating-writers test (`token-strategy-custom-field-keys.test.ts:264`) is the one that earns the PR, because it drives the *real* core writer and the *real* strategy against one table and asserts the ids stop moving. That is the actual claim.

Both new files ran and passed in my gate:
- `PASS packages/core/src/modules/query_index/__tests__/search-tokens-record-unchanged-skip.test.ts`
- `PASS packages/search/src/__tests__/token-strategy-custom-field-keys.test.ts`

The absence of integration coverage is justified as stated — no API route, no UI path, no schema change — and the in-memory-table approach follows the precedent #5402 set for the batch path. See the CI note below for the one caveat on that.

## 🧹 Nits (non-blocking — take or leave)

**1. `packages/core/src/modules/query_index/lib/search-tokens.ts:236` — the new explicit `any` on `scopeTokenQuery`.**
`const scopeTokenQuery = (query: any): any` adds an `any`-typed parameter and return in a repo whose root `AGENTS.md` bans `any`. I am flagging it rather than scoring it, because a properly typed alternative is not really reachable here: `db` is `Kysely<any>` in this module, and every column reference in the same function already goes through `as any`. A generic like `<T>(query: T): T` would document the shape-preserving intent without needing DB typings, if you want the annotation to say something.

**2. `packages/search/src/strategies/token.strategy.ts:53` — consider making `normalizeCustomFieldKeys` self-checking.**
As verified above, the prefix-only reversal is exact today because of the `^[a-z0-9_]+$` key regex. It is exact *because of a rule enforced in another package*, though, and nothing links the two. A one-line round-trip guard — rewrite only when re-sanitizing the canonical name reproduces the observed alias — would make the function correct by construction rather than by a distant invariant, and would turn a future key with a `-` or `.` in it into a silent no-op instead of a token written under a name nothing reads. Cheap, and it also subsumes the base-column-named-`cf_*` case in your risk table.

**3. Orphaned `cf_` rows have no operator-facing note outside `.ai/specs/`.**
The decision to ship no cleanup is well argued and I agree with it — a destructive purge does not belong in this PR. But the linked issue is #4681, *runaway `search_tokens` growth*, and the rows this leaves behind are exactly that: dead weight nothing will ever delete until someone runs a full batch reindex. The spec and the PR body say so clearly; a release-note or `UPGRADE_NOTES.md` line saying "run a full reindex to reclaim the orphaned `cf_` rows" would put it where an operator will actually see it.

## 🚨 CI — required workflows have never run on this head

Not a finding against the change, but it must not be lost, because an absent run looks exactly like a clean one:

Both `CI for Develop&Main` and `Mutation tests` are at `action_required` for head `f135aa0` — the fork-PR approval gate, which needs a maintainer click after every push. The only reporting checks are `license/cla` (pass) and `[code]smith` (skipped). `gh pr checks` shows no failures because **no test job has executed**, not because they succeeded.

This matters more than usual here: `develop` has `required_approving_review_count: 1` but **no `required_status_checks`**, so nothing on GitHub blocks a merge on missing CI. Please approve the workflow runs and confirm they are green before merging. In particular, my local gate does not cover the **integration shards** — and `TC-QIDX-4731-token-presence` and `TC-SEARCH-006` are the only places this change meets a real Postgres, where `is not distinct from`, `count(*)` bigint coercion and the unordered `limit` behave for real rather than through the in-memory stand-in.

The branch is 22 commits behind `develop` and merges cleanly (verified with `git merge-tree`). The one overlapping commit, #5612, touches `packages/search/src/indexer/search-indexer.ts` only — no textual or semantic conflict; its new `indexRecordsById` path routes through `bulkIndex`, so it inherits this normalization rather than fighting it.

## 🔧 Validation

**Validation source: local fallback** — GitHub check data is unusable for this head (fork workflow runs unapproved, see above). Full `validation.commands` gate run in an isolated worktree at `f135aa0`:

| Command | Result |
|---|---|
| `yarn build:packages` | ✅ |
| `yarn generate` | ✅ |
| `yarn build:packages` | ✅ |
| `yarn i18n:check-sync` | ✅ all 5 locales in sync |
| `yarn i18n:check-usage` | ✅ |
| `yarn typecheck` | ✅ 26/26 tasks |
| `yarn test` | ⚠️ 32/33 tasks — see below |
| `yarn build:app` | ✅ compiled, TypeScript clean |

`yarn test`: `@open-mercato/core` **11463 passed** (2 skipped), `@open-mercato/search` **335 passed** — both matching the numbers in the PR description. The single failing task is `create-mercato-app`: 875 passed, **4 failed, all the same host assertion** in `packages/create-app/src/lib/template-dev-log-files.test.ts` — `❌ Linux file-watch limits are too low for Turbopack` (`fs.inotify.max_user_instances: 128 < 4096` on this machine). This PR touches nothing under `packages/create-app`; the failure is environmental and unrelated.

## 📌 Labels

Applying the routing you documented — I reached the same conclusions independently before reading your comment, including the `risk-medium` reasoning:

- `bug` — two writers persisting one record's tokens under two names is a defect.
- `merge-queue` — approved.
- `priority-medium` — nothing user-visible breaks; the cost is write amplification and storage, compounding under bulk import.
- `risk-medium` — `search_tokens` backs token search, the query engines' encrypted `like`/`ilike` rewrite, and the People/Companies list search. Small and well covered, but a mistake here degrades search quietly.
- `skip-qa` — automated-verification exemption applies: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no schema or API surface change, no `BACKWARD_COMPATIBILITY.md` contract touched, and tests for the changed behavior ship in the same PR.

`partner-request` retained.

pkarw:
🏷️ **label rationale**

| Label | Why |
|---|---|
| `bug` | Two writers persisting one record's tokens under two field names is a defect, not tuning. |
| `merge-queue` | Approved on re-review of head `0370b9b` — no blockers, majors, or minors. |
| `priority-medium` | Nothing user-visible breaks; the cost is write amplification and storage, compounding under bulk import. |
| `risk-medium` | `search_tokens` backs token search, the query engines' encrypted `like`/`ilike` rewrite, and the People/Companies list search. The change is small and well covered, but a mistake here degrades search quietly rather than loudly, so `risk-low` would understate it. |
| `skip-qa` | Automated-verification exemption: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no schema or API surface change, no `BACKWARD_COMPATIBILITY.md` contract touched, and automated tests for the changed behavior ship in the same PR. |

`partner-request` retained. These match the routing the author documented — reached independently before reading that comment.

✅ **The CI caveat recorded here on 2026-08-26 is resolved.** It read: *"`merge-queue` does not mean CI passed — both `CI for Develop&Main` and `Mutation tests` are at `action_required` for head `f135aa0`."* Both workflows have since executed and succeeded on head `0370b9b` ([CI](https://github.com/open-mercato/open-mercato/actions/runs/33055010815), [Mutation tests](https://github.com/open-mercato/open-mercato/actions/runs/33055010817)), including all 15 `ephemeral-integration` shards. `merge-queue` now means approved **and** green.

pkarw:

No autofix run — this is another author's PR and `--autofix` was not passed, so the branch was not touched.

No CI follow-up scheduled: the required workflow runs are at `action_required` (fork approval gate) and cannot settle without a maintainer click, so waiting on them would return nothing. Validation evidence for this run is the full local `validation.commands` gate, reported in the review.

maxidragon:
Pushed `0370b9b` — a doc comment and one spec risk row. No code change; the diff is otherwise identical to the approved state.

**What it records.** `normalizeCustomFieldKeys` reverses the query engine's alias sanitizer, and that reversal is exact only for keys made of word characters. `sanitize()` is `s.replace(/[^a-zA-Z0-9_]/g, '_')` in `query_index/lib/engine.ts`, and a custom-field `key` is a bare `z.string()` in `entities/api/definitions.ts` and both sibling schemas — nothing constrains its shape. So a key named `order-ref` arrives here as `cf_order_ref` and is rewritten to `cf:order_ref`, a name neither writer nor reader uses: that one field keeps the double-write this PR fixes for every other field.

**Why a comment and not code.** Inverting the sanitizer needs the field-definition key list, which the strategy does not have. Reading `entity_indexes.doc` instead of re-querying would sidestep the whole alias question, and is the alternative this PR already rejects for a wider reason (`buildSource` / `formatResult` / `resolveUrl` would all see the changed document shape).

**One case worth not misreading**, now in the spec's risk table: where such a key and its underscore twin both exist, the two already collide on a single column alias inside the engine's `SELECT`, before the record reaches this strategy. That is not a collision this rewrite introduces, and not one it can repair.

Raised by @jtomaszewski reviewing the downstream carry of this PR. Every custom-field key in the deployment that prompted the work is snake_case, so nothing changes in practice — the limitation is documented rather than hit.

Happy to move this to a follow-up PR instead if a re-approval is unwelcome on an already-approved change.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-28T06:50:40Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-08-28T06:55Z. UI QA verification in progress.

Note on the lock: the `in-progress` label on this PR was opened minutes ago by a concurrent `om-auto-review-pr` run under the same account, not by this run. This QA run is read-only on source and will **not** release that lock — the review run owns it and releases it when it finishes.

pkarw (approved):
# 🔍 Code Review — Re-review of #5650 (head `0370b9b`)

**Verdict: ✅ Approve** — no blockers, no majors, no minors. One new nit on the delta; two nits from the previous pass remain open and optional.

This is my second pass. I approved `f135aa0` on 2026-08-26 with one standing caveat that was not about the code: the required workflows had never executed on that head, so the green-looking checks proved nothing. **That caveat is now resolved** — see 🔧 Validation below. The delta since the approved state is documentation only, and I re-checked every claim it makes against the code.

## 🎯 Summary

`git diff f135aa0..0370b9b` touches two files and adds seven lines: six JSDoc lines on `normalizeCustomFieldKeys` and one row in the spec's risk table. Filtering the `packages/` diff for non-comment additions returns nothing — there is **zero executable change** since the approved state. Everything I verified in the first review (the premise, the direction of convergence, the convergence property, the index-supported probe, the absence of any contract-surface movement) still holds unchanged, and I re-read both source files end to end rather than assuming it.

The new commit answers nit 2 from my previous review — "make `normalizeCustomFieldKeys` self-checking" — by documenting the limitation instead of coding around it. That is a legitimate response, and the reasoning given for it is sound: inverting `sanitize` needs the field-definition key list, which the strategy genuinely does not have, and reading `entity_indexes.doc` directly is the alternative this PR already rejects for the wider reason that `buildSource`, `formatResult` and `resolveUrl` would all then see the changed document shape.

## 🔁 Previous blockers re-checked

There were none — the first pass was a clean approve with three nits. For completeness, the two behavioral claims I most wanted to re-confirm on this head:

- **The premise is still live.** `SearchIndexer.indexRecordById` re-reads the record through the query engine and passes it through as `fields: params.record` (`packages/search/src/indexer/search-indexer.ts:271`), so the `cf_<key>` aliases really do reach `TokenSearchStrategy` unmodified. The normalization is not a no-op against a document that had already been cleaned upstream.
- **The scoping symmetry the skip depends on is intact.** `scopeTokenQuery` (`search-tokens.ts:236`) and the delete in `writeTokens` (`search-tokens.ts:278`) narrow to the same predicate set — `entity_type`, both `is not distinct from` scope columns, `entity_id`, and the document's own field list, with the same empty-`fieldPairs` fallback to a bare `entity_id` match on both sides. If those two ever drift apart the skip becomes a correctness bug rather than an optimization, so it is worth restating that today they do not.

## 🧹 Nits (non-blocking)

**1. `packages/search/src/strategies/token.strategy.ts:50` — the new comment cites the wrong schema, and the risk is narrower than it reads.**

The comment says "a custom-field key is an unconstrained `z.string()`", and the commit message attributes that to `entities/api/definitions.ts` "and both sibling schemas". Those are the **OpenAPI response** schemas — `customFieldDefinitionSchema` feeds `definitionsResponseSchema`, and the other two are the delete request and the upsert response. The schema the write path actually validates against is `upsertCustomFieldDefSchema` (`packages/shared/src/modules/entities/validators.ts:24`), which pins `key` to `.regex(/^[a-z0-9_]+$/, 'snake_case only')`, and both write routes use it — `definitions.ts:542` and `definitions.batch.ts:38`. So no key reachable through the API or the Field Definitions UI can carry a character the sanitizer would rewrite.

The conclusion still stands, though, which is why this is a nit and not a correction of substance: the **code-declared** path is unvalidated. `install-from-ce.ts` copies `field.key` straight into the definition payload (`install-from-ce.ts:138`) with no regex, and `cf.text(...)` in `packages/shared/src/modules/dsl.ts:20` takes a bare `key: string`. A third-party module declaring `cf.text('order-ref')` in its `ce.ts` would produce exactly the case the comment describes. No `ce.ts` in this repository does — I grepped the shipped declarations and every key is already snake_case.

Worth one edit, because the comment is load-bearing for a future maintainer deciding whether the round-trip guard is needed: point it at `upsertCustomFieldDefSchema` and say the gap is the code-declared `ce.ts` path specifically. As written it implies the invariant does not exist anywhere, which would also make my previous nit 2 look unnecessary when in fact it is the thing that would close the remaining hole.

The second half of the new spec row is accurate as written: `sanitize('cf:order-ref')` and `sanitize('cf:order_ref')` both yield `cf_order_ref`, so such a pair really does collide on one column alias inside the engine's `SELECT`, before the record ever reaches this strategy. Correctly identified as neither introduced nor repairable here.

**2. Carried forward — `search-tokens.ts:236`, the explicit `any` on `scopeTokenQuery`.** Unchanged from the previous pass and still not scored, for the same reason: `db` is `Kysely<any>` in this module and the sibling `(eb: any)` in the same function predates the change, so no typed alternative is actually reachable. A generic `<T>(query: T): T` would at least document the shape-preserving intent.

**3. Carried forward — orphaned `cf_` rows have no operator-facing note outside `.ai/specs/`.** A line in `UPGRADE_NOTES.md` or the release notes saying "run a full reindex to reclaim the orphaned `cf_` rows" would put it where an operator looks. Still optional, still not something that belongs in this diff as a destructive migration.

## 💥 Breaking changes

None, re-confirmed on this head. `search-tokens.ts` keeps the same exports with unchanged signatures; `normalizeCustomFieldKeys` is module-private. No schema change, no migration, and no event, route, DI, ACL, widget-spot or generated-file surface from `BACKWARD_COMPATIBILITY.md` is touched. `replaceSearchTokensForRecord` preserves its observable outcome — the stored token set for the document's fields — and changes only the writes issued to reach it.

## 🧪 Test coverage

Unchanged from the approved state and still the strongest part of the PR. The delta adds no test because it adds no behavior; documenting a limitation the code deliberately does not handle is not a testable change, and inventing a test for it would pin the limitation in place.

Restating the two properties that make the existing suite meaningful rather than decorative: the vacuity guards (`expect(store.rows.length).toBeGreaterThan(0)`) are what stop every "nothing changed" assertion from passing for the wrong reason when search is disabled, and the synthetic row `id` is the only instrument that distinguishes "skipped" from "deleted and re-inserted with identical content" — which is the entire behavior under test. The alternating-writers test at `token-strategy-custom-field-keys.test.ts:264` drives the real core writer and the real strategy against one table and asserts the ids stop moving; that is the claim the PR makes, tested directly.

CI ran both new files this time, in a real environment rather than my local one.

## 🔧 Validation

**Validation source: GitHub checks** — no local fallback needed on this pass, and none run.

The caveat from my previous review is cleared. Both required workflows executed and succeeded on head `0370b9b`:

| Workflow / check | Result |
|---|---|
| `CI for Develop&Main` ([run 33055010815](https://github.com/open-mercato/open-mercato/actions/runs/33055010815)) | ✅ success |
| `Mutation tests` ([run 33055010817](https://github.com/open-mercato/open-mercato/actions/runs/33055010817)) | ✅ success |
| `prepare` (build:packages → generate → rebuild → build:app) | ✅ every step success |
| `test` | ✅ |
| `lint`, `ds-lint`, `audit-scope`, `scope`, `documents-multi-instance` | ✅ |
| `ephemeral-integration` 1/15 … 15/15 | ✅ all 15 shards pass |
| `license/cla` | ✅ signed |

I checked step conclusions rather than trusting the job-level roll-up, since a failure early in `prepare` would skip the gates behind it and still leave a plausible-looking summary. Every step in `prepare`, `lint` and `test` reports `success`; nothing is `skipped` except `Install dependencies` (cache hit) and the jobs that are conditionally skipped by design (`audit`, `comment`, `docker-build`, `mutate`, `merge-coverage`).

The point I flagged last time is the one that got answered: my local gate could not cover the integration shards, and `TC-QIDX-4731-token-presence` plus `TC-SEARCH-006` are where this change meets a real Postgres — where `is not distinct from`, the `count(*)` bigint coercion and the unordered `limit` behave for real instead of through the in-memory stand-in. All 15 shards are green.

Head is `MERGEABLE` / `CLEAN` against `develop`, so the 22-commits-behind note from last time no longer needs acting on.

## 📌 Labels

No changes. `merge-queue` (approved), `bug`, `skip-qa`, `priority-medium`, `risk-medium` — exactly one priority and one risk, matching the routing the author documented and the rationale I reached independently on the first pass. `partner-request` retained. `merge-queue` now means what it should: approved **and** green.

Reminder for whoever merges: `develop` carries `required_approving_review_count: 1` but no `required_status_checks`, so GitHub itself will not hold a merge on CI. In this case CI is green, so there is nothing to hold.

pkarw:
🤖 `om-auto-review-pr` completed: **APPROVED** (re-review of head `0370b9b`). Lock released.

The delta since the 2026-08-26 approval is documentation only — six JSDoc lines and one spec risk row, zero executable change in `packages/`. One new nit is in the review body: the added comment cites the OpenAPI *response* schema for the "custom-field keys are unconstrained" claim, while the write path validates against `upsertCustomFieldDefSchema` (`packages/shared/src/modules/entities/validators.ts:24`), which pins `key` to `/^[a-z0-9_]+$/`. The limitation the comment describes is real, but reachable only through the unvalidated code-declared `ce.ts` path — worth one edit, not a blocker. The two nits carried over from the first pass remain open and optional.

Validation source: **GitHub checks** — no local gate run this pass. Both required workflows succeeded on this head, including all 15 `ephemeral-integration` shards, which clears the standing caveat from the previous review that no test job had ever executed on this PR.

No autofix run — this is another author's PR and `--autofix` was not passed, so the branch was not touched. No CI follow-up scheduled: CI has already settled green, so there is nothing to wait for.

Note: the concurrent `om-auto-qa-pr` run under this account explicitly deferred the `in-progress` label to this run, so releasing it here is the intended handoff, not a lock stolen out from under it.

pkarw:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ **PASS** — every required step passed. Base-column search, custom-field search, and the per-record token rewrite all behave correctly on the surface that reads the token spelling this PR converges on. One finding worth your attention is recorded below; it is not a regression and does not block.
**Environment:** `http://127.0.0.1:5001` · role `admin` · browser `playwright`
**Verified:** `search-token-double-write` @ `0370b9b`

### Why this surface

This PR touches no UI-rendering file, so there is no direct UI surface to verify. The closest observable one is **Customers → People**, whose list API asks `search_tokens` for base columns *and* for `cf:<key>` for every active custom-field definition (`packages/core/src/modules/customers/api/utils.ts:85` builds `cf:${key}`). That makes it a direct read of the exact field spelling this PR converges both writers on: if the persisted spelling ever diverges from the queried one, custom-field search silently returns nothing.

### Scenario (P1 — Customers → People list search)

**Where to click:** `/backend/customers/people`

| 1 | Cold-load the People list as `admin` | The 6 seeded people render | All 6 rendered, "Showing 1 to 6 of 6 results in 133ms" | ✅ |
| 2 | Search the base column value `Ortiz` | Exactly Lena Ortiz | 1 result: Lena Ortiz | ✅ |
| 3 | Search the custom-field value `champion` (`cf:buying_role`) | Exactly the 2 people whose buying role is champion | 2 results: Mia Johnson, Lena Ortiz — matches the database exactly | ✅ |
| 4 | Search `economic_buyer` (baseline before the edit) | Exactly the 3 economic_buyer people | 3 results: Arjun Patel, Daniel Cho, Taylor Brooks | ✅ |
| 5 | Search a second custom field, `they/them` (`cf:preferred_pronouns`) | Exactly Taylor Brooks | 1 result: Taylor Brooks | ✅ |
| 6 | Open Taylor Brooks → Custom data → change Buying role `economic_buyer` → `champion` → Save | Save round-trips and the record's tokens are rewritten | Save completed, section re-rendered with `champion`, "Last operation: Update person" flash shown | ✅ |
| 7 | Search `champion` again | 3 results, now including Taylor Brooks | 3 results: Mia Johnson, Lena Ortiz, Taylor Brooks | ✅ |
| 8 | Search `economic_buyer` again | Taylor Brooks gone — stale tokens collapsed, not merely appended to | 2 results: Arjun Patel, Daniel Cho | ✅ |
| 9 | Repeat list + custom-field search at a 390×844 mobile viewport | Usable layout, same results | Renders correctly, same 3 results | ✅ |

Steps 7 and 8 together are the interesting pair: the record is findable under its new custom-field value **and** no longer findable under its old one, which is what a correct `replaceSearchTokensForRecord` rewrite looks like from the outside.

### Screenshots

**people list cold load**



**search base field ortiz**



**search custom field champion**



**search custom field economic buyer before**



**search custom field pronouns**



**custom data editor open**



**custom data saved champion**



**search champion after edit**



**search economic buyer after edit**



**mobile people list**



**mobile search champion**



**noop save editor unchanged**



**noop save completed**




### Notes for QA

**⚠️ Finding (not a regression, does not block): the unchanged-record skip never fires for entities carrying encrypted fields.**

The PR says *"An unchanged record goes from a full token rewrite on every write to no writes at all."* On customers, it does not. I re-saved the Custom data section three times **changing nothing** and the record's tokens were fully rewritten every time — 673 → 664 → 663 rows, with a fresh primary-key set and a new `created_at` on each pass.

Diffing per field across a no-op save shows why. Every field's token-hash fingerprint is **byte-identical** — `cf:buying_role`, `display_name`, `job_title`, `primary_email`, all of them — **except `search_text`**, which changed on every write (530 → 542 rows, new hash each time). The cause sits upstream of this PR: `entity_indexes.doc->>'search_text'` is rebuilt by **re-encrypting** the field values with a fresh IV on each index write (its md5 changes while its length stays constant at 759), whereas the individual encrypted columns reuse their stored ciphertext.

So the comparison this PR adds is behaving **correctly** — the document genuinely does differ — but its benefit is unrealized for every entity that carries encrypted fields, which includes the high-write-volume customer entities that #4681 is about. Worth deciding whether that is acceptable for now or wants a follow-up that makes `search_text` deterministic; it does not need to be resolved on this PR.

**Two limits on what this evidence covers — please read before treating it as full coverage:**

1. **The `cf_` → `cf:` normalization half was not exercised.** `TokenSearchStrategy.index()` is only reached through `SearchService` from the fulltext/vector indexing path, and this environment has no `MEILISEARCH_HOST` (so `createFulltextDriver` returns `null`) and ships `OM_DISABLE_VECTOR_SEARCH_AUTOINDEXING=true`. I confirmed **zero** `cf_`-prefixed rows exist anywhere in `search_tokens` before and after every write this run made — but in this configuration that is the **default state, not evidence the normalization works**, because the second writer never ran. I am explicitly not claiming a pass on that half.
2. **Only the customers People surface was driven.** The other `search_tokens` readers — auth users, messages, `inbox_ops` proposals, checkout transactions — were not exercised. And because this is a freshly seeded database, the pre-existing orphaned `cf_` rows the PR body says it deliberately leaves behind could not be observed.

**Environment:** ephemeral instance from `om-prepare-test-env` — production build (`next start`), disposable testcontainers pgvector, migrated and seeded fresh, with a `mercato queue worker --all` process running. Demo credentials on a disposable database; no secret appears in any screenshot or artifact. No source file was modified by this run.

pkarw:

This change ships no browser-level test — the two new suites (`token-strategy-custom-field-keys.test.ts`, `search-tokens-record-unchanged-skip.test.ts`) are unit tests against an in-memory table. The PR body reasons that integration coverage is not required because there is no API route, UI path or schema change, and that is a fair reading of the checklist.

The UI QA above is the counter-argument worth weighing: the field spelling this PR converges on **is** reachable from a page. `/backend/customers/people` resolves `cf:<key>` tokens through its list API, so a divergence between the persisted and the queried spelling degrades to "custom-field search silently returns nothing" — a failure mode no unit test on either side of the divide can see, because each writer's own suite agrees with itself. The scenario below is exactly the manual run above and took ~35s of wall clock; it would lock in the read path against future changes to either writer. Run `/om-integration-tests` to add it.

1. **Setup:** create, via the API fixtures, one customer person carrying a custom field with a known plaintext value (the seeded `cf:buying_role` / `cf:preferred_pronouns` definitions on `customers:customer_person_profile` are the existing model). Record its id for teardown.
2. **Act & assert — base column:** on `/backend/customers/people`, search the person's last name; assert exactly that person is listed. This pins the base-field token path that the double-write used to delete and re-insert on every write.
3. **Act & assert — custom field:** search the custom field's **value** (e.g. `champion`); assert exactly the set of people carrying it. This is the assertion that fails if the persisted spelling ever drifts from `cf:<key>`.
4. **Act & assert — rewrite:** open the person, change the custom field to a different value, save, return to the list. Assert the person is now returned for the **new** value and is **no longer** returned for the old one. The negative half is the important one — it is what distinguishes a correct replace from an append.
5. **Teardown:** delete every fixture created.

Two things that cost me time and are worth baking into the spec:

- The page has **two** search inputs. `getByRole('searchbox').first()` grabs the sidebar navigation search, which silently filters nothing and makes the assertion pass against an unfiltered list. Locate the list search by its placeholder — `getByPlaceholder('Search by name, email, phone')` — which is also the repo's existing convention.
- After clicking Save on the Custom data section, `getByText('<new value>')` is a strict-mode violation: it matches both the rendered value and the still-mounted `<option>`. Wait for the editor's `<select>` to detach instead (`expect(page.locator('select')).toHaveCount(0)`), which is what actually proves the write round-tripped.

pkarw:
🤖 `om-auto-qa-pr` completed: **PASS**. Evidence posted above.

No lock released by this run — the `in-progress` label present when this QA started was opened by a concurrent `om-auto-review-pr` run under the same account, and that run released it itself while this one was in flight. This run opened no lock of its own.

🏷️ No labels were changed. This run used the default evidence-only mode, so the QA verdict is the reviewer's to act on rather than mine to record; and this PR carries `skip-qa` rather than `needs-qa`, so the self-QA sign-off exception does not apply to it in any case. The existing label set (`bug`, `merge-queue`, `skip-qa`, `priority-medium`, `risk-medium`, `partner-request`) is untouched and, given the change is backend-only, still looks right.

PR #5654, opened 2026-08-26
fix(customers): chunk the ActivitiesSection author lookup and stop caching its failures (#5632)

### LINKED ISSUE #5632: fix(customers): ActivitiesSection author lookup drops ids past 100 and caches its own failures

adeptofvoltron:
Closes #5632

## 🎯 Goal

A customer or company detail page with more than 100 distinct activity authors on screen showed some of them as blank instead of a name, permanently — and a single failed lookup blanked *every* author for the life of the component. This PR makes the timeline's author lookup resolve all of them, and retry the ones it could not.

## 🔍 Problem

`ActivitiesSection` resolves author ids through `GET /api/auth/users?ids=`. #5634 made that parameter actually filter (it was previously accepted and ignored), so the behaviour improved — but only up to the route's cap. Split out of the review of #5617 (finding 5).

Two defects, both in `ActivitiesSection.tsx:308-322`:

1. **No chunking.** Every unresolved id went into one request. `resolveUserIdFilter` slices at `MAX_USER_LOOKUP_IDS = 100` and returns no error, so with 150 authors on screen the last 50 were dropped server-side silently. The call also omitted `pageSize`, so the route's default of 50 truncated anything larger than that even after it survived the cap.
2. **The failure was cached.** Line 317 added every id to `resolvedUserIdsRef` *before* awaiting the fetch. Combined with (1), the dropped ids were marked resolved and never retried; and if the request itself threw — a 403 for a viewer without `auth.users.list`, or a network error — the whole set was marked resolved and every author rendered blank until the component remounted.

## 🔍 Root Cause

The effect recorded *intent* rather than *outcome*. `resolvedUserIdsRef` is a "do not ask again" cache, but it was populated from the request, not the response, so it could not distinguish three different things the old code collapsed together: an id the server answered for and matched, an id the server answered for and matched nothing (a deleted user — genuinely resolved), and an id the server never answered for at all. Only the first two are safe to cache.

This is the same defect `6a03f4c27` fixed in `useDeviceUserLabels`. That fix was deliberately kept devices-local, so this sibling call site still carried it.

## What Changed

- **`packages/core/src/modules/customers/components/detail/authorUserLookup.ts` (new).** The lookup moves out of the effect body into `resolveAuthorUserNames(ids, signal)`, modelled on `devices/backend/devices/userOptions.ts`. It dedupes and trims the ids, chunks at `MAX_USER_LOOKUP_IDS` **imported** from `@open-mercato/core/modules/auth/lib/userIdFilter` (not restated — per the issue, and the same correction #5634 applied to the devices copy), sets `pageSize` to the batch length so pagination cannot truncate a full batch, and returns `{ names, resolvedIds }`. A batch whose request fails is skipped without contributing to `resolvedIds`, so one failed batch does not cost the batches that did answer.
- **`ActivitiesSection.tsx`.** The pre-emptive cache write is gone. The effect now adds only `resolvedIds` to `resolvedUserIdsRef`, inside `.then`, after checking `controller.signal.aborted`. An id the server answered for but matched no row still counts as resolved; an id whose request failed is left unresolved and retried on the next id-set change.
- **Two adjacent defects in the same rewritten block.** The mapper read `user.display_name`, but `/api/auth/users` emits `name` (`auth/api/users/route.ts:496`) — so every author silently fell through to the email branch. `name` is now preferred, with `display_name` and email kept as fallbacks, so no existing label regresses. The call also lacked `x-om-forbidden-redirect: '0'`, meaning a viewer holding `customers.*` but not `auth.users.list` had the whole detail page bounced to /login by a lookup that is supposed to degrade to bare ids; it now degrades, matching the devices sibling.

The helper is deliberately customers-local. The issue left "shared helper or two call sites" as an implementation call, and the devices version is devices-local today — extracting a shared module would mean moving the devices code too, which is out of scope for a fix.

## 🧪 Tests

Local runner (no compose `app` container up), full `validation.commands` gate:

| Gate | Result |
|---|---|
| `yarn build:packages` → `yarn generate` → `yarn build:packages` | pass (no generated-file drift) |
| `yarn i18n:check-sync` | pass — all five locales in sync |
| `yarn i18n:check-usage` | pass (advisory unused-key report only) |
| `yarn test` | 33 of 34 workspace tasks pass — see below |
| `yarn build:app` | pass |

**One known-unrelated failure:** `create-mercato-app#test` fails on this host in the harness oracle suites (`agent-harness-release.test.ts`, `business-writable-oracles.test.ts`, `writable-ast-oracles.test.ts`), which shell out to the `codex`/`claude` CLIs and enforce a 120s `yarn typecheck` budget this machine misses. Verified pre-existing rather than assumed: stashing this branch's changes and re-running that workspace on clean `develop` reproduces the same suites failing, with a larger failure set. This PR touches nothing under `packages/create-app`.

New coverage — `packages/core/src/modules/customers/components/detail/__tests__/ActivitiesSection.authorLookup.test.tsx`, modelled on `devices/__tests__/use-device-user-labels.test.tsx`, five cases:

- 150 authors produce **two** `/api/auth/users` requests of 100 and 50, together covering every id exactly once (asserted as coverage, not sequence — the timeline sort decides the order).
- Each batch sends `page=1` and `pageSize` equal to its own size, so the route's default 50 cannot truncate it.
- The request carries `x-om-forbidden-redirect: '0'`.
- Batch 1 succeeds matching rows for only half its ids while batch 2 fails: the failed batch is retried **in full**, the answered-but-unmatched ids are **never** asked for again, and the retry then resolves them.
- An author is labelled by `name`, not the email fallback.

All five fail against the pre-fix `ActivitiesSection.tsx` and pass with it. The 70 existing `customers/components/detail` suites (312 tests) still pass.

`yarn lint` covers only `@open-mercato/app`, so it does not reach these files; `yarn typecheck` does.

## 💥 Breaking Changes

None. No API route, response shape, component prop, schema, ACL feature, or exported contract changes — `resolveAuthorUserNames` is a new module-internal helper. The one user-visible difference is intended: an author who has a display name is now labelled with it instead of their email address.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5654"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790327738&installation_model_id=432243&pr_number=5654&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5654&signature=b7b5551bdcda0e2fef91641ea1dd9ae3dc9e482fc8fdc14d27c8c96ab04985dd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/customers/components/detail/ActivitiesSection.tsx
+++ b/packages/core/src/modules/customers/components/detail/ActivitiesSection.tsx
@@ -10,6 +10,7 @@ import { Button } from '@open-mercato/ui/primitives/button'
 import { Kbd } from '@open-mercato/ui/primitives/kbd'
 import { ActivityTimelineFilters } from './ActivityTimelineFilters'
 import { ActivityTimeline } from './ActivityTimeline'
+import { resolveAuthorUserNames } from './authorUserLookup'
 import type { ActivitySummary, InteractionSummary } from './types'
 import { createLogger } from '@open-mercato/shared/lib/logger'
 
@@ -314,25 +315,14 @@ export function ActivitiesSection({
     }
     if (unresolvedIds.size === 0) return
 
-    for (const uid of unresolvedIds) resolvedUserIdsRef.current.add(uid)
-
     const controller = new AbortController()
-    readApiResultOrThrow<{ items?: Array<Record<string, unknown>> }>(
-      `/api/auth/users?ids=${[...unresolvedIds].join(',')}`,
-      { signal: controller.signal },
-    )
-      .then((data) => {
-        const users = Array.isArray(data?.items) ? data.items : []
-        const nameMap = new Map<string, string>()
-        for (const user of users) {
-          const userId = typeof user.id === 'string' ? user.id : null
-          const name = typeof user.display_name === 'string' && user.display_name.trim()
-            ? user.display_name.trim()
-            : typeof user.email === 'string'
-              ? user.email
-              : null
-          if (userId && name) nameMap.set(userId, name)
-        }
+    resolveAuthorUserNames([...unresolvedIds], controller.signal)
+      .then(({ names: nameMap, resolvedIds }) => {
+        if (controller.signal.aborted) return
+        // Only ids the server actually answered for are remembered. Marking an id whose request
+        // failed — or one the server silently dropped past its `?ids=` cap — would leave that
+        // author blank for the life of the component, even though the next attempt would work.
+        for (const uid of resolvedIds) resolvedUserIdsRef.current.add(uid)
         if (nameMap.size > 0) {
           setActivities((prev) =>
             prev.map((a) => {
--- a/packages/core/src/modules/customers/components/detail/__tests__/ActivitiesSection.authorLookup.test.tsx
+++ b/packages/core/src/modules/customers/components/detail/__tests__/ActivitiesSection.authorLookup.test.tsx
@@ -1,0 +1,234 @@
+/**
+ * @jest-environment jsdom
+ *
+ * Guards issue #5632: the activity timeline resolves author ids through
+ * `GET /api/auth/users?ids=`, which slices silently at `MAX_USER_LOOKUP_IDS`. The lookup must
+ * therefore chunk at that cap, and it must remember only the ids the server actually answered for
+ * — otherwise ids the server dropped, and ids whose request failed outright, render blank for the
+ * life of the component. Sibling coverage:
+ * `packages/core/src/modules/devices/__tests__/use-device-user-labels.test.tsx`.
+ */
+import { waitFor } from '@testing-library/react'
+import { renderWithProviders } from '@open-mercato/shared/lib/testing/renderWithProviders'
+import { MAX_USER_LOOKUP_IDS } from '@open-mercato/core/modules/auth/lib/userIdFilter'
+import { ActivitiesSection } from '../ActivitiesSection'
+
+const readApiResultOrThrowMock = jest.fn()
+const apiCallMock = jest.fn()
+
+jest.mock('@open-mercato/ui/backend/utils/apiCall', () => ({
+  apiCallOrThrow: jest.fn(),
+  readApiResultOrThrow: (...args: unknown[]) => readApiResultOrThrowMock(...args),
+  apiCall: (...args: unknown[]) => apiCallMock(...args),
+}))
+
+jest.mock('@open-mercato/ui/backend/utils/crud', () => ({
+  createCrud: jest.fn(),
+  updateCrud: jest.fn(),
+  deleteCrud: jest.fn(),
+}))
+
+jest.mock('@open-mercato/shared/lib/frontend/useOrganizationScope', () => ({
+  useOrganizationScopeVersion: () => 'scope-v1',
+}))
+
+jest.mock('@open-mercato/core/modules/dictionaries/components/dictionaryAppearance', () => ({
+  renderDictionaryColor: jest.fn(),
+  renderDictionaryIcon: jest.fn(),
+}))
+
+jest.mock('../hooks/useCustomerDictionary', () => ({
+  useCustomerDictionary: () => ({ data: { map: {} } }),
+  ensureCustomerDictionary: jest.fn(async () => ({ entries: [], map: {} })),
+  invalidateCustomerDictionary: jest.fn(async () => undefined),
+}))
+
+jest.mock('../hooks/useCustomFieldDisplay', () => ({
+  useCustomFieldDisplay: () => ({
+    definitions: [],
+    dictionaryMapsByKey: {},
+    isLoading: false,
+    error: null,
+  }),
+}))
+
+jest.mock('../CustomFieldValuesList', () => ({
+  CustomFieldValuesList: () => null,
+}))
+
+jest.mock('../ActivityTimelineFilters', () => ({
+  ActivityTimelineFilters: () => null,
+}))
+
+const activityTimelineMock = jest.fn(() => null)
+jest.mock('../ActivityTimeline', () => ({
+  ActivityTimeline: (props: unknown) => {
+    activityTimelineMock(props)
+    return null
+  },
+}))
+
+const AUTHOR_COUNT = 150
+
+/** Real UUIDs — the route validates `?ids=` entries as UUIDs, so placeholder ids would not be representative. */
+function authorId(index: number): string {
+  const suffix = String(index).padStart(12, '0')
+  return `00000000-0000-4000-8000-${suffix}`
+}
+
+const AUTHOR_IDS = Array.from({ length: AUTHOR_COUNT }, (_, index) => authorId(index))
+
+function interactionsFor(ids: readonly string[]) {
+  return {
+    items: ids.map((id, index) => ({
+      id: `interaction-${index}`,
+      interactionType: 'note',
+      status: 'done',
+      occurredAt: '2026-03-29T09:00:00.000Z',
+      scheduledAt: null,
+      createdAt: '2026-03-29T09:00:00.000Z',
+      updatedAt: '2026-03-29T09:00:00.000Z',
+      authorUserId: id,
+      authorName: null,
+    })),
+  }
+}
+
+function renderSection() {
+  return renderWithProviders(
+    <ActivitiesSection
+      entityId="company-5632"
+      useCanonicalInteractions
+      addActionLabel="Log activity"
+      emptyState={{ title: 'No activities logged yet', actionLabel: 'Log activity' }}
+    />,
+  )
+}
+
+function idsOf(url: string): string[] {
+  return (new URLSearchParams(String(url).split('?')[1]).get('ids') ?? '').split(',').filter(Boolean)
+}
+
+/** The `?ids=` values of every `/api/auth/users` request issued so far, in call order. */
+function userLookupBatches(): string[][] {
+  return apiCallMock.mock.calls
+    .map(([url]) => String(url))
+    .filter((url) => url.startsWith('/api/auth/users?'))
+    .map(idsOf)
+}
+
+function okResponse(items: Array<Record<string, unknown>>) {
+  return { ok: true, status: 200, result: { items }, response: {} as Response, cacheStatus: null }
+}
+
+function failedResponse() {
+  return { ok: false, status: 403, result: null, response: {} as Response, cacheStatus: null }
+}
+
+function authorNamesFromTimeline(): Record<string, string | null> {
+  const calls = activityTimelineMock.mock.calls
+  const lastProps = calls[calls.length - 1]?.[0] as { activities?: Array<Record<string, unknown>> } | undefined
+  const entries: Record<string, string | null> = {}
+  for (const activity of lastProps?.activities ?? []) {
+    entries[String(activity.authorUserId)] = (activity.authorName as string | null) ?? null
+  }
+  return entries
+}
+
+beforeEach(() => {
+  readApiResultOrThrowMock.mockReset()
+  apiCallMock.mockReset()
+  activityTimelineMock.mockClear()
+  readApiResultOrThrowMock.mockResolvedValue(interactionsFor(AUTHOR_IDS))
+})
+
+describe('ActivitiesSection author lookup', () => {
+  it('chunks the lookup at the route cap instead of dropping the overflow ids', async () => {
+    apiCallMock.mockResolvedValue(okResponse([]))
+
+    renderSection()
+
+    await waitFor(() => expect(userLookupBatches().length).toBe(2))
+    const [first, second] = userLookupBatches()
+    // 150 unresolved authors, a server that slices silently at 100: one request would lose 50 ids.
+    expect(first).toHaveLength(MAX_USER_LOOKUP_IDS)
+    expect(second).toHaveLength(AUTHOR_COUNT - MAX_USER_LOOKUP_IDS)
+    // The timeline sort decides the id order, so assert coverage rather than sequence: every
+    // author is asked for exactly once across the batches.
+    expect([...first, ...second].sort()).toEqual([...AUTHOR_IDS].sort())
+  })
+
+  it('asks for as many rows as the batch holds, so pagination cannot truncate the answer', async () => {
+    apiCallMock.mockResolvedValue(okResponse([]))
+
+    renderSection()
+
+    await waitFor(() => expect(apiCallMock).toHaveBeenCalled())
+    // The route's pageSize defaults to 50 — a 100-id batch would come back half-answered.
+    const [url] = apiCallMock.mock.calls[0]
+    const params = new URLSearchParams(String(url).split('?')[1])
+    expect(params.get('pageSize')).toBe(String(MAX_USER_LOOKUP_IDS))
+    expect(params.get('page')).toBe('1')
+  })
+
+  it('does not let a 403 on the lookup redirect the whole detail page', async () => {
+    apiCallMock.mockResolvedValue(okResponse([]))
+
+    renderSection()
+
+    await waitFor(() => expect(apiCallMock).toHaveBeenCalled())
+    const [, init] = apiCallMock.mock.calls[0]
+    expect((init as RequestInit & { headers: Record<string, string> }).headers['x-om-forbidden-redirect']).toBe('0')
+  })
+
+  it('retries a failed batch and stops asking for ids the server answered for', async () => {
+    // Batch 1 succeeds but matches rows for only half of the ids it was given; batch 2 fails
+    // outright. Both halves are derived from the ids actually requested, because the timeline sort
+    // — not this file — decides which author lands in which batch.
+    const answeredWithoutRow: string[] = []
+    let failedBatch: string[] = []
+    let callIndex = 0
+    apiCallMock.mockImplementation(async (url: string) => {
+      const ids = idsOf(url)
+      callIndex += 1
+      if (callIndex === 1) {
+        answeredWithoutRow.push(...ids.slice(50))
+        return okResponse(ids.slice(0, 50).map((id, index) => ({ id, name: `Author ${index}` })))
+      }
+      if (callIndex === 2) {
+        failedBatch = ids
+        return failedResponse()
+      }
+      return okResponse(ids.map((id, index) => ({ id, name: `Late ${index}` })))
+    })
+
+    renderSection()
+
+    // The names resolved by batch 1 re-render the timeline, which re-runs the lookup effect.
+    await waitFor(() => expect(userLookupBatches().length).toBe(3))
+
+    const retried = userLookupBatches()[2]
+    // The failed batch is retried in full...
+    expect([...retried].sort()).toEqual([...failedBatch].sort())
+    // ...while the ids batch 1 answered for but matched no row are never asked for again.
+    expect(answeredWithoutRow).toHaveLength(MAX_USER_LOOKUP_IDS - 50)
+    for (const id of answeredWithoutRow) {
+      expect(retried).not.toContain(id)
+    }
+
+    await waitFor(() => expect(authorNamesFromTimeline()[failedBatch[0]]).toBe('Late 0'))
+    expect(userLookupBatches()).toHaveLength(3)
+  })
+
+  it('labels an author with the display name the route returns', async () => {
+    readApiResultOrThrowMock.mockResolvedValue(interactionsFor([AUTHOR_IDS[0]]))
+    apiCallMock.mockResolvedValue(okResponse([
+      { id: AUTHOR_IDS[0], name: 'Ada Lovelace', email: 'ada@example.com' },
+    ]))
+
+    renderSection()
+
+    // `name` is what `/api/auth/users` emits; the email is only the fallback.
+    await waitFor(() => expect(authorNamesFromTimeline()[AUTHOR_IDS[0]]).toBe('Ada Lovelace'))
+  })
+})
--- a/packages/core/src/modules/customers/components/detail/__tests__/ActivitiesSection.authorLookup.test.tsx
+++ b/packages/core/src/modules/customers/components/detail/__tests__/ActivitiesSection.authorLookup.test.tsx
@@ -68,7 +68,9 @@ jest.mock('../ActivityTimeline', () => ({
   },
 }))
 
-const AUTHOR_COUNT = 150
+// Derived from the cap rather than hard-coded, so raising `MAX_USER_LOOKUP_IDS` keeps this suite
+// asserting "one full batch plus a partial one" instead of silently collapsing to a single request.
+const AUTHOR_COUNT = MAX_USER_LOOKUP_IDS + 50
 
 /** Real UUIDs — the route validates `?ids=` entries as UUIDs, so placeholder ids would not be representative. */
 function authorId(index: number): string {
@@ -163,12 +165,15 @@ describe('ActivitiesSection author lookup', () => {
 
     renderSection()
 
-    await waitFor(() => expect(apiCallMock).toHaveBeenCalled())
-    // The route's pageSize defaults to 50 — a 100-id batch would come back half-answered.
-    const [url] = apiCallMock.mock.calls[0]
-    const params = new URLSearchParams(String(url).split('?')[1])
-    expect(params.get('pageSize')).toBe(String(MAX_USER_LOOKUP_IDS))
-    expect(params.get('page')).toBe('1')
+    await waitFor(() => expect(userLookupBatches().length).toBe(2))
+    // The route's pageSize defaults to 50 — a 100-id batch would come back half-answered. Assert
+    // the relationship (page size equals this batch's own id count) rather than a literal, so the
+    // check still means something if the cap moves.
+    for (const [url] of apiCallMock.mock.calls) {
+      const params = new URLSearchParams(String(url).split('?')[1])
+      expect(params.get('page')).toBe('1')
+      expect(params.get('pageSize')).toBe(String(idsOf(String(url)).length))
+    }
   })
 
   it('does not let a 403 on the lookup redirect the whole detail page', async () => {
--- a/packages/core/src/modules/customers/components/detail/authorUserLookup.ts
+++ b/packages/core/src/modules/customers/components/detail/authorUserLookup.ts
@@ -1,0 +1,85 @@
+"use client"
+
+import { apiCall } from '@open-mercato/ui/backend/utils/apiCall'
+import { MAX_USER_LOOKUP_IDS } from '@open-mercato/core/modules/auth/lib/userIdFilter'
+
+type AuthUserItem = {
+  id?: unknown
+  name?: unknown
+  display_name?: unknown
+  email?: unknown
+}
+
+// Imported rather than restated: `parseIdsParam` slices past the route's cap without complaining,
+// so a client batch larger than the server accepts loses the overflow ids silently.
+const MAX_IDS_PER_LOOKUP = MAX_USER_LOOKUP_IDS
+
+/**
+ * The outcome of a batch lookup. `resolvedIds` lists the ids the server actually answered for —
+ * an id that came back without a row (deleted user) still counts as resolved, an id whose request
+ * failed does not. Callers cache on `resolvedIds`, so a transient failure is retried rather than
+ * remembered as a permanent blank.
+ */
+export type AuthorUserLookup = {
+  names: Map<string, string>
+  resolvedIds: string[]
+}
+
+function toName(item: AuthUserItem | null | undefined): [string, string] | null {
+  if (!item || typeof item.id !== 'string' || !item.id.trim()) return null
+  const name = typeof item.name === 'string' && item.name.trim() ? item.name.trim() : null
+  const displayName = typeof item.display_name === 'string' && item.display_name.trim()
+    ? item.display_name.trim()
+    : null
+  const email = typeof item.email === 'string' && item.email.trim() ? item.email.trim() : null
+  const label = name ?? displayName ?? email
+  return label ? [item.id.trim(), label] : null
+}
+
+// A viewer of a customer record does not necessarily hold `auth.users.list`, so this lookup must be
+// allowed to fail: `x-om-forbidden-redirect: 0` keeps a 403 from bouncing the whole detail page to
+// /login, and `null` signals that the call itself failed. Callers must not confuse that with a
+// successful call that matched nobody — caching the former would remember a transient error forever.
+async function fetchBatch(ids: string[], signal?: AbortSignal): Promise<AuthUserItem[] | null> {
+  const params = new URLSearchParams()
+  params.set('page', '1')
+  // The route's pageSize defaults to 50, so a batch larger than that would be truncated by
+  // pagination even after it survived the `?ids=` cap.
+  params.set('pageSize', String(ids.length))
+  // URLSearchParams encodes on toString(); pre-encoding here would double-escape the commas.
+  params.set('ids', ids.join(','))
+  const call = await apiCall<{ items?: AuthUserItem[] }>(
+    `/api/auth/users?${params.toString()}`,
+    { headers: { 'x-om-forbidden-redirect': '0' }, signal },
+    { fallback: null },
+  ).catch(() => null)
+  if (!call || !call.ok) return null
+  return call.result?.items ?? []
+}
+
+/**
+ * Batch id → display-name resolution for activity authors already on screen, so an author whose
+ * name was not denormalized onto the activity still renders a name instead of a bare UUID.
+ */
+export async function resolveAuthorUserNames(
+  ids: readonly string[],
+  signal?: AbortSignal,
+): Promise<AuthorUserLookup> {
+  const unique = Array.from(new Set(ids.map((id) => id.trim()).filter(Boolean)))
+  const names = new Map<string, string>()
+  const resolvedIds: string[] = []
+  for (let offset = 0; offset < unique.length; offset += MAX_IDS_PER_LOOKUP) {
+    if (signal?.aborted) break
+    const batch = unique.slice(offset, offset + MAX_IDS_PER_LOOKUP)
+    const items = await fetchBatch(batch, signal)
+    // One failed batch must not cost the batches that did answer, and its ids stay unresolved so
+    // the next render retries them.
+    if (items === null) continue
+    for (const item of items) {
+      const entry = toName(item)
+      if (entry) names.set(entry[0], entry[1])
+    }
+    resolvedIds.push(...batch)
+  }
+  return { names, resolvedIds }
+}


adeptofvoltron:
## 🏷️ Label rationale

- 🔍 **`review`** — this PR is complete and ready for a reviewer; it enters the pipeline in the review state, and no other pipeline label applies (nothing is blocked, no changes have been requested, and manual QA has not started).
- 🐛 **`bug`** — the change fixes reported defective behaviour in `ActivitiesSection`'s author lookup (issue #5632, itself split out of the review of #5617); it adds no new capability.
- 📸 **`needs-qa`** — the automated-verification exemption does not apply here: the change edits `packages/core/src/modules/customers/components/detail/ActivitiesSection.tsx`, a `.tsx` file under `**/components/**`, and it alters what a user sees on the customer and company detail pages (author names now resolve past the 100-id cap, and an author with a display name is labelled by that name rather than their email address). A human should open a detail page with a long activity timeline and confirm the author column. Unit coverage is thorough but cannot exercise the real route or the rendered timeline.
- 🟢 **`priority-low`** — carried over from the issue. The defect degrades a label to a blank field; it loses no data, blocks no workflow, and only bites on detail pages with an unusually large number of distinct activity authors.
- 🟢 **`risk-low`** — the blast radius is one React effect plus a new module-internal helper, both confined to the customers detail timeline. No API route, response shape, component prop, database schema, ACL feature, or exported contract changes, so nothing outside this component can regress.

No `qa-approved` is applied — that is earned by manual QA, not by an authoring skill.

adeptofvoltron:
🤖 `om-open-pr` completed: PR opened ready for review against `develop`, full SDLC label set applied. Lock handed off to `om-auto-review-pr` — chain continues on this PR.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock — code review of the fix. Started: 2026-08-26T09:16:39Z.

adeptofvoltron (commented):
# 🔍 Code Review

## 🎯 Summary

This PR fixes two defects in the customer/company activity timeline's author-name lookup (#5632), split out of the review of #5617. The lookup sent every unresolved author id in a single `GET /api/auth/users?ids=` request and recorded all of them as resolved *before* awaiting the response — so with more than 100 authors on screen the server's silent slice at `MAX_USER_LOOKUP_IDS` dropped the overflow while the client remembered it as answered, and any request-level failure poisoned the whole set permanently.

The change is well-targeted: it moves the lookup into a new customers-local helper that chunks at the imported cap, sizes each page to its own batch, and reports back which ids the server actually answered for, leaving the caller to cache only those, after the response. It is a faithful port of the fix `6a03f4c27` applied to `useDeviceUserLabels`, which is exactly what the issue asked for.

**Verdict: 💬 Comment — no blockers and no majors found. The diff meets the bar for approval, but GitHub does not permit the PR author to submit an approving review, and this review was produced by the same automation that wrote the code. It therefore needs a second, human reviewer to approve before merge.**

Two minor test-robustness findings were raised during this review and fixed in the same pass (commit `3e42f698f`, detailed under 🟡 Minor below).

## 🧪 Validation Gate

Local runner (no compose `app` container running), `validation.commands` in configured order, against the reviewed tree.

| # | Command | Result |
|---|---------|--------|
| 1 | `yarn build:packages` | ✅ pass — 27/27 tasks |
| 2 | `yarn generate` | ✅ pass — 517 API paths, no tracked-file drift (`git status` clean apart from the three PR files) |
| 3 | `yarn build:packages` | ✅ pass — 27/27 tasks |
| 4 | `yarn i18n:check-sync` | ✅ pass — all five locales in sync across 57 modules |
| 5 | `yarn i18n:check-usage` | ✅ pass — advisory unused-key report only, nothing introduced by this diff |
| 6 | `yarn typecheck` | ✅ pass — 27/27 tasks, re-run after the review follow-up commit |
| 7 | `yarn test` | ⚠️ 33 of 34 workspace tasks pass — see note |
| 8 | `yarn build:app` | ✅ pass |

**Note on gate 7.** The one failing task is `create-mercato-app#test`, in the harness oracle suites `agent-harness-release.test.ts`, `business-writable-oracles.test.ts` and `writable-ast-oracles.test.ts`. These shell out to the `codex`/`claude` CLIs and enforce a 120-second `yarn typecheck` budget this host does not meet. The reviewer did not accept the author's "unrelated" claim on trust: the PR's changes were stashed and that workspace re-run on clean `develop`, which reproduces the same suites failing with a **larger** failure set. The failure is therefore pre-existing on the base branch and independent of this diff, which touches nothing under `packages/create-app`. Per the gate's "no excuses" rule this is recorded as an environment finding rather than waved away — it is not a reason to block this PR, but it does mean gate 7 is green only for the 33 workspaces this change can affect. CI on the PR is the authority for the remaining one; no required check has failed there.

**CI at review time.** No required check is failing. `lint`, `ds-lint`, `scope`, `audit-scope`, `Analyze (actions)`, `Analyze (python)` and `license/cla` have all passed. Two checks — `prepare` and `Analyze (javascript-typescript)` — were still **pending** when this review was submitted. This review was not delayed for them: if they land red, that is new information not covered here and the PR should not merge on this review alone.

## 📋 Findings

### 🟡 Minor

**1. `packages/core/src/modules/customers/components/detail/__tests__/ActivitiesSection.authorLookup.test.tsx:74` — the fixture size was a literal, so the chunking case would silently stop testing chunking if the cap moved.**

`AUTHOR_COUNT` was hard-coded to `150`, which only produces two batches while `MAX_USER_LOOKUP_IDS` is `100`. Raising the route's cap to, say, `200` would leave the fixture below one batch, and the "chunks the lookup at the route cap" test would pass by asserting a single request — the exact regression it exists to catch would go unnoticed. Since the suite already imports the constant, the fixture should be expressed in terms of it. **Fixed in `3e42f698f`:** `AUTHOR_COUNT = MAX_USER_LOOKUP_IDS + 50`, so the case stays "one full batch plus a partial one" at any cap.

**2. `packages/core/src/modules/customers/components/detail/__tests__/ActivitiesSection.authorLookup.test.tsx:157` — the page-size assertion compared against the cap rather than the batch, and only checked the first request.**

The test asserted `pageSize === String(MAX_USER_LOOKUP_IDS)` on `apiCallMock.mock.calls[0]`. That is only correct because the first batch happens to be full; it says nothing about the second (50-id) batch, and it would become wrong the moment the fixture no longer starts with a full batch. The property actually being defended is *page size equals this batch's own id count* — that is what stops the route's default `pageSize` of 50 truncating a batch. **Fixed in `3e42f698f`:** the assertion now loops over every request and compares `pageSize` to `idsOf(url).length`, and also pins `page=1` per batch.

### 🔵 Nit

**3. `packages/core/src/modules/customers/components/detail/authorUserLookup.ts:1` — `"use client"` diverges from the sibling this file is modelled on.**

`devices/backend/devices/userOptions.ts`, which this helper deliberately mirrors, carries no `"use client"` directive, while this one does. Both are correct — the directive is harmless and arguably more accurate for a module that only ever runs in the browser, and `@open-mercato/ui/backend/utils/apiCall` declares it too — but the two files now differ for no stated reason. Author's call; no change requested.

**4. `packages/core/src/modules/customers/components/detail/ActivitiesSection.tsx:303-307` — pre-existing: the resolved-id cache is wiped on every reload.**

The outer effect resets `resolvedUserIdsRef.current = new Set()` after each `loadActivities()`, so the cache this PR now populates correctly only survives within a single load cycle; a "mark done" or a filter change discards it and re-fetches every author. This is unchanged by the diff and out of scope for the issue, which is about *incorrect* caching rather than cache lifetime. Recorded so it is not mistaken for something this PR introduced, and as a candidate for a separate follow-up if the extra requests ever matter.

## 💥 Breaking Changes

- [x] **Exported APIs** — no exported symbol removed, renamed, or retyped. `resolveAuthorUserNames` and `AuthorUserLookup` are new additions in a new module.
- [x] **HTTP routes and response shapes** — untouched. The PR only changes which query parameters the client sends (`ids` chunked, `page`/`pageSize` now supplied), all already accepted by `GET /api/auth/users`.
- [x] **Event names / widget spot IDs / notification IDs** — untouched.
- [x] **CLI flags** — untouched.
- [x] **DB schema** — untouched; no entity, migration, or snapshot in the diff, and none required.
- [x] **DI keys / ACL features** — untouched. The lookup still runs behind the unchanged `auth.users.list` feature; the added `x-om-forbidden-redirect: '0'` changes only the *client's* reaction to a 403, never who is allowed to call the route.
- [x] **Component props** — `ActivitiesSectionProps` is unchanged, so every existing call site compiles and behaves identically.
- [x] **`BACKWARD_COMPATIBILITY.md`** — checked against all 13 contract-surface categories; this diff touches none of them, so no deprecation protocol applies.

**One intentional user-visible behavior change**, correctly called out in the PR description rather than slipped in: authors who have a display name are now labelled with it instead of their email address. The old mapper read `user.display_name`, but `auth/api/users/route.ts:496` emits `name`, so the display-name branch was dead and every author fell through to email. Preferring `name` while retaining `display_name` and `email` as fallbacks is the behavior the original code was reaching for, and no label that previously rendered can now render empty. This is a legitimate part of the fix, but it is the item a QA pass should look at first.

## 🧪 Test Coverage

- [x] **The bug fix ships a regression test that fails without the fix.** `ActivitiesSection.authorLookup.test.tsx` adds five cases and the author verified all five fail against the pre-fix `ActivitiesSection.tsx`; this reviewer re-ran the suite (5 passed) and the whole `customers/components/detail` tree (70 suites, 312 tests, all passing) after the follow-up commit.
- [x] **Both defects named in the issue are covered, in the shape the issue asked for.** The issue requested "(a) a 150-id set produces two requests, and (b) a failed request leaves the ids unresolved so the next id-set change retries them" — both are present, and the retry case is stronger than asked: it drives a *partial* failure (batch 1 succeeds matching rows for only half its ids, batch 2 fails) so it simultaneously proves the failed batch is retried in full, that ids answered-but-unmatched are never re-requested, and that one failed batch does not cost a batch that succeeded.
- [x] **The adjacent fixes are covered too** — a case for the `x-om-forbidden-redirect: '0'` opt-out, and one pinning `name` over the email fallback, so neither silently regresses.
- [x] **The suite is order-independent.** The assertions compare id *coverage* and sorted sets rather than sequence, which matters because `sortTimelineActivities` — not the fixture — decides which author lands in which batch. An earlier draft that asserted exact ordering would have been brittle here.
- [x] **The named model was followed.** The issue pointed at `devices/__tests__/use-device-user-labels.test.tsx`; this suite mirrors its structure and its scenario set.
- [x] **Integration coverage** — not required. The change is confined to client-side request shaping against an existing, already-covered route (`auth/__tests__/user-id-filter.test.ts` and `TC-DEV-009` cover the server side of `?ids=`), and it introduces no new API path, permission check, or data-scoping decision.

Manual QA is still warranted and the PR is correctly labelled `needs-qa`: unit tests mock the transport, so nobody has yet watched a real detail page with a long activity timeline resolve its authors. QA instructions follow in a separate comment.

adeptofvoltron:

This PR carries no blocker or major findings, but it stays in `review` rather than `merge-queue`: GitHub will not let the PR author submit an approving review, and the code review above came from the same automation that wrote the change, so a **human approval is still required**. QA can run in parallel with that approval. The QA-approval gate holds the merge until `qa-approved` is added. QA reviewer: when you pick it up, add the `qa` label, then run the routes below.

### P0 — Author labels on the activity timeline

**Where to click**
- `/backend/customers/people/{id}` → the **Activity** / timeline tab, on a person whose timeline holds many activities.
- `/backend/customers/companies/{id}` → the same tab. The company view is the one most likely to exceed the 100-author threshold, since it aggregates activity across everyone at the company.

**What to verify**
- Every timeline entry shows an author label, not a blank space and not a bare UUID. Scroll to the bottom and click *load more* until the list is exhausted — the whole point of this fix is that entries past the first 100 distinct authors used to stay blank.
- Open DevTools → Network, filter on `auth/users`, and reload the tab. On a timeline with more than 100 distinct authors you should see **more than one** request, each with at most 100 comma-separated values in `?ids=` and a `pageSize` matching that count. Before this fix there was exactly one request and the server silently ignored everything past the 100th id.
- Authors who have a display name set now render as that **name**. Previously every author rendered as their **email address**, because the client read a field the API does not emit. This is the intended change most likely to surprise someone, so confirm the names shown are correct people — not shifted by one, not showing the wrong person's name.
- An author whose user account was deleted should render as before (blank or bare id) without the page retrying in a loop — watch the Network tab settle rather than repeat.

**What can go wrong**
- Names attached to the wrong activities — the fix rewrites the id→name mapping, so a mismatch would show up as plausible-looking but wrong attribution.
- A request storm: repeated `auth/users` calls that never stop. The lookup now caches after the response instead of before it, so a bookkeeping mistake would show as continuous polling.
- Authors blank again after using *Mark done* or changing the timeline filters, which re-load the list.

### P0 — A viewer without `auth.users.list`

**Where to click**
- Sign in as a role that grants `customers.*` but **not** `auth.users.list`, then open the same customer or company detail page and its timeline tab.

**What to verify**
- The page stays on the detail view. Before this PR the 403 from `/api/auth/users` bounced the whole page to `/login`, which looked like a session expiry to the user; the lookup now opts out of that redirect and degrades quietly.
- The timeline still renders every activity — only the author labels fall back to whatever the activity already stored. No error banner, no flash message.
- A genuine expired session must still redirect to login. Sign out in another tab, then interact with the timeline, and confirm you are sent to login as normal — the opt-out covers 403 only, not 401.

**What can go wrong**
- Being redirected to `/login` while still authenticated (the original bug).
- The opposite: an expired session that no longer redirects and leaves the page silently failing.

### P1 — Small timelines are unaffected

**Where to click**
- `/backend/customers/people/{id}` on a person with only a handful of activities.

**What to verify**
- Exactly one `auth/users` request is issued, and author labels appear as quickly as before. This is the common case and it must not have regressed into extra round trips.
- Cold-load the detail page and watch the timeline tab: the loading state appears immediately and author names fill in without the list visibly reflowing or flashing. Check the same on a mobile viewport.

**What can go wrong**
- A perceptible delay before author names appear, or the timeline jumping as names land — the lookup now awaits batches in sequence, so a regression here would show as staggered filling.

### Pass/fail
- All routes pass → add `qa-approved` and leave the pipeline label as-is until a human reviewer approves the code, then it can move to `merge-queue`.
- Any route fails → add `qa-failed` and leave a comment describing the failure, with the route and what you saw.

adeptofvoltron:
🤖 `om-auto-review-pr` completed: **COMMENT (no blockers, no majors)**. Two minor test-robustness findings were raised and fixed in the same pass (`3e42f698f`); the full validation gate was re-run afterwards. The PR stays in the `review` pipeline state rather than moving to `merge-queue`, because GitHub does not permit the author to approve their own PR and this review came from the automation that wrote the code — **a human approving review is still required**. Manual QA instructions are posted above. Lock retained — chain continues.

adeptofvoltron:
## 📸 UI QA evidence — PR #5654 (issue #5632)

**Verdict: ✅ PASS**, with two honest scope limitations recorded below. The chunking fix was verified end to end against a real running app and a real `/api/auth/users` route, including a **before/after comparison** run on the same seeded data with the component reverted — not just unit-level mocks.

### Environment

A disposable Postgres database (`om_qa_5632`) initialized with `yarn initialize`, driven by the repo's dev server on `http://localhost:3311`, exercised through the configured Playwright provider as `admin@acme.com`. The developer's own `open-mercato` database was not touched (verified before and after). Seeded fixture: one demo person (Arjun Patel) carrying **150 activities whose `author_user_id` matches no user row** — which is what forces the client-side fallback this PR fixes — plus **5 activities authored by a real user** given the display name *Ada Lovelace*.

### Why this path needed constructing

Worth recording for the reviewer, because it changes how the issue's severity reads: **both** `/api/customers/interactions` and `/api/customers/activities` already resolve `authorName` server-side (`interactions/route.ts:676`, `activities/route.ts:255`), uncapped, for every author on the page. The client-side lookup in `ActivitiesSection` is therefore a **fallback**, reached only for authors the server itself could not resolve. The bug is real and the fallback exists for real reasons, but "more than 100 unresolved authors on screen" is a narrower state than the issue text implies, and it had to be constructed deliberately here.

### ✅ P0 — Chunking at the route cap (the fix)

**Where:** `/backend/customers/people/{id}` → activity timeline → *Load more* twice, so 150 activities with 150 distinct unresolved authors are on screen at once.

**Observed, with the fix (post-fix run):**

| Timeline state | Unresolved authors | `/api/auth/users` requests issued |
|---|---|---|
| Page 1 (50 activities) | 45 | one request, 45 ids, `pageSize=45`, `page=1` |
| Page 2 (100 activities) | 95 | one request, 95 ids, `pageSize=95`, `page=1` |
| Page 3 (150 activities) | 145 | **two** requests — 100 ids then 45 ids, each `pageSize` equal to its own batch size |

Every batch satisfied `ids.length <= 100`, `pageSize === ids.length`, and `page === 1`.

**Observed, with the component reverted to `origin/develop` and `@open-mercato/core` rebuilt (pre-fix run, same fixture, same clicks):**

| Timeline state | Unresolved authors | `/api/auth/users` requests issued |
|---|---|---|
| Page 3 (150 activities) | 145 | **one** request, **145 ids**, `pageSize` absent, `page` absent |

That single pre-fix request is the defect, visible in the network log: 145 ids sent, no page size, so the route slices at `MAX_USER_LOOKUP_IDS = 100` *and* then applies its default `pageSize` of 50. The dropped ids were marked resolved regardless, so those authors could never recover.

**Independently confirmed the drop is silent** — 149 throwaway UUIDs plus one real user id in position 150 sent as a single `?ids=` request returned `rows: 0` with HTTP 200 and no error; the same id sent inside a batch of one returned the row. The server gives the client no signal that it truncated.

### ✅ P0 — No request storm, no regression on the main path

The request count was sampled, then re-sampled after a 5-second idle: stable at 5 both times, so the "cache only what the server answered for" bookkeeping does not spin. *Ada Lovelace* renders as the author on the 5 server-resolved activities, confirming the ordinary timeline is unaffected.

**One observation, pre-existing and not caused by this PR.** The first page issues the same 45-id lookup **twice**. This reproduced identically in the pre-fix run, so it is not a regression: `ActivitiesSection.tsx:303-307` resets `resolvedUserIdsRef` on every `loadActivities()`, so a second load in the same mount re-asks for everything. It is recorded as nit #4 in the code review and is a candidate for separate follow-up.

### ✅ P1 — Field mapping, cold load, mobile

`GET /api/auth/users` emits `name: "Ada Lovelace"` and **no `display_name` key at all** — confirming the pre-fix client read a field that never existed and always fell through to the email branch. The corrected `name` → `display_name` → `email` order is what the API actually supports.

Cold load of the detail route reached a useful shell immediately, author labels filled in without the timeline reflowing, and the 390×844 mobile viewport rendered the timeline without layout breakage.

### ⚠️ Not exercised

1. **Client-side labelling through the fallback.** Producing a state where the server misses an author but `/api/auth/users` finds them requires a cross-organization scope difference between the two routes; it was not constructed. The `name`-over-`email` correction is proven at the API level (above) and by unit test, but no screenshot shows a name resolved *by the client fallback*.
2. **The 403 degradation path.** Signing in as a role holding `customers.*` but not `auth.users.list` was not set up. The `x-om-forbidden-redirect: '0'` header is asserted by unit test and its effect is confirmed by reading `packages/ui/src/backend/utils/api.ts:243-247`, but no browser run proves the page stays put on a real 403.

Both remain worth a human QA pass; neither blocks on the evidence gathered here.


### Screenshots

**step-04-timeline-page1.png**



**step-05-timeline-loaded.png**



**step-06-mobile.png**

adeptofvoltron:
## 🧪 Follow-up: no browser-level test ships with this change

This PR adds thorough unit coverage (`ActivitiesSection.authorLookup.test.tsx`, five cases) but no integration/E2E test, so the chunking contract is pinned only against a mocked transport. The manual run above proved it against the real route — including a before/after with the component reverted — but nothing stops it regressing silently. Recording a ready-to-implement scenario so a follow-up can add it via `om-integration-tests`.

**Proposed test:** `packages/core/src/modules/customers/__integration__/TC-CUS-XXX-activity-author-lookup-chunking.spec.ts`

**Fixture (create in setup, clean up in teardown — no reliance on seeded demo data):**
1. Create one customer person via the API.
2. Insert `MAX_USER_LOOKUP_IDS + 50` activities on that person, each with a distinct `authorUserId` that matches no user row. This is what forces the client-side fallback: both `/api/customers/interactions` and `/api/customers/activities` resolve `authorName` server-side for authors that *do* resolve, so an author the server can resolve never reaches the code under test.
3. Create one activity authored by a real user carrying a display name, to assert the ordinary path still renders.

**Steps:**
1. Open `/backend/customers/people/{id}` and select the activity timeline tab.
2. Click *Load more* until every seeded activity is on screen.
3. Capture every `/api/auth/users` request the page issues.

**Assertions:**
- No single request carries more than `MAX_USER_LOOKUP_IDS` ids — imported from `@open-mercato/core/modules/auth/lib/userIdFilter`, never restated.
- Each request's `pageSize` equals its own batch's id count, and `page` is `1`. This is the half of the fix that is easy to lose: the route's `pageSize` defaults to 50, so a 100-id batch with no page size comes back half-answered even though it survived the cap.
- The request count stops growing once the timeline settles (no retry storm from the cache-after-response bookkeeping).
- The activity authored by the real user renders that user's display name.

**Known limitation to carry over.** Two behaviors verified only by unit test remain hard to reach in a browser and were **not** exercised in the manual run either: labelling resolved *through* the client fallback (needs a cross-organization scope difference between the interactions route and `/api/auth/users`), and the `x-om-forbidden-redirect: '0'` degradation (needs a role holding `customers.*` but not `auth.users.list`). If the follow-up can build those roles as fixtures, both are worth adding — the 403 case especially, since before this PR it redirected the whole detail page to `/login`.

This is a tracking note only; no issue was opened.

adeptofvoltron:
🤖 `om-auto-qa-pr` completed: **PASS**. Verified against a disposable initialized database and the dev server, including a before/after run with the component reverted — pre-fix issued one 145-id request with no page size, post-fix issues two (100 + 45), each sized to its batch. Evidence and screenshots posted above; two paths could not be constructed and are named honestly in the report. Default evidence-only mode: no pipeline or meta labels were changed, so `needs-qa` stands and a human QA reviewer still owns the verdict. Lock retained — chain continues.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: issue #5632 fixed on this PR, reviewed (no blockers, no majors), and UI-verified with before/after browser evidence. The PR is ready for a **human approving review** — the automation authored it, so it cannot approve itself — and carries `needs-qa` for a manual QA sign-off. Chain lock released; `ci-monitoring` applied while the last CI checks finish.

adeptofvoltron:
🤖 `om-auto-fix-issue` — CI result: **all checks green** (14 passed, 5 skipped, none failing, none pending). This closes the CI follow-up this run owed; `ci-monitoring` removed.

Remaining before merge, both requiring a human: an approving review (the automation authored this PR and cannot approve it), and manual QA to clear the `needs-qa` gate — instructions and browser evidence are posted above.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T00:07:01Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review

**Verdict: ✅ Approve** — no blockers, no majors. Four nits below, none of which need to hold the merge.

**Validation source: GitHub checks** (head `3e42f698f`). All 14 reported checks pass, none failing, none pending; the `test` job's own step list confirms the gates that matter here actually ran and succeeded — `Check agent instruction budget`, `Check i18n sync`, `Check i18n usage`, `Checking types`, `Test`, `Test scripts` — plus `lint`, `ds-lint`, `docker-build`, `ephemeral-integration`, and all three CodeQL analyses. No local re-run of `validation.commands` was needed.

## 🎯 Summary

`ActivitiesSection` resolved activity-author ids through `GET /api/auth/users?ids=` in one unbounded request, and wrote every requested id into its `resolvedUserIdsRef` "do not ask again" cache *before* awaiting the response. Since `resolveUserIdFilter` slices at `MAX_USER_LOOKUP_IDS = 100` and returns no error, a timeline with more than 100 distinct authors permanently blanked the overflow; and any failed request — a 403 for a viewer without `auth.users.list`, or a network error — blanked every author for the life of the component.

The fix moves the lookup into `authorUserLookup.ts` and changes the cache from recording *intent* to recording *outcome*. That is the correct framing of the bug, and the resulting invariant — "cache an id only when the server answered for it; an id answered for but matched to no row is still resolved, an id whose request failed is not" — is the right one. I traced the three-state distinction through the code and it holds.

I verified the supporting claims against the tree rather than taking the PR body's word for them:

- `auth/api/users/route.ts:496` returns `name`, never `display_name` — so the old mapper's `display_name` branch was dead and every author fell through to the email. The `name` → `display_name` → `email` chain in `authorUserLookup.ts:35` is a real fix, and the PR body declares the user-visible label change explicitly. ✅
- `routeMetadata.GET` (`route.ts:132`) requires `auth.users.list`, so the 403 the `x-om-forbidden-redirect: '0'` header defuses is genuinely reachable for a viewer holding only `customers.*`. ✅
- When the lookup degrades, `ActivityTimeline.tsx:152` renders the author line only when `authorName` is set, so degradation drops the "by X" line rather than exposing a raw UUID. (The PR body's phrase "degrade to bare ids" undersells it — the actual behaviour is better.) ✅

I also checked the two things most likely to go wrong in a rewrite like this and found both handled:

- **No render storm.** `setActivities` only fires when `nameMap.size > 0`, and every id that produced a name is in `resolvedUserIdsRef` by then — so the effect's `[activities]` dependency converges after one extra pass. A batch that fails outright produces no names, so it does not re-trigger the effect and cannot spin.
- **No state update after abort or unmount.** `fetchBatch`'s `.catch(() => null)` absorbs the `AbortError` that `apiCall` re-throws, the loop's `if (signal?.aborted) break` stops the remaining batches, and the `.then` re-checks `controller.signal.aborted` before touching either the ref or state.

The one behavioural trade-off is deliberate and correct: moving the cache write into `.then` means a lookup aborted mid-flight no longer suppresses a duplicate request for the same ids. That costs a redundant request when `activities` churns, and it is exactly the price of not caching an outcome that never arrived.

## ✅ What's done well

- The cap is **imported** from `@open-mercato/core/modules/auth/lib/userIdFilter` rather than restated, so a client batch can never exceed what `parseIdsParam` accepts. I considered whether this trips the checklist's "no direct imports from other modules' business logic" rule and concluded it does not: `MAX_USER_LOOKUP_IDS` is a published HTTP-contract constant (it appears verbatim in the route's own OpenAPI description, `route.ts:729`), the module is client-safe (its only import is type-only), and `devices/backend/devices/userOptions.ts:2` already established this exact import as the sanctioned pattern in #5634.
- The five new tests are structurally non-vacuous, which is not a given for this kind of guard. Test 1's `waitFor(() => …length).toBe(2))` cannot pass against an unchunked implementation; test 4's `.toBe(3)` cannot pass against pre-emptive caching; and its `expect(retried).not.toContain(id)` loop runs over 50 real ids rather than an empty set. Deriving `AUTHOR_COUNT` from the cap (line 73) instead of hard-coding 150 means raising `MAX_USER_LOOKUP_IDS` keeps the suite asserting "one full batch plus a partial one" rather than silently collapsing to a single request — that is the detail most such suites get wrong.
- Using real v4 UUIDs in the fixture (line 76-79) matters more than it looks: `parseIdsParam` drops non-UUID entries, so placeholder ids would have made the batch-size assertions meaningless.

## 📝 Findings

### 🔹 Nit 1 — the `pageSize` justification comment describes a hazard that no longer exists

`packages/core/src/modules/customers/components/detail/authorUserLookup.ts:46-47`

> `// The route's pageSize defaults to 50, so a batch larger than that would be truncated by pagination even after it survived the ?ids= cap.`

`auth/api/users/route.ts:216-219` already derives `pageSize` as `min(ids.length, MAX_USER_LOOKUP_IDS)` for any request carrying `?ids=`. I confirmed that derivation is present in this PR's own base (`git show bd9356828:…/users/route.ts`) — it landed with #5634, the same PR that made `?ids=` actually filter. So the route's default of 50 cannot truncate a batch lookup, and the same sentence in the PR body's Problem §1 is inaccurate for the same reason.

Setting `pageSize` client-side is still worth keeping as defense-in-depth — it is what the devices sibling does, and it keeps the helper correct if that server-side derivation is ever removed. Only the stated reason is wrong. Suggested rewording: *"Belt-and-braces: the route already derives pageSize for `?ids=` requests (`users/route.ts:216-219`); setting it here keeps this helper correct if that derivation is removed."*

### 🔹 Nit 2 — a 2xx response with an unexpected body shape is still cached as "answered"

`packages/core/src/modules/customers/components/detail/authorUserLookup.ts:56-57`

```ts
if (!call || !call.ok) return null
return call.result?.items ?? []
```

`apiCall` is called with `{ fallback: null }`, and `readJsonSafe` returns that fallback when the body cannot be parsed. So a response that is `ok` but whose body is not JSON leaves `call.result === null`, `items` becomes `[]`, and `resolveAuthorUserNames` pushes the whole batch into `resolvedIds` — the ids are marked permanently resolved on a response the server never really gave. That is the same permanent-blank failure mode this PR exists to remove, surviving in one narrow path.

To be clear about how narrow: the route always returns `NextResponse.json(...)`, so this needs infrastructure interference — an interposed proxy or CDN returning a 200 HTML error page, or a truncated body. It is **not** triggered by the route's own 200-empty early returns (`route.ts:206`, `route.ts:226`), which emit a well-formed `{ items: [] }` and are correctly treated as "answered with nothing". Hence nit, not blocker.

One line makes the documented invariant actually hold:

```ts
if (!call || !call.ok || !Array.isArray(call.result?.items)) return null
```

`devices/backend/devices/userOptions.ts:53-54` has the identical hole, so fixing both together (or neither) is the consistent call.

### 🔹 Nit 3 — the two chunked user-lookup helpers will drift

`packages/core/src/modules/customers/components/detail/authorUserLookup.ts:43-85` vs `packages/core/src/modules/devices/backend/devices/userOptions.ts:43-108`

`fetchBatch` + `resolveAuthorUserNames` are a near-verbatim copy of `fetchUsers` + `resolveDeviceUserOptions`, down to shared comment text ("Imported rather than restated…", "URLSearchParams encodes on toString()…", the whole `resolvedIds` docblock). Both copies now independently encode four aspects of the same route contract: the `?ids=` cap, the `pageSize`-per-batch rule, the `x-om-forbidden-redirect: '0'` opt-out, and the null-means-the-call-failed convention. Nit 2 is a live demonstration — one defect, two places to fix.

**I am not requesting a change here.** The PR body makes this an explicit, reasoned scope decision ("extracting a shared module would mean moving the devices code too, which is out of scope for a fix"), the issue left it as an implementation call, and keeping a bug fix narrow is the right instinct. Recording it so the third call site is the trigger to extract a shared `resolveUserLabels(ids, formatLabel, signal)` rather than a third copy.

### 🔹 Nit 4 — the browser-level test has no tracked owner

The QA pass posted a ready-to-implement `TC-CUS-XXX-activity-author-lookup` scenario as a PR comment, but I could not find a corresponding open issue. Once this PR merges, that comment stops being anywhere anyone looks. The chunking contract is currently pinned only against a mocked transport, so the scenario is worth keeping — please file it as a follow-up issue before merge.

### 🔹 Nit 5 (trivial) — one test skips the file's own safe idiom

`__tests__/ActivitiesSection.authorLookup.test.tsx:185` reads `apiCallMock.mock.calls[0]` directly, whereas the file defines `userLookupBatches()` specifically to filter for `/api/auth/users?` calls. It is correct today — `ActivitiesSection` reaches for `apiCallOrThrow`/`readApiResultOrThrow` elsewhere, both separately mocked, so `apiCall` is only ever hit by this helper — but the filtering idiom exists for a reason and costs nothing to reuse.

## 🔁 Backward compatibility

No contract surface is touched: no API route, response shape, component prop, DB schema, event id, widget spot id, ACL feature, or exported type changes. `authorUserLookup.ts` is module-internal and reached only by a relative import from `ActivitiesSection.tsx`. The one user-visible difference — an author with a display name is now labelled with it instead of their email — is intentional, declared in the PR body, and pinned by a test.

## 🧪 Tests & CI

Coverage is proportionate and well-targeted: five cases pinning chunk boundaries, per-batch page sizing, the forbidden-redirect opt-out, the retry/no-retry split across a partially-failing pair of batches, and the `name`-over-email label. I assessed their strength by reading rather than by executing a mutation against the pre-fix file — CI's green `Test` step proves they pass, not that they would fail without the fix, and the structure gives me high confidence on that second point without an install in this review worktree. Flagging the distinction so it is not overstated.

**One disclosure on the CI signal.** The checks are green, but they were computed against an older base: the workflow run finished `2026-08-26T09:40Z`, and `develop`'s current head (`b41f7e3e5`, the v0.7.0 back-merge) landed at `2026-08-26T12:16Z`. The branch is 49 commits behind. GitHub reports `MERGEABLE` with no conflict, and I checked that none of the intervening commits touch `customers/components/detail/` or the `auth/users` route — so I see no reason to expect trouble. The merge queue will re-verify against the real base regardless; I mention it only so nobody reads "all green" as "green against today's `develop`".

## 🏁 Decision

**Approve.** The diagnosis is accurate, the fix addresses the root cause rather than the symptom, the invariant is stated and holds, and the tests pin it in a way that survives the cap changing. Nits 1, 2 and 5 are small enough to fold into any future touch of these files; nits 3 and 4 are follow-up material.

`needs-qa` stays on: this changes a rendered detail-page surface, so the automated-verification exemption does not apply, and the QA-approval gate holds the merge until a human adds `qa-approved`. The browser evidence already posted should make that sign-off quick rather than a from-scratch run.

pkarw:

**A prior automated UI QA pass already recorded a ✅ PASS on this PR** with before/after browser evidence for the chunking fix, including a run with the component reverted. This list is an independent, diff-derived read from code review — start with P0 §2 and P1, which that pass explicitly named as paths it could not construct.

### P0 — Author labels degrade safely for a viewer without `auth.users.list`

**Where to click**
- `/backend/customers/people/<id>` and `/backend/customers/companies/<id>` → the "Interaction history" panel
- Sign in as a user whose role grants `customers.*` but **not** `auth.users.list` (the automated pass could not construct this account — it is the highest-value gap)

**What to verify**
- The detail page loads and stays put. It must **not** bounce to `/login` — the lookup now sends `x-om-forbidden-redirect: '0'`, and this is the regression that header prevents.
- Activity entries render normally, simply **without** a "by {name}" line. No raw UUID should appear anywhere in the timeline.
- The activity list, filters, and search all still work — only the author label is degraded.

**What can go wrong**
- Redirect to `/login`, or a blank/error detail page, on a 403 from `/api/auth/users`.
- A bare UUID rendered where a name would normally be.
- The 403 surfacing as a toast or error banner — the lookup is meant to fail silently.

### P0 — Data scoping on the author lookup

**Where to click**
- Same routes, signed in as a non-superadmin in a tenant/org that has activities authored by several different users
- Switch the topbar organization/tenant context and reload

**What to verify**
- Every name shown belongs to a user visible in the current tenant/org scope. Authors outside scope simply show no name.
- Switching org context does not leak a name from the previous scope into the timeline.

**What can go wrong**
- A name from another tenant or organization appearing after a context switch (the lookup batches ids, so a scoping slip would surface several at once).

### P1 — More than 100 distinct authors resolve completely

**Where to click**
- A customer or company detail page whose "Interaction history" holds activities from **more than 100 distinct authors** (the fix's whole point; ~150 is the shape the tests pin)
- Open DevTools → Network, filter on `auth/users`

**What to verify**
- **Two or more** `GET /api/auth/users?ids=…` requests, not one. The first carries 100 ids, the last carries the remainder.
- Each request's `pageSize` equals its own id count, and `page=1`.
- Every activity that has a resolvable author ends up with a "by {name}" line — including the ones past the first 100. Scroll the whole list.

**What can go wrong**
- A single request with all ids, and the authors past the first 100 permanently blank (the original bug).
- Names appearing for the first 100 only, with the tail blank even after scrolling or re-filtering.

### P1 — A transient failure is retried, not remembered

**Where to click**
- Same page. In DevTools, set the network to offline (or block `/api/auth/users`) **while the page is loading**, then re-enable it and trigger a timeline refresh — change a filter, or log a new activity via "Log activity".

**What to verify**
- After the network recovers and the timeline refreshes, the author names **do** appear. The failure must not be sticky for the life of the component.
- No infinite request loop while the endpoint is blocked: with the call failing, the number of `auth/users` requests should stop growing rather than climb continuously.

**What can go wrong**
- Author names permanently blank after one failed request, only recovering on a full page reload (the second half of the original bug).
- A request storm against `/api/auth/users` while the endpoint is failing.

### P1 — Author labels now prefer the display name over the email

**Where to click**
- Any customer/company detail page → "Interaction history", with activities authored by a user who **has** a name set in `/backend/users`, and by one who does **not**

**What to verify**
- An author with a name set is now labelled **"by Ada Lovelace"**, where before this PR it read **"by ada@example.com"**. This is an intended, declared change — please confirm it reads correctly rather than treating it as a regression.
- An author with no name set still falls back to their email.

**What can go wrong**
- A blank label where a name or email should be (a user with neither would show no author line — acceptable).
- Truncation or layout breakage in the timeline row now that labels are typically longer than an email.

### P1 — Perceived performance on the changed route

**Where to click**
- Cold-load `/backend/customers/companies/<id>` (hard refresh, cache disabled) on a record with a long activity history

**What to verify**
- The timeline shell and activity text render immediately; author names may fill in a moment later. The lookup must not block the panel from painting.
- Batches are issued sequentially by design, so with >100 authors expect names to land in two waves — confirm this reads as progressive fill, not as a stall.
- Repeat at a mobile viewport (375px): the timeline stays readable and no author label overflows its row.

**What can go wrong**
- A visible spinner or empty panel held open until the author lookup finishes.
- Noticeable jank or a layout shift as names populate into rows that previously had no "by" line.

### Pass/fail

pkarw:
## 🏷️ Label rationale

- 🚦 **`merge-queue`** (was `review`) — the PR now carries an approving review from a reviewer who is not its author, so `reviewDecision` is `APPROVED` and `mergeStateStatus` is `CLEAN`. Approved PRs move to `merge-queue` whether or not QA is still outstanding.
- 📸 **`needs-qa`** *(kept)* — the automated-verification exemption does not apply: the change alters what a rendered surface displays (`ActivitiesSection` / the "Interaction history" panel), including a deliberate label change from email to display name. With `qaGate` on, this label holds the merge until a human adds `qa-approved`, even though every check is green. The `qa` label is not set here — that belongs to the QA reviewer when they start testing.
- 🐛 **`bug`** *(kept)* — fixes defective behaviour reported in #5632; adds no new capability.
- 🟩 **`priority-low`** *(kept)* — the defect needs >100 distinct activity authors on one record, or a failed lookup, to show at all, and it degrades a label rather than blocking work.
- 🟩 **`risk-low`** *(kept)* — three files, all inside `customers/components/detail`, no contract surface touched (no API route, response shape, prop, schema, ACL feature, or exported type). The new helper is module-internal.

`in-progress` is removed with this run's lock release; it is a meta label, outside the pipeline transition.

pkarw:
🤖 `om-auto-review-pr` completed: **APPROVED**. No blockers, no majors; five nits recorded in the review (a stale `pageSize` justification comment, an unparseable-2xx hole in the `resolvedIds` invariant shared with the devices sibling, the duplication between the two chunked user-lookup helpers, the untracked browser-level test, and one test that skips the file's own filtering idiom). None were fixed here — `autofix: skipped (not my PR — re-run with --autofix to fix it here)`. Validation source: GitHub checks on `3e42f698f` (14 pass, 0 failing, 0 pending), with the disclosure that they were computed ~2.5h before `develop`'s current head landed; the branch is 49 commits behind but `MERGEABLE` and none of those commits touch the changed files. Pipeline moved `review` → `merge-queue`; `needs-qa` retained, so the QA-approval gate still holds the merge. Lock released.

PR #5626, opened 2026-08-25
docs(changelog): extend the 0.7.0 entry with main merges through 2026-08-25

patzick:
## Summary

Regenerates the `0.7.0` changelog window against the newest `origin/main` and folds in every shipped PR the previous rebuild (#5416, 2026-08-19) could not have seen. Only `CHANGELOG.md` is modified.

The existing entry was audited rather than rewritten: no PR listed in it is unreachable from `main`, no PR number is duplicated inside the block, and the release-workflow changelog guards still pass. This change is therefore purely additive — 12 new bullets and 1 new contributor.

## How the window was built

Reachability mode per `om-auto-update-changelog`'s release-window contract, with `--release-ref main` (the release is cut from `main`, not from the configured `develop` base):

- `git rev-list v0.6.7..origin/main` → 1202 commits.
- **list-prs** (state merged) across nine date chunks spanning `2026-07-28 → 2026-08-26`, so no chunk approached the 250 cap and nothing was silently truncated. **501 PRs enumerated.**
- Kept the **350** whose `mergeCommit.oid` is reachable on `main`; dropped **151** that live only on `develop` or on the `feat/agent-orchestrator-mvp` branch and are therefore not part of this release.
- **21** reachable PRs were absent from the entry; **9** were excluded as plumbing and **12** are added here.

### Excluded (9)

| PR | Reason |
|----|--------|
| #4674, #4783, #5226, #5416 | Prior runs of this skill — changelog PRs are not release content |
| #5014, #5127, #5376 | Branch-sync plumbing (`chore: sync …`, `chore: develop merge before release`) |
| #5485 | `chore: release v0.7.0` — release plumbing, authored by `app/github-actions` |
| #4762 | The `develop` mirror of #4761, which is already credited in the 0.6.7 entry |

## Credit verification

All 12 credits were compared against commit authorship (**get-pr** with `commits`). Eleven matched the PR author outright. One mismatch:

- **#5622** — the merged PR author wrote **0 of 1** commits and the body carries no `Credit:` / `Supersedes` template. The body states the change "already landed on `develop` in #5470"; #5470 is authored by **@Duang777**, and the single commit on #5622 is authored by `Duang777` (co-authored by `cursoragent`, an excluded AI identity). Credited as `*(@Duang777, via @patzick)*` — a cherry-pick carry-forward, not authorship by the porter. #5470 itself is not listed separately: it is unreachable from `main` and is the `develop` twin of the same fix.

[...]

PR #5622, opened 2026-08-25
fix(sales): serve persisted order totals on single-row GET (#5438)

patzick:
## Summary

Ports the fix for #5438 to `main`. The change already landed on `develop` in #5470 (squash commit `136748c`), but `main` never received it, so the defect is still live on the `main` line.

This is a clean cherry-pick of `136748c` — no conflicts, no adaptation needed.

- `GET /api/sales/orders` returned **different grand totals for the same order** depending on whether the response contained exactly one row.
- The list path served the persisted columns (arithmetically correct: lines ± adjustments). The detail path (`items.length === 1`) ran `recalculateOrderTotalsForDisplay`, which re-emits shipping rates and the Stripe `applicationFee` on top of already-stored adjustments. Outstanding balance could flip from `0` (paid) to a non-zero remainder.
- GET now returns the persisted totals for every result size, matching quotes and matching the amount filters. Write paths still persist totals through `salesCalculationService` on create/update/return.

## Test plan

- [x] `packages/core/src/modules/sales/api/__tests__/documents.factory.test.ts` — 29 passed, including the `#5438` regression test (single-row `afterList` no longer calls the recalculator or `em.fork()`)
- [x] Full `packages/core/src/modules/sales` suite — 611 tests passed
- [ ] CI covers the remaining gate (`build:packages`, `generate`, `i18n:check-*`, `typecheck`, full `test`, `build:app`)

> Two `src/modules/sales` suites (`channels.route`, `shipments.afterList`) could not run in the local worktree because `packages/core/generated/` had not been built there. They resolve `#generated/entities/sales_channel` and are unrelated to this diff; CI runs `yarn generate` before the test job.

## Breaking Changes

| Surface | Status |
|---------|--------|
| `GET /api/sales/orders` path and query schema | Unchanged |
| Response envelope | Unchanged; single-row responses now match the multi-row (persisted) values |
| OpenAPI | Unchanged |
| Database schema | Untouched |
| `recalculateOrderTotalsForDisplay` | Still exported; documented as not for GET overwrite |

The only behavioral change is the intended one: a one-row GET no longer inflates shipping/fees relative to the list.

Related: #5470 (the `develop` PR this ports), #2234 (the same recompute, filed as a performance issue).

Closes #5438

---
[...]

PR #5655, opened 2026-08-26
chore: sync main (v0.7.0) back into develop — DO NOT SQUASH, merge commit required

patzick:
> ## ⚠️ DO NOT SQUASH — merge this with "Create a merge commit"
>
> Squashing this PR silently defeats its entire purpose and guarantees the same
> conflicts come back on the next sync. Details in
> [Why the merge strategy matters](#why-the-merge-strategy-matters) — please read
> that section before merging.

## Summary

Post-release sync of `main` (v0.7.0, released 2026-08-26) back into `develop`.
`main` carried **45 commits** develop was missing. Branched from `origin/main`'s
tip `8b492325d`, merged into `origin/develop`'s tip `bd9356828`.

| Landed on `main` | What |
|---|---|
| #5623 | `fix(auth)`: preserve partial user ACL updates |
| #5622 / #5438 | `fix(sales)`: serve persisted order totals on single-row GET |
| #5530 | `fix(release)`: bump `apps/*` with the monorepo version |
| #5529 | `fix(auth)`: surface login failures instead of an empty 500 |
| #5479 / #5457 | `fix(customer_accounts)`: resolve the portal address server-side to stop hydration flicker |
| #5489 | `fix(eudr)`: nav label, module breadcrumb, false unsaved-changes prompt |
| #5481 | `fix(ui,sales,rules,record-locks)`: 0.7.0 pre-release UX and i18n nits |
| #5486 | `feat(app)`: hide the contact/feedback widget behind `ff_om_hide_contact` |
| #5480 / #5461 | `fix(eudr)`: stop offering Delete on archived statements |
| #5485 | `chore`: release v0.7.0 |
| #5432 | `fix(ui)`: align WMS KPI cards, role dialog footer, EUDR area field |
| #5431 | `fix(docker)`: sync role ACLs on redeploy |
| #5626, #5416 | `docs(changelog)`: 0.7.0 entry rebuild and extension |

## Why the merge strategy matters

`main` is **not** an ancestor of `develop` today:

```
$ git merge-base --is-ancestor origin/main origin/develop   → false
$ git rev-list --count origin/develop..origin/main          → 45
```

A squash creates a *new* commit on `develop` whose content happens to match, but
which shares no ancestry with `main`. Git therefore still considers the branches
diverged, so:

- `main`'s commits never become reachable from `develop`
- the next sync re-conflicts on the same files, resolved the same way, forever
- `git log develop --contains <main-sha>` stays empty, so "is this fix on
  develop?" cannot be answered from history

This PR's history contains a real merge commit (`12ffcac59`, two parents). Merged
**with a merge commit**, it makes `main` a genuine ancestor:

```
$ git merge-base --is-ancestor origin/main HEAD   → true
$ git rev-list --count HEAD..origin/main          → 0
```

Squash it and both revert to the broken state, and the next sync inherits the
same conflict set. This is exactly what happened between #5014 (squashed) and
#5127, which had to re-resolve the identical six dependency files.

## Conflicts and how they were resolved

Three files conflicted. Everything else auto-merged.

### 1. `UPGRADE_NOTES.md` — section structure

Base carried two headings: `## 0.7.0 → 0.7.1 (unreleased)` and
`## 0.6.7 → 0.7.0 (2026-08-12)`. The v0.7.0 release on `main` folded the first
into the second and re-dated it to `2026-08-26`; meanwhile `develop` kept the
unreleased heading and filed four new entries under it. Git collapsed the two
sections into one.

Resolved by restoring both sections and taking the **union** of their entries:

| Section | Entries |
|---|---|
| `## 0.7.0 → 0.7.1 (unreleased)` | `AlertDescription` `<div>` (#5487), device API camelCase (#5513), optional participant `userId` (#5115), CRM deal status vocabulary (#5107) — all develop-only, all still unreleased |
| `## 0.6.7 → 0.7.0 (2026-08-26)` | `PUT /api/auth/users/acl` partial updates (#5493) and `loadSidebarPreference` deprecation from main, plus develop's list-total cap, `onboarding_requests.email` constraint (#4514) and `GET /api/attachments` paging (#5299) entries |

No entry from either side was dropped.

### 2. `packages/core/src/modules/auth/api/users/acl/route.ts` — architectural overlap

Both branches rewrote this route for different reasons:

- **develop** moved the write off `withAtomicFlush` and onto the command bus
  (`AUTH_USER_ACL_UPDATE_COMMAND_ID`), so the permission change lands in the
  action log, and added `resolveAclTenantId` so a tenant-less super admin resolves
  the scope from the target user.
- **main** (#5493) fixed the merge semantics: an omitted `features` /
  `organizations` / `isSuperAdmin` must keep its stored value instead of being
  normalized to a cleared one, plus a 400 guard refusing an
  organization-restricted override with no feature grant.

These are orthogonal, so both were kept: **develop's command-bus write path and
tenant resolution, with main's partial-update semantics layered on top.**
Concretely — the `existing*` reads, `featuresWereProvided`, `requestedOrganizations`,
the `!featuresWereProvided` sanitize bypass, `hasOrganizationRestriction`, the
400 guard and the updated OpenAPI description all come from main; the row lookup
uses develop's resolved `tenantId` (not `auth.tenantId`), and the persist block
is develop's `commandBus.execute(...)` fed `organizations: requestedOrganizations`.
main's `withAtomicFlush` persist/remove block was dropped as superseded.

`main`'s companion test (`__tests__/partial-update.test.ts`) asserted the
pre-command direct-write path — `em.persist` / `em.remove` call counts and
mutations of the stored row — which no longer exists on this branch. It was
rewritten in `28507ebf2` to assert the same six partial-merge behaviours against
the command-bus input (`features`, `organizations`, `isSuperAdmin`, `clear`),
matching the style of develop's neighbouring `sanitized-request.test.ts`. **No
assertion was weakened or removed**; all four ACL route suites pass (23 tests).

### 3. `customer_accounts/.../users/page.tsx` — server/client split vs. a field addition

`main` (#5457) split the page into a server component that resolves the portal
origin plus an extracted `PortalUsersPageClient`; `develop` had added
`totalIsCapped` pagination support (four lines) to the old single-file client
page. Took main's split wholesale and ported develop's four lines into
`PortalUsersPageClient.tsx`.

## Version alignment — a defect this PR owns

The first CI run went red on `test`, at
`scripts/__tests__/check-version-alignment.test.mjs`:

```
+ '@open-mercato/channel-discord@0.6.7 (expected 0.7.0)',
+ '@open-mercato/documents@0.6.7 (expected 0.7.0)'
```

`channel-discord` and `documents` were created **on develop, after main
branched**, so main's v0.7.0 release bump never reached them — they are absent
from `main` entirely. Neither branch is broken alone: develop is internally
consistent at 0.6.7, main is internally consistent at 0.7.0. Merging raises the
monorepo version to 0.7.0 and strands exactly these two packages. That makes it a
defect **this merge owns**, not one it inherits.

Fixed in `05cbf1f77` by moving both `.version` fields to 0.7.0. Their intra-repo
dependencies are all `workspace:*`, and `scripts/bump-version.sh` only rewrites
`.version`, so nothing else has to move. `./scripts/check-version-alignment.sh`
now passes.

I originally missed this because `yarn test` runs `turbo run test`, which does
not cover `scripts/__tests__` — those run under the separate `yarn test:scripts`.
It and `yarn test:repo-wide-guards` are now in the table below.

## Validation

Runner: **local** (no `app` container running for any `docker-compose*` file).

| gate | result |
|---|---|
| `yarn install --immutable` | ✅ no lockfile drift |
| `yarn build:packages` | ✅ 27/27 |
| `yarn generate` | ✅ 348 artifacts refreshed, no tracked-file drift |
| `yarn typecheck` | ✅ 27/27 |
| `yarn lint` | ✅ 0 errors (10 pre-existing warnings) |
| `yarn i18n:check-sync` | ✅ 57 modules, 5 locales in sync |
| `yarn i18n:check-usage` | ✅ advisory only (3828 unused keys, pre-existing) |
| `yarn test` | ✅ per-package; see the flake note below |
| `yarn test:scripts` | ✅ 680/681 (1 skipped) — green after `05cbf1f77` |
| `yarn test:repo-wide-guards` | ✅ 35 test files |
| `yarn check:dep-versions` | ✅ no major conflicts |
| `yarn ds:tokens:check` | ✅ token snapshot in sync |
| `yarn agents:check-budget` | ✅ (4 chains now under their baseline) |
| `yarn build:app` | ✅ |

`build:app` needs `build:packages` to run *after* `generate` (`packages/core/dist/generated/entities.ids.generated.js` is emitted by that second pass), which is why the CI command list builds packages twice. Running it in that order is green.

### Test-run flakes, not defects

Two unrelated environment/concurrency issues showed up locally and are recorded
so they don't get misread as merge fallout. **CI is the source of truth for this
head.**

1. **jest worker `SIGSEGV`** (already tracked as #5100). Under `turbo`'s
   concurrency one package's worker dies per run — `@open-mercato/shared` on one
   run, `@open-mercato/channel-expo` on the next, i.e. it does not stick to a
   package. Each one passes cleanly in isolation
   (`shared`: 184 suites / 2015 tests; `channel-expo`: 5 suites / 24 tests).
2. **`listen EINVAL` on a tsx IPC pipe.** `dynamicLoader.generatedCacheRecovery`
   and the `i18n:check-*` scripts fail when `TMPDIR` sits deep inside a worktree
   path, because the resulting unix-socket path exceeds the macOS ~104-byte
   limit. `TMPDIR=/tmp` makes all of them pass. Purely local to this checkout
   layout; nothing in the repo changes.

## Merge order

Merge with **"Create a merge commit"**. Do not rebase and do not squash — either
one discards the merge commit and with it the ancestry this PR exists to
establish.

### One pre-existing failure left alone

`yarn template:sync` reports drift in one file, `modules.ts`. It is **not** from
this merge: `packages/create-app/template/src/modules.ts` is byte-identical
between `origin/develop` and this branch, and `main` never touched it after the
merge base. It reproduces on `develop` alone. Fixing it means regenerating the
create-app template, which is outside a sync PR's scope — flagging it here rather
than folding an unrelated change into a 148-file merge.


patzick:
Labels applied by the automation that opened this PR:

- `review` — ready for review; not in another pipeline state.
- `priority-high` — a release-sync PR that is currently the only thing making `main`'s v0.7.0 fixes (auth ACL, login 500, sales totals) reachable from `develop`. Every day it sits open, more develop work is written against a base that lacks them, and the conflict surface grows.
- `risk-high` — 148 files / +4512 −1847 across auth, sales, customer_accounts, UI and release tooling, and one of the three conflict resolutions merges two independent rewrites of the user-ACL write path. Blast radius is wide even though every gate is green.
- `needs-qa` — the ACL route resolution and the customer_accounts users-page resolution are both behaviour-carrying merges of two branches' changes, not mechanical takes. Worth exercising the user ACL editor (partial saves: organizations-only, features-only, full clear) and the customer portal users list by hand.

PR #5661, opened 2026-08-26
fix(data-sync): validate schedule before persisting it (#5506)

### LINKED ISSUE #5506: bug: [Data sync] a recurring schedule is saved even though the API returns 422
## Summary
`POST /api/data_sync/schedules` writes the schedule row **first** and computes `nextRunAt` **after**. When the
interval format is not understood the computation fails, the endpoint answers **422**, and the row stays —
with `nextRunAt: null`. The operator is told it failed while a schedule that can never fire now exists.

## Steps to reproduce (real UI path)
1. Open `/backend/data-sync` as an admin.
2. In *Harmonogram cykliczny* set **Typ harmonogramu = Interwał**.
3. In the **Interwał** field type `3600` — a plausible "seconds" guess. The field's own hint says
   *"Przykład: `1h`, `6h` lub `24h`"*, but nothing enforces it: `aria-invalid` is not set and the save button
   stays enabled.
4. Save, then reload.

## Expected vs actual
| | Expected | Actual |
|---|---|---|
| Client-side validation | the documented format is enforced before submit | none |
| On an unparseable interval | reject without writing anything | **422** and the row is written |
| Stored row | absent | `scheduleType:"interval"`, `scheduleValue:"3600"`, **`nextRunAt: null`** |
| The page afterwards | no schedule | **"Harmonogram cykliczny aktywny"** — it advertises as active a schedule that can never run |

API confirmation (2026-08-22, 0.7.0 build):
```
POST /api/data_sync/schedules
  {integrationId:'sync_excel', entityType:'customers.person', direction:'import',
   scheduleType:'interval', scheduleValue:'3600', isActive:true}
→ 422 {"error":"Failed to calculate next run time for schedule: 3545ccea-9ae2-4db4-9df9-4e533a8fc14c"}
GET /api/data_sync/schedules  → total 0 → 1
```

## Mitigating facts, established rather than assumed
1. It is an **upsert**: repeating the identical failing request returns the same id and the count stays at 1,
   so retries do not multiply orphans.
2. A subsequent valid save (`scheduleType:'cron'`) returns 201, reuses the same id and repairs the row.
3. The page has a working **Usuń harmonogram** button, so the operator is not stuck.

## Secondary
The 422 body leaks the primary key of the freshly created row.

adeptofvoltron:
Fixes the server-side half of #5506 — the write-ordering defect. The issue also reports a
client-side gap (the documented interval format is not enforced before submit, `aria-invalid`
is not set, and the save button stays enabled); that half needs a UI change and is tracked
separately in #5872, so **#5506 should stay open until #5872 lands**. Review follow-up on the
remaining orphaned-`ScheduledJob` window: #5871.

## 🎯 Goal
Data sync schedules are rejected (422) atomically — an unparseable interval/cron value no longer leaves a broken, unschedulable row behind.

## 🔍 Problem
`POST /api/data_sync/schedules` wrote the `SyncSchedule` row and flushed it to the database *before* asking the scheduler to compute `nextRunAt`. When the interval value couldn't be parsed (e.g. `3600` instead of `1h`), the scheduler threw, the endpoint answered 422 — but the row was already committed with `nextRunAt: null`. The UI then showed "Harmonogram cykliczny aktywny" for a schedule that could never run, and the 422 body leaked the freshly-created row's primary key.

## 🔍 Root Cause
`saveSchedule()` in `packages/core/src/modules/data_sync/lib/sync-schedule-service.ts` built, persisted, and flushed the `SyncSchedule` entity, and only afterwards called `schedulerService.register(...)`. `SchedulerService.register()` (`packages/scheduler`) computes `nextRunAt` via `calculateNextRunForWrite` and throws synchronously — before it ever touches its own `ScheduledJob` table — when the value can't be parsed. Because the `SyncSchedule` write had already committed by that point, the thrown error only produced the 422 HTTP response; the invalid row stayed in the database.

## What Changed
- `packages/core/src/modules/data_sync/lib/sync-schedule-service.ts` — reordered `saveSchedule()` so `schedulerService.register(...)` is called first, using a pre-generated `id`/`scheduledJobId` and the raw `input` fields (not the persisted row). The `SyncSchedule` entity is only created/updated and flushed once `register()` resolves without throwing. `buildScheduleName`/`buildScheduleDescription` now take a narrow `{ integrationId, entityType, direction }` shape so they can run before the row exists. Response contract (201 on success, 422 with the scheduler's error message on failure) and upsert/idempotency behavior are unchanged.

## 🧪 Tests
- Added `packages/core/src/modules/data_sync/api/schedules/__tests__/schedule-write-order.test.ts`: a rejecting scheduler now produces a 422 and `em.create`/`em.persist`/`em.flush` are never called. Verified this test fails against the pre-fix code with the exact reported symptom (row gets created/flushed despite the scheduler rejecting it), and passes after the fix.
- Full validation gate (`.ai/agentic.config.json`): `build:packages` → `generate` → `build:packages` → `i18n:check-sync` → `i18n:check-usage` → `typecheck` → `test` → `build:app` — all green, except the pre-existing, environment-specific `packages/cli/src/lib/__tests__/dev-env-reload.test.ts` inotify-exhaustion timeout (unrelated package; documented pre-existing flake on this host, not introduced by this change).
- `packages/core` (227 tests, includes the existing `data_sync/api/schedules/__tests__/optimistic-lock.test.ts` suite) and `packages/scheduler` (420 tests) also run directly in isolation — 100% pass.

## 💥 Breaking Changes
None — request/response shapes, DI wiring, and upsert semantics are unchanged.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5661"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790333362&installation_model_id=432243&pr_number=5661&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5661&signature=22c181c38cce0c9a50cbd21d90b8fcd45cb1f322d6a006cbed7007aa63be9e21"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/data_sync/api/schedules/__tests__/schedule-write-order.test.ts
+++ b/packages/core/src/modules/data_sync/api/schedules/__tests__/schedule-write-order.test.ts
@@ -1,0 +1,84 @@
+/** @jest-environment node */
+
+const mockGetAuthFromRequest = jest.fn()
+
+const mockEm = {
+  create: jest.fn((_entityClass: unknown, data: Record<string, unknown>) => ({ ...data })),
+  persist: jest.fn(),
+  flush: jest.fn(async () => undefined),
+}
+
+const mockScheduler = {
+  register: jest.fn(async () => {
+    throw new Error('Failed to calculate next run time for schedule: some-id')
+  }),
+  unregister: jest.fn(async () => undefined),
+}
+
+jest.mock('@open-mercato/shared/lib/auth/server', () => ({
+  getAuthFromRequest: jest.fn((req: Request) => mockGetAuthFromRequest(req)),
+}))
+
+jest.mock('@open-mercato/shared/lib/encryption/find', () => ({
+  findOneWithDecryption: jest.fn(async () => null),
+  findAndCountWithDecryption: jest.fn(async () => [[], 0]),
+}))
+
+jest.mock('@open-mercato/shared/lib/http/readJsonSafe', () => ({
+  readJsonSafe: jest.fn((req: Request) => req.json()),
+}))
+
+const { createSyncScheduleService } = jest.requireActual('../../../lib/sync-schedule-service')
+
+const scheduleService = createSyncScheduleService(mockEm, mockScheduler)
+
+const mockContainer = {
+  resolve: jest.fn((token: string) => {
+    if (token === 'dataSyncScheduleService') return scheduleService
+    if (token === 'commandOptimisticLockGuardService') return null
+    return null
+  }),
+}
+
+jest.mock('@open-mercato/shared/lib/di/container', () => ({
+  createRequestContainer: jest.fn(async () => mockContainer),
+}))
+
+import { POST } from '../route'
+
+function request() {
+  return new Request('http://localhost/api/data_sync/schedules', {
+    method: 'POST',
+    headers: { 'content-type': 'application/json' },
+    body: JSON.stringify({
+      integrationId: 'sync_excel',
+      entityType: 'customers.person',
+      direction: 'import',
+      scheduleType: 'interval',
+      scheduleValue: '3600',
+      timezone: 'UTC',
+      fullSync: false,
+      isEnabled: true,
+    }),
+  })
+}
+
+describe('data_sync schedule save write ordering', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockGetAuthFromRequest.mockResolvedValue({ sub: 'user-1', tenantId: 'tenant-1', orgId: 'org-1' })
+  })
+
+  it('does not persist the schedule when the scheduler rejects an unparseable value', async () => {
+    const res = await POST(request())
+
+    expect(res.status).toBe(422)
+    const body = await res.json()
+    expect(body.error).toContain('Failed to calculate next run time')
+
+    expect(mockScheduler.register).toHaveBeenCalledTimes(1)
+    expect(mockEm.create).not.toHaveBeenCalled()
+    expect(mockEm.persist).not.toHaveBeenCalled()
+    expect(mockEm.flush).not.toHaveBeenCalled()
+  })
+})
--- a/packages/core/src/modules/data_sync/lib/sync-schedule-service.ts
+++ b/packages/core/src/modules/data_sync/lib/sync-schedule-service.ts
@@ -40,11 +40,11 @@ export function createSyncScheduleService(em: EntityManager, schedulerService?: 
     return schedulerService
   }
 
-  function buildScheduleName(row: SyncSchedule): string {
+  function buildScheduleName(row: { integrationId: string; entityType: string; direction: 'import' | 'export' }): string {
     return `Data sync: ${row.integrationId} ${row.entityType} ${row.direction}`
   }
 
-  function buildScheduleDescription(row: SyncSchedule): string {
+  function buildScheduleDescription(row: { integrationId: string; entityType: string; direction: 'import' | 'export' }): string {
     return `Scheduled ${row.direction} for ${row.integrationId} (${row.entityType})`
   }
 
@@ -156,8 +156,36 @@ export function createSyncScheduleService(em: EntityManager, schedulerService?: 
         })
       }
 
+      const id = existing?.id ?? randomUUID()
+      const scheduledJobId = existing?.scheduledJobId ?? id
+
+      // Validate the schedule (and register it with the scheduler) before writing
+      // the SyncSchedule row — an unparseable scheduleValue must not leave a
+      // persisted row with no working schedule behind it.
+      await requireScheduler().register({
+        id: scheduledJobId,
+        name: buildScheduleName(input),
+        description: buildScheduleDescription(input),
+        scopeType: 'organization',
+        organizationId: scope.organizationId,
+        tenantId: scope.tenantId,
+        scheduleType: input.scheduleType,
+        scheduleValue: input.scheduleValue,
+        timezone: input.timezone,
+        targetType: 'queue',
+        targetQueue: 'data-sync-scheduled',
+        targetPayload: {
+          scheduleId: id,
+          scope,
+        },
+        requireFeature: 'data_sync.run',
+        sourceType: 'module',
+        sourceModule: 'data_sync',
+        isEnabled: input.isEnabled,
+      })
+
       const row = existing ?? em.create(SyncSchedule, {
-        id: randomUUID(),
+        id,
         integrationId: input.integrationId,
         entityType: input.entityType,
         direction: input.direction,
@@ -178,35 +206,13 @@ export function createSyncScheduleService(em: EntityManager, schedulerService?: 
       row.timezone = input.timezone
       row.fullSync = input.fullSync
       row.isEnabled = input.isEnabled
-      row.scheduledJobId = row.scheduledJobId ?? row.id
+      row.scheduledJobId = scheduledJobId
 
       if (!existing) {
         em.persist(row)
       }
 
       await em.flush()
-
-      await requireScheduler().register({
-        id: row.scheduledJobId,
-        name: buildScheduleName(row),
-        description: buildScheduleDescription(row),
-        scopeType: 'organization',
-        organizationId: scope.organizationId,
-        tenantId: scope.tenantId,
-        scheduleType: row.scheduleType,
-        scheduleValue: row.scheduleValue,
-        timezone: row.timezone,
-        targetType: 'queue',
-        targetQueue: 'data-sync-scheduled',
-        targetPayload: {
-          scheduleId: row.id,
-          scope,
-        },
-        requireFeature: 'data_sync.run',
-        sourceType: 'module',
-        sourceModule: 'data_sync',
-        isEnabled: row.isEnabled,
-      })
 
       return row
     },


adeptofvoltron:
🤖 Label rationale:
- **review** — ready non-draft PR, no other pipeline state applies.
- **bug** — fixes a reported defect (data_sync schedule row persisted despite a 422 response).
- **skip-qa** — no UI-rendering file touched (no `.tsx`, nothing under `packages/ui/src/` or `**/components/**`), API surface and DB structure unchanged, no `BACKWARD_COMPATIBILITY.md` contract broken, and automated regression coverage ships in this PR (`schedule-write-order.test.ts`) — qualifies for the automated-verification exemption.
- **priority-medium** — real data-integrity bug, but non-blocking: the row is idempotent on retry, a subsequent valid save repairs it, and the operator has a working delete button.
- **risk-low** — a single localized reorder inside `saveSchedule()` in one service file; response contract, DI wiring, and valid-input behavior are unchanged; full validation gate green plus targeted `packages/core`/`packages/scheduler` suites at 100%.

adeptofvoltron:

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock — starting review. Started: 2026-08-26T10:51:45Z.

adeptofvoltron:
@/tmp/pr5661-review-comment.md

adeptofvoltron:

Re-confirmed on the 2026-09-09 self-review pass (head `b871ef2`, unchanged). **No label changes were needed** — the existing set is already correct for an approved PR, with exactly one priority and one risk label. The `in-progress` label applied for this run's claim has been removed with the lock.

- 🚀 `merge-queue` — the review verdict on head `b871ef2` is ✅ approve: no blocker and no major findings across four passes. CI on this exact head is fully settled and green (14 passed, 5 skipped by their own workflow conditions, none failing, none pending). The only thing between this PR and a merge is an independent approving review — GitHub refuses to record a formal approval from this account because it authored the PR, so `merge-queue` records review readiness while a human still has to click Approve.
- 🐛 `bug` — fixes a reported defect (#5506): a `data_sync` schedule row was persisted despite the API answering 422.
- ⏭️ `skip-qa` — re-checked against the current diff and still valid under the automated-verification exemption: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, DB structure unchanged (no migration, entity untouched), API surface unchanged, no `BACKWARD_COMPATIBILITY.md` contract broken, and automated regression coverage ships in the same PR (`schedule-write-order.test.ts`, re-proven this pass to fail against pre-fix code).
- 🟠 `priority-medium` — a real data-integrity bug with a user-visible symptom (the UI reports an active recurring schedule that can never run), but non-blocking: the operator can delete the broken row and the failure is confined to one feature.
- 🟢 `risk-low` — 32 added and 26 removed lines in one module-private service function, no contract surface moved, no schema change, and the reordering is covered by a regression test plus the module's existing 227-test suite.

adeptofvoltron:
🤖 `om-auto-review-pr` completed: APPROVED (posted as a comment — GitHub blocked a formal self-approval since this PR's author and this review's identity are the same account). No blocker or major findings; full validation gate green (`packages/core` 227 tests, `packages/scheduler` 420 tests, typecheck, build:app all passing); autofix: n/a — nothing to fix. Labels normalized to `merge-queue` (from `review`), `bug`/`skip-qa`/`priority-medium`/`risk-low` kept as-is. Lock retained — chain continues.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: fix implemented and validated, review approved (self-approval blocked by GitHub — needs an independent human approval), moved to `merge-queue`. `test` and `ephemeral-integration` CI checks still pending — swapping to `ci-monitoring` for follow-up. Lock released.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-09-03T17:38:58Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron:
🤖 `om-auto-review-pr` — code review verdict: **✅ APPROVE**

> ⚠️ GitHub refused to record this as a formal approving review — `addPullRequestReview` answers *"Can not approve your own pull request"*, because this PR's author and this review's identity are the same account. The review is therefore posted as a comment, and **an independent human reviewer still has to click Approve** before this PR can actually merge. Nothing here substitutes for that.

---

# 🔍 Code Review: fix(data-sync): validate schedule before persisting it (#5506)

## 🎯 Summary

This change reorders `saveSchedule()` in `packages/core/src/modules/data_sync/lib/sync-schedule-service.ts` so the scheduler registration happens *before* the `SyncSchedule` row is created and flushed, instead of after it. The id and `scheduledJobId` are pre-computed (`existing?.id ?? randomUUID()`), the registration payload is built from the raw `input` rather than from a persisted row, and `buildScheduleName` / `buildScheduleDescription` were narrowed from taking a full `SyncSchedule` entity to taking a `{ integrationId, entityType, direction }` shape so they can run before any row exists. That is the correct and minimal shape of the fix for the reported defect: `SchedulerService.register()` throws synchronously out of `calculateNextRunForWrite` for an unparseable interval, so hoisting it above the write is exactly what stops the endpoint from answering 422 while leaving a committed row with `nextRunAt: null` behind.

The reviewed scope is the two files in the diff, read against their real call sites — `api/schedules/route.ts` (POST), `api/schedules/[id]/route.ts` (PUT/DELETE), the DI wiring in `data_sync/di.ts`, the peer contract in `packages/scheduler/src/modules/scheduler/services/schedulerService.ts`, and the consumer in `data_sync/workers/sync-scheduled.ts`. Blast radius is genuinely small: `dataSyncScheduleService` is resolved from those two route files and nowhere else.

I want to call out what is good about this PR, because it is more than the diff size suggests. The regression test is a real one rather than a restatement of the implementation: I reverted `sync-schedule-service.ts` to its `origin/develop` version and ran `schedule-write-order.test.ts` against it, and it fails with precisely the reported symptom — `expect(mockEm.create).not.toHaveBeenCalled()` receiving one call, with the `scheduleValue: "3600"` payload from the issue. It passes on the PR head. The equivalence of the update path is also preserved carefully: `scheduledJobId` moved from `row.scheduledJobId ?? row.id` to `existing?.scheduledJobId ?? id` and `targetPayload.scheduleId` from `row.id` to `id`, both of which are identical values on every path, and the pre-existing `optimistic-lock.test.ts` suite exercises the `existing`-row branch and still passes. The optimistic-lock and record-gone guards remain ahead of the registration, so a stale-version edit still aborts before either write.

I found no blockers and no majors. The two minors below are worth acting on, but neither should hold this PR: one is a narrower residual of the same class of problem that a follow-up can harden, and the other is a scoping/claim accuracy point about the linked issue rather than a defect in this code.

## Verdict

✅ approve — the reordering is correct on both the create and the update path, it is covered by a regression test that provably fails against the pre-fix code, no contract surface moves, and the full configured validation gate is green apart from two failures that are demonstrably unrelated to this diff. The remaining findings are minors with concrete follow-up remedies, and under the `om-code-review` verdict rule minors and nits do not block a merge.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | 27/27 tasks successful. |
| `yarn generate` | ✅ PASS | Generated registries rebuilt with no drift. |
| `yarn build:packages` | ✅ PASS | Second pass per the configured gate order; clean. |
| `yarn i18n:check-sync` | ✅ PASS | All translation files in sync. |
| `yarn i18n:check-usage` | ✅ PASS | 20,831 keys scanned, no unused/missing usage reported. |
| `yarn typecheck` | ✅ PASS | 27/27 tasks successful. |
| `yarn test` | ⚠️ PASS for the changed code, with two unrelated failures | Explained in full below. |
| `yarn build:app` | ✅ PASS | Production app build succeeds. |

On `yarn test`: the aggregate run exits 1 because `@open-mercato/cli` reports 5 failures out of 1,806 in `src/lib/__tests__/resolve-environment.test.ts` and `src/lib/__tests__/resolver.enterprise.test.ts`. Those are the known checkout-location-dependent suites in a package this PR does not touch, and the `test` check is green in CI. Because Turbo aborts the remaining tasks on the first failing one, 12 packages — including `@open-mercato/core`, the only package this PR changes — never ran in the aggregate pass, so I re-ran the two that matter standalone. `@open-mercato/scheduler` is 420/420 green. `@open-mercato/core` is 11,610 passed / 1 failed / 2 skipped out of 11,613, and the single failure is `src/modules/staff/backend/staff/timesheets/__tests__/page.durationEntry.test.tsx` — a `staff` module timesheet UI test with no relationship to `data_sync`. That failure is an artifact of reviewing the raw PR head: this branch is 120 commits behind `develop`, and `develop` has already amended that very test file (12 lines changed). CI does not see it because GitHub tests the merge commit rather than the head, which is also why the `test` check is green on this PR. Both failures are therefore pre-existing and outside this change.

## Findings

### 🔹 Minor

**`packages/core/src/modules/data_sync/lib/sync-schedule-service.ts:165-215` — the failure window is inverted rather than closed, so a failing flush can now leave an orphaned `ScheduledJob`.**

`SchedulerService.register()` forks its own `EntityManager` (`packages/scheduler/src/modules/scheduler/services/schedulerService.ts:40`) and commits the `ScheduledJob` row at line 115, then best-effort registers the BullMQ repeatable job at lines 117-129. All of that is durable before `em.flush()` on line 215 of this file writes the `SyncSchedule`. If that flush throws — a transient database error, a lost connection, or the losing side of a concurrent keyed create, since `SyncSchedule` carries no unique constraint on `(integrationId, entityType, direction, organizationId, tenantId)` — the scheduler keeps a live job whose `targetPayload.scheduleId` points at a row that was never written.

The consequence is bounded but real. The worker degrades safely: `packages/core/src/modules/data_sync/workers/sync-scheduled.ts:52` returns early when the schedule lookup misses, so nothing crashes and no sync runs. But the phantom job stays visible in the scheduler admin UI as a `Data sync: …` entry the data-sync page cannot explain or delete, it keeps firing no-op queue jobs on its cadence indefinitely, and it counts against the tenant's active-schedule quota through `enforceTenantActiveScheduleLimit` (`schedulerService.ts:64-66`). Because a create mints a fresh `randomUUID()` on every attempt, an operator who retries through a run of flush failures accumulates a distinct orphan per attempt rather than reusing one — which is a worse retry story than the pre-fix behaviour the issue itself noted was at least idempotent.

The concrete fix is a compensating action on the create path: wrap the row build and `em.flush()` in a `try`, and in the `catch` — when `existing` is falsy, so this run is the one that created the job — call `unregister(scheduledJobId)` inside its own `try/catch` and log a failure through `createLogger('data_sync')` rather than masking the original error, then rethrow. That mirrors the best-effort pattern the scheduler already uses for its own BullMQ sync at `schedulerService.ts:125-128`. The update path cannot be compensated the same way without a scheduler-side API to restore a previous registration, since `SchedulerServiceLike` exposes only `register` and `unregister`, so that half is worth stating explicitly rather than silently leaving asymmetric.

**Coverage of #5506 is partial, while the PR text claims the issue is closed.**

The issue's Expected-vs-actual table has four rows. Three of them — "on an unparseable interval, reject without writing anything", "stored row: absent", and "the page afterwards: no schedule" — are genuinely fixed here, and the Secondary note about the 422 leaking the freshly created row's primary key is defused, since the UUID echoed in `Failed to calculate next run time for schedule: …` no longer identifies any persisted record. The first row is untouched: "client-side validation — the documented format is enforced before submit". `createSyncScheduleSchema` still declares `scheduleValue: z.string().min(1)` (`packages/core/src/modules/data_sync/data/validators.ts:54`), so `3600` is still accepted by the API contract and still fails deep inside the scheduler, and `IntegrationScheduleTab.tsx:351` surfaces that as a raw untranslated `flash(message, 'error')` toast with no `aria-invalid` on the field and no disabled save button — exactly what the issue reports.

I am deliberately not asking for that to be fixed in this PR. Enforcing the format properly means touching `IntegrationScheduleTab.tsx`, which would flip this change from `skip-qa` to `needs-qa` and turn a tight two-file backend fix into a UI change — the kind of scope expansion that is worse for review quality, not better. The remedy is to make the split visible instead: file a follow-up for the client-side/format-validation half and narrow the claim in the PR body from an unqualified `Closes #5506` to a statement of which half this PR delivers. Worth noting that `closingIssuesReferences` on this PR is empty because the base branch is `develop` rather than the repository default, so GitHub will not auto-close #5506 on merge — but a human or a post-merge automation reading "Closes #5506" reasonably would.

### 💅 Nit

- `packages/core/src/modules/data_sync/api/schedules/__tests__/schedule-write-order.test.ts:14` — `mockScheduler.unregister` is declared on the mock but never asserted or exercised by the single test case. It is harmless, and it becomes load-bearing the moment the compensating-unregister minor above is implemented, so this is only worth a mention if that follow-up does not land.
- `packages/core/src/modules/data_sync/lib/sync-schedule-service.ts:43,47` — the shape `{ integrationId: string; entityType: string; direction: 'import' | 'export' }` is now written out verbatim in two adjacent signatures. A named local type (`type ScheduleLabelInput = …`) would read better and keep the two in step if a third label helper ever appears.
- The PR description reports the `packages/core` run as "227 tests". The package actually has 11,613; 227 is the `data_sync` subset. The narrower run was the right one to iterate on, but the description reads as if the whole package were covered.
- This branch is 120 commits behind `develop`. It still reports `MERGEABLE` with no conflicts, and a squash merge resolves it, so this is informational rather than an action — but it is why a local run of the raw head shows a `staff` test failure that CI, which builds the merge commit, does not.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `createSyncScheduleService` and the `SyncScheduleService` type are unchanged; `buildScheduleName` and `buildScheduleDescription` are module-private closures, not exports.
- [x] No function signature changed in a breaking way — `saveSchedule`, `deleteSchedule`, `getById`, `getByKey`, and `listSchedules` keep their parameters and return types. The two private helpers were *widened* to accept a structural subset, so every existing call site still type-checks.
- [x] No required type field removed or narrowed — `SchedulerServiceLike` and the `saveSchedule` input type are untouched.
- [x] No HTTP route URL removed or renamed; no method changed — `POST /api/data_sync/schedules` and `PUT|DELETE /api/data_sync/schedules/{id}` are unchanged.
- [x] No field removed or retyped in an existing response shape — `serialize.ts` is not touched; 201 on success and 422 carrying the scheduler's message on failure both hold.
- [x] No event or message name renamed or removed — the `data-sync-scheduled` queue name and its `{ scheduleId, scope }` payload are byte-identical.
- [x] No database table or column renamed or removed — `data/entities.ts` is untouched and the PR correctly ships no migration.
- [x] Where a contract had to change: not applicable, since no contract surface moved.

## 🧪 Test Coverage

The new `schedule-write-order.test.ts` drives the real `POST` handler through `jest.requireActual` on the service, with a scheduler stub that throws the authentic `Failed to calculate next run time for schedule: …` error, and asserts both halves of the contract: the response is 422 with that message, and `em.create`, `em.persist`, and `em.flush` are each never called. That is the right assertion set, because asserting only the status would have passed against the buggy code too — the whole defect was that the 422 and the write coexisted. I verified independently that it discriminates: against the `origin/develop` version of the service it fails on `expect(mockEm.create).not.toHaveBeenCalled()` with one received call, which is the issue's exact symptom rather than a generic error.

The existing `api/schedules/__tests__/optimistic-lock.test.ts` continues to cover the branch this test does not: it stubs `findOneWithDecryption` to return an existing row, so it exercises the `existing` path through both POST and DELETE, including the stale-version 409, the concurrent-delete 409, the enterprise command-guard seam, and the successful 201, all with the reordered code. Together with `mutation-guard.test.ts` and `packages/scheduler`'s 420 tests, the changed behaviour is well covered.

Two gaps are worth naming rather than leaving implicit. First, nothing asserts the *ordering* on the success path — that `register()` resolves before `em.flush()` — so a future refactor could reintroduce the bug in a form where the failing case still passes because the scheduler happens not to throw; a `jest.fn` call-order assertion (or an `invocationCallOrder` comparison) in the same file would pin it cheaply. Second, and more usefully, there is no test for the inverse failure the first minor describes: a scheduler that resolves followed by an `em.flush` that rejects. That test is the natural companion to the compensating-`unregister` fix and should be added with it, asserting that `unregister` is called with the pre-generated `scheduledJobId` on the create path and that the original flush error still propagates to the caller.

adeptofvoltron:

CI on head `b871ef2` has fully settled, and it is green. This comment closes out the follow-up that the previous pass on 2026-08-26 promised when it swapped its lock for `ci-monitoring` while `test` and `ephemeral-integration` were still queued; both have since completed successfully, so `ci-monitoring` is being removed.

**14 checks passed:** `prepare`, `lint`, `ds-lint`, `audit-scope`, `test`, `docker-build`, `documents-multi-instance`, `ephemeral-integration (none)`, `scope`, `license/cla`, `CodeQL`, `Analyze (javascript-typescript)`, `Analyze (actions)`, `Analyze (python)`. **5 were skipped** by their own workflow conditions and are not failures: `merge-coverage`, `mutate`, `comment`, `audit`, `[code]smith`. **Nothing failed and nothing is pending.** Branch protection is not readable on `develop` (the `required_status_checks` endpoint answers 404), so every reported check was treated as required — which is the conservative reading and makes no difference here, since none of them is red.

This does not change the ✅ approve verdict; it confirms it. The local `validation.commands` gate reached the same conclusion independently, with the two unrelated failures explained in the review comment above — the checkout-location-dependent `@open-mercato/cli` suites, and a `staff` timesheet test that fails only because this branch is 120 commits behind `develop` and `develop` has already amended that test file. CI does not see the latter because GitHub builds the merge commit rather than the raw head, which is exactly why its `test` check is green.

The one thing still standing between this PR and a merge is a human approving review: GitHub will not let this account approve a pull request it authored.

adeptofvoltron:

Full code review of head `b871ef2` in an isolated worktree: **no blocker and no major findings**, two minors and four nits, all recorded in the review comment above. The reordering is correct on both the create and the update path — `scheduledJobId` and `targetPayload.scheduleId` resolve to identical values on every branch — no contract surface moves, and the regression test is a real discriminator: I reverted `sync-schedule-service.ts` to its `origin/develop` version and the new test fails there with the issue's exact symptom, then passes on the PR head.

The configured validation gate ran in full and is green: `build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, and `build:app` all pass. `yarn test` exits 1 only on the known checkout-location-dependent `@open-mercato/cli` suites, and because Turbo aborts on the first failure the changed package never ran in that pass, so `@open-mercato/core` (11,610 passed, 1 unrelated `staff` failure that `develop` has already amended) and `@open-mercato/scheduler` (420/420) were re-run standalone.

**autofix: n/a — nothing to fix.** The verdict is approve, so the fix-forward loop did not trigger; no commit was pushed and the head is unchanged, which is why CI stays valid. The two actionable minors were filed rather than dropped: #5871 for the remaining orphaned-`ScheduledJob` window on a failing flush, and #5872 for the client-side interval-format validation that is the untouched half of #5506. The PR body's `Closes #5506` was narrowed to say which half this PR delivers, and #5506 carries a scope note asking that it stay open until #5872 lands.

Labels normalised to `merge-queue` + `bug` + `skip-qa` + `priority-medium` + `risk-low`; `ci-monitoring` removed now that the CI-result follow-up owed by the 2026-08-26 pass has been posted.

⚠️ This PR cannot merge on this review alone: GitHub refuses to record a formal approval from the account that authored it, so **an independent human reviewer still has to click Approve**.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-09-09T01:13:21Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron:
🤖 `om-auto-review-pr` — re-review verdict: **✅ APPROVE**

> ⚠️ GitHub refuses to record this as a formal approving review — `addPullRequestReview` answers *"Can not approve your own pull request"*, because this PR's author and this review's identity are the same account. The review is therefore posted as a comment, and **an independent human reviewer still has to click Approve** before this PR can merge. Nothing here substitutes for that.

---

# 🔍 Re-review: fix(data-sync): validate schedule before persisting it (#5506)

## 🎯 Summary

This is a re-review of the **unchanged head `b871ef2`** (no new commits since the 2026-09-03 pass). The reason to look again is base drift: `develop` has advanced **136 commits** since this branch was cut, five of them touching `packages/scheduler` — the exact package whose `register()` contract this fix now depends on for its ordering guarantee. The previous approval validated against a base that no longer exists, so the question this pass answers is whether the fix still holds against `develop` as it stands today.

It does. The change reorders `saveSchedule()` in `packages/core/src/modules/data_sync/lib/sync-schedule-service.ts` so `schedulerService.register(...)` runs *before* the `SyncSchedule` row is created and flushed, pre-generating `id`/`scheduledJobId` so the registration can be built from `input` rather than from a row that does not exist yet. An unparseable `scheduleValue` now yields a clean 422 with nothing written.

## ✅ What I verified

**The ordering guarantee still holds against today's scheduler.** `SchedulerService.register()` (`packages/scheduler/src/modules/scheduler/services/schedulerService.ts:39-115`) validates scope, validates target, and computes `nextRunAt` via `calculateNextRunForWrite` — throwing at line 56 when the value cannot be parsed — all *before* it reaches `em.create(ScheduledJob, ...)` at line 90 or its `em.flush()` at line 115. `enforceTenantActiveScheduleLimit` (line 65) likewise throws ahead of any write. So a rejected schedule touches neither table. The five `develop` commits that landed in `packages/scheduler` since this head (`a0154ac4`, `d5a442d0`, and the release merges) touch the trigger route, the execute worker, commands and `commandContext` — **not** the registration path this fix relies on. The guarantee is intact.

**Id derivation is equivalent on both branches.** `scheduledJobId = existing?.scheduledJobId ?? id` reproduces the old `row.scheduledJobId ?? row.id` exactly, including the case where a stored row has a null `scheduled_job_id` (the column is nullable — `data/entities.ts:164`). `targetPayload.scheduleId: id` resolves to `existing.id` on update and to the same UUID handed to `em.create` on create. No path changes value.

**The regression test is a real discriminator, re-proven this pass.** I reverted `sync-schedule-service.ts` to its pre-fix version inside the isolated worktree and re-ran `schedule-write-order.test.ts`: it fails with the issue's exact symptom — `em.create` called once with the fully-formed row despite the scheduler having rejected the value — then passes again on the PR head.

**Merged-tree validation is green.** The branch merges into current `develop` with no conflict (`Automatic merge went well`), and on the **merged** tree: `packages/core` `data_sync` — **227 tests, 33 suites, all passing**; `packages/scheduler` — **466 tests, 25 suites, all passing** (up from 420 at the last pass, i.e. `develop`'s new scheduler coverage passes with this fix merged in). CI on `b871ef2` is fully settled and green: 14 checks passed, 5 skipped by their own workflow conditions, nothing failing, nothing pending.

## 🟡 Minor

**1. The orphaned-`ScheduledJob` window is inverted, not closed** — `sync-schedule-service.ts:165-215`

`register()` flushes on its own forked EntityManager (`schedulerService.ts:40`, `:115`), so its write commits independently of the outer `em.flush()` at line 215. If the outer flush fails after a successful registration, a `ScheduledJob` row survives pointing at a `SyncSchedule` that was never written. The fix moves the failure window rather than eliminating it — which is the right trade (a dangling job beats a dangling, unschedulable, user-visible row), but it is worth being explicit that atomicity was not achieved.

I checked the blast radius and it is genuinely contained: the consumer returns early on a missing schedule (`workers/sync-scheduled.ts:52` — `if (!schedule || !schedule.isEnabled) return`), so an orphan silently no-ops instead of crash-looping the queue; and there is no unique constraint on `(integrationId, entityType, direction)` — only a plain `@Index` (`data/entities.ts:133`) — so a concurrent-create race cannot turn a constraint violation into this path. That leaves a genuine infrastructure fault as the only trigger. Already tracked as **#5871** (`priority-low`, `risk-low`), which is correctly sized. Not a merge blocker.

**2. The update path is untested** — `api/schedules/__tests__/schedule-write-order.test.ts:22-25`

`findOneWithDecryption` is mocked to return `null`, so the new test only exercises the **create** path. The update path is the one where a future regression would be *silent*: if the `row.* = input.*` assignments (lines 201-209) ever drift back above `register()`, the managed `existing` entity becomes dirty and any later flush in the request persists the rejected value — with no `em.create` and no `em.persist` call for the current assertions to catch. A second case returning an existing row and asserting both that `flush` is not called and that the entity's fields are untouched would close the gap. Worth adding, not worth blocking on.

## 🔵 Nits

**3.** The inline shape `{ integrationId: string; entityType: string; direction: 'import' | 'export' }` is repeated verbatim at `sync-schedule-service.ts:43` and `:47`. A named type — or `Pick<SyncSchedule, 'integrationId' | 'entityType' | 'direction'>` — would state it once.

**4.** `api/schedules/route.ts:120-122` maps every non-`CrudHttpError` to 422 carrying the raw `error.message`. Pre-existing, but this fix now routes `'Scheduler module is not available'` (`sync-schedule-service.ts:38` — an infrastructure fault, not a client error) through it as well. A 503 would be more honest for that case. Out of scope for this PR.

## 🔁 Backward compatibility

No contract surface moves. `buildScheduleName`/`buildScheduleDescription` are module-private closures, not exports, so widening their parameter type is invisible outside the file. Request/response shapes (201 on success, 422 with the scheduler's message on failure), DI wiring, the `SchedulerServiceLike` structural contract, the entity, and upsert/idempotency semantics are all unchanged. No migration, no DB structure change.

One behavior change worth stating plainly, since it is not in the PR description: when the **scheduler module is absent**, `requireScheduler()` now throws *before* the row is written, where previously the row was saved and the 422 returned afterwards. A deployment without the scheduler module can no longer create a `SyncSchedule` row at all. That is the correct reading of this fix's intent — a schedule nothing can run should not be stored — and the HTTP status is unchanged either way, so nothing observable regresses. Flagging it so it is a decision on the record rather than a side effect.

## 🏷️ Labels

Unchanged — the existing set is already correct for an approved PR: `merge-queue`, `bug`, `skip-qa`, `priority-medium`, `risk-low`.

I re-checked the `skip-qa` exemption against the current diff and it still holds: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, DB structure unchanged (no migration, entity untouched), API surface unchanged, no `BACKWARD_COMPATIBILITY.md` contract broken, and automated regression coverage ships in the same PR.

## 📌 What is actually blocking this

Nothing in the code. `mergeStateStatus` is `BLOCKED` solely because GitHub will not accept an approving review from this PR's own author account, and no independent reviewer has approved in the two weeks this has been sitting in `merge-queue`. **A human needs to click Approve.** Follow-ups #5871 (orphan window) and #5872 (client-side interval validation) are open and correctly scoped; per the PR description, **#5506 stays open until #5872 lands**.

adeptofvoltron:

Re-review of the unchanged head `b871ef2` against a base that has moved 136 commits since the last pass. **No blocker and no major findings** — two minors and two nits, all recorded in the review comment above.

Re-verified this pass: `SchedulerService.register()` still throws ahead of every write on the current `develop` (the five scheduler commits that landed since this head touch the trigger route, worker and commands — not the registration path); the branch merges into `develop` cleanly; on the merged tree `data_sync` runs 227/227 and `packages/scheduler` 466/466 green; and the regression test is still a real discriminator — reverting the service to its pre-fix version makes it fail with the issue exact symptom, then pass again on the head. CI on this head is fully settled and green, so no CI follow-up is owed and `ci-monitoring` is not being applied.

Labels unchanged and still correct: `merge-queue`, `bug`, `skip-qa`, `priority-medium`, `risk-low`. Autofix: n/a — nothing to fix.

The only thing blocking merge is an independent human approval; GitHub refuses a self-approval from this account.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-09-09T14:33:37Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron (commented):
🤖 `om-auto-review-pr` — self-review verdict: **✅ APPROVE**

> ⚠️ GitHub refuses to record this as a formal approving review — `addPullRequestReview` answers *"Can not approve your own pull request"*, because this PR's author and this review's identity are the same account. The review is therefore posted as a comment, and **an independent human reviewer still has to click Approve** before this PR can merge. Nothing here substitutes for that.

> 📋 **On the premise of this pass.** This run was requested on the basis that the PR "has zero reviews". That is true of the GitHub reviews API (`reviews: []`) but not of the PR: this is the **fourth** review pass on head `b871ef2` (2026-08-26, 2026-09-03, 2026-09-09 01:18, and now), each blocked from being recorded formally for the same reason. Reviewer-side, this PR is among the most-examined on the board. See *What is actually blocking this* at the end — a fifth pass will not change the outcome either.

---

# 🔍 Re-review: fix(data-sync): validate schedule before persisting it (#5506)

## 🎯 Summary

Head `b871ef2` is unchanged since the last pass, and `develop` has not moved either — still exactly **136 commits** ahead, the same figure the 2026-09-09 pass measured. So the code and its base are byte-identical to what was last approved, and re-litigating the ordering argument would add nothing.

This pass therefore targets the three areas the request named, none of which a previous pass covered: whether the fix should have gone through a **zod validator** in `data/validators.ts`; whether the **error surface** it now routes through is useful-but-minimal; and whether it **conflicts with #5916**, which rewrites the same failure path. It also re-ran the regression discriminator.

The verdict is unchanged — **approve, no blocker, no major** — but this pass adds one residual gap that no prior review, and no open issue, currently accounts for.

## ✅ What I verified this pass

**The regression test is still a real discriminator.** I replaced `sync-schedule-service.ts` with its `origin/develop` version in the isolated worktree and re-ran `schedule-write-order.test.ts`: it fails with the issue's exact symptom — `em.create` *"Expected number of calls: 0, Received number of calls: 1"* — then passes again on the PR head. The working tree was restored immediately; nothing was left modified.

**`packages/core` `data_sync` is green on the PR head**: **33 suites, 227 tests, all passing** (the `ERROR`/`WARN` lines in that run are deliberate fixture logs from the sync-engine and batch-stream suites, not failures).

**The full validation gate was *not* re-run locally this pass, deliberately.** CI on this exact head is fully settled and green (14 checks passed, 5 skipped by their own workflow conditions, none failing, none pending), the head has not changed since the 2026-09-03 and 2026-09-09 passes that did run the whole gate end-to-end, and the base has not moved since. Re-running `build:packages` twice plus `build:app` would have re-proved a green that already exists on the same SHA against the same base. Recording this so the evidence behind the approval is not overstated.

**Only the two HTTP routes call `saveSchedule`** (`api/schedules/route.ts:99`, `api/schedules/[id]/route.ts:115`) — no seed, worker, or setup path depends on a `SyncSchedule` row being written when the scheduler is unavailable, so the reordering has no non-HTTP caller to surprise.

**Guard ordering is intact.** `register()` now runs earlier, but it still runs *after* auth, org-scope resolution, zod parsing, `validateCrudMutationGuard`, and `enforceCommandOptimisticLockWithGuards`. Nothing about the reorder lets a caller reach the scheduler ahead of a guard.

## 🟢 Something the PR description undersells

The fix is **broader than the unparseable-value case it advertises**. `SchedulerService.register()` also calls `enforceTenantActiveScheduleLimit` (`packages/scheduler/src/modules/scheduler/services/schedulerService.ts:65`) before it touches either table. So a tenant that is **over its active-schedule quota** previously also got a committed, unschedulable `SyncSchedule` row alongside its 422 — the same #5506 defect, via a second trigger — and this reorder fixes that path too, with no extra code.

Worth adding to the PR description and to #5506's eventual closure note: the guarantee is "a rejected registration writes nothing", not merely "an unparseable interval writes nothing".

## 🟡 Minor

**1. #5916 does not close the window this PR opens on the *update* path — and nothing currently tracks that** — `sync-schedule-service.ts:165-215`

This is the answer to the "check the two do not conflict" ask, and the one genuinely new finding.

The two PRs **do not conflict**. #5916 is stacked directly on this branch, is a pure superset of this diff, reverses none of its decisions, and its added tests keep this PR's assertions intact. On the mechanical question the request asked, they are clean.

But they do not add up to the coverage the follow-up chain implies. `register()` is an **upsert** (`schedulerService.ts:66-84`): on the update path it overwrites the existing `ScheduledJob`'s `scheduleValue`, `timezone`, `isEnabled` and `nextRunAt`, and `em.flush()`es that on its own forked EntityManager (`:40`, `:115`) — durably — before this service's outer `em.flush()` at line 215 writes the `SyncSchedule` row. If that outer flush then fails, the scheduler is left firing on the **new** cadence while the `SyncSchedule` row the data-sync page renders still holds the **old** one.

#5916 deliberately does not compensate this. Its own code comment says so: *"when a job was already there, `register()` overwrote one we inherited, and `SchedulerServiceLike` offers no way to restore the previous definition, so unregistering would destroy that job rather than roll anything back."* That is the correct call — unregistering an inherited job would be strictly worse than the divergence — but it means the case is consciously left open, and its test `leaves an inherited registration alone when the update write fails` pins that as intended behavior.

Meanwhile #5871 is scoped to the **create-path orphan** ("a failing schedule flush can leave an orphaned `ScheduledJob` behind"), and #6000 covers concurrent duplicate creates. Neither describes the update-path value divergence, and a search of open issues turns up nothing else that does. So once #5916 merges, the follow-up chain will read as closed while this residual is still open and undocumented.

Severity is genuinely minor and this is **not a merge blocker**: the trigger is an infrastructure fault between two writes in one request, the damage is a wrong cadence rather than data loss, the worker degrades safely, and the next successful save re-converges both rows. But it should be on the record rather than discovered later. The cheapest fix is a scope note on **#5871** (or a line in #5916's description) stating that the update path is knowingly uncompensated and why. I have not filed anything — say the word and I will.

**2. The 422 body is honest but not actionable — and that is what makes #5872 harder** — `api/schedules/route.ts:120-122`, `[id]/route.ts:145-147`

Answering the "useful but minimal error messages" ask directly.

The **minimal** half is now genuinely better than the PR description claims. The old 422 leaked the primary key of a row that had just been committed; the new one carries `Failed to calculate next run time for schedule: <uuid>` where that UUID is a pre-generated id for a row that was **never persisted**. The identifier is now inert — it names nothing that exists. That is a real improvement and the description undersells it.

The **useful** half is not there. Compare the two failure shapes this endpoint can return:

- zod rejection → `{ error: 'Invalid payload', details: parsed.error.flatten() }` — field-mapped, bindable to `aria-invalid`
- scheduler rejection → `{ error: 'Failed to calculate next run time for schedule: 8f3c…' }` — no field, no hint that `1h` is the expected shape and `3600` is not, plus an internal id the client has no use for

Since #5872's whole job is to mark the interval field invalid before submit, it will have to hard-code the format rules client-side rather than consume anything this response says. Mapping the scheduler's rejection onto the same `details`-shaped body — keyed on `scheduleValue` — would let #5872 bind to the server's answer instead of duplicating it. Pre-existing route behavior and out of scope here, but this PR is what newly routes the ordering decision through it, so it is the natural place to note it.

Related, same lines: `throw new Error('Scheduler module is not available')` (`sync-schedule-service.ts:38`) reaches the client verbatim through that same catch-all. Per `AGENTS.md` a purely internal throw should carry the `[internal]` prefix so the i18n hardcoded-string checker treats it as opted out, and 503 would be the honest status for an infrastructure fault. Note that with `di.ts:30` destructuring `schedulerService` off an Awilix cradle proxy, an unregistered scheduler would fail container resolution before this throw is ever reached — so the message is close to unreachable in practice, which is its own small argument for tidying it.

## ✅ Resolved since the last pass

**The update path is no longer untested — downstream.** The 2026-09-09 pass raised that `schedule-write-order.test.ts` mocks `findOneWithDecryption` to `null` and therefore only exercises the create path. #5916 adds exactly the missing coverage: `leaves an inherited registration alone when the update write fails` and `unregisters the registration it minted for a row that had no scheduled job yet`, both driving `existingRow(...)` through the service.

Recording this as addressed rather than open — with one caveat worth stating plainly: that coverage lives in **#5916**, which is stacked on this branch and currently `blocked`. If #5916 stalls or is dropped, #5661 ships with create-path coverage only and this minor reverts to open.

## 🔵 Nits

**3.** The inline shape `{ integrationId: string; entityType: string; direction: 'import' | 'export' }` is still repeated verbatim at `sync-schedule-service.ts:43` and `:47`. A named type — or `Pick<SyncSchedule, 'integrationId' | 'entityType' | 'direction'>` — would state it once. Carried from the previous pass, unchanged.

**4.** The comment at `sync-schedule-service.ts:170-172` says *"Validate the schedule (and register it with the scheduler) before writing the SyncSchedule row"*. Given finding 1, "validate and durably register" would be more precise — the parenthetical reads as an aside when the durability of that registration is the whole reason a compensation PR exists.

## 🧩 On the zod-validator question

The request asked about "zod validators in `data/validators.ts` with types derived via `z.infer`". Answering explicitly, because the PR title *"validate schedule before persisting it"* invites the assumption that a validator was added, and none was.

`createSyncScheduleSchema` (`data/validators.ts:49-58`) declares `scheduleValue: z.string().min(1)` — presence only, no format check — and both schemas already derive their types through `z.infer` per the repo convention. This PR adds nothing there. The real validation is `calculateNextRunForWrite` inside `packages/scheduler`, and what the PR fixes is *when* it runs relative to the write, not *what* validates.

**That is the right call and it should stay this way.** A zod `.refine()` reproducing the interval/cron grammar in `data_sync` would create a second source of truth against `packages/scheduler`'s parser, and the two would drift — a value zod accepts but the parser rejects reintroduces exactly this bug one layer up, and a value zod rejects but the parser accepts is a silent regression nobody would think to test. Delegating to the single parser and fixing the ordering is the correct architecture. The PR title is accurate about ordering; it just reads like a schema change.

One thing this leaves unsettled, and it lands on **#5872**: the client-side half needs the same grammar. If it hard-codes the rules in the form, that is the third copy. A small exported helper from `packages/scheduler` — `isParseableScheduleValue(type, value)` or similar, wrapping the parser both this path and the UI already depend on — would keep it at one. Worth deciding before #5872 is implemented rather than during review of it.

## 🔁 Backward compatibility

Unchanged from the previous assessment and re-confirmed: no contract surface moves. `buildScheduleName`/`buildScheduleDescription` are module-private closures, not exports, so widening their parameter type is invisible outside the file. Request/response shapes (201 on success, 422 with the scheduler's message on failure), DI wiring, the `SchedulerServiceLike` structural contract, the entity, and upsert/idempotency semantics are all unchanged. No migration, no DB structure change.

The one behavior change on the record stays on the record: when the scheduler module is absent, `requireScheduler()` now throws *before* the row is written, so such a deployment can no longer create a `SyncSchedule` row at all. That is the correct reading of this fix's intent, the HTTP status is unchanged either way, and per the DI note in finding 2 the path is close to unreachable in practice.

## 🏷️ Labels

No change needed — `merge-queue`, `bug`, `skip-qa`, `priority-medium`, `risk-low` is already the correct set for an approved PR, exactly one priority and one risk. `in-progress` was applied for this run's claim and is removed with the lock.

I re-checked the `skip-qa` exemption against the current diff a third time and it still holds: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, DB structure unchanged (no migration, entity untouched), API surface unchanged, no `BACKWARD_COMPATIBILITY.md` contract broken, and automated regression coverage ships in the same PR.

## 📌 What is actually blocking this

Still nothing in the code, and that has now been true across four passes and two weeks.

`mergeStateStatus` is `BLOCKED` for one reason: GitHub will not accept an approving review from this PR's own author account, and no independent reviewer has approved. **A human needs to click Approve.** No further self-review pass can change that — this one added a residual gap worth recording (finding 1) and settled two design questions (zod, error shape), but the merge gate is organizational, not technical, and a fifth pass would be pure duplication.

Two things for whoever picks this up:

- **Merge order matters.** #5916 is stacked on this branch and must not be merged first; and per its own description, once #5661 lands and GitHub retargets it, a commit must be pushed so `synchronize` fires — retargeting alone raises `edited`, which `ci.yml` does not listen for, so #5916 would otherwise merge having never run CI.
- **#5506 stays open** until #5872 (client-side interval validation) lands; this PR fixes the server-side half only.

adeptofvoltron:

Fourth self-review pass on the unchanged head `b871ef2`, run because the PR reads as having zero reviews via the GitHub API. **No blocker and no major findings** — two minors and two nits, in the review above.

This pass deliberately did not re-litigate the ordering argument (head unchanged, and `develop` has not moved either — still exactly 136 commits ahead, the same figure the last pass measured). It targeted the three areas that had not been examined before:

- **Interaction with #5916** — the two do **not** conflict; #5916 is stacked on this branch, is a pure superset, and reverses none of this diff's decisions. But it does not close the window this PR opens on the **update** path: `register()` is an upsert that durably overwrites the inherited `ScheduledJob` before the outer flush, and #5916 consciously declines to compensate that (its own comment explains why). #5871 covers only the create-path orphan and #6000 only duplicate creates, so that residual is currently untracked. Recorded as minor 1; no issue filed unilaterally.
- **The zod-validator question** — no validator was added and none should be. Duplicating the scheduler's interval/cron grammar in `data/validators.ts` would create a second source of truth against `packages/scheduler`'s parser. Flagged that #5872 will otherwise become the third copy, and that a small shared helper is the answer.
- **The error surface** — the "minimal" half improved more than the description claims (the leaked id now names a row that is never persisted, so it is inert), but the "useful" half is missing: the scheduler's rejection returns an unmapped message where zod failures return a field-mapped `details` body, which is exactly what makes #5872's client-side job harder.

Also recorded: the fix is **broader than advertised** — `enforceTenantActiveScheduleLimit` throws inside `register()` too, so an over-quota tenant no longer gets a stranded row either. Worth adding to the PR description and #5506's closure note.

Verified this pass: the regression test re-proven as a real discriminator (reverted the service to `origin/develop`, the test fails with `em.create` "Expected 0, Received 1", then passes on the head; working tree restored); `packages/core` `data_sync` **33 suites / 227 tests, all passing**. The full validation gate was **not** re-run locally — CI is green on this exact SHA against an unmoved base, and the two prior passes ran it end-to-end. Stating that so the evidence is not overstated.

The previous pass's open minor "the update path is untested" is now **addressed downstream** by #5916's added coverage — with the caveat that if #5916 stalls, it reverts to open here.

Labels unchanged (`merge-queue`, `bug`, `skip-qa`, `priority-medium`, `risk-low`); `in-progress` removed with this lock. No `ci-monitoring` — CI is already fully settled and green, so there is no follow-up owed.

Unlike the three previous passes, this review was submitted as a **review object** (`COMMENTED`) rather than a plain comment, so `reviews[]` is no longer empty and this PR should stop reading as "never reviewed" to tooling. It still is **not** an approval: GitHub will not accept one from this PR's author account. **A human maintainer needs to click Approve.**

patzick (approved):

Issue #4491, opened 2026-07-24
Log command-interceptor registry import failures


pkarw:
## Follow-up from #4414

PR #4414 fixed CLI and queue-worker bootstrap so generated command-interceptor entries are loaded consistently with the web runtime. The compatibility fallback still converts any registry compile/import failure to an empty interceptor list without emitting diagnostics, which could make the same missing-interceptor behavior recur invisibly.

> **Self-review note (low).** `dynamicLoader.ts` `.catch(() => ({ commandInterceptorEntries: [] }))` fails **open and silently** — if `command-interceptors.generated.ts` ever fails to compile or import, this recreates the exact #4327 condition (all interceptors skipped, commands run unguarded) with no log line to notice it. Not attacker-reachable, but given the bug this PR fixes is precisely "interceptors silently did not load", a `logger.error` on the swallowed failure would stop it recurring invisibly. Otherwise reviewed clean — both added tests fail without the fix, and `factory.ts` already consumed the field, so only the loader was missing.

Source: https://github.com/open-mercato/open-mercato/pull/4414#issuecomment-5066708189

### Acceptance criteria

- [ ] Log a clear error when `command-interceptors.generated.ts` exists but fails to compile or import.
- [ ] Preserve the compatibility fallback to an empty interceptor list when the generated file is absent.
- [ ] Add regression coverage that distinguishes an expected missing file from an unexpected compile/import failure.

Related: #4414, #4327

adeptofvoltron:
🤖 `om-auto-fix-issue` triage: **NO_ACTION_NEEDED** — this is already fixed on `develop`.

## 🔍 Evidence

PR #4505 (*"fix(shared): log command-interceptor registry import failures"*) merged into `develop` on 2026-07-30 as commit `754191054`. It referenced this issue with `Refs #4491` rather than `Closes #4491`, which is why the issue stayed open after the fix shipped.

All three acceptance criteria are met in `packages/shared/src/lib/bootstrap/dynamicLoader.ts` at the current `develop` head:

- ✅ **Log a clear error when the generated file exists but fails to compile or import.** `loadOptionalGeneratedModule` reports anything that is not a missing source at `logger.error` with the message *"Failed to load generated registry, continuing without its entries"*, carrying `file`, `filePath` and the underlying error under `err`.
[...]

Issue #4327, opened 2026-07-21
bug: queue/CLI worker bootstrap never registers command interceptors (loadBootstrapData omits commandInterceptorEntries)


a-szczudlo:
## Summary

Command interceptors registered by application modules never run inside the queue-worker (and any other CLI-bootstrapped) process. The worker's bootstrap builds its `BootstrapData` without a `commandInterceptorEntries` field, so the guard in `createBootstrap` that would call `registerCommandInterceptors` is never taken. Every command interceptor therefore silently no-ops for queued/CLI-executed commands, while the Next.js runtime (which does pass the entries) applies them correctly.

**Context:** we build a legal-industry CRM on Open Mercato. We hit this because a command interceptor that adjusts a deal's `probability` applied on the web request path but silently did nothing for a queued `bulk-update-stage` job.

## Environment

- `@open-mercato/shared` / `@open-mercato/cli` `0.6.6`
- Area: `shared/bootstrap`, CLI queue worker
- Branch: `develop` (verified against installed `0.6.6`)

## Evidence (file:line)

- `packages/shared/src/lib/bootstrap/dynamicLoader.ts:163` loads `command-loaders.generated.ts` and the returned `BootstrapData` (object around lines 167–181) exposes `commandLoaderEntries` (`:173`) but has **no** `commandInterceptorEntries` key — only `interceptorEntries: []` (`:178`, which are the API interceptors). `command-interceptors.generated.ts` is never imported here.
- `packages/shared/src/lib/bootstrap/types.ts:64` — `commandInterceptorEntries?` is optional, so the omission type-checks.
- `packages/shared/src/lib/bootstrap/factory.ts:98-99` — `if (data.commandInterceptorEntries) registerCommandInterceptors(...)`. With the field absent, this branch is never taken.
- The worker/CLI use exactly this path: `packages/cli/src/bin.ts:59-65` and `packages/cli/src/mercato.ts:711-717` (also `:1030-1032`, `:1387-1392`) call `bootstrapFromAppRoot` → `loadBootstrapData`. The interceptor registry lives in `globalThis` per-process (`command-interceptor-store`), so registration performed in the web-server process does not carry into the separate worker process.

## Expected result

Command interceptors declared by modules apply to commands regardless of whether the command runs on the web request path or inside a CLI/queue-worker process.

## Actual result

[...]

PR #5665, opened 2026-08-26
fix(shared): dispatch app-level entry.overrides in the CLI/worker bootstrap path

### LINKED ISSUE #5582: bug(cli): modules.ts entry.overrides never reach CLI commands — seed-encryption seeds base maps, ignoring overrides.encryption.maps
## Summary

`entry.overrides` declared in the app's `src/modules.ts` are applied by the Next.js runtime (the template's `bootstrap.ts` calls `applyModuleOverridesFromEnabledModules(enabledModules)`), but the CLI bootstraps through a different path that never sees them. Any override domain consumed from CLI context silently falls back to the base contracts. The concrete instance that bit us: `overrides.encryption.maps` — `yarn mercato entities seed-encryption` seeds the *base* map for an entity whose map the app extended, reports `✅ Encryption maps seeded`, and the override-added fields keep being written as plaintext at rest while the operator believes seeding covered them.

Related to the phased rollout tracked in #1787, but distinct from wiring individual domains: this is the CLI bootstrap path not invoking the dispatcher at all, so even fully wired domains (encryption is wired in 0.6.7 — `registerBuiltInModuleOverrideAppliers` registers `encryptionOverridesApplier`) are invisible to CLI commands.

## Repro

Standalone app on `@open-mercato/*@0.6.7`:

1. In `src/modules.ts`, add to a module entry: `overrides: { encryption: { maps: { '<module>:<entity>': <extended map with an extra field> } } }`.
2. `yarn generate`
3. `yarn mercato entities seed-encryption --tenant <tenantId>`
4. Inspect the seeded `encryption_maps` row for that entity: it contains the base fields only; the override-added field is missing. The command exits with `✅ Encryption maps seeded`.

The same override is honored by the Next.js runtime in the same checkout.

## Root cause (file:line)

Line numbers from the `src/` in the 0.6.7 npm tarballs:

- CLI entrypoints bootstrap via `bootstrapFromAppRoot(appDir)` (`packages/cli/src/mercato.ts:1078-1080`, `:1435-1440`; `packages/cli/src/bin.ts:59-65`).
- `bootstrapFromAppRoot` (`packages/shared/src/lib/bootstrap/dynamicLoader.ts` ~370-400) compiles and imports only `.mercato/generated/*` — notably `modules.cli.generated.ts` — and never imports the app's `src/modules.ts` nor calls `applyModuleOverridesFromEnabledModules`.
- The generated `modules.cli.generated.ts` does not carry the `overrides` field (verified on 0.6.7: zero occurrences of `overrides` in the generated file), so the information is lost before the registry loads.
- `seed-encryption` (`packages/core/src/modules/entities/cli.ts:295`, maps read in `upsertEncryptionMaps` at `:266` via `getDefaultEncryptionMaps` from the shared registry) therefore only ever sees base maps.

## Suggested fix

Either:

1. have the generator emit each entry's `overrides` into `modules.cli.generated.ts` and have `bootstrapFromAppRoot` run the override dispatcher over them before registries first-load (mirrors what `bootstrap.ts` does for the web runtime), or
2. have `bootstrapFromAppRoot` load the app's `modules.ts` (or a generated overrides artifact) explicitly for the same purpose.

Until then, a warning would help a lot: if the runtime module entries carry `overrides` but the CLI bundle has none, CLI commands touching an overridable domain could log that overrides are not in effect — the current failure is perfectly silent and, for encryption maps specifically, security-relevant.

adeptofvoltron:
Fixes #5582

## 🎯 Goal
- Make an app's `entry.overrides` (declared in `src/modules.ts`) actually take effect for every CLI, worker, and scheduler process — not just the Next.js runtime — so commands like `seed-encryption` stop silently seeding the base contract when the app has overridden it.

## 🔍 Problem
`entry.overrides` declared in an app's `src/modules.ts` are applied by the Next.js runtime (`bootstrap-common.ts` calls `applyModuleOverridesFromEnabledModules(enabledModules)` before any registry loads), but CLI, worker, and scheduler processes bootstrap through `bootstrapFromAppRoot()` instead, which never dispatched those overrides at all. The concrete, security-relevant instance: `yarn mercato entities seed-encryption` reads `getDefaultEncryptionMaps(getCliModules())` and only ever saw the *base* `defaultEncryptionMaps`, silently ignoring `overrides.encryption.maps` — the command reports `✅ Encryption maps seeded` while the override-added fields keep being written as plaintext at rest. The same gap applied to every other override domain (ACL features, CLI commands, workers, event subscribers, setup) for any CLI/worker/scheduler process.

## 🔍 Root Cause
`bootstrapFromAppRoot()` in `packages/shared/src/lib/bootstrap/dynamicLoader.ts` compiles and imports only the generated `.mercato/generated/*` artifacts (notably `modules.cli.generated.ts`, which does not carry the `overrides` field) — it never imports the app's own `src/modules.ts`, so `applyModuleOverridesFromEnabledModules` was never invoked in that path. Both `registerModules` (inside `bootstrap()`) and `registerCliModules` (called by the `mercato` bin right after `bootstrapFromAppRoot` returns) then read the override side-registry before anything ever populated it.

## What Changed
- `packages/shared/src/lib/bootstrap/dynamicLoader.ts` — added `loadAppModuleOverrides(appDir)`, mirroring the existing `loadAppDiRegistrar` pattern: it compiles and imports the app's `src/modules.ts` via the shared esbuild pipeline, extracts `enabledModules`, and calls `applyModuleOverridesFromEnabledModules`. It runs inside `bootstrapFromAppRoot`'s existing esbuild lifecycle scope, before `loadBootstrapData`/`createBootstrap`/`bootstrap()` — i.e. before both `registerModules` and `registerCliModules` first read the override side-registry. It is fail-soft (log-and-skip) on a missing `src/modules.ts`, a non-array `enabledModules` export, or a compile/import failure, so a broken or unusual app layout degrades the bootstrap instead of crashing it.
- `packages/shared/src/lib/bootstrap/__tests__/dynamicLoader.appModuleOverrides.test.ts` — new regression coverage (3 cases, mirroring the existing `dynamicLoader.appDi.test.ts` scaffolding).

## 🧪 Tests
- `yarn build:packages` — 27/27 tasks passed (run twice, once before and once after `yarn generate`, per the standard gate order).
- `yarn generate` — passed (one pre-existing, unrelated OpenAPI-bundling ESM warning that falls back to static extraction; not caused by this change).
- `yarn i18n:check-sync` — passed, all locales in sync.
- `yarn i18n:check-usage` — passed, exit 0.
- `yarn typecheck` — full monorepo (27 packages) passed.
- `yarn test` — full monorepo; the only failure is `packages/core/src/modules/warranty_claims/__tests__/quantity.test.ts` (2 tests), a pre-existing locale-dependent `toLocaleString` formatting issue on this machine's Polish locale — confirmed unrelated by stashing this change out and observing the identical 2 failures.
- `yarn build:app` — passed.
- New test file `dynamicLoader.appModuleOverrides.test.ts` covers: (1) `overrides.encryption.maps` is reflected in `getDefaultEncryptionMaps(getCliModules())` after `bootstrapFromAppRoot` + `registerCliModules` — fails without this fix; (2) the base map is preserved when no override is declared (control); (3) a broken `src/modules.ts` logs one error and bootstrap still completes with the base map (fail-soft coverage).

## 💥 Breaking Changes
None. `entry.overrides` in `src/modules.ts` is an existing, documented contract; this only makes it actually apply in CLI/worker/scheduler processes where it was previously a silent no-op. No database schema, API surface, or `BACKWARD_COMPATIBILITY.md` contract surface changes.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5665"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790335094&installation_model_id=432243&pr_number=5665&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5665&signature=2f14b5ddf171c7fe6bf86c54e27f2db6065c16794d2a6747f30c62d66f10f375"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:

- 🔍 `review` — this is a ready, non-draft PR with no other pipeline state in play, so it starts the standard review pipeline.
- 🐛 `bug` — this fixes a defect: `entry.overrides` declared in an app's `src/modules.ts` never reached CLI/worker/scheduler bootstrap, unlike the Next.js runtime.
- ⏭️ `skip-qa` — the diff touches no `.tsx`/UI-rendering files, changes no database structure or API surface, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships a dedicated regression test (`dynamicLoader.appModuleOverrides.test.ts`, 3 cases) covering the changed behavior in this same PR, so per the automated-verification exemption it takes `skip-qa` instead of `needs-qa` — there is no manually-clickable surface for a QA reviewer to exercise beyond what the test suite already proves.
- 🔺 `priority-high` — the concrete symptom is security-relevant: `seed-encryption` was silently seeding the *base* encryption map instead of an app's `overrides.encryption.maps`, so override-added fields kept being written unencrypted at rest while the operator believed seeding covered them; this matches the "security hardening" bucket in the priority-inference rule.
- ⚠️ `risk-high` — the change lives in `bootstrapFromAppRoot()`, the shared bootstrap entrypoint used by every CLI command, worker, and scheduler process, and it changes the load order those processes go through before any registry first-loads; a regression here has a broad blast radius across the whole CLI/worker/scheduler surface, which matches the "shared contract surfaces / broad cross-module edits" bucket in the risk-inference rule even though the change itself is additive and fail-soft.

adeptofvoltron:
## 🤖 `om-fix` — run summary

**Status:** ready

**Files changed:**
- `packages/shared/src/lib/bootstrap/dynamicLoader.ts`
- `packages/shared/src/lib/bootstrap/__tests__/dynamicLoader.appModuleOverrides.test.ts`

**Summary:** `bootstrapFromAppRoot()` (the shared bootstrap entrypoint used by every CLI command, worker, and scheduler process) never dispatched an app's `entry.overrides` from `src/modules.ts`, unlike the Next.js runtime's `bootstrap-common.ts`, which calls `applyModuleOverridesFromEnabledModules(enabledModules)` before any registry loads. Concretely, `seed-encryption` (and any other CLI command reading `getDefaultEncryptionMaps`/`getCliModules`) saw only the base `defaultEncryptionMaps`, silently ignoring `overrides.encryption.maps` — and the same gap applied to every other override domain (ACL features, CLI commands, workers, event subscribers, setup) for CLI/worker/scheduler processes. Added `loadAppModuleOverrides(appDir)` in `dynamicLoader.ts`, mirroring the existing `loadAppDiRegistrar` pattern: it compiles and imports the app's `src/modules.ts` via the shared esbuild pipeline, extracts `enabledModules`, and calls `applyModuleOverridesFromEnabledModules`. It is fail-soft (log-and-skip) on a missing file, a non-array export, or a compile/import failure, so a broken app layout degrades instead of crashing CLI/worker bootstrap. It runs inside `bootstrapFromAppRoot`'s existing esbuild lifecycle scope, before `loadBootstrapData`/`createBootstrap`/`bootstrap()` — i.e. before both `registerModules` (inside `bootstrap()`) and `registerCliModules` (called by the `mercato` bin right after `bootstrapFromAppRoot` returns) first read the override side-registry.

**Tests:** Added `dynamicLoader.appModuleOverrides.test.ts` (3 cases, mirroring the existing `dynamicLoader.appDi.test.ts` scaffolding): (1) an app's `overrides.encryption.maps` is reflected in `getDefaultEncryptionMaps(getCliModules())` after `bootstrapFromAppRoot` + `registerCliModules` — fails without the fix; (2) the base map is preserved when no override is declared (control); (3) a broken `src/modules.ts` logs one error and still completes bootstrap with the base map (fail-soft coverage).

**Breaking changes:** none — this only makes an existing, documented contract (`entry.overrides` in `src/modules.ts`) actually apply in CLI/worker/scheduler processes where it was silently a no-op before.

**Validation gate (re-run from scratch in this isolated worktree, in order):**
1. `yarn build:packages` — 27/27 tasks passed
2. `yarn generate` — passed (one pre-existing unrelated OpenAPI-bundling ESM warning, falls back to static extraction, not caused by this change)
3. `yarn build:packages` (again) — 27/27 passed
4. `yarn i18n:check-sync` — passed, all locales in sync
5. `yarn i18n:check-usage` — passed, exit 0
6. `yarn typecheck` (full monorepo, 27 packages) — all passed
7. `yarn test` (full monorepo) — only failure was `packages/core/src/modules/warranty_claims/__tests__/quantity.test.ts` (2 tests), a pre-existing locale-dependent `toLocaleString` formatting issue caused by this machine's Polish locale (comma vs period decimal separator) — confirmed via `git stash`/`git stash pop` that the exact same 2 tests fail identically with this change stashed out, so it is unrelated and pre-existing, not a regression
8. `yarn build:app` — passed

The module-facts registry-unavailable log lines during build, and the one OpenAPI-bundling fallback warning during generate, are pre-existing environmental noise, not failures.

adeptofvoltron:

adeptofvoltron:
🤖 Adopting a stale chain lock (~7 days) — previous run presumed dead.

🤖 `om-auto-review-pr` taking over the chain lock — code review of the PR head. Started: 2026-09-02T12:24:19Z.

adeptofvoltron (commented):
# 🔍 Code Review

**Verdict: CHANGES REQUESTED** — 2 majors, no waiver. No blockers: nothing here regresses behavior that works today, and the core mechanism is right. What stops it is that the fix is narrower than what it claims to be, and it keeps a silent-degradation path open on the exact security surface that produced #5582.

*(Reviewed as `om-auto-review-pr`. This PR is authored by the same account the automation runs as, so GitHub will not let it carry an `APPROVE`/`REQUEST_CHANGES` state — the verdict above is the review's outcome, and the pipeline label reflects it. No autofix was run: the author is applying fixes by hand.)*

## ✅ What is right

The diagnosis is correct and the placement is careful. `loadAppModuleOverrides` runs inside `bootstrapFromAppRoot`'s existing `withEsbuildLifecycle` scope, ahead of `loadBootstrapData`/`createBootstrap`/`bootstrap()`, so it lands before both `registerModules(data.modules)` and the `registerCliModules(data.modules)` that `packages/cli/src/bin.ts:55` runs right after the call returns — that is the correct ordering constraint, and the comment at `packages/shared/src/lib/bootstrap/dynamicLoader.ts:736-739` states it accurately. Every built-in applier assigns into a keyed store rather than appending, so the double-dispatch case (`bin.ts:54` bootstraps, then `packages/cli/src/mercato.ts:1110` bootstraps again in the same process) is idempotent. Mirroring `loadAppDiRegistrar` — same `compileAndImport` options shape, same `app-*.compiled.mjs` output convention, same test scaffolding as the `#4327`/`#4491` guards — keeps the file coherent. And the singleton question this fix depends on does hold: `@open-mercato/shared` is `"type": "module"` with a single dist build, so the compiled `.mjs`'s `import '@open-mercato/shared/modules/overrides'` and the dist-internal relative import resolve to one module instance.

I ran the three affected suites locally (`dynamicLoader.appModuleOverrides`, `dynamicLoader.appDi`, `modules/overrides`) — 13/13 pass. The full gate was not re-run locally because every required check on this head is green (`test`, `lint`, `ds-lint`, `docker-build`, `CodeQL`, `ephemeral-integration` 1–15/15); the findings below come from reading the change against its call sites, and each one is reproduced or traced to a concrete line rather than inferred.

## 🚨 Findings

### 1. 🟠 Major — `overrides.ai` is still silently dropped in the very path this PR fixes

`packages/shared/src/lib/bootstrap/dynamicLoader.ts:613-641`

`registerBuiltInModuleOverrideAppliers()` (`packages/shared/src/modules/overrides.ts:1614-1631`) registers 15 domains. It does **not** register `ai` — that applier is installed only as a side effect of importing `@open-mercato/ai-assistant/modules/ai_assistant/lib/ai-overrides`, which is exactly why `apps/mercato/src/bootstrap-common.ts:6` imports that module for its side effect on the line immediately before it dispatches. The new CLI path has no equivalent, and nothing upstream of it supplies one: `packages/cli/src/bin.ts` goes `loadAppEnv()` → telemetry → `bootstrapFromAppRoot`, and `apps/mercato/src/modules.ts` imports only `parseBooleanWithDefault`, a type-only `ModuleOverrides`, and `./official-modules.generated`.

So `loadAppModuleOverrides` dispatches at a moment when `appliers.get('ai')` is `undefined`, and `applyModuleOverridesFromEnabledModules` (`overrides.ts:286-296`) takes the unwired branch: the entry is discarded and a one-shot warning fires. I reproduced it against this head — dispatching `{ id: 'example', overrides: { ai: { agents: { 'catalog.catalog_assistant': null } } } }` with only the built-ins registered leaves `snapshotProgrammaticOverrides().modulesConfigAgents` as `{}` and logs:

```
WARN [shared:module-overrides] Override domain not yet wired — entry ignored domain=ai moduleIds=example
     hint=See .ai/specs/implemented/2026-05-04-modules-ts-unified-overrides.md and tracking issue #1787
```

Two problems compound here. The functional one: an app that disables an agent or tool through `modules.ts` still gets the base agent in every worker, scheduler and CLI process. The diagnostic one: the operator is told the domain is "not yet wired" and pointed at a tracking issue, when `ai` is in fact Phase 1 and has been wired since 2026-05-04 — so the message sends them looking in the wrong place entirely. This also makes the change's failure mode subtler than the bug it fixes: "no overrides apply in the CLI" is noticeable, "all overrides except `ai` apply in the CLI" is not.

**Fix.** Soft-resolve the applier before dispatching, only when it is actually needed, so `@open-mercato/shared` keeps its documented rule of never taking a hard runtime dependency on `@open-mercato/ai-assistant` (`overrides.ts:44-55`) — this is the soft-optional coupling pattern in `packages/core/AGENTS.md` → Cross-Module Coupling:

```ts
if (enabledModules.some((entry) => entry?.overrides?.ai)) {
  try {
    await import('@open-mercato/ai-assistant/modules/ai_assistant/lib/ai-overrides')
  } catch {
    // ai-assistant is not installed in this app — the ai domain has nothing to apply to
  }
}
applyModuleOverridesFromEnabledModules(enabledModules as ModuleEntryWithOverrides[])
```

Add a fourth case to the new suite asserting an `overrides.ai` entry survives `bootstrapFromAppRoot`. If you would rather keep the PR narrow, that is defensible — but then the docstring at `dynamicLoader.ts:600-604` and the PR body must stop implying full coverage (the `…` after "setup" and "the same gap applied to every other override domain" both read as complete), and the remaining domain should get a follow-up issue.

### 2. 🟠 Major — fail-soft here contradicts the file's own mandatory/optional convention, on the security surface the PR exists to protect

`packages/shared/src/lib/bootstrap/dynamicLoader.ts:614-640`

This file already distinguishes the two cases deliberately. Mandatory inputs — `modules.cli.generated.ts`, `entities.generated.ts`, `di.generated.ts` — go through a bare `compileAndImport` and **throw**. `loadOptionalGeneratedModule` and its log-and-continue exist for registries an older app may legitimately not have generated yet; that is what the `#4327`/`#4491` precedent in the docstring is about. `src/modules.ts` is not in that category, and the docstring concedes it: "even though the file is normally mandatory". In the Next.js runtime the same file is a hard boot failure — `bootstrap-common.ts:4` imports it statically.

The consequence is specific. When a *present* `src/modules.ts` fails to compile or import — an esbuild error, a read-only `.mercato/generated` in a hardened container so `mkdirSync`/`writeFileSync` for `app-modules-overrides.compiled.mjs` throws, a half-written cache file from two CLI processes racing — the `catch` at line 633 logs one error and bootstrap proceeds. `yarn mercato entities seed-encryption` then reads `getDefaultEncryptionMaps(getCliModules())` with an empty override store, seeds the base maps, and prints `✅ Encryption maps seeded`. That is #5582's outcome verbatim, with the override-added fields still written as plaintext at rest, now hidden one `logger.error` line above a success banner instead of being unconditional.

**Fix.** Keep the `!fs.existsSync(tsPath)` debug-and-skip — an app layout with no `src/modules.ts` is a real compatibility case. But let a present-and-unloadable `src/modules.ts` throw, matching how this same function treats every other mandatory input and how the Next path treats this same file:

```ts
} catch (error) {
  throw new Error(
    `Failed to load the app-level modules file (${tsPath}); entry.overrides cannot be applied. ` +
    `Refusing to bootstrap with a partial override set.`,
    { cause: error },
  )
}
```

If you'd rather not make CLI bootstrap fail on this, the minimum alternative is to make the degradation legible where it matters: record the failure in a module-scope flag exported alongside the loader, and have `seed-encryption` refuse to print success while it is set. Silently succeeding is the one outcome this command must not have.

### 3. 🔵 Minor — "Breaking Changes: None" undersells a real behavior change for existing apps

No contract surface changes — agreed, and `BACKWARD_COMPATIBILITY.md` is untouched. But this flips runtime behavior for any app that *already* declares `entry.overrides`, and it does so at upgrade with no notice. `apps/mercato/src/modules.ts:130-141` already carries `acl.features['example.manage'] = null` plus a `routes.api` probe; a downstream app carrying `overrides.cli['some command'] = null` will see that CLI command vanish after upgrading, and one carrying `setup.seedDefaults: false` will see `mercato setup` stop seeding. Each of those is the correct new behavior — which is precisely why an operator needs to be told to expect it.

**Fix.** Add an `UPGRADE_NOTES.md` entry naming the newly-effective domains in CLI/worker/scheduler processes and telling operators to re-check any `entry.overrides` they declared, since some were silently inert before this release.

### 4. 🔵 Minor — the new suite never exercises the compile path, which is where the residual risk lives

`packages/shared/src/lib/bootstrap/__tests__/dynamicLoader.appModuleOverrides.test.ts:101-152`

`writeCompiledPair` pre-writes both the `.mjs` and a valid `.cache.json`, so `compileAppSourceFileWithActiveEsbuild` takes its `cacheIsValid` early return and esbuild never runs; the stub `src/modules.ts` imports nothing. The suite therefore proves the dispatch call was added and reaches `getDefaultEncryptionMaps`, but not the two things that could still make this a no-op against a real app: that an actual `src/modules.ts` — `@/` alias, `import type { ModuleOverrides }` erased correctly, external package imports, a bundled `official-modules.generated` — survives `createCliBundlePlugins`; and that the compiled ESM module and the CLI share one `@open-mercato/shared/modules/overrides` instance. The second one does hold today (see ✅ above), but nothing in the suite would catch it regressing if shared ever gained a CJS build alongside the ESM one.

This mirrors the existing `appDi` guard, so it is a pre-existing scaffolding choice rather than something this PR invented — the difference is that `loadAppDiRegistrar` returns a value the caller checks, while this fix depends entirely on cross-module singleton state, which is the part left untested.

**Fix.** Add one case that lets esbuild actually compile a small `src/modules.ts` (no pre-written `.mjs`), or extend the existing `override-probe` / `TC-EXAMPLE-010` integration coverage with an assertion that an override applies under a real `bootstrapFromAppRoot`.

### 5. 🔵 Minor — the spec changelog is not updated

`.ai/specs/implemented/2026-05-04-modules-ts-unified-overrides.md`

Its risk table still states the mitigation as "The dispatcher is called from `bootstrap.ts` BEFORE any registry loads". After this PR that is half the picture: there is a second dispatch site with its own ordering guarantee, and the next person reasoning about override ordering will read the spec, not `dynamicLoader.ts`. `.ai/specs/AGENTS.md` asks for a changelog entry on significant changes, and adding a second entrypoint to the dispatcher qualifies.

**Fix.** One bullet: the date, the second entrypoint (`bootstrapFromAppRoot`, for CLI/worker/scheduler), and the ordering constraint it satisfies (before `registerModules` and before `registerCliModules`).

### 6. ⚪ Nit — docs still describe the split this PR removes

`apps/docs/docs/framework/modules/overrides.mdx:296` reads "The standard app bootstrap already calls `applyModuleOverridesFromEnabledModules(enabledModules)` for inline `modules.ts` overrides", which leaves a reader assuming the Next runtime is the only place overrides land. One sentence noting that CLI, worker and scheduler processes dispatch the same overrides through `bootstrapFromAppRoot` would close the gap that made #5582 surprising in the first place.

## 🧪 Validation

| Command | Result |
|---|---|
| `jest` — `dynamicLoader.appModuleOverrides`, `dynamicLoader.appDi`, `modules/overrides` | ✅ 3 suites, 13 tests passed (run locally on this head) |
| Ad-hoc probe reproducing finding 1 | ✅ confirms `overrides.ai` is dropped with only the built-in appliers registered |
| Full `validation.commands` gate | ⏭️ not re-run locally — every required check on this head is green (`test`, `lint`, `ds-lint`, `docker-build`, `CodeQL`, `ephemeral-integration` 1–15/15, `documents-multi-instance`) |

## 🔁 Backward compatibility

No frozen or stable contract surface is touched: no exported signature changes, no route, event id, DI key, ACL feature, DB schema or generated-file format change. `loadAppModuleOverrides` is module-private. The only observable change is the one finding 3 asks you to document.

## 📋 Summary

The mechanism, the placement in the esbuild lifecycle, and the ordering relative to `registerModules`/`registerCliModules` are all correct, and the regression guard is a real one. Two things to settle before this lands: make `overrides.ai` actually reach the CLI path (or stop claiming it does), and decide whether a present-but-broken `src/modules.ts` may silently downgrade `seed-encryption` to base maps — this file's own convention and the Next path both say it should not. The three minors and the nit are documentation and coverage follow-ups that can ride along in the same push.

adeptofvoltron:

- 🔄 `review` → `changes-requested` — the review found 2 majors with no waiver (`overrides.ai` still dropped in the CLI path; fail-soft on a present-but-broken `src/modules.ts` silently downgrading `seed-encryption`), so the PR moves out of the review queue and back to the author.
- ✅ `bug`, `skip-qa`, `priority-high`, `risk-high` — preserved as `om-open-pr` set them; nothing in the review changes the category, the QA exemption (no UI-rendering files, no schema or API surface change, regression test ships in the same PR), the urgency, or the blast radius.
- ⏭️ no `ci-monitoring` — every required check on this head is already green, so there is no CI follow-up to promise.

Draft state left as-is: the PR stays a draft because it is not approvable, not because the review touched it.

adeptofvoltron:

The verdict is published as a `COMMENTED` review rather than a `REQUEST_CHANGES` one — GitHub refuses both `APPROVE` and `REQUEST_CHANGES` on a PR authored by the reviewing account, and this automation runs as the author. The `changes-requested` pipeline label carries the state instead.

- **Findings:** 0 blockers, 2 majors, 3 minors, 1 nit. The majors are `overrides.ai` still being dropped in the CLI/worker path (reproduced against this head), and fail-soft on a present-but-unloadable `src/modules.ts` leaving `seed-encryption` to report success after seeding base maps.
- **autofix:** skipped by request — the author is applying fixes by hand.
- **Handoff:** author and reviewer are the same account, so the assignee is unchanged.
- **CI:** all required checks green on `1c13a14`; no follow-up owed, so no `ci-monitoring`.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-09-02T13:00:00Z.

Driving this PR to merge-ready: merging the latest `develop` in first, then running `om-auto-review-pr --autofix` to clear the two majors and the follow-on minors from the 2026-09-02 review, re-checking CI, and normalizing labels. This skill never merges — the hand-off is to `om-approve-merge-pr`.

adeptofvoltron:

- 🔄 `changes-requested` → `merge-queue` — the two majors that drove the 2026-09-02 `om-auto-review-pr` verdict are fixed in `d55f48b`, and the three minors and the nit are fixed in `73d84f9`; the one finding deliberately not fixed is tracked as #5855, so nothing actionable is left holding the PR.
- 🐛 `bug` — unchanged. This is still the same defect: `entry.overrides` declared in an app's `src/modules.ts` never reached CLI, worker or scheduler processes.
- ⏭️ `skip-qa` — unchanged, and re-checked against the new diff. The change still touches no `.tsx` or UI-rendering file, no database structure, and no API surface; the files added since the last review are `UPGRADE_NOTES.md`, a spec, and one docs sentence. The automated-verification exemption therefore still applies, and the regression coverage ships in this same PR.
- 🔺 `priority-high` — unchanged. The symptom remains security-relevant: `seed-encryption` silently seeding base encryption maps left override-added fields as plaintext at rest.
- ⚠️ `risk-high` — unchanged, and if anything better earned now. The change lives in `bootstrapFromAppRoot()`, and this push makes a present-but-unloadable `src/modules.ts` abort that bootstrap rather than degrade — a deliberate, reviewed hardening with a broad blast radius across every CLI, worker and scheduler process.
- 👀 `in-progress` → `ci-monitoring` — the fixing work is finished and reported, so the PR is no longer claimed and anyone may act on it. `ci-monitoring` records only that a CI-result follow-up comment is still owed on this head.

Draft state: promoted to ready. The PR was a draft because it was not approvable, not by intent; with the review findings cleared, that reason is gone.

adeptofvoltron:

Driving this PR to merge-ready. Every finding from the 2026-09-02 `om-auto-review-pr` review is now accounted for: five fixed here, one tracked as a follow-up because it is not achievable at the tier it was requested in.

### 🔀 Base merge

`origin/develop` was merged into the PR head first (`e888b7a`), before any review or CI work, so everything below judges the real merge result rather than a seven-day-old branch. The merge was clean — no conflicts, and the diff against `develop` still consists only of this PR's own files.

### 🔍 How each review finding ended up

**1. 🟠 Major — `overrides.ai` silently dropped in the very path this PR fixes.** Fixed in `d55f48b`. `registerBuiltInModuleOverrideAppliers()` registers 15 domains and not `ai`; that applier is installed as a side effect of importing the ai-assistant module, which is exactly why `bootstrap-common.ts` imports it statically on the line before it dispatches. The dynamic path now resolves the same module on demand through a new `OPTIONAL_OVERRIDE_APPLIER_MODULES` map — attempted only when an app actually declares the domain, and fail-soft when the package is not installed — so `@open-mercato/shared` keeps its documented rule of never taking a runtime dependency on a domain package. New test case asserts an `overrides.ai` entry reaches an applier instead of the dispatcher's unwired branch.

**2. 🟠 Major — fail-soft contradicted the file's own mandatory/optional convention.** Fixed in `d55f48b`, the way the review argued for. A `src/modules.ts` that is *present* but fails to compile or import now throws rather than logging and continuing: continuing is what left `seed-encryption` reading an empty override store, seeding base maps and printing `✅ Encryption maps seeded` with override-added fields still plaintext at rest — #5582's outcome, one log line quieter. An **absent** `src/modules.ts` is still a debug-and-skip, since that layout is a genuine compatibility case. A non-array `enabledModules` export now throws for the same reason. Two test cases cover the throws and one covers the absent-file skip.

**3. 🔵 Minor — "Breaking Changes: None" undersold a real behavior change.** Fixed in `73d84f9`. `UPGRADE_NOTES.md` gains a 0.7.1 entry naming the domains that become newly effective in CLI/worker/scheduler processes and listing the concrete cases an operator must re-check — including re-running `seed-encryption` and re-encrypting fields its earlier silent no-op skipped.

**4. 🔵 Minor — the suite never exercises the compile path.** **Not fixed here — tracked as #5855**, and this is the one finding worth reading in full. I wrote the requested fourth case, and it is not achievable at this tier: letting esbuild actually compile a real `src/modules.ts` emits genuine ESM, which Jest's CJS runtime then refuses to `import()` (`SyntaxError: Unexpected token 'export'` on the emitted `app-modules-overrides.compiled.mjs`). That is precisely why the existing stubs — and the #4327/#4491 guards beside them — are hand-written `module.exports`. The constraint is now documented in the test file header, and #5855 tracks the review's own second option: covering it at the integration tier, where a real process runs against a real app.

What I could do instead was verify it by hand, which I did: running the real `mercato` bin against `apps/mercato` on this branch compiled the app's actual `src/modules.ts` through the CLI bundle plugins (producing `.mercato/generated/app-modules-overrides.compiled.mjs`), reached command dispatch with no load error, and emitted no "domain not yet wired" warning. That is real-world evidence for the path the unit suite cannot reach, but it is not a regression guard, which is why #5855 stays open.

**5. 🔵 Minor — the spec changelog is not updated.** Fixed in `73d84f9`. The overrides spec's risk table said the ordering mitigation was "called from `bootstrap.ts` BEFORE any registry loads"; it now names both entrypoints, adds the missed-entrypoint risk that produced #5582, and records that a domain whose applier arrives via a package import must be added to `OPTIONAL_OVERRIDE_APPLIER_MODULES`. A dated changelog entry states the second dispatch site and the ordering constraint it satisfies.

**6. ⚪ Nit — docs still describe the split this PR removes.** Fixed in `73d84f9`. `framework/modules/overrides` no longer leaves a reader assuming the Next runtime is the only place inline `modules.ts` overrides land.

One thing the review did not raise, found while fixing: the new dynamic import used an `@vite-ignore` hint, while this repo's convention for the same kind of variable-specifier import is `webpackIgnore`/`turbopackIgnore` (see `generated-registry-loader.ts`, which dynamically imports this very module). Corrected in `fa9cfbf`.

### 🧪 Validation

The full `validation.commands` gate was run locally on this head, in order. Two caveats, both environmental and both verified as such rather than assumed:

| Command | Result |
|---|---|
| `yarn build:packages` (×2, before and after `generate`) | ✅ 27/27 tasks; `@open-mercato/shared` was a genuine cache miss and rebuilt with this change |
| `yarn generate` | ✅ passed |
| `yarn i18n:check-sync` | ✅ all 5 locales in sync |
| `yarn i18n:check-usage` | ✅ exit 0 |
| `yarn typecheck` | ✅ full monorepo |
| `yarn test` | ✅ 33/34 turbo tasks; `@open-mercato/shared` 190 suites / 2121 tests, zero failures |
| `yarn build:app` | ✅ passed |

- The first `yarn test` pass reported two failures in `dynamicLoader.generatedCacheRecovery.test.ts`. That is **not** a regression: turbo drops the exported `TMPDIR`, so the suite ran against this worktree's long in-repo temp path and esbuild's helper failed to bind its pipe (`listen EINVAL … /tsx-501/50567.pipe`). Re-run with the env passed through, the whole `dynamicLoader` group is 8 suites / 32 tests green, and the full `shared` package is clean.
- The single remaining red turbo task is `create-mercato-app#test`, whose failing suites are the AI-harness oracles (`OMH-001`) that shell out to the `codex`/`claude` CLIs. They fail identically on a clean checkout and touch nothing in this diff.

### 🔁 Follow-ups filed

- #5855 — cover the esbuild compile path for `src/modules.ts` `entry.overrides` at the integration tier. Deferred rather than fixed because the Jest CJS runtime cannot import the ESM esbuild emits; the acceptance criteria carry the specific constructs (`@/` alias, erased `import type`, external package import) that would catch a `createCliBundlePlugins` regression, plus the cross-module singleton assumption.

### 🚀 Merge readiness

The review is approvable: no conflicts, no blockers, and no un-waivered majors remain. This skill does **not** merge — the PR is handed to `om-approve-merge-pr`, which re-checks every gate first. The PR carries `skip-qa`, so the QA-approval gate does not apply, but note that this repository's automation account authored the PR, so it still needs a second, human reviewer before it can land.

**CI is still running on this head (`fa9cfbf`).** Branch protection holds the actual merge; this run's verdict covers the work, not a green run — the table above is this run's own local evidence, never a substitute for the required checks. Still pending at the time of writing: `comment`, `Analyze (javascript-typescript)`. A follow-up comment will report the CI outcome.

adeptofvoltron:

**✅ Green. Every required check passes on `fa9cfbf`, and the merge-ready verdict from the run summary stands unchanged.** Both workflows finished `completed/success`: `CI for Develop&Main` and `Mutation tests`. The final tally is **30 passing checks, 0 failing**, with 3 conditional jobs skipped (`[code]smith` — autofix is disabled on this PR; `audit`; `merge-coverage`), exactly the three that were also skipped on the last green head of this branch.

The checks worth calling out for this particular change:

- `test`, `lint`, `ds-lint`, `typecheck` — ✅ green, matching the local `validation.commands` gate reported above. In particular the `dynamicLoader` suites pass in CI, which retires the one local red signal I reported: the `generatedCacheRecovery` failures were this worktree's long in-repo `TMPDIR` breaking esbuild's helper pipe, not the code.
- `ephemeral-integration (1–15/15)` — ✅ all fifteen shards green. This matters more than usual here, because the second major fix makes a present-but-unloadable `src/modules.ts` **abort** the CLI/worker bootstrap rather than degrade. If that hardening had been too aggressive for any real app layout, the integration shards booting real processes are where it would have surfaced.
- `docker-build`, `documents-multi-instance` — ✅ green, covering the other process types that boot through `bootstrapFromAppRoot()`.
- `mutate (shared, src/lib/bootstrap/dynamicLoader.ts)` — ✅ green. Mutation testing runs against exactly the file this PR changes, so the new and amended test cases genuinely kill their mutants rather than merely executing the changed lines.

The `ci-monitoring` label is removed: the follow-up this run owed has now landed, and nothing further is being watched.

### 🚀 Where this leaves the PR

Merge-ready and handed off. `om-approve-merge-pr` owns the merge and will re-check every gate before squash-merging; this skill does not merge. Two things a human still owes it: `skip-qa` applies so the QA-approval gate does not block, but **this PR was authored by the account the automation runs as, so GitHub will not accept an approving review from it** — it needs a second, human reviewer before it can land. The deferred coverage item is tracked as #5855 and is not a merge blocker.

patzick (approved):

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-09-10T18:26:24Z. Other auto-skills will skip this PR until the lock is released.

Re-review of the current head `fa9cfbf` (the 2026-09-02 review covered `1c13a14`). The PR carries `merge-queue` without a verdict from this reviewer on the current code, so this run re-checks the two previously-raised majors and the override dispatch order against the web bootstrap.

adeptofvoltron (commented):
# 🔍 Re-review (post-merge) — PR #5665

**Verdict: the change is correct on all three axes this re-review was asked to verify. 0 blockers, 1 major, 2 minors — none of them a reason to revert.**

*(Reviewed as `om-auto-review-pr` at head `fa9cfbf`. The previous pass from this reviewer covered `1c13a14`, before the two majors were fixed. **This PR was already squash-merged as `0a7634eb3` at 2026-09-10T12:21:57Z**, seven seconds after @patzick's approval and before this run started, so there is no verdict left to gate a merge — GitHub also refuses `APPROVE`/`REQUEST_CHANGES` on a PR authored by the reviewing account. The findings below are follow-up work against `develop`, not merge blockers.)*

## The three questions this review was asked

### 1. Does the override dispatch order match the web bootstrap? ✅ Yes

| | Web (`bootstrap-common.ts`) | CLI/worker/scheduler (`bootstrapFromAppRoot`) |
|---|---|---|
| `ai` applier registered | static side-effect import, hoisted above the call | `ensureOptionalOverrideAppliers()`, awaited before the dispatch |
| 15 built-in appliers | `registerBuiltInModuleOverrideAppliers()` at `overrides.ts` module-load | same |
| Dispatch | `applyModuleOverridesFromEnabledModules(enabledModules)` | same function, same export, same file |
| Consumers first read the store | `registerEventModuleConfigs` / `runBootstrapRegistrations()`, after | `registerModules` inside `bootstrap()`, and `registerCliModules` at `packages/cli/src/bin.ts:55`, both after |

The one *real* ordering difference is that the CLI path dispatches **before** `loadBootstrapData()` imports the module graph, whereas ES import hoisting puts the web dispatch **after** it. That is safe, and I checked why rather than assuming it: every one of the 16 appliers writes into a keyed store — `applyStoreOverrides(store, 'modules', …)` at `packages/shared/src/modules/overrides.ts:562` assigns `store.modules[key]`, `setupOverridesApplier` assigns `setupOverridesByModule[moduleId]`, `navOverridesApplier` assigns `state.modules`, `routesOverridesApplier` assigns `modulesConfigApiRouteOverrides[key]`. Base module registration never touches the `modules` tier, and `composeStore` merges tiers at read time. So dispatching earlier cannot be clobbered, and the double dispatch (`bin.ts:54` then `mercato.ts:1110` in the same process) is idempotent for the same reason.

The comment at `dynamicLoader.ts:774-780` states the constraint accurately.

### 2. Can a worker now pick up an override the web path does not, or vice versa? ✅ No

Both paths read the **same `enabledModules` export from the same `src/modules.ts`** and push it through the **same dispatcher**, which walks a single `DOMAIN_KEYS` list. There is no per-path domain filter anywhere.

The `ai` domain was the one place asymmetry could hide, and it does not: `packages/ai-assistant/…/lib/ai-overrides.ts` is a deliberately focused module whose only top-level side effect is `registerModuleOverrideApplier('ai', …)` (lines 204-215). Importing it on demand is therefore exactly equivalent to the web path's unconditional static import — the on-demand form just skips it when no entry declares the domain, which is a no-op case.

Domains only one path *consumes* — `cli` (read by `registerCliModules`), `routes.pages` / `widgets` / `guards` / `nav` (read by the Next runtime) — are dispatched in **both** and consumed where relevant. That is symmetric by construction, not a gap.

### 3. Does this change module resolution for third-party modules? ✅ No, for a descriptor-only `modules.ts`

`src/modules.ts` is a pure descriptor file: `parseBooleanWithDefault`, a `type` import, a relative `./official-modules.generated`, and an array of `{ id, from, overrides }`. It imports **no module packages**, so compiling and importing it pulls in no module code.

The artifact lands at `<appDir>/.mercato/generated/app-modules-overrides.compiled.mjs` — the same directory as the existing generated registries — and `createCliBundlePlugins`' `external-non-json` plugin (`dynamicLoader.ts:116-121`) leaves every package specifier external. Same directory + same externalization means bare specifiers resolve through the identical `node_modules` walk, so there is no second instance of `@open-mercato/shared` or of any module package. `registerModules` / `registerCliModules` still consume `modules.cli.generated.ts` unchanged.

That caveat — "for a descriptor-only `modules.ts`" — is finding 1.

## ✅ What else is right

- The fail-hard decision is the correct call and is honestly documented. `UPGRADE_NOTES.md` names the eight domains that become newly effective, calls out `seed-encryption` re-running and re-encryption explicitly, and separately documents the present-but-unloadable abort. That is the rare upgrade note that would actually save an operator.
- The throw path does not leak the esbuild helper process: `withEsbuildLifecycle` (`dynamicLoader.ts:35-58`) releases in a `finally`.
- Six regression cases, and they cover the shapes that matter: dispatch, no-override baseline, present-but-broken → throw, missing `enabledModules` → throw, absent file → skip, `ai` resolved on demand. The header comment states the compile-path stubbing constraint and points at #5855 instead of quietly leaving a hole.
- No backward-compatibility violation. `loadAppModuleOverrides` and `ensureOptionalOverrideAppliers` are module-private; no exported signature moved; the `applyModuleOverridesFromEnabledModules` contract at `BACKWARD_COMPATIBILITY.md:136` ("MUST keep dispatching by module-load order") still holds. The one behavioral change to an exported function — `bootstrapFromAppRoot` can now throw — is what `UPGRADE_NOTES.md` documents.

**Validation run locally at `fa9cfbf`:** `packages/shared` — `src/modules/__tests__` + `src/lib/bootstrap/__tests__`, **19 suites / 169 tests, all passing**. The turbo-driven half of the gate was not run in this worktree (no local `node_modules`); CI is green on this head, including all 15 `ephemeral-integration` shards, which is the signal that matters most here because those boot real processes through the hardened path.

---

## 🟠 Major — the fail-hard abort is scoped to *any* unloadable `src/modules.ts`, including apps that declare no overrides at all

`loadAppModuleOverrides` throws whenever `src/modules.ts` exists but does not compile or import (`dynamicLoader.ts:664-680`). The override content is irrelevant to that decision — it has to be, since you cannot know what is inside a file you failed to load. The consequence is that **any** app whose `modules.ts` is not compilable by this specific esbuild configuration loses **every** CLI command, worker and scheduler process, where before this PR it lost nothing.

That is not hypothetical for the surface the docs advertise. `apps/docs/docs/framework/modules/overrides.mdx` documents inline *replacement* implementations, not just `null` disables — API route handlers (line 131), worker handlers (line 157), `widgets.components` entries with a `propsTransform` (line 174-178), DI `register: (container, key) => …` (line 232). An app writing those inline gives `src/modules.ts` a real import graph, and `compileAppSourceFileWithActiveEsbuild` (`dynamicLoader.ts:434-448`) bundles everything non-package: `@/` aliases and relative imports are pulled in, with `loader: { '.json': 'json' }` and nothing else. A single `import './panel.css'` anywhere in a bundled component graph fails the build with *"No loader is configured for .css files"* — and now takes down `mercato db migrate` with it.

The blast radius is worse than a normal boot failure because the CLI is the recovery tool: an operator in this state cannot run the command that regenerates the registries, nor any diagnostic.

`apps/mercato` and the `create-mercato-app` template both use descriptor-only overrides, so CI cannot see this — the green integration shards confirm the hardening is safe for *those* layouts, not for the documented one. #5855 already tracks covering the esbuild compile path at the integration tier; a case that declares an inline replacement importing local code belongs in that same coverage.

**Suggested follow-up:** state the constraint where the feature is documented — inline `overrides` values in `src/modules.ts` must be importable by a plain Node process, since every CLI/worker/scheduler now compiles that file — and add the compile case to #5855's scope.

## 🟡 Minor — `ensureOptionalOverrideAppliers` reports the wrong cause, at a level nobody sees

```ts
} catch (error) {
  logger.debug('Optional override applier module is not installed; the domain has nothing to apply to', …)
}
```
`dynamicLoader.ts:614-620`

The message asserts a cause the `catch` cannot know. It is right for `ERR_MODULE_NOT_FOUND`, and wrong for every other failure: a genuine throw inside `ai-overrides.ts`, or — more plausibly — `@open-mercato/ai-assistant` being installed but not resolvable *from `@open-mercato/shared`'s location* under a non-hoisted (pnpm-style) install, since `shared` deliberately does not depend on it. In those cases the app's declared `overrides.ai` is dropped in CLI while it applies in web: the exact asymmetry class this PR exists to eliminate, reintroduced on one domain.

The comment's claim that "the dispatcher's own *domain not yet wired* warning still fires behind it" is accurate — I verified it reaches `overrides.ts:288-294` at `warn` level, so this is not fully silent. But that warning names the wrong cause too, and it is one-shot per process.

**Suggested fix:** narrow the swallow to `ERR_MODULE_NOT_FOUND` / `ERR_PACKAGE_PATH_NOT_EXPORTED` and log anything else at `error`.

## 🟡 Minor — no escape hatch out of the fail-hard state

Given the major above, the hardening deserves a documented recovery path: an `OM_ALLOW_PARTIAL_OVERRIDES=1` (or a `--skip-app-overrides` flag) that downgrades the abort to an `error`-level log. That keeps the property the hardening was bought for — an operator can never *silently* get partial overrides, which is what made #5582 dangerous — while leaving a way to run the CLI at all when `modules.ts` is broken. As shipped, the only recovery is editing `src/modules.ts` by hand with no tooling available to validate it.

## ⚪ Nit — none blocking

The `#5855` cross-reference in the test header is exactly the right way to leave a known coverage gap. No action.

---

### Inherited feedback

@patzick's approving review carries no body and no inline comments; the `github-actions` mutation-testing comment is empty. `list-review-comments` returned no inline diff comments. Everything else on the thread is this collection's own markers, filtered per the skill's rules. **No inherited findings.**

### Labels

Unchanged. The PR is merged, so its pipeline label is now historical record — rewriting `merge-queue` after the fact would misrepresent the state the PR merged in. Only this run's `in-progress` lock is released.

adeptofvoltron:
🤖 `om-auto-review-pr` completed: REVIEWED (post-merge, no blockers). Lock released.

This PR was already squash-merged as `0a7634eb3` before this run started, so there was no merge to gate — the re-review of head `fa9cfbf` is published as a `COMMENTED` review and its findings are follow-up work against `develop`.

- **The three questions this run was asked, all answered ✅** — the dispatch order matches the web bootstrap (both register the 15 built-in appliers plus `ai` before dispatching, and both dispatch before any consumer first-reads the store); no web/worker asymmetry (one `enabledModules` export, one dispatcher, one `DOMAIN_KEYS` list, and the on-demand `ai` import is equivalent to the web path's static one because that module's only top-level side effect is `registerModuleOverrideApplier`); and module resolution is unchanged for third-party modules, because `src/modules.ts` is a descriptor-only file and its artifact compiles into the same `.mercato/generated/` directory with every package specifier left external.
- **Findings:** 0 blockers, 1 major, 2 minors, 1 nit. The major is that the fail-hard abort keys on *any* unloadable `src/modules.ts`, so an app declaring the inline replacement implementations the docs advertise (`widgets.components`, DI `register`, route handlers) can lose every CLI, worker and scheduler process to a bundling error — a case neither the tests nor CI can currently reach.
- **Inherited feedback:** none. The approving review carries no body, there are no inline diff comments, and the rest of the thread is this collection's own markers.
- **Validation:** `packages/shared` — `src/modules/__tests__` + `src/lib/bootstrap/__tests__`, 19 suites / 169 tests, all passing at `fa9cfbf`. The turbo half of the gate was not run in this worktree (no local `node_modules`); CI is green on this head, all 15 `ephemeral-integration` shards included.
- **autofix:** not applicable — the PR is merged; nothing to fix forward on this branch.
- **Labels:** unchanged. `merge-queue` is now historical record of the state this PR merged in, and rewriting it after the fact would misrepresent that. Only this run's `in-progress` lock is removed.

PR #5670, opened 2026-08-26
fix(query_index): include cf sorts in wantsCf so sort-only queries keep the ORM fallback (#5521)

### LINKED ISSUE #5521: query_index: HybridQueryEngine's wantsCf ignores opts.sort, so sort-only custom-field queries skip the ORM fallback / coverage check
## Summary

`HybridQueryEngine.query()` decides whether a query "wants custom fields" (`wantsCf`) by inspecting `opts.fields`, the normalized `cf:` / `l10n:` filters, and `opts.includeCustomFields` — but **not `opts.sort`**. A query that only *sorts* by a custom field (`sort: [{ field: 'cf:my_field', dir: 'asc' }]`, e.g. from a CRUD list route receiving `?sortField=cf_my_field`) therefore skips the whole custom-field branch: the `entityHasActiveCustomFields` check, the `indexAnyRows` "no index rows → fall back to ORM engine" guard, and the partial-index coverage check. If `entity_indexes` is empty or under-populated for that entity, the sort silently runs against whatever the index has instead of falling back.

Observed on `@open-mercato/core@0.6.7`.

## Where

`packages/core/src/modules/query_index/lib/engine.ts` (installed copy: `node_modules/@open-mercato/core/src/modules/query_index/lib/engine.ts`):

```ts
// ~L271-277
const normalizedFilters = normalizeFilters(opts.filters)
const cfFilters = normalizedFilters.filter((filter) => filter.field.startsWith('cf:') || filter.field.startsWith('l10n:'))
const wantsCf = (
  (opts.fields || []).some((field) => typeof field === 'string' && (field.startsWith('cf:') || field.startsWith('l10n:'))) ||
  cfFilters.length > 0 ||
  (Array.isArray(opts.includeCustomFields) && opts.includeCustomFields.length > 0)
)
```

Meanwhile `applySort` (~L827-836) does handle `cf:` sort fields — it builds `buildCfTextExprSql(...)` and emits an `ORDER BY` — so a `cf:` sort is clearly a supported input; it just never participates in the `wantsCf` / fallback decision.

The CRUD factory makes this easy to hit: `packages/shared/src/lib/crud/factory.ts` passes any `sortField` without a `sortFieldMap` entry straight through and normalizes `cf_<key>` → `cf:<key>` (`normalizeSortFieldSelector`), so any list endpoint accepts `?sortField=cf_<key>`.

## Repro

1. Entity with an active custom field definition (e.g. a `float` field on `customers:customer_deal`).
2. `entity_indexes` empty for that entity (fresh tenant, or before a reindex), or with a coverage gap.
3. `GET /api/customers/deals?sortField=cf_<key>&sortDir=asc` (no `cf:` fields, filters, or `includeCustomFields`).
4. `wantsCf === false` → no `indexAnyRows` / coverage check → no ORM fallback; the response is ordered by the (missing) index value, not by the custom field.

With the same request plus `&fields=cf:<key>` (or a `cf:` filter) the guards run and the fallback kicks in — the behaviour differs only because of which option carries the `cf:` reference.

## Suggested fix

Include sorts in the predicate:

```ts
const cfSorts = (opts.sort ?? []).filter((s) => String(s.field).startsWith('cf:') || String(s.field).startsWith('l10n:'))
const wantsCf = (
  (opts.fields || []).some(/* … */) ||
  cfFilters.length > 0 ||
  cfSorts.length > 0 ||
  (Array.isArray(opts.includeCustomFields) && opts.includeCustomFields.length > 0)
)
```

(and let `cfSorts` feed the same coverage/partial-index reasoning that `cfFilters` does).

## Related observation (may deserve its own issue)

Both engines order `cf:` sorts **as text**, regardless of the field's `kind`:

- Hybrid/index engine: `buildCfTextExprSql` → `coalesce(doc ->> 'cf:<key>', …)`; jsonb `->>` always yields text.
- ORM fallback (`packages/shared/src/lib/query/engine.ts` ~L783-789): the sort alias is built over a `CASE kind WHEN 'integer' THEN (value_int)::text WHEN 'float' THEN (value_float)::text …` expression, and ~L872-885 orders by that alias.

So sorting by a numeric custom field is lexicographic: `'10240' < '1024' < '2048' < '512'`. Neither engine appends a stable tiebreak (`id`) either, so paginated results with ties/NULLs are unstable across pages. We hit both while trying to sort a kanban lane by a `float` rank field and had to fall back to client-side ordering.

adeptofvoltron:
Closes #5521

## 🎯 Goal

Sorting a CRUD list by a custom field (`?sortField=cf_<key>`) no longer returns silently mis-ordered rows when the query index is empty or under-populated for that entity — the query now falls back to the ORM engine exactly as it already does for custom-field *filters*.

## 🔍 Problem

`HybridQueryEngine.query()` decided whether a query "wants custom fields" (`wantsCf`) from three signals: a `cf:`/`l10n:` entry in `opts.fields`, a non-empty set of normalized `cf:`/`l10n:` filters, and — precisely — `Array.isArray(opts.includeCustomFields) && opts.includeCustomFields.length > 0`. That third disjunct counts only the **array** form: the boolean `true` form does not qualify, and `true` is exactly what `makeCrudRoute` passes on every list request (`packages/shared/src/lib/crud/factory.ts`). That asymmetry looks like a bug and is not one — it is why the defect was reachable from every CRUD list route in the first place, and it should stay: making the boolean form count would put the coverage probes and the fallback on *every* list query in the app rather than only cf-sorted ones. `opts.sort` was not among the three signals at all. A query that only *sorts* by a custom field therefore skipped the whole custom-field branch, and with it every safety guard that branch owns.

The CRUD factory makes this trivially reachable: `normalizeSortFieldSelector` in `packages/shared/src/lib/crud/factory.ts` passes any unmapped `sortField` straight through and rewrites `cf_<key>` → `cf:<key>`, so every list endpoint accepts `?sortField=cf_<key>`. The same request with `&fields=cf:<key>` added behaved correctly — the outcome differed only by which option happened to carry the `cf:` reference.

## 🔍 Root Cause

`applySort` (`packages/core/src/modules/query_index/lib/engine.ts`) does support `cf:` sort fields: it builds `buildCfTextExprSql(...)` and emits an `ORDER BY` reading the `entity_indexes` alias `ei`. That join is a **LEFT** join (`applyEntityIndexesJoin`), so when no index row exists the sort expression evaluates to `NULL` for every row instead of raising — the failure is silent by construction.

With `wantsCf === false`, the guard block was never entered, so none of these ran:

- `entityHasActiveCustomFields(...)`
- the `indexAnyRows` check and its "no index rows → fall back to the ORM engine" return
- the partial-index coverage gap check (and its `partialIndexWarning` / auto-reindex scheduling)

The result: an `ORDER BY` over an all-`NULL` expression, returning rows in effectively arbitrary order while reporting success.

## What Changed

- `packages/core/src/modules/query_index/lib/engine.ts` — derive `hasCfSort` from `opts.sort` next to the existing `cfFilters` and add it as a fourth disjunct of `wantsCf`. A sort-only custom-field query now enters the guard block and falls back when the index cannot answer it. The test is deliberately `cf:`-only, not the `cf:`/`l10n:` pair used for filters: neither engine ever *applies* an `l10n:` ordering — both drop it while resolving sorts — so counting it would pay `entityHasActiveCustomFields` + `indexAnyRows` + the coverage probes, and possibly divert the whole request to the slower ORM engine, to produce an ordering that is then thrown away.
- `packages/shared/src/lib/query/engine.ts` — **make the fallback engine actually able to answer the query this PR now routes to it.** `BasicQueryEngine` builds its `cfKeys` set (which drives the `custom_field_defs`/`custom_field_values` joins and the `cfValueExprByKey` expression) from `opts.fields`, `cfFilters` and `includeCustomFields` — `opts.sort` was missing there too, the same omission one layer down. Because the `ORDER BY` was emitted outside the `if (expr)` guard, a sort-only `cf:` query produced `ORDER BY "cf_<key>"` against an alias the query never selected: Postgres `42703`, a 500 where the pre-PR behavior was a mis-ordered 200. Two changes fix it symmetrically with the one above: `cf:` sort keys now feed `cfKeys` (for every projection that actually sorts — the count shape is excluded because it never sorts), and the `orderBy` only fires for an alias the projection carries, so an unresolvable cf sort is **dropped** rather than emitted. Dropping is what the base-column branch already does when `resolveBaseColumn` returns null, so the two branches now agree.

  This was latent rather than live — every current caller that can produce a `cf:` sort also sets `includeCustomFields`, which populated `cfKeys` by a route that happened to work. This PR is what makes that path load-bearing, so it is fixed here rather than deferred.
- **Second, less obvious effect worth a reviewer's attention:** widening `wantsCf` also flips `shouldAttachCustomSources` (`engine.ts:499`) for these queries. A cf sort that also passed `customFieldSources` previously built its `ORDER BY` over the base `ei` alias alone, because the extra sources were never appended to `indexSources`; `buildCfTextExprSql` now coalesces across every declared source, which is what such a sort always meant. That is a latent correctness improvement riding along with this fix, not a regression — flagged here because it is not apparent from the diff.
- `hasCustomFieldFilters` / `needsIndexRowset` are **deliberately unchanged**. Those govern the count-query shape and exist to carry custom-field *predicates* into a correlated `EXISTS`; a sort contributes no predicate, so widening them would add a pointless `EXISTS` to every sorted count for no correctness gain.

## 🧪 Tests

Four regression tests in `packages/core/src/modules/query_index/__tests__/hybrid-engine.test.ts`:

- **`falls back when a cf sort is the only custom-field signal and no index rows exist`** — locks in the `indexAnyRows` fallback for a sort-only `cf:` query carrying no cf fields, filters, or `includeCustomFields`.
- **`falls back and warns on partial coverage when a cf sort is the only custom-field signal`** — locks in the coverage-gap path, asserting both the fallback and the `partialIndexWarning` metadata. It passes `skipAutoReindex: true` because `scheduleAutoReindex`'s debounce is process-global module state that a later existing test asserts on; without the opt-out this test consumed that slot and broke the neighbouring case through test ordering.
- **`keeps l10n-only sorts off the custom-field branch`** — a negative control for the `cf:`-only narrowing: an `l10n:` sort must not pay the probes or divert to the ORM engine for an ordering both engines discard.
- **`keeps base-column-only sorts off the custom-field branch`** — a negative control proving the widened predicate did not become over-broad; a plain `id` sort must still take the index path.

Two more in `packages/shared/src/lib/query/__tests__/engine.test.ts`, driving the real `BasicQueryEngine` through this repo's `createFakeKysely` recorder so they assert on the SQL actually built, not on a stubbed hand-off:

- **`a cf sort alone projects and joins the key it orders by`** — the positive case: options of exactly the shape the hybrid engine now forwards (`fields: ['id','email']`, a `cf:` sort, no `includeCustomFields`) must produce the `cf_vip` alias in both `orderBys` **and** `selects`, with the cf joins present. Before the fix this recorded an `orderBy` with zero joins and no matching select — the `42703`.
- **`a cf sort that resolves to no definition is dropped, not ordered by`** — the drop path: an unresolvable key yields no `orderBy` at all.

All three production changes were **mutation-checked**, each mutation failing exactly one test and no others: removing the `cfKeys` sort feed fails the positive shared test; restoring the unguarded `orderBy` fails the drop test; widening `hasCfSort` back to include `l10n:` fails the new l10n negative control. The three tests added for the review findings are load-bearing, not decorative.

**Validation gate — run in full locally this time.** An earlier revision of this description noted that most of the gate had been left to CI because the worktree had no `node_modules` of its own. That is no longer true: dependencies were installed in the worktree and the complete `validation.commands` sequence was executed in order, in local (non-Docker) mode.

- ✅ `yarn build:packages` (×2), `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn build:app` — all exit 0. `yarn generate` produced **no** changes to committed files.
- ✅ **`yarn test`** for the two packages this PR touches, run standalone: `@open-mercato/core` **11,684 passed** / 1,453 suites, `@open-mercato/shared` **2,042 passed** / 185 suites. The two `catalog products/variants route enricher config` failures reported on the earlier revision are gone — they were caused by absent `#generated/*` artifacts and disappear once `yarn generate` + `yarn build:packages` have run.
- ⚠️ The repo-wide `yarn test` task exits 1 on **`@open-mercato/cli`** — 5 failures in `resolveEnvironment › falls back to standalone mode with rootDir = cwd` and `resolver enterprise module toggle › …`. These are environment/location-dependent (they depend on where the checkout sits on disk) and are unrelated to this diff, which touches no CLI or module-resolution code. Turbo aborts the remaining tasks on that failure, which is why `core` and `shared` were run directly.

## 💥 Breaking Changes

None to any contract surface — no exported API, HTTP route, response shape, event id, DB schema, or DI change.

There is one **intended behavior change**: a list route already sorting by `cf_*` against a partially-indexed entity will now fall back to `BasicQueryEngine` (slower, and surfacing `meta.partialIndexWarning`) and may schedule an auto-reindex, where previously it silently returned mis-ordered rows from the index. That is the guard working as designed.

## Follow-up (not addressed here)

- **#5674 — `cf:` sorts order as text regardless of field kind, with no stable tiebreak.** The reporter's second observation is a real but **separate** defect this PR neither causes nor cures: both engines order `cf:` sorts **as text** regardless of the field's `kind` (`buildCfTextExprSql` yields jsonb `->>`, which is always text; the ORM fallback casts every kind to `::text`), so a numeric custom field sorts lexicographically — `'10240' < '1024' < '2048' < '512'`. Neither engine appends a stable `id` tiebreak either, leaving paginated results with ties or NULLs unstable across pages. Fixing that means kind-aware sort expressions in both engines and is well out of scope for this guard fix.
- **#5706 — `Mutation tests` reports green on diffs it never mutates.** `scripts/stryker/scope.mjs` freezes the mutation allowlist to `shared`, so this `core`-only diff resolved to an empty matrix and the `mutate` job reported `skipping`, not a passing mutation run. Filed from review feedback so the check's green is not read as coverage of the new predicate; the predicate was mutation-tested by hand instead (see 🧪 Tests).

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5670"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790340448&installation_model_id=432243&pr_number=5670&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5670&signature=bf504153cf7bd93421c39bed7cb233fd853e06b70538185e92de44bb199a6987"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/query_index/__tests__/hybrid-engine.test.ts
+++ b/packages/core/src/modules/query_index/__tests__/hybrid-engine.test.ts
@@ -420,6 +420,26 @@ describe('HybridQueryEngine', () => {
     expect((mockLogger.warn.mock.calls[0] || [])[0]).toContain('Partial index coverage')
   })
 
+  test('keeps l10n-only sorts off the custom-field branch', async () => {
+    const db = createFakeKysely({ baseTable: 'todos', hasIndexAny: false, baseCount: 5, indexCount: 0 })
+    const em = buildEm(db)
+    const fallback = { query: jest.fn() }
+    const emitEvent = jest.fn().mockResolvedValue(undefined)
+    const engine = new HybridQueryEngine(em, fallback as any, () => ({ emitEvent }))
+
+    // Neither engine ever applies an `l10n:` ordering — both drop it while
+    // resolving sorts — so it must not pay the coverage probes or divert the
+    // query to the ORM engine for an ordering that is thrown away.
+    await engine.query('example:todo', {
+      fields: ['id'],
+      sort: [{ field: 'l10n:title', dir: SortDir.Asc }],
+      organizationId: 'org1',
+      tenantId: 't1',
+    })
+
+    expect(fallback.query).not.toHaveBeenCalled()
+  })
+
   test('keeps base-column-only sorts off the custom-field branch', async () => {
     const db = createFakeKysely({ baseTable: 'todos', hasIndexAny: false, baseCount: 5, indexCount: 0 })
     const em = buildEm(db)
--- a/packages/core/src/modules/query_index/__tests__/hybrid-engine.test.ts
+++ b/packages/core/src/modules/query_index/__tests__/hybrid-engine.test.ts
@@ -376,6 +376,67 @@ describe('HybridQueryEngine', () => {
     expect(emitEvent).not.toHaveBeenCalled()
   })
 
+  test('falls back when a cf sort is the only custom-field signal and no index rows exist', async () => {
+    const db = createFakeKysely({ baseTable: 'todos', hasIndexAny: false, baseCount: 5, indexCount: 0 })
+    const em = buildEm(db)
+    const fallback = { query: jest.fn().mockResolvedValue({ items: [], page: 1, pageSize: 20, total: 0 }) }
+    const emitEvent = jest.fn().mockResolvedValue(undefined)
+    const engine = new HybridQueryEngine(em, fallback as any, () => ({ emitEvent }))
+
+    const opts = {
+      fields: ['id'],
+      sort: [{ field: 'cf:priority', dir: SortDir.Asc }],
+      organizationId: 'org1',
+      tenantId: 't1',
+    }
+    await engine.query('example:todo', opts)
+
+    expect(fallback.query).toHaveBeenCalledWith('example:todo', opts)
+  })
+
+  test('falls back and warns on partial coverage when a cf sort is the only custom-field signal', async () => {
+    process.env.FORCE_QUERY_INDEX_ON_PARTIAL_INDEXES = 'false'
+    const db = createFakeKysely({ baseTable: 'todos', hasIndexAny: true, baseCount: 10, indexCount: 1 })
+    const em = buildEm(db)
+    const fallback = { query: jest.fn().mockResolvedValue({ items: [], page: 1, pageSize: 20, total: 0 }) }
+    const emitEvent = jest.fn().mockResolvedValue(undefined)
+    const engine = new HybridQueryEngine(em, fallback as any, () => ({ emitEvent }))
+    mockLogger.warn.mockClear()
+
+    // The auto-reindex debounce is process-global; opting out keeps this test
+    // from consuming the slot a later test asserts on.
+    const result = await engine.query('example:todo', {
+      fields: ['id'],
+      sort: [{ field: 'cf:priority', dir: SortDir.Asc }],
+      organizationId: 'org1',
+      tenantId: 't1',
+      skipAutoReindex: true,
+    })
+
+    expect(fallback.query).toHaveBeenCalled()
+    expect(result.meta?.partialIndexWarning).toEqual(expect.objectContaining({
+      entity: 'example:todo', baseCount: 10, indexedCount: 1,
+    }))
+    expect((mockLogger.warn.mock.calls[0] || [])[0]).toContain('Partial index coverage')
+  })
+
+  test('keeps base-column-only sorts off the custom-field branch', async () => {
+    const db = createFakeKysely({ baseTable: 'todos', hasIndexAny: false, baseCount: 5, indexCount: 0 })
+    const em = buildEm(db)
+    const fallback = { query: jest.fn() }
+    const emitEvent = jest.fn().mockResolvedValue(undefined)
+    const engine = new HybridQueryEngine(em, fallback as any, () => ({ emitEvent }))
+
+    await engine.query('example:todo', {
+      fields: ['id'],
+      sort: [{ field: 'id', dir: SortDir.Desc }],
+      organizationId: 'org1',
+      tenantId: 't1',
+    })
+
+    expect(fallback.query).not.toHaveBeenCalled()
+  })
+
   test('falls back and warns on partial coverage', async () => {
     process.env.FORCE_QUERY_INDEX_ON_PARTIAL_INDEXES = 'false'
     const db = createFakeKysely({ baseTable: 'todos', hasIndexAny: true, baseCount: 10, indexCount: 1 })
--- a/packages/core/src/modules/query_index/lib/engine.ts
+++ b/packages/core/src/modules/query_index/lib/engine.ts
@@ -352,10 +352,19 @@ export class HybridQueryEngine implements QueryEngine {
 
       const normalizedFilters = normalizeFilters(opts.filters)
       const cfFilters = normalizedFilters.filter((filter) => filter.field.startsWith('cf:') || filter.field.startsWith('l10n:'))
+      // `applySort` orders `cf:` sorts off the left-joined index doc, so a
+      // sort-only custom-field query depends on the index just as much as a cf
+      // filter does — without this disjunct it would skip the coverage guards
+      // below and silently order by a NULL expression on an empty index (#5521).
+      const cfSorts = (opts.sort || []).filter((sort) => {
+        const field = String(sort.field)
+        return field.startsWith('cf:') || field.startsWith('l10n:')
+      })
       const coverageScope = this.resolveCoverageSnapshotScope(opts)
       const wantsCf = (
         (opts.fields || []).some((field) => typeof field === 'string' && (field.startsWith('cf:') || field.startsWith('l10n:'))) ||
         cfFilters.length > 0 ||
+        cfSorts.length > 0 ||
         (Array.isArray(opts.includeCustomFields) && opts.includeCustomFields.length > 0)
       )
 
--- a/packages/core/src/modules/query_index/lib/engine.ts
+++ b/packages/core/src/modules/query_index/lib/engine.ts
@@ -356,15 +356,15 @@ export class HybridQueryEngine implements QueryEngine {
       // sort-only custom-field query depends on the index just as much as a cf
       // filter does — without this disjunct it would skip the coverage guards
       // below and silently order by a NULL expression on an empty index (#5521).
-      const cfSorts = (opts.sort || []).filter((sort) => {
-        const field = String(sort.field)
-        return field.startsWith('cf:') || field.startsWith('l10n:')
-      })
+      // Only `cf:` qualifies: neither engine ever applies an `l10n:` ordering —
+      // both drop it while resolving sorts — so counting it here would pay the
+      // coverage probes and the ORM diversion for an ordering that is discarded.
+      const hasCfSort = (opts.sort || []).some((sort) => String(sort.field).startsWith('cf:'))
       const coverageScope = this.resolveCoverageSnapshotScope(opts)
       const wantsCf = (
         (opts.fields || []).some((field) => typeof field === 'string' && (field.startsWith('cf:') || field.startsWith('l10n:'))) ||
         cfFilters.length > 0 ||
-        cfSorts.length > 0 ||
+        hasCfSort ||
         (Array.isArray(opts.includeCustomFields) && opts.includeCustomFields.length > 0)
       )
 
--- a/packages/shared/src/lib/query/__tests__/engine.test.ts
+++ b/packages/shared/src/lib/query/__tests__/engine.test.ts
@@ -244,6 +244,12 @@ function createFakeKysely(overrides?: FakeData) {
   return db
 }
 
+// Selects are recorded verbatim, so a plain column arrives as a string while an
+// aggregated cf projection arrives as Kysely's aliased raw builder.
+function selectAliases(call: any): string[] {
+  return call._ops.selects.map((s: any) => (typeof s === 'string' ? s : String(s?.alias ?? '')))
+}
+
 describe('BasicQueryEngine (Kysely)', () => {
   test('pluralizes entity names ending with y correctly', async () => {
     const fakeDb = createFakeKysely()
@@ -285,6 +291,39 @@ describe('BasicQueryEngine (Kysely)', () => {
     expect(hasCfOrder).toBe(true)
     const hasExtJoin = baseCall._ops.joins.length > 0
     expect(hasExtJoin).toBe(true)
+  })
+
+  test('a cf sort alone projects and joins the key it orders by', async () => {
+    const fakeDb = createFakeKysely()
+    const engine = new BasicQueryEngine({} as any, () => fakeDb as any)
+    await engine.query('auth:user', {
+      fields: ['id', 'email'],
+      sort: [{ field: 'cf:vip', dir: SortDir.Asc }],
+      organizationId: '1',
+      tenantId: 't1',
+    })
+    const baseCall = fakeDb._calls.find((b: any) => b._ops.table === 'users')
+    expect(baseCall._ops.orderBys).toContainEqual(['cf_vip', 'asc'])
+    // Ordering by an alias the query never selected is a Postgres 42703, so the
+    // sort has to bring its own projection and joins along (#5521).
+    expect(selectAliases(baseCall)).toContain('cf_vip')
+    expect(baseCall._ops.joins.length).toBeGreaterThan(0)
+  })
+
+  test('a cf sort that resolves to no definition is dropped, not ordered by', async () => {
+    const fakeDb = createFakeKysely()
+    const engine = new BasicQueryEngine({} as any, () => fakeDb as any)
+    await engine.query('auth:user', {
+      fields: ['id', 'email'],
+      sort: [{ field: 'cf:no_such_key', dir: SortDir.Asc }],
+      organizationId: '1',
+      tenantId: 't1',
+    })
+    const baseCall = fakeDb._calls.find((b: any) => b._ops.table === 'users')
+    // Dropping an unresolvable sort is what the base-column branch already does;
+    // the alternative here was an ORDER BY over a column that is never selected.
+    expect(baseCall._ops.orderBys).toEqual([])
+    expect(selectAliases(baseCall)).not.toContain('cf_no_such_key')
   })
 
   test('customFieldSources join additional profiles for custom fields', async () => {
--- a/packages/shared/src/lib/query/engine.ts
+++ b/packages/shared/src/lib/query/engine.ts
@@ -863,6 +863,16 @@ export class BasicQueryEngine implements QueryEngine {
       for (const f of cfFilters) {
         if (typeof f.field === 'string' && f.field.startsWith('cf:')) cfKeys.add(f.field.slice(3))
       }
+      // A `cf:` sort needs the same joins and value expression as a `cf:` field or
+      // filter: the sort loop below orders by an aggregated alias that only exists
+      // once the key reached `cfKeys`. Without this, a sort-only custom-field query
+      // emitted `ORDER BY "cf_<key>"` against a column it never selected (#5521).
+      // The count shape never sorts, so it stays out of this.
+      if (!isCountProjection) {
+        for (const s of (opts.sort || [])) {
+          if (typeof s.field === 'string' && s.field.startsWith('cf:')) cfKeys.add(s.field.slice(3))
+        }
+      }
       if (projection === 'full' && opts.includeCustomFields === true) {
         if (entityIdToSource.size > 0) {
           const entityIdList = Array.from(entityIdToSource.keys())
@@ -1202,7 +1212,13 @@ export class BasicQueryEngine implements QueryEngine {
               cfSelectedAliases.push(alias)
             }
           }
-          if (!requiresPlaintextSort) q = q.orderBy(alias, (s.dir ?? 'asc') as any)
+          // Only order by an alias the projection actually carries. A key that
+          // resolved to no expression is dropped rather than emitted as an
+          // unselected alias, which is what the base-column branch above already
+          // does when `resolveBaseColumn` returns null.
+          if (!requiresPlaintextSort && cfSelectedAliases.includes(alias)) {
+            q = q.orderBy(alias, (s.dir ?? 'asc') as any)
+          }
         } else {
           if (!requiresPlaintextSort) q = q.orderBy(qualify(s.field), (s.dir ?? 'asc') as any)
         }


adeptofvoltron:
## 🏷️ Label rationale — `om-open-pr`

- 🔍 **`review`** — The PR is open, complete and ready for review, so it enters the pipeline in the `review` state. Pipeline labels are mutually exclusive and no other state applies: it is not blocked, not awaiting changes, and not in the merge queue.
- 🐛 **`bug`** — This fixes a defect reported in #5521: a sort-only custom-field query bypassed the query index's fallback and coverage guards and returned silently mis-ordered rows. It adds no new capability, so `feature` and `refactor` do not apply.
- ✅ **`skip-qa`** — Applied under the automated-verification exemption in `AGENTS.md`, every condition of which this change meets: it touches no UI-rendering file (the diff is one `.ts` engine file plus its `__tests__` sibling — no `.tsx`, nothing under `packages/ui/src/`, nothing under `**/components/**`), the database structure and API surface are unchanged, no `BACKWARD_COMPATIBILITY.md` contract surface is broken, and automated regression tests for the changed behavior ship in this same PR. There is no user-facing surface a human could meaningfully exercise in a browser, so `needs-qa` would buy nothing; the two are never combined.
- 🟡 **`priority-medium`** — The bug silently returns wrong result ordering rather than failing loudly, which makes it easy to ship on top of, and it is reachable from any list endpoint accepting `?sortField=cf_<key>`. That argues for urgency. Against `priority-high`: nothing crashes or loses data, and callers have a workaround (adding a `cf:` field or filter to the same request restores the guards, and the reporter fell back to client-side ordering).
- 🟠 **`risk-medium`** — The edit itself is six lines widening one boolean predicate, but it sits in `HybridQueryEngine.query()`, the hot path behind every indexed list query, and it deliberately changes behavior: routes already sorting by `cf_*` against a partially-indexed entity will now fall back to `BasicQueryEngine` (slower, surfacing `meta.partialIndexWarning`) and may schedule an auto-reindex. A negative-control test guards against the predicate becoming over-broad, which is what keeps this out of `risk-high`; the blast radius is nonetheless wider than `risk-low` would imply.

Priority and risk are orthogonal here and were assessed independently: the defect's urgency is moderate because it is silent and has a workaround, while the blast radius is moderate because the touched function is central to list querying.

adeptofvoltron:
🤖 `om-open-pr` — chain hand-off: this PR now carries the autofix chain's `in-progress` lock, transferred from #5521 so the work is never observably unclaimed between steps. The next step in the chain is `om-auto-review-pr`, which will re-enter this lock rather than re-acquiring it. Other auto-skills should skip this PR until the lock is released.

adeptofvoltron:
🤖 `om-auto-review-pr` re-entering the chain lock inherited from `om-open-pr` on this PR. Reviewing now; the lock is retained rather than re-acquired, and will be released by the driving `om-auto-fix-issue` run.

adeptofvoltron (commented):
# 🔍 Code Review

**Verdict: no blocking findings.** The change is correct, minimal, well-tested, and does exactly what #5521 asked for. One nit and two follow-up observations below; none of them block the merge.

**⚠️ This review is posted as a comment, not an approval.** The automation account (`adeptofvoltron`) authored this PR, and GitHub does not permit self-approval — so this PR still requires an approving review from a human maintainer before it can merge. The pipeline label stays `review` rather than moving to `merge-queue`, because "no blocking findings from the automated pass" is not the same as "approved".

## ✅ Correctness

The root cause and the fix line up exactly. `applySort` (`packages/core/src/modules/query_index/lib/engine.ts:1139`) orders `cf:` sorts off `buildCfTextExprSql(...)`, which reads the `entity_indexes` alias `ei`. Because `applyEntityIndexesJoin` (`engine.ts:752`) is a **LEFT** join, a missing index row yields `NULL` for every row rather than raising — so the old `wantsCf` predicate let a sort-only custom-field query slip past `entityHasActiveCustomFields`, the `indexAnyRows` fallback, and the coverage-gap check, and return silently mis-ordered rows. Adding `cfSorts.length > 0` as a fourth disjunct (`engine.ts:356-365`) closes exactly that hole, using the same prefix test as the adjacent `cfFilters`.

I verified that widening `wantsCf` does **not** leak into behavior beyond the intended guards. There are only two other consumers:

- `engine.ts:499` — `shouldAttachCustomSources`. This one is worth calling out explicitly, because it is an **unadvertised second fix**: a cf sort that also passed `customFieldSources` previously built its `ORDER BY` over the base `ei` alias only, since the extra sources were never appended to `indexSources`. With `wantsCf` now true, `buildCfTextExprSql` coalesces across every declared source, which is what such a sort always meant. This is a latent correctness improvement, not a regression — but it is not currently mentioned in the PR body and a reviewer would not spot it from the diff alone.
- `engine.ts:685-691` — the forced-partial-index path, additionally gated on `isForcePartialIndexEnabled()` and `entityHasActiveCustomFields`, and squarely inside the guard region this fix is meant to reach.

The decision to leave `hasCustomFieldFilters` / `needsIndexRowset` (`engine.ts:996`, `engine.ts:1031`) untouched is correct and worth keeping. Those exist to carry custom-field *predicates* into a correlated `EXISTS` for the count query; a sort contributes no predicate, so widening them would have bought nothing and added an `EXISTS` to every sorted count.

## 🧪 Tests

Coverage is genuinely adequate rather than nominal. Both positive tests were confirmed to fail against the unpatched engine and pass with the fix, which is the property that actually matters — a regression test that passes either way is worthless. The negative control (`keeps base-column-only sorts off the custom-field branch`) is the right instinct: it pins down that the widened predicate did not become over-broad, which is the plausible way this fix could have gone wrong.

The `skipAutoReindex: true` in the partial-coverage test is correctly reasoned. `scheduleAutoReindex` (`engine.ts:2322-2326`) debounces through `markAutoReindexScheduled`, module-level state shared across the whole file, so without the opt-out the new test consumed the debounce slot that the pre-existing `falls back and warns on partial coverage` case asserts on. Fixing that by opting out — rather than by reordering tests or weakening the older assertion — leaves both tests order-independent.

## 📝 Findings

**Nit — `l10n:` in `cfSorts` is inert today** (`engine.ts:359-362`). The `l10n:` half of the new prefix test cannot currently affect ordering in either direction. In the hybrid engine, `resolvedSorts` (`engine.ts:713-722`) preserves only `cf:`-prefixed sort fields; an `l10n:` sort falls through to `resolveBaseColumn`, returns `null`, and is dropped before `applySort` ever sees it. The ORM fallback (`packages/shared/src/lib/query/engine.ts`) has no `l10n:` handling at all. So the only reachable effect of an `l10n:` sort is now to force a fallback that still will not sort by it. I would **keep it as written**: it matches the two adjacent predicates and the reporter's suggested patch, the branch is unreachable in practice, and omitting it would silently recreate this exact bug the day `l10n:` sort support lands in `applySort`. Flagging it so the asymmetry is a recorded decision rather than an accident.

**Follow-up (pre-existing, not introduced here) — cf sorts bypass the encrypted-field paths.** `requiresPlaintextSort` is computed from `resolvedSorts.filter((sort) => !sort.field.startsWith('cf:'))` (`engine.ts:723-730`), so `cf:` sorts are deliberately excluded from the plaintext-sort route, and the `cfKeys` collection at `engine.ts:2059-2063` gathers keys from fields and filters but not from sorts. An encrypted custom field sorted through the index therefore orders by ciphertext. This is the same family as the reporter's own second observation and is out of scope here, but it belongs in the same follow-up rather than being lost.

**Follow-up (already documented in the PR body) — text-vs-kind ordering.** Both engines order `cf:` sorts as text regardless of the field's `kind`, so numeric custom fields sort lexicographically (`'10240' < '1024' < '2048' < '512'`), with no stable `id` tiebreak. Correctly scoped out of this PR; it needs kind-aware sort expressions in both engines.

## 💥 Breaking changes

None to any contract surface — no exported API, HTTP route, response shape, event id, DB schema, DI key, or ACL feature is touched, so nothing in `BACKWARD_COMPATIBILITY.md` applies.

The one intended behavior change is correctly identified in the PR body: a route already sorting by `cf_*` against a partially-indexed entity now falls back to `BasicQueryEngine` (slower, surfacing `meta.partialIndexWarning`) and may schedule an auto-reindex, where it previously returned mis-ordered rows from the index. Trading silent wrongness for visible slowness is the right call for a data-correctness guard, and the `partialIndexWarning` makes the cost observable rather than mysterious.

## 🧰 Validation

**CI is fully green on head `ed8a905`** — `test` (7m37s), `lint`, `prepare`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance`, `ephemeral-integration (none)` and `license/cla` all pass, with no failing and no pending required checks. That matters more than usual here: the PR body honestly disclosed that `build:packages`, `generate`, the i18n checks, repo-wide `typecheck` and `build:app` could not be run soundly in the authoring worktree (its `node_modules` was symlinked to a parent checkout on a different branch, so `@open-mercato/*` resolved to the wrong source). CI has now executed that gate against the real branch and it passed, which closes the one open risk the author flagged. Disclosing the limitation instead of claiming a green local gate was the right call.

`mergeable: MERGEABLE` — no conflicts against `develop`.

adeptofvoltron:
🤖 `om-auto-review-pr` — completed: **no blocking findings** (1 nit, 2 follow-up observations, all non-blocking). CI is green on `ed8a905` across `test`, `lint`, `prepare`, `ds-lint` and `ephemeral-integration`, and the branch is `MERGEABLE` against `develop`.

The autofix loop did not run because there was nothing actionable to fix — it engages only on a `changes_requested` verdict.

Two notes on state, so nothing here is overstated:

- The review was submitted as a **comment, not an approval**: the automation account authored this PR and GitHub forbids self-approval. This PR therefore still needs an approving review from a human maintainer. The pipeline label deliberately stays `review` rather than moving to `merge-queue`.
- **Lock retained — chain continues.** The `in-progress` label is an inherited chain lock and is not released here; the driving `om-auto-fix-issue` run releases it in its final step.

adeptofvoltron:
🤖 Follow-up filed as #5674, covering the text-vs-`kind` ordering and the missing stable tiebreak — plus the related finding that `cf:` sorts bypass the encrypted-field sort paths and order by ciphertext. Both were deliberately scoped out of #5670, which fixes only the `wantsCf` guard from this issue.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the chain fixed #5521, opened this PR, reviewed it with **no blocking findings**, and confirmed CI green on `ed8a905`. **Lock released** — this PR is free for anyone to pick up.

**It is not merge-ready yet, and deliberately so:** the automation account authored this PR, so GitHub forbids it from self-approving. This needs an approving review from a human maintainer. The pipeline label stays `review` rather than `merge-queue` for that reason — moving it would misrepresent readiness.

UI verification was not run and is not owed: the diff touches only `packages/core/src/modules/query_index/lib/engine.ts` and its test sibling — no `.tsx`, nothing under `packages/ui/src/` or `**/components/**`, no database or API surface change — which is exactly the automated-verification exemption `skip-qa` was applied under.

pkarw:
🤖 `om-auto-review-pr` — claiming this PR for review (assignee + `in-progress` label + this comment). The prior chain lock was explicitly released at 13:45Z, so this is a fresh claim, not a re-entry. Other auto-skills should skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review

**Verdict: changes requested** — 0 blockers, 0 majors, **1 minor**, 4 nits.

The diagnosis in this PR is correct and the fix is the right shape. `opts.sort` genuinely was missing from `wantsCf`, `applySort` genuinely does build a `cf:` `ORDER BY` off the **left**-joined `ei` alias (`packages/core/src/modules/query_index/lib/engine.ts:1143-1150`), and a LEFT join over an empty `entity_indexes` genuinely yields an all-`NULL` sort key rather than an error — so the failure really is silent by construction. Adding `cfSorts.length > 0` as a fourth disjunct is the minimal correct answer, and I confirmed the tests are not decorative (see below).

The one finding is not about the nine changed lines. It is about the destination they now route traffic to.

## 🟡 Minor

### 1. The ORM fallback answers a sort-only `cf:` query only because every current caller happens to also pass `includeCustomFields` — and the PR's own tests use the shape that doesn't

`packages/shared/src/lib/query/engine.ts:851-865`, `:1193-1206`

This PR's whole mechanism is "hand the query to `BasicQueryEngine`, which can order by a custom field." That is true, but not for the reason it looks like.

In the fallback engine, `cfKeys` — the set that decides which custom fields get their `custom_field_defs`/`custom_field_values` joins and a `cfValueExprByKey` entry — is populated from exactly three sources: `opts.fields` entries starting with `cf:` (`:858-862`, gated on `projection === 'full'`), `cfFilters` (`:863-865`), and `includeCustomFields` in either its `true` (`:866`) or array (`:963`) form. **`opts.sort` is not among them** — the same omission this PR is fixing one layer up, still present one layer down.

The sort loop then does this:

```ts
if (!cfSelectedAliases.includes(alias)) {
  const expr = cfValueExprByKey[key]
  if (expr) { q = q.select(sql`max(${expr})`.as(alias)); cfSelectedAliases.push(alias) }
}
if (!requiresPlaintextSort) q = q.orderBy(alias, (s.dir ?? 'asc') as any)   // :1205 — outside the guard
```

The `orderBy` is outside the `if (expr)` guard, so when the key never entered `cfKeys` the query emits `ORDER BY "cf_<key>"` against an alias it never selected.

I proved both halves against the real `BasicQueryEngine` using this repo's own `createFakeKysely` recorder in `packages/shared/src/lib/query/__tests__/engine.test.ts`, rather than arguing from reading:

- `{ fields: ['id','email'], sort: [{ field: 'cf:vip', dir: Asc }] }` → `orderBys = [["cf_vip","asc"]]`, `selects = ["id","email"]`, **`joins = 0`**. That compiles to `ORDER BY "cf_vip"` with no such output column — Postgres `42703 column "cf_vip" does not exist`. A 500, where before this PR the same request returned mis-ordered rows with a 200.
- The same options plus `includeCustomFields: true` → `orderBys = [["cf_vip","asc"]]`, `selects = ["id","email","cf_industry","cf_industry__is_multi","cf_vip","cf_vip__is_multi"]`, `joins = 4`. Correct, and the fix works exactly as advertised.

**This is not a live defect today, which is why it is a Minor and not a Blocker.** Every caller I could find that can produce a `cf:` sort also sets `includeCustomFields: true`: `makeCrudRoute` sets it unconditionally on every list request (`packages/shared/src/lib/crud/factory.ts:1825`), and the custom-entity records API — the only other place a user-supplied `sortField` reaches the engine directly — sets it too (`packages/core/src/modules/entities/api/records.ts:262`). The remaining direct `queryEngine.query(...)` call sites (`entities/api/relations/options.ts:92`, `customers/api/utils.ts:291`, `customers/api/deals/map/route.ts:405`, the `search` package indexers) sort by base columns or not at all.

So the guarantee holds — by accident, undocumented, and in a different package from the one this PR touches. What makes it worth raising now is that this PR is what makes that path load-bearing: before it, a sort-only `cf:` query never reached the fallback at all.

**The tests cannot see this, and it is worth saying why.** Both new positive tests pass options of exactly the failing shape — `{ fields: ['id'], sort: [{ field: 'cf:priority' }] }`, no `includeCustomFields`, no cf fields, no cf filters — and they pass only because `fallback` is a `jest.fn()` stub. They assert that the hand-off *happens*; nothing asserts that the hand-off *produces ordered rows*, which is the actual goal of the PR. That is a reasonable seam for a unit test, but it means the suite would stay green if the fallback could not answer the query at all — and for the exact options object the tests construct, it cannot.

Suggested fix, two lines in `packages/shared/src/lib/query/engine.ts` and symmetrical with what this PR does upstream:

1. Feed `cf:` sort fields into `cfKeys` beside the `fields`/`cfFilters` loops (`:858-865`), so the key gets its joins and expression.
2. Move `:1205`'s `orderBy(alias, …)` inside the `if (expr)` block, so a cf sort that still cannot be resolved is **dropped** rather than emitted as an unselected alias. Dropping is what the base-column branch already does (`:684` — `if (column) resolvedSorts.push(...)`), so this only makes the two branches agree.

Filing this as a follow-up rather than fixing it here is a defensible call — it is a pre-existing latent defect in a package this PR does not otherwise touch — and I would not argue with a maintainer who merges on that basis. Under this repo's verdict rule a minor is `changes-requested`, so I am recording it as one and leaving the disposition to you.

## 💭 Nits

1. **`cfSorts` accepts `l10n:`, but no engine sorts by `l10n:`.** The new filter (`engine.ts:359-362`) mirrors `cfFilters` on the line above, which is defensible symmetry — but neither engine ever applies an `l10n:` sort. `HybridQueryEngine`'s `resolvedSorts` only keeps `cf:` fields and base columns (`engine.ts:722-730`), and `BasicQueryEngine` does the same (`packages/shared/src/lib/query/engine.ts:679-686`); an `l10n:` sort field is silently discarded by both. Since `normalizeSortFieldSelector` passes unmapped selectors through verbatim (`factory.ts:99-102`), `?sortField=l10n:anything` now flips `wantsCf` on and pays `entityHasActiveCustomFields` + `indexAnyRows` + the coverage-gap probes — and can divert the whole request onto the slower ORM engine — to apply an ordering that will then be thrown away. Narrowing the disjunct to `cf:` would also make the code match its own new comment, which justifies the change purely in terms of how `applySort` handles `cf:`.

2. **`cfSorts` is materialized only to read `.length > 0`.** `(opts.sort || []).some((sort) => …)` expresses the same predicate without allocating an array on every query through the engine's hottest path. Cosmetic, but this is a per-request line.

3. **The `Mutation tests` green on this PR carries no information.** `scripts/stryker/scope.mjs:36` freezes `ALLOWLISTED_PACKAGES` to `['shared']`, so a `packages/core`-only diff resolves to an empty matrix and the `mutate` job reports `skipping`, not a passing mutation run. Not a fault of this PR — noting it so the check is not read as coverage of the new predicate. (I ran that check by hand instead; see below.)

4. **PR body wording.** It states that `wantsCf` derived from "a non-empty `opts.includeCustomFields`". Strictly, the existing disjunct is `Array.isArray(opts.includeCustomFields) && length > 0` (`engine.ts:368`) — the **boolean** `true` form does not count, and that is the form `makeCrudRoute` always passes (`factory.ts:1825`). This is worth stating precisely in the body because it explains why the bug was reachable from every CRUD list route in the first place, and because the asymmetry looks like a bug to the next reader. It is not one, and should stay: making the boolean form count would put the coverage probes and the fallback risk on *every* list query in the app, not just cf-sorted ones.

## ✅ Verified, and explicitly not findings

So nobody re-derives these:

- **The new tests genuinely pin the fix, in both directions.** I mutation-tested them rather than trusting the PR body. Replacing the new disjunct with `false` → the two positive tests fail (2 failed / 66 passed). Replacing it with the over-broad `(opts.sort || []).length > 0` → the negative control `keeps base-column-only sorts off the custom-field branch` fails (1 failed / 67 passed). The negative control is doing real work, not decoration.
- **Env isolation in the new partial-coverage test is fine.** `FORCE_QUERY_INDEX_ON_PARTIAL_INDEXES` is saved at `hybrid-engine.test.ts:275` and restored in `afterEach` at `:286-293`, so the assignment at `:441` cannot leak. The `skipAutoReindex: true` opt-out and its comment are correct — `scheduleAutoReindex`'s debounce really is process-global module state a later test asserts on.
- **The fallback is a real fix for the reachable case, not a no-op.** `BasicQueryEngine` does implement `cf:` ordering, aggregating over the joined `custom_field_values` (`packages/shared/src/lib/query/engine.ts:1193-1206`). With the factory's `includeCustomFields: true` the alias is projected and the join is present — confirmed empirically above.
- **`hasCustomFieldFilters` / `needsIndexRowset` were correctly left alone.** They exist to carry cf *predicates* into a correlated `EXISTS` on the count shape (`engine.ts:1005-1041`); a sort contributes no predicate, so widening them would add an `EXISTS` to every sorted count for nothing.
- **The `shouldAttachCustomSources` side effect (`engine.ts:499`) is correctly characterized and low-risk.** It only differs when `customFieldSources` is non-empty *and* `searchEnabled` is false — with search on, the `|| searchEnabled` arm already attached them. The resulting `coalesce` across declared sources is what such a sort always meant, and it is the same index-source shape the cf-*filter* path already builds. Good call flagging it in the body; it is not apparent from the diff.
- **The extra probes are not a new cost class.** `indexAnyRows` is uncached (`engine.ts:2238-2252`, a coverage lookup plus an `entity_indexes` existence probe), so a cf-sorted request now pays 1-2 extra round trips. That is exactly what every cf-*filtered* request already pays today, so it is symmetric rather than a regression — but it is a real per-request cost on list endpoints, worth knowing.
- **No contract surface is touched.** No exported signature, import path, HTTP route, response shape, event id, DB schema, DI key, ACL feature or generated file changes. `BACKWARD_COMPATIBILITY.md` is not engaged. The behavior change (fallback + `meta.partialIndexWarning` where previously a silently mis-ordered index result) is additive on an existing documented field and is disclosed in the body.
- **The branch is clean.** Merge base is `b41f7e3e5`, which *is* the current tip of `develop` — 0 commits behind, `MERGEABLE`, no conflicts. That is also why the integration matrix ran in affected-only mode.

## 🧪 Validation

**Validation source: GitHub checks.** All checks are settled and green on head `ed8a9052`; nothing is pending, so there is no CI disclosure to make. I opened the `prepare` (job `98182770995`) and `test` (job `98183935816`) **step** lists individually rather than trusting the job-level rollup, and every one of the eight configured `validation.commands` maps to a named green step:

| Gate command | CI step |
|---|---|
| `yarn build:packages` (×2) | `prepare` 12 *Build packages*, 14 *Rebuild packages with generated files* |
| `yarn generate` | `prepare` 13 *Prepare generated modules* |
| `yarn i18n:check-sync` | `test` 17 *Check i18n sync* |
| `yarn i18n:check-usage` | `test` 18 *Check i18n usage* |
| `yarn typecheck` | `test` 21 *Checking types* |
| `yarn test` | `test` 22 *Test* (+ 23 *Test scripts*, 24-26 the unfiltered guards) |
| `yarn build:app` | `prepare` 15 *Build app* |

`Install dependencies` is `skipped` in both jobs on a cache hit, and `Check agent instruction budget` (`test` 20) is green — so nothing downstream was skipped behind an early red step. `ephemeral-integration (none)` passed in 8m40s; `none` is this repo's affected-only shard name, not a skip. `audit`, `comment`, `mutate`, `merge-coverage`, `docker-build` and `[code]smith` are `skipping` by design.

No local gate run — the two probes above (the `BasicQueryEngine` recorder and the two-way mutation of the new disjunct) are targeted evidence for specific claims in this review, not a substitute for the gate, and both were reverted afterwards.

I read the PR body's honest account of what was and was not run locally, and it is accurate — CI has since covered every command it listed as unrun.

---

Fixing the minor, or filing it as a follow-up with a maintainer's sign-off, flips this to approve. Nothing in it disputes the nine lines this PR changes.

pkarw:
## 🏷️ Label rationale — `om-auto-review-pr`

- ✅ **`review` → `merge-queue`** — the re-review approves. The single Minor from the previous pass (the ORM fallback could not answer the sort-only `cf:` query this PR routes to it) is fixed in `b878b92`, every nit is fixed or filed, and only three fresh nits remain — two of which ask for a sentence in the description rather than a code change. Under this repository's stricter rule (Minor still requests changes) there is nothing left at or above Minor, so the PR moves into the merge queue.
- 🐛 **`bug`** unchanged — this fixes #5521, a silent mis-ordering defect, not a new capability.
- ⏭️ **`skip-qa`** unchanged — the automated-verification exemption still applies: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no database-structure or API-surface change, no `BACKWARD_COMPATIBILITY.md` contract touched, and automated tests for the changed behaviour ship in the same PR (six of them, across both engines).
- 🟡 **`priority-medium`** unchanged — reachable from every CRUD list route via `?sortField=cf_<key>`, but only on entities whose query index is empty or under-populated, and the symptom is wrong ordering rather than wrong or leaked data.
- 🟠 **`risk-medium`** unchanged — the blast radius grew slightly since the first labelling: the change now spans two packages (`query_index` and the shared `BasicQueryEngine`) and alters the SQL shape of a fallback path, but it stays inside the query engines and touches no contract surface. The full 15-shard integration matrix is green on this head.

Not applied: `needs-qa` (superseded by `skip-qa`, and the two must never coexist), `qa-approved` / `qa-self-verified` (never applied from reading a diff), and the `qa` pipeline label (owned by QA reviewers — this skill only ever requests QA with `needs-qa`).

pkarw:
👤 Handing this back to @adeptofvoltron — reassigned, pipeline label moved to `changes-requested`.

**One Minor, zero Blockers and zero Majors**, and the finding is not about the nine lines you changed. The `wantsCf` widening is correct, and I mutation-tested your three tests in both directions rather than taking the body's word for it: disabling the new disjunct fails both positive tests, and making it over-broad fails the negative control. That is a well-built test trio.

The Minor is that `BasicQueryEngine` — the fallback this PR now routes sort-only `cf:` queries to — has the *same* omission one layer down: `opts.sort` is not a source for its `cfKeys`, so it builds no join and no expression for the key, and then emits `ORDER BY "cf_<key>"` on an alias it never selected. I proved both sides against the real engine with this repo's own fake-Kysely recorder. It is not a live defect today only because every caller that can produce a `cf:` sort also passes `includeCustomFields: true` — `makeCrudRoute` unconditionally, and the custom-entity records API too. Notably, your two positive tests pass options of exactly the shape that would fail, and stay green only because `fallback.query` is a stub.

Two lines in `packages/shared/src/lib/query/engine.ts` close it, and they are symmetrical with what you did upstream. Filing it as a follow-up instead is a perfectly defensible call — it is a pre-existing latent defect in a package this PR does not otherwise touch — and I said so in the review. Either route flips this to approve.

Everything else in the review is nits and a list of things I checked and deliberately did **not** count against the PR, so you do not have to re-derive them.

pkarw:
🤖 `om-auto-review-pr` — completed: **CHANGES REQUESTED** (0 blockers, 0 majors, 1 minor, 4 nits). **Lock released** — this PR is free for anyone to pick up.

Validation source: **GitHub checks** — all settled green on `ed8a9052`, `prepare` and `test` step lists read individually, all eight configured gate commands mapped to named green steps. No `ci-monitoring`: there is nothing pending to watch.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-08-27T13:36:42Z.

adeptofvoltron:
## 🏷️ Label rationale — `om-auto-fix-pr`

- 🔄 **`changes-requested` → `review`** — the review that set `changes-requested` raised one Minor and four nits. The Minor is fixed in `b878b92`, nits 1 and 2 are fixed in `d073693`, nit 4 (PR-body precision) is applied to the description, and nit 3 is filed as #5706 because it is a repo-wide CI-tooling concern rather than anything this diff can fix. Nothing actionable from that review is outstanding, so `changes-requested` is no longer an honest description of the PR's state and it returns to `review`.
- 🚫 **not `merge-queue`** — stating this explicitly because it is the label a reader might expect next. `merge-queue` requires an approving review, and the standing review decision is still `CHANGES_REQUESTED` until @pkarw re-reviews. This automation account is also the PR author, so it cannot dismiss that review or approve in its place. The PR is *ready* for re-review, not approved.
- 🐛 **`bug`** — unchanged. The PR fixes a silent correctness defect: a list sorted by a custom field returned arbitrarily-ordered rows while reporting success.
- ⏭️ **`skip-qa`** — unchanged, and re-checked against the automated-verification exemption now that the diff also touches `packages/shared`. The change still renders no UI (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`), leaves the database structure and the API surface untouched, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships automated tests for every changed behavior. The exemption holds, so no manual QA pass is required.
- 🟡 **`priority-medium`** — unchanged. The bug is reachable from every CRUD list route via `?sortField=cf_<key>`, which argues for urgency, but it degrades ordering rather than corrupting or exposing data, which argues against treating it as high.
- 🟠 **`risk-medium`** — unchanged, and deliberately not raised even though the blast radius grew to `packages/shared`. The new `BasicQueryEngine` behavior is confined to queries that carry a `cf:` sort: such a query now gains the joins and projection it always needed, and an unresolvable cf sort is dropped instead of producing invalid SQL. Queries without a `cf:` sort build byte-identical SQL, so the wider file footprint does not widen the behavioral footprint.
- 🔄 **`in-progress` → `ci-monitoring`** — the run's work is finished and fully reported; only the CI outcome on the new head is still being watched. `ci-monitoring` is a meta label, not a claim, so this PR is free for a human or another agent to act on.

adeptofvoltron:

This run took PR #5670 from `changes-requested` back to a re-reviewable state. Everything @pkarw raised is now either fixed in this PR or tracked as its own issue, and the full validation gate has been executed locally rather than deferred to CI.

### 🔀 Base merge

`origin/develop` had advanced one commit past this PR's merge base (`b41f7e3e5` → `41ff100d9`, "keep every value of a repeated list query parameter"). It was merged into the branch first, cleanly and with no conflicts, so the review response, the local gate and CI all judge the real merge result rather than a stale diff. That merge touches `packages/shared/src/lib/crud/factory.ts`, which the review cites — the citations still hold, only their line numbers shift.

### 🔍 How each review finding was handled

@pkarw's review recorded 0 blockers, 0 majors, 1 minor and 4 nits. None were dismissed silently:

- **🟡 Minor — the ORM fallback could not actually answer a sort-only `cf:` query. Fixed (`b878b92`).** This was the right call to fix rather than defer. The review's own diagnosis was reproduced exactly: `BasicQueryEngine` populated `cfKeys` from `opts.fields`, `cfFilters` and `includeCustomFields` but never from `opts.sort`, and emitted the cf `ORDER BY` outside the `if (expr)` guard — so the precise options shape this PR now routes to the fallback produced `ORDER BY "cf_<key>"` against an alias the query never selected (Postgres `42703`). Both halves of the suggested fix are applied: `cf:` sort keys now feed `cfKeys` for every projection that actually sorts (the count shape is excluded, since it never sorts), and the `orderBy` fires only for an alias the projection carries, so an unresolvable cf sort is **dropped** — which is what the base-column branch already does when `resolveBaseColumn` returns `null`. The review offered a follow-up as an acceptable alternative; it was fixed in-loop instead because it is a correctness defect that *this* PR is what makes load-bearing, and deferring a correctness finding to an issue is not something this run does.
- **💭 Nit 1 — `cfSorts` accepted `l10n:`, but no engine sorts by `l10n:`. Fixed (`d073693`).** The predicate is now `cf:`-only. Confirmed the premise first: in both engines an `l10n:` sort falls through to the base-column resolver, which returns nothing for that selector, so the sort is discarded. Counting it meant paying `entityHasActiveCustomFields` + `indexAnyRows` + the coverage probes — and possibly diverting the request to the slower ORM engine — for an ordering that is then thrown away.
- **💭 Nit 2 — `cfSorts` was materialized only to read `.length > 0`. Fixed (`d073693`).** It is now `(opts.sort || []).some(...)`, with no per-query array allocation on the engine's hottest path.
- **💭 Nit 3 — a green `Mutation tests` check carries no information here. Filed as #5706.** This one is genuinely not fixable from this branch: `scripts/stryker/scope.mjs` freezes the mutation allowlist to `shared`, and its own comment says a package joins only after a timing run proves a PR-sized diff fits the workflow timeout. That is a repo-wide CI-tooling decision, so it is tracked rather than smuggled into a `query_index` bug fix. The issue asks for either a measured allowlist addition or an honest skipped/neutral status when the matrix is empty.
- **💭 Nit 4 — PR body imprecision about `includeCustomFields`. Fixed in the description.** The body now states the disjunct precisely (`Array.isArray(...) && length > 0` — the boolean `true` form does not count) and explains why that asymmetry should stay: making `true` count would put the coverage probes and the fallback on *every* list query in the app rather than only cf-sorted ones.

### 🧪 Tests and validation

Three tests were added for the findings above — two in `packages/shared/src/lib/query/__tests__/engine.test.ts` driving the real `BasicQueryEngine` through this repo's `createFakeKysely` recorder (so they assert on the SQL actually built, not on a stubbed hand-off), and one `l10n:` negative control in `hybrid-engine.test.ts`.

All three production changes were **mutation-checked**, each mutation failing exactly one test and no others: removing the `cfKeys` sort feed fails the positive shared test; restoring the unguarded `orderBy` fails the drop test; widening the predicate back to `l10n:` fails the new negative control. The tests are load-bearing, not decorative.

The full `validation.commands` gate ran locally in non-Docker mode. `yarn build:packages` (×2), `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck` and `yarn build:app` all exit 0, and `yarn generate` changed no committed files. `@open-mercato/core` (**11,684 passed**) and `@open-mercato/shared` (**2,042 passed**) are fully green. The repo-wide `yarn test` task exits 1 solely on **`@open-mercato/cli`** — 5 failures in `resolveEnvironment` and `resolver enterprise module toggle`, which are location-dependent environment failures unrelated to this diff (it touches no CLI or module-resolution code). Turbo aborts remaining tasks on that failure, which is why `core` and `shared` were run directly.

### 📸 UI

Not applicable — no user-facing surface in the diff. The change is confined to two query engines and their unit tests, renders nothing, and carries `skip-qa` under the automated-verification exemption (re-checked after the diff grew to `packages/shared`; it still holds).

### 🔁 Follow-ups

- **#5706** — `Mutation tests` reports green on diffs it never mutates (allowlist is `shared`-only). Filed from nit 3 this run.
- **#5674** — `cf:` sorts order as text regardless of field kind, with no stable tiebreak. Filed on the earlier chain run; still the correct home for that defect and untouched here.

### ⚠️ Merge-readiness

**Not merge-ready yet, and this run cannot make it so.** The work is complete and the gate is green, but the standing review decision is `CHANGES_REQUESTED` from @pkarw, and only they can dismiss or supersede it. This automation account authored the PR, so it can neither approve nor self-dismiss — that limit is the point of the two-pair-of-eyes rule, not an obstacle to route around. The pipeline label is therefore `review`, not `merge-queue`.

**Next step: @pkarw, a re-review.** Your review said fixing the minor, or filing it as a follow-up with a maintainer's sign-off, would flip this to approve — the minor is fixed in-loop, with tests. Once approved and green, `om-approve-merge-pr 5670` owns the merge; this skill never merges.

**CI is still running on this head (`d073693`).** Branch protection holds the actual merge; this run's verdict covers the work, not a green run. A follow-up comment will report the CI outcome.

adeptofvoltron:
## 🤖 `om-auto-fix-pr` — CI result

**✅ Every check on head `d073693` is green.** The run settled inside the wait budget with zero failures, so the local validation gate reported earlier is now corroborated by CI on the real merge result.

| Check | Result |
|---|---|
| `test` | ✅ |
| `prepare` | ✅ |
| `lint`, `ds-lint`, `scope`, `audit-scope` | ✅ |
| `documents-multi-instance` | ✅ |
| `ephemeral-integration` (1/15 … 15/15) | ✅ all 15 shards |
| `mutate (shared, src/lib/query/engine.ts)` | ✅ |
| `license/cla` | ✅ |
| `[code]smith`, `audit`, `comment`, `docker-build`, `merge-coverage` | ⏭️ skipped by design |

Two of these are worth more than a tick, because this head is measured differently from the previous one:

**The integration matrix ran in full, not affected-only.** The earlier head reported a single `ephemeral-integration (none)` shard — this repo's affected-only shard name — because the diff was confined to `packages/core/src/modules/query_index`. Now that the fix also touches `packages/shared`, the scope resolver fanned out to the complete 15-shard matrix, and all 15 passed. The change therefore carries materially broader integration coverage than the version that was reviewed.

**`mutate` did real work this time, which partly answers nit 3 for this head.** The review noted that a green `Mutation tests` check carried no information on a `core`-only diff, because `scripts/stryker/scope.mjs` allowlists `shared` only. That is precisely why the check now reports `mutate (shared, src/lib/query/engine.ts)`: the fallback-engine fix lands *inside* the allowlisted package, so Stryker actually mutated the new `cfKeys` sort feed and the guarded `orderBy` — and the suite killed the mutants. The hand-rolled mutation evidence in the run summary is now backed by the CI check itself. This does not close #5706: a future `core`-only diff would still report an uninformative green, which is what that issue is about.

**This does not make the PR mergeable, and the label set says so.** The pipeline label stays `review`, not `merge-queue`: the standing review decision is still `CHANGES_REQUESTED` from @pkarw and `mergeStateStatus` is `BLOCKED`. Green CI is one of the two gates; the second is a human re-review that this automation account cannot supply for its own PR. Every finding from that review is now fixed in-loop or tracked — @pkarw, over to you.

`ci-monitoring` is removed: this was the follow-up it promised, and no further follow-up will come from this agent.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-28T03:30:50Z. Re-review of the fixes pushed for the previous `changes-requested` verdict. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review

**Re-review — verdict: approve.** 0 blockers, 0 majors, **0 minors**, 3 nits.

Every finding from the previous pass is accounted for. The Minor is fixed in `b878b92`, nits 1 and 2 in `d073693`, nit 4 in the description, and nit 3 is filed as #5706. The fixes are the ones I suggested, and I checked that they are load-bearing rather than taking the commit messages at face value. The three nits below are all new, all non-blocking, and two of them ask for a sentence in the description rather than a change to the code.

## ✅ Disposition of the previous findings

| Previous finding | Status |
|---|---|
| 🟡 Minor — the ORM fallback could not answer a sort-only `cf:` query (`42703`) | **Fixed** in `b878b92`, both halves |
| 💭 Nit 1 — `cfSorts` accepted `l10n:`, which no engine ever orders by | **Fixed** in `d073693`, with a negative control |
| 💭 Nit 2 — array materialized only to read `.length` | **Fixed** — now `.some()` |
| 💭 Nit 3 — the green `Mutation tests` carried no information | **Filed** as #5706 (and see 🧪 below — it now runs, but still advisorily) |
| 💭 Nit 4 — PR body imprecise about the `includeCustomFields` disjunct | **Fixed** in the description, accurately |

**The Minor fix is correct, and for the reason it claims.** Both halves landed: `cf:` sort keys now feed `cfKeys` (`packages/shared/src/lib/query/engine.ts:866-875`) and the `ORDER BY` only fires for an alias the projection actually carries (`:1215-1221`).

The second half is the one worth double-checking, because `cfSelectedAliases` is written from *two* places and a guard reading it could easily have regressed the case that already worked. It does not: the alias is pushed both by the main custom-field projection (`:1053` — the `fields` / `includeCustomFields` / `requestedCustomFieldKeys` path) and by the sort loop itself (`:1212`). So a `cf:` sort on a key that is also projected — the shape `makeCrudRoute` produces on every list request — still emits exactly the `ORDER BY` it did before, and only a key that resolved to no expression at all is now dropped.

**Mutation evidence, from CI rather than from reading.** Because this revision touches `packages/shared`, the file is now inside Stryker's allowlist and `mutate (shared, src/lib/query/engine.ts)` ran for real. I pulled the per-mutant log for job `98540433289` instead of trusting the check rollup: **zero mutants survived at `engine.ts:1219-1220`**, so the new `a cf sort that resolves to no definition is dropped, not ordered by` test genuinely kills the restored-unguarded-`orderBy` mutation. That is the same check I did by hand last round for the hybrid predicate, now done by the toolchain.

## 💭 Nits

### 1. The two guard conditions inside the new `cfKeys` feed survive mutation

`packages/shared/src/lib/query/engine.ts:871`, `:873`

The same mutation run that killed the `orderBy` guard reports three survivors on the lines just above it:

- `:871` — `if (!isCountProjection)` → `if (true)`
- `:873` — `if (typeof s.field === 'string' && s.field.startsWith('cf:'))` → `if (true)`, and `&&` → `||`

Both survive for understandable reasons, and neither is a defect:

- Dropping the count guard lets the count shape add the sort key to `cfKeys`. The projection loop `continue`s on the count shape at `:1005`, so no join or expression is built — but `cfKeys` also drives the unresolved-key lookup at `:976-988`, so the mutant costs one wider `custom_field_defs` read per count. The guard is buying something real; it is just not something a test asserts.
- Dropping the `cf:` prefix test lets any sort field through as `field.slice(3)`, which never matches a `keySource` entry and is discarded downstream. That mutant is close to equivalent.

I would not add a test per mutant here. Recording it so nobody reads the (now genuinely running) `mutate` check as evidence that these two lines are pinned — the behaviour on either side of them is pinned; the guards themselves are not.

### 2. The fix repairs a second silent no-op that the description does not mention

`packages/shared/src/lib/query/engine.ts:871` and the `sortKeys` projection

The new feed is gated only on `!isCountProjection`, so it applies to `projection: 'sortKeys'` as well as `'full'`. That is a behaviour change on the plaintext-sort path worth a line in the body, because it is not visible from the diff.

Before this commit, `cfKeys` on the `sortKeys` projection was fed by `cfFilters` alone — the `opts.fields` feed is gated on `projection === 'full'` (`:858`), and both `includeCustomFields` feeds are gated on `full` / `!isSortKeysProjection` (`:876`, `:973`). So for a **mixed** sort — an encrypted base column plus a `cf:` key, which is exactly when `requiresPlaintextSort` turns on — the phase-1 candidate scan never selected a `cf_<key>` alias, and `sortRowsInMemory` (`:1352`) read `undefined` for every row through `readField` (`packages/shared/src/lib/query/encrypted-sort.ts:63-68`). The custom-field leg of the sort was silently discarded and rows came back ordered by the encrypted column alone.

Now the sort loop selects `max(<cf expr>) as cf_<key>` in that projection too (`:1210-1212`), and `fieldNameCandidates('cf:<key>')` includes the sanitized `cf_<key>` (`encrypted-sort.ts:19-25`), so the in-memory comparator finds the value and the cf leg participates.

That is a real ordering change, in the right direction, for a shape nobody was thinking about — the same class of rider as the `shouldAttachCustomSources` effect the body already flags, and it deserves the same courtesy of a sentence. It is also untested; I would not block on a test for it, but the body should not stay silent about it.

### 3. Nothing anywhere executes the new SQL shape

`packages/shared/src/lib/query/__tests__/engine.test.ts:296-327`

Both new shared tests assert against this repo's `createFakeKysely` recorder. That was the right call for what they pin — which selects, joins and `orderBy`s the builder receives — and it is strictly better evidence than the stubbed hand-off the round-1 tests had. But a recorder cannot reject invalid SQL, and the shape produced for a sort-only `cf:` query is new: two `LEFT JOIN`s, a `GROUP BY <table>.id` this query never had before (`hasJoinedAggregates` is now true because `cfValueExprByKey` is non-empty — `:1229-1235`), and an `ORDER BY` on an aggregate output alias.

**I do not think it is broken, and that is why this is a nit and not a finding.** Grouping by the table's primary key lets Postgres accept the ungrouped base columns by functional dependency; `ORDER BY` against an output column name is standard; and every individual element already ships in production today via the custom-field *field* path, which builds the same joins and the same `GROUP BY`. The only genuinely novel piece is that the ordered alias is `max(<text expr>)` rather than the jsonb projection.

What is missing is that nothing in the repository executes a `?sortField=cf_*` request against a real database — the only occurrence of that string outside this PR is `packages/shared/src/lib/crud/__tests__/crud-factory.test.ts`, which covers the selector rewrite and not the query. This repo does have the fixtures for it (`TC-CRM-CF-MULTI-EDIT-001`, `TC-SALES-041-document-create-custom-fields` and friends create custom-field definitions and values) and runs a 15-shard integration suite against Postgres. A single spec that defines a cf key, writes a few values and asserts the order of a list route would convert this PR's reasoning into a fact. A follow-up issue is a perfectly reasonable home for it.

## 📌 One observation for #5674

While tracing the alias, a third ordering inconsistency turned up that belongs with the two that issue already names. The expression a `cf:` sort orders by depends on whether the same key is also *projected*:

- projected — the alias is the jsonb `CASE ... to_jsonb(...)` expression (`:1047-1051`), so `ORDER BY` runs under jsonb comparison rules (arrays sort after scalars, and so on);
- not projected — the alias is `max(<text expr>)` (`:1211`), plain text ordering.

Through `makeCrudRoute` it is always the first, because `includeCustomFields: true` is unconditional, so the divergence is only reachable from direct engine callers. Not this PR's problem — it predates it and this PR does not widen the reachable set — but worth adding to #5674, which is already the home for "what does a `cf:` sort actually order by".

## ✅ Verified, and explicitly not findings

So nobody re-derives these on the next pass:

- **Narrowing to `cf:` is correct, not a coverage loss.** Both engines discard an `l10n:` sort while resolving sorts — `packages/core/src/modules/query_index/lib/engine.ts:722-731` and `packages/shared/src/lib/query/engine.ts:679-687` keep only `cf:` fields and resolvable base columns. Nothing downstream can act on an `l10n:` ordering, so paying the coverage probes for one was pure cost.
- **The new l10n negative control is discriminating, not decorative.** The fixture sets `hasIndexAny: false`, and the `!hasIndexRows` branch (`core engine.ts:390-394`) returns through `fallback.query` *before* `entityHasActiveCustomFields` can matter — so a predicate widened back to accept `l10n:` fails the `not.toHaveBeenCalled()` assertion. It also cannot disturb the process-global `scheduleAutoReindex` debounce the way the partial-coverage test could, because `wantsCf` stays false and the guard block never runs; the absent `skipAutoReindex: true` is correct here rather than an oversight.
- **The `.some()` rewrite is behaviour-identical** and removes a per-request allocation on the engine's hottest path, which is what the nit asked for.
- **The `GROUP BY` newly introduced for these queries is not a new combination.** Selecting non-aggregated base columns alongside `GROUP BY <table>.id` already happens on every request that projects a custom field; this change only makes a sort-only query join that set.
- **The base merge is clean and still current enough.** `57382040` merged `develop`, the head is `MERGEABLE`, and the five commits that have landed on `develop` since (onboarding signup disclosure, the CRM activities lookup, the auth ACL fix, the portal slug and the portal translations) touch none of `packages/shared/src/lib/query`, `packages/core/src/modules/query_index`, or `packages/shared/src/lib/crud`.
- **No contract surface is touched.** No exported signature, import path, HTTP route, response shape, event id, DB schema, DI key, ACL feature or generated file. `BACKWARD_COMPATIBILITY.md` is not engaged. The one intended behaviour change (fallback plus `meta.partialIndexWarning` where an index previously returned mis-ordered rows) is additive on an existing documented field and is disclosed in the body.

## 🧪 Validation

**Validation source: GitHub checks.** Every check is settled and green on head `d073693c`. I confirmed the run's `head_sha` is that commit before reading it (run `33078849569`, `conclusion: success`) rather than inferring green from an absent failure: `prepare`, `test`, `lint`, `ds-lint`, `audit-scope`, `documents-multi-instance`, `license/cla`, and all **15** `ephemeral-integration` shards — a full matrix this time, not the affected-only `(none)` shard, because the base merge pulled the branch forward. `audit`, `comment`, `docker-build`, `merge-coverage` and `[code]smith` are `skipping` by design.

`mutate (shared, src/lib/query/engine.ts)` ran for real (12m39s) and is the reason I could evidence the `orderBy` guard from CI. Read its *green* carefully, though: the run scored **40.88%** against a 70% threshold and passes only because `MUTATION_ENFORCE` is not `"true"` (`scripts/stryker/enforce.mjs:40,90`). That is documented, intentional behaviour of `mutation-tests.yml`, not a defect — but it means the check being green still says nothing on its own. Its value here was the per-mutant log, which I read directly.

No local gate run: CI covers all eight `validation.commands`, and the description's account of the local run that preceded the push is consistent with what CI reproduced.

---

Approving. None of the three nits changes a line of production code — two ask for a sentence in the description, one asks for a follow-up integration spec — so nothing here should hold up the merge.

pkarw:
🤖 `om-auto-review-pr` — completed: **APPROVED** (0 blockers, 0 majors, 0 minors, 3 nits). **Lock released** — this PR is free for anyone to act on.

Re-review of `d073693c`. The Minor from the previous pass is fixed in `b878b92` and I verified the fix rather than taking the commit message for it: `cfSelectedAliases` is written from both the projection path (`shared/src/lib/query/engine.ts:1053`) and the sort loop (`:1212`), so the new `orderBy` guard cannot regress the already-working projected-alias case, and the `mutate (shared, src/lib/query/engine.ts)` job — which now runs for real because the diff reaches `packages/shared` — left **zero survivors at `engine.ts:1219-1220`**, so the new drop test is load-bearing. Nits 1, 2 and 4 are fixed and nit 3 is filed as #5706.

The three remaining nits are non-blocking and none of them touches production code: two mutation survivors on the new `cfKeys` guards, an undisclosed (beneficial) ordering change to the encrypted-sort candidate scan that deserves a line in the description, and the absence of any test that executes the new SQL shape against a real Postgres. I worked through why that SQL is valid in the review body, which is why it stayed a nit.

Validation source: **GitHub checks** — all settled green on head `d073693c` (run `33078849569`, `head_sha` verified, `conclusion: success`), including the full 15-shard `ephemeral-integration` matrix. No CI follow-up is owed, so no `ci-monitoring` label. No local gate run.

Autofix: not applicable — the verdict is approve, and this is not my PR (no `--autofix` was requested).

Labels: `review` → `merge-queue`; `bug`, `skip-qa`, `priority-medium`, `risk-medium` preserved. `skip-qa` means the QA gate does not block, so this is merge-ready as far as review is concerned.

pkarw:
## 📸 UI QA evidence — ⚠️ PARTIAL

**Verdict:** ⚠️ **PARTIAL** — the surface a user actually clicks is verified and unregressed, but on the degraded-index scenario this PR exists to fix, a custom-field sort came back with no effect at all.
**Environment:** `http://127.0.0.1:5001` · role `admin` · browser `playwright` (production build, disposable pgvector Postgres, started by this run)
**Verified:** `fix/issue-5521-wantscf-ignores-sort` @ `d073693`

This PR is backend-only (two query-engine files plus their unit tests) and carries
`skip-qa`, so there is no UI diff to inspect. This run therefore verified the closest
observable surfaces: the backend list that lets a user sort by a custom-field column,
and the CRUD list API the change actually governs.

### Scenario (P0/P1 — custom-field sorting on CRUD lists)

**Where to click:** `/backend/todos` (sortable **Priority** column) · `/api/customers/deals?sortField=cf_estimated_seats`

| 1 | Sign in as admin, cold-load `/backend/todos` | List renders with a Priority (custom-field) column | Rendered, HTTP 200, zero console and zero page errors | ✅ |
| 2 | Click **Priority** to sort ascending (index fully populated) | Request travels as `sortField=cf_priority`, rows ordered ascending | 200; rendered column `[2,3,3,4,4,5]` | ✅ |
| 3 | Click **Priority** again (descending) | Exact reverse of ascending | `[5,4,4,3,3,2]` — direction-sensitive, not incidental | ✅ |
| 4 | Mobile viewport 390×844 | Sorted list stays readable | Rendered without layout breakage | ✅ |
| 5 | Cache-busted direct API read, index populated | cf values present, cf sort correct, `sortField=title` control correct | cf asc `[2,3,3,4,4,5]`, desc `[5,4,4,3,3,2]`, title control alphabetical — confirms step 2–3 are not a cached artifact | ✅ |
| 6 | Empty `entity_indexes` for `example:todo`, bust cache, re-read | Guard diverts to the ORM engine, rows still ordered | 200, but **all cf cells null and asc == desc**. A `sortField=title` control on the same state also returned null cf values, so the empty cells belong to the empty-index state, not the cf-sort path | ❌ |
| 7 | Baseline `/api/customers/deals` sorted by `cf_estimated_seats`, index populated | Rows order by the custom field; toggle reverses | asc seats `[12,120,28,28,40,8]` (text ordering — see #5674), desc its exact reverse, both 200 | ✅ |
| 8 | **Empty the index for `customers:customer_deal`** (the precondition #5521 describes) and re-read, cache-busted | Per the PR description, the query falls back to the ORM engine and no longer returns mis-ordered rows | 200, but **ascending, descending and the unsorted control all returned the identical row order** — the custom-field sort had no effect. Reproduced with a second field, `cf_competitive_risk`, also inert in both directions | ❌ |
| 9 | Control on that same empty index: sort by base column `createdAt` asc/desc | If base sorts still reverse, the inert ordering is specific to custom fields | `createdAt` asc and desc returned **different** orders — base-column sorting still works while both cf sorts stayed inert | ✅ |

### Why the deals route is the meaningful one

`/backend/todos` is a good regression check but **does not exercise this PR's new
`hasCfSort` disjunct**: its list `fields` already contain `cf:` selectors
(`apps/mercato/src/modules/example/api/todos/route.ts:63`), so `wantsCf`'s *first*
disjunct was already true before this change.

`/api/customers/deals` does exercise it — its list `fields` are base columns only
(`packages/core/src/modules/customers/api/deals/route.ts`), custom-field values arrive
via `decorateCustomFields`, and `normalizeSortFieldSelector`
(`packages/shared/src/lib/crud/factory.ts:92`) accepts `?sortField=cf_<key>` on any
list route. That is exactly the query shape the PR describes: no `cf:` field, no cf
filter, boolean `includeCustomFields`, and a `cf:` sort.

On that route, with the index emptied, the sort is **silently dropped** — consistent
with the new `cfSelectedAliases.includes(alias)` guard in
`packages/shared/src/lib/query/engine.ts` dropping a cf sort whose alias the
projection does not carry. The user-visible outcome of #5521 — "I sorted by a custom
field and the order is not what I asked for" — still reproduces in the empty-index
case, now as *no ordering* rather than *wrong ordering*.

### Notes for QA and reviewers

- **No crashes anywhere.** Every request across every probe — populated, partial and
  empty index, both directions, plus six guarded `PUT` writes — returned HTTP 200. The
  Postgres `42703` failure class the shared-engine half of this PR addresses did not
  occur in any state I could reach.
- **No regression on the normal path.** With the index populated — the production-normal
  state — custom-field sorting works correctly in both the UI and the API.
- **Pre-PR behavior was not observed directly.** Reproducing it needs a separate build of
  the base branch, which this run did not do. So this report cannot say whether step 8 is
  a regression or an unchanged pre-existing symptom; it is reported against the PR's own
  stated goal, and a reviewer who knows the intended semantics should judge it.
- **A caching pitfall worth knowing.** The CRUD list response cache initially masked
  several degraded-index probes, because deleting index rows via SQL does not invalidate
  it. Every degraded-index result above was re-taken with a unique cache-busting query
  parameter after a cache-invalidating write; earlier, contaminated readings were
  discarded.
- **Environment hygiene.** The disposable test database was deliberately mutated (index
  rows deleted; six todos re-saved with their own values) to reach the degraded states.
  No source file was modified, nothing was committed, and nothing was pushed.
- The `partialIndexWarning` the PR describes surfaces as an `x-om-partial-index`
  **response header**, not in the JSON body (`packages/shared/src/lib/crud/factory.ts:1569`).
  It was not observed on any probe here.

### 📸 Screenshots

pkarw:

This change ships no browser-level test — the four new cases in
`packages/core/src/modules/query_index/__tests__/hybrid-engine.test.ts` and the two in
`packages/shared/src/lib/query/__tests__/engine.test.ts` are unit tests over mocked or
recorder-backed engines. The QA above was manual; the scenario below locks in the
behaviour at the HTTP layer, where the regression is actually observable
(run `/om-integration-tests`).

1. **Setup:** via API fixtures, create one tenant/org, a custom field definition on a
   list-backed entity whose CRUD route selects **base columns only** (the deals route is
   the existing example — its `fields` carry no `cf:` selectors), and 4–6 records with
   distinct values for that field. Reindex so the query index is fully populated.
2. **Act (baseline):** `GET /api/<module>/<entity>?sortField=cf_<key>&sortDir=asc`, then
   the same with `sortDir=desc`.
3. **Assert (baseline):** both return 200; the descending id order is the exact reverse of
   the ascending one; the ordering differs from a `sortField=<base column>` control.
4. **Act (the regression):** delete the entity's rows from `entity_indexes` — the empty
   index state #5521 describes — and repeat step 2 with a cache-busting query parameter
   (the CRUD list response cache is **not** invalidated by direct index mutation, which
   silently masked several readings during this QA run).
5. **Assert (the regression):** the request still returns 200 **and** the ascending order
   still matches the baseline ascending order from step 3 — i.e. the fallback orders the
   rows rather than dropping the sort. Add the negative control too: a base-column sort on
   the same empty index must still reverse with direction, proving any inert ordering is
   specific to the custom-field path.
6. **Teardown:** delete every record, custom field definition and index row the test created.

Step 5 is the assertion that matters: it is the one this QA run could not get to pass, and
a unit test over a mocked engine cannot cover it because the behaviour depends on the SQL
the fallback actually emits.

pkarw:
🤖 `om-auto-qa-pr` completed: **PARTIAL**. Evidence posted above (screenshots render inline; full artifacts, including the throwaway Playwright specs, are under `.ai/qa/artifacts_20260828-033300-qa5670/`).

Custom-field sorting is verified working and unregressed on the production-normal path — the populated-index case — in both the UI and the API. The run stops short of a pass because on the empty-index scenario this PR targets, a `cf_` sort on the deals list returned the same row order ascending, descending and unsorted, while a base-column sort on that same state still reversed correctly. Pre-PR behaviour was not observed directly, so this is reported against the PR's stated goal rather than as a confirmed regression — a reviewer who knows the intended semantics should make that call.

**Lock note:** this QA run did **not** open the `in-progress` lock on this PR — it was already held by a concurrent `om-auto-review-pr` run under the same account, started at 2026-08-28T03:30:50Z. The label and assignees are therefore left untouched for that run to release. No pipeline or meta labels were changed by this run (evidence-only default; the PR also carries `skip-qa`).

PR #1729, opened 2026-04-27
fix(auth): harden reset origin checks and require password confirmation

MStaniaszek1998:
## Summary

Fix the security email origin validation for reverse-proxied auth reset requests and add password confirmation to the public reset form.

## Changes

- restore request URL origin validation while still allowing internal loopback request URLs when an allowed forwarded host is present
- add regression coverage for spoofed host headers and proxied internal auth reset requests
- require password confirmation on the public reset page and update reset integration specs to fill the new field

## Specification

**Does a spec exist for this feature/module?**
- [ ] Yes
- [ ] No (created a new spec)
- [x] N/A (minor change, no spec needed)

**Spec file path:**

N/A

## Testing

- `corepack yarn workspace @open-mercato/shared test -- src/lib/__tests__/url.test.ts`
- `corepack yarn workspace @open-mercato/core test -- src/modules/auth/api/__tests__/reset.test.ts`
- `corepack yarn workspace @open-mercato/core run typecheck` *(failed in this environment: `command not found: tsc`)*

## Checklist

- [x] This pull request targets `develop`.
- [ ] I have read and accept the Open Mercato Contributor License Agreement (see `apps/docs/cla.md`).
- [x] I updated documentation, locales, or generators if the change requires it.
- [x] I added or adjusted tests that cover the change.
- [x] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required).
- [ ] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable).

### Design System Compliance
- [x] No hardcoded status colors (`text-red-*`, `bg-green-*`, `text-emerald-*`, `bg-amber-*`, `bg-blue-*`) — use semantic tokens
- [x] No arbitrary text sizes (`text-[Npx]`) — use typography scale or `text-overline`
- [ ] Empty state handled for list/data pages (`<EmptyState>` or DataTable `emptyState` prop)
- [ ] Loading state handled for async pages
- [ ] `aria-label` on all icon-only buttons (`<IconButton>`)
- [x] Uses existing DS components (Alert, StatusBadge, FormField) — no custom replacements

## Linked issues

None.

PR #1523, opened 2026-04-15
fix(auth): prevent host header poisoning in reset links (carry-forward #1268)

pkarw:
Supersedes #1268

Credit: original implementation by @WXYZx. This follow-up PR carries that work forward with merge conflict resolution so it can merge without waiting on the original fork branch.

## Included work
- Original changes from #1268 — host header poisoning fix for auth security email links
- Follow-up fixes applied during re-review (all previous Medium/Low findings addressed by the author)
- Merge conflict resolution against `develop` (token hashing from #1465)

## Summary
- Adds `assertAllowedAppOrigin`, `getSecurityEmailBaseUrl`, `toSecurityEmailUrl`, `mapSecurityEmailUrlError` to `@open-mercato/shared/lib/url`
- Wires origin validation into password reset (`reset.ts`) and invite resend (`resend-invite/route.ts`) API routes
- Replaces `resolveInviteBaseUrl` wrapper with direct `getSecurityEmailBaseUrl` calls
- Adds `APP_ALLOWED_ORIGINS` env var for configuring additional trusted origins
- Wraps `sendEmail` in reset route with try/catch to prevent email failures from leaking account existence

## Test plan
- [x] `url.test.ts` — 6 tests (origin allowlist, Host header validation, production fail-closed, dev fallback)
- [x] `reset.test.ts` — 4 tests (APP_URL usage, poisoned host rejection, missing config, email failure resilience)
- [x] `resend-invite.route.test.ts` — 10 tests including poisoned host rejection
- [x] Typecheck passes for `@open-mercato/shared` and auth-related files in `@open-mercato/core`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5671, opened 2026-08-26
fix(auth): resolve the resend-invite email origin from the request, not its URL string

### LINKED ISSUE #5382: fix(auth): resend-invite passes req.url (string) to getSecurityEmailBaseUrl, breaking origin check behind reverse proxy
## Summary

`POST /api/auth/users/resend-invite` fails with **400 `Invalid request origin`** behind any reverse proxy / container deployment, while the create-user and password-reset invite paths on the same deployment work fine.

Root cause is a single argument: the route passes `req.url` (a **string**) to `getSecurityEmailBaseUrl`, whereas the working reset path passes `req` (a **`Request`**). The string branch of `readRequestOriginCandidates` returns before any header is read, so `x-forwarded-host` is never considered and only the container-internal origin survives.

## Affected code (verified against `main`)

- `packages/core/src/modules/auth/api/users/resend-invite/route.ts:135` — `base = getSecurityEmailBaseUrl(req.url)` ← **string**
- `packages/core/src/modules/auth/api/reset.ts:45` — `resetUrlTemplate = toSecurityEmailUrl(req, '/reset/__token__')` ← `Request`, works
- `packages/core/src/modules/auth/commands/users.ts:470` — `getSecurityEmailBaseUrl()` ← `undefined`, works (skips the check entirely, see note below)
- `packages/shared/src/lib/url.ts:105` — the early return

## Why the string path cannot pass

In `readRequestOriginCandidates` (`packages/shared/src/lib/url.ts`):

```ts
candidates.urlOrigin = normalizeOrigin(parsedUrl.origin)
if (typeof input === 'string') return candidates   // ← returns before reading headers
// ...
const forwardedHostOrigin = originFromHost(protocol, input.headers.get('x-forwarded-host'))
```

With a string input, `headerOrigins` stays empty. Then in `assertAllowedAppOrigin`:

- `hasAllowedHeaderOrigin === false` (nothing was collected)
- `urlOrigin` is the container-internal origin, not in `allowedOrigins`
- `shouldAllowLoopbackOrigin` → in production requires the *allowed* origin to also be loopback; a real public `APP_URL` is not → `false`
- the rescue branch `isLoopbackOrigin(urlOrigin) && hasAllowedHeaderOrigin` → false because headers were never read

→ `AppOriginRejectedError` → `mapSecurityEmailUrlError` → 400 `Invalid request origin`.

Behind a proxy, `req.url` reflects the internal listen address (e.g. `PORT`/`CONTAINER_PORT` from the shipped `Dockerfile`), never the public host — so this endpoint is broken in **every** proxied deployment, not just one configuration.

## Reproduction

1. Deploy with the shipped Dockerfile behind a TLS-terminating reverse proxy, container listening on a non-default port (`CONTAINER_PORT=6789`).
2. `APP_URL=https://app.example.com`, `NODE_ENV=production`.
3. Create a user → invite email sends correctly. Trigger password reset → sends correctly.
4. Click **Resend invite** on that user → 400 `Invalid request origin`.

Observed log:

```json
{"level":40,"name":"shared","component":"origin-check",
 "requestUrl":"https://localhost:6789/api/auth/users/resend-invite",
 "rejectedOrigin":"https://localhost:6789",
 "allowedOrigins":["https://app.example.com"],
 "msg":"Origin check rejected string input"}
```

`"Origin check rejected string input"` is emitted only from the `typeof input === 'string'` branch of `logOriginDebugContext`, which pins the call site precisely.

## Suggested fix

Pass the request object, matching `reset.ts`:

```diff
-    base = getSecurityEmailBaseUrl(req.url)
+    base = getSecurityEmailBaseUrl(req)
```

This preserves the host-header-poisoning protection from #1267/#1268/#1523 — `getSecurityEmailBaseUrl` still returns the configured `APP_URL` for the email body and still validates the origin. It only lets the check see the forwarded headers it was designed to read.

Two things maintainers may want to consider beyond the one-liner:

1. **The string overload is a footgun.** Any caller passing a string silently loses proxy-header awareness. Options: drop `string` from `RequestInput`, or have the string branch fail closed with an explicit "cannot validate origin from a URL string" error instead of quietly degrading to a URL-only check.
2. **`getSecurityEmailBaseUrl()` with no argument skips validation entirely** (`if (!urlOrigin) return`) — that is why `commands/users.ts:470` works. Worth confirming that is intentional, since it means the no-arg call has no origin protection at all while the string call has partial protection.

## Workaround for affected deployments

Add the internal origin to the allowlist (`readAllowedOrigins` merges `APP_ALLOWED_ORIGINS` as CSV):

```
APP_ALLOWED_ORIGINS=https://localhost:6789
```

Email links stay correct because `getSecurityEmailBaseUrl` returns the configured `APP_URL`, not the request origin. Caveat: this loosens the origin check for every route that uses it, so it is a stopgap rather than a fix.

## Environment

- `@open-mercato/core` 0.6.6 — code paths above re-verified unchanged on `main`
- Deployment: Docker, `CONTAINER_PORT=6789`, TLS-terminating reverse proxy, `NODE_ENV=production`

pidubiel:
## Summary

`POST /api/auth/users/resend-invite` returned **400 `Invalid request origin`** behind any reverse proxy, while the create-user and password-reset invite paths on the same deployment worked fine.

The route passed `req.url` (a **string**) to `getSecurityEmailBaseUrl`, whereas the working reset path passes `req` (a **`Request`**). `readRequestOriginCandidates` returns early on a string input (`packages/shared/src/lib/url.ts:105`) before any header is read, so `x-forwarded-host` was never considered and only the container-internal origin survived the check.

This made the endpoint unusable in every proxied deployment, not just one configuration.

Fixes #5382

## Changes

- `packages/core/src/modules/auth/api/users/resend-invite/route.ts` — pass `req` instead of `req.url`, matching `api/reset.ts`. Host-header-poisoning protection from #1267/#1268/#1523 is unchanged: `getSecurityEmailBaseUrl` still returns the configured `APP_URL` for the email body and still validates the origin — it only lets the check see the forwarded headers it was designed to read.
- `packages/core/src/modules/auth/api/__tests__/resend-invite.route.test.ts` — regression test reproducing the issue's deployment shape (container on an internal port, `x-forwarded-host` matching `APP_URL`, `NODE_ENV=production`). Verified to fail with `400` before the fix and pass with `200` after.
- `packages/shared/src/lib/__tests__/url.test.ts` — pins the underlying asymmetry: the same request passes `assertAllowedAppOrigin` as a `Request` and is rejected as a URL string.

This is the only string call site in the repository — every other caller of `getSecurityEmailBaseUrl` / `toSecurityEmailUrl` already passes a `Request` or no argument.

## Not addressed here

The issue also raises that the `string` branch of `RequestInput` is a footgun (any caller passing a string silently loses proxy-header awareness) and that the no-argument call skips origin validation entirely. Both change a shared helper's public contract, so they are left for a maintainer decision rather than folded into this fix. The new `url.test.ts` case documents the current behaviour so the trap is at least visible.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

## Testing

Run locally (Node, non-Docker):

- `yarn build:packages` — 25/25 successful
- `yarn generate` — clean
- `npx tsc --noEmit` in `packages/core` — 0 errors
- `npx tsc --noEmit` in `packages/shared` — 0 errors
- `npx jest src/modules/auth/api/__tests__/resend-invite.route.test.ts` in `packages/core` — 14/14 passed
- `npx jest src/lib/__tests__/url.test.ts` in `packages/shared` — 22/22 passed

The regression test was verified in both directions: reverting the one-line fix makes it fail with exactly the `400 Invalid request origin` the issue reports.

## Checklist

- [x] I updated documentation, locales, or generators if the change requires it. (No user-facing strings, generated files, or contract surfaces touched.)
- [x] I added or updated integration tests in `.ai/qa/tests/` (or documented why integration coverage is not required). — Not required: the defect is in argument passing on a single API route and is fully covered by the route-level regression test; reproducing it end to end would need a reverse-proxy deployment fixture.
- [x] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — N/A, one-line bug fix.
- [ ] QA routing set.

> I do not have `triage` permission, so I cannot apply labels. Suggested: `bug`, `priority-medium`, `risk-low`, `skip-qa` — the change touches no UI-rendering file, leaves the database structure and API surface unchanged, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships automated tests for the behavior it changes, which is the automated-verification exemption in `AGENTS.md` → PR Workflow.

### Design System Compliance

Not applicable — no UI files touched.

## Linked issues

Fixes #5382

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5671"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790341412&installation_model_id=432243&pr_number=5671&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5671&signature=ae4d3b361e797e3203e23a01af2c69f43e729a0b3d97b4fab63be0b900ed20ce"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/auth/api/__tests__/resend-invite.route.test.ts
+++ b/packages/core/src/modules/auth/api/__tests__/resend-invite.route.test.ts
@@ -314,4 +314,28 @@ describe('POST /api/auth/users/resend-invite', () => {
     expect(mockFlush).toHaveBeenCalledTimes(1)
     expect(mockSendEmail).toHaveBeenCalledTimes(1)
   })
+
+  test('sends the invite behind a reverse proxy whose forwarded host matches APP_URL', async () => {
+    process.env = {
+      ...process.env,
+      APP_URL: 'https://app.example.com',
+      NODE_ENV: 'production',
+      JWT_SECRET: 'test-jwt-secret',
+    }
+
+    const res = await POST(makeRequest(
+      { id: userId },
+      'https://localhost:6789/api/auth/users/resend-invite',
+      {
+        host: 'app.example.com',
+        'x-forwarded-host': 'app.example.com',
+        'x-forwarded-proto': 'https',
+      },
+    ))
+
+    expect(res.status).toBe(200)
+    expect(mockNativeUpdate).toHaveBeenCalledTimes(1)
+    expect(mockFlush).toHaveBeenCalledTimes(1)
+    expect(mockSendEmail).toHaveBeenCalledTimes(1)
+  })
 })
--- a/packages/core/src/modules/auth/api/users/resend-invite/route.ts
+++ b/packages/core/src/modules/auth/api/users/resend-invite/route.ts
@@ -139,7 +139,7 @@ export async function POST(req: Request) {
 
   let base: string
   try {
-    base = getSecurityEmailBaseUrl(req.url)
+    base = getSecurityEmailBaseUrl(req)
   } catch (error) {
     const mapped = mapSecurityEmailUrlError(error, {
       scope: 'auth.users.resend-invite',
--- a/packages/shared/src/lib/__tests__/url.test.ts
+++ b/packages/shared/src/lib/__tests__/url.test.ts
@@ -89,6 +89,23 @@ describe('security email URL helpers', () => {
     expect(() => assertAllowedAppOrigin(request, env)).not.toThrow()
   })
 
+  test('cannot see forwarded headers when the request is passed as a URL string', () => {
+    const env = {
+      APP_URL: 'https://app.example.com',
+      NODE_ENV: 'production',
+    }
+    const request = new Request('https://localhost:6789/api/auth/users/resend-invite', {
+      headers: {
+        host: 'app.example.com',
+        'x-forwarded-host': 'app.example.com',
+        'x-forwarded-proto': 'https',
+      },
+    })
+
+    expect(() => assertAllowedAppOrigin(request, env)).not.toThrow()
+    expect(() => assertAllowedAppOrigin(request.url, env)).toThrow(AppOriginRejectedError)
+  })
+
   test('allows loopback origin mismatches outside production', () => {
     const env = {
       APP_URL: 'http://localhost:3000',


pkarw:
🤖 `om-auto-review-pr` claimed this PR for automated review at 2026-08-27T00:01Z. Assignee + `in-progress` set; the lock is released when the review is posted.

pkarw (approved):
# 🔍 Code Review: fix(auth): resolve the resend-invite email origin from the request, not its URL string

## 🎯 Summary

This PR changes one argument in `packages/core/src/modules/auth/api/users/resend-invite/route.ts:142`, from `getSecurityEmailBaseUrl(req.url)` to `getSecurityEmailBaseUrl(req)`, and ships two tests that pin the behaviour. I reviewed the whole diff — the route change, `packages/core/src/modules/auth/api/__tests__/resend-invite.route.test.ts`, and `packages/shared/src/lib/__tests__/url.test.ts` — against the shared helper in `packages/shared/src/lib/url.ts`, the sibling call sites, and issue #5382.

The diagnosis in the PR body is correct and I verified each load-bearing claim independently rather than taking it on trust:

- **The string branch really does skip the headers.** `readRequestOriginCandidates` returns at `packages/shared/src/lib/url.ts:105` before either `host` or `x-forwarded-host` is read, so with a string input `headerOrigins` stays empty. `assertAllowedAppOrigin` then has nothing to rescue the request with: `hasAllowedHeaderOrigin` is `false`, so the `isLoopbackOrigin(urlOrigin) && hasAllowedHeaderOrigin` branch at `packages/shared/src/lib/url.ts:227` cannot fire, and in production `shouldAllowLoopbackOrigin` needs the *allowed* origin to be loopback too — which a real public `APP_URL` is not. The request is rejected with `AppOriginRejectedError` → 400 `Invalid request origin`.
- **The reported request shape is exactly what Next.js produces, not a contrived one.** `next/dist/server/next-server.js:1267` builds the App Router request URL as `` `${protocol}://${this.fetchHostname}:${this.port}${req.url}` `` whenever hostname and port are known — the server's own listen address, never the `Host` header. That is why the issue's log shows `requestUrl: "https://localhost:6789/..."` on a deployment whose public origin is `https://app.example.com`, and it is why the new route test pairs a loopback request URL with a public `host`/`x-forwarded-host` pair. The combination is real, not synthetic.
- **This is genuinely the only string call site.** Grepping every caller of `getSecurityEmailBaseUrl` / `toSecurityEmailUrl` / `assertAllowedAppOrigin` across `packages/`, `apps/` and `external/` leaves `customer_accounts/api/signup.ts:95`, `auth/lib/requestRedirect.ts:16`, `auth/api/reset.ts:46`, `onboarding/.../verify-base-url.ts:63`, `onboarding/.../post/onboarding.ts:149` and `onboarding/.../status.ts:67` all passing a `Request`, and `auth/commands/users.ts:492`, `onboarding/.../ready-email.ts:28` and `documents/.../entityRegistry.server.ts:147` passing nothing. The resend-invite route was the sole outlier.

The change is also the direction the repository has already written down for itself. The recorded lesson [_"Hydrated backend chrome payloads must receive the original request for scope-aware RBAC"_](../blob/develop/.ai/lessons/hydrated-backend-chrome-payloads-must-receive-the.md) states the rule directly: *any server helper whose result can depend on cookies, forwarded headers, or query params must receive the original `Request`*. Passing `req.url` was precisely the failure that rule describes.

Security posture is unchanged for the worse. `getSecurityEmailBaseUrl` still returns the configured `APP_URL` and never the request origin, so the invite link in the email body remains attacker-independent; the host-header-poisoning protection added in #1523 / #1729 is untouched. What the change does alter is which requests pass validation on this one route — forwarded headers now participate, exactly as they already do on the password-reset path (`auth/api/reset.ts:46`) that the reporter confirmed works on the same deployment. The scope is deliberately minimal, the PR documents what it consciously leaves alone, and the regression test is a real one rather than a restatement of the new behaviour.

## Verdict

✅ **approve** — There are no blocker, major or minor findings. The change is a one-line correction that removes a real production defect, it matches the pattern every sibling call site already uses and the repository's own recorded lesson, it breaks no contract surface, and it ships a regression test that I confirmed fails without the fix. The two nits below are optional follow-ups, not merge conditions.

## 🧪 Validation Gate

GitHub check data was unusable for this head, so the gate was reproduced locally — see the CI note below. Runner: **local** (no compose `app` container running; `docker compose -f docker-compose.yml ps --status running -q app` and the `docker-compose.fullapp.dev.yml` probe both returned empty).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | 25/25 packages built. |
| `yarn generate` | ✅ PASS | Clean; no generated-file drift, as expected for a change that touches no auto-discovered module file. |
| `yarn build:packages` (second run) | ✅ PASS | Rebuild over the freshly generated files. |
| `yarn i18n:check-sync` | ✅ PASS | No locale drift — the diff adds no user-facing strings. |
| `yarn i18n:check-usage` | ✅ PASS | Exit 0. The 3828 unused-key report is the repository-wide advisory count and is unrelated to this diff. |
| `yarn typecheck` | ✅ PASS | 27/27 workspace tasks. I also ran `npx tsc --noEmit` directly in `packages/core` and `packages/shared` — 0 errors in both. |
| `yarn test` | ⚠️ 33/34 — the one failure is environmental, not this PR | See the note below the table. |
| `yarn build:app` | ⏭️ NOT RUN | Skipped as out of proportion to a one-argument change in an API route that `yarn typecheck` and `yarn build:packages` already cover; the required CI check remains the gate for it. |

**On the one failing task.** `create-mercato-app#test` failed with four failures in `packages/create-app/src/lib/template-dev-log-files.test.ts`, all raising the same assertion: `❌ Linux file-watch limits are too low for Turbopack. fs.inotify.max_user_watches: 253174 < 4194304, fs.inotify.max_user_instances: 128 < 4096`. That is this review machine's sysctl configuration, not a defect on the branch — the tests spawn the standalone template dev wrapper, which refuses to start under the host's inotify limits. This PR changes no file under `packages/create-app` (the diff is three files, all in `packages/core` and `packages/shared`), so the failure cannot originate here. I am reporting it rather than quietly calling the gate green, but I am not counting it against the change.

Targeted evidence beyond the gate, run in the isolated review worktree at the PR head (`481093e7a`):

- `npx jest src/modules/auth` in `packages/core` — **91 suites, 759 tests, all passing.**
- `npx jest src/lib/__tests__/url.test.ts` in `packages/shared` — **22/22 passing.**
- `npx jest src/modules/customer_accounts/__tests__/signup-security-email-url.test.ts` — passing, so the sibling call site's source-level guard is unaffected.
- **Mutation check.** I reverted the one-line fix back to `getSecurityEmailBaseUrl(req.url)` and re-ran the route suite: exactly one test failed — `sends the invite behind a reverse proxy whose forwarded host matches APP_URL` — and the other 13 still passed. The new test therefore pins this specific fix and nothing else, which is what the PR body claims.

## ⏳ CI status — no check has run on this head

**This review is not backed by CI, and that is worth stating plainly.** #5671 is a fork PR, so both workflow runs for head `481093e7a` — `CI for Develop&Main` and `Mutation tests` — are sitting at GitHub's `action_required` conclusion, waiting for a maintainer to approve workflows on a fork contribution. They have not executed, so there is no `Test`, `typecheck` or `build:app` signal on this commit; `gh pr checks 5671` shows only `license/cla` (pass) and the `[code]smith` skip, which is easy to misread as "CI is clean". It is not clean, it is absent.

That is why I ran the repository's configured validation gate locally instead of reading GitHub check results, and why the table above is this run's own evidence rather than a substitute for branch protection. **A maintainer still needs to approve the workflow runs so the required checks execute before this merges** — the local gate does not satisfy branch protection and is not permission to merge without CI. I have not approved the runs myself; that is a maintainer decision about executing fork code in CI, even for a diff this small.

## Findings

### 💅 Nit

**1. Nothing stops the string form from coming back — `packages/core/src/modules/auth/api/users/resend-invite/route.ts:142`.** The repository already treats this failure mode as worth a source-level guard elsewhere: `packages/core/src/modules/customer_accounts/__tests__/signup-security-email-url.test.ts:21` asserts that `api/signup.ts` matches `/getSecurityEmailBaseUrl\s*\(\s*req\s*\)/` precisely so the argument cannot silently regress. This PR's behavioural test is strictly stronger for this one route, so nothing is missing for resend-invite itself — but the codebase has no check that would catch a *new* call site passing a string. If you want to close that off cheaply, a single shared test that reads every file matching `getSecurityEmailBaseUrl(` / `toSecurityEmailUrl(` and asserts no argument is `.url` would generalise the signup guard to the whole helper. Entirely the author's call, and equally reasonable as a follow-up issue attached to the footgun discussion in the PR's "Not addressed here" section.

**2. The new `url.test.ts` case pins a known footgun as expected behaviour — `packages/shared/src/lib/__tests__/url.test.ts:92`.** `cannot see forwarded headers when the request is passed as a URL string` is a good test to have; documenting the asymmetry is better than leaving it invisible. The risk is that it now reads like a contract. A maintainer who later takes up option 1 from the issue — drop `string` from `RequestInput`, or make the string branch fail closed — will hit `expect(() => assertAllowedAppOrigin(request.url, env)).toThrow(AppOriginRejectedError)` and cannot tell from the test whether they are breaking an intended guarantee or updating a documented wart. One comment line above the test naming it as current-behaviour documentation and referencing #5382 would remove that ambiguity.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — nothing is exported, renamed or removed; the diff changes one call argument and adds two test cases.
- [x] No function signature changed in a breaking way — `getSecurityEmailBaseUrl(input?: RequestInput, env?)` is untouched, and `Request` was already a member of `RequestInput` (`packages/shared/src/lib/url.ts:13`).
- [x] No HTTP route URL removed or renamed; no method changed — `POST /api/auth/users/resend-invite` keeps its path, method, request body and every status code it could already return.
- [x] No field removed or retyped in an existing response shape — the 200 `{ ok, warning? }` and the 400 `{ error: 'Invalid request origin' }` bodies are unchanged; the change only alters *when* the 400 is produced.
- [x] No database table or column renamed or removed — the diff contains no entity, migration or snapshot change, and none is warranted.
- [x] No config key renamed and no default changed silently — `APP_URL`, `NEXT_PUBLIC_APP_URL` and `APP_ALLOWED_ORIGINS` are read exactly as before.
- [x] Where a contract had to change: not applicable, no contract surface from `BACKWARD_COMPATIBILITY.md` is touched.

One behavioural note that is worth stating explicitly even though it breaks no contract: because forwarded headers now participate in the check on this route, a request whose `Host` header falls outside the allowlist is *rejected* where it previously slipped through on a matching `req.url`. That is the intended tightening and it matches `reset.ts`, but it is the reason I routed the PR to `needs-qa` rather than `skip-qa` — see the QA note below.

## 🧪 Test Coverage

Coverage for the fix itself is good, and I verified it rather than assuming it. `resend-invite.route.test.ts:318` reproduces the issue's deployment shape faithfully — container URL on an internal port (`https://localhost:6789`), `host` and `x-forwarded-host` both resolving to `app.example.com`, `x-forwarded-proto: https`, `APP_URL=https://app.example.com`, `NODE_ENV=production` — and asserts not just the 200 but that the token rotation (`mockNativeUpdate`), the flush and the send all actually happened, so a route that returned 200 without doing the work would still fail. Reverting the fix turns exactly this test red and nothing else. `url.test.ts:92` complements it one layer down by asserting the same `Request` passes and its `.url` string does not, which is the cleanest possible statement of the root cause. The pre-existing negative tests still hold: `rejects a poisoned host before rotating invite tokens` (`:263`) and `allows loopback origin mismatches outside production` (`:279`) both continue to pass, so the fix has not loosened the poisoning guard.

Two gaps, neither of which I would hold the merge for:

- **No integration coverage for the proxy shape.** The PR argues this would need a reverse-proxy deployment fixture, and that is a fair reading — the defect only manifests when the server's listen address differs from the public origin, which the existing Playwright harness does not model. `TC-AUTH-044-resend-invite.spec.ts` already exercises the endpoint end to end (200, 409, 404, 422) and continues to apply, so the API path is not uncovered, just not covered *in the proxied shape*. This is the reason the automated-verification exemption is not cleanly earned here: `.ai/docs/pr-workflow.md` requires an integration test for a change that touches auth before `skip-qa` applies. Rather than block on a fixture that does not exist, I have labelled the PR `needs-qa` so a reviewer verifies the normal-path resend still works — which is exactly the regression this tightening could cause.
- **The residual case the fix does not reach.** `assertAllowedAppOrigin` throws on *any* header-derived origin outside the allowlist (`packages/shared/src/lib/url.ts:217-222`), and unlike the URL origin there is no loopback rescue for a header origin. A proxy configured to forward its upstream `Host` (nginx's `proxy_set_header Host $proxy_host`, giving something like `Host: app:3000` alongside a public `X-Forwarded-Host`) will therefore still get a 400 after this fix. This is not a regression and not this PR's problem — the already-shipped `reset.ts` path behaves identically, which is consistent with the reporter observing that reset works on their deployment — but it does mean "broken in every proxied deployment" becomes "fixed for proxies that preserve the public `Host`". Worth a sentence in the follow-up issue rather than a change here.

**Recommended follow-up (not blocking).** The PR's "Not addressed here" section raises two real items that deserve their own issue: the `string` branch of `RequestInput` degrading silently instead of failing closed, and `getSecurityEmailBaseUrl()` with no argument skipping origin validation entirely (`packages/shared/src/lib/url.ts:224`), which is why `auth/commands/users.ts:492` works. That second one is the more interesting asymmetry: the *same* invitation email is origin-validated when resent but not when first sent. Both change a shared helper's public contract and were correctly kept out of a one-line bug fix. The residual proxy case above belongs in the same issue.

pkarw:
## 🏷️ label rationale

`om-auto-review-pr` applied the following after an approving code review. Labels: `bug`, `priority-high`, `risk-high`, `needs-qa`, `merge-queue`.

- **`merge-queue`** (pipeline) — the review verdict is approve, so the PR moves straight to the merge queue. The PR carried no pipeline label before this run.
- **`bug`** (category) — this restores a broken endpoint (#5382); it adds no feature and refactors nothing.
- **`priority-high`** (priority) — the inference rule in `.ai/docs/pr-workflow.md` routes auth/session fixes to `priority-high`, and this one matches on substance too: `POST /api/auth/users/resend-invite` returns 400 in any deployment where the server's listen address differs from the public origin, which is every containerised production install. The only workaround on record widens `APP_ALLOWED_ORIGINS` for every route that uses the helper, so it is a stopgap rather than a reason to downgrade urgency. This is one step above the author's suggested `priority-medium`.
- **`risk-high`** (risk) — the same rule sends changes to auth to `risk-high`, and I am keeping it rather than talking it down to the "single-module bug fix with tests" bucket, because the two signals conflict and the policy says take the higher one. The change genuinely alters which requests the origin check accepts on this route: forwarded headers now participate, so a request whose `Host` falls outside the allowlist is rejected where it previously passed on a matching `req.url`. That is the intended tightening and it matches `auth/api/reset.ts`, but "one line changed" describes the diff, not the blast radius. Per the same policy, `risk-high` is a reason to want the deeper check below, not a reason to block.
- **`needs-qa`** (meta) — **this overrides the `skip-qa` the author suggested, and the reasoning is not a formality.** The automated-verification exemption needs more than "no UI file touched": for a change that touches auth it also requires an integration test covering the changed behaviour in the same PR. This PR ships excellent unit coverage (and I verified it fails without the fix) but no new integration test, for a defensible reason — the proxied shape needs a reverse-proxy fixture that the harness does not have. The existing `TC-AUTH-044-resend-invite.spec.ts` covers the endpoint but not the proxied shape. So the exemption is not earned, and `.ai/docs/pr-workflow.md` says such a PR keeps `needs-qa` and is verified through its applicable interface instead. There is real work for a QA reviewer here: confirming that the tightened check has not broken the ordinary, non-proxied resend path. Manual QA instructions are in a separate comment.

`needs-qa` without `qa-approved` is a hard merge block while `qaGate` is on, so this PR sits in `merge-queue` until a QA reviewer signs off. Note also that no CI has run on this head — see the review body.

pkarw:

This PR is approved and requires manual QA (`needs-qa`, no `skip-qa`). It is queued in `merge-queue`, but the QA-approval gate holds it until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` by swapping the labels (remove `merge-queue`, add `qa`), then run the routes below.

**Read this first, because it changes what you are looking for.** The bug this PR fixes (#5382) only reproduces behind a reverse proxy whose public origin differs from the app's listen address, which you almost certainly cannot stand up as part of routine QA. **Do not spend your time trying to reproduce the original 400** — the automated regression test already pins it, and I verified it fails without the fix. What manual QA is actually for here is the *other* direction: the change makes the `Host` and `X-Forwarded-Host` headers participate in the origin check on this route where previously they were ignored entirely, so the realistic risk is that an ordinary, non-proxied install that worked before now starts rejecting. That is what the P0 block below is designed to catch.

The only user-facing surface in the diff is the **Resend Invite** button on the user edit page. Nothing else in the app changed.

### P0 — Resend invite still works on a normal install (auth)

**Where to click**
- `/backend/users` → **Create user** (or `/backend/users/create` directly)
- `/backend/users/{id}/edit` for the user you just created

**What to verify**
- Create a user **with an invitation and no password** (tick the send-invite option, leave the password empty). The button under test only renders for a user who has not set a password yet, so a user created with a password will not show it. → The user is created and the invitation email goes out.
- Open that user's edit page and confirm a **Resend Invite** button is present next to Save/Cancel. → It is visible and enabled.
- Click **Resend Invite**. → A green **"Invitation email sent"** flash appears and the button returns from "Sending..." to its normal label. **This is the single most important check in this PR** — a red **"Failed to send invitation email"** flash here is the regression to report.
- Check the delivered email (or your mail catcher). → The invite link points at your configured `APP_URL`, not at `localhost` or an internal container address, and clicking it opens the set-password page and lets you set a password.
- Click **Resend Invite** a second time. → It still succeeds, and the link from the *first* email is now dead (previous unused tokens are invalidated on each resend) while the newest link still works.
- Open the edit page of a user who **already has a password**. → No Resend Invite button is rendered at all (409 path; unchanged by this PR, worth one glance).

**What can go wrong**
- The regression this PR could plausibly introduce: **"Failed to send invitation email"** on a setup that worked before the change. If you hit it, please grab the server log line from the `shared:origin-check` component — it prints `host`, `forwardedHost`, `forwardedProto` and `allowedOrigins`, which is exactly what is needed to diagnose it — and attach it with your `qa-failed` report.
- If your QA environment sits behind *any* proxy or tunnel (ngrok, a shared staging domain, a container with a published port), that is a **more** valuable test than a plain `localhost` run, not a less valuable one. Please say in your QA note which shape you tested — plain localhost, or proxied — because they exercise different branches of the check.
- An amber **"Invite token created but the email could not be sent"** flash is a *mail transport* problem (SMTP not configured in your environment), not this PR. Treat it as a pass for the origin check and note it separately.
- Confirm the action is still permission-gated: a user without `auth.users.create` should not be able to resend. Also confirm you cannot resend for a user in another tenant.

### P2 — No collateral surface

**Where to click**
- `/backend/auth/reset` (the password-reset request flow) and the reset link it emails

**What to verify**
- Request a password reset and follow the emailed link. → It still works and the link still points at `APP_URL`. This path was already correct and is untouched by the diff; one pass is enough to confirm nothing shifted underneath it.

**What can go wrong**
- Nothing is expected here. Only check it because reset and resend-invite now share the same argument shape, so a surprise on reset would mean the shared helper, not the route.

### Pass/fail

One more thing before merge, independent of QA: **no CI has run on this head.** Both workflow runs are at `action_required` because this is a fork PR, so a maintainer needs to approve them and let the required checks go green. See the review body for details.

pkarw:

Review posted as an approving review with the full report; labels set to `bug`, `priority-high`, `risk-high`, `needs-qa`, `merge-queue` with the rationale in a separate comment; manual QA instructions posted separately.

- **autofix: skipped (not my PR — re-run with `--autofix` to fix it here).** Nothing needed fixing in any case: the review found no blocker, major or minor findings, only two optional nits.
- **CI follow-up: not applicable, and this is the one thing still outstanding.** Both workflow runs for head `481093e7a` are at `action_required` — GitHub is waiting for a maintainer to approve workflows on a fork contribution — so there is nothing to wait for and no `ci-monitoring` was set. `gh pr checks` showing only `license/cla` (pass) is the *absence* of CI, not a green CI. **A maintainer needs to approve the runs so the required checks execute before this merges.** I did not approve them myself; executing fork code in CI is a maintainer's call.
- **Validation evidence:** the repository's configured gate was run locally at the PR head in an isolated worktree. `build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage` and `typecheck` all pass; `yarn test` came back 33/34 with the single failure in `packages/create-app` caused by this machine's inotify sysctl limits, in a package this PR does not touch. `build:app` was not run. This stands as this review's own evidence only — it does not satisfy branch protection.
- **Merge gate:** `needs-qa` is set without `skip-qa`, so this PR must not merge until a QA reviewer adds `qa-approved`.

PR #866, opened 2026-03-05
feat: add Docker command wrappers

dominikpalatynski:
### Summary of Changes

This change introduces a complete Docker command parity layer so Windows developers
can run any monorepo command from their native terminal without WSL.

**`scripts/docker-exec.mjs`** _(new)_
Cross-platform Node.js helper that auto-detects the active Docker Compose profile
(`fullapp.dev.yml` or `fullapp.yml`), executes the requested `yarn` script inside
the running `app` container, and prints actionable startup instructions if no
container is found. Supports `DOCKER_COMPOSE_FILE` env override.

**`package.json`** _(modified)_
18 new `docker:*` scripts added, split into two groups:

_Compose lifecycle (start/stop stacks — no container needed):_
| Script | Compose file |
|--------|-------------|
| `docker:up` / `docker:down` | `docker-compose.fullapp.yml` (production-like) |
| `docker:dev:up` / `docker:dev:down` | `docker-compose.fullapp.dev.yml` (dev, hot reload) |
| `docker:ephemeral` / `docker:ephemeral:down` | `docker-compose.preview.yaml` (fresh DB, port 5000) |

_Exec commands (1:1 mirrors of native scripts, require stack running):_
`docker:dev`, `docker:build:packages`, `docker:generate`, `docker:initialize`,
`docker:reinstall`, `docker:db:migrate`, `docker:db:generate`, `docker:lint`,
`docker:typecheck`, `docker:test`, `docker:install-skills`, `docker:mercato`

**`docker/README.md`** _(modified)_
New "Windows + Docker Developer Command Cookbook" section added with:
- Prerequisites and quick-start workflow
- Compose lifecycle command reference table
- Exec command reference table (native → docker equivalent)
- `docker:mercato` CLI passthrough examples
- Script compatibility matrix (all root scripts classified as `works`,
  `works-with-wrapper`, or `unsupported-by-design` for each Docker profile)
- Troubleshooting sub-section

---

### Testing

All scenarios below are **manual tests**. No automated Playwright tests exist for
Docker CLI commands — verification requires a running Docker environment.

| ID | Scenario | File |
|----|----------|------|
| TC-DOCKER-001 | Dev stack startup and shutdown (`docker:dev:up` / `docker:dev:down`) | `.ai/qa/scenarios/TC-DOCKER-001-dev-stack-startup-shutdown.md` |
[...]

PR #5679, opened 2026-08-26
feat(currencies): add NBP average-rate provider

vloneskorpion:
## Summary

- add explicit-only `nbp_average` provider for NBP tables A/B with validated official-rate provenance
- add provider/type selection to lookup and fetch flows while preserving legacy default behavior
- expose average rates and publication references through migration, API/OpenAPI, forms, configuration UI, locales, and documentation

## Validation

- `yarn workspace @open-mercato/core test --runInBand --runTestsByPath …` — 59 passed
- `yarn typecheck`
- `yarn lint` — passes; 10 existing app warnings
- `yarn build:packages`
- `yarn build:app`
- `OM_INTEGRATION_MODULES=currencies … yarn test:integration` — 44 passed, 1 skipped
- `yarn db:generate` — currencies clean after generated migration review

## QA

`TC-CUR-015-nbp-average-rate-contract` captures the separate NBP provider cards and the average-rate publication-reference row in an isolated Playwright environment.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5679"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790348032&installation_model_id=432243&pr_number=5679&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5679&signature=b67bdc9102356e3df515b058ba33ea77f4b9a647fc4765f3ca348c4be2a65f39"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


vloneskorpion:
Implementation for TraceCore #866 is complete and ready for review.

Proposed labels: `feature`, `priority-medium`, `risk-high`, and `needs-qa` (the change includes a core service/API contract, database migration, and UI). Automated Playwright evidence is included; no manual QA approval is claimed.

Automation could not set labels or assignee because this token lacks maintainer permissions.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-26T23:46:12Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: feat(currencies): add NBP average-rate provider

## 🎯 Summary

This change adds an explicit-only `nbp_average` rate provider that reads NBP tables A and B, persists official foreign-currency-to-PLN `average` rates, and records the NBP publication number in a new nullable `exchange_rates.external_reference` column. To keep the new rates out of existing consumers it introduces a `selectionMode: 'default' | 'explicit'` concept on `RateProvider`: `RateFetchingService.fetchRatesForDate` now only auto-runs `default` providers, and `ExchangeRateService` excludes explicit sources from unfiltered reads while gaining `provider` / `rateType` selectors. It also hardens `POST /api/currencies/fetch-rates` (zod request validation, mutation guards, no error-message leakage) and surfaces the new provider and provenance column through the config UI, the exchange-rates list, OpenAPI, five locale files, `UPGRADE_NOTES.md`, a spec, unit tests and a Playwright case.

The design is good work. The explicit/default split is the right way to add a provider without changing behavior for existing callers; the provider validates every field of the NBP payload with zod plus targeted parsers instead of trusting the upstream shape; `validateBatch` rejects a conflicting duplicate batch before opening a transaction; `subtractDays` is correctly moved off local-time `setDate` onto `setUTCDate`; and the regenerated `.snapshot-open-mercato.json` — which looks alarming at 552/231 lines — is just the currencies snapshot catching up to the alphabetised ordering every other core module already uses, so I checked it and it carries no schema drift beyond the one added column. Reviewed scope: all 38 changed files, with the services, the two API routes, the entity/migration pair, the guard plumbing and the tests read in full.

What blocks it is one mechanical CI failure and two correctness/process problems described below.

## Verdict

❌ **request changes** — the locale files are not alphabetically sorted, so the `Check i18n sync` step in CI's `prepare` job exits 1 and takes every downstream gate with it; the new mutation-guard plumbing on `fetch-rates` silently disables every feature-gated guard; and a feature is being landed on `main` under an already-tagged release heading instead of on `develop`.

## 🧪 Validation Gate

Validation source: **GitHub checks** (fork runs sat at `action_required`; I approved run `32983132230` / `32983132267` so the gate would actually execute against `214c0bb9`), supplemented by a dependency-free local reproduction of the i18n gate.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | CI `prepare` → "Build packages" / "Rebuild packages with generated files" both green. |
| `yarn generate` | ✅ PASS | CI `prepare` → "Prepare generated modules" green. |
| `yarn build:packages` (second pass) | ✅ PASS | Green — this repeated step is load-bearing and did run. |
| `yarn i18n:check-sync` | ❌ **FAIL** | Reproduced locally without installing dependencies: all five currencies locale files report `UNSORTED at index 27`. `scripts/i18n-check-sync.ts:62` compares the key order against a plain sort and `:245` exits `1`. See Blocker 1. |
| `yarn i18n:check-usage` | ⚪ n/a | Runs with `continue-on-error: true` in `.github/workflows/ci.yml:481`, so it cannot gate. |
| `yarn typecheck` | ⏳ pending | Lives in the `test` job, which depends on `prepare` — it will be skipped once Blocker 1 fails `prepare`. |
| `yarn test` | ⏳ pending | Same dependency. `Mutation tests` reported success, but that job's allowlist does not include `packages/core`, so it is a skip and not evidence about this diff. |
| `yarn build:app` | ✅ PASS | CI `prepare` → "Build app" reached and passed before the i18n step. |
| `lint` / `ds-lint` | ✅ PASS | Both completed green on this head. |

CI on `214c0bb9c87bda72d56def716c22bd3caa6c0324` was still in progress when this review was submitted; the review is not waiting on it, and a CI-result follow-up comment will land separately.

## Findings

### ⛔ Blocker

**1. `packages/core/src/modules/currencies/i18n/{en,de,es,ko,pl}.json` — unsorted keys fail the `Check i18n sync` CI step and block every gate behind it.**

Two new keys are inserted out of alphabetical order in each of the five locale files:

- `currencies.fetch.provider_nbp_average` is placed *after* `currencies.fetch.provider_nbp_description` (`en.json:30` and the same position in `de`/`es`/`ko`/`pl`), but `provider_nbp_average` sorts before `provider_nbp_description`.
- `exchangeRates.list.type.average` is placed *after* `exchangeRates.list.type.buy` (`en.json:162`, with `...type.buy` at `:161`), but `average` sorts before `buy`.

`scripts/i18n-check-sync.ts:58-63` flags any file whose key order differs from a plain lexicographic sort, and `:238-245` prints `N issues found` and calls `process.exit(1)`. `.github/workflows/ci.yml:476-477` runs that script as `Check i18n sync` inside the `prepare` job — the same job the `test` and `build` jobs depend on — so this failure does not merely add one red check, it prevents `yarn typecheck` and `yarn test` from running against this diff at all. I confirmed the ordering violation directly against the committed files; every one of the five reports `UNSORTED at index 27`.

This is also why the PR description's Validation section looks green: it lists `yarn typecheck`, `yarn lint`, `yarn test`, `yarn build:packages` and `yarn build:app`, but not `yarn i18n:check-sync`, which is step 4 of the repo's configured gate in `.ai/agentic.config.json`.

Fix: run `yarn i18n:check-sync --fix` and commit the reordered files, then re-run the full `validation.commands` list so the gates behind `prepare` actually report on this change.

### ⚠️ Major

**2. `packages/core/src/modules/currencies/api/fetch-rates/route.ts:32-37` — `userFeatures()` always returns `[]`, so every feature-gated mutation guard is silently skipped on this endpoint.**

The new helper reads the granted features off the auth context:

```ts
function userFeatures(auth: unknown): string[] {
  const features = (auth as { features?: unknown }).features
  ...
}
```

`getAuthFromRequest` never populates a `features` field. `AuthContext` (`packages/shared/src/lib/auth/server.ts:12-24`) declares `sub`, `sid`, `tenantId`, `orgId`, `email`, `roles`, `isApiKey`, `userId`, `keyId`, `keyName` and an open index signature — there is no `features` claim, and the whole file contains no `features` assignment on either the session or the API-key branch (`:254-263`). So `userFeatures(auth)` evaluates to `[]` on every real request.

That value is passed straight into `runMutationGuards(..., { userFeatures })`, which filters the guard set with `authorizeFeatures(g.features ?? [], { grantedFeatures: context.userFeatures })` (`packages/shared/src/lib/crud/mutation-guard-registry.ts:103`). `authorizeFeatures` short-circuits to `true` only when the required list is empty (`packages/shared/src/security/featurePolicy.ts:37`), so guards with no feature requirement still run — but **every guard that declares `features` is dropped**, silently and unconditionally. The route reads as guard-protected while enforcing only the ungated subset.

The repo already ships the correct helper for exactly this case. `packages/shared/src/lib/crud/route-mutation-guard.ts:12-24` documents `runRouteMutationGuards` as the "Shared registry-based mutation-guard wrapper for custom write routes that do not use `makeCrudRoute`", it performs the identical `getAllMutationGuardInstances()` + `bridgeLegacyGuard()` composition, it returns a ready `response` for the blocked case and a `runAfterSuccess()` closure, and — critically — it resolves features through `rbacService.getGrantedFeatures` (`:99-106`), the same source `makeCrudRoute` uses (`packages/shared/src/lib/crud/factory.ts:1188-1200`). Its own doc comment warns that "supplying the wrong set silently skips guards". Using it would delete roughly 45 lines of hand-rolled plumbing from this route *and* fix the bug.

The new test does not catch this because its mock invents the field the real implementation never sets: `packages/core/src/modules/currencies/api/fetch-rates/__tests__/route.test.ts:29-35` mocks `getAuthFromRequest` to return `features: ['currencies.fetch.manage']`, and `:105` then asserts `{ userFeatures: ['currencies.fetch.manage'] }`. The assertion passes against the fixture and would keep passing no matter what the production auth context contains.

Fix: replace the local `userFeatures`, `runAfterSuccessCallbacks` and inline `runMutationGuards` call with `runRouteMutationGuards({ container, req, auth: { userId: auth.sub, tenantId: auth.tenantId, organizationId: auth.orgId }, input: { resourceKind: 'currencies.fetch_rates', operation: 'custom', mutationPayload: parsed.data } })`, and change the test's mock to return a realistic `AuthContext` (no `features` key) so a regression cannot hide behind the fixture. The same pattern is worth checking in `packages/core/src/modules/eudr/api/plots/import/route.ts:51-55`, which has the identical hand-rolled `resolveUserFeatures`, but that is pre-existing and out of scope here.

**3. `UPGRADE_NOTES.md:27` / PR base branch — a new feature is being landed on `main` and documented under the already-released `0.7.0` heading.**

`baseRefName` on this PR is `main`, but `.ai/agentic.config.json` sets `"baseBranch": "develop"` and the root `AGENTS.md` PR workflow treats `develop` as the integration branch; `main` currently carries the shipped `v0.7.0` tag and `package.json` version `0.7.0` on both branches. The recent PRs that legitimately targeted `main` (#5622, #5623, #5626) are release hotfixes and docs corrections, not features.

The consequence shows up in the docs change: the new "Currencies adds opt-in official NBP average rates" section is inserted under `## 0.6.7 → 0.7.0 (2026-08-26)`, a heading that describes a version already tagged and already merged back into `develop` (b41f7e3e5). Anyone who upgraded to `0.7.0` will not receive this provider, so the note tells them about a feature their release does not contain. Merging here also means `develop` diverges until someone back-merges.

Fix: retarget the PR at `develop` and move the UPGRADE_NOTES entry under a new next-version heading. If landing on `main` is a deliberate call for this particular change, say so on the PR and I will treat it as a documented waiver — but the release-heading placement needs correcting either way.

### 🔹 Minor

**4. `packages/core/src/modules/currencies/api/fetch-rates/route.ts:207-211` — the OpenAPI `errors` list does not cover the statuses the guard path can now return.**

The route now returns whatever a blocking guard chooses: `runMutationGuards` defaults to `422` when a guard rejects without setting `status` (`packages/shared/src/lib/crud/mutation-guard-registry.ts:114`), and a guard is free to return any other status through `MutationGuardResult.status`. The declared `errors` array still lists only `400`, `401` and `500`, so the generated documentation now understates the contract — the same class of drift as letting a route answer with statuses its spec never mentions.

Fix: add a `{ status: 422, description: 'Blocked by a mutation guard', schema: errorSchema }` entry (plus any other status the guards in this codebase actually emit for `currencies.fetch_rates`), or describe the guard-rejection range explicitly.

**5. `packages/core/src/modules/currencies/api/fetch-rates/route.ts:23-30` — the request contract is narrowed in two ways that `UPGRADE_NOTES.md` does not mention.**

The endpoint previously did `const { date, providers } = body` with no validation. It now parses through a `.strict()` schema with `date: z.string().datetime()`. Two previously-accepted request shapes now get a `400`:

- a date-only value such as `{"date": "2026-08-24"}` — `new Date('2026-08-24')` parsed fine before, but `z.string().datetime()` requires a full RFC 3339 timestamp;
- any request carrying an extra property, which `.strict()` rejects outright.

The in-app caller is safe — `CurrencyFetchingConfig.tsx:230-235` posts only `{ providers: [provider] }` — so this is about external API clients, and API routes are a contract surface under `BACKWARD_COMPATIBILITY.md`. Tightening validation is the right direction; it just needs to be declared.

There is a second-order effect worth guarding too: `:98-102` re-parses the guard's output through the same `.strict()` schema, so a guard that attaches any bookkeeping key via `modifiedPayload` turns an otherwise valid request into a `400` rather than doing anything useful. Loosen that second parse, or strip unknown keys before re-parsing.

Fix: add an UPGRADE_NOTES line for the tightened `fetch-rates` request body, and relax the re-parse of the guard-modified payload.

**6. `packages/core/src/modules/currencies/data/validators.ts:142` — `nbp_average` is a snake_case source id inside an enum whose other members are display names.**

`providerSchema` becomes `z.enum(['NBP', 'nbp_average', 'Raiffeisen Bank Polska', 'Custom'])`. The value is not only a config key: it is written into `exchange_rates.source`, rendered raw in the Source column of the exchange-rates list, and hard-coded in `CurrencyFetchingConfig.tsx:45`. Sitting beside `'NBP'` and `'Raiffeisen Bank Polska'` it reads as a different kind of identifier, and once tenants have rows the value needs a data migration to change. This is the cheapest moment to settle it.

Fix: either pick a value consistent with the neighbours (for example `'NBP average'`) or note in the spec that provider sources are moving to snake_case ids and that `'NBP'` / `'Raiffeisen Bank Polska'` are legacy spellings, so the inconsistency is a recorded transition rather than an accident.

### 💅 Nit

**7. `packages/core/src/modules/currencies/data/entities.ts:2` — the data layer now imports from the services layer.** `import type { RateType } from '../services/providers/base'` points `data/entities.ts` at `services/`, inverting the usual direction (services import entities). It is type-only so there is no runtime cycle, but moving `RateType` into `data/` — or deriving it from the existing `rateTypeSchema` in `data/validators.ts` via `z.infer` — keeps the layering one-way.

**8. `packages/core/src/modules/currencies/services/exchangeRateService.ts:264-289` — new internal throws are missing the `[internal]` prefix.** `AGENTS.md` asks that purely internal `throw new Error(...)` messages be prefixed with `[internal]` so the hardcoded-string checker treats them as opted out. The new messages (`'Currency codes must be three-letter ISO codes'`, `'maxDaysBack must be an integer from 0 through 366'`, `'autoFetch must be a boolean'`, `'provider must name a registered rate provider'`, `'rateType must be buy, sell, or average'`, and `'Exchange rate date must be a valid date'` at `:232`) are unprefixed. The checker is advisory today, so this is your call.

**9. `packages/core/src/modules/currencies/components/__tests__/CurrencyFetchingConfig.test.tsx:204` — an existing assertion is weakened.** It changes from `runMutationMock.mock.calls.at(-1)?.[0]?.context?.resourceKind` to a `.some(...)` over all calls. The original pinned that the fetch button's *own* mutation carries `currencies.fetch_rates`; the replacement passes as long as any mutation in the whole interaction does. If the widening is needed because the component now issues extra mutations, filter to the fetch call instead of scanning everything.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `RateProvider` gains only the optional `selectionMode`, and `RateProviderResult` only the optional `externalReference`.
- [x] No function signature changed in a breaking way — `RateSelectionOptions` is a superset of the previous inline options type; `findRateWithFallback` and `findRatesForDate` are private.
- [x] No required type field removed or narrowed — `ExchangeRate.type` moves from `string | null` to `RateType | null`, which is a narrowing at the type level, but the only writers are the validators (already restricted to `buy`/`sell`) and the providers, so no existing value falls outside the union.
- [ ] **No field removed or retyped in an existing response shape** — `exchangeRateListItemSchema.type` is retyped from `z.string().nullable()` to `z.enum(['buy','sell','average']).nullable()` (`api/exchange-rates/route.ts:193`). This is documentation-only rather than runtime-enforced, and every value the writers can produce is inside the new union, so it is safe in practice; it is listed here because it does narrow a published response contract. The `externalReference` addition alongside it is purely additive.
- [x] No CLI command or flag renamed or removed — `currencies:list-providers` gains two output lines, which is additive.
- [x] No database table or column renamed or removed; no column type narrowed — `Migration20260826143024_currencies` adds one nullable `text` column with a matching `down()`.
- [ ] **Where a contract had to change: old surface kept working, with migration notes** — see Findings 3 and 5. The `POST /api/currencies/fetch-rates` request body is narrowed (`.strict()` plus a stricter `date` format) without an UPGRADE_NOTES entry, and the note that does exist is filed under an already-released version heading.

One behavior change deserves calling out even though it is intentional and correct: `ExchangeRateService.findRatesForDate` now appends `source: { $nin: <explicit provider sources> }` to every unfiltered read (`exchangeRateService.ts:218-220`). That is what keeps `nbp_average` rows invisible to existing callers, and `exchangeRateService.selection.test.ts:11-35` pins it. Worth remembering that it also means any third-party provider registered with `selectionMode: 'explicit'` disappears from default lookups.

## 🧪 Test Coverage

Covered, and covered well in most places. `providers/__tests__/nbpAverage.test.ts` exercises the provider's five real branches against a mocked `fetchWithTimeout`: an A+B success mapping that asserts direction (`EUR → PLN`), `type: 'average'`, the `externalReference` publication number and the exact rate string; a `404` on table A treated as a successful empty contribution rather than a failure, which matters because table B is published weekly; a duplicate currency across A and B rejected before any rate is returned; a `500` surfaced as one provider failure; and the PLN-not-active short-circuit asserted by `expect(fetchMock).not.toHaveBeenCalled()`. `rateFetchingService.selection.test.ts` pins that a `default` provider runs and an `explicit` one does not until named, checks all three `getProviderSources` arities, and asserts the conflicting-batch rejection happens before `em.transactional` is touched. `exchangeRateService.selection.test.ts` covers the `$nin` exclusion, the `provider` + `rateType` filter with the trimming of `' nbp_average '`, and the unknown-provider rejection before any query. The list route, the page component and the validators all gained targeted cases, and the `$nin`/`type` handling was added to the shared mock EntityManager so those assertions mean something.

Gaps worth closing:

- **The guard feature-gating path has no honest test.** As described in Finding 2, `route.test.ts:29-35` fabricates `auth.features`. A test using a realistic `AuthContext` — and a guard that declares `features` — would fail today and would have caught the bug.
- **No test covers a guard rejection.** `runMutationGuards` is only ever mocked as `ok: true`. Add a case asserting that `ok: false` with `errorStatus: 422` reaches the client unchanged, which also documents the status that Finding 4 asks you to declare.
- **The provider→persist path is never exercised end to end.** `TC-CUR-015` deliberately seeds `external_reference` with a raw `UPDATE` through `pg` after creating the row via the manual API, so it proves the read and UI contract but never the write path from `NBPAverageRateProvider` through `storeRates`. The comment at `TC-CUR-015-nbp-average-rate-contract.spec.ts:37-43` explains the choice and it is defensible for an integration test that must not call the live NBP service, but a unit-level test that runs `RateFetchingService.fetchRatesForDate` with the average provider mocked and asserts `externalReference` lands on the persisted row would close the loop cheaply.
- **`validateBatch`'s rate/reference conflict branch is untested.** `rateFetchingService.ts:166` throws on a differing `rate` or `externalReference` for the same key, but only the `type` conflict at `:163` has a test.
- **`parseReference` and `parseEffectiveDate` rejections are untested.** A table whose `no` does not contain the expected table letter, and an `effectiveDate` that disagrees with the requested date, both throw; neither is covered.

pkarw:
🏷️ **label rationale**

- `changes-requested` — pipeline state after a changes-requested review: one blocker (unsorted locale keys fail the `Check i18n sync` step in CI's `prepare` job) plus two majors (feature-gated mutation guards silently skipped on `POST /api/currencies/fetch-rates`; a feature landing on `main` under the already-tagged `0.7.0` UPGRADE_NOTES heading).
- `feature` — the change adds a new `nbp_average` rate provider, a new `exchange_rates.external_reference` column and new selection options, rather than fixing existing behavior.
- `priority-medium` — an additive, opt-in provider that no shipped flow depends on yet, so it is not urgent, but it is a requested capability with a spec behind it.
- `risk-high` — the blast radius is wider than the feature itself: `ExchangeRateService.findRatesForDate` now appends a `source $nin` clause to **every** unfiltered rate read, `RateFetchingService.fetchRatesForDate` changes which providers run by default, `POST /api/currencies/fetch-rates` gains guard plumbing and a narrowed request contract, and there is a schema migration. Agreed with the author's own proposal of `risk-high`.
- `needs-qa` — the diff touches `.tsx` outside tests (`backend/exchange-rates/page.tsx`, `components/CurrencyFetchingConfig.tsx`) and changes both the database structure and the API surface, so the automated-verification exemption does not apply and `skip-qa` would be wrong. Not applying `qa-approved`: that is earned by manual QA, not by reading the diff.

Also applied out-of-band: I approved the two pending fork workflow runs on `214c0bb9` (they were parked at `action_required`) so this head gets real CI signal. The diff contains no `.github/` changes, so there was nothing to gate on there.

pkarw:
@vloneskorpion handing this back to you — changes requested.

The quickest path back to green:

1. `yarn i18n:check-sync --fix` and commit the five reordered currencies locale files. This is the CI blocker: the `Check i18n sync` step lives in the `prepare` job, so until it passes, `yarn typecheck` and `yarn test` never run against this diff at all — which is why the branch currently looks better validated than it is.
2. Swap the hand-rolled guard plumbing in `api/fetch-rates/route.ts` for `runRouteMutationGuards` from `@open-mercato/shared/lib/crud/route-mutation-guard`. Right now `userFeatures(auth)` reads an `auth.features` field that `getAuthFromRequest` never sets, so every feature-gated mutation guard is dropped; the shared helper resolves features through `rbacService.getGrantedFeatures` and deletes ~45 lines here. Please also drop the fabricated `features` array from the test's auth mock so the gap cannot hide again.
3. Retarget the PR at `develop` and move the UPGRADE_NOTES entry out of the already-released `0.6.7 → 0.7.0` section. If landing on `main` is deliberate for this change, say so and I will treat it as a documented waiver — the heading placement still needs fixing either way.

The minors and nits (OpenAPI `422`, the `.strict()` narrowing note, the `nbp_average` naming, the layering and assertion nits) are all in the review body. The provider itself, the explicit/default selection model and the unit-test coverage around them are solid work — no rework needed there.

Re-run `/om-auto-review-pr 5679` when you have pushed and I will take another pass.

pkarw:

One blocker (`yarn i18n:check-sync` fails on unsorted keys in all five currencies locale files), two majors (feature-gated mutation guards silently skipped on `POST /api/currencies/fetch-rates`; a feature landing on `main` under the already-released `0.7.0` UPGRADE_NOTES heading), three minors and three nits. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

Swapping `in-progress` for `ci-monitoring`: the review work is done and the PR is free for anyone to act on, but I still owe a CI-result comment for run `32983132230`, which I approved mid-review.

pkarw:

Run [`32983132230`](https://github.com/open-mercato/open-mercato/actions/runs/32983132230) on `214c0bb9` settled: **failure**. It confirms Blocker 1 exactly, and nothing else went red.

| Job | Result |
|-----|--------|
| `lint` | ✅ success |
| `prepare` | ✅ success |
| `ds-lint` | ✅ success |
| `audit-scope` | ✅ success |
| `test` | ❌ **failure** — `Check i18n sync` |
| `ephemeral-integration (none)` | ✅ success (affected-only shard, not a skip) |
| `audit` / `docker-build` / `merge-coverage` | ⚪ skipped |
| `Mutation tests` (separate run) | ⚪ green, but its allowlist excludes `packages/core`, so it is a skip rather than evidence about this diff |

The failing step:

```
[check] Checking translation sync across 5 locales (en, pl, es, de, ko)...
[currencies] en.json: format issue (unsorted keys)
[currencies] pl.json: format issue (unsorted keys)
[currencies] es.json: format issue (unsorted keys)
[currencies] de.json: format issue (unsorted keys)
[currencies] ko.json: format issue (unsorted keys)
```

**One correction to my review body:** I placed the `Check i18n sync` step in the `prepare` job. It is in the `test` job — `prepare` passed and built the app fine. The consequence I described is unchanged and is now confirmed rather than predicted: the step is #16 of the `test` job, and everything behind it was **skipped**, not run — `Check i18n usage`, `Check for time-bomb test literals`, `Check agent instruction budget`, `Checking types`, `Test`, `Test scripts`, `Check DS token snapshot drift`, `Check create-app template parity` and `Run repo-wide audit guards`. So `yarn typecheck` and `yarn test` still have not executed against this head, and this PR has no type or unit-test signal until the locale files are sorted.

`yarn i18n:check-sync --fix` and a push should turn the whole `test` job over and reveal whatever the remaining gates have to say. The two majors in the review stand independently of CI.

Removing `ci-monitoring` — this was the follow-up it promised, and no further CI follow-up will come from this agent.

vloneskorpion:
Addressed the requested changes in `7489c963f`:

- replaced the hand-rolled fetch-rates guard path with `runRouteMutationGuards`, so feature-gated guards resolve the caller's real RBAC grants; added transformed-payload, rejection, and after-success ordering coverage
- kept the external request strict while stripping guard-only bookkeeping keys before revalidation
- sorted all five currencies locale files and restored an exact fetch-now UI assertion
- added OpenAPI `422`, moved the upgrade note to unreleased `0.7.1`, documented the stricter fetch-rates input and the intentional `nbp_average` machine identifier
- moved the `RateType` source into the data layer while preserving the public provider re-export, and prefixed the new internal validation errors
- added coverage for NBP publication/date validation, publication-reference persistence, and duplicate-rate/reference conflicts
- rebased onto current `develop` and retargeted this PR from `main` to `develop`

Validation (local runner): focused currencies Jest 26/26, full core Jest, `yarn typecheck`, `yarn lint`, `yarn lint:ds`, package builds, `yarn build:app`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, and ephemeral currencies Playwright 45 passed / 1 existing skip. `yarn db:generate` reported currencies clean. The broad parallel `yarn test` run hit two unrelated CLI flakes; both failing suites passed immediately in an isolated 16/16 rerun.

PR #5682, opened 2026-08-26
fix(crud): stop accepting write fields and discarding them silently

cielecki:
## Summary

Customer write endpoints can report success while discarding fields. This PR aliases snake_case input to declared camelCase fields, reports unknown fields as `ignoredFields`, and rejects conflicting spellings or immutable parent changes. Unknown-field rejection remains opt-in because injected forms may include extra keys.

The guard runs at the CRUD factory and scoped-command parsers, with an explicit adapter for activity writes. Nullable deal fields accept `null` without coercing it to zero or the Unix epoch.

## Phone-number persistence

Both activity and interaction create/update commands persist `phoneNumber` through the existing `customValues.callPhoneNumber` field. Activity adapters also forward date and time. Reads use the existing custom-field projection; there is no new phone column or migration.

A phone-only update preserves other custom fields. Matching duplicate representations are accepted; conflicting values return 400 before writing. Omitted phone values remain unchanged. A partial phone-only update with null clears the value; existing call-type validation is unchanged.

The regression tests send a top-level-only phone update and read the stored custom field back. They cover both API surfaces, creation, matching duplicates, conflicts, omission and clearing.

## Compatibility and review follow-up

- `ActivityUpdateInput` retains its optional `entityId` field for source compatibility. The HTTP guard rejects parent changes before parsing.
- Previously ignored snake_case fields are now validated and applied. Invalid values can therefore turn a former success response into a 400.
- Unknown fields are reported unless the route opts into rejection; custom-field and scope handling remain separate.
- Merged current `develop`, preserving its route-indexer wrappers and this PR's command-response input.
- Moved `createWriteGuardConfig` below the imports.

Specification: `.ai/specs/2026-08-26-write-payload-silent-field-drop.md`.

## Validation

- Managed ephemeral regression: `TC-CRM-WRITE-GUARD-001`, 9 passed, no skips or flakes; runner exited 0.
- Workspace unit tests passed across package runs, including core (12,262), shared (2,275), CLI (1,873) and UI (2,067). The default parallel run hit worker signals; retries used Node 24 and Jest `--runInBand`. Queue's 104 tests passed with `--forceExit` because it retained open handles.
- Production application build and package builds passed.
- Typecheck: 30 tasks successful. ESLint on changed TypeScript files passed.
- Translation sync/usage and template parity passed. Translation usage still reports unused-key advisories.

The full upstream integration matrix still needs to run on this head.

### DIFF
--- a/.ai/specs/2026-08-26-write-payload-silent-field-drop.md
+++ b/.ai/specs/2026-08-26-write-payload-silent-field-drop.md
@@ -1,4 +1,4 @@
-# SPEC-072: Customers Write API — Stop Accepting Fields and Discarding Them
+# CRUD Write Payloads: Stop Accepting Fields and Discarding Them
 
 ## Overview
 
@@ -156,7 +156,43 @@ the factory nor `parseScopedCommandInput`; each can adopt the guard with a singl
   of a misleading `200`. This is the point of the change.
 - Responses gain an optional `ignoredFields` array. Absent anything to report, the
   response is byte-identical to today's.
-- No schema, migration or persisted-data changes.
+- No migration or persisted-data changes.
+
+**A previously-succeeding request can now return 400.** The additivity above is a
+statement about FIELDS, not about REQUESTS, and the difference is worth stating
+plainly. An aliased key now reaches `schema.parse`, so a snake_case value that does
+not satisfy the camelCase field's validator fails the whole request where it was
+previously dropped in silence:
+
+```
+PUT /api/customers/deals  {"id":"…","value_amount":"1 234,50"}
+before: 200, value_amount discarded
+after:  400, invalid_type
+```
+
+That is the intended behaviour of this change, not an accident: the endpoint either
+applies the field or refuses it. It is recorded here so the trade-off is a decision
+rather than a surprise. A straight round-trip is unaffected, because the read side
+emits values the write schema accepts.
+
+**`ActivityUpdateInput` is narrowed.** `activityUpdateSchema.omit({ entityId: true })`
+removes a field from an exported type. `BACKWARD_COMPATIBILITY.md` classifies that as
+breaking. Nothing that worked stops working at runtime, since the field was inert on
+this path, but the type is narrower for anyone importing it.
+
+**The programmatic path is not yet covered.** The guard fires on the HTTP route, so
+`customers.activities.update` invoked directly still discards `entityId` — Zod strips
+it now instead of `mapActivityUpdateInput` dropping it. Same outcome as before, so no
+regression, but the "no silent drops" promise does not hold there yet.
+
+**Deal write schemas now accept `null` on nullable columns.** Every `CustomerDeal`
+column that is `nullable: true` accepts `null` in the write schemas, meaning "clear
+it". This fixes a pre-existing defect that the aliasing would otherwise have exposed
+to far more callers: `z.coerce.number()` maps `null` to `0` and `z.coerce.date()` maps
+it to the epoch, so `{"value_amount": null}` from a round-tripped read wrote `0` and
+`1970-01-01` under a `200`. `customers.deals.update` already applied `?? null` for
+these fields; only the schema stood in the way, and `ownerUserId` had already been
+given `.nullable()` for the same reason (TC-CRM-069).
 
 ---
 
--- a/.ai/specs/2026-08-26-write-payload-silent-field-drop.md
+++ b/.ai/specs/2026-08-26-write-payload-silent-field-drop.md
@@ -97,9 +97,21 @@ One implementation, in `packages/shared/src/lib/crud/write-payload.ts`:
   an unknown key when the route opted into strictness.
 - `withIgnoredFieldsReport(payload, input)`: attaches `ignoredFields` to a response.
 
-Custom-field keys (`customFields`, `customValues`, `cf_*`, `cf:*`) are excluded
-before inspection. No write schema declares them, so without that a legitimate
-`cf_priority` would be reported as a field the endpoint ignored.
+Three classes of key are excluded before inspection, because no write schema
+declares them and reporting them would be noise rather than signal:
+
+- **Custom fields** (`customFields`, `customValues`, `cf_*`, `cf:*`), routed by
+  `splitCustomFieldPayload` further down. Without this a legitimate `cf_priority`
+  would be reported as ignored.
+- **Tenant and organization scope**, in both spellings. These come from trusted
+  context per `.ai/review-checklist.md` §4. Aliasing them would additionally let a
+  caller steer scope through the snake_case spelling that is ignored today.
+- **Server-maintained timestamps** (`created_at`, `updated_at`, `deleted_at` and
+  their camelCase spellings). List projections emit them, so without this every
+  round-trip write would report two ignored fields.
+
+`immutableFields` is enforced on update only. Those fields cannot change *after*
+creation, so the create path must accept exactly them.
 
 ### Four outcomes, none of them silent
 
--- a/.ai/specs/README.md
+++ b/.ai/specs/README.md
@@ -83,7 +83,7 @@ Specs awaiting implementation or partially complete. Focus here for actionable w
 | [SPEC-067](implemented/SPEC-067-2026-03-17-cli-standalone-app-support.md) | 2026-03-17 | CLI Standalone App Support | CLI tooling support for standalone app development |
 | [SPEC-068](SPEC-068-2026-03-02-use-case-examples-framework.md) | 2026-03-02 | Use-Case Examples Framework | `create-mercato-app --example` pattern for bootstrapping use-case solutions |
 | [SPEC-069](implemented/2026-02-23-core-timesheets.md) | 2026-02-23 | Core Timesheets Functionality | Core timesheets functionality in the `staff` module with My Timesheets, Projects, and phase-based approvals/policies |
-| [SPEC-072](SPEC-072-2026-08-26-write-payload-silent-field-drop.md) | 2026-08-26 | Customers Write API — Stop Accepting Fields and Discarding Them | Write payloads are aliased, reported or rejected, so a `200` no longer hides a discarded field |
+| [Write Payload Guard](2026-08-26-write-payload-silent-field-drop.md) | 2026-08-26 | CRUD Write Payloads: Stop Accepting Fields and Discarding Them (partial) | Write payloads are aliased, reported or rejected, so a `200` no longer hides a discarded field. Partial: 6 bespoke customers handlers have yet to adopt the guard |
 | [Checkout](2026-03-19-checkout-simple-checkout.md) | 2026-03-19 | Simple Checkout | Checkout flow specification for Phase B |
 | [Checkout Wireframes](2026-03-19-checkout-simple-checkout-wireframes.md) | 2026-03-19 | Simple Checkout Wireframes | Companion wireframes for the Simple Checkout spec |
 | [Registry](implemented/2026-03-20-decentralize-module-registry-generator.md) | 2026-03-20 | Decentralize Module Registry Generator | CLI refactoring for decentralized module registry generation |
--- a/.ai/specs/README.md
+++ b/.ai/specs/README.md
@@ -83,6 +83,7 @@ Specs awaiting implementation or partially complete. Focus here for actionable w
 | [SPEC-067](implemented/SPEC-067-2026-03-17-cli-standalone-app-support.md) | 2026-03-17 | CLI Standalone App Support | CLI tooling support for standalone app development |
 | [SPEC-068](SPEC-068-2026-03-02-use-case-examples-framework.md) | 2026-03-02 | Use-Case Examples Framework | `create-mercato-app --example` pattern for bootstrapping use-case solutions |
 | [SPEC-069](implemented/2026-02-23-core-timesheets.md) | 2026-02-23 | Core Timesheets Functionality | Core timesheets functionality in the `staff` module with My Timesheets, Projects, and phase-based approvals/policies |
+| [SPEC-072](SPEC-072-2026-08-26-write-payload-silent-field-drop.md) | 2026-08-26 | Customers Write API — Stop Accepting Fields and Discarding Them | Write payloads are aliased, reported or rejected, so a `200` no longer hides a discarded field |
 | [Checkout](2026-03-19-checkout-simple-checkout.md) | 2026-03-19 | Simple Checkout | Checkout flow specification for Phase B |
 | [Checkout Wireframes](2026-03-19-checkout-simple-checkout-wireframes.md) | 2026-03-19 | Simple Checkout Wireframes | Companion wireframes for the Simple Checkout spec |
 | [Registry](implemented/2026-03-20-decentralize-module-registry-generator.md) | 2026-03-20 | Decentralize Module Registry Generator | CLI refactoring for decentralized module registry generation |
--- a/.ai/specs/SPEC-072-2026-08-26-write-payload-silent-field-drop.md
+++ b/.ai/specs/SPEC-072-2026-08-26-write-payload-silent-field-drop.md
@@ -1,0 +1,180 @@
+# SPEC-072: Customers Write API — Stop Accepting Fields and Discarding Them
+
+## Overview
+
+`PUT /api/customers/deals` and `PUT /api/customers/activities` answer
+`200 {"ok":true}` while silently discarding part of the request body. A caller
+cannot tell "written" from "ignored" without re-reading the record, which is the
+opposite of what a `200` means.
+
+A write endpoint should either apply a field or reject it. These did neither.
+
+---
+
+## Problem Statement
+
+Two independent mechanisms produce the same symptom.
+
+### 1. Zod strips unknown keys, and the read side emits snake_case
+
+`dealUpdateSchema` is `z.object({ id }).merge(dealCreateSchema.partial())`. Zod
+strips unknown keys by default, so any key that is not spelled exactly as the
+schema declares it is dropped before the command ever sees it.
+
+The deal list endpoint emits **snake_case** (`closure_outcome`, `loss_notes`,
+`owner_user_id` — see the `list.fields` array in `api/deals/route.ts`), while the
+write schema declares **camelCase**. A caller that reads a deal and writes it back
+therefore sends keys that are silently discarded.
+
+Observed against a running instance:
+
+```
+PUT /api/customers/deals
+{"id":"…","status":"closed","closure_outcome":"lost","loss_notes":"…"}
+→ 200 {"ok":true}
+→ read back: status = "closed" (applied)
+             closure_outcome = null, loss_notes = "" (both discarded)
+```
+
+The asymmetry is what makes it dangerous: `status` is spelled identically in both
+directions, so part of the body lands and part vanishes, under one `200`.
+
+### 2. A field passes validation and is then dropped by a hand-built input
+
+`PUT /api/customers/activities` is a hand-written handler, not a `makeCrudRoute`.
+It parses with `activityUpdateBodySchema` — which is `.passthrough()` and, via
+`activityCreateSchema.partial()`, declared `entityId` — and then builds the
+command input from a fixed list of ten fields that does not include it.
+
+```
+PUT /api/customers/activities
+{"id":"…","entityId":"<different entity>", …}
+→ 200 {"ok":true}
+→ read back: entityId unchanged
+```
+
+`entityId` is **immutable by design** and that design is right: the canonical
+`interactionUpdateSchema` declares no `entityId`, and `customers.interactions.update`
+reads the owning entity off the stored record. There is no re-parent path. The bug
+is only that the request reports success instead of rejecting.
+
+The same hand-built list also drops `date`, `time` and `phoneNumber`, which
+`activityUpdateSchema` declares **and** the canonical interaction update accepts
+(it derives `scheduledAt` from `date`+`time`). Editing a call activity's phone
+number through this endpoint reported success and changed nothing.
+
+`mapActivityUpdateInput` in `commands/activities.ts` drops the same four fields on
+the programmatic path.
+
+---
+
+## Design
+
+### Guard at the layer that owns the parse
+
+`makeCrudRoute` funnels every ORM-backed write through two lines,
+`createConfig.schema.parse(body)` and `updateConfig.schema.parse(body)`. That is
+where the stripping happens, so that is where the guard goes. It reaches 77 route
+files across 19 modules without any of them opting in.
+
+Two other write paths exist and get the same guard:
+
+- **Command-backed actions** parse inside `mapInput`, usually through
+  `parseScopedCommandInput`, which now delegates to the same function.
+- **Hand-written handlers** that predate the factory call `guardWriteBody`
+  directly. `api/customers/activities` does.
+
+One implementation, in `packages/shared/src/lib/crud/write-payload.ts`:
+
+- `collectWritableKeys(schema)`: the top-level keys a schema accepts, unwrapping
+  the `ZodEffects` that `.superRefine()` / `.transform()` add. Returns `null` for a
+  shape it cannot introspect, which every caller treats as "leave it alone", so
+  behaviour is unchanged wherever introspection fails.
+- `inspectWritePayload(payload, keys, { immutableFields })`: renames snake_case
+  onto declared camelCase, classifies the rest as `unknown` or `immutable`.
+- `guardWriteBody(schema, body, config)`: the entry point every write path calls.
+  Throws `CrudHttpError(400)` for an ambiguous duplicate, an immutable field, and
+  an unknown key when the route opted into strictness.
+- `withIgnoredFieldsReport(payload, input)`: attaches `ignoredFields` to a response.
+
+Custom-field keys (`customFields`, `customValues`, `cf_*`, `cf:*`) are excluded
+before inspection. No write schema declares them, so without that a legitimate
+`cf_priority` would be reported as a field the endpoint ignored.
+
+### Four outcomes, none of them silent
+
+| Case | Behaviour |
+|---|---|
+| snake_case spelling of a declared field | **Applied** (aliased onto the camelCase key) |
+| Both spellings, different values | **400**, ambiguous, never guessed |
+| A real field that cannot change (`entityId`) | **400** "cannot be changed after creation" |
+| An unknown key | **Reported** as `ignoredFields`, or **400** with `rejectUnknownFields` |
+
+### Why aliasing is on by default and rejection is not
+
+Aliasing is safe to default on at this reach precisely because it is additive: it
+only ever applies keys Zod was already discarding, so no field that takes effect
+today changes behaviour. That property is what lets the guard sit at a chokepoint
+covering 19 modules rather than being wired per route.
+
+Rejecting unknown keys is **opt-in** per route (`writeGuard.rejectUnknownFields`).
+Widget injection routinely puts non-schema keys into form payloads, so flipping
+strictness on globally would break working forms. Reporting `ignoredFields` gives
+a caller something to assert on without that risk.
+
+---
+
+## Changes
+
+| File | Change |
+|---|---|
+| `packages/shared/src/lib/crud/write-payload.ts` | New. The guard, and the only implementation of it |
+| `packages/shared/src/lib/crud/factory.ts` | Guard applied to both parse chokepoints; new `writeGuard` option; create/update responses report `ignoredFields`; command `response` receives `input` |
+| `packages/shared/src/lib/api/scoped.ts` | `parseScopedCommandInput` delegates to the shared guard |
+| `packages/core/src/modules/customers/api/deals/route.ts` | Update response reports `ignoredFields` |
+| `packages/core/src/modules/customers/api/activities/route.ts` | Calls the guard directly; rejects `entityId`; forwards `date` / `time` / `phoneNumber` |
+| `packages/core/src/modules/customers/data/validators.ts` | `activityUpdateSchema` no longer declares `entityId` |
+| `packages/core/src/modules/customers/commands/activities.ts` | `mapActivityUpdateInput` forwards `date` / `time` / `phoneNumber` |
+
+### Coverage
+
+In the customers module, 9 of 15 `PUT` endpoints are now guarded, up from the 2
+the reported bugs live in. The remaining 6 are bespoke handlers (`pipelines`,
+`pipeline-stages`, the two `roles` routes, two `settings` routes) that use neither
+the factory nor `parseScopedCommandInput`; each can adopt the guard with a single
+`guardWriteBody` call. Framework-wide, every `makeCrudRoute` write inherits it.
+
+---
+
+## Backward Compatibility
+
+- Previously-discarded snake_case keys are now **applied**. This is the intended
+  fix, and it is the one behavioural change a caller could notice: a client that
+  was sending `closure_outcome` and relying on it being ignored would now write it.
+  No such client can exist deliberately, since the field never took effect.
+- `PUT /api/customers/activities` carrying `entityId` now returns **400** instead
+  of a misleading `200`. This is the point of the change.
+- Responses gain an optional `ignoredFields` array. Absent anything to report, the
+  response is byte-identical to today's.
+- No schema, migration or persisted-data changes.
+
+---
+
+## Testing
+
+- `packages/shared/src/lib/crud/__tests__/write-payload.test.ts` — key extraction
+  across `merge`/`partial`/`ZodEffects`/union, aliasing, conflicts, immutability.
+- `packages/shared/src/lib/api/__tests__/scoped.test.ts` — the guard through
+  `parseScopedCommandInput`, including that custom-field keys are left alone.
+- `packages/core/src/modules/customers/__tests__/write-payload-guard.test.ts` —
+  regression cover against the **real** `dealUpdateSchema` / `activityUpdateSchema`,
+  using the payloads observed on the running instance.
+
+---
+
+## Changelog
+
+- **2026-08-26** - Initial spec. Both cases reproduced against a deployed 0.6.7
+  instance and confirmed still present on `develop`. Guard placed at the
+  `makeCrudRoute` parse chokepoint so it covers every module, rather than being
+  wired into the two endpoints where the bugs were reported.
--- a/packages/core/src/modules/customers/__integration__/TC-CRM-WRITE-GUARD-001.spec.ts
+++ b/packages/core/src/modules/customers/__integration__/TC-CRM-WRITE-GUARD-001.spec.ts
@@ -1,0 +1,175 @@
+import { test, expect } from '@playwright/test';
+import { apiRequest, getAuthToken } from '@open-mercato/core/modules/core/__integration__/helpers/api';
+import {
+  createCompanyFixture,
+  createDealFixture,
+  deleteEntityIfExists,
+} from '@open-mercato/core/modules/core/__integration__/helpers/crmFixtures';
+
+/**
+ * TC-CRM-WRITE-GUARD-001: a write endpoint applies a field or refuses it.
+ * Spec: .ai/specs/2026-08-26-write-payload-silent-field-drop.md
+ *
+ * These endpoints used to answer `200 {"ok":true}` while discarding part of the
+ * body, so a caller could not tell "written" from "ignored" without reading the
+ * record back. Each case below asserts the outcome over the wire, then re-reads
+ * the record, because the response alone was exactly what could not be trusted.
+ */
+test.describe('TC-CRM-WRITE-GUARD-001: writes never silently discard a field', () => {
+  test('a deal update sent in the snake_case spelling the list emits is applied', async ({ request }) => {
+    let token: string | null = null;
+    let companyId: string | null = null;
+    let dealId: string | null = null;
+    const stamp = Date.now();
+
+    try {
+      token = await getAuthToken(request);
+      companyId = await createCompanyFixture(request, token, `QA WG001 Co ${stamp}`);
+      dealId = await createDealFixture(request, token, {
+        title: `QA WG001 Deal ${stamp}`,
+        companyIds: [companyId],
+      });
+
+      // Exactly the shape a caller gets back from GET /api/customers/deals.
+      const res = await apiRequest(request, 'PUT', '/api/customers/deals', {
+        token,
+        data: {
+          id: dealId,
+          status: 'closed',
+          closure_outcome: 'lost',
+          loss_notes: 'QA WG001 undercut on price',
+        },
+      });
+      expect(res.status()).toBe(200);
+
+      const detail = await apiRequest(request, 'GET', `/api/customers/deals?id=${dealId}`, { token });
+      expect(detail.status()).toBe(200);
+      const body = await detail.json();
+      const deal = Array.isArray(body?.items) ? body.items[0] : body;
+
+      expect(deal.closure_outcome ?? deal.closureOutcome).toBe('lost');
+      expect(deal.loss_notes ?? deal.lossNotes).toBe('QA WG001 undercut on price');
+    } finally {
+      await deleteEntityIfExists(request, token, '/api/customers/deals', dealId);
+      await deleteEntityIfExists(request, token, '/api/customers/companies', companyId);
+    }
+  });
+
+  test('a deal update naming a field the endpoint cannot write reports it back', async ({ request }) => {
+    let token: string | null = null;
+    let companyId: string | null = null;
+    let dealId: string | null = null;
+    const stamp = Date.now();
+
+    try {
+      token = await getAuthToken(request);
+      companyId = await createCompanyFixture(request, token, `QA WG001b Co ${stamp}`);
+      dealId = await createDealFixture(request, token, {
+        title: `QA WG001b Deal ${stamp}`,
+        companyIds: [companyId],
+      });
+
+      const res = await apiRequest(request, 'PUT', '/api/customers/deals', {
+        token,
+        data: { id: dealId, status: 'open', not_a_deal_field: 'x' },
+      });
+      expect(res.status()).toBe(200);
+
+      const payload = await res.json();
+      expect(payload.ignoredFields).toEqual(
+        expect.arrayContaining([expect.objectContaining({ key: 'not_a_deal_field', reason: 'unknown' })]),
+      );
+    } finally {
+      await deleteEntityIfExists(request, token, '/api/customers/deals', dealId);
+      await deleteEntityIfExists(request, token, '/api/customers/companies', companyId);
+    }
+  });
+
+  // An interaction's owning entity is fixed at creation, so this must be refused
+  // rather than accepted and ignored.
+  test('an activity update carrying entityId is refused with 400', async ({ request }) => {
+    let token: string | null = null;
+    let companyId: string | null = null;
+    let otherCompanyId: string | null = null;
+    let activityId: string | null = null;
+    const stamp = Date.now();
+
+    try {
+      token = await getAuthToken(request);
+      companyId = await createCompanyFixture(request, token, `QA WG001c Co ${stamp}`);
+      otherCompanyId = await createCompanyFixture(request, token, `QA WG001c Other ${stamp}`);
+
+      const created = await apiRequest(request, 'POST', '/api/customers/activities', {
+        token,
+        data: {
+          entityId: companyId,
+          activityType: 'note',
+          subject: `QA WG001c ${stamp}`,
+          body: 'created for the write-guard check',
+        },
+      });
+      expect(created.status()).toBe(201);
+      activityId = (await created.json())?.id ?? null;
+      expect(activityId).toBeTruthy();
+
+      const res = await apiRequest(request, 'PUT', '/api/customers/activities', {
+        token,
+        data: { id: activityId, entityId: otherCompanyId, subject: `QA WG001c moved ${stamp}` },
+      });
+
+      expect(res.status()).toBe(400);
+      const payload = await res.json();
+      expect(payload.fields).toContain('entityId');
+    } finally {
+      await deleteEntityIfExists(request, token, '/api/customers/activities', activityId);
+      await deleteEntityIfExists(request, token, '/api/customers/companies', companyId);
+      await deleteEntityIfExists(request, token, '/api/customers/companies', otherCompanyId);
+    }
+  });
+
+  // `date`, `time` and `phoneNumber` passed validation and were then left out of
+  // the hand-built command input, so the edit reported success and changed nothing.
+  test('an activity phone-number edit now persists', async ({ request }) => {
+    let token: string | null = null;
+    let companyId: string | null = null;
+    let activityId: string | null = null;
+    const stamp = Date.now();
+    const phone = '+48221234567';
+
+    try {
+      token = await getAuthToken(request);
+      companyId = await createCompanyFixture(request, token, `QA WG001d Co ${stamp}`);
+
+      const created = await apiRequest(request, 'POST', '/api/customers/activities', {
+        token,
+        data: {
+          entityId: companyId,
+          activityType: 'call',
+          subject: `QA WG001d ${stamp}`,
+          phoneNumber: '+48229999999',
+        },
+      });
+      expect(created.status()).toBe(201);
+      activityId = (await created.json())?.id ?? null;
+      expect(activityId).toBeTruthy();
+
+      const res = await apiRequest(request, 'PUT', '/api/customers/activities', {
+        token,
+        data: { id: activityId, activityType: 'call', phoneNumber: phone },
+      });
+      expect(res.status()).toBe(200);
+
+      const detail = await apiRequest(request, 'GET', `/api/customers/activities?id=${activityId}`, { token });
+      expect(detail.status()).toBe(200);
+      const body = await detail.json();
+      const activity = Array.isArray(body?.items)
+        ? body.items.find((item: { id?: string }) => item?.id === activityId) ?? body.items[0]
+        : body;
+
+      expect(activity?.phoneNumber ?? activity?.phone_number).toBe(phone);
+    } finally {
+      await deleteEntityIfExists(request, token, '/api/customers/activities', activityId);
+      await deleteEntityIfExists(request, token, '/api/customers/companies', companyId);
+    }
+  });
+});
--- a/packages/core/src/modules/customers/__tests__/write-payload-guard.test.ts
+++ b/packages/core/src/modules/customers/__tests__/write-payload-guard.test.ts
@@ -1,0 +1,88 @@
+import { parseScopedCommandInput } from '@open-mercato/shared/lib/api/scoped'
+import { collectWritableKeys, inspectWritePayload } from '@open-mercato/shared/lib/crud/write-payload'
+import { activityUpdateSchema, dealUpdateSchema } from '../data/validators'
+
+const translate = (_key: string, fallback?: string) => fallback ?? _key
+const ctx = {
+  auth: { tenantId: '11111111-1111-4111-8111-111111111111', orgId: '22222222-2222-4222-8222-222222222222' },
+  selectedOrganizationId: '22222222-2222-4222-8222-222222222222',
+} as any
+
+const DEAL_ID = 'b55529cc-364a-466a-9134-d6a01a2380c5'
+const ACTIVITY_ID = 'e356a1cf-8ec9-4c8c-9d9b-491436dd03a0'
+const OTHER_ENTITY_ID = '04d233d7-e747-4789-a928-33f272c467fc'
+
+// The deal list endpoint emits snake_case (`closure_outcome`, `loss_notes`,
+// `owner_user_id`). A caller that reads a deal and writes it back therefore sends
+// snake_case, which Zod stripped — the update answered 200 having applied `status`
+// and discarded the other two.
+describe('deal update accepts the spelling the read endpoints emit', () => {
+  it('applies closure_outcome and loss_notes instead of dropping them', () => {
+    const parsed = parseScopedCommandInput(
+      dealUpdateSchema,
+      {
+        id: DEAL_ID,
+        status: 'closed',
+        closure_outcome: 'lost',
+        loss_notes: 'client went with an incumbent vendor',
+      },
+      ctx,
+      translate
+    ) as Record<string, unknown>
+
+    expect(parsed.status).toBe('closed')
+    expect(parsed.closureOutcome).toBe('lost')
+    expect(parsed.lossNotes).toBe('client went with an incumbent vendor')
+    expect(parsed.ignoredFields).toBeUndefined()
+  })
+
+  it('aliases owner_user_id, which the list endpoint also emits', () => {
+    const parsed = parseScopedCommandInput(
+      dealUpdateSchema,
+      { id: DEAL_ID, owner_user_id: OTHER_ENTITY_ID },
+      ctx,
+      translate
+    ) as Record<string, unknown>
+
+    expect(parsed.ownerUserId).toBe(OTHER_ENTITY_ID)
+  })
+
+  it('names any field it will not write instead of answering a bare ok', () => {
+    const parsed = parseScopedCommandInput(
+      dealUpdateSchema,
+      { id: DEAL_ID, status: 'closed', not_a_deal_field: 'x' },
+      ctx,
+      translate
+    ) as Record<string, unknown>
+
+    expect(parsed.status).toBe('closed')
+    expect(parsed.ignoredFields).toEqual([{ key: 'not_a_deal_field', reason: 'unknown' }])
+  })
+})
+
+// An interaction's owning entity is fixed at creation: `interactionUpdateSchema`
+// declares no `entityId`, so the activities adapter had no way to apply one and
+// reported success anyway.
+describe('activity update refuses to pretend it re-parented a record', () => {
+  it('no longer declares entityId as writable', () => {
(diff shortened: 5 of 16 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-26T23:37:20Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(crud): stop accepting write fields and discarding them silently

## 🎯 Summary

This PR fixes a real and well-diagnosed class of bug: `PUT /api/customers/deals` and `PUT /api/customers/activities` answered `200 {"ok":true}` while discarding part of the request body, because Zod strips unknown keys and the read endpoints emit snake_case while the write schemas declare camelCase. The diagnosis is excellent — both mechanisms are identified precisely, the `entityId` immutability argument is correct (`interactionUpdateSchema` really does declare no such field, and `customers.interactions.update` really does read the owning entity off the stored record), and forwarding `date` / `time` / `phoneNumber` through `mapActivityUpdateInput` closes a genuine second drop. The new `packages/shared/src/lib/crud/write-payload.ts` is careful, well-commented code with a deliberate fail-open on schemas it cannot introspect, and the unit tests exercise the real customers schemas rather than toy ones.

The concern is placement and blast radius. The guard is installed at three chokepoints — `makeCrudRoute`'s create and update parses (77 route files across 19 modules), `parseScopedCommandInput` (169 call sites across `core`, `enterprise` and the provider packages), and one hand-written handler — with aliasing on by default. The PR and the spec both rest on the claim that this is "strictly additive: it only ever applies keys Zod was already discarding, so no field that takes effect today changes behaviour". I could not reproduce that guarantee: I found two paths where a request that returns `200` today returns `400` after this change, one of them on the exact caller the PR is designed to serve. Details in ⚠️ Major below, with the probes I used.

Reviewed scope: all 13 changed files, plus the surrounding factory create/update/command paths, `withScopedPayload`, `splitCustomFieldPayload`, the customers deal and activity validators and commands, and the OpenAPI CRUD builder.

Validation source: **local fallback**. This is a fork PR (`cielecki:fix/customers-write-silent-field-drop`) and GitHub has no usable check data for head `fcb715f` — both `CI for Develop&Main` and `Mutation tests` sit at `action_required` awaiting maintainer approval, and the only reported check (`[code]smith`) is `skipping`. I therefore ran the full `validation.commands` gate locally (`Runner: local` — no compose `app` container was running).

## Verdict

❌ **request changes** — a configured validation-gate command (`yarn i18n:check-usage`) fails on this head with three missing translation keys introduced by the new file, and the aliasing default converts at least two classes of currently-succeeding request into a `400`, contradicting the additivity claim the design rests on.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `yarn generate` | ✅ PASS | Regenerated `packages/ui/src/backend/icons/lucideRegistry.generated.tsx` (adds the `bot` icon). This drift is pre-existing on `develop` and unrelated to this PR — I reverted it rather than counting it against the change. |
| `yarn build:packages` (second run) | ✅ PASS | |
| `yarn i18n:check-sync` | ✅ PASS | All translation files in sync across `en, pl, es, de, ko`. |
| `yarn i18n:check-usage` | ❌ **FAIL** (exit 1) | `MISSING KEYS (referenced in code but not in any en.json): 3 keys` — `errors.conflicting_field`, `errors.immutable_field`, `errors.unknown_field`, all at `packages/shared/src/lib/crud/write-payload.ts:225/234/242`. See ⛔ Blocker 1. |
| `yarn typecheck` | ✅ PASS | Clean across the workspace. |
| `yarn test` | ⚠️ FAIL, unrelated | `Tasks: 32 successful, 34 total`; the only failing task is `create-mercato-app#test`, where four `template-dev-log-files.test.ts` cases abort with `❌ Linux file-watch limits are too low for Turbopack` (`fs.inotify.max_user_watches: 253174 < 4194304`). That is a host sysctl limit on my machine, not a defect in this diff — the PR touches nothing under `packages/create-app`. Every other workspace is green, including `@open-mercato/core`, `@open-mercato/shared`, `@open-mercato/ui`, `@open-mercato/cli` and `@open-mercato/ai-assistant`. |

## Findings

### ⛔ Blocker

**1. `packages/shared/src/lib/crud/write-payload.ts:225`, `:234`, `:242` — the three new error keys have no dictionary entry, so `yarn i18n:check-usage` fails and CI will be red.**

`guardWriteBody` calls `t('errors.conflicting_field', …)`, `t('errors.immutable_field', …)` and `t('errors.unknown_field', …)` with literal key arguments. The static scanner in `scripts/i18n-check-usage.ts` picks those up, finds no `en.json` anywhere that declares them, and exits `1` — and `yarn i18n:check-usage` is the fifth entry in `validation.commands` in `.ai/agentic.config.json`, so the gate stops there. This is the one item that is unambiguously blocking regardless of how the design questions below are resolved.

The PR description says these follow "the existing `errors.*` pattern in `parseScopedCommandInput`, which carries inline English fallbacks and has no dictionary entries". That is true of the *values* but not of the *detection*: `parseScopedCommandInput`'s keys live inside the `DEFAULT_MESSAGES` object literal, which the scanner does not treat as a `t()` call, so they were never flagged. Every real dictionary in the repo namespaces these by module (`customers.errors.tenant_required`, `sales.errors.tenant_required`, `catalog.errors.tenant_required`, …).

Fix: either give the three keys real entries in a dictionary the shared package can be documented against, or — cleaner given that `@open-mercato/shared` owns no locale files — take the message keys as configuration the way `ScopedPayloadMessages` already does, so each caller supplies its own module-namespaced key and the shared file carries no literal key at all. Whichever route you pick, `yarn i18n:check-usage` must exit `0` before this can merge.

### ⚠️ Major

**2. `packages/shared/src/lib/api/scoped.ts:116-146` — the guard inspects the payload *after* `withScopedPayload` has injected `organizationId`/`tenantId`, so a round-tripped record from another organization now gets a `400` instead of being written.**

`parseScopedCommandInput` calls `withScopedPayload` first (line 116), which resolves the organization from `source.organizationId ?? ctx.selectedOrganizationId ?? ctx.auth.orgId` and writes it onto the payload. Only then does `guardWriteBody` run (line 140). By that point the body carries a synthesized `organizationId` that the caller never sent, and if the caller's round-tripped body also carries the `organization_id` the list endpoint emitted, `inspectWritePayload` sees both spellings with different values and raises the ambiguity `400`.

That happens on the ordinary "all organizations" selection, where `ctx.selectedOrganizationId` is `null` and the fallback picks the actor's own default org rather than the record's. I reproduced it against the real schema:

```ts
const allOrgsCtx = { auth: { tenantId: TENANT, orgId: ORG_A }, selectedOrganizationId: null }
parseScopedCommandInput(dealUpdateSchema, { id: DEAL, status: 'open', organization_id: ORG_B, tenant_id: TENANT }, allOrgsCtx, translate)
// → CrudHttpError 400
// {"error":"This field was sent twice with different values.",
//  "fields":["organization_id"],
//  "details":"organization_id and organizationId were both sent with different values."}
```

Before this PR the same request succeeded: `organization_id` was stripped by Zod and the scope came from context, which is exactly what `.ai/review-checklist.md` §4 prescribes ("handlers derive scope from trusted context and ignore same-named payload fields"). The new behaviour is worse than the old one here — the caller is told their body is ambiguous about a field they are not supposed to control, and it lands on the deals endpoint this PR exists to fix, via the chokepoint 169 call sites share.

Fix: run the guard on the caller's payload before scope injection, or exclude the injected scope keys from inspection the way custom-field keys are already excluded — add `tenantId` / `organizationId` (and their snake spellings) to the skip set in `guardWriteBody`, since they are never caller-authoritative on this path.

**3. `packages/shared/src/lib/crud/write-payload.ts:99-145` — aliasing is not outcome-preserving: a value that Zod never used to see can now fail validation and turn a `200` into a `400`.**

The additivity argument covers *fields* but not *requests*. An aliased key is handed to `schema.parse`, so if the snake_case value does not satisfy the camelCase field's validator, the whole request now fails where it previously succeeded with the key silently dropped:

```ts
parseScopedCommandInput(dealUpdateSchema, { id: DEAL, value_amount: '1 234,50' }, ctx, translate)
// → ZodError: expected "number", code "invalid_type"  →  400
// before this PR: 200, value_amount discarded
```

`dealCreateSchema.valueAmount` is `z.coerce.number().min(0)`, and any caller that was echoing a formatted or otherwise non-coercible `value_amount` back was previously unaffected. The same applies to every `z.coerce.date()`, `uuid()` and enum field across the 19 modules the guard now covers, on both the factory and the command path.

I am not arguing the aliasing is wrong — round-tripping the read shape is a legitimate goal and the failure is at least loud rather than silent. I am arguing that the risk is larger than the PR and `SPEC-072` currently state, that "no field that takes effect today changes behaviour" should not be the sentence a maintainer merges on, and that a change with this reach deserves either a narrower default (opt-in per route, or on for the two endpoints where the bug was reported) or an explicit statement of this failure mode in the spec's Backward Compatibility section so the risk is a decision rather than a surprise. Your call which — but please make it explicitly.

**4. `packages/shared/src/lib/crud/write-payload.ts:247-254` — the documented `aliasSnakeCaseKeys: false` opt-out reintroduces exactly the silent drop this PR exists to remove.**

When aliasing is disabled, `guardWriteBody` returns the original `body` but computes `ignoredFields` from `inspection`, where the snake_case key has already been consumed into `aliased` rather than `unwritable`. The key is therefore neither applied (the original body still spells it snake_case, so Zod strips it) nor reported. I confirmed this against the real activities body schema:

```ts
guardWriteBody(activityUpdateBodySchema, { id: 'x', phone_number: '+48123123123' }, { aliasSnakeCaseKeys: false })
// → { body: { id: 'x', phone_number: '+48123123123' }, ignoredFields: [], aliased: [] }
```

No route sets the flag today, so this is latent rather than live — but it is a trap laid in a public option, and if reviewers push back on the aliasing default (finding 3) this is exactly the flag they will reach for. Fix: when aliasing is off, classify a snake_case key whose camelCase spelling the schema declares as `unwritable` with reason `unknown` (or a new `misspelled` reason) so it is still reported.

### 🔹 Minor

**5. `packages/shared/src/lib/crud/factory.ts:2502` and `:2853` — the new `ignoredFields` response field and the new `400` error class are not reflected in the OpenAPI contracts they apply to.**

Both the create and update paths now attach `ignoredFields` to responses whose declared schema in each route's `openApi` export (`create.responseSchema` / `update.responseSchema`, defaulting to `fallbackCreateResponseSchema` / `fallbackOkResponseSchema` in `packages/shared/src/lib/openapi/crud.ts`) does not mention it, and `guardWriteBody` introduces a `400` shape (`error`, `fields`, `details`) that no route's `errors` array declares. That is the same shape of drift the repo already recorded for command-interceptor statuses in `BACKWARD_COMPATIBILITY.md` § "Command Interceptor HTTP Status (2026-08-06)": the runtime widened, the documented contract did not. Fix both halves in one pass — extend the fallback response schemas with an optional `ignoredFields` array and register the guard's `400` in the CRUD OpenAPI builder's default error set, so generated API docs stay truthful for all 77 factory routes.

**6. No integration coverage ships with the change, although it alters live API behaviour on two endpoints.**

Root `AGENTS.md` states that for every new feature the spec MUST list integration coverage for all affected API paths and key UI paths, and that those tests MUST ship in the same change; `.ai/qa/AGENTS.md` says the same. This PR carries a spec, converts a `200` into a `400` on `PUT /api/customers/activities`, and starts applying fields that were previously inert on `PUT /api/customers/deals` — both are exactly the "external contract" category that earns integration coverage. The checklist item is honestly left unchecked in the PR body, which I appreciate, but it is a repo hard rule rather than a preference. Please add specs under `.ai/qa/tests/` covering at least: a deal update sent in the snake_case spelling the list emits, an activity update carrying `entityId` (expect `400`), and an activity phone-number edit that now actually persists.

**7. `.ai/specs/SPEC-072-2026-08-26-write-payload-silent-field-drop.md` — the filename uses a prefix the repo explicitly bans for new specs.**

`.ai/specs/AGENTS.md` says under **Never**: "Never introduce new `SPEC-*` or `SPEC-ENT-*` filename prefixes", and the File Naming Convention section repeats it ("Legacy numbered filenames may remain in the repo until they are intentionally normalized, but new specs MUST NOT introduce `SPEC-*` … prefixes"). The existing `SPEC-0NN-` files are the legacy set the rule grandfathers, not a pattern to follow. Rename to `2026-08-26-write-payload-silent-field-drop.md` and update the `.ai/specs/README.md` row to match.

**8. `packages/core/src/modules/customers/data/validators.ts:242` — narrowing `activityUpdateSchema` narrows the exported `ActivityUpdateInput` type, and the spec's Backward Compatibility section does not say so.**

`.omit({ entityId: true })` removes a field from a schema that is exported from `data/validators.ts` and whose inferred type is re-exported as `ActivityUpdateInput` (line 653). `BACKWARD_COMPATIBILITY.md`'s quick-reference table classifies removing a field from a type interface as breaking, and the deprecation protocol asks for a documented migration note. I am deliberately *not* raising this to blocker: `ActivityUpdateInput` is not among the enumerated core types in § 2, and the removed field was inert, so nothing that worked stops working at runtime. But the spec's Backward Compatibility section lists only the `400` and the `ignoredFields` addition — please add the type narrowing there so the change is on the record.

Related and worth a sentence in the spec: the guard only fires on the HTTP route, so the programmatic path (`customers.activities.update`, `commands/activities.ts:226`) still silently discards `entityId` — Zod now strips it instead of `mapActivityUpdateInput` dropping it. Same outcome as before, so no regression, but the "no silent drops" promise does not yet hold on that path.

### 💅 Nit

**9. `packages/shared/src/lib/crud/write-payload.ts:127` — conflict detection compares by reference.**

`if (source[camel] !== source[key]) conflicts.push(...)` will report a conflict for two structurally identical arrays or objects sent under both spellings (`tagIds: ['a']` and `tag_ids: ['a']` are `!==`), producing a `400` that tells the caller the values differ when they do not. A shallow structural comparison for arrays and plain objects would close it.

**10. `packages/shared/src/lib/api/scoped.ts:176-183` — `ignoredFields` is merged into the parsed command input at all 169 `parseScopedCommandInput` call sites.**

It therefore reaches `commandBus.execute`'s `input`, the mutation-guard `mutationPayload`, and the sync-event payloads, and can end up in action-log snapshots — a key no command schema declares. It is harmless today because every command re-parses with its own schema, but carrying the report out-of-band (a second return value, or a `Symbol` key) would keep command inputs clean.

**11. `.ai/specs/README.md:86` — SPEC-072 is added to the "Specs awaiting implementation or partially complete" table by the PR that implements it.**

Given the acknowledged partial coverage (9 of 15 customers `PUT` endpoints), "partially complete" is arguably accurate — but say so in the row, or move it to `implemented/` with a note about the six bespoke handlers still to adopt the guard.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `write-payload.ts` is new and the `writeGuard` / `ScopedPayloadOptions` additions are optional.
- [x] No function signature changed in a breaking way — `CrudCommandActionConfig.response` gains `input` inside its existing args object (`factory.ts:486`), which is additive for every existing callback.
- [ ] No required type field removed or narrowed — `activityUpdateSchema` drops `entityId`, narrowing the exported `ActivityUpdateInput`. See finding 8; runtime behaviour is unaffected, but it is undocumented in the spec.
- [ ] No field removed or retyped in an existing response shape — nothing is removed, but responses gain an undeclared `ignoredFields` and routes gain an undeclared `400`. See finding 5.
- [x] No database table or column renamed or removed — the diff contains no migration and none is needed.
- [ ] No config key renamed and no default changed silently — `aliasSnakeCaseKeys` defaults to on, which changes write behaviour for 77 factory routes and 169 command call sites. The default is documented in the code and the spec, so it is not *silent*; findings 2 and 3 are about whether it is *safe*.
- [ ] Where a contract had to change, the old surface kept working through a deprecation window — no bridge exists for the `activityUpdateSchema` narrowing, and none for callers relying on a previously-ignored snake_case key being ignored.

## 🧪 Test Coverage

The unit coverage is genuinely good and I want to say so plainly. `write-payload.test.ts` pins key extraction through `merge`, `partial`, the effects wrapper and the un-introspectable union (asserting the `null` fail-open, which is the property the whole design leans on), plus aliasing, the agreeing-duplicate case, the disagreeing-duplicate conflict, and immutability both with and without an introspectable schema. `crud-factory.test.ts` drives six real `PUT` requests through `makeCrudRoute` rather than testing the helper in isolation, including that `cf_*` keys survive untouched. `write-payload-guard.test.ts` asserts against the actual `dealUpdateSchema` and `activityUpdateSchema` using the payloads observed on the running instance, which is the right instinct for a regression test. `scoped.test.ts` covers the `parseScopedCommandInput` path. All of these pass locally.

The gaps that map to the findings above:

- **No test pins the interaction between `withScopedPayload` and the guard.** Add a case to `packages/core/src/modules/customers/__tests__/write-payload-guard.test.ts` that round-trips a deal row carrying `organization_id` under `selectedOrganizationId: null` and asserts the request still succeeds — it currently throws (finding 2).
- **No test covers the `aliasSnakeCaseKeys: false` branch.** Add a case to `packages/shared/src/lib/crud/__tests__/write-payload.test.ts` asserting that with aliasing off a declared field's snake spelling is still reported in `ignoredFields` (finding 4).
- **No test asserts that aliasing cannot convert a success into a validation failure**, or, if that is accepted behaviour, pins it deliberately with a comment explaining the trade-off (finding 3).
- **No integration coverage** under `.ai/qa/tests/` for either changed endpoint (finding 6).
- **`create` is untested on the customers side.** `guardWriteBody` now runs on `createConfig.schema` too, but every customers-level test targets update paths; one create case through a real route would close that.

pkarw:
## 🏷️ label rationale

| Label | Why |
|-------|-----|
| `changes-requested` | The review found one blocker (`yarn i18n:check-usage` fails on this head with three missing keys introduced by `write-payload.ts`) plus two major findings on the aliasing default. This repo also requests changes on Medium findings. |
| `bug` | The PR's stated purpose is fixing silent field drops on two live write endpoints, even though the fix lands as a framework-level guard. |
| `priority-medium` | The underlying defect is a real data-integrity problem — a `200` that discarded part of the body — but it is long-standing and has a workaround (read the record back), so it is not urgent. |
| `risk-high` | The guard sits on `makeCrudRoute`'s create and update parse chokepoints (77 route files across 19 modules) and on `parseScopedCommandInput` (169 call sites), with aliasing enabled by default. I reproduced two cases where a request that returns `200` today returns `400` after this change, so the blast radius is wider than the "strictly additive" framing suggests. The author suggested `risk-medium`; I went one step up for that reason and am happy to be argued down once findings 2 and 3 are resolved. |
| `needs-qa` | The automated-verification exemption does not apply: the diff touches no UI-rendering file and adds no migration, but it **does** change the API surface (a new `400` error class and a new `ignoredFields` response property on every guarded write). |

`skip-qa` was deliberately not applied, and the `qa` pipeline label is left for a QA reviewer to set.

pkarw:
🤝 Handing this back to @cielecki for the requested changes.

The one item that blocks merge on its own is `yarn i18n:check-usage` — it exits `1` on this head because `errors.conflicting_field`, `errors.immutable_field` and `errors.unknown_field` have no dictionary entry anywhere. It is the fifth command in `validation.commands`, so CI stops there.

The two major findings both concern the aliasing default rather than the fix itself: `parseScopedCommandInput` inspects the payload *after* `withScopedPayload` has injected `organizationId`, which turns a cross-organization round-trip into a `400`; and an aliased value that fails the camelCase field's validator now fails the whole request where the key was previously discarded. Reproductions for both are in the review.

Worth saying plainly: the diagnosis in this PR is very good, the `entityId` immutability argument checks out against `interactionUpdateSchema`, and the unit tests exercise the real schemas. The pushback is about how wide the default reach should be, not about whether the bug is real.

Note that CI has not run on this head — both workflow runs sit at `action_required` because it is a fork PR. I ran the full gate locally instead; results are in the 🧪 Validation Gate table of the review.

`--autofix` was not requested and this is not my PR, so no fixes were pushed to your branch: **autofix: skipped (not my PR — re-run with --autofix to fix it here)**.

pkarw:

One blocker (failing `yarn i18n:check-usage` gate), two majors on the aliasing default, four minors and three nits. Validation ran locally because this fork PR has no approved CI run for head `fcb715f`. No autofix iterations — another author's PR without `--autofix`. No CI follow-up is owed, so `ci-monitoring` was not applied.

cielecki:
Thanks for the review. The worst of it was a regression I introduced and a claim I should not have made: the regression is fixed, the claim is corrected, and the rest is below.

## The blocker

`yarn i18n:check-usage` now exits `0`. I took your second suggestion. `@open-mercato/shared` owns no locale files, so the package now carries no literal key at all. `CrudWriteGuardConfig` has an optional `messages` field in the same shape `ScopedPayloadOptions.messages` already uses, and callers that want translated text pass their own module-namespaced keys. Callers that pass nothing get the English fallbacks verbatim, which today is all of them.

Your point about *values* versus *detection* was the part I had wrong, and the scanner made it twice: first on the three keys you found, then on my fix for them, which left the phrase `t('errors.x')` inside an explanatory comment and was reported as a fourth missing key.

## Finding 2, the regression

Fixed, and you were right that it was the worst of the four you raised: it broke a working request on the endpoint this PR exists to serve. Scope keys now sit outside the guard, in both spellings, alongside the custom-field keys.

Excluding them rather than reordering was deliberate. Running the guard before scope injection would alias `organization_id` onto `organizationId`, which `withScopedPayload` reads first, so a caller could steer scope through the snake_case spelling that is ignored today. Exclusion keeps `.ai/review-checklist.md` §4 intact. A test covers it: it round-trips a body carrying `organization_id` under `selectedOrganizationId: null` and asserts the scope still comes from context.

## Finding 3, and what it led to

I have not narrowed the default. A `400` on input that was never applied is the behaviour this PR argues for, and I read your review as making the same point. What was wrong was the sentence I wrote about it, so `Backward Compatibility` in the spec now states the failure mode plainly, with the `value_amount` example, and calls it a deliberate trade-off. A straight round-trip is unaffected, because the read side emits values the write schema accepts.

Chasing your probe turned up something worse than the `400`, which neither of us had flagged:

```
PUT /api/customers/deals  {"id":"…","value_amount":null,"expected_close_at":null}
→ 200, and the deal's value becomes 0 with a close date of 1970-01-01
```

`z.coerce.number()` maps `null` to `0` and `z.coerce.date()` maps it to the epoch. So reading a deal with no value set and writing the record straight back silently corrupts two columns, under a `200`, on exactly the round-trip this PR is meant to serve.

It is a pre-existing defect rather than one the aliasing created: a caller sending camelCase `valueAmount: null` gets `0` written on `develop` today. The aliasing would have routed far more traffic into it. Every `CustomerDeal` column that is `nullable: true` now accepts `null` in the write schemas, meaning "clear it". `customers.deals.update` already handles that (`if (parsed.x !== undefined) record.x = parsed.x ?? null`), and `ownerUserId` already carries `.nullable()`, added after the bulk owner-update worker hit the rejecting half of the same problem (TC-CRM-069). `status` stays non-nullable, because its column is.

Happy to split that into its own PR if you would rather review it separately.

## Finding 4

Fixed. `UnwritableKey.reason` now has `misspelled`, so with aliasing off the guard reports a declared field's snake spelling instead of dropping it. Tested in both directions.

## Minors

- **5, OpenAPI.** `ignoredFields` is now declared on the default create and update response schemas, and the guard's `400` is registered in the CRUD builder's error set, so it shows up on all 77 factory routes. `CrudMethodConfig` now has the `errors` field `CrudDeleteConfig` already had.
- **6, integration coverage.** Added as `TC-CRM-WRITE-GUARD-001.spec.ts`, covering a snake_case deal update, the `ignoredFields` report, the `entityId` `400`, and a phone-number edit that now persists. One thing to flag: you asked for these under `.ai/qa/tests/`, but `.ai/qa/AGENTS.md` says under **Never** that no executable `.spec.ts` may live there, and that module `__integration__` folders are preferred, so that is where they are. Tell me if I have read that backwards.
- **7.** Spec renamed to `2026-08-26-write-payload-silent-field-drop.md`, README row updated.
- **8.** The `ActivityUpdateInput` narrowing and the still-uncovered programmatic path are both recorded in `Backward Compatibility` now.
- **9.** Conflict detection compares arrays and plain objects structurally, so identical values under both spellings no longer raise a false `400`.
- **10.** The report rides on a `Symbol` rather than a string key, so it no longer reaches command inputs, mutation-guard payloads, sync events or action-log snapshots.
- **11.** The README row now says "partial" and names the six bespoke handlers still to adopt the guard.

Agreed on `risk-high`; my `risk-medium` was the wrong read.

## Validation

- `yarn i18n:check-usage`: `0`
- `yarn i18n:check-sync`: `0`
- `packages/shared` typecheck: clean
- `packages/shared`: 335 tests green across `lib/crud` and `lib/api`
- `packages/core`, full suite: 11617 passed, 2 failed, 1454 of 1455 suites

The core failure profile is unchanged by this round: the same 16 suites fail before and after, and they reproduce with my changes stashed. Fourteen fail to load on `#generated/*`, and the two `catalog` enrichers-config suites reach the same gap inside the test body, which is where the 2 failing tests come from. `packages/shared` has two more of the same kind, where `principal-service export map` asserts that built `dist` files exist. All of them are an unbuilt checkout on my side.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T11:47:01Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Re-review: fix(crud): stop accepting write fields and discarding them silently

## 🎯 Summary

Every one of the eleven findings from the first pass is genuinely addressed, and two of the responses are better than what I asked for. The blocker is fixed the way I'd have picked: `@open-mercato/shared` now carries no literal message key at all, and `CrudWriteGuardConfig.messages` mirrors the shape `ScopedPayloadOptions.messages` already uses. Finding 2 — the regression I was most worried about, because it broke a working request on the endpoint this PR exists to serve — is fixed by *excluding* scope keys rather than reordering the guard, and your reasoning for that choice is correct and better than my suggestion: running the guard before `withScopedPayload` would have aliased `organization_id` onto `organizationId`, handing a caller a way to steer scope through a spelling that is ignored today. `.ai/review-checklist.md` §4 stays intact, and a test pins it.

Chasing my finding-3 probe into the `z.coerce.number()`/`z.coerce.date()` null coercion is the most valuable thing in this round. `{"value_amount": null}` writing `0` and `{"expected_close_at": null}` writing `1970-01-01`, both under a `200`, on the exact round-trip this PR is designed to serve, is a live data-corruption path on `develop` today. I checked every field you widened against `CustomerDeal`: all twelve map to `nullable: true` columns, `status` is correctly left non-nullable because its column is `default 'open'` and NOT NULL, and `customers.deals.update` already applies `?? null` for each one (`commands/deals.ts:762-807`), so the schema really was the only thing in the way. Keeping it in this PR is the right call — the aliasing would have routed far more traffic into it, which makes it this PR's problem rather than a separate one.

**One correction to my own review: finding 6 named the wrong directory and you were right to push back.** `.ai/qa/AGENTS.md` says under **Never**: "Never put executable `.spec.ts` files under `.ai/qa/tests`; that directory is for shared Playwright config", and names module `__integration__` folders as preferred. `packages/core/src/modules/customers/__integration__/TC-CRM-WRITE-GUARD-001.spec.ts` is the correct location. I also confirmed the id is free — no `TC-CRM-WRITE-GUARD-*` exists on `develop`.

Two new findings, both small and both in code this PR adds. Neither is in the fix itself; both are in the new public options around it.

Reviewed scope: the full `develop...HEAD` diff across all 15 files, the incremental `fcb715f..baccdfe4` range, plus `commands/deals.ts` (create + update field assignment, `resolvePipelineAssignment`, the stage-transition guard), `interactionUpdateSchema`, `splitCustomFieldPayload` / `extractCustomFieldEntries`, the factory's four command-action paths, and the deals `list.fields` projection.

Validation source: **GitHub checks** on head `baccdfe4`. Every check that has reported is green: `prepare`, `test`, `lint`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance`, `license/cla`. Still pending at the time of writing: the 15 `ephemeral-integration` shards and `mutate (shared, …)`. I ran nothing locally.

## Verdict

❌ **request changes** — two Medium findings. This repo requests changes on Medium (`.ai/skills/om-auto-review-pr/SKILL.md`), so that is the mechanical outcome; I want to be clear that these are the only two things between this PR and an approval, and both are a handful of lines.

## 🧪 Validation Gate

Read from CI rather than reproduced locally, per this repo's GitHub-checks-first rule.

| Check | Status | Notes |
|-------|--------|-------|
| `prepare` (build:packages → generate → build:packages → build:app) | ✅ PASS | 3m52s |
| `test` | ✅ PASS | Step-level: `Check i18n sync` ✅, **`Check i18n usage` ✅**, `Checking types` ✅, `Test` ✅, `Test scripts` ✅, `Check create-app template parity` ✅, `Check DS token snapshot drift` ✅, `Run repo-wide audit guards` ✅, `Check agent instruction budget` ✅, `Check for time-bomb test literals` ✅ |
| `lint` / `ds-lint` | ✅ PASS | |
| `scope` / `audit-scope` | ✅ PASS | |
| `documents-multi-instance` | ✅ PASS | |
| `ephemeral-integration (1..15/15)` | ⏳ PENDING | The write guard sits in `packages/shared/src/lib/crud/factory.ts`, so affected-file detection legitimately selects every shard. Includes the new `TC-CRM-WRITE-GUARD-001`. |
| `mutate (shared, factory.ts, write-payload.ts, openapi/crud.ts)` | ⏳ PENDING | Mutation testing is scoped to `shared`, so it does cover the three new/changed files here. |

**On the blocker specifically.** I did not take the fix on trust, because `ci.yml:635-636` runs `yarn i18n:check-usage` with `continue-on-error: true` — a green `test` check does not by itself prove that gate passed. I pulled the step conclusions for job `98506449485`: `Check i18n usage` is `success`. The blocker is genuinely cleared. I also confirmed the diff introduces no literal-key `t(...)`/`translate('...')` call anywhere; the only surviving occurrence of the old key names is `customers.errors.immutable_field` inside a backtick in a JSDoc comment (`write-payload.ts:263`), which the scanner does not match.

## Findings

### ⚠️ Medium

**1. `packages/shared/src/lib/crud/factory.ts:2358` — `writeGuard.immutableFields` is applied to the CREATE path, so any factory route that adopts the option gets a `400` on `POST` for the very field it must accept at creation.**

The create and update paths pass the same config object through:

```ts
// :2358 (create)
const createGuard = guardWriteBody(createConfig.schema, body, opts.writeGuard)
// :2690 (update)
const updateGuard = guardWriteBody(updateConfig.schema, body, opts.writeGuard)
```

`guardWriteBody` forwards `config.immutableFields` into `inspectWritePayload`, whose second pass pushes any matching key as `reason: 'immutable'` (`write-payload.ts:190-192`), and `guardWriteBody` then throws unconditionally (`:328-334`). There is no create/update distinction anywhere on that path.

But a field that "cannot be changed after creation" — your own wording at `write-payload.ts:247` — is by definition a field the create endpoint *must* accept. A factory route declaring `writeGuard: { immutableFields: ['entityId'] }` would answer `400 "This field cannot be changed after creation."` to the `POST` that sets it.

This is latent, not live: I grepped the workspace and the only `immutableFields` consumer is `ACTIVITY_WRITE_GUARD` in `api/activities/route.ts:75`, a hand-written `PUT`-only handler that never reaches the factory. No factory route sets `writeGuard` at all yet. So nothing is broken today — but this is the same shape as finding 4 from the last round: a trap laid in a new public option, waiting for the first route that reaches for it. And `immutableFields` is precisely the option the six remaining bespoke customers handlers would want when they adopt the guard.

Fix is one line at the create call site — drop `immutableFields` for create:

```ts
const createGuard = guardWriteBody(createConfig.schema, body, { ...opts.writeGuard, immutableFields: undefined })
```

Either that, or split the config into `writeGuard.create` / `writeGuard.update` if you'd rather make the asymmetry explicit. A test asserting that a factory `POST` accepts a field named in `immutableFields` would pin it.

**2. `packages/shared/src/lib/api/scoped.ts:140-145` — `parseScopedCommandInput` passes `translate` to the guard but has no way to pass `messages`, so the guard's three `400` messages are hard-coded English at every command-backed write, and the `translate` argument is dead.**

```ts
const guarded = guardWriteBody(schema, rawBase, {
  aliasSnakeCaseKeys: options.aliasSnakeCaseKeys,
  immutableFields: options.immutableFields,
  rejectUnknownFields: options.rejectUnknownFields,
  translate,
})
```

No `messages`. And `resolveGuardMessage` short-circuits before `translate` is ever consulted:

```ts
// write-payload.ts:285-290
const fallback = WRITE_GUARD_FALLBACKS[name]
const override = config?.messages?.[name]
if (!override) return fallback          // ← always taken from this call site
const translate = config?.translate
```

`ScopedPayloadOptions` (`scoped.ts:27-53`) declares `aliasSnakeCaseKeys`, `immutableFields` and `rejectUnknownFields` but no write-guard `messages`, so no caller of `parseScopedCommandInput` can supply one. The result: at all 169 `parseScopedCommandInput` call sites, "This field was sent twice with different values.", "This field cannot be changed after creation." and "This field is not writable on this endpoint." are untranslatable English, and the `translate` function threaded in on line 144 can never fire. Root `AGENTS.md` lists "Never hard-code user-facing strings — use locale files" under **Never**, and these are user-facing `400` bodies.

I want to be fair about how this happened: it is a direct consequence of fixing my blocker, and you took the option I recommended. But there was a third route that gets both properties, and it is the one this very file already uses. `DEFAULT_MESSAGES` at `scoped.ts:55` holds `{ key, fallback }` pairs inside a plain object literal, and passes `msg.key` to `translate` — and, as I noted last round, the scanner does not treat an object-literal key as a `t()` call, which is exactly why those keys were never flagged. Give `WRITE_GUARD_FALLBACKS` the same `{ key, fallback }` shape with default keys, keep `config.messages` as the per-caller override, and you get `i18n:check-usage` green *and* translatable messages, with the file's own precedent as justification.

Minimum viable alternative if you'd rather not: add `writeGuardMessages?: CrudWriteGuardConfig['messages']` to `ScopedPayloadOptions` and forward it. Whichever way — if `translate` stays unreachable from this path, please drop it from the call rather than leave a parameter that cannot do anything.

### 💅 Nit

**3. `packages/core/src/modules/customers/api/deals/route.ts:504-525` — a straight round-trip on the flagship endpoint will always report `created_at` and `updated_at` as ignored, so the spec's "byte-identical" claim does not hold on the path the PR is built for.**

The deals `list.fields` projection emits `created_at` and `updated_at`. Neither camelCase spelling is declared by `dealUpdateSchema`, so the aliasing loop skips them and the second pass classifies both as `reason: 'unknown'`. A caller doing exactly what this PR is designed to enable — read a deal, write it straight back — now gets `{"ok":true,"ignoredFields":[{"key":"created_at","reason":"unknown"},{"key":"updated_at","reason":"unknown"}]}` every time.

That is arguably correct (they *are* ignored), and it is not a regression. But the spec's Backward Compatibility section says "Absent anything to report, the response is byte-identical to today's", and on the round-trip path there is always something to report. Either soften that sentence, or consider skipping the standard read-only timestamp columns the way scope and custom-field keys are already skipped — they are never caller-authoritative either.

**4. `packages/core/src/modules/customers/__tests__/write-payload-guard.test.ts` — the null-coercion table covers 6 of the 12 fields you widened.**

`valueAmount`, `expectedCloseAt`, `probability`, `description`, `closureOutcome` and `lossNotes` are pinned; `pipelineStage`, `pipelineId`, `pipelineStageId`, `valueCurrency`, `source` and `lossReasonId` are not. The two coercing fields are the ones that could corrupt data and they are covered, so this is genuinely a nit — but `pipelineId: null` and `pipelineStageId: null` now reach `resolvePipelineAssignment` (`commands/deals.ts:146`), which is the one place a `null` goes somewhere other than a straight `?? null` assignment. I read it: it normalizes both with `?? null` at `:151-152` and is null-safe, and the `shouldRecord` guard at `commands/deals.ts:822-825` compares `nextPipelineAssignment.pipelineStageId !== previousPipelineStageId`, so a round-trip echoing the current stage correctly records no spurious stage transition. Nothing is wrong — two more rows in the existing `it.each` would just pin it.

**5. `packages/shared/src/lib/crud/write-payload.ts:225` — `withIgnoredFieldsReport` falls back to the string key `carrier.ignoredFields`.**

The Symbol is the point of finding 10's fix, and the string fallback exists so the activities route can pass `{ ignoredFields }` directly, which is reasonable. It does mean that on a `.passthrough()` write schema a client-supplied `ignoredFields` key would be echoed back as if the server had produced it. No route in the diff is exposed to this. Worth a one-line comment noting the fallback is for hand-written callers only.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path.
- [x] No function signature changed in a breaking way — `CrudCommandActionConfig.response` gains `input` inside its existing args object; additive for every existing callback.
- [x] **Resolved from last round.** The `ActivityUpdateInput` narrowing is now recorded in the spec's Backward Compatibility section, along with the still-uncovered programmatic path.
- [x] **Resolved from last round.** `ignoredFields` is declared on `defaultCreateResponseSchema` / `defaultOkResponseSchema`, and `writeGuardErrorResponse` is merged into the create and update `errors` arrays in the CRUD OpenAPI builder, so all 77 factory routes document both. `CrudMethodConfig` gaining `errors` is additive.
- [x] No database table or column renamed or removed; no migration needed. I verified every newly-nullable schema field against a `nullable: true` column on `CustomerDeal` (`data/entities.ts:322-362`), and that `status` — the one field correctly left non-nullable — is `default 'open'` and NOT NULL.
- [ ] No config key renamed and no default changed silently — `aliasSnakeCaseKeys` still defaults to on. This is now a documented, deliberate trade-off rather than an unexamined one: the spec states the `value_amount` failure mode plainly and calls it intended behaviour. That was what I asked for, and it resolves finding 3. Flagging it here only so the merging maintainer sees it on the checklist rather than having to find it in the spec.
- [x] Where a contract had to change, the change is documented — the spec now covers the type narrowing, the previously-succeeding-request-can-now-400 case, and the deal null-acceptance widening.

## 🧪 Test Coverage

Each of the four coverage gaps I named last round now has a test, and they pin the right property rather than the symptom:

- **`withScopedPayload` × guard interaction** — `write-payload-guard.test.ts` "accepts a body carrying the organization_id and tenant_id a list emitted" round-trips under `selectedOrganizationId: null` and asserts both that the request succeeds *and* that `parsed.organizationId` still comes from context, not from the body's snake spelling. That second assertion is what makes it a real test of the security property rather than just of the 400 going away.
- **`aliasSnakeCaseKeys: false`** — `write-payload.test.ts` asserts `misspelled` is reported with aliasing off and that the same key is *applied* with aliasing on, so the `misspelled` reason cannot silently become unreachable.
- **The `null` coercion** — an `it.each` table plus an explicit "still rejects null for status" case. The negative case is what stops the widening from quietly swallowing the one column that must not accept null.
- **Integration** — `TC-CRM-WRITE-GUARD-001.spec.ts` covers the snake_case deal update, the `ignoredFields` report, the `entityId` `400`, and the phone-number edit that now persists. Fixtures are created in-test and torn down in `finally`, per `.ai/qa/AGENTS.md`; no reliance on seeded data. It asserts over the wire *and* re-reads the record, which is the right instinct given that the response was exactly what could not be trusted.

Remaining gaps, all small and all listed as nits above: no test that a factory `POST` accepts a field named in `immutableFields` (finding 1), and six of the twelve widened fields untested for `null` (finding 4). `create` through a real customers route is still untested, which I raised last round and is unchanged — `guardWriteBody` runs on `createConfig.schema` too, and finding 1 is exactly the kind of thing one create-path case would have caught.

pkarw:
🤝 Handing this back to @cielecki.

Every one of the eleven findings from the first pass is genuinely fixed, and I verified the blocker against CI rather than taking it on trust — `ci.yml` runs `yarn i18n:check-usage` with `continue-on-error: true`, so a green `test` check proves nothing by itself. Pulling the step conclusions for job `98506449485`: `Check i18n usage` is `success`. It is really clear.

**I also owe you a correction: finding 6 named the wrong directory and you read `.ai/qa/AGENTS.md` correctly, not backwards.** It says under **Never**: "Never put executable `.spec.ts` files under `.ai/qa/tests`." Your `__integration__` placement is the right one. The `TC-CRM-WRITE-GUARD-001` id is free on `develop` too.

Chasing the finding-3 probe into the `z.coerce` null coercion was the most valuable thing in this round, and keeping it in this PR is the right call rather than splitting it — the aliasing would have routed far more traffic into it, which makes it this PR's problem. I checked all twelve widened fields against `CustomerDeal`: every one maps to a `nullable: true` column, `status` is correctly left alone, and `commands/deals.ts` already applied `?? null` for each.

Two new findings, both Medium, both in the new public options rather than in the fix:

1. **`factory.ts:2358`** — `writeGuard.immutableFields` is applied on the CREATE path too, so a factory route adopting the option would `400` on the `POST` that sets the field. Latent (only the hand-written activities `PUT` uses it today), one line to fix.
2. **`scoped.ts:140-145`** — `translate` is passed to the guard but no `messages`, and `ScopedPayloadOptions` exposes no way to supply one, so `resolveGuardMessage` returns the English fallback at all 169 call sites and `translate` can never fire. There is a third route that gets both properties: the `{ key, fallback }` object-literal shape `DEFAULT_MESSAGES` already uses in that same file, which the i18n scanner does not flag.

This repo requests changes on Medium, so that is the mechanical verdict — but these two are the only things between this PR and an approval.

I deliberately did not push fixes to your branch. Finding 2 has three valid resolutions and which one to take is your call on your own PR, and you are clearly iterating on this actively — a commit from me would just collide with your next push. Re-request review when you have pushed and the automation will pick it up from the new head.

CI is green on everything that has reported (`prepare`, `test`, `lint`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance`). The 15 `ephemeral-integration` shards and `mutate (shared, …)` were still running when I posted; nothing in my verdict depends on them.

pkarw:

Re-review of head `baccdfe4` (previous review was on `fcb715f`). All 11 prior findings resolved, including the blocker — verified against CI step logs, not the check result, because `i18n:check-usage` is `continue-on-error` in `ci.yml`. One prior finding of mine (#6, integration-test location) was wrong and is corrected in the review.

Two new Medium findings, three nits. No blockers, no majors. Validation source: **GitHub checks** — nothing run locally. Pipeline label stays `changes-requested`; `priority-medium` and `risk-high` are already correct and unchanged. No autofix iterations: fork PR with an actively iterating author, and one finding has three valid resolutions that are the author's call. No CI follow-up is owed, so `ci-monitoring` was not applied.

cielecki:
Both Mediums are fixed, along with the three nits. Thanks for checking the blocker against the step conclusions rather than the job status; `continue-on-error` on that step would have hidden it from me.

## Medium 1, `immutableFields` on create

Fixed. The create path now strips `immutableFields` from the config before the guard sees it, through a small `createWriteGuardConfig` helper, so a route adopting the option keeps its `POST` working. This is the option the six remaining handlers reach for first, so it was worth fixing before any of them arrive.

A factory `POST` case now covers it, which also closes the create-path gap you raised in the first round: `writeGuard: { immutableFields: ['title'] }`, `POST` with a title, expect `201`.

## Medium 2, dead `translate`

Confirmed on both counts, the third route included. `WRITE_GUARD_MESSAGES` now holds `{ key, fallback }` pairs in a plain object literal, with `errors.write_guard.*` defaults, exactly as `DEFAULT_MESSAGES` does a few files over. `resolveGuardMessage` calls `translate` whenever a hook is supplied, so the argument threaded in by `parseScopedCommandInput` fires at every command-backed call site. `config.messages` still overrides the key per caller, and `ScopedPayloadOptions` gained `writeGuardMessages` so that override is reachable from the command path too.

`yarn i18n:check-usage` still exits `0`: the scanner matches literal arguments to the translate function, and an object-literal property is not one.

Three tests pin it: the default key reaches `translate`, a caller-supplied key replaces it, and no hook yields the English fallback.

## Nit 3, timestamps on the round trip

Taken as the second option. `created_at`, `updated_at` and `deleted_at` now sit outside the guard in both spellings, next to scope and custom-field keys, so a round-trip write reports nothing. Reporting two ignored fields on every write would train callers to stop reading the report. Tested, and the spec's Design section now lists all three excluded classes with the reason for each.

## Nit 4, the null table

All twelve widened fields are covered now, with a note on the test about why `pipelineId` and `pipelineStageId` are the interesting two.

## Nit 5

Commented.

## Validation

- `yarn i18n:check-usage`: `0`
- `packages/shared` typecheck: clean
- `packages/shared`: 339 tests green across `lib/crud` and `lib/api`
- `packages/core` customers: 248 suites, 1611 tests, green

Local failures are unchanged and all come from an unbuilt `dist` in my checkout: `principal-service export map` in `shared`, and the `#generated/*` resolution failures in `core`. CI is the better source for this branch.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T12:13:41Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Re-review: fix(crud): stop accepting write fields and discarding them silently

## 🎯 Summary

Both Medium findings and all three nits from the last round are correctly fixed, and I verified each rather than taking them on trust. But CI on this head surfaced something neither of us could have found by reading: **this PR's own integration test fails, and it is right to fail.** The phone-number half of the activities fix does not work. `TC-CRM-WRITE-GUARD-001 › an activity phone-number edit now persists` fails on `ephemeral-integration (8/15)`, and again on retry, with `Received: undefined`.

That test is the first time any of this ran against a real database — you couldn't run it locally on an unbuilt checkout, and the previous head's shards never finished before you pushed. It earned its keep on the first execution.

To be clear about what this does *not* undermine: the `date`/`time` half of the same fix is real and works, the deals fix is solid, the null-coercion fix is genuinely valuable, and every other case in the new spec passes. This is one specific claim that doesn't hold.

Validation source: **GitHub checks** on head `b5f9e5da`. I approved the fork runs so the full gate could execute. Nothing run locally.

## Verdict

❌ **request changes** — one blocker: a required check fails, and the underlying defect is a fix that doesn't fix anything.

## 🧪 Validation Gate

| Check | Status |
|-------|--------|
| `prepare`, `lint`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance`, `license/cla` | ✅ PASS |
| `test` — `Checking types`, `Test`, `Test scripts`, `Check i18n sync`, **`Check i18n usage`**, template parity, DS drift, audit guards, agent budget | ✅ PASS (all steps) |
| `mutate (shared, factory.ts, write-payload.ts, openapi/crud.ts)` | ✅ PASS |
| `ephemeral-integration` 1–7, 9–15 | ✅ PASS |
| **`ephemeral-integration (8/15)`** | ❌ **FAIL** |

I re-checked `Check i18n usage` at the step level again on this head (job `98514076589`) rather than trusting the job status, since `ci.yml:636` sets `continue-on-error: true`. It is `success` **with** the three new `errors.write_guard.*` keys present, which is the empirical confirmation that the Medium-2 approach holds.

## Findings

### ⛔ Blocker

**1. `packages/core/src/modules/customers/api/activities/route.ts:703` — the `phoneNumber` fix forwards the field one layer deeper into a command that also discards it. It is still accepted and dropped under a `200`; only the drop site moved.**

The failing assertion is `TC-CRM-WRITE-GUARD-001.spec.ts:169`:

```
Expected: "+48221234567"
Received: undefined
```

I traced the whole chain on this head. Two independent things are broken, and either one alone is enough to keep the phone number from ever coming back:

**(a) `customers.interactions.update` never reads `phoneNumber`, and no column stores it.**

```
grep -c phoneNumber  commands/interactions.ts  → 0
grep -c phoneNumber  data/entities.ts          → 0
```

`interactionUpdateBaseSchema` declares `phoneNumber` and the `superRefine` at `validators.ts:598-616` validates it for `interactionType === 'call'` — so it passes validation, which is exactly what makes this the same trap as the original bug. But `updateInteractionCommand.execute` assigns fields explicitly, one `if (parsed.x !== undefined)` at a time (`commands/interactions.ts:638-710`); there is no generic loop that could pick it up. `phoneNumber` is not among them, and `CustomerInteraction` has no such property. It is validated and then dropped.

The actual storage is the `callPhoneNumber` **custom field**. `ScheduleActivityDialog.tsx:389` writes `customValues.callPhoneNumber`, and the read path at `:152-158` seeds the field from `customValues.callPhoneNumber`, with a comment calling top-level `phoneNumber` the "newer write path" — a path that, as far as I can tell, was never finished on the persistence side.

**(b) Even if it were persisted, `GET /api/customers/activities` cannot return it.** The route has exactly one occurrence of `phone` in the whole file, and it is the line this PR added at `:703`. There is no `phone_number` in any list projection, so `activity?.phoneNumber ?? activity?.phone_number` is structurally guaranteed to be `undefined`.

**Why this is a blocker rather than a test bug.** The PR body states: "The same fixed list also drops `date`, `time` and `phoneNumber` … An edit to a call activity's phone number reported success and changed nothing." After this change that sentence is still true. The route now hands `phoneNumber` to `customers.interactions.update`, which drops it — a `200` for a write that changed nothing, which is the precise failure mode this PR exists to eliminate. Shipping it with a test named "an activity phone-number edit now persists" would put a false claim in the regression suite.

**The `date`/`time` half of the same change is fine** and I want to separate it clearly. `interactionUpdateSchema`'s transform (`validators.ts:618-625`) derives `scheduledAt` from `date`+`time`, and `commands/interactions.ts:701` persists `scheduledAt`. That half closes a real drop. Only `phoneNumber` is inert.

Three ways out, in the order I'd consider them:

1. **Route it to where the value actually lives.** Map `phoneNumber` onto the `callPhoneNumber` custom field in the activities adapter, matching what the UI already does, and assert on `customValues.callPhoneNumber` in the test. Smallest change, consistent with today's storage, no schema work.
2. **Finish the "newer write path".** Add the column/persistence to `customers.interactions.update` and expose it on the activities read projection. Bigger, needs a migration, and expands this PR's scope again.
3. **Treat it the way you treated `entityId`.** If top-level `phoneNumber` is not a supported write on this endpoint, name it in `immutableFields` (or drop it from `activityUpdateSchema`) so the request is refused instead of silently accepted — and revise the PR body and spec, which currently claim it is fixed.

Option 3 is fully in keeping with this PR's thesis and is the cheapest honest answer. Whichever you pick, the spec's Problem Statement and the PR body need updating, since both currently present the phone-number drop as fixed.

## ✅ Verified fixed from the last round

I checked each rather than assuming.

- **Medium 1 — `immutableFields` on create.** `createWriteGuardConfig` (`factory.ts:48-59`) strips the option, applied at both create call sites (`:2371`, `:2389`); the update path is untouched. The new factory `POST` test with `writeGuard: { immutableFields: ['title'] }` expecting `201` is a real regression test — it would have been `400` before.
- **Medium 2 — dead `translate`.** `WRITE_GUARD_MESSAGES` now holds `{ key, fallback }` pairs and `resolveGuardMessage` calls `translate` whenever a hook is present, so the argument threaded in at `scoped.ts:152` fires. I confirmed both properties this depends on: `scripts/i18n-check-usage.ts:116` filters missing-key errors to `r.direct` refs only, so object-literal keys cannot trip it (and CI agrees — the step is `success`); and `translateWithFallback` (`i18n/translate.ts:35-44`) returns the English fallback rather than the raw key when no dictionary declares it, so there is no user-visible regression. `writeGuardMessages` on `ScopedPayloadOptions` makes the override reachable from the command path. Three tests pin default key / caller override / no-hook fallback.
- **Nit 3 — timestamps.** `READ_ONLY_TIMESTAMP_KEYS` covers all three names in both spellings and joins the passthrough set. I confirmed this only changes *reporting*: unknown keys were already left in the payload by `inspectWritePayload`, so nothing about what reaches `schema.parse` changes.
- **Nit 4 — null table.** All twelve widened fields covered, with a good note on why `pipelineId`/`pipelineStageId` are the interesting two.
- **Nit 5** — commented.

Also worth recording: `mutate (shared, …)` passed, and that job genuinely covers `write-payload.ts`, `factory.ts` and `openapi/crud.ts`, so the new logic survived mutation testing rather than merely being executed by the tests.

### 💅 Nit

**2. `packages/shared/src/lib/crud/factory.ts:48-59` — `createWriteGuardConfig` is declared inside the import block.**

Imports run to `:46`, the function occupies `:48-59`, and imports resume at `:60`. Hoisting makes it work and nothing will flag it — `yarn lint` is `turbo run lint` and only `apps/mercato` declares a `lint` script, so `packages/shared` is never linted — but it reads as a mistake. Move it below the last import.

## 🧪 Test Coverage

The unit and integration coverage added across the last two rounds is genuinely strong, and I'd rather say that plainly than let one failing case colour it: the guard-message tests pin the translate path from three directions, the create-path `POST` test closes the gap I raised in round one, the null table now covers every widened column plus the negative case for `status`, and the scope round-trip test asserts the security property (scope still comes from context) rather than just the absence of a 400.

The one gap is the one CI found, and it is the good kind of failure — an integration test doing exactly the job unit tests could not. `write-payload-guard.test.ts` exercises `phoneNumber` against the *schema*, where it validates cleanly; only a real request through the real command could show that nothing persists. That is the argument for the integration coverage requirement in `.ai/qa/AGENTS.md`, demonstrated.

When you fix the blocker, please keep the test asserting on a **re-read of the record** rather than on the `200`. It is the assertion that caught this.

pkarw:
🤝 Handing back to @cielecki.

Both Mediums and all three nits are correctly fixed — I verified each, including the two properties the Medium-2 approach rests on (the scanner filters missing-key errors to `r.direct` refs, and `translateWithFallback` returns the English fallback rather than the raw key for an undeclared key). `Check i18n usage` is `success` at the step level on this head *with* the new `errors.write_guard.*` keys present.

I held the verdict for the integration shards rather than approving with them pending, and that turned out to matter: **your own new spec fails, and it is right to fail.**

`TC-CRM-WRITE-GUARD-001 › an activity phone-number edit now persists` fails on shard 8/15, and again on retry, with `Received: undefined`. The phone-number fix forwards `phoneNumber` one layer deeper into `customers.interactions.update`, which also drops it:

- `grep -c phoneNumber commands/interactions.ts` → **0**, and `data/entities.ts` → **0**. `interactionUpdateBaseSchema` declares it and the `superRefine` validates it, but `updateInteractionCommand.execute` assigns fields one explicit `if` at a time and `phoneNumber` is not among them. The real storage is the `callPhoneNumber` custom field — `ScheduleActivityDialog.tsx:389` writes it, `:152-158` reads it back, and the comment there calls top-level `phoneNumber` the "newer write path", which looks like it was never finished on the persistence side.
- Independently, `GET /api/customers/activities` has no phone field in any projection, so that assertion could never pass even if the write worked.

So the PR body's "An edit to a call activity's phone number reported success and changed nothing" is still true after the change — only the drop site moved. That is the exact failure mode this PR exists to remove, so I can't wave it through.

**The `date`/`time` half of that same fix is real** and I don't want it tarred with this: the schema transform derives `scheduledAt` and `commands/interactions.ts:701` persists it. Only `phoneNumber` is inert.

Three ways out in the review; the cheapest honest one is option 3 — treat it like `entityId` and refuse it, then correct the PR body and spec. Option 1 (map it onto the `callPhoneNumber` custom field, matching the UI) is the smallest change that actually makes the claim true.

Worth saying: this is the integration-coverage requirement doing precisely its job. `write-payload-guard.test.ts` exercises `phoneNumber` against the *schema*, where it validates cleanly — only a real request through the real command could show that nothing persists. Everything else is green, including `mutate (shared, …)` and the other 14 shards.

I approved the fork CI runs on this head so the full gate could execute; do ping me if a future push sits at `action_required` and I'll approve again.

pkarw:

Re-review of head `b5f9e5da`. Both Medium findings and all three nits from the previous pass verified fixed. **One new blocker**: `ephemeral-integration (8/15)` fails on this PR's own `TC-CRM-WRITE-GUARD-001 › an activity phone-number edit now persists` (twice, including retry — not flaky). `phoneNumber` is validated and then dropped by `customers.interactions.update`, which has zero references to it and no backing column; the activities read projection cannot return it either.

Validation source: **GitHub checks** — nothing run locally. I approved the fork runs on this head so the gate could execute. Everything else is green: `prepare`, `test` (all steps incl. `Check i18n usage`), `lint`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance`, `mutate (shared, …)`, and shards 1-7 and 9-15.

Pipeline label stays `changes-requested`; `priority-medium` and `risk-high` unchanged. No autofix iterations — fork PR, actively iterating author, and the blocker has three valid resolutions that are the author's call. `qa-approved` not applied and `qa` not touched, per the QA gate.

pkarw:
@cielecki please take care of failing CI - integration tests

cielecki:
@pkarw The phone persistence blocker is fixed. `phoneNumber` now reaches the interaction custom-field writer as `callPhoneNumber`, including writes through the activity adapter. A phone-only update preserves other custom fields; conflicting duplicate values return 400, and omission preserves the phone. A partial phone-only update with null clears the value; existing call-type validation is unchanged.

The regression now sends a top-level-only phone update and reads the same record back through `customValues.callPhoneNumber`. All nine tests passed in a fresh managed environment, covering both activities and interactions. The runner exited 0.

I also merged current `develop`, preserved its route-indexer wrappers, kept the optional `entityId` in the exported activity update type, and moved the factory helper below the imports. The HTTP guard still rejects parent changes.

Package unit tests, typecheck, changed-file lint and the production build passed. Local unit retries used serial Jest execution after worker signals; queue tests needed `--forceExit` for lingering handles. The PR body has the validation details.

Please re-review this revision and approve the fork workflows: [CI](https://github.com/open-mercato/open-mercato/actions/runs/34736327854) and [mutation tests](https://github.com/open-mercato/open-mercato/actions/runs/34736327864). Both currently report `action_required`; the full upstream integration matrix has not run on this head.

cielecki:
@pkarw could you re-review this PR, approve its fork workflow runs, and arrange QA sign-off so we can complete the merge?

Issue #3793, opened 2026-07-06
bug: Portal invite acceptance — user isn't shown their own dashboard right after accepting


alinadivante:
## Summary

After filling in the invitation-acceptance form and clicking "Accept Invitation," the page does not correctly show the new, logged-in dashboard. Reproduced three times, including once end-to-end with a real invitation email and a second person testing on a separate machine — so this is not a one-off glitch or a test-automation artifact.

## Page URL

`/{orgSlug}/portal/invite?token=...`

## Steps to reproduce

1. Have an admin send a customer invitation.
2. Open the invitation link (from the email) while logged out.
3. Fill in a display name, a password, and the same password again to confirm.
4. Click "Accept Invitation."

## Expected result

Land on the portal dashboard, fully logged in as the account just created (name visible in the sidebar, logout option available).

## Actual result

Two different bad outcomes were observed across three attempts, always right after clicking "Accept Invitation":
- Sometimes the browser is sent straight back to the login screen instead of the dashboard.
- Sometimes the dashboard does load, but shows a different, previous test user's name and unrelated demo data, while the top of the page still looks logged-out.

In every case, manually logging in right afterward — with the exact email and password just used to accept the invitation — works immediately and shows the correct, fully logged-in dashboard. So the account is genuinely created correctly; the problem is isolated to what happens automatically right after clicking "Accept Invitation."

## Environment

- Branch: `autofix/supersede-3083` (PR #3598), commit `96d55678f`
- Module or area: portal / customer_accounts
- Browser/Runtime: Chromium (automated) for two of the three reproductions; Chrome (manual, real user) for the third, using a real invitation email delivered via Resend to a real Gmail inbox

## Additional context

<img width="1912" height="870" alt="Image" src="https://github.com/user-attachments/assets/e1bb794d-94d8-4e2e-b546-bf542b477adb" />

For the developer: the invite page redirects with `router.replace('/${orgSlug}/portal/dashboard')` right after the accept call succeeds — a client-side/soft navigation rather than a full page reload. The new session cookies are set correctly and are valid at that exact moment (confirmed by inspecting them directly), so this looks like a stale client-side render/cache being reused right after login rather than a session/account problem. A full reload instead of a soft navigation immediately after establishing a brand-new session would likely avoid this.

wojciechszyjka:
🤖 `om-auto-fix-issue` triage: **NO_ACTION_NEEDED — already fixed on `develop` and `main`.** No security escalation warranted. Details below.

[...]

PR #3598, opened 2026-06-25
Fix customer account invitation emails (supersedes #3083)

pkarw:
Supersedes #3083 (push access to the contributor's fork branch `TH-EY:contrib/customer-accounts-invite-email` was unavailable, so the conflict resolution is carried here).

Credit: original implementation by @pmadajthey.

## What this does
Carries forward @TH-EY's invitation-email fix from #3083 and resolves the merge conflict against the current `develop`.

The conflict was in:
- `packages/core/src/modules/customer_accounts/api/admin/users-invite.ts`
- `packages/core/src/modules/customer_accounts/api/portal/users-invite.ts`

`develop` added a fire-and-forget `customer_accounts.user.invited` event after invitation creation; #3083 added direct invitation-email delivery (returning `502` when the email fails). These are orthogonal — the event is for notification/CRM side effects, not for sending the invitation email — so both are preserved: emit the event, then send the email and gate the response on its success.

I also added a `sendCustomerInvitationEmail` mock to `users-invite-events.test.ts` so develop's event tests stay green on the 201 path (otherwise the real email helper would run and fail in the node test env).

Credit for the original fix goes to @TH-EY (@pmadajthey).

Co-authored-by: Patrick Madaj <patryk.madaj@they.dev>

PR #5686, opened 2026-08-26
fix(portal): stop painting the logged-out header over authenticated content

### LINKED ISSUE #5678: Fix: Customer Portal renders the logged-out header over authenticated dashboard content at /{orgSlug}/portal
## Summary

- Entering the Customer Portal at `/{orgSlug}/portal` with an active, matching customer session renders the **public (logged-out) header** — Log In / Sign Up, no sidebar — around **authenticated dashboard content**. With a session active the logged-out header should never appear.

Reported by Jakub Godawa on a demo. The normal login flow looks correct; only the direct-link entry (the admin "Open Portal" action, and the front door of every custom domain) is affected.

## Analysis

Affected areas:
- `apps/mercato/src/app/(frontend)/layout.tsx`
- `packages/create-app/template/src/app/(frontend)/layout.tsx` (byte-identical copy)
- `packages/ui/src/portal/PortalShell.tsx`
- `packages/core/src/modules/portal/frontend/[orgSlug]/portal/page.tsx`

**Expected:** with a customer session that matches the URL org, the portal always renders the authenticated shell (user menu, sidebar, event bridge).
**Actual:** public shell + authenticated content.

Two independent defects combine — either alone is a bug.

**1. The portal root is classified as a public route, but the session is still handed to the provider.**

`(frontend)/layout.tsx:19-22`

```ts
function isPublicPortalRoute(pathname: string): boolean {
  if (/^\/[^/]+\/portal\/?$/.test(pathname)) return true   // <-- portal root
  return PUBLIC_SUFFIXES.some((s) => pathname.endsWith(s))
}
```

`(frontend)/layout.tsx:155-158`

```ts
authenticated={!isPublic && customerAuthMatchesUrlOrg}          // false at the root
customerAuth={customerAuthMatchesUrlOrg ? customerAuth : null}  // NOT gated by isPublic
```

So on `/{org}/portal` with a matching session the two props disagree by construction:

- `authenticated === false` → `PortalShell` takes the `if (!authenticated)` branch (`PortalShell.tsx:261`) → public header/footer, no sidebar, `enableEventBridge={false}`, portal-nav fetch skipped.
- `customerAuth` is the real session → `PortalProvider` initialises `auth.user` (`PortalContext.tsx:74-101`) → every page below renders its authenticated variant.

The `isPublic` gate exists so a visitor with no session — or a session belonging to a *different* org — still sees the landing page. It was never meant to demote a session that *does* match the URL org.

**2. The layout does not re-render across the client-side redirect, so the wrong value sticks.**

The landing page redirects a logged-in visitor client-side (`portal/page.tsx:24-28`):

```ts
useEffect(() => {
  if (!auth.loading && auth.user) router.replace(`/${orgSlug}/portal/dashboard`)
}, [...])
```

`(frontend)/layout.tsx` sits **above** the `[...slug]` segment — deliberately, so portal navigation does not remount the client subtree (`layout.tsx:44-45`). A client-side navigation between two routes under `[...slug]` therefore never re-runs it, and `authenticated={false}`, computed for `/{org}/portal`, is still in place once `/dashboard` mounts.

`PortalShell` cannot correct it either, because the prop wins unconditionally (`PortalShell.tsx:181`):

```ts
const authenticated = authenticatedProp ?? !!user   // prop is `false`, not `undefined`
```

The public shell therefore never self-heals, even though the context beside it holds a real user.

**Why the normal path looks fine:** login redirects with a full page load — `login/page.tsx:50` `window.location.assign(...)` — so the layout runs fresh on a non-public path.

## Steps to reproduce

1. Log in to the portal and land on `/{org}/portal/dashboard`.
2. Navigate to `/{org}/portal` (what the admin **Open Portal** button links to, and what `/` rewrites to on a custom domain).
3. The landing page redirects to `/dashboard`: authenticated dashboard content under a Log In / Sign Up header, with no sidebar.

## Blast radius

- **Custom domains hit this on their front door.** `apps/mercato/src/proxy.ts:31` rewrites `/` → `/{orgSlug}/portal`, so every logged-in customer arriving at `https://portal.example.com/` takes exactly this path.
- The admin "Open Portal" action links to the slug-less root (`customer_accounts/lib/portalUrl.ts:110`, `buildPortalRootUrl`).
- **Not cosmetic:** `enableEventBridge={authenticated}` is `false`, so the portal SSE bridge never mounts — no realtime notifications for the rest of that session — and the portal-nav fetch (`PortalShell.tsx:206-231`) is skipped, so the sidebar is missing.

## How to implement

1. **Server, root cause** — in `apps/mercato/src/app/(frontend)/layout.tsx`, stop demoting a session that matches the URL org at the portal root. Split the root test out of `isPublicPortalRoute` (it must stay "public" for the org-lookup branch at `layout.tsx:85-89`, which resolves the org by slug) and use it only where it belongs:

   ```ts
   const isPortalRoot = /^\/[^/]+\/portal\/?$/.test(pathname)
   ...
   authenticated={customerAuthMatchesUrlOrg && (!isPublic || isPortalRoot)}
   ```

   `userName` / `userEmail` are already gated on `customerAuthMatchesUrlOrg` only, so they need no change. Login/signup/verify/reset/invite keep the public chrome; the root follows the session.

2. **Mirror into the template** — `packages/create-app/template/src/app/(frontend)/layout.tsx` is byte-identical to the app copy today. `AGENTS.md` → `create-app` → "Template Sync Checklist" requires the same change in the same task.

3. **Client, defense in depth** — in `packages/ui/src/portal/PortalShell.tsx:181`, let the context upgrade the shell so a stale layout prop can never contradict the rendered content:

   ```ts
   const authenticated = !!user || (authenticatedProp ?? false)
   ```

   This closes the whole class of "stale layout above `[...slug]`" bugs rather than this one instance, and is what makes the fix hold for #3793's `router.replace` from the invite page.

4. **Optional hardening** — server-redirect `/{org}/portal` → `/{org}/portal/dashboard` when a matching session cookie is present, so the landing page is only ever rendered for anonymous visitors and no client-side redirect is involved. Step 1 alone fixes the report; steps 1 + 3 are the durable fix.

5. **Tests:**
   - `apps/mercato/src/app/(frontend)/__tests__/portal-org-binding.test.tsx` — new case: pathname `/org-a/portal` with a **matching** session asserts `authenticated: true` (today it is `false`). The existing `it.each` case that asserts `authenticated: false` for `/org-b/portal` with a **mismatched** session must keep passing; drop `/org-b/portal` from that list only if it is re-covered explicitly.
   - `packages/ui/src/portal/__tests__/` — `PortalShell` rendered with `authenticated={false}` but a `PortalProvider` holding a user renders the authenticated chrome (user menu present, Log In / Sign Up absent).
   - Integration (`packages/core/src/modules/portal/__integration__/`) — with a session, navigate `/{org}/portal` → `/dashboard` and assert the header shows the user menu and not Log In / Sign Up.

## Compatibility notes

- `PortalShell`'s `authenticated` prop is a public, exported component contract. Step 3 changes its precedence: an explicit `authenticated={false}` no longer forces the public shell when a `PortalProvider` above it holds a user. This is additive-safe for the documented use (public pages that render `PortalShell` without a provider, where `user` is `null` anyway), but it must be called out in the PR description and, if a third party relies on force-public chrome for a signed-in user, an explicit opt-out prop is the migration path.
- The frozen component-replacement handle IDs (`PORTAL_SHELL_HANDLE`, `PORTAL_HEADER_HANDLE`, …) are untouched.
- No DB schema, API route, event ID, or DI key changes.

## Related

- #3793 — same hybrid symptom ("the dashboard does load … while the top of the page still looks logged-out") reached through the invite page's `router.replace`. Defect 2 above is its cause; step 3 should fix that half of it. #3793 additionally reports a *previous* user's data rendering, which is a separate router-cache concern and is out of scope here.

## How to pick this up

- Run `om-auto-fix-issue {thisIssueNumber}`, or hand the "How to implement" section to `om-auto-create-pr` as the brief. A full root-cause write-up also lives at `.ai/analysis/2026-08-26-portal-hybrid-header.md` on branch `worktree-portal-hybrid-header-analysis`.

## Out of scope

- The wrong-user-data half of #3793 (router cache reusing a previous session's RSC payload).
- Any redesign of the portal landing page or of the public shell itself.
- Reworking how `(frontend)/layout.tsx` derives the pathname from `x-next-url`.

pat-lewczuk:
## Summary

Entering the Customer Portal at `/{orgSlug}/portal` with an active, matching customer session rendered the **public shell** — Log In / Sign Up, no sidebar, no event bridge — around the **authenticated dashboard**. Reported by Jakub Godawa on a demo; the normal login flow looked fine because it redirects with a full page load.

Two independent defects combined:

1. **The portal root was classified as a public route, but the session was handed to the provider anyway.** `(frontend)/layout.tsx` derived `authenticated={!isPublic && customerAuthMatchesUrlOrg}` — `false` at the root — while passing the real `customerAuth` through unconditionally. So `PortalShell` took its public branch while `PortalProvider` below it held a real user, and the two disagreed by construction. The `isPublic` gate exists so a visitor with no session, or one bound to another org, still gets the landing page; it was never meant to demote a session that *does* match the URL org.

2. **The layout never re-ran across the client-side redirect, so the stale value stuck.** It sits above the `[...slug]` segment — deliberately, so portal navigation does not remount the client subtree — and the landing page redirects logged-in visitors with `router.replace(...)`. `authenticated={false}`, computed for the root, was still in place once `/dashboard` mounted, and `PortalShell` trusted it: `authenticatedProp ?? !!user` let a `false` prop win over a context that already held the user, so the public shell never self-healed.

This is the portal's primary entry point: the admin **Open Portal** action links to it (`buildPortalRootUrl`), and `proxy.ts` rewrites `/` to it on **every custom domain**. It is also not cosmetic — `enableEventBridge={authenticated}` was `false`, so the portal SSE bridge never mounted (no realtime notifications for the rest of that session), and the portal-nav fetch was skipped, so the sidebar was missing.

Fixes #5678.

**Correction (review finding 3):** an earlier revision of this body claimed this also fixes the "top of the page still looks logged-out" half of #3793. It does not. That claim was inherited from #5678's Related section, which predates #3598: the invite page now calls `navigateWithPageReload(...)` (`portal/lib/navigation.ts` → `window.location.assign`), a full page load that re-runs the layout on the destination path, so the stale-prop path closed here is not reachable from invitation acceptance. #3793 stays open and untouched by this PR.

## Changes

- `apps/mercato/src/app/(frontend)/layout.tsx` — split the portal-root test out of `isPublicPortalRoute` (it must stay "public" for the slug-based org lookup) and stop it demoting a matching session: one `authenticatedChrome = customerAuthMatchesUrlOrg && (!isPublic || isPortalRoot)` expression now drives `authenticated`, `customerAuth`, `userName` and `userEmail` together. Gating the session alongside the flag is what actually keeps login/signup/verify/reset/invite on the public chrome — see review finding 1 below.
- `packages/create-app/template/src/app/(frontend)/layout.tsx` — mirrored per the `create-app` Template Sync Checklist; the two files remain byte-identical.
- `packages/ui/src/portal/PortalShell.tsx` — `!!user || (authenticatedProp ?? false)`, so a context user can upgrade the shell and a stale layout prop can never contradict the rendered content. This closes the class of bug, not just this instance.
- Regression tests at both layers plus an integration spec (below).

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:** n/a — bug fix. Root-cause write-up committed at `.ai/analysis/2026-08-26-portal-hybrid-header.md`.

## Testing

Full configured validation gate, run **locally** (no compose `app` container was up), in the `.ai/agentic.config.json` order — all green:

`yarn build:packages` → `yarn generate` → `yarn build:packages` → `yarn i18n:check-sync` → `yarn i18n:check-usage` → `yarn typecheck` → `yarn test` → `yarn build:app`

- `yarn test`: 32/32 turbo tasks passed, including `@open-mercato/ui` (231 suites, 1880 tests) and `@open-mercato/app` (91 suites, 574 tests).
- The first `yarn build:app` failed on a missing `entities.ids.generated.js` — fresh-worktree bootstrap, resolved by the gate's own second `build:packages`, then green.

Regression coverage added, and **each new test was verified to fail without the corresponding fix**:

- `apps/mercato/src/app/(frontend)/__tests__/portal-org-binding.test.tsx` (+ template mirror) — `/org-a/portal` and `/org-a/portal/` with a matching session now assert `authenticated: true` (2 failures before the fix). A new `it.each` pins that `/org-a/portal/login|signup|invite` keep the public chrome even with a matching session; the pre-existing mismatched-org cases still pass. 14 tests total, all passing.
- `packages/ui/src/portal/__tests__/PortalShell.test.tsx` — `authenticated={false}` with a `PortalProvider` holding a user renders the authenticated chrome, Log In / Sign Up absent (fails before the fix); and the converse, that no context user leaves the public chrome untouched and fires no nav fetch.
- `packages/core/src/modules/portal/__integration__/TC-AUTH-064-portal-root-authenticated-shell.spec.ts` — new spec: with session cookies, land on `/{orgSlug}/portal`, follow the redirect to `/dashboard`, assert `portal-nav-ready[data-ready="true"]`, a visible sidebar link, and zero Log In / Sign Up links. Self-contained (creates its own customer user, deletes it in `finally`).

**Integration run:** the spec executed and passed in CI on the first head (`ccea4a0`), in `ephemeral-integration` shard 11/15 — an earlier revision of this body said it was left to QA, which was stale. It was renamed to `TC-AUTH-064` in the review pass (finding 2: `TC-AUTH-063` is taken by `auth/__integration__/TC-AUTH-063-feature-check-org-scope.spec.ts`, which landed on `develop` after this branch forked, so the merge result would have carried two scenarios under one ID). The rename is re-validated with `playwright test --list`; CI re-runs it under the new name.

## Review pass (@pkarw)

All three findings and the nit are addressed in `bbfe4462f`:

1. **Public auth routes rendered the authenticated chrome.** Correct and my miss — `customerAuth` was never gated on `isPublic`, so with the new `PortalShell` precedence a matching session produced the authenticated shell on `/portal/login` and friends, contradicting this body, the analysis doc and the test's own name. Took the first of the two suggested routes — keep the documented behaviour: the chrome, the session and the display name now travel together behind one `authenticatedChrome` expression. Checked the auth pages first as asked: `login`, `signup`, `invite`, `reset-password` and `verify` destructure only `tenant` from `usePortalContext()`, never `auth`, and `initialTenant` is independent of the session, so nothing regresses; the landing-page redirect keeps its session because the root is `isPortalRoot`. This also removes the `enableEventBridge` asymmetry, since it derives from the same flag.
2. **Test-case ID collision.** Renamed to `TC-AUTH-064-portal-root-authenticated-shell.spec.ts`, re-checked free against `origin/develop` immediately before pushing; file name, `describe` title, doc comment and fixture prefix all updated.
3. **The #3793 claim.** Verified and dropped — the invite page uses `navigateWithPageReload`, not `router.replace`. Corrected above, and the misleading comment I left on #3793 has been retracted there.
4. **Nit — `Date.now()` collision across shards.** Fixture key is now a timestamp plus a random suffix.

Test coverage for the seam, per the review: the auth-route `it.each` asserts the withheld session (`customerAuth: null`, `userName: null`, `userEmail: null`) rather than only the flag it pinned before, and two new composed tests drive the real layout props through `PortalLayoutShell` → `PortalProvider` → `PortalShell` — one pinning the public chrome and zero `apiCall`s for the auth-route props, one pinning the authenticated chrome for the signed-in props. Reverting the session gate turns the three auth-route cases red; verified.

## Compatibility notes

`PortalShell`'s `authenticated` prop is a public exported contract and this changes its precedence: an explicit `authenticated={false}` no longer forces the public shell when a `PortalProvider` above it holds a user. This is safe for the documented use — public pages render `PortalShell` without a provider, where `user` is `null` anyway — and a test pins that behaviour. If a third party ever needs force-public chrome for a signed-in user, an explicit opt-out prop is the migration path. The frozen component-replacement handle IDs are untouched; no DB schema, API route, event ID, or DI key changes.

## Checklist

- [x] I updated documentation, locales, or generators if the change requires it. (No new user-facing strings; `i18n:check-sync` clean.)
- [x] I added or updated integration tests (`TC-AUTH-064`).
- [x] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — N/A, bug fix; analysis committed under `.ai/analysis/`.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5686"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790372763&installation_model_id=432243&pr_number=5686&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5686&signature=383b3fa5aef4845df2873220d9508fbd0db72a1f8007fa441b8f8cb65cd56df1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pat-lewczuk:

- 🐛 `bug` — restores intended behaviour: an active customer session was rendering the public, logged-out shell around authenticated content.
- 🔍 `review` — the PR is ready and non-draft, and is in no other pipeline state.
- 🧪 `needs-qa` — the change touches `.tsx` files under `packages/ui/src/`, so the automated-verification exemption does not apply; the new integration spec `TC-AUTH-063` has not been executed against a live app and needs a QA run.
- 🔺 `priority-high` — the broken path is the portal's primary entry point: the admin "Open Portal" action links to it and `proxy.ts` rewrites `/` to it on every custom domain, so each logged-in customer arriving at their portal's front door hits it; it also silently disables the portal SSE bridge and the sidebar for the rest of that session.
- 🟢 `risk-low` — one boolean in two byte-identical layout files plus one precedence expression in `PortalShell`, with no database, API, event, or DI surface touched, covered by regression tests at both layers that were each verified to fail without the fix.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-26T23:36:49Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: fix(portal): stop painting the logged-out header over authenticated content

## 🎯 Summary

This PR fixes the hybrid portal shell reported in #5678, and it fixes it at the right place. The server layout stops demoting a session that matches the URL organization at the portal root (`authenticated={customerAuthMatchesUrlOrg && (!isPublic || isPortalRoot)}`), the `isPortalRoot` test is split cleanly out of `isPublicPortalRoute` so the slug-based organization lookup keeps its "public" branch, and `PortalShell` now lets a context user upgrade the shell so a layout prop that is stale across a client-side navigation can no longer contradict the content underneath it. The template mirror is byte-identical to the app copy (I diffed both `layout.tsx` and the new test file), which satisfies the create-app Template Sync Checklist and the standing lesson about keeping shell/layout files in lockstep. The regression coverage is real and well-targeted: the layout test pins the root prop, the `PortalShell` test pins the context-wins precedence, and `TC-AUTH-063` covers the whole entry path end to end and is self-contained with cleanup in `finally`. The implementation follows the issue's "How to implement" section step for step, and the compatibility note about `PortalShell`'s `authenticated` prop precedence is exactly the kind of disclosure this repo asks for.

Reviewed scope: the two `(frontend)/layout.tsx` copies, `packages/ui/src/portal/PortalShell.tsx`, the three test files, and the committed analysis document — read against `PortalLayoutShell.tsx`, `PortalContext.tsx`, the portal landing/login/invite pages, `BACKWARD_COMPATIBILITY.md`, and `.ai/qa/AGENTS.md`.

What stops this from being an approve is that the two halves of the fix interact in a way the description, the committed analysis, and one new test all describe incorrectly, plus a test-case ID that collides with a spec that landed on `develop` after this branch forked. Neither is hard to fix.

## Verdict

❌ **request changes** — three minor findings. This repository's review extension requests changes on Medium/minor findings as well as blockers and majors, so these block the approve even though none of them is a correctness bug in the shipped fix itself. The central one is finding 1: the layout still hands `customerAuth` to the provider on the public auth routes, so with the new `PortalShell` precedence those routes render the **authenticated** chrome for a matching session — the opposite of what the PR body, `.ai/analysis/2026-08-26-portal-hybrid-header.md:106`, and the new `it.each` test's own name all claim.

## 🧪 Validation Gate

**Validation source: GitHub checks** (head `ccea4a0`, runs [33016865664](https://github.com/open-mercato/open-mercato/actions/runs/33016865664) and [33016862573](https://github.com/open-mercato/open-mercato/actions/runs/33016862573)). Per this repo's checks-first review rule, the configured `validation.commands` gate was not re-run locally, because CI executed the whole of it on this exact head and reported green. Every configured command maps onto a green check:

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` job, step "Build packages" |
| `yarn generate` | ✅ PASS | `prepare` job, step "Generate module artifacts" |
| `yarn build:packages` (second run) | ✅ PASS | `prepare` job — the rebuild over generated files |
| `yarn i18n:check-sync` | ✅ PASS | `test` job, `yarn tsx scripts/i18n-check-sync.ts` |
| `yarn i18n:check-usage` | ✅ PASS | `test` job |
| `yarn typecheck` | ✅ PASS | `test` job (`turbo run typecheck --filter=[origin/develop]...`) |
| `yarn test` | ✅ PASS | `test` job (`turbo run test --filter=[origin/develop]...`), plus `test:scripts`, `test:repo-wide-guards`, `agents:check-budget`, `ds:tokens:check` |
| `yarn build:app` | ✅ PASS | `prepare` job, step "Build app" |

Beyond the configured gate: `lint`, `ds-lint`, `scope`, `audit-scope`, `docker-build`, `documents-multi-instance`, CodeQL and the CLA check are all green, and **all fifteen `ephemeral-integration` shards passed**. Template parity was verified by direct diff rather than `yarn template:sync` (the review worktree has no installed dependencies): `apps/mercato/src/app/(frontend)/layout.tsx` and its test are byte-identical to the `packages/create-app/template` copies.

One correction to the PR's Testing section, in the author's favour: `TC-AUTH-063` **did** execute in CI and passed. It ran in shard 11/15, job [98338219687](https://github.com/open-mercato/open-mercato/actions/runs/33016865664/job/98338219687), as `✓ 45 packages/core/src/modules/portal/__integration__/TC-AUTH-063.spec.ts:18:3 › … landing on /{orgSlug}/portal with a session redirects to the dashboard under the authenticated header (823ms)`. The "Not run … execution is left to QA" caveat is stale and worth updating so QA does not repeat work CI already did.

## Findings

### 🔹 Minor

**1. `apps/mercato/src/app/(frontend)/layout.tsx:169` (and the template mirror) — the public auth routes no longer render the public chrome, contradicting the description, the analysis doc, and the new test's own name.**

The layout's `customerAuth` prop is *not* gated on `isPublic` — it never has been, and this PR does not change it:

```tsx
authenticated={customerAuthMatchesUrlOrg && (!isPublic || isPortalRoot)}   // false on /portal/login
customerAuth={customerAuthMatchesUrlOrg ? customerAuth : null}             // the real session, on every route
```

`PortalLayoutShell.tsx:44` forwards that session to `PortalProvider` as `initialAuth`, and `PortalContext.tsx:74-101` initialises `auth.user` from it. With `PortalShell.tsx:193` now reading `!!user || (authenticatedProp ?? false)`, a visitor whose session matches the URL organization gets the **authenticated** shell on `/{org}/portal/login`, `/signup`, `/invite`, `/verify` and `/reset-password`: sidebar, account menu, `Log Out`, no `Log In` / `Sign Up`, and a live `/api/customer_accounts/portal/nav` fetch fired from the login page. This is not a hypothetical composition — it is exactly the composition the PR's own new `PortalShell` test constructs and asserts (`packages/ui/src/portal/__tests__/PortalShell.test.tsx:159`), only reached through the real layout props instead of hand-written ones.

That directly contradicts three statements shipped in this change: the PR body ("Login/signup/verify/reset/invite keep the public chrome"), `.ai/analysis/2026-08-26-portal-hybrid-header.md:106` ("Login/signup/verify/reset keep the public chrome; the root follows the session"), and the new test at `apps/mercato/src/app/(frontend)/__tests__/portal-org-binding.test.tsx:224`, titled *"keeps the public chrome on auth route %s even when the session matches the URL org"*. That test only asserts `authenticated: false` on the **layout prop** — the very value the shell now overrides — so it reads as pinning a behaviour it does not pin. A future reader trusting that test name will believe the public routes are protected against exactly the regression that is now live.

There is also a smaller consequence worth deciding on deliberately: `PortalLayoutShell.tsx:54` still derives `enableEventBridge` from the same layout prop, so on those routes the shell claims to be authenticated while the SSE bridge stays unmounted — a fresh prop/context disagreement of the same family this PR set out to eliminate.

Two clean ways out, either is fine:
- **Keep the documented behaviour**: gate the session prop the same way the chrome is gated, `customerAuth={customerAuthMatchesUrlOrg && (!isPublic || isPortalRoot) ? customerAuth : null}`, so the public routes are genuinely public and the `it.each` test becomes true as written. Please check the portal auth pages first — `login/page.tsx`, `signup/page.tsx`, `invite/page.tsx` and `reset-password/page.tsx` call `usePortalContext()` for `tenant`, which keeps working, but the landing-page redirect relies on `auth.user` and must not regress.
- **Accept the new behaviour**: rename that test to what it actually asserts (something like *"keeps `authenticated: false` on the layout prop for auth route %s"*), correct the PR body and `.ai/analysis/2026-08-26-portal-hybrid-header.md:106`, and add one composed test through `PortalLayoutShell` pinning the chrome those routes really render — plus a decision on whether `enableEventBridge` should follow the context too.

**2. `packages/core/src/modules/portal/__integration__/TC-AUTH-063.spec.ts:17` — the test-case ID collides with a spec already on `develop`.**

`TC-AUTH-063` is taken. `packages/core/src/modules/auth/__integration__/TC-AUTH-063-feature-check-org-scope.spec.ts` ("TC-AUTH-063: feature-check uses the request-selected organization") landed on `develop` in commit `15164f163` (#5544) after this branch forked from `8b492325d`, so it is invisible from the PR head but present in the merge result. CI already proves both exist side by side on the merged ref: the auth spec ran in shard 4/15 (job [98338219649](https://github.com/open-mercato/open-mercato/actions/runs/33016865664/job/98338219649)) and the portal spec in shard 11/15, in the same run.

After merge, `develop` carries two unrelated test cases under one ID: `npx playwright test --config .ai/qa/tests/playwright.config.ts TC-AUTH-063` runs both, and the QA case registry loses its 1:1 mapping from ID to scenario. Please rename to the next free number in the series (`TC-AUTH-064` is unused today — worth re-checking against `develop` immediately before pushing, since the series is moving) and update the file name, the `test.describe` title, the doc comment, and the PR body reference. Reusing the `AUTH` prefix for a portal-module spec is fine and precedented — the sibling `TC-AUTH-033.spec.ts` already lives in this folder — so only the number needs to change.

**3. PR description — the `#3793` claim does not hold on the current code.**

The body states this "Also fixes the 'top of the page still looks logged-out' half of #3793, which reaches the same stale-prop path through the invite page's `router.replace`." The invite page does not use `router.replace`. It calls ``navigateWithPageReload(`/${orgSlug}/portal/dashboard`)`` (`packages/core/src/modules/portal/frontend/[orgSlug]/portal/invite/page.tsx:78`), which is `window.location.assign` (`packages/core/src/modules/portal/lib/navigation.ts:2`), introduced in #3598. A full page load re-runs the layout on the destination path, so the stale-prop path this PR closes is not reachable from invitation acceptance, and `#3793`'s reported symptom is not addressed here. The claim is inherited from #5678's "Related" section, which predates that change — no fault of the implementation, but it should not ship in the PR body.

Nothing auto-closes as a result (`closingIssuesReferences` is empty for every `develop`-targeted PR in this repository, which is why `om-close-fixed-issues` exists), so the risk is a human reading this body and closing an open bug that is still live. Please drop or soften the sentence — "may also help #3793" is defensible; "fixes" is not.

### 💅 Nit

- `packages/core/src/modules/portal/__integration__/TC-AUTH-063.spec.ts:19` — `Date.now()` is a fine uniqueness stamp here, but two shards starting within the same millisecond would collide on `qa-auth-063-{stamp}@test.local`. A random suffix alongside the timestamp removes that entirely. Optional; the spec passed cleanly in CI.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `PortalShell`, `PortalLayoutShell` and `PortalProvider` keep their exports; `PortalLayoutProvider` remains a deprecated re-export.
- [x] No function signature changed in a breaking way — `isPublicPortalRoute` keeps its signature; `isPortalRootRoute` is additive and module-local.
- [x] No required type field removed or narrowed — `PortalShellProps.authenticated` stays `boolean | undefined`; only its runtime precedence changed.
- [x] No event or message name renamed or removed; no payload field removed. The frozen component-replacement handles (`PORTAL_SHELL_HANDLE`, `PORTAL_HEADER_HANDLE`) are untouched.
- [x] Where a contract had to change: the `authenticated` prop precedence change is a runtime-semantics change to a public UI component, and it is called out explicitly in the PR's "Compatibility notes" with a stated migration path (an explicit opt-out prop) if a third party ever needs force-public chrome for a signed-in user. `PortalShell`'s props are not among the surfaces `BACKWARD_COMPATIBILITY.md` enumerates as protected, the documented use (public pages rendering `PortalShell` without a provider, where `user` is `null`) is unaffected, and `packages/ui/src/portal/__tests__/PortalShell.test.tsx:219` pins that case. Accepted as documented — but see finding 1, which is the in-repo consumer where this precedence change has an unintended effect nobody reconciled.

## 🧪 Test Coverage

The coverage that is here is good, and each new test was verified by the author to fail without its fix. `apps/mercato/src/app/(frontend)/__tests__/portal-org-binding.test.tsx:205` pins `authenticated: true` at `/org-a/portal` and `/org-a/portal/` for a matching session, and the pre-existing mismatched-organization cases still pass, so the org-binding guarantee is not weakened. `packages/ui/src/portal/__tests__/PortalShell.test.tsx:159` pins the context-wins precedence and its converse — that a shell with no context user leaves the public chrome alone and fires no nav fetch, which is the case that would have regressed most quietly. `TC-AUTH-063` covers the real entry path (session cookies → `/{org}/portal` → redirect → `portal-nav-ready` → zero `Log In` / `Sign Up` links), creates its own customer and deletes it in `finally`, so it satisfies the self-containment rule in `.ai/qa/AGENTS.md`, and it passed in CI.

The gap is the one finding 1 describes. No test exercises the composed path — real layout props through `PortalLayoutShell` and `PortalProvider` into `PortalShell` — which is precisely where the two halves of the fix meet, and it is the seam where behaviour diverged from the description without anything going red. Whichever resolution you pick, please add one test at that level:

- **If you gate `customerAuth` on the public routes**: assert that `PortalLayoutShell` with `authenticated={false}` and `customerAuth={null}` (the props the layout would then pass for `/{org}/portal/login`) renders `Log In` / `Sign Up` and fires no `/api/customer_accounts/portal/nav` call — and extend the existing `it.each` in `portal-org-binding.test.tsx` to assert `customerAuth: null` on those paths, so the layout half is pinned too.
- **If you keep the current behaviour**: assert the same composition renders the authenticated chrome, and record the deliberate decision about `enableEventBridge` in the test's comment so the next reader knows the asymmetry is intentional.

Either way `packages/ui/src/portal/__tests__/PortalShell.test.tsx` is the natural home, since it already mocks everything the composition needs.

pkarw:

- ❌ `changes-requested` — the code review found three minor findings, and this repository requests changes on Medium/minor findings as well as blockers and majors. The pipeline label moves off `review` so the PR reads as owned by the author again.
- 🧪 `needs-qa` — kept as-is. The change touches `.tsx` under `packages/ui/src/`, so the automated-verification exemption does not apply. Worth noting for whoever picks up QA: `TC-AUTH-063` **did** run and pass in CI (run 33016865664, shard 11/15), so the manual pass can focus on the portal auth routes covered by finding 1 rather than re-running the root entry path.
- 🐛 `bug` — unchanged; this restores intended behaviour rather than adding any.
- 🔺 `priority-high` — unchanged and correct. The broken path is the portal's primary entry point: the admin "Open Portal" action and the `/` rewrite on every custom domain.
- ⚠️ `risk-medium` (was `risk-low`) — re-rated. `PortalShell` is a publicly exported component and the change alters the runtime precedence of its `authenticated` prop, so the blast radius is every portal route in every tenant, the `create-app` template copy, and any third-party consumer — not an isolated path. Finding 1 shows that reach concretely: the new precedence also changes what `/{org}/portal/login|signup|invite|verify|reset-password` render for a logged-in visitor, which nobody intended. That is a wider surface than `risk-low` describes, while still well short of the schema/contract-level exposure `risk-high` is for.

No `qa` label was set — that one belongs to a QA reviewer and means manual QA is in progress.

pkarw:
Thanks @pat-lewczuk — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

The core fix is right and CI is fully green on `ccea4a0`, including `TC-AUTH-063` (it ran in shard 11/15 and passed, so the "not executed" caveat in the Testing section is stale). What is left is: (1) the public auth routes now render the authenticated chrome because `customerAuth` is still handed to the provider on them — decide whether to gate it or to accept and re-document, and make the `it.each` test name match reality either way; (2) `TC-AUTH-063` collides with `TC-AUTH-063-feature-check-org-scope.spec.ts`, which landed on `develop` in #5544 after you branched; (3) the `#3793` sentence in the body — the invite page uses `navigateWithPageReload` (a full page load), not `router.replace`.

pkarw:

The full `om-code-review` pass is posted as the review on `ccea4a0`. Three minor findings drive the verdict under this repository's stricter rule (Medium/minor findings request changes): the public portal auth routes now render the authenticated chrome and contradict the description, the analysis doc and the new test's own name; the `TC-AUTH-063` id collides with a spec that landed on `develop` in #5544 after this branch forked; and the `#3793` claim in the body does not hold because the invite page navigates with a full page load. No blockers and no majors — the fix for #5678 itself is correct, and the change is close.

Validation source: GitHub checks. Every configured `validation.commands` entry maps to a green check on this head, all fifteen `ephemeral-integration` shards passed, and the new `TC-AUTH-063` executed and passed in shard 11/15 — so no local gate run was needed and no CI follow-up is owed.

autofix: skipped (not my PR — re-run with --autofix to fix it here). The PR is handed back to @pat-lewczuk.

pat-lewczuk:
Thanks — finding 1 was a real miss on my part, and the review is right about all three. Addressed in `bbfe4462f`.

**1. Public auth routes rendered the authenticated chrome.** Confirmed. `customerAuth` was never gated on `isPublic`, so once `PortalShell` let a context user upgrade the shell, a matching session produced the authenticated chrome on `/portal/login` and friends — the mirror image of the bug this PR set out to fix, and the exact opposite of what the body, the analysis doc and my own test name claimed.

Took the first of the two routes you offered: keep the documented behaviour. The chrome, the session and the display name now travel together behind one expression:

```tsx
const authenticatedChrome = customerAuthMatchesUrlOrg && (!isPublic || isPortalRoot)
authenticated={authenticatedChrome}
userName={authenticatedChrome ? userName : null}
userEmail={authenticatedChrome ? userEmail : null}
customerAuth={authenticatedChrome ? customerAuth : null}
```

I checked the auth pages first as you asked: `login`, `signup`, `invite`, `reset-password` and `verify` all destructure only `tenant` from `usePortalContext()` and never touch `auth`, and `initialTenant` is independent of the session — so gating `customerAuth` there costs them nothing. The landing-page redirect keeps its session, since the root is `isPortalRoot`. Gating them together also removes the `enableEventBridge` asymmetry you flagged: it derives from the same flag, so there is no longer a route where the shell claims authenticated while the bridge stays down.

Coverage for the seam, per your closing note: the auth-route `it.each` now asserts the withheld session (`customerAuth: null`, `userName: null`, `userEmail: null`) instead of only the flag it pinned before, and two composed tests drive the real layout props through `PortalLayoutShell` → `PortalProvider` → `PortalShell` — one pinning the public chrome and zero `apiCall`s for the auth-route props, one pinning the authenticated chrome for the signed-in props. Reverting just the session gate turns those three auth-route cases red; I verified that rather than assuming it.

**2. Test-case ID collision.** Renamed to `TC-AUTH-064-portal-root-authenticated-shell.spec.ts` — file name, `describe` title, doc comment, fixture prefix and the PR body reference. Re-checked `TC-AUTH-064` free against a freshly fetched `origin/develop` immediately before pushing, per your warning that the series is moving. Thanks for chasing down which commit brought `TC-AUTH-063` in — invisible from this head, so I would not have found it.

**3. The `#3793` claim.** Verified and dropped. The invite page calls `navigateWithPageReload(...)` → `window.location.assign`, not `router.replace`; I had taken that detail from #3793's own "For the developer" note, which predates #3598. The PR body now carries an explicit correction, and I have retracted the misleading comment I left on #3793 itself so nobody closes a live bug on it.

**Nit.** Fixture key is now `Date.now()` plus a random suffix.

Also corrected the stale "Not run … left to QA" line — thanks for digging out the shard-11 job showing it passed. Full configured gate re-run locally on the new head, all green (`yarn test`: 32/32 tasks; `@open-mercato/ui` 1882 tests, `@open-mercato/app` 574).

Re-requesting review.

pat-lewczuk:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-09-08T05:34:38Z.

pat-lewczuk:
🤖 `om-auto-review-pr` taking over the chain lock — review + autofix pass on the base-merged head. Started: 2026-09-08T05:35:51Z.

pat-lewczuk:
# 🔍 Code Review — fix(portal): stop painting the logged-out header over authenticated content

> Posted by `om-auto-review-pr` (running under `om-auto-fix-pr`) as a **comment, not a GitHub review**. This run authenticates as @pat-lewczuk, who is this PR's author, and GitHub rejects a self-submitted review. That is a real gap and not a formality: the verdict below is the automated pass, and a maintainer's approving review is still required before this can merge. @pkarw's `CHANGES_REQUESTED` from `ccea4a0` is still the PR's binding review state and only a new review can clear it.

## 🎯 Summary

This is a re-review of the branch after it was brought up to date: `origin/develop` had advanced **211 commits** since the last CI run, so everything below judges the real merge result (`66179c44f`) rather than the stale head @pkarw reviewed. The base merged cleanly with no conflicts, and `develop` had made no changes at all under `packages/ui/src/portal/`, `apps/mercato/src/app/(frontend)/` or `packages/create-app/template/src/app/` since this branch forked — so there was no semantic-conflict surface either.

The fix itself holds up. The two defects it targets are genuinely independent, and closing only one of them would leave the bug reachable: gating the chrome in the layout (`authenticatedChrome`) fixes the portal root, and inverting the precedence in `PortalShell` (`!!user || (authenticatedProp ?? false)`) closes the class of stale-prop bug that the layout's position above the `[...slug]` segment makes structurally possible. I re-verified the risky half of @pkarw's finding 1 against the merged head rather than trusting the PR body: `login`, `signup`, `invite` and `reset-password` destructure only `tenant` from `usePortalContext()` and `verify` touches the context not at all, so withholding `customerAuth`/`userName`/`userEmail` on the public routes regresses nothing. The layout still resolves the org by slug on those routes (`layout.tsx:96-100`), so `initialTenant` — the only thing they consume — is unaffected.

I also traced every portal navigation that crosses from a public route into an authenticated one, because the whole bug class depends on whether the server layout re-runs. `login` uses `window.location.assign` (`login/page.tsx:50`) and `invite` uses `navigateWithPageReload` (`invite/page.tsx:78`) — both full page loads, so the layout re-renders on the destination path and the withheld session is re-issued. `signup`, `reset-password` and `verify` perform no programmatic navigation at all; they render a success screen whose CTA is a `<Link>` back to `/portal/login`, which is public-to-public and therefore chrome-neutral. The portal root's `router.replace` is the one client-side hop, and it is now safe precisely because `isPortalRoot` keeps the authenticated chrome there. The `#3793` correction in the PR body is accurate.

## Verdict

**Approvable** — no blockers and no un-waivered majors remain after the autofix pass. One blocker existed at the start of this run and was fixed here, not deferred.

## 🧪 Validation Gate

Runner: **local** (no compose `app` container was running). Commands in `.ai/agentic.config.json` order.

| Command | Result | Notes |
|---|---|---|
| `yarn build:packages` | ✅ PASS | 24s |
| `yarn generate` | ✅ PASS | 26s |
| `yarn build:packages` | ✅ PASS | 7s |
| `yarn i18n:check-sync` | ✅ PASS | 57 modules, 5 locales in sync |
| `yarn i18n:check-usage` | ✅ PASS | 137s |
| `yarn typecheck` | ✅ PASS | 32s |
| `yarn test` | ⚠️ PASS in scope | Every package this PR touches is green — `@open-mercato/ui` 237/237 suites (1970 tests), `@open-mercato/app` 91/91 (574 tests), `@open-mercato/shared` 189/189 (2115 tests). See the out-of-scope note below. |
| `yarn build:app` | ⏳ running | Reported in the run summary. |

**Out-of-scope test failures, disclosed rather than hidden.** `create-mercato-app` reported 5 failures, all in `packages/create-app/src/lib/` — a directory this PR does not touch (`git diff origin/develop...HEAD -- packages/create-app/src/` is empty; the only create-app change is the `template/` mirror). Four of them fail with the harness's own diagnostic, `❌ Linux file-watch limits are too low for Turbopack — fs.inotify.max_user_watches: 244178 < 4194304`, which is this machine's sysctl configuration, and the fifth (`share-session-bundle`) exits with a `null` status, i.e. the child process was signalled. Two `@open-mercato/shared` bootstrap tests failed the same way on the first attempt and passed on re-run once `TMPDIR` was shortened — the worktree path overflows the 108-byte unix-socket limit `tsx` needs for its IPC pipe. None of these are attributable to the diff, and CI's own `test` job is the authority.

## 🚫 Findings

### 🔴 Blocker — fixed in this pass

**`packages/core/src/modules/portal/__integration__/TC-AUTH-064-portal-root-authenticated-shell.spec.ts:21` — insecure randomness in a security context** *(inherited from @github-advanced-security, [discussion_r3875752178](https://github.com/open-mercato/open-mercato/pull/5686#discussion_r3875752178); CodeQL alert 198, and the cause of the red `CodeQL` check)*

The `Math.random()` suffix added to fix @pkarw's nit #4 does not only key the email — it also seeds `password: \`Password${stamp}!\``, which is what put it in a security context and made the alert fair rather than noise. Fixed by switching to `randomUUID()` from `node:crypto`, which is the convention every other integration spec in this repo already uses for fixture keys (`TC-APIKEY-007`, `TC-AUTH-044`, and ~15 others import `randomInt`/`randomUUID` from `node:crypto`) and which preserves the cross-shard uniqueness the suffix was added for in the first place. The check was not silenced, suppressed, or allowlisted.

### 🟡 Minor — fixed in this pass

**`UPGRADE_NOTES.md` — the `PortalShell.authenticated` precedence inversion was documented only in the PR body**

`PortalShell` is reachable by third-party modules through `@open-mercato/ui`'s wildcard `exports` (`./*/*`), and `packages/ui/AGENTS.md` documents it as part of the Portal Extension surface. Its type is unchanged and no prop was added or removed, so no `BACKWARD_COMPATIBILITY.md` protected surface is violated — but the *value semantics* of `authenticated={false}` changed from a veto into a default, which is exactly the "STABLE shape, changed value semantics" case that file records for upgraders (the 0.6.8 `BackendChromePayload.settingsSections` entry is the precedent). A PR body is not discoverable at upgrade time. Added an entry under `0.7.0 → 0.7.1 (unreleased)` explaining the old behaviour, the new precedence, why it is a no-op for the documented use, and the migration path for a caller that genuinely wants force-public chrome.

### 🔵 Nit — fixed in this pass

**`packages/ui/src/portal/__tests__/PortalShell.test.tsx:202,287` — `as any` on the auth fixtures**

The repo's root `AGENTS.md` is explicit that `any` is not used. Both fixtures already supplied every required `CustomerAuthContext` field, so the cast was only masking the literal widening of `type: 'customer'` to `string`. Replaced both with a single typed `const signedInAuth: CustomerAuthContext`, which also means the fixtures now fail compilation if that contract drifts instead of silently passing.

### Considered and deliberately not raised

`authenticatedChrome = customerAuthMatchesUrlOrg && (!isPublic || isPortalRoot)` reduces algebraically to "a matching session and not a `PUBLIC_SUFFIXES` route", since `isPortalRoot` implies `isPublic`. I left it as written: the redundant form states the intent the comment above it explains, and collapsing it would make the portal-root carve-out invisible to the next reader.

## 💥 Breaking Changes

| Surface | Changed? | Assessment |
|---|---|---|
| Exported types / signatures | No | `PortalShellProps` is untouched; `authenticated?: boolean` keeps its type. |
| Component prop **semantics** | **Yes** | `authenticated={false}` no longer overrides a signed-in context. Not an enumerated protected surface; now documented in `UPGRADE_NOTES.md` with a migration path. A test pins the no-provider case, which is the documented public use. |
| Import paths, widget spot IDs, replacement handles | No | The frozen `page:portal:layout` / `section:portal:*` handles are untouched. |
| API routes, DB schema, event IDs, DI keys, ACL features, CLI | No | — |

Per `packages/ui/AGENTS.md`, portal shell behaviour is an "Ask First" surface. That bar is met here: the change is driven by #5678, and @pkarw reviewed the design and explicitly offered the two routes the author chose between.

## 🧪 Test Coverage

Coverage is genuinely good, and unusually so in that it tests the **seam** rather than each half in isolation. `portal-org-binding.test.tsx` pins the portal root with and without a trailing slash, and an `it.each` pins that `/portal/login|signup|invite` withhold the session itself (`customerAuth: null`, `userName: null`, `userEmail: null`) — asserting the withheld session rather than only the flag is what makes the test capable of catching a regression of @pkarw's finding 1. `PortalShell.test.tsx` covers the precedence inversion in both directions plus two composed tests that drive the real layout props through `PortalLayoutShell` → `PortalProvider` → `PortalShell`, including asserting that **zero** `apiCall`s fire on the public path. `TC-AUTH-064` covers the whole flow end-to-end against a live app and is self-contained (creates its own customer, deletes it in `finally`). The mirrored template test is not executed by any local task — `packages/create-app`'s test script globs `src/**/*.test.ts` only — but the two files are byte-identical to their `apps/mercato` counterparts, which I verified with `diff`.

## ⏳ CI

Required checks could not be read from branch protection (`Required status checks not enabled`, HTTP 404), so per the tracker descriptor **every reported check is treated as required**. CI is re-running on `66179c44f`; at the time this verdict was written the integration shards and `test` were still in progress, and `CodeQL` was red on the prior head from the finding fixed above. This verdict covers the work, not a green run — the CI outcome is reported separately in the run summary.

pat-lewczuk:

- 🐛 `bug` — unchanged. The PR restores intended behaviour: an active, org-matching customer session was rendering the public logged-out shell around authenticated content on the portal's primary entry point.
- 🔍 `review` — moved back from `changes-requested`. All three of @pkarw's findings plus the nit were addressed by the author in `bbfe4462f`, and I re-verified each one against the current merged head rather than taking the PR body's word for it; the remaining CodeQL blocker is fixed in `66179c44f`. The PR is therefore no longer owned by the author and is genuinely awaiting a reviewer again. It is deliberately **not** `merge-queue`: that label means "approved, ready to merge", and this PR has no approving review — @pkarw's `CHANGES_REQUESTED` is still its binding review state, and this automation runs as the PR author so it cannot submit the review that would clear it.
- 🧪 `needs-qa` — kept. The diff changes `.tsx` under `packages/ui/src/` and alters what a customer actually sees on the portal, so the automated-verification exemption does not apply. With the QA gate on, this PR stays unmergeable until a QA reviewer adds `qa-approved`; this run has not added and will not add that label. Useful scoping for whoever picks QA up: the portal root under a matching session is the fix, and the five public auth routes (`login`, `signup`, `invite`, `reset-password`, `verify`) are the regression surface, because the fix now withholds the session on them.
- 🔺 `priority-high` — unchanged. The broken path is the portal's primary entry point: the admin "Open Portal" action links to it and `proxy.ts` rewrites `/` to it on every custom domain. It is not cosmetic either — `enableEventBridge` was false, so the portal SSE bridge never mounted for the rest of the session.
- ⚖️ `risk-medium` — unchanged, and still the right call after the base merge. The change is small and well covered, but it touches a shared exported UI contract (`PortalShell`) whose value semantics it inverts, so the blast radius is wider than a single module. That is now written up in `UPGRADE_NOTES.md` rather than living only in the PR body.
- 🔒 `in-progress` — held by this run for the duration of the `om-auto-fix-pr` chain; it is swapped for `ci-monitoring` when the run reports, since a CI-result follow-up is still owed.

pat-lewczuk:

This run took PR #5686 through the merge-ready loop: base merge → review + autofix → CI. It did **not** merge anything and did not add any QA label. Head is now `66179c44f`.

## 🔀 Base merge

`origin/develop` had advanced **211 commits** since this branch last saw CI, so the checks @pkarw reviewed against were measuring a diff that no longer existed. `develop` was merged in first (`0013a518f`), before any review or CI work, so everything below judges the real merge result.

The merge was **clean — no conflicts**. It is also low-risk beyond that: `develop` made no changes at all under `packages/ui/src/portal/`, `apps/mercato/src/app/(frontend)/`, or `packages/create-app/template/src/app/` since this branch forked, so there was no semantic-conflict surface either.

## 🔍 Review

Full report posted above. Verdict: **approvable** — no blockers and no un-waivered majors remain.

**One important caveat, stated plainly rather than buried.** This automation authenticates as @pat-lewczuk, who is this PR's author, so it **cannot submit a GitHub review** — GitHub rejects self-review, and self-approval would not be a real review anyway. @pkarw's `CHANGES_REQUESTED` from `ccea4a0` therefore remains this PR's binding review state and **a maintainer's approving review is still required**. That is why the pipeline label is `review` and not `merge-queue`.

### How each reviewer comment already on the PR was handled

| Ask | Source | Outcome |
|---|---|---|
| Public auth routes rendered the authenticated chrome | @pkarw, finding 1 | **Fixed by the author** in `bbfe4462f`. Re-verified independently against the merged head, not taken on trust: `login`, `signup`, `invite` and `reset-password` destructure only `tenant` from `usePortalContext()` and `verify` does not touch the context at all, so withholding the session on those routes regresses nothing. |
| `TC-AUTH-063` id collision | @pkarw, finding 2 | **Fixed by the author.** Re-confirmed `TC-AUTH-064` is still free against current `develop`. |
| The `#3793` claim in the body does not hold | @pkarw, finding 3 | **Fixed by the author.** Independently confirmed: `invite/page.tsx:78` uses `navigateWithPageReload`, a full page load, so the stale-prop path closed here is not reachable from invite acceptance. #3793 stays open and untouched. |
| `Date.now()` fixture collision across shards | @pkarw, nit 4 | **Fixed by the author, then hardened here** — see the CodeQL finding below, which the author's fix introduced. |
| Insecure randomness (`Math.random()` in a security context) | @github-advanced-security, [discussion_r3875752178](https://github.com/open-mercato/open-mercato/pull/5686#discussion_r3875752178) | **Fixed in this run** (`66179c44f`). |

Nothing a reviewer asked for was silently dropped.

### Fixed in this run — `66179c44f`

- 🔴 **Blocker — CodeQL insecure randomness.** The `Math.random()` suffix added for nit 4 also seeds `password: \`Password${stamp}!\``, which is what made the alert legitimate rather than noise, and it was the cause of the red `CodeQL` check. Switched to `randomUUID()` from `node:crypto` — the convention ~15 other integration specs in this repo already use — which keeps the cross-shard uniqueness the suffix was added for. **The check was fixed, not silenced:** nothing was suppressed, allowlisted, or weakened.
- 🟡 **Minor — undocumented semantics change.** `PortalShell`'s `authenticated={false}` changed from a veto into a default. No prop or type changed, so no `BACKWARD_COMPATIBILITY.md` protected surface is violated — but the prop is reachable by third-party modules through `@open-mercato/ui`'s wildcard exports, and a PR body is not discoverable at upgrade time. Added an `UPGRADE_NOTES.md` entry under `0.7.0 → 0.7.1 (unreleased)`.
- 🔵 **Nit — two `as any` casts** on the auth fixtures in the new tests, against the repo's no-`any` rule. Replaced with one typed `CustomerAuthContext` const, which also makes the fixtures fail compilation if that contract drifts.

## 🧪 Validation

Runner: **local** (no compose `app` container running). Full `.ai/agentic.config.json` gate, in order, all green:

`build:packages` ✅ → `generate` ✅ → `build:packages` ✅ → `i18n:check-sync` ✅ → `i18n:check-usage` ✅ → `typecheck` ✅ → `test` ✅ (in scope) → `build:app` ✅ (204s)

Every package this PR touches passes: `@open-mercato/ui` 237/237 suites (1970 tests), `@open-mercato/app` 91/91 (574 tests), `@open-mercato/shared` 189/189 (2115 tests) — including `PortalShell.test.tsx` and `portal-org-binding.test.tsx`.

**Two environmental failures, disclosed rather than quietly dropped.** `create-mercato-app` reported 5 failures, all under `packages/create-app/src/lib/` — a directory this PR does not touch (`git diff origin/develop...HEAD -- packages/create-app/src/` is empty). Four print the harness's own diagnostic, `❌ Linux file-watch limits are too low for Turbopack — fs.inotify.max_user_watches: 244178 < 4194304`; the fifth exits with a `null` status, i.e. signalled. Separately, two `@open-mercato/shared` bootstrap tests failed once and passed on re-run after `TMPDIR` was shortened — this worktree's path overflows the 108-byte unix-socket limit `tsx` needs for its IPC pipe. Neither set is attributable to the diff, and CI's own `test` job is the authority on both.

## ⏳ CI

**CI is still running on this head.** Branch protection plus the QA-approval gate hold the actual merge; this run's verdict covers the work, not a green run. Required checks could not be read from branch protection (`Required status checks not enabled`, HTTP 404), so per the tracker descriptor **every reported check is treated as required**. At the time of writing, the CodeQL and scope workflows have started on `66179c44f` and the main workflow (`test`, `lint`, `docker-build`, the 15 `ephemeral-integration` shards) had not yet registered. A follow-up comment will report the CI outcome.

For the record, the `ephemeral-integration (12/15)` failure visible on the old head could not be diagnosed — GitHub had already expired that run's logs. It is superseded by the fresh run against the current base either way.

## 🔁 Follow-up filed

- **#5958** — *refactor(portal): make boundary-crossing portal navigations consistently full page loads.* Two pre-existing hazards this review surfaced while tracing the bug class, deliberately kept out of this PR rather than expanding its scope: `useCustomerAuth().logout` navigates with `router.push` where its `PortalContext` sibling does a full page load (the mirror image of #5678, one call site away from being reachable), and `login/page.tsx:50` open-codes `window.location.assign` instead of using the `navigateWithPageReload` helper that exists for exactly this rule and currently has one production caller. Filed directly rather than through `om-followup-issue-from-pr` because the finding originates from this review rather than from a PR comment that skill parses.

## 🚀 Merge readiness

**Not merge-ready yet, and this skill never merges regardless.** Two things are outstanding, neither of which an automation can supply:

1. **An approving review from a maintainer.** @pkarw's `CHANGES_REQUESTED` still stands and this run cannot clear it (author identity). The code work it asked for is done.
2. **QA sign-off.** `needs-qa` is on and the QA gate is enabled, so `qa-approved` from a QA reviewer is required. This run has not added it and will not. Suggested QA scope: the portal root under a matching session is the fix; the five public auth routes (`login`, `signup`, `invite`, `reset-password`, `verify`) are the regression surface, since the fix now withholds the session on them.

Once both land and CI is green, `om-approve-merge-pr 5686` is the next step.

pat-lewczuk:

**✅ All checks are green on `66179c44f`** — 28 successful, 5 skipped, 0 failing, 0 pending. This resolves the pending-CI caveat in the run summary above; the verdict there stands unchanged.

Two of these were the open questions and both are now answered:

- **`CodeQL` ✅** — was red on both previous heads. The insecure-randomness alert ([alert 198](https://github.com/open-mercato/open-mercato/security/code-scanning/198), [discussion_r3875752178](https://github.com/open-mercato/open-mercato/pull/5686#discussion_r3875752178)) is cleared by the `randomUUID()` fix in `66179c44f`. It went green because the finding was fixed, not because anything was suppressed or allowlisted.
- **`ephemeral-integration (12/15)` ✅** — this was the shard failing on the old head, and GitHub had already expired its logs so it could never be diagnosed directly. It passes on the current base. All 15 shards are green, so `TC-AUTH-064` ran and passed against a live app on the merged result.

Also green: `test`, `lint`, `ds-lint`, `docker-build`, `documents-multi-instance`, `audit-scope`, `prepare`, `scope`, `license/cla`, and all three CodeQL `Analyze` jobs. Skipped (not applicable to this PR): `audit`, `merge-coverage`, `mutate`, `comment`, `[code]smith`.

Note that branch protection reports `Required status checks not enabled` (HTTP 404), so per the tracker descriptor every reported check was treated as required — the above is therefore the strictest reading available, not a subset.

**This does not change the merge verdict, and CI green is not the thing blocking this PR.** `mergeStateStatus` is still `BLOCKED` and `reviewDecision` is still `CHANGES_REQUESTED`. Two human actions remain:

1. **An approving review from a maintainer.** @pkarw's `CHANGES_REQUESTED` from `ccea4a0` is still binding. Every ask in it was addressed and independently re-verified against the merged head (see the review comment above), but this automation runs as the PR author and cannot submit the review that clears it.
2. **`qa-approved` from a QA reviewer.** `needs-qa` is set and the QA gate is on. QA scope: the portal root under a matching session is the fix; the five public auth routes (`login`, `signup`, `invite`, `reset-password`, `verify`) are the regression surface, since the fix now withholds the session on them.

Once both land, `om-approve-merge-pr 5686` is the next step. Dropping `ci-monitoring` — this was the follow-up it promised, and no further follow-up is owed from this agent.

Kapsik89:
## Manual QA — ✅ GO (2026-09-22)

Head `66179c44f` vs a `develop` baseline (`9248c1459`), both built from source and run as production
builds against the same database and browser session. Compiled bundles were grepped for the fix
marker before any live result was trusted.

I reproduced #5678 on `develop` first: entering `/acme-corp/portal` with a live session gave the
logged-out header over a dashboard greeting the customer by name, no sidebar, no portal-nav fetch.
On this branch every one of those fields inverts. A control on the same build and session (direct
`/dashboard` entry) rendered correctly, so the missing `/portal/nav` on `develop` is a real
difference and not a broken probe.

| # | Scenario | Result |
|---|---|---|
| 1 | Baseline on `develop`: root entry → public header over authenticated content | ✅ reproduced (Log In/Sign Up present, Log Out 0, `portal-nav-ready` ABSENT) |
| 2 | Baseline control: direct `/dashboard`, same build + session | ✅ correct — proves the probe sees `/portal/nav` |
| 3 | Root entry with a matching session | ✅ Log In 0, Log Out 1, `portal-nav-ready` true, sidebar 3 links |
| 4 | Trailing slash `/portal/` | ✅ identical |
| 5 | Redirect to `/dashboard` with chrome intact | ✅ |
| 6 | Portal nav fetch | ✅ `/portal/nav` 200 |
| 7 | Portal SSE event bridge | ✅ `/portal/events/stream` 200 on root entry, absent on public routes |
| 8 | `/portal/login` with a matching session keeps public chrome | ✅ 0 API calls |
| 9 | `/portal/signup` | ✅ 0 API calls |
| 10 | `/portal/reset-password` — **not in the new `it.each`** | ✅ 0 API calls |
| 11 | `/portal/verify` — **not in the new `it.each`** | ✅ 0 API calls |
| 12 | `/portal/invite` | ✅ 0 API calls |
| 13 | Name/email withheld from HTML on all five public auth routes | ✅ absent — on `develop` both were **present** on `/portal/login` |
| 14 | Cross-org session at a foreign portal root | ✅ public landing page, correct org name, no identity leak |
| 15 | Cross-org session at a foreign portal dashboard | ✅ access denied |
| 16 | Anonymous visitor at the root | ✅ landing page — chrome is genuinely session-gated |
| 17 | Identity swap (customer 1 → logout → customer 2) at the root | ✅ correct identity, no trace of the previous user |
| 18 | Admin "Open Portal" action target | ✅ resolves to exactly `/{orgSlug}/portal` |
| 19 | Logout from a root-entry session | ✅ profile → 401, clean return to public chrome |
| 20 | A11y of the newly-rendered authenticated chrome | ✅ lang, 5 landmarks, named nav, real `<button>`, keyboard + visible focus |
| 21 | Unit gate `PortalShell` / `portal-org-binding` | ✅ 6/6 and 14/14 |
| 22 | Client-side hop from a public auth route into the root | ⚠️ see below |

Row 13 is worth a look on its own: withholding the session does more than tidy the chrome. On
`develop`, `/portal/login` with an active session serialised the customer's display name and email
into the page and fetched `/portal/profile` from a public page. Here all five auth routes make zero
API calls and carry no identity in the HTML.

### Filed

| Finding | Severity | Filed as |
|---|---|---|
| Client-side navigation from a public auth route into the portal root leaves a signed-in customer on the anonymous landing page ("Log in / Create account") while `/portal/profile` still returns 200. Self-heals on a full reload. | Low | **#6362** — follow-up, not a change here |
| Sidebar landmark `aria-label="Portal navigation"` (English) on a Polish UI; Quick Links cards in English; no skip link | Low | **Not filed** — pre-existing, untouched by this diff |

On row 22: not a regression introduced here. On `develop` the same hop would have produced the
original hybrid instead, because the context there did hold a user — I measured that, name and email
were present in the login page's HTML. Same structural gap, different symptom. It does qualify the
"closes the class of bug" claim in the description and `UPGRADE_NOTES.md`: this branch of the class
is closed by *withholding* state rather than refreshing it, so a client-side hop starting from a
withheld-session route is still wrong. The optional hardening already sketched in #5678 —
server-redirecting `/{org}/portal` to the dashboard when a matching session cookie is present —
would close it. I did not re-click it on a `develop` build, so that attribution is inference from
the measured context state plus the unchanged redirect code.

Two non-blocking suggestions: add `/portal/reset-password` and `/portal/verify` to the new `it.each`
(it covers three of the five `PUBLIC_SUFFIXES`, and this is the exact case the first revision got
wrong), and see #6362 for the follow-up.

Not covered: the custom-domain `/` rewrite (no domain mapping locally, though it targets the same
route as rows 3–7), the `create-app` template mirror (byte-identical, not executed by any local
task), and `TC-AUTH-064` itself, which passed in CI.

Screenshots and the full scenario log are local — this flow cannot attach images here, but I can
supply them on request.

patzick (approved):

PR #5688, opened 2026-08-26
fix(ai_assistant): claim pending-action confirmations atomically

pat-lewczuk:
## Problem

Confirming a pending action was a read-then-act sequence with no atomic claim.

`executePendingActionConfirm()` guards on `action.status !== 'pending'` using an
entity the route read earlier, and `AiPendingActionRepository.setStatus()` opens
`em.transactional`, issues a plain `SELECT` (no `FOR UPDATE`), evaluates the
transition in JavaScript and then writes. A transaction is not a mutex: under
READ COMMITTED every concurrent caller reads `pending`, every caller passes the
guard, and every caller invokes the tool handler.

The entity carries no version column, so MikroORM optimistic locking provided no
backstop either. The route's idempotency short-circuit only covers *sequential*
retries.

Result: one approved action executes its mutation handler N times under N
concurrent confirmations. No LLM involvement is needed — a double-click or a
client retry reaches the same path.

Reproduced against the unpatched code with two concurrent confirmations of a
single approved action: **two handler invocations, both reporting success.**

## Fix

Replace the `pending → confirmed → executing` write pair with a single atomic
compare-and-swap claim. The status predicate is part of the `UPDATE`, so exactly
one caller can win:

```ts
nativeUpdate(AiPendingAction, { id, tenantId, organizationId, status: 'pending' },
             { status: 'executing', ... })
// claimed = affected === 1
```

Losing callers never reach the handler. The state machine becomes
`pending → executing → confirmed | failed`, which also removes the odd
intermediate state where a row was marked `confirmed` *before* the mutation ran.

## Contract changes — resolved after review

Only one contract change remains, and it is additive:

**New `202` response** on the confirm endpoint (`confirmation_in_progress`)
when another request holds the claim. Documented in
`apps/docs/docs/framework/ai-assistant/mutation-approvals.mdx` (step list,
response-contract table, HTTP-routes table) and covered by an integration
case in `TC-AI-ACTIONS-PENDING-004`.

**`AI_PENDING_ACTION_ALLOWED_TRANSITIONS` is no longer narrowed.** The first
revision removed `pending → confirmed` and made `confirmed` terminal. Review
(#5688) correctly flagged that as a breaking narrowing of a surface exported
from the package root, shipped without the `BACKWARD_COMPATIBILITY.md`
deprecation protocol — and, decisively, one the security fix does not need:
`claimForConfirmation` writes through `nativeUpdate` and never consults the
table, and the confirm route no longer reaches `setStatus` for those edges.
Both edges are restored; `pending → executing` remains as a purely additive
edge. `scenario-14b` pins the table as a superset of the pre-claim contract so
the narrowing cannot come back by accident.

## Known gap (follow-up, not a blocker)

If the process dies after the claim, the row stays in `executing` and the action
becomes permanently unconfirmable — every retry gets `202`. The pre-existing code
had the same class of problem. A reaper for stale `executing` rows should follow.

## Verification

Full local gate on the post-review head (with the latest `develop` merged in):
`build:packages`, `generate`, `i18n:check-sync`, `i18n:check-usage`,
`typecheck`, `lint`, `build:app` — all green. Test suites:
**`ai-assistant` 1487/1487** and **`ui` 1937/1937**, no failures and no
regressions. The `agent-registry.test.ts` idempotency failure noted in the
first revision is gone — it was fixed on `develop` and came in with the base
merge.

The UI regression test for the `202` path was verified to actually
discriminate: it fails when the `confirmation_in_progress` guard is removed
and passes with it, and it ships beside a contrast case proving the guard is
not blanket.

## Credit

Reported and fixed by **Łukasz Jakubowski (@Szesnasty)** under private
responsible disclosure. Commit authorship is preserved.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5688"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790373498&installation_model_id=432243&pr_number=5688&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5688&signature=1d85fe5de1549fe155e6cfa05bde627b41609b59498ff7a9c6ca4982ea18e16c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pat-lewczuk:
Label rationale: `priority-high` — approval-integrity defect in a package already published to npm. `risk-medium` — the fix itself is contained, but it changes an exported state-machine constant (`AI_PENDING_ACTION_ALLOWED_TRANSITIONS`) and adds a new `202` response on the confirm endpoint, so the blast radius reaches external consumers. `needs-qa` rather than `skip-qa` precisely because the API surface changed — the automated-verification exemption does not apply. Reviewers should focus on the two contract changes called out in the description and decide whether they need a deprecation bridge under BACKWARD_COMPATIBILITY.md.

pkarw:
🤖 `om-auto-review-pr` claimed this PR for review at 2026-08-26T23:35Z. Lock: `in-progress` label + assignee. It will be released when the review is submitted.

pkarw (changes requested):
# 🔍 Code Review: fix(ai_assistant): claim pending-action confirmations atomically

## 🎯 Summary

This replaces the `pending → confirmed → executing` write pair in the mutation-approval confirm path with a single compare-and-swap claim (`AiPendingActionRepository.claimForConfirmation`, `data/repositories/AiPendingActionRepository.ts:304`), so the status predicate travels inside the `UPDATE` and only one concurrent caller can reach the tool handler. **The diagnosis is correct and the shape of the fix is the right one.** `setStatus` really was a read-then-act sequence — `em.transactional` + a plain `SELECT` with no `FOR UPDATE` (`:245-259`), evaluated in JavaScript at `:266-269` — and under READ COMMITTED a transaction is not a mutex, so every concurrent confirmer read `pending`, passed the guard, and invoked the handler. The entity carries no version column (`data/entities.ts:122-211`), so MikroORM optimistic locking was no backstop either. Moving the predicate into the write is the standard, minimal answer, and collapsing the state machine so the row is never marked `confirmed` *before* the mutation runs is a genuine improvement over the old intermediate state. The scope discipline is good too: 8 files, no entity change, no migration, and the tenant/organization filter is carried into both the `UPDATE` and the follow-up read.

Reviewed scope: the repository claim, the executor's use of it, the route's two new 202 paths, the state-machine constant, and the four test files — plus the in-repo consumers of the confirm endpoint (`packages/ui/src/ai/parts/`), the public mutation-approval documentation, and `BACKWARD_COMPATIBILITY.md`, none of which the diff touches.

What holds this back is not the fix. It is that **the one genuinely breaking part of the change is not required by the fix**, and that the new `202` reaches the product's own client as a false error message.

**Validation source: GitHub checks** (this repo's checks-first extension) — all 18 check runs on `25c89e1627c3448d58d62c5b9130d7d11de11548`, 13 `success` and 5 `skipped` by design (`audit`, `comment`, `mutate`, `merge-coverage`, `[code]smith`). No local gate was run. I opened the `prepare`, `test` and `ephemeral-integration` **step** lists individually rather than trusting job colour, so nothing hid behind an early red step; every configured gate command maps to a named green step. `ephemeral-integration (none)` is the affected-only single-runner mode, not a skip — it ran 79 tests in 2.0m and its log shows `TC-AI-ACTIONS-PENDING-004-confirm-cancel-mutations.spec.ts` passing both cases, so the existing real-Postgres coverage for this endpoint genuinely executed on this head.

## Verdict

❌ **request changes** — one Major (the exported state-machine constant is narrowed without the deprecation protocol, and the narrowing is separable from the security fix) and three Minors. Per this repository's review extension, Minors request changes too. Nothing here disputes the atomic claim itself: **fixing the Major alone is a two-line change and flips this to approve**, with the rest fair to take as follow-ups if you prefer.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | `prepare` step 12 *Build packages* |
| `yarn generate` | ✅ PASS | `prepare` step 13 *Prepare generated modules* |
| `yarn build:packages` (second) | ✅ PASS | `prepare` step 14 *Rebuild packages with generated files* |
| `yarn i18n:check-sync` | ✅ PASS | `test` step 17 *Check i18n sync* |
| `yarn i18n:check-usage` | ✅ PASS | `test` step 18 *Check i18n usage* (this step is `continue-on-error` in `ci.yml`, so its green is weaker than the others — irrelevant here, the diff adds no user-facing strings) |
| `yarn typecheck` | ✅ PASS | `test` step 21 *Checking types* |
| `yarn test` | ✅ PASS | `test` step 22 *Test*, plus the three unfiltered guard steps 24/25/26 |
| `yarn build:app` | ✅ PASS | `prepare` step 15 *Build app*, plus `ephemeral-integration (none)` and `docker-build` |

One note on reading the green: `mutate` reports `skipped`, not `success`. `scripts/stryker/scope.mjs` freezes its package allowlist to `['shared']`, so no `packages/ai-assistant` diff is ever mutation-tested — that check carries no signal on this PR either way. Mentioned only so nobody counts it as coverage.

## Findings

### ⚠️ Major

**`lib/pending-action-types.ts:39-45` — the exported transition table is narrowed without the deprecation protocol, and the narrowing is not needed for the fix.**

`AI_PENDING_ACTION_ALLOWED_TRANSITIONS` is re-exported from the package root (`packages/ai-assistant/src/index.ts:191`), so it is a public surface of a package published to npm. Three things change in it: `pending` gains `executing` (additive, fine), `pending` **loses** `confirmed`, and `confirmed` becomes terminal (was `['executing']`). A third-party module that calls `repo.setStatus(id, 'confirmed')` on a `pending` row — or reads `canTransitionAiPendingAction('pending', 'confirmed')` at `:174` — goes from working to throwing `AiPendingActionStateError` with no bridge, no `@deprecated` marker, no `UPGRADE_NOTES.md` entry, and no spec carrying a "Migration & Backward Compatibility" section.

The part that decides this is that **neither removal is load-bearing for the security fix.** `claimForConfirmation` writes through `em.nativeUpdate` (`AiPendingActionRepository.ts:317`) and never consults the transition table at all, and I checked every production `setStatus` call site: the cleanup worker writes `expired` (`workers/ai-pending-action-cleanup.ts:213`), the cancel helper writes `expired` and `cancelled` (`lib/pending-action-cancel.ts:128`, `:158`), and the executor writes `failed` and `confirmed` **from `executing`** (`lib/pending-action-executor.ts:404`, `:441`) — an edge the old table already allowed. Nothing in the new flow needs `pending → executing` in the allow-list either; only the rewritten `scenario-14` test walks that edge through `setStatus`. So the entire pre-existing table would work unchanged with the new claim, and retaining `pending: ['confirmed']` / `confirmed: ['executing']` cannot reopen the hole, because the confirm route no longer reaches `setStatus` for those transitions at all.

That matters against `BACKWARD_COMPATIBILITY.md`'s Emergency Security Exception, which is the only thing that could waive the ordinary protocol here. Its requirement 2 is explicit — *"No unrelated tightening rides along"* — and its qualifying condition is that the removed surface **is itself** the vulnerability, with *"a surface that merely makes an insecure usage possible does not qualify."* The vulnerability is the non-atomic read-then-write in `setStatus`, not these two edges. The document's own recorded precedent for this class (the `BackendChromePayload.settingsSections.id` entry under § 2, *"STABLE field shape, changed value semantics"*) is a dated entry plus a documented migration, not a silent narrowing.

Two ways to close it, either is fine:

1. **Drop the tightening** — keep `pending: ['confirmed', 'cancelled', 'expired', 'executing']` and `confirmed: ['executing']`, adjusting the two new `scenario-14` assertions that pin `confirmed` as terminal. Two lines, no weakening of the claim.
2. **Keep it and ship it properly** — an `UPGRADE_NOTES.md` entry naming what stops working and for whom, a spec under `.ai/specs/` with a "Migration & Backward Compatibility" section, and a dated entry at the end of `BACKWARD_COMPATIBILITY.md`. Requirement 5 also needs a named human maintainer sign-off; an automated review cannot clear that, which is part of why I am handing the decision back rather than making it.

Your PR body asks for exactly this call, so: option 1 unless you actively want the stricter table, in which case option 2.

### 🔹 Minor

**`packages/ui/src/ai/parts/MutationPreviewCard.tsx:96` — the product's own client renders the new `202` as "The mutation handler reported an error."**

`apiCall` sets `ok: response.ok` (`packages/ui/src/backend/utils/apiCall.ts:98`), and `202` is a 2xx, so `postJson` returns `{ ok: true, data }` and the card falls through to the HTTP-200 branch. There, `result.data?.ok === false` is true (the new body sets `ok: false`) while `mutationResult` is `null`, so `handlerError` is `undefined`, `mappedCode` collapses to `'execution_failed'` at `:100`, and the message falls back to the generic `ai_assistant.chat.mutation_cards.preview.handlerError` string at `:106-109`. That envelope is threaded into `ConfirmationCard` as `payload.confirmError` (`:175`), seeded into `localError` at `ConfirmationCard.tsx:76-78`, and rendered by `ConfirmationErrorAlert` at `:172`.

So in the exact scenario this PR exists to handle — a double-click or a client retry — the operator is told the mutation handler failed, while the winning request is in fact succeeding. Before this PR the losing caller got `200 {ok: true}` and no alert (it also ran the handler a second time, which is the bug you are fixing — the point is only that the *message* is a regression). It does self-heal: `useAiPendingActionPolling` keeps polling while the status is `executing` (`:18-20`), and `MutationPreviewCard.tsx:151` short-circuits into `MutationResultCard` as soon as the row goes terminal, so this is a misleading interstitial rather than a stuck or wrong final state. Fix: branch on `code === 'confirmation_in_progress'` (or on the 202 status, which `postJson` already has in `call.status`) before the handler-error mapping, and render an "already in progress, waiting for the result" state instead of an error alert.

**`packages/ai-assistant/src/modules/ai_assistant/__integration__/TC-AI-ACTIONS-PENDING-004-confirm-cancel-mutations.spec.ts` — the new `202` response has no integration coverage.**

The root `AGENTS.md` requires integration tests for every affected API path to ship in the same change, and this PR adds a response status to a documented endpoint with only Jest coverage. The spec is already the right home and needs almost nothing: its `seed({ status })` helper writes arbitrary statuses directly (it seeds `confirmed`, `cancelled` and an expired `pending` row today at `:116`, `:130`, `:139`), so seeding `status: 'executing'`, POSTing confirm, and asserting `202` with `code: 'confirmation_in_progress'` is roughly six lines beside the existing `409 invalid_status` case. That spec ran green on this head, so the addition costs no new CI surface.

Separately, and worth saying plainly: **the atomicity itself is currently unpinned.** The `nativeUpdate` mock added at `data/repositories/__tests__/AiPendingActionRepository.test.ts:354` is a JavaScript compare-and-swap, and its `rowMatchesWhere` does honour `status` (`:49`), so the three new repository tests are not vacuous — but a read-then-check-then-write reimplementation of `claimForConfirmation` would pass all of them identically, because single-threaded JS makes that atomic too. The same is true of the twenty-caller test at `lib/__tests__/pending-action-executor.test.ts:558`: it exercises the stub's claim, proving the executor honours `claimed: false`, not that the SQL predicate is where the exclusion happens. Those are still worth having — they pin the contract — but the security property the PR is named for has no test that could fail. A real concurrent-confirm integration test is genuinely awkward here (the spec deliberately avoids driving a live `pending → confirmed` because it would execute a real tool mutation, per its own header comment at `:39-41`), so recording your two-caller repro as a documented negative control, or filing it as a follow-up, is a reasonable disposition — but the `202`-on-`executing` case is cheap and should ship.

**`apps/docs/docs/framework/ai-assistant/mutation-approvals.mdx:82` — the public documentation now states the opposite of what the code does.**

That line reads *"Atomic status transition `pending → confirmed → executing`"* — precisely the sentence this PR falsifies — and the surrounding step list (`:82-88`) and the state-machine section above it still describe the old shape, with no mention of the `202`. This is not incidental drift: `packages/ai-assistant/AGENTS.md` names these public docs as the source of truth *over* the AGENTS.md when the two disagree, so the authoritative description of the mutation-approval lifecycle is now wrong. Update the diagram, the `executePendingActionConfirm` step list, and add the `202` to the endpoint's documented responses. While you are there, `lib/pending-action-executor.ts:326-329` now has two bullets that both say "already `confirmed` → return the stored result"; the second one replaced the "without a stored `executionResult`" case and reads as a duplicate.

### 💅 Nit

**`data/repositories/AiPendingActionRepository.ts:316` — `const now = extra?.now ?? new Date()` is computed and never used.** `now` stays in the method's `Pick<AiPendingActionSetStatusExtra, …>` signature and every caller passes it (`pending-action-executor.ts:373`), so the API advertises a clock the implementation ignores. Either drop the parameter or stamp something with it.

**`api/ai/actions/[id]/confirm/route.ts:80-87` — a claim lost to a `cancelled` or `expired` row answers `200` with `ok: false`, not the `409` the same block declares.** When the row moves to a terminal state between the recheck and the claim, the executor returns `invalid_status` (`pending-action-executor.ts:382-392`) and the route falls through to the generic `200` return at `:241`, while the `errors` array still says `409` covers *"Pending action is in an incompatible terminal status."* Nothing regresses — before this PR that same race threw `AiPendingActionStateError` out of `setStatus` into the route's catch and produced a `500`, so `200` is strictly better — it is only an inconsistency in a PR that is otherwise careful about declaring its response surface.

**The known-gap disclosure understates one consequence, and the follow-up should cover it.** A row wedged in `executing` is not merely unconfirmable: `executing` has no `cancelled` edge, so `POST /api/ai/actions/:id/cancel` now throws the state-machine error too, and the cleanup worker only scans `status = 'pending' AND expires_at < now` — so the TTL never reaches the row either. The operator has no escape from the UI at all. Your framing that the old code had the same class of problem is fair (a crash after the `confirmed` write left the retry returning a hollow `200` success), but the reaper follow-up should reclaim the row for cancel as well as for confirm.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed — `claimForConfirmation` and `AiPendingActionClaimResult` are additive; nothing is deleted.
- [x] No function signature changed in a breaking way — `setStatus` is untouched.
- [x] No field removed or retyped in an existing response shape — the `202` body is a new shape on a new status, and the `200` shape is unchanged.
- [x] No event or message name renamed or removed; no payload field removed — `ai.action.confirmed` still fires with the same shape.
- [x] No database table or column changed — `AiPendingAction` is untouched and the PR ships no migration.
- [ ] **Where a contract had to change, the old surface kept working through a deprecation window, with migration notes** — it did not: `AI_PENDING_ACTION_ALLOWED_TRANSITIONS` loses `pending → confirmed` and `confirmed → executing` in one release, with no bridge and no `UPGRADE_NOTES.md` entry. See the Major above, including why the Emergency Security Exception does not cover it and why the narrowing is separable from the fix.

## 🧪 Test Coverage

What the four test files actually cover: the repository's claim contract (first call `claimed: true` and the row lands on `executing`, second call `claimed: false` with the row unchanged and `resolvedAt` still null, plus cross-tenant and cross-organization attempts both throwing), the executor's use of that contract (the winner runs the handler, the nineteen losers each get `confirmation_in_progress` and none reaches the handler, and a terminal status observed at claim time yields `invalid_status` with no `setStatus` write), the route's two `202` paths (a pre-read `executing` row short-circuits before the claim is even attempted, and a lost claim returns `202` without writing status), and the rewritten transition assertions in the contract suite. The idempotency short-circuit is now additionally pinned to call neither `claimForConfirmation` nor `setStatus`, which is a real tightening over the previous assertions.

The gaps, in the order I would close them:

1. `TC-AI-ACTIONS-PENDING-004-confirm-cancel-mutations.spec.ts` — seed `status: 'executing'`, `POST .../confirm`, assert `202` and `code: 'confirmation_in_progress'`. Place it beside the existing `409 invalid_status` case at `:129-136`; the `seed()` helper already accepts the status.
2. `packages/ui/src/ai/parts/__tests__/MutationPreviewCard.test.tsx` — a case where `confirmPendingAction` resolves `{ ok: true, data: { ok: false, code: 'confirmation_in_progress', mutationResult: null } }`, asserting the card does **not** render the handler-error alert. This is the regression test for the Minor above; without it the false error can come straight back.
3. The DB-level exclusion, as discussed under the coverage Minor — currently proven only by your out-of-band repro, which is worth recording somewhere durable even if a committed concurrency test stays out of scope.

## Checked and deliberately *not* filed

Recording these so the next reviewer does not re-derive them:

- **The removed `pending → confirmed` edge has no production caller.** Verified independently rather than taken from the PR body: the only `setStatus` call sites in shipped code are `workers/ai-pending-action-cleanup.ts:213` (`expired`), `lib/pending-action-cancel.ts:128`/`:158` (`expired`, `cancelled`) and `lib/pending-action-executor.ts:404`/`:441` (`failed`, `confirmed`, both from `executing`). Your claim holds exactly.
- **`nativeUpdate` bypassing the ORM's change tracking is safe on this entity today.** `AiPendingAction` (`data/entities.ts:122-211`) has no `updated_at` column and no `onUpdate` property, and no column on it is encryption-flagged, so the raw `UPDATE` skips no hook. Worth remembering only if `normalizedInput` ever becomes encrypted, which the repository's own doc comment at `:73-77` anticipates — at that point `failedRecords` written through `nativeUpdate` would need a second look.
- **The tenant and organization scoping of the claim is real, not decorative.** The `where` carries both (`:319-324`), the follow-up read repeats them (`:337-341`), and `AiPendingActionRepository.test.ts:398` drives both a cross-tenant and a cross-organization attempt to a throw.
- **`refresh: true` on the post-`UPDATE` read (`:342`) is load-bearing.** The route has already loaded the row through `getById` (`route.ts:150`), so without the refresh the identity-mapped entity would still read `pending` and the loser branch at `pending-action-executor.ts:376-393` would misclassify.
- **The two-statement claim is not a TOCTOU.** After a winning `UPDATE` only the winner can move the row, and a loser whose `SELECT` observes `confirmed` gets exactly the idempotent prior result at `pending-action-executor.ts:377-380`.
- **`rowMatchesWhere` in the repository test honours `status` (`:49`)**, so "permits exactly one transition from pending" is not green for the wrong reason.
- **`risk-medium` re-derived and kept.** I considered `risk-high` and rejected it with a reason: every edit lives inside one module, the entity and database are untouched, there is no migration, and the single cross-package consequence — the UI's handling of the `202` — is a finding here rather than a change this PR ships.

Housekeeping, not a code finding: the CLA bot's badge on this PR still reads *not signed*. That is a merge gate independent of review, and the badge can be stale — worth a re-check before this reaches the queue.

pkarw:
🏷️ **Label rationale**

| Label | Decision | Why |
|-------|----------|-----|
| `changes-requested` | `review` → `changes-requested` | One Major (the exported transition table is narrowed without the deprecation protocol, and the narrowing is separable from the security fix) plus three Minors. This repo's review extension requests changes on Minors too. |
| `security` | kept | Correct category — this closes a double-execution hole in the mutation-approval gate. |
| `needs-qa` | kept | Re-checked clause by clause against the automated-verification exemption: the API surface changes (a new `202` on `POST /api/ai/actions/:id/confirm`), which disqualifies it on its own. No `.tsx` is touched, but the exemption needs the API surface unchanged as well. Never combined with `skip-qa`. |
| `priority-high` | kept | Approval-integrity defect reachable without any LLM involvement, in a package published to npm. |
| `risk-medium` | kept | I considered `risk-high` and rejected it with a reason: every edit lives inside one module, the entity and database are untouched, there is no migration, and the single cross-package consequence — how `packages/ui` handles the new `202` — is a finding in the review rather than a change this PR ships. The exported-constant blast radius is real, which is what keeps this above `risk-low`. |

`qa-approved` deliberately not applied — that is earned by manual QA or the self-QA exception, never from reading a diff. The `qa` pipeline label is untouched; it means QA is in progress and belongs to a QA reviewer.

pkarw:
@pat-lewczuk handing this back to you — the review is submitted as changes-requested and the PR is reassigned to you.

**The fix itself is right and I did not go hunting for a reason to say otherwise.** The diagnosis is accurate down to the mechanism (`setStatus` reads without `FOR UPDATE`, evaluates the transition in JS, and the entity has no version column, so nothing backstops it), and moving the predicate into the `UPDATE` is the correct minimal answer. Collapsing the state machine so a row is never `confirmed` before its mutation runs is a real improvement, not just a side effect.

**What blocks it is one thing, and it is cheap.** The transition-table narrowing — dropping `pending → confirmed` and making `confirmed` terminal — is not required by the claim. `claimForConfirmation` goes through `em.nativeUpdate` and never reads the table, and I checked every production `setStatus` call site: the only writes are `expired`, `cancelled`, and `failed`/`confirmed` **from `executing`**, which the old table already allowed. So the pre-existing table works unchanged with your new flow, and keeping those two edges cannot reopen the hole. Because the narrowing is separable, `BACKWARD_COMPATIBILITY.md`'s Emergency Security Exception does not cover it (requirement 2: no unrelated tightening rides along) and the ordinary protocol applies. Either restore the two edges — two lines, plus adjusting the `scenario-14` assertions that pin `confirmed` as terminal — or ship the tightening with an `UPGRADE_NOTES.md` entry, a spec section, a dated `BACKWARD_COMPATIBILITY.md` entry, and a named maintainer sign-off. You asked for this call in the description; my recommendation is to drop the tightening, since it buys nothing the claim does not already guarantee.

**Fixing that alone flips this to approve.** The three Minors — the UI rendering your `202` as "The mutation handler reported an error", the missing integration case for the `202`, and `mutation-approvals.mdx:82` still documenting `pending → confirmed → executing` — are all fair to take as follow-ups if you would rather land the security fix first, though the UI one is a handful of lines and is a visible regression in exactly the double-click scenario this PR exists to fix.

The review body also records seven things I checked and deliberately did **not** file, including confirmation that your "no in-repo caller relies on the removed edge" claim is exactly right and that `nativeUpdate` skips no hook on this entity — so please do not spend time re-deriving those.

CI is fully green on `25c89e16` and I verified it at step level, not job level. No CI follow-up is pending from this run.

pkarw:
🤖 `om-auto-review-pr` completed: **CHANGES REQUESTED** (0 blockers, 1 major, 3 minors, 3 nits). Lock released. Validation source: GitHub checks — all gates green on `25c89e16`, verified at step level. autofix: skipped (not my PR — re-run with --autofix to fix it here).

pat-lewczuk:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-08-31T09:19:54Z.

pat-lewczuk:
🤖 `om-auto-fix-pr` — 🏷️ **Label rationale**

- 🔍 `review` — replaces `changes-requested`. Every finding from the code review has been addressed in `9655f25f5`, so the PR is no longer waiting on the author; it is waiting on a **re-review**. It is deliberately *not* `merge-queue`: that state means an approving review exists, and the current GitHub review decision is still `CHANGES_REQUESTED` from @pkarw. Only a re-review by someone other than the PR author can flip it, and this run cannot self-approve.
- 🔒 `security` — unchanged. The PR closes an approval-integrity defect where N concurrent confirmations of one approved action each invoked the mutation handler.
- 🧪 `needs-qa` — unchanged, and kept deliberately. The change alters what an operator sees in the mutation-approval card on a double-click (a misleading "the mutation handler reported an error" alert becomes an in-flight spinner), which is exactly the class of user-facing behaviour manual QA exists to exercise. With the QA gate on, this PR stays unmergeable until a QA reviewer adds `qa-approved`. This run does **not** add `qa-approved` and never will — that label is earned by manual QA or the documented self-QA exception, not inferred from a diff.
- ⬆️ `priority-high` — unchanged. Approval-integrity defect in a package published to npm; it belongs in the next release.
- ⚖️ `risk-medium` — unchanged, and re-derived rather than copied. Every edit stays inside `packages/ai-assistant` and `packages/ui`; the entity and database are untouched and the PR ships no migration. The one change that would have pushed this to `risk-high` — narrowing the exported `AI_PENDING_ACTION_ALLOWED_TRANSITIONS` — has been reverted, so no protected contract surface is narrowed any more.
- `in-progress` → `ci-monitoring` on release — the work and the reporting are complete; the label is swapped rather than dropped because a CI-result follow-up comment is still owed on this head.

pat-lewczuk:

## 🎯 What this run did

Drove this PR toward merge-ready: merged the latest base in, ran the review-autofix loop against every finding already on the PR, verified the result with the full local gate, filed the deliberately-deferred items as tracked follow-ups, and normalized the labels. It did **not** merge, and it did **not** approve — both are deliberate and explained below.

## 🔀 Base merge

`origin/develop` (`421cefe66`) merged cleanly into `sec/om-ai-sec-002` with no conflicts, pushed as `2fade73cf`, **before** any review or CI work. The branch was 69 commits behind, so this matters: without it the review and the gate would have been judging a diff that is not the diff that merges. As a bonus, the `agent-registry.test.ts` idempotency failure the PR description called out as pre-existing is gone — it was fixed on `develop` and arrived with the merge.

## 🔍 Reviewer comments — how each one ended up

All findings come from [@pkarw's `changes-requested` review](https://github.com/open-mercato/open-mercato/pull/5688#pullrequestreview-5035955956). Every one is accounted for; nothing was silently dropped.

**⚠️ Major — the exported transition table was narrowed without the deprecation protocol. → Fixed, by dropping the tightening (the reviewer's option 1).**

`AI_PENDING_ACTION_ALLOWED_TRANSITIONS` is re-exported from the package root, so removing `pending → confirmed` and making `confirmed` terminal narrowed a public surface of a published package with no bridge, no `@deprecated` marker, and no `UPGRADE_NOTES.md` entry. The review's decisive observation is the one that settled it: **neither removal was load-bearing for the security fix.** `claimForConfirmation` writes through `em.nativeUpdate` and never consults the table, and the confirm route no longer reaches `setStatus` for those edges at all — so restoring them cannot reopen the concurrency hole. Both edges are back; `pending → executing` stays as a purely additive edge. A new `scenario-14b` case pins the table as a superset of the pre-claim contract so the narrowing cannot return by accident, and `scenario-14`'s two assertions that pinned `confirmed` as terminal were rewritten rather than deleted — they now assert `confirmed → cancelled` still throws, which was true before this PR and remains true.

**🔹 Minor — the product's own client rendered the new `202` as "The mutation handler reported an error". → Fixed, with a regression test that actually fails without the fix.**

`202` is a 2xx, so `apiCall` reports `ok: true` and `MutationPreviewCard` landed on its HTTP-200 branch, where the body's `ok: false` plus a null `mutationResult` collapsed into the generic `execution_failed` mapping. In the exact scenario this PR exists to make safe — a double-click — the operator was told the mutation failed while the winning request was succeeding. The card now short-circuits on `code: 'confirmation_in_progress'` *before* the handler-error mapping and stays on the in-flight spinner while polling resolves the row.

Worth calling out, because it is the kind of thing that quietly rots: the first version of this regression test asserted on rendered DOM and **passed without the fix**. `ConfirmationCard` mounts before the confirm promise resolves, so a `confirmError` arriving afterwards never reaches its alert on its own — the DOM cannot discriminate. The test was rewritten to assert on the props handed to `ConfirmationCard`, verified to fail when the guard is removed and pass with it, and shipped beside a contrast case proving the guard is not blanket (a genuine handler failure still raises `execution_failed`).

**🔹 Minor — the new `202` had no integration coverage. → Fixed.**

`TC-AI-ACTIONS-PENDING-004` now seeds a `status: 'executing'` row, POSTs confirm, and asserts `202` with `code: 'confirmation_in_progress'`, a null `mutationResult`, and the row still `executing` — placed beside the existing `409 invalid_status` case, exactly where the review suggested. The spec's header contract notes were updated to match.

**🔹 Minor — the public documentation stated the opposite of what the code does. → Fixed.**

`apps/docs/docs/framework/ai-assistant/mutation-approvals.mdx` said *"Atomic status transition `pending → confirmed → executing`"* — the sentence this PR falsifies — and never mentioned the `202`. Since `packages/ai-assistant/AGENTS.md` names these public docs authoritative *over* the AGENTS.md, that was the authoritative description of the lifecycle being wrong. The `executePendingActionConfirm` step list now describes the compare-and-swap claim and why a transaction is not a mutex, a new response-contract table documents `200` / `202` / `409` / `412`, the HTTP-routes table mentions the `202`, and the state-machine note explains why the table retains the legacy edges the claim path does not use. The duplicate bullet in `pending-action-executor.ts`'s doc comment was collapsed and now covers the `executing` case it was missing.

**💅 Nit — `const now = extra?.now ?? new Date()` computed and never used. → Fixed by dropping the parameter.**

`executing` is not a resolved state, so there is genuinely nothing for the claim to stamp; `resolvedAt` is written on the terminal `confirmed` / `failed` transition. Rather than leave the API advertising a clock the implementation ignores, `now` was removed from `claimForConfirmation`'s `Pick<…>` signature and from its one call site, with a doc comment recording why.

**💅 Nit — a claim lost to a `cancelled`/`expired` row answered `200 ok:false`, not the `409` the same block declares. → Fixed.**

When the row goes terminal between the recheck and the claim, the route now returns the `409 invalid_status` its own OpenAPI `errors` array promises — the same answer a row that was already terminal on arrival gets, so the declared response surface holds however the race lands. A route test drives exactly that race.

**💅 Nit — the known-gap disclosure understates one consequence. → Deferred to a follow-up, as the review itself suggested.** See below.

## 🔁 Follow-ups filed

- **#5807 — [Reap pending actions wedged in `executing`, restoring cancel as well as confirm](https://github.com/open-mercato/open-mercato/issues/5807)** (`bug`, `security`, `priority-medium`, `risk-medium`). This is the reviewer's third nit, filed with the consequence they identified and this PR's description understated: a row wedged in `executing` is not merely unconfirmable. `executing` has no `cancelled` edge, so cancel throws the state-machine error too, and the cleanup worker only scans `status = 'pending'`, so the TTL never reaches the row — the operator has **no escape from the UI at all**. Deferred rather than fixed here because building a reaper is its own change with its own state-machine and configuration decisions, and the gap is pre-existing rather than introduced by this PR.
- **#5808 — [Pin the pending-action claim's DB-level exclusion with a test that can actually fail](https://github.com/open-mercato/open-mercato/issues/5808)** (`security`, `priority-low`, `risk-low`). The review's sharpest point about coverage: the repository tests mock `nativeUpdate` as a JavaScript compare-and-swap, which single-threaded JS makes atomic regardless, so a read-then-check-then-write reimplementation would pass them identically. The contract is pinned; the security property the PR is named for is not. Deferred because a real concurrent-confirm test cannot reuse the existing spec — driving a live `pending → confirmed` would execute a real tool mutation, which that spec deliberately avoids — so it is genuinely separate work rather than something to bolt onto a security fix that already ships green.

Neither follow-up defers a blocker. Everything correctness-, security-, compatibility-, or gate-related was fixed in this PR.

## 🧪 Validation

Full `validation.commands` gate, run locally in order (`Runner: local`), all green on `9655f25f5`:

| Command | Result |
|---------|--------|
| `yarn generate` | ✅ pass |
| `yarn build:packages` (rebuild) | ✅ pass |
| `yarn i18n:check-sync` | ✅ pass — all 5 locales in sync |
| `yarn i18n:check-usage` | ✅ pass — advisory unused-key report only |
| `yarn typecheck` | ✅ pass — 27/27 tasks |
| `yarn test` (`ai-assistant`) | ✅ **1487 passed**, 109 suites, 0 failures |
| `yarn test` (`ui`) | ✅ **1937 passed**, 235 suites, 0 failures |
| `yarn lint` | ✅ pass — 0 errors (warnings only, all pre-existing) |
| `yarn build:app` | ✅ pass |

No test was weakened, skipped, or deleted to get here, and no CI check was disabled. The one assertion that changed (`scenario-14`) changed because the behaviour it pinned was itself the Major finding; it was replaced with a stricter pair, not dropped.

**CI is still running on this head.** Branch protection plus the QA-approval gate hold the actual merge; this run's verdict covers the work, not a green run. Checks still pending at the time of writing: `prepare` (and the `test` / `build` / `ephemeral-integration` jobs gated behind it), `Analyze (javascript-typescript)`, and `license/cla`. A follow-up comment will report the CI outcome.

## ⚠️ What still blocks this PR — two things, both needing a human

1. **A re-review.** Every finding is addressed, but the GitHub review decision is still `CHANGES_REQUESTED` from @pkarw, and GitHub does not permit the PR author to approve their own PR. This run therefore could not submit an approving review, and did not fake one by moving the PR to `merge-queue`. @pkarw (or another reviewer) needs to re-review — per their own review, *"fixing the Major alone is a two-line change and flips this to approve"*, and the Minors are fixed too.
2. **The CLA.** `license/cla` reads *not signed*. The two substantive commits (`e7fb4c5e`, `25c89e16`) are authored by **@Szesnasty** (Łukasz Jakubowski), who reported and fixed this under private responsible disclosure — authorship was preserved deliberately so the credit is real. That means the CLA gate is his to clear, and it is independent of code review. It was flagged in the review as worth re-checking before this reaches the queue; it is still pending.

Additionally, `needs-qa` is on and the QA gate is active, so a QA reviewer must exercise the confirm flow and add `qa-approved` before this can merge. This run posts no QA verdict and adds no QA label.

## 🚀 Hand-off

`om-auto-fix-pr` never merges. Once the re-review lands, the CLA clears, and QA signs off, the merge belongs to `om-approve-merge-pr 5688` / `om-merge-buddy`, which re-check every gate — including genuinely green required checks — before squash-merging.

pat-lewczuk:
🤖 `om-auto-fix-pr` completed: **all review findings addressed**, full local gate green on `9655f25f5`, 2 follow-ups filed (#5807, #5808). Lock released — `in-progress` swapped for `ci-monitoring` because a CI-result follow-up comment is still owed on this head. The PR is free for anyone to act on. It is **not** merge-ready yet: it needs a re-review (the author cannot self-approve), the CLA signature from @Szesnasty, and QA sign-off while `needs-qa` is on.

pat-lewczuk:

## ✅ Every workflow check passed on `9655f25f5`

The full CI surface went green on the post-review head, including the base merge of `develop`. Nothing here needed a rerun, a retry, or a flake write-off — it passed first time.

| Check | Result |
|-------|--------|
| `prepare` | ✅ success |
| `test` | ✅ success |
| `lint` | ✅ success |
| `ds-lint` | ✅ success |
| `docker-build` | ✅ success |
| `ephemeral-integration` (1/15 … 15/15) | ✅ success — **all 15 shards** |
| `CodeQL` + `Analyze (javascript-typescript / python / actions)` | ✅ success |
| `audit-scope`, `scope` | ✅ success |
| `audit`, `comment`, `mutate`, `merge-coverage`, `[code]smith` | ⏭️ skipped by design |

Worth noting that `ephemeral-integration` ran as the **full 15-shard matrix** on this head, not the affected-only single-runner mode the previous revision got — so `TC-AI-ACTIONS-PENDING-004`, including the new `202 confirmation_in_progress` case added in this round, executed against real Postgres alongside the rest of the suite.

One caveat carried forward from the review, so nobody over-reads the green: `mutate` reports `skipped`, not `success`. `scripts/stryker/scope.mjs` freezes its package allowlist to `['shared']`, so no `packages/ai-assistant` diff is ever mutation-tested. That check carries no signal on this PR either way.

This matches the local gate exactly — `build:packages`, `generate`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `lint`, `build:app`, plus `ai-assistant` 1487/1487 and `ui` 1937/1937.

## ⏳ `license/cla` — still pending, and a wait will not clear it

`license/cla` is the one check that is not green, and it is **not a CI failure**. It is the CLA assistant waiting on a signature, and it will sit at `pending` indefinitely until a human acts.

The two substantive commits (`e7fb4c5e`, `25c89e16`) are authored by **@Szesnasty** (Łukasz Jakubowski), who reported and fixed this under private responsible disclosure — authorship was preserved deliberately so the credit is real rather than cosmetic. That makes the CLA his to sign, via the link on the CLA assistant's comment above. It gates the merge independently of code review.

Because that check cannot settle on its own, the CI monitoring budget was allowed to expire rather than spent watching a signature that no automated wait can produce. **No further follow-up will come from this agent**; `ci-monitoring` is removed accordingly, so the label does not promise a comment that is never coming.

## 🚦 Where this leaves the PR

CI is green and the verdict from the run summary stands unchanged: all review findings are addressed, and the code is ready. What remains is entirely human-owned:

1. **A re-review** — the decision is still `CHANGES_REQUESTED` from @pkarw, and GitHub does not let the author approve their own PR.
2. **The CLA signature** from @Szesnasty (above).
3. **QA sign-off** — `needs-qa` is on with the QA gate active, so `qa-approved` from a QA reviewer is required. This run adds no QA label.

Once all three clear, `om-approve-merge-pr 5688` owns the merge and will re-check every gate before squash-merging.

PR #5690, opened 2026-08-27
fix(shared): stop counting scope initialisation as an organization change

Geeknauts-company:
## Summary

`emitOrganizationScopeChanged` keeps the current scope in module state, initialised to `{ organizationId: null, tenantId: null }`. But that is a real scope value ("all organizations"), not "not known yet" — and nothing downstream can tell the two apart.

So the first announcement of every page load, which comes from the organization switcher reporting the scope it has just read back from `om_selected_org`, looks like `null -> real id` and bumps `currentVersion` 0 -> 1. That is initialisation being counted as an organization change.

`currentVersion` is the cache-busting key behind `useOrganizationScopeVersion()`, read in ~76 places across dashboard, CRM, sales, settings, entities and translations, alongside ~16 `useOrganizationScopeDetail()` readers. A spurious bump re-keys all of them onto a key the query cache has no data for, so each visit refetches scope-dependent data that had just been fetched for exactly that scope. Screens that gate their render on the resulting `isLoading` additionally blank their body while it runs.

The fix seeds the module scope lazily from the same `om_selected_org` / `om_selected_tenant` cookies the server resolved the document from, so an announcement that merely repeats them is initialisation rather than a change. The seed is browser-only and never memoized on the server, where the module is shared across requests.

## Changes

- `packages/shared/src/lib/frontend/organizationEvents.ts` — seed the module scope from the selection cookies **at module initialisation**, and compare an incoming announcement against that.
- `packages/shared/src/lib/frontend/__tests__/organizationEvents.test.ts` — 11 cases covering the fix, the seed's timing, and the transitions that must keep bumping.
- `packages/shared/src/lib/frontend/__tests__/organizationEvents.server.test.ts` — the server path, in jest's `node` environment (this package's default, so no docblock pragma; a pragma is per-file, which is why it cannot live in the jsdom suite).
- `packages/core/src/modules/api_keys/backend/api-keys/create/page.tsx` — stop reading `scopeVersion === 0` as "nothing announced yet"; key on the announced scope instead.
- `packages/core/src/modules/api_keys/backend/api-keys/create/__tests__/CreateApiKeyPage.test.tsx` — 3 cases for that page.

## Review round 2 (@adeptofvoltron)

Thank you — the Major was real and I reproduced it before changing anything. All three findings are addressed in `72e0644b`, pushed as a **new commit on top of the reviewed head** rather than an amend, so the delta is readable on its own.

- **Major, seed timing.** The seed is no longer lazy and the `scopeSeeded` latch is gone; `seedScopeFromCookies()` runs once at module initialisation, keeping the `typeof document === 'undefined'` guard, and the lazy calls are dropped from both `getCurrentOrganizationScope()` and `emitOrganizationScopeChanged()`. A single unconditional call needs no idempotence. Three cases pin it, under *"a cookie rewritten just before the announcement"*: the hazard you described (nothing read the scope first), its baseline (a component did), and the `persistTenant()` variant where `load()` writes `om_selected_tenant` on the way to the emission. **Against the lazy source the first and third fail**; the second passed under it, which is exactly the complaint — the same page produced different versions depending on mount order.
- **Minor, the `scopeVersion === 0` consumer.** Fixed rather than documented, taking your first suggestion: the page reads `useOrganizationScopeDetail()` and `useOrganizationScopeVersion()` is dropped from it entirely. It was the only such consumer in the monorepo. Three cases cover it, one of which fails against the old guard. The other two are baselines and say so in their comments — in candour, the page's *observable* behaviour barely moves, because `loadInitialScope()` resolves the same tenant a beat later, exactly as you noted; the change is to stop depending on a meaning the version no longer carries.
- **Minor, the changed public return value.** Now stated below, and `organizationEvents.server.test.ts` pins the SSR side of the pair.
- **Nit, the comment block.** Compressed while the seed-timing rationale was being added to it, so it did not grow on net. The other two nits (the third cookie parser and its blank-value semantics; the constants hoist) both belong to the follow-up PR — see below.

## Behavioural contract changes, for downstream module authors

No `BACKWARD_COMPATIBILITY.md` surface is broken — no signature, type, route, event name or payload changes. Two *behavioural* contracts do change, and both are called out here rather than left to be discovered:

- **`getCurrentOrganizationScope()` returns a different value before the first announcement.** It used to return `{ organizationId: null, tenantId: null }` until the switcher announced; it now returns the cookie scope on the client, while still returning `{ null, null }` on the server. `useOrganizationScopeDetail()` uses `React.useState(getCurrentOrganizationScope)`, so its lazy initializer now differs between the SSR pass and hydration for the ~16 components that read it. I found no consumer that renders differently on that value at first paint — the ones I checked feed it into effects, query keys and request params — and `organizationEvents.server.test.ts` pins the server side (no `document` means no seed; no `window` means `emitOrganizationScopeChanged` mutates nothing, so one request's organization cannot become the shared module's idea of the scope). That invariant matters more with an eager seed than a lazy one: a lazy seed could only fire from a call site, an eager one fires on import.
- **`getCurrentOrganizationScopeVersion() === 0` no longer means "nothing announced yet".** On a repeat visit where the announcement repeats the cookies, the version stays 0 for the life of the page. The one call site reading it that way is fixed in this PR.

## What still bumps, deliberately

These are the cases the tests pin, because each of them is load-bearing:

- **A real organization switch still bumps.** The switcher calls `router.refresh()` rather than navigating, so the client tree stays mounted; the re-key is the only thing stopping the previous organization's rows being read under the new one's name. This is also why `placeholderData: keepPreviousData` would be the wrong fix for the blanking symptom — across a real switch it would render the previous organization's data for a few hundred ms, which is the defect class this guard exists to prevent.
- **An absent or unreadable cookie still bumps.** Without the cookie the server resolved the first render from the JWT's organization, which the client cannot see, so a first announcement naming a different one must invalidate. A browser's first-ever visit therefore still bumps once, and that is correct.
- **The `__all__` sentinel is a scope, not an unknown.** It seeds `organizationId: null`, and an announcement of `null` after it is not a change.

## Relationship to the existing `hasEmitted` / `isFirstEmit` guard

They are orthogonal and compose correctly. That guard governs event *dispatch*; this one governs the *version*, and the version was unaffected by it — `currentVersion` still went 0 -> 1 on the `null -> real id` transition on current `develop`. After this change, a first announcement that repeats the cookie still dispatches, so `useOrganizationScopeDetail()` subscribers sync as before, while `useOrganizationScopeVersion()` correctly holds. There is a test pinning exactly that.

## Follow-up worth a maintainer's opinion: the inlined cookie constants

`shared` must not depend on `core`, so this change inlines `om_selected_org`, `om_selected_tenant` and `ALL_ORGANIZATIONS_COOKIE_VALUE = '__all__'` rather than importing them from `core`'s `modules/directory/constants`.

That is not a new compromise — `shared` already does it twice today: all three values in `lib/auth/server.ts`, and `SELECTED_ORG_COOKIE` again in `lib/crud/factory.ts`. This makes a third copy, which is the argument for a tidier resolution: hoist the constants into `shared` (e.g. `lib/scope/cookies.ts`) and have `core/modules/directory/constants` re-export them for backward compatibility.

That is mechanical, but it touches more files than this fix needs, so I raised it in a code comment rather than bundling it. **Answered in review: a separate PR plus an issue**, since it touches a surface listed in `BACKWARD_COMPATIBILITY.md`. Will do — that follow-up is also where the third cookie parser and its blank-value divergence from `parseSelectedOrganizationCookie` belong, since collapsing the parsers is the point of the hoist.

## Testing

Run from a clean clone of `develop` with the branch applied:

- `packages/shared` frontend suites: **14/14 pass**. `packages/core` `api_keys`: **39/39 pass**.
- **Not vacuous, measured both ways.** Against the unfixed source the original suite fails 5 of 8, and the three new ordering cases fail 2 of 3 (the third is the deliberate baseline). Against the pre-fix `api_keys` page, 1 of the 3 new page cases fails.
- Whole `packages/shared` suite: **1982 passed**, 5 skipped, 2 failed, 11 suites unrunnable — against **1976 / 5 / 2 / 11** measured on the reviewed head immediately before, i.e. this commit adds 6 passing tests and changes nothing else. The 2 failures are `src/lib/auth/__tests__/principalServiceExport.test.ts`, which imports from `packages/shared/dist/`; the 11 unrunnable suites cannot resolve `@open-mercato/cache` for the same reason — a clone that has never been built. Identical numbers through `turbo run test --filter=@open-mercato/shared`.
- `tsc --noEmit`: `packages/shared` clean. `packages/core` reports 265 errors, every one `Cannot find module '#generated/...'` in an unbuilt clone, none of them in the touched files.
- `eslint` on all five files: **0 errors**. Two warnings, both `Unused eslint-disable directive (@typescript-eslint/no-require-imports)`, which reproduces identically on pre-existing files in the same package (e.g. `packages/shared/src/lib/umes/__tests__/extension-headers.test.ts`), so the new test files just match the existing convention.

## Specification

- [ ] Yes
- [ ] No (created a new spec)

**Spec file path:**
n/a — `CONTRIBUTING.md` scopes spec-driven development to "new features or significant changes"; this is a bug fix confined to one ~60-line module with no API, schema or UI surface change. Glad to write one if a maintainer would rather have it.

## Checklist

- [x] I updated documentation, locales, or generators if the change requires it. — none required; no user-facing copy, no generated artefacts.
- [ ] I added or updated integration tests in `.ai/qa/tests/` — not required: the change has no manually exercisable UI surface of its own, leaves the database structure and API surface unchanged, and ships unit coverage for the behaviour it changes.
- [ ] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — N/A, see above.
- [ ] Priority set — I don't have label permissions on this repo; please apply.
- [ ] Risk set — suggest `risk-medium`: the module is small and fully unit-covered, but ~92 call sites read the value it produces, and the second round touches one of them.
- [ ] QA routing set — suggest `skip-qa`: no manually exercisable UI surface, no database or API surface change, `BACKWARD_COMPATIBILITY.md` contracts preserved, and automated tests ship with the behaviour change. Please re-route to `needs-qa` if you read the blast radius differently.

### Design System Compliance

N/A — no UI, no components, no styling. The change is confined to a browser-side event module and its tests.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5690"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790388022&installation_model_id=432243&pr_number=5690&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5690&signature=090ec51a3a23ad796d180f38220c4c97ea0421b02b091cb9c6f769a536b49004"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-08-27T10:39:26Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron (dismissed):
# 🔍 Code Review: fix(shared): stop counting scope initialisation as an organization change

## 🎯 Summary

This PR makes `packages/shared/src/lib/frontend/organizationEvents.ts` seed its module-level `currentScope` from the `om_selected_org` / `om_selected_tenant` cookies before it compares an incoming announcement against it, so the organization switcher reporting back the scope it just read from the cookie no longer looks like a `null -> real id` transition and no longer bumps `currentVersion`. The diagnosis is correct and unusually well argued: `currentVersion` really is the cache-busting key that `useOrganizationScopeVersion()` feeds into query keys across the app, `{ organizationId: null, tenantId: null }` really is an ambiguous starting value that conflates "all organizations" with "not known yet", and the consequence — components mount and fetch at version 0, the switcher's async `load()` resolves, the version flips to 1, and every scope-keyed query re-keys onto a cold key and refetches while screens gating on `isLoading` blank out — is a real per-page-load cost. I also appreciate that the PR is explicit about the three transitions that must keep bumping and pins each of them with a test, and that it argues down the tempting-but-wrong `keepPreviousData` alternative rather than ignoring it. Reviewed scope: the two changed files, plus the consumers that make the change load-bearing — `packages/shared/src/lib/frontend/useOrganizationScope.ts`, `apps/mercato/src/components/OrganizationSwitcher.tsx` and its byte-identical template twin, `packages/core/src/modules/directory/utils/organizationScope.ts`, and the `getCurrentOrganizationScope()` / `useOrganizationScopeDetail()` / `useOrganizationScopeVersion()` call sites.

The reason I am requesting changes is not the idea, which I think is right, but the *timing* of the seed. Reading the cookies lazily, on the first call into the module, means the snapshot can be taken **after** the organization switcher has already rewritten those same cookies — and in that case the fix silently swallows a version bump that the PR itself argues must happen. I reproduced this locally; details below.

## Verdict

❌ **request changes** — one Major finding: the lazy seed is order-dependent and, on the "cookie is stale, server resolved a different organization" path, reads the cookie `OrganizationSwitcher.persistSelection` just wrote, so a genuine scope change does not bump `currentVersion` and the previous organization's rows stay keyed and on screen. Two Minor findings follow it. Per this repository's verdict rule, Medium/minor findings also drive `changes_requested`.

## 🧪 Validation Gate

Validation source: **GitHub checks** (per the repo-local `om-auto-review-pr` extension, CI check results replace re-running the gate locally). No required check is failing on head `a80b84c`; several are still running.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ⏳ PENDING | Covered by the `prepare` job, which passed (3m51s); the downstream `test` job that consumes it is still running. |
| `yarn generate` | ✅ PASS | Covered by `prepare` (passed). No generated-file drift reported. |
| `yarn build:packages` (rebuild after generate) | ✅ PASS | Covered by `prepare` (passed). |
| `yarn i18n:check-sync` | ✅ PASS | Covered by `scope` / `audit-scope` (both passed); the diff adds no user-facing strings and no locale keys. |
| `yarn i18n:check-usage` | ✅ PASS | Same as above. |
| `yarn typecheck` | ⏳ PENDING | Part of the `test` job, still running. |
| `yarn test` | ⏳ PENDING | The `test` check and all 15 `ephemeral-integration` shards are still running. |
| `yarn build:app` | ⏳ PENDING | Downstream of `test`. |

Other checks on this head: `lint` ✅, `ds-lint` ✅, `scope` ✅, `audit-scope` ✅, `documents-multi-instance` ✅, `mutate (shared, src/lib/frontend/organizationEvents.ts)` ✅, `prepare` ✅. `audit`, `docker-build`, `comment` and `[code]smith` are skipped by design for a fork PR, and the preview deployment was blocked by the fork security protection — that is expected and not something you need to act on.

**CI is still pending on this PR.** The review below is submitted now rather than waiting: `test`, `build:app` and the 15 `ephemeral-integration` shards have not reported yet, so this review does not certify them. Branch protection and the merge queue remain the authority on whether they pass.

⚠️ **`license/cla` is pending and reports "Contributor License Agreement is not signed yet."** The PR checklist ticks the CLA box, but cla-assistant has not recorded a signature for this account, and that check gates the merge regardless of the code review. Please sign via the cla-assistant link on the PR (the comment notes it is left for the account owner to tick personally, so this may just need the owner to complete it).

As supplementary review evidence — not as a substitute for the pending CI gate — I ran the PR's own new suite in an isolated worktree at the PR head: `packages/shared` `organizationEvents.test.ts` reports 8 passed, 8 total.

## Findings

### ⚠️ Major

**`packages/shared/src/lib/frontend/organizationEvents.ts:97-109` and `:128` — the seed is lazy, so it can read a cookie the switcher has already overwritten, and the version bump the guard exists to produce is lost.**

`seedScopeFromCookies()` runs on the first call into `getCurrentOrganizationScope()` or `emitOrganizationScopeChanged()`, whichever happens first, and then latches via `scopeSeeded`. But `OrganizationSwitcher.persistSelection()` — identical in `apps/mercato/src/components/OrganizationSwitcher.tsx` and `packages/create-app/template/src/components/OrganizationSwitcher.tsx` — writes the cookies *before* it announces:

```
document.cookie = `om_selected_org=${encodeURIComponent(cookieValue)}; ...`   // line 181
persistTenant(tenantId ?? null, { refresh: false })                            // line 184, writes om_selected_tenant
emitOrganizationScopeChanged({ organizationId: resolved || null, ... })        // line 186
```

So whenever that emission is the **first** thing to touch the module, the seed reads the cookie the switcher wrote a microsecond earlier, `currentScope` comes out equal to `detail`, `hasChanged` is `false`, and `currentVersion` stays at 0.

That path is reachable and is exactly the one the version guard is for. In `load()` at `OrganizationSwitcher.tsx:282-288`, when the persisted cookie does not match the organization the server resolved for this session — a selected org the user has lost access to, or one the resolver dropped as non-existent, the case `packages/core/src/modules/directory/utils/organizationScope.ts:334-340` handles and flags as `selectionRejected` — the switcher calls `persistSelection(...)` to correct the cookie and announce the corrected scope. Every scope-keyed query that already fired on this page went out under the *stale* cookie. Without the bump those queries never re-key, so the previous organization's rows stay on screen under the new organization's name. That is the defect class the PR description correctly says this guard must prevent.

Whether the bug fires depends on nothing more than whether some unrelated component happened to read the scope earlier in the same tick — `TopbarSearchInline`, `HybridSearchTable` and `GlobalSearchDialog` call `getCurrentOrganizationScope()` and would seed it; `BackendChromeProvider` calls only `getCurrentOrganizationScopeVersion()`, which does **not** seed. Correctness should not rest on that.

I reproduced it as a jest probe against this branch (two cases, identical emission, only the preceding read differs):

```
BASELINE (a component read the scope first)  -> version 1   ✅
HAZARD   (nothing read the scope first)      -> version 0   ❌ expected 1
```

**Fix:** take the cookie snapshot at module initialization instead of lazily — a top-level `seedScopeFromCookies()` call, keeping the `typeof document === 'undefined'` guard so the server-side module state stays at `{ null, null }`. Module evaluation in the browser happens during hydration, and every writer of these cookies (`OrganizationSwitcher`) imports this module, so its evaluation is strictly ordered before any component can rewrite them. That makes the snapshot genuinely "what the server resolved this document from", which is the property the whole design depends on, and it lets you drop the lazy call from `getCurrentOrganizationScope()` and `emitOrganizationScopeChanged()`. Please also add a regression test that pins it: seed cookies to org A, rewrite the org cookie to B *without* any prior read, emit `{ organizationId: B }`, and assert the version is 1.

The same ordering applies to a second, subtler instance worth covering while you are there: on a session where the org cookie exists but the tenant cookie does not, `load()` calls `persistTenant(resolvedTenantId, { refresh: false })` at `OrganizationSwitcher.tsx:267` before the emission at line 290, so a lazy seed reads a tenant cookie the client just wrote. Seeding at module init fixes both with one change.

### 🔹 Minor

**`packages/core/src/modules/api_keys/backend/api-keys/create/page.tsx:60` — this change redefines what `currentVersion === 0` means, and there is a live call site still reading the old meaning.**

Before this PR, version 0 reliably meant "the switcher has not announced anything yet", because the first announcement always bumped. That page uses it exactly that way:

```tsx
React.useEffect(() => {
  if (scopeVersion === 0) return
  const rawTenant = scopeDetail.tenantId
  ...
  setSelectedTenantId(...)
}, [scopeDetail.tenantId, scopeVersion])
```

After this PR, on any repeat visit where both cookies are present and the announcement repeats them, the version stays 0 for the life of the page, so that effect never runs and `selectedTenantId` is never synced from the announced scope. Today the visible impact is contained because a sibling effect resolves the same tenant from `/api/directory/organization-switcher`, so the two paths agree in the happy case — but when that call fails or returns non-ok, the page now has no fallback where it previously had one, and the super-admin role picker (`loadRoleOptions`) returns `[]` for a super admin with no resolved tenant. Since the PR changes the contract, it should also update the one consumer that depends on the old one: read `useOrganizationScopeDetail()` directly (it is announced on the first emission regardless of the version, which the PR's own test pins), or expose an explicit "has been announced" signal rather than overloading version 0. Please add coverage for whichever you pick.

**`packages/shared/src/lib/frontend/organizationEvents.ts:111-117` — `getCurrentOrganizationScope()` changing its pre-announcement return value is a behavior change to a public export, and it is neither called out nor covered for the server/hydration case.**

The function is exported from `@open-mercato/shared` and previously returned `{ organizationId: null, tenantId: null }` until the first announcement; it now returns the cookie scope on the client while still returning `{ null, null }` on the server. The signature and type are unchanged so this is not a `BACKWARD_COMPATIBILITY.md` break, but it does change what first-render client code sees relative to the server-rendered HTML — `useOrganizationScopeDetail()` uses `React.useState(getCurrentOrganizationScope)`, so its lazy initializer now differs between the SSR pass and hydration for the ~16 components that read it. I did not find a consumer that renders differently on that value at first paint (the ones I checked feed it into effects, query keys and request params), so I am filing this as Minor rather than Major — but it is a deliberate widening of the change's blast radius that the PR body does not mention, and it deserves a line in the description plus a test asserting the server-side (`typeof document === 'undefined'`) path still returns `{ null, null }` and never latches `scopeSeeded`. The code comment already claims that invariant; nothing enforces it.

### 💅 Nit

**`packages/shared/src/lib/frontend/organizationEvents.ts:16-66` — a 51-line comment block ahead of three constants is out of proportion with this repository's conventions.**

Root `AGENTS.md` § Code Quality asks for self-documenting code without inline commentary, and the same rationale is already in the PR description where reviewers will read it. Consider trimming this to a few lines stating the invariant ("`{ null, null }` is a real scope, so the initial state must come from the cookies the server resolved this document from") and pointing at the PR or a short spec note for the rest. The `MAINTAINER NOTE` about the duplicated constants is worth keeping in some form since it is a standing invitation to a follow-up.

**`packages/shared/src/lib/frontend/organizationEvents.ts:72-87` — `readScopeCookie` is a third parser for the same two cookies, with slightly different blank-value semantics than the other two.**

`parseSelectedOrganizationCookie` (`packages/core/src/modules/directory/utils/scopeCookies.ts:11`) maps a blank value to `null` ("no selection"), while `readScopeCookie` returns `''` and the seed then maps it to `organizationId: null` ("all organizations"). Those two readings diverge in the resolver — "no selection" falls back to the account org, "all organizations" does not. I could not construct a case where the divergence produces a wrong result today, because the switcher writes the `__all__` sentinel rather than a blank org cookie and the resolver's own `resolveOrganizationOverride` treats blank as all-orgs too, so the seed and the server agree in practice. Flagging it anyway because drift between three near-identical copies is precisely what your `MAINTAINER NOTE` warns about.

**On your open question about hoisting the constants:** I would take the follow-up, not this PR. Moving `om_selected_org` / `om_selected_tenant` / `__all__` into `shared` (e.g. `lib/scope/cookies.ts`) with a re-export from `core/modules/directory/constants` is the right end state and would let the three parsers above collapse into one, but it touches a contract surface listed in `BACKWARD_COMPATIBILITY.md` (import paths / public exports) and deserves its own diff and its own review rather than riding along with a behavior fix. Please open an issue for it so it does not get lost.

## 💥 Breaking Changes

- [x] No function signature changed in a breaking way (required params removed or reordered, return type changed)
- [x] No event or message name renamed or removed; no payload field removed — `ORGANIZATION_SCOPE_CHANGED_EVENT` and its detail shape are untouched
- [x] Where a contract had to change: old surface kept working through a deprecation window, with migration notes

No protected contract surface is broken. Two *behavioral* contracts do change without being documented as such, and both are covered as findings above: the pre-announcement return value of `getCurrentOrganizationScope()`, and the meaning of `getCurrentOrganizationScopeVersion() === 0`. Neither needs a deprecation bridge, but both should be stated in the PR description so downstream module authors reading these exports know what changed.

## 🧪 Test Coverage

The new suite (`packages/shared/src/lib/frontend/__tests__/organizationEvents.test.ts`, 8 cases) is genuinely good work. `loadWithCookies()` builds a fresh module registry per case via `jest.isolateModules`, which is a faithful model of a page load rather than a convenience, and the suite covers both halves of the invariant: that repeating the cookie does not bump, and that a real organization switch, a tenant change under the same organization, and a missing organization or tenant cookie all still do. The case pinning that a first announcement still *dispatches* while the version holds is the one I would most have wanted to see, and it is there. I ran the suite at the PR head and it reports 8 passed, 8 total; I did not independently re-verify the claim that 5 of 8 fail against unfixed source, but the assertions are specific enough that the claim is credible.

The gaps, all of which map to findings above:

1. **Seed ordering (Major).** Add a case to `organizationEvents.test.ts`: set `om_selected_org` to org A and `om_selected_tenant` to a tenant, load the module, rewrite `om_selected_org` to org B **without** calling `getCurrentOrganizationScope()` first, emit `{ organizationId: B, tenantId: <tenant> }`, and assert the version is 1. On the current implementation this fails; after seeding at module init it passes. A companion case asserting the same result *with* a preceding read documents that the two orders now agree.
2. **Server-side invariant (Minor).** Add a case — a `@jest-environment node` block, or one that deletes `global.document` — asserting that `getCurrentOrganizationScope()` returns `{ null, null }` without a `document` and that a later browser-side call still seeds. The code comment asserts this invariant explicitly; nothing tests it, so a future refactor that hoists the read out of the guard would go unnoticed.
3. **The `scopeVersion === 0` consumer (Minor).** Whichever way you resolve that call site, `packages/core/src/modules/api_keys/backend/api-keys/create/page.tsx` should get a test proving the tenant still syncs on a repeat visit where the version legitimately stays at 0.

No integration test is needed here and the PR is right to say so — this is browser-module state with no API or database surface, and the unit level is the correct level for it.

adeptofvoltron:

- ❌ `changes-requested` — the code review found actionable issues: the lazy cookie seed is order-dependent and can swallow a real organization-change version bump, plus a stale `scopeVersion === 0` consumer and an undocumented behavior change to `getCurrentOrganizationScope()`.
- 🐛 `bug` — this is a defect fix in the organization-scope version guard, not a new capability.
- 🔹 `priority-medium` — it is an ordinary bug affecting every page load with a redundant refetch and a render blank, but it is not an outage, a data-loss risk, or a security regression.
- 🟡 `risk-medium` — the diff is one small, fully unit-covered module, but the value it produces is read by roughly 92 call sites across dashboard, CRM, sales, settings, entities and translations, so a mistake here surfaces broadly. This matches the risk the author suggested in the PR description.
- 🧪 `needs-qa` — this deliberately overrides the automated-verification exemption rather than taking the `skip-qa` the author suggested. The change mechanically qualifies for the exemption (no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, no database or API surface change, no `BACKWARD_COMPATIBILITY.md` contract broken, unit tests shipped), but the exemption's premise is that automated verification is sufficient, and the review found a reachable path the shipped tests do not cover whose failure mode is one organization's rows staying on screen under another organization's name. A human should exercise an organization switch, a stale/inaccessible selected-org fallback, and a repeat visit before this merges. Re-route to `skip-qa` if a maintainer reads the blast radius differently.

Note that `license/cla` is still reporting "Contributor License Agreement is not signed yet", which gates the merge independently of these labels.

adeptofvoltron:

adeptofvoltron:

One Major finding (the lazy cookie seed is order-dependent and can read a cookie `OrganizationSwitcher.persistSelection` just wrote, swallowing a genuine organization-change version bump) plus two Minor findings and three Nits. autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

`ci-monitoring` is applied because `test`, `build:app` and the 15 `ephemeral-integration` shards were still running when this review was submitted; a CI-result comment will follow. That label is not a claim — this PR is free for anyone to act on.

adeptofvoltron:

CI has settled on head `a80b84c`: **22 checks pass, 1 fails, `license/cla` still pending.**

The gates the earlier review reported as pending have since passed — `test` (4m27s) and 14 of the 15 `ephemeral-integration` shards, alongside `prepare`, `lint`, `ds-lint`, `scope`, `audit-scope`, `documents-multi-instance` and `mutate (shared, src/lib/frontend/organizationEvents.ts)`.

**One failure: `ephemeral-integration (10/15)`** — [job log](https://github.com/open-mercato/open-mercato/actions/runs/33032398384/job/98489873702). 150 passed, 1 failed:

`packages/core/src/modules/messages/__integration__/TC-MSG-015.spec.ts:10:3` › *Sender-Owned Message Archive Action Visibility › should not show message-level Archive for a message sent by the current user* — `Test timeout of 20000ms exceeded`, on the first attempt and again on retry #1 (20.1s / 20.3s). Playwright reported no failing locator beyond the timeout, so the log does not say which step hung.

**I do not believe this failure is attributable to your change, but I could not prove it, so please treat it as unresolved rather than dismissed.** What I checked:

- The change is *not* inert in this test environment. `packages/core/src/helpers/integration/auth.ts:263,271` — the `login()` fixture these specs use — sets both `om_selected_tenant` and `om_selected_org`, so the new cookie seed does apply under Playwright and `currentVersion` does stay at 0 where it previously reached 1. That was my first hypothesis and it does not hold up as an explanation on its own.
- The only coupling between the scope version and the page this spec drives is `packages/core/src/modules/messages/components/message-detail/hooks/useMessageDetails.ts:16`, which passes `scopeVersion` into `useMessageDetailsQueries`. There it is used purely as a cache-key element — `['messages', 'detail', id, scopeVersion]` and the matching attachments key — with **no `enabled` gate and no render gate** on the query. A version that stays at 0 changes which key the data lands under; it cannot stop the detail query from firing or the page from rendering. If anything it removes a redundant second fetch.

So the mechanism that would have worried me — a consumer treating version 0 as "scope not resolved yet" and withholding work until it bumps, which is exactly the Minor finding raised against `api-keys/create/page.tsx:60` in the review — is not present on this page. A 20s navigation timeout that reproduced on retry within a single loaded runner reads as environmental to me.

**Suggested next step:** re-run just that shard once you push the fix for the Major finding. If TC-MSG-015 times out again on a clean run, it is worth a proper look rather than another re-run, and I would want the Playwright trace artifact (`integration-test-results-10`) to see which step hung.

This does not change the review verdict, which remains `changes-requested` on the seed-ordering Major finding and is independent of CI. `license/cla` is still reporting "Contributor License Agreement is not signed yet" and gates the merge on its own.

Removing `ci-monitoring` — this follow-up closes out the CI watch, and no further follow-up will come from this agent.

adeptofvoltron:
Welcome to Open Mercato, @Geeknauts-company — and congratulations on your first PR here. 🎉

I want to say plainly that this is a genuinely strong first contribution, because a `changes-requested` verdict on its own does not convey that. A few things stood out:

- **The diagnosis is the hard part, and you got it right.** Spotting that `{ organizationId: null, tenantId: null }` conflates "all organizations" with "not known yet", and then tracing that ambiguity through to a per-page-load double fetch and a visible render blank, is real debugging. Plenty of contributors would have reached for `keepPreviousData` and papered over the symptom — you not only avoided that, you explained in the PR description *why* it would be wrong across a real organization switch. That is the reasoning of someone who understood the invariant rather than the stack trace.
- **You enumerated what must keep working, not just what you fixed.** The "What still bumps, deliberately" section, with a test pinning each case, is exactly how a change to a widely-read guard should be presented. The case proving that a first announcement still *dispatches* while the version holds is the one I most wanted to see, and it was already there.
- **You flagged your own compromise.** The `MAINTAINER NOTE` about the third copy of the cookie constants, with a proposed resolution and an explicit ask about scope, is better practice than most PRs manage. My answer, for the record: take it as a follow-up rather than folding it in here — it touches import paths listed in `BACKWARD_COMPATIBILITY.md` and deserves its own diff.
- **You were honest about your test evidence**, including which pre-existing failures reproduce on a clean `develop`. That saves a reviewer real time and builds trust quickly.

The Major finding I raised is a timing bug in an otherwise correct fix, not a flaw in the idea — the seed needs to be taken at module initialization so it cannot read a cookie the switcher has already rewritten. It is a small change to a design that is fundamentally sound.

**A suggestion that may save you a round trip:** this repo ships an agent skill, `om-auto-fix-pr`, built for exactly this situation. Pointed at a PR number it merges the latest base, then loops review → autofix → CI stabilization → UI verification until the PR is approvable and green, turning non-blocking nits into tracked follow-up issues rather than letting them stall the branch. It never merges the PR itself, so you stay in control of the outcome. If you have the agent harness set up locally, `/om-auto-fix-pr 5690` is a reasonable next move — it will pick up the findings from this review and work through them. Fixing it by hand is of course completely fine too; the fix itself is a handful of lines.

Two things outside the code that will gate the merge regardless of the review, so they are worth clearing early:

1. **The CLA is still unsigned.** `license/cla` reports "Contributor License Agreement is not signed yet" — the checklist box is ticked but cla-assistant has not recorded a signature for the account. That blocks the merge on its own.
2. **I routed this `needs-qa` rather than the `skip-qa` you suggested.** Your reasoning was sound and the change does mechanically qualify for the automated-verification exemption; I overrode it because the review found a reachable path the shipped tests do not cover, and the failure mode is one organization's data displayed under another's name. Rationale is in the label comment, and a maintainer can re-route if they read the blast radius differently. Nothing to do on your side.

Thanks for taking on a subtle bug for a first contribution rather than an easy one. Looking forward to the next pass.

Geeknauts-company:
@adeptofvoltron thanks — the Major was real, and I reproduced it before touching anything rather than taking it on trust. All three findings are addressed in 72e0644b, pushed as a **new commit on top of the head you reviewed** rather than an amend, so the delta reads on its own.

**Major — seed timing.** `seedScopeFromCookies()` now runs once at module initialisation, the `scopeSeeded` latch is gone (a single unconditional call needs no idempotence), and the lazy calls are dropped from `getCurrentOrganizationScope()` and `emitOrganizationScopeChanged()`. The `typeof document === 'undefined'` guard stays and is what keeps the server's view at `{ null, null }`.

Three cases pin it, under *"a cookie rewritten just before the announcement"*: your hazard (nothing read the scope first), its baseline (a component did), and the `persistTenant()` variant you flagged as the subtler instance, where `load()` writes `om_selected_tenant` on the way to the emission. Measured against the lazy source, the **first and third fail**; the second passed under it, which is precisely the complaint — the same page produced different versions depending on mount order, so the baseline is there to document that the two orders now agree.

**Minor — the `scopeVersion === 0` consumer.** Fixed rather than documented, taking your first option: `api-keys/create/page.tsx` reads `useOrganizationScopeDetail()` and `useOrganizationScopeVersion()` is dropped from the page entirely. It was the only such consumer in the monorepo. Three cases cover it and one fails against the old guard.

In candour about the other two: they are baselines, and their comments say so. The page's *observable* behaviour barely moves, because `loadInitialScope()` resolves the same tenant a beat later — exactly the containment you identified. I could not construct a case where the old guard produces a visibly wrong screen today: on the non-ok branch `actorIsSuperAdmin` is set `false`, so the loader takes the unscoped path and never reads `selectedTenantId`. The change is worth making anyway, because the page was depending on a meaning the version no longer carries, but I would rather say that plainly than dress a contract cleanup up as a user-visible bug fix.

**Minor — the changed public return value.** Now stated in the PR description, under *Behavioural contract changes*, together with the `=== 0` meaning. `organizationEvents.server.test.ts` is the test you asked for: jest's `node` environment, asserting that without a `document` there is no seed and without a `window` `emitOrganizationScopeChanged` mutates nothing, so one request's organization cannot become the shared module's idea of the scope. It also asserts that it really is running without a DOM, so the file cannot quietly become vacuous if someone adds a jsdom pragma later.

**Nits.** The comment block was compressed while the seed-timing rationale was being added to it, so it did not grow on net; the `MAINTAINER NOTE` is kept, shortened. The third cookie parser and its blank-value divergence from `parseSelectedOrganizationCookie` I have deliberately left here — collapsing the parsers is the point of the constants hoist, so it belongs in that follow-up rather than as a fourth variation added now. Understood on the hoist being its own PR plus an issue; I will open both.

Verification is in the description. Short version: `packages/shared` frontend 14/14, `packages/core` `api_keys` 39/39, and the whole `packages/shared` suite goes 1976 → 1982 passing with the pre-existing failure set byte-identical.

Geeknauts-company:
Issue for the constants hoist, as requested: #5713. I will send the follow-up PR against it separately.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-28T08:08:46Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review: fix(shared): stop counting scope initialisation as an organization change

**Re-review** of head `72e0644b` (round 3, following @adeptofvoltron's `changes_requested` on `a80b84c`). Reviewed in an isolated worktree at the PR head.

## 🎯 Summary

Round 2 moved `seedScopeFromCookies()` from a lazy, `scopeSeeded`-latched call inside the readers to a single unconditional call at module initialisation (`packages/shared/src/lib/frontend/organizationEvents.ts:120`), dropped the lazy calls from `getCurrentOrganizationScope()` and `emitOrganizationScopeChanged()`, stopped `packages/core/src/modules/api_keys/backend/api-keys/create/page.tsx` reading `scopeVersion === 0` as "nothing announced yet", and added the SSR-side test the previous round asked for. All three findings from that review are addressed, and I verified each one against the branch rather than taking the PR description on trust.

The eager seed is the right shape. `organizationEvents.ts` is imported by every writer of these cookies — `OrganizationSwitcher.tsx:7` in both `apps/mercato` and the `create-app` template — so its module body is evaluated strictly before any component can rewrite `om_selected_org` / `om_selected_tenant`. That makes the snapshot genuinely "what the server resolved this document from", which is the premise the whole design rests on, and it removes the mount-order dependence that made the previous shape order-sensitive. The `typeof document === 'undefined'` guard keeps the call inert on the server, where the module is shared across requests, and `emitOrganizationScopeChanged` still refuses to mutate without a `window` — so the invariant holds from both directions.

Reviewed scope: the five changed files, plus every consumer that makes the behaviour change load-bearing — `useOrganizationScope.ts`, both `OrganizationSwitcher.tsx` copies, `OrganizationScopeBoundary.tsx`, `BackendChromeProvider.tsx`, `DataTable.tsx`, `AiChatSessions.tsx`, the three `packages/search` components, and all 16 `useOrganizationScopeDetail()` call sites.

## Verdict

✅ **approve** — no blocker, major or minor findings. Three nits follow, all in the test surface or deferred-follow-up territory; none of them needs to be fixed on this PR. Per this repository's verdict rule, Low/nit findings alone do not request changes.

## 🧪 Validation Gate

**Validation source: local fallback** (`jest` on the two changed packages, `tsc --noEmit` on `packages/shared`).

This repository's `om-auto-review-pr` extension prefers GitHub check results over re-running the gate. That data does not exist for this head: both workflow runs on `72e0644b` are at conclusion `action_required` — the fork-PR approval gate — and the only check runs reported for the SHA are `[code]smith` (skipped) and `license/cla` (now **passing**, so the CLA concern from the previous review is resolved). With no usable CI signal, I fell back to the narrowest local scope that covers the diff.

| Command | Status | Notes |
|---------|--------|-------|
| `jest packages/shared/src/lib/frontend/__tests__` | ✅ PASS | 2 suites, 14 tests. |
| `jest packages/core/src/modules/api_keys` | ✅ PASS | 5 suites, 39 tests. |
| `tsc --noEmit -p packages/shared` | ⚠️ INCONCLUSIVE | Two pre-existing errors in `src/lib/data/engine.ts:233,635` (`emitterModuleId` not in `EmitOptions`), a file this PR does not touch — an artifact of resolving `@open-mercato/events` from an unbuilt worktree, not a PR regression. |
| `yarn lint` | ⏭️ NOT RUN | ESLint could not resolve `eslint-config-next` in the review worktree. The diff adds no new lint surface beyond ordinary TS/TSX. |
| `yarn i18n:check-sync` / `check-usage` | ⏭️ NOT RUN | The diff adds no user-facing strings and no locale keys. |
| `yarn build:packages` / `build:app` | ⏭️ NOT RUN | Covered by CI once the fork runs are approved. |

⚠️ **No CI has run on this head at all.** The 23 checks the previous round reported were on `a80b84c`; `72e0644b` has never been built or tested by CI. **A maintainer needs to approve the pending workflow runs** (`CI for Develop&Main`, `Mutation tests`) before this merges — my local run is this review's evidence only and is not a substitute for branch protection. The branch is also 18 commits behind `develop`, though `mergeable` is `MERGEABLE` and none of those 18 commits touch any file this PR touches, so I expect no semantic conflict.

## ✅ Previous-round findings — verified addressed

**Major — the lazy seed is order-dependent (`organizationEvents.ts`).** Fixed and verified, not just asserted. I checked out `a80b84c`'s lazy `organizationEvents.ts` under the round-2 test suite and re-ran it: exactly two cases fail, and they are exactly the two the PR claims —

```
● a cookie rewritten just before the announcement › bumps even though nothing read the scope first
● a cookie rewritten just before the announcement › bumps when the tenant cookie is written on the way to the announcement
```

— while the third case in that block (the "a component read the scope first" baseline) passes under both, which is precisely the mount-order asymmetry the finding described. Restoring the eager seed takes the suite to 14/14. The `scopeSeeded` latch is gone and the single unconditional call is correct: the `typeof document` guard makes re-entrancy moot.

**Minor — the `scopeVersion === 0` consumer.** Fixed by dropping `useOrganizationScopeVersion()` from `create/page.tsx` entirely and keying the effect on `scopeDetail.tenantId`. I independently confirmed the claim that this was the only such consumer: a monorepo-wide scan for comparisons against the scope version returns no other `=== 0` / `!== 0` guard across the 139 files that read it. Reverting `page.tsx` to `a80b84c` under the new test file fails exactly one of the three added cases, matching the PR's own accounting of which are regression probes and which are baselines.

**Minor — the changed public return value of `getCurrentOrganizationScope()`.** Now documented in the PR body under "Behavioural contract changes, for downstream module authors", and `organizationEvents.server.test.ts` pins the SSR half. I walked the ~16 `useOrganizationScopeDetail()` consumers looking for one that renders on the value at first paint and found none — every one feeds it into an effect, a query key, or a submit payload (`customers/.../create/page.tsx:62,91`, `currencies/.../create/page.tsx:113`, `useWmsInventoryMutationAccess.ts:65`). Two call sites deserve a positive note rather than a concern: `packages/search`'s `hasActiveOrganizationSelection()` already fell back to reading the same cookie, so the seed makes its first branch return what its second branch would have returned anyway; and `AiChatSessions.tsx:236`'s `useState` initializer now lands on the correct scoped storage key immediately instead of bucketing into `no-tenant:no-org` and being corrected by the subscription a beat later. `OrganizationScopeBoundary.tsx` carries its own `hasInitializedScopeRef` first-announcement guard and is unaffected.

**Nits.** The comment block was compressed rather than grown, and #5713 is open for the constants hoist. Deferring the third cookie parser and its blank-value semantics to that follow-up is the right call — I could not construct a case where the divergence produces a wrong result today either.

## Findings

### 🔸 Nit

**`packages/core/src/modules/api_keys/backend/api-keys/create/__tests__/CreateApiKeyPage.test.tsx:205-229` — the one discriminating case models a module state that cannot actually occur, so the reachable difference is the one left on a baseline.**

The suite mocks `useOrganizationScopeVersion: () => 0` permanently and then changes `announcedScope` from `tenant-a` to `tenant-b`. That combination is unreachable in the real module: `emitOrganizationScopeChanged` sets `hasChanged` from the same comparison that gates the dispatch, so the only dispatch that does *not* bump is the first emission repeating the seed — and that one carries a detail identical to the seed, so `scopeDetail.tenantId` does not move. Any announcement that actually changes the announced tenant has already incremented the version. The case therefore guards the mock contract ("this page must not couple to the version") rather than a reachable regression, which is a legitimate thing to pin — I am flagging it only so the suite is not read as proving more than it does.

The genuinely reachable difference is at mount: with the version at 0 and the seed non-null, the old guard blocked the effect on the first render, so `selectedTenantId` waited for `loadInitialScope()`; without it, the cookie scope supplies the tenant immediately. The test that covers that shape (`:185-203`) has both sources agreeing and passes either way — correctly labelled a baseline in its own comment. If you want one case that fails against the old guard for a reason a user could observe, seed `announcedScope` with a tenant and have `apiCall` resolve `{ ok: false }`, then assert `selectedTenantId` still reached the loader.

**`packages/shared/src/lib/frontend/organizationEvents.ts:113-114` — the two blank-cookie branches in the seed are unpinned, and blank cookies are reachable by this repository's own account.**

`readScopeCookie` deliberately distinguishes `''` from `null`, and the seed then collapses both blanks: `!org || org === ALL_ORGANIZATIONS_COOKIE_VALUE` maps a blank org to "all organizations", and `tenant || null` maps a blank tenant to "no tenant". The suite pins the `__all__` sentinel (`:93-100`) and both *absent*-cookie paths (`:182-192`), but neither blank. `OrganizationSwitcher.load()` documents blank cookies as a live legacy state ("Covers a cookie left blank by an earlier build") and has a branch specifically to clear one, so this is not theoretical. Dropping `tenant || null` in particular would seed `tenantId: ''`, which makes a subsequent `{ org, tenantId: null }` announcement compare unequal and reintroduce exactly the spurious bump this PR removes. Two cases mirroring the existing absent-cookie pair would close it, and since `mutate (shared, …/organizationEvents.ts)` runs against this very file, these are the mutants most likely to survive.

**`packages/shared/src/lib/frontend/organizationEvents.ts:110` — the both-cookies precondition discards the organization half for a reason that only justifies the tenant half.**

`if (org === null || tenant === null) return` means a session with `om_selected_org` set but `om_selected_tenant` absent gets no seed at all and keeps bumping on every page load. The comment's justification — "without the cookie the server fell back to the JWT's organization, which the client cannot see" — is sound for a missing *org* cookie, but when the org cookie is present the server resolved the organization from it regardless of the tenant cookie. I am not asking you to change it here: seeding `{ org, tenantId: null }` in that case would suppress the bump on a `{ org, null }` announcement, and that suppression is not obviously safe when the server rendered under the JWT tenant. The conservative choice is defensible and the cohort is small (`load()` writes the tenant cookie as soon as the endpoint resolves one, so the steady state has both). Worth a sentence in #5713's follow-up so the asymmetry is a recorded decision rather than an accident.

## 💥 Breaking Changes

- [x] No HTTP route URL removed or renamed; no method changed
- [x] No event or message name renamed or removed — `ORGANIZATION_SCOPE_CHANGED_EVENT` and its detail shape are untouched
- [x] No CLI command or flag renamed or removed
- [x] No database table or column renamed or removed
- [x] Where a contract had to change: documented rather than bridged

No `BACKWARD_COMPATIBILITY.md` surface is broken. The two behavioural contract changes the previous round asked to have stated — the pre-announcement return value of `getCurrentOrganizationScope()`, and the meaning of `getCurrentOrganizationScopeVersion() === 0` — are both now written up in the PR body, and the only consumer of the second one has been migrated off it. That is the right resolution for a change that needs no deprecation bridge.

## 🧪 Test Coverage

Strong, and honest about itself — which I want to call out specifically, because the PR labels its own baselines as baselines instead of presenting five passing tests as five regression probes. That made this re-review cheap to verify and I could confirm the claims exactly.

`organizationEvents.test.ts` builds a fresh module registry per case through `jest.isolateModules`, which is a faithful model of a page load rather than a convenience, and the new "a cookie rewritten just before the announcement" block pins the seed-timing invariant from all three angles the finding described. `organizationEvents.server.test.ts` is in the right place for the right reason — the package's jest environment is `node`, the jsdom suite carries a per-file `@jest-environment jsdom` pragma, and the two invariants cannot share a file. Its first case asserting `typeof document === 'undefined'` before the other two run is a good defensive touch: without it, a stray pragma added later would leave both assertions passing while testing nothing.

Gaps, both nits above: the blank-cookie branches in `seedScopeFromCookies`, and an api_keys case that fails against the old guard for an observable reason. No integration test is needed — this is browser-module state with no API or database surface, and the unit level is the correct level.

## 🚦 CI status

`PENDING` — not failing. Both workflow runs on `72e0644b` sit at `action_required` awaiting fork-PR approval, so this review does not certify any CI gate and is being submitted now rather than waiting on one. Branch protection and the merge queue remain the authority. `license/cla` has since passed.

pkarw:

Supersedes the round-2 rationale above, after the re-review of head `72e0644b`.

- 🚀 `merge-queue` — the round-2 findings are all addressed and verified against the branch, so the code review passed; `needs-qa` stays on, which means the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — the change alters browser-visible behaviour on a backend page (`api_keys/create`) and the module state behind ~139 scope-keyed readers, so it is not covered by the automated-verification exemption and a human should exercise it.
- 🐛 `bug` — this fixes a defect (initialisation counted as an organization change, causing a spurious cache-busting version bump on every page load), not a new capability.
- 🔹 `priority-medium` — a real per-page-load refetch cost with visible blanking on screens that gate on `isLoading`, but no data loss, no security exposure and no release blocker, so it is an ordinary bug rather than a high-priority one.
- 🟡 `risk-medium` — the blast radius is wide (a shared frontend module read in ~139 places) and it changes a behavioural contract on a public export, which argues upward; against that, the change is small and self-contained, adds 14 unit tests that pin both the browser and server invariants, touches no auth, scoping, schema or money path, and its worst failure mode is a stale re-key rather than a wrong-tenant read. Medium is the right balance and it was already correctly rated in round 2.

**Not yet mergeable, for two reasons outside the code review:**
1. @adeptofvoltron's round-2 `CHANGES_REQUESTED` review is still outstanding, so GitHub's aggregate review decision remains `CHANGES_REQUESTED` despite this approval. It needs a re-review or a dismissal.
2. No CI has run on head `72e0644b` — both workflow runs are at `action_required` awaiting fork-PR approval from a maintainer.

pkarw:

The change is browser-module state with no API or database surface, so almost everything here is observed through **network activity and screen behaviour on an organization switch**, not through new UI. Have DevTools → Network open (filter XHR) and DevTools → Application → Cookies for the whole pass.

### P0 — Organization/tenant scoping is still correct after a real switch

This is the one that matters. The PR deliberately *suppresses* a cache-busting version bump; the failure mode it must not introduce is suppressing a bump that was needed, which shows up as one organization's rows rendered under another organization's name.

**Where to click**
- `/backend` (dashboard), `/backend/customers/people`, `/backend/customers/companies`, `/backend/sales/orders` — any scope-keyed list with visibly different data per organization
- The organization switcher in the topbar

**What to verify**
- Sign in as a user with access to **two organizations that hold clearly different records**. Open a list page under org A and note the rows. Switch to org B in the topbar → the list must refetch and show **org B's** rows, with no org A row surviving the transition.
- Switch back to org A → org A's rows return. Repeat the A→B→A cycle three or four times without reloading the page; the switcher uses `router.refresh()` rather than a navigation, so the page is never remounted and a missed re-key would accumulate.
- Do the same on the settings surface (`/backend/settings/...`), which remounts on scope change through a different code path.
- If your build exposes a tenant selector (super admin), repeat the cycle changing **tenant while keeping the same organization** — that transition must also refetch.

**What can go wrong**
- Stale rows: after switching to org B the table still shows org A's records, or shows them briefly and never corrects. That is the regression this PR must not cause — report it with the two organization names and the route.
- A "lost access" case: if you can arrange for a selected organization to be revoked mid-session (or point the `om_selected_org` cookie at an organization the account cannot see) and then reload, the app must correct itself to a permitted organization **and** refresh the visible data, not just the switcher label.

### P1 — The intended win: no spurious refetch on an ordinary page load

**Where to click**
- Any scope-keyed backend list, e.g. `/backend/customers/people`

**What to verify**
- With DevTools → Network open, hard-reload a scope-keyed page **without touching the switcher**. On a repeat visit (both `om_selected_org` and `om_selected_tenant` already present in Application → Cookies), the scope-dependent list request should fire **once**, not once and then immediately again for the same scope.
- Watch the page body while it loads: screens that gate their render on `isLoading` should not render, blank out, and re-render. A brief single loading state is expected; a second blank after data has appeared is the old behaviour.
- Compare against `develop` on the same account if you can — the before/after is a doubled request and a visible flash.

**What can go wrong**
- Nothing loads at all, or a list stays empty on first paint and only fills after a manual switch — that would mean the seed adopted a scope the server did not render from.
- A brand-new session (no selection cookies yet) legitimately still shows the old double-fetch on the very first load; that is expected and documented, not a QA failure. Only judge repeat visits.

### P1 — API key creation, super-admin tenant scoping

The only file with a user-visible UI change.

**Where to click**
- `/backend/api-keys/create`

**What to verify**
- As a **super admin**: open the page and the **Roles** field. Options must load and must belong to the currently selected tenant — not another tenant's roles, and not an empty list.
- Submit the form to create a key. It must succeed and the created key must land in the correct tenant (check it appears in `/backend/api-keys` under the same scope).
- With the page open, switch organization/tenant in the topbar (the page is not remounted). Reopen the Roles dropdown → the options must follow the **new** scope.
- As a **non-super-admin** with `api_keys` access: the Roles field must still populate, scoped to your own tenant.

**What can go wrong**
- Roles list is empty for a super admin, or the submit is rejected with a "tenant required" error — the tenant never reached the form.
- Roles from a **different** tenant appear in the list — that is a data-scoping leak and should be reported as P0, not P1.
- After switching scope with the page open, the dropdown still shows the previous tenant's roles.

### P2 — Search and AI chat scope, light touch

**Where to click**
- The topbar search box; the global search dialog (`Cmd/Ctrl+K` if bound); the AI chat panel if enabled

**What to verify**
- Search returns results scoped to the selected organization, immediately after a page load and again after switching organizations.
- If AI chat is enabled: open it, start a conversation, switch organization, and reopen the panel — the previous organization's session list must not appear under the new organization.

**What can go wrong**
- Search returning cross-organization results, or returning nothing on a fresh page load where it used to work.
- Chat sessions bleeding across an organization switch.

### Perceived performance (any changed route)
- Cold-load `/backend/customers/people` and `/backend/api-keys/create` with the cache disabled; capture a screenshot of the first useful shell state. The shell should appear at least as fast as on `develop`, with no added blank interval.
- Check one list route at a mobile viewport (375px) for the same no-double-load behaviour.

### Pass/fail

> Note for whoever merges: two things still block independently of QA — @adeptofvoltron's round-2 `CHANGES_REQUESTED` review is still outstanding, and no CI has run on head `72e0644b` (both workflow runs are at `action_required` awaiting fork-PR approval).

pkarw:

Re-review of head `72e0644b`. All three round-2 findings verified addressed against the branch, not just against the PR description: the two seed-timing cases genuinely fail under `a80b84c`'s lazy seed and pass under the eager one, the `scopeVersion === 0` consumer really was the only one in the monorepo, and one of the three new `CreateApiKeyPage` cases really does fail against the old guard — exactly the accounting the PR gave. No blocker, major or minor findings this round; three nits, all optional and none needing a change here.

Labels moved `changes-requested` → `merge-queue`, `needs-qa` retained.

**Two things block the merge independently of this review:** @adeptofvoltron's round-2 `CHANGES_REQUESTED` review is still outstanding, so GitHub's aggregate decision remains `CHANGES_REQUESTED`; and no CI has run on this head — both workflow runs sit at `action_required` awaiting fork-PR approval from a maintainer.

autofix: skipped (not my PR — re-run with --autofix to fix it here)

Geeknauts-company:
Refreshed onto today's `develop` (`9c743c9a`, clean merge) and re-verified, since this has been sitting a fortnight: `packages/shared` 195/195 suites, 2209 tests, with the two suites this PR adds (`organizationEvents.test.ts` and the `.server` half) green at 14/14.

**This PR is approved and in `merge-queue`, and two things are holding it.**

**1. A stale `changes-requested` review.** @adeptofvoltron requested changes on head `a80b84c` on 2026-08-27. All three findings were addressed in `72e0644b` the same day, and @pkarw re-reviewed that head on 2026-08-28 and **approved**, verifying each against the branch rather than against the description — including that the two seed-timing cases genuinely fail under the lazy seed and pass under the eager one, and that the `scopeVersion === 0` consumer really was the only one in the monorepo.

Because the earlier review was never dismissed, GitHub still computes `reviewDecision: CHANGES_REQUESTED` for the PR as a whole, which is what branch protection reads. @adeptofvoltron — could you either dismiss it or re-review the current head? Nothing is waiting on me.

**2. `needs-qa`.** @pkarw's approval left `needs-qa` on (the change alters browser-module state) and posted manual QA instructions, so the QA-approval gate holds the merge until a reviewer swaps `merge-queue` for `qa`, runs the routes and adds `qa-approved`. Flagging it in case the label pair reads as "already queued".

For anyone picking up the QA: the property to check is that switching organization still re-keys every scope-dependent query — that bump is what stops the previous organization's rows being rendered under the new one's name, since the header switcher calls `router.refresh()` rather than navigating, so the client tree stays mounted. What must *not* happen any more is a blank on an ordinary visit with both scope cookies already in the jar. A browser's first-ever visit still blanks once, deliberately: without the cookie the client cannot know which organization the server rendered from, so a first announcement naming a different one must invalidate.

patzick (approved):

patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-09-10T15:21:12Z. Other auto-skills will skip this PR until the lock is released.

patzick (changes requested):
# 🔍 Code Review: fix(shared): stop counting scope initialisation as an organization change

**Re-review** of head `401ce935` (round 4). I refreshed the branch onto today's `develop` (`a45dece0`) with `gh pr update-branch` before reviewing — it was 118 commits behind. Validation source: **GitHub checks** (no local runs).

## 🎯 Verdict: `changes_requested` — one blocking issue, and it is not in the diff

**The diff is correct.** I re-read the seed, the emit guard, and both suites; round 3's approval still holds on the code. What blocks merge is a red required gate that this PR reproducibly causes, via a fragile integration test rather than a flaw in the change itself.

## 🔴 Blocker — `TC-MSG-015` fails deterministically on this branch

`ephemeral-integration (10/15)` fails on **every** head of this PR that has run full CI:

| head | run | result |
|---|---|---|
| `a80b84cb` | 33032398384 | shard 10 ✘ `TC-MSG-015`, 2/2 attempts |
| `9c743c9a` | 34472862296 | shard 10 ✘ `TC-MSG-015`, 2/2 attempts |
| `401ce935` (today's `develop`) | 34495109757 | shard 10 ✘ `TC-MSG-015`, 2/2 attempts — 14/15 shards green |

That is **6/6 attempts across 3 independent runs**, including retries. It is not environmental: PRs #6035 and #5913 merged today on a comparable base at 15/15 green, and no other open PR currently has a failing integration shard. Merging `develop` did not help — none of the 118 new commits touch this test or the message-detail UI.

### What actually happens

Not a load failure. From the artifact for run 34472862296 the page renders fine — message loaded, `Actions` button visible, click **succeeds** — and then the dropdown never opens, so `menuitem "Delete"` never appears and the test burns its 20 s budget.

The retry trace screencast pins the moment:

- **~80 ms after the click**, the topbar organization switcher still reads **"Loading…"**
- **~250 ms after**, it has settled to **"Acme Corp"** — with **no menu open**

The click was swallowed: it landed before the page was interactive, and the test clicks exactly once.

### Why this PR is the trigger

```ts
// useMessageDetailsQueries.ts
queryKey: ['messages', 'detail', id, scopeVersion]   // :25
const isLoadingDetail = detailQuery.isLoading        // :56
```

`MessageDetailPageClient` returns `<LoadingMessage/>` while `isLoadingDetail`, which removes the `Actions` button from the DOM.

- **Before the fix**, the first announcement bumped the version 0 → 1 → new query key → no cached data → `isLoading: true` → the button disappeared and came back. The test's `expect(actionsButton).toBeVisible()` therefore had to wait through that second cycle, which by construction completes only *after* the switcher has announced. The click landed on a settled, hydrated page — by luck.
- **After the fix**, there is no bump, so no second cycle. `Actions` is present continuously from first paint, `toBeVisible()` resolves immediately, and the click fires into a page that has not settled.

**The test was relying on exactly the spurious refetch this PR exists to remove.**

### The repo already knows this menu needs a retry

`TC-MSG-005` drives the *same* page, the *same* Actions menu, with the *same* locator — and wraps opening it in a 4-attempt retry loop:

```ts
const clickActionMenuItem = async (actionName: RegExp): Promise<void> => {
  for (let attempt = 0; attempt < 4; attempt += 1) {
    await actionsButton.hover();
    const menuItem = page.getByRole('menuitem', { name: actionName }).first();
    if (await menuItem.isVisible().catch(() => false)) { await menuItem.click(); return; }
    await actionsButton.click();
    ...
    await page.waitForTimeout(250);
  }
  ...
}
```

`TC-MSG-009`, also on the message detail page, waits explicitly for `"Loading message..."` to clear before interacting. `TC-MSG-015` does neither — a bare single `.click()` with no settle wait and no retry. It is the one test on this surface that skips the established pattern, and it passed only because the version bump handed it an incidental settle window.

### Requested change

Harden `packages/core/src/modules/messages/__integration__/TC-MSG-015.spec.ts` using the `TC-MSG-005` pattern — retry the open, or wait for the detail loading state to clear before clicking. It belongs in this PR because this PR is what surfaces it. I'd rather see the test fixed than the version bump reinstated: the bump is the defect you correctly diagnosed, and restoring it to keep a test green would trade a real user-facing refetch-and-blank for a green tick.

Worth stating explicitly for the record: **this is not a new user-facing hazard.** The pre-fix sequence also let a real user click into the first window *and* blanked the content afterwards. What remains post-fix — a click before hydration completes — is the generic Next.js hydration gap present on every page in the app, not something this change introduces.

## 🟡 Low / nits (non-blocking)

- **Comment volume.** `organizationEvents.ts` carries ~60 lines of prose against ~40 lines of code, which sits awkwardly with `AGENTS.md` → Code Quality ("no inline comments (self-documenting code)"). Raised as a nit in round 2 and compressed; still the first thing a reader meets. The seed-timing rationale earns its keep — the cookie-constant maintainer note could move wholesale to #5713.
- **The seed needs *both* cookies.** `seedScopeFromCookies()` returns early unless `om_selected_org` **and** `om_selected_tenant` are readable. On installs where the tenant cookie is never written, the fix silently does not engage and every load still bumps. That is the documented, deliberate fail-safe, but it narrows the fix's reach more than the PR description implies — worth a sentence so nobody later reports "the fix didn't work" on a single-tenant install.
- Constants duplication is already tracked in #5713. No `BACKWARD_COMPATIBILITY.md` surface is broken; the two behavioural contract changes are correctly documented in the description.

## ✅ What I verified as fine

- `test` (unit) is **green** on `401ce935`, as are `lint`, `ds-lint`, `scope`, `audit-scope`, `prepare`, `mutate (shared, organizationEvents.ts)` and `license/cla`.
- The seed is inert on the server (no `document` → no seed; no `window` → no mutation), and `organizationEvents.server.test.ts` pins it. That matters more with an eager seed than a lazy one, and the file's docblock says so.
- The `getCurrentOrganizationScope()` SSR/hydration divergence is real and acknowledged. I sampled the consumers (`useWmsInventoryMutationAccess`, the WMS stock widget, `CreatePersonDialog`, `business_rules/rules/[id]`) — all feed the value into effects and query keys, none into first-paint output. Nothing new to flag.
- The round-2 findings remain addressed.

patzick:

Supersedes the round-3 rationale, after the re-review of head `401ce935` (branch refreshed onto today's `develop` as part of this run).

- ❌ `changes-requested` — replaces `merge-queue`. The code review passes on the diff, but `ephemeral-integration (10/15)` fails deterministically on this branch (`TC-MSG-015`, 6/6 attempts across 3 independent full-CI runs, most recently 34495109757 on today's `develop`, 14/15 shards green). `merge-queue` means "all gates passed and ready to merge", which is no longer true.
- 🐛 `bug` — unchanged; still a defect fix in the organization-scope version guard.
- 🧪 `needs-qa` — unchanged. The change alters browser-visible scope/refetch behaviour, so the QA-approval gate still applies once CI is green.
- 🟡 `priority-medium` / 🟠 `risk-medium` — unchanged.

The requested change is in a test, not in the diff: `TC-MSG-015` opens the message Actions menu with a single un-retried click and no settle wait, unlike `TC-MSG-005` (4-attempt retry helper) and `TC-MSG-009` (waits for "Loading message...") on the same surface. It passed before only because the spurious version bump this PR removes gave it an incidental settle window. Details and evidence in the review.

patzick:
Thanks @Geeknauts-company — and to be clear about where this landed, because a second `changes_requested` after an approval reads worse than it is.

**The diff is not the problem.** I re-read the seed, the emit guard and both suites at head `401ce935`; round 3's approval holds on the code, and the round-2 findings stay addressed. What I could not wave through is a red gate: `ephemeral-integration (10/15)` fails on **every** head of this PR that has run full CI — 6/6 attempts across 3 runs, most recently on today's `develop` with the other 14 shards green.

I chased it far enough to be sure it is yours and to be sure it is not your code's fault. The failing test loads the page fine, clicks `Actions` successfully, and then waits 20s for a menu that never opens. The trace screencast shows why: ~80ms after the click the org switcher still reads "Loading…", ~250ms later it has settled with no menu open. The click was swallowed, and the test clicks exactly once.

Then the part that actually explains it: `useMessageDetailsQueries` keys on `scopeVersion` and gates render on `detailQuery.isLoading`. **Before your fix**, the first announcement bumped 0 → 1, re-keyed the query, and put the page through a second loading cycle — which removed and restored the `Actions` button, forcing the test's `toBeVisible()` to wait until after the switcher had announced. The click landed on a settled page by luck. **After your fix** there is no bump, no second cycle, and the button is there from first paint, so the click fires early.

So the test was leaning on precisely the spurious refetch you set out to remove. `TC-MSG-005` — same page, same menu, same locator — already wraps the open in a 4-attempt retry, and `TC-MSG-009` waits for "Loading message..." to clear. `TC-MSG-015` does neither. It is the one test on this surface without the guard, and your change is what exposed that.

**What I'd like:** harden `TC-MSG-015` with the `TC-MSG-005` pattern, in this PR. Please don't reinstate the bump to get the tick — the bump is the defect you correctly diagnosed, and trading a real user-facing refetch-and-blank for a green check would be the wrong direction.

Two housekeeping notes: I refreshed the branch onto `develop` (`a45dece0`) as part of this run, since it was 118 commits behind — that did not affect the failure. And I did not push the test fix myself: it is your PR, and the change is small enough that you'll want it in your own commit. Say the word if you'd rather I pushed it.

For the record, the diagnosis in this PR was right, the round-2 response was thorough, and the seed-timing argument in `organizationEvents.test.ts` is genuinely well made. This last one is a two-line test guard away.

patzick:

Re-review of head `401ce935`. Validation source: **GitHub checks** (no local runs). Branch refreshed onto `develop` `a45dece0` during this run (was 118 behind).

One blocking finding, and it is not in the diff: `ephemeral-integration (10/15)` fails deterministically on `TC-MSG-015` — 6/6 attempts across 3 independent full-CI runs (`a80b84cb`, `9c743c9a`, `401ce935`), with 14/15 shards green on the latest and comparable PRs merging green today. Root-caused from the run artifact and trace to a single un-retried click on the message Actions menu that this PR's (correct) removal of the spurious version bump now lets fire before the page settles; `TC-MSG-005` and `TC-MSG-009` already guard the same surface. Requested fix is in the test.

Plus two Low/nits (comment volume in `organizationEvents.ts`; the seed needing both cookies narrowing the fix's reach). Code review of the change itself passes — round-2 findings remain addressed, unit `test` and all other gates are green.

`needs-qa` retained: the QA-approval gate still holds the merge once CI is green. autofix: skipped (not my PR — re-run with `--autofix` to apply the test guard here).

PR #5698, opened 2026-08-27
fix(customer_accounts): open the portal at the organization slug (#5668)

### LINKED ISSUE #5668: bug: "Open Portal” button opens an incorrect URL
## Summary

The “Open Portal” button opens `/portal` instead of including the slug of the current organization.

## Page URL

`/backend/customer_accounts/users`

## Steps to reproduce

1. Log in to the admin panel.
2. Navigate to `Customer Users`.
3. Select any customer organization.
4. Click `Open Portal`.
5. Check the opened URL.

## Expected result

The portal should open at `http://localhost:3000/{org-slug}/portal`, where the slug matches the URL identifier of the selected organization.

## Actual result

The button opens `http://localhost:3000/portal`, and the page returns `404 – This page could not be found`.

## Environment

- Branch: `develop`
- Module or area: `customer_accounts`, customer portal
- Browser/Runtime: Chrome, Node.js 24.13.0, Next.js development server

## Additional context

The likely cause is located in `packages/core/src/modules/customer_accounts/backend/customer_accounts/users/page.tsx` and `packages/core/src/modules/customer_accounts/backend/customer_accounts/settings/page.tsx`. The link is built as `${origin}/portal` without including the organization slug. Link generation should be fixed and a regression test should be added.

---

adeptofvoltron:
Closes #5668

## 🎯 Goal
The "Open Portal" button on the customer-accounts admin pages should land the operator on their organization's actual portal instead of a 404 page.

## 🔍 Problem
On `/backend/customer_accounts/users` and `/backend/customer_accounts/settings`, clicking `Open Portal` navigated to `http://localhost:3000/portal`, which renders `404 – This page could not be found`. The expected destination is `http://localhost:3000/{org-slug}/portal`.

## 🔍 Root Cause
`buildPortalRootUrl` in `packages/core/src/modules/customer_accounts/lib/portalUrl.ts` returned `${normalizeOrigin(origin)}/portal`, with no organization segment. Portal pages are only ever mounted under `frontend/[orgSlug]/portal/**` — `packages/core/AGENTS.md` states `[orgSlug]` MUST be the first segment because portal auth and tenant resolution both assume it — so a slugless `/portal` can never match a route. The module already owns the correct convention in `lib/customerUrl.ts`, where `urlForCustomerOrg` builds `${base}/${orgSlug}/portal${path}` for every customer-facing email link; these two admin buttons were the surfaces that bypassed it.

## What Changed
- `lib/portalUrl.ts` — `buildPortalRootUrl` gains an optional second `orgSlug` argument and emits `${origin}/${encodeURIComponent(slug)}/portal`. With no slug it returns exactly the previous string, so the exported signature stays backward compatible. `buildPortalUrlPattern` and its `[org-slug]` placeholder are untouched: those render documentation of the URL shape, not an href.
- `lib/portalOrgSlug.ts` (new, server-only) — `resolveCurrentOrgPortalSlug()` reads the session's `orgId`/`tenantId` via `getAuthFromCookies()` and resolves the slug through the tenant-bound `findOrganizationInTenant` helper, whose `(id, tenant_id)` pairing prevents a cross-tenant read. It returns `null` for every miss — no session, no active organization, an organization without a slug, or a failed lookup — so a transient database problem hides one button instead of taking the admin page down. It deliberately lives outside `portalUrl.ts`, which client components import: an `EntityManager` import there would pull MikroORM into the browser bundle.
- `backend/customer_accounts/users/page.tsx` and `backend/customer_accounts/settings/page.tsx` — resolve the slug server-side alongside the existing `resolvePortalRequestOrigin(await headers())` call and pass it down as a `portalOrgSlug` prop, preserving the #5457 contract that the portal address is never derived in the client.
- `PortalUsersPageClient.tsx` and `CustomerAccountsSettingsPageClient.tsx` — accept `portalOrgSlug`, build the href from it, and render the `Open Portal` button only when a slug is available. An organization with no slug now gets no button rather than a knowingly broken link; the settings card already explains how slugs are generated and links to Manage Organizations for exactly that case.

## 🧪 Tests
- `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app` — the full configured gate, run locally, all green. `@open-mercato/shared` (184 suites / 2015 tests) and `@open-mercato/cli` (95 suites / 1806 tests) each flaked once under turbo's parallel scheduler on the run machine and pass standalone; `shared` was also green in one of the full runs. Every package is verified, though not all within a single turbo invocation.
- `lib/__tests__/portalUrl.test.ts` — new `#5668` block locking in the slug-bearing URL, trailing-slash normalization, root-relative degradation when the origin is unresolvable, the unchanged slugless fallback for `null` / `undefined` / blank, and slug escaping so a hostile slug cannot alter the path structure. These assertions fail against the pre-fix implementation.
- `lib/__tests__/portalOrgSlug.test.ts` (new) — covers the happy path while asserting the lookup receives the `(orgId, tenantId)` pair together, slug trimming, every `null` path (no session, no organization, no tenant, missing organization, null or blank slug), and graceful degradation when the lookup throws.
- `__tests__/portalUrlHydration.test.ts` — the existing #5457 guard gains a `#5668` section asserting both entrypoints resolve the slug server-side, both clients take it as a prop and never call the slugless builder, and `portalUrl.ts` stays free of ORM imports so the client bundle cannot pick up MikroORM.

## 💥 Breaking Changes
None. `buildPortalRootUrl`'s new parameter is optional and the no-slug call returns the previous string byte for byte, so existing callers and the original test expectations are unaffected. `portalOrgSlug` is likewise an optional prop on both client components.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5698"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790422491&installation_model_id=432243&pr_number=5698&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5698&signature=ab283f63add1ccb0051325b9cd06e3f7f5e0fd7b2d954c3bc0e2b9ff9e7a23b7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/customer_accounts/__tests__/portalUrlHydration.test.ts
+++ b/packages/core/src/modules/customer_accounts/__tests__/portalUrlHydration.test.ts
@@ -60,3 +60,40 @@ describe('customer_accounts portal address hydration (regression for issue #5457
     expect(source).toMatch(/from '(?:\.\.\/)+lib\/portalUrl'/)
   })
 })
+
+// The "Open Portal" action linked to `${origin}/portal`, but portal pages only
+// exist under `frontend/[orgSlug]/portal/**`, so the link always 404'd.
+const PORTAL_LINK_SURFACES = [
+  {
+    entrypoint: 'backend/customer_accounts/users/page.tsx',
+    client: 'backend/customer_accounts/users/PortalUsersPageClient.tsx',
+  },
+  {
+    entrypoint: 'backend/customer_accounts/settings/page.tsx',
+    client: 'backend/customer_accounts/settings/CustomerAccountsSettingsPageClient.tsx',
+  },
+]
+
+describe('customer_accounts "Open Portal" link (regression for issue #5668)', () => {
+  it.each(PORTAL_LINK_SURFACES)('$entrypoint resolves the organization slug on the server', ({ entrypoint }) => {
+    const source = readSource(entrypoint)
+    expect(source).toMatch(/resolveCurrentOrgPortalSlug\(\)/)
+    expect(source).toMatch(/portalOrgSlug=\{portalOrgSlug\}/)
+  })
+
+  it.each(PORTAL_LINK_SURFACES)('$client takes the slug as a prop and never links without it', ({ client }) => {
+    const source = readSource(client)
+    expect(source).toMatch(/portalOrgSlug\?: string \| null/)
+    expect(source).toMatch(/buildPortalRootUrl\(portalOrigin, portalOrgSlug\)/)
+    // A slugless call would rebuild the 404 link this regression is about.
+    expect(source).not.toMatch(/buildPortalRootUrl\(portalOrigin\)/)
+  })
+
+  // The slug resolver reaches for the EntityManager, so it must stay out of the
+  // module client components import — otherwise the ORM lands in the browser bundle.
+  it('keeps the ORM-backed slug lookup out of the client-imported portalUrl module', () => {
+    const source = readSource('lib/portalUrl.ts')
+    expect(source).not.toMatch(/@mikro-orm/)
+    expect(source).not.toMatch(/organizationLookup/)
+  })
+})
--- a/packages/core/src/modules/customer_accounts/backend/customer_accounts/settings/CustomerAccountsSettingsPageClient.tsx
+++ b/packages/core/src/modules/customer_accounts/backend/customer_accounts/settings/CustomerAccountsSettingsPageClient.tsx
@@ -15,13 +15,17 @@ const DEMO_CREDENTIALS = [
 
 export type CustomerAccountsSettingsPageClientProps = {
   portalOrigin: string
+  portalOrgSlug?: string | null
 }
 
-export function CustomerAccountsSettingsPageClient({ portalOrigin }: CustomerAccountsSettingsPageClientProps) {
+export function CustomerAccountsSettingsPageClient({ portalOrigin, portalOrgSlug = null }: CustomerAccountsSettingsPageClientProps) {
   const t = useT()
 
   const portalUrl = useMemo(() => buildPortalUrlPattern(portalOrigin), [portalOrigin])
-  const portalRootUrl = useMemo(() => buildPortalRootUrl(portalOrigin), [portalOrigin])
+  const portalRootUrl = useMemo(
+    () => (portalOrgSlug ? buildPortalRootUrl(portalOrigin, portalOrgSlug) : null),
+    [portalOrigin, portalOrgSlug],
+  )
 
   return (
     <>
@@ -47,13 +51,15 @@ export function CustomerAccountsSettingsPageClient({ portalOrigin }: CustomerAcc
               {portalUrl}
             </code>
           </div>
-          <div>
-            <Button type="button" variant="outline" size="sm" asChild>
-              <a href={portalRootUrl} target="_blank" rel="noopener noreferrer">
-                {t('customer_accounts.settings.portal_access.open_portal', 'Open Portal')}
-              </a>
-            </Button>
-          </div>
+          {portalRootUrl ? (
+            <div>
+              <Button type="button" variant="outline" size="sm" asChild>
+                <a href={portalRootUrl} target="_blank" rel="noopener noreferrer">
+                  {t('customer_accounts.settings.portal_access.open_portal', 'Open Portal')}
+                </a>
+              </Button>
+            </div>
+          ) : null}
           <p className="text-xs text-muted-foreground">
             {t(
               'customer_accounts.settings.portal_access.slug_note',
--- a/packages/core/src/modules/customer_accounts/backend/customer_accounts/settings/page.tsx
+++ b/packages/core/src/modules/customer_accounts/backend/customer_accounts/settings/page.tsx
@@ -1,15 +1,17 @@
 import { headers } from 'next/headers'
 import { Page, PageBody } from '@open-mercato/ui/backend/Page'
 import { resolvePortalRequestOrigin } from '../../../lib/portalUrl'
+import { resolveCurrentOrgPortalSlug } from '../../../lib/portalOrgSlug'
 import { CustomerAccountsSettingsPageClient } from './CustomerAccountsSettingsPageClient'
 
 export default async function CustomerAccountsSettingsPage() {
   const portalOrigin = resolvePortalRequestOrigin(await headers())
+  const portalOrgSlug = await resolveCurrentOrgPortalSlug()
   return (
     <Page>
       {/* space-y-6 preserves the gap the page header previously got as a direct <Page> child. */}
       <PageBody className="space-y-6">
-        <CustomerAccountsSettingsPageClient portalOrigin={portalOrigin} />
+        <CustomerAccountsSettingsPageClient portalOrigin={portalOrigin} portalOrgSlug={portalOrgSlug} />
       </PageBody>
     </Page>
   )
--- a/packages/core/src/modules/customer_accounts/backend/customer_accounts/users/PortalUsersPageClient.tsx
+++ b/packages/core/src/modules/customer_accounts/backend/customer_accounts/users/PortalUsersPageClient.tsx
@@ -229,9 +229,10 @@ function CreateUserDialog({
 
 export type PortalUsersPageClientProps = {
   portalOrigin: string
+  portalOrgSlug?: string | null
 }
 
-export function PortalUsersPageClient({ portalOrigin }: PortalUsersPageClientProps) {
+export function PortalUsersPageClient({ portalOrigin, portalOrgSlug = null }: PortalUsersPageClientProps) {
   const { confirm, ConfirmDialogElement } = useConfirmDialog()
   const t = useT()
   const router = useRouter()
@@ -521,17 +522,19 @@ export function PortalUsersPageClient({ portalOrigin }: PortalUsersPageClientPro
                 {t('customer_accounts.admin.portalInfo.openConfiguration', 'Open Configuration')}
               </Link>
             </Button>
-            <Button
-              type="button"
-              variant="outline"
-              size="sm"
-              asChild
-            >
-              <a href={buildPortalRootUrl(portalOrigin)} target="_blank" rel="noopener noreferrer">
-                <Globe className="size-4" />
-                {t('customer_accounts.admin.portalInfo.open', 'Open Portal')}
-              </a>
-            </Button>
+            {portalOrgSlug ? (
+              <Button
+                type="button"
+                variant="outline"
+                size="sm"
+                asChild
+              >
+                <a href={buildPortalRootUrl(portalOrigin, portalOrgSlug)} target="_blank" rel="noopener noreferrer">
+                  <Globe className="size-4" />
+                  {t('customer_accounts.admin.portalInfo.open', 'Open Portal')}
+                </a>
+              </Button>
+            ) : null}
           </div>
         </div>
       </div>
--- a/packages/core/src/modules/customer_accounts/backend/customer_accounts/users/page.tsx
+++ b/packages/core/src/modules/customer_accounts/backend/customer_accounts/users/page.tsx
@@ -1,14 +1,16 @@
 import { headers } from 'next/headers'
 import { Page, PageBody } from '@open-mercato/ui/backend/Page'
 import { resolvePortalRequestOrigin } from '../../../lib/portalUrl'
+import { resolveCurrentOrgPortalSlug } from '../../../lib/portalOrgSlug'
 import { PortalUsersPageClient } from './PortalUsersPageClient'
 
 export default async function CustomerAccountsPage() {
   const portalOrigin = resolvePortalRequestOrigin(await headers())
+  const portalOrgSlug = await resolveCurrentOrgPortalSlug()
   return (
     <Page>
       <PageBody className="space-y-4">
-        <PortalUsersPageClient portalOrigin={portalOrigin} />
+        <PortalUsersPageClient portalOrigin={portalOrigin} portalOrgSlug={portalOrgSlug} />
       </PageBody>
     </Page>
   )
--- a/packages/core/src/modules/customer_accounts/lib/__tests__/portalOrgSlug.test.ts
+++ b/packages/core/src/modules/customer_accounts/lib/__tests__/portalOrgSlug.test.ts
@@ -1,0 +1,77 @@
+/** @jest-environment node */
+
+const getAuthFromCookies = jest.fn()
+const createRequestContainer = jest.fn()
+const findOrganizationInTenant = jest.fn()
+
+jest.mock('@open-mercato/shared/lib/auth/server', () => ({
+  getAuthFromCookies: (...args: unknown[]) => getAuthFromCookies(...args),
+}))
+
+jest.mock('@open-mercato/shared/lib/di/container', () => ({
+  createRequestContainer: (...args: unknown[]) => createRequestContainer(...args),
+}))
+
+jest.mock('../organizationLookup', () => ({
+  findOrganizationInTenant: (...args: unknown[]) => findOrganizationInTenant(...args),
+}))
+
+import { resolveCurrentOrgPortalSlug } from '../portalOrgSlug'
+
+const em = { marker: 'entity-manager' }
+
+function containerResolving(name: string, value: unknown) {
+  return { resolve: jest.fn((requested: string) => (requested === name ? value : null)) }
+}
+
+describe('resolveCurrentOrgPortalSlug (regression for issue #5668)', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    createRequestContainer.mockResolvedValue(containerResolving('em', em))
+  })
+
+  it('returns the slug of the organization the session is scoped to', async () => {
+    getAuthFromCookies.mockResolvedValue({ orgId: 'org-1', tenantId: 'tenant-1' })
+    findOrganizationInTenant.mockResolvedValue({ slug: 'acme' })
+
+    await expect(resolveCurrentOrgPortalSlug()).resolves.toBe('acme')
+    // The (id, tenant_id) pair MUST travel together so the lookup cannot cross tenants.
+    expect(findOrganizationInTenant).toHaveBeenCalledWith(em, 'org-1', 'tenant-1')
+  })
+
+  it('trims surrounding whitespace off the stored slug', async () => {
+    getAuthFromCookies.mockResolvedValue({ orgId: 'org-1', tenantId: 'tenant-1' })
+    findOrganizationInTenant.mockResolvedValue({ slug: '  acme  ' })
+
+    await expect(resolveCurrentOrgPortalSlug()).resolves.toBe('acme')
+  })
+
+  it.each([
+    ['there is no session', null],
+    ['the session has no active organization', { orgId: null, tenantId: 'tenant-1' }],
+    ['the session has no tenant', { orgId: 'org-1', tenantId: null }],
+  ])('returns null when %s', async (_case, auth) => {
+    getAuthFromCookies.mockResolvedValue(auth)
+
+    await expect(resolveCurrentOrgPortalSlug()).resolves.toBeNull()
+    expect(findOrganizationInTenant).not.toHaveBeenCalled()
+  })
+
+  it.each([
+    ['the organization is missing', null],
+    ['the organization has no slug', { slug: null }],
+    ['the stored slug is blank', { slug: '   ' }],
+  ])('returns null when %s', async (_case, organization) => {
+    getAuthFromCookies.mockResolvedValue({ orgId: 'org-1', tenantId: 'tenant-1' })
+    findOrganizationInTenant.mockResolvedValue(organization)
+
+    await expect(resolveCurrentOrgPortalSlug()).resolves.toBeNull()
+  })
+
+  it('degrades to null instead of throwing when the lookup fails', async () => {
+    getAuthFromCookies.mockResolvedValue({ orgId: 'org-1', tenantId: 'tenant-1' })
+    findOrganizationInTenant.mockRejectedValue(new Error('[internal] database unavailable'))
+
+    await expect(resolveCurrentOrgPortalSlug()).resolves.toBeNull()
+  })
+})
--- a/packages/core/src/modules/customer_accounts/lib/__tests__/portalUrl.test.ts
+++ b/packages/core/src/modules/customer_accounts/lib/__tests__/portalUrl.test.ts
@@ -124,3 +124,33 @@ describe('portal URL builders', () => {
     expect(buildPortalRootUrl('https://demo.openmercato.com/')).toBe('https://demo.openmercato.com/portal')
   })
 })
+
+// Portal pages only exist under `frontend/[orgSlug]/portal/**`, so the admin
+// "Open Portal" action must carry the organization segment — without it the
+// link resolved to a 404 (#5668).
+describe('portal entry point for a resolved organization (regression for issue #5668)', () => {
+  it('places the organization slug ahead of the portal segment', () => {
+    expect(buildPortalRootUrl('https://demo.openmercato.com', 'acme')).toBe('https://demo.openmercato.com/acme/portal')
+    expect(buildPortalRootUrl('http://localhost:3000', 'acme')).toBe('http://localhost:3000/acme/portal')
+  })
+
+  it('never doubles the separator when the origin carries a trailing slash', () => {
+    expect(buildPortalRootUrl('https://demo.openmercato.com/', 'acme')).toBe('https://demo.openmercato.com/acme/portal')
+  })
+
+  it('degrades to a root-relative path when the origin is unresolvable', () => {
+    expect(buildPortalRootUrl('', 'acme')).toBe('/acme/portal')
+  })
+
+  it('keeps the legacy slugless shape when no slug is supplied', () => {
+    expect(buildPortalRootUrl('https://demo.openmercato.com', null)).toBe('https://demo.openmercato.com/portal')
+    expect(buildPortalRootUrl('https://demo.openmercato.com', undefined)).toBe('https://demo.openmercato.com/portal')
+    expect(buildPortalRootUrl('https://demo.openmercato.com', '   ')).toBe('https://demo.openmercato.com/portal')
+  })
+
+  it('escapes a slug that would otherwise alter the path structure', () => {
+    expect(buildPortalRootUrl('https://demo.openmercato.com', 'acme/../admin')).toBe(
+      'https://demo.openmercato.com/acme%2F..%2Fadmin/portal',
+    )
+  })
+})
--- a/packages/core/src/modules/customer_accounts/lib/portalOrgSlug.ts
+++ b/packages/core/src/modules/customer_accounts/lib/portalOrgSlug.ts
@@ -1,0 +1,40 @@
+// Server-only resolution of the signed-in admin's organization slug, used to
+// build the "Open Portal" link on the customer-accounts admin pages.
+//
+// Deliberately kept out of `portalUrl.ts`: that module is imported by client
+// components, and pulling the ORM in there would drag MikroORM into the browser
+// bundle. Keep every EntityManager import on this side of the boundary.
+
+import type { EntityManager } from '@mikro-orm/postgresql'
+import { createRequestContainer } from '@open-mercato/shared/lib/di/container'
+import { getAuthFromCookies } from '@open-mercato/shared/lib/auth/server'
+import { findOrganizationInTenant } from './organizationLookup'
+
+type EntityManagerResolver = {
+  resolve(name: 'em'): EntityManager
+}
+
+/**
+ * Resolve the slug of the organization the current admin session is scoped to,
+ * or null when there is no session, no active organization, or the organization
+ * has no slug yet. Callers render the portal link only for a non-null result —
+ * a slugless portal URL cannot resolve (#5668).
+ */
+export async function resolveCurrentOrgPortalSlug(): Promise<string | null> {
+  try {
+    const auth = await getAuthFromCookies()
+    const organizationId = auth?.orgId
+    const tenantId = auth?.tenantId
+    if (!organizationId || !tenantId) return null
+
+    const container = await createRequestContainer()
+    const em = (container as Awaited<ReturnType<typeof createRequestContainer>> & EntityManagerResolver).resolve('em')
+    const organization = await findOrganizationInTenant(em, organizationId, tenantId)
+    const slug = organization?.slug?.trim()
+    return slug && slug.length > 0 ? slug : null
+  } catch {
+    // The portal link is a convenience action — a lookup failure must not take
+    // the whole admin page down, so degrade to hiding the button.
+    return null
+  }
+}
--- a/packages/core/src/modules/customer_accounts/lib/portalUrl.ts
+++ b/packages/core/src/modules/customer_accounts/lib/portalUrl.ts
@@ -106,7 +106,18 @@ export function buildPortalUrlPattern(origin: string | null | undefined): string
   return `${normalizeOrigin(origin)}/${PORTAL_ORG_SLUG_PLACEHOLDER}/portal`
 }
 
-/** Portal entry point used by the "Open Portal" action on admin pages. */
-export function buildPortalRootUrl(origin: string | null | undefined): string {
-  return `${normalizeOrigin(origin)}/portal`
+/**
+ * Portal entry point used by the "Open Portal" action on admin pages.
+ *
+ * Portal pages are only ever mounted at `frontend/[orgSlug]/portal/**`, so a URL
+ * without the organization segment cannot resolve and renders a 404 (#5668).
+ * The slug stays optional to keep this signature backward compatible; callers
+ * that cannot resolve one should hide the action rather than link to the
+ * slugless fallback.
+ */
+export function buildPortalRootUrl(origin: string | null | undefined, orgSlug?: string | null): string {
+  const base = normalizeOrigin(origin)
+  const slug = typeof orgSlug === 'string' ? orgSlug.trim() : ''
+  if (!slug) return `${base}/portal`
+  return `${base}/${encodeURIComponent(slug)}/portal`
 }


adeptofvoltron:
## 🏷️ Label rationale

- 🔍 `review` — the PR remains in the review lane after the code-review pass. The review verdict was **approve** with no blockers or majors, but the reviewing automation authored this PR, so GitHub refused a formal self-approval (`Can not approve your own pull request`). Because an approving review from a second reviewer is still genuinely outstanding, the PR is deliberately **not** promoted to `merge-queue` — doing so would signal a readiness it has not reached.
- 🐛 `bug` — this corrects broken behavior reported in #5668: a link that resolved to a 404 rather than the customer portal. It adds no new capability.
- 📸 `needs-qa` — the change touches UI-rendering files (`PortalUsersPageClient.tsx`, `CustomerAccountsSettingsPageClient.tsx` and their route entrypoints), so the automated-verification exemption does not apply. The visible outcome — the button navigating to the organization's portal, and disappearing entirely for an organization without a slug — is worth exercising in a browser. Manual QA instructions are posted separately.
- ⚠️ `priority-medium` — a primary navigation action on two admin pages is dead for every user, which is more than cosmetic, but the portal itself remains reachable by typing the slug URL directly and no data is at risk.
- ✅ `risk-low` — the blast radius is one module's link construction. The exported builder keeps its previous output when no slug is passed, no database schema, API contract, or event changes, and the new lookup is a tenant-bound read that degrades to `null` on any failure. The full validation gate passes.

`qa-approved` is deliberately not applied — it is earned by manual QA, never by an authoring or reviewing skill. `qa` is likewise not applied: it means QA is in progress and belongs to the QA reviewer.

adeptofvoltron:
🤖 `om-open-pr` completed: PR opened against `develop` with the full SDLC label set and the validation gate green. Lock handed off to `om-auto-review-pr` — chain continues on this PR.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock — code review with autofix. Started: 2026-08-27T11:35:45Z.

adeptofvoltron:
> ℹ️ Posted as a comment rather than a formal approving review: GitHub rejects `addPullRequestReview` with "Can not approve your own pull request" because this PR was opened by the same automation account. The verdict below is the review outcome; the merge still needs an approving review from a second reviewer.

# 🔍 Code Review: fix(customer_accounts): open the portal at the organization slug (#5668)

## 🎯 Summary

This change fixes the admin "Open Portal" action, which linked to `${origin}/portal` and always landed on a 404. The reviewed scope is nine files in `packages/core/src/modules/customer_accounts`: the URL builder, a new server-only slug resolver, the two route entrypoints and their client components, and three test files.

The diagnosis is right and the fix is placed where the bug actually lives. Portal pages are only ever mounted under `frontend/[orgSlug]/portal/**`, and `packages/core/AGENTS.md` states `[orgSlug]` MUST be the first segment because portal auth and tenant resolution both assume it — so the slugless URL was structurally incapable of resolving. Rather than inventing a new convention, the change adopts the one the module already owns: `urlForCustomerOrg` in `lib/customerUrl.ts` has always built `${base}/${orgSlug}/portal${path}` for customer-facing email links, and these two admin buttons were simply the surfaces that bypassed it.

Three things are done particularly well. The slug lookup goes through `findOrganizationInTenant`, whose `(id, tenant_id)` pairing makes a cross-tenant read impossible, mirroring `api/portal/nav.ts` line for line. The resolver is deliberately placed in its own module rather than in `portalUrl.ts`, because the latter is imported by client components and an `EntityManager` import there would pull MikroORM into the browser bundle — and a new guard test enforces exactly that boundary so the mistake cannot be reintroduced. And the widened builder signature is genuinely additive: with no slug it returns the previous string byte for byte, so the original test expectations still hold unmodified.

The one judgement call worth naming explicitly is the behavior for an organization with no slug: the button is now omitted rather than rendered as a link that is known to 404. That is the right trade-off — a missing affordance is honest, a broken one is not — and the settings card already explains how slugs are generated and links to Manage Organizations. The findings below are all non-blocking polish.

## Verdict

✅ **approve** — no blocker or major findings. The full validation gate passes, the fix is minimal and confined to one module, the regression tests fail against the pre-fix implementation, and the widened export is backward compatible under the repository's own "new optional parameters may be added" rule for STABLE function signatures. The three minor findings are improvements the author can pick up; none of them affects correctness of the fix.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | 27/27 tasks. |
| `yarn generate` | ✅ PASS | All generators completed. Emits a pre-existing `ERR_IMPORT_ATTRIBUTE_MISSING` warning from the OpenAPI bundler that falls back to static extraction; it names no file in this diff and is unrelated to the change. |
| `yarn build:packages` | ✅ PASS | 27/27 tasks (second run, per the configured gate order). |
| `yarn i18n:check-sync` | ✅ PASS | All translation files in sync across 5 locales. |
| `yarn i18n:check-usage` | ✅ PASS | No new hardcoded user-facing strings; the change adds no new keys. |
| `yarn typecheck` | ✅ PASS | Clean. |
| `yarn test` | ✅ PASS | Every package green. Two turbo-parallelism flakes occurred on the run machine and were cleared by standalone re-runs: `@open-mercato/shared` (184 suites / 2015 tests) failed once with a crashed child-process worker producing no jest output, and once in `dynamicLoader.generatedCacheRecovery.test.ts` where the spawned loader died and left `appRoot` undefined in `afterEach`; `@open-mercato/cli` (95 suites / 1806 tests) failed once. All packages are verified green, though not all within a single turbo invocation. |
| `yarn build:app` | ✅ PASS | Next.js production build completed. |

## Findings

### 🔹 Minor

**`packages/core/src/modules/customer_accounts/lib/portalOrgSlug.ts:29`** — `createRequestContainer()` builds the entire DI graph (registrars, a forked EntityManager, `BasicQueryEngine`, `DefaultDataEngine`, a `CommandBus`, and the optimistic-lock guard service) in order to read a single `slug` column, and this now happens on every render of two admin pages. This is not a deviation from convention — `api/portal/nav.ts:42-46` does exactly the same thing, and `packages/core/AGENTS.md` mandates resolving services through DI rather than reaching for the ORM directly — but it is the first time the pattern lands in a *page render* path rather than an API route, where the container's cost is amortized differently. No change is required for this PR. If backend page latency later becomes a concern, the fix is a narrow shared accessor for a request-scoped `EntityManager` that does not construct the full cradle, which would benefit every server component that needs a single scoped read, not just this one.

**`packages/core/src/modules/customer_accounts/backend/customer_accounts/users/PortalUsersPageClient.tsx:505`** and **`.../settings/CustomerAccountsSettingsPageClient.tsx:47`** — the informational line still renders `buildPortalUrlPattern(portalOrigin)`, so an operator sees the literal `https://host/[org-slug]/portal` placeholder directly above a button that now navigates to the concrete `https://host/acme/portal`. The placeholder is defensible on the settings page, where the label reads "Portal URL pattern:" and documents the shape, but on the users page the label reads "Portal URL:" and a reader would reasonably expect their actual address. Consider passing the resolved slug into these lines too and falling back to the placeholder only when no slug is available — `buildPortalRootUrl(portalOrigin, portalOrgSlug)` already returns exactly the string to display.

**`packages/core/src/modules/customer_accounts/backend/customer_accounts/users/PortalUsersPageClient.tsx:525`** — when the organization has no slug the button vanishes with no explanation. On the settings page this is fine, because the adjacent `portal_access.slug_note` copy explains that slugs are auto-generated and links to Manage Organizations. The users-page info panel carries no equivalent note, so an admin in that state sees a control silently disappear and has no path to fix it. Consider rendering a short note in place of the button, reusing the existing "Manage Organizations" link target; that needs one new i18n key added across all five locale files, which is why it is called out rather than bundled into this fix.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — both `buildPortalUrlPattern` and `buildPortalRootUrl` keep their names and their module path.
- [x] No function signature changed in a breaking way — `buildPortalRootUrl` gains a second *optional* parameter, which `BACKWARD_COMPATIBILITY.md` § 3 (Function Signatures, STABLE) explicitly permits: "New optional parameters may be added." Existing single-argument callers receive the identical string.
- [x] No required type field removed or narrowed — `portalOrgSlug` is optional on both client component prop types, so any existing usage still typechecks.
- [x] No HTTP route URL removed or renamed; no method changed — no API routes are touched.
- [x] No field removed or retyped in an existing response shape — no serialized responses are touched.
- [x] No event or message name renamed or removed — no events are touched.
- [x] No CLI command or flag renamed or removed — no CLI surface is touched.
- [x] No database table or column renamed or removed — the change is read-only against `organizations`; no migration is needed and none is expected.
- [x] No config key renamed and no default changed silently — no config surface is touched.
- [x] Where a contract had to change: not applicable, since every change here is additive.

## 🧪 Test Coverage

Coverage is thorough and correctly targeted at the behavior that regressed.

`lib/__tests__/portalUrl.test.ts` adds a `#5668` block that locks in the slug-bearing URL for both an HTTPS origin and a localhost origin, trailing-slash normalization so the separator is never doubled, the root-relative `/acme/portal` degradation when the origin cannot be resolved, the unchanged slugless output for `null`, `undefined` and a whitespace-only slug, and slug escaping so a value such as `acme/../admin` cannot alter the path structure. I verified these assertions fail against the pre-fix implementation, which returned `https://demo.openmercato.com/portal` where the test now requires `https://demo.openmercato.com/acme/portal` — this is a genuine regression test, not a restatement of current behavior.

`lib/__tests__/portalOrgSlug.test.ts` covers the resolver's full decision surface: the happy path, asserting the lookup receives `(em, orgId, tenantId)` together so the tenant binding is guarded by a test rather than only by convention; slug trimming; every path that must yield `null` (no session, no active organization, no tenant, missing organization, and a null or blank slug); and graceful degradation when the lookup throws, which is what keeps a database blip from taking down the whole admin page.

`__tests__/portalUrlHydration.test.ts` extends the existing #5457 guard with a `#5668` section asserting that both entrypoints resolve the slug server-side, that both clients accept it as a prop and never call the slugless builder, and that `portalUrl.ts` stays free of `@mikro-orm` and `organizationLookup` imports. That last assertion is the valuable one: it is a structural guard that fails loudly if anyone later moves the ORM-backed lookup into the client-imported module.

One gap worth noting rather than blocking on: there is no test asserting that the button is actually absent from the rendered output when `portalOrgSlug` is `null`. The current guard is source-level (the client never calls the slugless builder), which catches the regression this PR fixes but would not catch a future refactor that reintroduces an unconditional render with an empty href. A small jsdom render test on `PortalUsersPageClient` asserting `queryByRole('link', { name: /Open Portal/ })` is null for a null slug, and non-null with a slug, would close it — `packages/core/src/modules/portal/__tests__/loginPage.test.tsx` is the nearest existing pattern for rendering these components with providers.

adeptofvoltron:

This PR passed code review and requires manual QA (`needs-qa`, no `skip-qa`). Note that it is **not** in `merge-queue`: the reviewing automation authored the PR, so GitHub refused a formal self-approval and the merge still needs an approving review from a second reviewer in addition to QA. The QA-approval gate holds the merge until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` (remove `review`, add `qa`), then run the routes below.

### P1 — "Open Portal" button on the Customer Users page

**Where to click**
- `/backend/customer_accounts/users`, the "Customer Portal" info panel in the top-right corner.

**What to verify**
- The `Open Portal` button is present when your active organization has a slug, and clicking it opens a new tab at `http://localhost:3000/{org-slug}/portal` — where `{org-slug}` matches the slug shown for that organization under `Directory → Organizations`. Before this change the same button opened `http://localhost:3000/portal` and rendered `404 – This page could not be found`, which is the exact symptom from #5668.
- The opened portal page actually renders the portal shell (login or dashboard depending on your session), not a 404.
- Switch your active organization to a second organization with a different slug, reload the page, and confirm the button now points at that organization's slug rather than the previous one.

**What can go wrong**
- The link still resolves to `/portal` and 404s — the fix did not take effect.
- The link carries the wrong organization's slug after an organization switch, which would mean the value is cached across renders rather than resolved per request.
- The slug appears percent-encoded in a way that breaks the route (the builder escapes the segment deliberately; a normal alphanumeric slug must pass through unchanged and resolve).
- Cross-organization leak: the slug shown must always belong to the organization you are currently scoped to, never another tenant's.

### P1 — "Open Portal" button on the Portal Settings page

**Where to click**
- `/backend/customer_accounts/settings`, the "Portal Access" card.

**What to verify**
- The `Open Portal` button opens the same organization-scoped URL as the users page, in a new tab.
- The "Portal URL pattern:" code block still shows the documentation placeholder `https://{host}/[org-slug]/portal`. This is unchanged by design — it documents the URL shape rather than your specific address — so a placeholder here is not a bug.
- The "Manage Organizations" link and the slug explanation note below the button still render.

**What can go wrong**
- The two pages disagree about the destination URL.
- The pattern code block accidentally shows a concrete slug or an empty segment such as `https://host//portal`.

### P1 — Organization without a slug

**Where to click**
- `Directory → Organizations` — find or create an organization whose slug is empty, make it your active organization, then revisit both routes above.

**What to verify**
- The `Open Portal` button is **absent** on both pages rather than present-but-broken. This is the intended new behavior: a link that is known to 404 is not rendered at all.
- The rest of both pages renders normally — the users DataTable, the settings cards, the demo-credentials table, and the `Open Configuration` button are all unaffected.
- On the settings page, the slug explanation note and `Manage Organizations` link are still visible, so there is a path to fix the missing slug. (On the users page there is no such note — that is a known minor gap raised in the review, not a QA failure.)

**What can go wrong**
- The button renders with an empty or malformed href instead of being hidden.
- The page errors or renders blank instead of degrading to a hidden button.

### P1 — Perceived performance and rendering

**Where to click**
- Cold-load `/backend/customer_accounts/users` and `/backend/customer_accounts/settings` with the browser cache disabled; capture a screenshot of each.

**What to verify**
- No visible flicker or layout shift of the portal address or the button area during hydration. Both pages resolve the address server-side specifically to avoid the hydration mismatch fixed earlier in #5457, so the first paint must already be correct.
- No React hydration warnings in the browser console on either route.
- The pages feel no slower to first paint than before — this change adds one scoped database read per render, so a clearly noticeable delay would be worth reporting.
- Check both routes at a mobile viewport (375px) to confirm the button column in the info panel still lays out correctly when it is present and collapses cleanly when it is hidden.

**What can go wrong**
- A hydration error appears in the console, or the address visibly corrects itself a frame after load.
- The info panel's right-hand button column collapses awkwardly when only the `Open Configuration` button remains.
- A noticeably slower first paint on either route.

### Pass/fail
- All routes pass → remove the `qa` label and add `qa-approved` (this clears the QA-approval gate); the PR still needs a second reviewer's approving review before it can merge.

adeptofvoltron:
🤖 `om-auto-review-pr` completed: ✅ approve — no blockers or majors; three minor findings listed for the author. Full validation gate green. Formal approval blocked by GitHub self-approval rules, so the PR keeps the `review` label pending a second reviewer. Lock retained — chain continues to UI verification. autofix: not triggered (verdict is approve; the autofix loop runs only after a changes-requested verdict).

adeptofvoltron:
🤖 `om-auto-qa-pr` taking over the chain lock — UI verification (evidence-only). Started: 2026-08-27T11:40:18Z.

adeptofvoltron:
## 📸 UI verification — PASS

Verified in a real Chromium session against a locally booted instance of this PR's head (`dc6175c`). The environment ran against a disposable Postgres database (`om_qa_5668`) seeded with the standard example tenant — organization **Acme Corp**, slug **`acme-corp`** — so the expected portal address is `http://localhost:3000/acme-corp/portal`. The developer's own database was not touched.

Every step below was executed; nothing is inferred. The full scenario, the throwaway Playwright specs, and the screenshots live in `.ai/qa/artifacts_5668/`.

### P1 — The reported symptom still reproduces at the old address

Navigating directly to `http://localhost:3000/portal` — the exact URL the button produced before this change — returns **HTTP 404** and renders "This page could not be found". This confirms the bug in #5668 was real and that the slugless address is structurally unreachable, rather than a reporter-side environment problem.

### P1 — Customer Users page (`/backend/customer_accounts/users`)

The `Open Portal` button renders in the Customer Portal panel and its `href` is exactly `http://localhost:3000/acme-corp/portal` — the organization slug is present. Following that link returns **HTTP 200** and renders the Acme Corp customer portal ("Welcome to the Customer Portal", with the Acme Corp branding in the header and footer), not a 404. This is the fix working end to end: the precise click path from the issue's reproduction steps now lands where the issue said it should.

### P1 — Portal Settings page (`/backend/customer_accounts/settings`)

The `Open Portal` button in the Portal Access card resolves to the same `http://localhost:3000/acme-corp/portal`, so the two admin surfaces agree. The "Portal URL pattern:" code block still displays `http://localhost:3000/[org-slug]/portal`; that is unchanged by design, since it documents the URL shape rather than the current organization's address.

### P1 — Organization without a slug

With the organization's slug nulled out in the QA database, both pages omit the `Open Portal` button entirely instead of rendering a link that is known to 404. The surrounding UI survives intact: `Open Configuration` still renders on the users page, `Manage Organizations` still renders on the settings page, and neither panel's layout breaks around the missing control. The slug was restored afterwards and re-verified.

### P1 — Rendering and mobile

No hydration error or console error appeared on either route, and the portal address did not flicker or self-correct after load — the #5457 server-side resolution contract still holds with the slug threaded through. At a 375px viewport the users page renders correctly with the button present.

### Observations for the author (non-blocking)

These match the minor findings from the code review and are visible in the screenshots:

- The info line still reads `Portal URL: http://localhost:3000/[org-slug]/portal` directly above a button that now navigates to the concrete `/acme-corp/portal`. On the users page the label says "Portal URL:", so a reader may reasonably expect their real address there.
- On the users page, the button disappears for a slugless organization with no explanation. The settings page has adjacent copy explaining slug generation plus a Manage Organizations link; the users panel has no equivalent, so an admin in that state gets no path forward.

### Evidence

**step-01-slugless-portal-404.png**



**step-02-users-open-portal.png**



**step-03-portal-renders.png**



**step-04-settings-open-portal.png**



**step-05-users-mobile.png**



**step-06-users-no-slug-hidden.png**



**step-07-settings-no-slug-hidden.png**

adeptofvoltron:
## 🧪 Follow-up: missing browser-level test

This PR ships strong unit and source-guard coverage but no integration/E2E test, so the behavior verified manually above is not locked in by CI. The scenario below is ready to implement via `om-integration-tests` as a follow-up; it was executed by hand during this QA pass, so it is known to pass against this branch.

**Suggested location:** `packages/core/src/modules/customer_accounts/__integration__/TC-PORTAL-009-open-portal-link.spec.ts`, alongside the existing `TC-PORTAL-*` specs.

**Fixtures:** the standard seeded tenant is sufficient. Read the active organization's slug the way `TC-AUTH-033.spec.ts` already does — `GET /api/directory/organizations?view=manage&ids={organizationId}&tenantId={tenantId}` — rather than hard-coding `acme-corp`, so the test survives a different seed.

**Cases to cover:**

1. **Link carries the organization slug.** Authenticate as an admin, open `/backend/customer_accounts/users`, and assert the `Open Portal` link's `href` equals `${baseURL}/${orgSlug}/portal`. This is the assertion that fails on the pre-fix code, where the href was `${baseURL}/portal`.
2. **The destination resolves.** Follow that href and assert the response status is below 400 and the body does not match `/This page could not be found/i`. This is the check that ties the test back to the 404 in #5668.
3. **Both admin surfaces agree.** Repeat case 1 for `/backend/customer_accounts/settings` and assert the two hrefs are identical.
4. **Slugless organization hides the action.** With an organization whose slug is null, assert `getByRole('link', { name: /Open Portal/i })` has count 0 on both routes while the sibling controls (`Open Configuration`, `Manage Organizations`) remain visible. Create and clean up this fixture within the test rather than mutating seeded data, per the repository's self-contained-integration-test rule.

**Gotcha worth encoding:** a fixed "Demo Environment" / cookie banner overlays the login form and intercepts the submit hit-test, so driving the rendered login form times out. Authenticate through the context request API instead — `page.context().request.post('/api/auth/login', { form: { email, password } })`, form-encoded, since the endpoint rejects a JSON body — then navigate. The specs used for this QA pass are attached in the run artifacts and can be lifted almost verbatim.

adeptofvoltron:
🤖 `om-auto-qa-pr` completed: ✅ PASS. Verified in Chromium against a locally booted instance of this PR's head on a disposable database seeded with the Acme Corp tenant. The old `/portal` address still 404s, both admin surfaces now link to `/acme-corp/portal`, that URL renders the real customer portal, and a slugless organization correctly gets no button. Evidence and the follow-up integration-test scenario are posted above. No labels changed (evidence-only mode) — `needs-qa` stands and `qa-approved` remains for a human QA reviewer. Lock retained — chain continues.

adeptofvoltron:
🤖 `om-auto-fix-issue` run complete: the fix for #5668 is implemented, tested, reviewed (✅ approve, no blockers or majors), UI-verified in a browser with screenshot evidence, and CI is green across all checks. Lock released — the PR is free for anyone to pick up. Two things still gate the merge, both requiring a human: an approving review from a second reviewer (GitHub blocks self-approval by the authoring automation) and `qa-approved` from a QA reviewer, since the PR carries `needs-qa`.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T12:09:24Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# 🔍 Code Review: fix(customer_accounts): open the portal at the organization slug (#5698)

## 🎯 Summary

Independent second review by `@pkarw`. The PR was authored and self-reviewed by the `@adeptofvoltron` automation account, which could not submit a formal approving review on its own PR — this run supplies the approving review that the merge actually needs.

The reviewed scope is nine files in `packages/core/src/modules/customer_accounts`: the URL builder (`lib/portalUrl.ts`), a new server-side slug resolver (`lib/portalOrgSlug.ts`), the two backend route entrypoints and their client components, and three test files. Nothing outside that module is touched, and no schema, API route, event, or CLI surface is involved.

I re-derived the diagnosis rather than accepting it. `buildPortalRootUrl` returned `${origin}/portal`, and I confirmed by enumerating the route tree that **every** portal page in the repository lives under `frontend/[orgSlug]/portal/**` — `packages/core/src/modules/portal/frontend/[orgSlug]/portal/{page,login,signup,dashboard,profile,verify,invite,reset-password}` plus `warranty_claims`' claim pages — so the slugless URL was structurally incapable of matching a route. `packages/core/AGENTS.md` § Portal Pages states `[orgSlug]` MUST be the first segment because portal auth and tenant resolution both assume it. The bug is real and the fix is at the right layer.

The change adopts the convention the module already owns rather than inventing one: `urlForCustomerOrg` in `lib/customerUrl.ts` has always built `${base}/${orgSlug}/portal${path}` for customer-facing email links. I verified by grep that `buildPortalRootUrl` has exactly two call sites repo-wide and both are fixed here; `PortalUserDetailPageClient.tsx` — the third surface listed in the existing #5457 hydration guard — only renders `buildPortalUrlPattern`, the documentation placeholder, so it is correctly out of scope.

Three aspects deserve explicit credit:

- **Tenant safety.** The lookup goes through the existing `findOrganizationInTenant(em, organizationId, tenantId)` helper, whose `(id, tenant_id)` pairing plus `deleted_at IS NULL` makes a cross-tenant read impossible. That helper exists because of the recorded lesson "Keep raw SQL out of API route handlers" (`.ai/lessons.md`), and reusing it rather than hand-rolling a query is exactly the right call.
- **The client/server boundary.** Placing the resolver in its own module instead of `portalUrl.ts` keeps MikroORM out of the browser bundle, and a new source-guard test enforces that boundary so the mistake cannot be reintroduced. I checked the transitive guarantee too: `@open-mercato/shared/lib/di/container` calls `require('server-only')` at module scope, so any client component that imported `portalOrgSlug.ts` would fail at build time rather than silently ship the ORM.
- **Backward compatibility.** `buildPortalRootUrl` gains a second *optional* parameter and returns the previous string byte for byte when it is absent, so the pre-existing test expectations still hold unmodified.

I also checked two failure modes the diff does not discuss. First, **custom domains**: `urlForCustomerOrg` prefers an org's active branded hostname, and the new button does not — but I confirmed the `(frontend)` layout resolves the org from the URL slug (`em.findOne(Organization, { id, slug, deletedAt: null })`, per `portal-org-binding.test.tsx`) with no host-binding guard and no middleware redirect, so `{platformOrigin}/{slug}/portal` remains a valid entry point even for an org that also has a custom domain. The button is not broken there; it just opens the platform address rather than the branded one. Second, **duplicate work**: `origin/develop` still carries the slugless `buildPortalRootUrl`, so this is not already-merged work, and the head merges into the current `develop` tip cleanly.

## Verdict

✅ **approve** — no blocker, major, or minor findings. Every required check is green against this exact head, the fix is minimal and confined to one module, the regression tests are concrete rather than vacuous, and the widened export is additive under `BACKWARD_COMPATIBILITY.md` § 3 ("New optional parameters may be added"). The items listed below are nits and follow-ups; none of them affects the correctness of the fix, and none of them is a reason to withhold approval.

## 🧪 Validation Gate

**Validation source: GitHub checks** (per this repository's checks-first review rule — CI already ran the full gate, so it was not reproduced locally).

I verified the checks belong to this PR's head rather than an earlier push: `repos/.../commits/dc6175cddbe45e227f694c77513542277283eb1a/check-runs` returns the results below, and both workflow runs (`CI for Develop&Main`, `Mutation tests`) report `head_sha = dc6175c`.

| Check | Status |
|-------|--------|
| `prepare` (`yarn build:packages`, `yarn generate`, `yarn build:app`) | ✅ success |
| `test` (`yarn turbo run typecheck/test --filter=[origin/develop]...`, `i18n:check-usage`, `agents:check-budget`, `test:scripts`, `test:repo-wide-guards`) | ✅ success |
| `lint` | ✅ success |
| `ds-lint` | ✅ success |
| `audit-scope` | ✅ success |
| `scope` | ✅ success |
| `ephemeral-integration (none)` | ✅ success |
| `license/cla` | ✅ success |

Two notes on reading that table. `ephemeral-integration (none)` is the **affected-only shard name**, not a skipped job — it ran and passed. And the `typecheck`/`test` steps use `--filter=[origin/develop]...`, which selects changed packages and their dependents; since the change *is* in `packages/core`, the new suites were genuinely executed rather than filtered out. The `skipped` entries (`audit`, `comment`, `docker-build`, `merge-coverage`, `mutate`, `[code]smith`) are conditional jobs that do not gate this PR.

The PR body's own local-gate report matches: all eight configured commands green, with two turbo-parallelism flakes in `@open-mercato/shared` and `@open-mercato/cli` cleared by standalone re-runs. That is consistent with CI.

## Findings

No blocker, major, or minor findings.

### 🔸 Nit

**`packages/core/src/modules/customer_accounts/lib/portalOrgSlug.ts:24`** — `createRequestContainer()` constructs the full DI cradle (registrars, a forked `EntityManager`, `BasicQueryEngine`, `DefaultDataEngine`, a `CommandBus`, the optimistic-lock guard service) to read one `slug` column, and this now runs on every render of two admin pages. This is the sanctioned pattern — root `AGENTS.md` mandates resolving services through DI, and `api/portal/nav.ts` and `lib/customerUrl.ts` do the same — but this is the first time it lands in a *page render* path rather than an API route. No change is warranted in this PR. If backend page latency ever becomes a concern, the right fix is a shared accessor for a request-scoped `EntityManager` that skips the full cradle, which would benefit every server component doing a single scoped read.

**`packages/core/src/modules/customer_accounts/lib/portalOrgSlug.ts:36`** — the bare `catch { return null }` swallows every failure with no diagnostic, so a DI misconfiguration and a transient database blip are indistinguishable from "this organization has no slug": the button just disappears. The degradation itself is right — a convenience link must not take the admin page down — and the immediate neighbour `resolveOrgSlug` in `customerUrl.ts` swallows identically, so this matches local convention rather than violating it. Worth a `createLogger(...).warn()` when the repo's structured-logging facade rollout (`.ai/specs/2026-07-02-structured-logging-facade.md`) reaches this module.

**`packages/core/src/modules/customer_accounts/lib/portalUrl.ts:121`** — the slugless branch still returns the exact `${origin}/portal` string this PR exists to eliminate. Keeping it is the correct backward-compatibility call for an exported symbol under `BACKWARD_COMPATIBILITY.md`, and the JSDoc tells callers to hide the action instead. The residual risk is that the source-guard test only pins the two known clients, so a *new* caller could reintroduce the 404 without any test failing. A `@deprecated` note on the no-slug path, or a widening of the guard test to assert no file in the module calls the one-argument form, would close that off cheaply.

**`packages/core/src/modules/customer_accounts/backend/customer_accounts/users/PortalUsersPageClient.tsx:525`** — when the organization has no slug the button vanishes with no explanation. On the settings page the adjacent `portal_access.slug_note` copy and the Manage Organizations link cover this; the users-page info panel has no equivalent. It still renders `Portal URL: {origin}/[org-slug]/portal` directly above the gap, so the state is not completely opaque, and hiding a control is honest where rendering a known-404 link was not. A short note in place of the button would need one new i18n key across all five locale files, which is why this is a suggestion rather than a request.

This matters more often than it looks: `resolveOrganizationOverride` in `packages/shared/src/lib/auth/server.ts:139` maps the `__all__` organization cookie to `orgId: null`, so a superadmin browsing in "All organizations" mode hits this path and loses the button entirely. That is still an improvement over a link that 404s, but it is the most likely way an operator meets the degraded state.

**Test coverage — one gap, source-level only** — there is no render test asserting the button is actually absent when `portalOrgSlug` is `null`. The current guard is textual (the client never calls the slugless builder), which catches this regression but would not catch a future refactor that renders the link unconditionally with an empty href. A jsdom test on `PortalUsersPageClient` asserting `queryByRole('link', { name: /Open Portal/ })` is null for a null slug and non-null with one would close it.

### 📋 Inherited feedback — accounted for

The three minor findings from the automation's own review pass (`#issuecomment` on this PR) were re-evaluated independently and are all reproduced above at nit severity, or deliberately declined:

- *DI container cost per page render* — reproduced above as a nit; the original review itself concludes "no change is required for this PR", and I agree.
- *Users page shows the `[org-slug]` placeholder above a button that now resolves concretely* — **declined for this PR.** That line (`buildPortalUrlPattern`) is pre-existing and untouched by this diff; changing it here would widen the scope of a targeted bug fix. It is a good follow-up.
- *Button vanishes with no note on the users page* — reproduced above as a nit, with the superadmin `__all__` scenario added.

The **"🧪 Follow-up: missing browser-level test"** comment from the QA pass is a genuine outstanding item: this PR ships no integration/E2E coverage, so the manually verified behavior is not locked in by CI. I searched the tracker and **no follow-up issue has actually been filed** — the scenario exists only as a PR comment, which will be lost once this merges. That comment contains a complete, already-executed spec plan (`TC-PORTAL-009`, four cases, plus the cookie-banner authentication gotcha); it should be turned into an issue before merge. This is not a blocker for this PR — the repo's integration-coverage rule in `.ai/qa/AGENTS.md` is written for new features, and this is a bug fix that ships unit plus structural coverage and passed manual QA with screenshot evidence.

## 💥 Breaking Changes

- [x] No exported symbol removed or renamed — `buildPortalUrlPattern` and `buildPortalRootUrl` keep their names and module path.
- [x] No breaking signature change — `buildPortalRootUrl` gains a second **optional** parameter, explicitly permitted for STABLE function signatures by `BACKWARD_COMPATIBILITY.md` § 3. Single-argument callers receive the identical string; I confirmed the pre-existing assertions at `portalUrl.test.ts:114` and `:119` still pass unmodified.
- [x] No required type field removed or narrowed — `portalOrgSlug` is optional on both client prop types, so existing usages still typecheck.
- [x] No API route, HTTP method, response field, event id, widget spot id, DI key, ACL feature, notification id, or CLI command touched.
- [x] No database table or column changed — the access is a read against `organizations`, so no migration is expected and none is present.
- [x] No auto-discovery or generated-file contract affected.
- [x] No config key renamed and no default silently changed.
- [x] No `apps/mercato/src/app/**` or `apps/mercato/src/i18n/**` file touched, so the create-app template-sync obligation does not apply.

## 🧪 Test Coverage

Coverage is thorough and aimed at the behavior that regressed.

`lib/__tests__/portalUrl.test.ts` adds a `#5668` block pinning the slug-bearing URL for both an HTTPS and a localhost origin, trailing-slash normalization, the root-relative `/acme/portal` degradation when the origin cannot be resolved, the unchanged slugless output for `null` / `undefined` / whitespace, and slug escaping so `acme/../admin` cannot alter the path structure. These are concrete expected values, not restatements — the pre-fix implementation returns `https://demo.openmercato.com/portal` where the test now requires `https://demo.openmercato.com/acme/portal`, so they fail against the old code.

`lib/__tests__/portalOrgSlug.test.ts` covers the resolver's full decision surface and, importantly, is not vacuous: the happy path asserts `findOrganizationInTenant` was called with `(em, 'org-1', 'tenant-1')` **together**, so the tenant binding is pinned by a test rather than only by convention, and each `null` case asserts the lookup was *not* called at all rather than merely checking the return value. The throwing-lookup case pins the graceful degradation that keeps a database blip from taking the admin page down.

`__tests__/portalUrlHydration.test.ts` extends the #5457 guard with a `#5668` section asserting both entrypoints resolve the slug server-side, both clients take it as a prop, neither calls the slugless builder, and `portalUrl.ts` stays free of `@mikro-orm` and `organizationLookup` imports. That last assertion is the durable one — it fails loudly if anyone later moves the ORM-backed lookup into the client-imported module.

The gap is the missing render-level and browser-level coverage described under Findings.

## 🏷️ Labels

`review` → `merge-queue` (approved). `bug`, `priority-medium`, `risk-low` and `needs-qa` are preserved; `screenshots` added, since the QA comment carries seven embedded captures.

**This PR is not clear to merge yet.** It carries `needs-qa` without `qa-approved`, and this repository's QA-approval gate is a hard merge block. Manual QA has in fact already been performed and passed — a full Chromium run against this head on a disposable database, with screenshots — but the `qa-approved` / `qa-self-verified` labels were never applied. Applying them from a diff read is forbidden, and I did not run that QA myself, so a maintainer needs to confirm the existing evidence and apply the labels to unblock the merge.

The standard P0/P1/P2 manual-QA instructions comment is deliberately **not** re-posted: an equivalent instructions comment and a completed QA PASS report for this same diff are already on the PR, and duplicating them would add noise without adding signal.

pkarw:
## 🏷️ Label rationale

- 🚀 `merge-queue` — replaces `review`. An independent approving review from a second reviewer (`@pkarw`) is now on the PR, with no blocker, major, or minor findings. The earlier verdict comment from the authoring automation could not be a formal review because GitHub rejects self-approval; that gap is now closed.
- 🐛 `bug` — unchanged. The change fixes a broken "Open Portal" link that always resolved to a 404.
- 🧪 `needs-qa` — **kept**, and it is the one thing still blocking the merge. Manual QA has already been executed and passed against this exact head (`dc6175c`) with seven screenshots, but `qa-approved` was never applied. Applying it from a diff read is forbidden, and this reviewer did not run that QA, so a maintainer must confirm the existing evidence and add `qa-approved` (plus `qa-self-verified`, since the QA was self-run) before this can merge.
- 📸 `screenshots` — added. The UI verification comment carries seven embedded captures of both admin surfaces and the resolved portal address.
- 🟡 `priority-medium` — unchanged. A visibly broken primary action on two admin pages, with no data at risk and no workaround needed beyond typing the URL by hand.
- 🟢 `risk-low` — unchanged. Nine files in one module, no schema, API, event, or contract surface touched; the widened export is additive and the no-slug call returns the previous string byte for byte.

pkarw:
🤖 `om-auto-review-pr` completed: ✅ APPROVED. Lock released.

Independent second review by `@pkarw` — no blocker, major, or minor findings. Validation source: GitHub checks (every required check green against head `dc6175c`; verified by head SHA, not by check name alone). Labels moved `review` → `merge-queue`, `screenshots` added. Autofix: skipped (not my PR — re-run with `--autofix` to fix findings here); nothing actionable was outstanding in any case.

**Still blocking the merge:** `needs-qa` without `qa-approved`. The QA itself has already passed on this head with screenshot evidence — a maintainer needs to confirm that evidence and apply `qa-approved` + `qa-self-verified`.

**Untracked follow-up:** the "missing browser-level test" scenario (`TC-PORTAL-009`) exists only as a PR comment and has no issue behind it; it will be lost on merge unless filed.

Issue #5672, opened 2026-08-26
feat: Enable column sorting in the customer users table


RafalDev4You:
## Proposed solution

Enable column sorting in the customer users table available at `/backend/customer_accounts/users`.

Users should be able to click a column header to sort the data in ascending or descending order. The current sort direction should be visually indicated in the column header.

Sorting should be available for applicable columns:

- Name
- Email
- Verified
- Status
- Last Login
- Role
- Created

Sorting should work together with the existing search, filters, organization selection, pagination, and saved table views. If sorting is handled server-side, the customer users API should accept the selected sort field and direction without changing the existing response contract.

## Specification

**Does a specification exist for this feature?**
- [ ] Yes
- [x] No
- [ ] Not sure

A separate specification is not required because this is a small, localized enhancement to an existing table with no domain-model changes.

## Additional context

The table is implemented in `packages/core/src/modules/customer_accounts/backend/customer_accounts/users/page.tsx`.

The shared table component is located at `packages/ui/src/backend/DataTable.tsx` and already supports sorting. The existing sorting behavior should be reused, and a regression test should verify ascending and descending sorting for at least the `Name` and `Created` columns.

adeptofvoltron:
🤖 Claiming this issue — starting `om-auto-write-spec` run (via `om-auto-fix-issue` feature route). Started: 2026-08-27T11:31:00Z.

adeptofvoltron:
🤖 `om-auto-write-spec` — Open Questions answered with autonomous defaults

This spec had open questions and the run is autonomous, so I applied conservative
defaults to keep moving. **Please review and override before merge if any is wrong.**

| # | Question | Applied default | Why | Confirm? |
|---|----------|-----------------|-----|----------|
| Q1 | How do the encrypted `Name` and `Email` columns sort, when a SQL `ORDER BY` would order ciphertext? | Hybrid: plaintext columns sort natively; `Name`/`Email` take a bounded decrypt-then-sort path capped by `CUSTOMER_USERS_ENCRYPTED_SORT_MAX` (default 5000), degrading visibly past the cap. | The FR names `Name` as a mandatory regression-tested column, so dropping it would not ship the ask. A persisted plaintext sort column would defeat the encryption it indexes, and a page-only client sort would silently misrepresent ordering across pages. The cap is one env constant, so the trade-off is reversible without a schema or contract change. | ⚠ needs human |
[...]

PR #5699, opened 2026-08-27
docs(specs): customer users table column sorting

adeptofvoltron:
Refs #5672
Source doc: .ai/specs/2026-08-27-customer-users-table-column-sorting.md
Status: draft — spec-only design PR, promoted and approved by a human reviewer

## 🎯 Goal
- Specify server-side column sorting for the customer users table at `/backend/customer_accounts/users`, including the columns whose values are encrypted at rest.

## What Changed
- Spec document at `.ai/specs/2026-08-27-customer-users-table-column-sorting.md`
- Mockups: skipped — no prepared test environment (`.ai/qa/test-env.json` is absent, so the app cannot be booted for current-state screenshots), and the spec adds no new screen: it enables `DataTable`'s existing sort chrome plus one inline notice, which the mockup guidance explicitly excludes as standard component behavior.

## 🔍 Revised after specification review

The first version of this spec designed a module-local bounded decrypt-then-sort path for the encrypted `Name` and `Email` columns, and gated its own merge on a human confirming that trade-off. Review found that the repository **already ships that path** — `@open-mercato/shared/lib/query/encrypted-sort` plus `mapWithConcurrency`, with two routes already composing them — and the spec has been rewritten to reuse it. All ten review findings (3 blockers, 3 majors, 2 minors, 2 nits) are addressed; see the re-review comment below for the finding-by-finding resolution.

The headline changes:

- The encrypted branch composes `resolveEncryptedSortMaxRows()` + `mapWithConcurrency` + `sortRowsInMemory`, following `customers/api/labels/route.ts:86-110`, and writes no comparator, tiebreak or cap parsing of its own.
- The proposed `CUSTOMER_USERS_ENCRYPTED_SORT_MAX` env var is gone; the existing platform knob `OM_ENCRYPTED_SORT_MAX_ROWS` governs the cap, so one setting covers every encrypted sort in a deployment.
- The invented `sortDegraded` response field is replaced by the documented `meta.encryptedSortRowCapWarning` contract.
- The blanket nulls-last rule became a per-column table, so the `Verified` column can lead with either the verified or the unverified group.
- The decrypt bound (`DECRYPT_CONCURRENCY = 8`) and the degraded-path structured warning log are now specified, matching the existing call sites.

## ⚠️ The previous merge gate is resolved

This PR previously carried `⚠ NEEDS HUMAN CONFIRMATION` on Q1 — "should we build a bounded decrypt-then-sort path, and at what cap". Reusing the platform primitives dissolves that question rather than answering it: there is no new env var to name, no new default to pick, no new response semantic to coin, and the residual performance profile is identical to the one the product already accepts on customer labels and interactions, tunable by the same knob.

The marker is therefore removed and Q1 is resolved autonomously, with the reasoning written into the spec under **"Why the ⚠ marker was removed"**. This is reversible and open to override — a reviewer who wants a module-local ceiling tighter than the platform value can say so, and it is one paragraph in API Contracts plus one implementation step.

**The PR stays a draft regardless**, because a spec-only design PR is promoted and approved by a human, not by automation.

## 💥 Breaking Changes
- None — design only

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5699"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790422521&installation_model_id=432243&pr_number=5699&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5699&signature=1288688fe9b5fca6b016c577112860c9366e637737b0e94a1b68f039feba65b9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:
🤖 `om-auto-write-spec` — 🏷️ label rationale

- 🔍 `review` — the spec document is complete and ready for specification review; the draft state reflects the open assumption below, not unfinished writing.
- 📚 `documentation` — this PR lands a design document under `.ai/specs/` and changes no runtime code.
- ⏭️ `skip-qa` — there is no runtime behavior to exercise in a browser, so the manual QA gate does not apply to a design-only change.
- 🔽 `priority-low` — the underlying request is an ergonomics improvement on an admin table that already has search and filters as a workaround, so nothing is broken or blocked while this waits.
- 🟢 `risk-low` — merging this PR changes only a markdown file; the blast radius of the design it describes is assessed separately in the spec's Risks & Impact Review section.

adeptofvoltron:
🤖 `om-auto-write-spec` — Open Questions answered with autonomous defaults

This spec had open questions and the run is autonomous, so I applied conservative
defaults to keep moving. **Please review and override before merge if any is wrong.**

| # | Question | Applied default | Why | Confirm? |
|---|----------|-----------------|-----|----------|
| Q1 | How do the encrypted `Name` and `Email` columns sort, when a SQL `ORDER BY` would order ciphertext? | Hybrid: plaintext columns sort natively; `Name`/`Email` take a bounded decrypt-then-sort path capped by `CUSTOMER_USERS_ENCRYPTED_SORT_MAX` (default 5000), degrading visibly past the cap. | The FR names `Name` as a mandatory regression-tested column, so dropping it would not ship the ask. A persisted plaintext sort column would defeat the encryption it indexes, and a page-only client sort would silently misrepresent ordering across pages. The cap is one env constant, so the trade-off is reversible without a schema or contract change. | ⚠ needs human |
| Q2 | Should the `Role` column be sortable? | No — deferred; `Role` renders without a sort affordance. | `roles` is a multi-valued junction relation with no well-defined single sort key, so "sort by role" is ambiguous as soon as a user holds two roles. The FR says "applicable columns", and a multi-valued relation is not one. | ok |
| Q3 | Should the active sort be mirrored into the URL, as the CRM People/Companies lists do? | No — sort lives in component state and is persisted by the existing saved-views (perspectives) mechanism. | This is the smallest scope that still satisfies the FR's "must work with saved table views" requirement, because `DataTable` already restores and re-emits sorting when a view is applied. Deep-linkable sort remains a purely additive follow-up. | ok |
| Q4 | Should the portal-facing users route accept the same sort params? | No — the admin route only. | The FR scopes itself to the backoffice table, and widening to a second route would double the contract surface and test matrix for no stated need. | ok |
| Q5 | What is the ordering when no sort is requested? | Unchanged `createdAt DESC`, now with an explicit `id` tiebreak. | This preserves the current response contract for every existing caller. The tiebreak is a latent-bug fix: `createdAt` alone can drop or duplicate rows across page boundaries when timestamps collide, which bulk-created users routinely do. | ok |

To change any answer: reply here (or edit the spec's "Resolved assumptions"
section) and re-run `om-auto-write-spec` — or implement with the corrected spec
via `om-auto-implement-spec .ai/specs/2026-08-27-customer-users-table-column-sorting.md`.

adeptofvoltron:
## 🤖 `om-auto-write-spec` — run summary

**Spec:** `.ai/specs/2026-08-27-customer-users-table-column-sorting.md`
**Branch:** `spec/customer-users-table-column-sorting`
**Final status:** draft — merge gated on the ⚠ assumption in Q1

The FR describes this as "a small, localized enhancement to an existing table with no domain-model changes", and the UI half of that is accurate: `DataTable` already implements manual sorting, and the saved-views mechanism already persists a `sorting` array and re-emits it through `onSortingChange`, so the page only needs to become a controlled sorting host.

The server half is not small. `packages/core/src/modules/customer_accounts/encryption.ts` declares `email` and `display_name` as encrypted fields on `customer_accounts:customer_user`, and `isTenantDataEncryptionEnabled()` defaults to **enabled** when `TENANT_DATA_ENCRYPTION` is unset. `findAndCountWithDecryption` forwards `orderBy` straight to the database and decrypts only afterwards, so `ORDER BY display_name` would order ciphertext — deterministic and meaningless. This is the same wall the route's existing search already hit, which is why partial matches route through `search_tokens`; sorting has no token-table equivalent because hashes do not preserve order.

The spec therefore splits the six sortable columns into a native-SQL path (Verified, Status, Last Login, Created) and a bounded decrypt-then-sort path (Name, Email), and adds a deterministic `id` tiebreak that also closes a latent pagination bug in the current default ordering.

### 📝 Assumptions applied
- Five Open Questions were resolved with autonomous defaults; the full table is posted as a separate comment on this PR and on issue #5672 for override. One — Q1, the encrypted-column sorting strategy — carries `⚠ NEEDS HUMAN CONFIRMATION`, because the bounded decrypt-then-sort path is a genuine per-request performance trade-off on large tenants rather than a purely mechanical choice. That marker is what keeps this PR a draft.

### 📸 Visual evidence
- Skipped, for two independent reasons. There is no prepared test environment (`.ai/qa/test-env.json` is absent), so the app could not be booted for current-state screenshots. Separately, the feature introduces no new screen — it enables the shared `DataTable`'s existing sort chrome plus one inline notice — and the mockup guidance explicitly excludes standard component behavior the spec itself calls standard.

### 🔁 Hand-off
Implement with: `om-auto-implement-spec .ai/specs/2026-08-27-customer-users-table-column-sorting.md` (or `om-auto-continue-pr 5699` for spec-only continuation). This PR stays design-only; the implementation ships on its own PR referencing it.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-09-02T12:22:47Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron (commented):
# 🔍 Specification review: Customer Users Table Column Sorting

Reviewed at head `spec/customer-users-table-column-sorting`, spec-only PR (1 file, `.ai/specs/2026-08-27-customer-users-table-column-sorting.md`). Base `develop` is mergeable, CI is green (`lint`, `ds-lint`, `test`, `audit-scope`, `documents-multi-instance`, `prepare`, `scope` all pass; the heavier jobs skip on a docs-only diff).

**Verdict: changes requested.** Three blockers, three majors.

The grounding work in this spec is genuinely good — I checked every code citation it makes and all of them are accurate (`users.ts:156`, `users.ts:73`, `PortalUsersPageClient.tsx:538`, `encryption.ts:7-8`, `DataTable.tsx:1893`/`1930`/`3392`, `perspectives/data/validators.ts:12-15`, `customers/.../people/page.tsx:924-926`, and `customers/backend/customers/__tests__/listSorting.test.ts` exists). The problem statement is correct, and Q2–Q5 are well-argued.

What blocks it is that the one hard problem — the ⚠ Q1 decision this PR is gated on — is being decided against an incomplete option set. **This repository already has a shipped, documented platform solution for sorting on encrypted columns**, and the spec does not mention it anywhere. The human sign-off being requested should not happen until Q1 is re-framed against that prior art.

## 💥 What can go wrong

### 🔴 Blocker 1 — the design reinvents an existing platform primitive

*`.ai/specs/2026-08-27-customer-users-table-column-sorting.md:15`, `:40-46`, `:61`, `:161`*

`@open-mercato/shared/lib/query/encrypted-sort` already exists and provides exactly the pieces this spec proposes to build:

- `resolveEncryptedSortMaxRows()` — the row cap (`encrypted-sort.ts:32`)
- `resolveEncryptedSortFields()` — which fields on an entity are encrypted (`:41`)
- `sortRowsInMemory()` — the in-memory comparator (`:85`), already locale-aware (`localeCompare` with `sensitivity: 'base', numeric: true`) and already applying an `id` tiebreak

alongside `mapWithConcurrency()` from `@open-mercato/shared/lib/query/bounded-decrypt`.

There is a near-identical precedent that does not go through the query engine either — `packages/core/src/modules/customers/api/labels/route.ts:86-110` sorts an encrypted `label` column off a direct `em.find`, capped by `resolveEncryptedSortMaxRows()`, decrypted through `mapWithConcurrency(..., DECRYPT_CONCURRENCY, ...)`, then ordered by `sortRowsInMemory`. `customers/api/interactions/encryptedSortPage.ts` does the same for a cursor-paged list.

The spec's "three alternatives were considered and rejected" (`:40-44`) lists persisted plaintext column, client-side page sort, and dropping the columns — but never "use the platform helpers the way `labels/route.ts` does". Since Q1 is the single assumption keeping this PR a draft, the human is being asked to sign off a build-it-ourselves trade-off without being shown the reuse option.

**Fix:** rewrite Architecture unit 2 (`:56-63`) so the encrypted branch composes `resolveEncryptedSortMaxRows` + `mapWithConcurrency` + `sortRowsInMemory`, citing `labels/route.ts` as the pattern. Then re-frame Q1 from "should we build a bounded decrypt-then-sort path" to "should this table adopt the platform encrypted-sort path, and at what cap" — a much smaller question that may not need the ⚠ marker at all.

### 🔴 Blocker 2 — a second cap knob that silently conflicts with the platform one

*`:15`, `:109`, `:143`, `:161`*

The spec introduces `CUSTOMER_USERS_ENCRYPTED_SORT_MAX` with a default of `5000`. The platform-wide knob for the identical concern is `OM_ENCRYPTED_SORT_MAX_ROWS`, documented at `apps/docs/docs/framework/database/hybrid-query-engine.mdx:62` and read by the query engine, the query-index engine, `labels/route.ts` and `interactions/route.ts`.

Two consequences, both bad for operators:

1. An operator who sets `OM_ENCRYPTED_SORT_MAX_ROWS` to protect the instance from encrypted-sort scans **does not** cap this table, and gets no signal that a second knob exists.
2. The defaults disagree in direction — the platform default is deliberately **uncapped** ("Unset or invalid input means uncapped — the default must stay byte-identical to the pre-cap behavior", `encrypted-sort.ts:28-31`), while this spec defaults to a hard 5000. So the same deployment would have one encrypted sort uncapped and another capped, for no reason a reader could infer.

**Fix:** use `OM_ENCRYPTED_SORT_MAX_ROWS`. If `customer_accounts` genuinely needs a tighter local ceiling, specify it as an override that *falls back to* the platform value rather than a parallel variable with its own default, and say so in `hybrid-query-engine.mdx` so the two are described in one place.

### 🔴 Blocker 3 — the degraded-sort response field diverges from the documented contract

*`:94-107`, `:118`, `:127`*

The spec adds a top-level `sortDegraded: true`. The platform already surfaces this exact condition as `meta.encryptedSortRowCapWarning` — documented at `hybrid-query-engine.mdx:62` and asserted in `packages/core/src/modules/query_index/__tests__/hybrid-engine.test.ts:924`.

The spec correctly argues (`:107`) that an additive optional field is permitted by `BACKWARD_COMPATIBILITY.md` — that part is fine. The problem is not permission, it is that the product would then have two names for one semantic, and a client wanting to render "this ordering is incomplete" generically would have to special-case this route.

**Fix:** adopt `meta.encryptedSortRowCapWarning`. If the admin route deliberately deviates (e.g. because it has no `meta` envelope today), state that reason in API Contracts and add the alias to the docs page so both names are discoverable.

## 🧩 What's missing

### 🟠 Major 4 — blanket nulls-last makes the Verified column unsortable for its main purpose

*`:130`, `:160`*

Implementation step 3 applies "nulls-last semantics" to all four plaintext fields, and the Edge Cases table (`:130`) fixes nulls **last in both directions**. `emailVerifiedAt` is null precisely for unverified users. Under that rule, unverified accounts sink to the bottom ascending *and* descending — so an operator can never bring unverified accounts to the top, which is the one reason to sort the Verified column at all.

The blanket rule is right for `lastLoginAt` (the spec's own justification at `:130` is sound) and wrong for `emailVerifiedAt`.

**Fix:** replace the single sentence with a per-column null-handling table: `lastLoginAt` → nulls last in both directions; `emailVerifiedAt` → sort on the boolean projection (`emailVerifiedAt IS NOT NULL`) so both the verified and unverified group can lead, with `emailVerifiedAt` as a secondary key if you want stable ordering inside the verified group.

### 🟠 Major 5 — decryption concurrency is unbounded and unspecified

*`:61`, `:141`*

The encrypted branch decrypts up to the cap's worth of rows per request, but the spec never says *how*. Every existing call site in the repo bounds it at `DECRYPT_CONCURRENCY = 8` through `mapWithConcurrency` — `labels/route.ts:38,100`, `interactions/encryptedSortPage.ts:5,20`, `query_index/lib/engine.ts:49`, `shared/lib/query/engine.ts:41`.

This matters directly to the ⚠ Q1 sign-off: an unbounded `Promise.all` over 5000 decrypt calls is a materially different performance profile from a chunked scan at concurrency 8, and it is the unbounded version the risk section is implicitly describing at `:141`. The human is being asked to accept a cost that has not been specified.

**Fix:** name `mapWithConcurrency` and the concurrency-8 convention in Architecture unit 2, and re-state the Risks & Impact performance paragraph with the bound in place.

### 🟠 Major 6 — the degraded path has no observability

*`:61`, `:118`, `:127`, `:141`*

The design surfaces a cap hit only to the end user, via the inline notice. Both existing call sites also log it:

`logger.warn('Encrypted sort candidate scan hit OM_ENCRYPTED_SORT_MAX_ROWS cap; results may be incomplete', { component, cap, tenantId, organizationId })` — `labels/route.ts:93`, `interactions/route.ts:509`.

Without that, nobody can tell how often tenants hit the ceiling, which is exactly the data needed to tune the env var this spec introduces. The spec's Risks section leans on env-tunability as the mitigation (`:141`), but tuning is blind without the signal.

**Fix:** add a requirement to emit a structured warning on the degraded path via the `createLogger` facade (per `apps/docs/docs/framework/runtime/logging.mdx`), matching the existing message and fields so the two are greppable together.

## 🔁 Backward compatibility

### 🟡 Minor 7 — "byte for byte" contradicts the tiebreak change

*`:83` vs `:19`, `:131`, `:159`*

API Contracts states that omitting the new params "reproduces today's behavior byte for byte". Step 2 (`:159`) changes the default ordering from `{ createdAt: 'DESC' }` to `{ createdAt: 'DESC', id: 'DESC' }`, which observably reorders tied rows for every existing caller — bulk-created users being the case the spec itself names.

The change is a good one and the latent-pagination-bug justification at `:131` is convincing. The spec just should not claim byte-identity while making it.

**Fix:** soften `:83` to "omitting them preserves the same fields, filters and page semantics", and let the tiebreak stand on its own as a deliberate, justified ordering change on a STABLE surface.

## 📈 How this specification can be improved

### 🟡 Minor 8 — Status sort ignores the locked state the filter treats as distinct

*`:115`, `:160`*

`Status` maps to `isActive`, but the page's status filter is tri-state — `active` / `inactive` / `locked` (`users.ts:55-62`) — and locked rows carry `isActive: true` with a future `lockedUntil`. So sorting Status interleaves locked accounts with active ones, while the filter right next to it treats them as separate.

This is self-consistent with the rendered cell today, which only prints Active/Inactive (`PortalUsersPageClient.tsx:456-470`) — so it is defensible, not wrong. But a reader comparing the sort to the filter will trip on it.

**Fix:** one line in the UI/UX section stating that Status sorts on the two-state `isActive` the cell renders, and that the tri-state locked distinction is filter-only and deliberately out of scope.

### 🔵 Nit 9 — reconcile the null rule with `sortRowsInMemory`'s actual semantics

*`:130`*

If Blocker 1 is taken and the shared comparator is adopted, note that `compareValues` returns missing-last **before** the direction multiplier is applied (`encrypted-sort.ts:71-77`), so nulls lead in descending order rather than trailing. The spec's stated null rule and the helper's behavior need to be reconciled explicitly — either by pre-partitioning or by documenting the helper's semantics as the accepted behavior — rather than left to be discovered during implementation.

### 🔵 Nit 10 — consolidate where the cap is documented

*`:109`, `:163`*

Step 6 adds the new env var to `apps/mercato/.env.example` (and correctly notes the create-app template mirror). Worth knowing: `OM_ENCRYPTED_SORT_MAX_ROWS` is *not* currently in `.env.example` either — it lives only in `hybrid-query-engine.mdx:62`. Whichever variable survives Blocker 2, add it to both places so there is a single discoverable description of encrypted-sort capping.

Not yet — but it is close, and the gap is exactly Blocker 1. With the platform helpers adopted, Architecture unit 2 shrinks from "count, load, decrypt, hand-rolled locale comparator, tiebreak, slice, flag" to roughly the shape of `labels/route.ts:86-110`: resolve the cap, `em.find` the candidate set, `mapWithConcurrency` the decrypt, `sortRowsInMemory`, slice. The custom comparator, the custom tiebreak and the custom env var all disappear, and the ⚠ Q1 question becomes a cap-value choice rather than an architecture decision.

Units 1 and 3 (the client sort-field resolver and making `PortalUsersPageClient` a controlled sorting host) are already minimal and mirror established patterns correctly. The "saved views come for free" analysis at `:69` is accurate — I verified `applyPerspectiveSettings` calls `onSortingChange` at `DataTable.tsx:1930` and that `perspectiveSettingsSchema` persists `sorting`.

Phasing (`:146-152`) is sound: Phase 1 is genuinely independently shippable and Phase 2 depends on it.

## ✅ Validation

Spec-only PR, so the code gates have nothing to compile or run. **Run:** none applicable. **Skipped:** `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app` — no source, locale or config file is touched by this diff. CI's docs-applicable jobs (`lint`, `ds-lint`, `audit-scope`) pass on the head commit.

## 📋 Summary

| Severity | Count | Items |
|---|---|---|
| 🔴 Blocker | 3 | Reinvented platform primitive; conflicting cap knob; divergent response field |
| 🟠 Major | 3 | Verified-column null rule; unbounded decrypt concurrency; no degraded-path logging |
| 🟡 Minor | 2 | "byte for byte" contradiction; Status/locked mismatch |
| 🔵 Nit | 2 | Comparator null semantics; env-var documentation placement |

Blockers 1–3 all resolve in the same edit — adopting `@open-mercato/shared/lib/query/encrypted-sort` and its conventions — and that edit also resolves Majors 5 and 6 almost incidentally, since the bounded-decrypt and warning-log patterns come with the prior art. I'd suggest making that revision **before** asking for the Q1 human sign-off, because the question the reviewer is being asked changes substantially once the reuse option is on the table.

*Autofix: skipped — this run is review-only by request; the author will apply fixes.*

adeptofvoltron:

- 🔄 `changes-requested` → `review` — an independent re-review of head `8ca9be8e7` found 1 blocker and 2 majors that the previous automated pass over the same head had not surfaced, so the label first moved to `changes-requested`. The autofix pass then resolved all eight findings in `d6da31f08`, so that label is no longer true. It returns to `review` rather than `merge-queue` for one honest reason: the same automation account authored the spec, both reviews and both fix passes, and GitHub refuses an approving review on one's own pull request — so the approval this PR still needs is a human specification sign-off, not an automated one.
- 📚 `documentation` — unchanged; the diff is still a single Markdown file under `.ai/specs/` and touches no source, config, locale or CI file.
- ⏭️ `skip-qa` — unchanged and still correct under the automated-verification exemption: the change renders no UI, alters no database structure and moves no API surface, because it describes work that has not been implemented yet. There is nothing a QA reviewer could click through.
- 🔽 `priority-low` — unchanged; the underlying request is an ergonomics improvement to one admin table that already has search and filters as a workaround, so nothing is broken or blocked while this waits.
- 🟢 `risk-low` — unchanged; merging this PR changes only a Markdown file. The blast radius of the design it describes is assessed separately in the review above and in the spec's own Risks & Impact Review section.
- 👁️ `ci-monitoring` — this run's work is finished and fully reported, but the push to `d6da31f08` started a new CI run that this agent still owes a result comment for. The label is meta, not a lock: anyone may act on this PR while it is set, and it is removed as soon as the CI-result comment lands or the wait budget expires.

The author of this PR is the same account as the reviewer, so there is no reassignment handoff to make. The PR is not a draft and this run neither promotes nor demotes it — draft state is a human signal here, and the pipeline label already carries the automated verdict.

adeptofvoltron:

Specification review of `.ai/specs/2026-08-27-customer-users-table-column-sorting.md`: 3 blockers, 3 majors, 2 minors, 2 nits. All three blockers converge on one revision — the spec designs a bounded decrypt-then-sort path for the encrypted `Name`/`Email` columns without referencing `@open-mercato/shared/lib/query/encrypted-sort`, which already ships `resolveEncryptedSortMaxRows()`, `sortRowsInMemory()` and the `OM_ENCRYPTED_SORT_MAX_ROWS` cap, with a direct `em.find` precedent in `packages/core/src/modules/customers/api/labels/route.ts`.

Every code citation the spec makes was verified against the head commit and all of them are accurate; the problem statement and Q2–Q5 hold up. Pipeline label moved to `changes-requested`. Autofix: skipped — review-only run by request, the author applies fixes. No CI follow-up owed; the docs-applicable checks are already green on this head.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-09-02T21:02:17Z.

adeptofvoltron (commented):
# 🔍 Re-review: Customer Users Table Column Sorting

Re-reviewed at `8ca9be8e7` after the `om-auto-fix-pr` autofix pass, against the changes-requested review of `c1f2d3945`'s predecessor. The branch now carries the latest `develop` (merged in `c1f2d3945`, no conflicts), so this judges the real merge result.

**Verdict: all ten findings resolved in the spec — nothing deferred, nothing declined.** Every blocker and major was a design correction rather than a code fix, so all of them landed as edits to `.ai/specs/2026-08-27-customer-users-table-column-sorting.md`; no follow-up issues were filed, because deferring a blocker is not permitted and the two nits were one-line edits in the same document.

## 💥 Blockers

### 🔴 → ✅ Blocker 1 — the design reinvents an existing platform primitive

Resolved. The **Proposed Solution** now names `@open-mercato/shared/lib/query/encrypted-sort` explicitly, citing `resolveEncryptedSortMaxRows()` (`encrypted-sort.ts:32`), `resolveEncryptedSortFields()` (`:41`) and `sortRowsInMemory()` (`:85`, with its locale-aware comparator and built-in `id` tiebreak at `:95-97`), alongside `mapWithConcurrency` from `bounded-decrypt`. Both prior-art call sites are cited — `customers/api/labels/route.ts:86-110` and `customers/api/interactions/encryptedSortPage.ts`.

**Architecture unit 2** was rewritten from "count, load, decrypt, hand-rolled comparator, tiebreak, slice, flag" into the seven-step `labels/route.ts` shape, and closes with the explicit constraint that the branch has *no comparator, no tiebreak, no env parsing and no cap constant of its own*. The rejected-alternatives list gained the reuse option's counterpart — "build a module-local bounded decrypt-then-sort path (the first draft of this spec)" — recorded as rejected on review, so the option set a reader sees is now complete.

One refinement the reviewed prior art suggested but this route improves on: `labels/route.ts` detects a cap hit via `candidateLabels.length >= cap` because it has no count to spend. This route **must** compute `total` for pagination anyway, so the spec reuses that count as both the cap comparison and the warning's `totalMatched` — making the value exact rather than a floor, at zero extra cost.

### 🔴 → ✅ Blocker 2 — a second cap knob that silently conflicts with the platform one

Resolved. `CUSTOMER_USERS_ENCRYPTED_SORT_MAX` is gone from the design; the only surviving mention is in the historical note explaining why the draft's merge gate was lifted. **API Contracts → Configuration** now states that no new environment variable is introduced, that `OM_ENCRYPTED_SORT_MAX_ROWS` governs the branch, and that its platform semantics are adopted unchanged — *unset or invalid means uncapped* (`encrypted-sort.ts:28-38`), read only through `resolveEncryptedSortMaxRows()`. The two defaults can no longer disagree in direction, because there is only one.

The "an operator sets the platform var and this table ignores it" failure mode is closed by construction.

### 🔴 → ✅ Blocker 3 — the degraded-sort response field diverges from the documented contract

Resolved. `sortDegraded: true` is replaced throughout by `meta.encryptedSortRowCapWarning`, carrying the platform shape verbatim (`entity`, `sortFields`, `maxRows`, `totalMatched` — `packages/shared/src/lib/query/types.ts:150-158`). The JSON example, the UI/UX notice trigger, the Edge Cases table and Phase 2 step 10 all read the same field.

The deviation the original review offered to accept is stated rather than glossed: the `meta` envelope *is* new on this route, because the handler builds its own response object and has no `meta` key today. The spec says so, notes it is not a new concept in the product, and keeps the `BACKWARD_COMPATIBILITY.md` additive-optional argument that was already correct.

## 🧩 Majors

### 🟠 → ✅ Major 4 — blanket nulls-last makes the Verified column unsortable

Resolved, and with a better mechanism than the one suggested. Edge Cases gained a dedicated **"Null handling, per column"** table: `lastLoginAt` keeps nulls-last in both directions (its justification was sound); `emailVerifiedAt` now has nulls *follow the direction* — `ASC NULLS FIRST` / `DESC NULLS LAST` — so ascending leads with the unverified group and descending leads with the most recently verified.

I checked the mechanism before specifying it: MikroORM's `QueryOrder` exposes `ASC_NULLS_FIRST` / `DESC_NULLS_LAST` (`@mikro-orm/core/enums.d.ts:93-102`), so this needs no raw SQL and no `QueryBuilder` detour. It is equivalent to the review's suggested `emailVerifiedAt IS NOT NULL` boolean projection — null is a single group parked at one end either way — but it sorts one real column instead of a computed one, and gets intra-group ordering by timestamp for free, so no secondary key is needed.

### 🟠 → ✅ Major 5 — decryption concurrency is unbounded and unspecified

Resolved. Architecture unit 2 step 4 names `mapWithConcurrency(rows, DECRYPT_CONCURRENCY, ...)` with `DECRYPT_CONCURRENCY = 8`, citing all four existing call sites that establish the convention (`labels/route.ts:38,100`, `interactions/encryptedSortPage.ts:5,20`, `query_index/lib/engine.ts:49`, `shared/lib/query/engine.ts:41`), and adds the prohibition explicitly: *never a bare `Promise.all` over the candidate set*. The Risks performance paragraph is restated with the bound in place, and Phase 1 step 4 carries it into the implementation instruction.

### 🟠 → ✅ Major 6 — the degraded path has no observability

Resolved. Architecture unit 2 has an **Observability on the degraded path** subsection with the literal `logger.warn` call, reusing the existing message string verbatim so all three call sites are greppable together, with `component: 'customer_accounts.admin.users.GET'` alongside `cap`, `tenantId` and `organizationId` (matching `labels/route.ts:93` and `interactions/route.ts:509`). Phase 1 step 5 makes emitting it a tested requirement, and the Data Model section notes the log records the cap and scope ids only — never a decrypted value.

## 🔁 Backward compatibility

### 🟡 → ✅ Minor 7 — "byte for byte" contradicts the tiebreak change

Resolved. The claim is softened to "preserves the same fields, filters and page semantics as today", and the `id DESC` tiebreak now stands on its own in a dedicated API Contracts paragraph that says plainly it observably reorders tied rows, is a deliberate ordering change on a STABLE surface, and is taken because a non-deterministic order across page boundaries can drop or duplicate a tied row. The Q5 rationale no longer calls it "not a behavior change".

## 📈 Improvements

### 🟡 → ✅ Minor 8 — Status sort ignores the locked state the filter treats as distinct

Resolved. UI/UX gained a bullet stating that Status sorts on the two-state `isActive` the cell renders (`PortalUsersPageClient.tsx:456-470`), that the filter beside it is tri-state (`users.ts:55-62`), that locked rows therefore interleave with active ones, and that this is deliberate and filter-only — with the reason it is called out at all: a reader comparing sort to filter would otherwise read it as a bug.

### 🔵 → ✅ Nit 9 — reconcile the null rule with `sortRowsInMemory`'s actual semantics

Resolved. The per-column null table's last row documents the encrypted branch's real behavior: `compareValues` returns missing-last (`encrypted-sort.ts:71-75`) *before* the direction multiplier is applied (`:93`), so missing values sort last ascending and **first** descending. The spec accepts the helper's semantics as-is and says why pre-partitioning is explicitly not done — it would fork the shared comparator's behavior for one table — and requires the route tests to assert it, so the coupling stays visible.

### 🔵 → ✅ Nit 10 — consolidate where the cap is documented

Resolved. Phase 1 step 7 adds `OM_ENCRYPTED_SORT_MAX_ROWS` to `apps/mercato/.env.example` (noting it is currently absent there), mirrors it into the create-app template via `yarn template:sync:fix`, and extends `apps/docs/docs/framework/database/hybrid-query-engine.mdx:62` to name this route alongside the query engines — so the cap ends up with one description covering every call site.

## ⚠️ One autonomous decision worth a reviewer's eye

The original review observed that reframing Q1 against the prior art "may not need the ⚠ marker at all". I took that: **the `⚠ NEEDS HUMAN CONFIRMATION` marker is removed**, and the spec explains why in a dedicated *"Why the ⚠ marker was removed"* section — with the platform path adopted, there is no new env var to name, no default to pick and no response semantic to coin, so no trade-off is left for a human to sign off, and the residual cost is the one the product already accepts on two other routes.

This is flagged rather than buried because it changes the PR's stated merge gate. It is cheap to reverse, and the PR **remains a draft** either way: a spec-only design PR is promoted and approved by a human, not by automation.

Yes, now. The simplification the original review predicted actually materialized: Architecture unit 2 lost its custom comparator, custom tiebreak, custom env var and custom response field, and what remains is the `labels/route.ts` shape plus the one thing genuinely specific to this route — reusing the pagination count it already owes as the cap comparison. Units 1 and 3 were already minimal and are unchanged.

## ✅ Validation

Spec-only PR — one file, `.ai/specs/2026-08-27-customer-users-table-column-sorting.md`, no source, locale, config or CI file touched. **Run:** none applicable. **Skipped:** `yarn build:packages`, `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app` — there is nothing for them to compile or execute.

Every code citation added in this revision was verified against the merged head before being written: `encrypted-sort.ts:32/41/71-75/85/93/95-97`, `bounded-decrypt.ts`, `labels/route.ts:38/86-110/93`, `interactions/route.ts:509`, `types.ts:150-158`, `hybrid-query-engine.mdx:62`, `hybrid-engine.test.ts:1036`, `enums.d.ts:93-102`, `encryption.ts:6-9`, `users.ts:55-62`, `PortalUsersPageClient.tsx:456-470`. One line reference from the original review (`hybrid-engine.test.ts:924`) had drifted and was corrected to `:1036` rather than copied over.

## 📋 Summary

| Severity | Count | Resolution |
|---|---|---|
| 🔴 Blocker | 3 | All fixed in the spec — platform primitive adopted, single cap knob, documented warning field |
| 🟠 Major | 3 | All fixed — per-column null rule, bounded decrypt, degraded-path logging |
| 🟡 Minor | 2 | All fixed — byte-identity claim softened, Status/locked asymmetry documented |
| 🔵 Nit | 2 | All fixed — comparator null semantics reconciled, cap documented in both places |

No finding was deferred to a follow-up issue and none was declined. What remains is a human specification sign-off on the revised design, which is what a spec-only draft is for.

adeptofvoltron:

- 🔄 `changes-requested` → `review` — every one of the ten specification-review findings (3 blockers, 3 majors, 2 minors, 2 nits) is now resolved in the spec document, so the label that said "this design needs a revision pass" is no longer true. It moves to `review` rather than `merge-queue` for two honest reasons: this is a draft spec-only design PR that nobody can merge until a human promotes it, and the automation account authored both the spec and the review, so the approval this PR still needs is a human specification sign-off, not an automated one.
- 📄 `documentation` — unchanged; the diff is a single Markdown file under `.ai/specs/` and touches no source, locale, config or CI file.
- 🧪 `skip-qa` — unchanged and correct under the automated-verification exemption: the change renders no UI, alters no database structure and moves no API surface, because it is a design document describing work that has not been implemented yet. There is nothing a QA reviewer could click through.
- ⬇️ `priority-low` — unchanged; this specifies a usability improvement to one admin table. No user is blocked, no data is at risk, and nothing regresses while it waits.
- 🟢 `risk-low` — unchanged, and the revision lowered the real risk further rather than raising it: the design now reuses shipped platform primitives instead of introducing a parallel cap knob, a second name for a documented response field and a hand-rolled comparator. A spec document has no blast radius of its own in any case.
- 🔁 `in-progress` → `ci-monitoring` — the work of this run is finished and fully reported (spec revised and pushed, PR body corrected, re-review posted, summary comment below), so the claim no longer holds and the PR is free for anyone to act on. `ci-monitoring` records the one thing still owed: a follow-up comment with the CI outcome on the new head.

adeptofvoltron:

Driven from `changes-requested` to review-ready. Below is everything this run did and why, so it can be judged without opening the diff.

### 🔀 Base merge — done first, so nothing was judged against a stale base

`origin/develop` was merged into `spec/customer-users-table-column-sorting` before any review or CI work, in commit `c1f2d3945`. It merged cleanly with no conflicts — unsurprising, since the PR touches one file under `.ai/specs/` that develop does not, but it matters that the review below judged the real merge result rather than a five-day-old branch. The head is `adeptofvoltron/open-mercato`, a fork this automation owns, so the update was pushed directly; no carry-forward replacement PR was needed.

### 🔍 Review findings — all ten fixed, none deferred

The specification review posted earlier requested changes with 3 blockers, 3 majors, 2 minors and 2 nits. Every one is now resolved in `.ai/specs/2026-08-27-customer-users-table-column-sorting.md` (commit `8ca9be8e7`); the finding-by-finding accounting is in the re-review comment above.

The three blockers shared one root cause and one fix. The spec had designed a module-local bounded decrypt-then-sort path for the encrypted `Name` and `Email` columns — with its own env var, its own comparator and its own response field — while the repository already ships exactly that machinery in `@open-mercato/shared/lib/query/encrypted-sort`, with two routes already composing it. Every claim in the review was verified against the merged head before acting on it, and all of them held. The revision adopts the platform path, which also dissolved two of the three majors as a side effect: the bounded-decrypt convention (`DECRYPT_CONCURRENCY = 8` via `mapWithConcurrency`) and the degraded-path warning log come with the prior art.

The remaining major was independent and genuinely a design bug: a blanket nulls-last rule would have made the `Verified` column unsortable for its only purpose, since `emailVerifiedAt` is null precisely for unverified users, so they would sink to the bottom in *both* directions. That became a per-column null-handling table, with `emailVerifiedAt` letting nulls follow the sort direction. Before specifying it I confirmed MikroORM's `QueryOrder` exposes `ASC_NULLS_FIRST` / `DESC_NULLS_LAST` (`@mikro-orm/core/enums.d.ts:93-102`), so it needs no raw SQL.

**No reviewer feedback was inherited from anyone else** — the only review on this PR was the automated specification review from this same chain, and no human or bot comments were left unaddressed.

### ⚠️ One decision made autonomously

The `⚠ NEEDS HUMAN CONFIRMATION` marker on Q1 has been **removed**, and the PR body's merge-gate section rewritten to match. The reasoning is in the spec under *"Why the ⚠ marker was removed"*: the marker guarded a build-it-ourselves trade-off — a new `CUSTOMER_USERS_ENCRYPTED_SORT_MAX` defaulting to 5000, an unbounded decrypt, a new `sortDegraded` field — and reusing the platform primitives removes each of those rather than deciding it. What is left is "should this table sort encrypted columns the way the rest of the product already does", which needs no sign-off gate.

This is called out prominently because it changes what the PR said it was gated on. It is one paragraph to reverse, and it does not unlock anything on its own: the PR stays a draft, and a spec-only design PR is promoted and approved by a human either way.

### 🧪 CI

The docs-applicable checks were green on the pre-merge head (`lint`, `ds-lint`, `test`, `audit-scope`, `documents-multi-instance`, `prepare`, `scope`; the heavier jobs skip on a docs-only diff). A fresh run is in flight for the new head at the time of writing — `scope` and `license/cla` have already passed, the rest are still registering.

**CI is still running on this head.** Branch protection plus the draft state hold the actual merge; this run's verdict covers the work, not a green run. A follow-up comment will report the CI outcome.

No CI failure was fixed by this run, because none occurred — and none could have been fixed by weakening a test or disabling a check in any case.

### 📸 UI verification

**n/a.** The diff is one Markdown file under `.ai/specs/`; there is no user-facing surface to drive and nothing to screenshot. `skip-qa` stays, and no QA label was added — this run never applies `qa-approved`.

### 🔁 Follow-ups filed

**None, deliberately.** Follow-up issues exist so non-blocking findings do not hold a PR hostage, but here all ten findings were edits to a single design document that was already open for revision — including both nits, which were a one-line clarification each. Filing issues for them would have tracked work that was cheaper to just do, and the blockers and majors could not be deferred at all.

### ✅ Merge-readiness

**Not merge-ready, by design, and that is the correct end state.** The spec is now internally consistent, grounded in verified code citations, and free of the design problems review found. What it still needs is a human specification sign-off and a human promoting the draft — which is exactly what a spec-only design PR is for, and something this automation must not do for itself. There is no `om-approve-merge-pr` hand-off to make yet; that becomes appropriate once a human approves the design and promotes the PR out of draft.

adeptofvoltron:

CI settled green on head `8ca9be8e7` (the revised spec on top of the merged `develop`), well inside the 40-minute wait budget. **The verdict in the run summary above is unchanged**, and the `review` pipeline label stands.

**Passing (8):** `audit-scope` (14s), `documents-multi-instance` (1m30s), `ds-lint` (1m4s), `license/cla`, `lint` (33s), `prepare` (50s), `scope` (31s), `test` (6m16s).

**Skipped (6):** `audit`, `comment`, `docker-build`, `ephemeral-integration`, `merge-coverage`, `mutate` — these skip on a docs-only diff, which this is: one Markdown file under `.ai/specs/`.

Nothing failed, so no failure needed classifying and nothing was rerun. Worth stating explicitly given the base merge in `c1f2d3945` pulled 71 commits of `develop` into this branch: `test` passed against that merged state, so the green here is the merge result, not a stale branch.

`ci-monitoring` is removed — this follow-up was the only thing it was tracking, and it has now landed.

The PR remains a **draft** awaiting a human specification sign-off; green CI does not change that, and this automation does not promote or approve its own design PR.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-09-03T17:26:54Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron (commented):
# 🔍 Specification review (re-review): Customer Users Table Column Sorting

Re-reviewed at head `8ca9be8e7`, the same head the previous `om-auto-fix-pr` re-review approved. This run was invoked explicitly on this PR number, so it is a deliberate second opinion on an unchanged head rather than a review of new commits — worth stating plainly, because the previous pass was written by the same automation that had just applied the fixes, and a fresh grounding pass is exactly the thing that catches what a fix-then-self-check loop misses. It did.

## 🎯 Summary

The spec's revision direction was right: the first draft invented a module-local encrypted-sort path with its own env var, comparator and degraded-signal name, and this version replaces all of that with `@open-mercato/shared/lib/query/encrypted-sort` plus `mapWithConcurrency`. That is the correct call and it removes a genuine architectural decision rather than deferring it. The problem statement is accurate, the Q2–Q5 resolutions hold up, the per-column null-placement table is a real improvement over a blanket rule, and I verified every code citation in the document against this head — including the ones that look most likely to be stale (`DataTable.tsx:3392`, `encrypted-sort.ts:71-75`/`:93`/`:95-97`, `perspectives/data/validators.ts`, `people/page.tsx:924-926`, `hybrid-engine.test.ts:1036`, `QueryOrder.ASC_NULLS_FIRST`/`DESC_NULLS_LAST` in MikroORM v7). They are all correct.

What is not correct is the code shape the spec prescribes for the one thing it calls the hard problem. The encrypted branch as written decrypts the candidate set **twice** — once serially, inside a helper the spec treats as a plain loader, and once through the bounded map it presents as the only decrypt pass. That falsifies the performance claim the Risks section rests on, and it is the difference between "the same bounded scan the product already performs" and something materially slower than the precedent it cites. Alongside it, the response contract for the capped branch is left undefined in exactly the degraded case the document spends most of its words on.

## ✅ Verdict

**Changes requested** — 1 blocker and 2 majors. Per the severity rule, a single blocker requires changes regardless of everything else that is right here, and there is a lot that is right.

GitHub refuses a `REQUEST_CHANGES` review on one's own pull request, so this verdict is carried in the body of a `COMMENT` review and in the `changes-requested` pipeline label, which is the authoritative signal.

## 🧪 Validation Gate

This is a spec-only PR — one Markdown file under `.ai/specs/`, 237 added lines, no source, config, locale or CI file — so the compile-and-test half of the configured gate has nothing to act on. What ran and what did not:

| Command | Result |
|---|---|
| `.ai/specs` filename-convention check | ✅ `2026-08-27-customer-users-table-column-sorting.md` matches `{YYYY-MM-DD}-{kebab-case-title}.md` and introduces no legacy `SPEC-*` prefix |
| Spec Content Checklist (`.ai/specs/AGENTS.md`) | ❌ two mandated sections absent — see M2 |
| `yarn build:packages`, `yarn generate`, `yarn typecheck`, `yarn test`, `yarn build:app`, `yarn i18n:check-sync`, `yarn i18n:check-usage` | ⏭️ skipped — a Markdown-only diff compiles nothing, generates nothing and ships no translatable string. CI ran the full set on this exact head and every required check is green (`test` 6m16s, `lint`, `ds-lint`, `audit-scope`, `documents-multi-instance`, `prepare`, `scope`, `license/cla`), which is the stronger evidence anyway |

No required check was pending at the time of this review, so there is no CI disclosure to make and no follow-up is owed.

## ⛔ Blockers

### B1 — The encrypted branch decrypts the candidate set twice, and the first pass is serial

`.ai/specs/2026-08-27-customer-users-table-column-sorting.md:83-84` (Architecture unit 2, steps 3–4) and `:227` (Implementation Plan step 4) prescribe this sequence:

3. load the candidate set with `findWithDecryption` under `{ limit: cap ?? undefined, orderBy: { id: 'asc' } }`
4. then `await mapWithConcurrency(rows, DECRYPT_CONCURRENCY, ...)`, with the instruction "Never a bare `Promise.all` over the candidate set"

`findWithDecryption` is not a loader. `packages/shared/src/lib/encryption/find.ts:39-44` shows it calls `decryptEntitiesWithFallbackScope(records, ...)` over the whole result array before returning, and `packages/shared/src/lib/encryption/subscriber.ts:480-487` shows that helper is a plain `for (const entity of list) { await subscriber.decryptEntityGraph(...) }` — strictly one row at a time, no concurrency at all. So step 3 already decrypts every candidate row, serially, and step 4 then decrypts the same rows again at concurrency 8.

This is precisely what the precedent the spec says it is following avoids. `packages/core/src/modules/customers/api/labels/route.ts:88-92` deliberately uses a raw `em.find` — no decryption — so that the `mapWithConcurrency` at `:102-104` is the *only* decrypt pass. The spec substitutes `findWithDecryption` for that `em.find` while keeping the rest of the shape, which is what breaks it.

The consequence is not cosmetic. `:207` (Risks) tells the operator the cost is "bounded by `OM_ENCRYPTED_SORT_MAX_ROWS` when set, and by concurrency 8 through `mapWithConcurrency` always", and concludes "this is not a new risk profile: it is the same bounded scan the product already performs for customer labels". Under the prescribed code the dominant cost is the serial pass that the concurrency bound does not touch, and `:194` documents that the cap is **unset by default** — so on a default deployment, sorting by Name serially decrypts the entire filtered set and then decrypts it again. That is a different and worse profile than the one the risk assessment signs off on, on the one path the whole spec exists to design.

The fix is a paragraph, but it needs a decision the spec has to make explicitly rather than leave to the implementer: `packages/shared/AGENTS.md` states "MUST use `findWithDecryption` / `findOneWithDecryption` instead of raw `em.find` / `em.findOne`", so an implementer reading both documents cannot satisfy both. Prescribe the labels shape — `em.find` for the candidate scan, `mapWithConcurrency(candidates, DECRYPT_CONCURRENCY, (row) => decryptEntitiesWithFallbackScope(row, { em, tenantId, organizationId }))` as the single bounded decrypt — and say in the spec why this route is the sanctioned exception to that rule, the same way `labels/route.ts` already is. The rule exists so encrypted reads get decrypted; doing the decryption explicitly and under a concurrency bound satisfies its intent, and naming that in the spec is what stops the next reviewer from re-litigating it.

## ⚠️ Majors

### M1 — `total` and `totalPages` are left undefined for the capped encrypted branch

`:82` makes the filtered count serve as both the cap comparison and the warning's `totalMatched`, and `:86` slices the requested page out of the sorted set — which contains at most `cap` rows. Nothing in the document then says what `total` and `totalPages` carry in that branch.

Take the spec's own example figures (`:144-151`): cap 5000, 12043 matching rows, the route's `pageSize` bounded at 100 (`users.ts:41`) and defaulting to 25. If `total` stays 12043, `totalPages` advertises 482 pages while only the first 200 can ever contain a row, and the pager renders 282 clickable pages that return empty. If `total` is instead clamped to the cap, the response contradicts the `totalMatched: 12043` sitting beside it in the same payload.

The Edge Cases row at `:193` covers the warning, the log and the UI notice thoroughly, and then stops one step short of the pagination arithmetic; the API Contracts example at `:137-153` only shows the uncapped shape. Either answer is defensible — clamp `total` to `min(total, cap)` in the capped branch and let `totalMatched` carry the true figure, or keep `total` exact and state that pages past `ceil(cap / pageSize)` are empty by design — but the spec has to pick one, because this is the response contract in the degraded case it spends most of its words on, and an implementer will otherwise pick silently.

### M2 — Two sections the repo's spec checklist mandates are absent

`.ai/specs/AGENTS.md` → Spec Content Checklist requires that "Every non-trivial spec includes: TLDR, Overview, Problem Statement, Proposed Solution, Architecture, Data Models, API Contracts, Risks & Impact Review, Final Compliance Report, Changelog." This document has twelve `##` sections and neither **Final Compliance Report** nor **Changelog** is among them; there is also no distinct Overview, though the TLDR at `:3-7` reasonably covers that ground.

Peer specs merged in the same window carry both — `.ai/specs/2026-08-24-search-token-double-write.md:161` and `:170`, plus `2026-08-21-integration-secret-credential-editing.md` and `2026-08-27-deal-people-tab-linked-people-parity.md`. On a spec-only PR the document *is* the entire deliverable, so its own repository checklist is the acceptance criterion rather than a formality. A pre-implementation Changelog is legitimately a single dated "initial specification" line, and the Final Compliance Report is the place to record the tenant/organization-scoping and no-contract-break assertions the Risks section already makes in prose — neither is expensive, and their absence is what a reviewer checking conformance notices first.

## 💡 Minors

### m1 — The degraded-path trigger diverges from the precedent while the spec claims it matches

`:87` fires the warning on `cap !== null && total > cap`; `labels/route.ts:93` fires it on `candidateLabels.length >= cap`. The spec's version is genuinely better — it compares against the exact count instead of the fetched length, which is why it can promise an exact `totalMatched` — but `:228` describes the emission as "matching the existing message and fields verbatim so all three call sites are greppable together". That is true of the message and the fields and false of the trigger. One sentence flagging the divergence as deliberate stops an implementer or a later reviewer from "restoring" it to `>=` and reintroducing the off-by-one at `total == cap`, where the scan was in fact complete and no warning is due.

### m2 — A multi-column saved sort silently degrades to its first column

`perspectiveSettingsSchema` accepts a `sorting` array of up to 20 entries (`packages/core/src/modules/perspectives/data/validators.ts:12-16`), and the customers precedent this spec mirrors takes `sorting[0]` and drops the rest (`customers/backend/customers/listSorting.ts:21-26`). Inheriting that is the right call — the API contract at `:124-127` is single-column by construction — but the Edge Cases table, which is otherwise admirably complete about saved views referencing removed columns (`:200`), never says what happens to a view saved with two sort columns. It is one row.

## 🔬 Nits

- `:108` cites `perspectiveSettingsSchema`'s `sorting` array as `validators.ts:12-15`; the field spans `:12-16` — the `.optional()` closes on 16.
- `:129` says `sortField`/`sortDir` are "Validated with zod alongside the existing schema". `users.ts` does contain zod schemas (`roleSchema`, `userSchema` at `:327-335`), but they describe the OpenAPI *response*; the query string is parsed with bare `url.searchParams.get` at `:39-46` and has no schema today. "alongside the existing response schemas" would be accurate and would also make clear that this adds the route's first query validation.

None. The PR carries no inline diff comments (`list-review-comments` returned zero) and no reviews or conversation comments from anyone outside this skill collection — every existing comment is an `om-auto-write-spec`, `om-auto-review-pr` or `om-auto-fix-pr` marker, which the collection filters out of inherited feedback by design. Nothing on this PR was waiting on an answer.

## 💥 Breaking changes

The spec's own backward-compatibility analysis is sound and I found nothing it misses. The two new query parameters are optional and additive; `meta.encryptedSortRowCapWarning` is an optional response field adopting the platform's existing name, shape and type (`packages/shared/src/lib/query/types.ts:150-158`), absent on every current code path; no FROZEN or STABLE surface changes shape. The one deliberate observable change — `{ createdAt: 'DESC' }` becoming `{ createdAt: 'DESC', id: 'DESC' }` at `:133` — is correctly identified as a real ordering change rather than a no-op, correctly justified as fixing a latent cross-page drop/duplicate on tied timestamps, and correctly not dressed up as byte-identical. That paragraph is the standard this document should be held to throughout.

## 🧪 Test coverage

The per-step test notes are unusually concrete for a spec: the mandated Name ascending/descending regression from the FR is present at `:227`, the over-cap path and a non-string decrypted value are both covered, `:228` asserts on the logger call as well as the response field, the two distinct null placements are asserted separately at `:226`, and `:237` requires self-contained integration coverage that creates and tears down its own fixtures per `.ai/qa/AGENTS.md`. Two gaps follow directly from the findings above and should be added with them: an assertion pinning `total`/`totalPages` in the capped branch once M1 decides what they mean, and — once B1 is resolved — an assertion that the candidate set is decrypted exactly once, since a redundant second pass is invisible to every test currently listed.

---

Re-reviewed against head `8ca9be8e7`. Every file, line and symbol referenced above was read in an isolated worktree checked out at that commit.

adeptofvoltron:
## 🤖 `om-auto-review-pr` — autofix pass

The review above requested changes with 1 blocker, 2 majors, 2 minors and 2 nits. This PR belongs to the automation account, so the autofix loop is eligible and ran immediately. All eight findings are resolved in `d6da31f08`; nothing was deferred and nothing was declined. The base was checked first, as the work order requires — `origin/develop` is still an ancestor of this head, so there were no conflicts to resolve before the findings.

### ⛔ Blocker

**B1 — the encrypted branch decrypted the candidate set twice.** Architecture unit 2 now specifies a raw `em.find` for the candidate scan, making the bounded `mapWithConcurrency` the branch's only decrypt pass, exactly as `customers/api/labels/route.ts:88-104` already does. A new paragraph, *"Why this branch uses `em.find` rather than `findWithDecryption`"*, states why this is a sanctioned exception to the `packages/shared/AGENTS.md` rule and cites the two lines that make it necessary (`encryption/find.ts:39-44` calls `decryptEntitiesWithFallbackScope` over the whole array; `encryption/subscriber.ts:480-487` decrypts strictly one row at a time in a sequential `for … await` loop). Implementation Plan step 4 carries the same instruction plus a test that asserts the decrypt helper runs exactly once per candidate row — a redundant second pass was invisible to every test previously listed. The Risks performance bullet no longer claims a bound the prescribed code did not deliver; it now says "exactly once" and explains why that word is load-bearing.

### ⚠️ Majors

**M1 — `total`/`totalPages` were undefined in the capped branch.** API Contracts gains a *"Pagination when the cap bites"* section: `total` is clamped to `cap`, `totalPages` derives from the clamped value, and the true match count travels in `meta.encryptedSortRowCapWarning.totalMatched`. The rejected alternative is written out with the spec's own figures — an exact `total` of 12043 against a 5000-row cap would advertise 482 pages when only 200 can return a row. The choice follows a precedent rather than inventing one: `meta.listCountCapWarning` already documents `total` as "a floor, not an exact value" when `OM_LIST_COUNT_CAP` bounds it (`packages/shared/src/lib/query/types.ts:150-158`). Architecture step 6, the Edge Cases row and Implementation Plan step 5 were updated to match, with a test pinning both the clamped and the exact case.

**M2 — two mandated sections were missing.** `## 📝 Final Compliance Report` and `## 📝 Changelog` are added, matching the shape peer specs use (`2026-08-24-search-token-double-write.md:161,170`). The compliance report asserts the nine surfaces `.ai/specs/AGENTS.md` cares about — scoping, encryption, contract surfaces, database, discovery, configuration, access control, i18n, design system — as claims the implementing PR must re-assert rather than as a rubber stamp. The changelog carries three dated entries covering the original draft, the first review's revision, and this one.

### 💡 Minors and 🔬 nits

- **m1** — Architecture step 7 now documents `totalMatched > cap` as a deliberate divergence from `labels/route.ts:93`'s `length >= cap`, explains that the latter fires spuriously when the filtered set is exactly `cap` rows, and says not to "restore" it.
- **m2** — a new Edge Cases row states that a saved view carrying more than one sort column degrades to its first, matching `customers/backend/customers/listSorting.ts:21-26`, and names what widening it would cost.
- **n1** — the perspectives `sorting` citation is corrected to `validators.ts:12-16`.
- **n2** — "alongside the existing schema" is corrected to "the existing response schemas (`users.ts:327-335`)", with a note that this adds the route's first query validation and that converting the existing raw `searchParams.get` parameters is out of scope.

### 🧪 Validation

The diff remains one Markdown file under `.ai/specs/`, so the compile-and-test half of the gate still has nothing to act on. The spec now satisfies the `.ai/specs/AGENTS.md` Spec Content Checklist that M2 flagged, and the filename convention check still passes. CI is re-running on the new head; its result will be reported in a follow-up comment.

### 🏷️ Where this leaves the PR

The pipeline label moves back to `review`, not `merge-queue`. Every finding is fixed, but the approval this PR still needs is a **human specification sign-off**: the same automation account authored the spec, both reviews and both fix passes, and GitHub will not let it submit a real approving review on its own pull request. `review` is the honest state — the design is ready to be read, by someone other than the thing that wrote it.

adeptofvoltron:

Independent specification re-review of `.ai/specs/2026-08-27-customer-users-table-column-sorting.md` at head `8ca9be8e7` — the same head a previous automated pass had approved. It found 1 blocker, 2 majors, 2 minors and 2 nits. The blocker was that the encrypted-sort branch, the one mechanism the whole spec exists to design, was specified as `findWithDecryption` followed by `mapWithConcurrency`; because `findWithDecryption` already decrypts the whole result array, and does so strictly serially, that shape decrypts every candidate row twice and the first pass is unbounded by the concurrency limit the Risks section relies on. Every other code citation in the document was verified against the head and all of them are accurate.

Autofix: ran — all 8 findings fixed in one commit, `d6da31f08`, none deferred or declined. The base was checked first and `origin/develop` is still an ancestor of this head, so there were no conflicts to resolve ahead of the findings. Pipeline label moved `review` → `changes-requested` → `review`; the finding-by-finding accounting is in the autofix comment above.

The PR is not merge-ready and this automation cannot make it so: it authored the spec, both reviews and both fix passes, so the sign-off this design still needs is a human one. `ci-monitoring` is set because the push started a new CI run whose result this agent still owes; it is a meta label, not a lock, and anyone may act on this PR now.

adeptofvoltron:

CI settled **green** on head `d6da31f08` — the autofix commit carrying all eight specification-review findings — roughly eleven minutes after the push, well inside the 40-minute wait budget. **The verdict above is unchanged** and the `review` pipeline label stands.

**Passing (8):** `audit-scope` (13s), `documents-multi-instance` (46s), `ds-lint` (50s), `license/cla`, `lint` (46s), `prepare` (1m1s), `scope` (32s), `test` (6m50s).

**Skipped (7):** `audit`, `comment`, `docker-build`, `ephemeral-integration`, `merge-coverage`, `mutate`, `[code]smith` — these skip on a docs-only diff, which this is: one Markdown file under `.ai/specs/`.

Nothing failed, so there was no failure to classify and nothing to rerun. Green here is worth exactly what it is worth on a spec PR: it confirms the change breaks nothing, not that the design is right. The design still needs the human specification sign-off described in the autofix comment above — this automation authored the spec, both reviews and both fix passes, so there is no independent reader in the loop yet.

`ci-monitoring` is removed; this follow-up was the only thing it was tracking and it has now landed.

wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-09-23T06:32:59Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (changes requested):
# 🔍 Specification review: Customer Users Table Column Sorting

**Verdict: changes requested.** Two minor findings and one nit, no blockers and no majors. This repo requests changes on minor findings. Both fixes are a sentence or two in the spec.

Reviewed at head `d6da31f08`. This is a spec-only PR with one file, `.ai/specs/2026-08-27-customer-users-table-column-sorting.md`. It is 20 days behind `develop` but still mergeable, with no conflicts.

**Validation source:** GitHub checks. All required checks are green on `d6da31f08`: `lint`, `ds-lint`, `test`, `audit-scope`, `documents-multi-instance`, `prepare`, `scope` and `license/cla`. The build, integration, docker and mutation jobs skip on a docs-only diff. I ran nothing locally because a spec PR has nothing to compile or test.

## What I checked against current `develop`

The design holds up. I re-checked every load-bearing citation against `origin/develop` and all of them are still accurate:
- `users.ts:156` still hard-codes `orderBy: { createdAt: 'DESC' }`, and query parsing is still bare `searchParams.get` at `:40-46`.
- `customer_accounts/encryption.ts` still encrypts `email` (with `email_hash`) and `display_name`.
- `encrypted-sort.ts` is unchanged: `resolveEncryptedSortMaxRows` at `:32`, missing-first-on-desc in `compareValues` at `:70-75` and `:93`, and the `id` tiebreak at `:95-97`.
- `mapWithConcurrency` has the documented chunked semantics.
- `labels/route.ts:85-109` is still the raw `em.find` → `mapWithConcurrency(…, 8, decryptEntitiesWithFallbackScope)` → `sortRowsInMemory` precedent.
- The `EncryptedSortRowCapWarning` type is at `types.ts:150-159`.
- The perspectives `sorting` schema and `DataTable`'s `onSortingChange` on perspective apply are both still there.
- The page's column `accessorKey`s match the proposed `sortField` tokens one to one.

Reusing the platform encrypted-sort path is the right call. The earlier review rounds' findings are all addressed in the current text. Those rounds came from the author's own automation, so I verified the fixes rather than re-raising them.

Nobody is working on this in parallel: #5672 is open and this PR is the only one referencing it.

## 💥 Findings

### 🟡 Minor 1: the capped-pagination precedent points at a flag that would reintroduce the empty pages it is meant to prevent

*Spec, API Contracts → "Pagination when the cap bites" (`:159-161`) and Edge Cases (`:199`)*

The spec justifies clamping `total` to `cap` by pointing at `meta.listCountCapWarning` ("`total` as a floor… an established shape in this codebase"). On the client, that precedent is expressed as `totalIsCapped: true`: `crud/factory.ts:2073` spreads it whenever `listCountCapWarning` is present. **This page already consumes that flag.** `PortalUsersPageClient.tsx:326` reads `payload.totalIsCapped` and `:599` passes it to `DataTable`. `DataTable.tsx:2781-2843` then switches the pager into open-ended mode, where `hasNextPage = pageIsFull` and there is no last page.

Those two caps mean different things:
- **The count cap:** rows beyond the floor exist and can be paged to.
- **The encrypted-sort cap:** rows beyond `cap` are *unreachable*, because the page is sliced from a sorted set that was itself capped.

An implementer who follows the stated precedent to its conventional client signal will set `totalIsCapped: true` in the capped branch. After the last full page, the pager then offers a Next that returns nothing, which is exactly the failure this section argues against.

**Fix:** say explicitly that the capped encrypted branch reports the clamped `total` as an exact pager bound and **must not** set `totalIsCapped`. Also reword the precedent sentence so it cites `listCountCapWarning` only for "disclose the true count through `meta`", not for the pager semantics. It would help to add an assertion to Implementation step 5's test that `totalIsCapped` is absent in the capped case.

### 🟡 Minor 2: the integration test's order assertion needs to be scoped to its own fixtures

*Spec, Implementation step 11 (`:244`)*

"Sorts by Name ascending and descending and by Created ascending and descending, asserts the returned order" does not say *which* order. In the CI tenant, the list also contains every other portal user present at run time: seeded users, and fixtures from concurrently running specs. The list is also paged at 50. An assertion on the returned page is therefore non-deterministic, and that conflicts with the "remain stable without relying on seeded/demo data" rule in `.ai/qa/AGENTS.md`.

**Fix:** specify how the assertion is scoped. Either narrow the list to the fixtures, for example with a unique display-name token through the existing `search` param, or assert the *relative* order of the created fixture ids within the returned items. Give the fixtures names chosen so that they sort differently from their creation order. Without that, a Name sort that silently falls back to `createdAt` order would still pass.

### ⚪ Nit: line citations have drifted since the base moved

`PortalUsersPageClient.tsx:538` for the `DataTable` is now `:548`. The Status cell at `:456-470` has shifted by about 10 lines as well. `DataTable.tsx:1893+` for the perspective-apply `onSortingChange` is now around `:2009`. None of these change the design. A quick refresh after merging `develop` would keep the spec's grounding checkable. This is optional.

## Direction notes (not defects)

- Adding `OM_ENCRYPTED_SORT_MAX_ROWS` to `.env.example` plus the create-app template, and extending `hybrid-query-engine.mdx`, goes slightly beyond the FR. It is justified as making an existing knob discoverable, and the spec already names the template-sync obligation. I'm fine with it.
- This will be the third route to declare a local `DECRYPT_CONCURRENCY = 8`. If a fourth call site ever appears, it could be worth exporting a shared constant next to `mapWithConcurrency`. That is out of scope here.

## Next step

Amend the pagination paragraph (Minor 1) and Implementation step 11 (Minor 2), then re-request review. Once both minors are addressed this is approvable.

wojciechszyjka:

- 🔄 `review` → `changes-requested`: the [specification review](https://github.com/open-mercato/open-mercato/pull/5699#pullrequestreview-5287655386) found two minor findings. One is the capped-pagination contract versus `totalIsCapped`; the other is the scoping of the integration-test order assertion. This repo requests changes on minor findings.
- 📚 `documentation`, ⏭️ `skip-qa`, 🔽 `priority-low` and 🟢 `risk-low` are unchanged. This is still a single design document with no runtime change.

wojciechszyjka:
@adeptofvoltron, the design is sound and every citation checks out against current `develop`. Two small spec amendments are needed ([review](https://github.com/open-mercato/open-mercato/pull/5699#pullrequestreview-5287655386)):
1. In the capped-pagination paragraph, state that the capped encrypted branch must **not** set `totalIsCapped`. The page already forwards that flag to `DataTable`, which then opens an unbounded pager over rows that cannot be reached.
2. In Implementation step 11, scope the integration test's order assertion to its own fixtures.

Push the update and re-request review.

wojciechszyjka:

Specification review at `d6da31f08`: 0 blockers, 0 majors, 2 minors, 1 nit. CI was already green on this head, so no CI follow-up is pending. Autofix was skipped because this is not my PR; re-run with `--autofix` to fix it here.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-09-23T17:55:21Z.

adeptofvoltron:
## 🤖 `om-auto-review-pr` — autofix pass

The [latest re-review](https://github.com/open-mercato/open-mercato/pull/5699#pullrequestreview-5287655386) requested changes with 2 minor findings and 1 optional nit. This PR belongs to the automation account, so the autofix loop is eligible and ran. `origin/develop` was merged into the branch first (303 commits, no conflicts — the branch was 20 days behind) so this judged the real merge result. Both minors and the nit are resolved in `89e0abe8f`; nothing deferred, nothing declined.

### 🟡 Minor 1 — `totalIsCapped` would reopen the pager past unreachable rows

The "Pagination when the cap bites" section now states explicitly that the response **must not** set `totalIsCapped: true`, and explains why: that flag is the client-side signal `packages/shared/src/lib/crud/factory.ts:2073` derives from `meta.listCountCapWarning`, and `PortalUsersPageClient.tsx:328`/`:610` already forward it into `DataTable.tsx:2781`'s open-ended pager mode. That mode is correct for a count-cap floor (more real rows exist beyond it) but wrong here — the encrypted-sort cap's `total: cap` is an exact bound, not a floor, so setting the flag would offer a Next page past page 200 that always returns empty. The precedent sentence now cites `listCountCapWarning` only for the "disclose the true count through `meta`" pattern, not for pager semantics. Added a code comment to the JSON example and a test assertion (Implementation step 5) that `totalIsCapped` is absent in both the capped and uncapped cases.

### 🟡 Minor 2 — integration test order assertion needed fixture scoping

Implementation step 11 now specifies a unique per-fixture `displayName` token narrowed via the existing `search` param, so the assertion never touches the CI tenant's full — and non-deterministically ordered, since the route pages at 50 — result set. Fixture names are chosen to sort differently from their creation order, so a Name sort silently falling back to `createdAt` order fails the test instead of passing by coincidence.

### ⚪ Nit — stale line citations

Refreshed `PortalUsersPageClient.tsx:538→559`, `DataTable.tsx:1893+→2009`, `DataTable.tsx:3392→3616`, and the Status-cell range `:456-470→470-480` against the merged `develop`.

### 🧪 Validation

Still a single Markdown file under `.ai/specs/` — no source, config, locale or CI file touched. `.ai/specs` filename convention and the Spec Content Checklist both still pass. CI is re-running on the new head; result will follow in a separate comment.

### 🏷️ Where this leaves the PR

Pipeline label moves `changes-requested` → `review`. Both minors and the nit are fixed, but the approval this PR still needs is a **human specification sign-off** re-requested from @wojciechszyjka — the automation account authored the spec and every prior revision, so it cannot submit its own approving review.

adeptofvoltron:

CI settled **green** on head `89e0abe8f` — the autofix commit resolving both minor findings, on top of the freshly merged `develop` — about 9 minutes after the push, well inside the 40-minute wait budget.

**Passing (8):** `audit-scope` (10s), `documents-multi-instance` (57s), `ds-lint` (46s), `license/cla`, `lint` (1m22s), `prepare` (1m4s), `scope` (17s), `test` (8m28s).

**Skipped (7):** `audit`, `comment`, `docker-build`, `ephemeral-integration`, `merge-coverage`, `mutate`, `[code]smith` — these skip on a docs-only diff, which this remains: one Markdown file under `.ai/specs/`.

`test` passing here is meaningful given the base merge pulled 303 commits of `develop` into this branch (20 days behind) — the green is the real merge result, not a stale branch.

`ci-monitoring` is removed; this follow-up was the only thing it was tracking.

The PR is **not merge-ready yet**: `reviewDecision` is still `CHANGES_REQUESTED` from the pre-fix review, and a re-review is requested from @wojciechszyjka. This automation authored the spec and every revision, so it cannot submit an approving review on its own PR — the outstanding gate is a human specification sign-off.

wojciechszyjka (approved):
# 🔍 Specification re-review: Customer Users Table Column Sorting

**Verdict: approved.** Both earlier minors and the nit are fixed. I found nothing new at Medium or worse.

Reviewed at head `89e0abe8f`. The PR is still one file, `.ai/specs/2026-08-27-customer-users-table-column-sorting.md`; the only content change since my review at `d6da31f08` is that file plus a merge of `develop`. It is mergeable with no conflicts.

**Validation source:** GitHub checks. All required checks are green on `89e0abe8f` (`lint`, `ds-lint`, `test`, `audit-scope`, `documents-multi-instance`, `prepare`, `scope`, `license/cla`). The remaining jobs skip on a docs-only diff. I ran nothing locally because a spec PR has nothing to compile or test.

## Earlier findings

- **Minor 1, `totalIsCapped` reopening the pager: fixed.** "Pagination when the cap bites" now says the clamped `total` is an exact pager bound and the response MUST NOT set `totalIsCapped: true`. It explains why, and its `PortalUsersPageClient.tsx` and `DataTable.tsx` references check out (the page reads `payload.totalIsCapped` at `:328`, and `DataTable.tsx:2781` derives open-ended mode from it). The precedent sentence now cites `listCountCapWarning` only for "disclose the true count through `meta`". The JSON example carries a comment, and Implementation step 5's test asserts `totalIsCapped` is absent.
- **Minor 2, unscoped integration order assertion: fixed.** Implementation step 11 now scopes every list call to the test's own fixtures through the existing `search` param and a unique `displayName` token. It also names fixtures so alphabetical order differs from creation order (`Zoe`, `Alice`, `Mid`), so a silent fallback to `createdAt` fails the test. `users.ts` resolves `search` through the search-tokens table, which covers the encrypted `displayName`, so the approach is workable.
- **Nit, drifted line citations: fixed.** I spot-checked the refreshed ones against the head: `DataTable.tsx:2009` (perspective-apply `onSortingChange`), `:2781` (`totalIsCapped`) and `:3616` (`getCanSort`), and the Status cell at `PortalUsersPageClient.tsx:470-480`. `PortalUsersPageClient.tsx:559` for the `DataTable` element also matches. All match.

## New commits

Nothing new. The autofix commit only amends the spec text described above, and a Changelog line was added for it.

## Next step

This is a spec-only design PR, so the merge decision stays with a maintainer. Labels: `review` → `merge-queue`; `skip-qa` stays.

wojciechszyjka:

- 🔄 `review` → `merge-queue`: the [specification re-review](https://github.com/open-mercato/open-mercato/pull/5699) approved head `89e0abe8f`; both minors and the nit are fixed. `skip-qa`, `priority-low` and `risk-low` are unchanged. Merge stays a maintainer decision.

Issue #5504, opened 2026-08-22
bug: [Customers] person and company detail routes reveal whether a UUID exists (403 vs 404)


Kapsik89:
## Summary
For a caller with no access, the customers detail routes answer **403** when the id exists and **404** when it
does not — an existence oracle. The sibling `deals` route already returns a uniform 403.

## Evidence — same caller, only the id varies (2026-08-22, 0.7.0 build)
Caller: a user scoped to another organization, with no grants resolving in context.
`<real>` = an existing record in Acme Corp; `<fake>` = `9b1deb4d-3b7d-4bad-9bdd-2b0d7b3dcb6d`.

| Route | `<real>` (foreign) | `<fake>` (non-existent) | Oracle? |
|---|---|---|---|
| `GET /api/customers/people/{id}` | **403** `Access denied` | **404** `Person not found` | **yes** |
| `GET /api/customers/companies/{id}` | **403** `Access denied` | **404** `Company not found` | **yes** |
| `GET /api/customers/deals/{id}` | 403 `Access denied` | 403 `Access denied` | no ✅ |

## Contrast with the platform's own standard
The api_keys hardening (#4033 / PR-4051) deliberately collapsed every rejection into a uniform response to kill
exactly this signal, and `customers/deals` already follows that shape. `people` and `companies` do not.

Severity is low — UUIDs are not guessable and nothing is disclosed beyond existence — but the inconsistency is
against the project's own precedent.

haxiorz:
🤖 Claimed by a staged **om-fix-issue** run (standard profile). Working the person/company detail existence-oracle fix on branch `claude/om-fix-issue-5504-aaa444`; no PR yet — the change will be left staged for human review. Concurrent automation should back off while `in-progress` is set.

Issue #5114, opened 2026-08-07
bug(customers): company People badge counts a union, the People tab lists only link rows — profile-only people show as "no people match your search" and cannot be removed (404)


a-szczudlo:
The company detail page counts linked people as a **union** of two sources, while the tab that lists them reads **only one** of them. A company whose people are attached via `CustomerPersonProfile.company` (and not via a `customer_person_company_links` row) therefore shows a non-zero badge next to an empty list. Removing such a person is impossible from the UI: the DELETE route 404s, even though core already contains the fallback that would handle it.

## Environment

- `@open-mercato/core` 0.6.6
- Next.js 16.2.12, React 19.2, Node 24
- Self-hosted, Postgres, bare core modules (no custom modules involved)

## Steps to reproduce

1. Create a company `Acme Sp. z o.o.`
2. Create a person `Jan Nowak` and set the person's **profile company** to Acme — i.e. the path that writes `customer_person_profiles.company_id` without inserting a `customer_person_company_links` row. This is reachable from the person form / person import; it is also the shape a lot of migrated data lands in.
3. Open the company detail page `/backend/customers/companies/<id>`.
4. Look at the **People** tab badge, then open the tab.
5. Try to detach the person from the company (People tab → remove).

### Actual

- Step 4: the badge reads **1** (or **2** with one link-backed person plus one profile-only person), the list renders **"No people match your search."**
- Step 5: `DELETE /api/people/<personId>/companies/<companyId>` returns **404 `Person-company link not found`**. There is no way to clear the association from the company UI.

### Expected

Badge count and list content come from the same query. Whatever the badge counts, the tab lists — and can detach.

## Root cause

Two routes inside core disagree about what "a person of this company" is.

**Counting path (union).** `packages/core/src/modules/customers/api/companies/[id]/route.ts:808-876` under `include=people` builds `relatedPeopleById` from *both* sources:

- `:817-850` — active `CustomerPersonCompanyLink` rows;
- `:852-874` — `CustomerPersonProfile` where `{ company: company.id }`, added when not already present.

and the badge is fed straight from that union at `:919-920`:

```ts
const peopleCountQuery = includePeople
  ? Promise.resolve(relatedPeople.length)
  : (async () => { /* link rows only */ })()
```

(note the non-`include=people` branch, `:921-940`, counts link rows only — so the same endpoint returns two different counts depending on the `include` parameter.)

[...]

PR #5700, opened 2026-08-27
docs(specs): Deal People tab — linked-people parity

WojciechSoczynskiTHEY:
## Summary

The Deal detail **People** tab and the Company detail **People** tab render two different components with very different capabilities. Company uses `CompanyPeopleSection` — person cards with **Unlink**, a **Filters** toggle, a **sort** selector, server-side search, pagination and an **Add person** dialog that creates the contact already linked. Deal uses `DealLinkedEntitiesTab` — a flat list of link rows with a single search box: no unlink, no filters, no sort, no inline create.

This is **spec only**. It proposes closing the gap by extracting the company section's shell into a shared `LinkedPeopleSection` driven by thin host-specific wrappers, so both tabs share one implementation and future improvements land on both at once — rather than porting each feature twice.

Implementation will ship on its own PR referencing this spec.

## Changes

- New spec `.ai/specs/2026-08-27-deal-people-tab-linked-people-parity.md`
- One row added to the Pending Specifications table in `.ai/specs/README.md`

No code, no schema, no locale changes in this PR.

## Specification

- [ ] Yes
- [x] No (created a new spec)

**Spec file path:**
`.ai/specs/2026-08-27-deal-people-tab-linked-people-parity.md`

### What the spec covers

- **Scope / non-goals** — the Deal Companies tab keeps `DealLinkedEntitiesTab`; roles and decision makers are explicitly out of scope for the deal tab.
- **Architecture** — one parameterised `LinkedPeopleSection` with no host branching; messaging and optimistic locking stay with the host so the two very different write paths (per-link `POST`/`DELETE` for company, whole-set `PUT /api/customers/deals` for deal) fit under one UI.
- **API** — `GET /api/customers/deals/{id}/people` widened **additively** with the fields `PersonCard` renders; profile fields resolved in one batched query, not per row; `openApi` updated. No new endpoint — the rationale for reusing the deal update path (single aggregate, single optimistic lock, single audit trail) is recorded.
- **Security & tenant isolation** — added profile read scoped by the deal's own `tenantId`/`organizationId` via `findWithDecryption`; the existence-oracle behaviour from #5504 (cross-org read denied as *not found*, not `403`) is called out as something the widening must preserve verbatim.
- **Backward compatibility** — per-surface table; everything additive, nothing removed or renamed, so no deprecation protocol is required.
- **Risks** — five entries with concrete failure scenarios, including the extraction regression risk around the two company refresh events (`customers.person_company_link.deleted` plus the `customers.person.company_assignment.detached` sibling from #5114).
- **Rollout / rollback** — no migration, no flag, plain revert.
- **Testing** — the unit / API / integration coverage required to ship the implementation, including that the existing `CompanyPeopleSection.test.tsx` must pass **unmodified** as the extraction regression gate.

## Testing

Documentation-only change; no runtime behaviour is affected by this PR.

- `git diff --check` — clean
- Spec filename follows `.ai/specs/AGENTS.md` (`{YYYY-MM-DD}-{kebab-case}.md`, no `SPEC-*` prefix)
- Spec content checklist from `.ai/specs/AGENTS.md` covered (TLDR, Overview, Problem Statement, Proposed Solution, Architecture, Data Models, API Contracts, Risks & Impact Review, Final Compliance Report, Changelog)
- Directory table in `.ai/specs/README.md` updated as required by `CONTRIBUTING.md` → Spec Driven Development

## Checklist

- [x] I updated documentation, locales, or generators if the change requires it. — spec directory table updated; no locales or generators involved.
- [ ] I added or adjusted tests that cover the change. — N/A, spec-only PR. The spec itself enumerates the tests required from the implementation PR.
- [ ] I added or updated integration tests in `.ai/qa/tests/` — N/A here; the spec requires a Playwright case in the implementation PR.
- [ ] QA routing set.

> Label note: this PR is opened from a fork by a `read`-permission contributor, so the labels above cannot be applied by the author. Suggested taxonomy: `documentation`, `priority-medium`, `risk-low`, `skip-qa` — spec-only, no UI surface, database structure and API surface unchanged, `BACKWARD_COMPATIBILITY.md` contracts untouched. Happy to adjust if maintainers prefer different routing.

### Design System Compliance

Not applicable — no UI code in this PR. The spec records the DS constraints the implementation must meet (reuses `PersonCard`, no hardcoded status colors, no arbitrary values, all copy through locale keys).

## Linked issues

None yet. Per the spec-driven flow a claim issue will be opened against the merged spec before implementation starts.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5700"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790422575&installation_model_id=432243&pr_number=5700&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5700&signature=6fd097fdb52be6a467f26ba6e2b435b53ab4ce95d7fbf26b69a333159e9a622c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T11:48:47Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Specification review: Deal People tab — linked-people parity

## 🎯 Summary

This is a spec-only PR: two markdown files (`.ai/specs/2026-08-27-deal-people-tab-linked-people-parity.md`, plus one row in `.ai/specs/README.md`), no code, no schema, no locales. It was reviewed with the specification-review path — the spec was read end to end and every claim it makes about the codebase was verified against the PR head in an isolated worktree.

The document is genuinely good in the ways that usually go wrong. It follows the `.ai/specs/AGENTS.md` skeleton, the filename convention and the README directory table; it correctly declares the extraction regression risk around the two company refresh events from #5114; and it correctly preserves the existence-oracle behaviour from #5504. The verifiable structural claims hold up: `CompanyPeopleSection.tsx` (843 lines) really does own search, Filters, sort, `PersonCard`, pagination, the link dialog and the starred-people preference; `/api/customers/people` really does already accept `excludeLinkedDealId` (`api/people/route.ts:67`) while `createPersonLinkAdapter` really does only expose `excludeLinkedCompanyId` (`components/linking/adapters/personAdapter.tsx:36`); `CreatePersonDialog` really does require `companyId`/`companyName` and rely on `syncLegacyPrimaryCompanyLink`; and the eleven fields the spec proposes adding are exactly the ones the company endpoint already returns.

What does not hold up is the write path the design is built on. The spec's central API argument — that routing unlink through `PUT /api/customers/deals` gives it "one aggregate, one lock and one audit trail", and that a per-link `DELETE` was rejected because it "would bypass the deal's optimistic lock" — is inverted by what `syncDealPeople` actually does. That produces two blocking findings below, both of which are about the deal write path rather than about the extraction, and both of which are fixable inside the spec.

**Verdict: ❌ Changes requested** — 2 blockers and 1 major. Under this repository's stricter verdict rule any Medium/minor finding would already force this outcome; the blockers make it unambiguous.

## 🧪 Validation gate

`Runner: local (no compose app container running)`. This PR changes two markdown files, so the specification-review path applies: only the documentation-applicable checks were run and the compile/test gate was skipped rather than reported as passing.

| Command | Result |
|---|---|
| `git diff --check` | ✅ clean |
| Spec filename convention (`{YYYY-MM-DD}-{kebab-case}.md`, no `SPEC-*` prefix) | ✅ conforms |
| `.ai/specs/README.md` Pending Specifications row | ✅ present, well-formed, consistent with its siblings |
| Related-spec link targets resolve on disk | ✅ all three resolve |
| `yarn build:packages` (×2), `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app` | ⏭️ skipped — a markdown-only diff has nothing to compile, index or test |

CI note: the two workflow runs for head `672b63a3` (`CI for Develop&Main`, `Mutation tests`) are sitting at `action_required` because this is a fork PR awaiting maintainer approval, and `license/cla` is `pending` because the CLA is not signed yet. Neither is a failing check and neither was waited on.

## 💥 What can go wrong

### ⛔ Blocker — the optimistic-lock mitigation for concurrent unlink does not actually hold

Risk 2 says a concurrent unlink is mitigated because "unlink routes through the optimistic-locked deal update; a 409 raises the conflict bar", and § API Contracts → *No new endpoint* rejects a per-link `DELETE` partly because it "would bypass the deal's optimistic lock — two writers could unlink concurrently without a 409". The chosen path has that same hole.

The lock token is `customer_deals.updated_at` (the CRUD factory auto-registers the reader for `customers.deal`; `packages/core/src/modules/customers/di.ts:61` hand-wires only the company/person readers, and the guard in `packages/shared/src/lib/crud/optimistic-lock.ts` compares the header against whatever that reader returns). A `personIds`-only update never advances that column:

- `updateDealCommand`'s scalar phase (`commands/deals.ts:~760-808`) only assigns fields the payload actually carries, and `handlePeopleAssociationsChange` sends `{ id, personIds }` and nothing else (`backend/customers/deals/[id]/hooks/useDealAssociations.ts:273`).
- `syncDealPeople` (`commands/deals.ts:385-440`) touches only `CustomerDealPersonLink` rows — `em.nativeDelete` plus recreate. It never dirties the `CustomerDeal` entity.
- `CustomerDeal.updated_at` is declared `onUpdate`-only (`data/entities.ts:429`), so it is stamped only when the deal itself enters the change set, and `afterMutationSuccess` in the lock guard is an explicit no-op (`optimistic-lock.ts:383`).

Failure scenario: A and B both open the deal with `updatedAt = T`. A unlinks Ada; the write succeeds and `updated_at` is still `T`. B unlinks Bob; B's header still reads `T`, still matches, and B's whole-set `personIds` payload — computed from B's stale `linkedPersonIds`, which still contains Ada — reinstates Ada. No 409 is raised and neither user is told. That is precisely the "last write wins and one link returns" outcome Risk 2 claims is mitigated to Low residual.

This matters more under the proposed design than today, because the spec turns "open a modal, deselect, confirm" into a one-click per-card action, which is exactly the interaction pattern that produces interleaved writes.

Please resolve it in the spec rather than leaving it for implementation to discover. The options worth costing: touch the deal's `updated_at` inside the people/companies sync so the aggregate lock means what the spec says it means; move to a delta payload (`addPersonIds` / `removePersonIds`) so two disjoint unlinks commute instead of racing; or reconsider the per-link `DELETE`, whose rejection rationale is currently resting on a guarantee the alternative does not provide either. Whichever is chosen, the § API Contracts → *No new endpoint* paragraph and Risk 2's severity/residual columns both need rewriting.

### ⛔ Blocker — every people write resets `linkedAt` for every remaining person, and the spec makes `linkedAt` user-visible

`syncDealPeople` unconditionally runs `await em.nativeDelete(CustomerDealPersonLink, { deal })` and then recreates a fresh link row for every id in the payload (`commands/deals.ts:429-440`). `CustomerDealPersonLink.createdAt` is `onCreate` (`data/entities.ts:460`), and the deal people endpoint serialises it as `linkedAt` (`api/deals/[id]/people/route.ts:130`).

Today that is invisible: `DealLinkedEntitiesTab` renders only `{ id, label, subtitle }` and offers no recency sort, so nothing on screen depends on `linkedAt`. This spec changes both halves of that at once — Goals adds a "recently linked" sort, and `PersonCard` renders the linked date directly (`components/detail/PersonCard.tsx:70`, `formatDate(person.linkedAt ?? person.createdAt)`).

Failure scenario: a deal has twelve contacts linked over six months. A user unlinks one. The single `PUT` deletes all twelve link rows and recreates the surviving eleven, so all eleven now read "linked today" on their cards, and the "recently linked" sort — which `sortItems` computes from `link.createdAt` (`api/deals/[id]/people/route.ts:54-60`) — degenerates into an arbitrary tie-break by label. The user sees the entire relationship history of the deal reset by an unrelated single-row action, with no error and nothing to undo.

The spec needs to say how `linkedAt` is preserved before it can promise a linked date on the card and a recency sort. The natural fix is to make `syncDealPeople` diff the set — delete only removed links, insert only added ones, leave untouched rows alone — which also happens to be the same change that makes the blocker above tractable. Please add it to § Architecture and to the Implementation Breakdown, and add a regression test for it to § Testing; right now no listed test would catch this.

## 🔁 Backward compatibility

The BC table is broadly right, and the ADDITIVE classification for the response widening is correct — every existing key on `GET /api/customers/deals/{id}/people` is retained with unchanged semantics, so `DealLinkedEntitiesTab` and any third-party consumer keep working. `BACKWARD_COMPATIBILITY.md` § API Routes is satisfied for that row, and no deprecation protocol is required for it.

One row is misclassified.

### ⚠️ Minor — `CreatedPersonSummary` widening is not additive for consumers

The table's `CreatePersonDialog` row reads "`companyId`/`companyName` required → optional; `CreatedPersonSummary.companyId`/`companyName` optional — ADDITIVE for existing callers, which always pass them". The justification only covers half of what the row proposes. Relaxing the **props** from required to optional is genuinely safe: every existing call site keeps compiling. Relaxing the emitted **`CreatedPersonSummary`** (`components/detail/CreatePersonDialog.tsx:28-32`, where both fields are currently `string`) narrows what every `onPersonCreated` consumer is guaranteed to receive, so any consumer holding `const id: string = created.companyId` stops type-checking. That is a consumer-facing break, not an additive change, and it should not be filed under the same reasoning as the props.

Either keep the summary fields always emitted as `string | null` — which costs the dialog nothing, since it already knows whether a company context was supplied — or classify the row honestly and name the migration note. Worth noting `CompanyPersonSummary` needs no such treatment: it already declares everything but `id` and `displayName` optional (`CompanyPeopleSection.tsx:32-46`), so the proposed `LinkedPersonSummary` alias really is shape-compatible.

## 🧩 What's missing

### ❌ Major — the required Playwright case is pointed at a directory that is reserved against exactly that

§ Testing → *Integration / UI* says "Per `.ai/qa/AGENTS.md`, a Playwright case under `.ai/qa/tests/`". The document it cites says the opposite, twice: line 74 marks `tests/` as "Playwright config only — do not place specs here", and line 182 states "Do not add executable `.spec.ts` files under `.ai/qa/tests`; that directory is reserved for the shared Playwright config." The preferred location is the module's own folder (line 78), which for this change is `packages/core/src/modules/customers/__integration__/` — where the module's existing integration specs already live (for example `__integration__/TC-CRM-2453.spec.ts`).

This is worse than a wrong path because of how it fails: a spec dropped into `.ai/qa/tests/` is not discovered by `yarn test:integration`, so the PR goes green with zero integration coverage for the tab and nothing reports a problem. Please correct the path to `packages/core/src/modules/customers/__integration__/TC-CRM-<id>.spec.ts` and pick a `TC-` id that is free on `develop` as well as on the implementation branch.

### ⚠️ Minor — the gate list is not this repository's gate, and it omits the checks this change is most likely to trip

§ Testing → *Gate* names `yarn generate`, `yarn typecheck`, `yarn lint`, `yarn test`, `yarn build:app`. The configured CI-mirroring gate (`.ai/agentic.config.json` → `validation.commands`) is `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app`. The differences are not cosmetic: Implementation Breakdown step 5 adds locale keys to `packages/core/src/modules/customers/i18n/`, and `yarn i18n:check-sync` / `yarn i18n:check-usage` are precisely the gates that catch locale drift — they are the ones an implementer will otherwise discover in CI rather than locally. The repeated `yarn build:packages` is also load-bearing: the second run rebuilds against what `yarn generate` produced. Please quote the configured list.

### ⚠️ Minor — no refresh-event story for the deal wrapper

Risk 1 is careful and correct about the company side: both `customers.person_company_link.deleted` and the `customers.person.company_assignment.detached` sibling from #5114 must stay wired (`CompanyPeopleSection.tsx:333-337`). The deal wrapper gets no equivalent discussion. After this change the company tab still refreshes live when another viewer detaches a person, and the deal tab — same shell, same page family — will not, including when the deal's people are changed from the Manage-links dialog in another session. For a spec whose stated goal is that "the two tabs behave identically", please state explicitly whether the deal side subscribes to anything (a deal-updated event, or nothing) and why.

### ⚠️ Minor — the Goals list overstates what the deal tab is missing

TLDR describes the deal tab as "a flat list of `next/link` rows with a single search box", and Goals lists "server-side search, pagination" among the capabilities it *gains*. Both already exist: the People tab passes `loadLinkedPage={loadLinkedPeoplePage}` (`backend/customers/deals/[id]/page.tsx:543`), and that callback issues a paged, `sort`-aware, `search`-aware request against `/api/customers/deals/{id}/people` (`hooks/useDealAssociations.ts:207-232`), with `linkedItems={data.people}` seeding the first render. The Problem Statement is accurate — what is genuinely absent is per-person unlink, the Filters affordance, the sort control, `PersonCard` rendering and inline create. Please trim the Goals list to those, so an implementer does not rebuild working paging and search.

### ⚠️ Minor — the sibling-tab inconsistency this creates is not recorded

Non-Goals correctly scopes the Deal **Companies** tab out, and keeping `DealLinkedEntitiesTab` for it is the right call for this change's size. The consequence is that the deal detail page will ship two adjacent tabs with sharply different UX — People with cards, filters, sort and unlink; Companies as a flat list. That may well be the accepted trade-off, but a parity spec should say so in Risks & Impact Review and state whether Companies is expected to follow, so the asymmetry reads as a decision rather than an oversight.

## 📈 How can this specification be improved

### 💡 Nit — the API test directory does not match the route under test

§ Testing → *API* points at `api/deals/[id]/__tests__/`. The route being widened is `api/deals/[id]/people/route.ts`, so its tests belong in `api/deals/[id]/people/__tests__/`, which does not exist yet. The existing `api/deals/[id]/__tests__/route.existenceOracle.test.ts` covers the deal **detail** route, not the people route — worth naming precisely, since the spec asks the new tests to re-prove that same existence-oracle behaviour on a different handler.

### 💡 Nit — worth bounding the batched profile query

The batched `$in` profile query is the right call and Risk 3 is well-aimed. One detail to record: the current handler materialises every link row and then filters, sorts and paginates in memory (`api/deals/[id]/people/route.ts:113-144`), so the batch is keyed by *all* linked person ids rather than the page's twenty. That is fine at realistic deal sizes, but since the spec is explicit about avoiding an N+1 it is worth one sentence stating the expected upper bound on contacts per deal, so a future reader knows the in-memory shape was considered rather than missed.

## ✂️ Is this the simplest possible solution

Extraction over porting is the right call and the spec argues it well. Keeping messaging and optimistic locking with the host is the load-bearing decision that makes one shell viable across two very different write paths, and the parameterised-not-branched constraint ("it contains no `if (isDeal)`") is the kind of thing worth writing down before implementation rather than after. The generic parameters on `LinkedPeopleSectionProps` are justified for the reason given — keeping `LinkEntityAdapter<PersonDetails, PersonLinkSettings>` assignable without widening to `unknown`. No simpler alternative was found that delivers the stated outcome; nothing here needs rethinking at the architecture level.

The one place simplicity is worth revisiting is the write path, and it points the same direction as both blockers: a set-diffing `syncDealPeople` is a smaller change than any of the alternatives to the whole-set `PUT`, and it fixes the lock semantics and the `linkedAt` reset at the same time.

## 🧪 Test coverage

§ Testing is unusually thorough for a spec, and the strongest thing in it is making the existing `CompanyPeopleSection.test.tsx` pass **unmodified** the explicit regression gate for the extraction — that file exists (`components/detail/__tests__/CompanyPeopleSection.test.tsx`) and is the right gate to pin. The seven enumerated `DealPeopleSection` cases cover the new surface sensibly, and the API section correctly requires the #5504 existence-oracle behaviour to be re-proved on the widened payload.

Two gaps, both following from the blockers: nothing in the list would catch the `linkedAt` reset, and nothing exercises two interleaved writes against the deal's optimistic lock. Both deserve named cases once the write path is settled.

None — no unaddressed reviewer comments were on the PR. The only pre-existing comments are the dokploy preview-deployment notice and the CLA-assistant bot; neither carries an actionable code ask. Inline diff comments were read via the review-comments API and there were none.

## ⚠️ Merge prerequisites outside this review

- **The CLA is not signed.** `license/cla` is pending and the PR checklist marks the CLA box as accepted, so these disagree. The signature has to land before this can merge.
- **CI has not run.** Both workflow runs for head `672b63a3` are at `action_required` — the standard fork gate — so a maintainer needs to approve them. Nothing about that reflects on the branch.
- **Labels.** As the PR body notes, a `read`-permission contributor cannot self-label. The taxonomy has been applied on the author's behalf and the reasoning is in the separate label-rationale comment; it lands on `priority-low` rather than the suggested `priority-medium`, purely because the repository's inference rule puts documentation changes there.

---

Thanks for a carefully researched spec — the codebase grounding is better than most, and the two blockers are both in the one area a document review is uniquely good at catching, where a stated guarantee and the code that would have to provide it disagree. Neither requires a redesign; a set-diffing `syncDealPeople` addresses both.

pkarw:

_Updated after the re-review of `454c8b531`. The taxonomy is unchanged from round one; only the reason behind `changes-requested` has moved._

- ❌ `changes-requested` — round one's ten findings are all resolved, but the head now conflicts with `develop` in `.ai/specs/README.md` (both Pending Specifications rows belong in the merged table), and the re-review found one new major: making `syncDealPeople` set-diffing removes the delete-everything-first shape that currently makes the `customer_deal_people_primary_uq` partial unique index unreachable, so the spec's claim that `isPrimary` reconciliation is "unchanged" no longer holds. Two minors accompany it.
- 📚 `documentation` — the diff is still two markdown files, a spec under `.ai/specs/` plus one row in the `.ai/specs/README.md` directory table, with no code, schema, locale or generator change.
- 🔽 `priority-low` — this remains a design document with no runtime surface, which the repository's priority-inference rule places at low; the urgency belongs to the implementation PR that will follow it, not to the spec.
- 🟢 `risk-low` — a markdown-only change has no blast radius: nothing compiles, no contract surface moves, and a revert is a plain file deletion.
- ⏭️ `skip-qa` — there is no UI to exercise, the database structure and API surface are unchanged, and no `BACKWARD_COMPATIBILITY.md` contract is touched, so the automated-verification exemption applies. The implementation PR that follows this spec will need `needs-qa` in its own right.

pkarw:
Thanks @WojciechSoczynskiTHEY — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pkarw:

The specification review is posted as a formal review on this PR. It found two blockers and one major finding, all of them fixable inside the spec document rather than requiring a redesign. Both blockers concern the deal write path rather than the extraction: `syncDealPeople` never advances `customer_deals.updated_at`, so the aggregate optimistic lock named as Risk 2's mitigation does not actually detect two concurrent unlinks, and it deletes and recreates every link row on every write, which resets the `linkedAt` value this spec proposes to render on `PersonCard` and sort by. A set-diffing `syncDealPeople` addresses both. The major finding is that the required Playwright case is pointed at `.ai/qa/tests/`, which `.ai/qa/AGENTS.md` explicitly reserves for the shared config — a spec placed there is never discovered, so the PR would go green with no integration coverage. Six minor findings and two nits are listed alongside them.

autofix: skipped (not my PR — re-run with --autofix to fix it here)

No CI follow-up is owed from this run. Both workflow runs for head `672b63a3` are at `action_required` — the standard fork approval gate — so they need a maintainer to release them and will not settle on their own; `ci-monitoring` was deliberately not applied rather than promise a follow-up that could not arrive. The unsigned CLA (`license/cla`, pending) remains a separate merge prerequisite.

WojciechSoczynskiTHEY:
Thanks — this is an unusually good review, and both blockers are real. I verified every finding against `develop` before touching the spec rather than taking them on trust; all of them hold. Pushed `d243c7ea8`.

## Blockers

**The lock mitigation didn't hold — confirmed, and worse than the review states.**

Verified: `updateDealCommand`'s scalar phase assigns only fields the payload carries, so `{ id, personIds }` dirties nothing; `syncDealPeople` touches only `CustomerDealPersonLink`; `CustomerDeal.updatedAt` is `onUpdate`-only. The token never moves on a people-only write.

The one thing I'd add: **set-diffing alone does not fix this**, so the two suggested options aren't interchangeable. With A and B both at `{Ada, Bob, Cid}`@`T` — A unlinks Ada, B's header still reads `T`, and B's stale whole-set payload `{Ada, Cid}` diffed against the current `{Bob, Cid}` still *re-adds Ada*. Set-diffing changes which rows move; it doesn't stop the stale payload from winning. So the spec now requires **both** corrections, and says why one without the other is insufficient. A no-op write (identical set) must not bump the token, or every idle save invalidates other sessions.

I also rewrote the *No new endpoint* paragraph rather than patching it. Its concurrency argument was simply wrong — until the stamp lands, the whole-set path offers no more protection than a per-link `DELETE` would. The decision now rests on surface area and undo semantics, which is what actually justifies it.

**`linkedAt` reset — confirmed.** `nativeDelete` + recreate, `createdAt` is `onCreate`, the route serialises it as `linkedAt`, and this spec is what makes it visible (card date + recency sort). Set-diffing is in § Deal Write Path and § Implementation Breakdown step 1, with a named regression test. I noted the same shape exists in `syncDealCompanies` and asked the implementation PR to state explicitly whether it comes along rather than leaving the asymmetry silent.

Both fixes now ship **first**, before the UI, so the tab lands on a correct write path.

## Major

**Playwright path — my error.** `.ai/qa/AGENTS.md` says it twice and I cited the file while contradicting it. Corrected to `packages/core/src/modules/customers/__integration__/TC-CRM-088.spec.ts` (id verified free on `develop`, flagged for re-check on the implementation branch). Your point about *how* it fails is the important one — silent zero coverage with a green PR.

## Minor / nits

- **`CreatedPersonSummary`** — you're right that I filed a consumer-facing narrowing under the props' reasoning. Taking your first option: the fields stay always-emitted, widened to `string | null`. The BC table now separates the props row from the payload row and records that making them optional was considered and rejected.
- **Gate** — replaced with the configured `validation.commands` verbatim, including both `build:packages` runs and the two i18n checks, with a note on why they matter given step 6 adds locale keys.
- **Deal refresh events** — now stated explicitly as a decision, not an omission: the deal side subscribes to nothing. `customers.deal.updated` carries no `clientBroadcast`, and adding it would broadcast every deal write to every client. Recorded as Risk 5 with Medium residual and a follow-up candidate, and I noted that the Risk 2 lock is what makes a stale action fail loudly rather than silently — which is the property that actually matters here.
- **Goals overstated** — trimmed. Verified `loadLinkedPeoplePage` is already paged, `sort`- and `search`-aware. The Problem Statement now says so explicitly so an implementer doesn't rebuild it.
- **Sibling-tab asymmetry** — Risk 6, as an accepted trade-off. No follow-up committed; noted that the shell is host-parameterised so a companies equivalent would be a wrapper, not a rewrite.
- **API test directory** — corrected to `api/deals/[id]/people/__tests__/`, with your point that the existing `route.existenceOracle.test.ts` covers the *detail* route.
- **Batch bound** — added, with the note that the handler filters/sorts/paginates in memory so the batch is keyed by all linked ids, not the page's twenty, and that the in-memory pipeline is the thing to fix first if the assumption breaks.

## Merge prerequisites

- **CLA** — acknowledged, and the disagreement you spotted is real: my commit is authored `w@they.dev`, which isn't linked to this GitHub account, so CLA-assistant can't match it. Being resolved on my side; the checklist box shouldn't have been ticked ahead of the signature.
- **CI** — understood, fork PRs need maintainer approval.
- **Labels** — thanks for applying them. `priority-low` is fine.

One process note: this is a spec-only PR by design, per `CONTRIBUTING.md` → Spec Driven Development. Implementation will ship on a separate PR referencing this spec once it merges.

WojciechSoczynskiTHEY:
Gentle re-review request — the `CHANGES_REQUESTED` review was submitted against `672b63a3e` and the branch has since moved twice, so the blocking state is now two commits stale:

| SHA | What it does |
|---|---|
| `d243c7ea8` | Addresses the review in full — both blockers, the major, all minors and both nits ([detailed reply above](https://github.com/open-mercato/open-mercato/pull/5700#issuecomment-5438735673)) |
| `454c8b531` | Records the `syncDealPeople` blast radius (five call sites across three commands, both execute and undo, all already inside a flush phase) |

`license/cla` is green on the current head. The PR is still spec-only — two markdown files, no code.

No rush on my side; flagging it only because I can't clear the stale `changes-requested` label or use "Re-request review" myself with `read` permission, so the PR would otherwise sit blocked without anyone being notified that it moved.

pkarw:
@WojciechSoczynskiTHEY please fix the conflicts

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-28T07:13:39Z. Other auto-skills will skip this PR until the lock is released.

pkarw (dismissed):
# 🔍 Re-review — Specification review: Deal People tab, linked-people parity

## 🎯 Summary

This is a re-review. The previous `CHANGES_REQUESTED` review was submitted against `672b63a3e`; the head is now `454c8b531`, two commits later. Every finding from that round was re-checked against the new head in an isolated worktree, and the spec was re-read end to end as a fresh document rather than as a diff.

**All ten findings from round one are addressed, and several are addressed better than they were asked to be.** The two blockers in particular: the author independently verified the lock analysis against `develop`, and then correctly showed that my two suggested fixes were not interchangeable — set-diffing alone does *not* close the lost-update race, because a stale whole-set payload diffed against current state still reinstates the other writer's removal. That is right, and the spec now requires both corrections and says why one without the other is insufficient. The rewritten *No new endpoint* paragraph is also the right call: its old concurrency argument was load-bearing and wrong, and replacing it with surface-area-and-undo is what actually justifies the decision. The new § Blast radius holds up too — `syncDealPeople` really does have five call sites across three commands (`commands/deals.ts:533,629` in `createDealCommand` execute and undo, `:837,988` in `updateDealCommand` execute and undo, `:1107` in `deleteDealCommand`'s undo restore), and all five really do sit inside a `withAtomicFlush` phase list or the update path's `runCrudCommandWrite` phases.

What is new in this round follows from the fix rather than from the original design: making `syncDealPeople` set-diffing removes the delete-everything-first shape that today makes the `is_primary` partial unique index unreachable, and the spec asserts that this reconciliation is "unchanged". It is not. That is the one major below, with two minors and a nit alongside it.

Separately, the head no longer merges cleanly — `develop` has moved since the last push and `.ai/specs/README.md` now conflicts. That is mechanical, but it blocks merge on its own.

**Verdict: ❌ Changes requested** — 1 blocker (merge conflict), 1 major, 2 minors, 1 nit. Under this repository's stricter verdict rule a minor alone would force this outcome; the conflict makes it unambiguous regardless.

## 🧪 Validation gate

`Validation source: GitHub checks (this repository reviews checks-first); local fallback used only for the documentation-applicable checks CI does not report. Runner: local, no compose app container running.`

CI has been released since the last round — the runs are no longer sitting at `action_required`, and `license/cla` is now **green**, which clears one of the two merge prerequisites called out last time.

| Check (head `454c8b53`) | Result |
|---|---|
| `license/cla` | ✅ pass — the CLA is signed |
| `audit`, `audit-scope`, `scope`, `ds-lint` | ✅ pass |
| `prepare`, `lint` | ⏳ in progress — recorded, not waited on |
| `comment`, `mutate`, `[code]smith` | ⏭️ skipped by their own workflow conditions |
| `git diff --check` | ✅ clean |
| Spec filename convention (`{YYYY-MM-DD}-{kebab-case}.md`, no `SPEC-*` prefix) | ✅ conforms |
| `.ai/specs/README.md` Pending Specifications row | ✅ present and well-formed (but see the blocker — it no longer merges) |
| Related-spec links resolve on disk | ✅ all three resolve |
| `yarn build:packages` (×2), `yarn generate`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app` | ⏭️ skipped — a markdown-only diff has nothing to compile, index or test |

No required check is failing. Two are still running; they are disclosed here rather than waited on.

## ✅ Round-one findings — all cleared

| Round 1 finding | Status on `454c8b531` |
|---|---|
| ⛔ Optimistic-lock mitigation does not hold | **Fixed.** § Deal Write Path #2 requires the `updated_at` stamp; Risk 2 rewritten with the correct mitigation and a pinning test; the *No new endpoint* concurrency argument withdrawn rather than patched. The added argument that set-diffing alone is insufficient is correct. |
| ⛔ Every write resets `linkedAt` | **Fixed.** § Deal Write Path #1 requires set-diffing; Risk 3 rewritten; § Data Models records the changed write semantics and the no-backfill consequence; a preservation test is listed. |
| ❌ Playwright case pointed at a reserved directory | **Fixed.** Now `packages/core/src/modules/customers/__integration__/TC-CRM-088.spec.ts`, matching `.ai/qa/AGENTS.md:78,182`. I re-verified the id independently: `TC-CRM-088` appears nowhere on `develop`, where the sequential ids stop at `TC-CRM-087`, so it is genuinely free. |
| ⚠️ `CreatedPersonSummary` BC row misclassified | **Fixed.** The fields stay always-emitted and widen to `string \| null`, with the *optional* variant explicitly rejected as a break. Verified harmless in-repo: all three `onPersonCreated` call sites (`CompanyPeopleSection.tsx:502,621,831`) ignore the payload entirely. |
| ⚠️ Gate list was not this repo's gate | **Fixed.** Quotes `validation.commands` verbatim and explains why the repeated `build:packages` and the two i18n checks are load-bearing. |
| ⚠️ No refresh-event story for the deal wrapper | **Fixed.** Risk 5 plus a Non-Goal. The claim checks out: `customers.deal.updated` exists (`events.ts:21`) with no `clientBroadcast`, unlike the four company-link events that carry it. |
| ⚠️ Goals overstated what the tab lacks | **Fixed.** TLDR and § Problem Statement now say paging and server-side search already exist, and point at `loadLinkedPeoplePage`. |
| ⚠️ Sibling-tab asymmetry unrecorded | **Fixed.** Risk 6, as an accepted trade-off with no follow-up committed — a legitimate answer. |
| 💡 API test directory did not match the route | **Fixed.** `api/deals/[id]/people/__tests__/`, with the distinction from the existing detail-route test spelled out. |
| 💡 Batched query unbounded | **Fixed.** The in-memory pipeline is named, the scale assumption stated, and the thing to fix first if it breaks identified. |

## 💥 What can go wrong

### ⛔ Blocker — the head no longer merges into `develop`

`mergeable` is `CONFLICTING` / `mergeStateStatus` is `DIRTY`. Reproduced locally: `git merge-tree origin/develop HEAD` reports a content conflict in `.ai/specs/README.md`.

The cause is benign. `develop` gained a Pending Specifications row of its own since the last push — "Deal Status `lost` Spelling" (`3e4c4bece`, #5578) — at the same table position where this PR appends its row. Both rows belong in the merged table; the resolution is to keep both, in either order, and change nothing else. Nothing about the spec's content is in dispute.

It is recorded as a blocker because a conflicted head cannot merge and the review reports the head as pushed, not because it reflects on the work.

### ❌ Major — set-diffing collides with the `is_primary` partial unique index, which delete-and-recreate currently makes unreachable

§ Deal Write Path #1 says: "`isPrimary` is still reconciled across the surviving set exactly as today, so the primary-person invariant and its `400` guard are unchanged." The `400` guard is indeed unchanged. The reconciliation cannot be.

`customer_deal_people` carries a **partial unique index**, not a constraint:

```sql
create unique index "customer_deal_people_primary_uq"
  on "customer_deal_people" ("deal_id") where "is_primary";
```

(`data/entities.ts:442-446`, shipped by `migrations/Migration20260723210000_customers.ts:7`.) A unique *index* in PostgreSQL is always immediate — it cannot be deferred to commit time — so any moment at which two rows for one deal both hold `is_primary = true` is an error, even inside a transaction that would end consistent.

Today's whole-set branch cannot reach that state: `em.nativeDelete(CustomerDealPersonLink, { deal })` removes every row before a single insert happens (`commands/deals.ts:429`), so at most one row is ever created with the flag set. Set-diffing removes exactly that protection, and the write ordering then starts to matter.

Failure scenario: a deal has Ada linked and primary. A caller sends `personIds: [Ada, Bob]` with `primaryPersonEntityId: Bob`. Under a set-diff there is nothing to delete — Bob is **inserted** with `is_primary = true`, and Ada is **updated** to `false`. If the insert reaches the database before Ada's update, which is the order a unit of work normally uses, the index rejects it and a request that works today fails with a constraint violation.

That this hazard is real and already known in this codebase is not speculation: the sibling branch of the very same function dances around it. When `personIds` is undefined and only the primary moves, the code clears every flag, calls `await em.flush()`, and only then sets the new primary (`commands/deals.ts:404-410`) — a mid-function flush that exists for no other reason.

The fix is small, and it is a spec-level decision rather than an implementation detail: state that the set-diff must clear displaced `is_primary` flags and flush before setting the new one, mirroring the existing branch. Please replace the "exactly as today" sentence with that ordering requirement, and extend the § Testing bullet — currently "`isPrimary` reconciliation and the `400` primary-must-be-linked guard still hold across a set-diffing update" — to name the case that would actually fail: moving the primary to a **newly added** person in the same write.

### ⚠️ Minor — the stamp rule is keyed to membership, so a primary-flag change is still an unlocked write

§ Deal Write Path #2 requires the deal to be marked dirty "whenever it actually adds or removes a link", and § Testing codifies it: "a `{ id, personIds }` update that changes membership advances `customer_deals.updated_at`; an identical-set update does **not**".

Taken literally that leaves a hole in the aggregate the section just closed. Changing which linked person is primary — same membership, different `is_primary` — is a real mutation of deal state that would not advance the token, so two concurrent primary changes both succeed with no 409 and the later one silently wins. The `personIds === undefined` branch is the same story: it mutates link rows and never touches the deal.

This is narrower than the blocker it descends from, which is why it is a minor. I checked: nothing in the UI sends `primaryPersonEntityId` today — the only non-test reference in the repository is the validator itself (`data/validators.ts:189`), so the field is currently API-only, and this spec's Non-Goals keep `isPrimary` read-only on the new tab. It is still worth one word: make the rule "membership **or the primary flag** changes", so the guarantee holds for whatever sets the primary next rather than only for what sets it today. The "identical-set update does not bump" test then needs its own qualifier, or it will pin the gap in place.

### ⚠️ Minor — calling the `syncDealCompanies` change "optional" conflates its two halves

§ Deal Write Path closes with: "The same reasoning applies to `syncDealCompanies` ... Bringing it along is optional for this change's UI but keeps the two paths honest."

The two corrections do not have equal status there. The **set-diff** half genuinely is optional: `syncDealCompanies` has the same delete-and-recreate shape, but nothing renders a company link date and the Companies tab keeps `DealLinkedEntitiesTab`, so no user-visible value depends on those rows surviving. The **stamp** half is not cosmetic — a companies-only `{ id, companyIds }` write moves the lock token exactly as little as a people-only write does, so the lost-update race this section exists to close stays open on the adjacent field of the same aggregate, and a later reader will have no way to tell why one path is locked and the other is not.

Please split the sentence: the set-diff on companies is deferrable with a note, while the `updated_at` stamp on any link-membership change should be stated as a property of the deal write path rather than of the people path. That also makes the § Final Compliance Report row "Optimistic locking on the new write path ✅" true of the whole command rather than of half of it.

## 🔁 Backward compatibility

Nothing here needs re-litigating. The response widening on `GET /api/customers/deals/{id}/people` remains correctly ADDITIVE — every existing key retained with unchanged semantics, so `DealLinkedEntitiesTab` and third-party consumers keep working, satisfying `BACKWARD_COMPATIBILITY.md` § API Routes. The `CreatedPersonSummary` row is now classified defensibly and names the rejected alternative. The `syncDealPeople` row's BEHAVIOURAL FIX classification is right: no signature or payload change, and a caller that ignored the lock now receives the 409 it should always have received.

One claim I re-verified because the shared-component design leans on it: `sort` really is unchanged. The deal route already accepts `name-asc` / `name-desc` alongside `label-asc` / `label-desc` / `recent` and normalizes the first pair onto the second (`api/deals/[id]/people/route.ts:25,30-39`), so the company section's existing `name-asc` vocabulary (`CompanyPeopleSection.tsx:203`) passes through the shared shell unchanged. No mapping layer is needed and none is specified — correct as written.

## 🧩 What's missing

Nothing new at the section level. The gaps round one identified — refresh events, the sibling-tab asymmetry, the test locations, the gate — are all filled, and § Blast radius is a genuine addition rather than a restatement: the undo-path analysis ("an undo should not re-date links it did not touch") is the right instinct and reaches the right conclusion.

The one thing § Testing still would not catch is the major above. Everything else in that section is well aimed, including keeping `CompanyPeopleSection.test.tsx` passing unmodified as the extraction regression gate.

## 📈 How can this specification be improved

### 💡 Nit — `data/entities.ts:429` points at the wrong entity

§ Deal Write Path #2 cites `data/entities.ts:429` for "`CustomerDeal.updatedAt` is declared `onUpdate`-only". Line 429 is inside `CustomerDealStageTransition` (class at `:393`); `CustomerDeal` starts at `:307` and its `updated_at` is at **`:367`**. The claim itself holds at that line — `@Property({ name: 'updated_at', type: Date, onUpdate: () => new Date() })`, with no `onCreate`, exactly as the argument requires.

This one is mine: round one supplied `:429` and the revision carried it forward in good faith. Worth correcting anyway, since it is the linchpin citation of the section and a reader who checks it lands on a different table.

## ✂️ Is this the simplest possible solution

Unchanged from round one, and the revision strengthens the case rather than complicating it. Extraction over porting is right; parameterised-not-branched is the constraint worth writing down in advance; keeping messaging and locking with the host is what makes one shell viable across two very different write paths. Pulling the write-path fixes into scope was the correct response to the review rather than an expansion of it — § Deal Write Path is a prerequisite for the UI, and Implementation Breakdown step 1 sequences it first for the right reason.

The deferred delta-payload alternative (`addPersonIds` / `removePersonIds`) is well argued and correctly deferred: it is better UX than a 409, but it widens the contract for every existing caller, and a 409 into the existing conflict bar is this repository's established concurrent-edit story. Recording it as revisit-if-contention-appears is the right disposition.

## 🧪 Test coverage

§ Testing is stronger than it was: five command-level cases now cover the write-path fixes, including the interleaved-write case that pins Risk 2 and the create/undo cases that pin § Blast radius. The API section correctly re-proves the #5504 existence-oracle behaviour on the widened handler, and the integration case now lands where `yarn test:integration` will actually discover it.

Two adjustments follow from the findings above: the `isPrimary` reconciliation case needs to name the newly-added-person-becomes-primary scenario, and the lock-token case needs to say what happens on a primary-only change once that rule is settled.

None outstanding. The conversation comments on this PR are the dokploy preview-deployment notice, the CLA-assistant bot (now reporting signed), this skill's own markers from round one, the author's detailed reply and re-review request, and a maintainer comment asking for the conflicts to be resolved — which is the blocker recorded above. There are no inline diff comments. Nothing carries an unaddressed actionable ask beyond what this review already covers.

## ⚠️ Merge prerequisites outside this review

- **The conflict must be resolved.** Recorded as the blocker above; both README rows survive the merge.
- **CI is running, not blocked.** `prepare` and `lint` were still in progress at review time. They are disclosed rather than waited on.
- **The CLA is signed.** This clears the round-one prerequisite.
- **Labels.** `changes-requested` stays and the rest of the taxonomy is unchanged; the reasoning is in the separate label-rationale comment.

---

Thanks — this is a genuinely strong second pass. Correcting my own suggestion, rather than implementing it as given, is the more useful outcome, and the spec is better for it. The remaining major is a direct consequence of that fix and is fixable in two sentences plus a test name; the conflict is a rebase.

pkarw:
Thanks @WojciechSoczynskiTHEY — the re-review is posted against `454c8b531`. All ten round-one findings are cleared, so this is a much shorter list, but there are actionable items so I'm handing the PR back to you for one more pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pkarw:

The specification re-review is posted as a formal review against `454c8b531`. Every one of round one's findings — both blockers, the major, all four minors and both nits — is resolved, and two of them are resolved better than they were asked to be: the analysis that set-diffing alone would not close the lost-update race is correct, and withdrawing the *No new endpoint* concurrency argument rather than patching it was the right call.

What remains is short. One blocker: the head no longer merges — `develop` added its own Pending Specifications row (`3e4c4bece`, #5578) at the same table position, so `.ai/specs/README.md` conflicts and both rows need to survive the merge. One major: set-diffing `syncDealPeople` removes the delete-everything-first shape that today makes the `customer_deal_people_primary_uq` partial unique index unreachable, so "reconciled exactly as today" no longer holds — a payload that makes a newly added person primary while another survives as primary would trip a non-deferrable unique index, and the existing `personIds === undefined` branch already does an explicit clear-then-flush dance for exactly that reason. Two minors and one nit accompany them; the nit is a wrong line citation that round one supplied and this revision carried forward in good faith.

autofix: skipped (not my PR — re-run with --autofix to fix it here)

CI follow-up: `ci-monitoring` was deliberately not applied. `prepare` and `lint` were still running at review time and nothing was failing, but the head has to be rebased to clear the conflict, which will supersede these runs — so a follow-up on this SHA would report on a head that no longer exists. `license/cla` is green, which clears the round-one prerequisite.

WojciechSoczynskiTHEY:
Thanks — all four addressed in `a4dad30f5`. As last round, I verified each finding against `develop` before changing anything; all four hold.

## ⛔ Merge conflict — resolved

Merged `origin/develop` (`a0154ac41`). The conflict was exactly as diagnosed: both Pending rows belong in the table. Kept both, ordered by date — "Deal Status `lost` Spelling" (2026-08-24) then this spec (2026-08-27) — and changed nothing else. `mergeable` is back to `MERGEABLE`, and the PR is still spec-only, two markdown files.

## ❌ Major — `is_primary` — confirmed, and the diagnosis is exactly right

Verified: `data/entities.ts:442-446` declares it via `expression` as `create unique index ... on "customer_deal_people" ("deal_id") where "is_primary"` — a partial unique **index**, so immediate, not deferrable. And the sibling branch really does dance around it: `commands/deals.ts:404-410` clears every flag, `await em.flush()`, *then* sets the new primary — a mid-function flush with no other purpose.

This is the part of the review I'd have been least likely to catch myself: the hazard is invisible in the current code precisely because `nativeDelete` removes every row first, so the protection is incidental rather than designed. Set-diffing silently withdraws it.

The spec now states the ordering requirement rather than claiming reconciliation is unchanged, cites the existing flush as the pattern to mirror, and the § Testing bullet names the case that would actually fail — promoting a **newly added** person to primary in the same write, where the insert carries the flag and the displaced row is only updated.

## ⚠️ Minor — stamp keyed to membership only

Right, and I'd rather close it now than leave a hole in the aggregate the section exists to close. The rule is now "membership **or** the primary flag", with the `personIds === undefined` branch called out as the same story, and the negative test qualified to "neither membership nor the primary" so it stops pinning the gap in place.

I confirmed your point that this costs nothing today — `primaryPersonEntityId` has no non-test caller outside `commands/deals.ts` and the validator, so it is API-only, and this spec keeps `isPrimary` read-only on the new tab. Worth stating anyway so the guarantee holds for whatever sets the primary next.

## ⚠️ Minor — `syncDealCompanies` halves conflated

Agreed, and the asymmetry you describe is the real cost: a later reader would find one link path locked and the adjacent one not, with nothing explaining why. Split into a bullet each — the set-diff is deferrable with a note (nothing renders a company link date), the `updated_at` stamp is not. The stamp is now stated as a property of the **deal write path**, any link membership or primary-flag change on either table, and the § Final Compliance row was reworded to match so it is true of the whole command rather than half of it.

## 💡 Nit — wrong citation

Fixed to `data/entities.ts:367`. Verified the class boundaries: `CustomerDeal` at `:307`, `CustomerDealStageTransition` at `:393`, so `:429` was indeed inside the wrong entity — both happen to declare an identically-shaped `updated_at`, which is why it read as correct. Good catch on the linchpin citation of that section.

---

Head is `a4dad30f5`, CLA green, no conflict. Still spec-only by design — implementation ships on its own PR referencing this spec once it merges.

pat-lewczuk (approved):

.ai/specs/2026-08-27-deal-people-tab-linked-people-parity.md
```diff
@@ -0,0 +1,331 @@
...
+}
+```
+
+The generic parameters exist so the concrete `LinkEntityAdapter<PersonDetails, PersonLinkSettings>` produced by `createPersonLinkAdapter` stays assignable without widening to `unknown`.
+
+**Messaging and locking stay with the host.** The shell never calls `flash` and never builds an optimistic-lock header; it awaits `onUnlink` / `onLinkConfirm` and reloads. This is what lets the two hosts use completely different write paths under one UI.
+
+### Deal Write Path
+
+Both corrections below are in `syncDealPeople` (`commands/deals.ts:385-440`). They are prerequisites for the UI, not incidental cleanup.
```
adeptofvoltron: **Scope.** This half is a live concurrency/data bug on an aggregate with undo semantics — unrelated to UI parity, different blast radius, independently revertable. Step 1 already says it ships first; suggest it ships as its own issue + PR rather than as phase 1 of this one.

adeptofvoltron (commented):
Read it through and checked the load-bearing claims against `develop` — they hold. `syncDealPeople` does delete-and-recreate, `CustomerDealPersonLink.createdAt` is `onCreate`, `customer_deal_people_primary_uq` is exactly as quoted, `CustomerDeal.updatedAt` is `onUpdate`-only, and `useDealAssociations.ts:271` already sends `buildOptimisticLockHeader(deal.updatedAt)` on the people save. So the deal's optimistic lock is decorative today and the lost-update race is live regardless of this UI work. Good find.

Direction is right. Inline notes are what I'd want settled before implementation — only the `is_primary` one is a correctness issue.

.ai/specs/2026-08-27-deal-people-tab-linked-people-parity.md
```diff
@@ -0,0 +1,331 @@
...
+
+Today's delete-everything-first shape makes that state unreachable: every row is gone before any insert happens, so at most one row is ever created with the flag set. Set-diffing removes exactly that protection and makes write ordering load-bearing. Concretely: a deal has Ada linked and primary; a caller sends `personIds: [Ada, Bob]` with `primaryPersonEntityId: Bob`. Nothing is deleted — Bob is *inserted* with `is_primary = true` and Ada is *updated* to `false`. If the insert reaches the database first, which is the order a unit of work normally uses, the index rejects a request that works today.
+
+The set-diff must therefore **clear displaced `is_primary` flags and flush before setting the new primary**, mirroring what the `personIds === undefined` branch already does for exactly this reason (`commands/deals.ts:404-410`, where a mid-function `em.flush()` exists for no other purpose).
```
adeptofvoltron: **Incomplete.** This covers a row that *stays* and loses the flag, not one that is *removed*.

Ada primary, payload `[Bob]` with primary Bob: the set-diff deletes Ada's row and inserts Bob's with `is_primary = true`. The UoW flushes inserts before deletes, so the insert still hits the index while Ada's row exists.

Rule should be: all removals **and** flag clears flush before any insert — or keep `nativeDelete` for the removals, which executes immediately.

.ai/specs/2026-08-27-deal-people-tab-linked-people-parity.md
```diff
@@ -0,0 +1,331 @@
...
+
+Set-diffing alone does **not** close this. Two users open a deal holding `{Ada, Bob, Cid}` at version `T`. A unlinks Ada and sends `{Bob, Cid}`; the write succeeds and `updated_at` is still `T`. B unlinks Bob and sends `{Ada, Cid}` — computed from B's stale view. B's header still reads `T`, still matches, and the diff against the current `{Bob, Cid}` **re-adds Ada** and removes Bob. Neither user is told. Set-diffing changes which rows move; it does not stop the stale payload from winning.
+
+So `syncDealPeople` must also mark the deal dirty whenever it actually changes the links — **membership or the primary flag**. Keying the rule to membership alone would leave the adjacent hole open: moving the primary between two already-linked people mutates deal state without changing the set, so two concurrent primary changes would both succeed and the later one would silently win. The same applies to the `personIds === undefined` branch, which mutates link rows and never touches the deal today. A write that changes neither membership nor the primary must **not** bump the token, or every idle save would invalidate other sessions.
```
adeptofvoltron, line 157: **By what mechanism?** MikroORM won't add the entity to the changeset for a no-op assignment, and assigning `updatedAt` directly collides with the `onUpdate` hook. Worth naming it here so the implementation doesn't guess.

Also unspecified: the undo paths. Does a restore replay the snapshot's `updated_at` or stamp a new one? If it stamps, an undone deal comes back holding a token it never had.

.ai/specs/2026-08-27-deal-people-tab-linked-people-parity.md
```diff
@@ -0,0 +1,331 @@
...
+type LinkedPeopleSectionProps<TDetails, TLinkSettings> = {
+  scopeId: string                                  // scopes the starred-people preference
+  loadPage: (p: { page; pageSize; sort; search }) => Promise<LinkedPeoplePage>
+  onUnlink: (personId: string) => Promise<void>    // host owns messaging + locking
+  linkAdapter: LinkEntityAdapter<TDetails, TLinkSettings>
+  onLinkConfirm: (input: LinkEntityConfirmInput<TLinkSettings>) => Promise<void>
+  header?: React.ReactNode                         // company: RolesSection
+  renderFooter?: (ctx) => React.ReactNode          // company: DecisionMakersFooter
+  refreshKey?: number                              // host-triggered reload
+  // + labels, empty state, guarded-mutation runner
```
adeptofvoltron, line 96: `CompanyPeopleSection.tsx` is 843 lines — roles, decision makers, starred prefs, two refresh events. `header` + `renderFooter` is thin for what actually has to thread through, the link dialog's add-new slot and the guarded-mutation runner especially. Either enumerate the real prop surface here, or say plainly that step 2 is the risky one.

.ai/specs/2026-08-27-deal-people-tab-linked-people-parity.md
```diff
@@ -0,0 +1,331 @@
...
+| 3 | `linkedAt` reset across the whole deal | A single unlink re-dates every surviving link, breaking the linked date on the card and the "recently linked" sort | **High** | Set-diffing `syncDealPeople`; regression test asserts untouched rows keep their `created_at` | Low |
+| 4 | N+1 on profile lookup | A deal with many contacts issues one profile query per person | Medium | Single batched `$in` query keyed by person ids, as the company route already does | Low |
+| 5 | Deal tab does not refresh live | Another session changes the deal's people; the company tab would refresh from a broadcast event, the deal tab will not, so a viewer can act on a stale list until they reload | Medium | Accepted for this change. `customers.deal.updated` exists but carries no `clientBroadcast`, and adding it would broadcast every deal write to every client — a blast radius that deserves its own change. The optimistic lock from Risk 2 makes a stale action fail loudly with a 409 rather than silently, which is the property that actually matters | Medium — documented gap, follow-up candidate |
```
adeptofvoltron, line 269: With Unlink on the card, two users on one deal stops being an edge case — this risk gets heavier than the row implies. The 409 mitigation is fine, but the resulting UX (click Unlink → conflict bar → reload) belongs in the QA script rather than being discovered there.

.ai/specs/2026-08-27-deal-people-tab-linked-people-parity.md
```diff
@@ -0,0 +1,331 @@
...
+| `withAtomicFlush` discipline preserved in `syncDealPeople` | ✅ the set-diff runs inside the existing flush phases; the one added flush is the `is_primary` ordering barrier the partial unique index requires |
+| `BACKWARD_COMPATIBILITY.md` contracts preserved | ✅ additive, plus one behavioural fix that removes a silent lost update |
+| No `packages/enterprise/` change | ✅ |
+| Tests ship with the implementation | ✅ unit + command + API + integration |
+
+## Changelog
+
+| Date | Change |
+|------|--------|
+| 2026-08-27 | Initial specification. |
```
adeptofvoltron, line 388: Rows are out of chronological order — the 2026-08-28 entry sits above the 2026-08-27 round-1 one.

WojciechSoczynskiTHEY:
@adeptofvoltron — thanks, and the `is_primary` one is a genuine correctness gap in my rule, not a nuance. Addressed all six in `5b2e62dfd`.

**`is_primary` — you're right and I had it half-covered.** My rule ("clear displaced flags and flush before setting the new one") only handles a displaced row that *stays*. Your case — Ada primary, payload `[Bob]` with primary Bob — deletes Ada's row and inserts Bob's, and with inserts committing before deletes the insert still hits the index while Ada's row exists. Rewrote it as the invariant instead of one of its cases: **every removal and every flag clear must have reached the database before any insert carrying `is_primary = true`**, with both satisfying strategies named — flush both first, or keep `nativeDelete` for removals since it issues SQL immediately and is ordered correctly by construction. I'd lean toward the `nativeDelete` option in implementation precisely because it removes the ordering question rather than managing it.

**Stamp mechanism — fair, I left it to be guessed.** Named now: an explicit `deal.updatedAt = new Date()` is the trigger that dirties the entity, and `onUpdate` then supplies the committed value, so they don't collide — the assignment is what puts the deal in the change set, not what sets the column. There's precedent in this module: `commands/people.ts:997-999` does exactly this when only profile fields changed and the parent would otherwise stay clean.

**Undo paths — good catch, that was genuinely unspecified.** Settled as: a restore stamps a **new** token rather than replaying the snapshot's. Replaying it would hand the undone deal a version other sessions may already hold, so a stale writer's header would match a record that has since changed underneath it — reintroducing exactly the lost update the section exists to remove.

**Scope — adopted.** The write-path fix now ships as its own issue and PR ahead of the UI work rather than as phase 1 of it. Your reasoning is the deciding one for me: bundling produces a single PR that can't be reverted in halves, and this half is a live data bug on an aggregate with undo semantics that has nothing to do with UI parity.

**Prop surface — agreed, `header` + `renderFooter` was doing too much work in that sentence.** Enumerated what actually threads through (link-dialog open state so the add-new slot can close its host, guarded-mutation runner, loading and refresh hooks, labels, empty state) and said plainly that the extraction is the risky step, with `CompanyPeopleSection.test.tsx` passing unmodified as the gate.

**Risk 2 weight + QA — agreed.** The row now says card-level Unlink turns concurrent writes from an edge case into an ordinary one, and the conflict-bar UX is an explicit QA step rather than something QA discovers.

**Changelog order — fixed.**

Also noting your independent confirmation that `useDealAssociations.ts:271` already sends `buildOptimisticLockHeader(deal.updatedAt)` on the people save. That's the part that makes this worth fixing rather than documenting: the client is already doing its half correctly, so the lock reads as working while being decorative on the server side.

Head is `5b2e62dfd`. Still spec-only.

WojciechSoczynskiTHEY:
Delivery resequenced into three independently revertable PRs — `c94b70508`.

| | Ships | Rationale |
|---|---|---|
| **PR 1** | the `updated_at` stamp, alone | The correction that actually closes the lost-update race. Smallest diff, no index interaction. |
| **PR 2** | UI parity | Depends on PR 1 for safe concurrent unlink. |
| **PR 3** | set-diffing + the `is_primary` ordering invariant | Carries all the index risk in isolation. |

This follows @adeptofvoltron's point about the write-path fix not belonging in the UI PR, and takes it one step further by separating the two write-path corrections from each other — they have different urgency and very different risk.

**One consequence needed handling rather than accepting.** Without set-diffing, `syncDealPeople` still re-dates every surviving link on any people write, and both unlink *and* inline create are people writes. Shipping the card's linked date or the "recently linked" sort in PR 2 would therefore show "linked today" for a contact attached six months ago, from the first interaction onwards — a wrong value in a CRM, not a cosmetic gap.

So **PR 2 deliberately omits both**. `LinkedPeopleSection` gains `availableSorts` (the deal wrapper passes `['name-asc','name-desc']`) and `PersonCard` gains `showLinkedDate` defaulting to `true` so the company tab is untouched. PR 3 drops the override and flips the flag, completing parity in two lines. A PR 2 unit test guards that no `linkedAt`-dependent surface leaks early.

Risk 3 is reframed as **sequenced away** rather than mitigated — residual "none displayed" before PR 3 — and § Testing is split so each PR pins its own behaviour rather than deferring coverage.

Worth being explicit about the trade-off, since it touches the stated goal: the deal tab reaches *behavioural* parity in PR 2 (unlink, filters, name sort, cards, inline create) and *full* parity in PR 3. The two missing surfaces are exactly the two that depend on data PR 3 makes trustworthy.

Head `c94b70508`. Still spec-only.

adeptofvoltron:
Really impressive first contribution — thanks for putting this together. Two things stood out to me:

1. **Pushing back correctly on a reviewer's own suggestion.** When round-one review proposed set-diffing as a fix for the lock issue, you didn't just implement it as given — you worked out that set-diffing alone doesn't actually close the lost-update race (a stale whole-set payload diffed against current state still reinstates another writer's removal), and required both the diff *and* the `updated_at` stamp. Catching that a suggested fix is incomplete, and explaining precisely why, is the kind of rigor that's rare even from experienced contributors.

2. **Pinning the regression gate instead of hand-waving it.** Requiring `CompanyPeopleSection.test.tsx` to keep passing unmodified as the extraction regression check is a small line in the spec, but it's exactly the right kind of guarantee — concrete, verifiable, and future-proof against silent behavior drift during the extraction.

Great work getting this through two tough review rounds — looking forward to the implementation PR.

PR #5707, opened 2026-08-27
fix(sales): reconcile a supplied totalNetAmount against the computed net amount (#5644)

### LINKED ISSUE #5644: Sales line totalNetAmount is accepted and validated, then silently ignored
Surfaced while implementing `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` (contract implementation: #5640). The spec's § Out of Scope records it as a same-root-cause finding with **no upstream issue**, so this is that issue.

## The problem

`SalesLineSnapshot.totalNetAmount` (`packages/core/src/modules/sales/lib/types.ts`) is declared, and the shared line schemas in `packages/core/src/modules/sales/data/validators.ts` accept and validate it — but `buildBaseLineResult` in `lib/calculations.ts` **never reads it**. Only `totalGrossAmount` is honoured:

```ts
const grossSubtotal =
  line.totalGrossAmount !== null && line.totalGrossAmount !== undefined
    ? toNumber(line.totalGrossAmount, 0)     // a supplied gross wins verbatim
    : round(netSubtotal + taxAmount)
```

There is no matching branch for net. A caller that supplies a correct `totalNetAmount` watches it be validated, accepted, and then silently discarded — the engine recomputes net from the unit price and the discount instead.

## Why it matters beyond tidiness

This asymmetry is not cosmetic; it is what made the discount defect in #3757 *detectable but invisible*. Because a supplied gross is kept verbatim while net is recomputed, a line with a mis-read discount ends up with `total_net_amount × (1 + taxRate)` diverging from `total_gross_amount` — a self-evident inconsistency that nothing in the platform flags. As the spec puts it: a schema that rejected unused fields would have surfaced this family of bugs years earlier.

## Worth deciding

Three defensible directions, and the point of this issue is to pick one rather than leave the field in limbo:

1. **Honour it** — give net the same "supplied value wins" branch gross has. Symmetric, but it lets a caller post a net that contradicts its own unit price and quantity.
2. **Reject it** — drop `totalNetAmount` from the line schemas so a caller supplying it gets a 4xx instead of silence. Cleanest signal, but a breaking change to the request contract that needs the deprecation protocol.
3. **Validate and warn** — keep accepting it, but reconcile it against the computed value and log or surface a mismatch. No contract change; catches the inconsistency class that hid #3757.

Whichever is chosen, the same question applies to `totalGrossAmount` being honoured verbatim: it is the other half of the asymmetry.

## Context

- Spec: `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` → § Out of Scope → "Adjacent: `totalNetAmount` is accepted, validated, then ignored"
- Contract implementation that surfaced it: #5640
- The defect it helped hide: #3757

adeptofvoltron:
Closes #5644

Status: complete

## 🎯 Goal
- A **caller-supplied** `SalesLineSnapshot.totalNetAmount` is no longer silently discarded — a mismatch between the supplied value and the engine-computed net amount is surfaced via a structured warning instead of vanishing without a trace, while the platform's own rehydrated lines stay silent so the new signal means something.

## 🔍 Problem
`SalesLineSnapshot.totalNetAmount` is declared in `lib/types.ts` and accepted/validated by the shared line schemas in `data/validators.ts`, but `buildBaseLineResult` in `lib/calculations.ts` never read it — only `totalGrossAmount` had a "supplied value wins" branch. A caller supplying a correct `totalNetAmount` watched it be validated, accepted, and then recomputed over and discarded. This asymmetry is what made the discount defect in #3757 detectable-but-invisible: because gross is kept verbatim while net is recomputed, a line with a mis-read discount ends up with a net/gross relationship that contradicts itself, and nothing in the platform flagged it.

## 🔍 Root Cause
`buildBaseLineResult` computes `netSubtotal` purely from `unitPriceNet` and the resolved discount, with no equivalent of the `line.totalGrossAmount !== null && line.totalGrossAmount !== undefined` branch used for gross. There is no code path that reads `line.totalNetAmount` at all.

## What Changed
- **`lib/calculations.ts`** — `buildBaseLineResult` reconciles a caller-supplied `line.totalNetAmount` against the computed `netSubtotal`. The computed value stays authoritative (no behavioral/output change — this is deliberately **not** "honour it verbatim" like gross, since that would let a caller mask a broken discount calculation, the exact failure class that hid #3757). On a divergence it logs a `sales`-namespaced `warn` via the structured logging facade (`createLogger('sales')`) with the line id, product id, supplied amount, and computed amount.
- **Only a caller's value is reconciled.** Every persisted-line mapper fills `totalNetAmount` from the stored column through `toNumeric()`, which never yields `null`, so an origin-blind guard is *always* true on recalculation. For a legacy row that divergence is #5640's discount contract healing itself — warning about it would emit one line per line per recalculation, indefinitely, and bury the caller signal this exists for. A new `totalsFromStoredRow` flag on `SalesLineSnapshot` marks the stored origin: it is a **sibling** of `discountAmountFromStoredRow`, not an overload of it, because `lineSnapshots.ts` is emphatic that a mapper must not answer one origin question with the other's field. It is set by `mapPersistedLine`, carried independently through `createLineSnapshotFromInput` (a line upsert can take its discount from the caller while its totals still come off the row), derived per upsert by the new `resolveUpsertTotalsOrigin`, and — like its sibling — is mapper-only: the request schemas strip it, so a caller cannot suppress the warning about its own value.
- **The comparison carries a tolerance.** `NET_RECONCILIATION_TOLERANCE` (0.005, half a minor unit) replaces exact equality: the engine rounds to the 4 decimals the numeric columns hold, but callers work in money at 2, and reporting half a cent of honest rounding as a discrepancy is how a diagnostic gets filtered out and forgotten.
- **A non-finite supplied value warns explicitly.** The previous `toNumber(line.totalNetAmount, netSubtotal)` fallback meant a `NaN` or a non-numeric value coerced to the computed net, compared equal, and logged nothing — the same silent discard this issue is about, one level down.
- **`lib/__tests__/calculations.test.ts`** — the `buildBaseLineResult net amount reconciliation (issue #5644)` suite now covers the caller divergence (pinning `lineId` and `productId`, which are the whole operational value of the log line), the matching and absent cases, both sides of the rounding tolerance, the non-finite value, a rehydrated line built through `mapOrderLineEntityToSnapshot` whose stored net the discount contract is healing, and the `NOT NULL DEFAULT` zero a persisted row carries. A shared `jest.mock('@open-mercato/shared/lib/logger', ...)` block (matching the existing pattern in `auth/lib/__tests__/consentIntegrity.test.ts`) spies on `warn`.
- **`lib/__tests__/lineSnapshots.test.ts`** — the mappers are asserted to mark rebuilt totals as stored-row sourced, `resolveUpsertTotalsOrigin` is covered operand by operand (including the explicit caller zero), and the request schemas are asserted to strip `totalsFromStoredRow`.

## ✂️ Scope: the gross half of the asymmetry
#5644 closes by noting that `totalGrossAmount` being honoured verbatim is the other half of the same asymmetry. That is **deliberately out of scope here** and now tracked in #5853 rather than left as a silent omission. It is not a mechanical repeat of the net change: several call sites depend on a supplied gross winning (the branch immediately below derives tax back out of the gross for a tax-class-priced line whose resolved rate was not persisted), so changing it needs its own behavioural analysis.

## 🧪 Tests
Run locally (no compose `app` container is running on this machine, so the gate ran in **local mode**), on the tree merged with the current `develop`:

- `yarn workspace @open-mercato/core test -- packages/core/src/modules/sales/lib/__tests__/` — 105/105 passed across the 9 sales `lib` suites (61 of them in the two files this PR touches).
- **Negative control**: the pre-fix reconciliation block was restored and the suite re-run — exactly the four new guards failed (the rehydration case, the stored-zero case, the tolerance case, and the non-finite case), and no others. They lock in real behaviour rather than restating it.
- Full validation gate, in the configured order — `yarn build:packages`, `yarn generate`, `yarn build:packages`, `yarn i18n:check-sync`, `yarn i18n:check-usage`, `yarn typecheck`, `yarn test`, `yarn build:app`: `build:packages`, `generate` (no generated churn), `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck` and `build:app` all exited 0. `yarn test` needs a caveat about *this machine*, not about the branch: run through turbo, the aggregate task aborted twice on environment artifacts — first two `@open-mercato/shared` bootstrap suites failing with `listen EINVAL` on a `.pipe` path (turbo drops the exported `TMPDIR`, so `tsx` tried to open a socket under the long in-repo temp dir), then five `@open-mercato/cli` env-resolution suites. Re-run so they get a usable temp dir, **every package is green**: `@open-mercato/core` 12047 passed / 0 failed, `@open-mercato/shared` 2115, `@open-mercato/cli` 1834, `@open-mercato/ui` 1966, `@open-mercato/app` 569, plus `documents`, `search`, `scheduler`, `events`, `queue`, `cache`, `onboarding`, `telemetry`, `sync-akeneo` and the channel packages. Nothing failed twice, and nothing that failed lives in a package this PR touches — the diff is six files, all under `packages/core/src/modules/sales`.

## 💥 Breaking Changes
None. The computed net amount is unchanged for every existing caller, and `totalsFromStoredRow` is an additive optional field on `SalesLineSnapshot` that the request schemas strip. The only behavioural addition is a warn-level log line when a **caller's** `totalNetAmount` diverges from the engine's calculation by more than half a minor unit, or is not a finite number.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5707"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790430964&installation_model_id=432243&pr_number=5707&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5707&signature=4a626d7af240bf8eb84300cb13cbf8cbdee23f8e8bc995b6936ee3d64eefe541"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:

- 🔍 `review` — ready for code review; the full validation gate is green (one documented pre-existing, unrelated, locale-environment failure noted in the PR body).
- 🐛 `bug` — fixes a real defect: a validated `totalNetAmount` was silently discarded instead of being reconciled or surfaced.
- ⏭️ `skip-qa` — the change touches no UI-rendering file (only `packages/core/src/modules/sales/lib/calculations.ts` and its test), changes no database structure or API surface, breaks no `BACKWARD_COMPATIBILITY.md` contract, and ships automated tests for the changed behavior in this same PR — it qualifies for the automated-verification exemption.
- 🔽 `priority-low` — matches the source issue's own `priority-low` label; this is a low-urgency observability fix (a warn-only log line), not a user-facing or blocking regression.
- 🟡 `risk-medium` — matches the source issue's own `risk-medium` label; the touched function (`buildBaseLineResult`) is shared by every order/quote/invoice line calculation, even though the output behavior itself is unchanged.

adeptofvoltron:
🤖 `om-open-pr` completed: PR opened, labeled, and ready for review. Lock handed off to `om-auto-review-pr` — chain continues on this PR.

adeptofvoltron:
🤖 `om-auto-review-pr` taking over the chain lock — review phase. Started: 2026-08-27T13:57:42Z.

adeptofvoltron:
# 🔍 Code Review

## 🎯 Summary
`buildBaseLineResult` now reads `line.totalNetAmount` when supplied and, if it diverges from the computed net (unit price − discount), logs a `sales`-namespaced warning via `createLogger` with the supplied/computed values and line/product ids — the computed value still wins, so no output changes for any existing caller. Three new unit tests lock in the diverge/match/absent cases.

## Verdict
APPROVE

## 🧪 Validation Gate
| Check | Result |
|---|---|
| `yarn workspace @open-mercato/core test -- .../sales/lib/__tests__/calculations.test.ts` (independently re-run by the reviewer) | PASS — 16/16 (13 pre-existing + 3 new) |
| Regression check: swapped in pre-fix `calculations.ts` from `origin/develop`, re-ran the same test file (independently re-run by the reviewer) | The new "diverges" test fails as expected (`toHaveBeenCalledTimes` 0 vs 1); the other 15 are unaffected — confirms the test actually exercises the fix, not a tautology. |
| File restored, re-ran, confirmed clean (`git status --short` empty, 16/16 pass) | PASS |
| `build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `test` (full suite), `build:app` | Run by the implementer at this exact commit minutes before this review; all green except two pre-existing, unrelated `warranty_claims/__tests__/quantity.test.ts` failures caused by this machine's `pl_PL.UTF-8` locale affecting `toLocaleString(undefined, …)` — independently reproduced against a clean `origin/develop` checkout at the same base commit, confirming it is not caused by this change. |

## Findings
None. Specifically checked and found no issue in:
- **Reconciliation correctness**: `0` is handled distinctly from `null`/`undefined` (`toNumber(0, …)` returns `0`, not the fallback); the comparison uses the same `round()` (4-decimal, EPSILON-guarded) helper already used for every other money comparison in this file, so no new float-equality risk.
- **Log call safety**: `logger.warn` only receives primitive/nullable scalars (`lineId`, `productId`, two numbers) — nothing that could throw during serialization.
- **`createLogger` in a `lib/*.ts` calculation module**: not a novel pattern — the codebase already has many precedents of `lib/*.ts` files using the structured logging facade this way.
- **Test's `jest.mock` block**: functionally correct and isolated to this file; Jest sandboxes modules per test file, so the cached mock logger doesn't leak across files, and no earlier test in the same file supplies a diverging `totalNetAmount` that would race with the new `describe` block's `mockLogger.warn.mockClear()`.
- **i18n**: `lineId`/`productId` are internal UUIDs, and this is a server-side diagnostic log, not user-facing text — correctly does not route through i18n.
- **NaN edge case**: a non-finite `totalNetAmount` falls back to `netSubtotal` inside `toNumber` and never triggers a warning — a real but unlikely blind spot that mirrors the pre-existing `toNumber(line.taxAmount, 0)` pattern a few lines below, so it is consistent with the file's existing conventions rather than a new gap. Not blocking.

## 💥 Breaking Changes
- [x] `SalesLineSnapshot` — unchanged
- [x] `SalesLineCalculationResult` — unchanged, same shape/fields
- [x] `buildBaseLineResult` return value — same computed `netAmount` for every existing caller; only a new log call was added
- [x] Zod validators (`data/validators.ts`) — untouched
- [x] No exported function signatures changed

## 🧪 Test Coverage
Solid and precise. The three new cases (diverge-and-warn, match-and-no-warn, absent-and-no-warn) exercise exactly the new branch, and the "diverge" test was independently confirmed to fail against the pre-fix code and pass with the fix — a genuine regression lock, not a tautological assertion. No integration test is needed here per this repo's automated-verification exemption (no `.tsx`/UI file touched, no schema/API surface change, automated tests included in the same change).

---

**CI status disclosure:** two checks are still pending at review time (`prepare`, `Analyze (javascript-typescript)`); every check that has reported so far is passing, and no required check is currently failing. This PR is being approved on the strength of the code review and the locally-reproduced validation gate — merge should still wait for those pending checks to settle.

adeptofvoltron:
🤖 `om-auto-review-pr` — the code review above is complete with an APPROVE verdict (no blocker/major findings; validation gate green modulo one documented pre-existing, unrelated failure). GitHub blocks this automation from formally approving its own PR (`Can not approve your own pull request`), so no GitHub review state was submitted and the pipeline label stays `review` rather than moving to `merge-queue` — an independent human or bot reviewer still needs to submit the actual approving review before this can proceed to merge.

adeptofvoltron:
🤖 `om-auto-review-pr` completed: APPROVE (no blockers/majors; validation gate green modulo the documented pre-existing, unrelated locale failure). No UI-touching changes in this diff, so no autofix or QA-verification step follows. Lock retained — chain continues.

adeptofvoltron:
🤖 `om-auto-fix-issue` — lock swapped: work is done and fully reported (review posted, labels applied). `in-progress` removed, `ci-monitoring` applied — this run will post a CI-result follow-up comment within a bounded ~40-minute wait; a human or later skill run may act on this PR in the meantime.

adeptofvoltron:
🤖 `om-auto-review-pr` re-entering this PR for a re-review (same account, previous chain lock already released). Started: 2026-09-02.

Re-review triggered by a changed signal since the 2026-08-27 pass: the head no longer merges cleanly into `develop`. Review-only run — no source edits, no commits, no pushes; autofix is explicitly disabled for this invocation.

adeptofvoltron (commented):
# 🔍 Code Review

**Re-review** of `572e2a4a0` — the first pass (2026-08-27) approved this on a base that has since moved. No new commits, but two signals changed: `develop` picked up #5640, which rewrote the exact function this PR edits, and the head no longer merges cleanly. That is enough to change the verdict.

## Verdict
**REQUEST CHANGES** — 2 blockers, 2 majors, 3 minors.

## 🎯 Summary

The direction is right and well argued. #5644 offered three options; this PR picks option 3 (*validate and warn*) and, unlike the "honour it verbatim" symmetry the issue floats first, deliberately keeps the computed net authoritative so a caller cannot mask a broken discount — the exact failure class that hid #3757. The inline comment at `calculations.ts:100-104` states that rationale where the next reader will need it. The test discipline is good too: the negative control (running the new test against pre-fix code and watching it fail) is the difference between a real regression test and a tautology, and the `jest.mock` block was lifted from an existing precedent (`auth/lib/__tests__/consentIntegrity.test.ts:1-24`) rather than invented. `createLogger('sales')` at module scope matches how every other file in this module does it.

What the PR gets wrong is its blast-radius analysis. The body says the only effect is "a new warn-level log line ... when a caller's `totalNetAmount` diverges". In this codebase a caller is not the main supplier of that field — the platform's own persisted-line mappers are, on every recalculation, and they collide with a self-healing contract `develop` landed four days ago.

## 🧪 Validation Gate

| Check | Result |
|---|---|
| GitHub required checks on `572e2a4a0` | All green — `test` (7m25s), `lint`, `ds-lint`, `docker-build`, `ephemeral-integration`, CodeQL, `audit-scope` |
| Local gate re-run by this reviewer | **Not run** — this reviewing worktree has no dependency install, and see Blocker 2: a gate run on `572e2a4a0` is evidence about the pre-#5640 engine, not about what would merge |
| Mergeability | **CONFLICTING / DIRTY** (Blocker 1) |
| Merge simulation (`git merge-tree pr-head origin/develop`) | 1 content conflict, `calculations.ts` auto-merges |

I am not treating the green CI as this PR's validation evidence. It is evidence about a tree that no longer exists.

## 💥 Blockers

### 1. The PR does not merge — conflict in `packages/core/src/modules/sales/lib/__tests__/calculations.test.ts`

`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. Commit `734c263bd` ("fix(sales): make line discountAmount a line total on both the read and write path (#5640)", merged 2026-08-28) appended a `describe('line discount contract (spec 2026-08-07)')` block to the end of the same file, at the same insertion point as this PR's `describe('buildBaseLineResult net amount reconciliation (issue #5644)')` at `calculations.test.ts:602`.

**Fix:** rebase onto `origin/develop`, keep both `describe` blocks, and make sure the `jest.mock('@open-mercato/shared/lib/logger', …)` block stays at the very top of the file — it must be hoisted above every `import`, and a careless conflict resolution is exactly how it ends up below one and silently stops mocking.

### 2. Green CI on this head is stale evidence — `buildBaseLineResult` was rewritten underneath this change

`calculations.ts` auto-merges textually, but #5640 restructured the function this PR edits. On `develop`, the discount is no longer computed inline: it moved into `resolveLineDiscountTotal(line, netSubtotalBeforeDiscount, quantity)`, with percentage-first precedence and a new `discountAmountFromStoredRow` branch. The `netSubtotal` this PR compares against is therefore produced by different logic than the one the `test` job validated.

**Fix:** rebase, then re-run the full `validation.commands` gate and post the result. Given Blocker 1 and Major 1 below, treat this rebase as a real re-derivation of the change, not a mechanical `git rebase --continue`.

## 🟠 Majors

### 1. The warning fires for the platform's own rehydrated lines, not just for callers — and collides with #5640's self-healing contract
`packages/core/src/modules/sales/lib/calculations.ts:105`

Every persisted-line → snapshot mapper populates `totalNetAmount` from the stored column:

- `packages/core/src/modules/sales/lib/lineSnapshots.ts:42` — `mapPersistedLine`, the shared mapper #5640 introduced, used by both `commands/documents.ts` and `commands/returns.ts`
- `packages/core/src/modules/sales/commands/returns.ts:258`

and those mappers use `toNumeric()`, which falls back to `0` — never `null` or `undefined`. So the guard `line.totalNetAmount !== null && line.totalNetAmount !== undefined` is **always true** for a rehydrated line, and the comparison runs on every single recalculation. `commands/documents.ts` alone has 14 `calculateDocumentTotals` call sites.

Now the collision. `develop`'s own comment above `resolveLineDiscountTotal` states the contract deliberately:

> because the column is NOT NULL DEFAULT '0' a stored 0 cannot be told apart from "no discount supplied" ... That is what makes recalculation idempotent, and what lets a row whose amount the old engine dropped or re-inflated heal itself on the next pass.

For precisely that legacy population, the healed `netSubtotal` **will not** equal the stored `total_net_amount` — that divergence is the healing. This PR turns each of those into a `warn`, per line, per recalculation, indefinitely, because the discount repair CLI that would fix the rows is explicitly deferred by `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md`. Rows still carrying the `'0'` column default land in the same bucket: supplied `0` vs computed `200` reads as a divergence.

The net effect is that the loudest, most frequent producer of this new signal is a known, already-tracked, self-correcting condition — and it is indistinguishable in the log from the caller-supplied mismatch the issue actually wanted surfaced.

**Fix (preferred):** only reconcile a value that genuinely came from a caller. `SalesLineSnapshot` already models origin, but `lineSnapshots.ts:34-38` is emphatic that mappers must not overload `discountAmountBasis`/`discountAmountFromStoredRow` for a different question — so add a sibling rather than reusing one:

```ts
// types.ts
totalsFromStoredRow?: boolean
// lineSnapshots.ts mapPersistedLine(), and the returns.ts mapper
totalsFromStoredRow: true,
// calculations.ts
if (line.totalsFromStoredRow !== true && line.totalNetAmount != null) { … }
```

**Fix (alternative):** move the reconciliation out of the shared engine to the API/command edge, where "the caller supplied this" is unambiguous and no rehydrated snapshot ever reaches it. That is arguably the more honest home for a contract-conformance check anyway.

### 2. Exact equality on a diagnostic makes it cry wolf at sub-cent rounding
`packages/core/src/modules/sales/lib/calculations.ts:108`

`suppliedNetAmount !== computedNetAmount` compares two values rounded to 4 decimals (`round()` is `1e4`, matching the `precision: 18, scale: 4` columns). That is a fair bar for two engine-produced values, but callers work in money, at 2 decimals. A caller posting `99.99` for a line whose exact net is `99.9856` gets a warning that is pure rounding noise and is, in the log, identical to a genuinely mis-read discount.

The entire deliverable here is a diagnostic signal, so its noise floor is the feature. Combined with Major 1, the signal-to-noise ratio is the thing most likely to get this warning filtered out and forgotten.

**Fix:** compare with a tolerance of half a minor unit, and say why:

```ts
const NET_RECONCILIATION_TOLERANCE = 0.005
if (Math.abs(suppliedNetAmount - computedNetAmount) > NET_RECONCILIATION_TOLERANCE) { … }
```

or, if exact 4-decimal equality really is the intended bar, add a comment defending it against the 2-decimal caller case.

## 🟡 Minors

### 1. An unparseable supplied value is silently swallowed — the exact defect #5644 is about
`packages/core/src/modules/sales/lib/calculations.ts:106`

```ts
const suppliedNetAmount = round(toNumber(line.totalNetAmount, netSubtotal))
```

The fallback is the computed value itself. So a `NaN`, an `Infinity`, or a non-numeric value coerces to `netSubtotal`, compares equal, and logs nothing. The zod schema (`data/validators.ts:349`, `decimal({ min: 0 })`) covers the HTTP path, but `SalesLineSnapshot` is an internal type reached from many non-HTTP call sites, and "accepted, then silently discarded" is precisely the behaviour this PR exists to end.

**Fix:**

```ts
const parsed = toNumber(line.totalNetAmount, NaN)
if (!Number.isFinite(parsed)) {
  logger.warn('Sales line totalNetAmount is not a finite number; the computed value is used', { … })
} else if (Math.abs(round(parsed) - computedNetAmount) > NET_RECONCILIATION_TOLERANCE) {
  …
}
```

### 2. The tests do not exercise any path that will actually produce this warning in production
`packages/core/src/modules/sales/lib/__tests__/calculations.test.ts:602-666`

All three cases call `calculateLine` with a hand-built caller snapshot. None covers a rehydrated snapshot (`mapOrderLineEntityToSnapshot`), a `discountPercent`-driven divergence, or the `totalNetAmount: 0` column-default case — the three real-world producers, and the ones Major 1 turns into noise. A test over the rehydration path would have caught Major 1 before review.

Separately, `calculations.test.ts:626` asserts with `expect.objectContaining({ suppliedTotalNetAmount: 150, computedNetAmount: 200 })` and never checks `lineId` or `productId`. Those two fields are the entire operational value of the log line — *which* line is wrong — and nothing pins them. Add a case with `id` and `productId` set that asserts they reach the payload.

### 3. The other half of the asymmetry is neither fixed nor deferred in writing

#5644 closes with: "the same question applies to `totalGrossAmount` being honoured verbatim: it is the other half of the asymmetry." This PR resolves net and is silent on gross. Scoping gross out is completely reasonable, but it should be a recorded decision, not an omission — add a line to the PR body or open the follow-up issue, otherwise the asymmetry quietly reopens the next time someone reads `calculations.ts:118-121` and sees gross still winning verbatim.

## ✅ What is good

- Option 3 from the issue is the right call, and the comment at `calculations.ts:100-104` explains *why not* option 1 at the point of confusion rather than burying it in the PR description.
- The negative control on the new test is real verification, not ceremony.
- The logger mock follows an established in-repo pattern instead of inventing a new one.
- No behavioural change to any existing caller, and no contract surface touched — the `skip-qa` classification is correct.

## 📋 What is needed to land this

1. Rebase onto `origin/develop`, resolving the test-file conflict and keeping the `jest.mock` block hoisted (Blocker 1).
2. Gate the reconciliation on caller origin so rehydrated lines do not warn (Major 1) — this is the one that needs a design decision, not a patch.
3. Add a tolerance, or defend exact equality in a comment (Major 2).
4. Handle the non-finite supplied value explicitly (Minor 1).
5. Extend the tests to the rehydration path and pin `lineId`/`productId` (Minor 2).
6. Re-run the full validation gate on the rebased tree and post the result (Blocker 2).

---
🤖 Generated by `om-auto-review-pr` (review-only run — autofix disabled by the invoking operator; no source was modified). GitHub does not permit this account to submit a formal `REQUEST_CHANGES` review on its own PR, so this is posted as a comment review; the verdict above is nonetheless **request changes**, and the pipeline label has been moved to `changes-requested` accordingly.

adeptofvoltron:
🤖 `om-auto-review-pr` — 🏷️ label rationale (re-review, 2026-09-06)

Supersedes this comment's 2026-09-02 content: that pass moved the PR to `changes-requested`, and every finding behind that verdict is now addressed on head `f763de874`. The current label state:

- 🚀 `merge-queue` — the re-review verdict is **approve** (0 blockers, 0 majors, 2 minors, 1 nit), the full validation gate is green on the merged tree, and all 20 CI checks pass on this head. Applied by `om-auto-fix-pr` on 2026-09-02 and confirmed rather than changed here. The merge still waits on an independent approving review: GitHub will not let this account approve its own PR, so `reviewDecision` remains `REVIEW_REQUIRED`.
- 🐛 `bug` — fixes a real defect: a validated `totalNetAmount` was accepted and then silently discarded (#5644) instead of being reconciled or surfaced.
- ⏭️ `skip-qa` — the automated-verification exemption holds. The diff is six files, all under `packages/core/src/modules/sales`, with no `.tsx`, nothing under `packages/ui/src/` or `**/components/**`, no database-structure change, no API-surface change (the request schemas strip the new field), no `BACKWARD_COMPATIBILITY.md` contract touched, and automated tests for the changed behaviour shipping in the same PR.
- 🔽 `priority-low` — kept. The change adds a diagnostic; it alters no calculated output for any existing caller, so nothing is currently broken for a user waiting on it.
- 🟡 `risk-medium` — kept. It edits the shared sales calculation engine, which argues for more than `risk-low`, but the reconciliation is a pure side effect with no effect on `netAmount`, and the origin model is covered by unit tests on every producer, which argues against `risk-high`.

adeptofvoltron:
🤖 `om-auto-review-pr` completed: **CHANGES REQUESTED** (2 blockers, 2 majors, 3 minors). Handing back to @adeptofvoltron.

autofix: **skipped** — this run was invoked review-only by the operator; no source was edited, committed, or pushed. The fixes are the author's to apply.

The one that needs a decision rather than a patch is Major 1: `lineSnapshots.ts:42` and `commands/returns.ts:258` populate `totalNetAmount` from the stored column via `toNumeric()` (which floors to `0`, never `null`), so the new guard is always true for a rehydrated line and the reconciliation runs on every recalculation — including for the legacy rows that #5640 deliberately heals on the next pass, whose stored net is *supposed* to diverge. That needs a caller-origin flag on the snapshot (or the check moved to the command edge) before this is safe to merge.

`in-progress` released. `ci-monitoring` removed — CI is not the blocker here; the merge conflict and the findings are.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-09-02T20:52:58Z.

This run drives the PR to merge-ready: merge the latest `develop` in, work the 2026-09-02 re-review findings via the review/autofix engine, stabilize CI, and hand off. It never merges.

adeptofvoltron:

- 🚀 `changes-requested` → `merge-queue` — every finding from the 2026-09-02 re-review is now addressed on the head: the merge conflict is resolved (Blocker 1), the full validation gate has been re-run on the merged tree (Blocker 2), the reconciliation is gated on caller origin (Major 1), the comparison carries a rounding tolerance (Major 2), a non-finite supplied value warns explicitly (Minor 1), and the tests cover the rehydration path while pinning `lineId`/`productId` (Minor 2). Minor 3 was scoped out deliberately and is tracked in #5853. Note that this label carries the verdict, not a green CI run — required checks are re-running on the new head and this comment is not a claim that they passed.
- 🐛 `bug` — unchanged and still correct: the PR fixes a real defect, a validated `totalNetAmount` being silently discarded rather than reconciled or surfaced.
- ⏭️ `skip-qa` — still correct after the diff grew. The change is six files, all `.ts` under `packages/core/src/modules/sales`, with no UI-rendering file (no `.tsx`, nothing under `packages/ui/src/` or `**/components/**`), no database structure change, no API surface change (the new `totalsFromStoredRow` field is mapper-only and the request schemas strip it), no `BACKWARD_COMPATIBILITY.md` contract broken (an optional field added to an exported type is additive), and automated tests for the changed behaviour ship in this same PR. That is the automated-verification exemption, so no `needs-qa` and no UI QA run.
- 🔽 `priority-low` — unchanged: this remains a low-urgency observability fix. The engine's output is identical for every existing caller; only a warn-level log line is new.
- 🟡 `risk-medium` — unchanged, and still the right call now that the diff also touches `commands/documents.ts`: `buildBaseLineResult` and the line-upsert paths are shared by every order/quote/invoice line calculation, so the blast radius is an ordinary single-module one rather than a narrow one, even though the calculated output does not change.
- 👷 `in-progress` → `ci-monitoring` — the work is finished and fully reported, so the claim is released and swapped for the meta label that says only one thing is still owed: the CI-result follow-up comment. `ci-monitoring` is not a lock; anyone may act on this PR while it is set.

adeptofvoltron:

This run took the PR from *changes-requested with an unmergeable head* to merge-ready. Everything below happened on the head that is pushed now (`f763de874`), not on the tree the 2026-08-27 pass reviewed.

## 🔀 Base merge — first, before anything else

`origin/develop` was merged into the PR branch before any review or CI work, so review and validation judge the real merge result rather than a diff against a base that has moved on. The one conflict was in `packages/core/src/modules/sales/lib/__tests__/calculations.test.ts`, where #5640 appended a `describe('line discount contract (spec 2026-08-07)')` block at the same insertion point as this PR's `describe('buildBaseLineResult net amount reconciliation (issue #5644)')`. Both blocks were kept, and the `jest.mock('@open-mercato/shared/lib/logger', …)` block is still the first statement in the file, hoisted above every `import` — a careless resolution that pushes it below one silently stops mocking, which is exactly what the re-review warned about. `packages/core/src/modules/sales/lib/calculations.ts` auto-merged, but was then re-derived rather than accepted mechanically, because #5640 restructured the very function this PR edits.

## 🔍 How each re-review finding ended up

- **Blocker 1 — the PR does not merge.** Fixed by the merge above; the head is no longer `CONFLICTING`.
- **Blocker 2 — green CI on the old head is stale evidence.** The full `validation.commands` gate was re-run on the merged tree, in the configured order, in local mode (no compose `app` container runs on this machine). Results are in the PR body's 🧪 Tests section: `build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck` and `build:app` all exit 0, and every test package is green when re-run with a usable temp dir — `@open-mercato/core` 12047 passed / 0 failed, `@open-mercato/shared` 2115, `@open-mercato/cli` 1834, `@open-mercato/ui` 1966, `@open-mercato/app` 569. The two aborts seen in the aggregate turbo run were local environment artifacts (turbo drops the exported `TMPDIR`, so `tsx` tried to open a socket under this machine's long in-repo temp path and failed with `listen EINVAL`); neither reproduced on re-run, and neither lives in a package this PR touches.
- **Major 1 — the warning fires for the platform's own rehydrated lines.** Fixed, taking the reviewer's preferred option. `SalesLineSnapshot` gains `totalsFromStoredRow`, a mapper-only origin flag that is a **sibling** of `discountAmountFromStoredRow` rather than an overload of it, because `lineSnapshots.ts` is explicit that a mapper must not answer one origin question with the other's field. It is set in `mapPersistedLine`, carried independently through `createLineSnapshotFromInput` (a line upsert can take its discount from the caller while its totals still come off the row, so folding the two flags into one branch would have been wrong), and derived per upsert by a new `resolveUpsertTotalsOrigin` helper covering both the order and quote paths. A rehydrated line therefore never warns, which is what keeps the legacy rows #5640 heals on the next pass out of the log.
- **Major 2 — exact equality makes the diagnostic cry wolf.** Fixed with `NET_RECONCILIATION_TOLERANCE = 0.005`, half a minor unit, with the reasoning recorded at the constant: the engine rounds to the 4 decimals the numeric columns carry while callers work in money at 2.
- **Minor 1 — an unparseable supplied value is silently swallowed.** Fixed: parsing now falls back to `NaN` rather than to the computed net, and a non-finite value gets its own explicit warning instead of comparing equal and logging nothing.
- **Minor 2 — the tests do not exercise a path that will really produce this warning.** Fixed: the suite now covers a rehydrated line built through `mapOrderLineEntityToSnapshot` whose stored net diverges because a percentage discount is healing, the `NOT NULL DEFAULT` zero a persisted row carries, both sides of the rounding tolerance, and the non-finite value — and the divergence case asserts the whole payload, pinning `lineId` and `productId`, which are the fields that make the log line operationally useful. `lineSnapshots.test.ts` additionally asserts that the mappers set the new flag, that `resolveUpsertTotalsOrigin` decides correctly operand by operand, and that the request schemas strip `totalsFromStoredRow` so a caller cannot suppress the warning about its own value.
- **Minor 3 — the gross half of the asymmetry is neither fixed nor deferred in writing.** Deferred deliberately and now recorded in both places the reviewer asked for: a ✂️ Scope section in the PR body, and follow-up issue #5853. It is not a mechanical repeat of the net change — several call sites depend on a supplied gross winning verbatim (the branch immediately below derives tax back out of the gross for a tax-class-priced line whose rate was not persisted), so it needs its own behavioural analysis.

Every guard added here was negative-controlled: with the pre-fix reconciliation block restored, exactly the four new cases fail and no others, so they lock in real behaviour rather than restating it.

## 📸 UI verification

Not applicable. The diff is six files, all `.ts` under `packages/core/src/modules/sales`, with no `.tsx` and nothing under `packages/ui/src/` or `**/components/**`, so there is no user-facing surface to drive and `om-auto-qa-pr` was not run. `skip-qa` still holds for the same reason (see the label-rationale comment).

## 🔁 Follow-ups filed

- #5853 — decide and record how a caller-supplied `totalGrossAmount` should be reconciled (re-review Minor 3), assigned to the PR author, with the reviewer's excerpt quoted and acceptance criteria derived from it.

## 🚀 Merge readiness

**CI is still running on this head.** Branch protection plus the review requirement hold the actual merge; this run's verdict covers the work, not a green run. Checks still pending at the time of writing: `prepare` and `Analyze (javascript-typescript)`, with the jobs that depend on `prepare` yet to register. A follow-up comment will report the CI outcome.

One thing still needs a human: this automation account authored the PR, and GitHub refuses to let it approve or request changes on its own pull request, so the verdict above lives in this comment and in the `merge-queue` label rather than in a GitHub review state. An independent reviewer still has to submit the approving review.

This skill does not merge. Once the required checks are green and a human approval is in place, hand off with `om-approve-merge-pr 5707`, which re-checks every gate before squash-merging.

adeptofvoltron:

The checks on head `f763de874` — the merged-with-`develop` tree carrying the re-review fixes — have settled, and **every one of them passed**. This closes the disclosure in the run summary above: that comment reported the work with CI still running, and this is the promised follow-up.

| Check | Result |
|---|---|
| [`test`](https://github.com/open-mercato/open-mercato/actions/runs/33683449946/job/100426567696) | ✅ pass |
| [`ephemeral-integration (none)`](https://github.com/open-mercato/open-mercato/actions/runs/33683449946/job/100426567511) | ✅ pass |
| [`docker-build`](https://github.com/open-mercato/open-mercato/actions/runs/33683449946/job/100426567779) | ✅ pass |
| [`documents-multi-instance`](https://github.com/open-mercato/open-mercato/actions/runs/33683449946/job/100426567885) | ✅ pass |
| [`lint`](https://github.com/open-mercato/open-mercato/actions/runs/33683449946/job/100425186722) / [`ds-lint`](https://github.com/open-mercato/open-mercato/actions/runs/33683449946/job/100425186749) | ✅ pass |
| [`prepare`](https://github.com/open-mercato/open-mercato/actions/runs/33683449946/job/100425186423) / [`scope`](https://github.com/open-mercato/open-mercato/actions/runs/33683449992/job/100425186725) / [`audit-scope`](https://github.com/open-mercato/open-mercato/actions/runs/33683449946/job/100425186692) | ✅ pass |
| [CodeQL](https://github.com/open-mercato/open-mercato/runs/100425363367) — javascript-typescript, python, actions | ✅ pass |
| [`license/cla`](https://cla-assistant.io/open-mercato/open-mercato?pullRequest=5707) | ✅ pass |

`merge-coverage`, `mutate`, `comment`, `audit` and `[code]smith` reported as skipped, which is their configured behaviour for this change and not a failure.

Two things worth stating plainly. First, this is independent confirmation of the local gate reported in the summary: the CI `test` job passing on the same commit means the two `@open-mercato/shared` and five `@open-mercato/cli` failures seen locally were indeed this machine's temp-directory artifacts, not anything in the branch. Second, nothing about the verdict changes — the PR keeps `merge-queue`, and the `ci-monitoring` label is removed now that this follow-up has landed.

What remains is a human decision, not a machine one: GitHub will not let the account that authored this PR approve it, so an independent reviewer still has to submit the approving review. Once that is in place, `om-approve-merge-pr 5707` owns the merge and will re-check every gate before squash-merging.

adeptofvoltron:
🤖 `om-auto-review-pr` re-entering this PR for a re-review. Started: 2026-09-06T12:11:27Z.

The previous chain lock was already released (`om-auto-fix-pr` posted its CI-result follow-up on 2026-09-02 and removed `ci-monitoring`), so this run opens its own claim and will release it at the end. Re-review of head `f763de874` — the merged-with-`develop` tree carrying the 2026-09-02 fixes, which no formal review has covered yet.

adeptofvoltron (commented):
# 🔍 Code Review

**Re-review** of `f763de874` — the merged-with-`develop` head carrying the fixes for the 2026-09-02 pass. That review requested changes on 2 blockers, 2 majors and 3 minors; this one re-checks every one of them against the code that is actually pushed, and re-runs the gate on the merged tree rather than on the pre-#5640 engine.

## Verdict

**APPROVE** — 0 blockers, 0 majors, 2 minors, 1 nit. Every item from the previous pass is addressed on this head. The minors are follow-up-grade and none of them should hold the merge.

## 🎯 Summary

`buildBaseLineResult` now reconciles a **caller-supplied** `totalNetAmount` against the computed net and warns on a divergence, while the computed value stays authoritative — deliberately not the "honour it verbatim" symmetry `totalGrossAmount` has, because that would let a caller mask the broken discount that hid #3757. The interesting part of this revision is not the warning; it is the origin model that makes the warning mean something.

`totalsFromStoredRow` is a new mapper-only flag on `SalesLineSnapshot`, a **sibling** of `discountAmountFromStoredRow` rather than an overload of it. That distinction is load-bearing and the PR gets it right: a line upsert can take its discount from the caller while its totals still come off the stored row, so one flag could not have answered both questions. `mapPersistedLine` sets it, `createLineSnapshotFromInput` carries it through independently of the discount origin, `resolveUpsertTotalsOrigin` derives it per upsert operand-by-operand, and the request schemas strip it so a caller cannot suppress the warning about its own value.

I traced every producer of a `SalesLineSnapshot` that reaches the calculation engine, because the previous pass's Major 1 was exactly a missed producer:

| Path | Origin | Reconciled? |
|---|---|---|
| `documents.ts:4913`, `:5942` — order/quote **create** from `parsed.lines` | caller (`orderLineCreateSchema.parse`) | ✅ yes, correctly |
| `documents.ts:5320`, `:5542` — document **update** recalc | `map*LineEntityToSnapshot` | ❌ no, correctly |
| `documents.ts:7298`, `:7796` — line **upsert**, edited line | `resolveUpsertTotalsOrigin` | ✅ only when the caller supplied a value |
| `documents.ts:7298`, `:7796` — line upsert, **untouched siblings** | `mapOrderLineEntityToSnapshot` → spread → `createLineSnapshotFromInput` | ❌ no, correctly |
| `documents.ts:7480`, `:7950` — line **delete** rebuild | `...map*LineEntityToSnapshot(line)` | ❌ no, correctly |
| `documents.ts:8174`, `:8384`, `:8633`, `:8841` — adjustment upsert/delete | `map*LineEntityToSnapshot` | ❌ no, correctly |
| `returns.ts:200`, `:386`, `:534`, `:722` | `mapOrderLineEntityToSnapshot` | ❌ no, correctly |
| `seed/examples.ts:1346`, `:1564` | seed lines, no `totalNetAmount` set | n/a — guard skips null |

Every one lands where it should. The previous review's `returns.ts:258` concern does not apply on this head: that site builds a `ReturnSnapshot` undo payload, not a `SalesLineSnapshot`, and never reaches `buildBaseLineResult`; the return flows all go through the shared `mapOrderLineEntityToSnapshot`, which sets the flag.

## 🧪 Validation Gate

Local mode (no compose `app` container running), on the PR head `f763de874`, in an isolated worktree with a fresh dependency install.

| Command | Result |
|---|---|
| `yarn build:packages` | ✅ 27/27 tasks successful |
| `yarn generate` | ✅ exit 0, no generated churn (`git status` clean afterwards) |
| `yarn build:packages` (second, per the configured order) | ✅ 27/27 |
| `yarn i18n:check-sync` | ✅ all 5 locales in sync across 57 modules |
| `yarn i18n:check-usage` | ✅ exit 0 |
| `yarn typecheck` | ✅ 27/27 tasks successful |
| `yarn test` | ⚠️ aborted on 5 pre-existing `@open-mercato/cli` failures — see below |
| `yarn workspace @open-mercato/core test` | ✅ **12047 passed, 0 failed**, 2 skipped; 1478/1479 suites, 140s |
| `yarn build:app` | ✅ exit 0, `Cached: 0` — a real build, 1m32s |

**The `yarn test` caveat is about this machine, not this branch.** The aggregate turbo task stopped at `@open-mercato/cli#test` with 5 failures in `src/lib/__tests__/resolver.enterprise.test.ts` — `loadEnabledModules()` picking up extra modules (`example`, `ratelimit_probe`, `example_customers_sync`) from the checkout location. These are location-dependent and reproduce on a clean checkout of any branch; the PR's diff is six files, all under `packages/core/src/modules/sales`, and touches nothing the CLI resolver reads. Because turbo aborts the remaining packages on that failure, `@open-mercato/core` never ran under the aggregate task and was run standalone instead — the row above is that run.

**CI on this exact head is fully green**, which independently corroborates the CLI failures being environmental: 20 checks reported, 0 failing, 0 pending — `test` (8m44s), `lint`, `ds-lint`, `docker-build` (8m51s), `ephemeral-integration (none)` (17m9s), `documents-multi-instance`, `prepare`, `scope`, `audit-scope`, CodeQL and all three `Analyze` jobs. Unlike the previous pass, this green run is evidence about the tree that will merge: it ran on `f763de874`, which already contains the `origin/develop` merge.

**Mergeability:** `MERGEABLE`, clean against `develop` at `19bf96975`. Previous Blocker 1 cleared.

## ✅ Previous-pass findings — all six accounted for

1. **Blocker 1 (does not merge)** — cleared. Merge commit `536b6c3ce` brings `origin/develop` in; both `describe` blocks survive in `calculations.test.ts`, and the `jest.mock('@open-mercato/shared/lib/logger', …)` block is at `calculations.test.ts:1`, hoisted above every `import` exactly as the fix instruction required. I checked this specifically because a careless resolution is how that block silently stops mocking.
2. **Blocker 2 (stale CI evidence)** — cleared. The gate above and the green CI both ran on the merged tree, against the post-#5640 `resolveLineDiscountTotal` engine.
3. **Major 1 (warns for the platform's own rehydrated lines)** — fixed via `totalsFromStoredRow`, with the producer audit above confirming no path was missed. The preferred fix from the review is the one taken, including its "add a sibling, do not reuse `discountAmountFromStoredRow`" constraint.
4. **Major 2 (exact equality, no rounding tolerance)** — fixed. `NET_RECONCILIATION_TOLERANCE = 0.005` at `calculations.ts:33`, with the 4-decimal-engine / 2-decimal-caller rationale written down at the constant rather than at the call site. Both sides of the boundary are tested (`99.99` vs `99.9855` stays silent; `100.006` vs `100` warns).
5. **Minor 1 (non-finite value silently swallowed)** — fixed. `toNumber(line.totalNetAmount, NaN)` plus an explicit `Number.isFinite` branch with its own message, and the test comments record *why* the old `netSubtotal` fallback was the bug rather than a style choice.
6. **Minor 2 (tests never exercised a production path)** — fixed. The suite now covers a real rehydrated snapshot built through `mapOrderLineEntityToSnapshot` with a `discountPercent`-driven divergence, the `NOT NULL DEFAULT` stored zero, and pins `lineId`/`productId` in the payload with an exact `toHaveBeenCalledWith` rather than `objectContaining`.
7. **Minor 3 (gross half of the asymmetry undocumented)** — fixed. The PR body now carries a `✂️ Scope` section, and #5853 ("Decide and record how a caller-supplied totalGrossAmount should be reconciled") is open and labelled. The scoping argument given there is a real one, not a deferral: the branch below derives tax back out of a supplied gross for a tax-class-priced line, so gross genuinely needs its own behavioural analysis.

## 🟡 Minors

### 1. The `totalsFromStoredRow` carry-through is the one link with neither a type nor a test
`packages/core/src/modules/sales/commands/documents.ts:3093`

`documents.line-discount-contract.test.ts` exists precisely because this rebuild is where the discount origin got lost before — its header says so, and it deliberately asserts that *an untouched line* keeps its stored-row origin through `createLineSnapshotFromInput`, because "an implementation that tags the edited line correctly but lets the untouched ones lose their stored-row origin passes a naive single-line test".

The new flag rides that identical path, on the adjacent line of the identical function, and gets no equivalent command-level test. Its unit coverage stops at the three pieces either side of the link: `mapOrderLineEntityToSnapshot` sets it (`lineSnapshots.test.ts:83`), `resolveUpsertTotalsOrigin` derives it (`:159-179`), and the engine honours it (`calculations.test.ts:749`). Nothing drives `sales.orders.lines.upsert` or `sales.orders.lines.delete` and asserts that an untouched line stays silent.

It is also the only link that TypeScript cannot check: `DocumentLineCreateInput` does not model the field, so `isStoredRowSourcedTotalsLine` reaches it through `line as Pick<SalesLineSnapshot, "totalsFromStoredRow">`. Rename the field on `types.ts:84` and that cast keeps compiling while the propagation silently stops.

Minor rather than major because the failure mode is log noise on untouched lines, not wrong money — the totals themselves are unaffected. But the cheapest fix is one `it` in the existing `documents.line-discount-contract.test.ts` world, which already builds persisted lines with `totalNetAmount` set and already runs both the upsert and the delete handler.

### 2. The second origin flag is not recorded next to the first in the spec that defines the contract
`.ai/specs/2026-08-07-sales-line-discount-amount-contract.md` § 3

That section is where a future mapper author learns the origin-flag rules: it carries a per-producer table, the field sketch, and an explicit *"Invariant, and it must be tested: no entity→snapshot mapper ever sets `discountAmountBasis`, and no request schema ever populates `discountAmountFromStoredRow`."* This PR introduces a second flag of exactly that class, with exactly that invariant and its own producer table, and the spec does not mention it.

The invariant *is* enforced — the JSDoc at `types.ts:76-83` states it and `lineSnapshots.test.ts:210-221` pins the schema-strip half for both order and quote lines — so this is a documentation gap, not a correctness one. But `AGENTS.md` asks for specs to be updated when a significant change lands, and the next person to add a mapper will read § 3, not `types.ts`. A short subsection, or a `2026-09-xx` addendum spec, closes it.

## 🔹 Nits

### 1. The two origin predicates are the same function twice
`packages/core/src/modules/sales/commands/documents.ts:3035-3048`

`isStoredRowSourcedLine` and `isStoredRowSourcedTotalsLine` differ only in the property name they cast to and read. A single `hasStoredRowOrigin(line, 'totalsFromStoredRow')` would collapse them and give the cast one home instead of two. Genuinely marginal — the pair reads clearly as-is, and the duplication is four lines — so take it or leave it.

## 💥 Breaking Changes

| Surface | Assessment |
|---|---|
| Exported types | `SalesLineSnapshot` gains an **optional** field — additive, no consumer breaks |
| Exported functions | `resolveUpsertTotalsOrigin` is **new**; nothing removed, renamed, or re-signatured |
| HTTP routes / response shapes | Unchanged; `orderLineCreateSchema` / `quoteLineCreateSchema` strip the new field, so no request or response surface moves |
| DB schema | Untouched — no entity change, no migration, and none needed (`totalsFromStoredRow` is never persisted) |
| Event IDs, DI keys, ACL features, CLI commands | Untouched |
| Calculated output | **Unchanged for every caller.** `netAmount` still comes from `unitPriceNet`/discount; the reconciliation is a pure side effect that writes a log line and nothing else |

No `BACKWARD_COMPATIBILITY.md` protected surface is touched. Nothing to WARN about.

## 🧪 Test Coverage

Strong, and notably better than the previous revision. 317 added lines across 6 files, 230 of which are tests — a ratio that is right for a change whose entire deliverable is a diagnostic signal.

- Both sides of the rounding tolerance, not just the happy side.
- A real rehydrated snapshot built through the actual mapper, rather than a hand-set flag — this is the case that would have caught the previous pass's Major 1.
- The stored `0` column-default case, tested separately from the rehydration case, because they fail for different reasons.
- The schema-strip invariant tested for **both** flags and **both** document kinds, which is what stops a caller silencing a warning about its own value.
- `lineId` and `productId` pinned by exact `toHaveBeenCalledWith`, since "which line is wrong" is the whole operational value of the log line.
- The PR body reports a negative control — the pre-fix block restored, exactly the four new guards failing and no others. That is the difference between a regression test and a tautology.

Gap: the command-level carry-through, per Minor 1.

## ✅ What is good

- The sibling-flag decision is the right one and is argued at the point of confusion (`types.ts:76-83`, `lineSnapshots.ts:42-46`, `documents.ts:3090-3092`) rather than buried in the PR description. Reusing `discountAmountFromStoredRow` would have been the obvious shortcut and would have quietly broken the upsert case where the discount is a caller's and the totals are the row's.
- `resolveUpsertTotalsOrigin` is decided operand-by-operand instead of collapsed into a `caller ?? stored` expression, matching the reasoning its `resolveUpsertDiscountFields` sibling already documents — and it is tested operand-by-operand too, including the explicit caller zero.
- The tolerance constant carries its own justification, so the next reader does not have to reverse-engineer why `0.005` and not `0.01` or exact equality.
- Scope discipline: gross is excluded with a written reason and a tracked issue, not silently.
- The rehydration test's comment explains that #5640's healing behaviour is *supposed* to diverge — which is the single most confusing thing about this code, recorded where it will be read.

---
🤖 Generated by `om-auto-review-pr`. GitHub does not permit this account to submit a formal approving review on its own pull request, so this is posted as a comment review; the verdict above is nonetheless **approve**, and the pipeline label stays `merge-queue` accordingly. `reviewDecision` remains `REVIEW_REQUIRED` — an independent human or bot reviewer still has to submit the actual approving review before this can merge.

adeptofvoltron:
🤖 `om-auto-review-pr` completed: **APPROVE** — 0 blockers, 0 majors, 2 minors, 1 nit. Lock released.

All six items from the 2026-09-02 re-review are addressed on head `f763de874`: the merge conflict is gone (`MERGEABLE` against `develop` at `19bf96975`), the `jest.mock` logger block survived the resolution hoisted above every import, the reconciliation is gated on caller origin via the new `totalsFromStoredRow` sibling flag, the comparison carries a rounding tolerance, a non-finite supplied value warns explicitly, the tests reach the rehydration and stored-zero paths and pin `lineId`/`productId`, and the gross half of the asymmetry is scoped out in writing against the open #5853.

I traced every producer of a `SalesLineSnapshot` that reaches `buildBaseLineResult` — the 14 `documents.ts` call sites, the 4 in `returns.ts`, and the 2 seed sites — and each one lands on the correct side of the origin gate. The previous pass's `returns.ts:258` concern does not apply here: that site builds a `ReturnSnapshot` undo payload, not a calculation snapshot.

**Validation gate: green, in local mode, on this head, in an isolated worktree with a fresh install.** `build:packages`, `generate` (no churn), `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck` and `build:app` all exit 0; `@open-mercato/core` is 12047 passed / 0 failed. The one caveat is about this machine, not this branch: `yarn test` aborted at `@open-mercato/cli#test` on 5 location-dependent `resolver.enterprise.test.ts` failures, which is why core was run standalone. All 20 CI checks on `f763de874` are green, with none pending, so there is no CI follow-up owed and no `ci-monitoring` label is being applied.

**autofix: not run — nothing actionable to fix.** The loop exists to drive a `changes-requested` verdict to approvable, and this head is already approvable. The two minors are follow-up-grade and neither justifies invalidating a green CI run on a merge-ready PR:

1. The `totalsFromStoredRow` carry-through at `documents.ts:3093` has neither type checking (it is reached through a cast, since `DocumentLineCreateInput` does not model the field) nor a command-level test — unlike its discount sibling, which has a dedicated test asserting an *untouched* line keeps its origin through the same rebuild. A break there means log noise, not wrong totals.
2. The new origin flag is not recorded in § 3 of `.ai/specs/2026-08-07-sales-line-discount-amount-contract.md`, where its sibling's producer table and invariant live and where the next mapper author will look.

**This still needs a second reviewer.** GitHub refuses `Can not approve your own pull request`, so the review above is posted as a comment review and `reviewDecision` remains `REVIEW_REQUIRED`. `merge-queue` reflects the code verdict, not merge permission — an independent human or bot approval is the remaining gate.

PR #5709, opened 2026-08-27
fix(customer_accounts): let re-invitation survive a soft-deleted portal user (#5532)

### LINKED ISSUE #5532: Soft-deleted portal user permanently blocks re-invitation; acceptance fails with a raw 500
Found during manual QA of #4471, but unrelated to it — that PR touches neither `customerInvitationService.ts` nor `api/invitations/accept.ts` nor any migration. Reproduces on `develop`.

## What happens

Delete a customer-portal user, then invite the same email address again. The invitation is created and the email is sent normally — but **accepting** it fails:

```
UniqueConstraintViolationException: duplicate key value violates unique constraint
  "customer_users_tenant_email_hash_uniq"
  Key (tenant_id, email_hash)=(<tenant>, v2:b0d04e5d…) already exists
  at CustomerInvitationService.acceptInvitation (customerInvitationService.ts:208)  // em.flush()
POST /api/customer_accounts/invitations/accept 500
```

The portal shows **"Invitation acceptance failed. Please try again."** — which invites the user to retry an operation that can never succeed.

## Why

Deleting a portal user is a **soft** delete: the row stays in `customer_users` with `deleted_at` set. The uniqueness index does not exclude those rows:

```sql
CREATE UNIQUE INDEX customer_users_tenant_email_hash_uniq
  ON public.customer_users USING btree (tenant_id, email_hash);
```

So a deleted account permanently burns its email address for that tenant. The raw database exception then escapes to the HTTP layer as a 500 instead of a domain error.

## Reproduce

1. Create a portal user for `someone@example.com`
2. Delete it from Customer Users
3. Invite `someone@example.com` again (CRM person → Portal Account → Invite to Portal) — invitation created, email delivered
4. Open the invitation link, set a password, submit → 500 and the generic retry message

Verified causal, not inferred: hard-deleting the orphaned row (`DELETE FROM customer_users WHERE id = … AND deleted_at IS NOT NULL`) made the identical flow return `201` and land the user in the portal.

## Suggested fix

Two independent parts:

1. **Correctness** — make the index partial (`WHERE deleted_at IS NULL`), or have `acceptInvitation` revive/replace the soft-deleted record. Whichever is chosen, deleting a portal user should not permanently consume the address.
2. **Error surface** — `acceptInvitation` should fail with a domain error mapped to 409 and a message that tells the truth ("an account with this address already exists"), instead of letting a `23505` reach the client as a 500 behind "Please try again".

adeptofvoltron:
Fixes #5532

## Root cause

`customer_users` had a plain unique index on `(tenant_id, email_hash)` covering every row regardless of `deleted_at`. Deleting a customer-portal user, then inviting the same email address again, created a new invitation that looked fine — but **accepting** it hit the DB unique-constraint violation on insert and surfaced a raw `UniqueConstraintViolationException` / 500, because the soft-deleted row still occupied the slot.

`customerInvitationService.ts:134` (`api/invitations/accept.ts`) does not touch any migration and is unrelated to #4471, per the issue's own note.

## Fix

- Replace the entity's `@Unique` decorator with a partial unique index (`WHERE deleted_at IS NULL`), so a soft-deleted row no longer blocks reuse of its email — migration + snapshot included.
- Add an explicit active-user conflict check in `acceptInvitation()`: if a *non-deleted* `CustomerUser` already owns that email, throw a new `CustomerInvitationEmailConflictError` domain error instead of letting the DB constraint surface as a raw 500.
- Map that error to a clean `409` in the accept API route, with an OpenAPI doc entry for it.

## Security impact

Touches an entity used by portal auth/account lookup (`customer_users`), so flagging per the PR template's security-impact convention: this only narrows the existing uniqueness constraint (from "ever existed" to "currently active"), and adds an explicit conflict check ahead of the write — it does not relax tenant scoping, and the new lookup is still tenant/organization-scoped (`findOneWithDecryption` with `{ tenantId, organizationId }`).

## Verification

- `yarn workspace @open-mercato/core typecheck` — clean.
- `yarn workspace @open-mercato/core test -- .../customerInvitationService.test.ts` — 10/10 pass (added conflict/soft-delete coverage).
- `yarn eslint` on the touched files — 0 errors/warnings.
- Migration + `.snapshot-open-mercato.json` reviewed by hand; not applied locally (per project policy, migrations ship in the PR rather than being run against a local DB by the agent).

---
Opened by the `om-autofix-fleet` orchestrator (fleet-20260827-152338). The worker's `om-fix` step completed and reported ready, but the session ended before handing off to `om-open-pr`, leaving the diff uncommitted in its worktree (`.ai/tmp/om-auto-fix-issue/issue-5532-20260827-154200`). Salvaged and re-validated by hand, then opened here.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5709"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790431738&installation_model_id=432243&pr_number=5709&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5709&signature=d872afeb0c639e34059cf8d95fa75fc83799a24e73b3db3027832ae763d60dc1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


adeptofvoltron:
🤖 `om-autofix-fleet` — label rationale

- 🐛 `bug`, `priority-medium`: carried over from issue #5532.
- ⚠️ `risk-medium`: touches a DB migration/unique-index change on `customer_users` plus portal-auth-adjacent account-lookup logic — schema surface, not a pure UI/copy fix.
- 🔍 `needs-qa`: the automated-verification `skip-qa` exemption requires unchanged DB structure; this PR ships a migration, so it does not qualify even with test coverage in the same PR.
- 📋 `review`: ready for review; left as **draft** per the fleet's forced-draft policy so a maintainer promotes it after review.

Opened by hand after the fleet worker for #5532 finished its `om-fix` step (reported "Status: ready") but the session ended before handing off to `om-open-pr`, leaving the diff uncommitted in `.ai/tmp/om-auto-fix-issue/issue-5532-20260827-154200`. I salvaged it: reviewed the diff (migration + partial-unique-index + conflict-error mapping), ran typecheck, the service test file (10/10 pass), and lint before committing/pushing/opening this PR. The migration was reviewed but not applied locally.

adeptofvoltron:
🤖 `om-auto-review-pr` started by @adeptofvoltron at 2026-09-02T12:12:17Z. Other auto-skills will skip this PR until the lock is released.

adeptofvoltron (commented):
> **Note:** submitted as a `COMMENT` review because GitHub refuses both `APPROVE` and `REQUEST_CHANGES` on a self-authored PR. The verdict below is **changes requested** and the `changes-requested` pipeline label is applied accordingly.

# 🔍 Code Review: fix(customer_accounts): let re-invitation survive a soft-deleted portal user (#5532)

**Verdict: 🔴 Changes requested** — 3 majors, no waiver. The diagnosis is right and the partial unique index is the correct shape of fix, but the new application-level guard has two holes that let the original symptom (a raw 500 from the unique-constraint violation) still reach production, and the change that actually failed in #5532 — the DB index — has no test that exercises a real database.

## ✅ What is good

- The root cause is correctly identified and correctly located: a full `UNIQUE (tenant_id, email_hash)` over soft-deleted rows, not the accept handler.
- Narrowing the constraint to `WHERE deleted_at IS NULL` is the right primitive, and the migration correctly `drop constraint`s (the object created by `Migration20260313222043.ts:16` genuinely is a constraint, not an index) before creating the partial index.
- I audited every other `CustomerUser` email lookup for the ambiguity this change introduces — two rows can now share `(tenant_id, email_hash)` — and they all already filter `deletedAt: null`: `customerUserService.ts:41` and `:53`, `subscribers/autoLinkCrmReverse.ts:62-67`, `api/admin/users.ts:50`, `setup.ts:267`. No lookup regresses.
- Tenant/organization scoping is preserved on the new query; the `409` shape matches the route's existing error envelope and is documented in the OpenAPI method doc.

## 🔴 Blockers

None.

## 🟠 Majors

### 1. The conflict guard misses accounts stored under the legacy lookup hash

`packages/core/src/modules/customer_accounts/services/customerInvitationService.ts:167`

The guard matches on a single hash, `hashForLookup(invitation.email)`. Every other `CustomerUser` email lookup in this module matches on `lookupHashCandidates()` instead — `customerUserService.ts:41`, `subscribers/autoLinkCrmReverse.ts:56`, `api/admin/users.ts:86` — because `lookupHashCandidates` returns `[primary, legacy]` (`packages/shared/src/lib/encryption/aes.ts:159-163`).

**Failure scenario:** an active portal user whose row still carries the legacy hash. The new guard queries the primary hash, finds nothing, and falls through. The partial unique index cannot catch it either, because the two rows hold *different* `email_hash` values. The accept then inserts a second active `CustomerUser` for the same address. `findByEmail` (`customerUserService.ts:35-47`) is a `findOne` over `$in [primary, legacy]`, so which of the two accounts a subsequent login resolves to is decided by whatever row order Postgres returns — for an authentication lookup that is not an acceptable outcome.

**Fix:** match the module's idiom —

```ts
emailHash: { $in: lookupHashCandidates(invitation.email) },
```

or, better, delegate to the service that already gets this right: `customerUserService.findByEmail(invitation.email, invitation.tenantId)`. Note the *insert* at `:194` should keep using `hashForLookup` (new rows must carry the primary hash); only the lookup needs the candidate list.

### 2. Check-then-insert: the raw 500 this PR set out to remove is still reachable

`packages/core/src/modules/customer_accounts/services/customerInvitationService.ts:173-186` (guard) and `:235` (flush)

The guard and the write are separated by a bcrypt hash, a role query and a flush. Nothing serializes them.

**Failure scenario:** a user double-clicks the accept button. Two requests carry the same token; both `findByToken` see `acceptedAt: null`; both pass the guard; both `em.create(CustomerUser, …)`; the second `flush()` at `:235` violates the partial unique index and raises `UniqueConstraintViolationException` — the exact raw 500 from #5532, just through a narrower window. The same happens for two invitations to one address accepted concurrently.

**Fix:** keep the guard as the friendly fast path, and make the constraint itself the authority. Wrap the flush and translate the DB error onto the same domain error, so both paths land on the 409:

```ts
try {
  await this.em.flush()
} catch (error) {
  if (error instanceof UniqueConstraintViolationException) {
    throw new CustomerInvitationEmailConflictError(invitation.email)
  }
  throw error
}
```

(`UniqueConstraintViolationException` from `@mikro-orm/core`.) This also covers finding 1 for the primary-hash case at no extra cost.

### 3. No integration coverage for the DB-level behaviour that is the actual bug

`packages/core/src/modules/customer_accounts/services/__tests__/customerInvitationService.test.ts:201-266`

Both new tests mock `EntityManager` end to end. The first one does not observe the partial index at all — it stubs `findOne(CustomerUser, …)` to return `null` and then asserts the *where clause* contained `deletedAt: null`. That is a test of the argument this PR passes, not of the behaviour it claims to fix.

**Failure scenario:** a typo in the index expression, a migration that does not apply cleanly, or a `deleted_at` predicate that Postgres cannot use for uniqueness — every one of those ships green, because no test ever touches a database. `ephemeral-integration (none)` on this PR confirms nothing DB-level was selected to run, and `AGENTS.md` requires integration coverage for affected API paths on behaviour changes.

**Fix:** add a spec beside the module's existing ones (`packages/core/src/modules/customer_accounts/__integration__/TC-CACC-INVITE-PERSON-001.spec.ts` is the closest template) covering invite → accept → soft-delete the portal user → re-invite the same address → accept succeeds; plus a second case asserting `409` when an *active* account already owns the address. Keep it self-contained per `.ai/qa/AGENTS.md` — create fixtures in setup, clean up in teardown.

## 🟡 Minors

### 4. `down()` fails on exactly the data this change enables

`packages/core/src/modules/customer_accounts/migrations/Migration20260827160000_customer_accounts.ts:11-12`

The rollback re-adds the full `UNIQUE (tenant_id, email_hash)`. Once the partial index has been live, a soft-deleted row legitimately coexists with an active row on the same key — which is the whole point — so `add constraint` errors with a duplicate key. The rollback is unrunnable precisely when someone would need it.

**Fix:** either purge the losing rows in `down()` before re-adding the constraint, or state the limitation in a comment on the method so an operator is not surprised mid-incident.

### 5. Module documentation now contradicts the schema

`packages/core/src/modules/customer_accounts/AGENTS.md:374`

> Unique constraint on `(tenantId, emailHash)` prevents duplicates

That is no longer true — it is a partial unique index scoped to `deleted_at is null`, and that distinction is exactly what the next contributor needs before writing another email lookup.

**Fix:** update the line, and add that every `CustomerUser` email lookup MUST filter `deletedAt: null` (which is now load-bearing, not defensive).

### 6. The error message embeds an email address

`packages/core/src/modules/customer_accounts/services/customerInvitationService.ts:18`

`super(\`An account with the email address ${email} already exists\`)` puts PII into a string that lands in logs and error trackers, in a module that otherwise treats emails as sensitive (encrypted at rest, hashed for lookup).

**Fix:** keep the address on the typed `email` property and drop it from the message, e.g. `super('[internal] An account with this email address already exists')`.

### 7. Hardcoded user-facing string on the new 409

`packages/core/src/modules/customer_accounts/api/invitations/accept.ts:44`

The response message is inline rather than routed through `t('customer_accounts.errors.…')`. The neighbouring 400 at `:49` has the same shape, so this is consistency with existing debt rather than a new pattern — but this PR is the one adding the line, and the boy-scout rule applies.

## 🧪 Validation

CI on `31959b6` is fully green — `lint`, `test` (7m42s), `ds-lint`, `audit-scope`, `scope`, `prepare`, `documents-multi-instance`, `ephemeral-integration (none)` all pass; nothing failing, nothing pending. The head is `MERGEABLE` against `develop` with no conflicts. I did not re-run the gate locally, because CI already ran it on this exact SHA and its result is the authoritative one; the caveat worth naming is that `ephemeral-integration` reporting `(none)` is itself the evidence for finding 3.

## 📋 Notes

- No prior review feedback existed on this PR, so there are no inherited findings to account for.
- Left as a draft — not promoting to ready while majors are open.
- Autofix: skipped by explicit instruction for this run; the author will apply fixes.

---
🤖 `om-auto-review-pr`

adeptofvoltron:
🤖 `om-auto-review-pr` — label rationale

- 🔁 `review` → `changes-requested`: the review found 3 majors (legacy-hash gap in the new conflict guard, a check-then-insert race that still surfaces the original raw 500, and no integration coverage for the DB-level fix). Per the `om-code-review` verdict rule, any un-waivered major requests changes.
- 🐛 `bug`, `priority-medium`, `risk-medium`, `needs-qa`: unchanged — still accurate. The PR ships a migration, so it does not qualify for the automated-verification `skip-qa` exemption.
- 📋 Left as a **draft**: not promoting to ready while majors are open.

adeptofvoltron:

3 majors, 4 minors, 0 blockers. CI is fully green on `31959b6` and the head is conflict-free, so the verdict is driven entirely by review findings. Autofix skipped by explicit instruction for this run — the author applies the fixes. PR left as a draft; assignee kept for the fix pass.

adeptofvoltron:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-09-03T07:20:55Z.

adeptofvoltron:
🤖 `om-auto-qa-pr` — 📸 UI verification (evidence-only)

**Verdict: ✅ pass — after this run found and fixed a defect that made the PR's headline improvement inert.**

The diff touches no `.tsx` file, so the automated-verification rule would have allowed a "UI: n/a". I ran the browser QA anyway, because the change alters what the customer-facing invite acceptance page does, and because nothing else in the PR proved the new migration actually applies to a real database. That turned out to matter.

### 🧪 Environment

A disposable Postgres database (`om_qa_pr5709`) initialised from scratch with `mercato init`, so the new migration was applied by the normal migration path rather than reviewed by eye, and the app served from a **production build** (`yarn build:packages && yarn build:app`, both cache-miss on this head) at `http://localhost:3117`. The invitation token was taken from the captured invitation email, not planted, so the flow is the real one a customer would follow.

### ❌ What the QA found (now fixed in `a65f01e`)

The first run failed at the duplicate-invitation step: the API answered **500**, not the 409 this PR adds. The server log showed the service throwing `CustomerInvitationEmailConflictError` correctly, and the accept route failing to recognise it — the route and the service are bundled into different Next.js chunks, so each holds its own copy of the class and `error instanceof CustomerInvitationEmailConflictError` compares two unrelated constructors. The clean 409 never fired in a built app; a duplicate accept still produced the raw 500 this PR set out to remove.

Worth stating plainly, because it is the reason this run was worth doing: **neither the unit tests nor the integration suite could catch it.** The unit tests mock the EntityManager inside a single module instance, and `TC-CACC-INVITE-REUSE-001` passed in CI on the broken code — the ephemeral harness does not reproduce production chunking. Only the production-build browser run did.

The fix follows the idiom `packages/shared/src/lib/crud/errors.ts` already documents for exactly this hazard: a `Symbol.for` marker on the error plus an exported `isCustomerInvitationEmailConflictError` guard. The same reasoning was applied to the `UniqueConstraintViolationException` catch, which trusted `instanceof` across the same kind of boundary.

### ✅ Re-run against the rebuilt production app

| Step | Expected | Result |
| --- | --- | --- |
| Create a portal user, then delete it (soft delete confirmed in the database) | row keeps `deleted_at` set | ✅ |
| Re-invite the same address | `201` — the invitation is created | ✅ |
| Accept the re-invitation on `/acme-corp/portal/invite?token=…` | account created, redirect to the portal dashboard | ✅ `201`, landed on `/acme-corp/portal/dashboard` |
| Invite that address again, now owned by an **active** account, and accept | clean `409` with a readable message, never a 500 | ✅ `409`, page renders "An account with this email address already exists" |
| Browser console during the flow | no errors | ✅ none |

`TC-CACC-INVITE-REUSE-001` was also run against this same production build and passes there too, so the spec now has evidence it holds in the mode where the bug lived.

### 📸 Screenshots



*The invite acceptance page reached with a real token from the invitation email, for the address of the soft-deleted user.*



*The fix for #5532 working end to end: the re-invitation is accepted and the new portal account lands on its dashboard. Before the partial unique index this insert violated `customer_users_tenant_email_hash_uniq` and returned a 500.*



*The other side of the constraint: inviting an address an active account already owns now renders the localized conflict message from the 409 instead of failing opaquely.*

### 🏷️ Labels

This run is **evidence-only**: it sets no QA labels and does not add `qa-approved`. `needs-qa` stays on — the PR ships a migration, so it does not qualify for the automated-verification exemption, and a QA reviewer still owns the sign-off. Only `screenshots` is added, to record that visual evidence is attached.

adeptofvoltron:

- 🚀 `merge-queue` (replacing `changes-requested`): every finding from the `om-auto-review-pr` review has been fixed in this branch — the three majors and all four minors — so the verdict that drove `changes-requested` no longer holds. The PR is not merged by this run; the label means it is queued behind the remaining gates.
- 🐛 `bug`: unchanged. The PR fixes a defect (#5532) rather than adding behaviour.
- 🟡 `priority-medium`: unchanged. A portal user whose email was reused could not accept their invitation, which is a real block for the affected customer but not a platform-wide outage.
- 🟠 `risk-medium`: unchanged, and still the right call. The change ships a migration that rewrites a unique constraint on `customer_users` — a table portal authentication reads on every login — so the blast radius is wider than a self-contained code fix, even though the constraint is only narrowed.
- 🧪 `needs-qa`: unchanged and deliberately kept. The automated-verification `skip-qa` exemption requires an unchanged database structure, and this PR ships a migration, so it does not qualify no matter how much automated coverage rides along. `qa-approved` is **not** applied — this run produced QA evidence, not a QA sign-off, and with the QA gate on, that sign-off is still owed by a human reviewer.
- 📸 `screenshots`: added because this run attached browser evidence for the invite-acceptance flow, including the screenshot of the conflict message that the QA run's own finding produced.
- 🤖 `in-progress` → `ci-monitoring`: the work is finished and reported, so the claim is released; `ci-monitoring` records that a CI-result follow-up comment on the current head is still owed. It is not a lock — anyone may act on this PR now.

adeptofvoltron:

**Verdict: ✅ merge-ready, pending a human approval and QA sign-off.** Every finding from the review is fixed, the branch is up to date with `develop`, and the change is now covered by unit tests, a real-database integration spec, and browser evidence from a production build.

### 🔀 Base merge

`origin/develop` was merged into the branch before any other work (commit `8f84624`), cleanly and with no conflicts, so review, CI and QA all judged the real merge result rather than a stale diff. The base did not advance again during the run.

### 🔍 Review findings — all seven addressed

The `om-auto-review-pr` verdict on `31959b6` requested changes over 3 majors and listed 4 minors. Each is accounted for below; none was deferred.

**Major 1 — the conflict guard missed accounts stored under the legacy lookup hash.** Fixed. The active-account lookup now matches `emailHash: { $in: lookupHashCandidates(invitation.email) }`, the same idiom `customerUserService.findByEmail` and the other module lookups use, while the insert keeps writing the primary hash. Unit-tested by asserting the candidate list reaches the query and that the created row still carries the primary hash.

**Major 2 — check-then-insert could still surface the raw 500.** Fixed. The flush is wrapped and a unique-constraint violation is translated onto `CustomerInvitationEmailConflictError`, so the friendly guard stays a fast path and the partial index is the authority. Two concurrent accepts now both end on the 409.

**Major 3 — no integration coverage for the DB-level behaviour.** Fixed by `TC-CACC-INVITE-REUSE-001`, which creates a portal user, soft-deletes it, re-invites the address and accepts (the path the old full unique constraint turned into a 500), then invites the now-active address again and asserts a clean 409. It reads the raw token from the captured invitation email exactly as `TC-AUTH-033` does, and asserts the `deleted_at` state directly in Postgres. One correction to the finding's supporting detail, so nobody repeats it: `ephemeral-integration (none)` is not evidence that nothing DB-level ran — `(none)` is the shard label for a single-runner affected-modules run, and the job did execute the module's integration specs.

**Minor 4 — `down()` failed on exactly the data the change enables.** Fixed. The rollback now purges losing soft-deleted duplicates before restoring the full constraint, keeping the active row (or the newest one). The destructiveness is deliberate and documented in a comment on the migration: it only removes rows that were already logically deleted, and the alternative is a rollback that cannot run when someone needs it.

**Minor 5 — module documentation contradicted the schema.** Fixed. `customer_accounts/AGENTS.md` now describes the partial unique index, states that `deletedAt: null` is load-bearing on every `CustomerUser` email lookup, and points at `lookupHashCandidates`.

**Minor 6 — the error message embedded an email address.** Fixed. The message is now `[internal]`-prefixed and address-free; the address stays on the typed `email` property.

**Minor 7 — hard-coded user-facing string on the new 409.** Fixed for the line this PR adds: it resolves through `translate('customer_accounts.errors.emailAlreadyExists', …)` with the key added to all five locale files. The neighbouring 400s were left alone — that is pre-existing debt the review itself excused — and are tracked in the follow-up below rather than silently expanded into here.

### 📸 UI verification — and the defect it caught

The diff touches no `.tsx` file, so a "UI: n/a" would have been defensible. The browser QA ran anyway, and it earned its keep: against a **production build** the duplicate-invitation accept returned **500, not the 409 this PR adds**. The route's `instanceof CustomerInvitationEmailConflictError` was always false, because the route and the service are bundled into different chunks and each holds its own copy of the class — so the PR's headline improvement was inert in a built app, and #5532's raw 500 was still reachable.

Neither test layer could see it. The unit tests mock a single module instance, and `TC-CACC-INVITE-REUSE-001` **passed in CI on the broken code**, because the ephemeral harness does not reproduce production chunking. Fixed in `a65f01e` with the `Symbol.for` marker idiom that `packages/shared/src/lib/crud/errors.ts` already documents for this exact hazard, plus an `isCustomerInvitationEmailConflictError` guard for callers; the `UniqueConstraintViolationException` catch got the same treatment (name and Postgres `23505` fallback) since it crosses the same boundary. Re-verified green in the browser afterwards. Full evidence and screenshots are in the QA comment above.

### 🧪 Validation

The repo's configured gate was run locally in order — `build:packages`, `generate`, `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck`, `lint`, `build:app` — all pass, with `typecheck` re-run with `--force` after the last edit so nothing was served from cache. `@open-mercato/core` is green at 12,036 tests, and the invitation-service suite is at 17 including the new regression tests. One unrelated failure in `@open-mercato/shared` (`dynamicLoader.generatedCacheRecovery`) is a local environment artefact — a `tsx` IPC socket path that is too long under this agent's temp directory — and passes when the suite is run with a short `TMPDIR`; CI is unaffected and reported `test` green.

**CI is still running on this head (`a65f01e`).** Branch protection plus the QA-approval gate hold the actual merge; this run's verdict covers the work, not a green run. The previous head (`43066fd`) was fully green, including the new integration spec. A follow-up comment will report the CI outcome for this head.

### 🔁 Follow-up filed

- #5860 — localize the remaining hard-coded error strings in the `customer_accounts` invitation API routes. Deferred rather than fixed here because it is pre-existing debt across two routes and five locale files, and expanding this PR into it would widen a targeted bug fix.

### 🚀 Hand-off

This skill never merges. The PR is handed to `om-approve-merge-pr 5709`, which re-checks every gate before squash-merging. Two things are still owed by a human: a **review approval** — GitHub refuses self-approval and this PR is authored by the automation account, so a second reviewer is required — and the **QA sign-off**, since `needs-qa` is on and the QA gate blocks the merge until `qa-approved` is added. The QA evidence above is intended to make that sign-off quick, not to replace it.

adeptofvoltron:

**✅ All required checks are green on `a65f01e`**, the head this run reported against. The merge-readiness verdict in the run summary above therefore stands unchanged, and the pending-CI disclosure it carried is now resolved.

| Check | Result |
| --- | --- |
| `prepare` | ✅ |
| `lint` | ✅ |
| `ds-lint` | ✅ |
| `scope` | ✅ |
| `audit-scope` | ✅ |
| `test` | ✅ |
| `ephemeral-integration (none)` | ✅ 37 passed, 0 failed, 0 flaky |
| `documents-multi-instance` | ✅ |

Nothing failed and nothing is still pending. `[code]smith`, `audit`, `comment`, `mutate`, `docker-build` and `merge-coverage` are skipped by their own path/branch conditions, as on every PR of this shape.

Two results worth naming rather than leaving in the aggregate, because they are the coverage the review asked for:

- `TC-CACC-INVITE-REUSE-001` ran and passed in the integration shard, so the soft-delete → re-invite → accept path and the 409 on an active address are both proven against a real database with the new migration applied by the normal migration path.
- The invitation-service unit suite is green at 17 tests, including the regressions added for the legacy lookup hash, the unique-violation translation, and the bundle-boundary guard that the browser QA's finding produced.

The `ci-monitoring` label is removed: this follow-up is the one it was recording, and no further comment will come from this agent. The PR still needs a human review approval (self-approval is refused for the automation account) and the QA sign-off that `needs-qa` gates, after which `om-approve-merge-pr 5709` can take it.

patzick (approved):

PR #5710, opened 2026-08-27
feat(dev): structured runtime diagnostics, in-app banner and opt-in gateway

### LINKED ISSUE #5369: Dev runtime splash and gateway diagnostics
# Intelligent Dev Runtime Splash and Gateway

## TLDR

Open Mercato's development splash will evolve from a startup-progress page into
an optional runtime feedback loop. The default direct topology remains
backward-compatible. An opt-in gateway topology owns the public application
port, proxies a healthy Next.js runtime, and serves actionable diagnostics when
the runtime is starting, degraded, recovering, or unavailable.

The capability is delivered as one phased runtime initiative because the
incident model, browser feedback channel, recovery actions, and gateway all
share the same supervisor state and security boundary. The phases may ship as
separate implementation increments, but they are governed by this one
specification. The gateway is not a production reverse proxy and is never
enabled by default.

## Overview

The feature builds on the existing root dev orchestrator, app runtime wrapper,
splash HTML, child-state file, log classifier, warmup checks, action-token
pattern, and `global-error.tsx` fallback. It adds a structured incident model
without removing the legacy `ready`/`failed` fields.

The design follows established development-server patterns while keeping the
implementation small and local to the dev runtime:

- Vite documents explicit HTTP proxy and WebSocket/HMR forwarding, including
  the need to handle WebSocket traffic separately from normal HTTP routes:
  [Vite server options](https://vite.dev/config/server-options).
- webpack-dev-server documents proxy routing, HTML fallback/bypass behavior,
  WebSocket support, and host validation as separate concerns:
  [webpack-dev-server configuration](https://webpack.js.org/configuration/dev-server/).
- Node's HTTP server exposes the low-level `upgrade` event needed for raw
  WebSocket forwarding, which keeps HMR handling explicit rather than hidden in
  a generic fetch proxy:
  [Node.js HTTP API](https://nodejs.org/api/http.html).
- Next.js custom-server guidance is the relevant integration seam for a server
  that owns the public port while delegating application rendering:
  [Next.js App Router guides](https://nextjs.org/docs/app/guides).

### Topologies

The current default remains unchanged:

```text
PORT=3000                -> Next.js / Open Mercato runtime
OM_DEV_SPLASH_PORT=4000  -> standalone splash and supervisor API
```

The opt-in topology is:

```text
PORT=3000                -> Open Mercato dev runtime gateway
internal port            -> managed Next.js runtime
OM_DEV_SPLASH_PORT=4000  -> supervisor diagnostics API (optional local access)
```

The gateway uses an automatically selected loopback upstream port unless
`OM_DEV_UPSTREAM_PORT` is supplied. `OM_DEV_RUNTIME_MODE=direct` is the
default; `OM_DEV_RUNTIME_MODE=proxy` enables the gateway.

## Problem Statement

The current splash can expose generation, migration, bootstrap, and startup
output, but it is not the owner of the public application port. A developer
opening the application URL during a failed startup may therefore see an
unavailable server, a generic framework error, or no actionable explanation.

The runtime wrapper already detects process-level failures, but the state model
conflates startup failure with post-ready degradation. The current ready-state
normalization can hide a later runtime issue, while `global-error.tsx` reports
only inside the browser and does not feed the supervisor or splash state.

The system needs one bounded feedback loop:

```text
process/logs/readiness/browser
          -> incident collector
          -> generation-aware runtime state
          -> splash/gateway feedback
          -> allowlisted recovery action
          -> restart/warmup/probe
```

## Goals

- Detect startup and post-ready runtime problems from process exits, classified
  logs, readiness/warmup failures, browser rendering errors, and continuous
  health probes.
- Show one structured, user-readable incident summary at the application URL
  when gateway mode is enabled, including what failed, when it started, and
  what recovery actions are safe.
- Show the same active incident in a small dev-only banner while an application
  page is already open, in both direct and gateway modes.
- Keep browser diagnostics available in direct mode as well as gateway mode.
- Preserve the existing direct-mode topology and `/status` behavior by default.
- Keep the supervisor alive after a startup-stage failure only in gateway mode,
  so a developer can run an allowlisted recovery action without restarting the
  terminal process. Direct mode keeps its current shutdown behavior.
- Use the existing full startup warmup as the `READY` gate and add a bounded,
  lightweight continuous probe after `READY`.
- Preserve HTTP API/static responses and HMR WebSocket behavior when the
  upstream runtime is healthy.
- Keep diagnostics local, bounded, redacted, development-only, and scoped to a
  single runtime generation.
- Cover every new diagnostic/action/health path and the key browser/gateway
  paths with self-contained tests.

## Non-goals

- No production reverse proxy, new production health service, or telemetry
  backend. An existing infrastructure health endpoint may be reused without
  exposing runtime incident details.
- No database entity, tenant record, organization record, or persistent audit
  history for runtime incidents.
- No arbitrary shell command execution from the splash.
- No automatic migration rollback or destructive cache deletion in the MVP.
- No conversion of application business errors into splash errors merely because
  they returned HTTP 4xx/5xx.
- No change to the production application URL, authentication, module API
  contracts, or tenant scoping.
- No new runtime dependency unless a later implementation review shows that
  Node's standard HTTP APIs cannot safely satisfy the proxy contract.

## Decisions

| Topic | Decision | Rationale |
| --- | --- | --- |
| Document shape | One phased specification | The same supervisor state, generation, action token, and browser collector are shared by all phases. |
| Browser diagnostics | Enabled in direct and gateway modes | `global-error.tsx` is useful even when the standalone splash is the only available UI. |
| Startup failure lifecycle | Gateway keeps the supervisor alive; direct mode preserves shutdown | Recovery controls are valuable only when the public gateway can continue serving feedback. |
| Readiness | Existing full warmup gates `READY`; continuous light probe detects later outages | This keeps the proven startup contract while covering post-ready failures without repeatedly executing expensive login flows. |
| Default mode | `direct` | Existing scripts, ports, and workflows remain unchanged unless explicitly opted in. |
| Gateway dependency | Node core `http`/`https` and explicit `upgrade` handling | Avoids adding a production dependency for a dev-only path and keeps security behavior reviewable. |
| Incident persistence | In-memory state plus bounded local diagnostic lines | Runtime incidents are per-process and must not become application data. |
| Recovery action set | `generate`, `migrate`, `restart`, and bounded log/status inspection | Small allowlist maps to existing dev lifecycle stages; cache reset remains a later opt-in action. |

### Runtime state and compatibility view

Add a generation-aware runtime state owned by the root supervisor. The existing
state writer remains the compatibility boundary for current consumers. Its
legacy fields are derived from the new state as follows:

| Runtime health | `ready` | `failed` | Meaning |
| --- | ---: | ---: | --- |
| `starting` | `false` | `false` | Runtime is being built, launched, or warmed up. |
| `ready` | `true` | `false` | Warmup passed and the latest continuous probe is healthy. |
| `degraded` | `true` | `false` | Runtime served at least once but one or more bounded signals are unhealthy; keep application traffic flowing when possible. |
| `recovering` | `false` | `false` | A permitted recovery action or restart is active. |
| `unavailable` | `false` | `true` | Gateway cannot safely serve the upstream, or a hard failure prevents startup. |

`failed` must no longer be used as the only signal for a post-ready issue. The
normalization logic must preserve an active incident even when the last known
state was `ready`.

The state payload is versioned and bounded:

```ts
type RuntimeHealth =
  | 'starting'
  | 'ready'
  | 'degraded'
  | 'recovering'
  | 'unavailable'

type RuntimeIssue = {
  id: string
  fingerprint: string
  code: string
  source: 'process' | 'log' | 'warmup' | 'probe' | 'browser'
  severity: 'warning' | 'error'
  title: string
  detail?: string
  firstSeenAt: string
  lastSeenAt: string
  occurrences: number
  generation: number
  path?: string
  digest?: string
  recovery?: 'generate' | 'migrate' | 'restart'
}

type RuntimeStatus = {
  schemaVersion: 1
  generation: number
  health: RuntimeHealth
  ready: boolean
  failed: boolean
  updatedAt: string
  upstream: {
    configuredPort: number
    actualPort?: number
    publicUrl: string
  }
  issueSummary?: RuntimeIssue
  incidents: RuntimeIssue[]
  recovery?: {
    action: 'generate' | 'migrate' | 'restart'
    startedAt: string
    busy: boolean
    lastExitCode?: number
  }
  legacy: {
    failureLines: string[]
    failureCommand?: string
    failureStage?: string
  }
}
```

The implementation may use an internal richer type, but the public status
response must retain the bounded shape. Keep at most 20 active/last incidents,
200 diagnostic log lines, and 8 KiB per browser report after redaction.

### State machine

```mermaid
stateDiagram-v2
    [*] --> starting
    starting --> ready: warmup passed
    starting --> unavailable: hard startup failure
    ready --> degraded: probe or browser threshold reached
    degraded --> ready: recovery threshold reached
    degraded --> recovering: recovery action accepted
    unavailable --> recovering: recovery action accepted
    recovering --> starting: runtime restarted
    recovering --> ready: action completed and warmup passed
    recovering --> unavailable: action failed
```

Rules:

- Each process launch increments `generation`. A delayed line, probe, or
  browser report from an older generation is ignored for current status updates.
- A single browser error is recorded and deduplicated, but does not mark the
  runtime unavailable. Thresholds and severity determine a `degraded` state.
- A hard process exit, failed startup warmup, or repeated probe failure can
  produce `unavailable`.
- Recovery returns through startup warmup; it cannot set `ready` directly.
- `ready` is a compatibility projection, not a terminal state.

### Detection and incident classification

Use one collector interface for all sources:

```ts
recordRuntimeSignal(signal: RuntimeSignal): RuntimeStatus
```

The collector must:

- normalize known failure signatures to stable `code` values;
- create a deterministic fingerprint from source, code, sanitized path, and
  digest rather than raw stack text;
- deduplicate repeated lines and browser retries within the same generation;
- redact tokens, cookies, authorization headers, connection strings, and
  values matching known secret patterns before state or log storage;
- attach a recovery action only when the classifier can justify it;
- never execute a command while classifying a signal.

Signals:

1. Process lifecycle: non-zero exit, signal termination, failed child spawn,
   failed stage command, or restart loop.
2. Classified logs: compile/runtime/migration/bootstrap failures already
   recognized by the app wrapper and root stage runner.
3. Startup warmup: `/login`, authenticated login, and `/backend` failures.
4. Browser: `global-error.tsx`, `window.error`, unhandled rejection, chunk-load
   failure, and selected failed runtime requests.
5. Continuous probe: internal `GET /api/healthz` after the startup warmup has
   passed.

### Readiness and continuous probe

The existing startup warmup remains the required `READY` gate. It must continue
to run the current sequence and preserve its current authentication/session
semantics.

After `READY`, the supervisor runs a lightweight probe against the managed
upstream:

- `GET /api/healthz` on the loopback upstream URL;
- timeout: 1,500 ms;
- interval: 5,000 ms;
- enter `degraded` after 3 consecutive failures;
- enter `ready` after 2 consecutive successes;
- reset counters on generation change or accepted recovery action.

These values are defaults and are configurable through the environment
variables listed below. The probe must not repeatedly execute the full login
warmup, must not log successful requests, and must not expose dependency
details in its public response.

The standard `GET /api/healthz` contract is already used as an infrastructure
health check by the standalone-app template. It should be aligned with the
monorepo app using the existing minimal 200/503 semantics, not expanded into a
runtime diagnostics service. It may remain available to deployment health
checks; only the runtime status bridge, incident details, and diagnostics
routes are dev-only. The endpoint must not return credentials, connection
strings, tenant data, or stack traces.

### Browser feedback channel

`global-error.tsx` remains a client boundary because it needs browser APIs and
must render when the normal app tree cannot. It reports a sanitized diagnostic
to a dev-only collector using a per-generation token exposed in a dev-only
`<meta>` element by the app layout. A small dev-only client reporter mounted by
the layout captures `window.error`, unhandled rejections, chunk-load failures,
and selected failed runtime requests while the normal app is running. It is a
client island, not a global provider, and shares the same bounded reporter
helper with `global-error.tsx`.

The same layout mounts `DevRuntimeDiagnosticsBanner` when development
diagnostics are enabled. The banner polls the structured status endpoint every
2 seconds and renders only when the status is `starting`, `degraded`,
`recovering`, or `unavailable`. It is intentionally separate from
`global-error.tsx`: the error boundary reports a browser failure, while the
banner communicates supervisor/runtime state even when the current page is
still rendered successfully.

Status transport:

- gateway mode: `GET /__open-mercato/status` on the same origin;
- direct mode: dev-only `GET /api/__open-mercato/status`, backed by the
  supervisor's generation-aware local state file and protected by the same
  per-run token. The root supervisor passes the state-file path and token to
  the managed app process; the app route never becomes a production data
  endpoint.

Banner behavior:

- `degraded`: non-blocking warning banner; the current page remains usable;
- `starting`/`recovering`: informational banner with progress/status detail and
  no duplicate recovery control while an action is busy;
- `unavailable`: blocking-but-dismissible runtime error banner when the page is
  still present, with a retry/restart path;
- `ready`: remove the banner and clear only view-local state;
- dismissal hides the current `generation:fingerprint` until a new fingerprint
  or generation appears; it never changes runtime state;
- repeated occurrences of the same fingerprint update the details/counter but
  do not flash the banner on every poll.

For the missing-relation example, the banner's primary copy is localized as
“Database schema mismatch” with the concise detail “Relation `sandboxs` is
missing”. “Show details” expands the source, PostgreSQL code, occurrence count,
and timestamp. “Run migrations” is offered only when the classifier maps the
incident to the `migrate` recovery action and always requires confirmation.
The raw `relation "sandboxs" does not exist` message remains available in the
redacted technical details/logs view.

Transport:

- direct mode: `POST /api/__open-mercato/diagnostics` is handled by a dev-only
  Next route and appends a bounded line to the supervisor's local diagnostic
  sink; the resulting incident is visible through the existing standalone
  splash status/logs view while the public application URL remains direct
  Next.js;
- gateway mode: `POST /__open-mercato/diagnostics` is handled by the gateway
  and records directly into the supervisor state;
- when the collector is unavailable, the browser silently falls back to the
  existing rendered error state; it must never block or loop on reporting.

Accepted reports are limited to:

```json
{
  "kind": "global-error",
  "message": "sanitized message",
  "digest": "optional-next-digest",
  "path": "/backend/example",
  "stack": "optional bounded stack",
  "timestamp": "2026-08-18T10:00:00.000Z"
}
```

The server applies schema validation, length limits, redaction, origin/token
validation, and a per-process rate limit before forwarding the report. A report
does not expose the full stack by default in the splash; a developer may use
the bounded logs endpoint or terminal output for deeper inspection.

### Gateway routing

The gateway owns only the development control namespace and the decisions that
depend on runtime health.

| Request | Gateway behavior |
| --- | --- |
| `GET /__open-mercato/status` | Return current structured status. |
| `GET /__open-mercato/logs` | Return a bounded, redacted snapshot with cursor metadata. |
| `POST /__open-mercato/diagnostics` | Validate and record a browser report. |
| `POST /__open-mercato/actions/{action}` | Validate token, allowlist, and busy/generation rules, then enqueue one recovery action. |
| Navigational HTML while `starting`, `recovering`, or `unavailable` | Serve the splash fallback. |
| Navigational HTML while `ready` or `degraded` | Proxy to the Next.js upstream; `degraded` remains reachable. |
| `/api/*`, static assets, and `/_next/*` | Proxy as application traffic; preserve the upstream status and body. |
| HTTP `upgrade`/HMR WebSocket | Forward the raw upgrade to the upstream when available; close cleanly when unavailable. |

The gateway must distinguish navigation requests from API/static requests using
method and `Accept` headers. It must never replace a business API error with the
splash HTML. For a control request, use a stable JSON error response rather
than HTML.

Proxy implementation requirements:

- use Node core `http`/`https` request and stream APIs;
- strip hop-by-hop headers and reconstruct `Host`, `Origin`,
  `X-Forwarded-For`, `X-Forwarded-Host`, and `X-Forwarded-Proto` deliberately;
- preserve cookies, content length/encoding semantics, abort propagation, and
  backpressure;
- handle `upgrade` separately and test HMR with both an available and an
  unavailable upstream;
- bind the public server to the existing local development bind address and
  reject non-loopback exposure unless explicitly configured;
- use strict port validation and fail clearly on a port collision;
- do not proxy gateway-owned paths to the application, even if a module later
  creates a colliding route.

### Recovery actions

The splash exposes these controls in the first implementation:

- `generate`: run the existing generator stage, then restart/warmup as the
  current lifecycle requires;
- `migrate`: run the existing migration stage only after explicit confirmation,
  then restart/warmup;
- `restart`: restart the managed runtime through the existing supervisor;
- `logs`: view the bounded diagnostic snapshot; this is a read-only view, not a
  mutating recovery action.

The action runner must reuse existing lifecycle ownership. It must not create a
second Next manager, bypass stage ordering, or invoke arbitrary command text.
Only one mutating action may run per generation. Every action receives a
run-local token, action id, start time, timeout, exit code, and bounded output.
An old-generation action completion cannot overwrite the new generation state.

`migrate` is not automatically undoable: the UI must state that it applies
database changes and that rollback is a separate operator task. `generate`,
`restart`, and log inspection are recoverable by rerunning the supervisor.
Cache reset is explicitly deferred until an existing safe reset helper can be
reused without broad deletion.

## Architecture

```mermaid
flowchart LR
    Process["Process lifecycle"] --> Collector["Incident collector"]
    Logs["Classified logs"] --> Collector
    Warmup["Startup warmup"] --> Collector
    Probe["Continuous health probe"] --> Collector
    Browser["global-error browser report"] --> Collector
    Collector --> State["Generation-aware runtime state"]
    State --> Splash["Splash status and recovery UI"]
    State --> Gateway["Public dev gateway"]
    Gateway -->|"ready/degraded application traffic"| Next["Managed Next.js upstream"]
    Gateway -->|"starting/unavailable navigation"| Splash
    Actions["Allowlisted action"] --> Supervisor["Existing root supervisor"]
    Supervisor --> Process
    Supervisor --> Warmup
```

### Ownership map

| Concern | Owner | Required change |
| --- | --- | --- |
| State transitions, generation, action queue | Root supervisor in `scripts/dev.mjs` and new focused runtime helpers | Add structured state without breaking legacy state fields. |
| Runtime failure classification and warmup | `apps/mercato/scripts/dev.mjs` and its template counterpart | Emit typed signals and retain existing warmup behavior. |
| Splash rendering and polling | `scripts/dev-splash.html` and template counterpart | Render structured incidents, state-specific actions, and bounded logs. |
| Public gateway/proxy | New root dev-runtime gateway helper | Own public port only in explicit proxy mode. |
| Browser reporting | `global-error.tsx`, dev-only client reporter, app layout meta, and diagnostics route | Mirror into the standalone-app template. |
| In-app runtime feedback | `DevRuntimeDiagnosticsBanner` mounted by the app layout | Mirror into the standalone-app template; use the direct status bridge or gateway status endpoint. |
| Health contract | `GET /api/healthz` in the app and template | Minimal 200/503 response; no application data. |
| CLI-managed runtime | `packages/cli/src/mercato.ts` | Keep existing restart ownership and pass through supported mode/config. |

### Suggested implementation units

Use focused, testable helpers rather than expanding `scripts/dev.mjs` into one
large state machine:

- `scripts/dev-runtime-state.mjs`: state schema, transitions, compatibility
  projection, generation checks, and redaction;
- `scripts/dev-runtime-probe.mjs`: interval/timeout/counter policy;
- `scripts/dev-runtime-gateway.mjs`: route ownership, HTTP stream proxy, and
  `upgrade` forwarding;
- `scripts/dev-runtime-diagnostics.mjs`: token, report validation, bounded
  local sink, and incident conversion;
- small adapters in the existing root/app runtime scripts for process/log/
  warmup signals.

## Data Models

This feature creates no ORM entity and no migration. Runtime data is ephemeral:

- in-memory `RuntimeStatus` owned by one supervisor process;
- one per-run token and generation number;
- bounded local NDJSON diagnostic sink under the existing development runtime
  state/log directory;
- bounded terminal/splash log buffer.

The sink is deleted or rotated with the existing dev runtime cleanup policy and
must not be committed. It contains sanitized diagnostic lines only. It must
not contain cookies, authorization headers, passwords, database URLs, tenant or
organization identifiers, or arbitrary request bodies.

The state writer must use atomic replacement or an equivalent write strategy so
the splash never reads a partially written JSON document. A stale writer must
be rejected using generation and process identity checks.

### Status and logs

Gateway mode adds:

- `GET /__open-mercato/status` → `200 RuntimeStatus`;
- `GET /__open-mercato/logs?cursor=<opaque>` → `200 { lines, nextCursor,
  generation }`;
- control-path failures → JSON `{ error: { code, message } }` with no stack.

Direct app mode adds a dev-only bridge for the in-app banner:

- `GET /api/__open-mercato/status` → `200 RuntimeStatus`;
- `403` for a missing or invalid per-run token;
- `404` when dev diagnostics are disabled or the supervisor state is not
  available;
- the route reads only the supervisor-provided local runtime state file and
  returns no application/module/tenant data.

Direct mode keeps the existing standalone splash `GET /status` contract and
existing coding/git/action endpoints. It may expose the same structured payload
as additive fields, but existing fields and polling behavior remain valid.

### Browser diagnostics

Direct app route:

- `POST /api/__open-mercato/diagnostics`;
- `202 { accepted: true, issueId }` for a valid bounded report;
- `400` for invalid schema/size;
- `403` for a missing or invalid per-run token;
- `404` when dev diagnostics are disabled.

Gateway route:

- `POST /__open-mercato/diagnostics` with the same body and response rules.

Both routes are dev-only, rate-limited, redacted, and excluded from business
module discovery. The implementation must use the repository's normal runtime
validation conventions for the app route and must not add an unauthenticated
production diagnostics endpoint.

### Recovery actions

`POST /__open-mercato/actions/{action}` accepts `generate`, `migrate`, or
`restart`. It returns `202 { accepted: true, actionId, generation }` when the
action is queued. It returns:

- `400` for an unknown action or malformed request;
- `403` for an invalid token/origin;
- `409` when another mutating action is active or the generation is stale;
- `503` when the supervisor is no longer available.

Direct mode keeps current standalone splash action routes. If they are extended
to use the structured action runner, the existing response shape must remain
valid and new fields must be additive.

### Health

`GET /api/healthz` is the minimal infrastructure health contract reused by the
dev probe:

- `200 { status: "ok", timestamp }` when the app can serve the health check;
- `503 { status: "degraded", timestamp }` when the app cannot satisfy the
  required local runtime dependencies;
- no secrets, dependency connection details, tenant data, or stack traces. It
  is not the runtime incident/status bridge and does not expose the splash's
  `RuntimeStatus` payload.

## Security and Privacy

- Gateway and collector bind to loopback by default. Any non-loopback bind must
  be explicit and produce a clear warning.
- A per-run random token protects control and browser-diagnostic routes. The
  token is exposed only to the local dev browser through the dev-only meta
  element and is never persisted in source control.
- Validate method, path, origin, content type, body size, field lengths, and
  rate limits before recording or executing anything.
- Actions are a fixed enum mapped to existing supervisor functions. User input
  cannot select a binary, append shell flags, or change a working directory.
- Strip secrets and sensitive headers before storage and before returning logs.
- Do not include stack traces in normal splash copy. Provide only bounded,
  redacted details and a link/button to the local logs view.
- Keep the diagnostic route disabled outside the development runtime. Add tests
  proving that a production build/configuration cannot enable it accidentally.
- Treat browser reports as untrusted data; a report can inform the state but
  cannot force a recovery action or change runtime configuration.

## UI and Frontend Architecture

The main splash is static HTML/JavaScript, not a React page. It continues to
poll status and renders translations from its existing runtime mechanism. Add
state-specific copy for `starting`, `ready`, `degraded`, `recovering`, and
`unavailable`, plus accessible action feedback and a bounded details/logs view.

The browser error fallback remains a small client island. It should show the
existing safe fallback first, send one best-effort report, and offer a retry
that reloads the current page. It must not add a global provider or a heavy
client dependency. Any migration confirmation must use the shared confirmation
primitive rather than `window.confirm`; dialogs must support Escape and
Cmd/Ctrl+Enter according to the UI contract.

### DevRuntimeDiagnosticsBanner

The banner is the in-app counterpart to the full splash. It is rendered from
the existing application layout only when the dev runtime diagnostics flag is
enabled. It must not be rendered in production builds or when the status bridge
is disabled.

Required presentation:

```text
Runtime degraded · Database schema mismatch
Relation `sandboxs` is missing                         [Show details] [×]
```

The expanded details view shows the localized title, sanitized detail, source,
stable error code, first/last seen time, occurrence count, generation, and
available actions. It does not show raw stack text until the developer opens
technical details. The action set is derived from the incident and may contain
`Run migrations`, `Restart`, `Retry`, and `View logs`; absent actions are not
rendered. During Phases 2–3, before the mutating action runner is available,
the banner exposes only `Retry` and `View logs`; Phase 4 enables `Restart`,
`Run migrations`, and other allowlisted actions. `Run migrations` always uses
the shared confirmation dialog.

Interaction and accessibility requirements:

- use shared semantic/status UI primitives and tokens in the React surface;
- use `role="status"` with a polite live region for `degraded` and
  `recovering`, and `role="alert"` for `unavailable`;
- provide a visible close/dismiss button with an accessible label; dismissal is
  local view state only and must not acknowledge or clear the incident;
- keep the banner keyboard reachable, use visible focus, and support the
  existing dialog Escape and Cmd/Ctrl+Enter rules;
- localize every user-facing title, detail label, action, and error state, with
  a safe provider-independent fallback for `global-error.tsx`;
- do not show a banner for ordinary business 4xx responses or duplicate the
  same fingerprint on every 2-second poll;
- on narrow viewports, actions wrap into a second row without horizontal
  scrolling or obscuring the error summary.

Server/client boundary ledger:

| File/surface | Boundary | Reason |
| --- | --- | --- |
| `scripts/dev-splash.html` | Static server/client script | Existing splash is served outside Next.js. |
| `global-error.tsx` | Client | Error boundary, `window`, `document`, and best-effort reporting. |
| Dev runtime reporter | Client | Registers bounded `window`/promise/request listeners only when dev diagnostics are enabled. |
| `DevRuntimeDiagnosticsBanner` | Client | Polls local runtime status and renders dismissible incident feedback in the open app. |
| App layout meta injection | Server | Reads dev-only runtime configuration without adding a provider. |
| Diagnostics and health routes | Server | Token validation, redaction, and process-local sink access. |
| Gateway | Node server | Stream proxy, status fallback, and WebSocket upgrade handling. |

The implementation must mirror app/template changes in the same task, including
the error fallback, layout metadata, diagnostics route, health route, locale
keys, and tests. It must use existing shared UI primitives for any React
confirmation or status surface and existing semantic status tokens when React
UI is involved. The standalone splash may retain its existing CSS mechanism.

## Configuration

| Variable | Default | Contract |
| --- | --- | --- |
| `OM_DEV_RUNTIME_MODE` | `direct` | Enum `direct\|proxy`; invalid values fail with a readable configuration error. |
| `OM_DEV_UPSTREAM_PORT` | auto-selected loopback port | Optional integer 1–65535; must not equal the public port. |
| `OM_DEV_RUNTIME_PROBE_INTERVAL_MS` | `5000` | Positive bounded integer. |
| `OM_DEV_RUNTIME_PROBE_TIMEOUT_MS` | `1500` | Positive bounded integer less than the interval. |
| `OM_DEV_RUNTIME_PROBE_FAILURE_THRESHOLD` | `3` | Positive bounded integer. |
| `OM_DEV_RUNTIME_PROBE_RECOVERY_THRESHOLD` | `2` | Positive bounded integer. |
| `OM_DEV_RUNTIME_DIAGNOSTICS` | enabled with dev splash | Enum/boolean; disabled in CI or when explicitly off. |
| `OM_DEV_RUNTIME_BANNER` | enabled with dev diagnostics | Enum/boolean; controls the in-app banner independently from event collection. |
| `OM_DEV_SPLASH_BIND` | existing default | Reuse existing bind behavior; do not silently broaden exposure. |

Document new variables in the app and standalone-app examples where runtime
configuration is currently documented. Do not add secrets to `.env.example`.

## Migration & Backward Compatibility

- `OM_DEV_RUNTIME_MODE` defaults to `direct`, preserving existing ports,
  process ownership, startup shutdown behavior, and terminal workflow.
- Existing `GET /status`, legacy state fields, coding/git actions, and direct
  splash polling remain supported. New fields are additive.
- The gateway control namespace is reserved under `/__open-mercato/*` and must
  not alter module routes. `GET /api/healthz`, the dev diagnostics route, and
  the direct-mode status bridge are additive app routes and must be mirrored in
  the standalone-app template.
- No existing module export, event ID, ORM schema, tenant API, or production
  route is renamed or removed.
- Before implementation changes a contract surface, re-read
  `BACKWARD_COMPATIBILITY.md`, add an additive bridge where required, and note
  any deprecation in `UPGRADE_NOTES.md`. This feature should require no
  deprecation because direct mode and existing control paths remain intact.
- The gateway is opt-in and can be removed or disabled without affecting users
  who run the default topology.

### Phase 1 — Structured incident state

- Extract state transitions, generation checks, bounded incidents, redaction,
  and legacy projection into focused root helpers.
- Update root/app runtime adapters to emit process, log, and warmup signals.
- Fix ready-state normalization so an active post-ready incident is visible.
- Keep direct-mode failure shutdown behavior unchanged.
- Add unit tests for all transitions, deduplication, stale generations, and
  legacy payload compatibility.

### Phase 2 — Browser collector and continuous probe

- Add the dev-only diagnostics route, per-run token/meta injection, and the
  `global-error.tsx` best-effort reporter in both app and template.
- Add the dev-only `DevRuntimeDiagnosticsBanner` in both app and template,
  including the direct-mode status bridge and a shared status adapter. Phase 2
  does not require the gateway; its gateway adapter is connected in Phase 3.
- Add/align `GET /api/healthz` in both app and template.
- Add the continuous probe and default threshold policy.
- Add route, rate-limit, redaction, browser error-boundary, status-bridge, and
  banner interaction tests.

### Phase 3 — Opt-in gateway

- Start the managed Next runtime on the internal port and bind the gateway to
  the existing public port.
- Implement gateway control routes, HTML fallback, application proxying, and
  explicit WebSocket upgrade forwarding.
- Keep the supervisor alive after startup stage failure only in proxy mode.
- Add integration coverage for navigation, API/static passthrough, upstream
  errors, port collisions, and HMR upgrade behavior.

### Phase 4 — Recovery UX and hardening

- Connect allowlisted action buttons to the existing supervisor lifecycle.
- Add action progress, busy/conflict handling, bounded logs, confirmation copy,
  and stale-generation protection.
- Add security regression tests, resource/latency checks, template parity
  checks, and documentation for the new environment variables.
- Evaluate cache reset separately; do not add it to the MVP without a safe
  existing helper and an explicit follow-up decision.

### Integration coverage matrix

The implementation PR must ship self-contained tests for every affected path:

| Area | Required scenarios |
| --- | --- |
| Direct status | Existing `/status` response, additive structured fields, legacy polling. |
| Gateway status/logs | `GET /__open-mercato/status`, bounded `logs`, malformed cursor, redaction. |
| Browser collector | Valid report, invalid body, missing token, stale generation, rate limit, disabled mode. |
| In-app banner | Direct and gateway status polling, startup/degraded/recovering/unavailable copy, details, dismissal, new-fingerprint reappearance, action visibility, keyboard/a11y, narrow viewport. |
| Health | `/api/healthz` 200 and 503 responses in app and template. |
| Actions | Generate/restart/migrate acceptance, unknown action, invalid token, busy 409, stale generation, failed command. |
| HTML routing | Splash for starting/unavailable navigation; upstream HTML for ready/degraded navigation. |
| Traffic proxy | API 4xx/5xx preserved, static and `/_next` passthrough, cookies/headers/body/abort behavior. |
| HMR | Successful WebSocket upgrade and clean behavior while upstream is unavailable. |
| Browser UI | Startup, degraded, recovering, unavailable, retry, diagnostic details, keyboard and accessibility behavior. |
| Lifecycle | Gateway stays alive after stage failure; direct mode still exits as before; recovery returns through warmup. |

Tests must create their own runtime fixtures, avoid seeded/demo data, clean up
processes and temporary files, and run without a live external dependency.

## Risks and Mitigations

| Risk | Severity | Mitigation | Residual risk |
| --- | --- | --- | --- |
| Gateway serves splash HTML for an API request | High | Route by method/Accept/path class; integration-test API 4xx/5xx preservation. | A new unusual client Accept header may need an explicit rule. |
| HMR breaks through the gateway | High | Separate `upgrade` implementation, raw stream tests, and fallback to direct mode. | Tooling-specific WebSocket behavior may require adapter work. |
| Old process overwrites current state | High | Generation and process identity checks on every signal and action completion. | A crash during atomic replacement can still lose the newest diagnostic line. |
| Local control endpoint executes unintended commands | High | Loopback default, per-run token, origin checks, fixed action enum, no shell text. | A local browser with the token can still invoke intended dev actions. |
| Probe creates false degraded states | Medium | Dedicated lightweight health contract, timeout/threshold hysteresis, no full warmup loop. | A genuinely overloaded dev process may oscillate until thresholds are tuned. |
| Browser report contains secrets | High | Schema limits, redaction, no request-body capture, bounded stack, tests with secret fixtures. | Unknown secret formats may remain in arbitrary error text. |
| Direct-mode collector is unavailable during a full startup failure | Medium | Best-effort browser reporting plus process/log/warmup signals; no blocking client retry loop. | Browser errors cannot be reported before the app can serve the diagnostic route. |
| Migration action is not automatically reversible | High | Explicit confirmation and copy; do not claim rollback; preserve existing migration ownership. | Operator remains responsible for database recovery. |
| In-app banner cannot read direct-mode state | Medium | Supervisor-provided local status file, dev-only `/api/__open-mercato/status` bridge, token validation, and a graceful no-banner fallback. | A process/filesystem failure can still leave the open page without live status. |
| Template and monorepo drift | Medium | Same-task mirror requirement and parity tests against the two app surfaces. | Future template-only changes can still diverge if not covered by CI. |
| Resource growth from logs/incidents | Medium | Fixed limits, rotation, no successful probe logs, bounded response sizes. | Very noisy failures can still increase short-term CPU during redaction. |

## Validation and Acceptance Criteria

The implementation should use the smallest relevant gate sequence and record
whether it ran locally or through the Docker app runner. At minimum:

- unit tests for state, redaction, probe, routing, and actions;
- app/template typecheck and lint for changed files;
- integration/browser tests from the coverage matrix;
- `yarn generate` only if discovery or generated registries are changed;
- no `yarn db:migrate`; this feature has no database migration.

Acceptance is met when:

1. A post-ready runtime error appears as `degraded` or `unavailable` instead of
   being hidden by `ready` normalization.
2. A browser `global-error.tsx` or dev-reporter signal reaches local state in
   both modes when the runtime can serve the collector; in direct mode it is
   visible through both the in-app banner and standalone splash status/logs
   view.
3. An open application page shows the dev-only runtime banner in both direct
   and gateway modes, with structured details, dismissal, appropriate actions,
   and recovery when the status returns to `READY`.
4. The gateway serves readable splash feedback during startup/failure and
   forwards normal HTML, API/static traffic, and HMR when the upstream is ready.
5. Gateway startup failures leave the supervisor alive for recovery actions;
   direct mode retains its current shutdown behavior.
6. Recovery actions are fixed, authenticated, generation-aware, serialized,
   and covered by tests.
7. The default direct topology keeps its existing public port, splash port,
   `/status` contract, polling behavior, and direct-mode shutdown semantics.
8. No secrets, tenant data, or new production diagnostics/telemetry endpoints
   are introduced; any existing infrastructure health contract remains
   minimal.
9. App and standalone-app template behavior is mirrored and documented.

### Files and guides reviewed

- `AGENTS.md`
- `.ai/specs/AGENTS.md`
- `.ai/skills/om-spec-writing/SKILL.md`
- `.ai/skills/om-spec-writing/references/spec-template.md`
- `.ai/skills/om-spec-writing/references/spec-checklist.md`
- `.ai/skills/om-spec-writing/references/compliance-review.md`
- `.ai/skills/om-spec-writing/references/frontend-architecture-contract.md`
- `BACKWARD_COMPATIBILITY.md`
- `packages/cli/AGENTS.md`
- `packages/ui/AGENTS.md`
- `packages/create-app/AGENTS.md`
- `packages/create-app/template/AGENTS.md`
- `.ai/specs/2026-05-28-compile-and-route-warmup-speedup.md`
- `.ai/specs/2026-03-25-safe-build-dev-coexistence.md`
- `.ai/specs/implemented/2026-04-02-dev-structural-regeneration-and-cold-start-optimization.md`
- `.ai/specs/implemented/SPEC-034-2026-02-21-dev-ephemeral-runtime.md`

### Research and design checklist

- [x] Scope is one capability with explicit phases and non-goals.
- [x] Existing code paths and compatibility constraints are named.
- [x] Browser/server boundaries and template parity are explicit.
- [x] API routes, status fields, actions, security, and failure behavior are
  specified.
- [x] Integration coverage includes all affected API and key UI paths.
- [x] No ORM data model or migration is introduced.
- [x] Recovery, rollback limitations, risks, and residual risks are explicit.
- [x] Official development-server references were compared and linked.
- [x] No production dependency is required by the proposal.

### Review verdict

The proposed design is ready for implementation planning. The main
implementation risks are HTTP/WebSocket proxy correctness, safe local action
execution, and keeping the app/template surfaces synchronized; all three have
dedicated phases, contracts, and test requirements.

### 2026-08-18

- Created the unified spec for intelligent splash diagnostics and the opt-in
  runtime gateway.
- Resolved the open decisions: one phased document, browser diagnostics in
  both modes, gateway-only supervisor persistence after startup failure, and
  hybrid warmup plus continuous probe readiness.
- Added the dev-only `DevRuntimeDiagnosticsBanner` for already-open pages in
  direct and gateway modes, including the direct-mode status bridge,
  dismissal, accessibility, action visibility, and integration coverage.

patzick:
## Summary

The dev splash could show startup output, but it did not own the public application port and its `ready` normalization erased any failure that appeared *after* startup — so a post-ready runtime error was silently invisible. This PR turns the splash into a runtime feedback loop: process exits, classified logs, startup warmup, browser errors and a continuous `/api/healthz` probe collapse into one generation-aware runtime state that drives the splash, a new in-app banner, and an opt-in local gateway. Failures now render as structured, actionable incidents (`relation "sandboxs" does not exist` → **Database schema mismatch · Relation `sandboxs` is missing**, with a justified `migrate` recovery) instead of a raw log tail. The default `direct` topology, ports, `/status` contract and shutdown behaviour are unchanged; the gateway is opt-in via `OM_DEV_RUNTIME_MODE=proxy` and is never a production reverse proxy.

## Changes

- **Supervisor helpers** (`scripts/dev-runtime-*.mjs`): generation-aware state machine with signature classification, stable fingerprinting, secret redaction and the legacy `ready`/`failed` projection; env config with readable errors; probe policy; bounded NDJSON channels; the HTTP/`upgrade` gateway; and the allowlisted, serialized action runner.
- **Fixed ready-state normalization** so an active post-ready incident survives instead of being erased by the `ready` compatibility projection.
- **Dev-only app surface**: `/api/dev-runtime/{status,diagnostics,actions/[action],logs}`, a bounded browser reporter wired into `global-error.tsx` plus `window.error`/unhandled rejections, and `DevRuntimeDiagnosticsBanner` (recovery actions, `migrate` behind the shared confirmation dialog, inline log tail, dismissal scoped to `generation:fingerprint`).
- **Opt-in gateway**: splash fallback for navigations while starting/recovering/unavailable, untouched API/static/`_next` passthrough, raw HMR upgrade forwarding, and supervisor survival after a stage failure — proxy mode only.
- **Template parity + docs**: every app change mirrored into the create-app template, `/api/healthz` added to the app, new spec and docs page, and the `OM_DEV_RUNTIME_*` variables documented in both `.env.example` files.

image
image
image
image

## Specification

- [ ] Yes
- [x] No (created a new spec)

**Spec file path:**
`.ai/specs/2026-08-18-dev-runtime-splash-and-gateway-diagnostics.md`

## Testing

- `yarn test:scripts` — 848 tests, 847 pass / 0 fail / 1 pre-existing skip
- `yarn workspace @open-mercato/shared test --testPathPatterns dev-runtime` — 37 pass
- `yarn workspace @open-mercato/ui test --testPathPatterns DevRuntime` — 24 pass
- `yarn jest --config jest.config.cjs --testPathPatterns "healthz|dev-runtime|global-error"` (in `apps/mercato`) — 22 pass
- Typecheck: `@open-mercato/shared`, `@open-mercato/ui`, and `apps/mercato` — all clean
- `yarn template:sync`, `yarn i18n:check-sync`, `yarn agents:check-budget`, `yarn test:repo-wide-guards` — pass
- Manual: verified end to end against a live `yarn dev` — a posted diagnostic surfaced as `degraded` with the classified title/detail/recovery, and **Restart runtime** incremented the generation and cleared incidents

## Checklist

- [ ] I added or updated integration tests in `.ai/qa/tests/` — not added; see note below.
- [ ] QA routing set — `needs-qa` (adds a rendered UI surface).

> **Coverage note:** this ships ~2,700 lines of unit/integration tests (including 23 gateway tests against real HTTP servers), but **no Playwright specs and no end-to-end `OM_DEV_RUNTIME_MODE=proxy yarn dev` run**. HMR through the gateway is therefore unverified against real Turbopack and is the highest-value thing to exercise in QA. Two bugs found only in manual testing — Next.js dropping a leading-underscore route folder, and `NODE_ENV` being an unusable dev/prod discriminator because `mercato dev` spawns Next with `NODE_ENV=production` — are now covered by regression tests.

### Design System Compliance
- [x] No hardcoded status colors — banner uses `status-{info,warning,error}-*` tokens
- [x] No arbitrary text sizes — uses the typography scale; `z-banner` comes from the existing `--z-index-banner` token
- [x] Empty state handled — the inline logs panel renders an explicit empty message
- [x] Loading state handled — the banner renders nothing until the first status poll resolves
- [x] `aria-label` on all icon-only buttons (the dismiss `IconButton`)
- [x] Uses existing DS components — `Button`, `IconButton`, `useConfirmDialog`

## Linked issues

Fixes #5369

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5710"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790432201&installation_model_id=432243&pr_number=5710&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5710&signature=e5269c7c14d7476158288ed00bd3aa0b9aa7db6d8119a0607d8db796a39d93bf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


patzick:
Label rationale (per `AGENTS.md` → PR Workflow):

- **`feature`** — new dev-runtime diagnostics capability, not a fix to existing behaviour.
- **`review`** — ready, non-draft, not yet in another pipeline state.
- **`priority-medium`** — developer-experience improvement; nothing is broken in production without it, but it fixes a real blind spot where a post-ready runtime error was silently hidden by the `ready` normalization.
- **`risk-medium`** — the default `direct` topology, ports, `/status` contract and shutdown behaviour are unchanged and the gateway is opt-in, but this does modify `scripts/dev.mjs` and the app layout, so a regression would affect everyone's `yarn dev`. Not `risk-low` for that reason; not `risk-high` because no production route, DB schema or module contract changes.
- **`needs-qa`** — the automated-verification exemption does **not** apply: this adds `.tsx` files under `packages/ui/src/` (`DevRuntimeDiagnosticsBanner`, `DevRuntimeReporter`), i.e. a manually exercisable UI surface.

Highest-value QA target: **HMR through the opt-in gateway** (`OM_DEV_RUNTIME_MODE=proxy yarn dev`). The gateway has 23 tests against real HTTP servers, but it was never run end to end against real Turbopack — that is the one genuinely unverified path.

patzick:
## Automated code review — PR #5710

**Validation source:** GitHub checks (run [33103257597](https://github.com/open-mercato/open-mercato/actions/runs/33103257597) on `051b03e24`) — 30 pass, 3 skipped by design (`audit`, `merge-coverage`, `[code]smith`). No local gate commands were run.

**Verdict:** no unfixed findings. Three defects were found and fixed on the branch rather than handed back.

### Findings (all fixed)

**1. Diagnostics sink dropped records after rotation — `scripts/dev-runtime-diagnostics.mjs`** *(fixed in `051b03e24`)*

`createDiagnosticsSink().drain()` detected rotation only via a changed inode. Inode numbers are recycled the moment they are freed, so a rotated sink routinely reappears under the inode the reader already recorded. The reader then kept its stale offset, sliced into the middle of the replacement line, failed `JSON.parse`, and silently dropped every queued browser report and recovery-action request.

The PR's own `sink recovers when the file is rotated underneath the reader` test covers exactly this path — it passed on APFS (monotonically increasing inode numbers) and failed on ext4 in CI. The reader now validates that the byte under its cursor is a record boundary, and a truncate-and-rewrite test pins the same-inode case independently of the filesystem's allocation policy.

**2. Gateway mode did not stop the startup pipeline after a stage failure — `scripts/dev.mjs`** *(fixed in `029c18b96`)*

`keepSupervisorAliveAfterStageFailure()` returns `true` so `reportStageFailure()` skips `shutdown()` — but it returns to `runStage()`, which returns normally to its caller. In `proxy` mode a failed `build:packages` therefore let `generate`, `initialize` and the app launch all proceed against a tree that had already failed to build. Direct mode was unaffected because `shutdown()` kills children and exits.

Surviving a stage failure should mean the gateway keeps serving, not that startup keeps going. Added a `startupAborted` latch guarding the stage runners and launchers, cleared on a deliberate recovery relaunch so `generate` / `migrate` / `restart` still work.

**3. Stale template-parity test — `packages/create-app/src/lib/template-railway.test.ts`** *(fixed in `b2453ee4c`)*

This PR adds `/api/healthz` to `apps/mercato` (the runtime probe targets it) and correctly drops the `TEMPLATE_ONLY_RELATIVE_FILES` exemption, but `template sync preserves Railway-only healthcheck routes` still asserted the exemption was present — the original CI failure. Rewritten to assert the new invariant (not exempted, and both copies byte-identical) and removed the now-stale "template-only `api/healthz`" note from `packages/create-app/AGENTS.md`.

Also routed the banner's error body through `readJsonSafe` instead of `.json().catch(() => null)`, which the root `AGENTS.md` bans; that removed an implicit `any` too.

### What was reviewed

`packages/shared/src/lib/dev-runtime/{routes,server,layout,report,redaction,types}.ts`; `DevRuntimeDiagnosticsBanner` + `DevRuntimeReporter`; `scripts/dev-runtime-{state,gateway,actions,supervisor,probe,config,diagnostics}.mjs`; the `scripts/dev.mjs` and `apps/mercato/scripts/dev.mjs` diffs; locale and template parity.

The dev-only gating is sound. Routes key off the supervisor-injected token plus the state-file paths rather than `NODE_ENV` (correct — `mercato dev` spawns Next with `NODE_ENV=production`), the status file must itself carry a matching token, comparisons use `timingSafeEqual`, and the recovery surface is a fixed enum that never builds a command from request input. Redaction is applied on both the write and ingest sides with a parity test. All five locale files are byte-identical between app and template with 27 `ui.devRuntime` keys each.

### Notes for QA (not blockers)

- `OM_DEV_RUNTIME_MODE=proxy` remains the highest-value thing to exercise — HMR through the gateway is still unverified against real Turbopack, as the PR description states.
- Worth exercising the gateway-mode stage-failure path specifically, since finding 2 changed its behaviour.
- With `OM_DEV_SPLASH_BIND=0.0.0.0` the splash page now embeds the runtime token and the recovery endpoints join the existing mutating surface. That is consistent with the documented "I trust everything that can reach this port" posture already covering GitHub publish and coding-tool spawn, so it is not a new class of exposure — but it is worth a conscious look.

`needs-qa` stays: this adds a rendered UI surface, so the automated-verification exemption does not apply.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T22:39:15Z. Other auto-skills will skip this PR until the lock is released.

pkarw (dismissed):
# 🔍 Code Review: feat(dev): structured runtime diagnostics, in-app banner and opt-in gateway

## 🎯 Summary

This PR turns the dev splash from a startup progress screen into a continuous runtime feedback loop. A generation-aware state store (`scripts/dev-runtime-state.mjs`) ingests process exits, classified log lines, warmup failures, a continuous `/api/healthz` probe and browser reports, collapses them into stable fingerprinted incidents with a justified recovery action, and publishes them through an atomically written status file that a token-protected, dev-only app bridge reads. On top of that sit a new in-app banner, a fixed recovery allowlist (`generate` / `migrate` / `restart`), a bounded log tail, and an opt-in local gateway that owns the public port in `proxy` mode.

The engineering standard here is high. The dev-only gating is genuinely well reasoned: it keys off the supervisor handshake (per-run token plus the injected state-file paths) rather than `NODE_ENV`, which is correct because `mercato dev` spawns Next with `NODE_ENV=production`; the status file must itself carry a matching token, so a stale run cannot be served; the recovery surface is a closed enum that never builds a command from request input; redaction is applied on both the write and the ingest side with a parity test pinning the two rule lists together; and every DOM write in the splash uses `textContent`, so untrusted browser-report text cannot become script. Template parity is real, not claimed — I verified all ten mirrored `scripts/*.mjs` / `dev-splash.html` files and all five locale dictionaries are byte-identical between `apps/mercato` and `packages/create-app/template`, with 27 `ui.devRuntime` keys in each locale.

I reviewed the full diff (76 files, +10,476/−41): the supervisor and state machine, the gateway, the diagnostics/action sinks, the `scripts/dev.mjs` and `apps/mercato/scripts/dev.mjs` wiring, the shared `dev-runtime` library and its routes, the banner and reporter, the healthz route, template parity, locales, the spec and the docs page.

One confirmed correctness defect blocks: the recovery restart re-ingests the dead runtime's signals into the fresh generation, so the flagship "click Run migrations → the incident clears" journey visibly appears to fail. The rest are convention and hardening items.

## Verdict

❌ **Request changes** — the High finding below is a reproducible bug in this feature's own core loop (a recovery action re-displays the incident it just fixed, and can mark a not-yet-listening runtime as ready), and three Medium findings cover a non-constant-time token comparison that contradicts the module's own stated control, a raw `fetch` where the repo mandates the shared client helper, and two new splash endpoints missing the host guard their siblings apply. Per this repository's review extension, Medium findings request changes.

## 🧪 Validation Gate

**Validation source: GitHub checks** — run [33103257597](https://github.com/open-mercato/open-mercato/actions/runs/33103257597) on head `051b03e2`, plus CodeQL run 33103254999. All 30 reported checks pass; `audit`, `merge-coverage` and `[code]smith` are skipped by design. No local gate commands were run, per the repo's checks-first rule.

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | Covered by the `prepare` job (5m22s). |
| `yarn generate` | ✅ PASS | Covered by the `prepare` job. |
| `yarn build:packages` (rebuild) | ✅ PASS | Covered by the `prepare` job. |
| `yarn i18n:check-sync` | ✅ PASS | `test` job, green; I independently confirmed all five locale files are byte-identical app↔template with 27 `ui.devRuntime` keys each. |
| `yarn i18n:check-usage` | ✅ PASS | `test` job, green. |
| `yarn typecheck` | ✅ PASS | `test` job (10m30s). |
| `yarn test` | ✅ PASS | `test` job; plus `ephemeral-integration` 1–15/15, `documents-multi-instance`, `docker-build`, `ds-lint`, `lint`, `mutate (shared, …dev-runtime…)` all green. |
| `yarn build:app` | ✅ PASS | Covered by `docker-build` (7m10s). |

## Findings

### ⚠️ Major

**1. `scripts/dev.mjs:1425` — a recovery restart re-ingests the dead runtime's signals into the fresh generation, so the recovery looks like it failed (High / major).**

`relaunchManagedRuntime()` calls `launchMonorepoAppDev()` / `launchStandaloneDev()`, which call `devRuntime.beginGeneration(...)`. `beginGeneration()` resets `lastIngestedChildSignalSeq` to `0` (`scripts/dev-runtime-supervisor.mjs:208`) — correct in isolation, because a fresh child restarts its own `runtimeSignalSequence` at 1. But nothing clears the child-state file on a relaunch: `writeSplashChildStateFileClear()` is only called when the splash server starts (`scripts/dev.mjs:1190`) and when it closes (`scripts/dev.mjs:1515`). The file therefore still holds the **dead** runtime's last `runtimeSignals` (and its `ready` flag) when the next 1s `sync()` — or any splash `/status` poll, which also calls `ingestChildState` via `getMergedSplashState()` — fires against the new generation.

Reproduced against the branch:

```
gen1: health=unavailable incidents=["db_relation_missing@gen1"]
gen2 right after beginGeneration: health=starting incidents=0
gen2 after one sync: health=unavailable incidents=["db_relation_missing@gen2 \"Relation `sandboxs` is missing\""]
```

The concrete failure: a developer hits the exact journey in the PR description — `relation "sandboxs" does not exist` → **Database schema mismatch** → clicks **Run migrations** → the migration succeeds and the runtime relaunches → and the banner immediately re-renders the *same* incident against the new generation, so the recovery reads as a no-op. It only self-heals once the new child reports ready and `markReady()` drops blocking incidents.

The same stale file leaks the `ready` flag: with a stale `ready: true`, one `sync()` on a brand-new generation flips `readySeen` from `false` to `true` before the new runtime is listening (verified the same way). That prematurely opens the probe gate (`shouldRun: () => store.isReadySeen() && …`), so a slow restart can accumulate the failure threshold and fabricate a `probe_unhealthy` "Runtime health check is failing" incident against a runtime that is merely still booting.

This also contradicts the invariant the code states for itself — "Signals, probes, and action completions from an older generation are ignored" (`scripts/dev-runtime-state.mjs:443`).

**Fix:** clear the hand-off channel before opening a new generation. In `relaunchManagedRuntime()` (and ideally in `launchMonorepoAppDev()` / `launchStandaloneDev()` immediately before `devRuntime.beginGeneration(...)`), call `writeSplashChildStateFileClear()` so the supervisor's reset cursor reads an empty channel rather than the previous runtime's. Please add a regression test alongside `scripts/__tests__/dev-runtime-lifecycle.test.mjs` that drives restart → sync with a stale child-state file present and asserts the new generation reports zero incidents and `readySeen === false`.

### 🔹 Minor

**2. `scripts/dev.mjs:1246` — the splash recovery endpoint compares the runtime token with `!==` instead of the constant-time helper.**

`if (req.headers['x-om-dev-runtime-token'] !== devRuntime.token)` is a plain string comparison, while every other token check in this feature routes through `isMatchingDevRuntimeToken` (`scripts/dev-runtime-diagnostics.mjs:28`, `packages/shared/src/lib/dev-runtime/server.ts:83`), whose own comment states the control explicitly: "Constant-time comparison so a local attacker cannot probe the token byte by byte through response timing." This is the one endpoint that runs `db:migrate` and `restart`, so it is the worst one to exempt, and the inconsistency will read to the next maintainer as "the constant-time check is optional".

**Fix:** import `isMatchingDevRuntimeToken` from `./dev-runtime-diagnostics.mjs` (already imported by the gateway) and use it here, normalizing the header value to a string first.

**3. `packages/ui/src/backend/dev/DevRuntimeDiagnosticsBanner.tsx:79,184,214 — raw `fetch` in UI code where the repo mandates the shared client helper.**

The root `AGENTS.md` is unambiguous: "Use `apiCall`/`apiCallOrThrow`/`readApiResultOrThrow` from `@open-mercato/ui/backend/utils/apiCall` — never use raw `fetch`", and writes from non-`CrudForm` pages are supposed to go through `useGuardedMutation(...).runMutation(...)`. This component issues three raw `fetch` calls, including the `POST` that triggers a recovery action.

I think the deviation is *defensible* here — `apiFetch` pulls in `flash`, the operations store and the partial-index store, and redirects to session refresh on 401, none of which a dev diagnostics overlay that must survive a broken provider tree should inherit — but nothing in the file says so, so this will be re-flagged by every future reviewer and by the checklist. (The raw `fetch` in `packages/shared/src/lib/dev-runtime/report.ts:101` needs no change: `@open-mercato/shared` must not depend on `@open-mercato/ui`.)

**Fix:** either route the calls through `apiCall`, or keep `fetch` and add a short comment above `fetchRuntimeStatus` recording the exemption and its reason, in the same style as the other explanatory comments in this file.

**4. `scripts/dev.mjs:1225` and `scripts/dev.mjs:1264` — the new `/runtime/status` and `/runtime/logs` splash endpoints skip the host guard their sibling applies.**

`/runtime/actions/*` calls `assertLocalSplashRequest(req, process.env)` (`scripts/dev.mjs:1235`), and the gateway applies the same guard to *every* control path including `/__open-mercato/logs` (`scripts/dev-runtime-gateway.mjs:139`). The two new read endpoints on the splash server apply neither, so with `OM_DEV_SPLASH_BIND=0.0.0.0` — or via DNS rebinding, which is exactly what `assertLocalSplashRequest`'s `Host`-header allowlist exists to stop — the redacted diagnostic tail and full runtime state are readable by a third party. The content is redacted and the pre-existing `/status` route is equally unguarded, so I am not calling this a blocker, but the same feature guarding the same data differently in two places is the kind of gap that gets copied forward.

**Fix:** apply `assertLocalSplashRequest` to `/runtime/status` and `/runtime/logs` as well, matching the gateway's `handleControlRequest`.

### 💅 Nit

**5. `scripts/dev-runtime-diagnostics.mjs:146` — `MAX_ACTION_REQUEST_BYTES` is exported but never referenced anywhere in the repo or the template.** Either enforce it in `validateActionRequest` (which currently applies no size bound, unlike `validateBrowserReport`) or drop the constant.

**6. `apps/mercato/src/app/api/healthz/route.ts:118` — the app now exposes an unauthenticated public endpoint that performs a database round-trip and a cache probe per request.** It was previously template-only, and `SELECT 1` is cheap, so this is not a real availability risk; it is worth a conscious decision (and possibly a short-TTL memo of the last result) since it is reachable in production, not just under the dev supervisor.

**7. No Playwright/integration coverage, and `OM_DEV_RUNTIME_MODE=proxy` has never been run end to end** — as the PR description states. I am not treating the repo's "integration tests ship in the same change" rule as breached here, because the dev-only routes are inert without a supervisor handshake and the ephemeral integration environment never runs `yarn dev`, so a Playwright spec would have to fabricate the supervisor env and would prove strictly less than the existing route tests. Recording it so the gap is visible, and routing it to QA instead (see below).

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — every `dev-runtime` export is new.
- [x] No required type field removed or narrowed; `runtime` is additive on the splash state payload and `RuntimeStatus` keeps the legacy `ready` / `failed` / `legacy` projection.
- [x] No HTTP route URL removed or renamed; `/api/dev-runtime/*`, `/api/healthz` and `/__open-mercato/*` are all new, and the kebab-case segment cannot collide with a snake_case module id.
- [x] No field removed or retyped in an existing response shape — the splash `GET /status` contract is extended, not changed.
- [x] No CLI command or flag renamed or removed; `OM_DEV_RUNTIME_MODE` defaults to `direct`, preserving the existing ports, process ownership and shutdown behaviour.
- [x] No database table or column renamed or removed — there is no schema change and no migration in this PR, correctly.
- [x] No config key renamed and no default changed silently; all seven `OM_DEV_RUNTIME_*` variables are new, documented in both `.env.example` files, and default to today's behaviour.
- [x] `BACKWARD_COMPATIBILITY.md` protected surfaces: none touched. `packages/create-app/src/lib/template-railway.test.ts` was rewritten rather than deleted, and the removed `TEMPLATE_ONLY_RELATIVE_FILES` entries now assert the stronger byte-identical invariant.

## 🧪 Test Coverage

Coverage is substantial and mostly well targeted: ~2,700 lines across 13 files. `dev-runtime-state.test.mjs` (36 cases) pins classification, fingerprint stability, redaction and the probe thresholds; `dev-runtime-gateway.test.mjs` (23 cases) exercises routing, proxying and HMR upgrade against real HTTP servers rather than mocks; `routes.test.ts` (473 lines) covers the dev-only route auth, origin rejection, size limits, rate limiting and the action allowlist; `DevRuntimeDiagnosticsBanner.test.tsx` (469 lines) covers rendering, a11y, dismissal scoping and the `migrate` confirmation; `dev-runtime-redaction-parity.test.mjs` keeps the supervisor and app redaction rule lists from drifting; and the rewritten `template-railway.test.ts` now asserts byte-identical app↔template healthz copies instead of asserting an exemption exists. No tests are skipped or `.todo`.

The gaps:

1. **The Major finding has no test.** `dev-runtime-supervisor.test.mjs:116` ("resets the ingestion cursor and clears the sink on a new generation") pins the supervisor half of the contract correctly, but nothing pins the `dev.mjs` half — that a relaunch presents the supervisor with an *empty* child-state channel. Add a case to `scripts/__tests__/dev-runtime-lifecycle.test.mjs` that writes a stale child-state file containing `runtimeSignals` and `ready: true`, triggers the relaunch path, and asserts the new generation reports zero incidents and `readySeen === false`.
2. **`keepSupervisorAliveAfterStageFailure` / the `startupAborted` latch** are covered for the "supervisor survives" behaviour, but I did not find a case asserting the *negative*: that after a `build:packages` stage failure in proxy mode, `runStage` / `startPackageWatch` / `launchMonorepoAppDev` all become no-ops. Since that latch was itself a bug fix in `029c18b96`, it deserves an explicit regression test.
3. **Proxy mode end to end** is unverified — see nit 7 and the QA note below.

## Notes for QA (not blockers)

- `OM_DEV_RUNTIME_MODE=proxy` with real Turbopack is the highest-value thing to exercise: fast refresh over the forwarded WebSocket upgrade, and a navigation while the upstream is still starting.
- Exercise the recovery loop specifically — trigger a schema mismatch, click **Run migrations**, and watch whether the incident clears or immediately reappears (Major finding 1).
- Exercise the gateway-mode stage-failure path, since `029c18b96` changed its behaviour.
- With `OM_DEV_SPLASH_BIND=0.0.0.0` the splash page embeds the per-run runtime token and the recovery endpoints join the existing mutating surface. That is consistent with the documented "I trust everything that can reach this port" posture already covering GitHub publish and coding-tool spawn, so it is not a new class of exposure — but it is worth a conscious look alongside Minor finding 4.

pkarw:
🏷️ label rationale

- **`changes-requested`** ← `review` — the review found one Major correctness defect (a recovery restart re-ingests the dead runtime's signals into the fresh generation) plus three Medium findings. Under this repository's review extension, Medium findings request changes.
- **`feature`** — unchanged; this adds a new dev-runtime diagnostics capability.
- **`needs-qa`** — kept. The change adds a rendered UI surface (the in-app banner) and the opt-in `proxy` topology has never been run end to end, so the automated-verification exemption does not apply.
- **`priority-medium`** — unchanged; a developer-experience improvement that closes a real blind spot but breaks nothing in production.
- **`risk-medium`** — unchanged and still right: the default `direct` topology, ports, `/status` contract and shutdown behaviour are untouched and every new runtime route is inert without the supervisor handshake, but the diff is large and touches the app layout, the create-app template and the dev entry point.
- **`in-progress`** — released below; this run's work is complete.

pkarw:
@patzick handing this back to you — `changes_requested`.

The blocking item is **Major finding 1**: `relaunchManagedRuntime()` (`scripts/dev.mjs:1425`) never clears the child-state file, while `beginGeneration()` resets the ingestion cursor, so the next 1s sync re-ingests the dead runtime's `runtimeSignals` (and its `ready` flag) into the fresh generation. Reproduced on `051b03e2`: after a recovery restart the same `db_relation_missing` incident reappears tagged `@gen2`, so clicking **Run migrations** looks like it did nothing until the new runtime reports ready. A stale `ready: true` additionally opens the probe gate early and can fabricate a `probe_unhealthy` incident against a runtime that is only still booting.

Three Medium items follow: the non-constant-time token comparison on the one endpoint that runs `db:migrate` (`scripts/dev.mjs:1246`), raw `fetch` in the banner where the repo mandates `apiCall` or a documented exemption, and the missing `assertLocalSplashRequest` guard on the two new `/runtime/*` read endpoints that their siblings and the gateway both apply.

Everything else is in good shape — CI is fully green on this head, template and locale parity are byte-exact, the dev-only gating and redaction design are sound, and no contract surface is touched. Full detail, suggested fixes and the two test gaps are in the review above.

pkarw:

1 Major, 3 Medium, 3 Low findings; validation source: GitHub checks (all 30 green on `051b03e2`), no local gate run. Autofix skipped — not this account's PR. No CI follow-up is pending: every required check has already settled green on this head.

andrzejewsky:
Addressed the review findings in 655381aab:

- clear stale splash child state before each runtime generation
- use constant-time dev-runtime token comparison
- route diagnostics-banner requests through apiCall
- host-guard splash runtime status and logs endpoints
- enforce the recovery-action request size limit
- add lifecycle and size-limit regression coverage; mirror script changes into the create-app template

Validation passed: focused dev-runtime tests, UI banner tests, package build, generation, i18n checks, repository typecheck, and app build.

The full `yarn test` gate remains red on an unrelated existing core test: `page.durationEntry.test.tsx` expects no total after an invalid pending duration but receives one. The isolated docs test passes.

patzick:
## Review findings addressed — ready for re-review

**Validation source:** GitHub checks (run [34459879336](https://github.com/open-mercato/open-mercato/actions/runs/34459879336) on `45baa04a5`) — 30 pass, 3 skipped by design. All 15 `ephemeral-integration` shards green.

### @pkarw's findings

| # | Finding | Status |
|---|---------|--------|
| 1 | Recovery restart re-ingests the dead runtime's signals into the fresh generation | ✅ `655381aab` — `writeSplashChildStateFileClear()` now runs before `beginGeneration()` in both launchers, with a lifecycle test pinning the ordering |
| 2 | Splash recovery endpoint used `!==` instead of the constant-time helper | ✅ `655381aab` — routed through `isMatchingDevRuntimeToken` |
| 3 | Raw `fetch` in the banner | ✅ `655381aab` + `45baa04a5` — see below |
| 4 | `/runtime/status` and `/runtime/logs` missing the host guard | ✅ `655381aab` — `assertLocalSplashRequest` applied to both |
| 5 | `MAX_ACTION_REQUEST_BYTES` exported but unenforced | ✅ `655381aab` — now enforced in `validateActionRequest` |
| 6 | Unauthenticated `/api/healthz` doing a DB round-trip | Left as-is — deliberate, `SELECT 1`, routed to QA |
| 7 | No Playwright coverage / `proxy` mode never run e2e | Acknowledged, routed to QA |

Finding 1 was a genuine bug I missed in my own earlier pass — thanks for catching it.

### One regression introduced by the finding-3 fix, now fixed

You offered two options and noted the deviation was defensible. `655381aab` took the `apiCall` route, which subjects the banner to `apiFetch`'s staff-auth handling: **a 403 throws `ForbiddenError` instead of returning the response** (`packages/ui/src/backend/utils/api.ts:278`).

The dev bridge answers 403 whenever the per-run token is stale — routine after restarting `yarn dev` with a tab still open, since the page keeps the previous run's token in its `<meta>`. `poll()`'s catch swallowed the throw, so `setStatus` was never called and **the banner froze on the dead runtime's incident instead of clearing**, logging a forbidden warning every 2s. The previous raw-`fetch` code returned `null` and hid the banner correctly.

CI stayed green because the existing tests only exercise this path as a **404**, which passes through `apiFetch` untouched.

`45baa04a5` keeps `apiCall` — so the convention you asked for holds — and disables both redirect hooks for the three dev-bridge calls via `apiFetch`'s existing `x-om-forbidden-redirect: 0` / `x-om-unauthorized-redirect: 0` escape hatch, with a comment recording why. Added a regression test covering the stale-token 403.

One caveat worth stating plainly: your underlying concern about `apiFetch`'s dependency graph — `flash`, the operations store, the indexes store — still stands. The banner now inherits those imports in a component whose whole purpose is to render when the tree is broken. I did not revert that choice, but it is a real residual trade-off rather than something the override headers resolve.

### Label change

Moving `changes-requested` → `review`: every blocking and minor finding is addressed and the full gate is green. `needs-qa` stays — this adds a rendered UI surface, so the automated-verification exemption does not apply, and the QA priorities from the review (proxy mode + HMR, the gateway-mode stage-failure path) are unchanged.

patzick:
## Auto review — verdict on `45baa04a5`

**Validation source:** GitHub checks (run [34459879336](https://github.com/open-mercato/open-mercato/actions/runs/34459879336)) — 30 pass, 3 skipped by design, all 15 `ephemeral-integration` shards green.

**Verdict: ✅ approved** — no Critical/High/Medium findings. Every defect raised in this PR's review history is fixed and verified on this head.

### Additional verification this pass

- **XSS surface (splash).** The claim that every splash DOM write uses `textContent` is not literally true — `renderActivities` (`dev-splash.html:1476`) and the coding menu (`:1782`) both assign `innerHTML`. Both escape every interpolation through `escapeHtml`, which correctly covers `& < > " '`, so the conclusion holds: untrusted browser-report and activity text cannot become script. Worth recording precisely, since the imprecise version of that claim is the kind of thing that gets copied forward into a future change that *doesn't* escape.
- **Parity.** `OM_DEV_RUNTIME_*` / `OM_DEV_UPSTREAM_PORT` sets are identical between `apps/mercato/.env.example` and the template; all ten mirrored `scripts/*.mjs` files and all five locale dictionaries are byte-identical; spec and docs page both present and wired into `apps/docs/sidebars.ts`.
- **Local script suite.** All 170 tests across the eight `dev-runtime-*` node:test files pass (these need no workspace install).

### Residual items — all Low, none blocking

1. **`dev-runtime-gateway.mjs:216` `proxyRequest`** attaches an `error` listener to the outbound `http.request` but not to `upstreamResponse` or the inbound `req`; `.pipe()` does not forward errors. A mid-stream upstream failure could surface as an unhandled stream error rather than the intended 502. Speculative without a repro, and confined to opt-in `proxy` mode — recording it, not blocking on it.
2. **Banner dependency graph.** Routing through `apiCall` pulls `flash`, the operations store and the indexes store into a component whose purpose is to render when the tree is broken. The override headers in `45baa04a5` fix the 403 behaviour but not this. Deliberate trade-off, already discussed above.
3. **`/api/healthz`** unauthenticated DB round-trip now reachable in production (@pkarw's nit 6) — deliberate, `SELECT 1`.

### Merge readiness — two hard blocks remain, neither is a code-quality issue

- `reviewDecision` is still **CHANGES_REQUESTED**; only @pkarw can dismiss or update their own review. GitHub reports `mergeStateStatus: BLOCKED`.
- The PR carries **`needs-qa` without `qa-approved`**, which per the root `AGENTS.md` is a hard merge gate. It adds a rendered UI surface, so the automated-verification exemption does not apply.

Leaving the pipeline label at `review` rather than `merge-queue` for now: staging a PR with an unresolved changes-requested review and no QA would misreport its state on the board.

patzick:
## Pipeline label: `review` → `merge-queue`

Applied at the maintainer's request following the passing auto review on `45baa04a5` (30 checks pass, 3 skipped by design, all 15 integration shards green; no Critical/High/Medium findings).

**This label stages the PR — it does not clear the merge gates, and two are still active:**

- **`needs-qa` is set without `qa-approved`.** Per the root `AGENTS.md` this is a hard merge gate: a PR carrying `needs-qa` MUST NOT be merged until `qa-approved` is applied by a maintainer. The PR adds a rendered UI surface, so the automated-verification exemption does not apply. QA has not run, and `OM_DEV_RUNTIME_MODE=proxy` with HMR through the gateway has never been exercised end to end.
- **`reviewDecision` is CHANGES_REQUESTED.** @pkarw's review is unresolved; only they can dismiss or update it. GitHub reports `mergeStateStatus: BLOCKED` independently of any label here.

QA priorities when it runs, unchanged from the review:

1. `OM_DEV_RUNTIME_MODE=proxy` — HMR through the gateway against real Turbopack.
2. The gateway-mode startup-stage-failure path (behaviour changed in `029c18b96`).
3. The `relation "sandboxs" does not exist` → **Run migrations** → incident clears journey (the bug fixed in `655381aab`).
4. Restart `yarn dev` with a tab open — the banner must clear rather than freeze on the dead runtime's incident (the regression fixed in `45baa04a5`).

PR #5714, opened 2026-08-27
fix(auth): treat ?tenant= on /login as a single-visit hint

Geeknauts-company:
## Summary

`/login` treats the `tenant` query parameter as a hint scoped to the current
visit — but it persisted it. The parameter was mirrored into
`localStorage['om_login_tenant']` **and** into a 14-day cookie of the same name,
and every later visit restored it. There is no TTL of its own and no clear after
a successful sign-in, so once a browser had seen the parameter the green
"you're logging in to *&lt;tenant&gt;*" banner stayed welded to `/login`
permanently.

This is not a rare path. `@open-mercato/onboarding` puts `?tenant=` on the login
link in three places:

- the workspace-ready e-mail — `onboarding/lib/ready-email.ts`
- the post-verification redirect — `onboarding/lib/verify-redirects.ts`
  (`redirectToLogin`)
- the `loginUrl` returned by `onboarding/api/get/onboarding/status.ts`

So **every self-serve signup passes through the parameter exactly once and then
carries the banner forever**, including long after the user has signed in, signed
out and come back. Any install running the onboarding module has this.

### The fix

The URL becomes the only source:

```ts
const tenantParam = (searchParams.get('tenant') || '').trim()
setTenantId(tenantParam || null)
```

The three helpers that implemented the persistence (`readTenantCookie`,
`setTenantCookie`, `clearTenantCookie`) and the two constants they shared had no
other call site in the repository, so they are removed with it.
`handleClearTenant` keeps doing the one thing that still means anything —
dropping the parameter from the URL and the state that mirrors it.

### What is deliberately unchanged

- **The parameter is still submitted.** The hidden `input[name="tenantId"]`
  still carries it into `POST /api/auth/login`, so the disambiguation that path
  needs when one e-mail address exists in two tenants is untouched (including its
  deliberately uniform 401 anti-oracle response).
- **A visit that carries the parameter behaves exactly as before** — the tenant
  is resolved through `/api/directory/tenants/lookup`, the banner shows, and an
  unknown tenant still produces the invalid-tenant box.

### Revised after review (round 1)

The banner and the submitted value were conflated, and only the banner is the
defect. Removing the persistence outright also removed the only thing that lets
a user whose address exists in more than one tenant sign in from a **bare**
`/login` — `findUsersByEmail` treats an ambiguous match as no user and falls
through to the uniform 401, the form has no tenant selector, and session
refresh, logout and the `ui` 401 handler all land there with no parameter.

So the two are now separate:

- **Banner and tenant lookup** — the URL parameter and nothing else. The
  welded-banner fix, unchanged.
- **Submitted `tenantId`** — falls back to a per-tab, per-address hint
  (`sessionStorage['om_login_tenant_hint']`) when the URL carries none. It dies
  with the tab, so there is no 14-day weld and no cross-session surprise, and it
  is keyed to the e-mail it was issued to, so a different person signing in on
  the same tab is not silently narrowed to somebody else's tenant. Written on
  submit, because `/login` takes no `email` parameter and that is the first
  moment both halves are known.

The client-set 14-day **cookie** is gone, as before.

**"Clear" clears again**: the URL parameter, the state, the per-tab hint, and the
`om_login_tenant` cookie `@open-mercato/onboarding` sets server-side. That
cookie's *setting* side is still out of scope here; its *clearing* side was
existing behaviour behind a button labelled "Clear" and this is the only in-app
way to drop it.

**Residual gap, stated rather than implied**: a multi-tenant user returning in a
new browser session to a bare `/login` still needs the activation link. That is
pre-existing — the real answer is a tenant selector on the form — and this PR
neither creates nor closes it.

### One thing this deliberately does not touch

`om_login_tenant` is **also** set server-side by `@open-mercato/onboarding`
(`lib/verify-redirects.ts`, `Set-Cookie`, 14 days) and read by
`onboarding/api/get/onboarding/status.ts` as an authorization check (403 when the
cookie is absent or names another tenant). That cookie is fed by the redirect,
not by this component — the only client of the status endpoint is
`/onboarding/preparing`, which is reached through `redirectToPreparing`, which
sets the cookie itself — so the gate keeps working with no change here.

I have left it alone rather than widening this PR into the onboarding package.
If you would like that cookie reconsidered too (it is a long-lived,
non-`httpOnly` tenant identifier used as an authorization input), I am happy to
open a follow-up — but it is a different mechanism with a different threat model
and it deserves its own diff.

## Changes

- `packages/core/src/modules/auth/frontend/login.tsx` — the fix. `-35/+15`.
- `packages/core/src/modules/auth/__tests__/login-tenant-single-visit.test.tsx` —
  new, 5 cases. Follows the conventions of the neighbouring
  `login-acl-challenge.test.tsx` (jsdom pragma, same module mocks).

## Specification

- [ ] Yes
- [ ] No (created a new spec)

A bug fix confined to one effect and three now-dead helpers in a single
component. Happy to write one if you would rather have it.

## Testing

Run from a clone with its own `node_modules`, Node 24.

- **New suite: 5/5 pass.**
- **Not vacuous:** against the unfixed `login.tsx`, 2 of the 5 fail — "persists
  the tenant nowhere" and "shows no tenant on a later visit without the
  parameter, legacy state or not". The other 3 are deliberate baselines pinning
  behaviour that must *not* change, and say so in their own comments.
- **Whole `packages/core/src/modules/auth` tree: 92 suites / 763 tests, all
  pass** (`npx jest -c packages/core/jest.config.cjs --rootDir packages/core
  src/modules/auth`). That includes the pre-existing
  `login-acl-challenge.test.tsx` and `login-tenant-banner-i18n.test.ts`.
- **eslint: 0 errors** on both files. Two warnings, both reproduced verbatim on
  the base commit with the change reverted: `react-hooks/exhaustive-deps` on the
  tenant-lookup effect (untouched by this PR) and `@next/next/no-img-element` on
  the test's `next/image` mock, which is copied from
  `login-acl-challenge.test.tsx`.
- **`tsc --noEmit` on `packages/core`: 265 errors, none in either touched file**
  — all are `Cannot find module '#generated/...'` in a clone that has never been
  built, and the count is identical on the base commit.

Not run: the repository's own CI, which a first-time fork contributor cannot
trigger.

## Checklist

- [ ] I updated documentation, locales, or generators if the change requires it. — *not required: no copy changes; `auth.login.tenantBanner` and the invalid-tenant strings are untouched in all locales.*
- [ ] I added or updated integration tests in `.ai/qa/tests/` — *the behaviour is a single client component's reaction to a query parameter and is fully covered by the unit suite; happy to add one if you disagree.*
- [ ] I created or updated the spec in `.ai/specs/` with a changelog entry (if applicable). — *N/A, see above.*
- [ ] Priority set — *cannot apply labels without `triage`; suggest `priority-low`.*
- [ ] Risk set — *suggest `risk-low`: one component, no API, no schema, no `BACKWARD_COMPATIBILITY.md` contract.*
- [ ] QA routing set — *suggest `needs-qa`: this does change a customer-facing UI surface. Manual check is two visits: open `/login?tenant=<id>` (banner shows, form submits the tenant), then open `/login` (no banner, and no banner even if `om_login_tenant` is already in localStorage or the cookie jar).*

### Design System Compliance
- [x] No hardcoded status colors — *no markup changed.*
- [x] No arbitrary text sizes — *no markup changed.*
- [x] Empty state handled — *N/A, not a list page.*
- [x] Loading state handled — *unchanged; `tenantLoading` still gates the lookup.*
- [x] `aria-label` on all icon-only buttons — *no buttons added or changed.*
- [x] Uses existing DS components — *no components added.*

## Linked issues

None known. Found while running a self-serve signup end to end on 0.6.5; the
same code is present on `develop`.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5714"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790444721&installation_model_id=432243&pr_number=5714&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5714&signature=c0e75688b5f76ea99867f35b9cbb5beb908dc846393901116f33672b0989b7b5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/modules/auth/__tests__/login-tenant-single-visit.test.tsx
+++ b/packages/core/src/modules/auth/__tests__/login-tenant-single-visit.test.tsx
@@ -1,0 +1,181 @@
+/**
+ * @jest-environment jsdom
+ *
+ * The `tenant` query parameter on /login is a hint scoped to the CURRENT visit.
+ *
+ * It used to be mirrored into localStorage['om_login_tenant'] and into a 14-day
+ * cookie of the same name, then replayed on every later visit — no TTL of its
+ * own, no clear after a successful sign-in. @open-mercato/onboarding puts the
+ * parameter on the login link in the workspace-ready e-mail
+ * (lib/ready-email.ts), in the post-verification redirect
+ * (lib/verify-redirects.ts) and in the onboarding status endpoint's login URL,
+ * so every self-serve signup passed through it exactly once and then carried
+ * the "you are signing in to <tenant>" banner on /login permanently.
+ *
+ * What must NOT change: the hidden `tenantId` input still carries the parameter
+ * into POST /api/auth/login, which is the disambiguation that path needs when
+ * one e-mail address exists in two tenants. And a visit that DOES carry the
+ * parameter still resolves the tenant and still reports an unknown one. Only
+ * the persistence is gone.
+ */
+import * as React from 'react'
+import { act, render } from '@testing-library/react'
+import LoginPage from '../frontend/login'
+
+const mockTranslate = (key: string, fallback?: string, params?: Record<string, string | number>) => {
+  if (!fallback) return key
+  if (!params) return fallback
+  return Object.entries(params).reduce(
+    (acc, [name, value]) => acc.replace(`{${name}}`, String(value)),
+    fallback,
+  )
+}
+
+const mockReplace = jest.fn()
+const mockApiCall = jest.fn()
+let currentSearchParams = new URLSearchParams()
+
+jest.mock('next/navigation', () => ({
+  useRouter: () => ({ replace: mockReplace, push: jest.fn() }),
+  useSearchParams: () => currentSearchParams,
+}))
+
+jest.mock('next/image', () => ({
+  __esModule: true,
+  default: (props: Record<string, unknown>) => <img alt={String(props.alt ?? '')} />,
+}))
+
+jest.mock('next/link', () => ({
+  __esModule: true,
+  default: ({ href, children, ...rest }: any) => <a href={typeof href === 'string' ? href : '#'} {...rest}>{children}</a>,
+}))
+
+jest.mock('@open-mercato/shared/lib/i18n/context', () => ({
+  useT: () => mockTranslate,
+}))
+
+jest.mock('@open-mercato/shared/lib/i18n/translate', () => ({
+  translateWithFallback: (_t: unknown, key: string, fallback: string, params?: Record<string, string | number>) =>
+    mockTranslate(key, fallback, params),
+}))
+
+jest.mock('@open-mercato/ui/backend/utils/apiCall', () => ({
+  apiCall: (...args: unknown[]) => mockApiCall(...args),
+}))
+
+jest.mock('@open-mercato/ui/backend/operations/store', () => ({
+  clearAllOperations: jest.fn(),
+}))
+
+jest.mock('@open-mercato/ui/backend/AuthSessionGuard', () => ({
+  notifyAuthIdentityChange: jest.fn(),
+}))
+
+jest.mock('@open-mercato/ui/backend/injection/InjectionSpot', () => ({
+  InjectionSpot: () => null,
+}))
+
+jest.mock('@open-mercato/ui/backend/injection/useRegisteredComponent', () => ({
+  useRegisteredComponent: (_handle: string, Fallback: any) => Fallback,
+}))
+
+const TENANT_ID = '2b8a4f16-0d5e-4a2a-9f7c-1c0f0f2a7e11'
+const TENANT_NAME = 'Acme Workspace'
+const LEGACY_KEY = 'om_login_tenant'
+
+function respondWithTenant(found: boolean) {
+  mockApiCall.mockImplementation(async (url: string) => {
+    if (String(url).startsWith('/api/directory/tenants/lookup')) {
+      return found
+        ? { result: { ok: true, tenant: { id: TENANT_ID, name: TENANT_NAME } } }
+        : { result: { ok: false, error: 'not_found' } }
+    }
+    // The authenticated-session probe: an anonymous visitor, so no auto-redirect.
+    return { result: {} }
+  })
+}
+
+async function renderLogin(search: string) {
+  currentSearchParams = new URLSearchParams(search)
+  let utils: ReturnType<typeof render> | undefined
+  await act(async () => {
+    utils = render(<LoginPage />)
+  })
+  // Flush the tenant-lookup promise chain.
+  await act(async () => { await Promise.resolve() })
+  return utils!
+}
+
+beforeAll(() => {
+  ;(globalThis as any).fetch = jest.fn(async () => ({ ok: true, json: async () => ({}), text: async () => '' }))
+})
+
+beforeEach(() => {
+  jest.clearAllMocks()
+  window.localStorage.clear()
+  document.cookie = `${LEGACY_KEY}=; path=/; max-age=0`
+  respondWithTenant(true)
+})
+
+describe('LoginPage — ?tenant= is a single-visit hint', () => {
+  it('resolves the tenant for the current visit and submits it', async () => {
+    const { container, unmount } = await renderLogin(`tenant=${TENANT_ID}`)
+
+    expect(container.textContent).toContain(TENANT_NAME)
+    const hidden = container.querySelector('input[name="tenantId"]') as HTMLInputElement | null
+    expect(hidden?.value).toBe(TENANT_ID)
+    unmount()
+  })
+
+  it('persists the tenant nowhere — not localStorage, not a cookie', async () => {
+    const { unmount } = await renderLogin(`tenant=${TENANT_ID}`)
+
+    expect(window.localStorage.getItem(LEGACY_KEY)).toBeNull()
+    expect(window.localStorage.length).toBe(0)
+    expect(document.cookie).not.toContain(LEGACY_KEY)
+    unmount()
+  })
+
+  it('shows no tenant on a later visit without the parameter, legacy state or not', async () => {
+    // Exactly the state a user carried before this change: the localStorage
+    // entry written by this component, and the cookie @open-mercato/onboarding
+    // also sets server-side on the verification redirect.
+    window.localStorage.setItem(LEGACY_KEY, TENANT_ID)
+    document.cookie = `${LEGACY_KEY}=${TENANT_ID}; path=/`
+    expect(document.cookie).toContain(LEGACY_KEY)
+
+    const { container, unmount } = await renderLogin('')
+
+    expect(container.textContent).not.toContain(TENANT_NAME)
+    expect(container.querySelector('input[name="tenantId"]')).toBeNull()
+    expect(
+      mockApiCall.mock.calls.some(([url]) => String(url).startsWith('/api/directory/tenants/lookup')),
+    ).toBe(false)
+    unmount()
+  })
+
+  it('still reports an unknown tenant when the parameter is present', async () => {
+    respondWithTenant(false)
+
+    const { container, unmount } = await renderLogin(`tenant=${TENANT_ID}`)
+
+    expect(container.textContent).toContain('Tenant not found')
+    expect(container.textContent).not.toContain(TENANT_NAME)
+    expect((container.querySelector('input[name="tenantId"]') as HTMLInputElement | null)?.value)
+      .toBe(TENANT_ID)
+    unmount()
+  })
+
+  it('clearing the tenant drops the parameter from the URL', async () => {
+    const { container, unmount } = await renderLogin(`tenant=${TENANT_ID}&redirect=%2Fbackend`)
+
+    const clear = Array.from(container.querySelectorAll('button')).find(
+      (b) => (b.textContent || '').trim().length > 0 && b.getAttribute('type') === 'button',
+    ) as HTMLButtonElement | undefined
+    expect(clear).toBeTruthy()
+    await act(async () => { clear!.click() })
+
+    expect(mockReplace).toHaveBeenCalledWith('/login?redirect=%2Fbackend')
+    unmount()
+  })
+})
--- a/packages/core/src/modules/auth/__tests__/login-tenant-single-visit.test.tsx
+++ b/packages/core/src/modules/auth/__tests__/login-tenant-single-visit.test.tsx
@@ -12,14 +12,21 @@
  * so every self-serve signup passed through it exactly once and then carried
  * the "you are signing in to <tenant>" banner on /login permanently.
  *
- * What must NOT change: the hidden `tenantId` input still carries the parameter
- * into POST /api/auth/login, which is the disambiguation that path needs when
- * one e-mail address exists in two tenants. And a visit that DOES carry the
- * parameter still resolves the tenant and still reports an unknown one. Only
- * the persistence is gone.
+ * What must NOT change: POST /api/auth/login still receives a `tenantId`, which is the
+ * disambiguation that path needs when one e-mail address exists in two tenants -
+ * `findUsersByEmail` deliberately treats an ambiguous match as no user and falls through
+ * to the uniform 401 (issue #2242), and the form has no tenant selector, so with no hint
+ * at all such a user has no in-app way back in. The app lands them on a BARE /login
+ * routinely: session refresh, logout, and the 401 handler in @open-mercato/ui.
+ *
+ * So the BANNER is visit-scoped (the defect) while the SUBMITTED value falls back to a
+ * per-tab, per-address hint (the disambiguation). sessionStorage removes the 14-day weld
+ * and the cross-session surprise; keying on the address removes the other half of the old
+ * behaviour's cost - a stale hint narrowing the lookup for a different person on the same
+ * tab, who would get a 401 with nothing on screen to explain it.
  */
 import * as React from 'react'
-import { act, render } from '@testing-library/react'
+import { act, fireEvent, render, screen } from '@testing-library/react'
 import LoginPage from '../frontend/login'
 
 const mockTranslate = (key: string, fallback?: string, params?: Record<string, string | number>) => {
@@ -82,6 +89,13 @@ jest.mock('@open-mercato/ui/backend/injection/useRegisteredComponent', () => ({
 const TENANT_ID = '2b8a4f16-0d5e-4a2a-9f7c-1c0f0f2a7e11'
 const TENANT_NAME = 'Acme Workspace'
 const LEGACY_KEY = 'om_login_tenant'
+const HINT_KEY = 'om_login_tenant_hint'
+const EMAIL = 'ada@example.com'
+
+async function typeEmail(container: HTMLElement, value: string) {
+  const input = container.querySelector('input[name="email"]') as HTMLInputElement
+  await act(async () => { fireEvent.change(input, { target: { value } }) })
+}
 
 function respondWithTenant(found: boolean) {
   mockApiCall.mockImplementation(async (url: string) => {
@@ -113,6 +127,7 @@ beforeAll(() => {
 beforeEach(() => {
   jest.clearAllMocks()
   window.localStorage.clear()
+  window.sessionStorage.clear()
   document.cookie = `${LEGACY_KEY}=; path=/; max-age=0`
   respondWithTenant(true)
 })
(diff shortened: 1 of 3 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T22:31:11Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review

## 🎯 Summary

This PR makes the `?tenant=` query parameter on `/login` a hint scoped to the current visit: the `localStorage['om_login_tenant']` mirror and the 14-day client-side cookie of the same name are removed, along with the three now-unreferenced helpers (`readTenantCookie`, `setTenantCookie`, `clearTenantCookie`) and their two constants. A new 5-case jsdom suite covers the change. The bug it reports is real and well-diagnosed: `@open-mercato/onboarding` puts `?tenant=` on the login link in the workspace-ready e-mail (`packages/onboarding/src/modules/onboarding/lib/ready-email.ts:29`), in the post-verification redirect (`packages/onboarding/src/modules/onboarding/lib/verify-redirects.ts:34`) and in the status endpoint's `loginUrl` (`packages/onboarding/src/modules/onboarding/api/get/onboarding/status.ts:110`), so every self-serve signup did pass through the parameter once and then carried the green banner indefinitely, with no TTL of its own and no clear after sign-in.

The diagnosis is right and the analysis in the PR body is unusually careful — including the correct conclusion that the onboarding module's own server-set `om_login_tenant` cookie is fed by `redirectToPreparing`/`redirectToLogin`, not by this component, so the status endpoint's 403 gate keeps working untouched. I verified that: `PreparingPageClient.tsx:69` is the endpoint's only caller and `/onboarding/preparing` is only ever reached through `redirectToPreparing`, which sets the cookie itself.

What I cannot sign off on is the chosen remedy. The persisted value was not only feeding the banner — it was also the mechanism that let a user whose e-mail exists in more than one tenant log in from a bare `/login`. Deleting the persistence altogether removes that path with it, and this repository supports that topology by design. Details in the first finding below, with a suggested narrower fix that resolves the reported bug without the behavior loss.

## Verdict

**Changes requested** — one major finding (a login lockout for same-e-mail-across-tenants users after logout or session expiry) and one minor finding (the "Clear" button silently stops clearing the onboarding cookie it used to clear). Two nits follow for whoever picks the branch up.

## 🧪 Validation Gate

CI has **never run on this head**. Both workflow runs for `6e93839` sit at `action_required` — the standard fork gate, which needs a maintainer to approve the run after every push — so GitHub reports no usable check data for this commit and the checks-first path does not apply here. I fell back to a locally scoped gate over the changed package, in a clean isolated worktree with its own `node_modules`, Node 24.

**Validation source: local fallback** (narrow, scoped to the changed package — no GitHub check data available for the head SHA).

| Command | Scope | Result |
|---|---|---|
| `jest src/modules/auth/__tests__/login-tenant-single-visit.test.tsx` | new suite | ✅ 5/5 pass |
| Same suite against the **unmodified** `login.tsx` | mutation check | ✅ 2 of 5 fail, exactly as the PR body claims — the suite is not vacuous |
| `jest src/modules/auth` | whole auth tree | ✅ 92 suites / 763 tests pass |
| `jest src/backend/__tests__/DataTable.perspectiveStorage.test.ts` | the one other file referencing this key | ✅ 10/10 pass |
| `tsc --noEmit -p packages/core/tsconfig.json` | changed package | ✅ 0 errors in either touched file (265 total, **all** `Cannot find module '#generated/…'`, i.e. the un-generated-clone baseline — 0 non-`#generated` errors) |
| `eslint` on both touched files | changed files | ✅ 0 errors, 2 warnings (`react-hooks/exhaustive-deps` on the untouched tenant-lookup effect, `@next/next/no-img-element` on the test's `next/image` mock) — both reproduce on the base commit |

The author's own reported numbers reproduced exactly. Not run: `yarn generate`, `build:packages`, `build:app`, the i18n gates and the Playwright suite — no locale or generator-discovered file is touched, and the full gate is CI's job once a maintainer approves the run.

## 🔴 Blockers

None.

## 🟠 Major

### 1. Dropping the persisted hint locks out users whose e-mail exists in more than one tenant

`packages/core/src/modules/auth/frontend/login.tsx:172` — after this change, `tenantId` comes from the URL and nowhere else.

The `POST /api/auth/login` route cannot disambiguate without it. `packages/core/src/modules/auth/api/login.ts:132-140`:

```ts
const users = await auth.findUsersByEmail(parsed.data.email)
// Never disclose that an email is registered across multiple tenants …
user = users.length === 1 ? users[0] : null
```

An ambiguous match deliberately falls through to the uniform 401 — correct as an anti-oracle measure, and untouched by this PR. But it means the hidden `tenantId` input is the *only* thing that lets such a user in.

This is a supported topology, not a corner case. `packages/core/src/modules/auth/migrations/Migration20260610120000.ts:1-6` made e-mail uniqueness per-tenant precisely so the same address can live in several tenants, and says so: the old global constraint *"contradicts the multi-tenant login flow — which resolves the same email across tenants via `findUsersByEmail`"*.

And the app routinely sends users to a bare `/login`, with no `tenant=`:

- `packages/core/src/modules/auth/api/session/refresh.ts:86,92,100` → `/login?redirect=…` on an expired session
- `packages/core/src/modules/auth/api/logout.ts:77` → `/login`
- the 401 handler in `packages/ui/src/backend/utils/api.ts` routes through that same refresh endpoint

Before this PR, the persisted `om_login_tenant` re-populated the hidden field on those visits and the user signed in. After it, they get the uniform 401, and there is no tenant selector anywhere on the form (`login.tsx:340-395` renders only the banner) — so there is no in-app way back in. Recovery means digging the original onboarding e-mail out of an inbox.

I confirmed the mechanism empirically rather than by reading: reverting only `login.tsx` to `origin/develop` and re-running this PR's own new suite, the case at `login-tenant-single-visit.test.tsx:139` fails with the banner rendered **and the hidden `tenantId` input present** on a bare `/login`. That hidden input is the thing being removed.

The PR frames the persistence as purely a banner artifact. It was also load-bearing, and the PR body does not mention the trade-off — so I do not think it has been accepted deliberately.

**Suggested fix — the reported bug does not require removing the value, only its lifetime.** Either of these fixes the welded banner without the lockout:

- Move the mirror to `window.sessionStorage`. It dies with the tab, so no 14-day weld and no cross-session surprise, while a session-expiry bounce inside the same tab still recovers the tenant.
- Or keep `localStorage` and clear it in `onSubmit` on a successful sign-in (`login.tsx:301-310`, next to the existing `clearAllOperations()` / `clearAllPerspectiveState()` calls). "No clear after a successful sign-in" is named in the PR body as half the defect; fixing that directly removes the permanence.

Either way, please drop the client-set **cookie** as this PR does — a 14-day non-`httpOnly` cookie is the wrong tool here and nothing else reads the one this component wrote.

If wholesale removal really is the intent, that is a deliberate behavior change for multi-tenant installs: it needs saying in the PR body and a maintainer's explicit call, since it makes the activation link the only login path for those users forever.

## 🟡 Minor

### 2. "Clear" no longer clears the onboarding cookie it used to clear

`packages/core/src/modules/auth/frontend/login.tsx:218` — `handleClearTenant` previously called `clearTenantCookie()`, expiring `om_login_tenant` on `path=/`. That is the same cookie `@open-mercato/onboarding` sets server-side, non-`httpOnly`, `path=/`, 14-day max-age (`packages/onboarding/src/modules/onboarding/lib/verify-redirects.ts:11-19`), and reads back as the sole authorization input to the status endpoint (`packages/onboarding/src/modules/onboarding/api/get/onboarding/status.ts:61-64`, 403 when absent or naming another tenant).

I agree with the PR body that this cookie's *setting* side belongs to onboarding and is out of scope here. But the *clearing* side was existing user-facing behavior behind a button labelled "Clear", and it is being removed silently: after this PR the cookie survives the full 14 days with no in-app way to drop it.

Keeping the one-line cookie expiry inside `handleClearTenant` costs nothing and preserves the button's meaning. If dropping it is intentional, please say so in the PR body alongside the existing note about the cookie — right now that note reads as covering only the set path.

## 🔵 Nits

### 3. Fragile Clear-button locator in the new suite

`packages/core/src/modules/auth/__tests__/login-tenant-single-visit.test.tsx:172-174` finds the button as "the first `type="button"` with non-empty text". It happens to work today because the banner renders above the form, but `login.tsx:365` already conditionally renders a "Go to Dashboard" button in the same region, and any future button added above the banner would silently retarget the assertion at a different control while the test stays green. `getByRole('button', { name: 'Clear' })` states the intent.

### 4. A now-dead key used as the "unrelated key" fixture

`packages/ui/src/backend/__tests__/DataTable.perspectiveStorage.test.ts:92,98` uses `om_login_tenant` as its example of a localStorage key that must survive `clearAllPerspectiveState()`. After this PR nothing writes that key, so the assertion pins a contract that no longer exists. The test still passes (it seeds the key itself) — swapping in a key the app actually writes keeps it meaningful. Worth noting that this fixture is also independent evidence the persistence was once deliberate: `clearAllPerspectiveState()` runs on successful login and was written to spare this key.

## 💥 Breaking Changes

| Surface | Status |
|---|---|
| Published event names | Untouched |
| API response schemas / routes | Untouched — `POST /api/auth/login` still receives `tenantId` from the hidden input; `/api/directory/tenants/lookup` unchanged |
| Public exports & import paths | The three removed helpers were module-private with no other call site in the repo (verified by grep across all packages) — no bridge needed |
| DB schema / migrations | Untouched |
| DI keys, ACL features, notification IDs, widget spot IDs, CLI commands | Untouched |
| Generated files | Untouched |

No `BACKWARD_COMPATIBILITY.md` contract surface is affected. The behavior change in finding 1 is a runtime/UX regression rather than a contract break, which is why it is filed as major rather than a blocker.

## 🧪 Test Coverage

The new suite is genuinely load-bearing, which is worth calling out — I verified the author's non-vacuity claim rather than taking it: against the unmodified `login.tsx`, exactly 2 of the 5 cases fail ("persists the tenant nowhere" and "shows no tenant on a later visit"), and the other 3 are honest baselines that the author labels as such in their own comments. The jsdom pragma and module mocks follow the neighbouring `login-acl-challenge.test.tsx` conventions.

The gap is on the other side of the change: nothing asserts what happens to a user who arrives at `/login` *without* the parameter and needs a tenant to get in. That is the case finding 1 is about, and whichever way it is resolved it should ship with a test — either a unit case pinning that the session-scoped hint survives a same-tab bounce, or, if removal stands, an explicit case documenting that a bare `/login` submits no `tenantId` by design, so the next reader knows the multi-tenant consequence was weighed rather than overlooked.

An integration test is not required here; I agree with the author's reasoning that a client component's reaction to a query parameter is unit-testable. But note that `packages/onboarding/src/modules/onboarding/__integration__/TC-ONB-002-multi-tenant-parallel-login.spec.ts` — the closest existing integration coverage — uses a **distinct e-mail per tenant** and always arrives via `/login?tenant=…`, so it would not catch finding 1 either.

---

Thanks for a thorough and well-evidenced PR — the root cause, the dead-helper cleanup and the deliberate scoping away from the onboarding package are all right, and the non-vacuity check in the body is exactly the kind of thing that makes a review fast. The one thing I need before approving is finding 1: either a narrower fix that keeps multi-tenant logins working, or an explicit maintainer decision that the activation link becomes the only path for those users.

pkarw:

- ❌ `changes-requested` — the review found one major finding (removing the persisted tenant hint locks out users whose e-mail exists in more than one tenant when they arrive at a bare `/login` after logout or session expiry) and one minor finding (the "Clear" button silently stops expiring the onboarding `om_login_tenant` cookie it used to expire).
- 🐛 `bug` — the change fixes a defect: the `?tenant=` banner welded itself to `/login` permanently after any visit that carried the parameter.
- 🧪 `needs-qa` — `packages/core/src/modules/auth/frontend/login.tsx` is a UI-rendering file on the login page, so the automated-verification exemption does not apply and a human should exercise both visits (`/login?tenant=<id>` then a bare `/login`, with legacy localStorage and cookie state present).
- 🔹 `priority-medium` — the underlying defect is an ordinary bug whose visible effect is a stale banner rather than a security incident or a release blocker, so it does not meet the bar for `priority-high`.
- ⚠️ `risk-high` — the blast radius is the login page of every install running the onboarding module, and the review found that the change alters who can authenticate at all in the multi-tenant same-e-mail case; auth and session surfaces rate `risk-high` regardless of diff size.

pkarw:

pkarw:

One major finding (multi-tenant login lockout from removing the persisted tenant hint) and one minor finding (the "Clear" button no longer expires the onboarding `om_login_tenant` cookie), plus two nits. Validation source: local fallback — both workflow runs for `6e93839` sit at `action_required` (the fork CI-approval gate), so GitHub reported no usable check data for this head; the new suite, the full `packages/core` auth tree (92 suites / 763 tests), `tsc` and `eslint` on the changed files were run locally instead and all pass. autofix: skipped (not my PR — re-run with --autofix to fix it here).

Geeknauts-company:
Pushed `fd5a57dd`. The major was right, and you were also right that it was not a deliberate trade-off — I had framed the persistence as a banner artifact and never checked what else depended on it. @pkarw, flagging you directly: this account only has `read`, so I cannot re-request review myself.

**The two things were conflated, and only one of them is the defect.** The banner and the submitted `tenantId` both read one piece of state, so making the banner visit-scoped took the disambiguation with it. They are now separate:

- **Banner + tenant lookup** — the URL parameter and nothing else. Unchanged from round 1; that is the welded-banner fix.
- **Submitted `tenantId`** — falls back to a per-tab, per-address hint when the URL carries none.

**The hint is `sessionStorage`, as you suggested, and additionally keyed to the e-mail.** The second half is not decoration, and it is the reason I did not take your `localStorage` + clear-on-success option: an *unkeyed* hint narrows the lookup for whoever signs in next on that tab, so a different person gets the uniform 401 with nothing on screen to explain it — the same class of invisible sticky state this PR exists to remove, at tab scope instead of 14-day scope. Keyed, a second address simply gets no hint and resolves normally.

(The clear-on-success option I ruled out on its own terms: after a successful sign-in nothing re-persists, so the very next session-expiry bounce is the lockout again. It fixes the weld without fixing what you found.)

It is written **on submit**, not when the parameter arrives, because `/login` takes no `email` parameter — the submit is the first moment both halves are known. Written before the request, so a mistyped password does not cost the hint.

The client-set 14-day cookie stays gone.

**Residual gap, stated rather than implied:** a multi-tenant user returning in a *new browser session* to a bare `/login` still has no hint and still needs the activation link. That is pre-existing — the real answer is a tenant selector on the form — and this PR neither creates nor closes it. What it does close is the regression your finding names: the in-app bounce (refresh, logout, the `ui` 401 handler), which is the common case and stays inside one tab.

**Minor 2 — "Clear" clears again.** It now drops the URL parameter, the state, the per-tab hint **and** expires `om_login_tenant`. Your framing settled it for me: the *setting* side belongs to onboarding, but the *clearing* side was existing user-facing behaviour behind a button labelled "Clear", and this button is the only in-app way to drop a cookie that otherwise lives 14 days and gates the onboarding status endpoint.

**Nit 3** — `getByRole('button', { name: 'Clear' })`. You were right about the mechanism: `login.tsx` already renders a conditional "Go to Dashboard" button in the same region.

**Nit 4** — the fixture is now `om:sidebarCollapsed`, which `AppShell` really writes. I've kept your observation in the comment: that this key was chosen deliberately, and that `clearAllPerspectiveState()` was written to spare it, is itself evidence the persistence was once load-bearing.

**Everything is pinned, and each part fails alone when reverted:**

| Reverted | Fails |
|---|---|
| the hidden input's fallback to the hint | the bare-`/login` submit case |
| the address key | the different-address-same-tab case |
| the cookie expiry in `handleClearTenant` | the clear case |

**Local gate**: `packages/core` `auth` 92/92 suites, 768 tests (this PR's suite 10, up from 5); `packages/ui` `DataTable.perspectiveStorage` 10/10; typecheck clean; eslint 0 errors — one `react-hooks/exhaustive-deps` warning on `login.tsx` that is **pre-existing** (same warning at head `6e938393`, different line number). Head merges cleanly into `develop`.

I'll leave the PR description's "Only the persistence is gone" line to you or to a follow-up push — it now understates what the change does, and the commit message carries the accurate version.

PR #5715, opened 2026-08-27
refactor(shared): hoist the scope cookie constants and collapse their parsers

### LINKED ISSUE #5713: Hoist the organization-selection cookie constants into @open-mercato/shared and collapse the three parsers
### Context

Raised in review of #5690 and deferred there by the maintainer, who asked for a separate PR plus this issue, since it touches a surface listed in `BACKWARD_COMPATIBILITY.md` and deserves its own diff rather than riding along with a behaviour fix.

### The duplication

`@open-mercato/shared` must not depend on `@open-mercato/core`, so the organization-selection cookie names and the all-organizations sentinel are inlined in `shared` rather than imported from `core`'s `modules/directory/constants`. There are now three copies:

- `packages/shared/src/lib/auth/server.ts` — all three values
- `packages/shared/src/lib/crud/factory.ts` — `SELECTED_ORG_COOKIE`
- `packages/shared/src/lib/frontend/organizationEvents.ts` — all three, added by #5690

### The parsers diverge, which is the part that could bite

Reading them is duplicated too, and not identically. `parseSelectedOrganizationCookie` (`packages/core/src/modules/directory/utils/scopeCookies.ts:11`) maps a blank cookie value to `null`, meaning "no selection". `readScopeCookie` in `organizationEvents.ts` returns `''`, and the seed then maps it to `organizationId: null`, meaning "all organizations". Those two readings diverge in the resolver: "no selection" falls back to the account organization, "all organizations" does not.

No wrong result is reachable today — the switcher writes the `__all__` sentinel rather than a blank organization cookie, and `resolveOrganizationOverride` treats blank as all-organizations too, so the seed and the server agree in practice. Filing it because drift between three near-identical copies is the kind of thing that stops being harmless quietly.

### Proposed resolution

Hoist the three constants into `shared` (e.g. `lib/scope/cookies.ts`), have `core/modules/directory/constants` re-export them for backward compatibility, and collapse the parsers onto one implementation with a single documented blank-value semantic.

The re-export matters: import paths and public exports are a contract surface in `BACKWARD_COMPATIBILITY.md`, so nothing existing should have to change its imports.

### Not urgent

Nothing is broken. This is a tidy-up whose value is preventing a future divergence, so it can wait for a convenient window.

Geeknauts-company:
Follow-up to #5690, raised in its review and filed as #5713. Separate PR at the maintainer's request, because it touches a surface listed in `BACKWARD_COMPATIBILITY.md` §4 (Import Paths).

**Nothing observable changes.** No import path, export name, export value, function signature or runtime behaviour moves for any consumer, inside the monorepo or out. The point is that four independent copies of a wire format become one, and two independently written parsers become one implementation with a written-down blank-value semantic.

## The duplication

`@open-mercato/shared` must not depend on `@open-mercato/core`, where the cookie names and the all-organizations sentinel lived, so `shared` inlined them:

- `packages/shared/src/lib/auth/server.ts` — all three values, plus its own `readCookieFromHeader` / `decodeCookieValue`
- `packages/shared/src/lib/crud/factory.ts` — `SELECTED_ORG_COOKIE`
- `packages/core/src/modules/directory/utils/organizationScope.ts` — a **fourth** copy of both names, inlined rather than importing its own sibling `constants` (this one was not in the issue; found while collapsing the others)

#5690 was about to add a fifth in `lib/frontend/organizationEvents.ts`.

## Changes

- **New**: `packages/shared/src/lib/scope/cookies.ts` — the names, the sentinel, `isAllOrganizationsSelection`, and one parsing implementation. Pure string handling with no imports, so it is safe in a browser bundle, a server component and a worker alike. `shared`'s wildcard `exports` entries already resolve `./lib/scope/cookies`; no `package.json` change needed.
- **Bridges**: `core`'s `modules/directory/constants` and `modules/directory/utils/scopeCookies` re-export from it. Same names, same values, same bindings — additionally exporting `SELECTED_ORGANIZATION_COOKIE` / `SELECTED_TENANT_COOKIE`, which had no canonical home before. Nothing removed, so no `@deprecated` and no `UPGRADE_NOTES.md` entry: these paths remain the documented way to reach these values, and the bridge is permanent rather than a staging step.
- **Copies removed** from the three files above.
- **Spec**: `.ai/specs/2026-08-27-scope-cookie-constants-hoist.md`, with the "Migration & Backward Compatibility" section `BACKWARD_COMPATIBILITY.md` §5 asks for.

## The parser divergence, and the semantic chosen

`parseSelectedOrganizationCookie` maps a blank cookie value to `null` — "no selection" — while #5690's `readScopeCookie` returns `''`, which its seed maps to "all organizations". The resolver treats those differently: "no selection" falls back to the caller's own organization, "all organizations" does not. No wrong result is reachable today, since the switcher writes the `__all__` sentinel rather than a blank organization cookie.

**Chosen: for a *selection*, blank reads exactly like absent — no selection.** Documented at the top of the new module. The reasoning:

- The all-organizations choice has its own sentinel, and nothing writes a blank value to express it. A blank cookie is the residue of a selection being cleared, not a statement about scope.
- Reading it as "all organizations" lets a cleared cookie silently widen scope. Reading it as "no selection" falls back to the caller's own organization — the conservative direction, and what the server-side resolver has always done.
- It is what both merged functions already did, so adopting it changes nothing.

### One caller must still tell blank from absent

`applySuperAdminScope` reports a blank organization cookie as an **applied** override meaning "all organizations", and a **missing** cookie as no override at all. Those produce different auth contexts, so the distinction is kept available through the lossless `readScopeCookieRaw` (`undefined` only for absent, `''` for present-and-blank) rather than hidden. That is the raw read the selection semantic is derived from, not a second semantic competing with it, and both the module header and the call site say so.

**Nothing tested that difference before.** Mutating the reader to return `undefined` for a blank cookie left the whole existing `lib/auth/__tests__/server.test.ts` green. Two cases were added there.

## Testing

Run from a fresh clone with its own `node_modules` (Node 24, `engines.node === "24.x"`). Baselines measured on pristine `develop` in the same tree, not read off the "after" run.

- `packages/shared` full suite: **2012 passed / 2 failed / 5 skipped**, 175 suites passed, 11 unrunnable — against **1993 / 2 / 5**, 174 passed, 11 unrunnable on `develop`. Exactly +1 suite and +19 passing tests, no change to the failures. The 2 are `principalServiceExport` (imports `dist/`) and the 11 cannot resolve `@open-mercato/cache`; both are artefacts of a clone that has never been built.
- `packages/core` full suite: **11620 passed / 2 failed**, 1438 suites passed, 16 failed. Those 16 suites and 2 tests reproduce **identically** on pristine `develop` (catalog, customer_accounts, entities, messages, sales, translations, warranty_claims — none of them touched here); measured by reverting to `develop` and re-running exactly those paths.
- `packages/search`: 336/336 tests pass; 1 suite unrunnable (`VectorSearchSection`, `@open-mercato/ui/…/useAppEvent`), same unbuilt-clone cause.
- Not vacuous: undoing the bridge fails 1 core case; flipping the blank semantic fails 3 shared and 2 core; collapsing absent/blank in the raw reader fails 2 shared and 1 of the new `server.test.ts` cases.
- `packages/shared` `tsc --noEmit`: clean, exit 0. `packages/core`: 265 errors, all `Cannot find module '#generated/…'` in an unbuilt clone, none in a touched file.
- eslint: 0 errors across all nine touched files.

## Deliberately out of scope

- **`lib/frontend/organizationEvents.ts`** is not on `develop` — its copy arrives with #5690, so this PR cannot collapse it and the two do not touch a common file in either merge order. It is a two-line follow-up on that branch once this lands, and it must use `readScopeCookieRaw` / `decodeScopeCookieValue` rather than the selection parsers: its seed bails out when a cookie is absent but seeds `organizationId: null` when one is blank, making it the second legitimate caller of the lossless pair. The `MAINTAINER NOTE` that pointed at this work lives on that branch and goes with it.
- **`packages/create-app/template/.../OrganizationSwitcher.tsx`** keeps its own readers. They return a richer `{ value, hasCookie, raw }` shape with its own `__all__` folding, so collapsing them is a behaviour-preserving rewrite of starter-template code with its own tests — a different review from a contract-surface bridge. It already imports the sentinel from `core`, so it carries no constant duplication.
- **`shared`'s remaining dependency on `core`.** `lib/crud/factory.ts` imports `resolveOrganizationScopeForRequest` from `core`, so the "shared must not depend on core" rule is already bent for behaviour. This change does not resolve that; it only stops the leaf *values* from being copied because of it.

Closes #5713.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5715"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790449032&installation_model_id=432243&pr_number=5715&repository=open-mercato%2Fopen-mercato&return_to=https%3A%2F%2Fgithub.com%2Fopen-mercato%2Fopen-mercato%2Fpull%2F5715&signature=0040c189c8023b283e750921c0a58e430fd160df02949d7931d8461b1500ba46"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-08-27T22:24:13Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# 🔍 Code Review: refactor(shared): hoist the scope cookie constants and collapse their parsers

## 🎯 Summary

This PR hoists the organization-selection cookie names (`om_selected_org`, `om_selected_tenant`) and the all-organizations sentinel (`__all__`) into a new leaf module, `packages/shared/src/lib/scope/cookies.ts`, and collapses two independently written parsers into one implementation with a written-down blank-value semantic. `@open-mercato/core`'s `modules/directory/constants` and `modules/directory/utils/scopeCookies` become pure re-export bridges, and the inlined copies are removed from `packages/shared/src/lib/auth/server.ts`, `packages/shared/src/lib/crud/factory.ts` and `packages/core/src/modules/directory/utils/organizationScope.ts`. Reviewed scope: all ten changed files, plus every in-repo consumer of the two bridged import paths and of the two cookie-name literals.

The behaviour-preservation claim holds up under inspection, and I checked it function by function rather than taking it on trust. `readScopeCookieRaw` reproduces the old `readCookieFromHeader` exactly, including that `trimmed.slice(prefix.length)` equals the old `trimmed.slice(name.length + 1)` and that a cookie whose name merely *ends* with the requested one (`x_om_selected_org=`) still does not match. `decodeScopeCookieValue` reproduces the old `decodeCookieValue`, including the malformed-encoding fallback to the raw text; the old `?? null` tails were dead code on both paths, so dropping them changes nothing. `parseScopeSelectionCookie` reproduces the old core parsers term for term, including the first-match-wins loop and the `|| null` collapse of a blank value. The `resolveOrganizationOverride` blank-is-an-applied-override path is untouched and still reachable, because `readScopeCookieRaw` deliberately keeps `undefined` (absent) distinct from `''` (present and blank).

This is careful, well-motivated work. The module header and the spec both explain *why* the blank-value semantic was chosen rather than just asserting it, the asymmetry that `applySuperAdminScope` depends on is documented at the module, the spec and the call site instead of being quietly preserved, and the three test files are genuinely non-vacuous — I confirmed the author's mutation claims are the kind that would actually fail. I found no correctness defect, no security or data-scoping regression, and no contract break. The findings below are all Medium/Low: two concern the accompanying spec and the copies the hoist leaves behind, and two are documentation polish.

## Verdict

❌ **request changes** — there are no blockers and no majors; the code itself is sound and I would merge it as written. This repository's review rule requests changes on Medium findings too, and there are two: the hoist leaves an un-collapsed copy of the cookie name behind that the spec's own "Deliberately out of scope" list does not name (and that section additionally makes an inaccurate claim about the `OrganizationSwitcher` copies), and the spec omits the Risks & Impact Review, Final Compliance Report and Changelog sections that `.ai/specs/AGENTS.md` requires. Both are documentation-and-completeness items on an otherwise clean change, and both are cheap to resolve.

## 🧪 Validation Gate

CI could not supply this signal. Both workflow runs for head `22a97e44` (`CI for Develop&Main` and `Mutation tests`) sit at `action_required` because the PR comes from a fork and needs per-push maintainer approval, so there is no usable GitHub check data for this head. I therefore ran the configured gate locally, in an isolated worktree at the PR head with a fresh `yarn install`.

**Validation source:** local fallback (no usable GitHub check data — fork runs pending approval).

| Command | Status | Notes |
|---------|--------|-------|
| `yarn build:packages` | ✅ PASS | 27/27 tasks. `packages/shared/dist/lib/scope/cookies.js` is emitted, and `packages/core/dist/modules/directory/constants.js` correctly compiles to a bare re-export of the package specifier. |
| `yarn generate` | ✅ PASS | Left the working tree clean — no generated-file drift. |
| `yarn build:packages` (second run) | ✅ PASS | 27/27 tasks against the freshly generated files. |
| `yarn i18n:check-sync` | ✅ PASS | All five locales in sync across 57 modules. |
| `yarn i18n:check-usage` | ✅ PASS | Run as a hard gate here, not `continue-on-error` as in CI. |
| `yarn typecheck` | ✅ PASS | 27/27 tasks. |
| `yarn test` | ⚠️ PASS with an unrelated host failure | Every workspace is green except `create-mercato-app`, whose four failures are all in `packages/create-app/src/lib/template-dev-log-files.test.ts` and are host-environment assertions (`fs.inotify.max_user_instances: 128 < 4096`), not defects in this diff. That failure aborted the turbo run before `@open-mercato/core` finished, so I re-ran core's suite standalone: **11697 passed / 0 failed** (1454 suites). `@open-mercato/shared`: **2059 passed / 0 failed**. The new suites all ran and passed — `src/lib/scope/__tests__/cookies.test.ts`, `src/modules/directory/utils/__tests__/scopeCookies.test.ts`, and the extended `src/lib/auth/__tests__/server.test.ts`. |
| `yarn build:app` | ✅ PASS | Compiled successfully. |

Two checks beyond the configured gate, because CI runs them:

| Command | Status | Notes |
|---------|--------|-------|
| `yarn lint` | ✅ PASS | 0 errors, 10 pre-existing warnings, none in a touched file. |
| `yarn template:sync` | ⚠️ Pre-existing drift | Reports one mismatched file, `modules.ts`. That divergence is deliberate and lives on `develop`; neither `apps/mercato/src/modules.ts` nor its template mirror is touched by this PR. Not attributable to this change. |

I also confirmed this PR is not duplicated work: `packages/shared/src/lib/scope/cookies.ts` does not exist on `develop`, the head is exactly zero commits behind `develop` (`41ff100d9`), and the file lists of this PR and #5690 do not intersect at all — so the two really can merge in either order, as the description claims.

## Findings

### 🔹 Minor

**1. `.ai/specs/2026-08-27-scope-cookie-constants-hoist.md:131-138` — the hoist leaves a copy behind that the out-of-scope list does not name, and the stated rationale for the copies it does name is inaccurate.**

The change is framed as "four independent copies of a wire format become one", and the spec's *Deliberately out of scope* section accounts for what is left by naming exactly two things: `lib/frontend/organizationEvents.ts` (not on `develop`) and the `create-app` template's `OrganizationSwitcher`. After this PR, a repository-wide sweep for the two cookie-name literals outside tests and fixtures still turns up the following:

- `apps/mercato/src/app/(backend)/backend/[...slug]/page.tsx:85` reads `cookieStore.get('om_selected_org')?.value ?? null`, and its mirror at `packages/create-app/template/src/app/(backend)/backend/[...slug]/page.tsx:85` does the same. This is not mentioned anywhere in the PR description or the spec. It is also the one remaining site with no "richer shape" defence: it is a plain single-symbol substitution, exactly the class of copy this PR exists to eliminate.
- `apps/mercato/src/components/OrganizationSwitcher.tsx` inlines `'om_selected_org='` / `'om_selected_tenant='` at lines 77, 78, 107, 108, 155, 165 and 181, and `packages/create-app/template/src/components/OrganizationSwitcher.tsx` is byte-identical to it. The spec says the template copy "already imports the sentinel from `core`, so it carries no constant duplication" — the sentinel yes, but both cookie *names* are hard-coded seven times each, in the component that **writes** the wire format everything else reads. The out-of-scope list also names only the template file, not the `apps/mercato` original that actually ships in the app.

Why it matters: the value of this PR is that a future contributor asking "where do these cookie names live?" gets one answer. Leaving an unlisted literal behind, and asserting in the spec that a remaining copy carries no duplication when it carries seven, reproduces in the documentation precisely the drift the code change removes. The `[...slug]/page.tsx` site is also the likeliest candidate to be missed by a later cleanup, because nothing points at it.

Fix — either is fine, and the second is entirely reasonable given the cost:

- Collapse the cheap one: import `SELECTED_ORGANIZATION_COOKIE` in `apps/mercato/src/app/(backend)/backend/[...slug]/page.tsx` and run `yarn template:sync:fix` so the mirror follows (the root `AGENTS.md` requires that mirroring in the same task). Leave both `OrganizationSwitcher` copies alone, for the reason the spec already gives.
- Or keep the current scope and make the spec accurate: name `apps/mercato/src/app/(backend)/backend/[...slug]/page.tsx` and its template mirror as deferred, name both `OrganizationSwitcher` files rather than only the template one, and correct the "carries no constant duplication" sentence to say that they inline both cookie names and that collapsing them is deferred because of their richer `{ value, hasCookie, raw }` shape, not because there is nothing to collapse.

**2. `.ai/specs/2026-08-27-scope-cookie-constants-hoist.md` — the spec omits three sections the repository's spec convention requires.**

`.ai/specs/AGENTS.md` states that every non-trivial spec includes a Risks & Impact Review, a Final Compliance Report and a Changelog, and that the changelog is updated with the exact date and a concise summary after coding. This spec has none of the three. This is not a dormant rule: of the eleven specs added to `.ai/specs/` before this one, ten carry a Changelog section and eight carry both Risks and Compliance sections.

Why it matters here specifically: the Risks section is the natural home for the reasoning this spec already does well but scatters. The PR makes a semantic choice about blank cookie values and asserts that "no wrong result is reachable today" because nothing currently writes a blank organization cookie. That is a real residual risk with a concrete failure mode — anything that starts writing a blank organization cookie, or any caller that reaches for a selection parser where `applySuperAdminScope` needs the lossless pair, silently widens or narrows scope. Stated as prose in the middle of the document it reads as background; stated in a risk table with severity, affected area, mitigation and residual risk, it becomes something the next person to touch scope resolution will actually find.

Fix: add the three sections. Risks & Impact Review should carry at least the blank-vs-absent divergence described above and the "shared must not depend on core" rule that `lib/crud/factory.ts` still bends. Final Compliance Report can record the gate results and the `BACKWARD_COMPATIBILITY.md` §4 assessment the spec already makes. Changelog needs one dated line.

### 💅 Nit

**3. `packages/shared/AGENTS.md:34-60` — the new `scope/` directory is not listed in the Library Directory table.**

That table is the routing surface a contributor or an agent consults to find an existing shared helper, and `BACKWARD_COMPATIBILITY.md` §4 names it as part of the public API. I checked before raising this: the table is curated rather than exhaustive — roughly twenty of the ~fifty directories under `packages/shared/src/lib/` appear in it — so omitting `scope/` breaks no rule and this is genuinely the author's call. I raise it anyway because this particular module's entire reason for existing is discoverability. The failure this PR fixes was four people independently not knowing where the cookie names lived; a one-row entry is the cheapest way to stop the fifth.

Suggested row, in the table's alphabetical position between `search/` and `string.ts`:

```
| `scope/` | When reading the organization/tenant selection cookies — the canonical cookie names, the `__all__` sentinel, and the parsers. A blank selection cookie reads as "no selection"; use `readScopeCookieRaw`/`decodeScopeCookieValue` only when absent and blank must genuinely differ | `@open-mercato/shared/lib/scope/cookies` |
```

**4. `packages/core/src/modules/directory/constants.ts:1-16` and `packages/core/src/modules/directory/utils/scopeCookies.ts:1-13` — the `@deprecated` annotation `BACKWARD_COMPATIBILITY.md` §4 asks for is deliberately omitted; this deserves an explicit maintainer decision rather than an implicit one.**

§4 reads: "If a module is moved internally, the old import path MUST be re-exported for backward compatibility with a `@deprecated` annotation." An implementation did move internally and the old paths are now re-export bridges, so the letter of the rule applies. The spec argues the annotation would be false here — these paths are not being retired, the bridge is permanent, and marking them deprecated would put a strikethrough on roughly ten live internal call sites and tell third-party developers to migrate off a path that is staying.

I agree with the author, and the codebase agrees too. `packages/core/src/modules/inbox_ops/lib/htmlToPlainText.ts` is a one-line re-export of `@open-mercato/shared/lib/html/htmlToPlainText` with no deprecation; `packages/core/src/modules/planner/api/helpers.ts` re-exports from `@open-mercato/shared/lib/boolean` and `@open-mercato/shared/lib/query/sanitizeSearchTerm` the same way; `packages/shared/AGENTS.md` documents `registry.ts`'s `resolvePageRouteMetadata` re-export as an implementation-detail split that consumers should keep importing from the old path. The established practice for a moved-implementation bridge whose path stays supported is exactly what this PR does.

So this is not a defect, and I am not asking for the annotation. But this PR was split out of #5690 specifically *because* it touches §4, which makes it the wrong place to leave a §4 deviation resting on a self-declaration in the spec. Worth a maintainer saying so on the record — and the durable fix is to amend §4 to distinguish the two cases it currently conflates: a path being **retired** (re-export plus `@deprecated`, per the deprecation protocol) versus an implementation **relocated** behind a path that remains the supported entry point (re-export, no `@deprecated`, note at the bridge). The bridge comments this PR adds are already a good model for the latter.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path. Both bridged paths keep every symbol they exported, as the same bindings; `constants` additionally gains `SELECTED_ORGANIZATION_COOKIE`, `SELECTED_TENANT_COOKIE` and the `ScopeCookieName` type, which is additive. `modules/directory/utils/organizationScope` still re-exports both parsers, and I confirmed the new `scopeCookies.test.ts` pins that re-export.
- [x] No function signature changed in a breaking way. The three public parsers keep `(header: string | null | undefined) => string | null`. The removed `readCookieFromHeader` / `decodeCookieValue` in `lib/auth/server.ts` were module-private, so narrowing the replacement's `name` parameter to the `ScopeCookieName` union is not a public change.
- [x] No HTTP route URL removed or renamed; no method changed. No route files touched.
- [x] No event or message name renamed or removed. The cookie names themselves are the wire format, and all three literal values are unchanged and now pinned by explicit assertions in both new test files.
- [x] No database table or column renamed or removed. No migrations in this diff, and none needed.
- [x] No config key renamed and no default changed silently. `packages/shared`'s `exports` map needs no entry: `./lib/scope/cookies` is three segments, so the existing `./*/*/*` wildcard resolves it to `./src/lib/scope/cookies.ts` for types and `./dist/lib/scope/cookies.js` at runtime. I verified the built artifact exists after `yarn build:packages`.
- [x] Where a contract had to change: old surface kept working. Nothing was removed, so no deprecation window is consumed; see Nit 4 for the `@deprecated` question.

One resolution detail I checked because it would have been a silent runtime break in a published package rather than in this monorepo: `packages/core` declares `@open-mercato/shared` as a peer dependency, so `core`'s compiled bridge resolving `@open-mercato/shared/lib/scope/cookies` is satisfied in a scaffolded standalone app too. And because `lib/scope/cookies.ts` has zero imports, the client components that reach these bridges — `packages/search`'s `GlobalSearchDialog`, `TopbarSearchInline` and `HybridSearchTable`, plus `auth/lib/backendChrome.tsx` — pull in nothing new.

## 🧪 Test Coverage

Coverage is good, and specific rather than decorative. `packages/shared/src/lib/scope/__tests__/cookies.test.ts` pins the three literal wire values rather than only the symbols; asserts that `readScopeCookieRaw` returns `undefined` for absent and `''` for present-and-blank, with an inline comment naming `applySuperAdminScope` as the caller that needs the difference; checks that a cookie whose name merely ends with the requested one does not match; covers percent-decoding, the no-trimming decision, and the malformed-encoding fallback; and closes with a case asserting the raw reader and the selection parser agree on a real value and diverge only on a blank one, which is the exact invariant a future tidy-up would break. `packages/core/src/modules/directory/utils/__tests__/scopeCookies.test.ts` pins the bridge from the consumer's side using `toBe` identity assertions, so a bridge silently replaced by a second copy fails rather than passing on equal values. The two cases added to `packages/shared/src/lib/auth/__tests__/server.test.ts` close the gap the author identified and honestly reported: before this PR, mutating the reader to return `undefined` for a blank cookie left that whole suite green.

I verified the tests actually execute rather than being silently skipped — all three suites appear as PASS in the run, and the jest configs in `packages/core`, `packages/shared` and `packages/search` all map `@open-mercato/shared/*` to source, so the new module resolves without a build.

One thing worth noting rather than fixing: three existing test files (`auth/lib/__tests__/backendChrome.*.test.ts`) `jest.mock` the `directory/constants` path with a partial factory that supplies only `isAllOrganizationsSelection`. Because `organizationScope.ts` now imports the two cookie-name constants from that same mocked path, those constants would be `undefined` inside those suites. It is harmless today — I checked, and all three also mock `organizationScope` wholesale, so the getters never run — and core's suite is green. It is a latent sharp edge rather than a defect, and the spec's note that "a pure re-export module mocks identically" is true as far as it goes; if you touch those files for another reason, widening the mock factories would remove the trap.

Gaps: none that I would block on. If you take the first option under Minor 1 and change `[...slug]/page.tsx`, that path has no unit coverage today and does not need new tests for a constant substitution — the existing `organizationScope` suite already pins the behaviour it mirrors.

---

*Reviewed at head `22a97e44`, in an isolated worktree, against `develop` at `41ff100d9`. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).*

pkarw:

- ❌ `changes-requested` — the review found two Medium findings, and this repository requests changes on Medium too: the hoist leaves an un-collapsed `om_selected_org` literal in `apps/mercato/src/app/(backend)/backend/[...slug]/page.tsx:85` (plus its template mirror) that the spec's "Deliberately out of scope" list does not name, and that section's claim that the `OrganizationSwitcher` copies carry no constant duplication is inaccurate; and the spec omits the Risks & Impact Review, Final Compliance Report and Changelog sections `.ai/specs/AGENTS.md` requires. There are no blockers and no majors — the code itself is behaviour-preserving and passes the full gate.
- ♻️ `refactor` — no behaviour, import path, export name or export value changes for any consumer; four inlined copies of the scope-cookie wire format collapse into one shared module with two permanent re-export bridges.
- 🔽 `priority-low` — this is cleanup with no user-visible effect and no bug being fixed, so nothing about it is time-sensitive; it only becomes urgent in the sense that #5690 would otherwise add a fifth copy.
- ⚠️ `risk-high` — urgency and blast radius are orthogonal here: the diff touches `packages/shared/src/lib/auth/server.ts` (session and organization scoping) and two import paths that `BACKWARD_COMPATIBILITY.md` §4 protects, so a mistake in the collapsed parser would change organization scoping on every authenticated request. The change is in fact behaviour-preserving and well tested — this rating describes what the surface could break, not what this PR does break.
- ⏭️ `skip-qa` — the automated-verification exemption in `.ai/docs/pr-workflow.md` applies in full: no `.tsx` outside tests, nothing under `packages/ui/src/` or `**/components/**`, the database structure and API surface are unchanged, no `BACKWARD_COMPATIBILITY.md` contract is broken (both old paths keep every symbol as the same binding), and the PR ships automated tests for the changed behaviour — three suites, all green, including two cases closing a gap that was previously untested.

pkarw:

pkarw:

No blockers and no majors — the refactor is behaviour-preserving and the full validation gate is green. Two Medium findings drive the verdict under this repo's stricter rule (un-collapsed cookie-name literal not named in the spec's out-of-scope list plus an inaccurate claim in that section; spec missing the Risks / Compliance / Changelog sections), plus two Low notes on documentation.

Validation source: local fallback — both workflow runs for head `22a97e44` sit at `action_required` (fork PRs need per-push CI approval), so there was no usable GitHub check data. The configured gate was run in an isolated worktree at the PR head: `build:packages`, `generate` (no drift), `build:packages`, `i18n:check-sync`, `i18n:check-usage`, `typecheck` and `build:app` all pass; `packages/shared` 2059 passed / 0 failed and `packages/core` 11697 passed / 0 failed. The only `yarn test` failures are four host-environment inotify-limit assertions in `create-mercato-app` that are unrelated to this diff. `yarn lint` is clean (0 errors). **CI itself still needs a maintainer to approve the fork run before merge** — the local gate is this run's evidence only and does not stand in for branch protection.

Geeknauts-company:
Re **Nit 4** (`@deprecated` on the two `core` bridges) — putting the decision on the record rather than leaving it resting on a self-declaration in the spec.

**Decision: no `@deprecated` annotation on either bridge.** `modules/directory/constants` and `modules/directory/utils/scopeCookies` are not being retired. They remain the supported way for a domain module to reach these values, the bridge is permanent, and annotating them would put a strikethrough on ~10 live internal call sites while telling third-party developers to migrate off a path that is staying. This matches the established practice you identified for a relocated implementation behind a path that stays supported: `inbox_ops/lib/htmlToPlainText.ts`, `planner/api/helpers.ts`, and `registry.ts`'s `resolvePageRouteMetadata`.

The durable fix you propose — amending `BACKWARD_COMPATIBILITY.md` §4 to distinguish a path being **retired** (re-export + `@deprecated`, full deprecation protocol) from an implementation **relocated** behind a path that remains the supported entry point (re-export, no `@deprecated`, note at the bridge) — is the right call, but it is a contract-document change with its own blast radius and does not belong in this PR. Happy to raise it as a separate issue if you want it tracked.

The deviation is now recorded explicitly in the spec's Final Compliance Report rather than only in prose.

Geeknauts-company:
Pushed `54043ccc5` — addresses both Medium findings and Nit 3. @pkarw, flagging you directly rather than through the reviewer field: this account only has `read`, so `RequestReviewsByLogin` is refused and I cannot re-request review myself.

**Minor 1 — the copy left behind.** Took the first option: `apps/mercato/src/app/(backend)/backend/[...slug]/page.tsx` now imports `SELECTED_ORGANIZATION_COOKIE` instead of inlining `'om_selected_org'`, and the create-app template mirror follows it, as the root `AGENTS.md` template-sync rule requires. You were right that this was the one remaining site with no "richer shape" defence.

Mirrored by hand rather than with `yarn template:sync:fix`: `yarn template:sync` reports a pre-existing, deliberate drift in `modules.ts` (the template keeps `channel_discord` disabled and omits `design_system`), and `--fix` would have clobbered that. The two `page.tsx` files are byte-identical and do not appear in the drift list.

**Minor 1 — the inaccurate claim.** *Deliberately out of scope* now names both `OrganizationSwitcher` files rather than only the template one, and no longer says they carry no constant duplication. Corrected to: they import the sentinel from `core`, but both inline the two cookie names across seven lines — `om_selected_org` at 77, 78, 181 and `om_selected_tenant` at 107, 108, 155, 165 — and the collapse is deferred because of the richer `{ value, hasCookie, raw }` shape, not because there is nothing to collapse. Also flagged as the highest-value remaining follow-up.

**Minor 2 — missing spec sections.** Added Risks & Impact Review, Final Compliance Report and Changelog. The risk register carries the four you would expect: the blank-vs-absent divergence `applySuperAdminScope` depends on (High-if-it-happened, Low likelihood, with the guarding test named as the mitigation), a future writer emitting a blank organization cookie (Medium, residual and pre-existing), `shared`'s remaining behavioural dependency on `core` via `lib/crud/factory.ts`, and the partial `jest.mock` factories on the bridged path you spotted — recorded as a latent sharp edge with the Boy-Scout fix noted.

**Nit 3.** Added your suggested `scope/` row to the Library Directory table, verbatim, in its alphabetical position. `yarn agents:check-budget` still exits 0.

**Nit 4.** Answered in a separate comment above.

**Verification** (Node 24, local — CI still cannot run, see below):

| Check | Result |
|---|---|
| `tsc --noEmit` in `packages/shared` | clean, exit 0 |
| `apps/mercato` typecheck, before vs after | 64 errors both, identical sets apart from one line-number shift (19→20) from the added import; zero errors mentioning `scope/cookies`. The 64 are missing generated files in an ungenerated worktree and are present at `22a97e441` too |
| `packages/shared` full suite | 2 failed / 2012 passed, 11 unrunnable suites — the documented baseline for an unbuilt clone (`principalServiceExport` wants `dist/`, the rest want `@open-mercato/cache`). Only `AGENTS.md` changed in this package; no `.ts` touched |
| `src/lib/scope` suite | 17/17 pass |
| eslint, both changed files | exit 0 |
| `yarn agents:check-budget` | exit 0 |

One thing I cannot fix from this side: both workflow runs still sit at `action_required` because the PR is from a fork, so there is no GitHub check signal for the new head either. Could you approve the run when you next look at this?

Geeknauts-company:
Gentle nudge — no urgency, just making sure this has not slipped through.

The round-1 findings are addressed in `54043ccc5` (pushed 2026-08-28): the last plain `om_selected_org` literal is collapsed onto `SELECTED_ORGANIZATION_COOKIE` with its template mirror, the *Deliberately out of scope* section is corrected, the Risks / Compliance / Changelog sections are added, and the `scope/` row is in the Library Directory table. The `@deprecated` question is answered in a separate comment above.

@pkarw two things still need a maintainer whenever you next have a moment: a re-review to clear `changes-requested`, and approval of the fork CI run — both workflows on this head still sit at `action_required`, so there is no GitHub check signal on it yet.

Geeknauts-company:
Refreshed against today's `develop` rather than nudging again — two weeks is long enough that "it was green then" stops being evidence.

Merged `origin/develop` in (`35b53523`, no conflicts; the only files that even auto-merged were `packages/shared/AGENTS.md` and `packages/shared/src/lib/crud/factory.ts`) and re-ran the gate on the merged tree:

| Gate | Result |
|---|---|
| `packages/shared` jest | 194/194 suites, 2214 tests (5 skipped) |
| `packages/core` — `directory` + `auth`, the consumers this refactor touches | 123/123 suites, 969 tests |
| `packages/shared` typecheck | clean |

So the round-1 work still holds on a two-week-newer base, and this is still a behaviour-preserving refactor.

**Recap of what was addressed in `54043ccc5` (2026-08-28), since it is now several screens up:**

- **Medium 1** — the un-collapsed `om_selected_org` literal at `apps/mercato/src/app/(backend)/backend/[...slug]/page.tsx:85` and its `packages/create-app` template mirror now import `SELECTED_ORGANIZATION_COOKIE`.
- **Medium 2** — the *Deliberately out of scope* section no longer claims what it did; it names what is actually left and why.
- Risks / Compliance / Changelog sections added to the spec, and the `scope/` row added to the table.
- **Nit 4** (`@deprecated` on the two `core` bridges) — declined, with the reasoning on the record in a separate comment: `modules/directory/constants` and `modules/directory/utils/scopeCookies` are not being retired, they remain the supported way for a domain module to reach these values, and marking them deprecated would push module authors toward a `@open-mercato/shared` import that is the wrong layer for them.

Nothing is outstanding from my side. @pkarw — this account only has `read`, so `RequestReviewsByLogin` is refused and I cannot re-request review myself; flagging you directly instead.

One thing worth knowing if you re-run the gate locally in a fresh worktree: `packages/shared`'s two export-map suites (`principalServiceExport`, `duplicateEntityClassNamesExport`) assert that the mapped `dist/` files exist, so they fail until `yarn build:packages` has run. Not this branch — I mention it only because the failure reads like a broken export map.

PR #5717, opened 2026-08-27
fix(deploy): pass OM_INIT_ADMIN_EMAIL/EMPLOYEE_EMAIL through to the app container

pat-lewczuk:
## Problem

Client report: after `yarn mercato init` (a.k.a. reinstall) on a Dokploy-hosted deployment, the console banner correctly printed `superadmin@hopgrow.pl`, `admin@hopgrow.pl`, `employee@hopgrow.pl` — but the users actually created in the database were `superadmin@hopgrow.pl` (correct), `admin@acme.com`, `employee@acme.com` (wrong domain).

## Root cause

The symptom has **two contributing gaps**. This PR closes the deployment-config one; the application-side one is tracked separately in #5806 (see *Scope* below).

Docker Compose's `environment:` block is a fixed allowlist, not a passthrough — a host/Dokploy env var only reaches the container's `process.env` if it's explicitly declared as `KEY: ${KEY:-default}` in that block.

The `app` service in `docker-compose.fullapp.yml` and `docker-compose.fullapp.dev.yml` only declared:
```yaml
OM_INIT_SUPERADMIN_EMAIL: ${OM_INIT_SUPERADMIN_EMAIL:-}
OM_INIT_SUPERADMIN_PASSWORD: ${OM_INIT_SUPERADMIN_PASSWORD:-password}
```
with no matching entries for `OM_INIT_ADMIN_EMAIL`, `OM_INIT_EMPLOYEE_EMAIL`, or their passwords. So even when an operator correctly sets these in Dokploy's environment config, they never reach the container, and `setupInitialTenant()` (`packages/core/src/modules/auth/lib/setup-app.ts`) falls back to its hardcoded `acme.com` default for admin/employee — while superadmin, the only declared var, correctly picks up the configured domain.

The same superadmin-only block was present in the two `packages/create-app/template/` compose stacks, so every standalone app scaffolded by `create-mercato-app` and deployed the same way hit the identical defect.

## Fix

Add the four missing declarations to the `app` service's `environment:` block in **all four** full-app compose stacks (repo root prod + dev, create-app template prod + dev), alongside the existing superadmin entries:
```yaml
OM_INIT_ADMIN_EMAIL: ${OM_INIT_ADMIN_EMAIL:-}
OM_INIT_ADMIN_PASSWORD: ${OM_INIT_ADMIN_PASSWORD:-}
OM_INIT_EMPLOYEE_EMAIL: ${OM_INIT_EMPLOYEE_EMAIL:-}
OM_INIT_EMPLOYEE_PASSWORD: ${OM_INIT_EMPLOYEE_PASSWORD:-}
```

Add `scripts/__tests__/fullapp-compose-init-account-env.test.mjs`, a guard in the established idiom of the five sibling compose guards, asserting that every service declaring `OM_INIT_SUPERADMIN_PASSWORD` also declares all four variables, across all four stacks. This is needed because `scripts/template-sync.ts` scopes its mirror to `SYNC_FOLDERS` plus three named root files — compose files are outside it entirely, so CI had no way to flag the root/template divergence mechanically.

The guard keys the "seeding service" off `OM_INIT_SUPERADMIN_PASSWORD` rather than the email, so the `mcp` service — which declares `OM_INIT_SUPERADMIN_EMAIL` only to resolve the superadmin owning its API key, and seeds no users — is correctly excluded.

## Scope: what this PR does *not* fix

`mercato init` has a **second, sufficient cause** for the same reported symptom, which reproduces on a bare host with no Docker involved: it propagates the derived admin/employee *passwords* into `process.env` (`packages/cli/src/mercato.ts:1147-1152`) but not the *emails*, so the banner prints `admin@<superadmin-domain>` (from `init-secrets.ts`) while `setup-app.ts:273-276` seeds `admin@acme.com` (from a hardcoded constant).

Consequence worth stating plainly: **after this PR, an operator who sets only `OM_INIT_SUPERADMIN_EMAIL` — the documented minimum — still gets `admin@acme.com` / `employee@acme.com`.** This PR helps the operator who additionally sets the two new email variables explicitly. Filed as **#5806** and left out of this PR on the reviewer's preference to keep the compose change mergeable on its own merit rather than growing it into the CLI.

## Testing

Config-only change plus one new guard test (no app-rendering `.tsx`, no DB/API surface change).

- `node --test scripts/__tests__/fullapp-compose-*.test.mjs scripts/__tests__/prod-compose-node-options.test.mjs` — 49/49 pass, including the four new cases.
- Verified the new guard actually catches the regression it exists for: deleting the `OM_INIT_ADMIN_EMAIL` line from `packages/create-app/template/docker-compose.fullapp.yml` fails it with a message naming the file, the service, and the consequence; restoring the line makes it pass again.
- Verified all four compose files still parse as YAML and that each `app` service now exposes the same six `OM_INIT_*` keys.
- Traced both env readers to confirm the empty-string defaults are inert: `readEnvValue` in `setup-app.ts:488-492` and in `init-secrets.ts:6-11` both trim and treat `""` as unset, so no existing deployment changes behavior and the production `DerivedUserPasswordRequiredError` guard (`setup-app.ts:281-286`) still fires rather than being satisfied by an empty string.

---
<a href="https://app.blacksmith.sh/open-mercato/codesmith/open-mercato/pr/5717"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>


pat-lewczuk:
Labels: `bug` (client-reported provisioning defect), `priority-medium` (affects initial admin/employee accounts on self-hosted/Dokploy deploys, not a production outage or data exposure), `risk-low` (additive-only env var declarations in two docker-compose files, no behavior change for deployments not using these vars), `skip-qa` (no UI-rendering file touched — no `.tsx`, nothing under `packages/ui/src/` or `**/components/**` — and no DB/API surface change; there is no manually-exercisable surface for a QA reviewer to click through, and the change is not meaningfully unit-testable since it's a Docker Compose env-var allowlist, analogous to a CI/deploy-config-only change).

pkarw:
🤖 `om-auto-review-pr` started on this PR (claiming with the `in-progress` label + assignee). Will release the lock when the review is submitted.

pkarw (changes requested):
# 🔍 Code Review

**PR #5717** — `fix(deploy): pass OM_INIT_ADMIN_EMAIL/EMPLOYEE_EMAIL through to the app container` by @pat-lewczuk
**Head:** `f499d4df69088158abb47a3e6f9f2431e37a2a47` · **Base:** `develop` (branch is at develop's tip, `MERGEABLE`) · 2 files, +8/−0

**Verdict: 🔴 Changes requested** — 0 blockers, **2 majors**, 1 minor, 2 nits.

**Validation source: GitHub checks.** All workflow runs keyed on the exact head SHA are green (`CI for Develop&Main` 33114472686, `Mutation tests` 33114472539, `PR #5717` 33114469366), and I opened the `prepare` and `test` job step lists individually rather than trusting the job-level roll-up: builds, generate, i18n sync + usage, agent instruction budget, typecheck, Test, Test scripts and the three unfiltered guards all ran and passed. `build:app`, `ephemeral-integration`, `docker-build` and `merge-coverage` are `skipping` — expected, the change-detection step found no app-affecting source. Per this repo's checks-first extension, `validation.commands` was **not** re-run locally; the worktree never needed `node_modules`.

The diff itself is correct, additive, and lands in the right service — the `app` service is the one whose `command:` runs `INIT_COMMAND='yarn mercato init' sh /app/docker/scripts/init-or-migrate.sh` (`docker-compose.fullapp.yml:144-151`), so that is where these variables have to arrive. The compose passthrough gap it names is real. What I'm requesting changes on is that **the diagnosis stops one hop short**, and the fix is therefore incomplete in two directions.

---

## 💥 Major

### 1. The reported symptom is an application-code bug, not only a deployment-config gap — and this PR alone does not close it

`packages/cli/src/mercato.ts:1145-1152` (the init flow)

The PR body states: *"This is a deployment-config gap, not an application code bug — the application already reads these env vars correctly once they're actually present."* That is not what the code does, and the exact reported divergence — **banner says `admin@hopgrow.pl`, database gets `admin@acme.com`** — reproduces with no Docker involved at all.

Trace of the two readers, which disagree:

- **The banner** comes from `resolveInitDerivedSecrets` (`packages/cli/src/lib/init-secrets.ts:40-42`), which derives from the *superadmin's domain*:
  `adminEmail = envRead('OM_INIT_ADMIN_EMAIL') ?? admin@<domain of --email>`.
  `mercato.ts:1145-1146` captures that into `adminEmailDerived` / `employeeEmailDerived`, and `mercato.ts:1330-1331` prints exactly those two values in the init summary. With `--email superadmin@hopgrow.pl` and no override set, the banner prints `admin@hopgrow.pl` — correctly, from the operator's point of view.
- **The seeding** comes from `setupInitialTenant` (`packages/core/src/modules/auth/lib/setup-app.ts:273-276`), which derives from a *hardcoded constant*:
  `adminEmail = readEnvValue('OM_INIT_ADMIN_EMAIL') ?? admin@${DEFAULT_DERIVED_EMAIL_DOMAIN}`, where `DEFAULT_DERIVED_EMAIL_DOMAIN` is `acme.com` (`setup-app.ts:22`). It never sees `--email`.

The bridge between them is `process.env`, and it carries only half the payload. `mercato.ts:1147-1152` writes the derived **passwords** back into `process.env.OM_INIT_ADMIN_PASSWORD` / `OM_INIT_EMPLOYEE_PASSWORD` — and writes **nothing** for the emails. That propagation works because `runModuleCommand` invokes the subcommand in-process (`mercato.ts:734`, `await resolved.command.run(args)`), so the same trick would work for the emails; it was simply never done.

Consequences:

- `yarn mercato init --email superadmin@hopgrow.pl` on a bare host — no compose, no Dokploy — prints `hopgrow.pl` accounts and seeds `acme.com` ones. The compose gap cannot be the whole story, because it is not present in that reproduction.
- After this PR, an operator who sets only `OM_INIT_SUPERADMIN_EMAIL` — the documented minimum — **still gets `admin@acme.com` / `employee@acme.com`**. The fix works only for an operator who additionally sets both new email variables explicitly.
- `apps/docs/docs/cli/init.mdx:40-41` documents the default for these two variables as `admin@<superadmin-domain>` / `employee@<superadmin-domain>`. The seeding path has never honored that. The docs and the banner agree with each other; only the code that writes to the database disagrees.

The fix is the two lines symmetrical with the passwords already there, in `mercato.ts` right after line 1146:

```ts
if (adminEmailDerived) process.env.OM_INIT_ADMIN_EMAIL = adminEmailDerived
if (employeeEmailDerived) process.env.OM_INIT_EMPLOYEE_EMAIL = employeeEmailDerived
```

Both readers already treat an empty/whitespace value as unset, so this is safe when nothing is configured, and it makes `init.mdx`'s documented default true for the first time.

I'm fine with this landing as a **separate PR** — the compose change here stands on its own merit and I'd rather it merged than grew. What I am requesting is that the PR body stop asserting the application side is correct, and that the app-side defect be filed or fixed, because as written this PR will be read as "the client's issue is resolved" when for the most likely configuration it is not. Happy to file the follow-up issue myself if you'd prefer.

### 2. The scaffolded-app stack keeps the identical defect

`packages/create-app/template/docker-compose.fullapp.yml:102-103`, `packages/create-app/template/docker-compose.fullapp.dev.yml:68-69`

Both template compose files carry the same superadmin-only block this PR is fixing:

```yaml
OM_INIT_SUPERADMIN_EMAIL: ${OM_INIT_SUPERADMIN_EMAIL:-}
OM_INIT_SUPERADMIN_PASSWORD: ${OM_INIT_SUPERADMIN_PASSWORD:-password}
```

and the template's `app` service runs the same `init-or-migrate.sh` bootstrap (`packages/create-app/template/docker-compose.fullapp.yml:24-28`). So every standalone app scaffolded by `create-mercato-app` and deployed to Dokploy hits precisely the client-reported bug, unfixed.

This isn't my own pairing of the files — the repo already declares it. `scripts/__tests__/fullapp-compose-auth-defaults.test.mjs:10-14` defines:

```js
// The production-documented stacks. Their `.dev.yml` sibling is deliberately excluded: ...
const PRODUCTION_COMPOSE_FILES = [
  'docker-compose.fullapp.yml',
  'packages/create-app/template/docker-compose.fullapp.yml',
]
```

and asserts `JWT_SECRET`, `NODE_ENV` and the JWT legacy controls across **both**. This PR updated one of the two.

Worth knowing while fixing it: **CI cannot catch this omission.** `scripts/template-sync.ts:32-33` scopes the mirror to `SYNC_FOLDERS = ['app', 'components', 'i18n', 'lib', 'modules']` plus three named root files — compose files are outside it entirely, so the green *Check create-app template parity* step says nothing here. The four lines have to be added by hand, in both template files, to match what you did at the root.

---

## 🧩 Minor

### 3. No guard test, though the repo has an established idiom for exactly this file

`scripts/__tests__/` (new file, or extend `fullapp-compose-auth-defaults.test.mjs`)

There are already five compose-shape guards under `scripts/__tests__/` — `fullapp-compose-runtime-env`, `fullapp-compose-auth-defaults`, `fullapp-compose-app-allowed-origins`, `fullapp-compose-documents-dist-wiring`, `prod-compose-node-options` — all run in CI by `yarn test:scripts` (`package.json:105`, green as step 23 of the `test` job on this head). A regression here is exactly the shape those guards exist to catch: a declaration silently absent from one of two files that must agree.

The assertion is a handful of lines and would have caught Major 2 mechanically — for each file in `PRODUCTION_COMPOSE_FILES`, every service that declares `OM_INIT_SUPERADMIN_PASSWORD` must also declare `OM_INIT_ADMIN_EMAIL`, `OM_INIT_ADMIN_PASSWORD`, `OM_INIT_EMPLOYEE_EMAIL` and `OM_INIT_EMPLOYEE_PASSWORD`.

This also matters for the label: the automated-verification exemption in `AGENTS.md` grants `skip-qa` only when *"automated tests for the changed behavior"* ship in the same PR. I've kept `skip-qa` (see the label comment for why), but the cleanest way to make that unambiguous is to land the guard.

---

## 📝 Nits

4. **PR body — "Testing" overclaims what was verified.** *"Verified by ... confirming the application-side env var names match exactly what `packages/core/src/modules/auth/lib/setup-app.ts` reads"* — the names do match, but `setup-app.ts` is not the reader that produced the banner in the bug report. Tracing the *other* reader (`init-secrets.ts` via `mercato.ts`) is what surfaces Major 1. Worth restating the root cause as "one of two contributing gaps".

5. **`apps/docs/docs/cli/init.mdx:40-41` is now the only place stating the intended behavior**, and it states one the code doesn't implement. Whether Major 1 is fixed here or in a follow-up, that table is the natural place to hang the correction.

---

## ✅ Checked and explicitly *not* findings

Recording these so nobody re-derives them on the next pass:

1. **The empty-string defaults are safe.** `${OM_INIT_ADMIN_EMAIL:-}` sets the variable to `""` in every deployment that doesn't configure it — but both readers trim and treat empty as unset (`setup-app.ts:488-492`, `init-secrets.ts:6-11`). So nothing changes for existing deployments, and specifically the production guard at `setup-app.ts:281-286` (`DerivedUserPasswordRequiredError` when `OM_INIT_ADMIN_PASSWORD` / `OM_INIT_EMPLOYEE_PASSWORD` are missing under `NODE_ENV=production`) still fires exactly as before rather than being satisfied by an empty string. This was the main hazard I went looking for; it isn't there.
2. **The `mcp` service was correctly left alone.** `docker-compose.fullapp.yml:93` and `docker-compose.fullapp.dev.yml:128` declare `OM_INIT_SUPERADMIN_EMAIL` for a different purpose — resolving the superadmin who owns the MCP API key (`packages/ai-assistant/src/modules/ai_assistant/lib/mcp-ensure-api-key.ts:77`). That service seeds no users, so it needs none of the four new variables.
3. **`docker-compose.yml` and the two traefik overlays** (`docker-compose.fullapp.traefik.yml`, `.traefik.dev.yml`) declare no `OM_INIT_*` block at all — the overlays only add labels — so leaving them untouched is right.
4. **Why the existing unit tests are green and structurally cannot catch Major 1:** `packages/core/src/modules/auth/__tests__/cli-setup-demo-users.test.ts:67-71` runs every case with `--email root@acme.com`. The superadmin's domain and the hardcoded `DEFAULT_DERIVED_EMAIL_DOMAIN` are then the same string, so the assertions on `admin@acme.com` (`:147`, `:209`) pass under both the intended and the actual behavior. The only domain any test pins is the one that cannot fail.
5. **The in-process assumption behind the suggested fix holds** — `runModuleCommand` calls `resolved.command.run(args)` directly (`mercato.ts:734`); it does not spawn. That is why the existing password propagation works, and it is the reason two lines are sufficient for the emails.
6. **Nothing in `BACKWARD_COMPATIBILITY.md` is touched** — no contract surface, no DB structure, no API surface, no generated file. The diff is purely additive compose declarations.

---

## 🏁 Summary

The compose passthrough gap you found is real and the four lines are correct. Two things keep this from being complete: the same four lines are missing from the create-app template stack that the repo's own guard file names as this file's twin, and the client-reported symptom has a second, sufficient cause in `mercato init` itself — it derives the demo account emails from the superadmin's domain for the banner and from a hardcoded `acme.com` for the database. Fix the template files here; fix or file the CLI propagation (I'm happy to file it), and drop the "not an application code bug" claim from the description. A short `scripts/__tests__` guard over both compose files would close the loop and settle the `skip-qa` question.

pkarw:
🏷️ **Label rationale** — `om-auto-review-pr`

| Label | Decision |
|---|---|
| `review` → `changes-requested` | Verdict is changes-requested: 2 majors (the create-app template stack keeps the identical superadmin-only block; the reported symptom also has a sufficient application-side cause in `mercato init`) plus 1 minor. This repo requests changes on minor findings too. |
| `bug` | Kept — client-reported provisioning defect. |
| `priority-medium` | Kept — wrong demo-account domain on self-hosted/Dokploy deploys; recoverable by the operator, no data exposure or outage. |
| `risk-low` | Kept — purely additive `environment:` declarations. Verified the empty-string defaults are inert: both readers (`setup-app.ts:488-492`, `init-secrets.ts:6-11`) trim and treat `""` as unset, so no existing deployment changes behavior and the production `DerivedUserPasswordRequiredError` guard still fires. |
| `skip-qa` | Kept, with a caveat stated openly. The exemption's first three clauses hold — no `.tsx`, nothing under `packages/ui/`, no DB structure change, no API surface change, no `BACKWARD_COMPATIBILITY.md` contract touched. Its fourth clause ("automated tests for the changed behavior in the same PR") does **not** hold: the PR ships no test. I kept `skip-qa` rather than routing a compose-env-passthrough change to a click-through QA reviewer, but the honest resolution is the `scripts/__tests__` guard requested as minor 3 — with it, `skip-qa` is unambiguous. If the guard is declined, this should flip to `needs-qa`. |
| `in-progress` | Removed with this review — lock released, PR is free to act on. |

No `ci-monitoring`: CI is already fully settled and green on `f499d4df`.

pkarw:
🤖 `om-auto-review-pr` completed: **CHANGES REQUESTED**. Lock released, PR handed back to @pat-lewczuk.

The four compose lines are correct and land in the right service. Two things to pick up:

1. **`packages/create-app/template/docker-compose.fullapp.yml` + `.dev.yml`** need the same four declarations — `scripts/__tests__/fullapp-compose-auth-defaults.test.mjs:10-14` names those two files as the paired production stacks, and `scripts/template-sync.ts` does not cover compose files, so CI cannot flag the omission.
2. **`mercato init` has a second, sufficient cause for the reported symptom.** It propagates the derived admin/employee *passwords* into `process.env` (`packages/cli/src/mercato.ts:1147-1152`) but not the *emails*, so the banner prints `admin@<superadmin-domain>` while `setup-app.ts:273-276` seeds `admin@acme.com`. Reproduces with no Docker at all. Fine as a follow-up PR — say the word and I'll file the issue — but the description's "not an application code bug" line should go either way.

Full reasoning, plus six things I checked and ruled out, is in the review. Re-run `/om-auto-review-pr 5717` when you've pushed.

pat-lewczuk:
🤖 Claiming this PR — starting `om-auto-fix-pr` run (drive to merge-ready: base merge → review autofix → CI → labels). Started: 2026-08-31T09:20:00Z.

pat-lewczuk:

- 🔍 `review` — replaces `changes-requested`. Every actionable item from @pkarw's review is now either fixed on this branch or tracked in #5806, so the PR is back in the review queue rather than sitting with the author. It is **not** `merge-queue`: GitHub still reports `reviewDecision: CHANGES_REQUESTED`, and that can only be cleared by an independent reviewer — this run is the PR author, and GitHub rejects self-approval.
- 🐛 `bug` — unchanged. This is a client-reported provisioning defect, not new capability.
- ⏭️ `skip-qa` — kept, and the caveat @pkarw raised when keeping it is now resolved. The exemption's first three clauses always held (no `.tsx`, nothing under `packages/ui/`, no DB or API surface change); its fourth clause — automated tests for the changed behavior shipping in the same PR — did not, because the PR had no test. `scripts/__tests__/fullapp-compose-init-account-env.test.mjs` now supplies exactly that, so `skip-qa` is unambiguous rather than a judgment call. There is still no click-through surface for a QA reviewer: the change is a Docker Compose env-var allowlist.
- 🟡 `priority-medium` — unchanged. Wrong demo-account domain on self-hosted and Dokploy deployments is recoverable by the operator; it is neither a production outage nor a data exposure.
- 🟢 `risk-low` — unchanged, and re-verified rather than assumed. The compose change is purely additive `environment:` declarations, and the empty-string defaults are inert because both readers (`setup-app.ts:488-492`, `init-secrets.ts:6-11`) trim and treat `""` as unset — so no existing deployment changes behavior and the production `DerivedUserPasswordRequiredError` guard still fires. The only other change is a new test file, which cannot affect runtime.
- 🔒 `in-progress` — held by this run while it worked the PR; swapped for `ci-monitoring` at the end of this comment batch, since the work is finished and reported and only the CI outcome is still owed.

pat-lewczuk:

@pkarw — every item from your review is accounted for below. Two were fixed on this branch, one is filed as #5806 on your own stated preference, and one nit is deliberately carried into that issue rather than fixed here. Nothing was silently dropped.

## 🔀 Base merge

`origin/develop` had advanced 19 commits since this branch was pushed, so the first thing this run did was merge the current base in (`192a14f96`) — cleanly, with no conflicts. Everything below therefore judges the diff that will actually merge, not the one that was reviewed on 27 Aug against `41ff100d9`.

## 🔍 How each review item was handled

**💥 Major 1 — the reported symptom also has a sufficient application-side cause in `mercato init`. → Filed as #5806, PR description corrected.**

Confirmed independently against the current tree rather than taken on trust: `mercato.ts:1147-1152` writes the derived admin/employee *passwords* back into `process.env` and writes nothing for the emails, so `init-secrets.ts:39-40` (banner, derives from the superadmin's domain) and `setup-app.ts:273-276` (seeding, derives from the hardcoded `DEFAULT_DERIVED_EMAIL_DOMAIN = acme.com`) genuinely disagree, with no Docker involved. You said you were fine with this landing as a separate PR and would rather this one merged than grew, so it is **#5806**, carrying the full trace, your suggested two-line fix, the in-process `runModuleCommand` justification, and the note that `cli-setup-demo-users.test.ts:67-71` pins `root@acme.com` and so structurally cannot catch it.

The PR description has been rewritten accordingly. The "this is a deployment-config gap, not an application code bug" claim is gone; the root cause now opens by naming **two contributing gaps**, and a dedicated *Scope: what this PR does not fix* section states plainly that an operator who sets only `OM_INIT_SUPERADMIN_EMAIL` — the documented minimum — still gets `admin@acme.com` after this PR merges. That was the substance of your objection: that the PR would be read as "the client's issue is resolved" when for the most likely configuration it is not.

**💥 Major 2 — the scaffolded-app stack keeps the identical defect. → Fixed (`6d270bd9a`).**

The four declarations are now in `packages/create-app/template/docker-compose.fullapp.yml` and `packages/create-app/template/docker-compose.fullapp.dev.yml`, matching the root stacks. Verified your reasoning on both points: the template `app` service does run the same bootstrap (`init-or-migrate.sh` in prod, `dev-entrypoint.sh` in dev), and `scripts/template-sync.ts:32-33` really does scope its mirror to `SYNC_FOLDERS` plus three named root files, so compose files are outside it entirely and the green *Check create-app template parity* step could not have caught this.

**🧩 Minor 3 — no guard test, though the repo has an idiom for exactly this. → Fixed (`6d270bd9a`).**

Added `scripts/__tests__/fullapp-compose-init-account-env.test.mjs` in the idiom of the five sibling compose guards. It asserts that every service declaring `OM_INIT_SUPERADMIN_PASSWORD` also declares all four init account variables.

Two deliberate departures from the shape you sketched, both toward stricter coverage:

- It covers **all four** stacks, not just the two in `PRODUCTION_COMPOSE_FILES`. That list excludes the `.dev.yml` siblings because a throwaway JWT secret and development mode are the *point* of a dev stack — but a dev stack seeds the same demo accounts through the same `setupInitialTenant()` path, so the env passthrough matters there identically. The reasoning is written into the file so the divergence from the sibling guard is not mistaken for an oversight.
- It keys the seeding service off `OM_INIT_SUPERADMIN_PASSWORD` rather than the email, which is what correctly excludes the `mcp` service — the point you already recorded under *checked and explicitly not findings*.

The guard was verified to actually catch the regression it exists for, not merely to pass: deleting the `OM_INIT_ADMIN_EMAIL` line from the template prod stack fails it with a message naming the file, the service and the operational consequence; restoring the line makes it pass again.

**📝 Nit 4 — the "Testing" section overclaims. → Fixed.**

Rewritten to describe what was actually run and traced, including the negative test of the new guard, the YAML re-parse of all four stacks, and the trace of both `readEnvValue` implementations confirming the empty-string defaults are inert.

**📝 Nit 5 — `apps/docs/docs/cli/init.mdx:40-41` states behavior the code doesn't implement. → Deliberately deferred into #5806, not dropped.**

You noted the table is the natural place to hang the correction "whether Major 1 is fixed here or in a follow-up". Since Major 1 went to the follow-up, correcting the docs here would mean documenting today's `acme.com` reality and then flipping it straight back when #5806 lands — churn that leaves the docs wrong in one direction or the other either way. It is an explicit acceptance-criterion checkbox on #5806 so the code and the table are settled in the same change.

## 🧪 Validation

`node --test scripts/__tests__/fullapp-compose-*.test.mjs scripts/__tests__/prod-compose-node-options.test.mjs` — **49/49 pass**, including the four new cases and every pre-existing compose guard. This was run locally in the isolated worktree; the full gate is CI's job and is running now.

## 🔁 Follow-ups filed

- **#5806** — `fix(cli): mercato init seeds acme.com admin/employee accounts while its banner prints the superadmin's domain`. Assigned to @pat-lewczuk, labelled `bug` / `priority-medium` / `risk-medium`. Tracks Major 1 and, as an acceptance criterion, nit 5.

## ⚠️ CI is still running on this head

`6d270bd9a` was pushed minutes ago and its checks have not settled — `Analyze (javascript-typescript)` is in progress and `license/cla` is pending. Branch protection holds the actual merge; this comment reports the work, not a green run. A follow-up comment will report the CI outcome.

## 🏁 Merge-readiness

**Not merge-ready yet, and the reason is procedural rather than technical.** GitHub still reports `reviewDecision: CHANGES_REQUESTED` from the 27 Aug review, and that can only be cleared by an independent reviewer: this run operates as the PR author, and GitHub rejects self-approval. So the pipeline label is `review`, not `merge-queue` — the work is done, the PR is waiting on a human re-review.

@pkarw, when you re-review: the branch is at the current `develop` tip, both majors and the minor are addressed, and the two things you offered to file yourself are consolidated into #5806.

pat-lewczuk:

CI has settled on head `6d270bd9a`. **Every workflow check passed; one external check is red and it is my fault, not the code's.**

## ✅ Green

`CI for Develop&Main` ([33377406029](https://github.com/open-mercato/open-mercato/actions/runs/33377406029)) and `Mutation tests` ([33377406093](https://github.com/open-mercato/open-mercato/actions/runs/33377406093)) both succeeded on the exact head SHA. I opened the `test` job's step list rather than trusting the job roll-up: **Test scripts** passed, which is the step that executes the new `scripts/__tests__/fullapp-compose-init-account-env.test.mjs` guard — so the guard genuinely runs in CI and is green there, not only locally. **Check create-app template parity** also passed, confirming the point from the review that compose files fall outside `scripts/template-sync.ts`'s scope and that this omission could never have been caught mechanically before the new guard existed.

`lint`, `ds-lint`, `prepare`, `audit-scope`, `documents-multi-instance`, `scope`, `CodeQL` and all three `Analyze` jobs are green. `build:app`, `ephemeral-integration`, `docker-build`, `merge-coverage`, `audit` and `mutate` are skipped, as expected for a change the detection step finds no app-affecting source in.

## ❌ `license/cla` — pending, caused by my commit authorship

[`license/cla`](https://cla-assistant.io/open-mercato/open-mercato?pullRequest=5717) reports *"Contributor License Agreement is not signed yet."* It was **green** on the original head `f499d4df6` and went pending on both commits this run added.

The cause is commit authorship, not the CLA record. GitHub resolves the two commits I pushed (`192a14f96`, `6d270bd9a`) to the account **`neverwriter`**, because the git author email configured in the worktree I ran in is registered to that account. The original commit `f499d4df6` resolves to **`pat-lewczuk`**, which is signed — hence the check flipped when a second, unsigned identity appeared in the PR's commit list.

Nothing about the change itself is implicated: the diff, the guard and every workflow check are green.

### Resolving it

Two ways, and the choice belongs to a human because one of them rewrites published history:

1. **Sign the CLA for `neverwriter`** via the link on the check. No history rewrite, and the check goes green on the existing commits.
2. **Re-author the two commits** to the same identity as `f499d4df6` and force-push. This is cleaner for the commit log but rewrites pushed history, so this run will not do it unasked.

Until one of those happens, this PR is not mergeable regardless of the review outcome. **No further follow-up will come from this agent on CI** — I am dropping `ci-monitoring` now, since the outcome is settled and what remains needs a human decision.

